PR #3004, opened 2016-05-19
[FIX] Fix wait_for_ajax

mkoura:
I've been running into
``[E] Message: Element <cfme.web_ui.Input _names=('hostname',), _use_id=False> not found on page.``
when adding the 'OpenShift Enterprise' containers provider. The 'hostname' field is displayed only after the provider type is selected using the drop down menu. The test was not waiting for the field to appear even though the 'wait_for_ajax' method was called in the process.
I debugged the the provider page, added new check to the 'wait_for_ajax' method and verified that the problem is no longer present.
Using the trace log it can be verified that the new check makes sense:
``2016-05-19 19:22:07,714 [I] Ajax running: miqQE: 0, jquery: 0, prototype: 0, miqProcessing: True, spinner: False, miq: None, autofocus: 0, document: complete (cfme/fixtures/pytest_selenium.py:335)``

### DIFF
--- a/cfme/fixtures/pytest_selenium.py
+++ b/cfme/fixtures/pytest_selenium.py
@@ -283,7 +283,8 @@ def in_flight():
 
     Returns:
         Dictionary of js-related keys and booleans as its values, depending on status.
-        The keys are: ``jquery, prototype, miq, spinner and document``.
+        The keys are: ``jquery, prototype, miq, spinner, document, autofocus, miqQE, miqProcessing
+                        and angular ``.
         The values are: ``True`` if running, ``False`` otherwise.
     """
     try:
@@ -293,6 +294,22 @@ def in_flight():
     except Exception:
         sleep(0.5)
         return execute_script(js.in_flight)
+
+
+def get_events():
+    """Get ids of events scheduled using setTimeout
+
+    Returns:
+        Dictionary of js-related keys and integers as its values.
+        The keys are: ``minTimeoutId, maxTimeoutId``.
+    """
+    try:
+        return execute_script(js.get_events)
+    except UnexpectedAlertPresentException:
+        raise
+    except Exception:
+        sleep(0.5)
+        return execute_script(js.get_events)
 
 
 def wait_for_ajax():
@@ -305,6 +322,14 @@ def wait_for_ajax():
     """
 
     _thread_local.ajax_log_msg = ''
+    orig_events = get_events()
+
+    def _events_not_finished():
+        events = get_events()
+        not_finished = False
+        not_finished |= orig_events["maxTimeoutId"] >= events['minTimeoutId'] > 0
+
+        return not_finished
 
     def _nothing_in_flight():
         """Checks if there is no ajax in flight and also logs current status
@@ -319,6 +344,8 @@ def wait_for_ajax():
             if "jquery" not in str(e).lower():
                 raise
             return True
+        running["events"] = _events_not_finished()
+
         anything_in_flight = False
         anything_in_flight |= running["jquery"] > 0
         anything_in_flight |= running["prototype"] > 0
@@ -328,6 +355,7 @@ def wait_for_ajax():
         anything_in_flight |= running["miqQE"] > 0
         anything_in_flight |= running["miqProcessing"]
         anything_in_flight |= running["angular"]
+        anything_in_flight |= running["events"]
         log_msg = ', '.join(["{}: {}".format(k, str(v)) for k, v in running.iteritems()])
         # Log the message only if it's different from the last one
         if prev_log_msg != log_msg:
--- a/cfme/fixtures/pytest_selenium.py
+++ b/cfme/fixtures/pytest_selenium.py
@@ -326,10 +326,9 @@ def wait_for_ajax():
 
     def _events_not_finished():
         events = get_events()
-        not_finished = False
-        not_finished |= orig_events["maxTimeoutId"] >= events['minTimeoutId'] > 0
 
-        return not_finished
+        return orig_events["maxTimeoutId"] >= events['minTimeoutId'] \
+            if events['minTimeoutId'] > 0 else False
 
     def _nothing_in_flight():
         """Checks if there is no ajax in flight and also logs current status
--- a/cfme/fixtures/pytest_selenium.py
+++ b/cfme/fixtures/pytest_selenium.py
@@ -326,6 +326,7 @@ def wait_for_ajax():
         anything_in_flight |= running["document"] != "complete"
         anything_in_flight |= running["autofocus"] > 0
         anything_in_flight |= running["miqQE"] > 0
+        anything_in_flight |= running["miqProcessing"]
         log_msg = ', '.join(["{}: {}".format(k, str(v)) for k, v in running.iteritems()])
         # Log the message only if it's different from the last one
         if prev_log_msg != log_msg:
--- a/cfme/fixtures/pytest_selenium.py
+++ b/cfme/fixtures/pytest_selenium.py
@@ -327,6 +327,7 @@ def wait_for_ajax():
         anything_in_flight |= running["autofocus"] > 0
         anything_in_flight |= running["miqQE"] > 0
         anything_in_flight |= running["miqProcessing"]
+        anything_in_flight |= running["angular"]
         log_msg = ', '.join(["{}: {}".format(k, str(v)) for k, v in running.iteritems()])
         # Log the message only if it's different from the last one
         if prev_log_msg != log_msg:
--- a/cfme/js.py
+++ b/cfme/js.py
@@ -12,6 +12,9 @@ function xpath(root, xpath) {
 
 in_flight = """
 function isHidden(el) {if(el === null) return true; return el.offsetParent === null;}
+function angularGet() {
+    try { return (ManageIQ.angular.scope.afterGet === false); } catch(err) { return false; };
+}
 
 return {
     jquery: jQuery.active,
@@ -22,7 +25,8 @@ return {
     document: document.readyState,
     autofocus: (typeof checkMiqQE === "undefined") ? 0 : checkMiqQE('autofocus'),
     miqQE: (typeof checkAllMiqQE === "undefined") ? 0 : checkAllMiqQE(),
-    miqProcessing: ManageIQ.observe.processing || ManageIQ.observe.queue.length > 0
+    miqProcessing: ManageIQ.observe.processing || ManageIQ.observe.queue.length > 0,
+    angular: angularGet()
 };
 """
 
--- a/cfme/js.py
+++ b/cfme/js.py
@@ -30,6 +30,27 @@ return {
 };
 """
 
+get_events = """
+function timeoutsMaxId() {
+    try {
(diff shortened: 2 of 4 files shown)


dajoRH:
Lint report for commit 87bb2449fbdbfd4b19dcc8bb71126dafe3be5c91:

*CFME QE Bot*

psav:
Very excited about this

jkrocil:
This is amazing!

jkrocil:
Would you mind adding the following check as well please? https://github.com/ManageIQ/manageiq/issues/8809

mkoura:
jkrocil: I'll look into that

psav:
Thanks @mkoura

himdel:
Looks good, should catch all pending miq_observe requests.. :)

quarckster:
Doesn't work for me ( inputs still are not filled correctly.

dajoRH:
Lint report for commit a59071986f5d9a7aee25c7ac40e2d6f076a03ccb:

*CFME QE Bot*

mfalesni:
@mkoura You might also try moving this wait_for_ajax into the beginning of the for loop here https://github.com/RedHatQE/cfme_tests/blob/master/cfme/web_ui/__init__.py#L1356

mkoura:
@quarckster @jkrocil @psav I added check for angular requests described in https://github.com/ManageIQ/manageiq/issues/8809.
I used that try... catch... because the 'ManageIQ.angular.scope.afterGet' is not declared most of the time.

mkoura:
@mfalesni good idea, done

dajoRH:
Lint report for commit 00667a3c60b620e802157a3708d096e760a79a2b:

*CFME QE Bot*

psav:
So howz this looking? Really want to get this merged :)

quarckster:
psav, doesn't work for me

psav:
@quarckster I'll try it on the form that was giving me troubles too and see if it fixes it there

mkoura:
I've been testing the approach outlined in
https://spin.atomicobject.com/2015/05/12/testing-asynchronous-behavior-javascript-selenium/
Our situation is little bit more complex as we always have some active timeouts so we can't simply wait till there is no 'setTimeout' active (i.e. we would be waiting forever for `Object.keys(window.testing.timeouts).length === 0`)

The reason is that the 'window.miqExpressionPrefill' function schedules itself repeatedly:
```
   ...
      setTimeout(function () {
        miqExpressionPrefill(expEditor, false);
      }, 200);
    ...
```
As we are generating unique increasing id for each timing event, I think we can just check what is the highest id and than wait until that timing event and all timing events with lower ids (i.e. all timing events started before our check) are finished.

cfme/js.py
```diff
@@ -21,7 +24,9 @@
         && isHidden(document.getElementById("lightbox_div")),
     document: document.readyState,
     autofocus: (typeof checkMiqQE === "undefined") ? 0 : checkMiqQE('autofocus'),
-    miqQE: (typeof checkAllMiqQE === "undefined") ? 0 : checkAllMiqQE()
+    miqQE: (typeof checkAllMiqQE === "undefined") ? 0 : checkAllMiqQE(),
+    miqProcessing: ManageIQ.observe.processing || ManageIQ.observe.queue.length > 0,
+    angular: angularGet()
```
mfalesni: As it grows it might be better now to jsmin it like eg. line 111.

cfme/js.py
```diff
@@ -21,7 +24,9 @@
         && isHidden(document.getElementById("lightbox_div")),
     document: document.readyState,
     autofocus: (typeof checkMiqQE === "undefined") ? 0 : checkMiqQE('autofocus'),
-    miqQE: (typeof checkAllMiqQE === "undefined") ? 0 : checkAllMiqQE()
+    miqQE: (typeof checkAllMiqQE === "undefined") ? 0 : checkAllMiqQE(),
+    miqProcessing: ManageIQ.observe.processing || ManageIQ.observe.queue.length > 0,
```
mfalesni: this check should also be guarded as it is not guaranteed that the ManageIQ is defined. Eg. when we see an error, then it would fail without letting us interact with it.

psav: Agreed, we should protect against this

dajoRH:
Lint report for commit 01d19203d32abcc025a1803baa1b5c7bd0ec42bb:

`cfme/js.py`:
- :bangbang: Line 36:101: **E501** *line too long (106 > 100 characters)*

*CFME QE Bot*

quarckster:
Tested commit 01d1920 and I'm still experiencing the bug. By the way it seems this bug doesn't affect on Chrome.

mkoura:
@quarckster I noticed a bug right after commit :( I'm fixing it now.

mkoura:
@quarckster Also there is a patch that should be applied to miq_application.js I just extended current diff and I'm not sure if the test will re-patch already patched applance.

cfme/js.py
```diff
@@ -21,7 +24,30 @@
...
+    try {
+        var maxid = Object.keys(window.testing.timeouts)[Object.keys(window.testing.timeouts).length - 1];
+        return (maxid) ? maxid : 0;
+    } catch(err) { return 0; }
+}
+
+function timeoutsMinId() {
+    try {
+        var minid = Object.keys(window.testing.timeouts)[0];
+        return (min) ? minid : 0;
```
jkrocil: `return (min)` or `return (minid)`?

mkoura:
@jkrocil yes, that's the bug. also the id is string and should be int

data/patches/miq_application.js.diff
```diff
@@ -35,5 +35,60 @@
...
++window.testing = {
++  timeouts: {}
++};
++
++// Make a copy of the original setTimeout function
++window._setTimeout = window.setTimeout;
++window.setTimeout = function(callback, timeout) {
++  // We need a handle to store our timeout under, we can't just use the
++  // timeout ID because we don't know it until after we create the timeout
++  var handle = _.uniqueId();
```
jkrocil: So... we have underscore.js available?

jkrocil: If so - http://underscorejs.org/#uniqueId.
Does it guarantee that the ID will increment? Just curious.

jkrocil: Ah, I see it now... we don't care about the handle, it's used as a key. We care about the IDs returned by setTimeout... got it.

dajoRH:
Lint report for commit 6184e8cc5a7093219109a7b06a26c13a65cb7224:

*CFME QE Bot*

cfme/fixtures/pytest_selenium.py
```diff
@@ -305,6 +322,14 @@ def wait_for_ajax():
     """
 
     _thread_local.ajax_log_msg = ''
+    orig_events = get_events()
+
+    def _events_not_finished():
+        events = get_events()
+        not_finished = False
+        not_finished |= orig_events["maxTimeoutId"] >= events['minTimeoutId'] > 0
```
jkrocil: More comments here please. Why do we check if `events['minTimeoutId'] > 0` here? This expression evaluates False in case maxTimeoutID is e.g. 5 and minTimeoutId is 0... `5 >= 0 > 0` means `not_finished = False` which means `finished = True`... is this correct?

jkrocil:
@quarckster @mkoura re: patching again - you either have to revert the patch, patch it manually or use a fresh appliance.

data/patches/miq_application.js.diff
mkoura: Actually we care about the handle as we are using the id to determine what's new event that we don't need to care about. We have lodash available (similar to underscore) and I checked that uniqueId increments.

dajoRH:
Lint report for commit 6184e8cc5a7093219109a7b06a26c13a65cb7224:

*CFME QE Bot*

cfme/fixtures/pytest_selenium.py
mkoura: That's correct. events['minTimeoutId'] is 0 only when there are no events or something went wrong and we were not able to read the value at all. Maybe I should have done it like this:
```
if events['minTimeoutId'] > 0:
  not_finished |= orig_events["maxTimeoutId"] >= events['minTimeoutId']
```
to make the purpose more obvious

jkrocil: I would definitely prefer it that way, yes.

mkoura: or since we agreed that we don't care about setInterval and so there will be no other check necessary, I can rewrite in a sense of:
```
return orig_events["maxTimeoutId"] >= events['minTimeoutId'] > 0
```
(with or without if)

jkrocil: Sound good but with if please, Id still like to get rid of the chained comparison.

psav:
Though this may help in some cases - it still doesn't seem to help on the Tag create page

dajoRH:
Lint report for commit 180d5c5ca75f999e19f81f65336d8fa392ccc265:

*CFME QE Bot*

dajoRH:
Lint report for commit 6ba2c0fd630e51a56088ae06b7e053c6b774d76d:

*CFME QE Bot*

data/patches/miq_application.js.diff
```diff
@@ -35,5 +35,60 @@
...
++      callback();
++
++      // Once a timeout completes, we need to remove our reference to it
++      delete window.testing.timeouts[handle];
++    },
++    timeout
++  );
++
++  // Store the id of the timeout we just created so it can be queried
++  window.testing.timeouts[handle] = timeoutId;
```
jkrocil: Shouldn't we store this before calling `window._setTimeout`?

jkrocil:
Let's get rid of all the "diapers" (the try-excepts that catch everything) and replace then with checks for 'undefined'. This has to be as transparent as possible.

data/patches/miq_application.js.diff
mkoura: `window._setTimeout` schedules an event, returns id of the timer and we store that id. So we know the id to store only after we call `window._setTimeout`. The actual callback passed to window._setTimeout is called in the future (after `timeout`).

jkrocil: Ah, right, sorry. Not sure where I was looking...

mkoura:
@jkrocil I was trying to avoid that as e.g. accessing 'ManageIQ.angular.scope.afterGet' would require me to check if `ManageIQ === "undefined"`, `ManageIQ.angular === "undefined"`, `ManageIQ.angular.scope === "undefined"` in order to avoid raising TypeError.

psav:
Couldn't we try accessing and catch?

mkoura:
@psav that's how it's done right now

psav:
Well my point was that maybe we could reduce the scope of the try/catch to be less diapery, and ..... i dunno :)

dajoRH:
Lint report for commit 9bee436c038755b85598839a6954890ff880e400:

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

mfalesni:
@jkrocil Do we close this PR for now, since we have already solved the issue?

mfalesni:
Closing for now.

PR #3006, opened 2016-05-20
check vm config modified time.

amavinag:
* added method to check .vmx file modified
    time. as the vmware python api does not
    provider vm creation time attrib.

### DIFF
--- a/scripts/cleanup_old_vms.py
+++ b/scripts/cleanup_old_vms.py
@@ -2,12 +2,18 @@
 
 import argparse
 import datetime
+import pytz
 import re
 import sys
 from collections import defaultdict
+from dateutil import parser
 from threading import Lock, Thread
+from tzlocal import get_localzone
 
 from utils.log import logger
+from utils.conf import cfme_data
+from utils.conf import credentials
+from utils.ssh import SSHClient
 from utils.providers import list_all_providers, get_mgmt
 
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit c88c9c1c751662daba99a2924708486190ac4b5e:

*CFME QE Bot*

psav:
What happens if we copy files from one datastore to another, or migrate the vm to a different datastore?

amavinag:
it gets two attributes, 
1. current host the vm is registered to,
2. the datastore url: the current datastore the vm files are registered.
so if we migrate the vm to different datastore/host then that attributes are fetched accordingly.

the modified time of the .vmx file changes in both the above cases though(and vm will be deleted after another 'delta' hours).

psav:
Right that was my only concern, however, it actually works in our favour as it essentially "resets" the timer, I'm fine if machines take a little longer to get deleted, rather than get deleted too soon.

psav:
@mfalesni I wonder if we can merge this in with your metadata set/retrieve, so that when we create a vm via our automation we "give" it a created time. Though I'm happy with this as it is right now.

dajoRH:
Lint report for commit 44d68ceecfc9145f4b2430433d78a93594a009f0:

*CFME QE Bot*

PR #3020, opened 2016-05-25
C&U grouped tables support

mfalesni:
For @nachandr : Please check this out, by using `.groups()` or `.find_group()` you should now be able to access the groups of the table. `.rows()` does not work because I did not modify that one, you have to go through `.groups()`. You also have to pass a `candu=True` option to the CustomReport or CannedSavedReport in order for the classes to pick up the correct table classes.


dajoRH:
Lint report for commit 35bfb37dd0619f537f435a7f43d7bf2999fca90d:

*CFME QE Bot*

PR #3030, opened 2016-05-26
Added feature, to navigate through provider for middleware servers, deployments

jkandasa:
* Added navigation feature for the pages `middleware servers` and `middleware deployments`.
* `servers` can navigate through `provider` page
* `deployments` can navigate through `provider` page [or] `provider >> server` page
* New test cases added for deployment page
* usage, 
```
MiddlewareProvider.deployments()  # lists all deployments
MiddlewareProvider.deployments(provider=_haw)  # lists all deployments from _haw provider
MiddlewareProvider.deployments(provider=_haw, server=_ser)  # lists all deployments from _haw provider >> _ser

MiddlewareProvider.servers()  # lists all servers
MiddlewareProvider.servers(provider=_haw)  # lists all servers from _haw provider
```

{{pytest: cfme/tests/middleware/ -v --long-running --use-provider hawkular}}

### DIFF
--- a/cfme/middleware/deployment.py
+++ b/cfme/middleware/deployment.py
@@ -1,9 +1,10 @@
 import re
 from cfme.fixtures import pytest_selenium as sel
+from cfme.middleware.server import MiddlewareServer
 from cfme.web_ui import CheckboxTable, paginator
 from cfme.web_ui.menu import nav, toolbar as tb
-from mgmtsystem.hawkular import Deployment, Path
 from utils.db import cfmedb
+from utils.providers import get_crud, get_provider_key, list_middleware_providers
 from utils.varmeth import variable
 from . import LIST_TABLE_LOCATOR, MiddlewareBase
 
@@ -11,26 +12,42 @@ list_tbl = CheckboxTable(table_locator=LIST_TABLE_LOCATOR)
 
 
 def _db_select_query(name=None, server=None, provider=None):
+    """Column order: `nativeid`, `name`, `server_name`, `feed`, `provider_name`"""
+    t_ems = cfmedb()['ext_management_systems']
     t_ms = cfmedb()['middleware_servers']
     t_md = cfmedb()['middleware_deployments']
-    t_ems = cfmedb()['ext_management_systems']
-    query = cfmedb().session.query(t_md.nativeid, t_md.name,
-                                   t_md.ems_ref, t_ms.name, t_ems.name).join(t_ms,
-                                            t_md.server_id == t_ms.id).join(t_ems,
-                                            t_md.ems_id == t_ems.id)
+    query = cfmedb().session.query(t_md.nativeid.label('nativeid'), t_md.name,
+                                   t_ms.name.label('server_name'), t_ms.feed.label('feed'),
+                                   t_ems.name.label('provider_name'))\
+        .join(t_ms, t_md.server_id == t_ms.id).join(t_ems, t_md.ems_id == t_ems.id)
     if name:
         query = query.filter(t_md.name == name)
     if server:
-        query = query.filter(t_md.nativeid.like('%{}%'.format(server)))
+        query = query.filter(t_ms.name == server.name)
+        if server.feed:
+            query = query.filter(t_ms.feed == server.feed)
     if provider:
-        query = query.filter(t_ems.name == provider)
+        query = query.filter(t_ems.name == provider.name)
     return query
+
+
+def _get_deployments_page(provider, server):
+    if server:  # if server instance is provided navigate through server page
+        server.summary.reload()
+        if server.summary.relationships.middleware_deployments.value == 0:
+            return
+        server.summary.relationships.middleware_deployments.click()
+    elif provider:  # if provider instance is provided navigate through provider page
+        provider.summary.reload()
+        if provider.summary.relationships.middleware_deployments.value == 0:
+            return
+        provider.summary.relationships.middleware_deployments.click()
+    else:  # if None(provider and server) given navigate through all middleware deployments page
+        sel.force_navigate('middleware_deployments')
 
 nav.add_branch(
     'middleware_deployments', {
         'middleware_deployment': lambda ctx: list_tbl.select_row('Deployment Name', ctx['name']),
-        'middleware_deployment_detail':
-            lambda ctx: list_tbl.click_row_by_cells({'Deployment Name': ctx['name']}),
     }
 )
 
@@ -43,78 +60,110 @@ class MiddlewareDeployment(MiddlewareBase):
     Args:
         name: Name of the deployment
         provider: Provider object (HawkularProvider)
-        id: Native id (internal id) of deployment
-        server: Server name of the deployment
+        server: Server object of the deployment (MiddlewareServer)
+        nativeid: Native id (internal id) of deployment
 
     Usage:
 
         mydeployment = MiddlewareDeployment(name='Foo.war',
-                                server='Bar-serv',
+                                server=ser_instance,
                                 provider=haw_provider)
 
-        deployments = MiddlewareDeployment.deployments()
+        deployments = MiddlewareDeployment.deployments() [or]
+        deployments = MiddlewareDeployment.deployments(provider=haw_provider) [or]
+        deployments = MiddlewareDeployment.deployments(provider=haw_provider,server=ser_instance)
 
     """
     property_tuples = [('name', 'name')]
 
-    def __init__(self, name, provider=None, **kwargs):
+    def __init__(self, name, server, provider=None, **kwargs):
         if name is None:
             raise KeyError("'name' should not be 'None'")
+        if not isinstance(server, MiddlewareServer):
+            raise KeyError("'server' should be an instance of MiddlewareServer")
         self.name = name
+        self.server = server
         self.provider = provider
-        self.id = kwargs['id'] if 'id' in kwargs else None
-        self.server = kwargs['server'] if 'server' in kwargs else None
+        self.nativeid = kwargs['nativeid'] if 'nativeid' in kwargs else None
 
     @classmethod
-    def deployments(cls, provider=None):
+    def deployments(cls, provider=None, server=None):
         deployments = []
-        if provider:
-            # if provider instance is provided try to navigate provider's deployments page
-            # if no deployments are registered in provider, returns empty list
-            if not provider.load_all_provider_deployments():
-                return []
-        else:
-            # if provider instance is not provided then navigates  to all deployments page
-            sel.force_navigate('middleware_deployments')
+        _get_deployments_page(provider=provider, server=server)
         if sel.is_displayed(list_tbl):
-            for page in paginator.pages():
+            _provider = provider  # In deployment UI, we cannot get provider name on list all page
+            for _ in paginator.pages():
                 for row in list_tbl.rows():
-                    deployments.append(MiddlewareDeployment(name=row.deployment_name.text,
-                                                            server=row.server.text))
+                    _server = MiddlewareServer(provider=provider, name=row.server.text)
+                    deployments.append(MiddlewareDeployment(provider=_provider, server=_server,
+                                                            name=row.deployment_name.text))
         return deployments
 
     @classmethod
-    def deployments_in_db(cls, server=None, provider=None):
+    def deployments_in_db(cls, server=None, provider=None, strict=True):
         deployments = []
-        rows = _db_select_query(server=server, provider=(provider.name if provider else None)).all()
+        rows = _db_select_query(server=server, provider=provider).all()
+        _provider = provider
         for deployment in rows:
-            deployments.append(MiddlewareDeployment(id=deployment[0], name=deployment[1],
-                                                    server=deployment[3], provider=deployment[4]))
+            if strict:
+                _provider = get_crud(get_provider_key(deployment.provider_name))
+            _server = MiddlewareServer(name=deployment.server_name, feed=deployment.feed,
+                                       provider=provider)
+            deployments.append(MiddlewareDeployment(nativeid=deployment.nativeid,
+                                                    name=deployment.name,
+                                                    server=_server, provider=_provider))
         return deployments
 
     @classmethod
-    def deployments_in_mgmt(cls, provider):
+    def _deployments_in_mgmt(cls, provider, server=None):
         deployments = []
         rows = provider.mgmt.list_server_deployment()
         for deployment in rows:
-            deployments.append(MiddlewareDeployment(provider=provider.name,
-                    id=deployment.id, name=deployment.name,
-                    server=re.sub(r'~~$', '', deployment.path.resource[0])))
+            _server = MiddlewareServer(name=re.sub(r'~~$', '', deployment.path.resource[0]),
+                                       feed=deployment.path.feed,
+                                       provider=provider)
+            _include = False
+            if server:
+                if server.name == _server.name:
+                    _include = True if not server.feed else server.feed == _server.feed
+            else:
+                _include = True
+            if _include:
+                deployments.append(MiddlewareDeployment(provider=provider, server=_server,
+                                                        nativeid=deployment.id,
+                                                        name=deployment.name))
+
         return deployments
+
+    @classmethod
+    def deployments_in_mgmt(cls, provider=None, server=None):
+        if provider is None:
+            deployments = []
+            for _provider in list_middleware_providers():
+                deployments.extend(cls._deployments_in_mgmt(get_crud(_provider), server))
+            return deployments
+        else:
+            return cls._deployments_in_mgmt(provider, server)
+
+    def _on_detail_page(self):
+        """Override existing `_on_detail_page` and return `False` always.
+        There is no uniqueness on summary page of this resource.
+        Refer: https://github.com/ManageIQ/manageiq/issues/9046
+        """
+        return False
 
     def load_details(self, refresh=False):
         if not self._on_detail_page():
-            sel.force_navigate('middleware_deployment_detail', context={'name': self.name})
+            _get_deployments_page(provider=self.provider, server=self.server)
+            list_tbl.click_row_by_cells({'Deployment Name': self.name, 'Server': self.server.name})
         if refresh:
             tb.refresh()
 
     @variable(alias='ui')
     def deployment(self):
         self.summary.reload()
-        if self.summary.properties:
-            return Deployment(name=self.summary.properties.name.text_value,
-                              id=self.summary.properties.nativeid.text_value, path=None)
-        return None
+        self.id = self.summary.properties.nativeid.text_value
+        return self
 
     @deployment.variant('mgmt')
     def deployment_in_mgmt(self):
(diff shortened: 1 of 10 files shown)


dajoRH:
Lint report for commit f46725858dea3f003d311fb6dc4cec56d6331e4b:

*CFME QE Bot*

dajoRH:
Lint report for commit 02592c3ce1dcfa937b02409e206845442273dc6a:

*CFME QE Bot*

cfme/middleware/deployment.py
```diff
@@ -43,91 +56,103 @@ class MiddlewareDeployment(MiddlewareBase):
...
+        deployments = MiddlewareDeployment.deployments(provider=haw_provider,server=ser_instance)
 
     """
     property_tuples = [('name', 'name')]
 
-    def __init__(self, name, provider=None, **kwargs):
+    def __init__(self, name, provider=None, server=None, **kwargs):
         if name is None:
             raise KeyError("'name' should not be 'None'")
+        if server is None or not isinstance(server, MiddlewareServer):
```
hhovsepy: can server be None?
if yes then I think check must be:
if server and not isinstance(server, MiddlewareServer):

cfme/middleware/deployment.py
```diff
@@ -43,91 +56,103 @@ class MiddlewareDeployment(MiddlewareBase):
...
-        rows = _db_select_query(server=server, provider=(provider.name if provider else None)).all()
+        rows = _db_select_query(server=server.name if server else None,
+                                provider=(provider.name if provider else None)).all()
+        _server = server
         for deployment in rows:
+            if not server:
+                _server = MiddlewareServer(name=deployment[3], provider=provider)
             deployments.append(MiddlewareDeployment(id=deployment[0], name=deployment[1],
-                                                    server=deployment[3], provider=deployment[4]))
+                                                    server=_server, provider=provider))
```
hhovsepy: here the idea is to construct object fully from db fields, to be validated later

cfme/middleware/deployment.py
jkandasa: @hhovsepy No, server can not be `None`. We should not allow server as `None`

dajoRH:
Lint report for commit 0ac977529d853c0e8018122004ddfa8d31dacaf7:

*CFME QE Bot*

cfme/middleware/server.py
```diff
@@ -59,25 +67,19 @@ def __init__(self, name, provider=None, **kwargs):
...
+        _get_servers_page(provider=provider)
         if sel.is_displayed(list_tbl):
-            for page in paginator.pages():
+            for _ in paginator.pages():
                 for row in list_tbl.rows():
                     servers.append(MiddlewareServer(name=row.server_name.text,
                                                     id=row.feed.text,
                                                     product=row.product.text,
-                                                    provider=row.provider.text))
+                                                    provider=provider))
```
hhovsepy: here provider must be get from page, to be validated

cfme/middleware/server.py
```diff
@@ -88,27 +90,28 @@ def headers(cls):
...
     @classmethod
-    def servers_in_db(cls, server=None, provider=None):
+    def servers_in_db(cls, name=None, provider=None):
         servers = []
-        rows = _db_select_query(server=server, provider=(provider.name if provider else None)).all()
+        rows = _db_select_query(server=name, provider=(provider.name if provider else None)).all()
         for server in rows:
             servers.append(MiddlewareServer(name=server[0], id=server[1],
-                                            product=server[2], provider=server[3]))
+                                            product=server[2], provider=provider))
```
hhovsepy: same here

cfme/middleware/server.py
jkandasa: @hhovsepy we should pass `provider` object, we cannot pass provider text in the place of provider object.

cfme/middleware/deployment.py
jkandasa: @hhovsepy Thank you! Updated.

dajoRH:
Lint report for commit ff4839f2728faa601a468aba26f00565c395c36a:

*CFME QE Bot*

cfme/tests/middleware/test_middleware_deployment.py
```diff
@@ -1,33 +1,48 @@
...
 import pytest
+from cfme.middleware import get_random_list
 from cfme.middleware.deployment import MiddlewareDeployment
+from cfme.middleware.server import MiddlewareServer
 from utils import testgen
 from utils.version import current_version
 
 pytestmark = [
-    pytest.mark.usefixtures('single_middleware_provider'),
+    pytest.mark.usefixtures('setup_provider'),
```
hhovsepy: we need to be sure that there are no any other providers

jkandasa: @hhovsepy Why do we want no other providers present?

cfme/middleware/deployment.py
hhovsepy: ok, then please update init and comments on instantiation

cfme/middleware/server.py
hhovsepy: ok, then either we need to construct provider from "row.provider.text" or other solution,
if we validate data what we provide, no reason of checking, we need to validate if data was shown correctly in UI and in DB

cfme/tests/middleware/test_middleware_deployment.py
hhovsepy: if my MIQ will have other providers, then more data will be shown in UI and in DB,
but we will not know which provider it was, as in yaml file we have single provider

cfme/tests/middleware/test_middleware_deployment.py
```diff
@@ -1,33 +1,48 @@
...
     Steps:
         * Get deployments list from UI
         * Get deployments list from Database
-        * Get deployments list from Management system(Hawkular)
-        * Compare size of all the list [UI, Database, Management system]
+        * Compare size of all the list [UI, Database]
     """
     ui_deps = _get_deployments_set(MiddlewareDeployment.deployments())
     db_deps = _get_deployments_set(MiddlewareDeployment.deployments_in_db())
-    mgmt_deps = _get_deployments_set(MiddlewareDeployment.deployments_in_mgmt(provider=provider))
```
hhovsepy: when you put back 'single_middleware_provider' you can return mgmt validation as well,
here we need to be sure that data is correct everywhere, the data we have in hawkular side, is correctly replicated into database, and then correctly shown in UI, otherwise no reason to verify

cfme/tests/middleware/test_middleware_server.py
```diff
@@ -4,33 +4,30 @@
 from utils.version import current_version
 
 pytestmark = [
-    pytest.mark.usefixtures('single_middleware_provider'),
+    pytest.mark.usefixtures('setup_provider'),
```
hhovsepy: same goes here

cfme/middleware/server.py
jkandasa: @hhovsepy we may not construct provider easily, because it needs more details(`ip`,`port`, etc.,). We may introduce in properties to hold these values.

cfme/tests/middleware/test_middleware_deployment.py
jkandasa: @hhovsepy we have two type of test cases here. one with `provider` and another without `provider`.
* When we go without `provider`, we no need to fetch data from `mgmt`. Because we need to query on multiple `mgmt` systems. So I have removed `mgmt` check in this test. Test happens only between `DB` and `UI`.
* When we go with `provider`, In `UI` we have navigation option to select particular `provider`, In `db` can filter with `provider` name. In `mgmt` nothing to do(we are getting `mgmt` from `provider` object).

When we go with `single_middleware_provider`, displays data only from single `provider`. There is no meaning to perform test through `provider`. We already filtered out other providers data by keeping only one `provider`.

cfme/tests/middleware/test_middleware_deployment.py
jkandasa: @hhovsepy `mgmt` check will be covered in `test_list_provider_deployments`.
Here testing between `UI` and `DB`. Later we will add `REST`

cfme/tests/middleware/test_middleware_deployment.py
hhovsepy: what if there is a bug in MIQ? it displays wrongly list without provider? 
each test should be itself final checking that everything is correct , without depending on other tests

dajoRH:
Lint report for commit aad4f5da62a42481d3878490508d522c23925a68:

*CFME QE Bot*

cfme/tests/middleware/test_middleware_deployment.py
hhovsepy: mgmt itself already is the REST to hawkular

jkandasa: @hhovsepy CFME `REST`

cfme/middleware/server.py
hhovsepy: then we need another approach, other classes for reading values and comparing

hhovsepy: what if provider is displayed wrong? test will always pass

cfme/tests/middleware/test_middleware_deployment.py
jkandasa: @hhovsepy When we do not have `provider`, we pick data from **list all**. I feel for **list all** page this enough.
If we have bug, lists provider-XYZ data in provider-ABC, how will `single_provider` do? ;)

cfme/middleware/deployment.py
jkandasa: @hhovsepy updated

cfme/middleware/server.py
mfalesni: Well, having other providers should not constitute a problem though :)

cfme/middleware/deployment.py
mfalesni: Yeah, should anything is required, then it should not have a default value.

cfme/middleware/server.py
mfalesni: By the way in utils/provider.py there should be a function to get provider by name so you use provider name instead of crud.

mfalesni: @hhovsepy *"provider is displayed wrong"* if you use the utils/providers' `provider = get_crud(get_provider_key(provider_name))` then it will give you an error.

mfalesni: You can possibly have a param saying that it should only append things that have a provider that is in yaml. Therefore the test would be able to run even though you would have some unknown provider in there.

cfme/middleware/deployment.py
jkandasa: @mfalesni I forget to remove default value. Updated now. Thank you.

cfme/tests/middleware/test_middleware_server.py
hhovsepy: does 'setup_provider' deletes and recreated provider each time?

dajoRH:
Lint report for commit e77ee68647cfad860aca067089655b0270411b1d:

*CFME QE Bot*

cfme/tests/middleware/test_middleware_server.py
jkandasa: @hhovsepy No, Creates only once, if the provider not available.

cfme/tests/middleware/test_middleware_deployment.py
hhovsepy: single provider will not resolve that case, for this we need other solution, probably will not even have,
but my intension is in other: 
1) comparing UI with DB, without knowing whether all these values are correct or not, do not relay on other tests, maybe they are disabled or failing for some other reason
2) make sure that we know what is happening in MIQ, we own single provider which we have just created, and have all tests based on assured info (this is general rule)

cfme/tests/middleware/test_middleware_server.py
hhovsepy: then, each test will have impact on other tests, this is potential intermittent failures

hhovsepy: see my comments above

cfme/tests/middleware/test_middleware_deployment.py
hhovsepy: 3) the result of some test (disable something, delete or edit) will not have impact on other test which uses the same provider

cfme/middleware/server.py
hhovsepy: my comment here means: do not ignore cell value displayed in the UI

cfme/tests/middleware/test_middleware_deployment.py
jkandasa: @hhovsepy as @mfalesni mentioned `provider = get_crud(get_provider_key(provider_name))` we can use this method to create provider object.
When doing operations like `edit`, `delete`, `disable`. at the end of the test we need to keep resource to original state [or] in that case we may do recreation of `provider`(for that test alone).

cfme/middleware/server.py
jkandasa: @hhovsepy  we can use like `provider = get_crud(get_provider_key(row.provider.text))`

cfme/tests/middleware/test_middleware_deployment.py
hhovsepy: question: from where we have 'provider_name', if tests give us default provider? can we rely on UI?

hhovsepy: my idea is that all providers mentioned in yaml file, have in test directly, without getting them from UI, so if there is bug in UI, test will not be based on this buggy UI

hhovsepy: @mfalesni can we have several middleware providers configured in yaml file, and have them in tests?
for instance here: "def test_hawkular_crud(provider):" to have set of providers?

mfalesni: @hhovsepy 
Yes, of course, that is the purpose of testgen ;)

*Can we rely on UI?*
I guess that not 100%, but that's why we test it in the first place :)

So you use testgen to parametrize the tests on providers and then base the tests off the provider, so you don't care about the other influences. That is why eg. our VM objects first ask self.provider to load the details and then it clicks through the relationship to see only the VMs of that provider so you are 100% sure you only see the vms of that provider when you are looking for that particular VM.

hhovsepy: OK then I misunderstood what @psav told me during our previous sessions. I thought that cfme_tests allow to have only single middleware provider. That is why I added fixture to clean all providers and leave only one.
This changes a lot.
We can have several providers in yaml, create them in fixtures, and "Provider" objects have in tests, not taken from UI via "get_crud(get_provider_key(provider_name))" but have them provided directly from config files.

mfalesni: Yes, that is exactly the approach that should be taken.

hhovsepy:
So my suggestion will be:
1. Create new fixture which will create multiple providers and remove all existing ones.
2. Modify those test cases which lists resources for all providers to use this fixture, and to receive providers set as a param, verify data in UI and DB with data from mgmt_system of each provider. Or even move these tests into separate class.

The rest of tests which depend on particular provider, leave as suggested in this PR.

Wdyt @jkandasa?

hhovsepy:
@mfalesni is it possible to create fixture which runs before test suite, once?

cfme/middleware/deployment.py
```diff
@@ -43,91 +56,101 @@ class MiddlewareDeployment(MiddlewareBase):
...
+        _get_deployments_page(provider=provider, server=server)
         if sel.is_displayed(list_tbl):
-            for page in paginator.pages():
+            _server = server
+            for _ in paginator.pages():
                 for row in list_tbl.rows():
-                    deployments.append(MiddlewareDeployment(name=row.deployment_name.text,
-                                                            server=row.server.text))
+                    if not server:
+                        _server = MiddlewareServer(provider=provider, name=row.server.text)
```
hhovsepy: I think it is better to create server for each row, _server = MiddlewareServer(provider=provider, name=row.server.text)
and put that instance in deployment object.
To verify each deployment's displayed server value.

hhovsepy:
@jkandasa  +1 I like the solutions provided in this commit.

cfme/middleware/deployment.py
```diff
@@ -43,91 +56,101 @@ class MiddlewareDeployment(MiddlewareBase):
...
-        if self.name:
-            deployment = _db_select_query(self.name, self.server).first()
-            if deployment:
-                return Deployment(deployment[0], deployment[1], Path(deployment[2]))
-            return None
-        raise KeyError('Variable "name" not set!')
+        deployment = _db_select_query(self.name, self.server).first()
+        if deployment:
+            _server = MiddlewareServer(name=deployment[3], provider=self.provider)
+            return MiddlewareServer(provider=self.provider, server=_server, id=deployment[0],
```
hhovsepy: MiddlewareDeployment I guess

dajoRH:
Lint report for commit 455542dcd6adc130bba5c37d31ff0ac226498514:

*CFME QE Bot*

cfme/tests/middleware/test_middleware_deployment.py
jkandasa: Added code to get `provider`'s object with `get_crud(get_provider_key(provider_name))`.
* For deployment page on `MiddlewareDeployment.deployments()` we will not get `provider` detail. This page does not have provider data on list all deployments.
* Added code to get deployments data from `mgmt` system

cfme/middleware/deployment.py
jkandasa: @hhovsepy Thank you! Fixed it.

mfalesni:
@hhovsepy 

1) A fixture should not bother about multiple providers unless you have some special test case where you must test interaction/interference between multiple providers
2) It is possible to have a fixture that runs once, if you session-scope it.

dajoRH:
Lint report for commit 9797f1ddb9c6657786d2348bf834f7e879fbc9be:

*CFME QE Bot*

jkandasa:
@hhovsepy I guess we no need to create any extra fixture for now.
We can write new test case(later some point) something like, get providers list from configuration file and add all providers, do some test.. so and so...
@hhovsepy @mfalesni Thank you so much for the review comments. I hope I have addressed/fixed all the review comments. Please let me know if I missed out any comments.

dajoRH:
*CFME QE Bot*

dajoRH:
Lint report for commit 402228e478560e98d26148429b32526002e39d2d:

*CFME QE Bot*

mfalesni:
`/me` *removed wip to exercise the PR tester ...*

mfalesni:
There are some errors, sending an e-mail

dajoRH:
Lint report for commit fdf37d409a9c4d4fa119f0063811f1564cbee46d:

*CFME QE Bot*

dajoRH:
Lint report for commit b52527352dd029fde86d47eb00ee93cc6ce1dafd:

*CFME QE Bot*

dajoRH:
Lint report for commit 2a7d145ab78b54f78bac70380dc55f4b530254ae:

*CFME QE Bot*

mfalesni:
Still errors :(

jkandasa:
@mfalesni seems http://livingontheedge.hawkular.org/ has some issues

mfalesni:
Yeah, when I log in it errors pop out in the flash messages.

dajoRH:
Lint report for commit 0181b933fd14ea22cbd8c0bf93a6afdc67d8df2a:

*CFME QE Bot*

dajoRH:
Lint report for commit 38a97857ee04d70c6ff4284fcd19cd047b4cf6fb:

*CFME QE Bot*

hhovsepy:
"test_hawkular_crud" failure is because of known issue in MiQ.
when there was provider before, deletion and recreate fails in UI because of some timeout issue.

jkandasa:
@hhovsepy Thank you!
@mfalesni Test failed in `test_hawkular_crud`, What shall I do now?

mfalesni:
So, per discussion with the guys I am gonna merge it, there is one failing test bu that is to be addressed.

PR #3045, opened 2016-06-01
compare datastore url with vm config file path.

amavinag:
* get vm config file path and datatstore url
    compare them, some times the local storage
    may change the config file path.
  * close the ssh connection after runcommand.


dajoRH:
Lint report for commit 99a29f0843db16785066141aafc7626de26fe475:

*CFME QE Bot*

PR #3053, opened 2016-06-03
Fix the SSH unavailable errors that were popping up

mfalesni:

PR #3066, opened 2016-06-07
Fix ssh for usecases not involving appliance.

amavinag:
* the recent changes in #3053 breaks ssh connection
    while template upload. these scripts does not need
    baseurl to be set. store.baseurl can be forced to
    read only when there is no kwargs passed with non-default
    ssh values.


dajoRH:
Lint report for commit 03d25015c38148b686ccfd0ece021389013858c4:

*CFME QE Bot*

PR #3072, opened 2016-06-10
Added new test cases to check "Download as x" toolbar action for all Middleware objects.

hhovsepy:


jkandasa:
@hhovsepy Where are you validating the downloaded content?

hhovsepy:
It is not supported to access file from test framework, so validating only download buttons.

cfme/tests/middleware/test_middleware_downloads.py
```diff
@@ -0,0 +1,53 @@
...
+    objecttype.download(filetype, provider=provider,
+                        server=_get_random_server(provider=provider))
+
+
+def verify_download(filetype):
+    """ Verifies whether download was successful and no error was shown.
+    Currently test framework does not allow to access download files.
+    So leaving file content checks later.
+    """
+    flash.assert_no_errors()
```
psav: Should we put a log warning here that it couldn't verify the file contents?

hhovsepy: I do not think so, it could confuse us.
I think better to have uncovered functionality commented in code.

hhovsepy:
Let's have it tested as much as possible for now, later on when downloads will be accessible, we will increase coverage here.

PR #3073, opened 2016-06-10
Automation for test_assign_events_to_container_image_control_policy

quarckster:


cfme/tests/control/test_basic.py
```diff
@@ -110,6 +110,18 @@ def random_host_condition():
 
 
 @pytest.yield_fixture
+def random_container_image_condition():
```
psav: I would like to remove the name "random" from here and the fixture below. If it is just string that is random, I don't think we should call it a random condition. It implies a level of unrepeatability.

cfme/tests/control/test_basic.py
```diff
@@ -428,6 +440,24 @@ def test_assign_host_condition_to_host_policy(
...
+        container_image_condition,
+        soft_assert):
+    """This test checks whether an event is assigned to container image control policy.
+    Steps:
+        * Create a container image control policy.
+        * Assign an event to the created policy.
+    """
+    random_container_image_control_policy.assign_conditions(
+        container_image_condition)
+    soft_assert(random_container_image_control_policy.is_condition_assigned(
```
mfalesni: If it checks only one single thing, it would be better to do:
```python
request.addfinalizer(random_container_image_control_policy.assign_conditions)
assert random_container_image_control_policy.is_condition_assigned(container_image_condition)
```

dajoRH:
Lint report for commit 40704a6c03c1f09bf9feb1a633422ed3f3ac166d:

*CFME QE Bot*

PR #3075, opened 2016-06-13
Changed address getter to use get_yaml_config

psav:
* get_yaml_file() call replaced with get_yaml_config() as the file
  doesn't exist on later appliances


dajoRH:
Lint report for commit 01e8082c38ff209528ac8318eb69f9a6910843b3:

*CFME QE Bot*

PR #3084, opened 2016-06-15
SSH container support

mfalesni:
DONE:
* Prefixing with docker exec with proper quoting
* file upload and download via intermediate location
* Pass the fact it is running in container using env.yaml

### DIFF
--- a/utils/ssh.py
+++ b/utils/ssh.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import fauxfactory
 import iso8601
 import re
 import socket
@@ -16,6 +17,7 @@ from utils.log import logger
 from utils.net import net_check
 from fixtures.pytest_store import store
 from utils.path import project_path
+from utils.quote import quote
 from utils.timeutil import parsetime
 
 
@@ -35,12 +37,16 @@ class SSHClient(paramiko.SSHClient):
 
     Allows copying/overriding and use as a context manager
     Constructor kwargs are handed directly to paramiko.SSHClient.connect()
+
+    If ``container`` param is specified, then it is assumed that the VM hosts a container of CFME.
+    The ``container`` param then contains the name of the container.
     """
     def __init__(self, stream_output=False, **connect_kwargs):
         super(SSHClient, self).__init__()
         self._streaming = stream_output
         # deprecated/useless karg, included for backward-compat
         self._keystate = connect_kwargs.pop('keystate', None)
+        self._container = connect_kwargs.pop('container', None)
 
         # load the defaults for ssh
         default_connect_kwargs = {
@@ -63,6 +69,10 @@ class SSHClient(paramiko.SSHClient):
         self._connect_kwargs = default_connect_kwargs
         self.set_missing_host_key_policy(paramiko.AutoAddPolicy())
         _client_session.append(self)
+
+    @property
+    def is_container(self):
+        return self._container is not None
 
     def __repr__(self):
         return "<SSHClient hostname={} port={}>".format(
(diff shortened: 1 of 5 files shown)


psav:
Could we just have the container vm have a "service" running on a port - if we detect that port, then we know we are hitting a containerized version?

dajoRH:
Lint report for commit 0dbf6939eeaf79f0e51cc0c80f7333638b282d2f:

*CFME QE Bot*

mfalesni:
@psav Possibly, it could be something REST that even serves you username and password :D

anewmanRH:
A file on disk I think would work just as well.

psav:
@anewmanRH the problem here is we have nobbled the SSH client to be container or no.....so we need to know that before we can use it :)

mfalesni:
Exactly.

psav:
A banner check on SSH could be sufficient

mfalesni:
Yeah, I have to figure that out, that sounds like a good idea.

dajoRH:
Lint report for commit 81ecfb3f9c6f1b5ae335b7da2643849c456af0de:

*CFME QE Bot*

mfalesni:
@RonnyPfannschmidt @psav This will probably have to be aligned somehow with the appliance stack efforts. This needs only one extra parameter by the way. I guess the pytest_store could have its own configure which would configure the appliance IP and ports and it would come first in the order of plugins.

dajoRH:
Lint report for commit ae0bb8fd56583e67d14dc67528791c416f14ec82:

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

dajoRH:
Lint report for commit 8864ede5c1eb5763c516426a8548c0a274d87cde:

*CFME QE Bot*

utils/ssh.py
```diff
@@ -135,10 +148,14 @@ def get_transport(self, *args, **kwargs):
             self.connect()
         return super(SSHClient, self).get_transport(*args, **kwargs)
 
-    def run_command(self, command, timeout=RUNCMD_TIMEOUT, reraise=False):
+    def run_command(self, command, timeout=RUNCMD_TIMEOUT, reraise=False, ensure_host=False):
         if isinstance(command, dict):
             command = version.pick(command)
         logger.info("Running command `%s`", command)
+        if self._container and not ensure_host:
```
psav: ```self.is_container``` here for readability???

mfalesni: It somehow makes sense to me to use the same variable for check and for the value later ... dunno. Of course I can change it.

utils/ssh.py
```diff
@@ -208,13 +225,28 @@ def run_rake_command(self, command, timeout=RUNCMD_TIMEOUT):
 
     def put_file(self, local_file, remote_file='.', **kwargs):
         logger.info("Transferring local file %s to remote %s", local_file, remote_file)
-        return SCPClient(self.get_transport(), progress=self._progress_callback).put(
-            local_file, remote_file, **kwargs)
+        if self.is_container:
+            tempfilename = '/share/temp_{}'.format(fauxfactory.gen_alpha())
+            scp = SCPClient(self.get_transport(), progress=self._progress_callback).put(
+                local_file, tempfilename, **kwargs)
+            self.run_command('mv {} {}'.format(tempfilename, remote_file))
```
psav: How does this "mv" it into the container? Do you have to know the remote path to the container? ie a mounted volume?

mfalesni: It assumes there is /share directory visible both in container and the host VM. A prerequisite.

psav: OK, so you dump it in share, and then retriving it is using the same name, but is run inside the docker context...got it

mfalesni: Exactly. Basically one additional step with a random name.

The docs I pushed clarify the prerequisities for the VM that hosts the docker container.

dajoRH:
Lint report for commit a380b651ee6fdc9c075fc07614d36d0860f3bc9c:

*CFME QE Bot*

psav:
:cake::punch:

PR #3085, opened 2016-06-15
On FlashMessageException dump all flash messages displayed

mfalesni:
Now when you do eg. `assert_message_match()` and if it does not match, it will modify the exception and do:

```
FlashMessageException: No matching flash message for 'asdf'
Present flash messages:
success: Infrastructure Provider "myprovider" was saved
```

{{pytest: cfme/tests/infrastructure/test_providers.py -v --use-provider complete}}

### DIFF
--- a/cfme/web_ui/flash.py
+++ b/cfme/web_ui/flash.py
@@ -136,6 +136,23 @@ def verpick_message(f):
     return g
 
 
+def onexception_printall(f):
+    """If FlashMessageException happens, appends all the present flash messages in the error text"""
+    @wraps(f)
+    def g(*args, **kwargs):
+        try:
+            return f(*args, **kwargs)
+        except FlashMessageException as e:
+            err_text = str(e)
+            messages = get_messages()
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit 3f110870e547682b3581e8e9b7f1e760616884d5:

*CFME QE Bot*

psav:
Can we get a PRT run on a test that uses flash messages......or does login count on this?

mfalesni:
@psav I am not entirely sure that we can point it to some "broken" test, this is unfortunately manual job to check it against UI ... :(

psav:
I just meant run it against something that uses flash messages so we know it doesn't break normal usage :)

mfalesni:
@psav added, retesting.

mfalesni:
@psav Your opinion? The validation tests are ruining it though because of the red Required instead of flash message :)

That reminds me we should probably abolish the sneaky avoidance of clicking Add ...

PR #3101, opened 2016-06-20
Fix analysis tests

mfalesni:


dajoRH:
Lint report for commit f0adcadb4534f84ef73d440fe83eb4b8e855f55b:

*CFME QE Bot*

cfme/web_ui/__init__.py
```diff
@@ -3056,6 +3056,10 @@ class DynamicTable(Pretty):
     """
     pretty_attrs = "root_loc", "default_row_item"
     ROWS = ".//tbody/tr[not(contains(@id, 'new_tr'))]"
+    DELETE_ALL = {
```
psav: Should it be called ```DELETE_ALL```? Doesn't it delete a single field? Is ```ALL``` referencing all versions?

mfalesni: All means it is used for deleting all the fields, since it references possibly all the Minus buttons that are used to delete stuff.

psav: OK

PR #3103, opened 2016-06-20
Fixed Orch template issues

sshveta:


dajoRH:
Lint report for commit 4dcc57f70c6f5ff7a4bd19a204dc2aef0a5313c3:

*CFME QE Bot*

cfme/tests/services/test_orchestration_template.py
```diff
@@ -49,12 +49,17 @@ def pytest_generate_tests(metafunc):
...
+    return fauxfactory.gen_alphanumeric()
+
+
 def test_orchestration_template_crud(provisioning):
     template_type = provisioning['stack_provisioning']['template_type']
     template = OrchestrationTemplate(template_type=template_type,
                                      template_name=fauxfactory.gen_alphanumeric(),
                                      description="my template")
-    template.create(METHOD_TORSO)
+    method = METHOD_TORSO.replace('CloudFormation', random_desc())
```
psav: If we are inserting a randomized string into the description, is there a need to have two ```METHOD_TORSO```s defined now? Also, as ```random_desc()``` does not hold any state etc, I don't see the real need to create a function for running that single line of code. I could be missing something though :)

cfme/tests/services/test_orchestration_template.py
```diff
@@ -77,9 +84,10 @@ def test_name_required_error_validation(provisioning):
     template = OrchestrationTemplate(template_type=template_type,
                                      template_name=None,
                                      description="my template")
+    method = METHOD_TORSO.replace('CloudFormation', random_desc())
 
     with error.expected(flash_msg):
-        template.create(METHOD_TORSO)
+        template.create(method)
```
psav: Thinking about this, why not create a function scoped fixture that would return a ```METHOD_TORSO``` with a randomized string in the description? That way you have eliminated the duplicate random calls, and making if function scoped will mean you don't need to worry about them colliding.

dajoRH:
Lint report for commit ac7097f604f6088b2f591410b3ad3807c9af8990:

*CFME QE Bot*

PR #3112, opened 2016-06-22
Fixed a bad dep call

psav:
* Why make a circular loop, when we can do this ;)


dajoRH:
Lint report for commit ee99c928da42b39e5bab8712aafa8f895cd18c18:

*CFME QE Bot*

PR #3118, opened 2016-06-23
Fix cloud provider form

mkoura:
{{pytest:  cfme/tests/cloud/test_providers.py -v -k test_region_required_validation}}

### DIFF
--- a/cfme/cloud/provider.py
+++ b/cfme/cloud/provider.py
@@ -43,7 +43,6 @@ properties_form_55 = Form(
             '5.5': AngularSelect("emstype")}),
         ('azure_tenant_id', Input("azure_tenant_id")),
         ('name_text', Input("name")),
-        ('azure_region_select', AngularSelect("provider_region")),
         ('hostname_text', Input("hostname")),
         ('ipaddress_text', Input("ipaddress"), {"removed_since": "5.4.0.0.15"}),
         ('region_select', {version.LOWEST: Select("select#provider_region"),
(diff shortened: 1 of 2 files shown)


dajoRH:
Lint report for commit 98e9d283570fff03e56cbd34d79d23c92d6275b4:

*CFME QE Bot*

cfme/cloud/provider.py
```diff
@@ -48,6 +48,7 @@
         ('ipaddress_text', Input("ipaddress"), {"removed_since": "5.4.0.0.15"}),
         ('region_select', {version.LOWEST: Select("select#provider_region"),
             "5.5": AngularSelect("provider_region")}),
+        ('amazon_region_select', AngularSelect("provider_region")),
```
mkoura: @psav I'm not sure if 'region_select' is still needed for 5.5, I just noticed that the 'amazon_region_select' is needed and is missing. Same goes for `properties_form_56`, there are both 'region_select' and 'amazon_region_select' with same value.

psav: Ahh I see what's happened.....personally I like to try this
* Remove both ```amazon_region_select``` and ```azure_region_select``` and replace them with just a generic ```region_select``` this will mean fixing up the provider forms, but there is no point having the same field defined twice.

cfme/cloud/provider.py
```diff
@@ -75,8 +76,7 @@
         ('google_project_text', Input("project")),
         ('azure_tenant_id', Input("azure_tenant_id")),
         ('azure_subscription_id', Input("subscription")),
-        ('amazon_region_select', {version.LOWEST: Select("select#provider_region"),
-            "5.5": AngularSelect("provider_region")}),
+        ('amazon_region_select', AngularSelect("ems_region")),
```
psav: As this is called ```ems_region``` implying that it doesn't matter which provider uses it, we should probably rename this one too. Those strings are easily greppable in the codebase and I'd have no worry merging those changes.

dajoRH:
Lint report for commit 425a7a7bfb08efa9576fe90459e739ba1d624f07:

*CFME QE Bot*

PR #3125, opened 2016-06-27
Yaml cleanup

psav:

PR #3134, opened 2016-06-29
Fixed group data call

psav:
In a recent PR #3125, the group_data object changed from a callable to a
dict. This is now reflected in testgen where it was used.

{{pytest: cfme/tests/integration/test_aws_iam_auth_and_roles.py --collect-only}}


dajoRH:
Lint report for commit 0f280b37a57806fc3b36ebcc0777e16cb37d741b:

*CFME QE Bot*

PR #3137, opened 2016-06-29
Added new list_tbl definitions to container modules

jdemon519:
Deleted list_tbl definition from cfme/containers/__init__.py
Added new list_tbl definitions to each module in cfme/containers to prevent caching issues
Modified test_relationships_tables.py by adding new list_tbl definition to each function


dajoRH:
Lint report for commit 872f8f90e50cf4abef316622904fb39cc6ae6560:

*CFME QE Bot*

mfalesni:
@jdemon Actually, #2984

psav:
Do we need this one if we have #3138? Which includes all these commits? If we do and one ie branched from the other please include that in the description. It makes things a lot easier. Having said that, we still need this one cleaned up.....ie squased.

PR #3141, opened 2016-06-30
Fix role setting

psav:
* Role setting is now done by SCP by default (could be riskier)
* Fixed another bug, where an SSH reconnect was being caused by the
  set_yaml_config function for <= 5.5

### DIFF
--- a/utils/appliance.py
+++ b/utils/appliance.py
@@ -1551,7 +1551,7 @@ class IPAppliance(object):
             else:
                 raise Exception('Only [vmdb] config is allowed from 5.6+')
         else:
-            return db.set_yaml_config(config_name, data_dict, self.address)
+            return db.set_yaml_config(config_name, data_dict)
 
     def get_yaml_file(self, yaml_path):
         """Get (and parse) a yaml file from the appliance, returning a python data structure"""
--- a/cfme/configure/configuration/__init__.py
+++ b/cfme/configure/configuration/__init__.py
@@ -1808,17 +1808,22 @@ class Tag(Pretty):
             sel.handle_alert()
 
 
-def set_server_roles(**roles):
+def set_server_roles(db=True, **roles):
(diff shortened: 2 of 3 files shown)


utils/appliance.py
```diff
@@ -1551,7 +1551,7 @@ def set_yaml_config(self, config_name, data_dict):
             else:
                 raise Exception('Only [vmdb] config is allowed from 5.6+')
         else:
-            return db.set_yaml_config(config_name, data_dict, self.address)
```
psav: This was causing the
```
workspace/default/lib/python2.7/site-packages/paramiko/client.pyc in _auth(self, username, password, pkey, key_filenames, allow_agent, look_for_keys, gss_auth, gss_kex, gss_deleg_creds, gss_host)
    602         if saved_exception is not None:
    603             raise saved_exception
--> 604         raise SSHException('No authentication methods available')
    605 
    606     def _log(self, level, msg):

SSHException: No authentication methods available
```

dajoRH:
Lint report for commit 1e845053898ddc66fb0c8248222b803109ad5e68:

*CFME QE Bot*

psav:
Please test thoroughly before merging, both 5.5 and 5.6 (I tested 5.5 and 5.6 should work)

dajoRH:
Lint report for commit 8247b1a42a4322de2814e39794259efba1296ebe:

*CFME QE Bot*

jkrocil:
Tested 5.5. and 5.6, looks good. Main reason for this change, as I understand it, being this 5.5 BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1351716
^ we were unable to change server roles through UI and needed an option to do it through yaml

psav:
Yes that BZ triggered this change, it also makes sense and is another step towards "fast" fixtures

PR #3168, opened 2016-07-11
Fix navigation for upstream UI changes.

amavinag:
* settings panel moved to top right in the upstream.
    added quick fix to unblock upstream smoke tests.

### DIFF
--- a/cfme/configure/configuration/__init__.py
+++ b/cfme/configure/configuration/__init__.py
@@ -5,6 +5,7 @@ from cached_property import cached_property
 import cfme.fixtures.pytest_selenium as sel
 from fixtures.pytest_store import store
 
+from cfme import dashboard
 import cfme.web_ui.tabstrip as tabs
 import cfme.web_ui.toolbar as tb
 from cfme.exceptions import ScheduleNotFound, AuthModeUnknown, ZoneNotFound, CandidateNotFound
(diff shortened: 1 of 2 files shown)


dajoRH:
Lint report for commit 67931a41bc68ddee8d56aae468815839cef70344:

*CFME QE Bot*

cfme/web_ui/menu.py
```diff
@@ -439,16 +439,19 @@ def f(_):
 
         def f2(_):
             args = [toplevel, secondlevel, thirdlevel]
-            loc = cls.ROOT
-            for arg in args:
-                if arg:
-                    if arg.startswith("/"):
-                        loc_to_use = cls.NAMED_LEV_HREF
+            if version.current_version() == 'master' and toplevel == 'Settings':
```
psav: I think I'd actually prefer to handle this outside of "menu". This is not related to the main "menu" anymore.....I try to knock something up.

jkrocil: Although this is good as a quickfix, I agree with Pete - this should be handled elsewhere, not in `cfme.web_ui.menu`.

amavinag: sure @psav @jkrocil I'm looking at dashboard and other pages to move this out of menu.py

psav:
I worked up #3174 to try and help simplify things a little, I think it is preferred

PR #3170, opened 2016-07-12
[FIX]  Found and added missing Azure Subscription ID reference

jteehan:
This is leftover from 2977.  It's a single line fix and works locally.  Nothing I can do about the location of these values until credentials allows for additional entries.

{{pytest: cfme/tests/cloud -k "test_provider_crud" --use-provider azure }}

### DIFF
--- a/cfme/cloud/provider.py
+++ b/cfme/cloud/provider.py
@@ -53,7 +53,6 @@ properties_form_55 = Form(
                 "5.5": "api_port",
             }
         )),
-        ('azure_subscription_id', Input("subscription")),
         ('api_version', AngularSelect("api_version"), {"appeared_in": "5.5"}),
         ('sec_protocol', AngularSelect("security_protocol"), {"appeared_in": "5.5"}),
         ('infra_provider', {
@@ -174,7 +173,7 @@ class AzureProvider(Provider):
         self.subscription_id = subscription_id
 
     def _form_mapping(self, create=None, **kwargs):
-        # Will still need to figure out where to put the tenant id.
+        # Tenant ID and Subscription ID will have to live in cfme_data for now.
         return {'name_text': kwargs.get('name'),
                 'type_select': create and 'Azure',
                 'region_select': kwargs.get('region'),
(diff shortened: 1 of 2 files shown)


dajoRH:
Lint report for commit 5b41f253859b8bd1e39e1a5a3012b6a448912a70:

*CFME QE Bot*

jteehan:
Downstream 5.6 is the best we're going to get for Azure.  This feature doesn't exist in 5.5

dajoRH:
@jteehan , per your comment, it would still need to work for upstream and Azure is supported for 55z as well so we would really want/need to see this passing on all three.  If something truly wasn't supported, we would uncollect it which would result in a pass for PRT and prevent a failed test from getting merged.  

With all that said, the failures are not related to your code, I just clicked retest for a second round to see if its the same result.

INFO:cfme:Running command `systemctl restart evmserverd`
INFO:cfme:Invoking callback for signal [server_details_changed]
INFO:cfme:Waiting for Web UI to start
ERROR:cfme:Couldn't complete appliance.is_web_ui_running at /cfme_tests_te/utils/appliance.py:1099 in time, took 319.68, 11 tries
ERROR:cfme:The last result of the call was: False
EERROR:cfme:@pytest.fixture
    def has_no_cloud_providers():

dajoRH:
Ok, I circled back to this and the upstream job hit the same error as before which smells like a separate issue, maybe @psav, @mfalesni , @jkrocil , have seen it?  Ultimately it probably will need a github issue.

The downstream 55z job did fail on provider.create()  ,  its trying to fill out the form which is missing the subscription id field.  I think we need a version pick on the cfme/cloud/provider.py:AzureProvider._form_mapping() returns based on the version.  What confuses me is its listed in cfme/cloud/provider.py:properties_form_55 dict , so was it there and was now removed @jteehan?  Is it a bug that its missing?

dajoRH:
Lint report for commit 15fcb8ba1b61cd82c50838ee42d8ca89d9ae94b7:

*CFME QE Bot*

jteehan:
Subscription ID should not be in the 5.5 properties_form_55. At the time, I was under the impression it was going to be back ported.  Removed that.  Upstream probably requires and issue to be created.  Not sure why it doesn't show but I looked and it's definitely not there.

jteehan:
Cool.  Now we have both downstreams passing.  Even better.

jkrocil:
The upstream issue is known - we need a new upstream template on sprout (WIP).
55z and 56z look good, Im afraid I have to merge this one.
:+1:

PR #3172, opened 2016-07-13
Fix testgen mark application

### LINKED ISSUE #3171: -m provider_related no longer works
We added the internally generated mark provider_related to be used in conjunction with --use-provider <provider> so we could only run tests that are specific to the provider.  Without it, the test collected would be anything non-parameterized through testgen, like general ui tests, along with the provider paramaterized tests specific to the provider that pass through testgen.  

Example (only run tests specific to rhevm36 provider integration):
py.test cfme/tests/infrastructure --use-provider rhevm36 -m provider_related

This appears to have broken at some point and need to be fixed.

A partial workaround is to pass "-m uses_testgen" instead.  It still pulls in some extra tests but limits it to test suites that use testgen.  For example, test_providers.py would run the specific crud tests but also run the provider validation tests which isn't ideal but far less than running without (when you only want to run the specific provider integration tests).  This all needs documentation in #3115

psav:
Previously the mark was applied correctly, however in the testgen
refactor this became broken. As the mark was applied outside of the
loop, merely running "through" testgen was enough to mark the
test. Unfortunately this was not accurate. This mark has now been moved
to be applied only if the "provider" parameter has actually been used
and hence the test has been properly parametrized.

Fixes #3171 

{{pytest: -m uses_testgen --collect-only --use-provider vsphere55}}

### DIFF
--- a/utils/testgen.py
+++ b/utils/testgen.py
@@ -331,8 +331,6 @@ def provider_by_type(metafunc, provider_types, required_fields=None):
 
     """
 
-    metafunc.function = pytest.mark.uses_testgen()(metafunc.function)
-
     argnames = []
     argvalues = []
     idlist = []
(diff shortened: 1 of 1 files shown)


psav:
This fixes #3171

PR #3174, opened 2016-07-14
Fixed nav upstream

psav:
Upstream nav needed some help, this work taks some of the ideas in
amavinag's branch and implements them slightly differently.


psav:
This failed on upstream because the upstream templates are too old for the patching to work and we can't get the template tester to get a new one because this fix is required.

mfalesni:
Ninja merge.

PR #3181, opened 2016-07-17
Test updated views 2

jdemon519:
Modified test_containers_properties.py, added test_provider_crud.py which used to be in cfme/tests/containers previously

### DIFF
--- a/cfme/tests/containers/test_containers_properties.py
+++ b/cfme/tests/containers/test_containers_properties.py
@@ -1,16 +1,13 @@
 # -*- coding: utf-8 -*-
 # the test performs field validation in Properties table
-# Polarion test 9911
-# Polarion test 9877
-# Polarion test 9867
 import pytest
 from cfme.fixtures import pytest_selenium as sel
-from cfme.containers.pod import Pod
-from cfme.containers.route import Route
-from cfme.containers.project import Project
+from cfme.containers.pod import Pod, list_tbl as list_tbl_pod
+from cfme.containers.route import Route, list_tbl as list_tbl_route
+from cfme.containers.project import Project, list_tbl as list_tbl_proj
 from utils import testgen
 from utils.version import current_version
-from cfme.web_ui import CheckboxTable
+
 
 pytestmark = [
     pytest.mark.uncollectif(
@@ -20,22 +17,23 @@ pytestmark = [
 pytest_generate_tests = testgen.generate(
     testgen.container_providers, scope="function")
 
-
+# Polarion test 9911
 # container pods Properties table field verification
+
+
 @pytest.mark.parametrize('rel',
-                         ['Name',
-                          'Phase',
-                          'Creation timestamp',
-                          'Resource version',
-                          'Restart policy',
-                          'DNS Policy',
-                          'IP Address'])
+                         ['name',
+                          'phase',
+                          'creation_timestamp',
+                          'resource_version',
+                          'restart_policy',
+                          'dns_policy',
+                          'ip_address'
+                          ])
 def test_pods_properties_rel(provider, rel):
     sel.force_navigate('containers_pods')
-    list_tbl_pod = CheckboxTable(table_locator="//div[@id='list_grid']//table")
     ui_pods = [r.name.text for r in list_tbl_pod.rows()]
     mgmt_objs = provider.mgmt.list_container_group()  # run only if table is not empty
-    validate_str = False
 
     if ui_pods:
         # verify that mgmt pods exist in ui listed pods
@@ -44,32 +42,23 @@ def test_pods_properties_rel(provider, rel):
 
     for name in ui_pods:
         obj = Pod(name, provider)
-        val = obj.get_detail('Properties', rel)
+        field_content = getattr(obj.summary.properties, rel).text_value
 
-        # the field should not be empty
-        try:
-            str(val)
-            validate_str = True
-        except ValueError:
-            pass
-
-        if validate_str:
-            assert len(val) != 0
+        if field_content:
+            assert len(field_content) != 0
 
 
-# container routes Properties table field verification
+# Polarion test 9877
 @pytest.mark.parametrize('rel',
-                         ['Name',
-                          'Creation timestamp',
-                          'Resource version',
-                          'Host Name'])
+                         ['name',
+                          'creation_timestamp',
+                          'resource_version',
+                          'host_name'
+                          ])
 def test_routes_properties_rel(provider, rel):
     sel.force_navigate('containers_routes')
-    list_tbl_route = CheckboxTable(
-        table_locator="//div[@id='list_grid']//table")
     ui_routes = [r.name.text for r in list_tbl_route.rows()]
     mgmt_objs = provider.mgmt.list_route()  # run only if table is not empty
-    validate_str = False
 
     if ui_routes:
         # verify that mgmt routes exist in ui listed routes
@@ -78,31 +67,22 @@ def test_routes_properties_rel(provider, rel):
 
     for name in ui_routes:
         obj = Route(name, provider)
-        val = str(obj.get_detail('Properties', rel))
+        field_content = getattr(obj.summary.properties, rel).text_value
 
-        # the field should not be empty
-        try:
-            str(val)
-            validate_str = True
-        except ValueError:
-            pass
-
-        if validate_str:
-            assert len(val) != 0
+        if field_content:
+            assert len(field_content) != 0
 
 
-# container projects Properties table validation
+# Polarion test 9867
 @pytest.mark.parametrize('rel',
-                         ['Name',
-                          'Creation timestamp',
-                          'Resource version'])
+                         ['name',
+                          'creation_timestamp',
+                          'resource_version'
+                          ])
 def test_projects_properties_rel(provider, rel):
     sel.force_navigate('containers_projects')
-    list_tbl_project = CheckboxTable(
-        table_locator="//div[@id='list_grid']//table")
-    ui_projects = [r.name.text for r in list_tbl_project.rows()]
+    ui_projects = [r.name.text for r in list_tbl_proj.rows()]
     mgmt_objs = provider.mgmt.list_project()  # run only if table is not empty
-    validate_str = False
 
     if ui_projects:
         # verify that mgmt projects exist in ui listed projects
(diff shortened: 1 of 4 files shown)


dajoRH:
Lint report for commit b6fe4013856106ddaa56d8fb0f956b9bbc8aac45:

*CFME QE Bot*

cfme/tests/containers/test_containers_properties.py
```diff
@@ -0,0 +1,124 @@
+# -*- coding: utf-8 -*-
+# the test performs field validation in Properties table
+# Polarion test 9911
+# Polarion test 9877
+# Polarion test 9867
```
nellyc: please put the ID comment on the relevant test - it will make life easier later

dajoRH:
Lint report for commit 82f8c115de5622c67b79e3e47adb930dcd6e4448:

*CFME QE Bot*

dajoRH:
Lint report for commit 4086645b63bb9ef2bf9506537320699771c08d22:

*CFME QE Bot*

cfme/tests/containers/test_containers_properties.py
```diff
@@ -0,0 +1,124 @@
...
+    validate_str = False
+
+    if ui_pods:
+        # verify that mgmt pods exist in ui listed pods
+        assert set(ui_pods).issubset(
+            [obj.name for obj in mgmt_objs]), 'Missing objects'
+
+    for name in ui_pods:
+        obj = Pod(name, provider)
+        val = obj.get_detail('Properties', rel)
```
mfalesni: Check if Pod has SummaryMixin. If not, you can add it to it as long as Pod implements load_details()

Then you could write something like getattr(obj.summary.provider, rel).text_value (or .value if you want to have it converted to number). It should handle it gracefully.

dajoRH:
Lint report for commit 99536075d0546a42a67561de8a14cb52d0dd44c9:

*CFME QE Bot*

dajoRH:
Lint report for commit 1dc49c983ed5785f19e2a8e4ae9987d00df72dc5:

*CFME QE Bot*

dajoRH:
Lint report for commit 7780cd32a800655d39e26095b13981b1ef18ec9f:

*CFME QE Bot*

cfme/tests/containers/test_one_image_analysis.py
```diff
@@ -0,0 +1,63 @@
...
+        lambda: current_version() < "5.6"),
+    pytest.mark.usefixtures('setup_provider'),
+    pytest.mark.tier(1)]
+pytest_generate_tests = testgen.generate(
+    testgen.container_providers, scope="function")
+
+
+def test_image_analysis_finished():
+    sel.force_navigate('containers_images')
+    checkbox = ".//input[@class='list-grid-checkbox']"
```
mfalesni: Can you move this to model? It is not very good having locators in tests, unless they are testing some specific UI glitch or similar thing.

cfme/tests/containers/test_one_image_analysis.py
```diff
@@ -0,0 +1,63 @@
...
+    sel.check(checkbox)
+    tb.select(
+        'Configuration',
+        'Perform SmartState Analysis',
+        invokes_alert=True)
+    sel.handle_alert()
+    flash.assert_message_contain('Analysis successfully initiated')
+    success_alert = ".//div[@class='alert alert-success']"
```
mfalesni: Why this one? Is our flash handling deficient? If so, I can address that.

cfme/tests/containers/test_one_image_analysis.py
```diff
@@ -0,0 +1,63 @@
...
+
+    if not success_alert:
+        wait_for(lambda: sel.is_displayed(success_alert), num_sec=10, delay=5)
+
+    is_task_finished()
+
+
+def is_task_finished():
```
mfalesni: I guess you could use the `all_vm_analysis_tasks` function inside to look up the task. It is located in the cfme/configure/tasks.py, line 156. It returns list of dicts, you can try that out in ipython.

cfme/tests/containers/test_containers_services.py
```diff
@@ -38,14 +39,7 @@ def test_service_summary_rel(provider, rel):
...
-            str(val)
-            validate_str = True
-        except ValueError:
-            pass
+        field_content = getattr(obj.summary.properties, rel).text_value
 
-        if validate_str:
-            assert len(val) != 0
+        if field_content:
+            assert len(field_content) != 0
```
jaryn: The code will never fail on this assert.

    assert getattr(obj.summary.properties, rel).text_value

is enough

cfme/tests/containers/test_containers_properties.py
```diff
@@ -0,0 +1,97 @@
...
+                          ])
+def test_pods_properties_rel(provider, rel):
+    sel.force_navigate('containers_pods')
+    ui_pods = [r.name.text for r in list_tbl_pod.rows()]
+    mgmt_objs = provider.mgmt.list_container_group()  # run only if table is not empty
+
+    if ui_pods:
+        # verify that mgmt pods exist in ui listed pods
+        assert set(ui_pods).issubset(
+            [obj.name for obj in mgmt_objs]), 'Missing objects'
```
jaryn: It is just a small think, but... you don't need to construct a list here. Please use just

    assert set(ui_pods).issubset(
        obj.name for obj in mgmt_objs), 'Missing objects'

cfme/tests/containers/test_containers_properties.py
```diff
@@ -0,0 +1,97 @@
...
+
+    for name in ui_pods:
+        obj = Pod(name, provider)
+        field_content = getattr(obj.summary.properties, rel).text_value
```
jaryn: Why not to check all the fileds per one page navigation?

    pod = Pod(name, provider)
    assert pod.summary.properties.name.text_value
    assert pod.summary.properties.phase.text_value
    assert pod.summary.properties.creation_timestamp.text_value
    ...

The benefit will be that the tests will finish much faster!

cfme/tests/containers/test_containers_properties.py
```diff
@@ -0,0 +1,97 @@
...
+        # verify that mgmt projects exist in ui listed projects
+        assert set(ui_projects).issubset(
+            [obj.name for obj in mgmt_objs]), 'Missing objects'
+
+    for name in ui_projects:
+        obj = Project(name, provider)
+        field_content = getattr(obj.summary.properties, rel).text_value
+
+        if field_content:
+            assert len(field_content) != 0
```
jaryn: Same as on other places. This assert will never fail as we will not reach line 97 because if `len(field_content) == 0`, the `if` will skip this line.

Remember that `if EXPRESSION` is doing `bool(EXPRESSION)`. `bool([])` evals to `False`

cfme/tests/containers/test_containers_properties.py
```diff
@@ -0,0 +1,97 @@
...
+                          ])
+def test_routes_properties_rel(provider, rel):
+    sel.force_navigate('containers_routes')
+    ui_routes = [r.name.text for r in list_tbl_route.rows()]
+    mgmt_objs = provider.mgmt.list_route()  # run only if table is not empty
+
+    if ui_routes:
+        # verify that mgmt routes exist in ui listed routes
+        assert set(ui_routes).issubset(
+            [obj.name for obj in mgmt_objs]), 'Missing objects'
```
jaryn: Same as on the other place. There is no need to create a array here.

cfme/tests/containers/test_containers_properties.py
```diff
@@ -0,0 +1,97 @@
...
+# Polarion test 9867
+@pytest.mark.parametrize('rel',
+                         ['name',
+                          'creation_timestamp',
+                          'resource_version'
+                          ])
+def test_projects_properties_rel(provider, rel):
+    sel.force_navigate('containers_projects')
+    ui_projects = [r.name.text for r in list_tbl_proj.rows()]
+    mgmt_objs = provider.mgmt.list_project()  # run only if table is not empty
```
jaryn: I do not get the meaning of the comment. Not empty what? The `mgmt_objs`? The test will continue even thouh the `mgmt_objs` is empty.

cfme/tests/containers/test_containers_services.py
```diff
@@ -0,0 +1,45 @@
...
+        # verify that mgmt services exist in ui listed services
+        assert set(ui_services).issubset(
+            [obj.name for obj in mgmt_objs]), 'Missing objects'
+
+    for name in ui_services:
+        obj = Service(name, provider)
+        field_content = getattr(obj.summary.properties, rel).text_value
+
+        if field_content:
+            assert len(field_content) != 0
```
jaryn: Same as above.

cfme/tests/containers/test_one_image_analysis.py
```diff
@@ -0,0 +1,63 @@
...
+        'task_name': task_name,
+        'state': expected_status
+    })
+
+    if not el:
+        try:
+            wait_for(is_task_finished, delay=20, timeout="5m",
+                     fail_func=lambda: tb.select('Reload'))
+        except TimedOutError:
+            pytest.fail("Analysis has not completed")
```
jaryn: The `pytest.fail` will raise an exception, so we won't do the cleanup. I am not sure about this but I think we should do the cleanup. What do you think @jkrocil @psav @mfalesni ?

cfme/tests/containers/test_containers_properties.py
```diff
@@ -0,0 +1,97 @@
...
+        # verify that mgmt routes exist in ui listed routes
+        assert set(ui_routes).issubset(
+            [obj.name for obj in mgmt_objs]), 'Missing objects'
+
+    for name in ui_routes:
+        obj = Route(name, provider)
+        field_content = getattr(obj.summary.properties, rel).text_value
+
+        if field_content:
+            assert len(field_content) != 0
```
jaryn: Same as above. The `assert` will never fail and I think we could optimize the tests running time by checking all the fields per one page navigation.

cfme/tests/containers/test_containers_properties.py
mfalesni: In that case, `soft_assert` is better for checking multiple values and failing taht at once at the end of the test, but I agree with the thought.

cfme/tests/containers/test_one_image_analysis.py
mfalesni: Add any cleanup code using `request.addfinalizer` before the wait_for starts (or do the cleanup after `yield` in a yield fixture. If possible, of course.

WRT the timeout, if there is a variable that affects the length of the task, then definitely that deserves variable timeout based on the variable. An example is infra provider discovery, which does I think `N * 30` where `N` is number of IP addresses scanned, but at least 300 seconds.

PR #3200, opened 2016-07-21
Add pytest.mark.polarion()

mfalesni:
This is for marking unparametrized test cases so far. This will inject the "polarion" key in the JSON that gets sent into Ostříž. If no marker is provided, it is None, otherwise it is a list of Polarion test cases associated with the test.

@psav - your opinion?


dajoRH:
Lint report for commit e059a0be67359ea13dc4e3f17b18758f846bde0b:

*CFME QE Bot*

psav:
Can you mark a test so it can be....tested?

PR #3203, opened 2016-07-21
Added new functionality to test Deploy/Undeploy/Redeploy/Stop/Start archive into Middleware server

hhovsepy:
Added new archive files as middleware resources to be used in middleware provider tests.
Refactored existing deployments tests to reuse common methods.
Fixed possibly failing tests.


dajoRH:
Lint report for commit 13e80bc6619d82627bfb827114a07cc2a8a82d75:

*CFME QE Bot*

dajoRH:
Lint report for commit a2b0b146fcc56c0ff45524933f27df04e1b864bb:

*CFME QE Bot*

dajoRH:
Lint report for commit d428f3e0427b1c372590ae463f1944eaf923bcac:

*CFME QE Bot*

cfme/tests/middleware/test_middleware_deployment.py
```diff
@@ -13,6 +16,10 @@
 pytest_generate_tests = testgen.generate(testgen.provider_by_type, ["hawkular"], scope="function")
 ITEMS_LIMIT = 5  # when we have big list, limit number of items to test
 
+RESOURCE_WAR_PATH = 'data/resources/middleware/cfme_test_war_middleware.war'
+RESOURCE_JAR_PATH = 'data/resources/middleware/cfme_test_jar_middleware.jar'
+RESOURCE_EAR_PATH = 'data/resources/middleware/cfme_test_ear_middleware.ear'
```
mfalesni: I guess you want to use objects from `utils.path` to generate the paths, that will be more reliable.

dajoRH:
Lint report for commit 055a0ee0d307f1bb6b7cb4cdd60c468c6cf72041:

*CFME QE Bot*

dajoRH:
Lint report for commit f0798dd44ce88d1d0e79277befd63ff4a2cf5385:

*CFME QE Bot*

dajoRH:
Lint report for commit f7adddccb06e31fcfeaa7057eedf97378a5ed2ad:

*CFME QE Bot*

cfme/middleware/__init__.py
```diff
@@ -3,14 +3,18 @@
 
 from cfme.common import Validatable
 from cfme.fixtures import pytest_selenium as sel
-from cfme.web_ui import toolbar as tb
+from cfme.web_ui import toolbar as tb, Form, fill, flash
+from cfme.web_ui.form_buttons import FormButton
 from utils.browser import ensure_browser_open
+import os
```
psav: This import should go on line 3 with the other builtins

psav:
50% pass rate on upstream...

dajoRH:
Lint report for commit 634cd77118bb3a63f7268fd6e8868e95d57dc0be:

*CFME QE Bot*

dajoRH:
Lint report for commit 0ce5e01d6663f77d70cc68516e738d5ab03b8b65:

`cfme/tests/middleware/test_middleware_deployment.py`:
- :bangbang: Line 216:1: **E302** *expected 2 blank lines, found 1*

*CFME QE Bot*

dajoRH:
Lint report for commit af932534cc1ed0b2a9180922144ffafef0f4fdc8:

*CFME QE Bot*

dajoRH:
Lint report for commit 19a1d72b8f60476bfa80f3aa4e7bdd7a46a7aaa8:

*CFME QE Bot*

cfme/middleware/__init__.py
```diff
@@ -27,6 +31,17 @@ def _on_detail_page(self):
         return sel.is_displayed('//h1[contains(., "{} (Summary)")]'.format(self.name))
 
 
+import_form = Form(
+    fields=[
+        ("file_select", "#upload_file"),
```
mfalesni: Nope nope nope. You want to use `FileInput` from `cfme.web_ui`

cfme/middleware/__init__.py
```diff
@@ -27,6 +31,17 @@ def _on_detail_page(self):
...
+        ("enable_deployment", "#enable_deployment_cb"),
+        ("runtime_name", "#runtime_name_input"),
```
mfalesni: Put these two as `Input`

cfme/middleware/__init__.py
mfalesni: FYI, FileInput nad Input classes take only the "name" of the field, so in this case, `FileInput('upload_file')` ...

cfme/middleware/__init__.py
```diff
@@ -53,3 +68,63 @@ def download(extension):
         download_btn("Download as {}".format(extensions_mapping[extension]))
     except:
         raise ValueError("Unknown extention. check the extentions_mapping")
+
+
+def add_deployment(filename, runtime_name=None, enable_deploy=True, cancel=False):
```
mfalesni: Wouldn't it make more sense to put this as a class method on a `Deployment` class? Then it could navigate to correct place, add the deployment and return an instance of `Deployment` class that will be further usable.

mfalesni: Same with the underlying functions, they seems to be fitting as methods for Deployment, but I can just be wrong, I am not that familiar with middleware UI.

dajoRH:
Lint report for commit 552442c5b653b8485e37b4ef04982367fb4cd6a6:

*CFME QE Bot*

cfme/tests/middleware/test_middleware_deployment.py
```diff
@@ -132,3 +133,228 @@ def test_tags(provider):
...
+        * Deplays some deployment archive into server.
+        * Select that deployment from deployments list.
+        * Performs "Undeploy" toolbar operation on it.
+        * Refreshes the provider.
+        * Lists all deployments on EAP server.
+        * Verified the recently undeployed archive's status is Disabled.
+        * Selects that archive from the list to load details.
+        * Verifies that properties of deplyment's summary page and the status is Disabled.
+    """
+    server = _get_eap_server(provider)
```
mfalesni: I think that tests should not use underscored functions/methods/things ...

cfme/middleware/__init__.py
hhovsepy: Inspired  from import_export.py

cfme/middleware/__init__.py
hhovsepy: Here it is core method for deployment adding, it is called from necessary pages.

cfme/tests/middleware/test_middleware_deployment.py
hhovsepy: workaround?
I need private methods there

cfme/middleware/__init__.py
mfalesni: Please use the FileInput, it is designed to control one of the "nastier" selenium background magic features ...

cfme/middleware/__init__.py
mfalesni: Deployments can be added from different objects? If so, wouldn't a mixin be a better option?

cfme/tests/middleware/test_middleware_deployment.py
mfalesni: Tests should be designed so they use the "public" interfaces of objects.

*I say "public" in quotes because everything is public in python :smile: *

hhovsepy: ok, just those functions without underscore

cfme/middleware/__init__.py
mfalesni: Oh, and the enable_deployment is actually a checkbox, that might stay is it is for now, seems we don't have a class for that yet.

cfme/middleware/__init__.py
hhovsepy: I think I wrongly explained.
Deployments can be done from different pages of Middleware provider, now from Summary, list of deployments, ..., this is toolbar button, I do not want to limit it in Deployment's class.
But I think it is Middleware specific and no need to have in mixins.

mfalesni: You can have a middleware-specific mixin that will live in the cfme/middleware/__init__.py

hhovsepy: I thought about it, but it can make mess because we will need 2 mixins.
1 Mixin can be extended by "Deployment", and it will contain deployment start/stop... already deployed objects operations
"add_deployment" method does not belong to "Deployment" and for that we will need second mixin for Server class.

Solution I gave was like cfme.web_ui.mixins but living in cfme/middleware/__init__.py

dajoRH:
Lint report for commit ef18c9bad296a7bd3e62c23dd0fcb65a44c5d88e:

*CFME QE Bot*

cfme/middleware/__init__.py
mfalesni: cfme.web_ui.mixins' name is picked wrong, mixin should be a class :)

So, if I understand it correctly, the add_deployment is specific for Server, is that right? Then why not make a mixin class without add_deployment and create a separate mixin class for the add_deployment specifically. Then the Server would inherit both ...

How does that sound?

hhovsepy: If it is not causing mess, as I said, it's ok for me.
Server extends new "Container" mixin with "add_deploymet", Deployment extends new "Deployable" mixin with rest of methods.

dajoRH:
Lint report for commit 98acccb0b0d61ddb11c3b661535eb8bceb762a01:

*CFME QE Bot*

dajoRH:
Lint report for commit 26b83cb32d7c9b5230a174126a3edf8af07aef5c:

*CFME QE Bot*

dajoRH:
Lint report for commit d49eee7e74672885fda1caad1455b2143c74c0c9:

*CFME QE Bot*

cfme/tests/middleware/test_middleware_deployment.py
```diff
@@ -132,3 +133,228 @@ def test_tags(provider):
...
+    runtime_name = deploy(provider, server, archive)
+    check_deployment_enabled(provider, server, runtime_name)
+    deployment = get_deployment_from_list(provider, server, runtime_name)
+    deployment.stop()
+    check_deployment_disabled(provider, server, runtime_name)
+    deployment = get_deployment_from_list(provider, server, runtime_name)
+    deployment.validate_properties()
+
+
+# disable for now as not stable test because of performance issue #10138
```
mfalesni: If this is a github issue, you can do:
```python
@pytest.mark.meta(blockers=['GH#owner/repo:10138'])
```

to skip the test with a message. GitHub skipping only works on upstream so far but you do run on upstream :)

dajoRH:
Lint report for commit 98e50f1b0d1609ac038b4b072bf2db21568b2133:

*CFME QE Bot*

cfme/middleware/__init__.py
```diff
@@ -27,6 +31,17 @@ def _on_detail_page(self):
         return sel.is_displayed('//h1[contains(., "{} (Summary)")]'.format(self.name))
 
 
+import_form = Form(
+    fields=[
+        ("file_select", FileInput("upload[file]")),
+        ("enable_deployment", "#enable_deployment_cb"),
```
mfalesni:
```python
from cfme.web_ui import CFMECheckbox

('enable_deployment', CFMECheckbox('enable_deployment_cb'))
```

cfme/middleware/__init__.py
```diff
@@ -53,3 +68,69 @@ def download(extension):
...
     except:
         raise ValueError("Unknown extention. check the extentions_mapping")
+
+
+class Container(SummaryMixin):
+
+    def add_deployment(self, filename, runtime_name=None, enable_deploy=True, cancel=False):
+        """Clicks to "Add Deployment" button, in opened window fills fields by provided parameters,
+        and deploys.
+        Args:
```
mfalesni: Newline before args please

cfme/tests/middleware/test_middleware_deployment.py
```diff
@@ -132,3 +149,228 @@ def test_tags(provider):
...
+    server = get_eap_server(provider)
+    deploy(provider, server, archive)
+    # disable for now as not stable test because of performance issue #10138
+    # _check_deployment_enabled(provider, server, runtime_name)
+    # deployment = _get_deployment_from_list(provider, server, runtime_name)
+    # deployment.validate_properties()
+
+
+@pytest.mark.meta(blockers=['GH#ManageIQ/manageiq:10138'])
+@pytest.mark.uncollect
```
mfalesni: why uncollect when the meta blockers skip this when that issue is still opened?

cfme/tests/middleware/test_middleware_deployment.py
```diff
@@ -132,3 +149,228 @@ def test_tags(provider):
...
+@pytest.mark.parametrize("archive", [RESOURCE_WAR_PATH, RESOURCE_JAR_PATH, RESOURCE_EAR_PATH])
```
mfalesni: parametrizing by full paths seems not a good option since if it gets run on a slightly different setup, the params change even though the tests is the same. I would rather recommend passing the filenames here and joining the path inside the test, that leads to more stable IDs.

cfme/tests/middleware/test_middleware_deployment.py
```diff
@@ -132,3 +149,228 @@ def test_tags(provider):
...
+    deployment = get_deployment_from_list(provider, server, runtime_name)
+    deployment.undeploy()
+    check_deployment_disabled(provider, server, runtime_name)
+    deployment = get_deployment_from_list(provider, server, runtime_name)
+    deployment.validate_properties()
+
+
+@pytest.mark.meta(blockers=['GH#ManageIQ/manageiq:10138'])
+@pytest.mark.uncollect
+@pytest.mark.parametrize("archive", [RESOURCE_WAR_PATH, RESOURCE_JAR_PATH, RESOURCE_EAR_PATH])
```
mfalesni: All these the same questions as above.

cfme/tests/middleware/test_middleware_deployment.py
hhovsepy: meta blocker was not skipping it

cfme/tests/middleware/test_middleware_deployment.py
```diff
@@ -132,3 +149,228 @@ def test_tags(provider):
...
+    server.add_deployment(file_path, runtime_name)
+    provider.refresh_provider_relationships(method='ui')
+    return runtime_name
+
+
+def get_eap_server(provider):
+    server = None
+    server_list = MiddlewareServer.servers(provider=provider)
+    for _server in server_list:
+        if _server.product == 'JBoss EAP':
```
mfalesni: Is this something that we can rely on being reasonably static? Same with the hawkular down there.

cfme/tests/middleware/test_middleware_deployment.py
```diff
@@ -132,3 +149,228 @@ def test_tags(provider):
...
+
+def get_local_server(provider):
+    server = None
+    server_list = MiddlewareServer.servers(provider=provider)
+    for _server in server_list:
+        if _server.product == 'Hawkular':
+            server = _server
+            break
+    assert server, 'Hawkular server was not found in servers list'
+    return server
```
mfalesni: These two functions can be DRYed:
```python
def get_server(provider, product):
    for server in MiddlewareServer.servers(provider=provider):
        if server.product == product:
            return server
    else:
        raise ValueError('{} server was not found in servers list'.format(provider))
```

cfme/tests/middleware/test_middleware_deployment.py
```diff
@@ -132,3 +149,228 @@ def test_tags(provider):
...
+
+
+def get_deployment_from_list(provider, server, runtime_name):
+    deployment = None
+    for _deployment in MiddlewareDeployment.deployments(provider=provider, server=server):
+        if _deployment.name == runtime_name:
+            deployment = _deployment
+            break
+    assert deployment, 'Recently deployed archive was not found in deployments list'
+    return deployment
```
mfalesni: You can use the comment above to refactor this function in similar manner. The logic is exactly the same.

cfme/tests/middleware/test_middleware_deployment.py
hhovsepy: probably yes, these are product names

cfme/tests/middleware/test_middleware_deployment.py
mfalesni: Weird, I will investigate.

jkrocil: Meta blocker was not skipping in upstream?

mfalesni:
```python
from utils.blockers import Blocker
b = Blocker.parse('GH#ManageIQ/manageiq:10138')  # => True
```

Worth noting this should only skip on upstream appliances.

dajoRH:
Lint report for commit 5690998b8fc82ec2769263290aff7313dc0dea62:

*CFME QE Bot*

cfme/tests/middleware/test_middleware_deployment.py
hhovsepy: @mfalesni it fails to check blocker: http://10.16.45.124/iWQue95I/artifacts/report.html

mfalesni:
@psav code-wise it looks okay, what do you think?

PR #3205, opened 2016-07-21
Ansible Entry point has to be default

sshveta:


dajoRH:
Lint report for commit e3eea53c789ea363f50c6093778b8f06a5e09ddd:

*CFME QE Bot*

psav:
Looks good, but do we have some more explanation as there are no PRT runs that back up the change.

PR #3206, opened 2016-07-22
Introduce summary unit parsing

mfalesni:
Sample usage:

```python
In [4]: from utils.providers import get_crud

In [5]: p = get_crud('provider')

In [6]: p.summary.properties.aggregate_host_cpus.value
Out[6]: 8

In [7]: p.summary.properties.aggregate_host_memory.value
Out[7]: Unit(464.0, 'G', 'B')

In [8]: p.summary.properties.aggregate_host_memory.value == '464 GB'
Out[8]: True

In [9]: p.summary.properties.aggregate_host_cpu_resources
Out[9]: '91.2 GHz'

In [10]: p.summary.properties.aggregate_host_cpu_resources.value
Out[10]: Unit(91.2, 'G', 'Hz')

In [11]: p.summary.properties.aggregate_host_memory.value > '400 GB'
Out[11]: True
```

So it is very easy to compare the numeric values. Even across the different prefixes.

The parsing also happens for the compared values.

### DIFF
--- a/utils/units.py
+++ b/utils/units.py
@@ -1,0 +1,94 @@
+# -*- coding: utf-8 -*-
+import math
+import re
+
+# TODO: Split the 1000 and 1024 factor out. Now it is not an issue as it is used FOR COMPARISON ONLY
+FACTOR = 1024
+PREFIXES = ['', 'K', 'M', 'G', 'T', 'P']
+FACTORS = {prefix: int(math.pow(FACTOR, i)) for i, prefix in enumerate(PREFIXES)}
+
+UNITS = ['Byte', 'Bytes', 'B', 'b', 'Hz']
+
+EQUAL_UNITS = {
+    'B': ('Byte', 'Bytes')
+}
+
+# Sanity check
+for target_unit, units in EQUAL_UNITS.iteritems():
+    assert target_unit in UNITS
+    for unit in units:
+        assert unit in UNITS
+
+REGEXP = re.compile(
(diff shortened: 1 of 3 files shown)


dajoRH:
Lint report for commit 5b4424d0fb0faebef4b4932d6073b8d80aee61a7:

*CFME QE Bot*

utils/units.py
```diff
@@ -0,0 +1,66 @@
+# -*- coding: utf-8 -*-
+import math
+import re
+
+
+FACTOR = 1024
+PREFIXES = ['', 'k', 'M', 'G', 'T']
```
psav: should we allow for ```('k', 'K')```? I guess we are ignoring that 'm' and 'M' would be different

mfalesni: I guess if capital K appears for kilo values, then that is a bug :)

psav: K is byte, k is bit right?

psav:
Is there an existing module for this?

mfalesni:
@psav There is once, but an extreme overkill it seems.

psav:
I'm just thinking if there is one already done and being maintained, why not just use it? but just personal opinion? What are you worried about with the overkill...?? mem usage?

utils/units.py
mfalesni: no, prefix does not act in this

dajoRH:
Lint report for commit 044114dbfe19293332f29948578089e0a68897b5:

*CFME QE Bot*

PR #3209, opened 2016-07-25
Containers Replicators Test

pavelzag:


dajoRH:
Lint report for commit ef46a18e6a14f3bb4570199fc5b9805346775349:

*CFME QE Bot*

cfme/tests/containers/test_containers_default_project_replicators.py
```diff
@@ -0,0 +1,63 @@
...
+    """ Default Project Replicator properties test
+        This test checks the properties fields of each Replicator
+        Steps:
+            * Goes to Containers --> Replicators
+            * Makes sure "router-1" is one of the existing Replicators
+             * Goes through each Replicator and checks each Properties fields
+        """
+    sel.force_navigate('containers_replicators')
+    replicator_name = [r.name.text for r in list_tbl.rows()]
+    assert "router-1" in replicator_name
```
mfalesni: Shouldn't the "router-1" be specified outside of the test scope? I am not familiar with openshift, but I guess this might be rather specified in the cfme_data along the provider, or retrieve that using mgmtsystem to eliminate hardcoded values.

pavelzag: Hey Milan, do you have an example of a test where there's a similar behaviour? 
The 'router-1' replicator is a replicator that being installed automatically with every advanced installation of OpenShift v3.1 and beyond

cfme/tests/containers/test_containers_default_project_replicators.py
```diff
@@ -0,0 +1,63 @@
...
+from cfme.containers.replicator import Replicator
+from itertools import product
+from utils import testgen
+from utils.version import current_version
+
+
+pytestmark = [
+    pytest.mark.uncollectif(
+        lambda: current_version() < "5.6"),
+    pytest.mark.uncollectif(lambda provider: provider.version < 3.2),
```
jkrocil: I noticed that this is not skipping the OSE 3.1 provider on PRT and I assume it's because the `uncollectif` marks are not stacking. What I'm trying to say is you will probably need to do `pytest.mark.uncollectif(lambda provider: current_version() < "5.6" or provider.version < 3.2)` instead of 2 separate `uncollectif` marks.

pavelzag: Yea, I've been meaning to ask you about that. I'll try it out, thanks!

dajoRH:
Lint report for commit 02eba3a3133b7760ebde2d0458fa90fc3c60d96e:

*CFME QE Bot*

cfme/tests/containers/test_containers_default_project_replicators.py
```diff
@@ -0,0 +1,62 @@
...
+        (lambda: current_version() < "5.6") or (lambda provider: provider.version < 3.2)),
+    pytest.mark.usefixtures('setup_provider'),
+    pytest.mark.tier(1)]
+pytest_generate_tests = testgen.generate(
+    testgen.container_providers, scope='function')
+
+list_tbl = CheckboxTable(table_locator="//div[@id='list_grid']//table")
+
+
+REPLICATORS_PROPERTIES_FIELDS = ['Name', 'Creation timestamp', 'Resource version',
```
jkrocil: Consider passing in IDs as well instead of passing only values.
You can turn this
```
test_containers_replicators_properties[cm-env2-replicators_properties_fields0]
test_containers_replicators_properties[cm-env2-replicators_properties_fields1]
test_containers_replicators_properties[cm-env2-replicators_properties_fields2]
```
into much more valuable
```
test_containers_replicators_properties[cm-env2-name]
test_containers_replicators_properties[cm-env2-creation_timestamp]
test_containers_replicators_properties[cm-env2-resource_version]
```
.
Next step would be to make the test names shorter and we can land at something like:
```
test_replicator_properties[cm-env2-name]
test_replicator_properties[cm-env2-creation_timestamp]
test_replicator_properties[cm-env2-resource_version]
```

jkrocil: Examples for IDs here: http://docs.pytest.org/en/latest/parametrize.html

cfme/tests/containers/test_containers_default_project_replicators.py
mfalesni: Any test that uses mgmtsystem to check something in the provider directly and then verifying that cfme displays the same. My point is that I don't want to have such provider-dependent constants in the tests, I would personally prefer if they lived in the cfme_data together with the provider data, but the best solution would be if we can read the router name from the provider so we are not dependent on some constant.

cfme/tests/containers/test_containers_default_project_replicators.py
mfalesni: The thing is, the product returns list of n-tuples but there is only one argument, therefore tuple gets passed in the test and the ugly id is generated. Consider splitting the product in its respective values by addign multiple arguments, then it will work as you would expected :)

cfme/tests/containers/test_containers_default_project_replicators.py
```diff
@@ -0,0 +1,62 @@
...
+                                 'Requested pods', 'Current pods']
+REPLICATORS_RELATIONSHIPS_FIELDS = ['Containers Provider', 'Project', 'Pods', 'Nodes',
+                                    'My Company Tags']
+
+# CMP - 9531
+
+
+@pytest.mark.parametrize('replicators_properties_fields', product(REPLICATORS_PROPERTIES_FIELDS))
+def test_containers_replicators_properties(provider, replicators_properties_fields):
```
mfalesni:
```python
@pytest.mark.parametrize(('prop', 'rel'), product(REPLICATORS_PROPERTIES_FIELDS))
def test_containers_replicators_properties(provider, prop, rel):
```

This is what I suggested in the previous comment.

mfalesni: I fixed it up

pavelzag: So I tried to go this way:
`@pytest.mark.parametrize(('prop', 'rel'), product(REPLICATORS_PROPERTIES_FIELDS, REPLICATORS_RELATIONSHIPS_FIELDS))
def test_replicators_properties(provider, prop, rel):
    """ Default Project Replicator properties test
        This test checks the properties fields of each Replicator
        Steps:
            * Goes to Containers --> Replicators
            * Makes sure "router-1" is one of the existing Replicators
             * Goes through each Replicator and checks each Properties fields
        """
    sel.force_navigate('containers_replicators')
    replicator_name = [r.name.text for r in list_tbl.rows()]
    assert router in replicator_name
    for name in replicator_name:
        obj = Replicator(name, provider)
        prop_val = obj.get_detail('Properties', ''.join(prop))
        rel_val = obj.get_detail('Relationships', ''.join(rel))
        assert prop_val
        assert rel_val`

But through the whole run I stuck with the changing values for 'rel' but same value for 'prop'.
What am I doing wrong here?

mfalesni: Do not do the joins in first place. The parameters are strings, not arrays so it makes no sense.

Send me the list of collected tests for this file (--collect-only)

pavelzag: Here's what I see:
http://pastebin.test.redhat.com/395657

mfalesni: That is correct.

dajoRH:
Lint report for commit de640fbcb9c33871554f527828bdfa88c318b863:

*CFME QE Bot*

dajoRH:
Lint report for commit 76602ddaded0657955d06d75331fc0dfe3286068:

`cfme/tests/containers/test_containers_default_project_replicators.py`:
- :bangbang: Line 29:101: **E501** *line too long (115 > 100 characters)*

*CFME QE Bot*

dajoRH:
Lint report for commit 8c7bb3be8e15a71e2a1aef36a1465eb49580c049:

*CFME QE Bot*

psav:
Lots of failures, but node ids are looking better

PR #3227, opened 2016-07-28
Ignore one kind of table for summaries

mfalesni:
For @jdemon519

### DIFF
--- a/cfme/common/__init__.py
+++ b/cfme/common/__init__.py
@@ -231,6 +231,8 @@ class Summary(object):
 class SummaryTable(object):
     ROWS = '../../../tbody/tr'
 
+    MULTIKEY_LOC = '../../../tbody/tr[1]/td/strong'
+
     def __init__(self, o, text, entry):
         self._object = o
         self._text = text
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit 7bd7b3e205aba8f579a1ecf3cd8015a3098923b5:

*CFME QE Bot*

cfme/common/__init__.py
```diff
@@ -246,6 +248,9 @@ def __repr__(self):
     def load(self):
         self._keys = []
         key_values = []
+        if sel.is_displayed(self.MULTIKEY_LOC, root=self._entry):
+            # WE cannot process this kind of table yet.
+            return
```
psav: Do we need to raise and error here or something?

mfalesni: @psav 

No. It is a kind of table I haven't coded yet, so it just won't read it. This one will be a bit tricky to code up though.

FYI, the table has columns, first is name, second key, last value. So a sort of key/value database for that name. It is obvious that it does not fit the current scheme so a new class will have to be oded. But since it is not required yet ... :)

PR #3246, opened 2016-08-03
Verpick AngularSelect

mfalesni:
Fun fact, Select() can work with the fancy selects, but since the .locate() must return the select tag, it stopped working with the form wait.

Therefore I verpicked some of the most prominent errors in hte jenkins here.


dajoRH:
Lint report for commit efbefdbc06799ba8fcd324169ea47e1675508d13:

*CFME QE Bot*

PR #3255, opened 2016-08-04
add image_url arg for template uploader.

amavinag:
* added image_url for template uploader.
  * sync template to trackerbot as soon as
    upload is successful, no need to wait until
    sync template tracker is run.

### DIFF
--- a/scripts/template_upload_all.py
+++ b/scripts/template_upload_all.py
@@ -40,6 +40,9 @@ def parse_cmd_line():
                              'can be specified as e.g downstream_542, downstream_532,'
                              'upstream',
                         default=None)
+    parser.add_argument('--image_url', dest='image_url',
+                        help='url for the image to be uploaded',
+                        default=None)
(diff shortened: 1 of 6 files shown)


dajoRH:
Lint report for commit ae2fc843356b491b32f3cc193010ddc946b74275:

*CFME QE Bot*

psav:
What does the image_url do?

amavinag:
@psav, we can pass "image_url" directly to upload the images.

1. currently, we have "stream" param and we get the url based on stream from cfme_data, if stream is not passed then we start uploading images for all the urls in cfme_data. (that's the case with our jobs now). 
Our jenkins jobs pass the "provider_type" and does not pass the "stream or url" to uploader. 

2. In image uploader jenkins jobs we can get which image has been uploaded to file share and pass that "image_url" to trigger the uploader. since image_url is available here, we can directly pass that rather calculating stream.

PR #3282, opened 2016-08-10
Projects Summary Tests

pavelzag:
{{pytest: cfme/tests/containers/test_containers_projects_summary.py -v --use-provider container }}

### DIFF
--- a/cfme/tests/containers/test_containers_projects_summary.py
+++ b/cfme/tests/containers/test_containers_projects_summary.py
@@ -1,0 +1,60 @@
+import pytest
+from cfme.fixtures import pytest_selenium as sel
+from cfme.containers.project import Project
+from cfme.web_ui import CheckboxTable
+from utils import testgen
+from utils.version import current_version
+
+
+pytestmark = [
+    pytest.mark.uncollectif(
+        lambda: current_version() < "5.6"),
+    pytest.mark.usefixtures('setup_provider'),
+    pytest.mark.tier(1)]
+pytest_generate_tests = testgen.generate(
+    testgen.container_providers, scope='function')
+
+
+projects_properties_fields = ['Name', 'Creation timestamp', 'Resource version']
+projects_relationships_fields = ['Containers Provider', 'Routes', 'Services', 'Replicators',
+                                 'Pods', 'Nodes']
+
+list_tbl = CheckboxTable(table_locator="//div[@id='list_grid']//table")
+
+
+# CMP - 9523
+
+
+def test_containers_projects_summary_relationships(provider):
+    """ Relationships fields tests in Project summary
+        This test checks correct population of the Relationships Fields in Containers Projects'
+        details menu
+        Steps:
+            * Goes to Containers -- > Projects menu
+            * Go through each Container Project in the menu and check validity of
+            Relationships fields
+        """
+    sel.force_navigate('containers_projects')
+    project_name = [r.name.text for r in list_tbl.rows()]
+    for name in project_name:
+        for field in projects_relationships_fields:
+            obj = Project(name, provider)
+            val = obj.get_detail('Relationships', field)
+            assert val
+
+
+def test_containers_projects_summary_properties(provider):
+    """ Properties fields tests in Project summary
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit e832dea00479746a1a1c2b92af12e484e74bc805:

`cfme/tests/containers/test_containers_projects_summary.py`:
- :red_circle: Line 2:1: **F401** *'product' imported but unused*

*CFME QE Bot*

dajoRH:
Lint report for commit d16f1c2d1bf48c1be8f370aafcf2147f979b9b4b:

*CFME QE Bot*

psav:
Squash and I think we can merge this, upstream has some failures but we can overlook for now

cfme/tests/containers/test_containers_projects_summary.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+projects_properties_fields = ['Name', 'Creation timestamp', 'Resource version']
+projects_relationships_fields = ['Containers Provider', 'Routes', 'Services', 'Replicators',
+                                 'Pods', 'Nodes']
+
+list_tbl = CheckboxTable(table_locator="//div[@id='list_grid']//table")
+
+
+# CMP - 9523
```
nellyc: is this one case in Polarion or two? if its two, please add the other TC number
also I think that you can use parametarization here instead of code duplication
its the same test just checking a different field name

pavelzag: Just one Nelly. I am checking what I can do

psav:
Squash and we can merge this

dajoRH:
Lint report for commit 9568efb433ecfb8c2f3c94eac186ae781bc58573:

*CFME QE Bot*

dajoRH:
Lint report for commit 001c4c5dd58546563e1f6fe6416da07ced3c3e0c:

*CFME QE Bot*

dajoRH:
Lint report for commit 7dfafc695904c6827065c5ef20d38ca7e2596bfb:

*CFME QE Bot*

cfme/tests/containers/test_containers_projects_summary.py
```diff
@@ -0,0 +1,60 @@
...
+            * Go through each Container Project in the menu and check validity of
+            Relationships fields
+        """
+    sel.force_navigate('containers_projects')
+    project_name = [r.name.text for r in list_tbl.rows()]
+    for name in project_name:
+        for field in projects_relationships_fields:
+            obj = Project(name, provider)
+            val = obj.get_detail('Relationships', field)
+            assert val
```
psav: This asserts that the value isn't equiv to False, so if the value is 0, then the assert will trigger.

pavelzag: So, maybe something like assert val != 0 ?

mfalesni: get_details returns strings, so you would have to cast it to integer first.

pavelzag: Wait, hold the horses.. The val here might return strings and numbers. Returning a zero here is not necessarily a problem. 
For example, Containers Provider is a string, Routes can definitely bring back 0 (Not a healthy situation but not necessarily a bug)

mfalesni: Maybe explain more?

pavelzag: The tests goes through various Properties and Relationships fields of Projects. These fields can contain either string or integers. They can even contain 0 which will not be necessarily a bug that should stop the test from running. Hence I am not sure we need to cast it to integer before the assertion

dajoRH:
Lint report for commit 7dfafc695904c6827065c5ef20d38ca7e2596bfb:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

pavelzag:
@dajoRH please merge

dajoRH:
Response to comment #246184563:

It seems like all the basic requirements were met, I have therefore merged your PR.

PR #3283, opened 2016-08-10
Fix selects in new catalog item form

mkoura:
{{pytest: cfme/tests/services/test_service_catalogs.py -k test_order_catalog_item_via_rest --use-provider vsphere55 --long-running -v}}

### DIFF
--- a/cfme/services/catalogs/catalog_item.py
+++ b/cfme/services/catalogs/catalog_item.py
@@ -73,7 +73,7 @@ request_form = tabstrip.TabStripForm(
             ('vm_filter', Select('//select[@id="service__vm_filter"]')),
             ('catalog_name', Table('//div[@id="prov_vm_div"]/table')),
             ('vm_name', '//input[@name="service__vm_name"]'),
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit 0a51a4f53019705a4ef0f7250307f00fc28eb2cf:

*CFME QE Bot*

psav:
Tweaked the prt metadata for test

cfme/services/catalogs/catalog_item.py
```diff
@@ -225,8 +225,8 @@ def create(self):
         sel.force_navigate('catalog_item_new',
                            context={'provider_type': self.item_type})
         sel.wait_for_element(basic_info_form.name_text)
-        catalog = fakeobject_or_object(self.catalog, "name", "<Unassigned>")
-        dialog = fakeobject_or_object(self.dialog, "name", "<No Dialog>")
+        catalog = fakeobject_or_object(self.catalog, "name", "Unassigned")
```
psav: Not sure this is the same for both 5.5 and 5.6, but we can wait and see what the result is on PRT

PR #3286, opened 2016-08-10
Improved SSHResult

mfalesni:
namedtuple() was good, but I saw a way to simplify the testing code. Old behaviour is kept. New behaviour just gets out of your way.

### DIFF
--- a/utils/ssh.py
+++ b/utils/ssh.py
@@ -24,7 +24,56 @@ from utils.timeutil import parsetime
 # Default blocking time before giving up on an ssh command execution,
 # in seconds (float)
 RUNCMD_TIMEOUT = 1200.0
-SSHResult = namedtuple("SSHResult", ["rc", "output"])
+
+
+class SSHResult(namedtuple("SSHResult", ["rc", "output"])):
+    """Allows rich comparison for more convenient testing.
+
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit d6c83bce6196c3276af369857c05394c8fe2fbe5:

*CFME QE Bot*

utils/ssh.py
```diff
@@ -24,7 +24,46 @@
...
+
+class SSHResult(namedtuple("SSHResult", ["rc", "output"])):
+    """Allows rich comparison for more convenient testing.
+
+    Given you have ``result`` which is an instance of :py:class:`SSHResult`, you can do as follows:
+
+    .. code-block:: python
+
+        assert result  # If $?=0, then the result evaluates to a truthy value and passes the assert
+        assert result == 'installed'  # direct matching of the output value
```
psav: Perhaps it would be good to have an "in" to check a substring also?
The Result is the output of the func right?

mfalesni: That is also doable. Yes, result is an output of all operations that invoke commands.

dajoRH:
Lint report for commit 35018fa306237ad7c25b73a0c02b711ec5ae280c:

*CFME QE Bot*

utils/ssh.py
mfalesni: It is done.

PR #3289, opened 2016-08-11
Parallelizer: Ensure the timer is always rescheudled

mfalesni:
A rogue exception could have caused the problems we had. When I wrote this, I forgot to put the reset timer function in the finally block. Since an unexpected non-sprout exception is something potentially dangerous, doing a short timeout in that case to "wait" for the good state to come back..


dajoRH:
Lint report for commit e0b71bdcb754daef5162229cec9e5f4d2fb30bbc:

*CFME QE Bot*

PR #3219, opened 2016-07-27
Initial Azure Power Control implementation - Revised

jteehan:
This is the short version now that most of the provisioning stuff is already in master.

{{pytest: cfme/tests/cloud/test_instance_power_control.py -k "test_stop" --long-running --use-provider azure }}

PR #3298, opened 2016-08-12
Azure does not display the tabs

mfalesni:
{{pytest: -v cfme/tests/cloud/test_instance_power_control.py --long-running --use-provider azure -k "not cancel" }}

### DIFF
--- a/cfme/cloud/instance.py
+++ b/cfme/cloud/instance.py
@@ -7,6 +7,7 @@ from cfme.web_ui import (
     accordion, fill, flash, paginator, toolbar, CheckboxTree, Region, Tree, Quadicon)
 from cfme.web_ui.menu import extend_nav
 from functools import partial
+from utils import deferred_verpick, version
 from utils.api import rest_api
 from utils.wait import wait_for
 
@@ -162,6 +163,7 @@ class Instance(VM):
 
     def get_collection_via_rest(self):
         return rest_api().collections.instances
+
 
 @VM.register_for_provider_type("openstack")
 class OpenStackInstance(Instance):
@@ -367,7 +369,12 @@ class AzureInstance(Instance):
     START = "Start"
     POWER_ON = START  # For compatibility with the infra objects.
     STOP = "Stop"
-    TERMINATE = "Terminate"
+    SUSPEND = "Suspend"
+    DELETE = "Delete"
+    TERMINATE = deferred_verpick({
+        version.LOWEST: 'Terminate',
+        '5.6.1': 'Delete',
+    })
     # CFME-only power control options
     SOFT_REBOOT = "Soft Reboot"
     # Provider-only power control options
@@ -449,6 +456,8 @@ class AzureInstance(Instance):
             self.provider.mgmt.stop_vm(self.name)
         elif option == AzureInstance.RESTART:
             self.provider.mgmt.restart_vm(self.name)
+        elif option == AzureInstance.SUSPEND:
+            self.provider.mgmt.suspend_vm(self.name)
         elif option == AzureInstance.TERMINATE:
             self.provider.mgmt.delete_vm(self.name)
         else:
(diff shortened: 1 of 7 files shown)


dajoRH:
Lint report for commit b7b69ea7b107c533da2e64959898366bdfe83ed4:

*CFME QE Bot*

jteehan:
I like this a lot better.  I agree that providers should not be worrying about the underlying intricacies of html changes.  I had been rebasing every day hoping my azure issue would just go away.  :)

mfalesni:
@jteehan So, I cherrypicked your commit without the "gut" changes and it seems to work. I will comment on some lines to show the differences between your and this implementation.

cfme/cloud/instance.py
```diff
@@ -367,7 +369,12 @@ class AzureInstance(Instance):
     START = "Start"
     POWER_ON = START  # For compatibility with the infra objects.
     STOP = "Stop"
-    TERMINATE = "Terminate"
+    SUSPEND = "Suspend"
+    DELETE = "Delete"
+    TERMINATE = deferred_verpick({
+        version.LOWEST: 'Terminate',
+        '5.6.1': 'Delete',
+    })
```
mfalesni: I used a deferred verpick here, which only accesses the SSH of the appliance after accessed for the first time.

cfme/cloud/provider/azure.py
```diff
@@ -27,7 +28,14 @@ def _form_mapping(self, create=None, **kwargs):
...
         credentials = cls.process_credential_yaml_key(credentials_key)
-        return cls(name=prov_config['name'],
+        # HACK: stray domain entry in credentials, so ensure it is not there
+        credentials.domain = None
+        region = prov_config.get('region', None)
+        if region is not None and isinstance(region, dict):
+            region = pick(region)
+        return cls(
+            name=prov_config['name'],
+            region=region,
```
mfalesni: @jteehan Here I did not bother with verpicking in the code, but rather keeping the values in the yaml. We can't verpick all the values in the yaml, but with region we know there can be differences. In the yaml, you don't have to use the dictionary, as you see here, it is optional. And it also passes the region so it then subsequently gets inserted in the UI.

dajoRH:
Lint report for commit c8fd9976e0e3309cbc37bbdb45a5da46b458cff7:

*CFME QE Bot*

cfme/common/provider.py
```diff
@@ -71,7 +71,7 @@ def form(self):
                 ('google_service_account', Input('service_account')),
             ]
             tab_fields = {
-                "Default": [
+                ("Default", ('default_when_no_tabs', )): [
```
mfalesni: @jteehan This is the architectural thing I was talking about. This sets a flag on the tab that if no tabs are there, then if accessing a field that is under this tab do nothing. And if accessing a different tab, cry.

cfme/tests/cloud/test_instance_power_control.py
```diff
@@ -26,12 +27,15 @@ def testing_instance(request, setup_provider, provider):
...
         request.addfinalizer(instance.delete_from_provider)
+        provider.refresh_provider_relationships()
     elif instance.provider.type == "ec2" and \
             provider.mgmt.is_vm_state(instance.name, provider.mgmt.states['deleted']):
         provider.mgmt.set_name(
             instance.name, 'test_terminated_{}'.format(fauxfactory.gen_alphanumeric(8)))
         instance.create_on_provider(allow_skip="default", find_in_cfme=True)
         request.addfinalizer(instance.delete_from_provider)
+        provider.refresh_provider_relationships()
+    instance.wait_to_appear()
```
mfalesni: @jteehan This file is almost all yours except these three lines I added.

dajoRH:
Lint report for commit 3a1ce3fae596ef1630169925bc61d09534f291a1:

*CFME QE Bot*

mfalesni:
Closing in favour of #3219 which contains the changes from this PR.

PR #3299, opened 2016-08-12
Notify user that the precompiling is happening and it is not stuck

mfalesni:


dajoRH:
Lint report for commit 21c6a370b3f98d1f1350477fa4f838c72216c9b1:

*CFME QE Bot*

mfalesni:
The messages are present in the log, so this should be all good for merging

PR #3330, opened 2016-08-19
Adding test_provider_crud.py back

jdemon519:
Purpose or Intent
=================

test_provider_crud.py somehow got deleted, adding it back

{{pytest: cfme/tests/containers/test_provider_crud.py -v --use-provider container }}

### DIFF
--- a/cfme/containers/image.py
+++ b/cfme/containers/image.py
@@ -8,6 +8,8 @@ from . import details_page
 
 list_tbl = CheckboxTable(table_locator="//div[@id='list_grid']//table")
 
+checkbox = ".//input[@class='list-grid-checkbox']"
+
 nav.add_branch(
     'containers_images',
     {
--- a/cfme/containers/image.py
+++ b/cfme/containers/image.py
@@ -5,6 +5,7 @@ from cfme.fixtures import pytest_selenium as sel
 from cfme.web_ui import toolbar as tb, CheckboxTable
 from cfme.web_ui.menu import nav
 from . import details_page
+from cfme.web_ui import flash
 
 list_tbl = CheckboxTable(table_locator="//div[@id='list_grid']//table")
 
@@ -22,6 +23,18 @@ nav.add_branch(
             {'Name': ctx['image'].name, 'Provider': ctx['image'].provider.name}),
     }
 )
+
+
+def run_smart_state_analysis(self):
+    sel.force_navigate('containers_images')
+    notification = "Analysis successfully initiated"
+    sel.check(checkbox)
+    tb.select(
+        'Configuration',
+        'Perform SmartState Analysis',
+        invokes_alert=True)
+    sel.handle_alert()
+    flash.assert_message_contain(notification)
 
 
 class Image(Taggable, SummaryMixin):
--- a/cfme/tests/containers/test_containers_properties.py
+++ b/cfme/tests/containers/test_containers_properties.py
@@ -1,0 +1,79 @@
+# -*- coding: utf-8 -*-
+
+""" This module verifies data integrity in the Properties table for:
+    pods, routes and projects
+"""
+import pytest
+from cfme.fixtures import pytest_selenium as sel
+from cfme.containers.pod import Pod, list_tbl as list_tbl_pod
+from cfme.containers.route import Route, list_tbl as list_tbl_route
+from cfme.containers.project import Project, list_tbl as list_tbl_proj
+from utils import testgen
+from utils.version import current_version
+
+
+pytestmark = [
+    pytest.mark.uncollectif(
+        lambda: current_version() < "5.6"),
+    pytest.mark.usefixtures('setup_provider'),
+    pytest.mark.tier(1)]
+pytest_generate_tests = testgen.generate(
+    testgen.container_providers, scope="function")
+
+""" CMP-9911
+"""
+
+
+@pytest.mark.parametrize('rel',
+                         ['name',
+                          'phase',
+                          'creation_timestamp',
+                          'resource_version',
+                          'restart_policy',
+                          'dns_policy',
+                          'ip_address'
+                          ])
+def test_pods_properties_rel(provider, rel):
+    sel.force_navigate('containers_pods')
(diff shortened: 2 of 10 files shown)


dajoRH:
Lint report for commit 0cc8386c069243eac16ff48a531d215059589128:

*CFME QE Bot*

cfme/tests/containers/test_containers_properties.py
```diff
@@ -0,0 +1,69 @@
...
+    pytest.mark.uncollectif(
+        lambda: current_version() < "5.6"),
+    pytest.mark.usefixtures('setup_provider'),
+    pytest.mark.tier(1)]
+pytest_generate_tests = testgen.generate(
+    testgen.container_providers, scope="function")
+
+# CMP-9911 # CMP-9877 # CMP-9867
+
+
```
psav: As you are navigating to the page showing _all_ pods, I think you either want to
a) use the ```has_no_container_providers``` fixture to remove all providers before you run the test, probably meaning you need to define some other fixtures, to get the dependency right
b) Navigate to Pods (through) the provider, so tat you only get pods that are related to the provider you are using.

cfme/tests/containers/test_relationships_tables.py
```diff
@@ -180,8 +166,7 @@ def test_images_rel(provider, rel):
         obj = Image(name, provider)
 
         val = obj.get_detail('Relationships', rel)
-        if val == '0' or val == 'Unknown image source':
-            continue
+        assert val != 'Unknown image source'
```
psav: Before it was allowed to continue if the val was equal to "Unknown image source", now we say it absolutely can't be?

cfme/tests/containers/test_one_image_analysis.py
```diff
@@ -0,0 +1,70 @@
...
+        On the Images page ---> put a checkmark next to the image
+        Click on Configuration ---> Perform SmartState Analysis
+        Verify the relevant notification is displayed
+        Settings ---> Tasks ---> All VM and Container Analysis Tasks
+        Verify the image is being scanned
+        Wait for the scan to finish ---> Delete All
+    """
+    sel.force_navigate('containers_images')
+    m = 'Analysis successfully initiated'
+    sel.check(image.checkbox)
```
psav: This shouldn't be here, but rather a method on the ```Image``` class should be created called ```run_smartstate_analysis```. In this way others can benefit from this too. The ```force_navigate``` at the beginning of the test and lines 35-40 should be part of this new method too. This test then becomes two steps:

```
image.run_smartstate_analysis()
is_task_finished()
```

cfme/tests/containers/test_one_image_analysis.py
```diff
@@ -0,0 +1,70 @@
...
+        'Configuration',
+        'Perform SmartState Analysis',
+        invokes_alert=True)
+    sel.handle_alert()
+    flash.assert_message_contain(m)
+
+    is_task_finished()
+
+
+def is_task_finished():
```
psav: I'm presuming there is no, identifying information to tell which image analysis is complete?
Regardless, this function belongs in the image.py, and not in the test.

psav:
This PR needs squashing

cfme/tests/containers/test_containers_properties.py
jdemon519: @psav I am trying to go with the second option (navigate to Pods through the provider), but what's the best way to do that?
existing_pods = provider.mgmt.list_container_group()  ?

cfme/tests/containers/test_relationships_tables.py
jdemon519: @psav I will reference the bug that was opened in the script

dajoRH:
Lint report for commit a36ea7bb861149861cd10d6e889cf56b19697395:

*CFME QE Bot*

dajoRH:
Lint report for commit adacb3d06ee378199101aa6be7362e23c640ac7e:

`cfme/tests/containers/test_one_image_analysis.py`:
- :red_circle: Line 6:1: **F401** *'flash' imported but unused*

*CFME QE Bot*

dajoRH:
Lint report for commit 9d298b9f1923d92b02787102d57b07d44b67e1ba:

*CFME QE Bot*

cfme/containers/image.py
```diff
@@ -22,6 +25,18 @@
 )
 
 
+def run_smart_state_analysis(self):
```
psav: Can this run against a single image?
It seems to me that this should be a classmethod on Image
Possibly even two methods, one that says ```run_smart_state_analysis```
And the other would be a class method that was called ```run_smart_state_analysis_all_images``` or something like that.
Regardless, I think this should be a method of the Image class, perhaps we need a mixin for SSA at some point

cfme/tests/containers/test_containers_properties.py
psav: well doing that will get you the list via mgmtsystem which will be outside of CFME. Presumably you want the list of pods associated with the provider from inside CFME, ie "what CFME says". In this case you probably want to go to the details page and click on one of the relationship links. The summary attribute of the provider should help you out here.

cfme/containers/image.py
jdemon519: Right, this can be run against a single image. I agree about the mixin for SSA, it's just not available right now.

psav: Right, so lets make it a method on Image, and we can make the SSA mixin later

psav: Add a ```# TODO``` comment on it that we need to make the mixin

jkrocil:
Description of this PR has **very** little to do with its content. You are adding 10 files with various content and description only says that you're re-adding a previously removed test. I'm not reviewing this until you describe what you're trying to do here.

I'd suggest (as I already have) to split this PR into 2.
The first adding the removed test (which could be merged immediately and I wrongly believed that to be a priority for you - to have a working smoke test) and the other adding additional tests that need reviewing, unlike the CRUD test.

PR #3335, opened 2016-08-21
StatusBox fix for menu_ui

pavelzag:
Purpose or Intent
=================
This PR fixes StatusBox handling in menu_ui to allow StatusBox tests in CFME
The problem was the XPATH was incorrect and returned nothing

### DIFF
--- a/cfme/web_ui/__init__.py
+++ b/cfme/web_ui/__init__.py
@@ -3647,4 +3647,3 @@ class StatusBox(object):
             "//span[contains(@class,'card-pf-aggregate-status-count')]"
             "/../../../../../div[contains(@status, '{}')]".format(self.name))
         return int(re.findall(r'\d+', elem_text)[0])
-
--- a/cfme/web_ui/__init__.py
+++ b/cfme/web_ui/__init__.py
@@ -3639,9 +3639,11 @@ class StatusBox(object):
 
     """
     def __init__(self, name):
-        self.name = name
+        assert type(name) == str
+        self.name = name.lower()
 
     def value(self):
-        return sel.element(
-            '//span[contains(@class, "card-pf-aggregate-status-count")]'
-            '/../../span[contains(., "{}")]/span'.format(self.name)).text
+        elem_text = sel.element(
+            "//span[contains(@class,'card-pf-aggregate-status-count')]"
+            "/../../../../../div[contains(@status, '{}')]".format(self.name)).text
+        return int(re.findall(r'\d+', elem_text)[0])
--- a/cfme/web_ui/__init__.py
+++ b/cfme/web_ui/__init__.py
@@ -3643,7 +3643,8 @@ class StatusBox(object):
         self.name = name.lower()
 
     def value(self):
-        elem_text = sel.element(
+        elem_text = sel.text(
             "//span[contains(@class,'card-pf-aggregate-status-count')]"
-            "/../../../../../div[contains(@status, '{}')]".format(self.name)).text
+            "/../../../../../div[contains(@status, '{}')]".format(self.name))
         return int(re.findall(r'\d+', elem_text)[0])
+


dajoRH:
Lint report for commit f9603037233cacd2c6408bbdbc54a53443b8a6b2:

*CFME QE Bot*

psav:
I have tried the current code on a 5.6.1 appliance and it works, can you show me an example of where it doesn't work? As I said when it was developed it doesn't work on the Providers box, but should work on all the others. We probably need a special case for the Providers one.

pavelzag:
The problem I had was with the StatusBox in Container Dashboard in Compute --> Containers --> Overview.
The XPATH that was in the code didn't actually return anything so I had to change it up a bit to return me the value I needed. Are there any other Status Boxes

psav:
Is this regarding the wait that we were discussing via email? I've run this multiple times and it always seemed to work

pavelzag:
Pretty sure that this locator has changed and since I don't see any other test at the moment that uses StatusBox I think it's safe to assume that this PR will not break anything

cfme/web_ui/__init__.py
```diff
@@ -3639,9 +3639,11 @@ class StatusBox(object):
...
+        assert type(name) == str
+        self.name = name.lower()
 
     def value(self):
-        return sel.element(
-            '//span[contains(@class, "card-pf-aggregate-status-count")]'
-            '/../../span[contains(., "{}")]/span'.format(self.name)).text
+        elem_text = sel.element(
+            "//span[contains(@class,'card-pf-aggregate-status-count')]"
+            "/../../../../../div[contains(@status, '{}')]".format(self.name)).text
```
mfalesni: You can improve this one step further by doing `sel.text('locator')` directly. That can also handle cases when the element is not visible. And in one ofthe PRs, I have improved the `text()` function even more.

pavelzag: Not entirely understanding your suggestion.. Can you give me please a PR for example?

mfalesni: We tend to not use sel.element().something but we have wrapper functions that evetually can solve selenium issues. One of them is sel.text() which essentially does sel.element().text but does it better.

mfalesni: You see that I am refining the text() here: https://github.com/ManageIQ/integration_tests/pull/3338/files#diff-54c758ec2117f433d2c05581244ceb74R652 since if the element is not visible, then the text can't be read by selenium itself and it needs some help of JS, so that is why I recommended that :)

pavelzag: Ok, trying.. 
Thanks

pavelzag:
@mfalesni 
Something like this?

dajoRH:
Lint report for commit ab2641355814f57be44495b1ea6999fbaf700736:

`cfme/web_ui/__init__.py`:
- :heavy_exclamation_mark: Line 3650:1: **W391** *blank line at end of file*

*CFME QE Bot*

dajoRH:
Lint report for commit a5b34e4fc9227bd8f53f32e9c9479b519e7b1f16:

*CFME QE Bot*

cfme/web_ui/__init__.py
```diff
@@ -3639,9 +3639,11 @@ class StatusBox(object):
 
     """
     def __init__(self, name):
-        self.name = name
+        assert type(name) == str
```
mfalesni: We suppose that user passes a string, therefore I would recommend replacing this construct with:

```python
self.name = str(name).lower()
```

cfme/web_ui/__init__.py
```diff
@@ -3639,9 +3639,11 @@ class StatusBox(object):
...
+        self.name = name.lower()
 
     def value(self):
-        return sel.element(
-            '//span[contains(@class, "card-pf-aggregate-status-count")]'
-            '/../../span[contains(., "{}")]/span'.format(self.name)).text
+        elem_text = sel.text(
+            "//span[contains(@class,'card-pf-aggregate-status-count')]"
+            "/../../../../../div[contains(@status, '{}')]".format(self.name))
+        return int(re.findall(r'\d+', elem_text)[0])
```
jaryn: I gues you can use https://docs.python.org/2/library/re.html#re.search here instead

mfalesni:
@pavelzag yes, exactly that

pavelzag:
Fixed here
https://github.com/ManageIQ/integration_tests/pull/3362

PR #3341, opened 2016-08-23
container summary test

pavelzag:
Purpose or Intent
=================
This tests checks status boxes in the Containers Overview menu and verifies it shows the correct data for:
Pods, Containers and Nodes

{{pytest: cfme/tests/containers/test_containers_summary.py -v --use-provider container }}

### DIFF
--- a/cfme/tests/containers/test_containers_summary.py
+++ b/cfme/tests/containers/test_containers_summary.py
@@ -1,0 +1,75 @@
+import pytest
+from cfme.fixtures import pytest_selenium as sel
+from cfme.web_ui import StatusBox
+from utils import testgen
+from utils.version import current_version
+import time
+
+pytestmark = [
+    pytest.mark.uncollectif(
+        lambda: current_version() < "5.6"),
+    pytest.mark.usefixtures('setup_provider'),
+    pytest.mark.tier(1)]
+pytest_generate_tests = testgen.generate(
+    testgen.container_providers, scope='function')
+
+
+# container_dashboard
+# CMP-9822
+def test_containers_summary_containers(provider):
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit 8136bc4fe2ba7527dffa15936a4c658b2136c40b:

*CFME QE Bot*

cfme/tests/containers/test_containers_summary.py
```diff
@@ -0,0 +1,75 @@
...
+       This test checks that the amount of containers in the system is shown correctly
+        in the Containers widget in the
+       Overview menu
+       Steps:
+           * Goes to Compute --> Containers --> Overview
+           * Checks how many Containers are shown in the Containers Widget
+           * Goes to Containers summary page and checks how many Containers are shown there.
+           * Checks the amount is equal
+       """
+    sel.force_navigate('container_dashboard')
```
psav: When this test is run on PRT, it will return 0 for the values, because you didn't include the setup_provider fixture. Without it, you will get the provider fixture, but the provider won't be added to CFME, you also may want the fixture to ```has_no_container_providers```

pavelzag:
Not sure what you recommend here, I tried this:
http://pastebin.test.redhat.com/405483
And all it done is deleted and added providers over and over for the test cases. Looks a bit unproductive and effectively kills the test because it takes time to drop deleted objects and rescan them. :|

psav:
I actually did not see it up in the top section of the file. Ignore me. Regardless yes it will add the provider and delete each time, perhaps we can make this module scoped, @mfalesni ??

pavelzag:
Replaced with https://github.com/ManageIQ/integration_tests/pull/3369

PR #3338, opened 2016-08-23
History button | Improved toolbar escaping

mfalesni:
Purpose or intent
=================

* Add history button handling
* Improve the escaping in toolbar module (&)
* Make the text() function in selenium work EVERY FREAKIN TIME

```python
from cfme.web_ui import history


history.history_items()  # Returns a list of all history items
history.select_history_item('Provider xyz summary')  # Selects by the history item text
history.select_nth_history_item(0)  # Or selects the history items based on the position, 0 is the latest
```

I hope it works in all screens, I made the locator quite generic.

@pavelzag Here it is. You should be able to cherrypick this commit to play with it and once this gets merged, you should be able to remoe that commit and rebase on top of `master`.
@mmojzis Sorry, we needed this quick :)

PR #3351, opened 2016-08-25
Data integrity tests

pavelzag:
Purpose or Intent
=================
This test checks correct population of the Properties Fields in Containers/Image Registries/Container Images and Container Nodes details menu

image_registry.py file was edited to allow get_detail on this object

{{pytest: cfme/tests/containers/test_containers_data_integrity_properties.py -v --use-provider container }}

### DIFF
--- a/cfme/tests/containers/test_containers_data_integrity_properties.py
+++ b/cfme/tests/containers/test_containers_data_integrity_properties.py
@@ -1,0 +1,107 @@
+import pytest
+from cfme.fixtures import pytest_selenium as sel
+from cfme.web_ui import CheckboxTable
+from cfme.containers.image import Image
+from cfme.containers.image_registry import ImageRegistry
+from cfme.containers.pod import Pod
+from cfme.containers.provider import Provider
+from cfme.containers.node import Node
+from cfme.containers.container import Container
+from utils import testgen
+from utils.version import current_version
+from cfme.web_ui import history
+
+list_tbl = CheckboxTable(table_locator="//div[@id='list_grid']//table")
+
+pytestmark = [
+    pytest.mark.uncollectif(
+        lambda: current_version() < "5.6"),
+    pytest.mark.usefixtures('setup_provider'),
+    pytest.mark.tier(1)]
+pytest_generate_tests = testgen.generate(
+    testgen.container_providers, scope='function')
+
+container_image_fields = ['Name', 'Image Id', 'Full Name']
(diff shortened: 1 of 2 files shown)


dajoRH:
Lint report for commit 47d97f32ec6101c2b96a138d531e5a8c66d3154b:

*CFME QE Bot*

pavelzag:
The tests will fail because they use the history module from 
https://github.com/ManageIQ/integration_tests/pull/3338
When it's merged it shall work

cfme/tests/containers/test_containers_data_integrity_properties.py
```diff
@@ -0,0 +1,107 @@
...
+    """ Properties fields test in Containers
+        This test checks correct population of the Properties Fields in Containers' details menu
+        Steps:
+            * Goes to Containers -- > Containers menu
+            * Go through each Container in the menu and check validity of Properties fields
+        """
+    sel.force_navigate('containers_containers')
+    container_name = [r.name.text for r in list_tbl.rows()]
+    container_pod_name = [r.pod_name.text for r in list_tbl.rows()]
+    dictionary = dict(zip(container_name, container_pod_name))
```
psav: This code is also cropping up in other container PRs, I strongly recommend you guys get together and make a ```get_all_pods``` method for your provider object, which will return for you a list of Pod objects. Or is this something different?

pavelzag: Can you give me an example of another PR where it happens? I'll check what I can do about it

mfalesni: Yeah, I have a strong illusion that all the Container PRs add the same code when I am looking at them so I am lost in that. Pete made a good point. What repeats, needs DRY :)

pavelzag: I understand what you want me to do, but I am not sure where can I stick that part of the code that should scan through content of the table and give me a list of all the containers/pods/images in the table. Should it be in Provider object? Is there something similar in the code? I don't know the infra code enough to decide where to stick it.

psav: Yup that's exactly what I think would be a good idea. Container Provider object has a ```get_all_pods``` method that returns a list of Pod objects.

pavelzag: I'll try to concoct something, thanks

pavelzag:
Replaced by https://github.com/ManageIQ/integration_tests/pull/3367

PR #3326, opened 2016-08-19
Fixing Power action tests on 5.6+

lkhomenk:
#Purpose or Intent
=================

__Fixing__ test_vm_power_control.py and test_instance_power_control.py tests because they were failing

{{pytest: cfme/tests/cloud/test_instance_power_control.py -k test_terminate --long-running}}

PR #3364, opened 2016-08-29
[WIPTEST]fixing finding Quadicon by short name in upstream

lkhomenk:
Purpose or Intent
=================

fixing search Quadicon by short name if name is greater than 12 symbols

A lot of test were failing as title in upstream was changed to title=test_...xlkjT(5 last symbols)

{{pytest: cfme/tests/cloud/test_instance_power_control.py::test_stop --long-running}}

### DIFF
--- a/cfme/web_ui/__init__.py
+++ b/cfme/web_ui/__init__.py
@@ -2342,7 +2342,10 @@ class Quadicon(Pretty):
             return "contains(normalize-space(@title), {name})"\
                 .format(name=quoteattr('Name: {}'.format(self._name)))
         else:
-            return "@title={name} or @data-original-title={name}".format(name=quoteattr(self._name))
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit 466bf6d12f1c8026789b311881036893b7b67d31:

*CFME QE Bot*

lkhomenk:
This PR is needed for #3326 for Upstream

psav:
We have a GH Issue open for this, unfortunately this fix could result in the wrong name being chosen. It is likely that we will have to wait for this to be fixed upstream.

mfalesni:
I saw that it should be possible to retrieve the quadicon by name but that would required an overhaul of the locator structure.

jkrocil:
It is a bug in upstream ManageIQ. This affects all quadicon types so letsl wait for a fix.

PR #3383, opened 2016-09-02
fix provider setup in advanced search tests

mkoura:
Fixing `setup_a_provider`call in advanced search tests.
Fixing `setup_a_provider` function so that it doesn't iterate over chars of "required_keys" when "required_keys" is a string, not list (in `all(key in providers_data[provider] for key in required_keys)`).

### DIFF
--- a/cfme/tests/infrastructure/test_advanced_search_vms.py
+++ b/cfme/tests/infrastructure/test_advanced_search_vms.py
@@ -16,7 +16,7 @@ from cfme.web_ui.cfme_exception import (assert_no_cfme_exception,
 def vms():
     """Ensure the infra providers are set up and get list of vms"""
     try:
-        setup_a_provider(prov_type="infra", required_keys="large")
+        setup_a_provider(prov_class="infra", required_keys=["large"])
     except Exception:
         pytest.skip("It's not possible to set up any providers, therefore skipping")
     pytest.sel.force_navigate("infra_vms")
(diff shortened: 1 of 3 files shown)


dajoRH:
Lint report for commit f2d0838e9a897b3dc9dae0491e1cc3fdc763d175:

Commit inspection report:
**No commit flaws detected**.

*CFME QE Bot*

jkrocil:
Looks good, merging. +1

cfme/tests/infrastructure/test_advanced_search_vms.py
```diff
@@ -16,7 +16,7 @@
 def vms():
     """Ensure the infra providers are set up and get list of vms"""
     try:
-        setup_a_provider(prov_type="infra", required_keys="large")
+        setup_a_provider(prov_class="infra", required_keys="large")
```
jkrocil: Fix this to pass a list instead of a string, please.

utils/providers.py
```diff
@@ -135,6 +135,8 @@ def setup_a_provider(prov_class="infra", prov_type=None, validate=True, check_ex
     """
     if not required_keys:
         required_keys = []
+    elif not hasattr(required_keys, '__iter__'):
+        required_keys = [required_keys]
```
jkrocil: This is not necessary. We cannot safeguard every function against strings. Instead, we should fix the incorrect fn calls.

dajoRH:
Lint report for commit 6d6f1cd57b1dba3eb3b1ea2f84ef4b128b79ee57:

Commit inspection report:
**No commit flaws detected**.

*CFME QE Bot*

jkrocil:
Merging :+1:

PR #3389, opened 2016-09-06
fix gce template_deploy

dajohnso:
Purpose or Intent
=================
Added GCE support for deploy_templates through the framework.

The way we do this probably need to be re-worked but this is how it is
today. 

(Requires wrapanapi PR#65)


dajoRH:
Lint report for commit bbd8fbe42618bea315afb8cfaf1a338ceb68078c:

Commit inspection report:
**No commit flaws detected**.

*CFME QE Bot*

PR #3395, opened 2016-09-07
Fix docs content test. Added dict for content

otsuman:
{{pytest: cfme/tests/configure/test_docs.py -k test_contents}}

### DIFF
--- a/cfme/tests/configure/test_docs.py
+++ b/cfme/tests/configure/test_docs.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 import pytest
-import re
 import requests
 try:
     # Faster, C-ext
@@ -46,7 +45,6 @@ def guides():
         ) is not None]
 
 
-@pytest.mark.tier(3)
 @pytest.fixture(scope="session")
 def docs_info():
     if version.current_version() < "5.4.0.1":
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit 40722106b3a2ce7700a256c5613fda7a88399372:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

dajoRH:
Lint report for commit 136c2e6f992e0692f904d0926adceb195c63918d:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

cfme/tests/configure/test_docs.py
```diff
@@ -84,8 +83,21 @@ def docs_info():
         return []
 
 
+@pytest.mark.tier(3)
```
jkrocil: Fixtures shouldn't be marked with the `tier` marker, only tests, so please remove the tier marker from here and from the `docs_info` fixture as well.
Also, I assume you were inspired by the `docs_info` fixture above when creating `docs_content` but static data usually should be simply stored as constant data in the module, not inside a session-scoped fixture. The difference with `docs_info` is that there is version-dependent data in there and we don't want to run version.pick on module import because it needs to connect to a running appliance to decide the version, so it needs to be in a fixture which is not run on module import. Your static data inside `docs_content` could be stored just as a constant dict, but... since we can safely assume that it will need to be version picked in the near future (with next release) lets keep it as a session-scoped fixture. :+1: So just remove the unneeded markers, please, and this can be merged.

dajoRH:
Lint report for commit 44c290dd184046da3622d3289081304cfc00232b:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

cfme/tests/configure/test_docs.py
```diff
@@ -84,8 +82,20 @@ def docs_info():
...
+            'lifecycle_and_automation_guide': 'methods available for automation',
+            'insight_guide': 'managing infrastructure and inventory',
+            'providers_guide': 'managing providers',
+            'scripting_actions_guide': 'scripting actions in cloudforms',
+            'control_guide': 'policies and profiles guide'}
+
+
 @pytest.mark.tier(2)
-@pytest.mark.meta(blockers=[1272618])
+@pytest.mark.ignore_stream("upstream")
```
jkrocil: One last thing - since we mark all tests in this module with `ignore_stream('upstream')`, we should use `pytestmark = pytest.mark.ignore_stream("upstream")` at the top of this module instead. https://github.com/ManageIQ/integration_tests/blob/a97cf316a683009c9caa4db5c08d90bf5b1fa47a/cfme/tests/cloud/test_stack.py#L8

PR #3410, opened 2016-09-12
Basic Metrics Test

pavelzag:
Purpose or Intent
=================

This is a basic test that checks if Metrics for OSE are up
{{pytest: cfme/tests/containers/test_basic_metrics.py -v --use-provider container }}

PR #3414, opened 2016-09-13
Basic Metrics Test

pavelzag:
Purpose or Intent
=================

This is a basic test that checks if Metrics for OSE are up
{{pytest: cfme/tests/containers/test_basic_metrics.py -v --use-provider container }}

### DIFF
--- a/cfme/tests/containers/test_basic_metrics.py
+++ b/cfme/tests/containers/test_basic_metrics.py
@@ -1,0 +1,17 @@
+import pytest
+from utils import testgen
+from utils.version import current_version
+
+pytestmark = [
+    pytest.mark.uncollectif(lambda provider: current_version() < "5.6"),
+    pytest.mark.tier(1)]
+pytest_generate_tests = testgen.generate(
+    testgen.container_providers, scope='function')
+
+
+def test_basic_metrics(provider):
+    """ Basic Metrics availability test
+        This test checks that the Metrics service is up
+        Opens the hawkular status page and checks if it's up
+        """
+    assert provider.check_metrics(provider)
--- a/cfme/containers/provider/__init__.py
+++ b/cfme/containers/provider/__init__.py
@@ -5,6 +5,7 @@ from cfme.web_ui import (
 )
 from cfme.web_ui.menu import nav
 from cfme.web_ui.tabstrip import TabStripForm
+from utils import conf
 from utils import deferred_verpick, version
 from utils.browser import ensure_browser_open
 from utils.db import cfmedb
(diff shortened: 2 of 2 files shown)


dajoRH:
*CFME QE Bot*

dajoRH:
Lint report for commit a6e2b5c4ac910ee5a1eedf974f75c48496068593:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

cfme/tests/containers/test_basic_metrics.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def test_basic_metrics(provider):
+    """ Basic Metrics availability test
+        This test checks that the Metrics service is up
+        Opens the hawkular status page and checks if it's up
+        """
+    hostname = 'https://' + conf.cfme_data.get('management_systems', {})[provider.key]\
+        .get('hostname', []) + '/hawkular/metrics'
+    ensure_browser_open()
```
psav: This shouldn't be called inside a test ever

psav: It probably makes sense to make a method on the provider object called ```check_metrics``` or something like that

psav: Or you could use the browser fixture inside another fixture to obtain the right browser location

pavelzag: I could add it to a provider object but I fear it would take more time to go through approval :D

mfalesni: That is the point of the approval, that the code is right ...

pavelzag: Yes.. I guess

jkrocil:
Sorry for jumping in but:
```
Basic Metrics availability test
  This test checks that the Metrics service is up
  Opens the hawkular status page and checks if it's up
```
What does this have to do with ManageIQ testing at all? Am I missing something here?
If you are writing a test against ManageIQ and you want to make sure that your infrastructure is running, you can check that as part of a fixture for the real test. But I don't see the "real test" anywhere.
What I see here is just a fixture checking environment's availability wrapped as a test and that's it.

pavelzag:
The Metrics part of the OSE is a part that breaks often. We'd like to have heads up every time sometimes in deployment changes and the Metrics system is down. We wasted a lot of time in the past to find what went wrong and I proposed that we'd have a simple test that checks if Hawkular is up and Metrics is working.
It's a super simple test but believe me it'll spare us majore headaches going forward

pavelzag:
Added a check_metrics function to provider object. Please let me know if it's better

dajoRH:
Lint report for commit 634f5a1840dabc98a57e95eb83a329ac52b28bd5:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

jkrocil:
@pavelzag #3410 alredy got merged so what is the point of this anyway? Like I said before, this can be a fixture checking environment availability right before the test itself but thats about it.

pavelzag:
@jkrocil This change adds check_metrics function in provider's __init__.py per request. 
I'd like this to remain a test so I can see in Jenkins' reports if Metrics is up or down

jkrocil:
But this functionality is already present in master, see here: https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/containers/test_basic_metrics.py
I prefer to have this code as a test the way you've already merged it rather than having it as part of provider.

pavelzag:
@jkrocil I have no problems with that. Thanks

mfalesni:
*The Metrics part of the OSE is a part that breaks often.* - I would not say that in public :D

pavelzag:
True, let's give it a codename to confuse the enemy. :D

Issue #123, opened 2013-08-19
Move Utilization Tests to CFME_Tests


jwadkins:
This test was deleted from CFME_Pages, by the commit below, because it is a destructive test. It should go in CFME_Tests instead.

https://github.com/seandst/cfme_pages/commit/62f11f676e3561eaaa8dd725719932c3ced294ed

This also had the following issues opened in CFME_Pages:

https://github.com/RedHatQE/cfme_pages/issues/150
https://github.com/RedHatQE/cfme_pages/issues/149

PR #3415, opened 2016-09-13
Containers summary test3

pavelzag:
Purpose or Intent
=================

or

or

or

--


pavelzag:
Bad bad bad

PR #3424, opened 2016-09-14
Fix selector for vm ownership for upstream

otsuman:
{{pytest: cfme/tests/cloud_infra_common/test_vm_ownership.py -k test_form_button_validation --use-provider vsphere55 }}

### DIFF
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -32,8 +32,8 @@ pwr_btn = partial(toolbar.select, "Power")
 retire_remove_button = "//span[@id='remove_button']/a/img|//a/img[contains(@src, '/clear')]"
 
 set_ownership_form = Form(fields=[
-    ('user_name', Select("//select[@id='user_name']")),
-    ('group_name', Select("//select[@id='group_name']")),
+    ('user_name', AngularSelect('user_name')),
+    ('group_name', AngularSelect('group_name')),
     ('create_button', form_buttons.save),
     ('reset_button', form_buttons.reset),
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit 8c92dfb88a44908f61d8fa771fcc58ac8be47793:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

cfme/common/vm.py
```diff
@@ -32,8 +32,8 @@
 retire_remove_button = "//span[@id='remove_button']/a/img|//a/img[contains(@src, '/clear')]"
 
 set_ownership_form = Form(fields=[
-    ('user_name', Select("//select[@id='user_name']")),
-    ('group_name', Select("//select[@id='group_name']")),
+    ('user_name', AngularSelect('user_name')),
```
psav: Does this still work in the old versions, if not, we need to use a verpick

otsuman: @psav but does we support old versions? I thought for now we supporting 5.6+ and upstream for tests

psav: That's what I mean, we support 5.6+ (is AngularSelect present in 5.6?)

otsuman: @psav ci/downstream-56z — success. I think AngularSelect works with 5.5+ versions 
https://github.com/ManageIQ/integration_tests/blob/master/cfme/cloud/provider/__init__.py#L64

psav: Right I'm merging 5.7 failure is due to another issue

PR #3448, opened 2016-09-19
Fix Chargeback Rate API tests

mkoura:
{{pytest: '''cfme/tests/intelligence/test_chargeback.py::TestRatesViaREST'' -v'}}

Fix reflecting changes in the API.
See https://bugzilla.redhat.com/show_bug.cgi?id=1366019

### DIFF
--- a/cfme/rest/__init__.py
+++ b/cfme/rest/__init__.py
@@ -168,7 +168,7 @@ def rates(request, rest_api):
     chargeback = rest_api.collections.chargebacks.get(rate_type='Compute')
     data = [{
         'description': 'test_rate_{}_{}'.format(_index, fauxfactory.gen_alphanumeric()),
-        'rate': 1,
+        'source': 'allocated',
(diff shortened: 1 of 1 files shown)


cfme/rest/__init__.py
```diff
@@ -168,7 +168,7 @@ def rates(request, rest_api):
     chargeback = rest_api.collections.chargebacks.get(rate_type='Compute')
     data = [{
         'description': 'test_rate_{}_{}'.format(_index, fauxfactory.gen_alphanumeric()),
-        'rate': 1,
```
psav: What did this ```rate``` used to represent?

mkoura: I guess it corresponded to settings in UI where you can specify Fixed rate or Variable rate.
As described in https://github.com/ManageIQ/manageiq/issues/10501 current API `/rates` does not support tiered rates at the moment and therefore "is hardly usable".

PR #3451, opened 2016-09-19
Added Middleware Destinations

psav:
Purpose or Intent
=================

Extending Middleware Providers to provider navmazing functionality

@hhovsepy - this is going to be the new navigation system moving forward, we will provide full documentation and training shortly

{{pytest: cfme/tests/middleware/test_middleware_provider.py --use-provider=hawkular}}

### DIFF
--- a/cfme/middleware/provider/__init__.py
+++ b/cfme/middleware/provider/__init__.py
@@ -1,12 +1,17 @@
+from navmazing import NavigateToSibling, NavigateToAttribute
+
+from cfme.common.provider import import_all_modules_of
+from cfme.common.provider import BaseProvider
 from cfme.fixtures import pytest_selenium as sel
 from cfme.web_ui import (
-    Region, Form, AngularSelect, Input, Quadicon, form_buttons
+    Region, Form, AngularSelect, Input, Quadicon, form_buttons, toolbar as tb, paginator
 )
 from cfme.web_ui.menu import nav
-from cfme.common.provider import BaseProvider
 from utils import version
+from utils.appliance import CurrentAppliance
+from utils.appliance.endpoints.ui import navigator, CFMENavigateStep
 from utils.db import cfmedb
-from cfme.common.provider import import_all_modules_of
+
 from .. import cfg_btn, mon_btn, pol_btn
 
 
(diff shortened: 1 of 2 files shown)


dajoRH:
Lint report for commit 1e8b37cbe5e82aeb9e629cc67e1b7b8d7c2e3ae8:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

psav:
Failures all seem unrelated to navigation destinations

cfme/middleware/provider/__init__.py
```diff
@@ -82,6 +87,80 @@ class MiddlewareProvider(BaseProvider):
...
+@navigator.register(MiddlewareProvider, 'EditFromDetails')
+class EditFromDetails(CFMENavigateStep):
+    prerequisite = NavigateToSibling('Details')
+
+    def step(self):
+        cfg_btn('Edit this Middleware Provider')
+
+
+@navigator.register(MiddlewareProvider, 'EditTags')
+class EditTags(CFMENavigateStep):
```
hhovsepy: what if we keep 'Edit Tag' of provider in Taggable?

psav: hmm......just thinking if that works. I think the only thing would need would be fo the provider to support the "Details" destination, then we should be able to make it work as the MRO walk whould pick it up if we use NavigateToSibling

hhovsepy:
new PR sent to fix failing tests: https://github.com/ManageIQ/integration_tests/pull/3456

cfme/middleware/provider/__init__.py
hhovsepy: @psav will you guys do that changes on all pages or it is on each provider team?

psav: We will do them if you want us to, it's us who is requesting the change, please let me know ASAP which you'd prefer

hhovsepy: @psav please go on with changes

cfme/middleware/provider/__init__.py
```diff
@@ -82,6 +87,80 @@ class MiddlewareProvider(BaseProvider):
...
+        sel.uncheck(paginator.check_all())
+
+
+@navigator.register(MiddlewareProvider, 'Add')
+class Add(CFMENavigateStep):
+    prerequisite = NavigateToSibling('All')
+
+    def step(self):
+        cfg_btn('Add a New Middleware Provider')
+
```
mshriver: Not seeing middleware on my appliance so I can't answer my own question - is 'Discover' a valid destination here, even though it may not be used in existing tests?

cfme/middleware/provider/__init__.py
```diff
@@ -82,6 +87,80 @@ class MiddlewareProvider(BaseProvider):
...
+@navigator.register(MiddlewareProvider, 'EditTagsFromDetails')
+class EditTagsFromDetails(CFMENavigateStep):
+    prerequisite = NavigateToSibling('Details')
+
+    def step(self):
+        pol_btn('Edit Tags')
+
+
+@navigator.register(MiddlewareProvider, 'TimelinesFromDetails')
```
mshriver: In cloud provider I just called this destination 'Timelines', since there was only one way to navigate there (from details).

That being said, I just want to be consistent, I don't have strong feelings about doing it either way, and there is definitely some value to the path being described by the destination name. Runs the risk of ending up with a name that's too long, but we can address that as we find it.

cfme/middleware/provider/__init__.py
psav: There is no discover

cfme/middleware/provider/__init__.py
psav: Yeh, I was thinking this too, I think you should be able to get to Timelines, without Details, granted I didn't look, but you 'used' to be able to.

cfme/middleware/provider/__init__.py
mshriver: @psav We may be able to extend navigation to the sub-pages for Details in a similar way, if there is (or we add) a class structure equivalent to Taggable.

^ by subpages, I mean those reached through the accordion tree, not through the pol, cfg, or mon buttons.

cfme/middleware/provider/__init__.py
mshriver: I'm only seeing the mon button on the details screen. cfg, pol, and auth on 'All' screens.

The naming is at its root rehashing the argument of whether to put 'Edit Tag' in Taggable.  How to extend these navigation destinations that are 2, 3 or even 4 levels deep.

psav: Right, I'm OK for now with leaving them here, and trying the Taggable thing later on, next week perhaps, this is all iterative remember navmazing and its implementation are still in their infancy.

izapolsk: The navigation approach in current state is easy and covers 90-95% of our needs.
So, let's focus on it now.

I also need a manner to specify path/chain of actions how to get to some particular component. But we can think over this later.

PR #3497, opened 2016-09-26
Test hosts/nodes users and groups

rrasouli:
Purpose or Intent
=================
Testing the security section on infra provider nodes that currently includes users and groups

### DIFF
--- a/cfme/tests/openstack/test_host_security.py
+++ b/cfme/tests/openstack/test_host_security.py
@@ -1,0 +1,33 @@
+import pytest
+from utils import testgen
+from cfme.web_ui import Quadicon, toolbar
+from cfme.infrastructure.host import Host
+from cfme.web_ui import InfoBlock
+from cfme.fixtures import pytest_selenium as sel
+from cfme.configure.tasks import is_host_analysis_finished
+from utils.wait import wait_for
+
+
+pytest_generate_tests = testgen.generate(testgen.provider_by_type,
+                                         ['openstack-infra'],
+                                         scope='module')
+
+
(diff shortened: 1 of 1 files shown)


rrasouli:
@dajoRH please merge

dajoRH:
Lint report for commit bf417c9cd977651958d92fc5492f49c38e201c72:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

dajoRH:
Response to comment #249517273:

It seems like all the basic requirements were met, I have therefore merged your PR.

cfme/tests/openstack/test_host_security.py
```diff
@@ -0,0 +1,33 @@
...
+from utils.wait import wait_for
+
+
+pytest_generate_tests = testgen.generate(testgen.provider_by_type,
+                                         ['openstack-infra'],
+                                         scope='module')
+
+
+@pytest.mark.usefixtures("setup_provider_modscope")
+def test_host_security(provider, soft_assert):
```
mfalesni: This test does not do what it is supposed to do by its name. It says it checks *security* but in fact it just checks if SSA reveals any user and any group, so sorry but this is wrong.

rrasouli: @mfalesni I can understand the name is confusing, however the test purpose to check that users and groups appear under the security section, indeed after SSA

mfalesni: Wouldn't it make more sense to unify such tests in a test file called eg. test_host_summary where you would check various UI values?

PR #3500, opened 2016-09-26
Update cfme.common.provider with navmazing

mshriver:
{{pytest: -k test_provider_crud }}
Purpose or Intent
=================

Integrate navmazing into the common provider class.

@pavelzag has tested this patch with additional updates where the common create is called.

### DIFF
--- a/cfme/common/provider.py
+++ b/cfme/common/provider.py
@@ -18,6 +18,7 @@ from utils import conf
 from utils import version
 from utils.api import rest_api
 from utils.appliance import Navigatable
+from utils.appliance.endpoints.ui import navigate_to
 from utils.browser import ensure_browser_open
 from utils.db import cfmedb
 from utils.log import logger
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit 7e28dc15992828ba097c7869a9afc5f06718575d:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

cfme/common/provider.py
```diff
@@ -437,15 +433,7 @@ def _on_detail_page(self):
 
     def load_details(self, refresh=False):
```
psav: What are you doing with refresh here? We hsould still honour the ask to refresh until we remove it.

mshriver: Bad assumption on my part that the navigate_to() would inadvertently refresh - if am_i_here() is defined, it may or may not. Fixing

dajoRH:
Lint report for commit 9669de17be0eeedc9ffa2529bbb266854edd3d8c:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

PR #3525, opened 2016-09-28
navigation to cloud provider volumes

rrasouli:
Adding a method to navigate to cloud -> provider -> volumes page

### DIFF
--- a/cfme/common/provider.py
+++ b/cfme/common/provider.py
@@ -582,6 +582,7 @@ def get_paginator_value():
 class CloudInfraProvider(BaseProvider, PolicyProfileAssignable):
     vm_name = ""
     template_name = ""
+    volume_name = "Cloud volumes"
     detail_page_suffix = 'provider'
     edit_page_suffix = 'provider_edit'
     refresh_text = "Refresh Relationships and Power States"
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit 121c3a85b5ba0727c8bfa1fd77ae8a8cc44fb911:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

cfme/common/provider.py
```diff
@@ -582,6 +582,7 @@ def get_paginator_value():
 class CloudInfraProvider(BaseProvider, PolicyProfileAssignable):
     vm_name = ""
     template_name = ""
+    volume_name = "Cloud volumes"
```
psav: There is no need for this here, the only reason you'd want to do this, is if the method was on the BaseProvider and you wanted to add specific volume name that differed between the various sub classes. Plus, it seems it is only relevant for Cloud, judging by the naming, in which case, putting it in the cloud module would be better. Having said all of this, I think we want to make this a navmazing destination, @mshriver what say you?

mshriver: Agreed - This looks like it should be in Cloud, and it looks like the intent of the function using it is solely to navigate to Cloud Volumes via the details page for a specific provider.

I do think it would be best to represent these pages as destinations, though it may mean many destinations for an object. Ideally we want it version picked as well since we've seen the InfoBox rows change, for example 'Images' to 'Container Images'.

If not a navmazing destination, then a load_<name>() function similar to load_details, that uses the base Details destination and then interacts with the InfoBox - again only relevant for Cloud providers for this specific instance.

psav:
This PR will not be merged in its current state

mshriver:
@rrasouli @psav I'm working on a replacement for this.

psav:
Closed, will be implemented by @mshriver with navmazing

PR #3546, opened 2016-09-30
removing invalid on_hosts and on_vms methods 

izapolsk:
on_hosts and on_vms methods don't work now because those compare summary when browser window even doesn't appear.


dajoRH:
Lint report for commit db02b2f1ff445c03f9de6c67abfef94aef60d54b:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

PR #3595, opened 2016-10-06
added darga-41 version string to configure method picker.

amavinag:
* added darga-41 version string to configure method picker as darga41 needs to be configured as upstream appliance else it defaults to configure_cfme().


dajoRH:
Lint report for commit c098c421074f1a0b369e6635606330f32a02dab7:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

PR #3649, opened 2016-10-17
disabled marionette for local Firefox

izapolsk:
Purpose or Intent
=================
I'm about to disable marionette since selenium 3+ uses it by default for local Firefox. 
Such a param will be added to desired capabilities of default env.yaml


dajoRH:
Lint report for commit a6e3d2499c8c52e60118317ba2e270f25ede14f6:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

izapolsk:
selenium will be pin to former version instead of fixing this

PR #3652, opened 2016-10-17
Update web_ui.expression_editor, make count_form key an exact match

mshriver:
Purpose or Intent
=================

The expression editor was finding ">=" in the selector before finding ">" when ">" was provided as the key, resulting in unexpected filter results.

Use the exact argument to change the behavior to find an exact match, preventing selection of the wrong key.

The tests I found to be using this have other failures present, so it will be difficult to get relevant PRT results. These test failures are being addressed in separate branches, and many of the tests fail in part because of the behavior addressed in this PR.

Impacts RHCFQE-823 + 998

### DIFF
--- a/cfme/web_ui/expression_editor.py
+++ b/cfme/web_ui/expression_editor.py
@@ -166,7 +166,7 @@ count_form = Form(
     fields=[
         ("type", AngularSelect("chosen_typ")),
         ("count", AngularSelect("chosen_count")),
-        ("key", AngularSelect("chosen_key")),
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit 0c42a3f0e3a4b1a78b4e7dea87160351a3c5e82f:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

mshriver:
upstream failure in run 7835 is unrelated to PR.

PR #3656, opened 2016-10-18
FIX: cryptography fails to build w/ AssertionError

anewmanRH:
Defect in pycparser 2.14/2.15 causes cryptography build to fail with
"AssertionError: sorry, but this version only supports 100 named groups"


dajoRH:
Lint report for commit b9914a32601a3260d26403e765fcc783db860e1b:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

psav:
@anewmanRH I didn't see an issue here, I pip updated, removed cryptography and reinstalled it with the --no-cache-dir option, forcing a rebuild, no error here.

PR #3660, opened 2016-10-19
FIX: 'Getting Started' instructions

anewmanRH:
Using a clean RHEL 7.2 install, the directions in the current 'Getting Started' produce 2 errors.

The cryptography package fails to build w/ AssertionError, and the docker package has an issue w/ pip.

This PR updates pip the address the docker package issue, and sets the 'no-cache-dir' when installing deps for force rebuild.

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -30,7 +30,8 @@ file, place it in the ``cfme_tests`` repository (along ``conftest.py``):
 
   . ./.cfme_tests/bin/activate
   python scripts/disable-bytecode.py
-  PYCURL_SSL_LIBRARY=nss pip install -Ur ./requirements.txt
+  pip install --upgrade pip
+  PYCURL_SSL_LIBRARY=nss pip install -Ur ./requirements.txt --no-cache-dir
   echo "Run '. ./.cfme_tests/bin/activate' to load the virtualenv"
 
 Detailed steps (manual environment setup):


dajoRH:
Lint report for commit 9be3d82d7c81bade0704d792253f1a6b767bcaea:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

docs/getting_started.rst
```diff
@@ -30,7 +30,8 @@ file, place it in the ``cfme_tests`` repository (along ``conftest.py``):
 
   . ./.cfme_tests/bin/activate
   python scripts/disable-bytecode.py
-  PYCURL_SSL_LIBRARY=nss pip install -Ur ./requirements.txt
+  pip install --upgrade pip
+  PYCURL_SSL_LIBRARY=nss pip install -Ur ./requirements.txt --no-cache-dir
```
psav: No no, I'm not saying you _should_ put --no-cache-dir, I'm saying that when I ran it, to ensure I wasn't using a cahced version, ie to force a rebuild, that's what I did, the upgrade pip is a good idea, but as this is a getting started guide, wouldn't one expect them to have a fresh pip anyway?

anewmanRH: For 'no-cache-dir', since this would only possibly increase the time for the initial setup and could resolve possible issues beforehand for people less experienced or whom skim the instructions. I still think it should be added.

Then for the pip part. I just tested this since I did not have useful data.

- New untouched venv  pip version - 1.4.1

- Post pip upgrade version - 8.1.2

- System pip version - 7.1.0

PR #3695, opened 2016-10-26
Fixing wt shortcut

psav:
Purpose or Intent
=================

wt was assign as the current wt instance to be used during the pre_navigate
and go functions of the CFMENavigateStep. The problem with this is that
after a quit() the widgetastic instance gets replaced, but this replacement
doesn't get propagated to the wt attribute.


dajoRH:
Lint report for commit d1fc63c2f9e578b848dac7ac1f3c7cb1d7b97c9f:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

PR #3705, opened 2016-10-28
Fixed login for force navigate

psav:
Purpose or Intent
=================

* Force navigate (which will be gone soon) calls the login function on
  the current_user. If we are already logged in, this should not log us
  out.
* This was happening because the navigate_to for hte LoginScreen was
  placed BEFORE the check to see if we were already logged in as the
  current user.
* This obviously then led to a logout happening to get us to hte
  LoginScreen, from where a normal login occured.

### DIFF
--- a/cfme/login.py
+++ b/cfme/login.py
@@ -141,9 +141,6 @@ def login(user, submit_method=LOGIN_METHODS[-1]):
         RuntimeError: If the login fails, ie. if a flash message appears
     """
     # Circular import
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit b93fca9ed8d60ae74cf713678ebfc51c7cef3113:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

PR #3706, opened 2016-10-28
added timestamp to console logs

izapolsk:
1. We don't setup rootLogger. So, it is setup and inherited from different packages. F.e. I see such issue when use get_crud. Stdout/stderror is formatted by formatter definded in cinder client package.
2. From time to time I need to know how much time was some command run or measure the time between two commands.
So, I added default logging format in order to fix such issues.


dajoRH:
Lint report for commit 94af30368b4c166c051406b260541729458644f6:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

PR #3721, opened 2016-10-31
Fix uncollect

psav:
Purpose or Intent
=================

So, there was an issue where uncollection would mess up paralleizer if the slave was restarted after some tests passed earlier in the run. This fixes that.

### DIFF
--- a/fixtures/parallelizer/remote.py
+++ b/fixtures/parallelizer/remote.py
@@ -199,6 +199,7 @@ if __name__ == '__main__':
     parser = argparse.ArgumentParser()
     parser.add_argument('slaveid', help='The name of this slave')
     parser.add_argument('base_url', help='The base URL for this slave to use')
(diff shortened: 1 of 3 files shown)


dajoRH:
Lint report for commit c0f181d3f0b0a1d12ced49240a008fb8821ca385:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

fixtures/parallelizer/remote.py
```diff
@@ -199,6 +199,7 @@ def _init_config(slave_options, slave_args):
     parser = argparse.ArgumentParser()
     parser.add_argument('slaveid', help='The name of this slave')
     parser.add_argument('base_url', help='The base URL for this slave to use')
+    parser.add_argument('ts', help='The timestap to use for collections')
```
mshriver: typo, timestamp

PR #3755, opened 2016-11-08
[1LP][RFR]added ip and host_credentials check in cleanup scripts.

amavinag:
Purpose or Intent
=================
Credentials check is necessary when host_credentials are added to cfme_data, otherwise default to host_default credentials. Similarly, IP check is necessary for the providers in cleanup script, if IP is not reachable then there is no necessary to try invoking that provider.

### DIFF
--- a/scripts/cleanup_old_vms.py
+++ b/scripts/cleanup_old_vms.py
@@ -10,6 +10,7 @@ from dateutil import parser
 from threading import Lock, Thread
 from tzlocal import get_localzone
 
+from utils import net
 from utils.log import logger
 from utils.conf import cfme_data
 from utils.conf import credentials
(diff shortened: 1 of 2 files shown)


dajoRH:
Lint report for commit 0e509cb7124e571e8852dd596561fe2151f53be0:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

scripts/cleanup_old_vms.py
```diff
@@ -147,6 +149,9 @@ def cleanup_vms(texts, max_hours=24, providers=None, prompt=True):
     matchers = [re.compile(text) for text in texts]
 
     for provider_key in providers:
+        ipaddress = cfme_data['management_systems'][provider_key].get('ipaddress', None)
```
mshriver: :+1:

scripts/sync_template_tracker.py
```diff
@@ -22,6 +22,9 @@ def main(trackerbot_url, mark_usable=None):
     unresponsive_providers = set()
     # Queue up list_template calls
     for provider_key in all_providers:
+        ipaddress = cfme_data['management_systems'][provider_key].get('ipaddress', None)
```
psav: Won't this prevent us from tracking templates on thins like azure and ec2? as they don't have IP addresses? Though, I don't think we track those any way do we?

psav: However the comment may be related to the above usage

PR #3763, opened 2016-11-09
Don't inc tries on patching

psav:
Purpose or Intent
=================

When patching we shouldn't inc the tries, this was leading to problems by tries running out during navigation if both patches needed applying.


dajoRH:
Lint report for commit 122aeaf6c0a2f24c931da15604837ffee0b1b036:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

PR #3765, opened 2016-11-09
[WIP]Replacing is_displayed property by assert_is_displayed() method

quarckster:
Purpose or Intent
=================

I propose to replace `is_displayed` boolean property by `assert_is_displayed()` method. `assert_is_displayed()` should be just a bunch of assertions. It's much easier to catch an error if you know in which place it occurred. Check an example of the implementation in this commit 06cb408

### DIFF
--- a/utils/appliance/implementations/ui.py
+++ b/utils/appliance/implementations/ui.py
@@ -137,9 +137,11 @@ class CFMENavigateStep(NavigateStep):
 
     def am_i_here(self):
         try:
-            return self.view.is_displayed
-        except (AttributeError, NoSuchElementException):
+            self.view.assert_is_displayed()
+        except (AttributeError, NoSuchElementException, AssertionError):
             return False
+        else:
+            return True
 
     def pre_navigate(self, _tries=0, *args, **kwargs):
         if _tries > 2:


dajoRH:
Lint report for commit cb5d24c1be810187fdde872dd228eb800aa354ea:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

utils/appliance/implementations/ui.py
```diff
@@ -137,9 +137,11 @@ def create_view(self, *args, **kwargs):
 
     def am_i_here(self):
         try:
-            return self.view.is_displayed
-        except (AttributeError, NoSuchElementException):
+            self.view.assert_is_displayed()
```
mfalesni: `return self.view.assert_is_displayed()`

I reckon this would also require changing other views to have this method. Or add a fallback here (first try the assert_is_displayed, then is_displayed) ...

quarckster: Yes, views should be changed, because `is_displayed` tell you nothing about why view is not displayed. But if we have both `is_displayed` and `assert_is_displayed()` it will double the code.

mfalesni: The problem is that there is a convention about that widgets (of which view are a subclass) provide `is_displayed` ... I was thinking, maybe we could create a decorator that would be supplied instead of property for the is_displayed that would unify this.

PR #3810, opened 2016-11-20
list existing ec2 instances in all the regions.

amavinag:
Purpose or Intent
=================
  * ec2 instances cost, this cleanup script is leveraged to list all the existing instances on ec2 across
    regions. If regex and max hours are provided then it deletes the instances matching the regex.


dajoRH:
Lint report for commit cd4d939d781ea470a430bd38bef71be71797ee91:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

PR #3841, opened 2016-11-30
Update utils.testgen.provider_by_type, add provider_keys

mshriver:
Purpose or Intent
=================

Prototype update to testgen.provider_by_type to add provider_keys arg.

Should probably be a unique function, wanted to prototype usage and concept first before duplicating all the behavior here for uncollect etc.

### DIFF
--- a/utils/testgen.py
+++ b/utils/testgen.py
@@ -274,7 +274,7 @@ def _uncollect_since_version(data, metafunc, required_fields):
     return False
 
 
-def provider_by_type(metafunc, provider_types, required_fields=None):
+def provider_by_type(metafunc, provider_types=None, provider_keys=None, required_fields=None):
     """Get the values of the named field keys from ``cfme_data.get('management_systems', {})``
 
     ``required_fields`` is special and can take many forms, it is used to ensure that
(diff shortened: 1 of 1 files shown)


utils/testgen.py
```diff
@@ -340,7 +341,9 @@ def pytest_generate_tests(metafunc):
     argvalues = []
     idlist = []
 
-    for provider in cfme_data.get('management_systems', {}):
+    providers = cfme_data.get('management_systems', {})
```
mshriver: this was for debugging and stepping, ignore.

mshriver:
This prototype is driven by comments from PR #3805, and a desire to generate a cloud views test on provider keys

dajoRH:
Lint report for commit be91303e26bf043395299987fadaf48d54868d6e:

Commit inspection report:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

utils/testgen.py
```diff
@@ -359,6 +362,10 @@ def pytest_generate_tests(metafunc):
             if not(prov_obj.type_tclass in provider_types or prov_obj.type_name in provider_types):
                 continue
 
+        if provider_keys is not None:
```
psav: I presume this was so that we could use a provider key in the test to target a specific provider? However this is not something we would want to do. The key could change or be absent entirely.

mshriver: Oh? I misunderstood the intent of your other comments though, where a test was scoped for type 'openstack' but we wouldn't want to run against all versions of openstack. This PR can probably be closed then, not the bullet we're looking for.

psav: Yes I think so

PR #3858, opened 2016-12-06
Fix test_vm_ownership.py

otsuman:
{{pytest: cfme/tests/cloud_infra_common/test_vm_ownership.py}}

### DIFF
--- a/cfme/web_ui/__init__.py
+++ b/cfme/web_ui/__init__.py
@@ -1957,8 +1957,7 @@ class BootstrapTreeview(object):
         See :py:meth:`expand_path` for more informations about synopsis.
         """
         node = self.expand_path(*path, **kwargs)
-        if node is not None:
-            sel.click(node)
+        sel.click(node)
         return node
 
     def read_contents(self, nodeid=None, include_images=False, collapse_after_read=False):
@@ -2062,6 +2061,8 @@ class BootstrapTreeview(object):
 
 @fill.method((BootstrapTreeview, Sequence))
 def _fill_bstree_seq(tree, values):
+    if not values:
+        return None
     try:
         if (isinstance(values[0][0], list) or isinstance(values[0][0], tuple)) \
                 and not isinstance(values[0][0], types.StringTypes):
--- a/cfme/web_ui/__init__.py
+++ b/cfme/web_ui/__init__.py
@@ -1957,7 +1957,8 @@ class BootstrapTreeview(object):
         See :py:meth:`expand_path` for more informations about synopsis.
         """
         node = self.expand_path(*path, **kwargs)
-        sel.click(node)
+        if node is not None:
+            sel.click(node)
         return node
 
     def read_contents(self, nodeid=None, include_images=False, collapse_after_read=False):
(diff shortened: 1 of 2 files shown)


dajoRH:
Lint report for commit ab3a0c944bd102ef8e51521757a603b8637ef3b5:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

cfme/web_ui/__init__.py
```diff
@@ -1957,7 +1957,8 @@ def click_path(self, *path, **kwargs):
         See :py:meth:`expand_path` for more informations about synopsis.
         """
         node = self.expand_path(*path, **kwargs)
-        sel.click(node)
+        if node is not None:
```
psav: Why is node None?

psav: What awesome bug did you find?

mshriver: This inherently changes the behavior, have you audited usages of click_path() to handle the case where the node returned is None? Before these callers wouldn't have had to account for that fact, since this would throw an exception if you tried to call sel.click(None).

mshriver: I think node=None is the root failure in Jenkins job 248, configure.test_access_control.test_rolename_required_error_validation.

mshriver: Specific to test_rolename_required_error_validation, the registered fill method for the BootstrapTree ends up calling click_path() on an empty list, so `node` ends up being None.

This change would only mitigate the failure. I think I'd like to update `_fill_bstree_seq` to return None when the values list is empty. Prototyping now.

dajoRH:
Lint report for commit a8d58893793769a646cdcd06c4819a6131d1a050:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

cfme/web_ui/__init__.py
```diff
@@ -2061,6 +2061,8 @@ def _find_in_tree(t, p=None):
 
 @fill.method((BootstrapTreeview, Sequence))
 def _fill_bstree_seq(tree, values):
+    if not values:
```
mshriver: :+1:

psav: Where are we trying to fill with a None set of values though???

otsuman: @psav 
https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/cloud_infra_common/test_vm_ownership.py#L25
https://github.com/ManageIQ/integration_tests/blob/master/cfme/configure/access_control.py#L406
When no need to change default values for product features

mshriver: This is where I considered setting the default for `Role.product_features` to `[(['Everything'], true)]` instead of `[]` to match the UI default and prevent the exception.

psav:
Can we explain failures?

dajoRH:
Lint report for commit 2d471523462d425bcd8536b8112d0a608282648b:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

PR #3304, opened 2016-08-16
[RFR]SSHClient supports nonroot users

mfalesni:
The only condition is paswordless sudo.

I also managed to smuggle in an optimization:

``` python
Filename: string_concat.py

Line #    Mem usage    Increment   Line Contents
================================================
     7     44.0 MiB      0.0 MiB   @profile
     8                             def use_strings():
     9     44.0 MiB      0.0 MiB       result = ''
    10     58.6 MiB     14.6 MiB       for thing in STRINGS:
    11     58.6 MiB      0.0 MiB           result += thing
    12     53.9 MiB     -4.8 MiB       return result

Filename: string_concat.py

Line #    Mem usage    Increment   Line Contents
================================================
    15     44.0 MiB      0.0 MiB   @profile
    16                             def use_list():
    17     44.0 MiB      0.0 MiB       result = []
    18     44.1 MiB      0.1 MiB       for thing in STRINGS:
    19     44.1 MiB      0.0 MiB           result.append(thing)
    20     49.2 MiB      5.1 MiB       return ''.join(result)
```

putting things in a list and then concatenating saves memory and string creation/deletion every time.

I also tried to benchmark it by time. I was quite surprised:

``` bash
$ python -m timeit -n 5 -r 1 -s 'import string_concat' 'string_concat.use_strings()'
5 loops, best of 1: 3.25 sec per loop
$ python -m timeit -n 5 -r 1 -s 'import string_concat' 'string_concat.use_list()'
5 loops, best of 1: 3.04 msec per loop
```

1000x faster in "controlled" conditions :D

PR #3867, opened 2016-12-09
[1LP][RFR] Changed configuration files to include gce instance to smoke tests.

amavinag:
Purpose or Intent
=================
  * clone_deploy is changed to specify ssh public key while appliance deployment on gce.
  * appliance/__init__: configure_gce method is included, this can be improvised though.
  * ssh.py is same as https://github.com/ManageIQ/integration_tests/pull/3304 but modified it to fix smoke test failures on non-root user login.

    https://github.com/ManageIQ/wrapanapi/pull/90 is related to this PR

### DIFF
--- a/scripts/clone_template.py
+++ b/scripts/clone_template.py
@@ -8,6 +8,7 @@ import sys
 import utils
 from utils.appliance import Appliance, IPAppliance
 from utils.conf import cfme_data
+from utils.conf import credentials as cred
 from utils.log import logger
 from utils.path import log_path
 from utils.providers import destroy_vm, get_mgmt
@@ -146,7 +147,10 @@ def main(**kwargs):
             deploy_args["allowed_datastores"] = provider_dict["allowed_datastores"]
     elif provider_type == 'scvmm':
         deploy_args["host_group"] = provider_dict["provisioning"]['host_group']
-
+    elif provider_type == 'gce':
+        deploy_args['ssh_key'] = '{user_name}:{public_key}'.format(
+            user_name=cred['ssh']['ssh-user'],
+            public_key=cred['ssh']['public_key'])
     # Do it!
     try:
         logger.info('Cloning {} to {} on {}'.format(deploy_args['template'], deploy_args['vm_name'],
@@ -182,7 +186,10 @@ def main(**kwargs):
                 app = IPAppliance(address=ip)
             else:
                 app = Appliance(kwargs['provider'], deploy_args['vm_name'])
-            app.configure()
+            if provider_type == 'gce':
+                app.configure_gce()
+            else:
+                app.configure()
             logger.info('Successfully Configured the appliance.')
     except Exception as e:
         logger.exception(e)
--- a/utils/appliance/__init__.py
+++ b/utils/appliance/__init__.py
@@ -241,6 +241,21 @@ class IPAppliance(object):
         self.restart_evm_service(log_callback=log_callback)
         self.wait_for_web_ui(timeout=1800, log_callback=log_callback)
 
+    # TODO: this method eventually needs to be moved to provider class..
+    @logger_wrap("Configure GCE IPAppliance: {}")
+    def configure_gce(self, log_callback=None):
+        self.wait_for_ssh(timeout=1200)
+        self.deploy_merkyl(start=True, log_callback=log_callback)
+        # TODO: Fix NTP on GCE instances.
+        # self.fix_ntp_clock(log_callback=log_callback)
+        self.enable_internal_db(log_callback=log_callback)
+        # evm serverd does not auto start on GCE instance..
+        self.start_evm_service(log_callback=log_callback)
+        self.wait_for_evm_service(timeout=1200, log_callback=log_callback)
+        self.wait_for_web_ui(timeout=1800, log_callback=log_callback)
+        self.loosen_pgssl(log_callback=log_callback)
+        self.wait_for_web_ui(timeout=1800, log_callback=log_callback)
+
     def seal_for_templatizing(self):
         """Prepares the VM to be "generalized" for saving as a template."""
         with self.ssh_client as ssh_client:
@@ -456,6 +471,16 @@ class IPAppliance(object):
     def coverage(self):
         return ui_coverage.CoverageManager(self)
 
+    def ssh_client_with_privatekey(self):
+        with open(conf_path.join('appliance_private_key').strpath, 'w') as key:
+            key.write(conf.credentials['ssh']['private_key'])
+        connect_kwargs = {
+            'hostname': self.hostname,
+            'username': conf.credentials['ssh']['ssh-user'],
+            'key_filename': conf_path.join('appliance_private_key').strpath,
+        }
+        return ssh.SSHClient(**connect_kwargs)
+
     @cached_property
     def ssh_client(self):
         """Creates an ssh client connected to this appliance
@@ -483,6 +508,14 @@ class IPAppliance(object):
             'container': self.container,
         }
         ssh_client = ssh.SSHClient(**connect_kwargs)
+        try:
+            ssh_client.get_transport().is_active()
+            logger.info('default appliance ssh credentials are valid')
+        except Exception as e:
+            logger.error(e)
+            logger.error('default appliance ssh credentials failed, trying establish ssh connection'
+                         ' using ssh private key')
+            ssh_client = self.ssh_client_with_privatekey()
         # FIXME: propperly store ssh clients we made
         store.ssh_clients_to_close.append(ssh_client)
         return ssh_client
--- a/utils/ssh.py
+++ b/utils/ssh.py
@@ -123,6 +123,10 @@ class SSHClient(paramiko.SSHClient):
     def is_container(self):
         return self._container is not None
 
+    @property
+    def username(self):
+        return self._connect_kwargs.get('username', None)
+
     def __repr__(self):
         return "<SSHClient hostname={} port={}>".format(
             repr(self._connect_kwargs.get("hostname")),
@@ -197,20 +201,45 @@ class SSHClient(paramiko.SSHClient):
             self.connect()
         return super(SSHClient, self).get_transport(*args, **kwargs)
 
-    def run_command(self, command, timeout=RUNCMD_TIMEOUT, reraise=False, ensure_host=False):
+    def run_command(
+            self, command, timeout=RUNCMD_TIMEOUT, reraise=False, ensure_host=False,
+            ensure_user=False):
+        """Run a command over SSH.
+
+        Args:
+            command: The command. Supports taking dicts as version picking.
+            timeout: Timeout after which the command execution fails.
+            reraise: Does not muffle the paramiko exceptions in the log.
+            ensure_host: Ensure that the command is run on the machine with the IP given, not any
+                container or such that we might be using by default.
+            ensure_user: Ensure that the command is run as the user we logged in, so in case we are
+                not root, setting this to True will prevent from running sudo.
+
+        Returns:
+            A :py:class:`SSHResult` instance.
+        """
         if isinstance(command, dict):
             command = version.pick(command)
-        logger.info("Running command `%s`", command)
+        logger.info("Parsing command `{command}`".format(command=command))
+        uses_sudo = False
         if self.is_container and not ensure_host:
             command = 'docker exec {} bash -c {}'.format(self._container, quote(
                 'source /etc/default/evm; ' + command))
-            logger.info("Actually running command `%s`", command)
-        template = '{}\n'
-        command = template.format(command)
 
-        output = ''
+        if self.username != 'root' and not ensure_user:
+            # We need sudo
+            command = 'sudo -i bash -c {command}'.format(command=quote(command))
+            uses_sudo = True
+
+        logger.info("Running command `{command}`".format(command=command))
+        command += '\n'
+
+        output = []
         try:
             session = self.get_transport().open_session()
+            if uses_sudo:
+                # We need a pseudo-tty for sudo
+                session.get_pty()
             if timeout:
                 session.settimeout(float(timeout))
             session.exec_command(command)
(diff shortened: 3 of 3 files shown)


dajoRH:
Lint report for commit eba904bb32a93120309502ce2083ca3d20e7c622:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

dajoRH:
Lint report for commit 9f1d9227d9312f3254d1dd34c42412a821efa31a:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

scripts/clone_template.py
```diff
@@ -146,7 +147,10 @@ def main(**kwargs):
             deploy_args["allowed_datastores"] = provider_dict["allowed_datastores"]
     elif provider_type == 'scvmm':
         deploy_args["host_group"] = provider_dict["provisioning"]['host_group']
-
+    elif provider_type == 'gce':
+        deploy_args['ssh_key'] = '{user_name}:{public_key}'.format(
+            user_name=cred['ssh']['ssh-user'],
+            public_key=cred['ssh']['publick_key'])
```
mshriver: is `'publick-key'` a typo?

utils/appliance/__init__.py
```diff
@@ -483,6 +497,21 @@ def ssh_client(self):
             'container': self.container,
         }
         ssh_client = ssh.SSHClient(**connect_kwargs)
+        try:
+            ssh_client.get_transport().is_active()
+            logger.error('default appliance ssh credentials are valid')
```
mshriver: Why is this an error in the log, sounds successful?

utils/appliance/__init__.py
```diff
@@ -483,6 +497,21 @@ def ssh_client(self):
...
+            logger.error('default appliance ssh credentials failed, trying establish ssh connection'
+                         ' using ssh private key')
+            with open(conf_path.join('appliance_private_key').strpath, 'w') as key:
+                key.write(conf.credentials['ssh']['private_key'])
+            connect_kwargs = {
+                'hostname': self.hostname,
+                'username': conf.credentials['ssh']['ssh-user'],
+                'key_filename': conf_path.join('appliance_private_key').strpath,
+            }
+            ssh_client = ssh.SSHClient(**connect_kwargs)
```
mshriver: Consider writing a small function for this since its fairly complex for an except clause. There are chances for more exceptions to occur here.

You could also make the log message at the start of this clause a warning, since you're trying a second connection. If the second fails, then elevate to 'error' level.

utils/ssh.py
```diff
@@ -197,20 +201,56 @@ def get_transport(self, *args, **kwargs):
...
+            timeout: Timeout after which the command execution fails.
+            reraise: Does not muffle the paramiko exceptions in the log.
+            ensure_host: Ensure that the command is run on the machine with the IP given, not any
+                container or such that we might be using by default.
+            ensure_user: Ensure that the command is run as the user we logged in, so in case we are
+                not root, setting this to True will prevent from running sudo.
+
+        Returns:
+            A :py:class:`SSHResult` instance.
+        """
```
mshriver: :+1:

utils/ssh.py
```diff
@@ -197,20 +201,56 @@ def get_transport(self, *args, **kwargs):
...
+        command_changed = False
+        uses_sudo = False
         if self.is_container and not ensure_host:
             command = 'docker exec {} bash -c {}'.format(self._container, quote(
                 'source /etc/default/evm; ' + command))
+            command_changed = True
+
+        if self.username != 'root' and not ensure_user:
+            # We need sudo
+            if ';' in command:
```
mshriver: There's some potential here for an escaped ';' in a command to break this parsing.

utils/ssh.py
```diff
@@ -197,20 +201,56 @@ def get_transport(self, *args, **kwargs):
...
+                container or such that we might be using by default.
+            ensure_user: Ensure that the command is run as the user we logged in, so in case we are
+                not root, setting this to True will prevent from running sudo.
+
+        Returns:
+            A :py:class:`SSHResult` instance.
+        """
         if isinstance(command, dict):
             command = version.pick(command)
         logger.info("Running command `%s`", command)
```
mshriver: Consider updating this to read "Parsing command: (...)" and then change 'actually running' below to state 'Running command' regardless of if command_changed is true or false

utils/ssh.py
```diff
@@ -219,20 +259,20 @@ def run_command(self, command, timeout=RUNCMD_TIMEOUT, reraise=False, ensure_hos
             while True:
                 if session.recv_ready:
                     for line in stdout:
-                        output += line
+                        output.append(line)
```
mshriver: :+1:

utils/ssh.py
```diff
@@ -241,11 +281,11 @@ def run_command(self, command, timeout=RUNCMD_TIMEOUT, reraise=False, ensure_hos
         except socket.timeout as e:
             logger.error("Command `%s` timed out.", command)
             logger.exception(e)
-            logger.error("Output of the command before it failed was:\n%s", output)
+            logger.error("Output of the command before it failed was:\n%s", ''.join(output))
```
mshriver: Switch to .format() usage?

utils/ssh.py
```diff
@@ -262,15 +302,16 @@ def cpu_spike(self, seconds=60, cpus=2, **kwargs):
             "for ((i=0; i<instances; i++)) do while (($(date +%s) < $endtime)); "
             "do :; done & done".format(seconds, cpus), **kwargs)
 
-    def run_rails_command(self, command, timeout=RUNCMD_TIMEOUT):
+    def run_rails_command(self, command, timeout=RUNCMD_TIMEOUT, **kwargs):
         logger.info("Running rails command `%s`", command)
-        return self.run_command('cd /var/www/miq/vmdb; bin/rails runner {}'.format(command),
-            timeout=timeout)
+        return self.run_command('/var/www/miq/vmdb/bin/rails runner {}'.format(command),
```
mshriver: Ensure current working directory does not effect the running/environment of rails since you're no longer CDing

utils/ssh.py
```diff
@@ -281,8 +322,15 @@ def put_file(self, local_file, remote_file='.', **kwargs):
             self.run_command('mv {} {}'.format(tempfilename, remote_file))
             return scp
         else:
-            return SCPClient(self.get_transport(), progress=self._progress_callback).put(
-                local_file, remote_file, **kwargs)
+            if not self.username != 'root':
```
mshriver: double negative is confusing - do you mean `if self.username == 'root'`?

scripts/clone_template.py
amavinag: @mshriver yes, it is.

utils/ssh.py
amavinag: :smile: :+1:

utils/ssh.py
```diff
@@ -262,15 +302,16 @@ def cpu_spike(self, seconds=60, cpus=2, **kwargs):
...
+            timeout=timeout, **kwargs)
 
-    def run_rake_command(self, command, timeout=RUNCMD_TIMEOUT):
+    def run_rake_command(self, command, timeout=RUNCMD_TIMEOUT, **kwargs):
         logger.info("Running rake command `%s`", command)
-        return self.run_command('cd /var/www/miq/vmdb; bin/rake {}'.format(command),
-            timeout=timeout)
+        return self.run_command(
+            '/var/www/miq/vmdb/bin/rake -f /var/www/miq/vmdb/Rakefile {}'.format(command),
+            timeout=timeout, **kwargs)
```
amavinag: @mshriver @psav "cd" is removed here too and invoking the rake by its absolute path.

utils/ssh.py
amavinag: @mshriver It should not affect, as we are invoking rails directly. why I removed "cd" is because it does not work with "sudo". 

@psav are you OK with this change?

utils/ssh.py
amavinag: @mshriver :+1: Agree

utils/appliance/__init__.py
amavinag: @mshriver again, this was not intended.

utils/appliance/__init__.py
amavinag: @mshriver Agree :+1: I can port that to new method and handle any new exceptions there.

dajoRH:
Lint report for commit c84283262dfc4a2cc045235edb05d641721069b7:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

utils/ssh.py
```diff
@@ -219,33 +255,34 @@ def run_command(self, command, timeout=RUNCMD_TIMEOUT, reraise=False, ensure_hos
...
+            logger.error("Command `{command}` timed out.".format(command=command))
             logger.exception(e)
-            logger.error("Output of the command before it failed was:\n%s", output)
+            logger.error("Output of the command before it failed was:\n{output}".format(
+                output=''.join(output)))
             raise
 
         # Returning two things so tuple unpacking the return works even if the ssh client fails
-        return SSHResult(1, None)
+        return SSHResult(1, '')
```
psav: I'm a little concerned about this change...it's so tiny it could just break things in new and interesting ways, what's the rationale?

utils/ssh.py
```diff
@@ -197,20 +201,52 @@ def get_transport(self, *args, **kwargs):
...
+            ensure_user=False):
+        """Run a command over SSH.
+
+        Args:
+            command: The command. Supports taking dicts as version picking.
+            timeout: Timeout after which the command execution fails.
+            reraise: Does not muffle the paramiko exceptions in the log.
+            ensure_host: Ensure that the command is run on the machine with the IP given, not any
+                container or such that we might be using by default.
+            ensure_user: Ensure that the command is run as the user we logged in, so in case we are
```
psav: What is the use case for needing this?

utils/ssh.py
```diff
@@ -262,30 +299,38 @@ def cpu_spike(self, seconds=60, cpus=2, **kwargs):
...
             self.run_command('mv {} {}'.format(tempfilename, remote_file))
             return scp
         else:
-            return SCPClient(self.get_transport(), progress=self._progress_callback).put(
-                local_file, remote_file, **kwargs)
+            if self.username == 'root':
+                return SCPClient(self.get_transport(), progress=self._progress_callback).put(
+                    local_file, remote_file, **kwargs)
+            # scp client is not sudo, may not work for non sudo
+            SCPClient(self.get_transport(), progress=self._progress_callback).put(
```
psav: Should we have a return here? Or the line after?

utils/appliance/__init__.py
```diff
@@ -241,6 +241,20 @@ def _configure_upstream(self, log_callback=None):
         self.restart_evm_service(log_callback=log_callback)
         self.wait_for_web_ui(timeout=1800, log_callback=log_callback)
 
+    @logger_wrap("Configure GCE IPAppliance: {}")
+    def configure_gce(self, log_callback=None):
```
psav: Can we put a message for a TODO in here, eventually we will want this moved out as a function/method that is in the provider class, appliance shouldn't know anything special about GCE (eventually)

utils/ssh.py
amavinag: @psav, By default we switch to "sudo" if the user is not root (at line# 229). "ensure_user " adds check if we want to run the command as the non-root user without adding sudo to it.

@mfalesni please mention if there is specific usecase you added it for in #3304

utils/ssh.py
amavinag: @psav, yes, will add return.

utils/ssh.py
```diff
@@ -262,30 +299,38 @@ def cpu_spike(self, seconds=60, cpus=2, **kwargs):
...
+            command=command), timeout=timeout, **kwargs)
 
-    def run_rake_command(self, command, timeout=RUNCMD_TIMEOUT):
-        logger.info("Running rake command `%s`", command)
-        return self.run_command('cd /var/www/miq/vmdb; bin/rake {}'.format(command),
-            timeout=timeout)
+    def run_rake_command(self, command, timeout=RUNCMD_TIMEOUT, **kwargs):
+        logger.info("Running rake command `{command}`".format(command=command))
+        return self.run_command(
+            '/var/www/miq/vmdb/bin/rake -f /var/www/miq/vmdb/Rakefile {command}'.format(
```
amavinag: @psav, same with rake command, removed "cd" here and using absolute path. and referencing Rakefile with -f option.

utils/ssh.py
```diff
@@ -262,30 +299,38 @@ def cpu_spike(self, seconds=60, cpus=2, **kwargs):
             "for ((i=0; i<instances; i++)) do while (($(date +%s) < $endtime)); "
             "do :; done & done".format(seconds, cpus), **kwargs)
 
-    def run_rails_command(self, command, timeout=RUNCMD_TIMEOUT):
-        logger.info("Running rails command `%s`", command)
-        return self.run_command('cd /var/www/miq/vmdb; bin/rails runner {}'.format(command),
-            timeout=timeout)
+    def run_rails_command(self, command, timeout=RUNCMD_TIMEOUT, **kwargs):
+        logger.info("Running rails command `{command}`".format(command=command))
+        return self.run_command('/var/www/miq/vmdb/bin/rails runner {command}'.format(
```
amavinag: @psav, are you OK with this change? removed "cd" here and using absolute path

psav: I _think_ so, I mean you tested it right?

utils/ssh.py
mfalesni: There is no specific usecase yet but I want to be sure that we can also use the login user if necessary and do not cut ourselves away from it.

dajoRH:
Lint report for commit a3a371d4c07e59012ef87220e795a3ee042f2d51:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

PR #3870, opened 2016-12-13
[1LP][RFR] Handle confirmation alert in host power controls

rrasouli:
Purpose or Intent
=================
host.power_off(), host.power_on() could not be invoked since there is a confirmation message.

Now with this change the host power controls handle confirm message and immediately invoke the power action

### DIFF
--- a/cfme/infrastructure/host.py
+++ b/cfme/infrastructure/host.py
@@ -226,12 +226,12 @@ class Host(Updateable, Pretty, Navigatable):
 
     def power_on(self):
         navigate_to(self, 'Details')
-        pow_btn('Power On')
+        pow_btn('Power On', invokes_alert=True)
         sel.handle_alert()
 
     def power_off(self):
         navigate_to(self, 'Details')
-        pow_btn('Power Off')
+        pow_btn('Power Off', invokes_alert=True)
         sel.handle_alert()
 
     def get_ipmi(self):


dajoRH:
Lint report for commit 45f249bd911ff14da9c791b28ae6a076e140aceb:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

mfalesni:
I would not call this PR "avoid" as you aren't technically avoiding the alert. You are just preparing the framework for the actual handle_alert

cfme/infrastructure/host.py
```diff
@@ -226,12 +226,12 @@ def execute_button(self, button_group, button, cancel=True):
 
     def power_on(self):
         navigate_to(self, 'Details')
-        pow_btn('Power On')
+        pow_btn('Power On', invokes_alert=True)
```
mshriver: Is the alert present in both CFME 5.6 and 5.7?

rrasouli: @mshriver this feature is new on 5.7 and RHOS10

psav: So doe we need to "version pick" behaviour dependant on 5.6 and 5.7 because the alert won't be present in 5.6?

rrasouli: @psav the functionality exist on 5.6, the test will fail probably in 5.6

mfalesni: I think this should not do much harm anyway.

PR #3872, opened 2016-12-13
adding order to the Provision Form filling

lkhomenk:
Purpose or Intent
=================

__Fixing__ order in Provisioning Form filling because it goes random based on dict hashes. which works for usual provisioning but fails on PXE provision. On PXE order is Network/Customize(which depends on  Custom's provision type)

{{pytest: cfme/tests/infrastructure/test_pxe_provisioning.py --use-provider vsphere55 }}

### DIFF
--- a/cfme/provisioning.py
+++ b/cfme/provisioning.py
@@ -249,7 +249,9 @@ provisioning_form = tabstrip.TabStripForm(
             # Infra
             ('stateless', ui.Input('schedule__stateless')),
         ])
-    ])
+    ]),
+    order=['Request', 'Purpose', 'Catalog', 'Environment', 'Hardware', 'Network', 'Properties',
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit 2cb5b83452d1e096b2fc7db6526bc47ceb6d1da6:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

psav:
tabfields is already an Ordered Dict. I thought that was enough?

lkhomenk:
@psav yes. tabfields are ordered but it doesn't work at all when we take tabstrip.fields during filling. "fields" attribute is affected by passed "order" key

psav:
I fixed this elsewhere on Friday ;) #3882

PR #3886, opened 2016-12-18
[WIP] Relationships table data integrity (provider)

jdemon519:
{{pytest: cfme/tests/containers/test_provider_add_token.py -v --use-provider cm-env2 }}

Added the delete function to make sure the test starts out with a clean slate.
Added wait time to provider.py

### DIFF
--- a/cfme/common/provider.py
+++ b/cfme/common/provider.py
@@ -310,7 +310,7 @@ class BaseProvider(Taggable, Updateable, SummaryMixin, Navigatable):
                      {'ui': ui},
                      message="do_stats_match_db",
                      num_sec=1000,
-                     delay=60)
+                     delay=120)
 
         client.disconnect()
 
--- a/cfme/common/provider.py
+++ b/cfme/common/provider.py
@@ -765,4 +765,3 @@ def import_all_modules_of(loc):
     path = project_path.join('{}'.format(loc.replace('.', '/'))).strpath
     for _, name, _ in pkgutil.iter_modules([path]):
         importlib.import_module('{}.{}'.format(loc, name))
-
--- a/cfme/common/provider.py
+++ b/cfme/common/provider.py
@@ -310,7 +310,7 @@ class BaseProvider(Taggable, Updateable, SummaryMixin, Navigatable):
                      {'ui': ui},
                      message="do_stats_match_db",
                      num_sec=1000,
-                     delay=120)
+                     delay=60)
 
         client.disconnect()
 
@@ -765,3 +765,4 @@ def import_all_modules_of(loc):
     path = project_path.join('{}'.format(loc.replace('.', '/'))).strpath
     for _, name, _ in pkgutil.iter_modules([path]):
         importlib.import_module('{}.{}'.format(loc, name))
+
--- a/cfme/containers/image.py
+++ b/cfme/containers/image.py
@@ -49,7 +49,7 @@ class All(CFMENavigateStep):
         nav._nav_to_fn('Compute', 'Containers', 'Container Images')(None)
 
     def resetter(self):
-        tb.select('Grid View')
+        tb.select('List View')
         sel.check(paginator.check_all())
         sel.uncheck(paginator.check_all())
 
--- a/cfme/containers/provider/__init__.py
+++ b/cfme/containers/provider/__init__.py
@@ -109,9 +109,8 @@ class ContainersProvider(BaseProvider, Pretty):
         'num_service',
         'num_replication_controller',
         'num_pod',
-        'num_node',
         'num_container',
-        'num_image']
+        'num_node']
     # TODO add 'num_image_registry' and 'num_volume'
     string_name = "Containers"
     page_name = "containers"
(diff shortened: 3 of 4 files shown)


dajoRH:
Lint report for commit fe7ed6f50afcae369ee91e855a097a27660937d9:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

cfme/common/provider.py
```diff
@@ -310,7 +310,7 @@ def validate_stats(self, ui=False):
                      {'ui': ui},
                      message="do_stats_match_db",
                      num_sec=1000,
-                     delay=60)
+                     delay=120)
```
psav: This will make tests potentially run slower because we are only checking every two minutes and not every minute to see if the match is complete. Is there a reason for this?

jdemon519: @psav True, but for some reason this solved the issue for CFME 5.7. The test ran fine on 5.6, then it failed on 5.7 due to timeout error. Changing the delay to 120 solved the problem. Maybe 80 would have been a better choice?

jdemon519: @psav Just tested it on 5.6  with the delay set to 60 and it passed, I'll give it another spin on 5.7 right now, just to be sure.

jdemon519: It  fails with a timeout error on 5.7. (when the delay was set to 120). I ran it without the UI and it fails on Images

provider.validate_stats(ui=False)

Any ideas on what might be causing this? It keeps restarting the API, but the num_image() fails every time on 5.7. 
@psav @mfalesni @jkrocil

psav: @jdemon519 so you need to look at what images are being returned by both wrapanapi and CFME. Obviously this didn't run on PRT so I can't debug, but the cfme.log should also show you the mismatch, ie.....```3 != 4``` or something like that. Then you need to find out if we need to modify wrapanapi to produce the right numbering/listing for CFME.

jdemon519: @psav Does num_image() check the number of images in the API of Openshift or CFME? Apparently, there is a new feature in CFME 5.7 that displays the number of images that is larger than the number of images in Openshift. For instance, when I check the Openshift, it shows me 87 images vs. 100 images in CFME. When I checked num_image() in iPython, I got 100 images as well. So there is a discrepancy here.

psav: Right, these come along from time to time and we need to see where the discrepancy lies.
So, if CFME gives 100 Images.....
if you do something like.....
```python
len(provider.mgmt.list_images())
```
what do you get?

psav: is images the right thing there for openshift? Maybe I chose the wrong wrapanapi function?

jdemon519: @psav No, it's actually the right function. Here is what Erez says: "Before: We only collected the images from the pods
Now: We collect the images from the pods AND also from Openshift. Images from Openshift have more information to them. If a pod is running an image that we can link to an Openshift image and we unite them."
So it's pulling images from Openshift but there are also additional images that are being added. The problem is. the validation we are doing right now is using this function:
    def list_image(self):
        """Returns list of images (derived from pods)"""
        entities = []
        entities_j = self.api.get('pod')[1]['items']
        for entity_j in entities_j:
            imgs_j = entity_j['status'].get('containerStatuses', [])
            for img_j in imgs_j:
                _, name, tag = self._parse_image_info(img_j['image'])
                img = Image(name, tag, img_j['imageID'])
                if img not in entities:
                    entities.append(img)
        return entities

But it only pulls the Openshift images, while ignoring the other ones. In CFME 5.6, it was just the Openshift images.

jdemon519:
Would it make sense to skip images for this kind of validation? Right now, it is one of the expected values in STATS_TO_MATCH. I can write a separate test for the images.

cfme/common/provider.py
psav: Ahh ok, so, we need to update wrapanapi to include a parameter to get the images from all places, then  modify the test to versionpick and either include that parameter in the list_images call, or not.

dajoRH:
Lint report for commit 40d9b209c980af076067e75f9756c4dafd55e86a:

`cfme/common/provider.py`:
- :heavy_exclamation_mark: Line 768:1: **W391** *blank line at end of file*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
Lint report for commit 2fb2dd3b917c358986167207c048b8a2a94f1e9f:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

psav:
If you prefer, the idea was that it made it easy to check because you basically didn't have to do anything extra to get that check, but by all means, I think you can just override the STATS_TO_MATCH for that one, up to you.

jdemon519:
Fixed https://github.com/ManageIQ/wrapanapi/pull/91
Overriding STATS_TO_MATCH is a good idea.

dajoRH:
Lint report for commit 859d7f74f8e1a926c89a0f682c4a4409b9e91653:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

PR #3889, opened 2016-12-20
Change provider type

sshveta:
Recent PR 3805 changed provider_type to provider .
While creating catalog item for vmware and Rhev provider field is not an option in UI .
Hence , provider_type should be there which is none for these providers.

### DIFF
--- a/cfme/tests/cloud/test_providers.py
+++ b/cfme/tests/cloud/test_providers.py
@@ -311,3 +311,42 @@ def test_openstack_provider_has_api_version():
     pytest.sel.wait_for_ajax()
     assert pytest.sel.is_displayed(
         prov.properties_form.api_version), "API version select is not visible"
+
+
+class TestProvidersRESTAPI(object):
+    @pytest.mark.tier(3)
+    @pytest.mark.uncollectif(lambda: version.current_version() < '5.7')
+    @pytest.mark.parametrize(
+        "from_detail", [True, False],
+        ids=["from_detail", "from_collection"])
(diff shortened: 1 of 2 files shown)


dajoRH:
Lint report for commit ba935ef6bb44aea7c508cf56a850a24355a41e25:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

cfme/services/catalogs/catalog_item.py
```diff
@@ -152,7 +152,7 @@ def create(self):
                                'select_catalog': catalog.name,
                                'select_dialog': dialog.name,
                                'select_orch_template': self.orch_template,
-                               'select_provider': self.provider,
```
psav: Just checking,

Before self.provider would have been a provider object no? Or perhaps I am thinking of another place. I remember recently we changed a very similar call so that it no longer took the ```provider``` and ```provider_type``` args, but just the ```provider``` object. This is much more preferable as then you can use ```provider.name``` or ```provider.type``` when needed.

sshveta: Yes but here we need the provider for some catalog items (like ec2 and openstack) and don't need it for other (like vmware and Rhev) . All of these catalog items also need provider to select template.
So I need to pass the provider object to select the template but also need to pass provider_type which will be none for vmware and Rhev and will have some value (same as provider ) for ec2 and openstack.

psav: Well then it's easy enough to do
```'select_provider': self.provider.type if self.provider else None```

PR #3905, opened 2016-12-25
[WIPTEST] Modified test_containers_properties.py to utilize API/allow for faster execution

jdemon519:
{{pytest: cfme/tests/containers/test_containers_properties.py -v --use-provider cm-env2 }}

Instead of checking each field in the properties table, the test makes calls to the relevant APIs


dajoRH:
Lint report for commit e8f507df2337f35a206c5941f9c47d1d9d592aac:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

PR #3909, opened 2016-12-30
[RFR] replaced force_navigate calls with navigate_to

izapolsk:
Purpose or Intent
=================
1. replaced force_navigate in several tests

{{pytest: --use-provider=vsphere55 cfme/tests/webui/test_splitter.py cfme/tests/intelligence/reports/test_menus.py }}

### DIFF
--- a/cfme/tests/webui/test_splitter.py
+++ b/cfme/tests/webui/test_splitter.py
@@ -1,18 +1,42 @@
 import pytest
-from cfme.web_ui.splitter import pull_splitter_left, pull_splitter_right
 from xml.sax.saxutils import quoteattr, unescape
+
+from cfme.base.ui import Server
+from cfme.cloud.instance import Instance
+from cfme.infrastructure.config_management import ConfigManager
+from cfme.infrastructure.datastore import Datastore
+from cfme.infrastructure.pxe import ISODatastore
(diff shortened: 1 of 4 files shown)


dajoRH:
Lint report for commit a46925d721b5016a15cd2ac38990d790c929c687:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

cfme/tests/webui/test_splitter.py
```diff
@@ -1,20 +1,44 @@
...
+def nav_to(location):
+    if isinstance(location, basestring):
+        pytest.sel.force_navigate(location)
+    else:
+        dest = 'All'
+        if location is Vm:
+            dest = 'VMsOnly'
+        elif location is Server:
+            dest = 'Dashboard'
+
```
mshriver: Is there a fallback if the destination doesn't exist for a location object? Ex if VMsOnly isn't implemented, do we want to fall back to Dashboard, or skip the test?

izapolsk: the test fails in such case and that's expected.

izapolsk:
all the issues in tests are fixed and green when I run them locally. But there can be some test failures in PRT due to overall test instability.

PR #3922, opened 2017-01-04
[WIPTEST] Cloud tenants CRUD 

rrasouli:
Purpose or Intent
=================

A draft 
Adding openstack cloud provider tenant CRUD functions.
Ignore the commit message 'fixed refresh function'

### DIFF
--- a/cfme/cloud/tenant.py
+++ b/cfme/cloud/tenant.py
@@ -7,27 +7,61 @@
 
 from utils.conf import cfme_data
 from cfme.fixtures import pytest_selenium as sel
-from cfme.web_ui import Region, SplitTable
-
+from cfme.web_ui import Region, SplitTable, toolbar as tb, CheckboxTable, Form, form_buttons
+from utils.appliance import Navigatable
+from utils.appliance.implementations.ui import navigator, CFMENavigateStep, navigate_to
+from navmazing import NavigateToAttribute, NavigateToSibling
+from cfme.cloud.instance import match_page
+from cfme.common import SummaryMixin, Taggable
 
 # Page specific locators
 list_page = Region(
     locators={
         'tenant_table': SplitTable(header_data=('//div[@class="xhdr"]/table/tbody', 1),
-            body_data=('//div[@class="objbox"]/table/tbody', 1))
+                                   body_data=('//div[@class="objbox"]/table/tbody', 1))
     },
     title='Cloud Tenants')
 
-
+create_form = Form([('cloud_provider', ("//button[@id='cloud_tenant_id']"),
+                     ('tenant_name', "//*[@name='name']"))])
+list_tbl = CheckboxTable(table_locator="//div[@id='list_grid']//table")
 details_page = Region(infoblock_type='detail')
 
 
-class Tenant(object):
+class Tenant(Taggable, SummaryMixin, Navigatable):
     def __init__(self, name, description, provider_key):
         """Base class for a Tenant"""
         self.name = name
         self.description = description
         self.provider_key = provider_key
+        Navigatable.__init__(self)
+
+    def create(self, provider_key):
+        """
+            create a new tenant
+        """
+        navigate_to(Tenant, 'All')
+        tb.select('Configuration', 'Create Cloud Tenant')
+        tenant_data = dict(tenant_name=self.name, provider_name=provider_key)
+        create_form.fill(tenant_data)
+        sel.click(form_buttons.save)
+
+    def update(self, tenant_name, new_name):
+        """
+            update a given cloud tenant name to new tenant
+        """
+        navigate_to(Tenant, 'All')
+        list_tbl.select_row('Name', tenant_name)
+        tb.select('Configuration', 'Edit Selected Cloud Tenant')
+        tenant_data = dict(tenant_name=new_name)
+        create_form.fill(tenant_data)
+        sel.click(form_buttons.save)
+
+    def delete(self, tenant_name):
+        navigate_to(Tenant, 'All')
+        list_tbl.select_row('Name', tenant_name)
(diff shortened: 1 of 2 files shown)


mshriver:
@rrasouli Please check out #3942, I didn't see your PR and picked up a JIRA card for adding these destinations. Your implementation is good, but is missing some things like paginator handling when there are >20 tenants, or the multiple delete/edit paths (from All, from Details).  Don't want to throw out all of either of our work - its definitely my fault for duplicating your effort though, sorry about that.

Let me know what you think about my changes, and thoughts on how to proceed. Maybe you can focus your efforts on growing the test_tenant test module and I can cover the navigation and object modeling?

rrasouli:
@mshriver I can see your PR and tried to run it, I wonder if we can merge both PR's
I would have split each CRUD test. update tenant is missing, as well as some navigation utilities

mshriver:
@rrasouli I saw your comments on my PR that got merged - I guess it wasn't quite ready but we can resolve that quickly. Do you want to rebase against master and improve the CRUD test and I'll tackle the exception I caused by not setting result when `cancel=True`?

rrasouli:
@mshriver I could not rebase my code here, perhaps we can merge the code in #3942

cfme/cloud/tenant.py
```diff
@@ -7,27 +7,61 @@
...
+        """
+            create a new tenant
+        """
+        navigate_to(Tenant, 'All')
+        tb.select('Configuration', 'Create Cloud Tenant')
+        tenant_data = dict(tenant_name=self.name, provider_name=provider_key)
+        create_form.fill(tenant_data)
+        sel.click(form_buttons.save)
+
+    def update(self, tenant_name, new_name):
```
mfalesni: Update has a special syntax.

```python
def create(self, updates):
    ...
    ...
    fill(create_form, updates)
    flash.assert_message('tenant {} was saved'.format(updates.get('name', self.name))
```

it is not exact with strings etc but it shows the usage

cfme/tests/openstack/cloud/test_tenant_crud.py
```diff
@@ -0,0 +1,34 @@
...
+    new_tenant.create(cloud_provider)
+    wait_for(provider.is_refreshed, [None, 10], delay=5)
+    sel.refresh()
+    assert new_tenant.exists(), "New cloud tenant hasn't been created"
+
+
+@pytest.mark.usefixtures("setup_provider_modscope")
+def test_update_cloud_tenant(cloud_tenant_name, new_tenant_name, cloud_provider):
+    my_tenant = Tenant(cloud_tenant_name, "", cloud_provider)
+    my_tenant.update(cloud_provider, new_tenant_name)
```
mfalesni:
```python
from utils.update import update

with update(my_tenant):
    tenant.name = 'foo'
    tenant.whatever = 'something'
```

By the way, why not having a complete crud test in one? Like we do it?

cfme/cloud/tenant.py
```diff
@@ -7,27 +7,61 @@
...
         self.description = description
         self.provider_key = provider_key
+        Navigatable.__init__(self)
+
+    def create(self, provider_key):
+        """
+            create a new tenant
+        """
+        navigate_to(Tenant, 'All')
+        tb.select('Configuration', 'Create Cloud Tenant')
```
mfalesni: You should use an Add location instead of All->click table

cfme/cloud/tenant.py
```diff
@@ -7,27 +7,61 @@
...
     locators={
         'tenant_table': SplitTable(header_data=('//div[@class="xhdr"]/table/tbody', 1),
-            body_data=('//div[@class="objbox"]/table/tbody', 1))
+                                   body_data=('//div[@class="objbox"]/table/tbody', 1))
     },
     title='Cloud Tenants')
 
-
+create_form = Form([('cloud_provider', ("//button[@id='cloud_tenant_id']"),
+                     ('tenant_name', "//*[@name='name']"))])
```
mfalesni: we have an Input class in cfme.web_ui

cfme/tests/openstack/cloud/test_tenant_crud.py
```diff
@@ -0,0 +1,34 @@
...
+import cfme.fixtures.pytest_selenium as sel
+from cfme.cloud.tenant import Tenant
+from utils.wait import wait_for
+
+
+@pytest.mark.usefixtures("setup_provider_modscope")
+def test_create_cloud_tenant(provider):
+    """
+        create a new cloud tenant
+    """
```
mfalesni: Can you follow some reasonable test descriptions?
Like this one https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/test_ui_problems.py#L26

cfme/cloud/tenant.py
```diff
@@ -7,27 +7,61 @@
...
         """Base class for a Tenant"""
         self.name = name
         self.description = description
         self.provider_key = provider_key
+        Navigatable.__init__(self)
+
+    def create(self, provider_key):
+        """
+            create a new tenant
+        """
```
mfalesni: Comments should look like this: https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments

cfme/cloud/tenant.py
```diff
@@ -38,3 +72,36 @@ def exists(self):
...
+    def resetter(self):
+        tb.select('List View')
+
+
+@navigator.register(Tenant, 'Details')
+class Details(CFMENavigateStep):
+    prerequisite = NavigateToSibling('All')
+
+    def step(self):
+        tb.select('List View')
```
mfalesni: This one should be as a resetter, shouldnt it?

rrasouli:
Will continue from #3942

PR #3931, opened 2017-01-04
[1LP][RFR] Service Catalog Navmazing

sshveta:
Some other files need to be modified because of this navamzing change like : test_provision_stack and test_service_catalogs , which have some other failures too . 
So fixing them up in a separate PR coming soon.

### DIFF
--- a/cfme/services/catalogs/service_catalogs.py
+++ b/cfme/services/catalogs/service_catalogs.py
@@ -1,8 +1,14 @@
 # -*- coding: utf-8 -*-
 from functools import partial
 
+from navmazing import NavigateToSibling, NavigateToAttribute
+
 import cfme.fixtures.pytest_selenium as sel
-from cfme.web_ui import accordion, flash, menu, form_buttons, Form, Input, Select
+from cfme.web_ui import(accordion, flash, form_buttons, Form, Input, Select,
+ match_location, toolbar as tb, PagedTable)
+from selenium.common.exceptions import NoSuchElementException
+from utils.appliance.implementations.ui import CFMENavigateStep, navigate_to, navigator
+from utils.appliance import Navigatable
 from utils.update import Updateable
 from utils.pretty import Pretty
 from utils.wait import wait_for
@@ -13,6 +19,7 @@ order_button = {
     '5.4': "//button[@title='Order this Service']"
 }
 accordion_tree = partial(accordion.tree, "Service Catalogs")
+list_tbl = PagedTable(table_locator="//div[@id='list_grid']//table")
 
 # Forms
 stack_form = Form(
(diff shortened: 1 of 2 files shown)


dajoRH:
Lint report for commit 419db164c91fdc6680e6296a851cf0b6f2af62c9:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

dajoRH:
Lint report for commit 78db844345d5fb42bc34feb68b0a9daa7100bb16:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

cfme/services/catalogs/service_catalogs.py
```diff
@@ -1,8 +1,13 @@
 # -*- coding: utf-8 -*-
 from functools import partial
 
+from navmazing import NavigateToSibling, NavigateToAttribute
+
 import cfme.fixtures.pytest_selenium as sel
-from cfme.web_ui import accordion, flash, menu, form_buttons, Form, Input, Select
+from cfme.web_ui import accordion, flash, form_buttons, Form, Input, Select, match_location,\
```
mshriver: When the list gets this long, lets format it like below. Not something we've got everywhere yet, but as we're changing these lists its nice to format them consistently. 

```python
from cfme.web_ui import  (
    accordion, flash, blah, blah, blahto100chars,
    more_blah)
```

cfme/services/catalogs/service_catalogs.py
```diff
@@ -33,47 +39,74 @@
...
         wait_for(flash.get_messages, num_sec=10, delay=2)
         flash.assert_success_message("Order Request was Submitted")
 
-    def order_stack_item(self, catalog, catalog_item):
-        sel.force_navigate('order_service_catalog',
-                           context={'catalog': catalog,
-                                    'catalog_item': catalog_item})
+    def order_stack_item(self):
+        navigate_to(self, 'Order')
         stack_form.fill(self.stack_data)
```
mshriver: This looks like the only difference between the order and order_stack_item functions at this point, it would be nice to condense this into a single `order` function that takes a boolean `fill` that triggers calling `stack_form.fill()`. I don't think its a change that needs to be made in this PR, but otherwise the code is duplicated in each function.

cfme/services/catalogs/service_catalogs.py
```diff
@@ -33,47 +39,74 @@
...
         sel.click(form_buttons.submit)
         wait_for(flash.get_messages, num_sec=10, delay=2)
         flash.assert_success_message("Order Request was Submitted")
+
+
+@navigator.register(ServiceCatalogs, 'All')
+class ServiceCatalogsAll(CFMENavigateStep):
+    prerequisite = NavigateToAttribute('appliance.server', 'LoggedIn')
+
+    def am_i_here(self):
```
mshriver: :+1:

cfme/services/catalogs/service_catalogs.py
```diff
@@ -33,47 +39,74 @@
...
-        ]
-    }
-)
-
-
-class ServiceCatalogs(Updateable, Pretty):
+match_page = partial(match_location, title='Catalogs', controller='catalog')
+
+
+def nav_to_all():
```
mshriver: I think since there is only one class in this module, and the other catalogs classes are already in different modules in the services.catalogs package, this won't be used outside of the 'All' destination registered for ServiceCatalogs. Thus you can drop this function and do it directly in the `All` destination step. We used this when there were multiple classes in a module with 2+ `All` destinations that shared navigation steps.

cfme/services/catalogs/service_catalogs.py
```diff
@@ -33,47 +39,74 @@
...
+            sel.check(paginator.check_all())
+            sel.uncheck(paginator.check_all())
+
+
+@navigator.register(ServiceCatalogs, 'Details')
+class ServiceCatalogDetails(CFMENavigateStep):
+    prerequisite = NavigateToSibling('All')
+
+    def step(self):
+        listview_table.click_row_by_cells({'Name': self.obj.service_name})
```
mshriver: Since this is a paged table, check out implementing `listview_table` as a `PagedTable` and calling `PagedTable.find_row_by_cell_on_all_pages()`. Doesn't look like the service catalogs table is a checkbox table, and you haven't used any checkbox table specific calls, so it should be an easy switch.

cfme/services/catalogs/service_catalogs.py
```diff
@@ -33,47 +39,74 @@
...
+
+    def resetter(self):
+        tb.refresh()
+
+
+@navigator.register(ServiceCatalogs, 'Order')
+class ServiceCatalogOrder(CFMENavigateStep):
+    prerequisite = NavigateToSibling('Details')
+
+    def am_i_here(self):
```
mshriver: This `am_i_here()` should be in the `Details` destination - the summary for the Order page is 'Order Service "{<name>}"'

mshriver:
This PR is missing a test that was using the old navigation branches. You'll need to audit for existing usages of the nav branches that you're replacing so we don't start seeing new exceptions in the tests. I did a cursory search and this was the only one I found: cfme.tests.test_ui_problems

cfme/tests/services/test_generic_service_catalogs.py
```diff
@@ -43,14 +43,14 @@ def test_delete_catalog_deletes_service(dialog, catalog):
     catalog.delete()
     service_catalogs = ServiceCatalogs("service_name")
     with error.expected(CandidateNotFound):
-        service_catalogs.order(catalog.name, catalog_item)
+        service_catalogs.order()
```
mshriver: You raised this exception handling as an area of concern in an email.

From what I'm seeing of how this was handled for catalog_item, the default exception raised by the table's select/click method is `NoSuchElementException`. The test yeild_fixture then catch and pass this exception when attempting to delete after yielding.

You can either modify your destinations to catch NoSuchElementException thrown by the table's click_row methods and re-raise as CandidateNotFound, or you can change the test here to expect a NoSuchElementException instead. 

CandidateNotFound exception was written specifically for accordion tree navigation failure, which is no longer the case here since you're navigating by the table and not by the accordion. That was likely the reason for not re-raising in catalog_item and just allowing the NoSuchElementException to pass through.

dajoRH:
Lint report for commit cda2ba717c365d2c5b247eaeaf81258cbf93901b:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

cfme/services/catalogs/service_catalogs.py
sshveta: I click on the row first (details page) and then the order button. the summary at this page is "Service 'name'"

mshriver: service_catalog_order
service_catalog_details


@sshveta These screenshots were taken in CFME 5.7, and show the service catalog details page with the summary 'Service "testing"', and the service catalog order page with a summary 'Order Service "testing"'. Can you double check your findings? From what I'm seeing this am_i_here() check would match on the details page, not the order page.

sshveta: screenshot from 2017-01-06 07 41 14

sshveta: @mshriver , I am using different navigation. This page shows up when I click on any service row . I check if "I am here" and then click on order

mshriver: So, the screenshot you included is the 'Details' page for the ServiceCatalog. That's the prerequisite for this `Order` destination.

This ServiceCatalogOrder.am_i_here() function will be run after the prerequisite is met, and **before** the `step()` is executed.  If it returns True, the `step()` will not be executed.

This am_i_here() will return true while on the ServiceCatalogDetails destination, and this class will never execute the step() function and click Order.

mshriver: ServiceCatalogDetails class should have:
```python
def am_i_here(self):
    return match_page(summary='Service "{}"'.format(self.service_name))
```

ServiceCatalogOrder class should have:
```python
def am_i_here(self):
    return match_page(summary='Order Service "{}"'.format(self.service_name))

Comment (mshriver) [inline cfme/services/catalogs/service_catalogs.py]: I'm testing localling and seeing other small problems (that I duplicated in my example above) like `self.service_name` instead of `self.obj.service_name`

Comment (mshriver) [inline cfme/services/catalogs/service_catalogs.py]: The exception when checking am_i_here() is suppressed and it defaults to returning False, which is why your tests passed.

**Comment by dajoRH** [comment]
Lint report for commit 1267b999e65513b780a9894aa2b821f3c9f1f9c5:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

File: cfme/tests/services/test_generic_service_catalogs.py [hunk 8]
```diff
@@ -8,7 +8,7 @@
 from cfme.services.catalogs.service_catalogs import ServiceCatalogs
 from cfme.services.catalogs.catalog_item import CatalogBundle
 from cfme.services import requests
-from cfme.exceptions import CandidateNotFound
+from widgetastic.exceptions import NoSuchElementException
```
mshriver: I think you actually want `from selenium.common.exceptions import NoSuchElementException`

dajoRH:
Lint report for commit 767879562a52b80cc3faf50bee152b8b15ba2785:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

PR #3935, opened 2017-01-05
[RFR] Test Basic Metrics - Fixed assertions

pavelzag:
{{pytest: cfme/tests/containers/test_basic_metrics.py -v --use-provider cm-env2 }}

This fixes the Basic Metrics test that allows him to actually fail when the OSE is down

### DIFF
--- a/cfme/tests/containers/test_basic_metrics.py
+++ b/cfme/tests/containers/test_basic_metrics.py
@@ -7,6 +7,7 @@ from utils.version import current_version
 
 pytestmark = [
     pytest.mark.uncollectif(lambda provider: current_version() < "5.6"),
+    pytest.mark.usefixtures('setup_provider'),
     pytest.mark.tier(1)]
 pytest_generate_tests = testgen.generate(
     testgen.container_providers, scope='function')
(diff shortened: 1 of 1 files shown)


cfme/tests/containers/test_basic_metrics.py
```diff
@@ -1,19 +1,17 @@
 import pytest
+import re
 from utils import testgen
 from utils import conf
 from utils.version import current_version
 
 
 pytestmark = [
     pytest.mark.uncollectif(lambda provider: current_version() < "5.6"),
-    pytest.mark.usefixtures('setup_provider'),
```
psav: I think you'll still need to setup the provider? No? Else what's the point of having the provider. Also what's the purpose of this test, it doesn't interact with ManageIQ at all.

pavelzag: Yes, no idea why it was deleted from the test. 
The purpose of the test is very simple, to check whether the OpenShift provider we work against has functioning metrics (We have this recurring issue and would like to detect it as a part of our tests run)

dajoRH:
Lint report for commit d760f1705ccedbd26260a93ce8cde72cc557fbc5:

**No commit flaws detected**.
<mergeable/>

*CFME QE Bot*

PR #3958, opened 2017-01-09
[1LP][RFR] Navmazing for Automate Explorer

mfalesni:
{{pytest: -v -m crud cfme/tests/automate/ -k 'domain or namespace or class_ or method or instance'}}


psav:
Just need the appliance kwargs on all the classes....then good to go.

PR #3942, opened 2017-01-05
[RFR] cloud.tenant CRUD and test

mshriver:
{{pytest: cfme/tests/cloud/test_tenant.py --use-provider rhos7-ga}}

Improving the create/delete methods for cloud.tenant and wrote a little CRUD test.

**PRT Results**
* 5.6: uncollected purposefully
* 5.7 and upstream: rhos7-ga is cleaning up the tenants, which is causing failures. create/delete are working as expected otherwise

Issue #3953, opened 2017-01-08
test_tenant failed, delete :UnboundLocalError: local variable 'result' referenced before assignment


rrasouli:
In my setup RHOS10 and CFME 5.7 latest, the cloud tenant crud failed on delete method, the provider tenant has already deleted, in test the failure is immediate.

@mshriver @psav 
py.test test_tenant.py --use-provider oc_ci
no test annotation found in None
============================================= test session starts ==============================================
platform linux2 -- Python 2.7.5, pytest-2.9.2, py-1.4.32, pluggy-0.3.1 -- /root/cfme_tests/bin/python
cachedir: ../../../.cache
rootdir: /root/cfme_tests/cfme_tests, inifile: pytest.ini
plugins: wait-for-1.0.8
collected 1 items 

Appliance's streams: [downstream, 5.7]

test_tenant.py::test_tenant_crud[oc_ci] FAILED

=================================================== FAILURES ===================================================
___________________________________________ test_tenant_crud[oc_ci] ____________________________________________

tenant = <cfme.cloud.tenant.Tenant object at 0x7f02fc049b90>

    @pytest.mark.uncollectif(lambda: current_version() < '5.7')
    def test_tenant_crud(tenant):
        """ Tests tenant create and delete
    
        Metadata:
            test_flag: tenant
        """
        tenant.create(cancel=True)
        assert not tenant.exists()
    
        tenant.create()
        assert tenant.exists()
    
>       tenant.delete(from_details=False, cancel=True)

test_tenant.py:43: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <cfme.cloud.tenant.Tenant object at 0x7f02fc049b90>, cancel = True, from_details = False, wait = True

    def delete(self, cancel=False, from_details=True, wait=True):
        if current_version() < '5.7':
            raise OptionNotAvailable('Cannot delete cloud tenants in CFME < 5.7')
        if from_details:
            # Generates no alert or confirmation
            try:
                navigate_to(self, 'Details')
            except Exception as ex:
                # Catch general navigation exceptions and raise
                raise TenantNotFound('Exception while navigating to Tenant details: {}'
                                     .format(ex))
            cfg_btn('Delete Cloud Tenant')
        else:
[...]

PR #3979, opened 2017-01-11
[1LP][RFR] added update test in tenant CRUD

rrasouli:
Continuing #3942 added update test.
In test code a helper refresh provider function, since the delete and update requires a refresh provider before we query the tenant 'All' page

### DIFF
--- a/cfme/cloud/tenant.py
+++ b/cfme/cloud/tenant.py
@@ -67,6 +67,24 @@ class Tenant(Navigatable):
         except TimedOutError:
             logger.error('Timed out waiting for tenant to disappear, continuing')
 
+    def wait_for(self, timeout=300):
+        try:
+            return wait_for(self.exists,
+                            fail_condition=False,
+                            timeout=timeout,
+                            message='Wait for cloud tenant to appear',
+                            delay=10,
+                            fail_func=sel.refresh)
+        except TimedOutError:
+            logger.error('Timed out waiting for tenant to appear, continuing')
+
+    def update(self, updates, wait=True):
+        navigate_to(self, 'Edit')
+        updated_name = updates.get('name', self.name + '_edited')
+        create_tenant_form.fill({'name': updated_name})
+        sel.click(create_tenant_form.save_button)
+        flash.assert_success_message('Cloud Tenant "{}" updated'.format(updated_name))
+
     def delete(self, cancel=False, from_details=True, wait=True):
         if current_version() < '5.7':
             raise OptionNotAvailable('Cannot delete cloud tenants in CFME < 5.7')
--- a/cfme/cloud/tenant.py
+++ b/cfme/cloud/tenant.py
@@ -83,7 +83,7 @@ class Tenant(Navigatable):
         updated_name = updates.get('name', self.name + '_edited')
         create_tenant_form.fill({'name': updated_name})
         sel.click(create_tenant_form.save_button)
-        flash.assert_success_message('Cloud Tenant "{}" updated'.format(updated_name))
+        self.wait_for(600)
 
     def delete(self, cancel=False, from_details=True, wait=True):
         if current_version() < '5.7':
@@ -126,7 +126,7 @@ class Tenant(Navigatable):
             # Flash message is the same whether deleted from details or by selection
             result = flash.assert_success_message('Delete initiated for 1 Cloud Tenant.')
             if wait:
-                result = self.wait_for_disappear()
+                result = self.wait_for_disappear(600)
             return result
 
     def exists(self):
--- a/cfme/cloud/tenant.py
+++ b/cfme/cloud/tenant.py
@@ -67,23 +67,20 @@ class Tenant(Navigatable):
         except TimedOutError:
             logger.error('Timed out waiting for tenant to disappear, continuing')
 
-    def wait_for(self, timeout=300):
-        try:
-            return wait_for(self.exists,
-                            fail_condition=False,
-                            timeout=timeout,
-                            message='Wait for cloud tenant to appear',
-                            delay=10,
-                            fail_func=sel.refresh)
-        except TimedOutError:
-            logger.error('Timed out waiting for tenant to appear, continuing')
+    # def wait_for(self, timeout=600):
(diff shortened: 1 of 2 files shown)


izapolsk:
However, your test failed in 5.7 and it isn't executed in 5.6. 
Please make sure you fixed it.

rrasouli:
@izapolsk fixed the failures found a bug in my run.

cfme/tests/cloud/test_tenant.py
```diff
@@ -34,14 +40,23 @@ def test_tenant_crud(tenant):
...
     """
+
     tenant.create(cancel=True)
     assert not tenant.exists()
 
     tenant.create()
     assert tenant.exists()
 
+    tenant.update({'name': 'my_new_name'})
+    tenant.wait_for()
```
psav: Why is there a need for a wait_for here?? Is this an openstack tenant? Are renaming it inside openstack? If so, the wait_for function probably doesn't need to be there and instead you can just put a wait around .exists in the update method.

cfme/cloud/tenant.py
```diff
@@ -67,6 +67,24 @@ def wait_for_disappear(self, timeout=300):
...
+                            fail_func=sel.refresh)
+        except TimedOutError:
+            logger.error('Timed out waiting for tenant to appear, continuing')
+
+    def update(self, updates, wait=True):
+        navigate_to(self, 'Edit')
+        updated_name = updates.get('name', self.name + '_edited')
+        create_tenant_form.fill({'name': updated_name})
+        sel.click(create_tenant_form.save_button)
+        self.wait_for(600)
```
psav: **REQUIRED:** So, the comment previously said, there is no need for a wait_for method, and to put the actual wait_for at the end of the update method. 
**REQUIRED:** The method also above is bad as it just silently passes if the provider doesn't appear. This is going to cause you bad debugging in the future as you will not see the error when it happens.

cfme/tests/cloud/test_tenant.py
```diff
@@ -34,14 +37,21 @@ def test_tenant_crud(tenant):
...
     """
+
     tenant.create(cancel=True)
     assert not tenant.exists()
 
     tenant.create()
     assert tenant.exists()
 
+    tenant.update({'name': 'my_new_name'})
+    refresh_provider(tenant.provider)
```
psav: **REQUIRED:** I was 100% sure I wrote a comment for this, but we don't need that ```refresh_provider``` function. it is a one liner. Just better to put ```tenant.provider.refresh_provider_relationships()``` instead.

cfme/tests/cloud/test_tenant.py
```diff
@@ -34,14 +37,21 @@ def test_tenant_crud(tenant):
...
+    assert tenant.exists()
```
psav: **REQUIRED:** One could argue that you do not need the refresh here as the exists should call it. In actual fact the fact that you don't call the refresh in the update method before starting the wait_for function is confusing as to why you need it later on here. If we didn't need a refresh to make sure that the provider existed in the loop, why do we need it now?

cfme/tests/cloud/test_tenant.py
rrasouli: @psav I will remove the helper function in case BZ#1411112 fixed

cfme/tests/cloud/test_tenant.py
mfalesni: @psav Also `with update(tenant):` should be used instead, right?

cfme/tests/cloud/test_tenant.py
rrasouli: @psav removed the helper function to one line on test

rrasouli:
@psav please review

cfme/cloud/tenant.py
```diff
@@ -67,6 +67,18 @@ def wait_for_disappear(self, timeout=300):
...
+        return wait_for(self.exists, fail_condition=False, timeout=timeout,
+                        message='Wait for cloud tenant to appear', delay=10,
+                        fail_func=sel.refresh)
+
+    def update(self, updates, wait=True):
+        navigate_to(self, 'Edit')
+        updated_name = updates.get('name', self.name + '_edited')
+        create_tenant_form.fill({'name': updated_name})
+        sel.click(create_tenant_form.save_button)
+        # self.provider.refresh_provider_relationships()
```
psav: 1) If it is required that the refresh take place for the update to show itself, it should be here as part of the update mechanism
2) I have already commented twice that the wait_for should take place inside the update method and should not be a separate method. For multiple reasons 
 a) You are only using it once, so there is no need to abstract it out
 b) The name is ambiguous
I don't even like the ```wait_for_disappear()``` that is used below. This really should just be a wait_for inside the method too, however since that is not really touched in this PR, I will ignore it.

rrasouli:
@mshriver https://github.com/ManageIQ/integration_tests/issues/3953 Please review the issue and suggest a solution

dajoRH:
*CFME QE Bot*

cfme/cloud/tenant.py
```diff
@@ -67,6 +67,19 @@ def wait_for_disappear(self, timeout=300):
...
+                        message='Wait for cloud tenant to appear', delay=10,
+                        fail_func=sel.refresh)
+
+    def update(self, updates, wait=True):
+        navigate_to(self, 'Edit')
+        updated_name = updates.get('name', self.name + '_edited')
+        create_tenant_form.fill({'name': updated_name})
+        sel.click(create_tenant_form.save_button)
+        self.provider.refresh_provider_relationships()
+        self.wait_for(600)
```
psav: As stated several times before

```python
    def update(self, updates, wait=True):
        navigate_to(self, 'Edit')
        updated_name = updates.get('name', self.name + '_edited')
        create_tenant_form.fill({'name': updated_name})
        sel.click(create_tenant_form.save_button)
        wait_for(self.exists, fail_condition=False, timeout=600,
            message='Wait for cloud tenant to appear', delay=10,
            fail_func=sel.refresh)
```

PR #3983, opened 2017-01-11
[1LP][WIP] addressing stack failures

sshveta:
{{pytest: -v --long-running --use-provider ec2 cfme/tests/services/test_provision_stack.py }}

### DIFF
--- a/cfme/fixtures/pytest_selenium.py
+++ b/cfme/fixtures/pytest_selenium.py
@@ -889,7 +889,7 @@ def checkbox(loc, set_to=False):
             selected = el.is_selected()
 
         if selected is not set_to:
-            logger.debug("Setting checkbox %s to %s", str(loc), str(set_to))
+            logger.debug("Setting checkbox to {}".format(set_to))
             click(el)
         return selected
 
--- a/cfme/services/catalogs/catalog_item.py
+++ b/cfme/services/catalogs/catalog_item.py
@@ -119,6 +119,8 @@ class CatalogItem(Updateable, Pretty, Navigatable):
     def create(self):
         # Create has sequential forms, the first is only the provider type
         navigate_to(self, 'Add')
+        # For element not found exception (To be removed)
+        sel.sleep(5)
         sel.select("//select[@id='st_prov_type']",
                    self.provider_type or self.item_type or 'Generic')
 
@@ -138,18 +140,15 @@ class CatalogItem(Updateable, Pretty, Navigatable):
                                'display_checkbox': self.display_in,
                                'select_catalog': catalog.name,
                                'select_dialog': dialog.name,
-                               'select_orch_template': self.orch_template,
+                               'select_orch_template': self.orch_template.template_name,
                                'select_provider': provider_name,
                                'select_config_template': self.config_template})
-        if sel.text(basic_info_form.field_entry_point) == "":
+        if not (self.item_type in provider_required_types):
             sel.click(basic_info_form.field_entry_point)
             if version.current_version() < "5.7":
                 dynamic_tree.click_path("Datastore", self.domain, "Service", "Provisioning",
                                      "StateMachines", "ServiceProvision_Template", "default")
-            else:
-                entry_tree.click_path("Datastore", self.domain, "Service", "Provisioning",
-                                     "StateMachines", "ServiceProvision_Template", "default")
-            sel.click(basic_info_form.apply_btn)
+                sel.click(basic_info_form.apply_btn)
         if self.catalog_name is not None \
                 and self.provisioning_data is not None \
                 and not isinstance(self.provider, NoneType):
--- a/cfme/services/catalogs/service_catalogs.py
+++ b/cfme/services/catalogs/service_catalogs.py
@@ -11,7 +11,6 @@ from utils.appliance.implementations.ui import CFMENavigateStep, navigate_to, na
 from utils.appliance import Navigatable
 from utils.update import Updateable
 from utils.pretty import Pretty
-from utils.wait import wait_for
 
 order_button = "//button[@title='Order this Service']"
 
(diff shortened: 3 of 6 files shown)


cfme/tests/services/test_provision_stack.py
```diff
@@ -164,12 +163,6 @@ def template(provider, provisioning, dialog_name):
 
     yield template
 
-    try:
```
mshriver: Why is this being removed - we should be cleaning up after tests?

sshveta: Yes It needs to be cleaned up after stack

mshriver: I'm not sure what you mean by your comment. Its scoped for function, so for each test it will create+delete the template that's used?

cfme/tests/services/test_provision_stack.py
```diff
@@ -259,6 +252,7 @@ def _cleanup_vms():
                 wait_for(lambda: provider.mgmt.delete_stack(stack_data['stack_name']),
                          delay=10, num_sec=800, message="wait for stack delete")
             stack_data['vm_name'].delete_from_provider()
+            catalog_item.orch_template.delete()
```
sshveta: Deleting template here as it needs to be cleaned up after stack . If the clean up code is inside the template method , it gets executed before stack gets cleaned up and hence fails .

cfme/tests/services/test_provision_stack.py
sshveta: Cleaning template at line 255

cfme/tests/services/test_provision_stack.py
mshriver: Gotcha, thanks for explaining.

In that case, you'll need to repeat this for each _cleanup_vms() in the module. `test_reconfigure_service` and `test_retire_stack` both use catalog_item fixture, which uses template fixture. I don't see a the template getting cleaned up in their finalizers?

cfme/services/catalogs/service_catalogs.py
```diff
@@ -53,7 +53,7 @@ def order(self):
         if self.stack_data:
             stack_form.fill(self.stack_data)
         sel.click(form_buttons.submit)
-        wait_for(flash.get_messages, num_sec=10, delay=2, fail_condition=[], fail_func=tb.refresh())
+        sel.sleep(5)
```
sshveta: wait_for(flash.get_messages, num_sec=10, delay=2, fail_condition=[], fail_func=tb.refresh()) and 
wait_for(flash.get_messages, num_sec=10, delay=2) are not working and test fails with 
FlashMessageException: No matching info flash message for 'Order Request was Submitted', instead got []

with sel.sleep(5) it worked but  should it be used ?

cfme/fixtures/pytest_selenium.py
```diff
@@ -889,7 +889,7 @@ def checkbox(loc, set_to=False):
             selected = el.is_selected()
 
         if selected is not set_to:
-            logger.debug("Setting checkbox %s to %s", str(loc), str(set_to))
+            logger.debug("Setting checkbox to %s", str(set_to))
```
sshveta: This is to fix : 
if selected is not set_to:
>               logger.debug("Setting checkbox %s to %s", str(loc), str(set_to))
E               TypeError: __str__ returned non-string (type WebElement)

sshveta:
Test pass locally for 5.6 and 5.7 
============
(cfme_vir_env)[sshveta@localhost cfme_tests]$ py.test --long-running -v --use-provider ec2  cfme/tests/services/test_provision_stack.py 
Loading templates from trackerbot                                                                                                                                                                                  
no test annotation found in None
=============================================================================================== test session starts ===============================================================================================
platform linux2 -- Python 2.7.5, pytest-2.9.2, py-1.4.32, pluggy-0.3.1 -- /home/sshveta/Work/cfme_vir_env/bin/python2
cachedir: .cache
rootdir: /home/sshveta/Work/cfme_vir_env/cfme_tests, inifile: 
plugins: wait-for-1.0.8
collected 4 items 

Appliance's streams: [5.6, downstream]

cfme/tests/services/test_provision_stack.py::test_provision_stack[ec2west] PASSED
cfme/tests/services/test_provision_stack.py::test_reconfigure_service[ec2west] PASSED
cfme/tests/services/test_provision_stack.py::test_remove_template_provisioning[ec2west] PASSED
cfme/tests/services/test_provision_stack.py::test_retire_stack[ec2west] PASSED


But it never passes PRT :/

cfme/fixtures/pytest_selenium.py
mshriver: Consider keeping the same logging and switching to use format() instead of attempting to cast vars to string. This should handle loc being a WebElement.

cfme/services/catalogs/service_catalogs.py
mshriver: Throw in a TODO comment so this gets removed when we can.

cfme/fixtures/pytest_selenium.py
```diff
@@ -889,7 +889,7 @@ def checkbox(loc, set_to=False):
             selected = el.is_selected()
 
         if selected is not set_to:
-            logger.debug("Setting checkbox %s to %s", str(loc), str(set_to))
+            logger.debug("Setting checkbox {} to {}".format(loc, set_to))
```
sshveta: Still getting  TypeError: __str__ returned non-string (type WebElement)

sshveta:
Ec2 validation fails is causing the failures

cfme/services/catalogs/catalog_item.py
```diff
@@ -119,6 +119,8 @@ def __str__(self):
     def create(self):
         # Create has sequential forms, the first is only the provider type
         navigate_to(self, 'Add')
+        # For element not found exception (To be removed)
+        sel.sleep(5)
```
psav: Can't we just put is displayed in a wait for for the element below?

sshveta: @psav I tried that, it inconsistently still failed on PRT .

cfme/services/catalogs/catalog_item.py
```diff
@@ -138,18 +140,15 @@ def create(self):
                                'display_checkbox': self.display_in,
                                'select_catalog': catalog.name,
                                'select_dialog': dialog.name,
-                               'select_orch_template': self.orch_template,
+                               'select_orch_template': self.orch_template.template_name,
                                'select_provider': provider_name,
                                'select_config_template': self.config_template})
-        if sel.text(basic_info_form.field_entry_point) == "":
+        if not (self.item_type in provider_required_types):
```
izapolsk: you can use "not in" operator

PR #3988, opened 2017-01-12
[RFR][1LP] BootstrapTreeview: Fix text matcher

mfalesni:
This might fail but it should not fail on tree operations

{{pytest: cfme/tests/control/test_basic.py::test_policy_crud -k VMControlPolicy -v --long-running}}


mfalesni:
Note: the locator is used just to pre-filter items in tree to make the scope narrower to speed up individual node matching.

PR #3989, opened 2017-01-12
[1LP][RFR] Moved finalizer to correct place

bargulg:
Previously there were problems with snapshot tests not removing VM they
created on RHV golden environment, this should fix it.

Also, not removing the VM if it was already present, to make sure it doesn't break stable environments, where you always want to have that VM...

### DIFF
--- a/cfme/tests/infrastructure/test_snapshot.py
+++ b/cfme/tests/infrastructure/test_snapshot.py
@@ -39,12 +39,13 @@ def domain(request):
 
 
 @pytest.fixture(scope="module")
-def test_vm(setup_provider_modscope, provider, vm_name):
+def test_vm(setup_provider_modscope, provider, vm_name, request):
     """Fixture to provision appliance to the provider being tested if necessary"""
     vm = VM.factory(vm_name, provider, template_name=provider.data['small_template'])
 
     if not provider.mgmt.does_vm_exist(vm_name):
         vm.create_on_provider(find_in_cfme=True, allow_skip="default")
+        request.addfinalizer(test_vm.delete_from_provider)
     return vm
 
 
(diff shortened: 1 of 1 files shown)


cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -39,12 +39,13 @@ def domain(request):
...
 
 @pytest.fixture(scope="module")
-def test_vm(setup_provider_modscope, provider, vm_name):
+def test_vm(setup_provider_modscope, provider, vm_name, request):
     """Fixture to provision appliance to the provider being tested if necessary"""
     vm = VM.factory(vm_name, provider, template_name=provider.data['small_template'])
 
     if not provider.mgmt.does_vm_exist(vm_name):
         vm.create_on_provider(find_in_cfme=True, allow_skip="default")
+        request.addfinalizer(test_vm.delete_from_provider)
```
mfalesni: test_vm vs vm?

mfalesni:
@bargulg When you create a new PR, add a tag to its title as per our PR review policy. [WIP], [WIPTEST] or [RFR] (when you are ready to review).

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -39,12 +39,13 @@ def domain(request):
...
 
 @pytest.fixture(scope="module")
-def test_vm(setup_provider_modscope, provider, vm_name):
+def test_vm(setup_provider_modscope, provider, vm_name, request):
     """Fixture to provision appliance to the provider being tested if necessary"""
     vm = VM.factory(vm_name, provider, template_name=provider.data['small_template'])
 
     if not provider.mgmt.does_vm_exist(vm_name):
         vm.create_on_provider(find_in_cfme=True, allow_skip="default")
+        request.addfinalizer(vm.delete_from_provider)
```
mshriver: Great improvement to the test structure.

Consider a yield_fixture, where you can do more to wrap the delete in a try block and do things like log+continue if it fails, etc. Just a consideration, not necessary.

cfme/tests/infrastructure/test_snapshot.py
bargulg: done

PR #3993, opened 2017-01-12
[1LP][RFR] Update utils.providers._setup_providers, use navigate_to

mshriver:
Add class map for prov types coming from prov_class

### DIFF
--- a/utils/providers.py
+++ b/utils/providers.py
@@ -12,12 +12,14 @@ import cfme.fixtures.pytest_selenium as sel
 from fixtures.pytest_store import store
 from cfme.web_ui import Quadicon, paginator
 from cfme.common.provider import BaseProvider
-from cfme.containers import provider as container_providers  # NOQA
-from cfme.cloud import provider as cloud_providers  # NOQA
-from cfme.infrastructure import provider as infrastructure_providers  # NOQA
-from cfme.middleware import provider as middleware_providers  # NOQA
+from cfme.containers import provider as container_providers # NOQA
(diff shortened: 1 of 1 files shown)


utils/providers.py
```diff
@@ -264,7 +266,7 @@ def setup_provider_by_name(provider_name, *args, **kwargs):
     return setup_provider(get_provider_key(provider_name), *args, **kwargs)
 
 
-def setup_providers(prov_classes=('cloud', 'infra'), validate=True, check_existing=True):
+def setup_providers(prov_classes=['cloud', 'infra'], validate=True, check_existing=True):
```
mshriver: Changed this to a list, when it was called with a single item in a tuple prov_classes was getting set to the item (a string), and was passing the first character of the string as `pclass`

utils/providers.py
```diff
@@ -307,14 +309,19 @@ def _setup_providers(prov_class, validate, check_existing):
 
     """
 
+    prov_class_map = {
```
psav: There is another way to do this, because those are imported at the top you can do this,
```python
from cfme.common.provider import BaseProvider
```
and then the ```prov_class_map``` IS ```BaseProvider.type_mapping```

mshriver: okay, I did look at BaseProvider.type_mapping, but I couldn't confirm that it was getting populated with the strings 'cloud', 'infra', 'middleware', 'container'. I'll test this.

mshriver: fixed

PR #3997, opened 2017-01-12
[1LP][RFR] Access Control Tenant Nav

mshriver:
Added new ChargebackView to chargeback, registered Navigation against Server.

Added Access Control Tenant navigation

Refactored tests accordingly

Using WIPTEST and RFR because of the priority push for force_navigate() removal.

{{ pytest: cfme/tests/configure/test_access_control.py -v --long-running }}


mshriver:
1LP, psav looked at it too

PR #4005, opened 2017-01-13
[1LP][RFR]Navmazing in Workloads

lkhomenk:
__adding__ new navigation to Services/Workloads and replacing old force_navigates

### DIFF
--- a/cfme/services/workloads.py
+++ b/cfme/services/workloads.py
@@ -2,8 +2,8 @@
 """ A model of Workloads page in CFME
 """
 from functools import partial
-
 from navmazing import NavigateToSibling, NavigateToAttribute  # NOQA
+
 from cfme.exceptions import DestinationNotFound
 from cfme.web_ui import accordion, match_location, toolbar
 from utils.appliance import Navigatable
@@ -12,15 +12,12 @@ from utils.appliance.implementations.ui import navigator, CFMENavigateStep
 vm_instances_tree = partial(accordion.tree, "VMs & Instances", 'All VMs & Instances')
 templates_images_tree = partial(accordion.tree, "Templates & Images", 'All Templates & Images')
 
-QUADICON_TITLE_LOCATOR = ("//div[@id='quadicon']/../../../tr/td/a[contains(@href,'vm_infra/x_show')"
-                          " or contains(@href, '/show/')]")
-
 # Only this buttons present. may be used in future
 cfg_btn = partial(toolbar.select, 'Configuration')
 pol_btn = partial(toolbar.select, 'Policy')
 lcl_btn = partial(toolbar.select, 'Lifecycle')
 
-match_page = partial(match_location, controller='vm_or_template')
+match_page = partial(match_location, controller='vm_or_template', title='Workloads')
 
 
 class VmsInstances(Navigatable):
--- a/cfme/services/workloads.py
+++ b/cfme/services/workloads.py
@@ -2,28 +2,79 @@
 """ A model of Workloads page in CFME
 """
 from functools import partial
-from cfme.web_ui import accordion, menu
 
-vm_instances_tree = partial(accordion.tree, "VMs & Instances")
-templates_images_tree = partial(accordion.tree, "Templates & Images")
+from navmazing import NavigateToSibling, NavigateToAttribute  # NOQA
+from cfme.exceptions import DestinationNotFound
+from cfme.web_ui import accordion, match_location, toolbar
+from utils.appliance import Navigatable
+from utils.appliance.implementations.ui import navigator, CFMENavigateStep
+
+vm_instances_tree = partial(accordion.tree, "VMs & Instances", 'All VMs & Instances')
+templates_images_tree = partial(accordion.tree, "Templates & Images", 'All Templates & Images')
+
+QUADICON_TITLE_LOCATOR = ("//div[@id='quadicon']/../../../tr/td/a[contains(@href,'vm_infra/x_show')"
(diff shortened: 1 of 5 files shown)


lkhomenk:
0a3241f - to make test_default_filers work.before we were losing kwargs passed to navigate_to

cfme/services/workloads.py
```diff
@@ -2,28 +2,79 @@
...
+from utils.appliance.implementations.ui import navigator, CFMENavigateStep
 
+vm_instances_tree = partial(accordion.tree, "VMs & Instances", 'All VMs & Instances')
+templates_images_tree = partial(accordion.tree, "Templates & Images", 'All Templates & Images')
 
-@menu.extend_nav
-class services_workloads:
-    def service_vms_instances(_):
-        vm_instances_tree("All VMs & Instances")
+QUADICON_TITLE_LOCATOR = ("//div[@id='quadicon']/../../../tr/td/a[contains(@href,'vm_infra/x_show')"
```
mshriver: I don't think this is needed? Not used in your code here at all, and the only other place I see it in the codebase is in cfme.infrastructure.virtual_machines

cfme/services/workloads.py
```diff
@@ -2,28 +2,79 @@
...
-    def service_vms_instances_filter_folder(ctx):
-        vm_instances_tree("All VMs & Instances", ctx["folder_name"])
+# Only this buttons present. may be used in future
+cfg_btn = partial(toolbar.select, 'Configuration')
+pol_btn = partial(toolbar.select, 'Policy')
+lcl_btn = partial(toolbar.select, 'Lifecycle')
 
-    def service_vms_instances_filter(ctx):
-        vm_instances_tree("All VMs & Instances", ctx["folder_name"], ctx["filter_name"])
+match_page = partial(match_location, controller='vm_or_template')
```
mshriver: Include `title=''` argument as well, matches against xpath:/html/head/title with the 'CFME: ' stripped.

cfme/services/workloads.py
```diff
@@ -2,28 +2,79 @@
 """ A model of Workloads page in CFME
 """
 from functools import partial
-from cfme.web_ui import accordion, menu
 
-vm_instances_tree = partial(accordion.tree, "VMs & Instances")
-templates_images_tree = partial(accordion.tree, "Templates & Images")
+from navmazing import NavigateToSibling, NavigateToAttribute  # NOQA
```
izapolsk: this is external lib, please move it to external libraries.

cfme/services/workloads.py
```diff
@@ -2,28 +2,79 @@
...
-        templates_images_tree("All Templates & Images", ctx["folder_name"])
+class VmsInstances(Navigatable):
+    """
+        This is fake class mainly needed for navmazing navigation
 
-    def service_templates_images_filter(ctx):
-        templates_images_tree("All Templates & Images", ctx["folder_name"], ctx["filter_name"])
+    """
+    def __init__(self, appliance=None):
+        Navigatable.__init__(self, appliance)
```
izapolsk: we would need to replace all such statements to use super instead. But it isn't necessary for now

cfme/tests/configure/test_default_views_infra.py
```diff
@@ -34,9 +33,8 @@ def setup_a_provider():
     'VMs': Vm,
     'My Services': MyService,
     'Catalog Items': CatalogItem,
-    'VMs & Instances': 'service_vms_instances',
-    'Templates & Images': 'service_templates_images'
-}
+    'VMs & Instances': VmsInstances,
+    'Templates & Images': TemplatesImages}
```
izapolsk: :+1:  but consider making this like it was done by psav in the following commit
https://github.com/ManageIQ/integration_tests/pull/4003/files#diff-707235ec81985bc8d13d47291b348740

utils/appliance/implementations/ui.py
```diff
@@ -232,7 +232,7 @@ def do_nav(self, _tries=0, *args, **kwargs):
         br = self.appliance.browser
 
         try:
-            self.step()
+            self.step(*args, **kwargs)
```
izapolsk: :1st_place_medal: , two beer this man. Thanks. I forgot to push my PR with kwargs when worked on virtual machines.

cfme/services/workloads.py
lkhomenk: yes. I've put it there just in case to not search it - workloads are a bit mysterious. Workloads is a combination of Infra VMs +  Instances / Infra Templates + Images - it might be used for all testing.
But it stays without use from our side at all. 
I'll remove it

cfme/tests/configure/test_default_views_infra.py
lkhomenk: @izapolsk that was done to fit others locations in that test. it has further some login with conditions which changes destination.

cfme/services/workloads.py
lkhomenk: done

cfme/services/workloads.py
lkhomenk: done

cfme/services/workloads.py
lkhomenk: done (damn - locator was terribly wrong for this page)

lkhomenk:
needs few more minutes - issue in am_i_here

cfme/services/workloads.py
mshriver: Part of why I suggested removing  :)

PR #4010, opened 2017-01-16
[1LP][RFR] fix for href menu navigation

izapolsk:
Purpose or Intent
=================
some of nav destinations use href instead of title.
This small patch fixes that issue.

{{pytest: --long-running cfme/tests/control/test_alerts.py:test_alert_vm_turned_on_more_than_twice_in_past_15_minutes}}

### DIFF
--- a/widgetastic_patternfly.py
+++ b/widgetastic_patternfly.py
@@ -314,7 +314,7 @@ class VerticalNavigation(Widget):
     CURRENTLY_SELECTED = './/li[contains(@class, "active")]/a'
     LINKS = './li/a'
     ITEMS_MATCHING = './li[a[normalize-space(.)={}]]'
-    DIV_LINKS_MATCHING = './ul/li/a[span[normalize-space(.)={}]]'
+    DIV_LINKS_MATCHING = './ul/li/a[span[normalize-space(.)={txt}] or @href={txt}]'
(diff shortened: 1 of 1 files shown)


widgetastic_patternfly.py
```diff
@@ -392,7 +392,7 @@ def select(self, *levels, **kwargs):
             passed_levels.append(level)
             finished = passed_levels == levels
             link = self.browser.element(
-                self.DIV_LINKS_MATCHING.format(quote(level)), parent=current_div)
+                self.DIV_LINKS_MATCHING.format(txt=quote(level)), parent=current_div)
```
mfalesni: Assuming no menu item text can start with / it should be fine I guess ...

izapolsk: what's wrong if menu item starts with / ?
There are such menu items but I haven't experienced any issues during testing them.
self.parent_view.navigation.select('Compute', 'Infrastructure', '/vm_infra/explorer')

mfalesni: For this use case it is perfectly fine. I think we can merge this and I might eventually split the locators to two for each use case to combat the ambiguity.

jkrocil:
:cake: :fist_oncoming:

PR #4011, opened 2017-01-16
[1LP][RFR] New test test_check_compliance_history

quarckster:
{{pytest: -v -k 'test_check_compliance_history' --long-running}}

Purpose or Intent
=================

__New test__ for feature Control to cover case test_check_compliance_history. https://bugzilla.redhat.com/show_bug.cgi?id=1375093

### DIFF
--- a/cfme/control/explorer/policies.py
+++ b/cfme/control/explorer/policies.py
@@ -422,6 +422,14 @@ class BasePolicy(Updateable, Navigatable, Pretty):
         view.flash.assert_message('Actions for Policy Event "{}" were saved'.format(
             event))
 
+    @property
+    def exists(self):
+        policies = self.appliance.db["miq_policies"]
+        return self.appliance.db.session\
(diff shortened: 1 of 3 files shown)


cfme/control/explorer/policies.py
```diff
@@ -421,6 +422,14 @@ def assign_actions_to_event(self, event, actions):
         view.flash.assert_message('Actions for Policy Event "{}" were saved'.format(
             event))
 
+    @property
```
izapolsk: wouldn't it be better to check this somehow thru UI ?

quarckster: @izapolsk  It's checked here https://github.com/ManageIQ/integration_tests/blob/master/cfme/control/explorer/policies.py#L138, but due this BZ https://bugzilla.redhat.com/show_bug.cgi?id=1359909 for now it's broken

PR #4024, opened 2017-01-18
[RFR] Moved fixtures and adding new test coverage for appliance_console_cli

lcouzens:
Moved fixtures such as sprout provisioning, so new tests can call them rather then adding additional fixtures to each file.

### DIFF
--- a/cfme/fixtures/cli.py
+++ b/cfme/fixtures/cli.py
@@ -4,6 +4,22 @@ from utils.appliance import current_appliance
 from utils.conf import cfme_data, credentials
 from utils.log import logger
 import pytest
+import paramiko
+
+
+@pytest.fixture()
+def dedicated_db(appliance, app_creds):
+    HOST = appliance.address
+    USER = app_creds['username']
+    PASS = app_creds['password']
+    client = paramiko.SSHClient()
+    client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+    client.connect(HOST, username=USER, password=PASS)
+    channel = client.invoke_shell()
+    stdin = channel.makefile('wb')
+    stdin.write("ap \n 8 \n 1 \n 1 \n 1 \n y \n {PASS} \n {PASS} \n \n")
+
+    return appliance
 
 
 @pytest.yield_fixture(scope="module")
@@ -43,6 +59,16 @@ def fqdn_appliance():
     yield apps[0]
 
     sp.destroy_pool(pool_id)
+
+
+@pytest.yield_fixture()
+def ipa_crud(fqdn_appliance, app_creds, ipa_creds):
+    fqdn_appliance.ap_cli.configure_ipa(ipa_creds['ipaserver'], ipa_creds['username'],
+        ipa_creds['password'], ipa_creds['domain'], ipa_creds['realm'])
+
+    yield(fqdn_appliance)
+
+    fqdn_appliance.ap_cli.uninstall_ipa_client()
 
 
 @pytest.fixture()
--- a/cfme/fixtures/cli.py
+++ b/cfme/fixtures/cli.py
@@ -1,0 +1,69 @@
+from utils.version import get_stream
+from cfme.test_framework.sprout.client import SproutClient
+from utils.appliance import current_appliance
+from utils.conf import cfme_data, credentials
+from utils.log import logger
+import pytest
+
+
+@pytest.yield_fixture(scope="module")
+def appliance():
+    sp = SproutClient.from_config()
+    version = current_appliance.version.vstring
+    stream = get_stream(current_appliance.version)
+    apps, pool_id = sp.provision_appliances(
+        count=1, preconfigured=False, version=version, stream=stream)
+
+    yield apps[0]
+
+    sp.destroy_pool(pool_id)
+
+
+@pytest.yield_fixture(scope="module")
+def fqdn_appliance():
+    sp = SproutClient.from_config()
+    available_providers = set(sp.call_method('available_providers'))
+    required_providers = set(cfme_data['fqdn_providers'])
+    usable_providers = available_providers & required_providers
+    version = current_appliance.version.vstring
+    stream = get_stream(current_appliance.version)
+
+    for provider in usable_providers:
+        try:
+            apps, pool_id = sp.provision_appliances(
+                count=1, preconfigured=True, version=version, stream=stream, provider=provider)
+            break
+        except Exception as e:
+            logger.warning("Couldn't provision appliance with following error:")
+            logger.warning({}.format(e))
+            continue
+    else:
+        logger.error("Couldn't provision an appliance at all")
+
+    yield apps[0]
+
+    sp.destroy_pool(pool_id)
+
+
+@pytest.fixture()
+def app_creds():
+    return {
+        'username': credentials['database']['username'],
+        'password': credentials['database']['password'],
+        'sshlogin': credentials['ssh']['username'],
+        'sshpass': credentials['ssh']['password']
+    }
+
+
+@pytest.fixture()
+def ipa_creds():
+    fqdn = cfme_data['auth_modes']['ext_ipa']['ipaserver'].split('.', 1)
+    creds_key = cfme_data['auth_modes']['ext_ipa']['credentials']
+    return{
+        'hostname': fqdn[0],
+        'domain': fqdn[1],
+        'realm': cfme_data['auth_modes']['ext_ipa']['iparealm'],
+        'ipaserver': cfme_data['auth_modes']['ext_ipa']['ipaserver'],
+        'username': credentials[creds_key]['principal'],
+        'password': credentials[creds_key]['password']
+    }
--- a/cfme/fixtures/cli.py
+++ b/cfme/fixtures/cli.py
@@ -27,6 +27,10 @@ def dedicated_db_appliance(app_creds):
     sp.destroy_pool(pool_id)
 
 
+""" The Following fixture 'fqdn_appliance' provisions one appliance for testing from an FQDN
+    provider unless there are no provisions available"""
+
+
 @pytest.yield_fixture(scope="function")
 def fqdn_appliance():
     sp = SproutClient.from_config()
@@ -35,7 +39,6 @@ def fqdn_appliance():
     usable_providers = available_providers & required_providers
     version = current_appliance.version.vstring
     stream = get_stream(current_appliance.version)
-
     for provider in usable_providers:
         try:
             apps, pool_id = sp.provision_appliances(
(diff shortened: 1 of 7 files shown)


dajoRH:
Lint report for commit ffef1cee4d98bd2ca218b84c81e029495972c146:

`cfme/fixtures/cli.py`:
- :red_circle: Line 12:5: **N806** *variable in function should be lowercase*
- :red_circle: Line 13:5: **N806** *variable in function should be lowercase*
- :red_circle: Line 14:5: **N806** *variable in function should be lowercase*

`cfme/tests/cli/test_appliance_cli.py`:
- :red_circle: Line 33:5: **N806** *variable in function should be lowercase*

**No commit flaws detected**.

*CFME QE Bot*

cfme/fixtures/cli.py
```diff
@@ -0,0 +1,95 @@
...
+import pytest
+import paramiko
+
+
+@pytest.fixture()
+def dedicated_db(appliance, app_creds):
+    hostname = appliance.address
+    username = app_creds['username']
+    password = app_creds['password']
+    client = paramiko.SSHClient()
```
psav: Why not just use ```appliance.ssh_client``` here?

cfme/fixtures/cli.py
```diff
@@ -0,0 +1,95 @@
...
+    stream = get_stream(current_appliance.version)
+
+    for provider in usable_providers:
+        try:
+            apps, pool_id = sp.provision_appliances(
+                count=1, preconfigured=True, version=version, stream=stream, provider=provider)
+            break
+        except Exception as e:
+            logger.warning("Couldn't provision appliance with following error:")
+            logger.warning({}.format(e))
```
psav: ```"{}".format(e)``` ?

utils/log_validator.py
```diff
@@ -37,10 +37,13 @@ class LogValidator(object):
     """
 
     def __init__(self, remote_filename, **kwargs):
-        self._remote_file_tail = SSHTail(remote_filename)
-        self.skip_patterns = kwargs['skip_patterns'] if 'skip_patterns' in kwargs else []
-        self.failure_patterns = kwargs['failure_patterns'] if 'failure_patterns' in kwargs else []
-        self.matched_patterns = kwargs['matched_patterns'] if 'matched_patterns' in kwargs else []
+        self.skip_patterns = kwargs.pop('skip_patterns') if 'skip_patterns' in kwargs else []
```
mshriver: Suggest you use kwargs built-in get() method, which takes a default value as a second parameter.

```python
self.var = kwargs.get('var', [])
```

utils/appliance/__init__.py
```diff
@@ -70,17 +70,34 @@ def run(self, ap_cli_command):
     def set_hostname(self, hostname):
         self.run("-H {}".format(hostname))
 
+    def configure_appliance_external_join(self, dbhostname,
+            username, password, dbname, fetch_key, sshlogin, sshpass):
+        self.run("-h {} -U {} -p {} -d {} -v -K {} -s {} -a {}".format(
```
mshriver: consider naming the fields since there are 7 of them

dajoRH:
*CFME QE Bot*

cfme/fixtures/cli.py
```diff
@@ -0,0 +1,116 @@
...
+    channel = client.invoke_shell()
+    stdin = channel.makefile('wb')
+    stdin.write("ap \n 8 \n 1 \n 1 \n 1 \n y \n {} \n {} \n \n".format(pwd, pwd))
+    wait_for(is_dedicated_db_active, func_args=[appliance2])
+
+    return appliance2
+
+
+@pytest.yield_fixture(scope="function")
+def appliance():
```
otsuman: **REQUIRED:**  use here and below 
https://github.com/ManageIQ/integration_tests/blob/master/fixtures/appliance.py#L60

dajoRH:
Lint report for commit 37bddfb2c9487ef472b7132eea7272dffe687b7b:

`utils/appliance/__init__.py`:
- :red_circle: Line 1409:9: **N805** *first argument of a method should be named 'self'*

**No commit flaws detected**.

*CFME QE Bot*

cfme/fixtures/cli.py
```diff
@@ -0,0 +1,87 @@
...
+    sp = SproutClient.from_config()
+    version = current_appliance.version.vstring
+    stream = get_stream(current_appliance.version)
+    apps, pool_id = sp.provision_appliances(
+        count=1, preconfigured=False, version=version, stream=stream)
+    pwd = app_creds['password']
+    client = apps[0].ssh_client
+    channel = client.invoke_shell()
+    stdin = channel.makefile('wb')
+    stdin.write("ap \n 8 \n 1 \n 1 \n 1 \n y \n {} \n {} \n \n".format(pwd, pwd))
```
psav: **REQUIRED:** Are these options likely to change for different versions, should we have a map, do they differ already?

cfme/fixtures/cli.py
```diff
@@ -0,0 +1,87 @@
...
+@pytest.yield_fixture(scope="function")
+def fqdn_appliance():
+    sp = SproutClient.from_config()
+    available_providers = set(sp.call_method('available_providers'))
+    required_providers = set(cfme_data['fqdn_providers'])
+    usable_providers = available_providers & required_providers
+    version = current_appliance.version.vstring
+    stream = get_stream(current_appliance.version)
+
+    for provider in usable_providers:
```
psav: **REQUIRED:** Some comment around this would be nice, it looks at first glance as though you are requesting multiple appliances because of the loop.

cfme/fixtures/cli.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+@pytest.yield_fixture()
+def ipa_crud(fqdn_appliance, app_creds, ipa_creds):
+    fqdn_appliance.ap_cli.configure_ipa(ipa_creds['ipaserver'], ipa_creds['username'],
+        ipa_creds['password'], ipa_creds['domain'], ipa_creds['realm'])
+
+    yield(fqdn_appliance)
+
+    fqdn_appliance.ap_cli.uninstall_ipa_client()
```
psav: If this appliance is going to get destroyed at the end of the test, is there a need to uninstall? Though it is cleaner and I guess if other people use it.......discussion point ;)

utils/appliance/__init__.py
```diff
@@ -73,19 +74,42 @@ def run(self, ap_cli_command):
             "appliance_console_cli {}".format(ap_cli_command))
 
     def set_hostname(self, hostname):
-        self.run("-H {}".format(hostname))
+        self.run("-H {host}".format(host=hostname))
```
psav: **REQUIRED:** Looking at this now, I really think we should change ```.run()``` to be ```._run()```. There shouldn't be a reason for the average user to run this method. So let's make it less obvious.

utils/appliance/__init__.py
```diff
@@ -1382,6 +1406,11 @@ def enable_external_db(self, db_address, region=0, db_name=None,
 
         return status, out
 
+    def is_dedicated_db_active(self):
+        return_code, output = self.ssh_client.run_command(
+            "systemctl status {}-postgresql.service | grep running".format(scl_name()))
```
psav: Is this for 5.7 only?

utils/log_validator.py
```diff
@@ -39,10 +39,11 @@ class LogValidator(object):
     """
 
     def __init__(self, remote_filename, **kwargs):
-        self._remote_file_tail = SSHTail(remote_filename)
-        self.skip_patterns = kwargs['skip_patterns'] if 'skip_patterns' in kwargs else []
-        self.failure_patterns = kwargs['failure_patterns'] if 'failure_patterns' in kwargs else []
-        self.matched_patterns = kwargs['matched_patterns'] if 'matched_patterns' in kwargs else []
+        self.skip_patterns = kwargs.pop('skip_patterns', [])
```
psav: Beautiful!

cfme/fixtures/cli.py
lcouzens: The options may change in each version, but for the moment they work on 5.7 builds. The plan is to change this to setup the appliance via the cli but that option is not available _yet_

cfme/fixtures/cli.py
lcouzens: True there is no reason to uninstall ipa again this is now removed

utils/appliance/__init__.py
lcouzens: Agreed.

utils/appliance/__init__.py
lcouzens: this is indeed only for 5.7+

cfme/fixtures/cli.py
lcouzens: Where should I put the comment? line 38 or above the fixture?

cfme/fixtures/cli.py
psav: I think then you should add a check in there and raise an error if we try to run this on a 5.6....because otherwise we could run any old option.

cfme/fixtures/cli.py
```diff
@@ -0,0 +1,90 @@
...
+        sp = SproutClient.from_config()
+        version = current_appliance.version.vstring
+        stream = get_stream(current_appliance.version)
+        apps, pool_id = sp.provision_appliances(
+            count=1, preconfigured=False, version=version, stream=stream)
+        pwd = app_creds['password']
+        client = apps[0].ssh_client
+        channel = client.invoke_shell()
+        stdin = channel.makefile('wb')
+        stdin.write("ap \n 8 \n 1 \n 1 \n 1 \n y \n {} \n {} \n \n".format(pwd, pwd))
```
jkrocil: I believe we want test case for not sending any password at all as well. If you dont pick a password, it breaks the appliance afaik.

jkrocil: **REQUIRED** ^

lcouzens: I agree we should have a test for that, but I think we should add that to the appliance_console PR I have opened rather than this cli one. I mean atm this is just a go between fixture which I hope to replace with a new cli method once we get some more feature parity.

cfme/fixtures/cli.py
```diff
@@ -0,0 +1,90 @@
...
+from utils.appliance import current_appliance
+from utils.conf import cfme_data, credentials
+from utils.log import logger
+import pytest
+from wait_for import wait_for
+from cfme.test_framework.sprout.client import SproutException
+
+
+@pytest.yield_fixture(scope="function")
+def dedicated_db_appliance(app_creds):
```
lcouzens: Is this also required?

lcouzens: I just tried this, it works for the fixture but sprout_client.destroy_pool(request_id) on appliance.py is called before the test is run against the dedicated_db_appliance.

cfme/fixtures/cli.py
jkrocil: Sounds good to me, I don't mind if its handled here or in a different PR but we should add it.

PR #4013, opened 2017-01-16
[WIP] Widgetastic conversion of cfme.cloud.instance

mshriver:
Conversion of cfme.cloud.instance class to widgetastic views/forms.

Nested the forms in the views for each navigation destination. Not all of the navigation destinations are used by tests.

I do not plan to fix all failing tests within this PR, only to convert the navigation/form/widget use and to fix any failures that are a result of that conversion.

{{ pytest: cfme/tests/cloud/test_instance_power_control.py cfme/tests/cloud/test_provisioning.py --long-running --use-provider rhos7-ga --use-provider ec2west --use-provider azure }}

## PRT

Fixing deferred_verpick errors in tests with PR #4544 
Other failures/skips that I'm seeing aren't related to this migration but are in effected tests so they show up here.

PR #4031, opened 2017-01-19
[1LP][RFR] Widgetastic ProvisioningForm 

mshriver:
Split this from PR #4013 to get it merged sooner since its imported by several modules that we'll be converting to widgetastic.

I've tested locally with instance provisioning and the form was properly instantiated and filled. I didn't hit every provisioning field in testing.

Leaving the old form in place for now, it can be removed later once we've removed the dependencies on it.

# PRT
Upstream PRT failure  can be ignored. 5.6 and 5.7 passed, basically irrelevant.

### DIFF
--- a/cfme/provisioning.py
+++ b/cfme/provisioning.py
@@ -1,7 +1,12 @@
 # -*- coding: utf-8 -*-
 from collections import OrderedDict
 
+from widgetastic.widget import View
+from widgetastic_patternfly import Tab, BootstrapSelect, Input, BootstrapTreeview
+from widgetastic_manageiq import VersionPick, Version, CheckboxSelect, Table, Calendar
+
 from cfme import web_ui as ui
+from cfme import BaseLoggedInPage
 from cfme.fixtures import pytest_selenium as sel
 from cfme.infrastructure.virtual_machines import Vm
 from cfme.services import requests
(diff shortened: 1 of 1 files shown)


cfme/provisioning.py
```diff
@@ -35,6 +40,137 @@ def select_security_group(sg):
...
+
+    @View.nested
+    class catalog(Tab):  # noqa
+        TAB_NAME = 'Catalog'
+        num_instances = BootstrapSelect('service__number_of_vms')
+        instance_name = Input(name='service__vm_name')
+        instance_description = Input(name='service__vm_description')
+        vm_filter = BootstrapSelect('service__vm_filter')
+        num_vms = BootstrapSelect('service__number_of_vms')
+        vm_name = Input(name='service__vm_name')
```
otsuman: Maybe instead of "vm_name" and "instance_name" better to add one "name" input? They have same locator.
Same with description

mshriver: Agreed, I didn't audit the form too much from its existing fields. I'll condense these though.

mshriver: pushed a 2nd commit, there were a few more duplicates as well in the other tabs.

cfme/provisioning.py
```diff
@@ -35,6 +40,126 @@ def select_security_group(sg):
...
+    @View.nested
+    class properties(Tab):  # noqa
+        TAB_NAME = 'Properties'
+        instance_type = BootstrapSelect('hardware__instance_type')
+        guest_keypair = BootstrapSelect('hardware__guest_access_key_pair')
+        hardware_monitoring = BootstrapSelect('hardware__monitoring')
+        boot_disk_size = BootstrapSelect('hardware__boot_disk_size')
+        # GCE
+        is_preemtible = VersionPick({
+            Version.lowest(): None,
```
mfalesni: I am not sure how that is gonna be handled. It might need some more investigation :)

PR #4033, opened 2017-01-19
[1LP][RFR] REST API tests for orchestration templates

mkoura:
Adding tests for /api/orchestration_templates collection.

{{pytest: '''cfme/tests/services/test_rest_services.py::TestOrchestrationTemplatesRESTAPI'' -v'}}

**PRT results**
test_delete_orchestration_templates[from_detail] is expected to fail until https://bugzilla.redhat.com/show_bug.cgi?id=1414881 is fixed

### DIFF
--- a/cfme/rest/gen_data.py
+++ b/cfme/rest/gen_data.py
@@ -15,6 +15,20 @@ from utils.log import logger
 from utils import version
 
 
+_TEMPLATE_TORSO = """{
+  "AWSTemplateFormatVersion" : "2010-09-09",
+  "Description" : "AWS CloudFormation Sample Template Rails_Single_Instance.",
+
+  "Parameters" : {
+    "KeyName": {
+      "Description" : "Name of an existing EC2 KeyPair to enable SSH access to the instances",
+      "Type": "AWS::EC2::KeyPair::KeyName",
+      "ConstraintDescription" : "must be the name of an existing EC2 KeyPair."
+    }
+  }
+}
+"""
+
 def service_catalogs(request, rest_api):
     name = fauxfactory.gen_alphanumeric()
     scls_data = [{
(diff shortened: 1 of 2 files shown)


dajoRH:
Lint report for commit 794632e64c16ed6c1c6b9d54a1b06493063b7004:

`cfme/rest/gen_data.py`:
- :bangbang: Line 406:24: **E203** *whitespace before ':'*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

cfme/rest/gen_data.py
```diff
@@ -403,3 +417,28 @@ def _finished():
             logger.warning("Failed to delete arbitration settings.")
 
     return response
+
+
+def orchestration_templates(request, rest_api, num=2):
+    collection = rest_api.collections.orchestration_templates
+    body = []
+    for _ in range(num):
```
mshriver: Consider using the number here in a meaningful way by including it in the name somewhere - for example test_{num}_{unique}

cfme/tests/services/test_rest_services.py
```diff
@@ -449,3 +455,73 @@ def test_edit_blueprints(self, rest_api, blueprints, from_detail):
...
+    @pytest.mark.tier(3)
+    @pytest.mark.uncollectif(lambda: version.current_version() < '5.7')
+    def test_create_orchestration_templates(self, rest_api, orchestration_templates):
+        """Tests creation of orchestration templates.
+
+        Metadata:
+            test_flag: rest
+        """
+        assert len(orchestration_templates) > 0
+        record = rest_api.collections.orchestration_templates.get(id=orchestration_templates[0].id)
```
mshriver: Why not assert against all of the created templates instead of just the first?

mkoura: @mshriver The elements are created the same way regardless of the number, so check that the first one is ok was enough in my opinion. Edit and delete tests are more complex and are using the same fixture for resources creation, so this is just kind of sanity test. But I'm happy to change that if you have different opinion.

cfme/rest/gen_data.py
mkoura: @mshriver The purpose is to create unique name and description for each test and just number would not help, so I don't see any advantage in including number in addition to unique string. Using underscore where value is not needed is idiomatic in python.

cfme/tests/services/test_rest_services.py
psav: I think it's worth checking them all, if there was an error where only the first was processed, and we've all seen issues like that before......

PR #4040, opened 2017-01-22
[1LP][RFR] unblock added for cla != ContainersProvider

gshefer:
{{pytest: cfme/tests/containers/test_containers_main_pages_sort.py -v }}

### DIFF
--- a/cfme/tests/containers/test_containers_main_pages_sort.py
+++ b/cfme/tests/containers/test_containers_main_pages_sort.py
@@ -8,6 +8,7 @@ from cfme.containers.service import Service
 from cfme.containers.route import Route
 from cfme.containers.provider import ContainersProvider
 from utils.appliance.implementations.ui import navigate_to
+from utils.blockers import BZ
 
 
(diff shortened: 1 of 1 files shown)


gshefer:
There is a bug in the framework that raise exception once we are trying to sort a table column with empty fields:
http://10.16.45.124/xgIo2PGx/artifacts/report.html

mshriver:
Can you revise your pytest command to run against non-container providers so we can confirm this uncollects appropriately? Thanks.

gshefer:
Do you mean '{{pytest: cfme/tests/containers/test_containers_main_pages_sort.py -v }}' right?

mshriver:
Yep, or specify a non-container type, just so we see both sides.

Looks like the failing PRT results in the last run (with container provider) are unrelated to scoping this test.

Please open a GH issue against the framework bug you mentioned in an earlier comment.

gshefer:
Done - https://github.com/ManageIQ/integration_tests/issues/4052

PR #4065, opened 2017-01-24
[1LP][RFR] Fix openstack infrastructure tests

okolisny:
Fix openstack infrastructure tests, re-organize tests under separate infrastructure directory


cfme/infrastructure/cluster.py
```diff
@@ -117,7 +117,12 @@ class All(CFMENavigateStep):
     prerequisite = NavigateToAttribute('appliance.server', 'LoggedIn')
 
     def step(self):
-        self.parent_view.navigation.select('Compute', 'Infrastructure', 'Clusters')
+        try:
```
mshriver: Is this version dependent? If so I think it would be better to navigate based the version and not just catch the exception.

cfme/tests/openstack/infrastructure/test_credentials_quad.py
```diff
@@ -13,7 +13,6 @@
 
 @pytest.mark.usefixtures("setup_provider_modscope")
 def test_credentials_quads(provider):
-    provider.load_details()
```
mshriver: :+1:

cfme/tests/openstack/infrastructure/test_hosts.py
```diff
@@ -0,0 +1,112 @@
+import pytest
+from utils import testgen
```
mshriver: Leave an empty line between the types of imports (builtin, installed, local). Move utils.testgen down with the other utils.wait import so things are more organized. If you're feeling extra pedantic organize all the local imports into alphabetical.

cfme/infrastructure/cluster.py
okolisny: Nope. It depends on what type of provider did we connect. For all providers it is 'Clusters', and only for Openstack - 'Deployment Roles'

cfme/tests/openstack/infrastructure/test_hosts.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+pytest_generate_tests = testgen.generate(testgen.providers_by_class,
+                                         [OpenstackInfraProvider],
+                                         scope='module')
+pytestmark = [pytest.mark.usefixtures("setup_provider_modscope")]
+
+
+def test_host_configuration(provider, soft_assert):
+    provider.load_details()
```
mshriver: Please use `navigate_to(provider, 'Details')` here, we plan to remove load_details() class methods.

cfme/tests/openstack/infrastructure/test_hosts.py
```diff
@@ -0,0 +1,112 @@
...
+    sel.click(InfoBlock.element("Relationships", "Nodes"))
+    my_quads = list(Quadicon.all())
+    assert len(my_quads) > 0
+    for quad in my_quads:
+        host = Host(name=quad.name)
+        host.run_smartstate_analysis()
+        wait_for(lambda: is_host_analysis_finished(host.name), delay=15,
```
mshriver: You can remove the lambda function here and pass host.name as a funcarg, wait_for() supports it.

Same for the fail_func, just set `fail_func=toolbar.refresh`

This also covers instances where the toolbar doesn't have Reload, and defaults to a selenium browser refresh.

cfme/tests/openstack/infrastructure/test_hosts.py
```diff
@@ -0,0 +1,112 @@
...
+                 timeout="10m", fail_func=lambda: toolbar.select('Reload'))
+        soft_assert(
```
mshriver: You could condense this to iterate over the list `['Packages', 'Services', 'Files']` to reduce the duplication.

cfme/tests/openstack/infrastructure/test_hosts.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def test_host_cpu_resources(provider, soft_assert):
+    provider.load_details()
+    sel.click(InfoBlock.element("Relationships", "Nodes"))
+    my_quads = list(Quadicon.all())
+    assert len(my_quads) > 0
+    for quad in my_quads:
+        host = Host(name=quad.name)
+        soft_assert(int(host.get_detail("Properties", "Number of CPUs")) > 0,
```
mshriver: The soft_assert failure messages don't quite match up with the detail you're getting, which could lead to confusion in failure debug. Recommend matching the strings exactly - which is also easier if you iterate like my comment from like 28 above. :)

cfme/infrastructure/cluster.py
mshriver: Noted, I would like to see that logic here instead of the try/except. `self.obj.provider` should be accessible to base the navigation off of the provider type.

mshriver:
@okolisny Note PRT results are false-positive, no tests were collected. I believe you'll want to specify tests and a provider with {{ pytest: }} in your description.

cfme/tests/openstack/infrastructure/test_roles.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def test_host_role_type(provider):
+    provider.load_details()
+    sel.click(InfoBlock.element("Relationships", "Nodes"))
+    my_quads = list(Quadicon.all())
+    assert len(my_quads) > 0
+    for quad in my_quads:
+        role_name = str(quad.name)
+        role_name = re.search(r'\((\w+)\)', role_name).group(1)
```
mshriver: It looks like the intent here is to strip parenthesis from role_name. Consider string.translate() instead of using `re`.

rrasouli: @mfalesni big fan of re :-)

dajoRH:
*CFME QE Bot*

dajoRH:
Lint report for commit e4332195fe637795f944445be4a9f5fc6be42075:

* e4332195fe637795f944445be4a9f5fc6be42075

*CFME QE Bot*

dajoRH:
Lint report for commit 1bc74797930b2cb22907831b5829092f4179bab0:

* e4332195fe637795f944445be4a9f5fc6be42075

*CFME QE Bot*

dajoRH:
Lint report for commit eca9f0075fd468edcd27953adec7617a80273941:

* e4332195fe637795f944445be4a9f5fc6be42075

*CFME QE Bot*

cfme/infrastructure/cluster.py
okolisny: self.obj.provider raises
AttributeError: type object 'Cluster' has no attribute 'provider'
Don't know how can it be

mshriver:
@okolisny you need to rebase (not merge) against master, right now you've got commits from master showing up in your branch :)

dajoRH:
Lint report for commit 987243f195a1662a47c8461e3fc9720bc5140f4c:

* e4332195fe637795f944445be4a9f5fc6be42075

*CFME QE Bot*

cfme/infrastructure/cluster.py
mshriver: Did you navigate on a Cluster object or class? (message seems to indicate object). Its not optional, and the object attribute is set in __init__.... so that is strange.

That aside, I think this may warrant a separate navigation destination, 'Deployment Roles', to be used by Openstack clusters.  Otherwise this 'All' destination, which should be navigatable by the class, would now require an instantiated Cluster to navigate.

cfme/tests/openstack/infrastructure/test_roles.py
okolisny: I don't understand how to use 'translate' here. Maybe using of re is more readable?

okolisny:
@mshriver All done. I've created a new nav page for Deployment roles. I think it is better than adding a lot of hooks into Cluster :)

cfme/tests/openstack/infrastructure/test_roles.py
mshriver: That's fine, it can stay as-is.

To use translate here, you would use None as the first argument - the 'translation table' - which will pass the non-deleted characters through unchanged.

```python
string = '(my_role_name)'
print string.translate(None, '()')
# output: my_role_name
```

cfme/infrastructure/deployment_roles.py
```diff
@@ -0,0 +1,76 @@
...
+    Args:
+        name: Name of the role.
+        provider: provider this role is attached to
+            (deployment roles available only for Openstack!).
+    """
+    pretty_attrs = ['name', 'provider']
+
+    def __init__(self, name, provider, appliance=None):
+        Navigatable.__init__(self, appliance=appliance)
+        if not isinstance(provider, OpenstackInfraProvider):
```
mshriver: I don't think this is quite the right place to catch this exception. Possibly a `NotImplementedError` would be more appropriate.

Is the navigation menu available only when a provider has been configured?  It looks like you can navigate to 'All' without having called `__init__`, just by importing DeploymentRoles.

cfme/infrastructure/deployment_roles.py
```diff
@@ -0,0 +1,76 @@
...
+    def am_i_here(self):
+        return match_location(match_page)
+
+
+@navigator.register(DeploymentRoles, 'Details')
+class Details(CFMENavigateStep):
+    prerequisite = NavigateToSibling('All')
+
+    def step(self):
+        sel.click(Quadicon(self.obj.name))
```
mshriver: In the list view, can you match both the provider and deployment role names? That would be preferable since two different providers could have the same name for their unique deployment roles.

cfme/infrastructure/deployment_roles.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def am_i_here(self):
+        return match_page(summary="{} (Summary)".format(self.obj.name))
+
+
+@navigator.register(DeploymentRoles, 'DetailsFromProvider')
+class DetailsFromProvider(CFMENavigateStep):
```
mshriver: Reconsider the destinations and your use of am_i_here()

I think you want to capture 4 destinations, but only 1 Details page. In either case the details page is identical, what you want to capture is the navigation path for Details.

Using am_i_here() in both of your Details destinations means if you're on the page and attempt to re-navigate from the other path, it will short-circuit and not navigate because the page is the same.

Consider these destinations, naming them as you see fit:
* All
* AllByProvider
* Details
* DetailsFromProvider

cfme/infrastructure/deployment_roles.py
okolisny: > Is the navigation menu available only when a provider has been configured?
yes, deployment roles exist only if Openstack infra provider was added.

> It looks like you can navigate to 'All' without having called __init__
yes, and so? No concern, as for me.

cfme/infrastructure/deployment_roles.py
okolisny: Ah, sounds reasonable. Btw, I was taking cluster.py as example, so there are incorrect navigation too..

cfme/infrastructure/deployment_roles.py
mshriver: The concern is that this exception will not be raised during that navigation if no Openstack infra provider is configured.

I can do the following, which would result in a `NoSuchElement` exception, not a DestinationNotFound.
```python
from cfme.infrastructure.deployment_roles import DeploymentRoles
from utils.appliance.implementations.ui import navigate_to

navigate_to(DeploymentRoles, 'All')
```

I think you want to use a different type of exception here, and then in the step() method for the 'all' destination either catch the NoSuchElement exception and raise it was a DestinationNotFound.

okolisny: Understood. So, catching NoSuchElement and then raising DestinationNotFound is a good way to do, right?

cfme/infrastructure/deployment_roles.py
okolisny: > In the list view, can you match both the provider and deployment role names?

I'm not sure, there is no List View on that page.
screenshot from 2017-01-31 19-35-52

cfme/infrastructure/deployment_roles.py
mshriver: Noted - that was converted kinda early, and we were mostly driven by what navigation destinations were needed at the time.

Since this looks like a navigation path that is commonly used, I think its worth having the separate destinations for. We can update cluster as the destination is needed there.

cfme/infrastructure/deployment_roles.py
mshriver: I think so (in the navigation step, not here). It keeps the 'All' destination independent from an instantiated object and provides a deterministic behavior for when the destination isn't in the menu.  This allows you to test/verify that the menu item is unavailable until the provider is configured

cfme/infrastructure/deployment_roles.py
mshriver: Ah, but there is! Check out the buttons in the top right, in the toolbar. Grid View, Tile View, and List View.

cfme.web_ui provides 'toolbar' (often imported as 'tb' alias), which will allow you to change the view. Grid is just the default in this case. Usually the List view (a table) provides a more accurate way to select a specific resource.

okolisny: Yes, but there is
screenshot from 2017-01-31 19-51-34
 no relation to provider anyway..

okolisny:
@mshriver Set to RFR, I've rewrote navigations as you mentioned above.

cfme/infrastructure/deployment_roles.py
okolisny: done

cfme/infrastructure/deployment_roles.py
rrasouli: @mshriver I will be adding AllByProvider on PR #3907 , if we could have it reviewed and merged

okolisny: @rrasouli are you talking about Deployment Roles or Cluster?

rrasouli:
I was referring the provider navigate to all nodes

Thanks,

Ronnie Rasouli
Quality Engineer RHOS CloudForms
Red Hat

cfme/infrastructure/deployment_roles.py
mshriver: @rrasouli Let's go with 'AllForProvider' for that destination name. @izapolsk chose this for the common VM destination and we should keep things consistent. I've updated cloud instances to use 'AllForProvider'. There's once instance of 'AllByProvider' in cloud.volume but we can convert it later.

cfme/infrastructure/deployment_roles.py
```diff
@@ -0,0 +1,97 @@
...
+        try:
+            self.prerequisite_view.navigation.select(*nav_path)
+        except NoSuchElementException:
+            raise DestinationNotFound('Navigation {} not found'.format(nav_path))
+
+    def am_i_here(self):
+        return match_location(match_page)
+
+
+@navigator.register(DeploymentRoles, 'AllByProvider')
```
mshriver: slight tweak, let's go with 'AllForProvider'

cfme/infrastructure/deployment_roles.py
okolisny: @mshriver does it mean that I have to rename AllByProvider navigation to AllForProvider for Deployment Roles?

cfme/infrastructure/deployment_roles.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def step(self):
+        nav_path = ('Compute', 'Infrastructure', 'Deployment Roles')
+        try:
+            self.prerequisite_view.navigation.select(*nav_path)
+        except NoSuchElementException:
+            raise DestinationNotFound('Navigation {} not found'.format(nav_path))
+
+    def am_i_here(self):
+        return match_location(match_page)
```
mshriver: **REQUIRED**
Check your use of the partial here, you want to call `match_page(summary='[unique summary]')`

cfme/infrastructure/deployment_roles.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+@navigator.register(DeploymentRoles, 'AllByProvider')
+class AllByProvider(CFMENavigateStep):
+
+    prerequisite = NavigateToAttribute('appliance.server', 'LoggedIn')
+
+    def step(self):
+        navigate_to(self.obj.provider, 'Details')
```
mshriver: **REQUIRED**
This is your prerequisite, not LoggedIn

cfme/infrastructure/deployment_roles.py
```diff
@@ -0,0 +1,97 @@
...
+        return match_page(summary=summary)
+
+
+@navigator.register(DeploymentRoles, 'Details')
+class Details(CFMENavigateStep):
+
+    prerequisite = NavigateToSibling('All')
+
+    def step(self):
+        sel.click(Quadicon(self.obj.name))
```
mshriver: This assumes the page is on Grid View, which could have been changed before navigating. You should set the view here before trying to interact with the quadicon.

cfme/infrastructure/deployment_roles.py
```diff
@@ -0,0 +1,97 @@
...
+        return match_page(summary="{} (Summary)".format(self.obj.name))
+
+
+@navigator.register(DeploymentRoles, 'DetailsFromProvider')
+class DetailsFromProvider(CFMENavigateStep):
+
+    prerequisite = NavigateToSibling('AllByProvider')
+
+    def step(self):
+        sel.click(Quadicon(self.obj.name))
```
mshriver: **REQUIRED**
Again, assuming grid view. set it first

cfme/infrastructure/deployment_roles.py
mshriver: Yeah, that's my fault for originally suggesting AllByProvider, apologies. I commented in-line below for the change as well.

cfme/tests/openstack/infrastructure/test_api_port.py
```diff
@@ -11,5 +11,6 @@
 
 @pytest.mark.usefixtures("setup_provider_modscope")
 def test_api_port(provider, soft_assert):
-    soft_assert(provider.summary.properties.api_port.value.isdigit(),
+    port = provider.get_yaml_data()['port']
+    soft_assert(provider.summary.properties.api_port.value == port,
```
mshriver: **REQUIRED**
No use in having a soft_assert at the end of the test, since nothing else is dependent on it. I think it should be a full assert.

cfme/tests/openstack/infrastructure/test_relationships.py
```diff
@@ -4,52 +4,50 @@
...
-    provider.load_details()
-    result = provider.summary.relationships.orchestration_stacks.value
+    navigate_to(provider, 'Details')
     """
     todo get the list of tenants from external resource and compare
     it with result - currently not 0
     """
 
-    assert result > 0
+    assert int(provider.get_detail('Relationships', 'Orchestration stacks')) > 0
```
mshriver: **OPTIONAL**
Consider parametrizing on ['Deployment Roles', 'Nodes', 'Templates', 'Orchestration stacks'] since all of these tests are the same and only the 2nd arg to get_detail changes.

cfme/tests/openstack/infrastructure/test_resources.py
```diff
@@ -0,0 +1,29 @@
...
+
+pytest_generate_tests = testgen.generate([OpenstackInfraProvider],
+                                         scope='module')
+
+
+@pytest.mark.usefixtures("setup_provider_modscope")
+def test_number_of_cpu(provider, soft_assert):
+    navigate_to(provider, 'Details')
+    v = provider.get_detail('Properties', 'Aggregate Node CPU Resources')
+    soft_assert(int(v.split()[0]) > 0, "Aggregate Node CPU Resources is 0")
```
mshriver: why soft_assert and not assert here?

cfme/tests/openstack/infrastructure/test_resources.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+pytestmark = [pytest.mark.meta(server_roles='+smartproxy +smartstate')]
+
+
+pytest_generate_tests = testgen.generate([OpenstackInfraProvider],
+                                         scope='module')
+
+
+@pytest.mark.usefixtures("setup_provider_modscope")
```
mshriver: **REQUIRED**
move this into the pytestmark list above.

cfme/tests/openstack/infrastructure/test_hosts.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def test_host_hostname(provider, soft_assert):
+    provider.summary.relationships.nodes.click()
+    my_quads = list(Quadicon.all())
+    assert len(my_quads) > 0
+    for quad in my_quads:
+        host = Host(name=quad.name)
+        result = host.get_detail("Properties", "Hostname")
+        soft_assert(result) != '', "Missing hostname in: " + str(host)
```
mshriver: is this a bug if the hostname is missing? suggest a hard assert if so.

cfme/tests/openstack/infrastructure/test_hosts.py
```diff
@@ -0,0 +1,105 @@
...
+def test_host_zones_assigned(provider):
+    navigate_to(provider, 'Details')
+    sel.click(InfoBlock.element("Relationships", "Nodes"))
+    my_quads = list(Quadicon.all())
+    assert len(my_quads) > 0
+    my_quads = filter(lambda q: True if q.name == 'Compute' else False,
+                      my_quads)
+    for quad in my_quads:
+        host = Host(name=quad.name)
+        result = int(host.get_detail('Relationships', 'Availability Zone'))
```
mshriver: You're casting the result to int() and then asserting against string 'nova'?

cfme/infrastructure/deployment_roles.py
okolisny: How can I access provider obj there ? I can't refer to 'self' in class fields

cfme/infrastructure/deployment_roles.py
okolisny: hm, done

cfme/tests/openstack/infrastructure/test_hosts.py
okolisny: This is something strange that I've missed. Rewriting..

cfme/tests/openstack/infrastructure/test_hosts.py
okolisny: Oh, thank you for this. Rewriting

cfme/tests/openstack/infrastructure/test_relationships.py
okolisny: I'd like to keep them separately as there are different test cases in Polarion

cfme/tests/openstack/infrastructure/test_resources.py
okolisny: Replaced last soft_assert with assert

cfme/infrastructure/deployment_roles.py
mshriver: You can define prerequisite as a method like `step` or `am_i_here`. We just generally use the shorter syntax since its almost always a NavigateToAttribute or NavigateToSibling call.

cfme/tests/openstack/infrastructure/test_relationships.py
mshriver: Understood. There may be a way around that? You'll still have multiple runs of the test that have unique names, so maybe its possible to correctly identify the cases in Polarion. I just started using it this week so maybe someone like @anewmanRH can help us out here?

cfme/infrastructure/deployment_roles.py
```diff
@@ -0,0 +1,99 @@
...
+        return match_page(summary=summary)
+
+
+@navigator.register(DeploymentRoles, 'Details')
+class Details(CFMENavigateStep):
+
+    prerequisite = NavigateToSibling('All')
+
+    def step(self):
+        tb.select('Grid View')
```
mshriver: :+1:

cfme/tests/openstack/infrastructure/test_relationships.py
mshriver: Keep in mind this is optional, it can stay as-is but I'm curious while the PR is open because I'll be running into the same thing before too long.

cfme/infrastructure/deployment_roles.py
okolisny: @mshriver so prerequisite can be implemented as a simple instance method? it should not be static only, right?

cfme/tests/openstack/infrastructure/test_relationships.py
okolisny: To be fair enough, we need to implement more realistic asserts there, for example get stacks/templtes etc. from mgmt_system for Openstack, and compare value with the result from UI. In this case tests will not look so simple, but this is not a priority for us now. I'd like to have exactly these tests merged

cfme/infrastructure/deployment_roles.py
okolisny: @mshriver ah, I found the examples. Ok

okolisny:
@mshriver All done, you might consider to review it end-to-end :-)

cfme/infrastructure/deployment_roles.py
```diff
@@ -89,7 +89,7 @@ def am_i_here(self):
 @navigator.register(DeploymentRoles, 'DetailsFromProvider')
 class DetailsFromProvider(CFMENavigateStep):
 
-    prerequisite = NavigateToSibling('AllByProvider')
+    prerequisite = NavigateToSibling('AllForProvider')
```
mshriver: :wink:

okolisny: it was close enough.. :)

cfme/tests/openstack/infrastructure/test_relationships.py
psav: We will have a fix for this soon so that we can parametrize

cfme/tests/openstack/infrastructure/test_api_port.py
```diff
@@ -10,6 +10,7 @@
 
 
 @pytest.mark.usefixtures("setup_provider_modscope")
-def test_api_port(provider, soft_assert):
-    soft_assert(provider.summary.properties.api_port.value.isdigit(),
-                "Invalid API Port")
+def test_api_port(provider):
+    port = provider.get_yaml_data()['port']
```
rrasouli: I found this yaml file value useless, since in case of SSL port changes to 13000 by default. 
Better way doing it by parsing the auth URL port
provider.get_yaml_data()['auth_url'].split(':')[1].split('/')[0]

okolisny: @rrasouli hm, yes, but it is better to change yaml I think

rrasouli:
I believe so, however this value is useless we do not fill the API port
from yaml files - done automatically from the security settings

okolisny:
Hm looks like I do not know about it
At least I was filling this  value by myself always

PR #3869, opened 2016-12-11
Added provider custom attributes functions and tests

gshefer:
{{pytest: cfme/tests/containers/test_static_custom_attributes.py -v --use-provider cm-env2 }}

Added custom attributes CRUD functionality to openshift provider.
Added custom attributes CRUD tests.

PR #4077, opened 2017-01-26
[1LP][RFR] Custom attributes - new tests

gshefer:
{{pytest: cfme/tests/containers/test_static_custom_attributes.py -v --use-provider cm-env1 }}

### DIFF
--- a/cfme/containers/provider/openshift.py
+++ b/cfme/containers/provider/openshift.py
@@ -30,7 +30,7 @@ class OpenshiftProvider(ContainersProvider):
         # probably a js wait issue, not reproducible manually
         super(OpenshiftProvider, self).create(validate_credentials=validate_credentials, **kwargs)
 
-    def _href(self):
+    def href(self):
         return self.appliance.rest_api.collections.providers\
             .find_by(name=self.name).resources[0].href
 
@@ -73,7 +73,7 @@ class OpenshiftProvider(ContainersProvider):
     def custom_attributes(self):
         """returns custom attributes"""
         response = self.appliance.rest_api.get(
-            path.join(self._href(), 'custom_attributes'))
+            path.join(self.href(), 'custom_attributes'))
         out = []
         for attr_dict in response['resources']:
             attr = self.appliance.rest_api.get(attr_dict['href'])
@@ -109,7 +109,7 @@ class OpenshiftProvider(ContainersProvider):
             if fld_tp:
                 payload['resources'][i]['field_type'] = fld_tp
         return self.appliance.rest_api.post(
-            path.join(self._href(), 'custom_attributes'), **payload)
+            path.join(self.href(), 'custom_attributes'), **payload)
 
     def edit_custom_attributes(self, *custom_attributes):
         """Editing static custom attributes in provider.
@@ -132,18 +132,24 @@ class OpenshiftProvider(ContainersProvider):
                 "value": ca.value
             } for ca in custom_attributes]}
         return self.appliance.rest_api.post(
-            path.join(self._href(), 'custom_attributes'), **payload)
+            path.join(self.href(), 'custom_attributes'), **payload)
 
-    def delete_custom_attributes(self, *names):
+    def delete_custom_attributes(self, *custom_attributes):
         """Deleting static custom attributes from provider.
         Args:
-            names: The names of the custom attributes to delete.
+            custom_attributes: The custom attributes to delete.
+                               (Could be also names (str))
         returns: response.
         """
-        for name in names:
-            if type(name) is not str:
-                raise TypeError('Type of names should be {}. ({} != {})'
-                                .format(str, type(name), str))
+        names = []
+        for attr in custom_attributes:
+            attr_type = type(attr)
+            if attr_type in (str, CustomAttribute):
+                names.append(attr if attr_type is str else attr.name)
+            else:
+                raise TypeError('Type of arguments should be either'
+                                'str or CustomAttribute. ({} not in [str, CustomAttribute])'
+                                .format(type(attr)))
         attribs = self.custom_attributes()
         if not names:
             names = [attr.name for attr in attribs]
@@ -153,4 +159,4 @@ class OpenshiftProvider(ContainersProvider):
                 "href": attr.href,
             } for attr in attribs if attr.name in names]}
         return self.appliance.rest_api.post(
-            path.join(self._href(), 'custom_attributes'), **payload)
+            path.join(self.href(), 'custom_attributes'), **payload)
--- a/cfme/tests/containers/test_static_custom_attributes.py
+++ b/cfme/tests/containers/test_static_custom_attributes.py
@@ -1,3 +1,8 @@
+from string import digits, ascii_letters
+from random import choice
+import sys
+import inspect
+import re
 import pytest
 
 from manageiq_client.api import APIException
@@ -6,6 +11,8 @@ from utils.version import current_version
 from utils import testgen
 from cfme.containers.provider import ContainersProvider
 from cfme.containers.provider.openshift import CustomAttribute
+from traceback import format_exception_only
+from os import path
 
 pytestmark = [
     pytest.mark.uncollectif(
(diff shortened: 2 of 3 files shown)


cfme/containers/provider/openshift.py
```diff
@@ -30,7 +30,7 @@ def create(self, validate_credentials=True, **kwargs):
         # probably a js wait issue, not reproducible manually
         super(OpenshiftProvider, self).create(validate_credentials=validate_credentials, **kwargs)
 
-    def _href(self):
+    def href(self):
```
mshriver: Any intent here beyond just changing the name?

gshefer: Yes: Private (protected) -> Public.
I need to get this outside the class...

cfme/tests/containers/test_static_custom_attributes.py
```diff
@@ -6,6 +11,8 @@
 from utils import testgen
 from cfme.containers.provider import ContainersProvider
 from cfme.containers.provider.openshift import CustomAttribute
+from traceback import format_exception_only
```
mshriver: Move these up with the other built-in imports.

cfme/tests/containers/test_static_custom_attributes.py
```diff
@@ -14,14 +21,43 @@
     pytest.mark.tier(2)]
 pytest_generate_tests = testgen.generate([ContainersProvider], scope='function')
 
+
+def get_random_string(length):
```
mshriver: It looks like `fauxfactory.gen_alphanumeric()` would be sufficient here? If you require some symbols you can always concatenate those, and you can still replace multiple spaces with a single.  Was there a reason you decided to implement instead of using the built-in?

gshefer: well, didn't know about fauxfactory - I'll check this and probably use it.
and yes - there are some printable characters that I want to prevent from using them.
Let me investigate fauxfactory, if it'll yield what I want I'll change accordingly :)

cfme/containers/provider/openshift.py
mshriver: https://docs.python.org/2/tutorial/classes.html#private-variables-and-class-local-references

It shouldn't need to change in order to access the method outside of the class. I'm not against the change - did you find that you were unable to call the method with the underscore in the name?

cfme/tests/containers/test_static_custom_attributes.py
```diff
@@ -14,14 +21,43 @@
...
 ATTRIBUTES_DATASET = [
     CustomAttribute('exp_date', '2017-01-02', 'Date'),
     CustomAttribute('sales_force_acount', 'ADF231VRWQ1', None),
-    CustomAttribute('expected_num_of_nodes', '2', None),
+    CustomAttribute('expected_num_of_nodes', '2', None)
 ]
 VALUE_UPDATES = ['2018-07-12', 'ADF231VRWQ1', '1']
 
 
+def setup_dataset(provider, verify_exists=False):
```
mshriver: Still reading the new test functions, but I'm skeptical of this method name.

I'm thinking it would be better to implement the setup as a fixture, and create a separate method to delete DATASET attributes (verify_exists=False mode)

cfme/tests/containers/test_static_custom_attributes.py
```diff
@@ -77,12 +112,9 @@ def test_delete_static_custom_attributes(provider):
...
         (you should not see a custom attributes table)
     """
-    # Checking that all ATTRIBUTES_DATASET was added.
-    attribs_names = [attr.name for attr in provider.custom_attributes()]
-    assert all([attr.name in attribs_names
-                for attr in ATTRIBUTES_DATASET])
+    setup_dataset(provider, verify_exists=True)
 
-    provider.delete_custom_attributes()
+    provider.delete_custom_attributes(*ATTRIBUTES_DATASET)
```
mshriver: setup_dataset() makes the same call, but with the added protection of only passing attributes that are in DATASET and are present. Intentionally skipping that protection here?

cfme/containers/provider/openshift.py
gshefer: I didn't event tried (It'll work) but it just a convention:
http://docs.quantifiedcode.com/python-code-patterns/correctness/accessing_a_protected_member_from_outside_the_class.html

I think this method should be public and not protected.
If you insist - I can keep it protected and call it outside - but this is not recommended...

mshriver: Oh I don't see a problem with it - its just that it wasn't actually 'protected' by python. Wanted to make sure there wasn't a secondary intent here, something else I was missing.

cfme/tests/containers/test_static_custom_attributes.py
gshefer: You're right - better to make this as a fixture - I'll change accordingly...

cfme/tests/containers/test_static_custom_attributes.py
```diff
@@ -124,3 +158,79 @@ def test_add_date_value_with_wrong_value(provider):
...
+        #       specify the href and yield the exception
+        payload = {
+            "action": "edit",
+            "resources": [{
+                "href": '{}/custom_attributes/9876543210000000'
+                        .format(provider.href()),
+                "value": ca.value
+            }]}
+        provider.appliance.rest_api.post(
+            path.join(provider.href(), 'custom_attributes'), **payload)
```
mshriver: check out pytest.raises() instead of this try/except format.

cfme/containers/provider/openshift.py
gshefer: LOL, forgot to say that I've added this method (https://github.com/ManageIQ/integration_tests/pull/3869/files) so there is no secondary intent AFAIK :)

cfme/tests/containers/test_static_custom_attributes.py
```diff
@@ -131,14 +131,12 @@ def test_add_attribute_with_empty_name(provider):
         * You should get an error
         * You should not see this attribute in the custom  attributes table
     """
-    try:
+    with pytest.raises(APIException):
         provider.add_custom_attributes(
             CustomAttribute('', "17")
         )
         pytest.fail('You have added custom attribute with empty name'
```
mshriver: This will take some verification but... I think the pytest.fail isn't necessary anymore.  If the exception isn't raised the test will fail with a message like: `Failed: Expecting APIException`.

Did you test locally and find that this pytest.fail was evaluated on no raised exception?

Doesn't look like raises() takes an option for a failure message if the exception isn't raised.

mshriver: The current message doesn't provide much additional context, if you want some more context or assertion behavior on failure check out the examples again, they have 1-2 with an assert following a pytest.raises  - but its not inside the `with():` block.

cfme/containers/provider/openshift.py
mshriver: seems our message order got jumbled for some reason - but we agree, this change is fine.

cfme/tests/containers/test_static_custom_attributes.py
mshriver: Good switch to pytest.raises - but a note. in general, try to limit what is in the `with()` block to the call that will actually raise the exception. So setup `payload` beforehand and only make your rest_api call in the `with()`

cfme/tests/containers/test_static_custom_attributes.py
gshefer: After deep investigation, looks like fauxfactory cannot yield what I need, I need EXACTLY what get_random_string does and I think its functionality is not generic enough to add it to fauxfactory.
Let's keep this function here, it serves only the test scope.

cfme/tests/containers/test_static_custom_attributes.py
```diff
@@ -14,10 +19,17 @@
     pytest.mark.tier(2)]
 pytest_generate_tests = testgen.generate([ContainersProvider], scope='function')
 
+
+def get_random_string(length):
+    valid_chars = digits + ascii_letters + ' !@#$%^&*()'
+    out = ''.join([choice(valid_chars) for _ in xrange(length)])
+    return re.sub('\s+', ' ', out)
```
quarckster: `re.sub('\s{2,}', ' ', out)`

PR #4080, opened 2017-01-26
[1LP] [RFR] ExpressionEditor fix

quarckster:
{{pytest: -v -k "test_modify_vm_condition_expression" --long-running}}

Purpose or Intent
=================

__Added__ `VersionPick` into commit and discard buttons. Also `ExpressionEditor` has been converted to View.

### DIFF
--- a/cfme/web_ui/expression_editor_widgetastic.py
+++ b/cfme/web_ui/expression_editor_widgetastic.py
@@ -14,70 +14,70 @@ from widgetastic_manageiq import Calendar
 from widgetastic.utils import VersionPick, Version
 
 
-class FieldFormView(View):
-    type = BootstrapSelect("chosen_typ")
-    field = BootstrapSelect("chosen_field")
-    key = BootstrapSelect("chosen_key")
-    value = Input("chosen_value")
-    user_input = Input("user_input")
-
-
-class FieldDateForm(View):
-    dropdown_select = BootstrapSelect("chosen_from_1")
-    input_select_date = Calendar("miq_date_1_0")
-    input_select_time = BootstrapSelect("miq_time_1_0")
-
-
-class CountFormView(View):
-    type = BootstrapSelect("chosen_typ")
-    count = BootstrapSelect("chosen_count")
-    key = BootstrapSelect("chosen_key")
-    value = Input("chosen_value")
-    user_input = Input("user_input")
-
-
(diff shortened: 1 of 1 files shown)


quarckster:
cast @mfalesni

cfme/web_ui/expression_editor_widgetastic.py
```diff
@@ -8,12 +8,71 @@
...
 from widgetastic_manageiq import Calendar
+from widgetastic.utils import VersionPick, Version
 
 
-class ExpressionEditor(Widget, Pretty):
+class FieldFormView(View):
+    type = BootstrapSelect("chosen_typ")
+    field = BootstrapSelect("chosen_field")
+    key = BootstrapSelect("chosen_key")
+    value = Input("chosen_value")
```
mshriver: use `name=` for Input

cfme/web_ui/expression_editor_widgetastic.py
```diff
@@ -8,12 +8,71 @@
...
+    user_input = Input("user_input")
+
+
+class FieldDateForm(View):
+    dropdown_select = BootstrapSelect("chosen_from_1")
+    input_select_date = Calendar("miq_date_1_0")
```
mshriver: use `name=` for Calendar since it extends TextInput

cfme/web_ui/expression_editor_widgetastic.py
```diff
@@ -8,12 +8,71 @@
...
+
+class RegistryFormView(View):
+    type = BootstrapSelect("chosen_typ")
+    key = Input("chosen_regkey")
+    value = Input("chosen_regval")
+    operation = BootstrapSelect("chosen_key")
+    contents = Input("chosen_value")
+
+
+class DateSpecificForm(View):
```
mshriver: Two things:

1. Naming consistency - ThingFormView, ThingFormView, ThingForm - make them all one or the other, I'm partial to calling them 'Form'
2. These appear to be nested views - why define these outside of the ExpressionEditor, and then set class attributes for each? This treats them like widgets instead of like nested views. This may be correct, I'm still learning, @mfalesni should probably weigh in on this.

mshriver: I guess you liked thing #2 :)

PR #4083, opened 2017-01-26
[1LP][RFR] Remove old setup_a_provider

jkrocil:
WIP - a follow-up commit will enforce (hopefully) all provider setup through fixtures. That's the plan, anyway.

Edit: will be a separate PR. This one is ready.

### DIFF
--- a/cfme/rest/gen_data.py
+++ b/cfme/rest/gen_data.py
@@ -5,10 +5,11 @@ from manageiq_client.api import APIException
 
 from cfme.automate.service_dialogs import ServiceDialog
 from cfme.exceptions import OptionNotAvailable
+from cfme.infrastructure.provider import InfraProvider
 from cfme.services.catalogs.catalog_item import CatalogItem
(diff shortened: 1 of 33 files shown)


cfme/rest/gen_data.py
```diff
@@ -238,7 +240,7 @@ def _finished():
 
 
 def a_provider():
```
psav: Wondering if we should normalize all these
we have ```a_provider``` here and ```setup_a_provider``` else where. I guess we should look at what the fixture actually does. ```a_provider``` is possibly better than ```setup_a_provider```, thoughts?

jkrocil: Yes. There is also stuff like `setup_first_provider`, `setup_single_provider`, `vmware_provider` and also just `provider`. All of that will (like I said "hopefully) come together nicely and be unified.

dajoRH:
*CFME QE Bot*

dajoRH:
*CFME QE Bot*

PR #4084, opened 2017-01-26
[1LP][RFR] Update widgetastic_miq.PaginationPane, bug fixes

mshriver:
No tests using this widget yet, PRT results are just smoke tests.

Updated PaginationPane.exists to just return self.is_displayed - I'm not sure why it wasn't this simple to begin with? Previous code raised exception on screens without paginator.

Updated _parse_pages() calculation of current page and page amount:
* Removed 'corner' cases of modulo and adding 1 (was necessary for page amount and missing)
* Changed to float division instead of integer (floored) division and used math.ceil to round up

I tested locally with:
* `.pages()` on a screen without pages
* `pages()` on a screen with 1 page
* `pages()` on a screen with multiple pages
* `exists` on a screen without paginator (False)
* `exists` on a screen with paginator (True)

### DIFF
--- a/widgetastic_manageiq.py
+++ b/widgetastic_manageiq.py
@@ -875,6 +875,21 @@ class PaginationPane(View):
     def last_page(self):
         self.paginator.last_page()
 
+    def pages(self):
+        """Generator to iterate over pages, yielding after moving to the next page"""
+        if self.exists:
+            # start iterating at the first page
+            self.first_page()
+            while self.cur_page < self.pages_amount:
+                # Aren't on the last page yet, advance
+                self.next_page()
+                yield
+            else:
+                # last page reached
+                pass
+        else:
+            yield
+
     @property
     def items_amount(self):
         return self.paginator.page_info()[1]
--- a/widgetastic_manageiq.py
+++ b/widgetastic_manageiq.py
@@ -3,6 +3,7 @@ import re
 from datetime import date
 from jsmin import jsmin
 from selenium.common.exceptions import WebDriverException
+from math import ceil
 
 from widgetastic.exceptions import NoSuchElementException
 from widgetastic.utils import VersionPick, Version
@@ -808,8 +809,7 @@ class PaginationPane(View):
 
     @property
     def exists(self):
-        cur_view = self.browser.element(self)
-        return False not in self.browser.classes(cur_view)
+        return self.is_displayed
 
     def check_all(self):
         self.check_all_items.fill(True)
@@ -827,7 +827,8 @@ class PaginationPane(View):
 
     @property
     def items_per_page(self):
-        return int(self.items_on_page.selected_option)
+        selected = self.items_on_page.selected_option
+        return int(re.sub(r'\sitems', '', selected))
 
     def set_items_per_page(self, value):
         self.items_on_page.select_by_visible_text(str(value))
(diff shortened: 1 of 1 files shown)


mshriver:
Tagged @otsuman and @izapolsk for first review, up to you guys.

@izapolsk check out `exists` if nothing else - I may be missing something as to why it was written that way.

widgetastic_manageiq.py
```diff
@@ -808,8 +809,7 @@ class PaginationPane(View):
 
     @property
     def exists(self):
-        cur_view = self.browser.element(self)
-        return False not in self.browser.classes(cur_view)
+        return self.is_displayed
```
izapolsk: awesome , I have missed this obvious way. Thanks.
I'd add some smoke tests for every widget if more things were converted to widgetastic.

widgetastic_manageiq.py
```diff
@@ -827,7 +827,8 @@ def sorted_by(self):
 
     @property
     def items_per_page(self):
-        return int(self.items_on_page.selected_option)
+        selected = self.items_on_page.selected_option
+        return int(re.sub(r'\sitems', '', selected))
```
izapolsk: I'd add here some quantifier like '\s+'

widgetastic_manageiq.py
```diff
@@ -842,16 +843,16 @@ def _parse_pages(self):
         # obtaining amount of existing pages, there is 1 page by default
         if item_amt == 0:
             page_amt = 1
-        elif item_amt % items_per_page != 0:
-            page_amt = item_amt // items_per_page + 1
         else:
-            page_amt = item_amt // items_per_page
+            # round up after dividing total item count by per-page
+            page_amt = int(ceil(float(item_amt) / float(items_per_page)))
```
izapolsk: looks more gracefully :+1: . I thought of using math but didn't want to add additional dependencies.

widgetastic_manageiq.py
```diff
@@ -875,6 +876,27 @@ def first_page(self):
     def last_page(self):
         self.paginator.last_page()
 
+    def pages(self):
```
izapolsk: awesome. this thing came to my mind then but I prefer to new code only when it is really necessary.

widgetastic_manageiq.py
```diff
@@ -875,6 +876,27 @@ def first_page(self):
...
+        """Generator to iterate over pages, yielding after moving to the next page"""
+        if self.exists:
+            # start iterating at the first page
+            if self.cur_page != 1:
+                logger.debug('Resetting paginator to first page')
+                self.first_page()
+
+            # Adding 1 to pages_amount to include the last page in loop
+            for page in range(1, self.pages_amount + 1):
```
izapolsk: you can use just range(self._pages_amount) since you don't use the number somewhere.

izapolsk:
@mshriver, you really made it better, thanks.

widgetastic_manageiq.py
mshriver: Not quite sure what you mean here - I still need to add 1 to `pages_amount` to get the correct behavior for the last page, do you mean drop the 'page' assignment like:

```python
for _ in (self.pages_amount + 1):
```

widgetastic_manageiq.py
mfalesni: Neat :laughing:

widgetastic_manageiq.py
```diff
@@ -875,6 +876,27 @@ def first_page(self):
     def last_page(self):
         self.paginator.last_page()
 
+    def pages(self):
+        """Generator to iterate over pages, yielding after moving to the next page"""
+        if self.exists:
+            # start iterating at the first page
+            if self.cur_page != 1:
+                logger.debug('Resetting paginator to first page')
```
mfalesni: `self.logger` ! :)

widgetastic_manageiq.py
```diff
@@ -875,6 +876,27 @@ def first_page(self):
...
+                self.first_page()
+
+            # Adding 1 to pages_amount to include the last page in loop
+            for page in range(1, self.pages_amount + 1):
+                yield self.cur_page
+                if self.cur_page == self.pages_amount:
+                    # last or only page, stop looping
+                    break
+                else:
+                    logger.debug('Paginator advancing to next page')
```
mfalesni: dtto.

widgetastic_manageiq.py
mshriver: ah okay I'll check this out

mshriver:
@mfalesni loggers fixed

PR #4097, opened 2017-01-30
[WIPTEST] Get metrics hostname from openshift and check it is up.

### LINKED ISSUE #3934: Metrics test made to always pass
See the https://github.com/ManageIQ/integration_tests/pull/3906#pullrequestreview-15303144

jaryn:
Fixes #3934

Test assumed the metrics to be running on the master which may not be the case. I also think that if we need to test metrics in cfme_tests, we should be using the DNS route. This can be discovered by asking the openshift using the kubeshift client in the test running time to avoid the need to configure it.

Patch is adding a yet another dependency -- on kubeshift. I think though that the kubeshift may be used in other tests to get the information about the provider to compare it with the CFME UI / API which could justify it's addition.

{{ pytest: cfme/tests/containers/test_basic_metrics.py --use-provider cm-env1 }}

### DIFF
--- a/cfme/tests/containers/test_basic_metrics.py
+++ b/cfme/tests/containers/test_basic_metrics.py
@@ -1,8 +1,9 @@
+import kubeshift
+import requests
 import pytest
-import re
+
 from cfme.containers.provider import ContainersProvider
 from utils import testgen
-from utils import conf
 from utils.version import current_version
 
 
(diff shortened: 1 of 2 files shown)


mshriver:
PRT results were a false-positive for run 10096.

I modified your description to add a pytest directive and specify a provider. Hopefully this will produce some meaningful PRT results.

cfme/tests/containers/test_basic_metrics.py
```diff
@@ -18,11 +19,23 @@ def test_basic_metrics(provider, ssh_client):
...
-    host_url = 'https://' + hostname + '/hawkular/metrics/'
-    command = 'curl -X GET ' + host_url + ' --insecure'
-    ssh_client = ssh_client(hostname=hostname, username=username, password=password)
-    assert re.search("Hawkular[ -]Metrics", str(ssh_client.run_command(command)))
+
+    os_api_url = "{0.rest_protocol}://{0.hostname}:{0.port}".format(
+        provider.provider_data)
+
+    # Prepare the creds for the kubeshift. Note we don't need to store
+    # need to store the credentials in any file, hence /dev/null is used.
```
mshriver: typo, duplicate 'need to store'

mshriver:
PRT uncovered an exception, was local testing successful?

        kubeshift_config.set_credentials(creds_name,
>                                        token=provider.provider_data["credentials"]["token"])
E       TypeError: string indices must be integers, not str

jaryn:
@mshriver I had no problem with this on our jenkins. We probably use other structure for the creds then I used.

Recently I found that the routes are probably available to get trough mgmt_system, so maybe I should replace the kubeshift with it.

mshriver:
If you can get the data through mgmt_system, that would be preferable as it would also remove the additional requirement kubeshift brings.

dajoRH:
*CFME QE Bot*

jkrocil:
Is this still under active development? Close this PR if not, please.

jkrocil:
Feel free to reopen once you spend some more time on this. Thanks!

PR #4098, opened 2017-01-30
[1LP][RFR]Patch for image comparison in cfme/tests/containers/test_containers_overv…

jdemon519:
{{pytest: cfme/tests/containers/test_containers_overview_data_integrity.py -v --use-provider cm-env1 }}

Fixed functionality to collect images from the Openshift, as well as the images from other registries if they don't already exist in the list of images collected from the Openshift.

### DIFF
--- a/cfme/tests/containers/test_containers_overview_data_integrity.py
+++ b/cfme/tests/containers/test_containers_overview_data_integrity.py
@@ -93,11 +93,13 @@ def test_containers_overview_data_integrity(provider):
     list_img_from_openshift_parsed = [i[7:].split(
         '@sha256:')[-1] for i in list_img_from_openshift_splitted]
 
-    list_img_from_openshift_parsed.extend([item for item in list_img_from_registry_splitted
-                                           if item not in list_img_from_openshift_parsed])
+    list_img_from_registry_splitted_new = set(list_img_from_registry_splitted)
+    list_img_from_openshift_parsed_new = set(list_img_from_openshift_parsed)
+
+    list_img_from_openshift_parsed_new.update(list_img_from_registry_splitted_new)
 
     num_img_cfme_56 = len(provider.mgmt.list_image())
-    num_img_cfme_57 = len(list_img_from_openshift_parsed)
+    num_img_cfme_57 = len(list_img_from_openshift_parsed_new)
 
     assert StatusBox('images').value() == version.pick({version.LOWEST:
                                                         num_img_cfme_56, '5.7': num_img_cfme_57})
--- a/cfme/tests/containers/test_containers_overview_data_integrity.py
+++ b/cfme/tests/containers/test_containers_overview_data_integrity.py
@@ -84,7 +84,7 @@ def test_containers_overview_data_integrity(provider):
         list_providers_by_class(ContainersProvider))
 
     list_img_from_registry = provider.mgmt.list_image()
-    list_img_from_registry_splitted = [i[2].split(
+    list_img_from_registry_splitted = [i.id.split(
         '@sha256:')[-1] for i in list_img_from_registry]
 
     list_img_from_openshift = provider.mgmt.list_image_openshift()
@@ -102,10 +102,7 @@ def test_containers_overview_data_integrity(provider):
 
     results = {}
     for cls in DATA_SETS:
-        results[
-            cls.object] = api_values[
-            cls.object] == statusbox_values[
-            cls.object]
+        results[cls.object] = api_values[cls.object] == statusbox_values[cls.object]
     if not all(results.values()):
         pytest.fail('There is a mismatch between API and UI values:\n{}'.format(
             '\n'.join(['{}: {} (API) != {} (UI)'.format(
--- a/cfme/tests/containers/test_containers_overview_data_integrity.py
+++ b/cfme/tests/containers/test_containers_overview_data_integrity.py
@@ -1,12 +1,11 @@
 import pytest
+
 from utils import testgen
 from utils.version import current_version
 from cfme.containers.provider import ContainersProvider
 from cfme.containers.node import Node
 from cfme.containers.pod import Pod
 from cfme.containers.service import Service
-from cfme.containers.image_registry import ImageRegistry
-from cfme.containers.image import Image
 from cfme.containers.project import Project
 from cfme.containers.route import Route
 from collections import namedtuple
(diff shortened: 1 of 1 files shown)


cfme/tests/containers/test_containers_overview_data_integrity.py
```diff
@@ -81,15 +75,39 @@ def test_containers_overview_data_integrity():
...
+
+    for s in list_img_from_openshift_parsed:
+        for item in list_img_from_registry_splitted:
+            if item not in list_img_from_openshift_parsed:
+                list_img_from_openshift_parsed.append(item)
+
+    assert len(list_img_from_openshift_parsed) == StatusBox('images').value()
+
     results = {}
     for cls in DATA_SETS:
```
jaryn: I am not entirely sure what are you trying to do here as the code is quite cumbersome, but my guess is you wanna use sets:

```
In [21]: abcd = set("abcd")

In [22]: bcde = set("bcde")

In [23]: abcd.symmetric_difference(bcde)
Out[23]: {'a', 'e'}

In [24]: abcd.difference(bcde)
Out[24]: {'a'}

```

So you can just do simply something similar to:
```
conflicts = abcd.symmetric_difference(bcde)
if conflicts:
  pytest.fail("These API and UI imags SHA's are disjoint {}.format(conflicts)")
```

cfme/tests/containers/test_containers_overview_data_integrity.py
```diff
@@ -81,15 +75,39 @@ def test_containers_overview_data_integrity():
...
+        list_providers_by_class(ContainersProvider))
+
+    list_img_from_registry = provider.mgmt.list_image()
+    list_img_from_registry_splitted = [i[2].split(
+        '@sha256:')[-1] for i in list_img_from_registry]
+
+    list_img_from_openshift = provider.mgmt.list_image_openshift()
+    list_img_from_openshift_splitted = [d['name']
+                                        for d in list_img_from_openshift]
+    list_img_from_openshift_parsed = [i[7:].split(
```
jaryn: Hmm, let me check wouldn't `re` make this more readable.

cfme/tests/containers/test_containers_overview_data_integrity.py
```diff
@@ -81,15 +75,39 @@ def test_containers_overview_data_integrity():
...
+        for item in list_img_from_registry_splitted:
+            if item not in list_img_from_openshift_parsed:
+                list_img_from_openshift_parsed.append(item)
+
+    assert len(list_img_from_openshift_parsed) == StatusBox('images').value()
+
     results = {}
     for cls in DATA_SETS:
-        results[cls.object] = api_values[cls.object] == statusbox_values[cls.object]
+        results[
```
jaryn: This formatting is crazy.

cfme/tests/containers/test_containers_overview_data_integrity.py
jaryn: Well it seems `re` wouldn't help much, but I would suggest that instead of using `i[2]` to use `i.id`.

jdemon519: @jaryn Good call, will fix that

cfme/tests/containers/test_containers_overview_data_integrity.py
jdemon519: @jaryn, agree), just fixed that

cfme/tests/containers/test_containers_overview_data_integrity.py
jdemon519: @jaryn When it comes to images, I am comparing two lists. If the item from the fist list already exists in the second list, then we do nothing, if it does not exist - we add the item to the second list.

jaryn: Yes and that's exactly a result of `set.difference.` I suggested `symteric_difference` because it is more strict and quite probably the thing that you want to use here.

cfme/tests/containers/test_containers_overview_data_integrity.py
```diff
@@ -81,15 +75,42 @@ def test_containers_overview_data_integrity():
...
+    list_img_from_openshift = provider.mgmt.list_image_openshift()
+    list_img_from_openshift_splitted = [d['name']
+                                        for d in list_img_from_openshift]
+    list_img_from_openshift_parsed = [i[7:].split(
+        '@sha256:')[-1] for i in list_img_from_openshift_splitted]
+
+    for s in list_img_from_openshift_parsed:
+        for item in list_img_from_registry_splitted:
+            if item not in list_img_from_openshift_parsed:
+                list_img_from_openshift_parsed.append(item)
```
gshefer: Why are you iterating over list_img_from_openshift_parsed if you don't use 's'? it causes to that you're appending every 'item' to list_img_from_openshift_parsed many times instead of once.
also, you have already defined list_img_from_openshift_parsed in line 93 so why are you do it again (with "not splitted" items - do we need them splitted or not?)?
anyway, I'd suggest to extend it with list comprehension:
`list_img_from_openshift_parsed.extend([item for item in list_img_from_registry_splitted if item not in list_img_from_openshift_parsed])`

cfme/tests/containers/test_containers_overview_data_integrity.py
```diff
@@ -81,15 +75,42 @@ def test_containers_overview_data_integrity():
...
     for cls in DATA_SETS:
         results[cls.object] = api_values[cls.object] == statusbox_values[cls.object]
     if not all(results.values()):
         pytest.fail('There is a mismatch between API and UI values:\n{}'.format(
-                    '\n'.join(['{}: {} (API) != {} (UI)'.format(
-                        obj.__name__, api_values[obj], statusbox_values[obj])
-                        for obj, is_pass in results.items() if not is_pass])))
+            '\n'.join(['{}: {} (API) != {} (UI)'.format(
+                obj.__name__, api_values[obj], statusbox_values[obj])
+                for obj, is_pass in results.items() if not is_pass])))
```
gshefer: I would keep the indentation as before, but this is not critical...

jdemon519: @gshefer Indeed, changed the nested loop to comprehension, will do the same in the other test

mshriver:
Tagged WIPTEST since it looked like you were still planning to address other comments. Set [1LP][RFR] when you're ready for merging

psav:
Shoud we merge this or are the other changes requested going to land?

jdemon519:
@psav Unless you guys think I should change something in the code, it's ready to be merged.

PR #4100, opened 2017-01-30
[1LP][RFR] Update cfme.automate, move self.obj to self.context in views

mshriver:
Found cfme.automate was using `self.obj` in views, which DNE.

Changed to self.context['object'].


mfalesni:
@mshriver so I shall do Yet Another Rebase™ when you merge this, right? :smile:

mshriver:
Unless you want to include it in your PR and merge today :D

PR #4101, opened 2017-01-30
[1LP][RFR] remove redundant TestPowerControlRESTAPI tests

mkoura:
Removing tests for power control using REST API that are superseded by tests in `cfme/tests/cloud_infra_common/test_power_control_rest.py` that do the same for both "from_detail" and "from_collection".

**PRT:** nothing to test since this PR is only removing tests.


psav:
Please also send an email to request these are removed from the test management system

PR #4102, opened 2017-01-31
[RFR] Update utils trackerbot and version for cfme 58

mshriver:
Support 5.8 in trackerbot and version utils.


utils/version.py
```diff
@@ -345,6 +345,7 @@ def product_version(self):
     '5.5': SPTuple('downstream-55z', '4.0'),
     '5.6': SPTuple('downstream-56z', '4.1'),
     '5.7': SPTuple('downstream-57z', '4.2'),
+    '5.8': SPTuple('downstream-58z', '4.5'),
```
jkrocil: Well well well, this is interesting! Can you confirm that we are indeed skipping `4.3` and `4.4`?

mshriver: I can - contacting on IRC with source.

PR #4109, opened 2017-01-31
[1LP] [RFR] Changed strings for 5.7.1

bargulg:
Some strings changed in version 5.7.1 and tests are failing, this should fix (at least some of) those issues

### DIFF
--- a/cfme/infrastructure/provider/__init__.py
+++ b/cfme/infrastructure/provider/__init__.py
@@ -165,8 +165,7 @@ class InfraProvider(Pretty, CloudInfraProvider):
     @cached_property
     def vm_name(self):
         return version.pick({
-            version.LOWEST: "VMs",
-            '5.5': "VMs and Instances",
+            version.LOWEST: "VMs and Instances",
             '5.7.1': "Virtual Machines"})
 
     @variable(alias='db')
--- a/cfme/infrastructure/provider/__init__.py
+++ b/cfme/infrastructure/provider/__init__.py
@@ -167,7 +167,7 @@ class InfraProvider(Pretty, CloudInfraProvider):
         return version.pick({
             version.LOWEST: "VMs",
             '5.5': "VMs and Instances",
-            '5.7': "Virtual Machines"})
+            '5.7.1': "Virtual Machines"})
 
     @variable(alias='db')
     def num_datastore(self):
--- a/cfme/infrastructure/provider/rhevm.py
+++ b/cfme/infrastructure/provider/rhevm.py
@@ -1,3 +1,4 @@
+from utils import version
 from . import InfraProvider, prop_region
 from mgmtsystem.rhevm import RHEVMSystem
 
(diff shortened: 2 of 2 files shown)


cfme/infrastructure/provider/__init__.py
```diff
@@ -167,7 +167,7 @@ def vm_name(self):
         return version.pick({
             version.LOWEST: "VMs",
             '5.5': "VMs and Instances",
-            '5.7': "Virtual Machines"})
+            '5.7.1': "Virtual Machines"})
```
mshriver: What am I missing here, these strings look the exact same?

I think you can go further here though, removing the `version.LOWEST` and `5.5` keys from the dictionary, just leaving 5.7 (which should be selected for 5.7.1 as well)

cfme/infrastructure/provider/rhevm.py
```diff
@@ -21,8 +22,14 @@ def __init__(self, name=None, credentials=None, zone=None, key=None, hostname=No
         self.end_ip = end_ip
 
     def _form_mapping(self, create=None, **kwargs):
+        provider_name = version.pick(
+            {
```
mshriver: Condense this syntax a bit, like the version.pick call in provider.__init__:

```python
provider_name = version.pick({
    <blah>: <blah>
    <blah: <blah>})

File: cfme/infrastructure/provider/__init__.py (hunk 1 above)
Comment (bargulg) [inline cfme/infrastructure/provider/__init__.py]: I changed the version for new string to 5.7.1, because 5.7.0 still has the old string.
Why would we remove support for 5.5? We might still need to test 5.5, and I don't see any reason to intentionally remove the support.

Comment (mshriver) [inline cfme/infrastructure/provider/__init__.py]: Noted, your 5.7 vs 5.7.1 change makes sense now.

We've been removing any version picking prior to 5.6 for some time now.  If we need to test against 5.5 we can do so on an older version of the repository.  In which case this would move 'VMs and Instances' to version.LOWEST which would cover 5.5, 5.6, and 5.7.0

Comment (bargulg) [inline cfme/infrastructure/provider/__init__.py]: Well, if you started doing that everywhere else, I will do the same here. Thanks for info

**Comment by mshriver** [comment]
Thanks for updates, love quick turn-around :)

waiting on PRT results, looks good.

File: cfme/infrastructure/provider/__init__.py [hunk 3]
```diff
@@ -165,9 +165,8 @@ def _form_mapping(self, create=None, **kwargs):
     @cached_property
     def vm_name(self):
         return version.pick({
-            version.LOWEST: "VMs",
-            '5.5': "VMs and Instances",
-            '5.7': "Virtual Machines"})
+            version.LOWEST: "VMs and Instances",
+            '5.7.1': "Virtual Machines"})
```
psav: This should be ```'5.7'``` not ```5.7.1``` else the auto will fail for 5.7.0, though it is unlikely we should ever use it against it, we shouldn't break backwards compatibility.

bargulg: As I already answered to comment from mshriver, in 5.7.0 it's actually still "VMs and Instances" and someone changed it to "Virtual Machines" for all 5.7 versions, which caused automation failed on latest 5.7.0 version. The string was changed to "Virtual Machines" only in version 5.7.1. not 5.7.0.

PR #4113, opened 2017-01-31
[RFR] rhevm export cleanup

mshriver:
Taken from Amogh's PR #3863

# Testing
I have tested the script, executing locally. It successfully cleaned up old templates on rhevm36, and found no templates to delete on rhvqe40

### DIFF
--- a/scripts/cleanup_edomain_templates.py
+++ b/scripts/cleanup_edomain_templates.py
@@ -12,14 +12,24 @@ import pytz
 from threading import Lock, Thread
 
 from utils import net
-from utils.conf import cfme_data
-from utils.conf import credentials
-from utils.ssh import SSHClient
+from utils.conf import cfme_data, credentials
 from utils.log import logger
 from utils.providers import get_mgmt
+from utils.ssh import SSHClient
 from utils.wait import wait_for
 
 lock = Lock()
+
+
+def make_ssh_client(provider_mgmt):
+    creds = credentials[provider_mgmt.kwargs.get('ssh_creds', None)]
+
+    connect_kwargs = {
+        'username': creds['username'],
+        'password': creds['password'],
+        'hostname': provider_mgmt.kwargs.get('ipaddress')
+    }
+    return SSHClient(**connect_kwargs)
 
 
 def parse_cmd_line():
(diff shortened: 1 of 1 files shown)


scripts/cleanup_edomain_templates.py
```diff
@@ -73,13 +75,16 @@ def change_edomain_state(api, state, edomain):
                 elif state == 'active' and export_domain.get_status().state != 'active':
                     dc.storagedomains.get(edomain).activate()
 
-                wait_for(is_edomain_template_deleted,
-                         [api, state, edomain], fail_condition=False, delay=5)
-                print('{} successfully set to {} state'.format(edomain, state))
+                wait_for(is_edomain_in_state,
+                        [api, state, edomain], fail_condition=False, delay=5)
```
psav: Curious as to why we need to pass in api here. Can wrapanapi not be extended to include this function?

scripts/cleanup_edomain_templates.py
```diff
@@ -100,25 +105,36 @@ def get_edomain_path(api, edomain):
...
+        provider_obj: provider object under execution
     """
     try:
-        print("RHEVM: Deleting the empty directories on edomain/vms file...")
-        ssh_client = make_ssh_client(edomainip, sshname, sshpass)
-        command = 'cd {}/master/vms && find . -maxdepth 1 -type d -empty -delete'.format(path)
+        provider_ip = provider_obj.kwargs.get('ipaddress', None)
+        provider_name = provider_obj.kwargs.get('name', None)
+        ssh_client = make_ssh_client(provider_ip, sshname, sshpass)
+        edomain_path = edomainip + ':' + path
```
psav: I'd prefer ```"{}:{}".format(...```

scripts/cleanup_edomain_templates.py
```diff
@@ -100,25 +105,36 @@ def get_edomain_path(api, edomain):
...
         sshpass: edomain ssh credentials.
+        provider_obj: provider object under execution
     """
     try:
-        print("RHEVM: Deleting the empty directories on edomain/vms file...")
-        ssh_client = make_ssh_client(edomainip, sshname, sshpass)
-        command = 'cd {}/master/vms && find . -maxdepth 1 -type d -empty -delete'.format(path)
+        provider_ip = provider_obj.kwargs.get('ipaddress', None)
+        provider_name = provider_obj.kwargs.get('name', None)
+        ssh_client = make_ssh_client(provider_ip, sshname, sshpass)
```
psav: Shouldn't we already have access to ssh_client for a provider under ```provider_obj.ssh_client``` or is this object just a dict? if it is just the dict of data, we should rename it to provider_dict.

scripts/cleanup_edomain_templates.py
```diff
@@ -278,22 +294,22 @@ def run(**kwargs):
             if args.provider != provider:
                 continue
 
-        mgmt_sys = cfme_data['management_systems'][provider]
-        ssh_rhevm_creds = mgmt_sys['hosts'][0]['credentials']
+        provider_obj = get_mgmt(provider)
```
psav: Same here, this is misleading. We should call it ```provider_mgmt```

mshriver:
@psav I'm looking into your comments, ripped this code from Amogh's branch.

scripts/cleanup_edomain_templates.py
mshriver: Its an object, but there is no ssh_client attribute. ssh_client is defined for utils.appliance and utils.composite but not in the provider classes.

scripts/cleanup_edomain_templates.py
mshriver: agreed

mshriver:
@dajohnso 

Script is working from local branch, would like to get merged so Jenkins picks it up and starts maintaining rhevm edomain again.

PR #4119, opened 2017-02-01
[1LP] [RFR] fix test_delete_orchestration_templates test

mkoura:
Adding BZ check to workaround the issue reported in https://bugzilla.redhat.com/show_bug.cgi?id=1414881 and leaving the test useful.

{{pytest: cfme/tests/services/test_rest_services.py -v -k TestOrchestrationTemplatesRESTAPI --long-running}}

### DIFF
--- a/cfme/tests/services/test_rest_services.py
+++ b/cfme/tests/services/test_rest_services.py
@@ -67,11 +67,6 @@ def service_data(request, rest_api, a_provider, dialog, service_catalogs):
     return _service_data(request, rest_api, a_provider, dialog, service_catalogs)
 
 
-@pytest.fixture(scope='function')
-def orchestration_templates(request, rest_api):
-    return _orchestration_templates(request, rest_api, num=2)
-
-
 class TestServiceRESTAPI(object):
     def test_edit_service(self, rest_api, services):
         """Tests editing a service.
@@ -457,6 +452,12 @@ class TestBlueprintsRESTAPI(object):
 
 
 class TestOrchestrationTemplatesRESTAPI(object):
+    @pytest.fixture(scope='function')
+    def orchestration_templates(self, request, rest_api):
+        response = _orchestration_templates(request, rest_api, num=2)
+        assert len(response) == 2
+        return response
+
     @pytest.mark.tier(3)
     @pytest.mark.uncollectif(lambda: version.current_version() < '5.7')
     def test_create_orchestration_templates(self, rest_api, orchestration_templates):
@@ -465,7 +466,6 @@ class TestOrchestrationTemplatesRESTAPI(object):
         Metadata:
             test_flag: rest
         """
-        assert len(orchestration_templates) == 2
         for template in orchestration_templates:
             record = rest_api.collections.orchestration_templates.get(id=template.id)
             assert record.name == template.name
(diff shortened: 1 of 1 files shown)


cfme/tests/services/test_rest_services.py
```diff
@@ -486,6 +487,8 @@ def test_delete_orchestration_templates(self, rest_api, orchestration_templates,
         collection = rest_api.collections.orchestration_templates
         if from_detail:
             methods = ['post', 'delete']
+            if BZ('1414881', forced_streams=['5.7', 'upstream']).blocks:
+                methods[0] = 'delete'  # because the 'post' method fails
```
mfalesni: `methods = ['delete']` ? :D (why run it twice)

mkoura: it's workaround for BZ 1414881. 'post' is not working, so until it's fixed, only 'delete' can be used for "from_detail"

mkoura: Basically I want to use both "post" and "delete" methods. Now I can use just "delete" until the bug is fixed, but I want to leave the selection there.

mshriver: Personally I'd rather see post removed from methods and a TODO to add it back when the bug is resolved.

mkoura: I can do that however reviewers wish, but if I add TODO, there's big chance it will get noticed long time after the bug is resolved. The way it is now the test will start working correctly right after the bug is resolved.
Also I wanted to leave the original code intact so that when it's time to clean up old workarounds and BZs, it's easy just to delete two lines without any other changes in the test necessary.

cfme/tests/services/test_rest_services.py
```diff
@@ -473,27 +474,33 @@ def test_create_orchestration_templates(self, rest_api, orchestration_templates)
...
+    @pytest.mark.parametrize('method', ['post', 'delete'])
+    def test_delete_orchestration_templates_from_detail(self, orchestration_templates, method):
+        """Tests delete orchestration templates from detail.
+
+        Metadata:
+            test_flag: rest
+        """
+        if method == 'post' and BZ('1414881', forced_streams=['5.7', 'upstream']).blocks:
+            pytest.skip("Affected by BZ1414881, cannot test.")
+        for ent in orchestration_templates:
```
mshriver: **OPTIONAL**
Consider a more descriptive variable here other than `ent`, `template` maybe?

cfme/tests/services/test_rest_services.py
```diff
@@ -473,27 +474,33 @@ def test_create_orchestration_templates(self, rest_api, orchestration_templates)
...
+    @pytest.mark.tier(3)
+    @pytest.mark.uncollectif(lambda: version.current_version() < '5.7')
+    @pytest.mark.parametrize('method', ['post', 'delete'])
+    def test_delete_orchestration_templates_from_detail(self, orchestration_templates, method):
+        """Tests delete orchestration templates from detail.
+
+        Metadata:
+            test_flag: rest
+        """
+        if method == 'post' and BZ('1414881', forced_streams=['5.7', 'upstream']).blocks:
```
mshriver: This is a lot cleaner than before, :+1:

PR #4121, opened 2017-02-01
SSH .status: Ignore warning line for workers

mfalesni:
{{pytest: cfme/tests/cli/test_evmserverd.py::test_evmserverd_stop -v --long-running}}


psav:
Doesn't look like this is enough - failed on 5.6 and 5.7 looks like there are some more messages that can sometimes come up - so perhaps we need that more robust solution.

PR #4123, opened 2017-02-02
[1LP] [RFR] Fixed Navigation, Replaced DynamoDB, Fixed bad forms

jteehan:
__Fixing__ 

Navigation got screwed up when the widgetastic stuff got added.
Forms were not being provided correct data from yamls

__Extending__ 

Replaced Template JSON from DynamoDb (which is expensive) with plain old fashion test-asdf1243 which also allows the cleanup script to function properly against these VMs and not leave databases all over the place.

__NOTE__

This PR is not intended to fix everything.  It's simply to get the module back in service so that other broken items can be fixed in a timely manner.  The current version of this module in integration_tests CANNOT possibly work in any manner due to the above referenced fixes.

{{pytest: cfme/tests/services/test_provision_stack.py::test_provision_stack -v --use-provider ec2 }}

### DIFF
--- a/cfme/tests/services/test_provision_stack.py
+++ b/cfme/tests/services/test_provision_stack.py
@@ -26,72 +26,126 @@ pytestmark = [
 AWS_TEMPLATE = """
 {
   "AWSTemplateFormatVersion" : "2010-09-09",
-  "Description" : "AWS CloudFormation Sample Template DynamoDB_Table",
+  "Description" : "AWS CloudFormation Sample Template EC2InstanceWithSecurityGroupSample:",
   "Parameters" : {
-    "HaskKeyElementName" : {
-      "Description" : "HashType PrimaryKey Name",
+    "KeyName": {
+      "Description" : "Name of an existing EC2 KeyPair to enable SSH access to the instance",
+      "Type": "AWS::EC2::KeyPair::KeyName",
+      "ConstraintDescription" : "must be the name of an existing EC2 KeyPair."
+    },
+    "virtualMachineName" : {
+      "Description" : "Name of the Virtual Machine",
       "Type" : "String",
-      "AllowedPattern" : "[a-zA-Z0-9]*",
-      "Default" : "SSS",
-      "MinLength": "1",
-      "MaxLength": "2048",
-      "ConstraintDescription" : "must contain only alphanumberic characters"
+      "ConstraintDescription" : "must be a valid EC2 instance name."
     },
-
-    "HaskKeyElementType" : {
-      "Description" : "HashType PrimaryKey Type",
+    "InstanceType" : {
+      "Description" : "WebServer EC2 instance type",
       "Type" : "String",
-      "Default" : "S",
-      "AllowedPattern" : "[S|N]",
-      "MinLength": "1",
-      "MaxLength": "1",
-      "ConstraintDescription" : "must be either S or N"
+      "Default" : "m1.small",
+      "AllowedValues" : ["m1.small", "t1.micro", "t2.nano", "t2.micro", "t2.small", "t2.large"],
+      "ConstraintDescription" : "must be a valid EC2 instance type."
     },
-
-    "ReadCapacityUnits" : {
-      "Description" : "Provisioned read throughput",
-      "Type" : "Number",
-      "Default" : "5",
-      "MinValue": "5",
-      "MaxValue": "10000",
-      "ConstraintDescription" : "must be between 5 and 10000"
+    "SSHLocation" : {
+      "Description" : "The IP address range that can be used to SSH to the EC2 instances",
+      "Type": "String",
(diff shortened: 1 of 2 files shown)


cfme/services/catalogs/orchestration_template.py
```diff
@@ -237,7 +239,7 @@ class AddDialog(CFMENavigateStep):
     VIEW = AddDialogView
 
     def step(self):
-        self.view.configuration.item_select('Create Service Dialog from Orchestration Template')
+        cfg_btn("Create Service Dialog from Orchestration Template")
```
mshriver: @jteehan This is a regression, if the previous call wasn't working we need to debug why, not remove it.

How was the navigation failing before? Can I help debug?

jteehan: It worked just fine before we made it pretty.  Navmazing and widgetastic were the regressions in this particular case as no one ever bothered to test the changes.  All I did was use the exact same structure as is in catalog_item.py which is working with the new navigation and already approved.  The way it was written above, each View class would have to initialize the same self.view.configuration. to work, which it never did, and really doesn't make sense to do it that way.  There is precedent for the way I did it, but I'm open to suggestions.

mfalesni: @jteehan What exactly was the error it was ending with? I just checked if it is a Widgetastic problem, but the dropdown works fine with this string both on 5.6 and 5.7 in that location. Maybe the navigation did something wrong and then the configuration button did not match?

With regards of initializing the self.view.configuration ... well that is the point of the whole design, that the things just get lazily regenerated per view to avoid any possible validity problems :)

jteehan: This is the error when using the currently implemented methods.

    def move_to_element(self, locator, *args, **kwargs):
        """Moves the mouse cursor to the middle of the element represented by the locator.
    
            Can handle moving to the ``<option>`` tags or Firefox being pissy and thus making it utilize
            a JS workaround, ...
    
            Args: See :py:meth:`elements`
    
            Returns:
                :py:class:`selenium.webdriver.remote.webelement.WebElement`
            """
        self.logger.debug('move_to_element: %r', locator)
        el = self.element(locator, *args, **kwargs)
        if el.tag_name == "option":
            # Instead of option, let's move on its parent <select> if possible
            parent = self.element("..", parent=el)
            if parent.tag_name == "select":
                self.move_to_element(parent)
                return el
        move_to = ActionChains(self.selenium).move_to_element(el)
        try:
            move_to.perform()
        except MoveTargetOutOfBoundsException:
            # ff workaround
            self.execute_script("arguments[0].scrollIntoView();", el)
            try:
                move_to.perform()
            except MoveTargetOutOfBoundsException:  # This has become desperate now.
                raise MoveTargetOutOfBoundsException(
                    "Despite all the workarounds, scrolling to `{}` was unsuccessful.".format(
>                       locator))
E               MoveTargetOutOfBoundsException: Message: Despite all the workarounds, scrolling to `<selenium.webdriver.remote.webelement.WebElement (session="1dc853fb-1046-401f-87c2-788c93661a6a", element="47")>` was unsuccessful.

../lib/python2.7/site-packages/widgetastic/browser.py:373: MoveTargetOutOfBoundsException

jteehan: Even if I add this:

class AddTemplateView(TemplateForm):

    add_button = Button("Add")
    configuration = Dropdown("Configuration")

dajoRH:
Lint report for commit 9af85669e7d5bc0e9dbeb884b009ea45979a2f30:

- :red_circle: Line 3:1: **F401** *'widgetastic_patternfly.Dropdown' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

jteehan:
I nuked all the pyc files and it started working, so I have no idea.  Anyway, all of PRT is passing with this so the fix was to revert the orchestration file.  I can see real VMs being created on Amazon rather than those expensive DynamoDB instances.

cfme/tests/services/test_provision_stack.py
```diff
@@ -26,73 +26,127 @@
 AWS_TEMPLATE = """
```
psav: **OPTIONAL:** This is getting big enough that it should probably be an external file now, ie not part of the test file. We should make an issue to do this.

jteehan: I completely agree.  Especially since I need to start expanding the number of json templates for azure as well.  Added JIRA card RHCFQE-1750 and assigned to myself.

Do I just create a directory inside this directory?  I'll find you when I start on the task and we'll find a nice place to organize these.

psav: There is a data directory and a file loader to help with this, but in a second PR is fine

Issue #4114, opened 2017-02-01
REST test doesn't descriminate in a_provider


psav:
This call to setup_a_provider https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/services/test_rest_services.py#L41
Is not discriminating against providers which don't have the right fields and so it is getting a provider which doens't have the right data and then failing here https://github.com/ManageIQ/integration_tests/blob/master/cfme/rest/gen_data.py#L132 if the provider doesn't have the ```provisioning``` data for example. We need to possibly use the new_setup_a_provider which allows provider filters which I think will allow the specification of the required_fields.

PR #4129, opened 2017-02-02
[1LP] [RFR] Fix provider selection in test_rest_services.py

mkoura:
Fixing provider selection in `test_rest_services.py` to avoid fails like
```
    def service_data(request, rest_api, a_provider, dialog, service_catalogs):
        """
        The attempt to add the service entities via web
        """
        template, host, datastore, iso_file, vlan, catalog_item_type = map(
>           a_provider.data.get("provisioning").get,
            ('template', 'host', 'datastore', 'iso_file', 'vlan', 'catalog_item_type'))
E       AttributeError: 'NoneType' object has no attribute 'get'

cfme/rest/gen_data.py:137: AttributeError
```
See https://github.com/ManageIQ/integration_tests/issues/4114

I hope I understood how `testgen` is working and that it's not necessary to use it in this module as no tests are parameterized with provider.

### DIFF
--- a/cfme/rest/gen_data.py
+++ b/cfme/rest/gen_data.py
@@ -133,9 +133,9 @@ def service_data(request, rest_api, a_provider, dialog, service_catalogs):
     """
     The attempt to add the service entities via web
     """
-    template, host, datastore, iso_file, vlan, catalog_item_type = map(
-        a_provider.data.get("provisioning").get,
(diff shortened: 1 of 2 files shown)


mshriver:
5.7 failure in PRT is a single test case failure, the test was properly parametrized.

PR #4131, opened 2017-02-03
[RFR] Added explanation of Selenium Standalone Server to the Guides

jteehan:
Purpose or Intent
=================

I seem to look this same information up about once a quarter so probably others may want the information as well.  I've never actually added to the docs, so feel free to help me if it needs work or isn't necessary.


jteehan:
I think this latest commit organizes it better.  See what you think.

PR #4139, opened 2017-02-06
[WIP] Fix finalizers for provisioning tests

bargulg:
Later these functions should be moved somewhere else, but we need a fix, fast

### DIFF
--- a/cfme/tests/infrastructure/test_provisioning_dialog.py
+++ b/cfme/tests/infrastructure/test_provisioning_dialog.py
@@ -6,7 +6,6 @@ from datetime import datetime, timedelta
 import fauxfactory
 import pytest
 from cfme import test_requirements
-from cfme.common.provider import cleanup_vm
 from cfme.infrastructure.virtual_machines import Vm
 from cfme.infrastructure.provider import InfraProvider
 from cfme.infrastructure.provider.rhevm import RHEVMProvider
@@ -33,6 +32,13 @@ pytestmark = [
     ]),
     pytest.mark.tier(3)
 ]
+
+
+def remove_vm(provider, vm_name):
+    if provider.mgmt.does_vm_exist(vm_name):
+        if provider.mgmt.is_vm_running(vm_name):
+            provider.mgmt.stop_vm(vm_name)
+        provider.mgmt.delete_vm(vm_name)
 
 
 pytest_generate_tests = testgen.generate([InfraProvider], required_fields=[
@@ -83,7 +89,7 @@ def provisioner(request, setup_provider, provider, vm_name):
         fill(provisioning_form, provisioning_data, action=provisioning_form.submit_button)
         flash.assert_no_errors()
 
-        request.addfinalizer(lambda: cleanup_vm(vm_name, provider))
+        request.addfinalizer(lambda: remove_vm(vm_name, provider))
         if delayed is not None:
             total_seconds = (delayed - datetime.utcnow()).total_seconds()
             row_description = 'Provision from [{}] to [{}]'.format(template, vm_name)
--- a/cfme/tests/infrastructure/test_provisioning_dialog.py
+++ b/cfme/tests/infrastructure/test_provisioning_dialog.py
@@ -34,7 +34,7 @@ pytestmark = [
 ]
 
 
-def remove_vm(provider, vm_name):
+def remove_vm(vm_name, provider):
     if provider.mgmt.does_vm_exist(vm_name):
         if provider.mgmt.is_vm_running(vm_name):
             provider.mgmt.stop_vm(vm_name)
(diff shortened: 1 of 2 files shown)


dajoRH:
Lint report for commit 6d84fa8fea13ff4189b398dc46a8bedef8abfd75:

* 6d84fa8fea13ff4189b398dc46a8bedef8abfd75

*CFME QE Bot*

cfme/tests/infrastructure/test_provisioning_dialog.py
```diff
@@ -35,6 +34,13 @@
 ]
 
 
+def remove_vm(provider, vm_name):
+    if provider.mgmt.does_vm_exist(vm_name):
+        if provider.mgmt.is_vm_running(vm_name):
+            provider.mgmt.stop_vm(vm_name)
+        provider.mgmt.delete_vm(vm_name)
```
mfalesni: I think that instead of creating two new instances of the same code you might like to put your changes into the cleanup_vm function. It was not deleting your VMs in your env?

dajoRH:
Lint report for commit aed0fc2def87e2846269cf18d0a178efaa3305a4:

* 6d84fa8fea13ff4189b398dc46a8bedef8abfd75

*CFME QE Bot*

cfme/tests/infrastructure/test_provisioning_dialog.py
bargulg: No, it wasn't, because in 5.7.1 VMs on RHV provider are started after provisioning by default, which means it can't be deleted before it's stopped.
There are 2 reasons why I didn't put this into cleanup_vm:
1. cleanup_vm catches the exception and hides the information in logger.warning, where we can't see it, because the test where it failed to delete a VM didn't fail. This causes other tests to fail and it's hard to investigate, because the root problem is not seen in any of the tests which failed.
2. I don't know whether I can use does_vm_exist, is_vm_running, stop_vm in cleanup_vm, because it's used by other providers, and I'm not sure whether they have these methods defined

bargulg: Still, I'm aware we need to have this somewhere in one place, and I think the best place would be in class RHEVMSystem in mgmtsystem/rhevm.py in wrapanapi, because it wouldn't break anything for other providers, but this PR is the fastest way for me to make tests work correctly while I work on a better fix. For now we can just checkout directly from PR in jenkins, with no intent to get this PR merged in its current state

cfme/tests/infrastructure/test_provisioning_dialog.py
```diff
@@ -83,7 +89,7 @@ def _provisioner(template, provisioning_data, delayed=None):
         fill(provisioning_form, provisioning_data, action=provisioning_form.submit_button)
         flash.assert_no_errors()
 
-        request.addfinalizer(lambda: cleanup_vm(vm_name, provider))
+        request.addfinalizer(lambda: remove_vm(vm_name, provider))
```
lkhomenk: cleanup_vm already has this check in cfme/common/provider. so this one more check is a bit excessive

bargulg: cleanup_vm doesn't stop VM before removing it, and additionally it catches all exceptions and makes VM stay there silently when something goes wrong. We can't have that in our infrastructure.

lkhomenk: @bargulg  . cleanup_vm uses delete_vm which stops vm if it's not stopped 
https://github.com/ManageIQ/wrapanapi/blob/master/mgmtsystem/rhevm.py#L210 . it works just perfect on VC or SCVMM. have you met any problems else where?

bargulg: For some reason it failed to delete VM on RHV. When running with this patch, it didn't fail to delete it. I will send you link to jenkins run in email.

lkhomenk: @bargulg I have replied. with old cleanup_vm test run successfully on rhevm36

bargulg: It fails for us with our environment which is using CMFE 5.7.1 and RHV 4.0 (AFAIK, this env is not in your yaml, because we use local yaml)

lkhomenk: @bargulg I had a talk with istein. this is an issue in 5.7 and rhevm-s. it's not related to vm cleanup. cleanup runs successfully on 5.6.4.0. this PR is not needed I guess

bargulg: Well, it somehow works as a workaround for some of the tests (like test_provision from provisioning_rest), so can we just keep this open for now? As I already said in some comment, I have no intent to get this PR merged, it's just here as a temporary workaround that we check out in jenkins

dajoRH:
*CFME QE Bot*

jkrocil:
Closing; @slaykovsky feel free to open a new PR.

PR #4141, opened 2017-02-06
[1LP][RFR] Update cfme.infra.host, add provider arg

### LINKED ISSUE #4137: Host should have an infra provider object associated with
Currently there are no relationships between host and infra provider.
provider_crud shouldn't be necessary host should pull the provider from the self.provider.

mshriver:
Modify calls to instantiate to include a provider object if available

Intent is not to modify behavior at this point but to add provider as an argument for Host constructor and to pass it where available in the current code.

# PRT Results
Failures unrelated to this specific change, but in general a very low pass rate for the included tests.

FIXES #4137

### DIFF
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 """Module containing classes with common behaviour for both VMs and Instances of all types."""
-from contextlib import contextmanager
 from datetime import date
 from functools import partial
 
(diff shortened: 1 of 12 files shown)


mshriver:
Fixes issue #4137

psav:
LOVE IT!!

cfme/common/vm.py
```diff
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 """Module containing classes with common behaviour for both VMs and Instances of all types."""
-from contextlib import contextmanager
```
mshriver: Removed for Lint.

cfme/common/vm.py
```diff
@@ -124,6 +123,10 @@ def factory(cls, vm_name, provider, template_name=None, template=False):
         """
         # Ensure the classes are loaded:
         import cfme.cloud.instance  # NOQA
+        from cfme.cloud.instance.azure import AzureInstance  # NOQA
```
mshriver: Found that vm.factory calls weren't selecting the correct class for cloud instances. Since these weren't imported their classes aren't registered.

PR #4147, opened 2017-02-07
[RFR][1LP] add support for subcollections and coverage info to apishow.py

mkoura:
* add support for subcollections
* add possibility to add coverage info taken from cfme log file to each action (handy for review what's tested / automated)

Output example
```
...
=======
* users
  Collection actions:
    * edit OK
    * query OK
    * create OK
    * delete OK
  Collection entity actions:
    * edit OK
    * delete
        POST OK
        DELETE OK
...
```
with "OK" where the action is covered.

It's a standalone script, so PRT isn't necessary.

### DIFF
--- a/scripts/apishow.py
+++ b/scripts/apishow.py
@@ -40,6 +40,15 @@ def parse_coverage_line(line):
         # line not in expected format
         return
 
+    if method not in ('POST', 'DELETE'):
+        return
+
+    try:
+        action = re.search(r'\'action\': u?\'([a-z_]*)', line).group(1)
+    except AttributeError:
+        # line not in expected format
+        return
+
     searches = [
         # collection, e.g. /api/vms
         r'/api/([a-z_]*) ',
@@ -61,15 +70,10 @@ def parse_coverage_line(line):
         except (AttributeError, IndexError):
             pass
         if collection:
+            # found matching expression
             break
     else:
         return
-
-    if '[RESTAPI] POST http' in line:
-        try:
-            action = re.search(r'\'action\': u?\'([a-z_]*)', line).group(1)
-        except AttributeError:
-            pass
 
     return Coverage(
         method=method,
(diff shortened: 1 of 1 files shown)


mkoura:
@izapolsk @mshriver can you please review/approve?

mkoura:
@mshriver @izapolsk any questions/feedback on this one?

scripts/apishow.py
```diff
@@ -1,55 +1,235 @@
...
+        _print_resource('Collection', collection)
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter)
+    parser.add_argument(
+        '--address',
+        default=conf.env.get('base_url', None),
+        help="hostname or ip address of target appliance")
```
mshriver: **optional**
'(...) target appliance, default pulled from local environment conf'

scripts/apishow.py
```diff
@@ -1,55 +1,235 @@
...
+    parser.add_argument(
+        '--logfile',
+        metavar='FILE',
+        default=None,
```
mshriver: Why not pull `utils.path.log_path + 'cfme.log'` for this default?

scripts/apishow.py
```diff
@@ -1,55 +1,235 @@
...
+        help="path to cfme log file")
+    args = parser.parse_args()
+
+    # we are really not interested in any warnings and "warnings.simplefilter('ignore')"
+    # doesn't work when it's redefined later in the REST API client
+    warnings.showwarning = lambda *args, **kwargs: None
+
+    api = MiqApi(
+        '{}/api'.format(
+            args.address if not args.address.endswith('/') else re.sub(r'/+$', '', args.address)),
```
mshriver: **optional**
Why not just use args.address.rstrip('/') here? It will strip all trailing slashes or just return the address as-is without using re.sub() or a ternary statement

scripts/apishow.py
```diff
@@ -1,55 +1,235 @@
...
+        ('admin', 'smartvm'),
```
mshriver: **required**
These credentials should be pulled from creds not hardcoded.

scripts/apishow.py
```diff
@@ -1,55 +1,235 @@
...
+            for key, subcollection in sorted(res_dict.iteritems()):
+                if key in ('actions', 'actions_avail', 'entity', 'methods'):
+                    continue
+                _print_resource('Subcollection "{}"'.format(key), subcollection)
+
+        _print_resource('Collection', collection)
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
```
mshriver: **optional**
Put argparse calls in a method and just call it here.

scripts/apishow.py
mkoura: @mshriver good idea, I originally wanted the coverage to be optional, but since parsing the logfile has negligible overhead compared to getting all the collections/subcollections/entities info out of the REST API client, it can be on by default.

PR #4149, opened 2017-02-07
[1LP][WIP] Added the new Ansible role from upstream

mfalesni:


cfme/configure/configuration/__init__.py
```diff
@@ -77,7 +77,8 @@
         ("storage_metrics_coordinator", CFMECheckbox("server_roles_storage_metrics_coordinator")),
         ("storage_inventory", CFMECheckbox("server_roles_storage_inventory")),
         ("vmdb_storage_bridge", CFMECheckbox("server_roles_vmdb_storage_bridge")),
-
+        # NEW ANSIBLE ROLE
```
mshriver: personally don't like comments like 'new' - consider 'Ansible role added in euwe'?

psav:
https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/configure/test_server_roles.py#L20
Do we need it adjusted here and perhaps in the yamls?

PR #4156, opened 2017-02-08
[1LP][RFR]Ansible retire test

sshveta:
Ansible retire test automated

### DIFF
--- a/cfme/tests/services/test_config_provider_servicecatalogs.py
+++ b/cfme/tests/services/test_config_provider_servicecatalogs.py
@@ -6,11 +6,13 @@ from cfme.automate.service_dialogs import ServiceDialog
 from cfme.configure.settings import DefaultView
 from cfme.services import requests
 from cfme.services.catalogs.service_catalogs import ServiceCatalogs
+from cfme.services.myservice import MyService
 from cfme.services.catalogs.catalog import Catalog
 from cfme.services.catalogs.catalog_item import CatalogItem
 from utils import testgen, version
 from utils.log import logger
 from utils.wait import wait_for
+from utils.blockers import BZ
 
 pytestmark = [
     test_requirements.service,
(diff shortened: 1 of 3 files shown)


sshveta:
Getting "AttributeError: 'RHBugzilla' object has no attribute 'getbugsimple'" 
Not getting this error when run locally .
Any ideas , anybody ?

cfme/tests/services/test_config_provider_servicecatalogs.py
```diff
@@ -114,3 +121,28 @@ def test_order_catalog_item(catalog_item, request):
...
+@pytest.mark.meta(blockers=[BZ(1420179, forced_streams=["5.6", "5.7", "upstream"])])
+def test_retire_ansible_service(catalog_item, request):
+    """Tests order catalog item
+    Metadata:
+        test_flag: provision
+    """
+    catalog_item.create()
+    service_catalogs = ServiceCatalogs(catalog_item.name)
+    service_catalogs.order()
+    logger.info('Waiting for cfme provision request for service %s', catalog_item.name)
```
mshriver: **REQUIRED**
use `'{}'.format()` please

cfme/tests/services/test_config_provider_servicecatalogs.py
```diff
@@ -114,3 +121,28 @@ def test_order_catalog_item(catalog_item, request):
...
+    row_description = catalog_item.name
+    cells = {'Description': row_description}
```
mshriver: **OPTIONAL**
Since this is only used here just put `catalog_item.name` in the dictionary directly

cfme/tests/services/test_config_provider_servicecatalogs.py
```diff
@@ -47,22 +51,23 @@ def config_manager(config_manager_obj):
...
+        default_text_box=service_name
     )
-    service_dialog = ServiceDialog(label=dialog, description="my dialog",
+    service_dialog = ServiceDialog(label=dialog_name, description="my dialog",
                      submit=True, cancel=True,
                      tab_label="tab_" + fauxfactory.gen_alphanumeric(), tab_desc="my tab desc",
                      box_label="box_" + fauxfactory.gen_alphanumeric(), box_desc="my box desc")
     service_dialog.create(element_data)
-    return dialog
+    yield dialog_name, service_name
```
mshriver: Any reason not to clean up after yielding? wrap service_dialog.delete() in a try?

cfme/tests/services/test_config_provider_servicecatalogs.py
```diff
@@ -107,10 +117,35 @@ def test_order_catalog_item(catalog_item, request):
     catalog_item.create()
     service_catalogs = ServiceCatalogs(catalog_item.name)
     service_catalogs.order()
-    logger.info('Waiting for cfme provision request for service %s', catalog_item.name)
-    row_description = catalog_item.name
-    cells = {'Description': row_description}
+    logger.info('Waiting for cfme provision request for service {}'.format(catalog_item.name))
```
mshriver: I think you may have missed my update to my comment. Milan corrected me, for logging we do want to use %s, because logger takes the argument.

sshveta: Oh :(

cfme/tests/services/test_config_provider_servicecatalogs.py
```diff
@@ -108,9 +118,32 @@ def test_order_catalog_item(catalog_item, request):
...
+    service_catalogs.order()
+    logger.info('Waiting for cfme provision request for service %s', catalog_item.name)
+    cells = {'Description': catalog_item.name}
+    row, __ = wait_for(requests.wait_for_request, [cells, True],
+        fail_func=requests.reload, num_sec=1400, delay=20)
+    assert 'Provisioned Successfully' in row.last_message.text
+    myservice = MyService(catalog_item.name)
+    myservice.retire()
+    # this is to assert that service is deleted from VMDB after retirement
+    with error.expected(NoSuchElementException):
```
psav: **REQUIRED: ** I really was on the fence with putting this as an OPTIONAL but I really think we should solve this now. We shouldn't be importing anything from selenium in the test. If there is an error from deleting it we should either, make the delete return ```False```, this is something that we don't do across the board yet, OR catch the error inside the method and reraise it as another error that is inside our framework. I know it seems like a very picky request, but it is important to have the lines of delineation clear. Tests **shouldn't** ever need to talk to selenium, that's why we abstract things out. This means that one day if we were to replace selenium, we wouldn't have to go rooting through tests to fix it. Are we there everywhere yet? NO! but we really don't want to introduce more technical debt.

PR #4157, opened 2017-02-08
[RFR] Fix requirements

psav:
Fixing again for docs


mshriver:
PRT failures are `ImportError: Error importing plugin "fixtures.prov_filter": cannot import name HTTPClient`, which you haven't touched in this PR from the looks of it.

PR #4159, opened 2017-02-08
[1LP][RFR] Update cloud_infra_common.test_retirement, ProviderFilter

mshriver:
Update test_retirement.test_set_and_unset_retirement_date_tomorrow, move to test_unset_retirement_date.

I'm scoping this commit for the changes that I would like to make to the test generation and fixtures.

I will address test updates, additions, fixes, etc in a separate commit to keep them reasonable in size. That being said I'm testing against all providers here for a baseline.

# PRT
Various failures from timeouts during wait, azure missing a template, VM not found on provider.
Increased wait time for retirement in new commit, and covered an upstream change to the flash message. Otherwise I would prefer to leave additional test fixes to a 2nd PR to limit size/scope.

{{ pytest: cfme/tests/cloud_infra_common/test_retirement.py -v --long-running --use-provider complete }}

### DIFF
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -530,8 +530,10 @@ class VM(BaseVM):
         self.load_details(refresh=True)
         lcl_btn(self.TO_RETIRE, invokes_alert=True)
         sel.handle_alert()
-        flash.assert_success_message('Retirement initiated for 1 VM and Instance from the CFME '
-                                     'Database')
+        flash.assert_success_message(
+            'Retirement initiated for 1 VM and Instance from the {} Database'.format(version.pick({
+                version.LOWEST: 'CFME',
+                'upstream': 'ManageIQ'})))
 
     def power_control_from_provider(self):
         raise NotImplementedError("You have to implement power_control_from_provider!")
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -522,7 +522,7 @@ class VM(BaseVM):
 
     retire_form = Form(fields=[
         ('date_retire', AngularCalendarInput(
-            "retirement_date", "//label[contains(normalize-space(.), 'Retirement')]")),
+            "retirement_date", "//label[contains(normalize-space(.), 'Retirement Date')]")),
         ('warn', AngularSelect('retirementWarning'))
     ])
 
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 """Module containing classes with common behaviour for both VMs and Instances of all types."""
-from contextlib import contextmanager
 from datetime import date
 from functools import partial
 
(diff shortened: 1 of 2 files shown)


cfme/tests/cloud_infra_common/test_retirement.py
```diff
@@ -3,87 +3,92 @@
...
+    complete_filter = ProviderFilter(required_tags=['complete'])
+    no_container_filter = ProviderFilter(required_tags=['container'], inverted=True)
+    argnames, argvalues, idlist = testgen.providers(metafunc, filters=[complete_filter,
+                                                                       no_container_filter])
+    testgen.parametrize(metafunc, argnames, argvalues, ids=idlist, scope='module')
 
 
-@pytest.fixture(scope="function")
-def vm(request, provider, setup_provider, small_template):
+#@pytest.fixture(scope='module')
```
mshriver: testing, prototyping...ignore...

dajoRH:
Lint report for commit a7233628f4fc99283416c05c681b59bc1de7679c:

`cfme/tests/cloud_infra_common/test_retirement.py`:
- :red_circle: Line 10:1: **F401** *'utils.providers.new_setup_a_provider' imported but unused*
- :bangbang: Line 23:1: **E265** *block comment should start with '# '*
- :bangbang: Line 24:1: **E265** *block comment should start with '# '*
- :bangbang: Line 26:1: **E266** *too many leading '#' for block comment*

**No commit flaws detected**.

*CFME QE Bot*

cfme/common/vm.py
```diff
@@ -522,20 +521,19 @@ class VM(BaseVM):
     TO_RETIRE = None
 
     retire_form = Form(fields=[
-        ('date_retire', {
-            version.LOWEST: date_retire_element,
-            "5.5": AngularCalendarInput(
-                "retirement_date", "//label[contains(normalize-space(.), 'Retirement Date')]")}),
+        ('date_retire', AngularCalendarInput(
+            "retirement_date", "//label[contains(normalize-space(.), 'Retirement')]")),
```
otsuman: **REQUIRED:** Specify locator
`"//label[contains(normalize-space(.), 'Retirement')]"` has two possible values in page, something can be added in future with  'Retirement' in locator

mshriver: So, check out AngularCalendarInput, the second argument is a 'click away element'. It's clicked after the calendar element itself is interacted with on a fill, to ensure the calendar is closed.

I think in this case its fine that the locator matches two labels on the page, clicking either one will produce the desired result.

otsuman: But what if new element will be added before calendar widget with similar locator? It will start to interact with it  and can redirect you to another page/menu, for example

mshriver: I see what you mean, the label returned by that locator could be clickable? I could match on 'Retirement Date' but there's also no guarantee the UI won't be changed to make that specific label clickable?  I could set to some other text on the page, like the h3 summary?

otsuman: From my point of view, locator should be unique, because if something will be changed we can easily track this changes. 
So i'm okay with 'Retirement Date'  or h3 summary.

mshriver: Done, f93b39f

mshriver:
@dajohnso

This opens up provider coverage for retirement tests, to match what I'm currently adding in Polarion.

PR #4181, opened 2017-02-14
[1LP][RFR] Catalog Widgetastic Conversion

sshveta:
{{pytest: -v --long-running cfme/tests/services/test_catalog.py }}

### DIFF
--- a/cfme/services/catalogs/catalog.py
+++ b/cfme/services/catalogs/catalog.py
@@ -1,115 +1,120 @@
-# -*- coding: utf-8 -*-
-from functools import partial
-
+from widgetastic.widget import Text
+from widgetastic_patternfly import Button, Input
 from navmazing import NavigateToAttribute, NavigateToSibling
-
-from cfme import web_ui
-from cfme.exceptions import DestinationNotFound
-from cfme.fixtures import pytest_selenium as sel
-from cfme.web_ui import accordion, flash, form_buttons, Input
-from cfme.web_ui import toolbar as tb, paginator, CheckboxTable, match_location
-from utils.appliance import Navigatable
-from utils.appliance.implementations.ui import navigator, navigate_to, CFMENavigateStep
-from utils.blockers import BZ
-from utils.pretty import Pretty
 from utils.update import Updateable
+from utils.pretty import Pretty
+from utils.appliance import Navigatable
+from utils.appliance.implementations.ui import navigator, CFMENavigateStep, navigate_to
 from utils import version
 
 
-cfg_btn = partial(tb.select, "Configuration")
-catalog_tree = partial(accordion.tree, "Catalogs")
-
-listview_table = CheckboxTable(table_locator='//div[@id="list_grid"]/table')
-
-item_multiselect = web_ui.MultiSelect(
-    "//select[@id='available_fields']",
-    "//select[@id='selected_fields']",
-    "//div[@id='column_lists']//a[contains(@href, 'button=right')]/img",
-    "//div[@id='column_lists']//a[contains(@href, 'button=left')]/img")
-
-form = web_ui.Form(
-    fields=[('name_text', Input("name")),
-            ('description_text', Input("description")),
-            ('button_multiselect', item_multiselect),
-            ('add_button', form_buttons.add),
-            ('save_button', form_buttons.save),
-            ('cancel_button', form_buttons.cancel)])
-
-item_form = web_ui.Form(
-    fields=[('type_select', "//select[@id='st_prov_type']"),
-            ('name_text', Input("name")),
-            ('description_text', Input("description")),
-            ('display_checkbox', Input("display")),
-            ('add_button', form_buttons.add)])
-
-match_page = partial(match_location, controller='catalog', title='Catalogs')
+from . import ServicesCatalogView
 
 
-def _all_catalogs_add_new(_):
-    if BZ(1213863).blocks:
-        sel.pytest.skip("Blocker on the bug 1213863")
-    catalog_tree('All Catalogs')
-    cfg_btn('Add a New Catalog')
+class CatalogForm(ServicesCatalogView):
+    title = Text('#explorer_title_text')
+
+    name = Input(name='name')
+    description = Input(name="description")
+
+    save_button = Button('Save')
+    cancel_button = Button('Cancel')
+
+
+class CatalogsView(ServicesCatalogView):
+    title = Text("#explorer_title_text")
+
+    @property
+    def is_displayed(self):
+        return (
+            self.in_explorer and
+            self.title.text == 'All Catalogs' and
+            self.catalogs.is_opened and
+            self.catalogs.tree.currently_selected == ["All Catalogs"])
+
+
+class DetailsCatalogView(ServicesCatalogView):
+    title = Text("#explorer_title_text")
+
+    @property
+    def is_displayed(self):
+        return (
+            self.in_explorer and self.catalogs.is_opened and
+            self.title.text == 'Catalog "{}"'.format(self.context['object'].name)
+        )
+
+
+class AddCatalogView(CatalogForm):
+
+    add_button = Button("Add")
+
+    @property
+    def is_displayed(self):
+        return (
+            self.in_explorer and self.catalogs.is_opened and
+            self.title.text == "Adding a new Catalog"
+        )
+
+
+class EditCatalogView(CatalogForm):
+
+    save_button = Button('Save')
+    reset_button = Button('Reset')
+
+    @property
+    def is_displayed(self):
+        return (
+            self.in_explorer and self.catalogs.is_opened and
+            self.title.text == "Editing Catalog {}".format(self.name)
+        )
 
 
 class Catalog(Updateable, Pretty, Navigatable):
-    """Represents a Catalog"""
-    pretty_attrs = ['name', 'items']
 
     def __init__(self, name=None, description=None, items=None, appliance=None):
+        Navigatable.__init__(self, appliance=appliance)
         self.name = name
         self.description = description
         self.items = items
-        Navigatable.__init__(self, appliance=appliance)
-
-    def __str__(self):
-        return self.name
 
     def create(self):
-        navigate_to(self, 'Add')
-        # For element not found exception (To be removed)
-        sel.sleep(5)
-        web_ui.fill(form, {'name_text': self.name,
-                           'description_text': self.description,
-                           'button_multiselect': self.items},
-                    action=form.add_button)
-        flash_str = 'Catalog "{}" was saved'
-        flash.assert_success_message(flash_str.format(self.name))
+        view = navigate_to(self, 'Add')
+        view.fill({'name': self.name,
+                   'description': self.description})
+        view.add_button.click()
+        view.flash.assert_success_message('Catalog "{}" was saved'.format(self.name))
+        view = self.create_view(CatalogsView)
+        assert view.is_displayed
+        view.flash.assert_no_error()
 
     def update(self, updates):
-        navigate_to(self, 'Edit')
-        web_ui.fill(form, {'name_text': updates.get('name', None),
-                           'description_text': updates.get('description', None),
-                           'button_multiselect': updates.get('items', None)},
-                    action=form.save_button)
-        flash.assert_success_message('Catalog "{}" was saved'.format(self.name))
-
-    def delete(self, from_dest='All'):
-        """
-        Delete the catalog, starting from the destination provided by from_dest
-        Throws cfme.DestinationNotFound exception
-
-        :param from_dest: A valid navigation destination to start the delete from
-        :return: none
-        """
-        if from_dest in navigator.list_destinations(self):
-            navigate_to(self, from_dest)
+        view = navigate_to(self, 'Edit')
+        changed = view.fill(updates)
+        if changed:
+            view.save_button.click()
         else:
-            msg = 'cfme.services.catalogs.catalog does not have destination {}'.format(from_dest)
-            raise DestinationNotFound(msg)
+            view.cancel_button.click()
+        view = self.create_view(DetailsCatalogView, override=updates)
+        assert view.is_displayed
+        view.flash.assert_no_error()
+        if changed:
+            view.flash.assert_message(
+                'Catalog "{}" was saved'.format(updates.get('name', self.name)))
+        else:
+            view.flash.assert_message(
+                'Edit of Catalog "{}" was cancelled by the user'.format(self.name))
 
-        # Delete using the appropriate method
-        if from_dest == 'All':
-            # Select the row to delete, assuming default List View for All
-            listview_table.select_row_by_cells({'Name': self.name, 'Description': self.description})
-            cfg_btn(version.pick({version.LOWEST: 'Remove Items from the VMDB',
-                    '5.7': 'Remove Catalogs'}), invokes_alert=True)
-        elif from_dest == 'Details':
-            cfg_btn(version.pick({version.LOWEST: 'Remove Item from the VMDB',
-                    '5.7': 'Remove Catalog'}), invokes_alert=True)
-
-        sel.handle_alert()
-        flash.assert_success_message(
+    def delete(self):
+        view = navigate_to(self, "Details")
+        view.configuration.item_select(
+            version.pick({
+                version.LOWEST: 'Remove Item from the VMDB',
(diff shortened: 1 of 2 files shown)


cfme/services/catalogs/catalog.py
```diff
@@ -1,115 +1,104 @@
...
 
-def _all_catalogs_add_new(_):
-    if BZ(1213863).blocks:
-        sel.pytest.skip("Blocker on the bug 1213863")
-    catalog_tree('All Catalogs')
-    cfg_btn('Add a New Catalog')
+    @property
+    def is_displayed(self):
+        return (
+            super(CatalogsView, self).is_displayed and
```
mfalesni: **REQUIRED**: Do not use super() like this. ServicesCatalogView should check that it is in ServicesCatalog and it should also check that accordion is enabled and such things - that it actually is usable for the steps to happen since that ensures correct matching of locations. I strongly advocate for creating a property, let's say `in_servies_catalogs` and use that one instead in the subclasses.

Then the parent class would look like.

```python
@property
def in_services_catalogs(self):
    return 'logged in as current user and correct menu item selected'

@property
def is_displayed(self):
    return self.in_services_catalogs and not accordion.is_dimmed and 'whatever else to ensure that the view is correct'
```

Then in the child class, like this one, you will just replace this line with

```python
self.in_services_catalogs and  # ...
```

You might see the super() usage in some places, yes, somewhere it makes sense and in the other places I am going to remove them :)

cfme/services/catalogs/catalog.py
```diff
@@ -1,115 +1,104 @@
...
 
-item_form = web_ui.Form(
-    fields=[('type_select', "//select[@id='st_prov_type']"),
-            ('name_text', Input("name")),
-            ('description_text', Input("description")),
-            ('display_checkbox', Input("display")),
-            ('add_button', form_buttons.add)])
+    add_button = Button('Add')
+    save_button = Button('Save')
+    cancel_button = Button('Cancel')
```
mfalesni: **REQUIRED**: The CatalogForm child classes also specify the buttons. Currently it is customary to have them in the child classes only (for the different between add/save). So you should remove them here.

cfme/services/catalogs/catalog.py
```diff
@@ -1,115 +1,104 @@
...
+        view.fill({'name_text': updates.get('name', None),
+                   'description_text': updates.get('description', None),
+                   'button_multiselect': updates.get('items', None)})
+        view.save_button.click()
+        view.flash.assert_success_message('Catalog "{}" was saved'.format(self.name))
+
+    def delete(self):
+        view = navigate_to(self, "Details")
+        view.configuration.item_select(version.pick({version.LOWEST: 'Remove Item from the VMDB',
+                    '5.7': 'Remove Catalog'}), handle_alert=True)
```
mfalesni: **REQUIRED**: Indenting

```python
    version.pick({
        version.LOWEST: 'Remove Item from the VMDB',
        '5.7': 'Remove Catalog'}),
    handle_alert=True)
```

cfme/services/catalogs/catalog.py
```diff
@@ -126,56 +115,38 @@ def exists(self):
...
-        if paginator.page_controls_exist():
-            # Make sure nothing is selected
-            sel.check(paginator.check_all())
-            sel.uncheck(paginator.check_all())
+        self.view.catalogs.tree.click_path("All Catalogs")
 
 
-@navigator.register(Catalog, 'Add')
-class Add(CFMENavigateStep):
+@navigator.register(Catalog, 'AddCatalog')
```
mfalesni: Why the change? Convention says that we should use Add/Edit/Details and such names, so you never have to look it up. If you have multiple classes defined in one module and the Add steps' classes' names clash, you can rename the class but keep the step name as Add ...

cfme/services/catalogs/catalog.py
```diff
@@ -126,56 +115,38 @@ def exists(self):
...
-        return match_page(summary='Catalog "{}"'.format(self.obj.name))
+    VIEW = DetailsCatalogView
 
     def step(self):
-        # Best effort to make sure the catalog is in the list
-        if paginator.page_controls_exist():
-            paginator.results_per_page(1000)
-        listview_table.click_row_by_cells(
-            {'Name': self.obj.name, 'Description': self.obj.description})
+        self.view.catalogs.tree.click_path("All Catalogs", self.obj.name)
```
mfalesni: **REQUIRED**: Use `self.prerequisite_view` since to get to the location, you usually need the previous view and do something with it to get to the location this step represents. In this case it probably works because the views are inherited, but this leads to bad copypasting and it might also lead to pulled hair when something changes :)

cfme/services/catalogs/catalog.py
```diff
@@ -126,56 +115,38 @@ def exists(self):
...
+class EditCatalog(CFMENavigateStep):
     prerequisite = NavigateToSibling('Details')
 
-    def am_i_here(self):
-        match_page(summary='Editing Catalog "{}"'.format(self.obj.name))
+    VIEW = EditCatalogView
 
     def step(self):
-        cfg_btn('Edit this Item')
+        self.view.configuration.item_select('Edit this Item')
```
mfalesni: **REQUIRED**: Same with the prerequisite view as above.

sshveta:
Done , Done and Done , Please review again

cfme/services/catalogs/catalog.py
```diff
@@ -1,115 +1,106 @@
...
-        # For element not found exception (To be removed)
-        sel.sleep(5)
-        web_ui.fill(form, {'name_text': self.name,
-                           'description_text': self.description,
-                           'button_multiselect': self.items},
-                    action=form.add_button)
+        view = navigate_to(self, 'Add')
+        view.fill({'name_text': self.name,
+                   'description_text': self.description,
+                   'button_multiselect': self.items})
```
izapolsk: REQUIRED: I haven't found 'button_multiselect' among view or form controls. Are you sure you haven't missed it ?

cfme/services/catalogs/catalog.py
```diff
@@ -1,115 +1,106 @@
...
-        navigate_to(self, 'Add')
-        # For element not found exception (To be removed)
-        sel.sleep(5)
-        web_ui.fill(form, {'name_text': self.name,
-                           'description_text': self.description,
-                           'button_multiselect': self.items},
-                    action=form.add_button)
+        view = navigate_to(self, 'Add')
+        view.fill({'name_text': self.name,
+                   'description_text': self.description,
```
izapolsk: OPTIONAL: it make sense to remove text from control names.

cfme/services/catalogs/catalog.py
```diff
@@ -1,115 +1,106 @@
...
+                   'button_multiselect': self.items})
+        view.add_button.click()
         flash_str = 'Catalog "{}" was saved'
-        flash.assert_success_message(flash_str.format(self.name))
+        view.flash.assert_success_message(flash_str.format(self.name))
```
izapolsk: OPTIONAL: I'd prefer to see formatted string and format statement together. readability counts.

cfme/services/catalogs/catalog.py
```diff
@@ -1,115 +1,106 @@
...
-        elif from_dest == 'Details':
-            cfg_btn(version.pick({version.LOWEST: 'Remove Item from the VMDB',
-                    '5.7': 'Remove Catalog'}), invokes_alert=True)
-
-        sel.handle_alert()
-        flash.assert_success_message(
+        view = navigate_to(self, 'Edit')
+        view.fill({'name_text': updates.get('name', None),
+                   'description_text': updates.get('description', None),
+                   'button_multiselect': updates.get('items', None)})
```
izapolsk: REQUIRED: ^

cfme/services/catalogs/catalog.py
```diff
@@ -126,56 +117,38 @@ def exists(self):
 class All(CFMENavigateStep):
     prerequisite = NavigateToAttribute('appliance.server', 'LoggedIn')
 
-    def am_i_here(self):
-        return match_page(summary='All Catalogs')
+    VIEW = CatalogsView
```
izapolsk: REQUIRED: could you please put VIEW before prerequisite ?

cfme/services/catalogs/catalog.py
```diff
@@ -126,56 +117,38 @@ def exists(self):
...
+        self.view.catalogs.tree.click_path("All Catalogs")
 
 
 @navigator.register(Catalog, 'Add')
 class Add(CFMENavigateStep):
     prerequisite = NavigateToSibling('All')
 
-    def am_i_here(self):
-        return match_page(summary='Adding a new Catalog')
+    VIEW = AddCatalogView
```
izapolsk: ^

cfme/services/catalogs/catalog.py
```diff
@@ -126,56 +117,38 @@ def exists(self):
...
 
 
 @navigator.register(Catalog, 'Details')
 class Details(CFMENavigateStep):
     prerequisite = NavigateToSibling('All')
 
-    def am_i_here(self):
-        # header text as separate var to reduce quote escaping
-        return match_page(summary='Catalog "{}"'.format(self.obj.name))
+    VIEW = DetailsCatalogView
```
izapolsk: ^

cfme/services/catalogs/catalog.py
```diff
@@ -126,56 +117,38 @@ def exists(self):
...
+        self.prerequisite_view.catalogs.tree.click_path("All Catalogs", self.obj.name)
 
 
 @navigator.register(Catalog, 'Edit')
 class Edit(CFMENavigateStep):
     prerequisite = NavigateToSibling('Details')
 
-    def am_i_here(self):
-        match_page(summary='Editing Catalog "{}"'.format(self.obj.name))
+    VIEW = EditCatalogView
```
izapolsk: ^

cfme/services/catalogs/catalog.py
sshveta: removing multiselect , not required in making catalog

cfme/services/catalogs/catalog.py
sshveta: Done

cfme/services/catalogs/catalog.py
sshveta: Done

cfme/services/catalogs/catalog.py
sshveta: Got it

cfme/services/catalogs/catalog.py
```diff
@@ -1,115 +1,103 @@
...
-            cfg_btn(version.pick({version.LOWEST: 'Remove Item from the VMDB',
-                    '5.7': 'Remove Catalog'}), invokes_alert=True)
-
-        sel.handle_alert()
-        flash.assert_success_message(
+        view = navigate_to(self, 'Edit')
+        view.fill({'name': updates.get('name', None),
+                   'description': updates.get('description', None)})
+        view.save_button.click()
+        view.flash.assert_success_message('Catalog "{}" was saved'.format(self.name))
```
mfalesni: Check this out: https://github.com/ManageIQ/integration_tests/blob/master/cfme/automate/explorer/namespace.py#L217

This is how I rewrote the CRUD methods for use with widgetastic.

cfme/services/catalogs/catalog.py
```diff
@@ -1,115 +1,103 @@
...
-                    '5.7': 'Remove Catalogs'}), invokes_alert=True)
-        elif from_dest == 'Details':
-            cfg_btn(version.pick({version.LOWEST: 'Remove Item from the VMDB',
-                    '5.7': 'Remove Catalog'}), invokes_alert=True)
-
-        sel.handle_alert()
-        flash.assert_success_message(
+        view = navigate_to(self, 'Edit')
+        view.fill({'name': updates.get('name', None),
+                   'description': updates.get('description', None)})
```
mfalesni: No need to tinker with the updates dict, just throw it in fill :)

cfme/services/catalogs/catalog.py
```diff
@@ -1,115 +1,110 @@
...
+            self.catalogs.is_opened
+        )
 
 
 class Catalog(Updateable, Pretty, Navigatable):
-    """Represents a Catalog"""
-    pretty_attrs = ['name', 'items']
 
     def __init__(self, name=None, description=None, items=None, appliance=None):
+        Navigatable.__init__(self, appliance)
```
mfalesni: Can we keep the keyword reference here? `Navigatable.__init__(self, appliance=appliance)`

cfme/services/catalogs/catalog.py
```diff
@@ -1,115 +1,110 @@
...
-                           'description_text': self.description,
-                           'button_multiselect': self.items},
-                    action=form.add_button)
-        flash_str = 'Catalog "{}" was saved'
-        flash.assert_success_message(flash_str.format(self.name))
+        view = navigate_to(self, 'Add')
+        view.fill({'name': self.name,
+                   'description': self.description})
+        view.add_button.click()
+        view.flash.assert_success_message('Catalog "{}" was saved'.format(self.name))
```
mfalesni: So, it should actually create a view that it is supposed to have landed on and then assert is_displayed on it. Then use that view for checking flash. Something like:

```python
view.fill(...)
view.add_button.click()
view = self.create_view(ViewItShouldHaveLandedOn)  # In case of update(self, updates) you should also pass override=updates
assert view.is_displayed
view.flash.assert_no_error()
view.flash.assert_success_message('Catalog "{}" was saved'.format(self.name))
```

create_view passes self automatically in the view as the context object.

From the latest widgetastic version you can use fill_with which triggers the appropriate actions upon filling:
```python
if view.fill_with({...}, on_change=view.add_button, no_change=view.cancel_button):
    view = self.create_view(ViewItShouldHaveLandedOn)  # In case of update(self, updates) you should also pass override=updates
    assert view.is_displayed
    view.flash.assert_no_error()
    view.flash.assert_success_message('Catalog "{}" was saved'.format(self.name))
else:  # when cancel
    view = self.create_view(AnotherViewItShouldHaveLandedOn)  # In case of update(self, updates) DO NOT pass updates here
    assert view.is_displayed
    view.flash.assert_no_error()
    view.flash.assert_success_message('Add was cancelled')
```

@psav is this the way how we should go forward? More checking? I do it like this with automate.

Also the self.name, in some places the display names, descriptions and such things change the used string based on what is and what is not specified. I think that since this is a conversion it should not be the case but just reminding :)

psav: Yeh I think this is right :) Makes me think we could have a special fn that is used to assert actions as there is some DRY inside that, but I'm happy with what you have suggested, it looks like what I would expect. The only confusing thing is the reuse of the view variable.

mfalesni: Well, you know, only one view can be valid at the time so if you consider the `view` variable as something you are working on and supposed to be there ...

sshveta: Did this for Edit and delete but cannot do it for Add method as the view is different when I try to add a duplicate catalog in a test.

sshveta: I have done assert flash message before assert view in Add method so that while creating duplicate catalog item I can catch the error and view does not conflict . Please see if its okay.

cfme/services/catalogs/catalog.py
```diff
@@ -1,115 +1,121 @@
...
-    fields=[('type_select', "//select[@id='st_prov_type']"),
-            ('name_text', Input("name")),
-            ('description_text', Input("description")),
-            ('display_checkbox', Input("display")),
-            ('add_button', form_buttons.add)])
+    @property
+    def is_displayed(self):
+        return (
+            self.title.text == "Adding a new Catalog" and
+            self.catalogs.is_opened
```
mfalesni: so, it should also check in_explorer. And the is_opened should come before the title text. So the order then would be in_explorer, catalogs, title text.

cfme/services/catalogs/catalog.py
```diff
@@ -1,115 +1,121 @@
...
-    catalog_tree('All Catalogs')
-    cfg_btn('Add a New Catalog')
+    save_button = Button('Save')
+    reset_button = Button('Reset')
+
+    @property
+    def is_displayed(self):
+        return (
+            self.title.text == "Editing Catalog {}".format(self.name) and
+            self.catalogs.is_opened
```
mfalesni: Same as my previous comment.

PR #4185, opened 2017-02-14
[1LP][RFR] check response status code in REST API tests for cloud providers

mkoura:
* response status code checking
* refactoring, enhancements

{{pytest: cfme/tests/cloud/test_providers.py -v -k TestProvidersRESTAPI}}

### DIFF
--- a/cfme/tests/cloud/test_providers.py
+++ b/cfme/tests/cloud/test_providers.py
@@ -320,15 +320,17 @@ def test_openstack_provider_has_api_version():
 class TestProvidersRESTAPI(object):
     @pytest.yield_fixture(scope="function")
     def arbitration_profiles(self, rest_api, setup_a_provider):
+        num_profiles = 2
         provider = rest_api.collections.providers.get(name=setup_a_provider.name)
         body = []
         providers = [{'id': provider.id}, {'href': provider.href}]
-        for i in range(2):
+        for i in range(num_profiles):
             body.append({
                 'name': 'test_settings_{}'.format(fauxfactory.gen_alphanumeric(5)),
                 'provider': providers[i % 2]
             })
         response = rest_api.collections.arbitration_profiles.action.create(*body)
+        assert len(response) == num_profiles
 
         yield response
 
@@ -340,9 +342,7 @@ class TestProvidersRESTAPI(object):
 
     @pytest.mark.tier(3)
     @pytest.mark.uncollectif(lambda: version.current_version() < '5.7')
-    @pytest.mark.parametrize(
-        "from_detail", [True, False],
-        ids=["from_detail", "from_collection"])
+    @pytest.mark.parametrize('from_detail', [True, False], ids=['from_detail', 'from_collection'])
     def test_cloud_networks_query(self, setup_a_provider, rest_api, from_detail):
         """Tests querying cloud providers and cloud_networks collection for network info.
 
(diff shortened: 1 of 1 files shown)


cfme/tests/cloud/test_providers.py
```diff
@@ -320,15 +320,17 @@ def test_openstack_provider_has_api_version():
 class TestProvidersRESTAPI(object):
     @pytest.yield_fixture(scope="function")
     def arbitration_profiles(self, rest_api, setup_a_provider):
+        num_profiles = 2
         provider = rest_api.collections.providers.get(name=setup_a_provider.name)
         body = []
         providers = [{'id': provider.id}, {'href': provider.href}]
-        for i in range(2):
+        for i in range(num_profiles):
```

cfme/tests/cloud/test_providers.py
```diff
@@ -320,15 +320,17 @@ def test_openstack_provider_has_api_version():
...
             body.append({
                 'name': 'test_settings_{}'.format(fauxfactory.gen_alphanumeric(5)),
                 'provider': providers[i % 2]
             })
         response = rest_api.collections.arbitration_profiles.action.create(*body)
+        assert len(response) == num_profiles
```
mshriver: **Optional**
Consider a soft_assert here, looks like the tests don't have a hard dependency on the number of profiles given by the fixture

This ties in with my comments below. If this is a soft_assert it could create ambiguity in the tests as to whether or not the test executed against 0, 1, or `num_profiles` profiles.

cfme/tests/cloud/test_providers.py
```diff
@@ -383,59 +384,67 @@ def test_create_arbitration_profiles(self, rest_api, arbitration_profiles):
         Metadata:
             test_flag: rest
         """
-        assert len(arbitration_profiles) > 0
```
mshriver: **Optional**
If the fixture uses a soft assert, this assert should be kept.

cfme/tests/cloud/test_providers.py
```diff
@@ -383,59 +384,67 @@ def test_create_arbitration_profiles(self, rest_api, arbitration_profiles):
...
         collection = rest_api.collections.arbitration_profiles
-        if from_detail:
-            methods = ['post', 'delete']
-            for i, ent in enumerate(arbitration_profiles):
-                ent.action.delete(force_method=methods[i % 2])
-                with error.expected("ActiveRecord::RecordNotFound"):
-                    ent.action.delete()
-        else:
+        collection.action.delete(*arbitration_profiles)
```
mshriver: **Optional**
This actually seems like two tests to me - one for deleting a single item from the collection, and one for deleting multiple items from the collection.

Can definitely be a TODO or something like that and doesn't have to be done in this PR.

I see you've moved the assert against the length of arbitration_profiles into the fixture, and it all plays together. I'm thinking it may be best to have a soft_assert in the fixture if the number of created doesn't match the intended, and then keep a hard assert inside the cases where the test won't actually assert/do anything if the list is empty.

cfme/tests/cloud/test_providers.py
```diff
@@ -383,59 +384,67 @@ def test_create_arbitration_profiles(self, rest_api, arbitration_profiles):
...
-        "from_detail", [True, False],
-        ids=["from_detail", "from_collection"])
+    @pytest.mark.parametrize('from_detail', [True, False], ids=['from_detail', 'from_collection'])
     def test_edit_arbitration_profiles(self, rest_api, arbitration_profiles, from_detail):
         """Tests editing of arbitration profiles.
 
         Metadata:
             test_flag: rest
         """
         response_len = len(arbitration_profiles)
```
izapolsk: OPTIONAL: I would rename response_len to profile_number or something similar

PR #4187, opened 2017-02-14
[RFR] Fix erroneous partial matches on sprout pools for jenkins jobs

mshriver:
The function to check if a pool already exists for a given Jenkins job uses the partial match from sprout.find_pools_by_description. It doesn't support full regex match, and a partial match will return jenkins-job-name-12345 and jenkins-job-name-other-56789

Verify that the pool with a partially matching description is an exact match for jenkins-job-name-[pool] and there isn't any other text between the pool ID and the job name

**This fix is untested**

I could use some help in testing this change. @mfalesni ?

FIXES RHCFQE-1871

### DIFF
--- a/cfme/test_framework/sprout/plugin.py
+++ b/cfme/test_framework/sprout/plugin.py
@@ -206,8 +206,16 @@ class SproutManager(object):
                 "Check if pool already exists for this %r Jenkins job", jenkins_job[0])
             jenkins_job_pools = self.client.find_pools_by_description(jenkins_job[0], partial=True)
             for pool in jenkins_job_pools:
-                log.info("Destroying the old pool %s for %r job.", pool, jenkins_job[0])
-                self.client.destroy_pool(pool)
+                # Some jobs have overlapping descriptions, sprout API doesn't support regex
+                # job-name-12345 vs job-name-master-12345
+                # the partial match alone will catch both of these, use regex to confirm pool
+                # description is an accurate match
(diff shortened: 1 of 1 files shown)


cfme/test_framework/sprout/plugin.py
```diff
@@ -206,8 +206,16 @@ def clean_jenkins_job(self, jenkins_job):
                 "Check if pool already exists for this %r Jenkins job", jenkins_job[0])
             jenkins_job_pools = self.client.find_pools_by_description(jenkins_job[0], partial=True)
             for pool in jenkins_job_pools:
-                log.info("Destroying the old pool %s for %r job.", pool, jenkins_job[0])
-                self.client.destroy_pool(pool)
+                # Some jobs have overlapping descriptions, sprout API doesn't support regex
+                # job-name-12345 vs job-name-master-12345
+                # the partial match alone will catch both of these, use regex to confirm pool
+                # description is an accurate match
+                if self.client.get_pool_description(pool) == '{}-{}'.format(jenkins_job[0], pool):
```
psav: Looks to me like the ```-``` is already in the name eg: ```Jenkins tag: jenkins-downstream-57z-tests-master-```

mshriver: I'll have to be smarter about adding that dynamically

mshriver: done

PR #4189, opened 2017-02-15
[1LP][RFR] utils/providers.py: Accept and propagate appliance= param

mfalesni:
This is required for Sprout provider scaenging functioning again, because managed_providers now returns crud objects which do not work by default on Sprout because Sprout's base_url is None, therefore get_or_create_current_appliance does not work.


mfalesni:
Ultimately we should end up with provider collection and these module methods will be removed :)

mfalesni:
I have also added an exception when old-style setup_provider ocde tries to sneak in to have a meaningful message.

PR #4192, opened 2017-02-15
[1LP][RFR]Fix and resetter for InfraProvider Details 

lkhomenk:
adding checks for paginator to avoid BZ 1422449 block and adding 'Summary View' to Details (it was added in 5.8)

This is mainly for 5.8 - as InfraProviders received Dashboard/Summary views in provider details.
There's a NEEDINFO regarding CloudProviders in that BZ .  It is always good to have  additional checks

### DIFF
--- a/cfme/infrastructure/provider/__init__.py
+++ b/cfme/infrastructure/provider/__init__.py
@@ -314,9 +314,11 @@ class All(CFMENavigateStep):
 
     def resetter(self):
         # Reset view and selection
-        tb.select("Grid View")
-        sel.check(paginator.check_all())
-        sel.uncheck(paginator.check_all())
+        if tb.exists("Grid View"):
+            tb.select("Grid View")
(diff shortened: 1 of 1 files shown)


lkhomenk:
update: Dashboard View became Default View in 5.8, it was added in 5.7 also for InfraProviders only

cfme/infrastructure/provider/__init__.py
```diff
@@ -314,9 +314,11 @@ def step(self):
 
     def resetter(self):
         # Reset view and selection
-        tb.select("Grid View")
-        sel.check(paginator.check_all())
-        sel.uncheck(paginator.check_all())
+        if tb.exists("Grid View"):
```
psav: I'm going to merge this but @mfalesni should we move the ```exists``` assertion inside the ```select```?

PR #4196, opened 2017-02-15
Fix test_provisioning emails

bargulg:
This should fix the problem with multiple emails having same subject, while still checking that there aren't multiple same emails.
There are still multiple emails with the same body, which is a bug - that's why I added BZ marker


dajoRH:
Lint report for commit 20b1f5109d31d5967193dff442fd3e2ff2ab3b6b:

* 2a57da17f6d1c738c0d965f5febf7c1c793abfe4

*CFME QE Bot*

PR #4203, opened 2017-02-15
[1LP][WIP] add 'action_exists' function for usage in REST API tests

mkoura:
There's at least one REST API collection where there is an action available, but is not listed in "actions". See https://bugzilla.redhat.com/show_bug.cgi?id=1422596

For REST API tests that are deleting items, I usually test "delete" with both DELETE and POST methods. In case of the bug above, the DELETE action is not listed, but is available, so the test would pass silently. For that reason I want to add this simple function for checking if action is listed in the collection and use it (mainly) in delete tests.

I'm not sure what's the best location for such function, `utils/rest.py` or `rest/utils.py`?

### DIFF
--- a/utils/rest.py
+++ b/utils/rest.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+"""Helper functions for tests using REST API."""
+
+
+def action_exists(collection, name, method='post'):
+    """Check if action is present for given collection/subcollection.
+
+    You can do following:
+
+    .. code-block:: python
+
+        assert action_exists(rest_api.collections.providers[0], 'delete', 'delete')
+
+    """
+    try:
+        collection_actions = collection._actions
+    except AttributeError:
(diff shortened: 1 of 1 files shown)


mfalesni:
@mkoura Wouldn't it be better to push it through as a method in the REST API client?

mkoura:
@mfalesni Could be, but it seemed to me that it's not that useful outside of tests development to be included in general REST API client.

mshriver:
Looks good if you decide to keep it here and not move it elsewhere.

mkoura:
@mfalesni @mshriver I've got different idea, will see if reviewers of the manageiq-api-client-python will like it.
https://github.com/ManageIQ/manageiq-api-client-python/pull/20

mkoura:
superseded by https://github.com/ManageIQ/manageiq-api-client-python/pull/20

PR #4204, opened 2017-02-16
[1LP][RFR] fix creating templates in test_rest_templates

mkoura:
* tests fix:
These tests often failed with errors like
`E           ValueError: No such 'vms' matching query {'name': 'test_rest_vm_CwrY'}!`
The reason is that the `mark_vm_as_template` function tried to preserved the original vm, which doesn't work on some providers and the vm is destroyed.

* "delete" tests expanded to test both POST and DELETE methods
There is a bug that prevents testing of delete with POST. I created two separate tests instead of one with parameters, because this way I'm able to block the test before fixtures are resolved (as opposed to `if BZ(...): pytest.skip(...)` for the affected test variant).

* add http response status checks

**PRT:**
tested on vsphere55 because of timeouts on rhevm36

{{pytest: cfme/tests/infrastructure/test_rest_templates.py -v --use-provider vsphere55}}

### DIFF
--- a/cfme/rest/gen_data.py
+++ b/cfme/rest/gen_data.py
@@ -248,7 +248,15 @@ def vm(request, a_provider, rest_api):
     vm_name = deploy_template(
         a_provider.key,
         "test_rest_vm_{}".format(fauxfactory.gen_alphanumeric(length=4)))
-    request.addfinalizer(lambda: a_provider.mgmt.delete_vm(vm_name))
+
+    @request.addfinalizer
+    def _finished():
+        try:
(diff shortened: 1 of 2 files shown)


dajoRH:
Lint report for commit e7b27b1261b1f9ee5255b4454bd8ec6513f18f47:

`cfme/tests/infrastructure/test_rest_templates.py`:
- :red_circle: Line 67:5: **N802** *function name should be lowercase*
- :red_circle: Line 81:5: **N802** *function name should be lowercase*

**No commit flaws detected**.

*CFME QE Bot*

PR #4205, opened 2017-02-16
[1LP][RFR] check response status code in test_provisioning_rest

mkoura:
HTTP response status code checking

**PRT:**
@mshriver 5.6 error is from teardown, the test itself passed. The `delete_vm` from mgmtsystem module is supposed to stop the vm before deleting it but for some reason it failed to do so. I could wrap `provider.mgmt.delete_vm(vm_name)` in try except and mask such issues, but I think it's better to notice them and maybe make mgmtsystem more robust over time. Looking briefly at the code responsible for stopping vm before deleting, it always returns True:
```
vm.stop()
self.wait_vm_stopped(vm_name)
return True
```

{{pytest: cfme/tests/infrastructure/test_provisioning_rest.py -v}}


mshriver:
5.6 PRT failure isn't clear to me and is in this test, please assess.

PR #4208, opened 2017-02-16
[1LP][RFR] Merging all table view tests into 1 file

gshefer:
{{pytest: cfme/tests/containers/test_table_views.py -v --use-provider cm-env1 }}

- merging all table view tests into 1 file
- delete replaced files
- added polarion marking


cfme/configure/settings.py
```diff
@@ -349,12 +349,17 @@ def __init__(self, appliance=None):
         Navigatable.__init__(self, appliance=appliance)
 
     @classmethod
-    def set_default_view(cls, button_group_name, default):
+    def save(cls):
```
mshriver: **Required**
Is the intent here to speed up the default view setting in the test when you're changing multiple views?

If so, I'm pretty sure I'd rather see this class take a list of button_group_names. Should be minor to refactor existing uses to pass a list, or to handle both strings and lists here.

gshefer: NP - WIP

gshefer: Well, Note that I need to set different default for each object, There is a problem with that since I'll also have to change the 'default' arg to 'defaults' as well, this can cause serious complexity right?
I can do it if you want.

gshefer: it's going to be something like:
`@classmethod
    def set_default_view(cls, button_group_names, defaults):

        if not isinstance(button_group_names, (list, tuple)):
            button_group_names = [button_group_names]
        if not isinstance(defaults, (list, tuple)):
            defaults = [defaults]*len(button_group_names)
        assert len(button_group_names) == len(defaults)

        for button_group_name, default in zip(button_group_names, defaults):
            bg = ButtonGroup(button_group_name)
            navigate_to(cls, 'All')
            if bg.active != default:
                bg.choose(default)
        sel.click(form_buttons.save)`

dajoRH:
*CFME QE Bot*

cfme/configure/settings.py
```diff
@@ -349,12 +349,20 @@ def __init__(self, appliance=None):
...
+        if not isinstance(defaults, (list, tuple)):
+            defaults = [defaults] * len(button_group_names)
+        assert len(button_group_names) == len(defaults)
+
+        for button_group_name, default in zip(button_group_names, defaults):
+            bg = ButtonGroup(button_group_name)
+            navigate_to(cls, 'All')
+            if bg.active != default:
+                bg.choose(default)
+        sel.click(form_buttons.save)
```
mshriver: **Required**
Save button will only be enabled if the form is changed, either set a boolean in the `bs.active != default` block, or detect the state of the button before clicking.

cfme/configure/settings.py
```diff
@@ -349,12 +349,20 @@ def __init__(self, appliance=None):
...
 
     @classmethod
-    def set_default_view(cls, button_group_name, default):
-        bg = ButtonGroup(button_group_name)
-        navigate_to(cls, 'All')
-        if bg.active != default:
-            bg.choose(default)
-            sel.click(form_buttons.save)
+    def set_default_view(cls, button_group_names, defaults):
+
```
mshriver: I like this enhancement, doesn't impact existing callers and allows for setting a list of views to a single given default. cool!

cfme/configure/settings.py
```diff
@@ -367,5 +375,9 @@ def get_default_view(cls, button_group_name):
 class DefaultViewAll(CFMENavigateStep):
     prerequisite = NavigateToAttribute('appliance.server', 'MySettings')
 
+    def am_i_here(self):
+        if match_location(title='Configuration'):
```
mshriver: **Required**
Add `controller='configuration'` param to `match_location` since 'Configuration' title actually matches a few pages.

cfme/configure/settings.py
```diff
@@ -349,11 +349,23 @@ def __init__(self, appliance=None):
         Navigatable.__init__(self, appliance=appliance)
 
     @classmethod
-    def set_default_view(cls, button_group_name, default):
-        bg = ButtonGroup(button_group_name)
-        navigate_to(cls, 'All')
-        if bg.active != default:
-            bg.choose(default)
+    def set_default_view(cls, button_group_names, defaults):
+
```
psav: **REQUIRED:** For something that seems to have a fair amount of restrictions placed on it below, it would be very useful to have a docstring here, with some example usage. I'm only saying this because the default behaviour (ie having one default that gets expanded for multiple ones seems a little unintuitive.

**OPTIONAL:** I would have preferred to go for tuples for the (button_grp_name, default) and then a fallback default, so that just button_grp_name could be passed if the default was perhaps unchanged. However, I'm happy to merge as is. Just need the docs above.

gshefer: WIP

PR #4220, opened 2017-02-17
[WIPTEST] fixed two wrong navigate destinations

izapolsk:
'Global Filters' is no longer selectable. So, I removed it from the navigation.


dajoRH:
*CFME QE Bot*

mfalesni:
I think that this was resolved in #4216. Reopen if that is not true :)

izapolsk:
Yes, your fixed is convenient but I also planned to fix another thing :)

PR #4221, opened 2017-02-17
[1LP][RFR]Update blocker BZ info for RHEV

nachandr:
Purpose or Intent
=================
Updating blocker BZ info for RHEV.


mshriver:
The intended cases were blocked in PRT, won't hold you to the scope of fixing all the test failures there if this is all you want to merge right now.

PR #4225, opened 2017-02-20
[1LP][RFR] fix test_power_control_rest for scvmm provider

mkoura:
Fixing the test_power_control_rest tests for scvmm provider.

PRT:
some test still randomly timeout on PRT even though I eventually increased the timeout to 4000 sec (66min!)
I've tried many times to have all tests passed, but I can't beat PRT :( Results combined from multiple runs will have to do.
downstream-56z: http://10.16.4.32/trackerbot/pr/run/10710#sN6Vs8lA
downstream-57z: http://10.16.4.32/trackerbot/pr/run/10800#UqvzTpPd
upstream: http://10.16.4.32/trackerbot/pr/run/10630#phezGfFH

{{pytest: cfme/tests/cloud_infra_common/test_power_control_rest.py -v --use-provider scvmm2016}}

### DIFF
--- a/cfme/tests/cloud_infra_common/test_power_control_rest.py
+++ b/cfme/tests/cloud_infra_common/test_power_control_rest.py
@@ -42,7 +42,7 @@ def vm_obj(request, provider, setup_provider_modscope, small_template_modscope):
 def verify_vm_power_state(vm_obj, state, action=None):
     vm = vm_obj.get_vm_via_rest()
     if vm_obj.provider.one_of(GCEProvider, EC2Provider, SCVMMProvider):
-        num_sec = 2500  # extra time for slow providers
+        num_sec = 4000  # extra time for slow providers
     else:
         num_sec = 1000
     if action:
--- a/cfme/tests/cloud_infra_common/test_power_control_rest.py
+++ b/cfme/tests/cloud_infra_common/test_power_control_rest.py
@@ -9,6 +9,7 @@ from utils.wait import wait_for
 from utils.log import logger
 from cfme.cloud.provider.gce import GCEProvider
 from cfme.cloud.provider.ec2 import EC2Provider
+from cfme.infrastructure.provider.scvmm import SCVMMProvider
 
 
 pytestmark = [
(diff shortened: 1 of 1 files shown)


cfme/tests/cloud_infra_common/test_power_control_rest.py
```diff
@@ -40,15 +41,18 @@ def _delete_vm():
...
+        num_sec = 4000  # extra time for slow providers
     else:
         num_sec = 1000
     if action:
         action()
-    wait_for(lambda: vm.power_state == state,
-        num_sec=num_sec, delay=20, fail_func=vm.reload,
-        message='Wait for VM state `{}` (current state: {})'.format(state, vm.power_state))
+
+    def _state_changed():
```
mshriver: **Optional**

Scope is really limited but the name is a bit misleading as it's not detecting state changes but a specific state. Consider `_check_state()`

cfme/tests/cloud_infra_common/test_power_control_rest.py
```diff
@@ -40,15 +41,18 @@ def _delete_vm():
 
 def verify_vm_power_state(vm_obj, state, action=None):
     vm = vm_obj.get_vm_via_rest()
-    if vm_obj.provider.one_of(GCEProvider, EC2Provider):
-        num_sec = 1600  # extra time for slow providers
+    if vm_obj.provider.one_of(GCEProvider, EC2Provider, SCVMMProvider):
```
mshriver: Since this set is used more than once, consider making it a constant for the module.

cfme/tests/cloud_infra_common/test_power_control_rest.py
mkoura: but it reads well together with wait_for - `wait_for(_state_change` :)

cfme/tests/cloud_infra_common/test_power_control_rest.py
mkoura: There's also `vm_obj.provider.one_of(GCEProvider, EC2Provider)` in other place, so I would rather leave it explicit as it is since there are different combinations in different places.

cfme/tests/cloud_infra_common/test_power_control_rest.py
mshriver: heh, you're right, it does.

Can I suggest removing this and replacing it with common.VM.wait_for_vm_state_change()?

mshriver: hehe, it does. Its optional from my point of view. :)

cfme/tests/cloud_infra_common/test_power_control_rest.py
```diff
@@ -40,15 +41,18 @@ def _delete_vm():
...
     if action:
         action()
-    wait_for(lambda: vm.power_state == state,
-        num_sec=num_sec, delay=20, fail_func=vm.reload,
-        message='Wait for VM state `{}` (current state: {})'.format(state, vm.power_state))
+
+    def _state_changed():
+        vm.reload()
+        return vm.power_state == state
+    wait_for(_state_changed, num_sec=num_sec, delay=20,
```
mshriver: This is failing in PRT, looks like it is assuming that we're on the details page already and the screenshots show that we're on the VM's page filtered for the device under test.

mkoura:
@mshriver where do you see that? All I can see are `TimedOutError`s and related issues. If scvmm is in such a bad shape that it can't change power state in 4000 seconds, there's not much I can do about that. This PR fixes real issue and is waiting for more that two weeks to be merged because these timeouts.

cfme/tests/cloud_infra_common/test_power_control_rest.py
mkoura: It doesn't care, the state is read from REST API.

mshriver:
Noted about the timeouts, they are failures for svcmm2016 to bring the VM to the 'On' state.

@jteehan just updated the scvmm2016 provider map in cfme_data, so I've restarted PRT.

PR #4180, opened 2017-02-13
[1LP][RFR] Update Advanced Search buttons, BZ1380430

mshriver:
alt-text for the advanced search buttons will be fixed here:
https://github.com/ManageIQ/manageiq-ui-classic/pull/367

The required bug has been fixed in 57z and 58z cloudforms, and is in upstream MIQ. With the announcement of EOL of version support of CFME 56z, this can now be merged.
https://bugzilla.redhat.com/show_bug.cgi?id=1380430

This fulfills the 'automate_bug' flag in the BZ and reverts to previous design using FormButton instead of specific locators for advanced search buttons.

## Pytest Results
Marginally better than current master/downstream-stable results against 58z.  Failures well outside the scope of what I'm trying to accomplish here, that should be addressed by focus area owner of CFME QE.

* 57z: 73% overall passing rate, test_adv_search_vms failing because search widget isn't rendered on page (35% pass)
* 58z: 76% overall passing rate, test_adv_search_vms failing because search widget isn't rendered on page (35% pass)

{{ pytest: -k "advanced_search" }}

PR #4227, opened 2017-02-20
[RFR] fix for advanced search issue

izapolsk:
Purpose or Intent
=================
We have 20-30+ tests failing due to changes in advanced search view.
This PR should fix most such issues in 5.7.
{pytest: --use-provider=vsphere55 cfme/tests/infrastructure/test_advanced_search_vms.py}

This is just a quick fix. There will be more full one #4180 soon.

Yaml fix - https://gitlab.cee.redhat.com/cfme-qe/cfme-qe-yamls/merge_requests/96

### DIFF
--- a/cfme/web_ui/expression_editor.py
+++ b/cfme/web_ui/expression_editor.py
@@ -34,8 +34,8 @@ def _expressions_root():
 # Buttons container
 buttons = Region(
     locators=dict(
-        commit="//img[@alt='Commit expression element changes']",
-        discard="//img[@alt='Discard expression element changes']",
+        commit="//button[@title='Commit expression element changes']",
+        discard="//button[@title='Discard expression element changes']",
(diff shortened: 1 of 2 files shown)


cfme/web_ui/expression_editor.py
```diff
@@ -34,8 +35,12 @@ def _expressions_root():
 # Buttons container
 buttons = Region(
     locators=dict(
-        commit="//img[@alt='Commit expression element changes']",
-        discard="//img[@alt='Discard expression element changes']",
+        commit=deferred_verpick(
+            {version.LOWEST: "//img[@alt='Commit expression element changes']",
+             '5.7': "//button[@title='Commit expression element changes']"}),
```
mfalesni: Doe sdeferred verpick work here? Region should be able to process dict as verpick automatically anyways ... deferred_verpick is only for placing on classes.

PR #4238, opened 2017-02-20
[1LP][RFR]Fix 57z test_provider_discovery

mshriver:
The provider.exists call was always false because the discovered provider doesn't have a matching name.

Added IPAppliance.managed_provider_names and used it to validate that each discovered provider's IP address was in a managed provider's name.

Tested locally against 5.6.4 and 5.7.1.1


izapolsk:
looks ok but test results aren't ready yet

PR #4244, opened 2017-02-21
[1LP][RFR] unify all automation_requests tests into one file

mkoura:
There are actions for automation requests available in both `/api/automation_requests` and `/api/requests`.
This PR unifies tests using these two collections for automation requests into single file. It reuses the same test code where possible.
The "edit" action is available only in `/api/requests`

Testing on scvmm2016 due to problems with small template on other infrastructure providers.

{{pytest: cfme/tests/infrastructure/test_rest_automation_request.py -v --use-provider vsphere6-nested}}


mshriver:
Quite a large PR, I'm thinking we might want to have a 2nd review on this. All tests executed and passed though, nice work!

mkoura:
@izapolsk can you please also review?

dajoRH:
*CFME QE Bot*

dajoRH:
*CFME QE Bot*

cfme/tests/infrastructure/test_rest_automation_request.py
```diff
@@ -1,123 +1,295 @@
...
+        def _gen_automation_requests(output):
+            api = MiqApi(entry_point, auth, verify_ssl=False)
+            requests_data = _automation_requests_data(
+                'nonexistent_vm', requests_collection=True, approve=False)
+            api.collections.requests.action.create(*requests_data[:2])
+            result = (api.response.status_code, api.response.json())
+            output.put(result)
+
+        processes = [
+            mp.Process(target=_gen_automation_requests, args=(output,))
```
izapolsk: OPTIONAL: I would use threads instead processes

mkoura: I wanted to make sure that it would be really parallel. With threads it might not be the case due to GIL.

izapolsk: I don't think you would suffer from GIL in such an easy case with ~ 8 threads.

mkoura: I wanted to emulate situation similar to what's described in https://bugzilla.redhat.com/show_bug.cgi?id=1418445
While in theory GIL will not affect I/O bound task, using multiprocessing works for sure and doesn't pose significant overhead for tasks like this one.
I'll investigate this more as I have plans to write other tests like this one, maybe I'll switch to threads in the end.

izapolsk: btw, as far as I'm aware there are plans to move on Py3 where you can encounter GIL much more rarely.

PR #4246, opened 2017-02-21
[RFR][1LP] Updating metrics collection fixture

nachandr:
With the existing code and the  the condition checking before host_rising or vm_rising can be set to 
True ,this exception would be occasionally thrown.

Timed out waiting for metrics to be collected"

Since we just want to make sure metrics are being collected for VMs and hosts, the condition checking made through this PR should suffice.


nachandr:
@jkrocil , Could you review this when you get a chance?

PR #4247, opened 2017-02-21
[RFR] one more fix for search tests

izapolsk:
Purpose or Intent
=================
one more fix for search tests.

{{pytest: cfme/tests/infrastructure/test_advanced_search_vms.py}}


psav:
Just initiated a retest

izapolsk:
rest of test fail due to BZ(1425840). I haven't added blockers because when those would run in different slaved, the couldn't appear.

PR #4249, opened 2017-02-21
[RFR][1LP] Fix the typo: quit -> close. Diaper hid it, so diaper removed and logging added

mfalesni:


utils/browser.py
```diff
@@ -310,11 +310,10 @@ def quit(self):
...
         self._consume_cleanups()
         try:
-            self.factory.quit(self.browser)
-        except:
-            # Due to the multitude of exceptions can be thrown when attempting to kill the browser,
-            # Diaper Pattern!
-            pass
+            self.factory.close(self.browser)
+        except Exception as e:
+            log.error('An exception happened during browser shutdown:')
```
jaryn:
```
log.exception('An exception happened during browser shutdown:')
```
Should be enough
```
In [9]: try:
   ...:     raise Exception("Foo")
   ...: except Exception as e:
   ...:     logging.exception("Oh crap! It broke.")
   ...:     
ERROR:root:Oh crap! It broke.
  File "<ipython-input-9-be4eb48ce636>", line 2, in <module>
    raise Exception("Foo")
Exception: Foo
```

PR #4255, opened 2017-02-22
[1LP][RFR] check response status code in test_vm_retirement_rest

mkoura:
* check HTTP response status code
* refactoring

{{pytest: cfme/tests/infrastructure/test_vm_retirement_rest.py -v --use-provider rhevm36}}

### DIFF
--- a/cfme/tests/infrastructure/test_vm_retirement_rest.py
+++ b/cfme/tests/infrastructure/test_vm_retirement_rest.py
@@ -1,43 +1,31 @@
 # -*- coding: utf-8 -*-
 import datetime
 import pytest
-import fauxfactory
 
 from cfme import test_requirements
 from cfme.rest.gen_data import a_provider as _a_provider
-from utils.version import current_version
-from utils.virtual_machines import deploy_template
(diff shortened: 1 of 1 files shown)


cfme/tests/infrastructure/test_vm_retirement_rest.py
```diff
@@ -1,43 +1,31 @@
...
-        num_sec=600, delay=5)
-    return vm_name
+    return _vm(request, a_provider, rest_api)
 
 
 @pytest.mark.tier(3)
-@pytest.mark.uncollectif(lambda: current_version() <= "5.5.2.4")
 @pytest.mark.parametrize(
-    "multiple", [True, False],
+    "from_collection", [True, False],
```
psav: **FUTURE:** I am not 100% loving the fixture name here. We should call it something like ```source```, the only reason I say this is that ```from_collection``` conflicts with the actual ID ```from_collection``` yielding confusion around having two different IDs that always provide the same fixture, and one of them matches the name.......gah that sounded terrible....hence my confusion initially. I'd considering changing it to ```method``` or ```source``` but not in this PR.

PR #4259, opened 2017-02-22
[1LP][RFR] refactoring; new tests; check response status code for test_tag_category

mkoura:
* check HTTP response status code
* refactoring
* test also DELETE method

{{pytest: cfme/tests/configure/test_tag_category.py -v -k TestCategoriesViaREST}}

### DIFF
--- a/cfme/tests/configure/test_tag_category.py
+++ b/cfme/tests/configure/test_tag_category.py
@@ -5,7 +5,7 @@ from cfme.rest.gen_data import categories as _categories
 import pytest
 from utils.update import update
 from utils.wait import wait_for
-from utils import error, version
+from utils import error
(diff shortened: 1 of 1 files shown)


cfme/tests/configure/test_tag_category.py
```diff
@@ -21,63 +21,90 @@ def test_category_crud():
 
 
 class TestCategoriesViaREST(object):
-
     @pytest.fixture(scope="function")
     def categories(self, request, rest_api):
-        return _categories(request, rest_api, num=5)
+        response = _categories(request, rest_api, num=5)
```
mshriver: **optional**
store the number in a variable since its used twice.

PR #4260, opened 2017-02-22
[WIPTEST]Label creation and verification

jdemon519:
{{pytest: cfme/tests/containers/test_label.py -v --use-provider cm-env1 }}

The test adds a label on the master node in Openshift, then it logs in to CFME and verifies the existence of the created label.

### DIFF
--- a/cfme/tests/containers/test_label.py
+++ b/cfme/tests/containers/test_label.py
@@ -1,7 +1,7 @@
 import pytest
 
 from cfme.fixtures import pytest_selenium as sel
-from cfme.containers.node import Node
+from cfme.containers.node import NodeCollection, Node
 from utils import testgen, version
 from cfme.web_ui import CheckboxTable, toolbar as tb
 from utils.appliance.implementations.ui import navigate_to
@@ -26,7 +26,6 @@ def test_create_label_check(ssh_client, provider):
         then it verifies the label exists and is visible on the summary
         page.There is a sufficient time provided so that CFME is updated
         with the latest data
-
     """
 
     hostname = conf.cfme_data.get('management_systems', {})[provider.key] \
@@ -55,7 +54,7 @@ def test_create_label_check(ssh_client, provider):
     provider.validate_stats(ui=True)
     time.sleep(6)
 
-    navigate_to(Node, 'All')
+    navigate_to(NodeCollection, 'All')
     tb.select('List View')
 
     list_tbl = CheckboxTable(table_locator="//div[@id='list_grid']//table")
@@ -65,10 +64,4 @@ def test_create_label_check(ssh_client, provider):
 
     for name in names:
         obj = Node(name, provider)
-        obj.summary.reload()
-        labels_all = obj.summary.labels
-        lbls_cfme = list(labels_all)
-        lbls_cfme_new = [i for j in lbls_cfme for i in j]
-        lst_lbls_found = ['kube', 'master']
-        found = any(x in lst_lbls_found for x in lbls_cfme_new)
-        assert found
+        assert obj.get_detail('Labels', 'kube') == 'master'
--- a/cfme/tests/containers/test_label.py
+++ b/cfme/tests/containers/test_label.py
@@ -41,7 +41,7 @@ def test_create_label_check(ssh_client, provider):
 
     ssh_client.run_command(
         "master_node=$(oc get nodes | awk 'NR==2' | sed -e 's/\s.*$//'); "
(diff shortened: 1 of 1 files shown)


cfme/tests/containers/test_label.py
```diff
@@ -0,0 +1,67 @@
...
+
+    hostname = conf.cfme_data.get('management_systems', {})[provider.key] \
+        .get('hostname', [])
+    username, password = provider.credentials['token'].principal, \
+        provider.credentials['token'].secret
+
+    ssh_client = ssh_client(
+        hostname=hostname,
+        username=username,
+        password=password)
```
gshefer: ssh_client fixture is doing the work for you, lines 30-39 are redundant.

cfme/tests/containers/test_label.py
```diff
@@ -0,0 +1,67 @@
...
+        "master_node=$(oc get nodes | awk 'NR==2' | sed -e 's/\s.*$//'); "
+        "oc label nodes $master_node kube=master")
+
+    navigate_to(provider, 'Details')
+    tb.select(
+        'Configuration',
+        'Refresh items and relationships',
+        invokes_alert=True)
+    sel.handle_alert()
+    time.sleep(6)
```
gshefer: Why is the sleep here? if you are waiting for something use wait_for

cfme/tests/containers/test_label.py
```diff
@@ -0,0 +1,67 @@
...
+
+    tb.select('Reload Current Display')
+    provider.validate_stats(ui=True)
+    time.sleep(6)
```
gshefer: Why is the sleep here? if you are waiting for something use wait_for

cfme/tests/containers/test_label.py
```diff
@@ -0,0 +1,67 @@
...
+    pytest.mark.uncollectif(
+        lambda provider: version.current_version() < "5.6"),
+    pytest.mark.usefixtures('setup_provider'),
+    pytest.mark.tier(2)]
+pytest_generate_tests = testgen.generate(
+    [ContainersProvider], scope='function')
+
+# CMP-10572
+
+
```
gshefer: please use polarion marker!

cfme/tests/containers/test_label.py
```diff
@@ -0,0 +1,67 @@
...
+    tb.select('List View')
+
+    list_tbl = CheckboxTable(table_locator="//div[@id='list_grid']//table")
+    names = [r.name.text for r in list_tbl.rows()]
+    remove_names = [1, 2]
+    names = [v for i, v in enumerate(names) if i not in remove_names]
+
+    for name in names:
+        obj = Node(name, provider)
+        assert obj.get_detail('Labels', 'kube') == 'master'
```
gshefer: use obj.summary.labels.kube instead. more safe.
use soft_assert (a fixture) so you will be able to test all names. also provide an information once failed, something like:
kube = obj.summary.labels.kube.lower(); soft_assert(=='high', 'Expected for "kube" label to be "master", got {} instead' .format(kube))

gshefer:
I would create a fixture that produce a (random) label and pass it into the function and delete it after the test.

cfme/tests/containers/test_label.py
```diff
@@ -0,0 +1,67 @@
...
+        page.There is a sufficient time provided so that CFME is updated
+        with the latest data
+    """
+
+    hostname = conf.cfme_data.get('management_systems', {})[provider.key] \
+        .get('hostname', [])
+    username, password = provider.credentials['token'].principal, \
+        provider.credentials['token'].secret
+
+    ssh_client = ssh_client(
```
mshriver: On top of @gshefer  comment, I'd actually like to hold this PR until #4316 is merged, and you use the appliance fixture instead and access the ssh_client through it.

jdemon519: @mshriver I don't think appliance fixture would make any difference since I am logging in to Openshift, not the appliance.

mshriver: :smh:

gshefer: @mshriver So #4316 is merged. but as Tony said this ssh_client is of the appliance right? if so, lets create (a global - not only for this test) ssh client fixture for OCP client.

jdemon519: @gshefer I am finishing up the test that creates labels for all objects and thus will replace the two labels PRs. Indeed, we do need an ssh client fixture for the OCP client, so I'll create that as well. Good call.

PR #4265, opened 2017-02-23
[WIP]Fix host drift analysis

niyazRedhat:
Fixing host drift locators.
{{pytest: -v -k 'test_host_drift_analysis.py'}}

### DIFF
--- a/cfme/infrastructure/host.py
+++ b/cfme/infrastructure/host.py
@@ -402,16 +402,28 @@ class Host(Updateable, Pretty, Navigatable, PolicyProfileAssignable):
         tb.select("Select up to 10 timestamps for Drift Analysis")
 
         # Make sure the section we need is active/open
-        sec_loc_map = {
-            'Properties': 'Properties',
-            'Security': 'Security',
-            'Configuration': 'Configuration',
-            'My Company Tags': 'Categories'}
-        active_sec_loc = "//div[@id='all_sections_treebox']//li[contains(@id, 'group_{}')]"\
-            "/span[contains(@class, 'dynatree-selected')]".format(sec_loc_map[section])
-        sec_checkbox_loc = "//div[@id='all_sections_treebox']//li[contains(@id, 'group_{}')]"\
-            "//span[contains(@class, 'dynatree-checkbox')]".format(sec_loc_map[section])
-        sec_apply_btn = "//div[@id='accordion']/a[contains(normalize-space(text()), 'Apply')]"
+        if version.current_version() >= '5.7':
+            sec_loc_map = {
+                'Properties': 'Properties',
+                'Security': 'Security',
+                'Configuration': 'Configuration',
+                'My Company Tags': 'Categories'}
+            active_sec_loc = "//div[@id='all_sectionsbox']//li[@title='{}']"\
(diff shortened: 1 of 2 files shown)


cfme/infrastructure/host.py
```diff
@@ -402,16 +402,28 @@ def equal_drift_results(self, row_text, section, *indexes):
...
-        sec_checkbox_loc = "//div[@id='all_sections_treebox']//li[contains(@id, 'group_{}')]"\
-            "//span[contains(@class, 'dynatree-checkbox')]".format(sec_loc_map[section])
-        sec_apply_btn = "//div[@id='accordion']/a[contains(normalize-space(text()), 'Apply')]"
+        if version.current_version() >= '5.7':
+            sec_loc_map = {
+                'Properties': 'Properties',
+                'Security': 'Security',
+                'Configuration': 'Configuration',
+                'My Company Tags': 'Categories'}
+            active_sec_loc = "//div[@id='all_sectionsbox']//li[@title='Categories']"\
```
mshriver: **Required** 
This is now hard coded to locate the 'Categories' list item, which is 'My Company Tags'. There is no formatting metachar, `{}`, in the xpath string,  so the `format()` does nothing here. It should format the sec_loc_map value in where 'Categories' is in the xpath.

Won't break for the test you're changing now, but it'll break for any other section.

cfme/web_ui/__init__.py
```diff
@@ -3399,7 +3399,8 @@ def get_cell(self, row_text, col_index):
             Selenium element of the cell.
         """
         self.expand_all_sections()
-        cell_loc = ".//div/div[1][contains(., '{}')]/../div[{}]".format(row_text, col_index + 2)
+        cell_loc = ".//div/div[1][contains(., '{}')]//tbody/tr[{}]/td".format(
```
mshriver: This xpath isn't matching the driftgrid in 5.7.

suggest trying something more like the following, with a bit appended for the value cell instead of the row_text cell.
`'.//div[@id="compare-grid"]/table//th[contains(normalize-space(.), "{}")]'.format(row_text)`

niyazRedhat: @mshriver As you said I tried locally but got the same results then I removed the `root=self.loc` at line number 3404 and ran my test so it moves into next steps means the path matching the dirftGrid but unable to check drift differences and return false.
I got the following error:
`>                   raise SoftAssertionError(_thread_locals.caught_asserts)
E                   SoftAssertionError: 
E                   Drift analysis results are equal when they shouldn't be (cfme/tests/infrastructure/test_host_drift_analysis.py:133)

../../../fixtures/soft_assert.py:62: SoftAssertionError
========================================================== 1 passed, 1 error in 1317.31 seconds ===========================================================`

niyazRedhat: For just debugging purpose i also tried this one:

`cell_loc = "//div[contains(normalize-space(.), '{}')]/table/tbody/tr[12]/td[{}]".format(row_text, col_index + 1))`
getting the same result `SoftAssertionError`

niyazRedhat:
The PR is quite old.I am closing this one and later on I will open new pr with latest fix.

PR #4271, opened 2017-02-23
[1LP][RFR] Add Resume from Retirement Tests

mshriver:
Add new automation for resuming a retired instance.

Skip for known bug, unblock for infra providers

Parametrize on resume via two methods:
* Set a new retirement date in the future
* Remove the set retirement date

{{ pytest: cfme/tests/cloud_infra_common/test_retirement.py --use-provider vsphere6-nested --use-provider ec2west --use-provider rhos7-ga --use-provider rhevm36 --long-running }}

## PRT Results
5.7 setup failures due to provider problems. It's BZ skipping where I want it to now though.

### DIFF
--- a/cfme/tests/cloud_infra_common/test_retirement.py
+++ b/cfme/tests/cloud_infra_common/test_retirement.py
@@ -5,7 +5,9 @@ from datetime import date, timedelta, datetime
 from cfme import test_requirements
 from cfme.common.provider import CloudInfraProvider
 from cfme.common.vm import VM
+from cfme.infrastructure.provider import InfraProvider
 from cfme.web_ui import toolbar as tb
+
 from utils import testgen
 from utils.blockers import BZ
 from utils.generators import random_vm_name
@@ -18,12 +20,11 @@ from utils.version import pick, current_version
 
 pytest_generate_tests = testgen.generate(
     gen_func=testgen.providers,
-    filters=[ProviderFilter(classes=[CloudInfraProvider], required_flags=['provision', 'retire'])],
-    scope='module')
+    filters=[ProviderFilter(classes=[CloudInfraProvider], required_flags=['provision', 'retire'])])
 
 
 pytestmark = [
-    pytest.mark.usefixtures('setup_provider_modscope'),
+    pytest.mark.usefixtures('setup_provider'),
     pytest.mark.tier(2),
     pytest.mark.long_running
 ]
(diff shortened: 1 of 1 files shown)


cfme/tests/cloud_infra_common/test_retirement.py
```diff
@@ -141,3 +141,29 @@ def test_unset_retirement_date(test_vm):
...
+    Two methods to resume:
+    1. Set a retirement date in the future
+    2. Remove the set retirement date
+    """
+    num_days = 5
+
+    test_vm.retire()
+    verify_retirement_state(test_vm)
+
+    if remove_date:
```
mshriver: probably going to turn this into a ternary.

cfme/tests/cloud_infra_common/test_retirement.py
```diff
@@ -141,3 +142,31 @@ def test_unset_retirement_date(test_vm):
...
+
+    retire_date = None if remove_date else generate_retirement_date(delta=num_days)
+    test_vm.set_retirement_date(retire_date)
+
+    if BZ(1419150, forced_streams='5.6').blocks and not remove_date:
+        # The date is wrong, but we can still test unset
+        logger.warning('Skipping test step verification for BZ 1419150')
+    else:
+        verify_retirement_date(test_vm, expected_date=retire_date if retire_date else 'Never')
+    assert test_vm.is_retired is False
```
izapolsk: REQUIRED: "is" shouldn't be used here. you compare value not identity

mshriver:
@izapolsk 5.6 failures are due to sprout, SSH unavailable exception killed the test at the very start.

cfme/tests/cloud_infra_common/test_retirement.py
mshriver: I think I disagree with you here Ievgen, because although 'is' makes an identity comparison, I want to assert that the response from is_retired is boolean False, and doesn't equate to boolean false (ex. `0`).

Identity comparison is used throughout our codebase for booleans. Definitely not for non-boolean value comparison though.

psav:
Still see failures - rerunning

cfme/tests/cloud_infra_common/test_retirement.py
```diff
@@ -133,11 +134,39 @@ def test_unset_retirement_date(test_vm):
...
     else:
         verify_retirement_date(test_vm, expected_date=retire_date)
 
     test_vm.set_retirement_date(None)
     verify_retirement_date(test_vm, expected_date='Never')
+
+
+@test_requirements.retirement
+@pytest.mark.meta(blockers=[BZ(1306471, forced_streams=['5.6'],
+                               unblock=lambda provider: isinstance(provider, InfraProvider))])
```
jkrocil: You can do `unblock=lambda provider: provider.one_of(InfraProvider)` here.

mshriver:
Hm, first try 'Dismissing' a review @jkrocil. Probably won't do that again.

PR #4286, opened 2017-02-27
[1LP][RFR] Removed test_folder_field_scope

quarckster:
Purpose or Intent
=================

test_folder_field_scope should be ~~uncollected not skipped~~ removed.

### DIFF
--- a/cfme/tests/control/test_bugs.py
+++ b/cfme/tests/control/test_bugs.py
@@ -12,9 +12,6 @@ from cfme.control.explorer.conditions import VMCondition
 from cfme.control.explorer.alert_profiles import VMInstanceAlertProfile
 from cfme.infrastructure.virtual_machines import Vm
 from utils.appliance.implementations.ui import navigate_to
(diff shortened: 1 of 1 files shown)


cfme/tests/control/test_bugs.py
```diff
@@ -175,7 +175,7 @@ def test_scope_windows_registry_stuck(request, setup_a_provider):
 
 
 @pytest.mark.meta(blockers=[1209538], automates=[1209538])
-@pytest.mark.skipif(current_version() > "5.5", reason="requires cfme 5.5 and lower")
+@pytest.mark.uncollectif(current_version() > "5.5", reason="requires cfme 5.5 and lower")
```
otsuman: Do you need this test at all? Because we don't support versions lower than 5.6

dajoRH:
*CFME QE Bot*

mshriver:
Ignoring PRT results since you're just removing a test. other failures are outside your scope here.

PR #4292, opened 2017-02-28
[1LP][RFR] test_rest_services fixes and refactoring

mkoura:
* tests fixes
* refactoring
* check HTTP response statuses

**PRT:**
(updated 14:17 10.3.2017)
See run http://10.16.4.32/trackerbot/pr/run/11160
The only failed test is "TestServiceTemplateRESTAPI.test_delete_service_template_post" because BZ 1427338 is no longer blocker but the fix is not available in the tested version - this was solved by last commit.

{{pytest: cfme/tests/services/test_rest_services.py -v --long-running --use-provider vsphere6-nested}}

### DIFF
--- a/cfme/tests/services/test_rest_services.py
+++ b/cfme/tests/services/test_rest_services.py
@@ -150,7 +150,7 @@ class TestServiceRESTAPI(object):
             service.reload()
             assert service.name == new_names[i]
 
-    @pytest.mark.meta(blockers=[BZ(1414852, forced_streams=['5.7', 'upstream'])])
+    @pytest.mark.meta(blockers=[BZ(1414852, forced_streams=['5.7', '5.8', 'upstream'])])
     def test_delete_service_post(self, rest_api, services):
         """Tests deleting services from detail using POST method.
 
@@ -311,14 +311,14 @@ class TestServiceRESTAPI(object):
             assert rest_api.response.status_code == 200
             wait_for_vm_power_state(vm, resulting_state)
 
-        _action_and_check('start', 'on')
+        wait_for_vm_power_state(vm, 'on')
         _action_and_check('stop', 'off')
         _action_and_check('start', 'on')
         _action_and_check('suspend', 'suspended')
         _action_and_check('start', 'on')
 
-    @pytest.mark.meta(blockers=[BZ(1416146, forced_streams=['5.7', 'upstream'])])
     @pytest.mark.uncollectif(lambda: version.current_version() < '5.7')
+    @pytest.mark.meta(blockers=[BZ(1416146, forced_streams=['5.7', '5.8', 'upstream'])])
     def test_create_service_from_parent(self, request, rest_api):
         """Tests creation of new service that reference existing service.
 
@@ -376,8 +376,8 @@ class TestServiceRESTAPI(object):
         child.reload()
         assert child.ancestry == str(parent.id)
 
-    @pytest.mark.meta(blockers=[BZ(1416903, forced_streams=['5.7', 'upstream'])])
     @pytest.mark.uncollectif(lambda: version.current_version() < '5.7')
+    @pytest.mark.meta(blockers=[BZ(1416903, forced_streams=['5.7', '5.8', 'upstream'])])
     def test_power_parent_service(self, request, rest_api, service_data):
         """Tests that power operations triggered on service parent affects child service.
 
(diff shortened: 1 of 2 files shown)


mshriver:
I liked the <100 line PRs better ;)

cfme/tests/services/test_rest_services.py
```diff
@@ -3,20 +3,18 @@
 import fauxfactory
 import pytest
 
-from manageiq_client.api import APIException
-
 from cfme.rest.gen_data import dialog as _dialog
```
mshriver: **Optional**
Consider combining these imports:

```
from cfme.rest.gen_data import (
    dialog as _dialog, services as _services, service_data as _service_data)
```

cfme/tests/services/test_rest_services.py
```diff
@@ -41,49 +39,73 @@ def a_provider():
         pytest.skip("It's not possible to set up any providers, therefore skipping")
 
 
+def wait_for_vm_power_state(vm, resulting_state):
+    wait_for(
+        lambda: vm.power_state == resulting_state,
```
mshriver: **Optional**

Change this up a bit, providing `vm.power_state` as `func` and `resulting_state` as `fail_condition`.
From the wait_for doc:
fail_condition: An object describing the failure condition that should be tested
            against the output of func. If func() == fail_condition, wait_for continues
            to wait.

cfme/tests/services/test_rest_services.py
```diff
@@ -116,121 +137,156 @@ def test_edit_multiple_services(self, rest_api, services):
...
+            for service in services:
+                service.action.retire()
+                assert rest_api.response.status_code == 200
+        else:
+            rest_api.collections.services.action.retire(*services)
+            assert rest_api.response.status_code == 200
+
+        for service in services:
+            wait_for(
+                lambda: not rest_api.collections.services.find_by(name=service.name),
```
mshriver: **Optional**
Consider using func_args here and dropping the lambda, it will wait while find_by() returns a Trueish value.

cfme/tests/services/test_rest_services.py
```diff
@@ -116,121 +137,156 @@ def test_edit_multiple_services(self, rest_api, services):
...
-            * Retrieve list of entities using GET /api/services , pick the first one
-            * POST /api/service/<id> (method ``retire``) with the ``retire_date``
         Metadata:
             test_flag: rest
         """
-        assert "retire" in rest_api.collections.services.action.all
-
-        retire_service = services[0]
-        date = (datetime.datetime.now() + datetime.timedelta(days=5)).strftime('%m/%d/%y')
+        date = (datetime.datetime.now() + datetime.timedelta(days=5)).strftime("%Y/%m/%d")
```
mshriver: **Optional**
Check out utils.parsetime, it has CFME standard formatters for time.

cfme/tests/services/test_rest_services.py
```diff
@@ -116,121 +137,156 @@ def test_edit_multiple_services(self, rest_api, services):
...
-            return False
+        if from_detail:
+            for service in services:
+                service.action.retire(**future)
+                assert rest_api.response.status_code == 200
+        else:
+            rest_api.collections.services.action.retire(*services, **future)
+            assert rest_api.response.status_code == 200
+
+        def _finished(ser):
```
mshriver: **Optional**
Two things:

* use `service` here instead of `ser` (looks like you made this change other places)
* combine the conditionals for returning False with `or`

mshriver:
Mark RFR once you've had a chance to consider the optional comments.

cfme/tests/services/test_rest_services.py
mkoura: Here I'm waiting until it doesn't return Trueish value. In this case I would have to use
```
wait_for(rest_api.collections.services.find_by, func_kwargs={"name": service.name}, fail_condition=True, ...
```
for the same effect, which I don't find more readable.

cfme/tests/services/test_rest_services.py
mkoura: I would need to use datetime anyway for timedelta, so I would rather leave it as it is.

dajoRH:
*CFME QE Bot*

psav:
Conflict needs resolving

mkoura:
@psav resolved

PR #4293, opened 2017-02-28
[1LP][WIPTEST] Widgetastic Slider widget

izapolsk:
Purpose or Intent
=================
one small Slider widget 

I've checked all the methods manually.

### DIFF
--- a/widgetastic_manageiq.py
+++ b/widgetastic_manageiq.py
@@ -1309,3 +1309,46 @@ class AlertEmail(View):
 
     def read(self):
         return list(self.all_emails)
+
+
+class Slider(Widget):
+    """This control represents HTML5 Slider
+
+    """
+    ROOT = ParametrizedLocator('{@locator}')
+
+    def __init__(self, parent, locator, logger=None):
+        Widget.__init__(self, parent, logger=logger)
+        self.locator = locator
+
+    @property
+    def value(self):
+        return float(self.browser.get_attribute('value', self))
+
+    @property
+    def max(self):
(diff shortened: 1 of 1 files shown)


widgetastic_manageiq.py
```diff
@@ -1309,3 +1309,46 @@ def all_emails(self):
...
+    def min(self):
+        return float(self.browser.get_attribute('min', self))
+
+    @property
+    def step(self):
+        return float(self.browser.get_attribute('step', self))
+
+    @value.setter
+    def value(self, val):
+        self.browser.set_attribute('value', val, self)
```
otsuman: **REQUIRED**: This setter doesn't work. It will not change "value" field in GUI. It must be changed

izapolsk: it changes value in gui but javascript doesn't pick up changes for some reason.

otsuman: I checked same control in Bottlenecks page. It works exactly like this one.

widgetastic_manageiq.py
```diff
@@ -1309,3 +1309,46 @@ def all_emails(self):
...
+        self.value += self.step
+
+    def decrease(self):
+        self.value -= self.step
+
+    def read(self):
+        return self.value
+
+    def fill(self, val):
+        self.value = val
```
mfalesni: **REQUIRED**: fill must return true or false.

izapolsk:
It seems this control appears only along with timelines chart. Additionally, it is implemented in strange way.
So, I'm closing this PR and it will come along with timelines view.

PR #4319, opened 2017-03-04
[RFR] Fixed splitter test to use real Server instance

psav:
* The test navigation for Utilization at least, relies on having access
  to the zone attribute which the base class doesn't have


psav:
This is only intended to fix Utilization

PR #4325, opened 2017-03-05
[1LP][RFR]Cockpit button accessibility/cockpit login

jdemon519:
{{pytest: cfme/tests/containers/test_cockpit.py -v --use-provider cm-env1 }}

The test verifies the accessibility of the cockpit (Web Console) button on the master node, then logs in to cockpit. It will not work until the single sign-on bug is fixed.

### DIFF
--- a/cfme/tests/containers/test_cockpit.py
+++ b/cfme/tests/containers/test_cockpit.py
@@ -17,11 +17,12 @@ pytest_generate_tests = testgen.generate(
 
 
 @pytest.mark.polarion('CMP-10255')
-def test_cockpit_button_access(provider):
+def test_cockpit_button_access(provider, soft_assert):
     """ The test verifies the existence of cockpit "Web Console"
         button on each node
 
     """
+
     navigate_to(NodeCollection, 'All')
     tb.select('List View')
 
@@ -31,6 +32,8 @@ def test_cockpit_button_access(provider):
     for name in names:
         obj = Node(name, provider)
         obj.load_details()
-        assert tb.exists(
-            'Open a new browser window with Cockpit for this '
-            'VM.  This requires that Cockpit is pre-configured on the VM.')
+        soft_assert(
+            tb.exists(
+                'Open a new browser window with Cockpit for this '
+                'VM.  This requires that Cockpit is pre-configured on the VM.'),
+            'Cockpit "Web Console" button is not found on details page.'.format(name))
--- a/cfme/tests/containers/test_cockpit.py
+++ b/cfme/tests/containers/test_cockpit.py
@@ -36,4 +36,4 @@ def test_cockpit_button_access(provider, soft_assert):
             tb.exists(
                 'Open a new browser window with Cockpit for this '
                 'VM.  This requires that Cockpit is pre-configured on the VM.'),
-            'Cockpit "Web Console" button is not found on details page.'.format(name))
+            'Cockpit "Web Console" button {} is not found on details page.'.format(name))
--- a/cfme/tests/containers/test_cockpit.py
+++ b/cfme/tests/containers/test_cockpit.py
@@ -1,10 +1,10 @@
 import pytest
 
-from cfme.containers.node import NodeCollection, Node
-from utils import testgen, version
-from cfme.web_ui import CheckboxTable, toolbar as tb
-from utils.appliance.implementations.ui import navigate_to
+from cfme.containers.node import Node, NodeCollection
 from cfme.containers.provider import ContainersProvider
+from cfme.web_ui import CheckboxTable, toolbar as tb
+from utils import testgen, version
+from utils.appliance.implementations.ui import navigate_to
 
 
 pytestmark = [
--- a/cfme/tests/containers/test_cockpit.py
+++ b/cfme/tests/containers/test_cockpit.py
@@ -2,7 +2,7 @@ import pytest
 
 from cfme.containers.node import Node, NodeCollection
 from cfme.containers.provider import ContainersProvider
-from cfme.web_ui import CheckboxTable, toolbar as tb
+from cfme.web_ui import toolbar as tb
 from utils import testgen, version
 from utils.appliance.implementations.ui import navigate_to
 
@@ -23,11 +23,9 @@ def test_cockpit_button_access(provider, soft_assert):
 
     """
 
-    navigate_to(NodeCollection, 'All')
-    tb.select('List View')
+    view = navigate_to(NodeCollection, 'All')
 
-    list_tbl = CheckboxTable(table_locator="//div[@id='list_grid']//table")
-    names = [r.name.text for r in list_tbl.rows()]
+    names = [r.name.text for r in view.nodes]
 
     for name in names:
         obj = Node(name, provider)
--- a/cfme/tests/containers/test_cockpit.py
+++ b/cfme/tests/containers/test_cockpit.py
@@ -15,9 +15,8 @@ pytestmark = [
 pytest_generate_tests = testgen.generate(
     [ContainersProvider], scope='function')
 
-# CMP-10255
 
-
+@pytest.mark.polarion('CMP-10255')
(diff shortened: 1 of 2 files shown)


cfme/tests/containers/test_cockpit_button_access.py
```diff
@@ -0,0 +1,37 @@
...
+    pytest.mark.uncollectif(
+        lambda provider: version.current_version() < "5.6"),
+    pytest.mark.usefixtures('setup_provider'),
+    pytest.mark.tier(1)]
+pytest_generate_tests = testgen.generate(
+    [ContainersProvider], scope='function')
+
+# CMP-10255
+
+
```
gshefer: Please use @pytest.mark.polarion('CMP-10255')

gshefer:
Change file name to test_cockpit.py

jdemon519:
@gshefer done :)

cfme/tests/containers/test_cockpit.py
```diff
@@ -0,0 +1,36 @@
...
+    navigate_to(NodeCollection, 'All')
+    tb.select('List View')
+
+    list_tbl = CheckboxTable(table_locator="//div[@id='list_grid']//table")
+    names = [r.name.text for r in list_tbl.rows()]
+
+    for name in names:
+        obj = Node(name, provider)
+        obj.load_details()
+        assert tb.exists(
```
gshefer: 1. add soft_assert, so you will be able to check all Nodes.
2. Add an informative error message once assertion fail.

I'd do something like:
```
soft_assert( tb.exists('Open a new browser window with Cockpit for this '
                                    'VM.  This requires that Cockpit is pre-configured on the VM.'), 
                   'cockpit "Web Console" button not found in Node "{}" details page, probably Cockpit is not pre-configured?'.format(name))
```

jdemon519: @gshefer, agree will change it shortly.

cfme/tests/containers/test_cockpit.py
```diff
@@ -0,0 +1,39 @@
...
+    names = [r.name.text for r in list_tbl.rows()]
+
+    for name in names:
+        obj = Node(name, provider)
+        obj.load_details()
+        soft_assert(
+            tb.exists(
+                'Open a new browser window with Cockpit for this '
+                'VM.  This requires that Cockpit is pre-configured on the VM.'),
+            'Cockpit "Web Console" button is not found on details page.'.format(name))
```
gshefer: Please add "{}" for the name formatting, else it'll crash.

cfme/tests/containers/test_cockpit.py
```diff
@@ -0,0 +1,39 @@
+import pytest
+
+from cfme.containers.node import NodeCollection, Node
```
psav: **REQUIRED:** Alphabetical ordering is nice

cfme/tests/containers/test_cockpit.py
```diff
@@ -0,0 +1,39 @@
...
+from utils.appliance.implementations.ui import navigate_to
+from cfme.containers.provider import ContainersProvider
+
+
+pytestmark = [
+    pytest.mark.uncollectif(
+        lambda provider: version.current_version() < "5.6"),
+    pytest.mark.usefixtures('setup_provider'),
+    pytest.mark.tier(1)]
+pytest_generate_tests = testgen.generate(
```
psav: **OPTIONAL:** No reason not to put this all on one line

cfme/tests/containers/test_cockpit.py
jdemon519: @psav done

jdemon519:
@psav I alphabetized the imports with flake8 plugin so it should be good to go. Thanks!

cfme/tests/containers/test_cockpit.py
```diff
@@ -0,0 +1,39 @@
...
+
+@pytest.mark.polarion('CMP-10255')
+def test_cockpit_button_access(provider, soft_assert):
+    """ The test verifies the existence of cockpit "Web Console"
+        button on each node
+
+    """
+
+    navigate_to(NodeCollection, 'All')
+    tb.select('List View')
```
mshriver: This table is modeled already in the NodeView, no need to create it again.  This is why we're moving to thoroughly modeling all UI elements as widgets - to access them through the views generated by navigation.

The 'List View' selection is redundant - the 'All' destination currently calls it in resetter, which is called by default.

https://github.com/ManageIQ/integration_tests/blob/master/cfme/containers/node.py#L34

```python
# might need slight tweaks but this is how you would use the view and widgets after navigating
view = navigate_to(NodeCollection, 'All')

names = [r.name.text for r in view.nodes]
```

`view.nodes`, the widgetastic_manageiq.Table class, returns rows() through `__iter__`

jdemon519:
@mshriver I changed the script accordingly, thanks

jdemon519:
@psav @gshefer @mshriver Guys, let me know if there is anything I need to change (the script won't work until the single sign-on issue is resolved). Thanks.

mshriver:
@jdemon519 Your update didn't actually implement my suggestion, you still have the CheckboxTable dependency and aren't using the view returned by navigation.

jdemon519:
@mshriver I implemented it in the previous version, but then it was decided to combine the cockpit login PR and this one, meaning I only need the master node. I tried clicking on the first node(master) but it did not work with the node view. I'll check the implementation of the node view again.

mshriver:
@jdemon519 Noted - Don't hesitate to gather up a screenshot, trace of the failure, and cfme.log and get in touch for some debugging support.

It sounds like what you actually need to do is navigate to the 'Details' destination for a node in the collection.  Generating a list of the names in the table isn't actually needed at all. The NodeCollection class can provide you with a Node object, which you can then call `navigate_to()` with.

mshriver:
Moving it back to WIPTEST for now, we can re-evaluate if using the new widgets is too troublesome at this time.

jdemon519:
@mshriver Here is the error:

        view = navigate_to(NodeCollection, 'All')
    
        names = [r.name.text for r in view.nodes]
>       names[0].row_element.click()
E       AttributeError: 'unicode' object has no attribute 'row_element'

jdemon519:
@psav @mshriver @gshefer Please review. Also, the env1 is the false positive, there is no "web console" button in 5.6

cfme/tests/containers/test_cockpit.py
```diff
@@ -0,0 +1,49 @@
...
+            'VM.  This requires that Cockpit is pre-configured on the VM.'),
+        'No "Web Console" button found')
+    tb.select('Open a new browser window with Cockpit for this '
+              'VM.  This requires that Cockpit is pre-configured on the VM.')
+
+    port_num = ':9090/system'
+    url_cockpit = provider.hostname + port_num
+    sel.get(url_cockpit)
+    title_pg = sel.title()
+    soft_assert(title_pg == 'Cockpit', 'Cockpit main page failed to open')
```
mshriver: Looks like in 5.7 the window opened but isn't correctly identified here. If this is opening a new window or tab I'm not sure this will work as intended in the current framework.

That being said, the test structure looks fine.

cfme/tests/containers/test_cockpit.py
```diff
@@ -0,0 +1,49 @@
...
+        opens up the cockpit main page in a new window. Then
+        we verify the title of the main cockpit page. The test
+        will not work until the single sign-on bug is fixed
+
+    """
+
+    collection = NodeCollection()
+    nodes = collection.all()
+    node = [node for node in nodes if 'master' in node.name][0]
+    navigate_to(node, 'Details')
```
mshriver: :+1:

jdemon519:
Right, it will fail until the bug is fixed, so that's an expected outcome.

mfalesni:
@jistefan you might get aware of this test as you also test cockpit.

PR #4331, opened 2017-03-06
[1LP][RFR] Remove providers through rest & fix provider exists check for renamed providers

jkrocil:
This adds mapping between our classes and MiQ classes so that when we call `clear_providers` to delete for example all `InfraProviders`, it will look for all `InfraManager` providers and remove them. Until now, we only deleted known managed providers (part of this PR is also renaming the method to make its function 100% clear from the name) which is not sufficient when those are renamed. Same with checking with `.exists`.
It also removes `clear_providers_by_class` and keeps only `clear_providers` which becomes a `classmethod` -> the providers removed depend on which class you call it from. `InfraProvider.clear_providers()` will remove all infra providers, `CloudProvider.clear_providers()` will remove all cloud providers...

{{pytest: -k test_provider}}

### DIFF
--- a/cfme/common/provider.py
+++ b/cfme/common/provider.py
@@ -1,5 +1,6 @@
 import datetime
 from functools import partial
+from manageiq_client.api import APIException
 
 import cfme
 import cfme.fixtures.pytest_selenium as sel
@@ -65,6 +66,7 @@ class BaseProvider(Taggable, Updateable, SummaryMixin, Navigatable):
     _properties_region = None
     add_provider_button = None
     save_button = None
+    db_types = ["Providers"]
 
     class Credential(cfme.Credential, Updateable):
         """Provider credentials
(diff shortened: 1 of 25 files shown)


dajoRH:
*CFME QE Bot*

cfme/common/provider.py
```diff
@@ -553,30 +554,22 @@ def process_credential_yaml_key(cls, cred_yaml_key, cred_type=None):
...
-
-    # Move to collection
-    @staticmethod
-    def clear_providers():
-        """ Clear all providers on an appliance using UI """
-        BaseProvider.clear_providers_by_class(BaseProvider)
+    @classmethod
+    def clear_providers(cls):
+        """ Clear all providers of given class on the appliance """
+        from utils.appliance import current_appliance as app
```
izapolsk: OPTIONAL: we arranged to not use current_appliance. is it ok

jkrocil: Yes but I assume we will have appliance available once we turn this into a collection method?

jkrocil: I don't have a better way around this as is. If you can think of something, I'm definitely all ears.

PR #4335, opened 2017-03-06
[1LP][RFR] Add Openstack cloud instances tests

okolisny:
Adding tests for Openstack cloud instances: provision and power control

### DIFF
--- a/cfme/provisioning.py
+++ b/cfme/provisioning.py
@@ -217,6 +217,9 @@ provisioning_form = tabstrip.TabStripForm(
             ('automatic_placement', ui.Input('environment__placement_auto')),
 
             # Cloud
+            ('cloud_tenant', {
+                version.LOWEST: ui.Select('select#environment__cloud_tenant'),
+                '5.5': AngularSelect('environment__cloud_tenant')}),
             ('availability_zone', {
                 version.LOWEST: ui.Select('select#environment__placement_availability_zone'),
                 '5.5': AngularSelect('environment__placement_availability_zone')}),
--- a/cfme/provisioning.py
+++ b/cfme/provisioning.py
@@ -186,29 +186,19 @@ provisioning_form = tabstrip.TabStripForm(
 
         ('Catalog', [
             # Cloud
-            ('num_instances', {
-                version.LOWEST: ui.Select('select#service__number_of_vms'),
-                '5.5': AngularSelect('service__number_of_vms')}),
+            ('num_instances', AngularSelect('service__number_of_vms')),
             ('instance_name', '//input[@name="service__vm_name"]'),
             ('instance_description', ui.Input('service__vm_description')),
 
             # Infra
-            ('vm_filter', {
-                version.LOWEST: ui.Select('select#service__vm_filter'),
-                '5.5': AngularSelect('service__vm_filter')}),
-            ('num_vms', {
-                version.LOWEST: ui.Select('select#service__number_of_vms'),
-                '5.5': AngularSelect('service__number_of_vms')}),
+            ('vm_filter', AngularSelect('service__vm_filter')),
+            ('num_vms', AngularSelect('service__number_of_vms')),
             ('vm_name', '//input[@name="service__vm_name"]'),
             ('vm_description', ui.Input('service__vm_description')),
             ('catalog_name', ui.Table('//div[@id="prov_vm_div"]/table')),
-            ('provision_type', {
-                version.LOWEST: ui.Select('select#service__provision_type'),
-                '5.5': AngularSelect('service__provision_type')}),
+            ('provision_type', AngularSelect('service__provision_type')),
             ('linked_clone', ui.Input('service__linked_clone')),
-            ('pxe_server', {
-                version.LOWEST: ui.Select('select#service__pxe_server_id'),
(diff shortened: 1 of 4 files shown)


cfme/provisioning.py
```diff
@@ -217,6 +217,9 @@ class schedule(Tab):  # noqa
             ('automatic_placement', ui.Input('environment__placement_auto')),
 
             # Cloud
+            ('cloud_tenant', {
+                version.LOWEST: ui.Select('select#environment__cloud_tenant'),
+                '5.5': AngularSelect('environment__cloud_tenant')}),
```
mshriver: **Required**
We don't need to account for anything lower than 5.5, this can just be the AngularSelect.

cfme/tests/openstack/cloud/test_instances.py
```diff
@@ -0,0 +1,176 @@
...
+from utils.appliance.implementations.ui import navigate_to
+from utils.version import current_version
+
+pytest_generate_tests = testgen.generate([OpenStackProvider],
+                                         scope='module')
+
+pytestmark = [pytest.mark.usefixtures("setup_provider_modscope")]
+
+
+@pytest.mark.usefixtures("setup_provider_modscope")
```
mshriver: Why duplicate the mark here and at line 36 when its already on line 17?

cfme/tests/openstack/cloud/test_instances.py
```diff
@@ -0,0 +1,176 @@
...
+@pytest.fixture(scope='function')
+def active_instance(provider):
```
mshriver: Sorry but I don't agree with this fixture at all. Its taking an already running instance, or picking a random one and trying to start it, and then the tests are shutting the instance down (regardless of its initial state) and never recovering the initial state.  What if it picks an instance in use by a test running in parallel? Or a C&U instance that should be on 24x7?

Drop this and provision an instance to test on every time, or use a module instance and change the tests to get the instance in the starting state.

cfme/provisioning.py
okolisny: Deleted variants for versions lower 5.5 all over the from. Does it ok ?

cfme/tests/openstack/cloud/test_instances.py
okolisny: Well, now it creates an instance for each test.

cfme/tests/openstack/cloud/test_instances.py
```diff
@@ -0,0 +1,151 @@
...
+from utils.appliance.implementations.ui import navigate_to
+from utils.version import current_version
+
+pytest_generate_tests = testgen.generate([OpenStackProvider],
+                                         scope='module')
+
+pytestmark = [pytest.mark.usefixtures("setup_provider_modscope")]
+
+
+@pytest.fixture(scope='function')
```
mshriver: much better! thanks for re-writing the fixtures.

cfme/tests/openstack/cloud/test_instances.py
```diff
@@ -0,0 +1,151 @@
...
+             ('Flavor', 'instance_type')]
+
+    if current_version() >= '5.7':
+        props.append(('Virtual Private Cloud', 'cloud_network'))
+
+    for p in props:
+        v = new_instance.get_detail(properties=('Relationships', p[0]))
+        soft_assert(v == prov_data[p[1]])
+
+
```
mshriver: It looks like some of these tests duplicate test coverage that already exists, and add some new coverage that's OpenStack specific.

https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/cloud/test_instance_power_control.py

Consider removing the duplicate coverage, and only cover power control specific to openstack here.  Or use the existing test and add functions that are only collected against openstack provider types.

okolisny: Yes, I saw it. I would like to keep current tests separately, if you do not mind.

mshriver: Okay, that's fine for now, just wanted to make sure you were aware of the existing tests. We're working on solving this at a higher level to reduce duplication across integration_tests and any eventual repo splits.

cfme/tests/openstack/cloud/test_instances.py
```diff
@@ -0,0 +1,151 @@
...
+                                         option=OpenStackInstance.HARD_REBOOT)
+    new_instance.wait_for_instance_state_change(OpenStackInstance.STATE_REBOOTING)
+
+    state = new_instance.get_detail(properties=('Power Management',
+                                                'Power State'))
+    assert state in (OpenStackInstance.STATE_ON,
+                     OpenStackInstance.STATE_REBOOTING)
+
+
+def test_delete_instance(new_instance):
```
mshriver: **Optional**

Consider test_terminate_instance since the power state you're using is `TERMINATE`

okolisny: `TERMINATE ` is just a name of var, which can point on "Delete" or "Terminate" power option on the page, it depends on CF version. I've named test as delete because this is the option which displayed in latest versions.

okolisny: Anyway, it doesn't matter really :)

PR #4338, opened 2017-03-07
[1LP][RFR] Update ec2cleanup, overwrite instead of append log

mshriver:
Don't really need to append to the log file every time, just overwrite it.


izapolsk:
@mshriver , could you please explain what is this change for ?

mshriver:
@izapolsk We were appending to a log in the Jenkins workspace instead of writing a new one, and I don't see much use in continually appending to the log every time this is executed. We've got job history of each log file as well as the emailed reports.

PR #4348, opened 2017-03-09
[1LP][RFR] dockerbot should use frozen.txt instead of requrements.txt

izapolsk:
Purpose or Intent
=================
slightly updated dockerbot to use frozen.txt instead of requrements.txt

### DIFF
--- a/scripts/dockerbot/dockerbot.py
+++ b/scripts/dockerbot/dockerbot.py
@@ -240,7 +240,7 @@ class DockerBot(object):
                             if filen['filename'].startswith('cfme/tests') or \
                                filen['filename'].startswith('utils/tests'):
                                 files.append(filen['filename'])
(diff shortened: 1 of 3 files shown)


scripts/dockerbot/dockerbot.py
```diff
@@ -240,7 +240,7 @@ def find_files_by_pr(self, pr=None):
                             if filen['filename'].startswith('cfme/tests') or \
                                filen['filename'].startswith('utils/tests'):
                                 files.append(filen['filename'])
-                        if filen['filename'] == 'requirements.txt':
+                        if filen['filename'] == 'frozen.txt':
```
psav: we may need a path in here

izapolsk: agree, updated.

PR #4350, opened 2017-03-09
[1LP][RFR] new tests test_ec2_manual_placement_using_rest

mkoura:
New test that selects particular VPC while creating VM in AWS through REST API.

{{pytest: -v -k test_ec2_manual_placement_using_rest --use-provider ec2west}}

### DIFF
--- a/cfme/tests/cloud/test_provisioning.py
+++ b/cfme/tests/cloud/test_provisioning.py
@@ -16,7 +16,6 @@ from cfme.cloud.instance.gce import GCEInstance  # NOQA
 from cfme.cloud.provider import CloudProvider
 from cfme.cloud.provider.azure import AzureProvider
 from cfme.cloud.provider.gce import GCEProvider
-from cfme.cloud.provider.ec2 import EC2Provider
 from cfme.cloud.provider.openstack import OpenStackProvider
 from utils import testgen
 from utils.generators import random_vm_name
@@ -208,8 +207,6 @@ def test_provision_from_template_using_rest(
         num_sec=600, delay=5, message="VM {} becomes visible".format(vm_name))
 
 
-@pytest.mark.uncollectif(lambda provider:
-        not provider.one_of(EC2Provider) or current_version() < "5.7")
 def test_ec2_manual_placement_using_rest(
         request, setup_provider, provider, vm_name, rest_api, provisioning):
     """ Tests provisioning ec2 instance with manual placement using the REST API.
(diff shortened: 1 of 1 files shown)


cfme/tests/cloud/test_provisioning.py
```diff
@@ -207,6 +207,107 @@ def _finished():
         num_sec=600, delay=5, message="VM {} becomes visible".format(vm_name))
 
 
+@pytest.mark.uncollectif(lambda provider: provider.type != 'ec2' or current_version() < "5.7")
```
izapolsk: REQUIRED: provider.type is deprecated please use one_of method instead.
"not provider.one_of(EC2Provider)"

cfme/tests/cloud/test_provisioning.py
```diff
@@ -207,6 +207,107 @@ def _finished():
...
+    Metadata:
+        test_flag: provision
+    """
+    image_guid = rest_api.collections.templates.get(name=provisioning['image']['name']).guid
+    provider_id = rest_api.collections.providers.get(name=provider.name).id
+    instance_type = provisioning['instance_type'].split(':')[0].strip()
+
+    flavors = rest_api.collections.flavors.find_by(name=instance_type)
+    assert len(flavors) > 0
+    flavor = None
```
izapolsk: OPTIONAL: do you mind me asking what is it for ?

cfme/tests/cloud/test_provisioning.py
```diff
@@ -207,6 +207,107 @@ def _finished():
...
+    for flavor in flavors:
+        if flavor.ems_id == provider_id:
+            break
+    else:
+        pytest.fail("Cannot find flavour.")
+
+    cloud_networks = rest_api.collections.cloud_networks.find_by(
+        type='ManageIQ::Providers::Amazon::NetworkManager::CloudNetwork')
+    assert len(cloud_networks) > 0
+    cloud_network = None
```
izapolsk: OPTIONAL: ^^

cfme/tests/cloud/test_provisioning.py
```diff
@@ -207,6 +207,107 @@ def _finished():
...
+
+    request.addfinalizer(
+        lambda: provider.mgmt.delete_vm(vm_name) if provider.mgmt.does_vm_exist(vm_name) else None)
+
+    request = rest_api.collections.provision_requests.action.create(**provision_data)[0]
+    assert rest_api.response.status_code == 200
+
+    def _finished():
+        request.reload()
+        if request.status.lower() in {"error"}:
```
izapolsk: OPTIONAL: I'm wondering what is the reason for using set here.  Could you please explain ?

cfme/tests/cloud/test_provisioning.py
mkoura: I'm using the loop variable later in the code. This assignment is there to make it explicit that the loop variable is defined even in case when there's nothing to iterate over. It's not necessary at the moment as there's the assert above and else in the for loop, but pylint still insists. So at the moment it's there just to make pylint happy (otherwise it reports ``[undefined-loop-variable] Using possibly undefined loop variable 'flavor'``).

cfme/tests/cloud/test_provisioning.py
mkoura: No particular reason, the check is taken verbatim from the `test_provision_from_template_using_rest` above. Now looking at it, it would be better with just ``if "error" in request.status.lower():``. Will change that.

cfme/tests/cloud/test_provisioning.py
```diff
@@ -207,6 +207,112 @@ def _finished():
...
+    assert len(flavors) > 0
+    flavor = None
+    for flavor in flavors:
+        if flavor.ems_id == provider_rest.id:
+            break
+    else:
+        pytest.fail("Cannot find flavour.")
+
+    provider_data = rest_api.get(provider_rest._href +
+        '?attributes=cloud_networks,cloud_subnets,security_groups')
```
mfalesni: the rest api client does not have the means of passing attributes?

mkoura: I could have used ``rest_api.get_entity(rest_api.collections.providers, provider_rest.id, attributes="...")``, but I wanted to get just unprocessed dictionary. I had a mysterious issue with missing keys and wanted to be absolutely sure it's not stripped during result processing in the api client.

PR #4354, opened 2017-03-10
[WIP]extending button's locators for upcomming 5.8

lkhomenk:
our framework was failing on adding providers - buttons were updated there

### DIFF
--- a/cfme/web_ui/form_buttons.py
+++ b/cfme/web_ui/form_buttons.py
@@ -28,13 +28,15 @@ class FormButton(Pretty):
 
     class Button:
         """Holds pieces of the XPath to be assembled."""
-        TAG_TYPES = "//a | //button | //img | //input"
+        TAG_TYPES = "//a | //button | //img | //input | //miq-button"
         TYPE_CONDITION = (
-            "(contains(@class, 'button') or contains(@class, 'btn') or contains(@src, 'button'))"
+            "(contains(@class, 'button') or contains(@class, 'btn') or contains(@src, 'button') "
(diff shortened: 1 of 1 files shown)


lkhomenk:
this is for 5.8 but it doesn't break our current automation

cfme/web_ui/form_buttons.py
```diff
@@ -28,13 +28,15 @@ class FormButton(Pretty):
...
         TYPE_CONDITION = (
-            "(contains(@class, 'button') or contains(@class, 'btn') or contains(@src, 'button'))"
+            "(contains(@class, 'button') or contains(@class, 'btn') or contains(@src, 'button') "
+            "or contains(@class,'ng-isolate-scope'))"
         )
         DIMMED = "(contains(@class, 'dimmed') " \
             "or contains(@class, 'disabled') " \
-            "or contains(@class, 'btn-disabled'))"
+            "or contains(@class, 'btn-disabled')" \
+            "or contains(@class, 'ng-isolate-scope ng-hide'))"
```
jkrocil: This is not correct - at least not for Infra Provider Add page. `.is_dimmed` check doesn't work because of that.

mkoura:
The old button is still there, so why trying to use the miq-button? The original button is now missing the alt text, but text() can be used instead.
I tried to fix it without much testing, just to be able to run my tests - https://github.com/ManageIQ/integration_tests/pull/4360

jkrocil:
Superseded by #4360, closing.

PR #4355, opened 2017-03-10
[1LP][RFR] Removed older option and simplified codebase

psav:
Removing an older option that is no longer present in 5.6+
Removed get_yaml_file() as it is no longer used

### DIFF
--- a/utils/appliance/__init__.py
+++ b/utils/appliance/__init__.py
@@ -1687,9 +1687,9 @@ class IPAppliance(object):
             ssh.run_command('ldconfig')
 
             log_callback('Modifying YAML configuration')
-            c_yaml = self.get_yaml_config('vmdb')
+            c_yaml = self.get_yaml_config()
             c_yaml['product']['storage'] = True
-            self.set_yaml_config('vmdb', c_yaml)
+            self.set_yaml_config(c_yaml)
 
             # To mark that we installed netapp
             ssh.run_command("touch /var/www/miq/vmdb/HAS_NETAPP")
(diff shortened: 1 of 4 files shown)


utils/appliance/__init__.py
```diff
@@ -1931,64 +1927,48 @@ def host_id(self, hostname):
...
+        temp_ruby = load_data_file(ruby_template.strpath, ruby_replacements)
+        self.ssh_client.put_file(temp_ruby.name, dest_ruby)
+
+        # Run it
+        if self.ssh_client.run_rails_command(dest_ruby):
+            self.server_details_changed()
         else:
-            # 2 = errno.ENOENT
-            raise IOError(2, 'Remote yaml not found or permission denied')
+            raise Exception('Unable to set config')
```
izapolsk: OPTIONAL: "ApplianceException" ?

utils/appliance/__init__.py
```diff
@@ -1931,64 +1927,48 @@ def host_id(self, hostname):
...
+    def get_yaml_config(self):
+        writeout = self.ssh_client.run_rails_command(
+            '"File.open(\'/tmp/yam_dump.yaml\', \'w\') '
+            '{|f| f.write(Settings.to_hash.deep_stringify_keys.to_yaml) }"'
+        )
+        if writeout.rc:
+            logger.error("Config couldn't be found")
+            logger.error(writeout.output)
+            raise Exception('Error obtaining config')
+        base_data = self.ssh_client.run_command('cat /tmp/yam_dump.yaml')
```
izapolsk: OPTIONAL: we have the number of commands which we run over ssh everywhere. I'd create an SSHAgent with pre-defined commands

utils/appliance/__init__.py
psav: I'll leave as is as for now there may be something relying on this exception

utils/appliance/__init__.py
psav: I think we could, not sure if we'd just end up with a bunch of mishmash commands. It doesn't seem too horrible here to keep this......but I'll definitely consider it

PR #4357, opened 2017-03-13
[RFR][1LP] Optimize renew

psav:
* Renew flow is optimized
* Renew timer starts immediately and doesn't try
 * Means no errors at start of test runs

### DIFF
--- a/utils/browser.py
+++ b/utils/browser.py
@@ -231,20 +231,21 @@ class BrowserKeepAliveThread(threading.Thread):
 
     def run(self):
         while not self.stopped():
+            time.sleep(THIRTY_SECONDS)
             with self.manager.factory.lock:
-                try:
-                    log.debug('renew')
-                    self.manager.browser.current_url
-                except Exception as e:
-                    log.error('something bad happened')
-                    log.error(e)
+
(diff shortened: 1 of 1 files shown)


utils/browser.py
```diff
@@ -231,20 +231,16 @@ def __init__(self, manager):
...
         while not self.stopped():
+            time.sleep(THIRTY_SECONDS)
             with self.manager.factory.lock:
-                try:
-                    log.debug('renew')
-                    self.manager.browser.current_url
-                except Exception as e:
-                    log.error('something bad happened')
-                    log.error(e)
+                for _ in range(2):
```
izapolsk: OPTIONAL: please add a comment what insists you to check state 3 times ?

psav: I actually JUST added this and it's 2 times, not 3 :)

izapolsk: @psav, range(2) == [0,1,2] :)

psav:
```
In [1]: range(2)
Out[1]: [0, 1]
```

izapolsk: yeah, forgot that it uses strict more. ignore me

PR #4281, opened 2017-02-27
[1LP][RFR]Catalog Item Widgetastic Conversion

sshveta:

PR #4367, opened 2017-03-14
[1LP][RFR] BootstrapSelect: Can ignore NoSuchElementException when closing (select hided)

mfalesni:
This will resolve the issue of #4281


mfalesni:
Rebased.

psav:
Why do we try to close one if it has already disappeared?

mfalesni:
@psav It does not know that it disappeared or not. So I have put the responsibility to close, the close method checks it for itself and does nothing where applicable

PR #4383, opened 2017-03-15
[1LP][RFR] Patternfly ViewChangeButton widget for 56z

mshriver:
The view change buttons in cfme 56z don't have the proper patternfly markup. Add  Version picking to widgetastic_manageIQ's ItemsToolbarViewSelector

### DIFF
--- a/widgetastic_manageiq.py
+++ b/widgetastic_manageiq.py
@@ -27,7 +27,8 @@ from widgetastic.widget import (
 from widgetastic.utils import ParametrizedLocator, Parameter, attributize_string
 from widgetastic.xpath import quote
 from widgetastic_patternfly import (
-    Accordion as PFAccordion, CandidateNotFound, BootstrapTreeview, Button, Input, BootstrapSelect)
+    Accordion as PFAccordion, CandidateNotFound, BootstrapTreeview, Button, ViewChangeButton, Input,
+    BootstrapSelect)
 from cached_property import cached_property
 
 
@@ -1205,9 +1206,15 @@ class ItemsToolBarViewSelector(View):
         view_selector.selected
     """
     ROOT = './/div[contains(@class, "toolbar-pf-view-selector")]'
-    grid_button = Button(title='Grid View')
-    tile_button = Button(title='Tile View')
-    list_button = Button(title='List View')
+    grid_button = VersionPick({
+        Version.lowest(): ViewChangeButton(title='Grid View'),
+        '5.7.0': Button(title='Grid View')})
+    tile_button = VersionPick({
+        Version.lowest(): ViewChangeButton(title='Tile View'),
+        '5.7.0': Button(title='Tile View')})
+    list_button = VersionPick({
+        Version.lowest(): ViewChangeButton(title='List View'),
+        '5.7.0': Button(title='List View')})
 
     @property
     def _view_buttons(self):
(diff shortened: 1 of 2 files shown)


widgetastic_patternfly.py
```diff
@@ -93,6 +93,35 @@ def title(self):
         return self.browser.get_attribute('title', self)
 
 
+class ViewChangeButton(Widget, ClickableMixin):
+    """A PatternFly/Bootstrap view selection button in CFME 56z
+
+        .. code-block:: python
+
+            Button('Text of button (unless it is an input ...)')
```
mshriver: copypasta, need to clean up.

widgetastic_manageiq.py
```diff
@@ -1205,15 +1206,15 @@ class ItemsToolBarViewSelector(View):
         view_selector.selected
     """
     ROOT = './/div[contains(@class, "toolbar-pf-view-selector")]'
-    grid_button = Button(title='Grid View')
```
izapolsk: REQUIRED: there is one more view with such buttons - DetailsToolBarViewSelector. It waits for your attention as well.

widgetastic_manageiq.py
```diff
@@ -1205,15 +1206,15 @@ class ItemsToolBarViewSelector(View):
...
-    tile_button = Button(title='Tile View')
-    list_button = Button(title='List View')
+    btn_class = VersionPick({
```
izapolsk: OPTIONAL: wouldn't it be better to make one class ViewChangeButton which supports both 5.6 and 5.7 ?

abra kadabra summoning @mfalesni. What's your opinion ?

widgetastic_patternfly.py
```diff
@@ -93,6 +93,28 @@ def title(self):
         return self.browser.get_attribute('title', self)
 
 
+class ViewChangeButton(Widget, ClickableMixin):
+    """A PatternFly/Bootstrap view selection button in CFME 56z
+
+        .. code-block:: python
+
+            Button(title='Show xyz')
```
izapolsk: REQUIRED: please update description

widgetastic_patternfly.py
```diff
@@ -93,6 +93,28 @@ def title(self):
...
+            assert button.active
+    """
+    CHECK_VISIBILITY = True
+
+    def __init__(self, parent, title, **kwargs):
+        Widget.__init__(self, parent, logger=kwargs.pop('logger', None))
+        self.title = title
+
+    def __locator__(self):
+        return './/a[(@title={})]/i[contains(@class, "fa")]/..'.format(quote(self.title))
```
izapolsk: OPTIONAL: I'd put "i" inside "a" in this case and get rid of ..

widgetastic_patternfly.py
```diff
@@ -93,6 +93,28 @@ def title(self):
...
+
+    @property
+    def active(self):
+        return 'active' in self.browser.classes(self.parent)
```
izapolsk: REQUIRED: Are you sure you need to check parent here ?

widgetastic_manageiq.py
mfalesni: It can be done like with ManageIQTree ... btw I would avoid the btn_class here and just place the verpicked buttons on the class.

mshriver: @izapolsk the Button class works for 5.7, I would be version picking all of the element interaction in the class that works for both - that was my first implementation but @mfalesni objected to any version picking in the widgetastic_patternfly module.

widgetastic_patternfly.py
mshriver: hm yes that would be better xpath

widgetastic_patternfly.py
mshriver: Yep, the anchor element doesn't have the active class, the anchors are in a list and the <li> parent element holds the 'active' class.

widgetastic_manageiq.py
mshriver: I wasn't seeing these on the infra providers, but you're right the containers provider details page has them so I'll have to implement there as well.

mshriver:
Updated with comments/discussion and re-tagged reviewers.

widgetastic_manageiq.py
mshriver: (done)

PR #4385, opened 2017-03-15
[1LP][RFR] better way to cleanup custom attributes

mkoura:
Changing the way how custom attributes are deleted after test is finished. Now it's consistent with how it's done for entities in other collections.

{{pytest: cfme/tests/infrastructure/test_providers.py -v -k TestProvidersRESTAPI}}

### DIFF
--- a/cfme/tests/infrastructure/test_providers.py
+++ b/cfme/tests/infrastructure/test_providers.py
@@ -1,10 +1,8 @@
 # -*- coding: utf-8 -*-
-import fauxfactory
 import uuid
+import fauxfactory
 
 import pytest
-
-from manageiq_client.api import APIException
 
 import cfme.web_ui.flash as flash
 import utils.error as error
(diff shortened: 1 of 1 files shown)


cfme/tests/infrastructure/test_providers.py
```diff
@@ -256,15 +253,11 @@ def custom_attributes(self, rest_api, infra_provider):
...
-                provider.custom_attributes.action.delete(attr)
-            except APIException:
-                # custom attributes can be deleted by tests, just log warning
-                log_warn = True
-        if log_warn:
-            logger.warning("Failed to delete custom attribute.")
+        provider.custom_attributes.reload()
+        ids = [attr.id for attr in attrs]
+        delete_attrs = [attr for attr in provider.custom_attributes if attr.id in ids]
+        if len(delete_attrs) != 0:
```
jkrocil: PEP8 "For sequences, (strings, lists, tuples), use the fact that empty sequences are false."

```
if delete_attrs:
    ...
```

PR #4389, opened 2017-03-16
[RFR] Pass args and kwargs in SummaryTable init

mshriver:
Found that when using SummaryTable for cloud.instances widgetastic refactor, and exception was raised because of logger being passed as a kwarg.


mfalesni:
@mfalesni is sorry

PR #4396, opened 2017-03-17
[1LP][RFR] new REST API test for resources hiding

mkoura:
Adding new REST API test for resources hiding.

{{pytest: -v -k test_resources_hiding}}

### DIFF
--- a/cfme/tests/test_rest.py
+++ b/cfme/tests/test_rest.py
@@ -173,6 +173,23 @@ def test_user_settings(rest_api):
     assert isinstance(rest_api.settings, dict)
 
 
+@pytest.mark.uncollectif(lambda: current_version() < '5.8')
+def test_resources_hiding(rest_api):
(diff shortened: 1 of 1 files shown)


cfme/tests/test_rest.py
```diff
@@ -173,6 +173,23 @@ def test_user_settings(rest_api):
...
+    """Test that it's possible to hide resources in response.
+
+    Metadata:
+        test_flag: rest
+    """
+    roles = rest_api.collections.roles
+    resources_visible = rest_api.get(roles._href + '?filter[]=read_only=true')
+    assert rest_api.response.status_code == 200
+    assert 'resources' in resources_visible
+    resources_hidden = rest_api.get(roles._href + '?filter[]=read_only=true&hide=resources')
```
jkrocil: @mkoura It might not be a bad idea to add this functionality to the manageiq rest_api package instead of doing it like this.

PR #4400, opened 2017-03-20
[1LP][RFR] Fixes

gshefer:
{{pytest: cfme/tests/containers/test_search_bar.py -v --use-provider cm-env1 }}

### DIFF
--- a/cfme/tests/containers/test_search_bar.py
+++ b/cfme/tests/containers/test_search_bar.py
@@ -73,4 +73,6 @@ def test_search_bar(provider, soft_assert):
                         'Should not find records, search results: "{}"'
                         .format(search_string, results_row_names))
         finally:
-            search.ensure_no_filter_applied()
+            # search.ensure_no_filter_applied() -> TimedOutError
+            # https://github.com/ManageIQ/integration_tests/issues/4401
+            search.normal_search("")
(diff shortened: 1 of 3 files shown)


cfme/tests/containers/test_search_bar.py
```diff
@@ -73,4 +73,6 @@ def test_search_bar(provider, soft_assert):
                         'Should not find records, search results: "{}"'
                         .format(search_string, results_row_names))
         finally:
-            search.ensure_no_filter_applied()
+            # search.ensure_no_filter_applied() -> "TimedOutError: Could not do Waiting
+            # for advanced search to open at...", so better for this test to cleanup by:
+            search.normal_search("")
```
mfalesni: Hmm, I think we have a function to click the "default" link which should reset all filtering, or is it not applicable here?

mfalesni:
@gshefer One comment, otherwise I am willing to ack it :)

cfme/tests/containers/test_search_bar.py
gshefer: The right way is to fix this issue but I really have to merge something work first, is this worth to open an (framework) issue for that? (if so, I'll open and change the comment to point on that):
 search.ensure_no_filter_applied() -> "TimedOutError: Could not do Waiting for advanced search to open at....
Anyway it's not applicable right now.

PR #4402, opened 2017-03-20
[1LP][RFR] add some additional assert to tests

okolisny:
Resolving some old TODOs


jkrocil:
The PR didn't run on PRT because it is not signed.

PR #4404, opened 2017-03-20
[1LP][RFR] update test_query_simple_collections with collections new in 5.8

mkoura:
The test_query_simple_collections updated with collections added in 5.8.

{{pytest: -v -k test_query_simple_collections}}

### DIFF
--- a/cfme/tests/test_rest.py
+++ b/cfme/tests/test_rest.py
@@ -65,14 +65,14 @@ def test_vm_scan(rest_api, vm, from_detail):
         return response.task.state.lower() == 'finished'
 
 
-COLLECTIONS_IGNORED_56 = {
+COLLECTIONS_ADDED_IN_57 = {
     "arbitration_profiles", "arbitration_rules", "arbitration_settings", "automate",
     "automate_domains", "blueprints", "cloud_networks", "container_deployments", "currencies",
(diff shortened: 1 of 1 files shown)


cfme/tests/test_rest.py
```diff
@@ -72,22 +72,36 @@ def _finished():
...
+     "service_orders", "service_requests", "service_templates", "services",
+     "tags", "tasks", "templates", "tenants", "users", "virtual_templates",
+     "vms", "zones"])
 @pytest.mark.uncollectif(
-    lambda collection_name: (
-        collection_name in COLLECTIONS_IGNORED_56 and current_version() < "5.7"))
+    lambda collection_name:
+        (collection_name in COLLECTIONS_IGNORED_56 and current_version() < "5.7") or
+        (collection_name in COLLECTIONS_IGNORED_57 and current_version() < "5.8")
+)
```
jkrocil: The naming here is misleading because `COLLECTIONS_IGNORED_57` for example doesn't really contain collections ignored only in 5.7. If current_version is `5.6` and collection name can be found only in `COLLECTIONS_IGNORED_57`, it will still uncollect.

**REQUIRED:**
They should really be called `COLLECTIONS_ADDED_IN_57` and `COLLECTIONS_ADDED_IN_58` to describe the content correctly. Then it will make sense.

PR #4411, opened 2017-03-21
[WIPTEST] ManageIQForm Prototype to reduce button re-definition

mshriver:
Provide a generic ManageIQForm view that defines the basic form buttons. To be inherited/composited into form views to prevent defining these form buttons hundreds of times across the page models.

### DIFF
--- a/widgetastic_manageiq.py
+++ b/widgetastic_manageiq.py
@@ -1589,3 +1589,19 @@ class TimelinesView(View):
     @property
     def is_displayed(self):
         return self.title.text == 'Timelines'
+
+
+class ManageIQForm(View):
+    """Generic form view with common action buttons
+
+    .. code-block:: python
+
+        class PageWithFormView(View):
+            @View.nested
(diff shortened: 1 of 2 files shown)


widgetastic_manageiq.py
```diff
@@ -1589,3 +1589,19 @@ class chart(TimelinesChart):  # NOQA
...
+    """Generic form view with common action buttons
+
+    .. code-block:: python
+
+        class PageWithFormView(View):
+            @View.nested
+            class form(ManageIQForm):
+                # widgets
+    """
+    continue_button = Button('Continue')
```
mshriver: This creates widgets that might not exist on the form inheriting it, big downside if we don't want exceptions from trying to access buttons that aren't on the page.

dajoRH:
*CFME QE Bot*

mshriver:
Closing, lack of interest and decisions to declare buttons on every form so that we can assert all the widgets are displayed.

PR #4418, opened 2017-03-21
[1LP][RFR] Ansible upstream fix

sshveta:
Changes made for ansible menu changes in 5.8

### DIFF
--- a/cfme/infrastructure/config_management.py
+++ b/cfme/infrastructure/config_management.py
@@ -82,8 +82,9 @@ class ConfigManager(Updateable, Pretty, Navigatable):
         self.key = key or name
 
     def _form_mapping(self, create=None, **kwargs):
+        provider_type = None if self.appliance.version >= '5.8' else (create and self.type)
         return {'name_text': kwargs.get('name'),
-                'type_select': create and self.type,
+                'type_select': provider_type,
                 'url_text': kwargs.get('url'),
                 'ssl_checkbox': kwargs.get('ssl')}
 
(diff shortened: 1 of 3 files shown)


mshriver:
@sshveta The PR title is 'upstream fix' but these tests are uncollected in upstream. Did you mean to remove the decorator ignoring upstream?  They don't execute in 5.8 on PRT either because of the forced BZ.

cfme/infrastructure/config_management.py
```diff
@@ -82,8 +82,12 @@ def __init__(self, name=None, url=None, ssl=None, credentials=None, key=None, ap
         self.key = key or name
 
     def _form_mapping(self, create=None, **kwargs):
+        if self.appliance.version < '5.8':
```
mshriver: **Optional**
Consider 1-lining this:

```python
# Parenthesis for clarity, lint might complain about them being redundant.
provider_type = None if self.appliance.version >= '5.8' else (create and self.type)
```

cfme/infrastructure/config_management.py
```diff
@@ -441,17 +445,26 @@ class MgrAll(CFMENavigateStep):
...
 
     def step(self):
-        if self.obj.appliance.version > '5.7.0.8':
-            self.prerequisite_view.navigation.select('Configuration', 'Management')
-        else:
+        if self.obj.appliance.version < '5.7':
             self.prerequisite_view.navigation.select('Configuration', 'Configuration Management')
+        if self.obj.appliance.version >= '5.8':
+            self.prerequisite_view.navigation.select('Automation', 'Ansible Tower', 'Explorer')
+        if self.obj.appliance.version > '5.7' and self.obj.appliance.version < '5.8':
```
mshriver: **Optional**
Tempted to say required but I don't think we have anything for this in our coding standards.  I think using `elif` should be a required change here though.
Could you switch the order here so its:

```python
if x < 5.7:
elseif x >= 5.7 and < 5.8
elseif x >= 5.8

File: cfme/services/catalogs/catalog_item.py [hunk 3]
```diff
@@ -194,10 +194,10 @@ def create(self):
         view = navigate_to(self, 'Add')
         # Need to provide the (optional) provider name to the form, not the object
         provider_name = None
-        provider_required_types = ['AnsibleTower', 'Orchestration']
-        if self.item_type in provider_required_types \
-                or self.provider_type in provider_required_types:
+        if self.item_type == 'Orchestration':
             provider_name = self.provider.name
+        if self.item_type == 'AnsibleTower':
```
mshriver: **Required**
`elif` and change the variable name here since its not just a string anymore. maybe `provider_formvalue`

mshriver:
@sshveta this looks good, just some minor tweaks. Ping me when you've updated and it should be a quick 1LP

sshveta:
@mshriver : Yes this PR is include the menu changes for Ansible in upstream and 5.8 .
But as of now they are skipped/ignored because of a bug while adding provider. WIll remove the blocker once it is fixed.

Will implement other comments :+1:

sshveta:
The failures for ansible_tower_243 is coming from provider where data is not refreshed from provider .
@pakotvan (Pavol) is looking into the issue

PR #4426, opened 2017-03-22
[1LP][RFR] Update mgmtsystem in requirements/frozen.txt

mshriver:
First time updating in frozen.txt instead of requirements.txt

Pick up @jteehan fix

Outside of release process, but is preventing CI testing on scvmm providers for CFME QE.


mshriver:
Going to let @jteehan own this in #4427

PR #4432, opened 2017-03-23
[RFR] Add class filter support in the old buttons

### LINKED ISSUE #4409: In chargeback rate assignment - 'Add' button

mfalesni:
Fixes #4409 (again)

Usage should be the same like with the new buttons, I olnly added the PRIMARY class there, add another when needed ...

### DIFF
--- a/cfme/web_ui/form_buttons.py
+++ b/cfme/web_ui/form_buttons.py
@@ -6,6 +6,8 @@ You can use it also for the other buttons with same shape like those CRUD ones.
 from selenium.common.exceptions import NoSuchElementException
 from xml.sax.saxutils import quoteattr
 
+from widgetastic.xpath import quote
+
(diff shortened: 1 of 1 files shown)


gshefer:
On the way, Can you please change line 147 to:
`add = FormButton("Add", classes=[FormButton.PRIMARY])`

mfalesni:
@gshefer I am not entirely sure that every use of this `add` button is against btn-primary. I would hope so, but I would prefer creating a separate object for that ...

PR #4433, opened 2017-03-23
[1LP][RFR] Deferred version pick for provider condition string

quarckster:
{{pytest: -v -k 'test_condition_crud' --long-running}}

Purpose or Intent
=================

Fixing strings in conditions.

### DIFF
--- a/cfme/control/explorer/conditions.py
+++ b/cfme/control/explorer/conditions.py
@@ -304,6 +304,9 @@ class ContainerImageCondition(BaseCondition):
 
 class ProviderCondition(BaseCondition):
 
-    TREE_NODE = "Container Provider"
+    TREE_NODE = deferred_verpick({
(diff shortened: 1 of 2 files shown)


cfme/control/explorer/conditions.py
```diff
@@ -304,6 +304,9 @@ class ContainerImageCondition(BaseCondition):
 
 class ProviderCondition(BaseCondition):
 
-    TREE_NODE = "Container Provider"
+    TREE_NODE = deferred_verpick({
+        version.LOWEST: "Container Provider",
+        '5.7.2': "Provider",
+    })
```
mfalesni: Is it the same case for 5.8 too?

quarckster: Yes, for 5.8 it's just a "Provider" instead of "Container Provider"

PR #4436, opened 2017-03-23
[RFR]Ec2cleanup - Added ELB & ENI cleanup

mmojzis:
Waiting for merge in wrapanapi:
https://github.com/ManageIQ/wrapanapi/pull/118

### DIFF
--- a/scripts/ec2cleanup.py
+++ b/scripts/ec2cleanup.py
@@ -16,8 +16,9 @@ def parse_cmd_line():
                         help='List of EIPs, which should be '
                              'excluded. Allocation_id or public IP are allowed.')
     parser.add_argument('--exclude-elbs', nargs='+',
-                        help='List of ELBs, which should be '
-                             'excluded.')
+                        help='List of ELBs, which should be excluded.')
+    parser.add_argument('--exclude-enis', nargs='+',
+                        help='List of ENIs, which should be excluded. ENI ID is allowed.')
     parser.add_argument("--output", dest="output", help="target file name, default "
                                                         "'cleanup_ec2.log' in utils.path.log_path",
                         default=log_path.join('cleanup_ec2.log').strpath)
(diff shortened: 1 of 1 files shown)


mshriver:
marked [WIP] instead of [NOTEST]

mshriver:
Blocked by PR #4471, including mgmtsystem 1.5.0 with wrapanapi PR 118

scripts/ec2cleanup.py
```diff
@@ -75,23 +79,76 @@ def delete_unattached_volumes(provider_mgmt, excluded_volumes, output):
...
+                provider_mgmt.delete_loadbalancer(loadbalancer=elb)
+        print "  Deleted Elastic LoadBalancers: {}".format(elb_list)  # noqa
+        with open(output, 'a+') as report:
+            if elb_list:
+                # tabulate volume_list and write it
+                report.write(tabulate(tabular_data=elb_list,
+                                      headers=['Provider Key', 'ELB name'],
+                                      tablefmt='orgtbl'))
+    except Exception as e:
+        logger.error(e)
```
mshriver: **Required**
Let's print something here as well as using logger, so its apparent in the job output that there was an issue. Otherwise its a 'silent' failure as the job output and the script log (not the same as logger here) don't have any failure messages.

I'm going to create an issue to actually remove the logger use in this, and instead print everything so that its quickly and easily accessible when reviewing the job in Jenkins.

mshriver:
Tested locally against resources @mmojzis setup

scripts/ec2cleanup.py
```diff
@@ -72,26 +76,79 @@ def delete_unattached_volumes(provider_mgmt, excluded_volumes, output):
...
 
 
-def ec2cleanup(exclude_volumes, exclude_eips, output):
+def delete_unused_loadbalancers(provider_mgmt, excluded_elbs, output):
+    elb_list = []
+    provider_name = provider_mgmt.kwargs['name']
+    try:
+        for elb in provider_mgmt.get_all_unused_loadbalancers():
+            if excluded_elbs and elb.name in excluded_elbs:
+                print "  Excluding Elastic LoadBalancer id: {}".format(elb.name)  # noqa
```
mshriver: @psav raised a point offline - these should be py3 compatible print statements.

At least update the modified ones here, if you'd like you can update all the prints in the script.

PR #4442, opened 2017-03-24
[1LP][RFR]Fixed timeprofile tests for 5.8 version

mmojzis:
Add button was changed with Save button in 5.8 so this is fix for it.

### DIFF
--- a/cfme/tests/configure/test_timeprofile.py
+++ b/cfme/tests/configure/test_timeprofile.py
@@ -3,7 +3,6 @@ import fauxfactory
 import cfme.configure.settings as st
 import pytest
 import utils.error as error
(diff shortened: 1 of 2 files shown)


cfme/tests/configure/test_timeprofile.py
```diff
@@ -65,7 +64,10 @@ def test_days_required_error_validation():
         tp.create(cancel=True)
         assert "At least one day needs to be selected" == \
                tp.timeprofile_form.days.angular_help_block
-        assert form_buttons.add.is_dimmed
+        if version.current_version() > "5.8":
```
izapolsk: please use >= in such cases. it looks really confusing.

PR #4449, opened 2017-03-27
[1LP][RFR] RHEV template upload: Slightly more verbose; Exceptions with tracebacks

mfalesni:


scripts/template_upload_rhevm.py
```diff
@@ -66,11 +67,10 @@ def is_ovirt_engine_running(rhevm_ip, sshname, sshpass):
...
         stdout = ssh_client.run_command('service ovirt-engine status')[1]
         ssh_client.close()
-        if 'running' not in stdout:
-            return False
-        return True
-    except Exception as e:
-        print(e)
+        return 'running' in stdout
+    except Exception:
+        print('RHEVM: While checking status of ovirt engine, an exception happened')
```
mshriver: Let's print `rhevm_ip` so we can associate this message with a provider

mshriver:
@mfalesni Ran against CM QE provider from your branch, there may be further updates or improvements but I think this should be merged so Jenkins starts dumping more debug on new builds.

PR #4451, opened 2017-03-27
[1LP][RFR] Removed ssh_client fixture from test_actions.py

quarckster:
{{pytest: -v -k 'test_action_power_on_logged' --long-running --use-provider=vsphere6-nested}}

Purpose or Intent
=================

`ssh_client` fixture was removed and in this PR that fixture has been replaced by `appliance.ssh_client`.


mshriver:
dammit, I was too slow (#4455) :) Got assigned a JIRA card to fix these for 58z

PR #4452, opened 2017-03-27
[WIPTEST] Update BootstrapSelect, support partial text

mshriver:
Widget takes boolean 'partial', modifies locator with contains()

FIXES RHCFQE-2286

Currently testing


mshriver:
Shelving for now, shall implement with partial passed to the method instead of the widget instance.

PR #4460, opened 2017-03-28
[1LP][RFR] fix test_power_control_rest for scvmm and make testing more robust

mkoura:
* reset is supported on scvmm in 5.8
* more robust checking for power state changes
* make sure we are always using /api/vms and not /api/instances even for cloud providers

**PRT results**:
OK apart from failures to create vm on providers (not these tests faults) and semi-expected `SoftAssertionError`s (timeouts when waiting for change states - this is unreliable hence soft asserts).

{{pytest: cfme/tests/cloud_infra_common/test_power_control_rest.py -v --use-provider scvmm}}

### DIFF
--- a/cfme/tests/cloud_infra_common/test_power_control_rest.py
+++ b/cfme/tests/cloud_infra_common/test_power_control_rest.py
@@ -10,6 +10,7 @@ from utils.log import logger
 from cfme.cloud.provider.gce import GCEProvider
 from cfme.cloud.provider.ec2 import EC2Provider
 from cfme.infrastructure.provider.scvmm import SCVMMProvider
+from utils.version import current_version
 
 
 pytestmark = [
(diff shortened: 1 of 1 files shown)


cfme/tests/cloud_infra_common/test_power_control_rest.py
```diff
@@ -180,16 +190,20 @@ def test_reset_vm(rest_api, vm_obj, from_detail):
...
 
     old_date = vm.updated_on
     if from_detail:
         vm.action.reset()
     else:
-        vm_obj.get_collection_via_rest().action.reset(vm)
+        rest_api.collections.vms.action.reset(vm)
     success, message = verify_action_result(rest_api, assert_success=False)
-    if vm_obj.provider.one_of(GCEProvider, EC2Provider, SCVMMProvider):
+    if current_version() < '5.7':
```
psav: **REQUIRED**: we should rather use the appliance fixture and do 
```if appliance.version < '5.7'```

File: cfme/tests/cloud_infra_common/test_power_control_rest.py [hunk 2]
```diff
@@ -85,19 +93,20 @@ def test_stop(rest_api, vm_obj, from_detail):
...
 
     if from_detail:
         vm.action.stop()
     else:
-        vm_obj.get_collection_via_rest().action.stop(vm)
+        rest_api.collections.vms.action.stop(vm)
     verify_action_result(rest_api)
-    verify_vm_power_state(vm_obj, state=vm_obj.STATE_OFF)
+    wait_for_vm_state_change(vm_obj, vm_obj.STATE_OFF)
+    soft_assert(not verify_vm_power_state(vm, vm_obj.STATE_ON), "vm still running")
```
izapolsk: OPTIONAL: I don't think you need soft_assert here since this is only existing check in this test

PR #4467, opened 2017-03-29
[1LP][RFR] Container reports tests

gshefer:
{{pytest: cfme/tests/containers/test_reports.py -v --use-provider cm-env1}}

Added Container reports tests

- Added fixture vporizer to grabbing and manipulating VPOR db table. will be used later in more tests.
- Fix ContainerTestItem.pretty_id so support also objects without __name__ attrribute

### DIFF
--- a/cfme/containers/provider/__init__.py
+++ b/cfme/containers/provider/__init__.py
@@ -354,7 +354,9 @@ class ContainersTestItem(object):
         self.polarion_id = polarion_id
 
     def pretty_id(self):
-        return '{} ({})'.format(self.obj.__name__, self.polarion_id)
+        name = (self.obj.__name__ if hasattr(self.obj, '__name__')
+                else str(self.obj))
+        return '{} ({})'.format(name, self.polarion_id)
 
 
 def navigate_and_get_rows(provider, obj, count, table_class=CheckboxTable):
--- a/cfme/containers/provider/__init__.py
+++ b/cfme/containers/provider/__init__.py
@@ -354,9 +354,9 @@ class ContainersTestItem(object):
         self.polarion_id = polarion_id
 
     def pretty_id(self):
-        name = (self.obj.__name__ if hasattr(self.obj, '__name__')
-                else str(self.obj))
-        return '{} ({})'.format(name, self.polarion_id)
+        return '{} ({})'.format(
+            getattr(self.obj, '__name__', str(self.obj)),
+            self.polarion_id)
 
 
 def navigate_and_get_rows(provider, obj, count, table_class=CheckboxTable):
--- a/cfme/fixtures/vporizer.py
+++ b/cfme/fixtures/vporizer.py
@@ -2,7 +2,6 @@ from datetime import datetime
 import re
 from collections import namedtuple
 from random import random
-from cfme.fixtures.base import appliance
 
 import pytest
 
@@ -31,27 +30,33 @@ vpor_data_instance = namedtuple('vpor_data_instance',
                                     'max_mem_usage_absolute_average'
                                 ])
 
+cpu_usagemhz_rate_average__data_range = 8000
+derived_memory_used__data_range = 100
+max_cpu_usage_rate_average__data_range = 8000
+max_mem_usage_absolute_average__data_range = 100
+
+
+def gen_vpor_values():
+    return vpor_values_pattern.format(
+        random(), random() * cpu_usagemhz_rate_average__data_range,
+        random() * 100, random() * derived_memory_used__data_range,
+        random(), random() * max_cpu_usage_rate_average__data_range,
+        random() * 100, random() * max_mem_usage_absolute_average__data_range
+    )
+
 
 @pytest.yield_fixture(scope='module')
-def vporizer():
+def vporizer(appliance):
 
     """Grabbing vim_performance_operating_ranges table data for nodes and projects.
     In case that no such data exists, inserting fake rows"""
 
-    db = appliance().db
-    vpor = db.get('vim_performance_operating_ranges')
+    db = appliance.db
+    vpor = db['vim_performance_operating_ranges']
 
-    def gen_vpor_values():
-        return vpor_values_pattern.format(
-            random(), random() * 8000,
-            random() * 100, random() * 100,
-            random(), random() * 8000,
-            random() * 100, random() * 100
-        )
-
-    container_nodes = db.get('container_nodes')
-    container_pods = db.get('container_groups')
-    container_projects = db.get('container_projects')
+    container_nodes = db['container_nodes']
+    container_pods = db['container_groups']
+    container_projects = db['container_projects']
     created_at = datetime.now()
     vpor_data_list = []
 
@@ -60,8 +65,7 @@ def vporizer():
         ('ContainerNode', 'ContainerProject', 'ContainerGroup')
     ):
 
-        for resource in table.__table__.select() \
-                .where(True).execute().fetchall():
+        for resource in db.session.query(table).all():
             def get_resource_vpor_data():
                 return vpor.__table__.select().where(
                     (vpor.resource_id == resource.id) &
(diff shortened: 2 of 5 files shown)


cfme/containers/provider/__init__.py
```diff
@@ -331,7 +331,9 @@ def __init__(self, obj, polarion_id):
         self.polarion_id = polarion_id
 
     def pretty_id(self):
-        return '{} ({})'.format(self.obj.__name__, self.polarion_id)
+        name = (self.obj.__name__ if hasattr(self.obj, '__name__')
```
mshriver: **Required**
Use getattr() here and provide default of `str(self.obj)`

mshriver: Optionally keep it as a single statement as well, and stick `getattr()` into the `format()` call.

cfme/fixtures/vporizer.py
```diff
@@ -0,0 +1,97 @@
+from datetime import datetime
```
mshriver: I think you should split this fixture into its own PR and set this PR blocked by it so that it may be reviewed separately.

cfme/fixtures/vporizer.py
```diff
@@ -0,0 +1,97 @@
...
+                                ])
+
+
+@pytest.yield_fixture(scope='module')
+def vporizer():
+
+    """Grabbing vim_performance_operating_ranges table data for nodes and projects.
+    In case that no such data exists, inserting fake rows"""
+
+    db = appliance().db
```
mshriver: Why not have appliance as a parameter in your fixture definition?

cfme/fixtures/vporizer.py
```diff
@@ -0,0 +1,97 @@
...
+    vpor = db.get('vim_performance_operating_ranges')
+
+    def gen_vpor_values():
+        return vpor_values_pattern.format(
+            random(), random() * 8000,
```
mshriver: **Required**
Set module constants for these magic numbers.

cfme/tests/containers/test_reports.py
```diff
@@ -0,0 +1,201 @@
...
+    pytest.mark.tier(1)]
+pytest_generate_tests = testgen.generate([ContainersProvider], scope='function')
+
+
+@pytest.fixture(scope='module')
+def node_hardwares_db_data():
+
+    """Grabbing hardwares table data for nodes"""
+
+    db = appliance().db
```
mshriver: Again, why import and call this explicitly instead of just providing it as a parameter in your fixture definition?

cfme/tests/containers/test_reports.py
```diff
@@ -0,0 +1,201 @@
...
+    hardwares_table = db.get('hardwares')
+    container_nodes = db.get('container_nodes')
+
+    out = {}
+    for node in container_nodes.__table__.select().where(True).execute().fetchall():
```
mshriver: **Optional**
Consider changing up this syntax and DB queries to something more like:

```python
hardwares = db.session.query(db['hardwares']).all()

File: cfme/tests/containers/test_reports.py [hunk 7]
```diff
@@ -0,0 +1,201 @@
...
+# @pytest.mark.meta(blockers=[BZ(1435958, forced_streams=["5.8"])])
+@pytest.mark.polarion('CMP-9533')
+def test_pods_per_ready_status(soft_assert, pods_per_ready_status):
+
+    report = get_report('Pods per Ready Status')
+    for row in report.data.rows:
+        name = row['# Pods per Ready Status']
+        readiness_ui = (True if row['Ready Condition Status'].lower() == 'true'
+                        else False)
+        if name not in pods_per_ready_status:  # this check based on BZ#1435958
```
mshriver: **Optional**
Consider structuring this like:

```python
if soft_assert(name in pods_per_ready_status, 'Could not yada yada'):
    soft_assert(pods_per_ready_status[name] == readiness_ui, (...other code...)

File: cfme/fixtures/vporizer.py (hunk 2 above)
Comment (gshefer) [inline cfme/fixtures/vporizer.py]: I really prefer to keep it under this PR.

File: cfme/fixtures/vporizer.py (hunk 4 above)
Comment (gshefer) [inline cfme/fixtures/vporizer.py]: But there are not constants...

Comment (mshriver) [inline cfme/fixtures/vporizer.py]: what is 8000? why is it 8000?

File: cfme/fixtures/vporizer.py [hunk 8]
```diff
@@ -0,0 +1,101 @@
...
+            def get_resource_vpor_data():
+                return vpor.__table__.select().where(
+                    (vpor.resource_id == resource.id) &
+                    (vpor.resource_type == resource_type)
+                ).execute().fetchone()
+            # Checking whether values has already inserted to the table for this resource.
+            # Should be True only in case that the appliance age is greater than 24h
+            # if there isn't such row, insert...
+            if not get_resource_vpor_data():
+                vpor.__table__.insert().values(
```
mfalesni: Using "private" attributes raises questions - it cannot be done without using `__table__`?

cfme/tests/containers/test_reports.py
```diff
@@ -0,0 +1,198 @@
...
+                    .format(row['Name'], usage_db, usage_report))
+
+
+@pytest.mark.polarion('CMP-10033')
+def test_report_nodes_by_memory_usage(appliance, soft_assert, vporizer):
+
+    report = get_report('Nodes By Memory Usage')
+    for row in report.data.rows:
+
+        vpor_values = [vals for vals in vporizer if vals.resource_name == row["Name"]][0]
```
mfalesni: This deserves a shortcut, some function that does the task of this expression.

cfme/fixtures/vporizer.py
```diff
@@ -0,0 +1,101 @@
...
+    container_nodes = db['container_nodes']
+    container_pods = db['container_groups']
+    container_projects = db['container_projects']
+    created_at = datetime.now()
+    vpor_data_list = []
+
+    for table, resource_type in zip(
+        (container_nodes, container_projects, container_pods),
+        ('ContainerNode', 'ContainerProject', 'ContainerGroup')
+    ):
```
mfalesni: Can you zip this into an intermediate variable so the for does not look that ugly?

cfme/fixtures/vporizer.py
```diff
@@ -0,0 +1,101 @@
...
+                vpor_data_instance(
+                    resource_vpor_data['resource_type'],
+                    resource_vpor_data['resource_id'], resource.name,
+                    vpor_values[0], vpor_values[1], vpor_values[2], vpor_values[3]
+                )
+            )
+
+    yield vpor_data_list
+
+    vpor.__table__.delete().where(vpor.created_at == created_at).execute()
```
mfalesni: This is a bit of a nitpick: Not sure if that can be done, but it would be better to grab the IDs of the created entries and use them in this query to specifically reference those. I think the query to create the entries should somehow return the created entries with their IDs set. `created_at` does not have the uniqueness constraint so it could theoretically also reference some entries that were created by other process.

cfme/tests/containers/test_reports.py
```diff
@@ -0,0 +1,198 @@
...
+
+        out[node.name] = hardwares_table.__table__.select().where(
+            hardwares_table.id == node.id
+        ).execute().fetchone()
+
+    return out
+
+
+@pytest.fixture(scope='function')
+def pods_per_ready_status(provider):
```
mfalesni: This looks like a candidate for including in the mgmtsystem library.

cfme/fixtures/vporizer.py
gshefer: Not AFAIK

cfme/fixtures/vporizer.py
gshefer: Agree - WIP

cfme/tests/containers/test_reports.py
gshefer: I'll do it later, I just wan't to prevent bureaucracy and add reports testing coverage ASAP, I'm adding a comment for that

cfme/tests/containers/test_reports.py
gshefer: Absolutely

PR #4474, opened 2017-03-29
[1LP][RFR] new REST API tests checking that it's possible to limit returned attributes

mkoura:
new REST API tests checking that it's possible to limit returned attributes for every collection

{{pytest: -v -k "test_select_attributes or test_query_simple_collections"}}

### DIFF
--- a/cfme/tests/test_rest.py
+++ b/cfme/tests/test_rest.py
@@ -76,27 +76,34 @@ COLLECTIONS_ADDED_IN_58 = {
 }
 
 
+COLLECTIONS_ALL = {
+    "actions", "alert_definitions", "alerts", "arbitration_profiles",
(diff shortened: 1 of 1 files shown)


cfme/tests/test_rest.py
```diff
@@ -114,6 +121,34 @@ def test_query_simple_collections(rest_api, collection_name):
...
+
+    Metadata:
+        test_flag: rest
+    """
+    collection = getattr(rest_api.collections, collection_name)
+    response = rest_api.get('{}{}'.format(collection._href, '?expand=resources&attributes=id'))
+    assert rest_api.response.status_code == 200
+    for resource in response['resources']:
+        assert 'id' in resource
+        expected_len = 2 if 'href' in resource else 1
```
mshriver: **Optional**
Since this isn't used again just collapse it into the line below.

PR #4475, opened 2017-03-30
[1LP][RFR] Strings comparing must be replaced by type checking

quarckster:
Purpose or Intent
=================

Small refactoring in basic control tests. We should check types, not strings.


mshriver:
@quarckster Can you comment on the 56z failures in PRT please? Looks like some tests may not have been uncollected related to providers and middleware

quarckster:
@mshriver I made some changes and now tests are successfully uncollected. Failed tests for 5.6 doesn't depend on this PR.

PR #4476, opened 2017-03-30
[RFR][1LP] Updates for MIQ Fine release

mshriver:
Updating utils.version, utils.trackerbot for the newest MIQ 'Fine' release.

Update template_upload_all to dynamically parse MIQ release name and version string when not on the MIQ master branch.

Testing locally at the moment, requires YAML updates for CFME QE.

### DIFF
--- a/scripts/template_upload_all.py
+++ b/scripts/template_upload_all.py
@@ -94,12 +94,16 @@ def template_name(image_link, image_ts, checksum_link, version=None):
             # manageiq-pppp-bbbbbb-yyyymmddhhmm.ova => miq-nightly-vvvv-yyyymmddhhmm
             return "miq-nightly-{}-{}".format(version, result[0])
         elif "stable" in image_link:
-            if 'euwe' in image_link:
-                pattern = re.compile(r'[^\d]*?-euwe-(.*)-(?P<year>\d{4})(?P<month>\d{2})('
-                                     r'?P<day>\d{2})')
-                result = pattern.findall(image_name)
-                return "miq-stable-euwe-{}-{}{}{}".format(result[0][0], result[0][1], result[0][2],
-                                                      result[0][3])
+            # Handle named MIQ releases, dropping provider and capturing release and date
+            if 'master' not in image_link:
+                pattern = re.compile(r'manageiq-[^\d]*?-(?P<release>[-\w]*?)'
+                                     r'-(?P<year>\d{4})(?P<month>\d{2})(?P<day>\d{2})')
+                # Use match so we can use regex group names
+                result = pattern.match(image_name)
+                return "miq-stable-{}-{}{}{}".format(result.group('release'),
+                                                     result.group('year'),
+                                                     result.group('month'),
+                                                     result.group('day'))
             return "miq-stable-{}-{}".format(result[0][0], result[0][2])
         else:
             # manageiq-pppp-bbbbbb-yyyymmddhhmm.ova => miq-nightly-yyyymmddhhmm
--- a/scripts/template_upload_gce.py
+++ b/scripts/template_upload_gce.py
@@ -15,6 +15,7 @@ import sys
 import os
 import urllib2
 from threading import Lock, Thread
+from time import sleep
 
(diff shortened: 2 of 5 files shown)


scripts/template_upload_all.py
```diff
@@ -378,6 +382,8 @@ def main():
 
         print("TEMPLATE_UPLOAD_ALL:------End of {} upload on: {}--------".format(
             kwargs['template_name'], provider_type))
+    else:
```
mshriver: if we made it all the way through the for loop without returning, we never matched. Without this the 'failure' there is silent

scripts/template_upload_rhevm.py
```diff
@@ -81,9 +81,16 @@ def change_edomain_state(api, state, edomain, provider):
             export_domain = dc.storagedomains.get(edomain)
             if export_domain:
                 if state == 'maintenance' and export_domain.get_status().state == 'active':
-                    dc.storagedomains.get(edomain).deactivate()
+                    # may be tasks on the storage, try multiple times
+                    print('RHEVM:{} {} in active, waiting for deactivate...').format(
+                        provider, edomain)
+                    wait_for(lambda: dc.storagedomains.get(edomain).deactivate(), delay=5,
+                             num_sec=600, handle_exception=True)
```
mshriver: This does have a negative side effect of burying the exceptions from the `activate` or `deactivate` call so we won't get tracebacks on them, but a timeout exception instead.

scripts/template_upload_gce.py
```diff
@@ -77,23 +78,21 @@ def download_image_file(image_url, destination=None):
             return (file_name, file_path)
         os.remove(file_name)
     print("Downloading: {} Bytes: {}".format(file_name, file_size))
-    f = open(file_name, 'wb')
-    os.system('cls')
```
mshriver: This line removed. was generating an error

scripts/template_upload_gce.py
```diff
@@ -77,23 +78,21 @@ def download_image_file(image_url, destination=None):
...
-    while True:
-        buffer_f = u.read(block_sz)
-        if not buffer_f:
-            break
-
-        file_size_dl += len(buffer_f)
-        f.write(buffer_f)
-        status = r"{:.2f}".format(file_size_dl, file_size_dl * 100. / file_size)
-        status = status + chr(8) * (len(status) + 1)
-        print('r')
```
mshriver: This was printing an obscene amount of status lines.

utils/trackerbot.py
```diff
@@ -146,7 +146,7 @@ def mark_provider_template(api, provider, template, tested=None, usable=None,
     Returns the response of the API request
 
     """
-    provider_template = _as_providertemplate(provider, template)
+    provider_template = _as_providertemplate(provider, template, group=stream)
```
mshriver: I think this portion, connecting the stream from `trackerbot_add_provider_template` to `mark_provider_template`,  will need more testing on the various streams.

While testing the upstream_stable changes, the templates were getting added to the unknown group, and staying there after sync_template_tracker.py was run because the template already existed and `parse_template` was never called.

This connected the stream provided to the group, and the template was marked correctly in trackerbot.

@jkrocil I could use your help determining what's happening with trackerbot, and the templates not getting identified for the right group (upstream_stable in this case).

utils/version.py
```diff
@@ -173,8 +173,9 @@ def parse(self, vstring):
             vstring = ".".join(map(str, vstring))
         elif vstring:
             vstring = str(vstring).strip()
-        if vstring in ('master', 'latest', 'upstream') or 'euwe' in vstring:
+        if vstring in ('master', 'latest', 'upstream') or 'fine' in vstring:
             vstring = 'master'
+        # TODO These aren't used anywhere - remove?
```
mfalesni: dunno, this should have been a dictionary on the top of the file since its inception anyway.

jkrocil: I'm pretty sure it started out as a single conditional and then stacked up...

PR #4485, opened 2017-03-31
[RFR][1LP] Update test_appliance_console, grep for MIQ+CFME

mshriver:
This is causing failures in template testing for upstream.

@lcouzens should check this out

# PRT
{{ pytest: cfme/tests -m smoke --use-provider default }}


lcouzens:
Can we not set this test to be downstream only? 

I assume your change greps for ManageIQ or CFME to pass? This mean any downstream appliance with ManageIQ would pass when it should fail, correct?

@mshriver

mshriver:
ping @lcouzens, passing PRT up/downstream with explicitly picked 'CFME' or 'ManageIQ' via `utils.version.is_appliance_downstream`

lcouzens:
@mshriver Looks good to me!

PR #4490, opened 2017-04-02
[RFR]Add provider with SSL fixes

pavelzag:
Purpose or Intent
=================
This PR fixes the problem that appeared by adding the Hawkular validation tab.


jkrocil:
Can be merged if it works for you.

PR #4492, opened 2017-04-03
[1LP][RFR] fix unpacking of element_data

mkoura:
UPDATE:
I changed the PR to include just the framework fix to make the change obvious and speed up the review/merge process.

The issue is that the ``element_data`` can be list or tuple, in which case the current code fails.

### DIFF
--- a/cfme/automate/service_dialogs.py
+++ b/cfme/automate/service_dialogs.py
@@ -125,7 +125,10 @@ class ServiceDialog(Updateable, Pretty, Navigatable, Fillable):
         sel.wait_for_element(box_form.box_label)
         fill(box_form, {'box_label': self.box_label,
                         'box_desc': self.box_desc})
-        self.add_element(self.element_data)
+        if isinstance(self.element_data, (list, tuple)):
+            self.add_element(*self.element_data)
+        else:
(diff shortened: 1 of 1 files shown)


izapolsk:
all tests failed. I've restarted PRT tests

izapolsk:
@mkoura , tests failed again. Please either fix them or confirm that those are expected and why.

mkoura:
@izapolsk the PR4281 changed `service_dialog.create()` call and broke many tests, including mine. While I was at fixing my tests, I fixed all the remaining calls in other tests, too.
I wasn't trying to fix other problems in these tests.

mkoura:
This PR fixes ``service_dialog.create()`` calls (`create` no longer accepts `element_data`) and `element_data` unpacking - now it's broken when `element_data` is not single item.
There were other issues caused by https://github.com/ManageIQ/integration_tests/pull/4281 like changes to ``ServiceCatalogs`` instantiation that I'm addressing in https://github.com/ManageIQ/integration_tests/pull/4497
PRT results are not really relevant here, tests are broken for various other reasons and this PR is only solving issues with ``dialog.create()`` calls.

PR #4497, opened 2017-04-03
[RFR] fixes for service_catalogs

mkoura:
UPDATE:
I included tests fixes from https://github.com/ManageIQ/integration_tests/pull/4492 into this PR.
Sorry for the mess, I figured it will be easier to review/merge the PRs if it would be just tests fixes vs framework fixes.

The purpose of this PR is to fix calls to ``CatalogItem()``, ``ServiceCatalogs()`` and ``ServiceDialog.create()``.

This PR fixes **just the calls**, not all the other things that may be broken == tests are not expected to pass in PRT

### DIFF
--- a/cfme/tests/automate/test_service_dialog.py
+++ b/cfme/tests/automate/test_service_dialog.py
@@ -20,12 +20,13 @@ def test_create_service_dialog():
         'default_text_box': "Default text"
     }
     dialog = ServiceDialog(label=fauxfactory.gen_alphanumeric(),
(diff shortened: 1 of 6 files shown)


mkoura:
The PRT is failing on `service_dialog.create()` calls - fixes for these are in https://github.com/ManageIQ/integration_tests/pull/4492

psav:
Please can we get this fixed up ASAP - it seems to be affecting many tests

PR #4500, opened 2017-04-03
[1LP][RFR] Entrypoints for instances/templates

psav:
Entrypoints for Instance


cfme/cloud/instance/azure.py
```diff
@@ -31,6 +29,15 @@ class AzureInstance(Instance):
     STATE_UNKNOWN = "unknown"
     STATE_ARCHIVED = "archived"
 
+    UI_POWERSTATES_AVAILABLE = {
+        'on': [STOP, SUSPEND, SOFT_REBOOT, TERMINATE],
```
mfalesni: I guess we should also constantize the UI strings.

PR #4506, opened 2017-04-04
[1LP][RFR] Fixes to ssh client connection

lcouzens:
Fixing tests that cause ssh connections to be left open

### DIFF
--- a/cfme/fixtures/cli.py
+++ b/cfme/fixtures/cli.py
@@ -53,7 +53,7 @@ def fqdn_appliance(appliance):
         raise SproutException('No provision available')
     yield apps[0]
 
-    fqdn_appliance.ssh_client.close()
+    apps[0].ssh_client.close()
     sp.destroy_pool(pool_id)
 
 
--- a/cfme/fixtures/cli.py
+++ b/cfme/fixtures/cli.py
@@ -53,6 +53,7 @@ def fqdn_appliance(appliance):
         raise SproutException('No provision available')
     yield apps[0]
 
+    fqdn_appliance.ssh_client.close()
     sp.destroy_pool(pool_id)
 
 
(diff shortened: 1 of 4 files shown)


cfme/fixtures/cli.py
```diff
@@ -53,6 +53,7 @@ def fqdn_appliance(appliance):
         raise SproutException('No provision available')
     yield apps[0]
 
+    fqdn_appliance.ssh_client.close()
```
psav: This won't work, as ```fqdn_appliance``` isn't defined locally here and is the fixtures name. You probably want to use ```apps[0].ssh_client.close()``` instead

fixtures/appliance.py
```diff
@@ -32,6 +32,7 @@ def temp_appliances(count=1, preconfigured=True, lease_time=180):
             count=count, lease_time=lease_time, preconfigured=preconfigured)
         yield apps
     finally:
+        temp_appliances.ssh_client.close()
```
psav: You'll need an iterator here on ```apps```

psav:
Waiting on PRT results but looks good from here

cfme/tests/cli/test_appliance_cli.py
```diff
@@ -4,9 +4,10 @@
 
 
 def test_set_hostname(request, appliance):
-    appliance.appliance_console_cli.set_hostname('test.example.com')
+    hostname = 'Elite-QE.redhat.com'
```
jkrocil: Let's not.

quarckster: 🥇

lcouzens: @jkrocil Better?

PR #4509, opened 2017-04-04
[RFR] Testgen: Configurable provider limit for setting up providers

mfalesni:
{{pytest: cfme/tests/infrastructure/test_provisioning.py -v --long-running --use-provider complete --provider-limit 1 -s}}

### DIFF
--- a/fixtures/provider.py
+++ b/fixtures/provider.py
@@ -58,6 +58,15 @@ _setup_failures = defaultdict(lambda: 0)
 SETUP_FAIL_LIMIT = 3
 
 
+def pytest_addoption(parser):
+    # Create the cfme option group for use in other plugins
+    parser.getgroup('cfme')
+    parser.addoption("--provider-limit", action="store", default=0, type=int,
+        help=(
+            "Number of cloud/infra providers allowed to coexist on appliance. 0 means no limit. "
+            "Use 1 or 2 when running on a single appliance, depending on HW configuration."))
+
+
 def _artifactor_skip_providers(request, providers, skip_msg):
     node = request.node
     name, location = get_test_idents(node)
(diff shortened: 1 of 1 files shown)


fixtures/provider.py
```diff
@@ -58,6 +58,15 @@
 SETUP_FAIL_LIMIT = 3
 
 
+def pytest_addoption(parser):
+    # Create the cfme option group for use in other plugins
+    parser.getgroup('cfme')
+    parser.addoption("--provider-limit", action="store", default=0, type=int,
+        help=(
+            "Number of cloud/infra providers allowed to coexist on appliance. 0 means no limit. "
```
jkrocil: **REQUIRED** It's not just cloud and infra actually. It applies to anything derived from`BaseProvider` - so containers and middleware as well.

fixtures/provider.py
```diff
@@ -67,8 +76,26 @@ def _artifactor_skip_providers(request, providers, skip_msg):
     pytest.skip(skip_msg)
 
 
-def _setup_provider_verbose(provider):
+def _setup_provider_verbose(request, provider):
     try:
+        if request.config.option.provider_limit > 0:
+            existing_providers = [
+                p for p in list_providers(use_global_filters=False)
+                if p.exists and p.key != provider.key]
```
jkrocil: I think ideally we would want to get current appliance in here through the `appliance` fixture and use `appliance.managed_known_providers` but all that might still change once we get collections so... this is good enough.

fixtures/provider.py
jkrocil: Unless you turn the `list_providers(...)` below into `list_providers_by_class(CloudInfraProvider, use_global_filter=False)`. Then it would be just the two.

mfalesni: I guess I will just remove the cloud/infra mention in the comment.

fixtures/provider.py
```diff
@@ -67,8 +76,27 @@ def _artifactor_skip_providers(request, providers, skip_msg):
...
 
 
-def _setup_provider_verbose(provider):
+def _setup_provider_verbose(request, provider):
     try:
+        if request.config.option.provider_limit > 0:
+            # TODO: Use appliance.
+            existing_providers = [
+                p for p in list_providers(use_global_filters=False)
+                if p.exists and p.key != provider.key]
```
jkrocil: **OPTIONAL** You dont need to compare provider keys to check for equality. You can compare them directly `...and p != provider]` since `BaseProvider` has `__eq__` defined.

PR #4513, opened 2017-04-05
[1LP][RFR] sysV changes from service command to systemctl

lcouzens:
Changes to sysV commands to match newer rhel version


psav:
Is this because service is no longer available?

lcouzens:
@psav Yes. Well being depreciated I believe.

PR #4516, opened 2017-04-05
[1LP][RFR] Changes to cli variables to match new setup

lcouzens:
Changes to Variables in class to better match the cli regardless whether the short options are changed.
Also a change to one appliance_console test option due to latest build changes.

### DIFF
--- a/utils/appliance/__init__.py
+++ b/utils/appliance/__init__.py
@@ -126,7 +126,7 @@ class ApplianceConsoleCli(object):
                 dbname=dbname, fetch_key=fetch_key, sshlogin=sshlogin, sshpass=sshpass))
 
     def configure_ipa(self, ipaserver, username, password, domain, realm):
-        self._run("--ipaserver {ipaserver} --ipaprincapal {username} --ipapassword {password}"
+        self._run("--ipaserver {ipaserver} --ipaprincipal {username} --ipapassword {password}"
             " --ipadomain {domain} --iparealm {realm}".format(
                 ipaserver=ipaserver, username=username, password=password, domain=domain,
                 realm=realm))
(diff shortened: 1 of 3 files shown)


psav:
please explain failures

lcouzens:
Spelling mistake on my part ;)

psav:
Still fails on most.....explanation?

lcouzens:
@psav so the last fails I believe are down to provisions again like before, we still seem to have issues with rhos appliances.

izapolsk:
I've restarted tests to check whether the same tests fail again.

izapolsk:
@lcouzens , I've moved PR to 1LP but please either fix or expain errors.

lcouzens:
I think I have figured out the issue with this, I will try to come up with a fix.

utils/appliance/__init__.py
```diff
@@ -89,6 +89,8 @@ def run_commands(self, commands, autoreturn=True, timeout=20, channel=None):
             try:
                 while True:
                     result += channel.recv(1)
+                    if result is 'Press any key to continue':
```
izapolsk: @lcouzens: this will work in this certain case but please use == to check that objects have equal values. "is" is for other things.

izapolsk:
I've restarted tests since current PRT job hung

utils/appliance/__init__.py
mshriver: I'm going to call this **Required** please.

mshriver:
PRT results look pretty solid for run 12056, upstream has sprout provision exceptions and an exception that's occuring because there are multiple calls in a try block and the finally assumes the 2nd command ran (not in your changes here).  57z failures were ssh failures that look related to infrastructure, not your changes.

izapolsk:
@lcouzens , please look into errors in tests against 5.6. Two tests failed on apps var.

lcouzens:
Those failed because there were no provisions available from sprout.

izapolsk:
okay, I restarted tests then.

PR #4537, opened 2017-04-10
[1LP][RFR] Fix for Openstack nodes navigation

okolisny:
Fixing the issue with Openstack infrastructure nodes navigation (Hosts become Nodes for Openstack)

### DIFF
--- a/cfme/infrastructure/host.py
+++ b/cfme/infrastructure/host.py
@@ -17,6 +17,7 @@ import cfme.web_ui.flash as flash
 import cfme.web_ui.toolbar as tb
 from utils import conf
 from cfme.exceptions import HostNotFound
+from cfme.infrastructure.provider.openstack_infra import OpenstackInfraProvider
 from cfme.web_ui import (
     AngularSelect, Region, Quadicon, Form, Select, CheckboxTree, CheckboxTable, DriftGrid, fill,
     form_buttons, paginator, Input, mixins, match_location
@@ -453,7 +454,10 @@ class All(CFMENavigateStep):
     prerequisite = NavigateToAttribute('appliance.server', 'LoggedIn')
 
     def step(self):
-        self.prerequisite_view.navigation.select('Compute', 'Infrastructure', 'Hosts')
+        dest = 'Hosts'
+        if hasattr(self.obj, 'provider') and self.obj.provider.one_of(OpenstackInfraProvider):
+            dest = 'Nodes'
+        self.prerequisite_view.navigation.select('Compute', 'Infrastructure', dest)
 
     def resetter(self):
         tb.select("Grid View")
--- a/cfme/infrastructure/host.py
+++ b/cfme/infrastructure/host.py
@@ -622,7 +622,7 @@ def find_quadicon(host, do_not_navigate=False):
         raise HostNotFound("Host '{}' not found in UI!".format(host))
 
 
-def navigate_and_select_all_hosts(host_names, provider):
+def navigate_and_select_all_hosts(host_names, provider=None):
     """ Reduces some redundant code shared between methods """
     if isinstance(host_names, basestring):
         host_names = [host_names]
--- a/cfme/infrastructure/host.py
+++ b/cfme/infrastructure/host.py
@@ -17,7 +17,6 @@ import cfme.web_ui.flash as flash
 import cfme.web_ui.toolbar as tb
 from utils import conf
 from cfme.exceptions import HostNotFound
-from cfme.infrastructure.provider.openstack_infra import OpenstackInfraProvider
 from cfme.web_ui import (
     AngularSelect, Region, Quadicon, Form, Select, CheckboxTree, CheckboxTable, DriftGrid, fill,
     form_buttons, paginator, Input, mixins, match_location
@@ -454,8 +453,9 @@ class All(CFMENavigateStep):
     prerequisite = NavigateToAttribute('appliance.server', 'LoggedIn')
 
     def step(self):
+        from cfme.infrastructure.provider.openstack_infra import OpenstackInfraProvider
         dest = 'Hosts'
-        if hasattr(self.obj, 'provider') and self.obj.provider.one_of(OpenstackInfraProvider):
+        if self.obj.provider and self.obj.provider.one_of(OpenstackInfraProvider):
             dest = 'Nodes'
         self.prerequisite_view.navigation.select('Compute', 'Infrastructure', dest)
 
--- a/cfme/infrastructure/host.py
+++ b/cfme/infrastructure/host.py
@@ -274,8 +274,7 @@ class Host(Updateable, Pretty, Navigatable, PolicyProfileAssignable):
 
         def _looking_for_state_change():
             tb.refresh()
-            return 'currentstate-' + desired_state in find_quadicon(self.name,
-                                                                    do_not_navigate=False).state
+            return 'currentstate-' + desired_state in find_quadicon(self).state
 
         navigate_and_select_all_hosts(self)
         return wait_for(_looking_for_state_change, num_sec=timeout)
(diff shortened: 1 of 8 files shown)


izapolsk:
@okolisny, all PRT tests fail please check what's wrong

okolisny:
@izapolsk it was a cycling import error, fixed by moving import to a local scope. Also, I've refactored some methods in host.py, so you probably want to review it all from the beginning

cfme/infrastructure/host.py
```diff
@@ -452,7 +448,12 @@ class All(CFMENavigateStep):
     prerequisite = NavigateToAttribute('appliance.server', 'LoggedIn')
 
     def step(self):
-        self.prerequisite_view.navigation.select('Compute', 'Infrastructure', 'Hosts')
+        from cfme.infrastructure.provider.openstack_infra import OpenstackInfraProvider
+        dest = 'Hosts'
+        if hasattr(self.obj, 'provider') and self.obj.provider:
```
mshriver: I don't think you need a `hasattr` call here, the parameter is optional but the class attribute will be set regardless of if its None or not.

Looks like `if self.obj.provider` is sufficient, since it will either be None or a provider object.

mshriver:
Knocking back to RFR and re-tagging @izapolsk for review since you stated you've done some additional refactoring.

cfme/infrastructure/host.py
```diff
@@ -607,29 +608,25 @@ def find_quadicon(host, do_not_navigate=False):
     """Find and return a quadicon belonging to a specific host
 
     Args:
-        host: Host name as displayed at the quadicon
+        host: Host object
```
mshriver: Have you audited the codebase for all uses of this find_quadicon, dynamic and static, since this change will break existing calls?

cfme/infrastructure/host.py
```diff
@@ -607,29 +608,25 @@ def find_quadicon(host, do_not_navigate=False):
...
-    else:
-        navigate_to(Host, 'All')
+    if not isinstance(hosts, list):
+        hosts = [hosts]
+    navigate_to(hosts[0], 'All')
     if paginator.page_controls_exist():
         paginator.results_per_page(1000)
-    for host_name in host_names:
-        sel.check(Quadicon(host_name, 'host').checkbox())
+    for host in hosts:
```
mshriver: **Required**
Use `paginator.check_all()`, this will take forever comparatively if there actually are 1000 hosts.

cfme/infrastructure/host.py
```diff
@@ -607,29 +608,25 @@ def find_quadicon(host, do_not_navigate=False):
...
-    if isinstance(host_names, basestring):
-        host_names = [host_names]
-
-    if provider_crud:
-        navigate_to(provider_crud, 'ProviderNodes')
-    else:
-        navigate_to(Host, 'All')
+    if not isinstance(hosts, list):
+        hosts = [hosts]
+    navigate_to(hosts[0], 'All')
```
mshriver: Consider just navigating to `Host` class here.

cfme/infrastructure/host.py
okolisny: `self.obj.provider` raises 'AttributeError' when navigating with class object. For example:
`navigate_to(Host, 'All')`

cfme/infrastructure/host.py
okolisny: Yes, I was checking, but looks like miss one call at **test_advanced_search_host.py**
Changing a call there

cfme/infrastructure/host.py
okolisny: It will not work for Openstack, because it demands now `isinstance(self.obj.provider, OpenstackInfraProvider)` inside the navigation class!

cfme/infrastructure/host.py
mfalesni: then try/except may be a better and more pythonic choice?

We should move away from class navigation in future anyway having the collections in place. But that is for future :)

okolisny: do you mean to wrap actual navigation into try/except? for example:
`try:
    self.prerequisite_view.navigation.select('Compute', 'Infrastructure', 'Hosts')
except NavigationDestinationError:
    self.prerequisite_view.navigation.select('Compute', 'Infrastructure', 'Nodes')`

okolisny: I'm sorry for 'code style' :)

mfalesni: @okolisny I updated your comment, look how it is done :)

mfalesni: In python it is "better ask for forgiveness than for permission"

okolisny: @mfalesni well, I agree with your suggestion. In this case, by the way, most changes from this PR are excess. So we can easily revert it.
@mshriver are you okay with it? I remember we were talking about 'asking for permission' variant at first

dajoRH:
Lint report for commit c7e35121b5538081e0b634db2ca07e9bc832d126:

`cfme/infrastructure/host.py`:
- :red_circle: Line 12:1: **F401** *'selenium.common.exceptions.NoSuchElementException' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

okolisny:
Setting to RFR. Reverted some changes in host.py

mshriver:
Restarted tests, run 12212. Previous run was a false-positive with provider skips.

mshriver:
Lots of NoSuchElement exceptions running against the RHOS7 provider in our PRT environment, can you explain or address?

okolisny:
@mshriver It seems like your Openstack provider is in a bad form:) rerun the tests.
Btw, everything is ok in our [CI jobs](https://rhos-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/rhos10_cfme58_integration/TEST_REPORT/) (RHOS 10 and latest 5.8)

okolisny:
@mshriver The problem is that your Openstack doesn't have resources which pointed in yamls. Setting back to RFR

cfme/infrastructure/host.py
```diff
@@ -620,16 +621,17 @@ def find_quadicon(host, do_not_navigate=False):
         raise HostNotFound("Host '{}' not found in UI!".format(host))
 
 
-def navigate_and_select_all_hosts(host_names, provider_crud=None):
+def navigate_and_select_all_hosts(host_names, provider):
```
mshriver: **Required**
This was an optional parameter, and you kept the logic that deals with it being empty. it should remain an optional parameter. `provider=None`

izapolsk:
@okolisny, almost all tests failed against 5.6 and 5.8.
Please try to sort out errors and move it back to RFR when tests are good enough.

cfme/configure/tasks.py
```diff
@@ -131,7 +131,7 @@ def is_analysis_finished(name, task_type='vm', clear_tasks_after_success=True):
             'state': 'finished'
         },
         'host': {
-            'tab': 'MyOther',
```
izapolsk: are you sure this is true for all versions 5.6-5.8 ?

cfme/infrastructure/host.py
```diff
@@ -621,16 +622,17 @@ def find_quadicon(host, do_not_navigate=False):
...
 
-    if provider_crud:
-        navigate_to(provider_crud, 'ProviderNodes')
+    if provider:
+        navigate_to(provider, 'ProviderNodes')
     else:
         navigate_to(Host, 'All')
+
     if paginator.page_controls_exist():
-        paginator.results_per_page(1000)
```
izapolsk: REQUIRED: You have removed row which made cfme show you all objects if there are several pages.
And added row paginator.check_all which does nothing. It just gives you select_all input. btw, you left statement where you select all passed elements.
If I understand correctly,
you need to do something like:

 ```
if paginator.page_controls_exist():
      paginator.results_per_page(1000)
      sel.click(paginator.check_all())
else:
    for host_name in host_names:
          sel.check(Quadicon(host_name, 'host').checkbox())
```

cfme/configure/tasks.py
okolisny: Yes, this string is the actual navigation described in this module for Task.

okolisny:
Tests are failing because it can't find 'private3' cloud network on instance provision form. You need update your yamls with actual data for Openstack provider

izapolsk:
@okolisny, all tests are skipped due to problems with rhos7 ga provider. I cant merge PR before PRT tests are ok.

btw, rhos provider has private3 network.

mshriver:
I've made a change to the rhos provider used by PRT to share the 'private3' network defined by the YAMLs used by PRT and restarted tests.

izapolsk:
@mshriver,  rhos_us, our only infra rhos provider and  private3 existed in it when I added it last time.

okolisny:
@izapolsk infra provider? we were talking about cloud provider..

PR #4544, opened 2017-04-11
[1LP][RFR] Implement instance ui_powerstates_* as properties

mshriver:
The power control tests fail when a deferred_verpick isn't picked on access because its defined for a class constant.

Thanks @mfalesni for the help brainstorming this change!

{{ pytest: cfme/tests/cloud/test_instance_power_control.py --long-running --use-provider azure -k test_power_options_from }}

## PRT
Failures are either provider setup (GCE and EC2 access down at the moment, rhos7-ga+rhos8 unstable) or a VM not hitting the desired power state in time.

There is an Azure softassert failure against the available power options. Need to consult with FA owner for this. Its outside the scope of my intended fixes here.

### DIFF
--- a/cfme/cloud/instance/azure.py
+++ b/cfme/cloud/instance/azure.py
@@ -29,14 +29,17 @@ class AzureInstance(Instance):
     STATE_UNKNOWN = "unknown"
     STATE_ARCHIVED = "archived"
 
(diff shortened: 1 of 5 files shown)


mshriver:
@psav tagged you because IIRC you had added these class vars.

PR #4545, opened 2017-04-11
[RFR] Jenkins Template Tester script updates

mshriver:
Fix template_upload_all regex to capture build filenames with '.' in the version.

Also strip the trailing slash when comparing build URL's, to avoid silly failures.

Add verbosity to engine_image_uploader call in template_upload_rhevm to get more status information from the script in the event of a failure.

Changes tested locally against the euwe-2.1 build.


mshriver:
@dajohnso gave me a sanity check, ninja-merging.

PR #4551, opened 2017-04-13
changed blocker mark to correct blocker bz

jistefan:
Changed bz number from 5.8 verified bug to 5.7 blocker clone which is reason tests are skipped


quarckster:
@jistefan Please use [WIP] or [WIPTEST] in title

PR #4558, opened 2017-04-17
[WIPTEST] Allow use of CFME_TESTS_KEY env var to store yaml encryption key

LandoCalrizzian:
Fix for storing the yaml encryption key in the CFME_TESTS_KEY environment variable instead of .yaml_key. 

Getting started guide states that the yaml encryption can be stored in .yaml_key OR  CFME_TESTS_KEY but the environment variable is never read and automation fails when .yaml_key doesn't exist.  This will allow either method to be used and CFME_TEST_KEY will override .yaml_key if it exists. Update affects pytest and dockerbot automation.

This is a quality of life fix and can be disregarded if you want to settle on using .yaml_key only (w/ update to docs).

### DIFF
--- a/cfme/test_framework/config.py
+++ b/cfme/test_framework/config.py
@@ -15,13 +15,16 @@ class Configuration(object):
     def __init__(self):
         self.yaycl_config = None
 
-    def configure(self, config_dir, crypt_key_file=None):
+    def configure(self, config_dir, crypt_key_file=None, crypt_key_env_var='CFME_TESTS_KEY'):
         """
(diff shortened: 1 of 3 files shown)


dajoRH:
*CFME QE Bot*

LandoCalrizzian:
Closing this PR since https://github.com/ManageIQ/integration_tests/commit/ab8516c6d9e4fca6154fcd104a3fd99273d92891 removed any reference to the ability to store the yaml encryption key in CFME_TESTS_KEY envvar

PR #4559, opened 2017-04-17
[1LP][RFR]Updated tests to conform to the Navmazing changes in service_ca…

jteehan:
…talogs

Purpose or Intent
=================

Navigation was failing on orchestration tests due to changes made to service_catalogs for NavMazing, which changed the parameters that I needed to send when initializing.  Tested locally and is working fine.

{{pytest: cfme/tests/services/test_provision_stack.py::test_provision_stack -v --use-provider azure }}


jteehan:
All the PRT tests passed, but I accidently hit the retest button.  Should still pass again, but this is ready to merge.

PR #4566, opened 2017-04-19
[1LP][RFR]Change provider type for RHV

bargulg:
Provider type naming is changing all the time, so here's the latest dict that should hopefully work with all versions for some time

### DIFF
--- a/cfme/infrastructure/provider/rhevm.py
+++ b/cfme/infrastructure/provider/rhevm.py
@@ -28,7 +28,11 @@ class RHEVMProvider(InfraProvider):
     def _form_mapping(self, create=None, **kwargs):
         provider_name = version.pick({
             version.LOWEST: 'Red Hat Enterprise Virtualization Manager',
-            '5.7.1': 'Red Hat Virtualization Manager'})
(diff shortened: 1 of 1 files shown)


mfalesni:
@bargulg Why did you put 1LP in the description?

bargulg:
@mfalesni because I've seen people put 1LP in the description for 5-line patches if they were simple enough

mfalesni:
@bargulg Really? Only people that were given the right to do the first or second level review should set the 1LP and definitely not on their own PRs. I have sent you a link in IRC.

bargulg:
Ok, I'm sorry if this patch is too complicated for you to consider it as 1-line patch, and I'm sorry for wasting your time on this.

bargulg:
Oh, I just noticed what 1LP means, sorry

bargulg:
This code review proces reminds me of gerrit review process quite a lot, except it's less automated, less clear, and we're using a tool which isn't designed for this - maybe we should migrate to gerrit to simplify this?

mfalesni:
@bargulg Nope.

mfalesni:
This looks good for merging, so merging.

PR #4569, opened 2017-04-20
[1LP][RFR]Marked test with BZ

sshveta:
{{pytest: -v --long-running --use-provider azure cfme/tests/services/test_provision_stack.py }}

### DIFF
--- a/cfme/services/myservice.py
+++ b/cfme/services/myservice.py
@@ -137,7 +137,7 @@ class ServiceRetirementView(ServiceRetirementForm):
 class ReconfigureServiceView(SetOwnershipForm):
     title = Text("#explorer_title_text")
 
-    save_button = Button('Save')
+    submit_button = Button('Submit')
 
     @property
     def is_displayed(self):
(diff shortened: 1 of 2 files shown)


cfme/services/myservice.py
```diff
@@ -250,6 +250,12 @@ def download_file(self, extension):
         view.download_choice.item_select("Download as " + extension)
         view.flash.assert_no_error()
 
+    def reconfigure_service(self):
+        view = navigate_to(self, 'Reconfigure')
+        view.submit_button.click()
+        # Request page is displayed so not asserting for view
+        view.flash.assert_no_error()
```
mshriver: **Required**
Add a TODO here to create view and assert against the redirect once Requests is modeled with widgetastic.  The flash widget here is inherited through several views.

mshriver:
@sshveta  Just add a TODO and then ping me and I'll merge.

mshriver:
56z result isn't reported yet but it passed, and skipped in 57z, 58z, upstream as intended.

PR #4573, opened 2017-04-21
[RFR] sprout: setup time sync in appliances after provisioning

izapolsk:
Purpose or Intent
=================
CFME experiences different issues when system time is incorrect.
Sprout currently doesn't fix time sync daemon after provisioning appliances. 
This PR will make sprout run setup for time sync daemon after provisioning.

### DIFF
--- a/sprout/appliances/tasks.py
+++ b/sprout/appliances/tasks.py
@@ -894,6 +894,8 @@ def appliance_power_on(self, appliance_id):
                 # VM is running now.
                 sync_appliance_hw.delay(appliance.id)
                 sync_provider_hw.delay(appliance.template.provider.id)
+                # fixes time synchronization
+                appliance.ipapp.fix_ntp_clock()
                 return
             else:
                 # IP not present yet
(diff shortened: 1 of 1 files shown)


sprout/appliances/tasks.py
```diff
@@ -907,6 +907,10 @@ def appliance_power_on(self, appliance_id):
             appliance.set_status("Powering on.")
             appliance.provider_api.start_vm(appliance.name)
             self.retry(args=(appliance_id, ), countdown=20, max_retries=40)
+
+        # fixes time synchronization
+        cfme_appliance = CFMEAppliance(appliance.provider_name, appliance.name)
```
mfalesni: Look in models.py, I bet there is an IPAppliance invocation in one of the properties. Otherwise I think it is good.

izapolsk: I missed it when I was looking through Appliance class. ty.

sprout/appliances/tasks.py
```diff
@@ -907,6 +907,9 @@ def appliance_power_on(self, appliance_id):
             appliance.set_status("Powering on.")
             appliance.provider_api.start_vm(appliance.name)
             self.retry(args=(appliance_id, ), countdown=20, max_retries=40)
+
+        # fixes time synchronization
+        appliance.ipapp.fix_ntp_clock()
```
mfalesni: last thing, put this line just after `sync_provider_hw.delay(appliance.template.provider.id)` ... this place is dead - it can never be reached.

izapolsk: hm, I thought the rest of if/elif statements should lead to this destination. fixed.

PR #4574, opened 2017-04-21
[RFR][1LP] More metadata into Ostriz

mfalesni:


fixtures/artifactor_plugin.py
```diff
@@ -187,13 +188,24 @@ def pytest_runtest_protocol(item):
     # This pre_start_test hook is needed so that filedump is able to make get the test
     # object set up before the logger starts logging. As the logger fires a nested hook
     # to the filedumper, and we can't specify order inriggerlib.
+    meta = item.get_marker('meta')
+    if meta and 'blockers' in meta.kwargs:
+        blocker_spec = meta.kwargs['blockers']
+        blockers = []
+        for blocker in blocker_spec:
+            if isinstance(blocker, int):
```
mshriver: **Optional**
```python
blockers.append(BZ(blocker).url if isinstance(blocker, int) else Blocker.parse(blocker).url)

mshriver:
Go ahead and merge if you don't like my optional comment.

PR #4575, opened 2017-04-23
[1LP][RFR] Chargeback tests

gshefer:
{{pytest: cfme/tests/containers/test_chargeback_for_images.py cfme/tests/containers/test_chargeback_for_projects.py -v -m "not long_running_env and not long_running_provider" --use-provider cm-env1 }}

Added chargeback tests for container images and projects

Note: failures in some tests are expected (required long running CFME).


cfme/tests/containers/test_chargeback_for_images.py
```diff
@@ -0,0 +1,123 @@
...
+    }
+    ccb = cb.ComputeRate(description, fields=data, appliance=appliance)
+    ccb.create()
+    return ccb
+
+
+@pytest.fixture(scope='module')
+def provider_name():
+    # Since 'provider' fixture is a 'function' scoped, we use this
+    # fixture to yield the provider name
```
mkoura: I believe you can use ``pytest.mark.usefixtures("setup_provider_modscope")`` (instead of "setup_provider")

cfme/tests/containers/test_chargeback_for_projects.py
```diff
@@ -0,0 +1,236 @@
...
+    }
+    ccb = cb.ComputeRate(description, fields=data, appliance=appliance)
+    ccb.create()
+    return ccb
+
+
+@pytest.fixture(scope='module')
+def provider_name():
+    # Since 'provider' fixture is a 'function' scoped, we use this
+    # fixture to yield the provider name
```
mkoura: same as above

mkoura:
You might need to tweak the PRT settings as it tests nothing atm. Something like
```
{{pytest: cfme/tests/containers/test_chargeback_for_images.py cfme/tests/containers/test_chargeback_for_projects.py -v --long-running --use-provider cm-env1 }}
```

gshefer:
@mkoura OK but note that some of them are going to fail since they are used for long running environment only.

gshefer:
@mkoura I run the tests in PRT with and without the -m ...
In the running without, tests are fails but that expected because they should running in long running env, you can see run # 12412.
In the running with -m ... (currently) it passes because it's running only the tests that are not long running so I prefer to keep it as now.
Thanks!

PR #4580, opened 2017-04-24
[1LP][RFR] update manageiq-client in requirements/frozen.txt

mkoura:
Updating to v0.3.0 so better subcollection support and HTTP method selection can be used in new tests.


quarckster:
@mkoura I can't find  0.3 version in https://pypi.python.org/pypi/manageiq-client

mshriver:
Marking this WIP, please update title to RFR when the package is available on pypi.

mkoura:
https://pypi.python.org/pypi/manageiq-client/0.3.0

PR #4605, opened 2017-04-28
[1LP][RFR] Removed references to 5.3

psav:
Removed some very old references

### DIFF
--- a/utils/appliance/__init__.py
+++ b/utils/appliance/__init__.py
@@ -343,30 +343,7 @@ class IPAppliance(object):
         log_callback("Configuring appliance {}".format(self.address))
         with self as ipapp:
             ipapp.wait_for_ssh()
-            configure_function = pick({
-                '5.2': self._configure_5_2,
-                '5.3': self._configure_5_3,
-                LATEST: self._configure_upstream,
-            })
-            configure_function(log_callback=log_callback)
-
-    # When calling any of these methods, `with self:` context must be entered.
-    def _configure_5_2(self, log_callback=None):
-        self.deploy_merkyl(start=True, log_callback=log_callback)
-        self.fix_ntp_clock(log_callback=log_callback)
-        self.enable_internal_db(log_callback=log_callback)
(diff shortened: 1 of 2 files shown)


utils/appliance/__init__.py
```diff
@@ -343,30 +343,7 @@ def configure(self, log_callback=None, **kwargs):
...
-
-    def _configure_5_3(self, log_callback=None):
-        self.deploy_merkyl(start=True, log_callback=log_callback)
-        self.fix_ntp_clock(log_callback=log_callback)
-        self.enable_internal_db(log_callback=log_callback)
-        self.wait_for_web_ui(timeout=1800, log_callback=log_callback)
-        self.loosen_pgssl(log_callback=log_callback)
-        # self.ipapp.update_rhel(log_callback=log_callback)
-        self.wait_for_web_ui(timeout=1800, log_callback=log_callback)
+            self._configure_upstream(log_callback=log_callback)
```
mshriver: this leaves me thinking we need to reconsider the name 'configure_upstream', the method name itself shouldn't include the version, the actions within the 'configure' method should be selected based on the version. ANYWAY not relevant here.

utils/appliance/__init__.py
```diff
@@ -343,30 +343,7 @@ def configure(self, log_callback=None, **kwargs):
...
-    # When calling any of these methods, `with self:` context must be entered.
-    def _configure_5_2(self, log_callback=None):
-        self.deploy_merkyl(start=True, log_callback=log_callback)
-        self.fix_ntp_clock(log_callback=log_callback)
-        self.enable_internal_db(log_callback=log_callback)
-        # need to skip_broken here until/unless we see a newer 52z build
-        self.update_rhel(log_callback=log_callback, skip_broken=True)
-        self.wait_for_web_ui(timeout=1800, log_callback=log_callback)
-
-    def _configure_5_3(self, log_callback=None):
```
mshriver: @jamesoden You may want to pay attention here, I believe you were dropping your SSL configs into this method.

PR #4609, opened 2017-04-28
[1LP][RFR] reworking requests

lkhomenk:
__rework__ for requests.
Sometimes we just need to find Request in the table - without waiting it to finish etc

### DIFF
--- a/cfme/services/requests.py
+++ b/cfme/services/requests.py
@@ -180,30 +180,7 @@ def wait_for_request(cells, partial_check=False):
     Returns:
          The matching :py:class:`cfme.web_ui.Table.Row` if found, ``False`` otherwise.
     """
-    for page in paginator.pages():
-        # We check only for the SplitTable. Can't think of better detection.
-        if version.current_version() < "5.5.0.8"\
-                and sel.elements(fields.request_list._header_loc) and\
-                not sel.is_displayed(fields.request_list):
-            # The table exists but it is hidden - no cells
-            return False
-        results = fields.request_list.find_rows_by_cells(cells, partial_check)
-        if len(results) == 0:
-            # row not on this page, assume it has yet to appear
-            continue
-        elif len(results) > 1:
-            raise RequestException(
-                'Multiple requests with matching content found - be more specific!'
-            )
-        else:
-            # found the row!
-            row = results[0]
-            logger.debug(' Request Message: %s', row.last_message.text)
-            break
-    else:
-        # Request not found at all, can't continue
-        return False
-
+    row = find_request(cells, partial_check)
     if row.request_state.text in REQUEST_FINISHED_STATES:
         return row
     else:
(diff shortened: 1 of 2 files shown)


cfme/services/requests.py
```diff
@@ -217,28 +194,40 @@ def debug_requests():
...
     """
     navigate_to(Request, 'All')
     for page in paginator.pages():
-        try:
-            # found the row!
-            row, = fields.request_list.find_rows_by_cells(cells)
-            return True
-        except ValueError:
+        # We check only for the SplitTable. Can't think of better detection.
+        if version.current_version() < "5.5.0.8"\
```
psav: **REQUIRED:** As we are reworking this, let's remove the 5.5 logic as 5.5 is no longer supported

lkhomenk: removed

cfme/tests/services/test_provision_stack.py
```diff
@@ -200,7 +200,6 @@ def test_remove_template_provisioning(provider, provisioning, catalog, catalog_i
     template.delete()
     row_description = 'Provisioning Service [{}] from [{}]'.format(item_name, item_name)
     cells = {'Description': row_description}
-    wait_for(lambda: requests.find_request(cells), num_sec=500, delay=20)
```
lkhomenk: added if_exists for such cases. but this particular doesn't need it at all - wait_for_request does all the same. so saving time a bit

cfme/services/requests.py
```diff
@@ -217,28 +194,38 @@ def debug_requests():
...
             continue
+        elif len(results) > 1:
+            raise RequestException(
+                'Multiple requests with matching content found - be more specific!'
+            )
+        else:
+            # found the row!
+            row = results[0]
+            logger.debug(' Request Message: %s', row.last_message.text)
+            break
```
mshriver: **Required**
Return here instead of breaking

cfme/services/requests.py
```diff
@@ -250,19 +237,12 @@ def go_to_request(cells, partial_check=False):
...
-            # row not on this page, assume it has yet to appear
-            # it might be nice to add an option to fail at this point
-            continue
-    else:
-        # Request not found at all, can't continue
+    try:
+        row = find_request(cells, partial_check)
+        sel.click(row)
+        return True
+    except Exception as e:
```
mshriver: **Required**
Its not clear which of these calls may result in an exception that gets put in the log and ignored.

If the row isn't found, find_request returns false, which means its the sel.click throwing an exception when really the failure was in find_request.

I don't think all exceptions should be hidden here, and the point of failure should be more clear.

Consider something structured more like:
```python
row = find_request(cells, partial_check)
if row:
    sel.click(row)
    return True
else:
    return False
```

lkhomenk:
widgetastic and navmazing will come apart from this PR

mshriver:
PRT results have failures but so far look unrelated to your changes, I'll merge once PRT is complete and I can confirm none of the failures are within the methods you've updated here.

mshriver:
squashing commits since the history isn't very clean and only 2/5 are verified.

PR #4610, opened 2017-04-28
[1LP][RFR] Fixed a test for 5.8

psav:
Fixed test where flash message changed

### DIFF
--- a/cfme/tests/configure/test_access_control.py
+++ b/cfme/tests/configure/test_access_control.py
@@ -92,10 +92,15 @@ def test_user_login():
 
 
 @pytest.mark.tier(3)
-def test_user_duplicate_name():
+def test_user_duplicate_name(appliance):
+    region = appliance.server_region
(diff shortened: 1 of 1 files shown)


cfme/tests/configure/test_access_control.py
```diff
@@ -243,10 +243,15 @@ def test_group_crud():
...
+def test_group_duplicate_name(appliance):
+    region = appliance.server_region
     group = new_group()
     group.create()
-    with error.expected("Description has already been taken"):
+    msg = version.pick({
+        version.LOWEST: "Description has already been taken",
+        '5.8': "Userid is not unique within region {}".format(region)
+    })
+    with error.expected(msg):
```
mfalesni: I think error.expected does verpick for us, doesn't it?

psav: No it doens't

PR #4626, opened 2017-05-04
[1LP][RFR] More SSA tests

gshefer:
{{pytest: cfme/tests/containers/test_containers_smartstate_analysis.py -v --use-provider cm-env1 }}

2 Additional test cases:
- check compliance.
- Profile setting
- new blocker bug for SSA
- Fixes in SSA test
- Changed container image instantiation from tag to id

### DIFF
--- a/cfme/containers/image.py
+++ b/cfme/containers/image.py
@@ -10,7 +10,7 @@ from cfme.web_ui import toolbar as tb, CheckboxTable, paginator, match_location,
 from utils.appliance.implementations.ui import CFMENavigateStep, navigator, navigate_to
 from utils.appliance import Navigatable
 from cfme.configure import tasks
-from wait_for import TimedOutError
+from utils.wait import wait_for, TimedOutError
 
 list_tbl = CheckboxTable(table_locator="//div[@id='list_grid']//table")
 paged_tbl = PagedTable(table_locator="//div[@id='list_grid']//table")
@@ -58,6 +58,44 @@ class Image(Taggable, SummaryMixin, Navigatable, PolicyProfileAssignable):
             except TimedOutError:
                 raise TimedOutError('Timeout exceeded, Waited too much time for SSA to finish ({}).'
                                     .format(ssa_timeout))
+
+    def check_compliance(self, timeout=240):
+        """Initiates compliance check and waits for it to finish."""
+        navigate_to(self, 'Details')
+        original_state = self.compliance_status
+        tb.select('Policy',
+                  "Check Compliance of Last Known Configuration",
+                  invokes_alert=True)
+        sel.handle_alert()
+        flash.assert_no_errors()
+        wait_for(
+            lambda: self.compliance_status != original_state,
+            num_sec=timeout, delay=5, fail_func=sel.refresh,
+            message="compliance of {} checked".format(self.name)
+        )
+        return self.compliant
+
+    @property
+    def compliance_status(self):
+        self.summary.reload()
+        return self.summary.compliance.status.value
+
+    @property
+    def compliant(self):
+        """Check if the image is compliant
+
+        Returns:
+            :py:class:`NoneType` if the image was never verified, otherwise :py:class:`bool`
+        """
+        text = self.compliance_status.strip().lower()
+        if text == "never verified":
+            return None
+        elif text.startswith("non-compliant"):
+            return False
+        elif text.startswith("compliant"):
+            return True
+        else:
+            raise ValueError("{} is not a known state for compliance".format(text))
 
 
 @navigator.register(Image, 'All')
--- a/cfme/containers/image.py
+++ b/cfme/containers/image.py
@@ -22,9 +22,9 @@ match_page = partial(match_location, controller='container_image',
 
 class Image(Taggable, SummaryMixin, Navigatable, PolicyProfileAssignable):
 
-    def __init__(self, name, tag, provider, appliance=None):
+    def __init__(self, name, id_, provider, appliance=None):
         self.name = name
-        self.tag = tag
+        self.id = id_
         self.provider = provider
         Navigatable.__init__(self, appliance=appliance)
 
@@ -51,7 +51,7 @@ class Image(Taggable, SummaryMixin, Navigatable, PolicyProfileAssignable):
         sel.handle_alert()
         flash.assert_message_contain('Analysis successfully initiated')
         if wait_for_finish:
-            ssa_timeout = '10M'
+            ssa_timeout = '20M'
             try:
                 tasks.wait_analysis_finished('Container image analysis',
                                              'container', timeout=ssa_timeout)
@@ -59,7 +59,7 @@ class Image(Taggable, SummaryMixin, Navigatable, PolicyProfileAssignable):
                 raise TimedOutError('Timeout exceeded, Waited too much time for SSA to finish ({}).'
                                     .format(ssa_timeout))
 
-    def check_compliance(self, timeout=240):
+    def check_compliance(self, wait_for_finish=True, timeout=240):
         """Initiates compliance check and waits for it to finish."""
         navigate_to(self, 'Details')
         original_state = self.compliance_status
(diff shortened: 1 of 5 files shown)


dajoRH:
Lint report for commit 78b166fb7b8931fecf73ee58904983e4466da780:

`cfme/containers/image.py`:
- :bangbang: Line 101:1: **E303** *too many blank lines (3)*

**No commit flaws detected**.

*CFME QE Bot*

cfme/containers/image.py
```diff
@@ -59,6 +59,43 @@ def perform_smartstate_analysis(self, wait_for_finish=False):
...
+        wait_for(
+            lambda: self.compliance_status != original_state,
+            num_sec=timeout, delay=5, fail_func=sel.refresh,
+            message="compliance of {} checked".format(self.name)
+        )
+        return self.compliant
+
+    @property
+    def compliance_status(self):
+        self.summary.reload()
```
mshriver: Does this reload method work if you're not on the containers image page?

cfme/containers/image.py
```diff
@@ -59,6 +59,43 @@ def perform_smartstate_analysis(self, wait_for_finish=False):
                 raise TimedOutError('Timeout exceeded, Waited too much time for SSA to finish ({}).'
                                     .format(ssa_timeout))
 
+    def check_compliance(self, timeout=240):
+        """Initiates compliance check and waits for it to finish."""
+        original_state = self.compliance_status
+        tb.select('Policy',
```
mshriver: **Required**
There's no navigation here, so you're relying on the self.compliance_status call to navigate for you. I don't think this navigation dependency is a good thing to have, and you should navigate explicitly here to the page where you want to click the toolbar.

cfme/containers/image.py
gshefer: yes. summary reload [is calling to self.load_details](https://github.com/ManageIQ/integration_tests/blob/master/cfme/common/__init__.py#L211), which [is doing the navigation](https://github.com/ManageIQ/integration_tests/blob/master/cfme/containers/image.py#L33).

gshefer:
@mshriver In addition to the tests I added few additional changes & fixes. I did it in new commit so it'll probably be easier.

mshriver:
Bumped back down to 1st level and re-tagged @mkoura since the changes in the new commit weren't trivial.

dajoRH:
Lint report for commit b00a9f8ae8b9db500d98f222dc8d6a8a4e6cd334:

`cfme/tests/containers/test_containers_smartstate_analysis.py`:
- :bangbang: Line 98:17: **E226** *missing whitespace around arithmetic operator*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/containers/test_containers_smartstate_analysis.py
```diff
@@ -68,9 +93,9 @@ def test_containers_smartstate_analysis(provider, test_item, soft_assert):
                     'Traceback: \n{}'.format(format_exc()))
 
     evm_tail.fix_before_start()
-
     image_obj.perform_smartstate_analysis(wait_for_finish=True)
-
+    #  evm_tail.validate_logs()
+    time.sleep(5 * 60)  # temp debug!
```
mshriver: Is there a BZ associated with this? I'd put a TODO or FIXME on it.  In your tests folder, up to you.

cfme/tests/containers/test_containers_smartstate_analysis.py
```diff
@@ -79,6 +104,15 @@ def test_containers_smartstate_analysis(provider, test_item, soft_assert):
         if not soft_assert(hasattr(table, attr),
                 '{} table has missing attribute \'{}\''.format(tbl, attr)):
             continue
-        value = getattr(table, attr).value
+        wait_for_retval = wait_for(lambda: getattr(table, attr).value,
+                                   message='Trying to get attribute "{}" of table "{}"'.format(
+                                       attr, tbl),
+                                   fail_func=image_obj.summary.reload,
+                                   delay=5, num_sec=120, silent_failure=True)
```
mshriver: why use silent_failure here?

cfme/tests/containers/test_containers_smartstate_analysis.py
```diff
@@ -68,9 +93,9 @@ def test_containers_smartstate_analysis(provider, test_item, soft_assert):
                     'Traceback: \n{}'.format(format_exc()))
 
     evm_tail.fix_before_start()
-
     image_obj.perform_smartstate_analysis(wait_for_finish=True)
```
mshriver: This task errored out, causing an exception in the 58z PRT results. please address here if possible.

cfme/containers/image.py
```diff
@@ -59,6 +59,44 @@ def perform_smartstate_analysis(self, wait_for_finish=False):
                 raise TimedOutError('Timeout exceeded, Waited too much time for SSA to finish ({}).'
                                     .format(ssa_timeout))
 
+    def check_compliance(self, timeout=240):
```
mshriver: **Required**
Make the wait optional, match `wait_for_finish` attribute in `perform_smartstate_analysis` above for consistency.

cfme/containers/image.py
```diff
@@ -59,6 +59,44 @@ def perform_smartstate_analysis(self, wait_for_finish=False):
...
+        original_state = self.compliance_status
+        tb.select('Policy',
+                  "Check Compliance of Last Known Configuration",
+                  invokes_alert=True)
+        sel.handle_alert()
+        flash.assert_no_errors()
+        wait_for(
+            lambda: self.compliance_status != original_state,
+            num_sec=timeout, delay=5, fail_func=sel.refresh,
+            message="compliance of {} checked".format(self.name)
```
mshriver: Let's tweak the behavior of this wait_for a bit.

The call to self.compliance_status is calling self.load_details, as you've stated below. This means that there's no need for the fail_func refresh parameter here as its already going to load on each execution of the func.  5 seconds is also potentially cutting it a little close for depending on nested navigation calls. Consider bumping to 10.

I'd also like to see the `compliance_status` and `original_state` included in the log message, but this poses a challenge as we don't have the current value of `compliance_status` stored anywhere that we can access it, and calling it again from `self` will force another navigation and won't give the value that was just compared.

What do you think about changing the message to at least include the original_state, and to be more descriptive of what its doing.  Consider 'compliance state of {} still matches {}'.format(self.name, original_state)

cfme/tests/containers/test_containers_smartstate_analysis.py
gshefer: Because I would like it to continue to the next checks and not just fail here. see the next ~4 lines (soft_assert)

cfme/containers/image.py
```diff
@@ -22,9 +22,9 @@
 
 class Image(Taggable, SummaryMixin, Navigatable, PolicyProfileAssignable):
 
-    def __init__(self, name, tag, provider, appliance=None):
+    def __init__(self, name, id_, provider, appliance=None):
```
izapolsk: honestly, I don't like such parameter names. I'd advice you to use something like image_id instead

cfme/containers/image.py
```diff
@@ -51,19 +51,62 @@ def perform_smartstate_analysis(self, wait_for_finish=False):
...
+        return self.summary.compliance.status.value
+
+    @property
+    def compliant(self):
+        """Check if the image is compliant
+
+        Returns:
+            :py:class:`NoneType` if the image was never verified, otherwise :py:class:`bool`
+        """
+        text = self.compliance_status.strip().lower()
```
izapolsk: it makes sense to move strip call to compliance_status

cfme/tests/containers/test_containers_smartstate_analysis.py
```diff
@@ -47,28 +45,40 @@
...
+def test_check_compliance(provider):
+    chosen_row = navigate_and_get_rows(provider, Image, 1).pop()
+    image_obj = Image(chosen_row.name.text, chosen_row.id.text, provider)
+    image_obj.assign_policy_profiles('OpenSCAP profile')
+    image_obj.check_compliance()
+
+
 @pytest.mark.meta(blockers=[BZ(1382326), BZ(1408255), BZ(1371896),
-                            BZ(1437128, forced_streams=['5.6', '5.7'])])
+                            BZ(1447655, forced_streams=['5.6', '5.7', '5.8', 'upstream'])])
```
izapolsk: you can remove 5.6 since it is out of scope now

cfme/containers/image.py
mshriver: Agreed on requiring this change.

PR #4627, opened 2017-05-04
[1LP][RFR] New bottleneck tests

otsuman:
{{ pytest: cfme/tests/optimize/ -k "test_bottlenecks_summary_time_zone or test_bottlenecks_summary_show_host_events or test_bottlenecks_summary_event_groups"}}

### DIFF
--- a/cfme/tests/optimize/test_bottlenecks.py
+++ b/cfme/tests/optimize/test_bottlenecks.py
@@ -41,7 +41,7 @@ def db_restore(temp_appliance_extended_db):
     db_storage_ssh = SSHClient(hostname=db_storage_hostname, **conf.credentials['bottlenecks'])
     with db_storage_ssh as ssh_client:
         # Different files for different versions
-        ver = "_56" if current_version() < '5.7' else ""
+        ver = "_58" if current_version() > '5.7' else ""
         rand_filename = "/tmp/v2_key_{}".format(fauxfactory.gen_alphanumeric())
         ssh_client.get_file("/home/backups/otsuman_db_bottlenecks/v2_key{}".format(ver),
                             rand_filename)
@@ -105,9 +105,9 @@ def test_bottlenecks_report_time_zome(temp_appliance_extended_db, db_restore, db
         # Compare bottleneck's table timestamp with db
         assert row[0].text == db_row[0][0].strftime("%m/%d/%y %H:%M:%S UTC")
         # Changing time zone
-        view.report.time_zone.fill('(GMT+02:00) Kyiv')
+        view.report.time_zone.fill('(GMT-04:00) La Paz')
         row = view.report.event_details[0]
-        assert row[0].text == (db_row[0][0] - timedelta(hours=4)).strftime("%m/%d/%y %H:%M:%S EEST")
+        assert row[0].text == (db_row[0][0] - timedelta(hours=4)).strftime("%m/%d/%y %H:%M:%S -04")
 
 
 @pytest.mark.tier(2)
@@ -150,7 +150,7 @@ def test_bottlenecks_summary_time_zome(temp_appliance_extended_db, db_restore, d
         # Compare event timestamp with db
         assert events[0].time_stamp == db_row[0][0].strftime("%Y-%m-%d %H:%M:%S UTC")
         # Changing time zone
-        view.report.time_zone.fill('(GMT+02:00) Kyiv')
+        view.summary.time_zone.fill('(GMT-04:00) La Paz')
         events = view.summary.chart.get_events()
         assert events[0].time_stamp == (db_row[0][0] - timedelta(hours=4)).strftime("%Y-%m-%d "
-                                                                                    "%H:%M:%S EEST")
+                                                                                    "%H:%M:%S -04")
--- a/cfme/tests/optimize/test_bottlenecks.py
+++ b/cfme/tests/optimize/test_bottlenecks.py
@@ -6,6 +6,7 @@ from datetime import timedelta
 from cfme.optimize.bottlenecks import Bottlenecks
 from utils import conf
 from utils.appliance.implementations.ui import navigate_to
+from utils.timeutil import parsetime
 from utils.ssh import SSHClient
 
 
@@ -40,7 +41,7 @@ def db_restore(temp_appliance_extended_db):
     db_storage_ssh = SSHClient(hostname=db_storage_hostname, **conf.credentials['bottlenecks'])
     with db_storage_ssh as ssh_client:
         # Different files for different versions
-        ver = "_58" if temp_appliance_extended_db.version > '5.7' else ""
+        ver = "_58" if temp_appliance_extended_db.version > '5.7' else "_57"
         rand_filename = "/tmp/v2_key_{}".format(fauxfactory.gen_alphanumeric())
         ssh_client.get_file("/home/backups/otsuman_db_bottlenecks/v2_key{}".format(ver),
                             rand_filename)
@@ -105,7 +106,7 @@ def test_bottlenecks_report_time_zone(temp_appliance_extended_db, db_restore, db
         # Selecting row by uniq value
         db_row = db_events.filter(db_tbl.message == row[5].text)
         # Compare bottleneck's table timestamp with db
-        assert row[0].text == db_row[0][0].strftime("%m/%d/%y %H:%M:%S UTC")
+        assert row[0].text == db_row[0][0].strftime(parsetime.american_with_utc_format)
         # Changing time zone
         view.report.time_zone.fill('(GMT-04:00) La Paz')
         row = view.report.event_details[0]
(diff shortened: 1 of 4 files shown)


utils/appliance/__init__.py
```diff
@@ -1062,6 +1062,11 @@ def restore_database(self, database_path="/tmp/evm_db.backup", log_callback=None
 
         """
         log_callback('Restoring database')
+        if current_version() > '5.8':
```
psav: We should be checking this against the appliance itself using ```self``` right? ```self.version```

otsuman: done

dajoRH:
*CFME QE Bot*

cfme/tests/optimize/test_bottlenecks.py
```diff
@@ -41,7 +41,7 @@ def db_restore(temp_appliance_extended_db):
     db_storage_ssh = SSHClient(hostname=db_storage_hostname, **conf.credentials['bottlenecks'])
     with db_storage_ssh as ssh_client:
         # Different files for different versions
-        ver = "_56" if current_version() < '5.7' else ""
+        ver = "_58" if current_version() > '5.7' else ""
```
izapolsk: REQUIRED: please get current_version from appliance

cfme/tests/optimize/test_bottlenecks.py
```diff
@@ -107,4 +107,50 @@ def test_bottlenecks_time_zome(temp_appliance_extended_db, db_restore, db_tbl, d
         # Changing time zone
         view.report.time_zone.fill('(GMT-04:00) La Paz')
         row = view.report.event_details[0]
-        assert row[0].text == (db_row[0][0] - timedelta(hours=4)).strftime("%m/%d/%y %H:%M:%S BOT")
+        assert row[0].text == (db_row[0][0] - timedelta(hours=4)).strftime("%m/%d/%y %H:%M:%S -04")
+
+
+@pytest.mark.tier(2)
+def test_bottlenecks_summary_event_groups(temp_appliance_extended_db, db_restore, db_tbl,
+                                          db_events):
```
izapolsk: REQUIRED: please add description to tests

cfme/tests/optimize/test_bottlenecks.py
```diff
@@ -107,4 +107,50 @@ def test_bottlenecks_time_zome(temp_appliance_extended_db, db_restore, db_tbl, d
...
+
+
+@pytest.mark.tier(2)
+def test_bottlenecks_summary_show_host_events(temp_appliance_extended_db, db_restore, db_events):
+    with temp_appliance_extended_db:
+        view = navigate_to(Bottlenecks, 'All')
+        view.summary.show_host_events.fill(False)
+        # Checking that events with value 'Host / Node' absent in table
+        events = view.summary.chart.get_events()
+        assert not sum(1 for event in events if event.type == 'Host')
```
izapolsk: :+1:

cfme/tests/optimize/test_bottlenecks.py
```diff
@@ -107,4 +107,50 @@ def test_bottlenecks_time_zome(temp_appliance_extended_db, db_restore, db_tbl, d
...
+        view.summary.show_host_events.fill(True)
+        events = view.summary.chart.get_events()
+        # Compare number of events in chart with number of rows in db
+        assert len(events) == db_events.count()
```
izapolsk: I'm not sure I understand where you generate events. I don't think it is good idea to rely on presence of some events in appliance db

cfme/tests/optimize/test_bottlenecks.py
```diff
@@ -107,4 +107,50 @@ def test_bottlenecks_time_zome(temp_appliance_extended_db, db_restore, db_tbl, d
...
+
+
+@pytest.mark.tier(2)
+def test_bottlenecks_summary_time_zome(temp_appliance_extended_db, db_restore, db_tbl, db_events):
```
izapolsk: REQUIRED: typo

cfme/tests/optimize/test_bottlenecks.py
```diff
@@ -107,4 +107,50 @@ def test_bottlenecks_time_zome(temp_appliance_extended_db, db_restore, db_tbl, d
...
+    with temp_appliance_extended_db:
+        view = navigate_to(Bottlenecks, 'All')
+        events = view.summary.chart.get_events()
+        # Selecting row by uniq value
+        db_row = db_events.filter(db_tbl.message == events[0].message)
+        # Compare event timestamp with db
+        assert events[0].time_stamp == db_row[0][0].strftime("%Y-%m-%d %H:%M:%S UTC")
```
izapolsk: it makes sense to use variable for 
event = events[0]
db_record_timestamp = db_row[0][0]

utils/appliance/__init__.py
```diff
@@ -1062,6 +1062,12 @@ def restore_database(self, database_path="/tmp/evm_db.backup", log_callback=None
             log_callback(msg)
             raise ApplianceException(msg)
 
+        if self.version > '5.8':
+            status, output = self.ssh_client.run_command("fix_auth --databaseyml -i {}".format(
+                conf.credentials['database'].password), timeout=45)
+            if status != 0:
```
izapolsk: I would use SSHResult.failed here

cfme/tests/optimize/test_bottlenecks.py
otsuman: It takes  ~2 days to gather data. So the best way for checking these controls it to restore db

cfme/tests/optimize/test_bottlenecks.py
otsuman: I think it is readable enough and not frequently used

utils/appliance/__init__.py
otsuman: I don't see a reason whyI should use it here

cfme/tests/optimize/test_bottlenecks.py
otsuman: done

cfme/tests/optimize/test_bottlenecks.py
izapolsk: I don't plan to insist. But consider using variable with meaningful name when you use complex constructions like db_row[0][0] more than 2 times.
here you use for example db_row[0][0] 3 times. 
btw, variable db_row[0][0] doesn't bring any meaning about what is in it. So, when you see it first time in code (assignment statement) it is more or less obvious but later it looses its meaning especially if there are several variables with similar names.

cfme/tests/optimize/test_bottlenecks.py
mshriver: ^ditto

utils/appliance/__init__.py
mshriver: more pythonic, plain english instead of comparators and integers ;)

mshriver:
**Required**
Fix two QuantifiedCode issues, should be real quick and easy ones (docstring and kwargs.get default)

cfme/tests/optimize/test_bottlenecks.py
```diff
@@ -107,4 +109,53 @@ def test_bottlenecks_time_zome(temp_appliance_extended_db, db_restore, db_tbl, d
...
+@pytest.mark.tier(2)
+def test_bottlenecks_summary_time_zone(temp_appliance_extended_db, db_restore, db_tbl, db_events):
+    """ Checks time zone selectbox in summary tab. It should change time zone of events in chart """
+    with temp_appliance_extended_db:
+        view = navigate_to(Bottlenecks, 'All')
+        events = view.summary.chart.get_events()
+        # Selecting row by uniq value
+        db_row = db_events.filter(db_tbl.message == events[0].message)
+        # Compare event timestamp with db
+        assert events[0].time_stamp == db_row[0][0].strftime("%Y-%m-%d %H:%M:%S UTC")
```
mshriver: This looks like ISO with UTC: 
https://github.com/ManageIQ/integration_tests/blob/master/utils/timeutil.py#L19

import parsetime and use that instead.  Doesn't really help for the test above unfortunately, where you're working with a different offset not in this util.

widgetastic_manageiq.py
```diff
@@ -1589,8 +1589,9 @@ def __repr__(self):
             params = ", ".join(["{}={}".format(attr, getattr(self, attr)) for attr in attrs])
             return "TimelinesEvent({})".format(params)
 
-    def __init__(self, parent, logger=None):
+    def __init__(self, parent, locator=None, logger=None):
         super(TimelinesChart, self).__init__(parent=parent, logger=logger)
+        self.locator = locator or '//div[contains(@class, "timeline-container")]'
```
mshriver: **Required**

It looks like the whole reason you've made this change is to support adding the chart with a `timeline-pf-chart` div.
(https://github.com/ManageIQ/integration_tests/pull/4627/files#diff-e25760b09cb1951870a11138d80a45f3R34)


It seems this chart element is patternfly, which means this change isn't needed for manageiq:
https://github.com/patternfly/patternfly-timeline/blob/27f0a72733238ebebf8ab69a5803be2ad7e6441c/src/timeline.js#L57

Unfortunately this is a major change if I'm right.

@mfalesni @izapolsk Thoughts? This should be added to patternfly and we need to bump that version before accepting this PR.

dajoRH:
*CFME QE Bot*

dajoRH:
*CFME QE Bot*

widgetastic_manageiq.py
mfalesni: @mshriver @otsuman to save time, I would personally keep it here for now and then backport it to PF, then remove and bump ...

PR #4628, opened 2017-05-04
[1LP][RFR] Fixed SSH closing on one test

psav:
Close of SSH on finishing of test

#### mshriver: limiting PRT to the test that uses the ssh fixture
{{ pytest: cfme/tests/control/test_alerts.py -k  test_alert_timeline_cpu --long-running --use-provider vsphere6-nested }}


mshriver:
test setup & teardown passed using this modified yield_fixture, even though the test itself failed - not related to the SSH client close.  The test did run past the point where the ssh client fixture was used. Calling it good.

PR #4643, opened 2017-05-09
[1LP][RFR] Fix the coverage hook setup

mfalesni:
The miq/lib/ directory was removed, making the installation of coverage hook fail.

Moved the installation into vmdb/lib/ and modified the preinitializer to require_relative.

{{pytest: -v -m smoke --ui-coverage}}

### DIFF
--- a/utils/ssh.py
+++ b/utils/ssh.py
@@ -309,13 +309,13 @@ class SSHClient(paramiko.SSHClient):
 
     def run_rails_command(self, command, timeout=RUNCMD_TIMEOUT, **kwargs):
         logger.info("Running rails command %r", command)
-        return self.run_command('/var/www/miq/vmdb/bin/rails runner {command}'.format(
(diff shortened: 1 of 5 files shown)


psav:
Looks good to me

mfalesni:
Initiated a retest with the hook enabled though I tested it locally.

utils/ssh.py
```diff
@@ -309,13 +309,13 @@ def cpu_spike(self, seconds=60, cpus=2, **kwargs):
 
     def run_rails_command(self, command, timeout=RUNCMD_TIMEOUT, **kwargs):
         logger.info("Running rails command %r", command)
-        return self.run_command('/var/www/miq/vmdb/bin/rails runner {command}'.format(
+        return self.run_command('cd /var/www/miq/vmdb; bin/rails runner {command}'.format(
```
mfalesni: This is required as when the absolute path is used it will not take the installed gem into the account.

PR #4649, opened 2017-05-10
[RFR][1LP] Adds extra jenkins info to Ostriz

psav:
Gives ostriz extra jenkins data


mfalesni:
Do you think #4574 could go in as well?

PR #4652, opened 2017-05-10
[1LP][RFR] remove rest_api fixture from test_rest_services

mkoura:
Replacing the rest_api fixture with appliance.rest_api in test_rest_services.

{{pytest: -v --long-running cfme/tests/services/test_rest_services.py}}

### DIFF
--- a/cfme/tests/services/test_rest_services.py
+++ b/cfme/tests/services/test_rest_services.py
@@ -72,20 +72,20 @@ def dialog():
 
 
 @pytest.fixture(scope="function")
-def service_catalogs(request, rest_api):
-    response = _service_catalogs(request, rest_api)
-    assert rest_api.response.status_code == 200
+def service_catalogs(request, appliance):
+    response = _service_catalogs(request, appliance.rest_api)
+    assert appliance.rest_api.response.status_code == 200
     return response
 
 
 @pytest.fixture(scope="function")
-def services(request, rest_api, a_provider):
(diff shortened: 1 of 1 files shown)


mkoura:
PRT: one timeout on 5.8, all other passed.

mkoura:
@izapolsk It's timeout issue, I think it's safe to merge. Other variant of this test  (`TestServiceRESTAPI::test_power_service[from_collection]`) succeeded, so I wouldn't increase timeouts further atm. The test is cycling between service power states (on -> off -> on -> suspended). It's maybe too complex, I'll consider splitting it to several tests testing just one transition at a time (in other PR).

cfme/tests/services/test_rest_services.py
```diff
@@ -40,7 +40,7 @@ def a_provider(request):
 def wait_for_vm_power_state(vm, resulting_state):
     wait_for(
         lambda: vm.power_state == resulting_state,
-        num_sec=600, delay=20, fail_func=vm.reload,
+        num_sec=1200, delay=45, fail_func=vm.reload,
```
mshriver: **Optional**
Keep the delay lower, save 25 seconds sometimes! Up to you, I don't see much of a problem checking with the same frequency for longer.

mshriver:
Single timeout failure in 58z, upstream failure because of flash message changes in catalog item creation, not part of this PR.

PR #4657, opened 2017-05-10
[1LP][RFR] Update scripts, close ssh clients after use

mshriver:
Still need to test some of this. Not sure what PRT will look like, some of these scripts are used by the test runner.

Pretty simple changes. @rsnyman check it out, should complement your PRs #4633 and #4634 

All of these scripts have been tested locally

### DIFF
--- a/scripts/clean_appliance.py
+++ b/scripts/clean_appliance.py
@@ -22,13 +22,13 @@ from utils.ssh import SSHClient
 
 def main():
     parser = argparse.ArgumentParser(epilog=__doc__,
-        formatter_class=argparse.RawDescriptionHelpFormatter)
+                                     formatter_class=argparse.RawDescriptionHelpFormatter)
     parser.add_argument('hostname', nargs='?', default=None,
-        help='hostname or ip address of target appliance')
+                        help='hostname or ip address of target appliance')
     parser.add_argument('username', nargs='?', default=credentials['ssh']['username'],
-        help='SSH username for target appliance')
+                        help='SSH username for target appliance')
     parser.add_argument('password', nargs='?', default=credentials['ssh']['password'],
-        help='SSH password for target appliance')
+                        help='SSH password for target appliance')
 
     args = parser.parse_args()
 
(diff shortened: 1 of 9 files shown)


scripts/clean_appliance.py
```diff
@@ -22,13 +22,13 @@
 
 def main():
     parser = argparse.ArgumentParser(epilog=__doc__,
-        formatter_class=argparse.RawDescriptionHelpFormatter)
+                                     formatter_class=argparse.RawDescriptionHelpFormatter)
```
mshriver: this is all for lint

scripts/template_upload_vsphere.py
```diff
@@ -78,7 +78,7 @@ def upload_ova(hostname, username, password, name, datastore,
                cluster, datacenter, url, provider, proxy,
                ovf_tool_client, default_user, default_pass):
 
-    cmd_args = ['ovftool --noSSLVerify']
+    cmd_args = list('ovftool --noSSLVerify')
```
mkoura: Why is this better? Just curious.

mshriver: Initially I thought it was a PEP standard but I went back and checked, its not. Just a suggestion pycharm gives.  I see both styles throughout the codebase. Personally I guess I kinda like this style more.

mfalesni: This will turn it into:
```python
['o', 'v', 'f', 't', ....]
```

mfalesni: IPython says:
```
In [1]: list('ovftool --noSSLVerify')
Out[1]: 
['o',
 'v',
 'f',
 't',
 'o',
 'o',
 'l',
 ' ',
 '-',
 '-',
 'n',
 'o',
 'S',
 'S',
 'L',
 'V',
 'e',
 'r',
 'i',
 'f',
 'y']
```

mshriver: dammit. fixed.

scripts/get_unregistered_vmware_files.py
```diff
@@ -34,55 +34,53 @@ def list_orphaned_files_per_host(host_name, host_datastore_urls, provider_key, v
...
-                    if file_extension:
-                        file_type = file_extension[0]
-                        number_of_files = 1
-                    if int(output.strip()) == 0:
-                        command = 'ls ~/{}/{} | wc -l'.format(datastore_path[0], fil)
+        with SSHClient(**connect_kwargs) as ssh_client:
+            for datastore_url in host_datastore_urls:
+                datastore_path = re.findall(r'([^ds:`/*].*)', str(datastore_url))
+
+                command = 'ls ~/{}'.format(datastore_path[0])
```
mkoura: OPTIONAL: I would explicitly use `ls -1` just to be on the safe side. But I guess it's safe to assume that ls will always correctly detect that the standard output is not terminal and the output will be single-column.

scripts/template_upload_ec2.py
```diff
@@ -124,53 +124,54 @@ def create_image(template_name, image_description, bucket_name, key_name):
...
+        print("AMAZON EC2: Creating JSON file beofre importing the image ...")
+        upload_json = """[
+      {{
+        "Description": "{description}",
+        "Format": "vhd",
+        "UserBucket": {{
+            "S3Bucket": "{bucket_name}",
+            "S3Key": "{key_name}"
+        }}
+    }}]""".format(description=image_description, bucket_name=bucket_name,
```
psav: Anyway we can format this better? It doesn't look good like this.

mshriver: Added TODO above the block, agree but would rather not address here.

scripts/get_unregistered_vmware_files.py
mshriver: noted, going to leave off for now though. keeping the scope small and focused.

mfalesni: We might have a problem with this one on non-root sudo user access as we use pseudo-tty in that case. But I guess we don't care that much about it now.

PR #4672, opened 2017-05-12
[RFR] Composite Source selector

psav:
Can now choose single source for uncollect

### DIFF
--- a/utils/trackerbot.py
+++ b/utils/trackerbot.py
@@ -307,10 +307,11 @@ def depaginate(api, result):
     }
 
 
-def composite_uncollect(build):
+def composite_uncollect(build, source='jenkins'):
     """Composite build function"""
     since = env.get('ts', time.time())
-    url = "{}?build={}&source=jenkins&since={}".format(conf['ostriz'], urllib.quote(build), since)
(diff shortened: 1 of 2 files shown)


utils/trackerbot.py
```diff
@@ -307,10 +307,13 @@ def depaginate(api, result):
     }
 
 
-def composite_uncollect(build):
+def composite_uncollect(build, source=None):
     """Composite build function"""
+    if not source:
+        source = 'jenkins'
```
mfalesni: String is immutable, so it will be perfectly fine to put it as a default value in the function definition.

utils/trackerbot.py
```diff
@@ -307,10 +307,13 @@ def depaginate(api, result):
...
     since = env.get('ts', time.time())
-    url = "{}?build={}&source=jenkins&since={}".format(conf['ostriz'], urllib.quote(build), since)
+    url = "{0}?build={1}&source={2}&since={3}".format(
+        conf['ostriz'], urllib.quote(build), urllib.quote(source), since)
```
mfalesni: I know it was not before but whouldn't `since` be escaped as well?

PR #4674, opened 2017-05-12
[RFR]Block SCVMM providers from provider_crud due to BZ

jteehan:
Purpose or Intent
=================
__Updating tests__ to block provider_crud for scvmm while bug is being fixed.  CFME is returning temporary templates in addition to real templates.
https://bugzilla.redhat.com/show_bug.cgi?id=1450527

### DIFF
--- a/cfme/tests/infrastructure/test_providers.py
+++ b/cfme/tests/infrastructure/test_providers.py
@@ -210,6 +210,7 @@ def test_provider_add_with_bad_credentials(provider):
 @pytest.mark.usefixtures('has_no_infra_providers')
 @pytest.mark.tier(1)
 @test_requirements.provider_discovery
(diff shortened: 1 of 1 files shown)


mshriver:
@jteehan good find and well done getting a quick working update.

There was a fairly recent framework update that made this check even more robust - next time, check out provider.one_of() for such situations where you'd like verify a provider type is (or isn't) a certain type.

https://github.com/ManageIQ/integration_tests/blob/master/cfme/common/provider.py#L521

It uses the provider class instead of its string `type` attribute. You may have to import the class that you're checking against [SCVMMProvider](https://github.com/ManageIQ/integration_tests/blob/master/cfme/infrastructure/provider/scvmm.py#L5) to call it, if not already imported to your module.

PR #4698, opened 2017-05-17
[RFR][1LP] UI coverage, fix one problem

mfalesni:
A file was not committed and therefore it did not work properly

{{pytest: -v -m smoke --ui-coverage}}


mfalesni:
This looks good, logs look like they should look.

PR #4700, opened 2017-05-17
[1LP][RFR] Enhancement to ssh

psav:
* You can already do: if res, but it is not as explicit as if
  res.success or res.failed


utils/ssh.py
```diff
@@ -75,6 +75,14 @@ def __cmp__(self, other):
         else:
             raise ValueError('You can only compare SSHResult with str or int')
 
+    @property
+    def success(self):
+        return self.rc == 0
+
+    @property
+    def failed(self):
+        return self.rc != 0
```
mfalesni: Could have been not self.success but that would be too dry :D

PR #4702, opened 2017-05-17
[RFR] Fixed the uncollect again

psav:
* getoption in Config in pytest, will not return the "notset" on a
  store optin if a default is set


dajoRH:
*CFME QE Bot*

PR #4705, opened 2017-05-18
[WIP]Upgrade CFME script

shalomnaim1:
# Upgrade CFME script
## summary

one of our testing and everyday task is upgrading CFME version on active appliance.
The script I wrote ruunig system upgrade between CFME version (for example from 5.7 to 5.8)

After this code will merged I planing to use it on my automation test of CFME upgrade task

## Upgrade step in short:
1.stop CFME engine 
2. setting up yum repos with the relevant version of CFME
3. signup to subscription-mameger
4. execute yum update which download and install the new CFME version
5. restarting so components of the system such as the DB and more
6. start CFME engine

## Script usage:
the script get YAML file path with the following key:
1. address: appliance IP 
2. username: user name for the subscription manager
3. password:  password for the subscription manager
4. skip_validation: skip version salivation after upgrade finish (boolean, if not exist assuming validation needed)
5. dest_version: destination version (used for after script's version validation, not mandatory if skip_validation is true
6. repo: list of yum repos with the required version rpms


dajoRH:
Lint report for commit 2d442e5fff427f5afbe7f4fcf2bb8d2beb355ac0:

`scripts/upgrade_cfme.py`:
- :red_circle: Line 24:7: **N801** *class names should use CapWords convention*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
Lint report for commit d476a0277a2ca687c56a6444e3d8bb007ef916e7:

`scripts/upgrade_cfme.py`:
- :red_circle: Line 26:7: **N801** *class names should use CapWords convention*

**No commit flaws detected**.

*CFME QE Bot*

scripts/upgrade_cfme.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+class CfmeUpgradeManeger(IPAppliance):
+
+    def __init__(self, appliance_ip, repo_list, dest_version):
+        super(CfmeUpgradeManeger, self).__init__(address=appliance_ip)
+        self.repo_list = repo_list
+        self.dest_version = dest_version
+
+    def exec_commaned(self, cmd, expect_failure=False, ignore_failure=False):
```
jaryn: typo: `commaned`

scripts/upgrade_cfme.py
```diff
@@ -0,0 +1,216 @@
...
+        '''
+        execute command on the appliance and logging the command with it output to log file
+        :param cmd: command
+        :param expect_failure: is the command have to fail
+        :param ignore_failure: not raise exception on command failure
+        :return: command results
+        '''
+
+        logger.debug("exec_commaned: {cmd}".format(cmd=cmd))
```
jaryn: You were mangling with the messages formatting. Why not to use `%(funcName)s` then?

scripts/upgrade_cfme.py
```diff
@@ -0,0 +1,216 @@
...
+        result = self.ssh_client.run_command(cmd)
+        try:
+            if bool(result.rc) != expect_failure and not ignore_failure:
+                exception_msg = "command {cmd} failed!!\n\n\n" \
```
jaryn: When command fails and it is expected to fail, then you will shout "command failed". I find this confusing.

scripts/upgrade_cfme.py
```diff
@@ -0,0 +1,216 @@
...
+                                "{command_output}".format(cmd=cmd,
+                                                          command_output=result.output)
+                raise RuntimeError(exception_msg)
```
jaryn: https://docs.python.org/2/library/exceptions.html#exceptions.RuntimeWarning

I might be wrong but I think this is invalid use of `RuntimeError`. Just use Exception or make
`class CommandFailedError(Exception)`

scripts/upgrade_cfme.py
```diff
@@ -0,0 +1,216 @@
...
+
+    def stop_cfme(self):
+        '''
+        stop CFME engine process
+        :return:
+        '''
+        logger.info("stopping the cfme engine (evmserverd)")
+
+        result = self.exec_commaned("systemctl status evmserverd | "
+                                    "grep Active: | awk {'print $2'}", ignore_failure=True)
```
jaryn: `awk '/Active:/ {print $2}'` to save one extra call.

scripts/upgrade_cfme.py
jaryn: I also think single `"\n"` is well enough if not too much and I like `(``)` rather then `\`:
http://stackoverflow.com/a/1874679

scripts/upgrade_cfme.py
```diff
@@ -0,0 +1,216 @@
...
+        logger.debug("exec_commaned: {cmd}".format(cmd=cmd))
+        result = self.ssh_client.run_command(cmd)
+        try:
+            if bool(result.rc) != expect_failure and not ignore_failure:
+                exception_msg = "command {cmd} failed!!\n\n\n" \
+                                "{command_output}".format(cmd=cmd,
+                                                          command_output=result.output)
+                raise RuntimeError(exception_msg)
+        except Exception as ex:
+            logger.exception(ex.message)
```
jaryn: logger.exception("It broke") will log the exception class and the formatted exception. So your code is producing doubled log lines which is no good.

```
In [5]: try: raise Exception("It was wind")
except Exception:
     logging.exception("It fell!")
   ...:     
ERROR:root:It fell!
  File "<ipython-input-5-8a089cb33723>", line 1, in <module>
    try: raise Exception("It was wind")
Exception: It was wind
```

scripts/upgrade_cfme.py
```diff
@@ -0,0 +1,216 @@
...
+        :return: None
+        '''
+
+        logger.info("adding repositories to yum")
+        for curr_repo_url in self.repo_list:
+            # Add all the repos to yum repo
+            self.exec_commaned("yum-config-manager "
+                               "--add-repo {repo_url}".format(repo_url=curr_repo_url))
+
+    def update_yum(self):
```
jaryn: misleading name

scripts/upgrade_cfme.py
```diff
@@ -0,0 +1,216 @@
...
+            res = self.exec_commaned(""" ps -ef | grep "yum update" | grep -v grep """,
+                                     ignore_failure=True)
+            while not bool(res.rc):
+                logger.info("yum still running")
+                time.sleep(10)
+                res = self.exec_commaned(""" ps -ef | grep "yum update" | grep -v grep """,
+                                         ignore_failure=True)
+
+        # start thr monitoring thread
+        monitor = threading.Thread(target=monitor_yum)
```
jaryn: You are sure you need threads? I found that it is good to try to avoid them as they are rising the code complexity so much. Shouldn't you set `daemon=True`? And/or shouldn't you join the thread at some point? Or you just start start `yum` and just after that restart and have hopeful fingers crossed your binaries got updated before the service restarted?

Threads are advanced stuff. Really.

scripts/upgrade_cfme.py
```diff
@@ -0,0 +1,216 @@
...
+                           "--force".format(username=username, password=password))
+
+    def validate_cfme_version(self):
+        '''
+        Validating if the cfme version  is fit
+        :return:
+        '''
+        logger.info("waiting for 120 seconds for main process to load "
+                    "before running version validation")
+        time.sleep(120)
```
jaryn: 120 seconds should be enough for anybody (:

scripts/upgrade_cfme.py
```diff
@@ -0,0 +1,216 @@
...
+        logger.info("start CFME version validation")
+        api_url = "https://{ip}/api/".format(ip=self.address)
+        auth = ("admin", "smartvm")
```
jaryn: hardcoded passwords?

scripts/upgrade_cfme.py
```diff
@@ -0,0 +1,216 @@
...
+                                "is_mandatory": True, "default": None}}
+    for key in requierd_fildes:
+        if key not in params:
+            try:
+                if requierd_fildes[key]["is_mandatory"]:
+                    raise RuntimeError(requierd_fildes.get(key).get("msg"))
+            except Exception as ex:
+                logger.exception(ex.message)
+            else:
+                params.update({key: requierd_fildes.get(key).get("default")})
```
jaryn: Isn't this equivalent to
`params[key] = requierd_fildes.get(key).get("default")`
?

scripts/upgrade_cfme.py
```diff
@@ -0,0 +1,216 @@
...
+    params = {}
+    try:
+        logger.info("loading config file")
+        with open(args.config, 'r') as stream:
+            params = yaml.load(stream)
+    except Exception as ex:
+        logger.exception(ex.message)
+
+    # filed name: (error msg, is mandatory, default value)
+    requierd_fildes = {"address": {"msg": "Appliance address is missing",
```
mfalesni: Typo: required_fields.

scripts/upgrade_cfme.py
mfalesni: Don't you want to use wait_for and wait until the api starts responding instead?

scripts/upgrade_cfme.py
```diff
@@ -0,0 +1,216 @@
...
+        self.dest_version = dest_version
+
+    def exec_commaned(self, cmd, expect_failure=False, ignore_failure=False):
+        '''
+        execute command on the appliance and logging the command with it output to log file
+        :param cmd: command
+        :param expect_failure: is the command have to fail
+        :param ignore_failure: not raise exception on command failure
+        :return: command results
+        '''
```
mfalesni: We use google style comments: https://google.github.io/styleguide/pyguide.html#Comments

scripts/upgrade_cfme.py
```diff
@@ -0,0 +1,216 @@
...
+
+            time.sleep(5)
+            # do as long yum is running on the appliance
+            res = self.exec_commaned(""" ps -ef | grep "yum update" | grep -v grep """,
+                                     ignore_failure=True)
+            while not bool(res.rc):
+                logger.info("yum still running")
+                time.sleep(10)
+                res = self.exec_commaned(""" ps -ef | grep "yum update" | grep -v grep """,
+                                         ignore_failure=True)
```
mfalesni: We use wait_for for these purposes.

mfalesni:
IPAppliance already has these methods that I believe would be idal candidates for using and not reinventing wheel:

- update_rhel
- stop_evm_service
- start_evm_service
- restart_evm_service
- is_evm_service_running

We do not use RHSM as we receive links to repos themselves on the download server (as like when the new release comes out, you can see all the links in the e-mail).

I am not sure about practical usability of automated approach as for each combination of from/to versions there may be a manual step involved and each update must be done first and documented so any intermediate steps can be performed next time.

@lcouzens am I right?

lcouzens:
Mfalesni is correct, we don't bother registering/subscribing to RHSM we just use links to repos themselves. Also for the most part the upgrade process should be the same but there is often additional steps such as postgres upgrades etc.

scripts/upgrade_cfme.py
epacific1: self.exec_commaned typo?
self.exec_command

PR #4710, opened 2017-05-18
[1LP][RFR] Fix tenant_crud test

okolisny:
Added support for 5.8

{{ pytest: cfme/tests/cloud/test_tenant.py --use-provider rhos7-ga }}

### DIFF
--- a/cfme/cloud/tenant.py
+++ b/cfme/cloud/tenant.py
@@ -74,7 +74,7 @@ class Tenant(Navigatable):
 
     def wait_for_appear(self, timeout=600):
         return wait_for(self.exists, timeout=timeout, message='Wait for cloud tenant to appear',
-                        delay=10)
+                        delay=10, fail_func=sel.refresh)
 
     def update(self, updates, wait=True):
         navigate_to(self, 'Edit')
--- a/cfme/cloud/tenant.py
+++ b/cfme/cloud/tenant.py
@@ -53,7 +53,7 @@ class Tenant(Navigatable):
         sel.click(form_buttons.cancel if cancel else create_tenant_form.save_button)
 
         if cancel:
-            return flash.assert_success_message('Add of new Cloud Tenant was cancelled by the user')
+            return flash.assert_no_errors()
         else:
             return flash.assert_success_message('Cloud Tenant "{}" created'.format(self.name))
 
--- a/cfme/cloud/tenant.py
+++ b/cfme/cloud/tenant.py
@@ -74,7 +74,7 @@ class Tenant(Navigatable):
 
     def wait_for_appear(self, timeout=600):
         return wait_for(self.exists, timeout=timeout, message='Wait for cloud tenant to appear',
-                        delay=10, fail_func=sel.refresh)
+                        delay=10)
 
     def update(self, updates, wait=True):
         navigate_to(self, 'Edit')
--- a/cfme/cloud/tenant.py
+++ b/cfme/cloud/tenant.py
@@ -14,6 +14,7 @@ from cfme.fixtures import pytest_selenium as sel
 from cfme.web_ui import (
     PagedTable, CheckboxTable, toolbar as tb, match_location, Form, AngularSelect, Input,
     form_buttons, flash, paginator)
+from utils import version
 from utils.appliance import Navigatable
 from utils.appliance.implementations.ui import CFMENavigateStep, navigator, navigate_to
 from utils.log import logger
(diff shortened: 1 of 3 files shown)


okolisny:
I don't know why tests in PRT were not run. Tested on local 5.7 and 5.8

mkoura:
please try to use different provider for PRT. rhos7-ga maybe?

okolisny:
@mkoura Test "Skipped". As I understand it demands some special tag in yaml in order to run it

mshriver:
@okolisny please ensure all commits are signed.

cfme/cloud/tenant.py
```diff
@@ -68,10 +72,10 @@ def wait_for_disappear(self, timeout=300):
         except TimedOutError:
             logger.error('Timed out waiting for tenant to disappear, continuing')
 
-    # def wait_for(self, timeout=600):
-    #     return wait_for(self.exists, fail_condition=False, timeout=timeout,
-    #                     message='Wait for cloud tenant to appear', delay=10,
-    #                     fail_func=sel.refresh)
+    def wait_for_appear(self, timeout=600):
+        return wait_for(self.exists, fail_condition=False, timeout=timeout,
```
mshriver: **Required**
Drop `fail_condition` and `fail_func` parameters. fail_condition is False by default and self.exists calls navigate, which refreshes the page. This fail_func refresh would be executed right before the navigation that occurs during self.exists anyway.

cfme/tests/cloud/test_tenant.py
```diff
@@ -37,7 +37,7 @@ def test_tenant_crud(tenant):
     assert not tenant.exists()
 
     tenant.create()
-    assert tenant.exists()
+    tenant.wait_for_appear()
```
mshriver: Consider keeping `assert tenant.exists()` after the wait_for_appear.

cfme/cloud/tenant.py
```diff
@@ -17,13 +17,15 @@
...
+from utils import version
 from utils.wait import wait_for, TimedOutError
 
 create_tenant_form = Form(
     fields=[
         ('prov_select', AngularSelect("ems_id")),
         ('name', Input('name')),
-        ('save_button', form_buttons.angular_save),
+        ('save_button', version.pick({version.LOWEST: form_buttons.angular_save,
+                                      '5.8': form_buttons.simple_save})),
```
mfalesni: remove version.pick, Form knows how to handle dict as version pick itself.

okolisny:
@mshriver I returned fail_func for wait_for, because self.exists does not navigate. As I understand, am_i_here method of TenantAll returns True each time

mshriver:
@okolisny you're correct - the am_i_here will return true.  BUT! :)  The resetter() is executed even when am_i_here() is true, unless `use_resetter=False` is explicitly passed to the navigate_to call. So it'll get refreshed even if its already on that page. Make sense?  Check out here for the navigation cycle:
https://github.com/ManageIQ/integration_tests/blob/master/utils/appliance/implementations/ui.py#L368

cfme/cloud/tenant.py
```diff
@@ -68,10 +72,9 @@ def wait_for_disappear(self, timeout=300):
         except TimedOutError:
             logger.error('Timed out waiting for tenant to disappear, continuing')
 
-    # def wait_for(self, timeout=600):
-    #     return wait_for(self.exists, fail_condition=False, timeout=timeout,
-    #                     message='Wait for cloud tenant to appear', delay=10,
-    #                     fail_func=sel.refresh)
+    def wait_for_appear(self, timeout=600):
+        return wait_for(self.exists, timeout=timeout, message='Wait for cloud tenant to appear',
+                        delay=10, fail_func=sel.refresh)
```
mshriver: Still convincing you to remove the fail_func call ;)

okolisny: I understood you, and it must work as you said, but it doesn't. I don't know the reason, but what I mentioned - it doesn't refresh the page within self.exists().
Here is the part of [cfme.log](http://pastebin.test.redhat.com/486907), started from wait_for_appear call, without fail_func param. As I understand, ressetter() for Tenant/Details was not called.

mshriver: Discussed offline - the 'All' destination needs a resetter method for this to work as I was pretending it did.

okolisny:
@mshriver added resetter for 'All' destination, deleted fail_func param inside wait_for_appear
full [cfme.log](http://pastebin.test.redhat.com/486929) of local re-run

mshriver:
Will merge after commits cleaned up a bit, two unsigned in the history

PR #4713, opened 2017-05-18
[1LP][RFR] Removed bz and changed common test name

sshveta:
{{pytest: -v --long-running --use-provider ansible cfme/tests/services/test_config_provider_servicecatalogs.py }}

Changed common  test case name and removed bz flag

### DIFF
--- a/cfme/infrastructure/config_management.py
+++ b/cfme/infrastructure/config_management.py
@@ -48,6 +48,7 @@ page = Region(locators={
 add_manager_btn = form_buttons.FormButton('Add')
 edit_manager_btn = form_buttons.FormButton('Save changes')
 cfg_btn = partial(tb.select, 'Configuration')
+RELOAD_LOC = "//button[@name='summary_reload']"
 
 
 match_page = partial(match_location, controller='provider_foreman',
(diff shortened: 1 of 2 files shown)


cfme/infrastructure/config_management.py
```diff
@@ -226,6 +229,8 @@ def _does_profile_exist(self):
     def config_profiles(self):
         """Returns 'ConfigProfile' configuration profiles (hostgroups) available on this manager"""
         navigate_to(self, 'Details')
+        # workaround for BZ 1452425
```
mshriver: **Required**
Make this a TODO so we can find and remove it easier.

cfme/tests/services/test_config_provider_servicecatalogs.py
```diff
@@ -49,20 +49,20 @@ def config_manager(config_manager_obj):
...
     item_name = dialog.element_data.get("default_text_box")
     catalog_item = CatalogItem(item_type=item_type,
                                name=item_name,
                                description="my catalog",
                                display_in=True,
                                catalog=catalog,
                                dialog=dialog,
-                               provider=provider,
+                               provider=version.pick({
+                                   '5.7': provider_name + ' Configuration Manager',
```
mshriver: Use a consistent syntax, this is the same exact message as printed in the quad_name property above, I would stick with the string.format() syntax.

If the 'name' attribute isn't found, `provider_name` will be None, and this will raise exception for invalid operand

mshriver:
Commit is unverified, @sshveta please ensure the commit is signed and the same key is added to your GH account.

PR #4721, opened 2017-05-19
[1LP][RFR] new REST API test for vms service subcollection; removing support for versions < 5.7

mkoura:
* new test checking that VM added to service is also accessible using VMs subcollection.

Tested locally. This test depends on manageiq-client==0.3.0 which is in requirements/frozen.txt, but for some reason is not installed on PRT. It seems requirements are not being installed at all.

* also removing support for versions < 5.7

{{pytest: -v -k test_service_vm_subcollection --long-running}}

### DIFF
--- a/cfme/tests/services/test_rest_services.py
+++ b/cfme/tests/services/test_rest_services.py
@@ -323,6 +323,17 @@ class TestServiceRESTAPI(object):
         _action_and_check('start', 'on')
 
     @pytest.mark.uncollectif(lambda: version.current_version() < '5.7')
+    def test_service_vm_subcollection(self, appliance, service_data):
+        """Tests /api/services/:id/vms.
+
+        Metadata:
+            test_flag: rest
+        """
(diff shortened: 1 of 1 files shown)


cfme/tests/services/test_rest_services.py
```diff
@@ -323,6 +323,17 @@ def _action_and_check(action, resulting_state):
...
+        """Tests /api/services/:id/vms.
+
+        Metadata:
+            test_flag: rest
+        """
+        service = appliance.rest_api.collections.services.get(name=service_data['service_name'])
+        vm = appliance.rest_api.collections.vms.get(name=service_data['vm_name'])
+        assert service.vms[0].id == vm.id
+
+    @pytest.mark.uncollectif(lambda: version.current_version() < '5.7')
```
izapolsk: REQUIRED: @mkoura, as far as I know testing of cfme less than 5.7 has been dropped. So please remove these uncollects.

PR #4726, opened 2017-05-22
[RFR][1LP] Update scripts/clone_template, handle deploy failures

mshriver:
More cleanly handle the case where a VM deploy fails explicitly or
silently.

Cleanup some CQ .get('thing', None) cases to .get('thing')

FIXES RHCFQE-2825

### DIFF
--- a/scripts/clone_template.py
+++ b/scripts/clone_template.py
@@ -83,7 +83,7 @@ def destroy_vm(provider_mgmt, vm_name):
 
 def main(**kwargs):
     # get_mgmt validates, since it will explode without an existing key or type
-    if kwargs.get('deploy', None):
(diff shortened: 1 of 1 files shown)


scripts/clone_template.py
```diff
@@ -178,11 +178,15 @@ def main(**kwargs):
         provider.deploy_template(**deploy_args)
     except Exception as e:
         logger.exception(e)
-        logger.error('Clone failed')
-        if kwargs.get('cleanup', None):
+        logger.error('provider.deploy_template failed')
```
mshriver: Here and the new block below are the core of the PR, everything else is CQ cleanup.

PR #4740, opened 2017-05-25
[1LP][RFR] Fix delete canned saved reports 

niyazRedhat:
This Fix deletes saved canned reports.
Due to parsetime it was failing.
{{pytest: --use-provider=vsphere55 cfme/tests/intelligence/reports/test_views.py}}

### DIFF
--- a/cfme/intelligence/reports/reports.py
+++ b/cfme/intelligence/reports/reports.py
@@ -421,7 +421,7 @@ class CannedSavedReport(CustomSavedReport, Navigatable):
 
     def delete(self):
         navigate_to(self, 'Saved')
-        self.saved_table.select_row(header='Run At', value=self.datetime_in_tree)
+        self.saved_table.select_row(header='Run At', value=self.datetime)
         cfg_btn("Delete this Saved Report from the Database", invokes_alert=True)
         sel.handle_alert()
         flash.assert_no_errors()


cfme/intelligence/reports/reports.py
```diff
@@ -367,6 +367,9 @@ def __init__(self, path_to_report, datetime, candu=False, appliance=None):
         self.candu = candu
         self.datetime_in_tree = version.pick({"5.6": self.datetime,
                         "5.7": parsetime.from_american_with_utc(self.datetime).to_iso_with_utc()})
+        self.datetime_in_tree_delete = version.pick({"5.6": self.datetime,
+                        "5.7": parsetime.from_american_with_utc(
+                            self.datetime).to_american_with_utc()})
```
mkoura: Why is it necessary to convert `from_american_with_utc` -> `to_american_with_utc`? What difference does it make to  `self.datetime`?
Why does the datetime differs for delete? It's the same tree, right?

mkoura:
Looking at `cfme/tests/intelligence/reports/test_views.py` are you sure the change is really tested there? The only use of `report.delete()` there is guarded by try..catch.

cfme/intelligence/reports/reports.py
niyazRedhat: When i used `from_american_with_utc(self.datetime).to_iso_with_utc` for delete reports its failing.
because of it unable to find proper date time format on the tree to select the report. That's why I am converting `from_american_with_utc -> to_american_with_utc`.

PR #4742, opened 2017-05-25
Renaming timelines test cases

izapolsk:
* Gave timelines test cases unique names
* renamed wrongly called fixtures


izapolsk:
PRT errors aren't related to these changes:
1. rhos7-ga errors are caused by unstable provider
2. cluster index issue is something new and I'll check and fix it in other PR

PR #4743, opened 2017-05-25
[1LP][RFR] remove rest_api fixture from test_rest_templates

mkoura:
removing the `res_api` fixture + old cruft

{{pytest: -v cfme/tests/infrastructure/test_rest_templates.py}}


mkoura:
Failure on upstream is unrelated to this PR.

PR #4750, opened 2017-05-26
[1LP][RFR] fix wait_for calls to return False on failure

mkoura:
fix `wait_for` calls to return False on failure. I mistakenly assumed that any false value means failure, but it's just False by default.

### DIFF
--- a/cfme/tests/configure/test_rest_access_control.py
+++ b/cfme/tests/configure/test_rest_access_control.py
@@ -66,7 +66,7 @@ class TestTenantsViaREST(object):
         assert tenants_len == len(edited)
         for index in range(tenants_len):
             record, _ = wait_for(
-                lambda: collection.find_by(name=new[index]['name']),
+                lambda: collection.find_by(name=new[index]['name']) or False,
                 num_sec=180,
                 delay=10,
             )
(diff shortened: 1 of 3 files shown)


dajoRH:
*CFME QE Bot*

cfme/tests/configure/test_rest_access_control.py
```diff
@@ -66,7 +66,7 @@ def test_edit_tenants(self, appliance, tenants, multiple):
         assert tenants_len == len(edited)
         for index in range(tenants_len):
             record, _ = wait_for(
-                lambda: collection.find_by(name=new[index]['name']),
+                lambda: collection.find_by(name=new[index]['name']) or False,
```
mshriver: While this is one option, and I do like it, I at least wanted to make sure you were aware of the method of using the fail_condition argument (L101). Combined with the behavior of the fail condition check setting based on the type (L72) you might be able to get the behavior that you want without using the lambda at all or having to be a bit 'tricky'.

https://github.com/RedHatQE/wait_for/blob/master/wait_for/__init__.py#L101
https://github.com/RedHatQE/wait_for/blob/master/wait_for/__init__.py#L72

Once you've responded that you're aware of the argument/behavior ahead, but want to stick with this implementation I'm willing to merge. So totally optional but a bit more traditional to the method's intended use.

mkoura: Thanks for comment, I was not able to come up with nice `fail_condition` in this situation. When nothing is found the `collection.find_by` returns empty `SearchResult` which is false. So I would need to use something like `fail_condition=lambda result: not result` if I'm not mistaken.

PR #4757, opened 2017-05-29
[RFR] Superspeed

mfalesni:
dismiss_any_alerts takes ~2s and is invoked on every step. By putting it inside the following try/except we only execute it when there is a possibility it actually happened. If there is an alert and you try to interact witha page, you get an exception, which we handle, currently we do nothing but this patch moves the alert check in there.


mfalesni:
Checking against some older PRs that do smoke tests this reduces the running time ~7:46 to ~6:15 on 5.7. The numbers aren't directly comparable due to environment influence, but I guess it is a good start.

I will investigate why 5.8 failed.

PR #4767, opened 2017-05-31
[1LP][RFR][STABLE] Disable appliance destroy with UI coverage enabled

mfalesni:
Please add this into the stable branch as well.


mfalesni:
@mshriver That should be the case.

PR #4773, opened 2017-06-01
[RFR][1LP] Refactor template_upload_gce

mshriver:
- Use gcloud/gsutil from cli-tool box that already exists for CFME
- more robust download/upload/error handling
- clean up bucket to not waste storage costs

FIXES RHCFQE-2850

### DIFF
--- a/scripts/template_upload_gce.py
+++ b/scripts/template_upload_gce.py
@@ -12,6 +12,7 @@ normally be placed in main function, are located in function run(**kwargs).
 import argparse
 import re
 import sys
+from os.path import join
 from threading import Lock, Thread
 from datetime import datetime
 
@@ -98,11 +99,11 @@ def download_image_file(image_url, ssh_client):
     """
     # TODO: add a --local option to the script and run commands locally to download
     # Download to the amazon/gce upload machine in the lab because its fast
-    target_dir = '/root/downloads'
+    target_dir = '/var/tmp/templates/'
     file_name = image_url.split('/')[-1]
     # check if file exists
     print('INFO: Checking if file exists on cli-tool-client...')
-    result = ssh_client.run_command('ls -1 {}/{}'.format(target_dir, file_name))
+    result = ssh_client.run_command('ls -1 {}'.format(join(target_dir, file_name)))
     if result.success:
         print('INFO: File exists on cli-tool-client, skipping download...')
         return file_name, target_dir
@@ -111,7 +112,7 @@ def download_image_file(image_url, ssh_client):
     print('INFO: Prepping cli-tool-client machine for download...')
     assert ssh_client.run_command('mkdir -p {}'.format(target_dir))
     # This should keep the downloads directory clean
-    assert ssh_client.run_command('rm -f {}/*.gz'.format(target_dir))
+    assert ssh_client.run_command('rm -f {}'.format(join(target_dir, '*.gz')))
 
     # get the file
     download_cmd = 'cd {}; ' \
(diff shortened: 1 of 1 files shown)


scripts/template_upload_gce.py
```diff
@@ -104,51 +131,98 @@ def check_template_name(name):
...
+
+        # IMAGE CREATION
+        log_detail('Creating template {}...'.format(template_name), provider)
         template_name = check_template_name(template_name)
-        gcloud.create_image(image_name=template_name, bucket_url=blob_name.get('selfLink'))
-        print("GCE:{} Successfully uploaded the template.".format(provider))
+        image_cmd = 'gcloud compute images create {} --source-uri gs://{}/{}'\
+            .format(template_name, bucket, file_name)
+        result = ssh_client.run_command(image_cmd)
+        assert result.success
```
izapolsk: :+1:

scripts/template_upload_gce.py
```diff
@@ -67,33 +73,54 @@ def make_kwargs(args, **kwargs):
...
+def download_image_file(image_url, ssh_client):
+    """
+    Download the file to the cli-tool-client and return the file path + file name
+    Default destinations are None and are set to the cli-tool-client if left that way
+    :param image_url: string url to the tar.gz image
+    :return: tuple: file_name, file_path strings
+    """
+    # TODO: add a --local option to the script and run commands locally to download
+    # Download to the amazon/gce upload machine in the lab because its fast
+    target_dir = '/root/downloads'
```
mshriver: Possibly, but /tmp only has 1G on this server so that won't be enough.  /var and /root are in the same logical volume, so I could move this to /var.

scripts/template_upload_gce.py
```diff
@@ -67,33 +73,54 @@ def make_kwargs(args, **kwargs):
...
-            f.write(buffer_f)
-            status = r"{:.2f}".format(file_size_dl, file_size_dl * 100. / file_size)
-            status += chr(8) * (len(status) + 1)
-            print(status)
-            sleep(5)
-
-    return (file_name, file_path)
+    # check if file exists
+    print('INFO: Checking if file exists on cli-tool-client...')
+    result = ssh_client.run_command('ls -1 {}/{}'.format(target_dir, file_name))
```
mshriver: ah yes, definitely. will do.

scripts/template_upload_gce.py
```diff
@@ -104,51 +131,98 @@ def check_template_name(name):
...
-        blob_name = gcloud.get_file_from_bucket(bucket_name, file_name)
-        if not blob_name:
-            gcloud.upload_file_to_bucket(bucket_name, file_path)
-            blob_name = gcloud.get_file_from_bucket(bucket_name, file_name)
-        print("GCE:{} File uploading done ...".format(provider))
-
-        print("GCE:{} Creating template/image {}...".format(provider, template_name))
+        # IMAGE CHECK
+        log_detail('Checking if template {} present...'.format(template_name), provider)
+        result = ssh_client.run_command('gcloud compute images list {}'.format(template_name))
```
mshriver: talked offline, this is the only one that returns 0 even when it doesn't find anything, so I'm forced to rely just on stdout.
[root@[host] ~]# gcloud compute images list cfme-5-6-0-11-junk
Listed 0 items.
[root@[host] ~]# echo $?
0

scripts/template_upload_gce.py
mshriver: Did this everywhere that I had concatenated {}/{}, for local paths and those on the GCE bucket.

PR #4800, opened 2017-06-06
[RFR][1LP] Update cleanup_old_vms, detailed reporting

mshriver:
Refactor the cleanup_old_vms script to:
* Remove list mode - to be replaced by separate script
* Collect report data for deleted VMs, and write that to outfile

This way the script is more focused on cleanup, and reporting cleanup actions/results, and its use/function doesn't get convoluted.

I've been testing locally with combinations of singular and multiple text matchers with no issues so far, other than uncovering problem VM's that are in bad states and can't be cleaned up via normal calls.

### DIFF
--- a/scripts/cleanup_old_vms.py
+++ b/scripts/cleanup_old_vms.py
@@ -279,7 +279,7 @@ def cleanup_vms(texts, max_hours=24, providers=None, prompt=True):
     for thread in thread_queue:
         thread.join()
 
-    with open(args.outfile, 'a') as report:
+    with open(args.outfile, 'w') as report:
         report.write('## VM/Instances deleted via:\n'
                      '##   text matches: {}\n'
                      '##   age matches: {}\n'
--- a/scripts/cleanup_old_vms.py
+++ b/scripts/cleanup_old_vms.py
@@ -97,7 +97,7 @@ def get_vm_config_modified_time(host_name, vm_name, datastore_url, provider_key)
             'hostname': hostname
         }
         datastore_path = re.findall(r'([^ds:`/*].*)', str(datastore_url))
-        command = 'find ~/{} -name {}.vmx | xargs  date -r'.format(
+        command = 'find ~/{} -exec date -r {} \; -name {}.vmx | xargs  date -r'.format(
             pathjoin(datastore_path[0], vm_name),
             vm_name)
 
@@ -166,7 +166,7 @@ def process_provider_vms(provider_key, matchers, delta, vms_to_delete):
                 continue
 
         with lock:
-            print('{} finished'.format(provider_key))
+            print('{} finished processing for matches'.format(provider_key))
     except Exception as ex:
         with lock:
             # Print out the error message too because logs in the job get deleted
@@ -204,7 +204,7 @@ def delete_provider_vms(provider_key, provider_mgmt, names_ages):
             with lock:
                 deleted_vms_list.append([provider_key, vm_name, age, status, result])
     with lock:
-        print("{} is done!".format(provider_key))
+        print("{} is done deleting vms".format(provider_key))
 
 
 def cleanup_vms(texts, max_hours=24, providers=None, prompt=True):
@@ -289,8 +289,9 @@ def cleanup_vms(texts, max_hours=24, providers=None, prompt=True):
                            tablefmt='orgtbl')
         report.write(message)
     print(message)
-
     print("Deleting finished")
+
+    return 0
 
 
 if __name__ == "__main__":
--- a/scripts/cleanup_old_vms.py
+++ b/scripts/cleanup_old_vms.py
@@ -279,7 +279,7 @@ def cleanup_vms(texts, max_hours=24, providers=None, prompt=True):
     for thread in thread_queue:
         thread.join()
 
-    with open(args.outfile, 'w') as report:
+    with open(args.outfile, 'a') as report:
         report.write('## VM/Instances deleted via:\n'
                      '##   text matches: {}\n'
                      '##   age matches: {}\n'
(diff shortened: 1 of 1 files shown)


scripts/cleanup_old_vms.py
```diff
@@ -56,6 +62,14 @@ def match(matchers, vm_name):
 
 
 def get_vm_config_modified_time(name, vm_name, datastore_url, provider_key):
+    """
```
mshriver: Will update this for CQ, didn't feel like finishing out tonight.

mshriver:
Have made some significant commits, so bumped myself back down to level 1 and requested re-review from mkoura.  WIPTEST for now until I'm done testing changes.

dajoRH:
*CFME QE Bot*

mshriver:
@mkoura Thanks for pointing that out, I'll go through and clean up for that.

scripts/cleanup_old_vms.py
```diff
@@ -55,51 +68,59 @@ def match(matchers, vm_name):
...
         datastore_path = re.findall(r'([^ds:`/*].*)', str(datastore_url))
-        ssh_client = SSHClient(**connect_kwargs)
-        command = 'find ~/{}/{} -name {} | xargs  date -r'.format(
-            datastore_path[0], str(vm_name), str(vm_name) + '.vmx')
-        exit_status, output = ssh_client.run_command(command)
-        ssh_client.close()
-        modified_time = parser.parse(output.rstrip())
-        modified_time = modified_time.astimezone(pytz.timezone(str(get_localzone())))
-        return modified_time.replace(tzinfo=None)
+        command = 'find ~/{} -name {}.vmx | xargs  date -r'.format(
```
mshriver: hm agreed, just left the command as-is from old script. updating.

scripts/cleanup_old_vms.py
```diff
@@ -55,51 +68,59 @@ def match(matchers, vm_name):
...
+            pathjoin(datastore_path[0], vm_name),
+            vm_name)
+
+        with SSHClient(**connect_kwargs) as ssh_client:
+            result = ssh_client.run_command(command)
+        parsed_time = parser.parse(str(result).rstrip())
```
izapolsk: are you sure you will always get here only one row ?

scripts/cleanup_old_vms.py
```diff
@@ -125,6 +145,16 @@ def process_provider_vms(provider_key, provider_type, matchers,
...
 
+                if vm_creation_time is False:
+                    # This VM must have some problem, include in report even though we can't delete
+                    status = provider.vm_status(vm_name)
+                    deleted_vms_list.append([provider_key,
+                                             vm_name,
+                                             NULL,  # can't know age, failed getting creation date
+                                             status,
+                                             FAIL])
+                    raise Exception  # the except block message is accurate in this case
```
izapolsk: IMHO, it would look better if it was implemented using for/continue instead of this try/catch block.

scripts/cleanup_old_vms.py
```diff
@@ -145,54 +175,72 @@ def process_provider_vms(provider_key, provider_type, matchers,
         logger.exception(ex)
 
 
-def delete_provider_vms(provider_key, vm_names):
+def delete_provider_vms(provider_key, provider_mgmt, names_ages):
+    """
+    lock, lock, more locking. Need to convert caller to Process+Queue
```
izapolsk: +:100:

scripts/cleanup_old_vms.py
```diff
@@ -145,54 +175,72 @@ def process_provider_vms(provider_key, provider_type, matchers,
...
+    Checks provider connectivity (using ping)
+    Threads process_provider_vms to build list of vms to delete
+    Prompts user to continue with delete
+    Threads deleting of the vms
+
+    :param texts: list of strings to match against
+    :param max_hours: integer maximum number of hours that the VM can exist for
+    :param providers: list of provider keys
+    :param prompt: boolean, whether or not to prompt the user for each delete
+    :return: 0 if user declines delete when prompt is True
```
izapolsk: I don't see where it returns something.

scripts/cleanup_old_vms.py
mshriver: I'll check this out.

scripts/cleanup_old_vms.py
```diff
@@ -239,10 +279,20 @@ def cleanup_vms(texts, max_hours=24, providers=None, prompt=True):
...
+    print(message)
+
     print("Deleting finished")
 
 
 if __name__ == "__main__":
     args = parse_cmd_line()
-    sys.exit(cleanup_vms(args.text_to_match, args.max_hours,
-                         args.providers, args.prompt))
+    sys.exit(cleanup_vms(args.text_to_match, args.max_hours, args.providers, args.prompt))
```
izapolsk: I suppose cleanup_vms should always return something If you pass this procedure results to sys.exit.
0 - success
1-n error

scripts/cleanup_old_vms.py
mshriver: Agreed on both counts here, though I'm not too worried about getting more than 1 row since there should only be 1 vmx file here.

Since mgmtsystem doesn't return consistent timestamp formatting for vm_creation_time, I think I would like to leave refactoring the timestamp/age handling until wrapanapi issue #133 is fixed:
https://github.com/ManageIQ/wrapanapi/issues/133

mshriver: Right now I'm not worried about this, there should only be 1 vmx file found and thus 1 date.

Going to add a TODO to improve the datetime handling and parsing a bit cleaner, sound okay?

scripts/cleanup_old_vms.py
mshriver: @izapolsk we'll have to chat about this tomorrow because looking closer I'm not quite sure what you mean.

mshriver:
@psav @mfalesni as much as I want to merge this so I can see it run tonight in Jenkins, I think @izapolsk is right and one of you should check this over, because of the potential for bad deleting.

PR #4805, opened 2017-06-06
[RFR] test_config_management fix

izapolsk:
There are ~ 7 tests failing in jenkins with NavigationError error.
This PR fixes those tests

{{pytest: cfme/tests/infrastructure/test_config_management.py}}


izapolsk:
I was fixing NavigationError and it is fixed. PRT fail but I think this is the responsibility of test owner.

PR #4820, opened 2017-06-08
[1LP][RFR] Script to list provider vms

mshriver:
This simple script is for monitoring of active vms on providers.
Takes --tag and positional provider key options
tags correspond to those in cfme_data for management_system entries
Multiple tags are additive, not combination
Can specify tags and providers

Takes --outfile to write a tabulated report to a file, also writes it to
stdout along with progress updates

Lots of other things this _could_ do, but keeping it simple for now.

### DIFF
--- a/scripts/list_provider_vms.py
+++ b/scripts/list_provider_vms.py
@@ -1,0 +1,141 @@
+import argparse
+from tabulate import tabulate
+from multiprocessing import Process, Queue
+from multiprocessing.queues import Empty
+
+from utils.conf import cfme_data
+from utils.path import log_path
+from utils.providers import get_mgmt, list_provider_keys
+
+
+# Constant for report
+NULL = '--'
+
+
+def parse_cmd_line():
+    parser = argparse.ArgumentParser(argument_default=None)
+    parser.add_argument('--outfile',
+                        default=log_path.join('list_provider_vms.log').strpath,
+                        dest='outfile')
+    parser.add_argument('--tag',
+                        default=None,
+                        dest='tag',
+                        action='append',
+                        help='A provider tag to match a group of providers instead of all '
+                             'providers from cfme_data. Can be used multiple times')
+    parser.add_argument('provider',
+                        default=None,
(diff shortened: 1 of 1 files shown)


scripts/list_provider_vms.py
```diff
@@ -0,0 +1,145 @@
...
+                        default=None,
+                        nargs='*',
+                        help='Provider keys, can be user multiple times. If none are given '
+                             'the script will use all providers from cfme_data or match tags')
+
+    args = parser.parse_args()
+    return args
+
+
+def process_tags(provider_keys, tags=[]):
```
quarckster: You should be careful using mutable type as a default value in a function argument. Consider to use `None` instead.

scripts/list_provider_vms.py
```diff
@@ -0,0 +1,145 @@
...
+    Build list of lists with basic vm info: [[provider, vm, status, age, type], [etc]]
+    :param provider_key: string provider key
+    :param output_queue: a multiprocessing.Queue object to add results to
+    :return: list of lists of vms and basic statistics
+    """
+    output_list = []
+
+    print('Listing VMS on provider {}'.format(provider_key))
+    provider = get_mgmt(provider_key)
+    try:
```
quarckster: **OPTIONAL:**
```python
try:
    list_vm = provider.list_vm()
except NotImplementedError:
    ...
else:
    for vm_name in list_vm:
        ...
```

mshriver: ah, yeah good call. I shoehorned in the NotImplementedCatch at the end and its ugly as-is.

mfalesni:
@mshriver Since psav made #4825, you may possibly add it to the master script :)

scripts/list_provider_vms.py
mshriver: will do.

mshriver:
@mfalesni agreed that it should be in the miq scripts, but an enhancement for another time :)

PR #4840, opened 2017-06-12
[RFR]Fix mutable args

psav:
Though some of these are not necessary, it would perhaps be better to
set a standard of no mutable args. These changes shouldn't break
anything and are really to keep QC happier.

### DIFF
--- a/cfme/control/explorer/actions.py
+++ b/cfme/control/explorer/actions.py
@@ -119,8 +119,9 @@ class Action(Updateable, Navigatable, Pretty):
         description: Action name.
         action_type: Type of the action, value from the dropdown select.
     """
-    def __init__(self, description, action_type, action_values={}, appliance=None):
+    def __init__(self, description, action_type, action_values=None, appliance=None):
         Navigatable.__init__(self, appliance=appliance)
+        action_values = action_values or {}
         self.description = description
         self.action_type = action_type
         self.snapshot_name = action_values.get("snapshot_name")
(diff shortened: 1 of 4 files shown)


cfme/control/explorer/actions.py
```diff
@@ -119,8 +119,9 @@ class Action(Updateable, Navigatable, Pretty):
         description: Action name.
         action_type: Type of the action, value from the dropdown select.
     """
-    def __init__(self, description, action_type, action_values={}, appliance=None):
+    def __init__(self, description, action_type, action_values=None, appliance=None):
         Navigatable.__init__(self, appliance=appliance)
+        action_values = {} if not action_values else action_values
```
mfalesni: `action_values = action_values or {}`

utils/bz.py
```diff
@@ -135,8 +135,10 @@ def get_bug_variants(self, id):
                 expanded.add(b)
         return found
 
-    def resolve_blocker(self, blocker, version=None, ignore_bugs=set([]), force_block_streams=[]):
+    def resolve_blocker(self, blocker, version=None, ignore_bugs=None, force_block_streams=None):
         # ignore_bugs is mutable but is not mutated here! Same force_block_streams
+        force_block_streams = [] if not force_block_streams else force_block_streams
+        ignore_bugs = set([]) if not ignore_bugs else ignore_bugs
```
mfalesni: `ignore_bugs = ignore_bugs or set()`

psav:
It is, I went into silly mode ;)

psav:
Fixed

PR #4842, opened 2017-06-12
[RFR][1LP] create legacy testcases

mkoura:
Generate xml files with legacy testcases (without parametrization) for xunit importer.

To be run with ``pytest --collect-only --generate-legacy-xmls --xmls-testrun-id <testrun id>``

{{pytest: --collect-only --use-provider complete --generate-legacy-xmls --xmls-testrun-id 5_7_3_01}}

### DIFF
--- a/fixtures/xunit_tools.py
+++ b/fixtures/xunit_tools.py
@@ -23,6 +23,12 @@ def pytest_addoption(parser):
     parser.getgroup('cfme')
     parser.addoption("--generate-xmls", action="store_true", default=False,
         help="generate the xml files for import")
+    parser.addoption("--generate-legacy-xmls", action="store_true", default=False,
+        help="generate the legacy xml files for import")
+    parser.addoption("--xmls-testrun-id",
+        help="testrun id")
+    parser.addoption("--xmls-testrun-title",
+        help="testrun title")
 
 
 def testcase_gen(
@@ -49,8 +55,8 @@ def testcase_gen(
     test_steps.append(test_step)
     testcase.append(test_steps)
     custom_fields_el = etree.Element('custom-fields')
-    for id, content in custom_fields.iteritems():
-        custom_field = etree.Element('custom-field', id=id, content=content)
+    for tc_id, content in custom_fields.iteritems():
+        custom_field = etree.Element('custom-field', id=tc_id, content=content)
         custom_fields_el.append(custom_field)
     testcase.append(custom_fields_el)
     if linked_items:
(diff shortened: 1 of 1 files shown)


fixtures/xunit_tools.py
```diff
@@ -239,13 +251,21 @@ def pytest_collection_modifyitems(session, config, items):
...
-        except:
-            param_dict = {}
+        if legacy:
+            if item.name in test_name:
+                continue
+            name = item.name
+            param_dict = None
+            test_name.append(name)
+        else:
+            name = re.sub(r'\[.*\]', '', item.name)
```
mfalesni: that is parameter stripping, right?

mkoura: yes, this is when we are generating xmls with parametrization

mfalesni: I am wondering whether this simple solution can cause any problems.

mkoura: @mfalesni I don't think it can, there are neither nested square brackets nor square brackets used in test name (other than for test parameter).

fixtures/xunit_tools.py
```diff
@@ -18,15 +18,36 @@
...
 
 
+blacklist = [
+    'cfme/tests/containers/',
+    'cfme/tests/middleware/',
+    'cfme/tests/openstack/',
+    'utils/',
+    'rhos',
+    'rhev',
+    'hawkular',
```
jkrocil: What about `cfme/scripting` here?

fixtures/xunit_tools.py
```diff
@@ -18,15 +18,36 @@
...
     parser.getgroup('cfme')
     parser.addoption("--generate-xmls", action="store_true", default=False,
         help="generate the xml files for import")
+    parser.addoption("--generate-legacy-xmls", action="store_true", default=False,
+        help="generate the legacy xml files for import")
+    parser.addoption("--xmls-testrun-id",
+        help="testrun id")
+    parser.addoption("--xmls-testrun-title",
+        help="testrun title")
+    parser.addoption("--xmls-apply-blacklist", action="store_true", default=False,
```
jkrocil: We might want to switch this to `True` by default, otherwise one incorrect invocation will create test cases we don't care about that will have to be cleaned up.

mkoura: when generating the XMLs to be used for submit to polarion, this option is used - https://github.com/mkoura/cfme-testcases/blob/master/cfme_testcases/gen_xmls.py#L66

mkoura:
I refactored the code into multiple functions with a goal to make it more readable/maintainable and to be able to iterate over the items just once when generating both XML files.

The blacklist now matches only testcases in `cfme/tests` that are tested by external teams. All other tests (`utils/tests`, `scripts/tests`, etc.) are excluded always even when `--xmls-apply-blacklist` option is not used.

PR #4848, opened 2017-06-13
[RFR][NOTEST] Fulfill access_controll class, method docs

rlbabyuk:
Purpose or Intent
=================
Fullfill class, method docs

### DIFF
--- a/cfme/configure/access_control.py
+++ b/cfme/configure/access_control.py
@@ -22,6 +22,10 @@ def simple_user(userid, password):
 
 
 class UserForm(ConfigurationView):
+    """ User Form View.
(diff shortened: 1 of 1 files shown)


cfme/configure/access_control.py
```diff
@@ -22,6 +22,10 @@ def simple_user(userid, password):
 
 
 class UserForm(ConfigurationView):
+    """ User Form View.
+        Args:
```
psav: For these forms/views, i don't think Args is necesssary

rlbabyuk: Done, removed Args from views

PR #4856, opened 2017-06-15
[1LP][RFR] NTP configuration fix

patchkez:
Purpose or Intent
=================
All our Sprout appliances (configured & unconfigured) are not able to pick up NTP servers offered by DHCP client. Root cause of the problem are 2 missing lines in chrony.conf:
```
commandkey 1
keyfile /etc/chrony.keys
```
During appliance provisioning, we overwrite chronyd configuration file (and we miss above 2 lines). We should not overwrite but amend this config. This fix is trying to use default chronyd config which comes with rpm and adds only those NTP servers (from yamls), which are not defined in chrony.conf already.

More details about the problem is captured here - [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1439755).

### DIFF
--- a/utils/appliance/__init__.py
+++ b/utils/appliance/__init__.py
@@ -924,36 +924,30 @@ class IPAppliance(object):
 
         # Retrieve time servers from yamls
         server_template = 'server {srv} iburst'
-        time_servers = []
+        time_servers = set()
         try:
             logger.debug('obtaining clock servers from config file')
-            clock_servers = conf.cfme_data.get('clock_servers', {})
+            clock_servers = conf.cfme_data.get('clock_servers')
             for clock_server in clock_servers:
-                time_servers.append(server_template.format(srv=clock_server))
-        except IndexError:
+                time_servers.add(server_template.format(srv=clock_server))
+        except TypeError:
             msg = 'No clock servers configured in cfme_data.yaml'
             log_callback(msg)
             raise ApplianceException(msg)
 
         filename = '/etc/chrony.conf'
-        chrony_conf = client.run_command("cat {f}".format(f=filename)).output.strip().split('\n')
+        chrony_conf = set(client.run_command("cat {f}".format(f=filename)).output.strip()
+                    .split('\n'))
 
-        # Get list of servers which will be added into chrony.conf
-        for line in chrony_conf:
-            for time_server in time_servers:
-                if time_server in line:
-                    # modified_chrony_file.append(line)
-                    index_of_element = time_servers.index(time_server)
-                    time_servers.pop(index_of_element)
-
-        if len(time_servers) > 0:
-            for time_server in time_servers:
-                chrony_conf.append(time_server)
-
-        modified_chrony_conf = "\n".join(chrony_conf)
-        client.run_command('echo "{txt}" > {f}'.format(txt=modified_chrony_conf, f=filename))
-        logger.info("chrony's config file updated")
-        conf_file_updated = True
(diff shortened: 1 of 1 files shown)


utils/appliance/__init__.py
```diff
@@ -922,28 +922,38 @@ def fix_ntp_clock(self, log_callback=None):
...
+        # Retrieve time servers from yamls
         server_template = 'server {srv} iburst'
-        base_config = ['driftfile /var/lib/chrony/drift', 'makestep 10 10', 'rtcsync']
+        time_servers = []
         try:
             logger.debug('obtaining clock servers from config file')
             clock_servers = conf.cfme_data.get('clock_servers', {})
             for clock_server in clock_servers:
-                base_config.append(server_template.format(srv=clock_server))
+                time_servers.append(server_template.format(srv=clock_server))
```
quarckster:
```python
time_servers.add(server_template.format(srv=clock_server))
```

utils/appliance/__init__.py
```diff
@@ -922,28 +922,38 @@ def fix_ntp_clock(self, log_callback=None):
...
             client.run_command('systemctl daemon-reload')
 
-        # # deploy config and start chrony if it isn't running
+        # Retrieve time servers from yamls
         server_template = 'server {srv} iburst'
-        base_config = ['driftfile /var/lib/chrony/drift', 'makestep 10 10', 'rtcsync']
+        time_servers = []
         try:
             logger.debug('obtaining clock servers from config file')
             clock_servers = conf.cfme_data.get('clock_servers', {})
```
quarckster:
```python
clock_servers = conf.cfme_data.get('clock_servers')

File: utils/appliance/__init__.py [hunk 3]
```diff
@@ -922,28 +922,38 @@ def fix_ntp_clock(self, log_callback=None):
...
             for clock_server in clock_servers:
-                base_config.append(server_template.format(srv=clock_server))
+                time_servers.append(server_template.format(srv=clock_server))
         except IndexError:
```
quarckster:
```python
except TypeError:
```

utils/appliance/__init__.py
```diff
@@ -922,28 +922,38 @@ def fix_ntp_clock(self, log_callback=None):
...
+                    time_servers.pop(index_of_element)
+
+        if len(time_servers) > 0:
+            for time_server in time_servers:
+                chrony_conf.append(time_server)
+
+        modified_chrony_conf = "\n".join(chrony_conf)
+        client.run_command('echo "{txt}" > {f}'.format(txt=modified_chrony_conf, f=filename))
+        logger.info("chrony's config file updated")
+        conf_file_updated = True
```
quarckster:
```python
modified_chrony_conf = chrony_conf.union(time_servers)
if modified_chrony_conf != chrony_conf:
    modified_chrony_conf = "\n".join(list(modified_chrony_conf))
    client.run_command('echo "{txt}" > {f}'.format(txt=modified_chrony_conf, f=filename))
    logger.info("chrony's config file updated")
    conf_file_updated = True
else:
    logger.info("chrony's config file hasn't been changed")
    conf_file_updated = False
```

izapolsk:
@patchkez, I have merged your changes but the question about checking our existing servers is still open.

patchkez:
@izapolsk Last time when I checked, appliances running on rhos7-sprout and on RHEV environment had NTP servers pre-configured in chrony.conf. NTP servers offered by DHCP were not used because of missing directives in chrony.conf.

PR #4862, opened 2017-06-16
[WIPTEST] Fixed import errors

quarckster:
Purpose
=======

__Fixing__ import errors of `login` and `login_admin`


psav:
These need to be changed to no longer import, but to use the appliance objects, like ```appliance.server.login()```

PR #4880, opened 2017-06-19
[1LP][RFR] Openstack test instances fix

okolisny:
Fixing tests/openstack/cloud/test_instances.py because of changes in Instance class

{{ pytest: cfme/tests/openstack/cloud/test_instances.py --use-provider rhos7-ga }}

### DIFF
--- a/cfme/cloud/instance/__init__.py
+++ b/cfme/cloud/instance/__init__.py
@@ -493,9 +493,6 @@ class Provision(CFMENavigateStep):
     def step(self, *args, **kwargs):
         self.prerequisite_view.toolbar.lifecycle.item_select('Provision Instances')
 
-    def am_i_here(self):
-        return match_location(summary='Provision Instances - Select an Image')
-
 
 @navigator.register(Instance, 'Reconfigure')
 class Reconfigure(CFMENavigateStep):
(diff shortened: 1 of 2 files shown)


izapolsk:
@okolisny, all PRT are uncollected. Please either fix this or confirm that this fix is working in your env.
I'll merge it then.

okolisny:
@izapolsk PRT run finished, some tests failed, and I don't know why. But it looks like a race condition, IMO
Also, there is a run in Jenkins - [REPORT](https://rhos-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/Director/view/CFME_RHOS_Integration/job/rhos9_cfme57_integration/TEST_REPORT/)
Again, some tests failed, and again the reason are race conditions - 1) error when calling Openstack API, and 2) test couldn't wait for instance to appear

izapolsk:
@okolisny, as far as I see this PR fixes all test_instance tests but some of them still fail by another reason.
So, we can go two ways
1. I merge this PR and you work on failing tests in another PR
2. I wait until you all failing tests in this PR

Which way would you prefer ?

okolisny:
@izapolsk lets go the first way. thanks

PR #4884, opened 2017-06-19
[1LP][RFR] Smoke tests for Control

quarckster:
Intent
=====

__Adding smoke tests__ for Control tab.

{{py.test: -v -k 'test_control_navigation or test_control_explorer_tree'}}

### DIFF
--- a/cfme/control/simulation.py
+++ b/cfme/control/simulation.py
@@ -1,0 +1,36 @@
+# -*- coding: utf-8 -*-
+from navmazing import NavigateToSibling
+from widgetastic.widget import Select
+from widgetastic_patternfly import Button
+
+from cfme.base import Server
+from cfme.base.login import BaseLoggedInPage
+from utils.appliance.implementations.ui import navigator, CFMENavigateStep
+
+
+class ControlSimulationView(BaseLoggedInPage):
+    """Basic view for Control/Simulation tab."""
+    event_selection = Select(id="event_typ")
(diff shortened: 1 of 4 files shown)


cfme/control/simulation.py
```diff
@@ -0,0 +1,34 @@
...
+    reset_button = Button("Reset")
+
+    @property
+    def is_displayed(self):
+        return (
+            self.event_selection.is_displayed and
+            self.vm_selection.is_displayed and
+            self.submit_button.is_displayed and
+            self.reset_button.is_displayed
+        )
```
mshriver: At least a TODO here for the results pane

cfme/tests/control/test_smoke_control.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+@pytest.mark.parametrize("destination", destinations)
+def test_control_navigation(destination):
+    """This test verifies presence of destinations of Control tab.
+
+    Steps:
+        * Open each destination of Control tab.
+    """
+    navigate_to(Server, destination)
```
mshriver: **Optional**
Capture the view and add an assert against view.is_displayed for an explicit 'test' here.

cfme/control/simulation.py
quarckster: There is no results pane when you open Simulation tab, the tree with simulation results will be shown only after filling that widgets.

quarckster: Done

PR #4886, opened 2017-06-20
[RFR][1LP] Fix overview tests

gshefer:
{{pytest: cfme/tests/containers/test_breadcrumbs.py cfme/tests/containers/test_containers_summary.py cfme/tests/containers/test_overview_data_integrity.py -v --use-provider cm-env2}}

- Added status_boxs to cfme/containers/overview.py
- Fix overview tests

### DIFF
--- a/cfme/containers/overview.py
+++ b/cfme/containers/overview.py
@@ -1,16 +1,19 @@
 # -*- coding: utf-8 -*-
 from functools import partial
-from utils.appliance import Navigatable
-from cfme.common import Taggable
-from utils.appliance.implementations.ui import CFMENavigateStep, navigator
+
 from navmazing import NavigateToAttribute
-from cfme.web_ui import match_location
+
+from cfme.containers.provider import ContainersProvider
+from cfme.web_ui import match_location, StatusBox
+from utils.appliance.implementations.ui import CFMENavigateStep, navigator
+from utils.appliance import Navigatable
+from utils.wait import wait_for
 
 
 match_page = partial(match_location, controller='container_dashboard', title='Container Dashboards')
 
 
-class ContainersOverview(Taggable, Navigatable):
+class ContainersOverview(Navigatable):
     pass
 
 
@@ -25,4 +28,6 @@ class All(CFMENavigateStep):
         self.prerequisite_view.navigation.select('Compute', 'Containers', 'Overview')
 
     def resetter(self):
-        pass
+        # We should wait ~2 seconds for the StatusBox population
+        wait_for(lambda: StatusBox(ContainersProvider.PLURAL.split(' ')[-1]).value(),
+                 num_sec=10, delay=1)
--- a/cfme/tests/containers/test_containers_summary.py
+++ b/cfme/tests/containers/test_containers_summary.py
@@ -1,30 +1,28 @@
 import pytest
+from widgetastic.utils import attributize_string
+
 from cfme.containers.overview import ContainersOverview
 from cfme.containers.provider import ContainersProvider
+from cfme.containers.container import Container
+from cfme.containers.service import Service
+from cfme.containers.route import Route
+from cfme.containers.project import Project
+from cfme.containers.pod import Pod
+from cfme.containers.image import Image
+from cfme.containers.image_registry import ImageRegistry
+from cfme.containers.node import Node
 from cfme.web_ui import StatusBox
-from utils import testgen, version
+
+from utils import testgen
 from utils.appliance.implementations.ui import navigate_to
-from utils.version import current_version
 from utils.blockers import BZ
 
 pytestmark = [
-    pytest.mark.uncollectif(
-        lambda: current_version() < "5.6"),
     pytest.mark.usefixtures('setup_provider'),
     pytest.mark.tier(1)]
 pytest_generate_tests = testgen.generate([ContainersProvider], scope='function')
 
-
-container_object_types = \
-    ['routes', 'projects', 'container_images', 'image_registries',
-     'container_services', 'containers', 'pods', 'nodes']
-container_object_types_lowest = \
-    ['routes', 'projects', 'images', 'image_registries',
-     'services', 'containers', 'pods', 'nodes']
-objects_key = ({
-    version.LOWEST: container_object_types_lowest,
-    '5.7': container_object_types
-})
+tested_objects = [Service, Route, Project, Pod, Image, Container, ImageRegistry, Node]
 
 
 @pytest.mark.polarion('CMP-10575')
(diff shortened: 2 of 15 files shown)


cfme/tests/containers/test_overview_data_integrity.py
```diff
@@ -27,42 +25,37 @@
...
+        Node: 0,
         Pod: 0,
         Service: 0,
         Project: 0,
-        Route: 0,
-        ImageRegistryAll: 0
+        Route: 0
     }
-    for provider in providers:
+    for provider in list_providers_by_class(ContainersProvider):
```
psav: **REQUIRED**: This will do it for ALL providers in the yamls, not one specific to the test being run. Does it require this?

utils/ocp_cli.py
```diff
@@ -10,19 +10,21 @@ def __init__(self, provider):
...
+            self.password = creds.get('password', None)
         else:
-            # Try with known hosts
-            self.ssh_client = SSHClient()
-            self.ssh_client.load_system_host_keys()
-            self.ssh_client.connect(self.hostname)
+            self.username = credentials[creds].get('username', None)
+            self.password = credentials[creds].get('password', None)
+
+        self.ssh_client = SSHClient()
```
psav: @mfalesni can you take a look at this usage pls

mfalesni: `with` will probably be preferable, but I believe th connect args should be passed through the constructor. @gshefer can you check the definition of SSHClient and pass all the info through the constructor and use `with`? Also I believe load_system_host_keys should happen automatically somehow.

cfme/tests/containers/test_overview_data_integrity.py
gshefer: I would like to do this for all the providers since the status box presents objects count for all the provider, but, not all the providers in yaml, I need all the -added- providers, so this a good point and I'll modify accordingly.

utils/ocp_cli.py
gshefer: @mfalesni sure...

gshefer: @mfalesni if possible, I really prefer to keep it as it is since this is currently work well in Jenkins and PRT and this PR include very  urgent issues. hope that this is not critical to use With.

psav:
We also need the 5 new issues that QC found to be fixed

utils/ocp_cli.py
gshefer: @mfalesni Done.

cfme/containers/overview.py
```diff
@@ -1,15 +1,40 @@
...
-from cfme.web_ui import match_location
+from utils.appliance import Navigatable
 
 
 match_page = partial(match_location, controller='container_dashboard', title='Container Dashboards')
 
 
+# TODO: Switch the entire file to support widgetastic
+status_boxs = {
+    Route: StatusBox('Routes'),
```
psav: **OPTIONAL**: This is OK, but since this may come up elsewhere, perhaps a better idea is to put a "plural" string on each of these objects. That way you can just access Route.plural, and get back "Routes". We did this before in many places.

cfme/tests/containers/test_containers_summary.py
```diff
@@ -1,30 +1,37 @@
...
-    '5.7': container_object_types
-})
+object_names_mapping = {
+    Service: 'service',
+    Route: 'route',
+    Project: 'project',
+    Pod: 'pod',
+    Image: 'image',
+    Container: 'container',
+    ImageRegistry: 'image_regist',
```
psav: **QUESTION:** ```regist```? This could also be an attribute on the classes too as above

cfme/tests/containers/test_containers_summary.py
```diff
@@ -40,14 +47,18 @@ def test_containers_summary_objects(provider, soft_assert):
...
-    container_object = version.pick(objects_key)
-    prov_ui_values, status_box_values = dict(), dict()
     navigate_to(ContainersOverview, 'All')
-    for obj_type in container_object:
-        status_box_values[obj_type] = StatusBox(obj_type).value()
-    for obj_type in container_object:
-        prov_ui_values[obj_type] = getattr(provider.summary.relationships, obj_type).value
-        soft_assert(status_box_values[obj_type] == prov_ui_values[obj_type],
+    # We should wait ~2 seconds for the StatusBox population
+    wait_for(lambda: status_boxs[ContainersProvider].value(), num_sec=10, delay=1)
```
psav: Nice using ```wait_for``` :)

cfme/tests/containers/test_containers_summary.py
```diff
@@ -40,14 +47,18 @@ def test_containers_summary_objects(provider, soft_assert):
...
+    # Collecting status boxs values:
+    status_box_values = {}
+    for obj, sb in status_boxs.items():
+        status_box_values[obj] = sb.value()
```
psav: **OPTIONAL**: For such a small loop as this a comprehension may be better
```status_box_values = {obj: sb.value() for obj, sb in status_boxs.items()}```
Replaces 3 lines with just one, which is equally as readable

File: cfme/tests/containers/test_overview_data_integrity.py [hunk 7]
```diff
@@ -73,42 +66,14 @@ def test_containers_overview_data_integrity(provider):
...
-    list_img_from_openshift_parsed_new = set(list_img_from_openshift_parsed)
-
-    list_img_from_openshift_parsed_new.update(list_img_from_registry_splitted_new)
-
-    num_img_cfme_56 = len(provider.mgmt.list_image())
-    num_img_cfme_57 = len(list_img_from_openshift_parsed_new)
-
-    assert StatusBox('images').value() == version.pick({version.LOWEST: num_img_cfme_56,
-                                                        '5.7': num_img_cfme_57})
+    wait_for(lambda: status_boxs[ContainersProvider].value(), num_sec=10, delay=1)
```
psav: **OPTIONAL**: Since I have now seen this pattern two times navigating to the same page, you could add the wait as part of the step to the navigation. I was tempted to make this required, but I think optional will suffice.

psav:
Setting to wip until the comments have been seen

cfme/tests/containers/test_containers_summary.py
gshefer: This is becuase the soft_get, but I prefer what you suggested. WIP

cfme/containers/overview.py
gshefer: I fully agree.

gshefer:
@psav Done. Note that it takes too much time for the provider to populate.  the test_breadcrumbs has been set in order to let it enough time, so the failure is expected. anyway, we should find a way to fix this.

cfme/containers/overview.py
```diff
@@ -25,4 +51,5 @@ def step(self):
         self.prerequisite_view.navigation.select('Compute', 'Containers', 'Overview')
 
     def resetter(self):
-        pass
+        # We should wait ~2 seconds for the StatusBox population
+        wait_for(lambda: status_boxs[ContainersProvider].value(), num_sec=10, delay=1)
```
psav: My suggestion above was to avoid the above status_boxs variable entirely and be able to do this
```wait_for(lambda: ContainersProvider.PLURAL, num_sec=10, delay=1)```

PR #4887, opened 2017-06-20
fix version float

apagac:
This is a fix for get_version when 'obj' is float.

During my testing an error occurred with get_version function. For whatever reason it was getting a float number on input (obj == 5.7) and it did not expect that to happen. My fix propagates this number through the function and fixes the tests for both 5.7 and 5.8.

### DIFF
--- a/utils/version.py
+++ b/utils/version.py
@@ -41,7 +41,7 @@ def get_version(obj=None):
     If obj is None, the version will be retrieved from the current appliance
 
     """
-    if isinstance(obj, Version):
+    if isinstance(obj, Version) or isinstance(obj, float):
         return obj
     if obj.startswith('master'):
         return Version.latest()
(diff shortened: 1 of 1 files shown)


utils/version.py
```diff
@@ -41,7 +41,7 @@ def get_version(obj=None):
     If obj is None, the version will be retrieved from the current appliance
 
     """
-    if isinstance(obj, Version):
+    if isinstance(obj, Version) or isinstance(obj, float):
```
quarckster: **REQUIRED**
```python
if isinstance(obj, (Version, float)):
    ...
```
:)

utils/version.py
```diff
@@ -41,8 +42,10 @@ def get_version(obj=None):
     If obj is None, the version will be retrieved from the current appliance
 
     """
-    if isinstance(obj, Version) or isinstance(obj, float):
+    if isinstance(obj, Version):
         return obj
+    if not isinstance(obj, string_types):
+        obj = str(obj)
```
apagac: We have decided to just convert obj to string in case it is not already a string, This way it can be then inspected with startswith as a normal string and then passed to Version.

PR #4895, opened 2017-06-21
[RFR] RHV provider: Fix test_multiple_host_bad_creds error message

istein1:
__Fixing__ test_multiple_host_bad_creds, for RHV provider, in which the error message is different than other providers.


lkhomenk:
REQUIRED: still needs right signature

istein1:
Moving back to [RFR] as signature is now added.

PR #4898, opened 2017-06-21
[1LP][WIP] Cloud provider widgetastic conversion -  rest of views

izapolsk:
1. Converted rest of cloud provider views to widgetastic
2. fixed all broken tests
3. Removed old code from both cloud and infra providers
4. fixed various issues

{{pytest: -v cfme/tests/cloud/test_providers.py cfme/tests/infrastructure/test_providers.py cfme/tests/cloud_infra_common/test_discovery.py cfme/tests/cloud_infra_common/test_tag_objects.py cfme/tests/cloud_infra_common/test_tag_visibility.py }}

### DIFF
--- a/cfme/cloud/provider/__init__.py
+++ b/cfme/cloud/provider/__init__.py
@@ -220,7 +220,7 @@ def get_all_providers():
     return [item.name for item in view.items.get_all(surf_pages=True)]
 
 
-def discover(credential, cancel=False, d_type="Amazon"):
+def discover(credential, cancel=False, d_type="Amazon EC2"):
     """
     Discover cloud providers. Note: only starts discovery, doesn't
     wait for it to finish.
@@ -231,10 +231,19 @@ def discover(credential, cancel=False, d_type="Amazon"):
       d_type: provider name which will be discovered
     """
     view = navigate_to(CloudProvider, 'Discover')
-    view.fill({'discover_type': d_type,
-               'username': credential.principal,
-               'password': credential.secret,
-               'password_verify': credential.verify_secret})
+    view.fill({'discover_type': d_type})
+    if d_type == 'Amazon EC2':
+        view.fields.fill({'username': getattr(credential, 'principal', None),
+                          'password': getattr(credential, 'secret', None),
+                          'password_verify': getattr(credential, 'verify_secret', None)})
+    elif d_type == 'Azure':
+        view.fields.fill({'client_id': getattr(credential, 'principal', None),
+                          'client_key': getattr(credential, 'secret', None),
+                          'tenant_id': getattr(credential, 'tenant_id', None),
+                          'subscription': getattr(credential, 'subscription_id', None)})
+    else:
+        raise ValueError('Wrong provider type or credential')
+
     if cancel:
         view.cancel.click()
     else:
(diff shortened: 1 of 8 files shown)


dajoRH:
Lint report for commit 87098e632b32f664f7fe2593611881ef3690e0d3:

`cfme/cloud/provider/__init__.py`:
- :red_circle: Line 27:1: **F401** *'cfme.web_ui.form_buttons' imported but unused*
- :red_circle: Line 29:1: **F401** *'cfme.web_ui.fill' imported but unused*
- :red_circle: Line 103:9: **F811** *redefinition of unused 'tb' from line 28*

**No commit flaws detected**.

*CFME QE Bot*

cfme/cloud/provider/__init__.py
```diff
@@ -309,41 +214,37 @@ def step(self, *args, **kwargs):
...
+    view = navigate_to(CloudProvider, 'Discover')
+
+    view.fill({'discover_type': d_type,
+               'username': getattr(credential, 'principal', None),
+               'password': getattr(credential, 'secret', None),
+               'password_verify': getattr(credential, 'verify_secret', None)})
+    if cancel:
+        view.cancel.click()
+    else:
+        view.start.click()
```
quarckster: I believe some assertion(s) should be here, after the button click the view has to be changed. Also I think this function could return a view, but it can be done later.

cfme/tests/cloud/test_providers.py
```diff
@@ -8,14 +8,16 @@
 
 from utils import error
 from cfme.base.credential import Credential
-from cfme.cloud.provider import (discover, wait_for_a_provider, CloudProvider, prop_region)
+from cfme.cloud.provider import (discover, wait_for_a_provider, CloudProvider)
```
quarckster: Parentheses are not needed here.

cfme/cloud/provider/__init__.py
izapolsk: okay, I'll add checking messages

izapolsk: @quarckster, I've just tried. I can't do that. I'll have to change all tests which use this function then.
Those do all asserts inside

cfme/cloud/provider/__init__.py
```diff
@@ -309,41 +214,45 @@ def step(self, *args, **kwargs):
...
-            providers.add(sel.get_attribute(title, "title"))
-    return providers
-
-
-def discover(credential, cancel=False, d_type="Amazon"):
+    view = navigate_to(CloudProvider, 'All')
+    return [item.name for item in view.items.get_all(surf_pages=True)]
+
+
+def discover(credential, cancel=False, d_type="Amazon EC2"):
```
psav: Did we do this like this in the infra, we should pass in the provider type and then inside this method map that to a string, probably with a dict?

cfme/cloud/provider/__init__.py
```diff
@@ -309,41 +214,45 @@ def step(self, *args, **kwargs):
...
-    if credential:
-        form_data.update({'username': credential.principal,
-                          'password': credential.secret,
-                          'password_verify': credential.verify_secret})
-    fill(discover_form, form_data,
-         action=form_buttons.cancel if cancel else discover_form.start_button,
-         action_always=True)
+    view = navigate_to(CloudProvider, 'Discover')
+    view.fill({'discover_type': d_type})
+    if d_type == 'Amazon EC2':
```
psav: I want to avoid provider specific behaviour in a common method, we should more move these out as helper functions on the providers, this way we don't need to modify the core to support a new type of provider.

PR #4902, opened 2017-06-21
[1LP][RFR] create new REST API test for server_info

mkoura:
Basic test for new items in `server_info`.

{{pytest: -v -k test_server_info_href}}

### DIFF
--- a/cfme/tests/test_rest.py
+++ b/cfme/tests/test_rest.py
@@ -250,6 +250,19 @@ def test_server_info(appliance):
     assert all(item in appliance.rest_api.server_info for item in ('appliance', 'build', 'version'))
 
 
+@pytest.mark.uncollectif(lambda: current_version() < '5.8')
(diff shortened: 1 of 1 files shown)


cfme/tests/test_rest.py
```diff
@@ -250,6 +250,19 @@ def test_server_info(appliance):
...
+def test_server_info_href(appliance):
+    """Check that appliance's server, zone and region is present.
+
+    Metadata:
+        test_flag: rest
+    """
+    items = ('server_href', 'zone_href', 'region_href')
+    for item in items:
+        assert item in appliance.rest_api.server_info
+        assert 'id' in appliance.rest_api.get(appliance.rest_api.server_info[item])
```
quarckster: Why is this test not parametrized?

mkoura: Because it would be overkill to have distinct test for every new attribute.

PR #4909, opened 2017-06-22
[RFR] rhos in sprout doesn't require events endpoint

izapolsk:
Events endpoint won't be necessary for RHOS cloud providers if it is absent in config_file


mshriver:
PRT py.test execution was started, which is verification of the fix, were before it was raising exception after template checks.

PR #4915, opened 2017-06-26
[RFR] fix for providers which don't have ipaddress or hostname

izapolsk:
it turned out that my changes in Cloud providers break  appliance.check_no_conflicting_providers in some cases. This is the fix for that issue.


izapolsk:
I can't reproduce this issue locally
So, it has been tested by @gshefer  in his environment.

PR #4916, opened 2017-06-26
[1LP][RFR] check that default region is present in REST API

mkoura:
Test checking that the default region is present in the REST API

{{pytest: -v -k test_default_region}}

### DIFF
--- a/cfme/tests/test_rest.py
+++ b/cfme/tests/test_rest.py
@@ -263,6 +263,17 @@ def test_server_info_href(appliance):
         assert 'id' in appliance.rest_api.get(appliance.rest_api.server_info[item])
 
 
+@pytest.mark.uncollectif(lambda: current_version() < '5.8')
+def test_default_region(appliance):
+    """Check that the default region is present.
+
+    Metadata:
+        test_flag: rest
+    """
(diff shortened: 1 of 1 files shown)


cfme/tests/test_rest.py
```diff
@@ -263,6 +263,17 @@ def test_server_info_href(appliance):
...
 
 
+@pytest.mark.uncollectif(lambda: current_version() < '5.8')
+def test_default_region(appliance):
+    """Check that the default region is present.
+
+    Metadata:
+        test_flag: rest
+    """
+    reg = appliance.rest_api.collections.regions.get(region=0)
```
mfalesni: This assumption is wrong as we only have region 0 in Sprout, in reality, the region id can be any.

mkoura: @mfalesni Are you saying that region 0 might not be present at all in manageiq? On my appliances there's just one region by default and it's always 0 (it's value of the "region" attribute, not id of the resource).

mkoura: @mfalesni After re-reading I think I get it, region 0 is there because Sprout set it that way, correct?

PR #4918, opened 2017-06-27
[RFR] fix for check_no_conflicting_providers which fails on gce

izapolsk:
It turned out that appliance.check_no_conflicting_providers didn't check existance of gce properly.
When GCE was converted to endpoints/widgetastic  + some consequent changes, that method started to fail.
This PR makes check_no_conflicting_providers ignore gce providers.
I expect to fix checking for existance of gce providers a bit later when I figure out how to compare gce providers.

{{pytest: -v cfme/tests/cloud/test_providers.py cfme/tests/infrastructure/test_providers.py}}

### DIFF
--- a/utils/appliance/__init__.py
+++ b/utils/appliance/__init__.py
@@ -564,6 +564,7 @@ class IPAppliance(object):
          automation issues.
         """
         from utils.providers import list_providers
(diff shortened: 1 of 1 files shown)


izapolsk:
58 tests fail because of absent BreadCrumb

PR #4920, opened 2017-06-28
[1LP][RFR] Fix test_verify_revert_snapshot

apagac:
This fixes test_verify_revert_snapshot in test_snapshot.py:

 - changed used template. VMs based on DamnSmallLinux were without networking and I needed to ssh to the VM
 - created waiting for ssh to become available using wait_for. Comments in code describe how it works
 - changed soft assert checking. It was failing, so I introduced regular expressions and it's working now
 - changed checking for result of the test; added asserts; added checking for both txt files

### DIFF
--- a/cfme/tests/infrastructure/test_snapshot.py
+++ b/cfme/tests/infrastructure/test_snapshot.py
@@ -41,7 +41,7 @@ def domain(request):
 @pytest.fixture(scope="module")
 def test_vm(setup_provider_modscope, provider, vm_name, request):
     """Fixture to provision appliance to the provider being tested if necessary"""
-    vm = VM.factory(vm_name, provider, template_name=provider.data['small_template'])
+    vm = VM.factory(vm_name, provider, template_name=provider.data['full_template']['name'])
 
     if not provider.mgmt.does_vm_exist(vm_name):
         vm.create_on_provider(find_in_cfme=True, allow_skip="default")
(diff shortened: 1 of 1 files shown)


cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -110,6 +111,14 @@ def test_verify_revert_snapshot(test_vm, provider, soft_assert, register_event,
...
     with SSHClient(**ssh_kwargs) as ssh_client:
+        # We need to wait for ssh to become available on the vm, it can take a while. Without
+        # this wait, the ssh command would fail with 'port 22 not available' error.
+        # Easiest way to solve this is just mask the exception with 'handle_exception = True'
+        # and wait for successful completition of the ssh command.
+        # The 'fail_func' ensures we close the connection that failed with exception.
+        # Without this, the connection would hang there and wait_for would fail with timeout.
+        wait_for(lambda: ssh_client.run_command('touch snapshot1.txt').rc == 0, num_sec=300,
+                 delay=20, handle_exception=True, fail_func=ssh.close())
         ssh_client.run_command('touch snapshot1.txt')
```
mfalesni: With the previous change, you technically do not need to run this command since the successful execution of the wait_for lambda will do exactly the same.

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -109,33 +109,41 @@ def test_verify_revert_snapshot(test_vm, provider, soft_assert, register_event,
...
         'hostname': ip
     }
-    with SSHClient(**ssh_kwargs) as ssh_client:
-        ssh_client.run_command('touch snapshot1.txt')
-        snapshot1.create()
-        ssh_client.run_command('touch snapshot2.txt')
-        snapshot2 = new_snapshot(test_vm)
-        snapshot2.create()
-        snapshot1.revert_to()
+    ssh_client = SSHClient(**ssh_kwargs)
```
apagac: I can't use `with` statement here. `SSHClient` has in its `__enter__` method `connect`, which is failing. `wait_for` is suited for masking these failures and thus I can wait for successful ssh connection.
In this case, ssh connection is failing because the VM we just created on a provider does not yet have its IP assigned, although it's already visible in CFME UI.

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -109,33 +109,41 @@ def test_verify_revert_snapshot(test_vm, provider, soft_assert, register_event,
...
-            logger.info('Revert to snapshot %s Failed', snapshot1.name)
+    current_state = test_vm.find_quadicon().state
+    soft_assert(current_state.startswith('currentstate-on'),
+                "Quadicon state is {}".format(current_state))
+    soft_assert(test_vm.provider.mgmt.is_vm_running(test_vm.name), "vm not running")
+    wait_for(lambda: ssh_client.run_command('test -e snapshot1.txt').rc == 0,
+             num_sec=400, delay=20, handle_exception=True, fail_func=ssh_client.close())
+    try:
+        result = ssh_client.run_command('test -e snapshot1.txt')
+        assert result.rc == 0
```
izapolsk: 0 is always treated as False, whereas any other number is treated as True. So, you don't need this comparison statement. the same for statement below

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -109,33 +109,41 @@ def test_verify_revert_snapshot(test_vm, provider, soft_assert, register_event,
...
+    snapshot1.create()
+    ssh_client.run_command('touch snapshot2.txt')
+    snapshot2 = new_snapshot(test_vm)
+    snapshot2.create()
+    snapshot1.revert_to()
     # Wait for the snapshot to become active
     logger.info('Waiting for vm %s to become active', snapshot1.name)
     wait_for(snapshot1.wait_for_snapshot_active, num_sec=300, delay=20, fail_func=sel.refresh)
     test_vm.wait_for_vm_state_change(desired_state=test_vm.STATE_OFF, timeout=720)
-    register_event(target_type='VmOrTemplate', target_name=test_vm.name,
```
izapolsk: I agree that you don't need to check these events. But it would be good to check that snapshot related events come

PR #4921, opened 2017-06-28
[RFR] Fix containers navigation in 5.8

gshefer:
{{pytest: cfme/tests/containers/test_properties.py -v --use-provider cm-env2}}

- Fix navigation to Containers ('All' view) which crashes due to a change in the accordion that introduced in 5.8.1+

### DIFF
--- a/cfme/containers/container.py
+++ b/cfme/containers/container.py
@@ -4,15 +4,20 @@ import random
 import itertools
 
 from navmazing import NavigateToSibling, NavigateToAttribute
+from widgetastic_manageiq import Accordion, ManageIQTree, View, Table
+from widgetastic_patternfly import VerticalNavigation
 
+from cfme.base.login import BaseLoggedInPage
+from cfme.containers.provider import details_page, pol_btn, mon_btn
 from cfme.common import SummaryMixin, Taggable
 from cfme.fixtures import pytest_selenium as sel
-from cfme.web_ui import CheckboxTable, toolbar as tb, paginator, match_location, accordion,\
-    PagedTable
-from utils import version
+from cfme.web_ui import CheckboxTable, toolbar as tb, paginator, match_location, PagedTable
 from utils.appliance import Navigatable
 from utils.appliance.implementations.ui import CFMENavigateStep, navigator, navigate_to
-from cfme.containers.provider import details_page, pol_btn, mon_btn
+from utils import version
+from widgetastic.widget import Text
+from widgetastic.xpath import quote
+from widgetastic.utils import Version, VersionPick
 
 
 list_tbl = CheckboxTable(table_locator="//div[@id='list_grid']//table")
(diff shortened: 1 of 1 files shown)


gshefer:
This PR depends on this one: https://github.com/ManageIQ/integration_tests/pull/4928

cfme/containers/container.py
```diff
@@ -59,21 +61,45 @@ def get_random_instances(cls, provider, count=1, appliance=None):
                 for obj in itertools.islice(containers_list, count)]
 
 
+class ContainerAllView(BaseLoggedInPage):
+
+    containers = Table(locator="//div[@id='list_grid']//table")
+
+    @View.nested
+    class filters(Accordion):  # noqa
+
```
mshriver: **Optional**
Condense this a bit, no need for all the empty lines

cfme/containers/container.py
```diff
@@ -59,21 +61,45 @@ def get_random_instances(cls, provider, count=1, appliance=None):
...
+        ACCORDION_NAME = "Filters"
+
+        @View.nested
+        class navigation(VerticalNavigation):
+
+            DIV_LINKS_MATCHING = './/div/ul/li/a[contains(text(), {txt})]'
```
mshriver: @rsnyman Check this out - one way of 'fixing' the bug we were talking about today.

@gshefer rsynman is working on a widgetsatic VerticalNavigation change to support this on multiple pages instead of having to replicate your overriding here.  Plan is to define a second locator constant for contains (including href like the original one) and defining a parameter for partial matches (`VerticalNavigation(partial=True)`)

mshriver: Depending on the timing, we may merge this with a TODO referencing @rsnyman coming PR.

cfme/containers/container.py
```diff
@@ -59,21 +61,45 @@ def get_random_instances(cls, provider, count=1, appliance=None):
...
+
+            def __init__(self, parent, logger=None):
+                VerticalNavigation.__init__(self, parent, '#Container_def_searches', logger=logger)
+        tree = ManageIQTree()
```
mshriver: Do you have a tree and vertical navigation here? I thought it was just the vertical nav elements.

cfme/containers/container.py
```diff
@@ -59,21 +61,45 @@ def get_random_instances(cls, provider, count=1, appliance=None):
...
+
+    @property
+    def is_displayed(self):
+        return match_page(summary=version.pick({
```
mshriver: This works, but I definitely recommend defining a Text widget for the page title, and validating against self.navigation.currently_selected and self.title. You can still version pick the title using widgetastic version support.

cfme/containers/container.py
```diff
@@ -59,21 +61,45 @@ def get_random_instances(cls, provider, count=1, appliance=None):
...
+        }))
+
+
 @navigator.register(Container, 'All')
 class ContainerAll(CFMENavigateStep):
     prerequisite = NavigateToAttribute('appliance.server', 'LoggedIn')
-
-    def am_i_here(self):
-        return match_page(summary='All Containers')
+    VIEW = ContainerAllView
```
mshriver: please stick this line above prerequisite

cfme/containers/container.py
mshriver: IGNORE this comment, I see below where its version dependent.

cfme/containers/container.py
gshefer: @mshriver I really prefer to merge this ASAP with a TODO because this PR fix blocker for ~50 tests in our CI (all the tests that have navigation to this page).

cfme/containers/container.py
gshefer: Yep,
5.7 -> tree
5.8 -> navigation
(look at ContainerAll.resetter)

cfme/containers/container.py
mshriver: Given your time constraints, a TODO would be acceptable here as the page is fully converted.

cfme/containers/container.py
```diff
@@ -59,21 +61,45 @@ def get_random_instances(cls, provider, count=1, appliance=None):
                 for obj in itertools.islice(containers_list, count)]
 
 
+class ContainerAllView(BaseLoggedInPage):
+
```
mshriver: **Required**
Add docstrings to new classes

gshefer:
@mshriver @psav I moved this PR to BLOCKED since we have no option to test it due to [BZ#1467639](https://bugzilla.redhat.com/show_bug.cgi?id=1467639). All our tests are navigating twice to ContainersAll so if we would have some option to run PRT with some navigate_to(Container, 'All') only to verify - we could test this and merge.

gshefer:
@psav @mshriver moved from BLOCKER to RFR. seems like the issue fixed. All PRT tests are passing.

cfme/containers/container.py
```diff
@@ -61,21 +66,44 @@ def get_random_instances(cls, provider, count=1, appliance=None):
...
+    @View.nested
+    class Filters(Accordion):  # noqa
+        ACCORDION_NAME = "Filters"
+
+        @View.nested
+        class Navigation(VerticalNavigation):
+            DIV_LINKS_MATCHING = './/div/ul/li/a[contains(text(), {txt})]'
+
+            def __init__(self, parent, logger=None):
+                VerticalNavigation.__init__(self, parent, '#Container_def_searches', logger=logger)
```
psav: **REQUIRED:** Can we have a line break here to split them up

gshefer: Done.

PR #4922, opened 2017-06-28
[RFR] Added Pod.pods_per_ready_status + Fix smart management test

gshefer:
{{pytest: cfme/tests/containers/test_reload_button_provider.py cfme/tests/containers/test_properties.py cfme/tests/containers/test_reports.py cfme/tests/containers/test_smart_management.py -v --use-provider cm-env2}}

- Created Pod.pods_per_ready_status in order to be able to share this usage between test_pods_conditions and test_report_pods_per_ready_status
- Fixed test_smart_management which didn't wait enough time to the tags to be populated.

### DIFF
--- a/cfme/containers/provider/__init__.py
+++ b/cfme/containers/provider/__init__.py
@@ -6,12 +6,12 @@ from random import sample
 
 
 from navmazing import NavigateToSibling, NavigateToAttribute
-from cfme.base.login import BaseLoggedInPage
 from widgetastic_patternfly import SelectorDropdown, Button, Dropdown
 from widgetastic.widget import Text
-from utils.wait import wait_for
+from wrapanapi.utils import eval_strings
 
 
+from cfme.base.login import BaseLoggedInPage
 from cfme.common.provider import BaseProvider
 from cfme import exceptions
 from cfme.fixtures import pytest_selenium as sel
@@ -26,6 +26,7 @@ from utils.browser import ensure_browser_open
 from utils.pretty import Pretty
 from utils.varmeth import variable
 from utils.log import logger
+from utils.wait import wait_for
 
 
 paged_tbl = PagedTable(table_locator="//div[@id='list_grid']//table")
@@ -269,6 +270,18 @@ class ContainersProvider(BaseProvider, Pretty):
     def num_image_registry_ui(self):
         return int(self.get_detail("Relationships", "Image Registries"))
 
+    def pods_per_ready_status(self):
+        """Grabing the Container Statuses Summary of the pods from API"""
+        #  TODO: Add later this logic to wrapanapi
+        entities = self.mgmt.api.get('pod')[1]['items']
+        out = {}
+        for entity_j in entities:
+            out[entity_j['metadata']['name']] = {
+                condition['type']: eval_strings([condition['status']]).pop()
+                for condition in entity_j['status'].get('conditions', [])
+            }
+        return out
+
 
 @navigator.register(ContainersProvider, 'All')
 class All(CFMENavigateStep):
@@ -430,6 +443,25 @@ class ContainersTestItem(object):
         return '{} ({})'.format(
             getattr(self.obj, '__name__', str(self.obj)),
             self.polarion_id)
+
+    @classmethod
+    def get_pretty_id(cls, obj):
+        """Since sometimes the test object is wrapped within markers,
+        it's difficult to find get it inside the args tree.
+        hence we use this to get the object and all pretty_id function.
+
+        Args:
+            * obj: Either a ContainersTestItem or a marker that include it
+        returns:
+            str pretty id
+        """
+        if isinstance(obj, cls):
+            return obj.pretty_id()
+        elif hasattr(obj, 'args') and hasattr(obj, '__iter__'):
+            for arg in obj.args:
+                pretty_id = cls.get_pretty_id(arg)
+                if pretty_id:
+                    return pretty_id
 
 
 class Labelable(object):
--- a/cfme/tests/containers/test_properties.py
+++ b/cfme/tests/containers/test_properties.py
@@ -1,12 +1,12 @@
 # -*- coding: utf-8 -*-
 import pytest
+from wrapanapi.utils import eval_strings
 
 from cfme.containers.provider import ContainersProvider,\
     ContainersTestItem
 from cfme.containers.route import Route
 from cfme.containers.project import Project
 from cfme.containers.service import Service
-from cfme.containers.container import Container
 from cfme.containers.node import Node
 from cfme.containers.image import Image
 from cfme.containers.image_registry import ImageRegistry
@@ -17,7 +17,6 @@ from cfme.containers.volume import Volume
 from utils import testgen, version
 from utils.version import current_version
 from utils.soft_get import soft_get
-from utils.appliance.implementations.ui import navigate_to
 
 
 pytestmark = [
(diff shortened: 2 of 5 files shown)


gshefer:
Dependency: https://github.com/ManageIQ/integration_tests/pull/4921

cfme/tests/containers/test_properties.py
```diff
@@ -156,27 +155,24 @@ def test_properties(provider, test_item, soft_assert):
...
+        ose_pod_condition = pods_per_ready_status[pod_name]
+        cfme_pod_condition = {_type:
+                              eval_strings(
+                                  [getattr(getattr(cfme_pod.summary.conditions, _type), "Status")]
+                              ).pop()
+                              for _type in ose_pod_condition}
+
+        for status in cfme_pod_condition:
+            soft_assert(ose_pod_condition[status] == cfme_pod_condition[status],
+                        'Expected "{}" status for pod {} is {}. Found {} instead.'
```
jaryn: The English grammar is bad in the error message and it doesn't match the assert condition. Why it is not saying something like

    The Pod {} status mismatch: It is "{}" in openshift while cfme sees "{}".

Not mentioning that it doesn't make much sense to me why one needs to first group the pods by it status to compare them then. It smells like the Polarion prescription is bad, but I cannot check it in the moment.

cfme/tests/containers/test_smart_management.py
```diff
@@ -58,13 +60,22 @@ def set_random_tag(instance):
     return Tag(display_name=random_tag.text, category=random_cat.text)
 
 
+def wait_for_tag(obj_inst):
+    # Waiting for some tag to appear at "My Company Tags" and return pop'ed last tag
+    try:
+        return wait_for(lambda: getattr(obj_inst.summary.smart_management,
+                                        'my_company_tags', []), fail_condition=[],
+                        num_sec=30, delay=5, fail_func=obj_inst.summary.reload).out.pop()
+    except TimedOutError:
```
jaryn: 1) Why did you choose to silence the exception?

2) Returning none will cause problems at lines 89 and 98

cfme/tests/containers/test_properties.py
gshefer: No problem I'll fix that.

gshefer:
@jaryn This PR includes

gshefer:
@jaryn this PR includes several fixes and changes which are t

cfme/tests/containers/test_properties.py
```diff
@@ -132,8 +134,7 @@
 ]
 
 
-@pytest.mark.parametrize('test_item', TEST_ITEMS,
-                         ids=[ti.args[1].pretty_id() for ti in TEST_ITEMS])
+@pytest.mark.parametrize('test_item', TEST_ITEMS)
```
psav: Why remove this?

cfme/tests/containers/test_properties.py
```diff
@@ -143,7 +144,6 @@ def test_properties(provider, test_item, soft_assert):
 
     for instance in instances:
 
-        navigate_to(instance, 'Details')
```
psav: We don't need to navigate anymore?

cfme/tests/containers/test_reload_button_provider.py
```diff
@@ -15,42 +13,10 @@
...
-        Images are being validated separately, since the total
-        number of images in CFME 5.7 and CFME 5.8 includes all images from the OSE registry as well
-        as the images that are being created from the running pods. The images are searched
-        according to the @sha.
+        button.
     """
 
-    navigate_to(provider, 'Details')
-    tb.select('Reload Current Display')
     provider.validate_stats(ui=True)
```
psav: Is the refresh done as part of this func call?

cfme/tests/containers/test_reports.py
```diff
@@ -91,24 +76,24 @@ def test_container_reports_base_on_options(soft_assert):
         soft_assert(option, 'Could not find option "{}" for base report on.'.format(base_on))
 
 
-@pytest.mark.meta(blockers=[BZ(1435958, forced_streams=["5.8"])])
+@pytest.mark.meta(blockers=[BZ(1467059, forced_streams=["5.8"])])
```
psav: If the bug applies to all the tests in the file, you can mark the test at the beginning of the file and apply it to all

cfme/tests/containers/test_properties.py
gshefer: No, The instance.summary below is already do it for us.

cfme/tests/containers/test_properties.py
gshefer: causes problem when there is a bugzilla marker, it's too complicating the way to call that (sometimes args[1].pretty_id, sometimes args[0].args[1].pretty_id). I can create some function that treat this but i think it's too much for this purpose. what you think?

cfme/tests/containers/test_reload_button_provider.py
gshefer: [https://github.com/ManageIQ/integration_tests/blob/master/cfme/common/provider.py#L454](https://github.com/ManageIQ/integration_tests/blob/master/cfme/common/provider.py#L454)

cfme/fixtures/pods_per_ready_status.py
```diff
@@ -0,0 +1,16 @@
+import pytest
+from wrapanapi.utils import eval_strings
+
+
+@pytest.fixture(scope='function')
+def pods_per_ready_status(provider):
```
psav: I dont see a need for this to be a fixture, this could just be a method on the container object

gshefer: OK, so it should be on the Pod object as static method for all pods (since we want just 1 API call)

cfme/tests/containers/test_properties.py
psav: I need to have something that is either a string, or has a .name attribute for a parameter else it will screw up future reporting.

cfme/containers/pod.py
```diff
@@ -57,6 +59,19 @@ def get_random_instances(cls, provider, count=1, appliance=None):
         return [cls(obj.name, obj.project_name, provider, appliance=appliance)
                 for obj in itertools.islice(pod_list, count)]
 
+    @staticmethod
```
psav: So, a few things here,
1) This function/method is dealing with pods from a specific provider
Therefore, either this function should be a method of the provider object, then there is no need to pass a provider to it.

cfme/containers/pod.py
```diff
@@ -3,6 +3,8 @@
 import random
 import itertools
 
+from wrapanapi.utils import eval_strings
```
mshriver: lint bot isn't running right now, this isn't used anywhere in the module?

cfme/containers/provider/__init__.py
```diff
@@ -6,12 +6,12 @@
...
 
 from navmazing import NavigateToSibling, NavigateToAttribute
-from cfme.base.login import BaseLoggedInPage
 from widgetastic_patternfly import SelectorDropdown, Button, Dropdown
 from widgetastic.widget import Text
-from utils.wait import wait_for
+from wrapanapi.utils import eval_strings
 
 
+from cfme.base.login import BaseLoggedInPage
```
mshriver: :+1:

cfme/containers/provider/__init__.py
```diff
@@ -269,6 +270,18 @@ def num_image_registry(self):
     def num_image_registry_ui(self):
         return int(self.get_detail("Relationships", "Image Registries"))
 
+    def pods_per_ready_status(self):
+        """Grabing the Container Statuses Summary of the pods from API"""
+        #  TODO: Add later this logic to wrapanapi
+        entities_j = self.mgmt.api.get('pod')[1]['items']
```
mshriver: why `_j` on this variable name? nitpicking, but it doesn't seem to make sense, `entities` would be sufficient.

cfme/containers/provider/__init__.py
```diff
@@ -269,6 +270,18 @@ def num_image_registry(self):
     def num_image_registry_ui(self):
         return int(self.get_detail("Relationships", "Image Registries"))
 
+    def pods_per_ready_status(self):
+        """Grabing the Container Statuses Summary of the pods from API"""
+        #  TODO: Add later this logic to wrapanapi
```
mshriver: https://github.com/ManageIQ/wrapanapi/issues/152

gshefer: (+ @psav ) Indeed, I didn't forgot that, my next step will be to migrate all these functions to wrapanapi (this is not the only one, I have similar TODOS all over the framework)

PR #4937, opened 2017-06-30
[1LP][RFR] Discovery Infra generify

psav:
Needs more work to remove the Form based stuff too

### DIFF
--- a/cfme/infrastructure/provider/rhevm.py
+++ b/cfme/infrastructure/provider/rhevm.py
@@ -96,3 +96,8 @@ class RHEVMProvider(InfraProvider):
                    start_ip=start_ip,
                    end_ip=end_ip,
                    appliance=appliance)
+
+    @staticmethod
+    def discover_dict():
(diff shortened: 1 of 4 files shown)


izapolsk:
:+1:

cfme/infrastructure/provider/rhevm.py
```diff
@@ -96,3 +96,8 @@ def from_config(cls, prov_config, prov_key, appliance=None):
                    start_ip=start_ip,
                    end_ip=end_ip,
                    appliance=appliance)
+
+    @staticmethod
+    def discover_dict():
+        """Returns the discovery dict"""
+        return {"rhevm": True}
```
mfalesni: This might as well be just a class attribute, I see no reason why this should be hidden in a function call :)

psav: Yeh I can, it was just to maintain consistency with the others, but sure

PR #4946, opened 2017-07-04
[RFR] New test coverage for Timezones

lcouzens:
New test coverage for timezone setting through appliance_console
{{pytest: -k test_black_console_set_timezone}}

### DIFF
--- a/cfme/tests/cli/test_appliance_console.py
+++ b/cfme/tests/cli/test_appliance_console.py
@@ -11,6 +11,20 @@ import yaml
 
 TimedCommand = namedtuple('TimedCommand', ['command', 'timeout'])
 LoginOption = namedtuple('LoginOption', ['name', 'option', 'index'])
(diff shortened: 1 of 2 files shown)


cfme/tests/cli/test_appliance_console.py
```diff
@@ -46,6 +46,18 @@ def is_hostname_set(appliance):
     assert return_code == 0
```
psav:
```python
from collections import namedtuple

TZ = namedtuple('TimeZone', ['name', 'option'])
tzs = [
    TZ('Antarctica/Casey', '3'),
    TZ('Arctic/Longyearsyen', '4')
    ...
]
@pytest.mark.parametrize('timezone', tzs, idlist=[tz.name for tz in tzs])
def test_black.......
    ...
    command_set = ('ap', '', opt, timezone.option......
    ...
```

PR #4952, opened 2017-07-06
[1LP][RFR]fix for azure provision

lkhomenk:
Purpose or Intent
=================

__Fixing__ auto_placement provision for azure - one argument was missed

{{pytest: --long-running --use-provider azure  -k "azure-Auto"}}


lkhomenk:
manual provision can't pass after another test. as we are still blocked by sticky "instance details" page.

izapolsk:
PRT is green. It hung after running tests

PR #4953, opened 2017-07-06
[1LP][RFR] Update class.exists() methods to be properties

mshriver:
most were, just making it consistent.
Fixed all the calls using .exists() for .exists

Not including sentaku implementations here because of: 
https://github.com/RonnyPfannschmidt/Sentaku/issues/12

If Sentaku PR is merged to fix ContextualProperty, Zone.exists can be implemented as a property as well. https://github.com/RonnyPfannschmidt/Sentaku/pull/14/files

FIXES RHCFQE-3249

PRT might be a bit crazy because I'm not intending to fix any modified tests that may not be working, limiting scope to just fixing the exists calls.

# PRT Results
Failures, but nothing to do with any of the calls to exists properties that I've modified here.

### DIFF
--- a/cfme/infrastructure/pxe.py
+++ b/cfme/infrastructure/pxe.py
@@ -167,6 +167,7 @@ class PXEServer(Updateable, Pretty, Navigatable):
             flash.assert_message_match('Add of new PXE Server was cancelled by the user')
 
     @variable(alias="db")
+    @property
     def exists(self):
         """
         Checks if the PXE server already exists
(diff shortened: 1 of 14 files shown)


cfme/infrastructure/pxe.py
```diff
@@ -369,6 +372,7 @@ def exists(self):
         return self.name in [s.name for s in candidates]
 
     @exists.variant('ui')
+    @property
```
mfalesni: @mshriver hmm, does this combination actually work?

mshriver: It may not, like with the Sentaku instances. Our tests for pxe never get to calling it because rhevm36 infra isn't setup correctly for CFME to display the pxe servers

mfalesni: I looked into it and varmeth probably cannot work with properties properly.

PR #4956, opened 2017-07-07
[1LP][RFR][NOTEST]Refactor cleanup_old_vms

mshriver:
Depends on: ~~https://github.com/ManageIQ/wrapanapi/pull/151~~

vsphere no longer requires connecting to the host to get
vm_creation_date for a powered off VM, greatly simplifies this script,
removing the need for a separate method for these vm types.

Converted to using MultiProcessing Process + Queue instead of threads
and global dicts

MP applied to delete with a process for each prov+vm tuple, instead of
threaded per-provider which is really slow (vm deletes were sequential)

FIXES RHCFQE-3256

### DIFF
--- a/scripts/cleanup_old_vms.py
+++ b/scripts/cleanup_old_vms.py
@@ -5,7 +5,7 @@ import re
 import sys
 from collections import namedtuple
 from operator import attrgetter
-from multiprocessing import Process, Queue
+from multiprocessing import Process, Manager, Pool
 
 import pytz
 from tabulate import tabulate
@@ -100,7 +100,7 @@ def scan_vm(provider_key, vm_name, delta, match_queue, non_match_queue, scan_fai
             for message in messages:
                 logger.exception(message)
                 print(message)
-            scan_failure_queue.put([VmReport(provider_key, vm_name, FAIL, status, NULL)])
+            scan_failure_queue.put(VmReport(provider_key, vm_name, FAIL, status, NULL))
             return
 
     # None of this should raise exceptions
@@ -110,7 +110,7 @@ def scan_vm(provider_key, vm_name, delta, match_queue, non_match_queue, scan_fai
 
     # test age to determine which queue it goes in
     queue = match_queue if delta < vm_delta else non_match_queue
-    queue.put([data])
+    queue.put(data)
 
 
 def scan_provider(provider_key, matchers, delta, match_queue, scan_failure_queue):
@@ -140,24 +140,19 @@ def scan_provider(provider_key, matchers, delta, match_queue, scan_failure_queue
 
         # MP for vm scanning, could be many VMs per provider
         # match_queue is passed through and processed by this method's caller
-        non_time_match_queue = Queue()
-        vm_proc_list = []
-        for vm_name in text_matched_vms:
-            vm_proc_list.append(
-                Process(target=scan_vm,
-                        args=(provider_key, vm_name, delta, match_queue,
-                              non_time_match_queue, scan_failure_queue)))
-
-        for proc in vm_proc_list:
-            proc.start()
-        for proc in vm_proc_list:
-            proc.join()
+        manager = Manager()
+        non_time_match_queue = manager.Queue()
+        vm_scan_pool = Pool(8)
+        vm_scan_args = [
+            (provider_key, vm_name, delta, match_queue, non_time_match_queue, scan_failure_queue)
+            for vm_name in text_matched_vms]
+        vm_scan_pool.map_async(scan_vm, vm_scan_args)
 
         # each item on scan_queue is a VmData tuple
         # process non_match_queue and log its contents now
         non_time_matching = []
         while not non_time_match_queue.empty():
-            non_time_matching.extend(non_time_match_queue.get())
+            non_time_matching.append(non_time_match_queue.get())
 
         print('{}: Finished scanning for matches'.format(provider_key))
         for non_match in non_time_matching:
@@ -209,7 +204,7 @@ def delete_provider_vm(provider_key, vm_name, age, queue):
         logger.exception(message)
     finally:
         print(message)
-        queue.put([VmReport(provider_key, vm_name, age, status, result)])
+        queue.put(VmReport(provider_key, vm_name, age, status, result))
 
 
 def cleanup_vms(texts, max_hours=24, providers=None, prompt=True):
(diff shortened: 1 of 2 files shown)


scripts/cleanup_old_vms.py
```diff
@@ -68,143 +57,141 @@ def match(matchers, vm_name):
...
+        print('{}: Matching text filters: {}'.format(provider_key, text_matched_vms))
+
+        # MP for vm scanning, could be many VMs per provider
+        # match_queue is passed through and processed by this method's caller
+        non_time_match_queue = Queue()
+        vm_proc_list = [
+            Process(target=scan_vm,
+                    args=(provider_key, vm_name, delta, match_queue,
+                          non_time_match_queue, scan_failure_queue))
+            for vm_name in text_matched_vms]
```
mkoura: Consider using normal loop as is recommended when comprehension doesn't fit on single line

scripts/cleanup_old_vms.py
```diff
@@ -68,143 +57,141 @@ def match(matchers, vm_name):
...
-        finally:
-            with lock:
-                deleted_vms_list.append([provider_key, vm_name, age, status, result])
-    with lock:
-        print("{} is done deleting vms".format(provider_key))
+    except Exception:  # ANYTHING could happen!
+        result = FAIL  # set this here to cover anywhere the exception could happen
+        message = '{}: Exception during delete: {}'.format(provider_key, vm_name)
+        logger.exception(message)
+        pass
```
mkoura: REQUIRED: remove

scripts/cleanup_old_vms.py
mshriver: I'm torn about this in general, because a lot of places it does make sense (and we have in our codebase) to use comprehension on multiple lines.

Here I think you're right, and it will be more clear in a regular loop.

scripts/cleanup_old_vms.py
mshriver: yup. thanks :)

scripts/cleanup_old_vms.py
```diff
@@ -44,167 +35,181 @@ def parse_cmd_line():
...
+        non_text_matching = set(vm_list) - set(text_matched_vms)
+        print('{}: NOT matching text filters: {}'.format(provider_key, non_text_matching))
+        print('{}: Matching text filters: {}'.format(provider_key, text_matched_vms))
+
+        # MP for vm scanning, could be many VMs per provider
+        # match_queue is passed through and processed by this method's caller
+        non_time_match_queue = Queue()
+        vm_proc_list = []
+        for vm_name in text_matched_vms:
+            vm_proc_list.append(
```
izapolsk: REQUIRED: so, if there are 100 matching vms, you will start 100 processes. There should be some limitation to 10 processes for example.

scripts/cleanup_old_vms.py
```diff
@@ -44,167 +35,181 @@ def parse_cmd_line():
...
+        for proc in vm_proc_list:
+            proc.start()
+        for proc in vm_proc_list:
+            proc.join()
+
+        # each item on scan_queue is a VmData tuple
+        # process non_match_queue and log its contents now
+        non_time_matching = []
+        while not non_time_match_queue.empty():
+            non_time_matching.extend(non_time_match_queue.get())
```
izapolsk: could you please explain why you put list with one element everywhere and use extend when retrieve it from queue ?

scripts/cleanup_old_vms.py
```diff
@@ -217,75 +222,87 @@ def cleanup_vms(texts, max_hours=24, providers=None, prompt=True):
...
+    scan_proc_list = []
+    for provider_key in providers_to_scan:
+        scan_proc_list.append(
+            Process(target=scan_provider,
+                    args=(provider_key, matchers, delta, scan_queue, scan_failure_queue)))
+
+    for proc in scan_proc_list:
+        proc.start()
+    for proc in scan_proc_list:
+        proc.join()
```
izapolsk: actually, it doesn't make much sense here but this is potential bottleneck.

scripts/cleanup_old_vms.py
```diff
@@ -217,75 +222,87 @@ def cleanup_vms(texts, max_hours=24, providers=None, prompt=True):
...
+
+    # each item on scan_queue is a VmData tuple
+    vms_to_delete = []
+    while not scan_queue.empty():
+        vms_to_delete.extend(scan_queue.get())
 
     if vms_to_delete and prompt:
-        yesno = raw_input('Delete these VMs? [y/N]: ')
+        yesno = raw_input('Delete the-se VMs? [y/N]: ')
```
izapolsk: typo ?

scripts/cleanup_old_vms.py
```diff
@@ -217,75 +222,87 @@ def cleanup_vms(texts, max_hours=24, providers=None, prompt=True):
...
 
     if not vms_to_delete:
         print('No VMs to delete.')
 
-    thread_queue = []
-    for provider_key, vm_set in vms_to_delete.items():
-        provider_mgmt = get_mgmt(provider_key)
+    delete_queue = Queue()
+    delete_proc_list = []
+    for provider_key, vm_name, age in vms_to_delete:
```
izapolsk: ^^

scripts/list_provider_vms.py
```diff
@@ -136,11 +136,15 @@ def list_vms(provider_key, output_queue):
     # Now pull all the results off of the queue
     # Stacking the generator this way is equivalent to using list.extend instead of list.append
     # Need to check queue.empty since a call to get will raise an Empty exception
-    output_data = [data for _ in proc_list if not queue.empty() for data in queue.get()]
+    output_data = []
+    for _ in proc_list:
```
izapolsk: can you please expain why you iterate over proc_list here ?

scripts/cleanup_old_vms.py
izapolsk: I implemented such things in past by creating master process/thread and number of workers

scripts/cleanup_old_vms.py
```diff
@@ -44,167 +35,181 @@ def parse_cmd_line():
...
 
+    # None of this should raise exceptions
+    vm_delta = now - vm_creation_time
+    print('{}: VM {} age: {}'.format(provider_key, vm_name, vm_delta))
+    data = VmData(provider_key, vm_name, str(vm_delta))
 
-def process_provider_vms(provider_key, matchers, delta, vms_to_delete):
+    # test age to determine which queue it goes in
+    queue = match_queue if delta < vm_delta else non_match_queue
+    queue.put([data])
```
mshriver: Because of how the queue data is handled on `get()`.

scripts/cleanup_old_vms.py
mshriver: heh, yep.

scripts/cleanup_old_vms.py
mshriver: explain? I posted this pattern in the dev chat a while ago and didn't get any negative feedback?

scripts/cleanup_old_vms.py
mshriver: not sure what you're pointing out here?

scripts/list_provider_vms.py
mshriver: thought I had removed these, no need.

scripts/cleanup_old_vms.py
mshriver: IIRC the way things are pulled off the Queue was causing issues where only the first item in the tuple was fetched, instead of the tuple that is the first item of a list.

I then extend here to handle the fact that its a list, but I want the element from inside the list.

scripts/cleanup_old_vms.py
mshriver: Going to try this.

scripts/cleanup_old_vms.py
izapolsk: not sure I got, you use extend on get(). What interferes you to use append on get and item instead of list here ?

scripts/cleanup_old_vms.py
izapolsk: Mike this is just comment. I don't ask you to change this. It is ok for this script.
However,  you understand that script is stuck here until all jobs finish.

scripts/cleanup_old_vms.py
izapolsk: I just highlighted hte piece of code where you probably fork a lot of processes again

scripts/cleanup_old_vms.py
mshriver: ah gotcha - absolutely, and that's the intent, to wait until all the scans are done to process the Queue of the matching VM's

scripts/cleanup_old_vms.py
mshriver: gotcha gotcha.

mshriver:
Bumping myself back down to 1st level, @izapolsk raised a great point about controlling the potentially high number of processes.  Implementing a much cleaner multiprocessing.Pool implementation with a better overall control flow.

dajoRH:
Lint report for commit c57337e6f4c84aabbd80a6952be34e26da597772:

`scripts/cleanup_old_vms.py`:
- :red_circle: Line 10:1: **F401** *'multiprocessing.Manager' imported but unused*
- :bangbang: Line 258:19: **E225** *missing whitespace around operator*

**No commit flaws detected**.

*CFME QE Bot*

mshriver:
Thanks @izapolsk, was a bit of a headache but I'm much happier with what's here now, control/flow wise.

Will be testing thoroughly today!

PR #4964, opened 2017-07-10
[RFR]Replace change_stored_password

rlbabyuk:
Purpose or Intent
=================
Remove cfme.web_ui change_stored_password() usage from access_control.py
{{pytest: -v -k test_user_crud}}


cfme/configure/access_control.py
```diff
@@ -212,7 +213,6 @@ def copy(self):
         view = self.create_view(AddUserView)
         new_user = User(name="{}copy".format(self.name),
                         credential=Credential(principal='redhat', secret='redhat'))
-        change_stored_password()
```
psav: This is no longer required??

rlbabyuk: no, on copy user page we don't have 'change password' option, as we didn't set up any password at all. We have it only for edit user page

PR #4969, opened 2017-07-11
[1LP][RFR] Add widgetastic common taggable class

rlbabyuk:
Purpose or Intent
=================
Add common class for tagging, for already converted classes to widgetastic.
{{pytest: -v -k 'test_tagvis_catalog_items or test_edit_tags'}}

### DIFF
--- a/cfme/common/__init__.py
+++ b/cfme/common/__init__.py
@@ -184,13 +184,16 @@ class TagPageView(BaseLoggedInPage):
 class WidgetasticTaggable(object):
     """
         This class can be inherited by any class that honors tagging.
-        Class should have 'EditTags' navigate step registered.
+        Class should have 'EditTags' navigate step registered,
+        also 'Details' view should be present with tag attribute(widget with tag
+            information on details page)
+
         This class should be used for already converted to widgetastic classes
 
         It provides functionality to assign and unassigned tags.
     """
 
-    def edit_tag(self, category=None, tag=None, cancel=False, reset=False):
+    def add_tag(self, category=None, tag=None, cancel=False, reset=False):
         """ Add tag to tested item
             Args:
                 category: category(str)
@@ -217,7 +220,7 @@ class WidgetasticTaggable(object):
             cancel = True
         self._tags_action(view, cancel, reset)
 
-    def edit_tags(self, tags):
+    def add_tags(self, tags):
         """Add multiple tags
             Args:
                 tags: pass dict with category name as key, and tag as value,
@@ -225,10 +228,10 @@ class WidgetasticTaggable(object):
         """
         if isinstance(tags, dict):
             for category, tag in tags.items():
-                self.edit_tag(category=category, tag=tag)
+                self.add_tag(category=category, tag=tag)
         elif isinstance(tags, (list, tuple)):
             for tag in tags:
-                self.edit_tag(tag=tag)
+                self.add_tag(tag=tag)
 
     def remove_tag(self, category=None, tag=None, cancel=False, reset=False):
         """ Remove tag of tested item
--- a/cfme/common/__init__.py
+++ b/cfme/common/__init__.py
@@ -203,15 +203,18 @@ class WidgetasticTaggable(object):
             category = tag.category.display_name
             tag = tag.display_name
         try:
-            view.fill({
+            updated = view.fill({
                 "tag_category": '{} *'.format(category),
                 "tag_name": tag
             })
         except NoSuchElementException:
-            view.fill({
+            updated = view.fill({
                 "tag_category": category,
                 "tag_name": tag
             })
+        # In case if field is not updated cancel the edition
+        if not updated:
+            cancel = True
         self._tags_action(view, cancel, reset)
 
     def edit_tags(self, tags):
--- a/cfme/common/__init__.py
+++ b/cfme/common/__init__.py
@@ -1,8 +1,10 @@
 # -*- coding: utf-8 -*-
 from functools import partial
+import re
 from urlparse import urlparse
 
 from cached_property import cached_property
+from cfme.base.login import BaseLoggedInPage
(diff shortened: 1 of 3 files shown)


dajoRH:
Lint report for commit 476689617dfea5996670eddc78e1105c833da2ef:

`cfme/common/__init__.py`:
- :bangbang: Line 163:1: **E302** *expected 2 blank lines, found 1*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

cfme/common/__init__.py
```diff
@@ -155,6 +161,129 @@ def get_tags_db(self):
...
+        """ Add tag to tested item
+            Args:
+                tag: list, tuple with tag category and tag name or Tag object
+                cancel: set True to cancel tag assigment
+                reset: set True to reset already set up tag
+        """
+        view = navigate_to(self, 'EditTags')
+        if isinstance(tag, (list, tuple, set)):
+            category = tag[0]
+            tag_name = tag[1]
```
quarckster: **REQUIRED**
TypeError: 'set' object does not support indexing

cfme/common/__init__.py
```diff
@@ -155,6 +161,129 @@ def get_tags_db(self):
...
+class WidgetasticTaggable(object):
+    """
+        This class can be inherited by any class that honors tagging.
+        Class should have 'EditTags' navigate step registered.
+        This class should be used for already converted to widgetastic classes
+
+        It provides functionality to assign and unassigned tags.
+    """
+
+    def edit_tag(self, tag, cancel=False, reset=False):
```
quarckster: **OPTIONAL**
I don't like `edit_tag()` has tag argument as a list. Is there any reason to do that? I would do something like that:
```python
def edit_tag(self, category, tag_name, cancel=False, reset=False)
```

rlbabyuk: agree, maybe the list is not good idea to use here

cfme/common/__init__.py
rlbabyuk: removed that part, as updated edit tag parameters not to be a list

cfme/common/__init__.py
```diff
@@ -155,6 +161,135 @@ def get_tags_db(self):
...
+                tag: tag(str) or Tag object
+                cancel: set True to cancel tag assigment
+                reset: set True to reset already set up tag
+        """
+        view = navigate_to(self, 'EditTags')
+        if isinstance(tag, Tag):
+            category = tag.category.display_name
+            tag = tag.display_name
+        try:
+            view.fill({
```
quarckster: **REQUIRED**
```python
cancel = not view.fill({
    "tag_category": '{} *'.format(category),
    "tag_name": tag
})

File: cfme/common/__init__.py [hunk 4]
```diff
@@ -155,6 +161,135 @@ def get_tags_db(self):
...
+                "tag_category": '{} *'.format(category),
+                "tag_name": tag
+            })
+        except NoSuchElementException:
+            view.fill({
```
quarckster:
```python
cancel = not view.fill()
```

cfme/common/__init__.py
```diff
@@ -155,6 +161,135 @@ def get_tags_db(self):
...
+
+    def get_tags(self, tenant="My Company Tags"):
+        """ Get list of tags assigned to item
+            Args:
+                tenant: string, tags tenant, default is "My Company Tags"
+            Returns:
+                List of tags in format "Tag_category: Tag_name"
+        """
+        view = navigate_to(self, 'Details')
+        tags = view.tag.read()
```
quarckster: Does it mean that any WidgetasticTaggable object has to have some details view with `tag` attribute?

rlbabyuk: yes, as this tag summary widget can be different ex. for group tag with use SummaryFormItem, and for vm/instances another one, so it better to add it to Details view

cfme/common/__init__.py
rlbabyuk: not fully covered the case with successful fields fill and cancel. 
I would like to do it like:
updated = view.fill()
if not updated:
    cancel = True
@quarckster what do you think, can it be like that?

cfme/common/__init__.py
```diff
@@ -155,6 +161,138 @@ def get_tags_db(self):
...
+class WidgetasticTaggable(object):
+    """
+        This class can be inherited by any class that honors tagging.
+        Class should have 'EditTags' navigate step registered.
+        This class should be used for already converted to widgetastic classes
+
+        It provides functionality to assign and unassigned tags.
+    """
+
+    def edit_tag(self, category=None, tag=None, cancel=False, reset=False):
```
quarckster: Maybe better to rename it to `add_tag`?

cfme/common/__init__.py
quarckster: It should be added to the class description.

cfme/common/__init__.py
rlbabyuk: yes, it better to use add_tag

cfme/common/__init__.py
```diff
@@ -155,6 +161,141 @@ def get_tags_db(self):
...
+        """
+        if reset:
+            view.reset_button.click()
+            view.flash.assert_message('All changes have been reset')
+        elif cancel:
+            view.cancel_button.click()
+            view.flash.assert_success_message('Tag Edit was cancelled by the user')
+        else:
+            view.save_button.click()
+            view.flash.assert_success_message('Tag edits were successfully saved')
```
jkrocil: I'd like to see one change here. Currently, if the user passes `cancel=True, reset=True`, they will remain on the edit tags page after clicking the reset button and the cancel button is not clicked anymore. Let's change this conditional so that `cancel` is still considered and it will be clicked, if set to `True`.

PR #4984, opened 2017-07-13
[1LP][RFR] removing the rest_api fixture

mkoura:
Removing the ``rest_api`` fixture as it has been replaced with ``appliance.rest_api`` in all places already.


psav:
This is amazing

PR #4988, opened 2017-07-14
[1LP][RFR] Update infra.pxe, remove property decorator

mshriver:
@property doesn't play with @variant

Thanks @mfalesni


mfalesni:
@mshriver dowe need to update the calls as well to use parentheses?

mshriver:
@mfalesni updated callers

PR #4991, opened 2017-07-17
[WIP] Provider Abstraction

vatsalparekh:
Purpose or Intent
=================
For the scenarios where we need to provision a VM, where we need work with multiple providers of different types(Cloud, Infra of containers maybe), we would have to check the type of the provider that is coming to the function pytest parameterized, and provision by its according class, like for Cloud provider, we would use cfme.cloud.instance.Instance, and similarly for infra; 
A better design, suggested by @psav would solve this where we define the instance/VM class inside the provider class itself, so we can easily use it

### DIFF
--- a/cfme/tests/infrastructure/test_infra_quota.py
+++ b/cfme/tests/infrastructure/test_infra_quota.py
@@ -7,7 +7,6 @@ from cfme import test_requirements
 from cfme.configure.access_control import Tenant
 from cfme.fixtures import pytest_selenium as sel
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
-from cfme.infrastructure.virtual_machines import Vm
 from cfme.provisioning import provisioning_form
 from cfme.services import requests
 from cfme.web_ui import fill, flash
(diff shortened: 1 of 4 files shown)


cfme/tests/infrastructure/test_infra_quota.py
```diff
@@ -114,7 +113,7 @@ def template_name(provisioning):
 @pytest.fixture(scope="function")
 def provisioner(request, setup_provider, provider):
     def _provisioner(template, provisioning_data, delayed=None):
-        vm = Vm(name=vm_name, provider=provider, template_name=template)
+        vm = provider.vm_type(name=vm_name, provider=provider, template_name=template)
```
mfalesni: Looks good, though I have an idea - Can you create a method in the provider that will use the vm_type and also injectthe provider itself?

So you could then just do:
```python
provider.instantiate_vm(name=vm_name, template_name=template)
```

vatsalparekh: +1 @mfalesni

vatsalparekh:
@lkhomenk is working on this in detail, #5113 is one the part

PR #4993, opened 2017-07-18
[RFR] Fix DisksTable locator

jkrocil:
`Table.__locator__()` used to return `str` but now it returns `Locator` obj, which doesn't work with `.format`. This fixes the issue because `Table.locator` is a `str`.

{{pytest: -k test_vm_reconfig_add_remove_disk_cold --use-provider vsphere6-nested --long-running}}


jkrocil:
This is good to go @mfalesni 2 tests already passed, we don't need to wait for the rest. As long as at least one passes, it means the locator works. :+1:

PR #4998, opened 2017-07-18
[1LP][RFR] HTML5 Remote Console Test - Initial Prototype

kedark3:
Purpose or Intent
=================
Adding HTML5 Remote Console tests.  It does following:
- For VMware: 
Configures VMware Console Support option on  Appliance Config to use VNC

- For VMware and RHOSP: 
Deploys a VM small_template and opens HTML5 Console to it, login using credentials, and touch a file. 

-For VMware: 
It can SSH to VM and try to remove the file which was created through HTML5 console, which if succeeds indicates that file was created and removes it, and if fails, indicates there was some issue with console and could not create file. 

- For VMware and RHOSP:
Tests Ctrl+Alt+Del Button functionality as well
Note:

-  Requires tesseract to be installed via dnf/yum.
- It might work well with RHEV, but not tested. 
- Tests do not run back to back, please run them against a single provider at a time using "--use-provider", back to back execution will be supported soon. 
- Need to run VMware tests against PhotonOS Template(currently yamls take 'damnSmallLinux' as value for small_template) and OpenStack Tests against Cirros.

**UPDATE**: The code should now support testing on multiple providers back to back.

**PRT Expected Result** 
- Against RHV the test is expected to fail
- Against OSP(rhos7-ga) it is expected to Pass - not 100% since sometimes console drops some characters while file creation and then SSH can not fine it to delete it. 
- Against VMware(vsphere6-nested) the test is expected to Pass - unless websocket hangs in "Connected" state for console with the "Blank screen"

**mshriver** Limiting execution for PRT to currently supported providers in CFME QE.
{{ pytest: --long-running --use-provider=vsphere6-nested --use-provider=rhos7-ga cfme/tests/cloud_infra_common/test_html5_vm_console.py }}

### DIFF
--- a/cfme/cloud/provider/openstack.py
+++ b/cfme/cloud/provider/openstack.py
@@ -3,6 +3,7 @@ from wrapanapi.openstack import OpenstackSystem
 from . import CloudProvider
 from cfme.common.provider import EventsEndpoint
 from cfme.infrastructure.provider.openstack_infra import RHOSEndpoint, OpenStackInfraEndpointForm
+from cfme.exceptions import ItemNotFound
 
 
 class OpenStackProvider(CloudProvider):
@@ -14,6 +15,10 @@ class OpenStackProvider(CloudProvider):
     mgmt_class = OpenstackSystem
     db_types = ["Openstack::CloudManager"]
     endpoints_form = OpenStackInfraEndpointForm
+    # xpath locators for elements, to be used by selenium
+    _console_connection_status_element = '//*[@id="noVNC_status"]'
+    _canvas_element = '//*[@id="noVNC_canvas"]'
+    _ctrl_alt_del_xpath = '//*[@id="sendCtrlAltDelButton"]'
 
     def __init__(self, name=None, endpoints=None, zone=None, key=None, hostname=None,
                  ip_address=None, api_port=None, sec_protocol=None, amqp_sec_protocol=None,
@@ -85,3 +90,27 @@ class OpenStackProvider(CloudProvider):
                    tenant_mapping=prov_config.get('tenant_mapping', False),
                    infra_provider=infra_provider,
                    appliance=appliance)
+
+    # Following methods will only work if the remote console window is open
+    # and if selenium focused on it. These will not work if the selenium is
+    # focused on Appliance window.
+    def get_console_connection_status(self):
+        try:
+            return self.appliance.browser.widgetastic.selenium.find_element_by_xpath(
+                self._console_connection_status_element).text
+        except:
+            raise ItemNotFound("Element not found on screen, is current focus on console window?")
+
+    def get_remote_console_canvas(self):
+        try:
+            return self.appliance.browser.widgetastic.selenium.find_element_by_xpath(
+                self._canvas_element)
+        except:
+            raise ItemNotFound("Element not found on screen, is current focus on console window?")
+
+    def get_console_ctrl_alt_del_btn(self):
+        try:
+            return self.appliance.browser.widgetastic.selenium.find_element_by_xpath(
+                self._ctrl_alt_del_xpath)
+        except:
+            raise ItemNotFound("Element not found on screen, is current focus on console window?")
--- a/cfme/common/provider.py
+++ b/cfme/common/provider.py
@@ -718,6 +718,16 @@ class BaseProvider(Taggable, Updateable, SummaryMixin, Navigatable):
         else:
             raise ValueError("Endpoints should be either dict or endpoint class")
 
+    # These methods need to be overridden in the provider specific classes
+    def get_console_connection_status(self):
+        raise NotImplementedError("This method is not implemented for given provider")
+
+    def get_remote_console_canvas(self):
+        raise NotImplementedError("This method is not implemented for given provider")
+
+    def get_console_ctrl_alt_del_btn(self):
+        raise NotImplementedError("This method is not implemented for given provider")
+
 
 def get_paginator_value():
     return paginator.rec_total()
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -7,6 +7,7 @@ from wrapanapi import exceptions
 
 from cfme import js
 from cfme.base.login import BaseLoggedInPage
+from cfme.common.vm_console import VMConsole
 from cfme.exceptions import (
     VmOrInstanceNotFound, TemplateNotFound, OptionNotAvailable, UnknownProviderType)
 from cfme.fixtures import pytest_selenium as sel
@@ -31,6 +32,7 @@ from widgetastic_patternfly import Button, BootstrapSelect
 
 from . import PolicyProfileAssignable, Taggable, SummaryMixin
 
+access_btn = partial(toolbar.select, "Access")
 cfg_btn = partial(toolbar.select, "Configuration")
 lcl_btn = partial(toolbar.select, "Lifecycle")
 mon_btn = partial(toolbar.select, 'Monitoring')
@@ -237,6 +239,25 @@ class BaseVM(Pretty, Updateable, PolicyProfileAssignable, Taggable, SummaryMixin
         else:
             raise ValueError("{} is not a known state for compliance".format(text))
 
+    @property
+    def console_handle(self):
+        '''
+        The basic algorithm for getting the consoles window handle is to get the
+        appliances window handle and then iterate through the window_handles till we find
+        one that is not the appliances window handle.   Once we find this check that it has
+        a canvas widget with a specific ID
+        '''
+        browser = self.appliance.browser.widgetastic
+        appliance_handle = browser.window_handle
+        cur_handles = browser.selenium.window_handles
+        logger.info("Current Window Handles:  {}".format(cur_handles))
+
+        for handle in cur_handles:
+            if handle != appliance_handle:
+                # FIXME: Add code to verify the tab has the correct widget
+                #      for a console tab.
+                return handle
+
     def delete(self, cancel=False, from_details=False):
         """Deletes the VM/Instance from the VMDB.
 
@@ -346,6 +367,46 @@ class BaseVM(Pretty, Updateable, PolicyProfileAssignable, Taggable, SummaryMixin
             return InfoBlock.icon_href(*properties)
         else:
             return InfoBlock.text(*properties)
+
+    def open_console(self, console='VM Console', invokes_alert=False, cancel=False):
+        """
+        Initiates the opening of one of the console types supported by the Access
+        button.   Presently we only support VM Console, which is the HTML5 Console.
+        In case of VMware provider it could be VMRC, VNC/HTML5, WebMKS, but we only
+        support VNC/HTML5.
+        Possible values for 'console' could be 'VM Console' and 'Web Console', but Web
+        Console is not supported as well.
+
+        Args:
+            console       - one of the supported console types given by the Access button.
+            invokes_alert - If the particular console will invoke a CFME popup/alert
+                            setting this to true will handle this.
+            cancel        - Allows one to cancel the operation if the popup/alert occurs.
+        """
+        # TODO: implement vmrc vm console
+        if console not in ['VM Console']:
+            raise NotImplementedError('Not supported console type: {}'.format(console))
+
+        view = navigate_to(self, 'Details')
+
+        # Click console button given by type
+        view.toolbar.access.item_select(console)
+
+        # Get the consoles window handle, and then create a VMConsole object, and store
+        # the VMConsole object aside.
+        console_handle = self.console_handle
+        if console_handle is None:
+            raise TypeError("Console handle should not be None")
+        appliance_handle = self.appliance.browser.widgetastic.window_handle
+        logger.info("Creating VMConsole:")
+        logger.info("   appliance_handle: {}".format(appliance_handle))
+        logger.info("     console_handle: {}".format(console_handle))
+        logger.info("               name: {}".format(self.name))
+        self.vm_console = VMConsole(
+            appliance_handle=appliance_handle,
+            console_handle=console_handle,
+            vm=self
+        )
 
     def open_details(self, properties=None):
         """Clicks on details infoblock"""
--- a/cfme/common/vm_console.py
+++ b/cfme/common/vm_console.py
@@ -39,13 +39,13 @@ class VMConsole(Pretty):
         # due to an exception being thrown.
         try:
             text = self.provider.get_console_connection_status()
-            logger.info('Read following text from console banner: {}'.format(text))
-        except NoSuchElementException as e:
-            logger.info('Could not find banner element.')
-            logger.info('Exception: {}'.format(e))
+        except NoSuchElementException:
+            logger.exception('Could not find banner element.')
             return None
+        finally:
+            self.switch_to_appliance()
 
-        self.switch_to_appliance()
+        logger.info('Read following text from console banner: %s', text)
         return text
 
     def get_screen(self):
--- a/cfme/common/vm_console.py
+++ b/cfme/common/vm_console.py
@@ -1,0 +1,203 @@
+# -*- coding: utf-8 -*-
+"""
+Module containing classes with common behaviour for consoles of both VMs and Instances of all types.
+"""
+
+import base64
+import re
+import tempfile
+
+from PIL import Image, ImageFilter
+from pytesseract import image_to_string
+from utils.log import logger
+from utils.pretty import Pretty
+from wait_for import wait_for, TimedOutError
+from selenium.common.exceptions import NoSuchElementException
+from selenium.webdriver.common.keys import Keys
+
+
+class VMConsole(Pretty):
+    """Class to manage the VM Console.   Presently, only support HTML5 Console."""
+
+    pretty_attrs = ['appliance_handle', 'browser', 'console_handle', 'name']
+
+    def __init__(self, vm, console_handle, appliance_handle):
+        self.name = vm.name
+        self.selenium = vm.appliance.browser.widgetastic.selenium
+        self.console_handle = console_handle
+        self.appliance_handle = appliance_handle
+        self.provider = vm.provider
+
+    ###
+    # Methods
+    #
+    def get_banner(self):
+        """Get the text of the banner above the console screen."""
+        self.switch_to_console()
+        # We know the widget may or may not be available right away
+        # so we do this in a try-catch to ensure the code is not stopped
+        # due to an exception being thrown.
+        try:
+            text = self.provider.get_console_connection_status()
+            logger.info('Read following text from console banner: {}'.format(text))
+        except NoSuchElementException as e:
+            logger.info('Could not find banner element.')
+            logger.info('Exception: {}'.format(e))
+            return None
+
+        self.switch_to_appliance()
+        return text
+
+    def get_screen(self):
+        """
+        Retrieve the bit map from the canvas widget that represents the console screen.
+
+        Returns it as a binary string.
+
+        Implementation:
+        The canvas tag has a method toDataURL() which one can use in javascript to
+        obtain the canvas image  base64 encoded.   Examples of how to do this can be
+        seen here:
+
+            https://qxf2.com/blog/selenium-html5-canvas-verify-what-was-drawn/
+            https://stackoverflow.com/questions/38316402/how-to-save-a-canvas-as-png-in-selenium
+        """
+        self.switch_to_console()
+
+        # Get the canvas element
+        canvas = self.provider.get_remote_console_canvas()
+
+        # Now run some java script to get the contents of the canvas element
+        # base 64 encoded.
+        image_base64_url = self.selenium.execute_script(
+            "return arguments[0].toDataURL('image/jpeg',1);",
+            canvas
+        )
+
+        # The results will look like:
+        #
+        #   data:image/jpeg;base64,iVBORw0KGgoAAAANSUhEUgAAAfQAAABkCAYAAABwx8J9AA...
+        #
+        # So parse out the data from the non image data from the URL:
+        image_base64 = image_base64_url.split(",")[1]
+
+        # Now convert to binary:
+        image_jpeg = base64.b64decode(image_base64)
+
+        self.switch_to_appliance()
+        return image_jpeg
+
+    def get_screen_text(self):
+        """
+        Return the text from a text console.
+
+        Uses OCR to scrape the text from the console image taken at the time of the call.
+        """
+        image_str = self.get_screen()
+
+        # Write the image string to a file as pytesseract requires
+        # a file, and doesn't take a string.
+        tmp_file = tempfile.NamedTemporaryFile(suffix='.jpeg')
+        tmp_file.write(image_str)
+        tmp_file.flush()
+        tmp_file_name = tmp_file.name
+        # Open Image file, resize it to high resolution, sharpen it for clearer text
+        # and then run image_to_string operation which returns unicode that needs to
+        # be converted to utf-8 which gives us text [typr(text) == 'str']
+        # higher resolution allows tesseract to recognize text correctly
+        text = (image_to_string(((Image.open(tmp_file_name)).resize((7680, 4320),
+         Image.ANTIALIAS)).filter(ImageFilter.SHARPEN), lang='eng',
+         config='--user-words eng.user-words')).encode('utf-8')
+        tmp_file.close()
+
+        logger.info('screen text:{}'.format(text))
+        return text
+
+    def is_connected(self):
+        """Wait for the banner on the console to say the console is connected."""
+        banner = self.get_banner()
+        if banner is None:
+            return False
+        return re.match('Connected', banner) is not None
+
+    def send_keys(self, text):
+        """Send text to the console."""
+        self.switch_to_console()
+        canvas = self.provider.get_remote_console_canvas()
+        canvas.send_keys(text)
+        canvas.send_keys(Keys.ENTER)
+        logger.info("Sending following Keys to Console {}".format(text))
+        self.switch_to_appliance()
+
+    def send_ctrl_alt_delete(self):
+        """Press the ctrl-alt-delete button in the console tab."""
+        self.switch_to_console()
+        ctrl_alt_del_btn = self.provider.get_console_ctrl_alt_del_btn()
+        logger.info("Sending following Keys to Console CTRL+ALT+DEL")
+        ctrl_alt_del_btn.click()
+        self.switch_to_appliance()
+
+    def switch_to_appliance(self):
+        """Switch focus to appliance tab/window."""
+        logger.info("Switching to appliance: window handle = {}".format(self.appliance_handle))
+        self.selenium.switch_to_window(self.appliance_handle)
+
+    def switch_to_console(self):
+        """Switch focus to console tab/window."""
+        logger.info("Switching to console: window handle = {}".format(self.console_handle))
+        self.selenium.switch_to_window(self.console_handle)
+
+    def wait_for_connect(self, timeout=15):
+        """Wait for as long as the specified/default timeout for the console to be connected."""
+        try:
+            logger.info('Waiting for console connection (timeout={})'.format(timeout))
+            wait_for(func=lambda: self.is_connected(),
+                     delay=1, handle_exceptions=True,
+                     num_sec=timeout)
+            return True
+        except TimedOutError:
+            return False
+
+    def close_console_window(self):
+        """Attempt to close Console window at the end of test."""
+        if self.console_handle is not None:
+            self.switch_to_console()
+            self.selenium.close()
+            logger.info("Browser window/tab containing Console was closed.")
+            self.switch_to_appliance()
+
+    def find_text_on_screen(self, text_to_find, current_line=False):
+        """Find particular text is present on Screen.
+
+        This function uses get_screen_text function to get string containing
+        the text on the screen and then tries to match it against the 'text_to_find'.
+
+        Args:
+            text_to_find   - This is what re.search will try to search for on screen.
+        Returns:
+            If the match is found returns True else False.
+        """
+        if current_line:
+            return re.search(text_to_find, self.get_screen_text().split('\n')[-1]) is not None
+        return re.search(text_to_find, self.get_screen_text()) is not None
+
+    def wait_for_text(self, timeout=45, text_to_find="", to_disappear=False):
+        """Wait for as long as the specified/default timeout for the 'text' to show up on screen.
+
+        Args:
+            timeout         - Wait Time before wait_for function times out.
+            text_to_find    - value passed to find_text_on_screen function
+            to_disappear    - if set to True, function will wait for text_to_find to disappear
+                          from screen.
+        """
+        if not text_to_find:
+            return None
+        try:
+            if to_disappear:
+                logger.info("Waiting for {} to disappear from screen".format(text_to_find))
+            result = wait_for(func=lambda: to_disappear != self.find_text_on_screen(text_to_find),
+                     delay=5,
+                     num_sec=timeout)
+            return result.out
+        except TimedOutError:
+            return None
--- a/cfme/infrastructure/provider/rhevm.py
+++ b/cfme/infrastructure/provider/rhevm.py
@@ -6,7 +6,7 @@ from utils import version
 from . import InfraProvider
 from cfme.common.provider import CANDUEndpoint, DefaultEndpoint, DefaultEndpointForm
 from wrapanapi.rhevm import RHEVMSystem
-
+from cfme.exceptions import ItemNotFound
 
 class RHEVMEndpoint(DefaultEndpoint):
     @property
@@ -48,6 +48,10 @@ class RHEVMProvider(InfraProvider):
     db_types = ["Redhat::InfraManager"]
     endpoints_form = RHEVMEndpointForm
     discover_dict = {"rhevm": True}
+    # xpath locators for elements, to be used by selenium
+    _console_connection_status_element = '//*[@id="connection-status"]'
+    _canvas_element = '//*[@id="remote-console"]/canvas'
+    _ctrl_alt_del_xpath = '//*[@id="ctrlaltdel"]'
 
     def __init__(self, name=None, endpoints=None, zone=None, key=None, hostname=None,
                  ip_address=None, start_ip=None, end_ip=None, provider_data=None, appliance=None):
@@ -97,3 +101,27 @@ class RHEVMProvider(InfraProvider):
                    start_ip=start_ip,
                    end_ip=end_ip,
                    appliance=appliance)
+
+    # Following methods will only work if the remote console window is open
+    # and if selenium focused on it. These will not work if the selenium is
+    # focused on Appliance window.
+    def get_console_connection_status(self):
+        try:
+            return self.appliance.browser.widgetastic.selenium.find_element_by_xpath(
+                self._console_connection_status_element).text
+        except:
+            raise ItemNotFound("Element not found on screen, is current focus on console window?")
+
+    def get_remote_console_canvas(self):
+        try:
+            return self.appliance.browser.widgetastic.selenium.find_element_by_xpath(
+                self._canvas_element)
+        except:
+            raise ItemNotFound("Element not found on screen, is current focus on console window?")
+
+    def get_console_ctrl_alt_del_btn(self):
+        try:
+            return self.appliance.browser.widgetastic.selenium.find_element_by_xpath(
+                self._ctrl_alt_del_xpath)
+        except:
+            raise ItemNotFound("Element not found on screen, is current focus on console window?")
--- a/cfme/infrastructure/provider/virtualcenter.py
+++ b/cfme/infrastructure/provider/virtualcenter.py
@@ -6,6 +6,7 @@ from cfme.common.provider_views import ProviderNodesView
 from cfme.exceptions import DestinationNotFound
 from utils.appliance.implementations.ui import CFMENavigateStep, navigator
 from . import InfraProvider
+from cfme.exceptions import ItemNotFound
 
 
 class VirtualCenterEndpoint(DefaultEndpoint):
@@ -22,6 +23,10 @@ class VMwareProvider(InfraProvider):
     db_types = ["Vmware::InfraManager"]
     endpoints_form = VirtualCenterEndpointForm
     discover_dict = {"vmware": True}
+    # xpath locators for elements, to be used by selenium
+    _console_connection_status_element = '//*[@id="connection-status"]'
+    _canvas_element = '//*[@id="remote-console"]/canvas'
+    _ctrl_alt_del_xpath = '//*[@id="ctrlaltdel"]'
 
     def __init__(self, name=None, endpoints=None, key=None, zone=None, hostname=None,
                  ip_address=None, start_ip=None, end_ip=None, provider_data=None, appliance=None):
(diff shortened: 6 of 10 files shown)


cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -0,0 +1,168 @@
...
+
+        - The console's status will be checked.
+        - A command that creates a file will be sent through the console.
+        - Using ssh we will check that the command worked (i.e. that the file
+          was created.
+    """
+    # temporarily getting credentials from vspherehost-nested because they
+    # match with the PhotonOS Credentials, but need to add credentials for
+    # templates in the credentials.eyaml
+    photon_os_username = credentials['vspherehost-nested'].get('username')
```
kedark3: These should be replaced with template credentials, but since we do not have template credentials in our yamls, I have used these credentials of "vspherehost-nested" here, but I will put up a separate GitLab PR to add Template credentials in our yamls and then those will be referenced here.

cfme/common/vm.py
```diff
@@ -342,6 +363,53 @@ def get_detail(self, properties=None, icon_href=False):
...
+        navigate_to(self, 'Details')
+
+        # Click console button given by type
+        access_btn(console, invokes_alert=invokes_alert)
+
+        # Some access types in the future may require a CFME popup be
+        # handled.  The one that is supported now does not require this,
+        # and hence the default is not do this.
+        if invokes_alert:
+            sel.handle_alert(cancel)
```
psav: **REQUIRED** This needs to be converted to WT instead

cfme/common/vm.py
```diff
@@ -342,6 +363,53 @@ def get_detail(self, properties=None, icon_href=False):
...
+
+        # Get the consoles window handle, and then create a VMConsole object, and store
+        # the VMConsole object aside.
+        console_handle = self.console_handle
+        appliance_handle = self.appliance.browser.widgetastic.window_handle
+        logger.info("Creating VMConsole:")
+        logger.info("   appliance_handle: {}".format(appliance_handle))
+        logger.info("     console_handle: {}".format(console_handle))
+        logger.info("               name: {}".format(self.name))
+        vm_console = VMConsole(
```
psav: **QUESTION:** Why not just ```self.vm_console = vm_console``` also, as it can be so easily disrupted, I wonder if it's a good idea to store it on the vm object....

cfme/common/vm_console.py
```diff
@@ -0,0 +1,179 @@
...
+canvas_xapth = '//*[@id="remote-console"]/canvas'  # specific to VMware/RHEV Provider
+ctrl_alt_del_xpath = '//*[@id="ctrlaltdel"]'  # specific to VMware/RHEV Provider
+
+
+class VMConsole(Pretty):
+    """Class to manage the VM Console.   Presently, only support HTML5 Console."""
+
+    pretty_attrs = ['appliance_handle', 'browser', 'console_handle', 'name']
+
+    def __init__(self, name, selenium, console_handle, appliance_handle, provider):
```
psav: **REQUIRED:** As the VMconsole object requires the provider, it can obtain selenium from ```provider.appliance```. In fact why do we pass in the provider? Wouldn't it make more sense to pass in the VM object, which would have access to the provider anyway? As the console is "perVM"?

cfme/common/vm_console.py
```diff
@@ -0,0 +1,179 @@
...
+        self.name = name
+        self.selenium = selenium
+        self.console_handle = console_handle
+        self.appliance_handle = appliance_handle
+        self.provider = provider
+
+    ###
+    # Methods
+    #
+    def _get_openstack_banner(self):
```
psav: **REQUIRED**:These methods here I would propose to move into the provider objects themselves. It can be called the same name then on ALL providers, needs no conditional, and is invoke by simply doing, ```self.provider.get_banner()``` If you chose to do it like that. You can also put a Base Provider object to raise NotImplemented exception. In this way, adding a new provider with a new way of getting the banner, requires no more changes to core files.

cfme/common/vm_console.py
```diff
@@ -0,0 +1,179 @@
...
+        self.switch_to_console()
+        if self.provider.one_of(OpenStackProvider):
+            text = self._get_openstack_banner()
+        else:
+            text = self._get_vmware_rhev_connection_status_text()
+        logger.info('Read following text from console banner: {}'.format(text))
+        self.switch_to_appliance()
+        return text
+
+    def _get_openstack_canvas(self):
```
psav: **REQUIRED:** same as argument above

cfme/common/vm_console.py
```diff
@@ -0,0 +1,179 @@
...
+        tmp_file.flush()
+        tmp_file_name = tmp_file.name
+        # Open Image file, resize it to high resolution, sharpen it for clearer text
+        # and then run image_to_string operation which returns unicode that needs to
+        # be converted to utf-8 which gives us text [typr(text) == 'str']
+        text = (image_to_string(((Image.open(tmp_file_name)).resize((7680, 4320),
+         Image.ANTIALIAS)).filter(ImageFilter.SHARPEN), lang='eng')).encode('utf-8')
+        tmp_file.close()
+
+        logger.info('screen text:{}'.format(text))
```
psav: **QUESTION:** How reliable is this?

cfme/common/vm_console.py
```diff
@@ -0,0 +1,179 @@
...
+        canvas = self.get_canvas()
+        canvas.send_keys(text)
+        logger.info("Sending following Keys to Console {}".format(text))
+        self.switch_to_appliance()
+
+    def send_ctrl_alt_delete(self):
+        """Press the ctrl-alt-delete button in the console tab."""
+        self.switch_to_console()
+
+        if self.provider.one_of(OpenStackProvider):
```
psav: As above

cfme/common/vm_console.py
```diff
@@ -0,0 +1,179 @@
...
+        obtain the canvas image  base64 encoded.   Examples of how to do this can be
+        seen here:
+
+            https://qxf2.com/blog/selenium-html5-canvas-verify-what-was-drawn/
+            https://stackoverflow.com/questions/38316402/how-to-save-a-canvas-as-png-in-selenium
+        """
+        self.switch_to_console()
+
+        # Get the canvas element
+        canvas = self.get_canvas()
```
psav: **QUESTION:** Do we at some point "touch" the canvas, if a screensaver is active and we wanted to scrape the screen, what happens?

cfme/common/vm_console.py
```diff
@@ -0,0 +1,179 @@
...
+        """
+        Return the text from a text console.
+
+        Uses OCR to scrape the text from the console image taken at the time of the call.
+        """
+        image_str = self.get_screen()
+
+        # Write the image string to a file as pytesseract requires
+        # a file, and doesn't take a string.
+        tmp_file = tempfile.NamedTemporaryFile(suffix='.jpeg')
```
psav: **REQUIRED:** We should rather do this with an in-memory object, rather than writing out to a file.

https://docs.python.org/release/2.7/library/io.html#binary-i-o

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -0,0 +1,168 @@
...
+        try:
+            vm_obj.delete_from_provider()
+        except Exception:
+            logger.warning("Failed to delete vm `{}`.".format(vm_obj.name))
+
+    vm_obj.create_on_provider(timeout=2400, find_in_cfme=True, allow_skip="default")
+    return vm_obj
+
+
+def _configure_vmware_console_for_test(appliance):
```
psav: **REQUIRED:** I think we should have a method on the Appliance, in the configuration page called something like ```set_console_mode``` and probably with its counterpart, ```get_console_mode```, It then allows other people to use the function you have written here, currently it's hidden.

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -0,0 +1,168 @@
...
+def _configure_server_roles_for_test():
+    """
+    Enable websocket role if it is disabled.
+
+    Currently the fixture cfme/fixtures/base.py,
+    disables the websocket role to avoid intrusive popups.
+    """
+    # TODO:Find correct place to put this method at.
+    roles = configuration.get_server_roles()
+    if 'websocket' in roles and not roles['websocket']:
```
psav: **REQUIRED:** Right, so I think at the moment having this as a fixture is the right way to go, except we should disable it again after the test. So, perhaps make it a yield fixture and do the cleanup afterwards.

cfme/tests/cloud_infra_common/test_html5_vm_console.py
psav: Yes please add those, it won't be merged as is.

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -0,0 +1,168 @@
...
+        - Using ssh we will check that the command worked (i.e. that the file
+          was created.
+    """
+    # temporarily getting credentials from vspherehost-nested because they
+    # match with the PhotonOS Credentials, but need to add credentials for
+    # templates in the credentials.eyaml
+    photon_os_username = credentials['vspherehost-nested'].get('username')
+    photon_os_password = credentials['vspherehost-nested'].get('password')
+    _configure_server_roles_for_test()  # make sure websocket is enabled
+    if provider.one_of(VMwareProvider):
```
psav: **REQUIRED:** I'd prefer the conditional moved into the configure_mode fixture above

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -0,0 +1,168 @@
...
+
+    # Try to get the login screen text. We should have the text
+    # login there:
+    console_text = vm_console.get_screen_text()
+    assert re.search("login:", console_text)
+
+    # Enter Username:
+    # XXX: This is hard coded, we need to add some new yaml to track the
+    #      template credentials.
+    if provider.one_of(OpenStackProvider):
```
psav: Yes new yaml needs adding before we merge this, add a new substanza on the provider

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -0,0 +1,168 @@
...
+
+    # Check if console prompted for password
+    wait_for(func=lambda: re.search("Password:", vm_console.get_screen_text()),
+            delay=1,
+            num_sec=20)
+    console_text = vm_console.get_screen_text()
+    # Enter Password if prompted for password
+    assert re.search("Password:", console_text)
+
+    if provider.one_of(OpenStackProvider):
```
psav: And here

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -0,0 +1,168 @@
...
+        vm_console.send_keys("cubswin:)\n")  # For Cirros Template
+    else:
+        vm_console.send_keys("{}\n".format(photon_os_password))  # For PhotonOS Template
+
+    logger.info("Wait 15 secs to get the '$' prompt")
```
psav: **REQUIRED:** Rather than an arbitrary sleep time, let's make use of the ```wait_for``` for all the places you sleep. We can read for that $ prompt at the end? So can't we continually scrape until it arrives?

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -0,0 +1,168 @@
...
+    time.sleep(5)  # Wait for file creation
+
+    # Test pressing ctrl-alt-delete...we should be able to get a new login prompt:
+    vm_console.send_ctrl_alt_delete()
+    logger.info("Wait 60 sec to get 'login:' prompt since ctrl+alt+del resets VM/Instance")
+    time.sleep(60)   # Wait for "login" prompt
+    console_text = vm_console.get_screen_text()
+    assert re.search("login:", console_text)
+    # Code below only works for VMware and not tested against other providers
+    connect_ip, tc = wait_for(provider.mgmt.get_ip_address, [vm_obj.name], num_sec=300,
```
psav: **REQUIRED:** If it only works for vmware, why don't we have a conditional?

cfme/common/vm_console.py
kedark3: These tests are entirely dependent on the fix set of Templates, PhotonOS for VMware and Cirros for OpenStack. So with those we won't face screensaver situation. For GUI based VM Consoles, I will have to find some other way to deal with it. My primary goal was being able to interact with some VM console, so I chose to go with the command line based first. It was collective decision of myself and @jamesooden . Hope it makes sense.

cfme/common/vm_console.py
kedark3: This does not give 100% accurate results but it is well tested to correctly recognize words such as login, password prompts, which was a basic necessity in this case.

cfme/common/vm.py
kedark3: Hello Pete, I apologize but I am not very much familiar with the WT (widgetastic) kindly provide me some more info on what is expected here.

cfme/common/vm.py
kedark3: do you mean: 
`self.vm_console = VMConsole(...)`?
Also, I am not sure what does it mean by easily disrupted. Kindly elaborate.

cfme/tests/cloud_infra_common/test_html5_vm_console.py
kedark3: in utils/appliance/__init__.py, class Appliance or class IPAppliance?

cfme/tests/cloud_infra_common/test_html5_vm_console.py
kedark3: I thought adding continuous scraping would be too much of work overhead, hence decided to go with sleep. Initially I was using wait_for only. For Login and Password prompts I can use wait_for(should I?), but for $ prompt, it might not be entirely possible since tesseract is screwing up $ while it converts image to string, as "3" so basically for $ I get 3 as character recognition.

cfme/tests/cloud_infra_common/test_html5_vm_console.py
kedark3: It only works for VMware because for OpenStack, I get ip address as "None", so I am trying to figure out why. So far, I have understood, correct me if I am wrong, but I will need a floating IP to be associated to OpenStack instance then it will not return None for get_ip_address. Also, this code might work as is with the RHEV, not tested, hence not put conditional.

cfme/tests/cloud_infra_common/test_html5_vm_console.py
kedark3: Please Check MR on GitLab for the same.

cfme/common/vm.py
```diff
@@ -342,6 +363,53 @@ def get_detail(self, properties=None, icon_href=False):
...
+        In case of VMware provider it could be VMRC, VNC/HTML5, WebMKS, but we only
+        support VNC/HTML5.
+        Possible values for 'console' could be 'VM Console' and 'Web Console', but Web
+        Console is not supported as well.
+
+        Args:
+            console       - one of the supported console types given by the Access button.
+            invokes_alert - If the particular console will invoke a CFME popup/alert
+                            setting this to true will handle this.
+            cancel        - Allows one to cancel the operation if the pope/alert occurs.
```
jamesooden: pope should be popup.

kedark3: Pope will become popup :100:

cfme/common/vm_console.py
```diff
@@ -0,0 +1,217 @@
...
+        tmp_file.write(image_str)
+        tmp_file.flush()
+        tmp_file_name = tmp_file.name
+        # Open Image file, resize it to high resolution, sharpen it for clearer text
+        # and then run image_to_string operation which returns unicode that needs to
+        # be converted to utf-8 which gives us text [typr(text) == 'str']
```
jamesooden: Let's add a comment explaining why we had to change the resolution of the image.   Also, could we make this resolution a configurable thing?

cfme/common/vm_console.py
jamesooden: Same response as above.

dajoRH:
Lint report for commit 706fc2c0ce271d14d2d29657ffe2c53590579261:

`cfme/common/vm_console.py`:
- :red_circle: Line 216:16: **F821** *undefined name 'negation'*

**No commit flaws detected**.

*CFME QE Bot*

cfme/common/vm_console.py
jamesooden: Discussed this with @psav and @kedark3.   We do want to do what @psav suggested, however we want to keep the code that handles the window switching in vm_console.py.   This means we will end up with the same subroutines (save for the provider specific ones) but after doing the console switch, they will call a new function off of the provider object to do the provider specific stuff.

cfme/common/vm_console.py
jamesooden: See new response above.

cfme/cloud/provider/openstack.py
```diff
@@ -14,6 +14,9 @@ class OpenStackProvider(CloudProvider):
     mgmt_class = OpenstackSystem
     db_types = ["Openstack::CloudManager"]
     endpoints_form = OpenStackInfraEndpointForm
+    _console_connection_status_element = '//*[@id="noVNC_status"]'
```
jamesooden: Maybe add a comment saying these are selenium locators

cfme/common/provider.py
```diff
@@ -707,6 +707,15 @@ def _prepare_endpoints(endpoints):
         else:
             raise ValueError("Endpoints should be either dict or endpoint class")
 
+    def get_console_connection_status(self):
```
jamesooden: Add a comment explaining that these will only work if the console window is the active window.

cfme/infrastructure/provider/rhevm.py
```diff
@@ -48,6 +48,9 @@ class RHEVMProvider(InfraProvider):
     db_types = ["Redhat::InfraManager"]
     endpoints_form = RHEVMEndpointForm
     discover_dict = {"rhevm": True}
+    _connection_status_xpath = '//*[@id="connection-status"]'
```
jamesooden: Same comment as in openstack provider

cfme/infrastructure/provider/rhevm.py
```diff
@@ -97,3 +100,15 @@ def from_config(cls, prov_config, prov_key, appliance=None):
                    start_ip=start_ip,
                    end_ip=end_ip,
                    appliance=appliance)
+
+    def get_console_connection_status(self):
```
jamesooden: Same comment as in openstack provider

cfme/infrastructure/provider/virtualcenter.py
```diff
@@ -17,6 +17,9 @@ class VMwareProvider(InfraProvider):
     db_types = ["Vmware::InfraManager"]
     endpoints_form = VirtualCenterEndpointForm
     discover_dict = {"vmware": True}
+    _console_connection_status_element = '//*[@id="connection-status"]'
```
jamesooden: Same comment as in openstack provider

cfme/infrastructure/provider/virtualcenter.py
```diff
@@ -62,3 +65,15 @@ def view_value_mapping(self):
         return {'name': self.name,
                 'prov_type': 'VMware vCenter'
                 }
+
+    def get_console_connection_status(self):
```
jamesooden: Same comment as in openstack provider

cfme/cloud/provider/openstack.py
kedark3: Done

cfme/common/provider.py
kedark3: Done

cfme/infrastructure/provider/rhevm.py
kedark3: Done

cfme/infrastructure/provider/rhevm.py
kedark3: Done

cfme/infrastructure/provider/virtualcenter.py
kedark3: Done

cfme/infrastructure/provider/virtualcenter.py
kedark3: Done

cfme/common/vm_console.py
```diff
@@ -0,0 +1,202 @@
...
+            to_disappear    - if set to True, function will wait for text_to_find to disappear
+                          from screen.
+        """
+        if text_to_find == "":
+            return None
+        try:
+            if to_disappear:
+                logger.info("Waiting for {} to disappear from screen".format(text_to_find))
+            result = wait_for(func=lambda: to_disappear != self.find_text_on_screen(text_to_find),
+                     delay=1,
```
mkoura: Given the amount of work necessary (OCR etc.), consider increasing the delay.

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -0,0 +1,188 @@
...
+            # last line of the console will contain one of the following words:
+            # [login, password, incorrect]
+            # if so, regex_for_login_password will find it and result will not be []
+            # .split('\n')[-1] splits the console text on '\n' & picks last item of resulting list
+            result = regex_for_login_password.findall(vm_console.get_screen_text().split('\n')[-1])
+            return result == []
+
+        # if _try_login() returns True, it means we did not find any of words
+        # [login, password, incorrect] on last line of console text, which implies login success
+        wait_for(func=lambda: _try_login(), timeout=300)
```
mkoura: Given the amount of work necessary (OCR etc.), consider increasing the delay.

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -0,0 +1,188 @@
...
+
+        logger.info("Wait to get the '$' prompt")
+        if provider.one_of(VMwareProvider):
+            vm_console.wait_for_text(text_to_find="root@photon", timeout=200)
+        else:
+            time.sleep(15)
```
mkoura: Is there a better way to detect successful login & prompt? Like printing e.g. LOGGED IN after successful login (running ``echo 'echo LOGGED IN' >> .bashrc`` before logging in and ``sed -i '$ d' .bashrc`` during cleanup)?

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -0,0 +1,188 @@
...
+        vm_console.wait_for_text(text_to_find="login:", timeout=200, to_disappear=True)
+        assert vm_console.wait_for_text(text_to_find="login:", timeout=200) is not None,\
+            "VM Console didn't prompt for Login"
+
+        with ssh.SSHClient(hostname=vm_obj.ip_address, username=console_vm_username,
+                password=console_vm_password) as ssh_client:
+            # if file was created in previous steps it will be removed here
+            # we will get instance of SSHResult
+            command_result = ssh_client.run_command("rm bla*", ensure_user=True)
+            assert command_result
```
kedark3: 1. I felt like keeping ssh operations at the end. But yes,  it can be moved further up. 
2. Full name of the file was not used because OpenStack console sometimes drops some characters and creates file with partial name, @jamesooden is investigating that and once resolved, we will be updating that with full name. 
3. I think we do not need to add finalizer, since we destroy this VM irrespective of what SSH result ended up looking like. And for new test we will start with a fresh deployment of VM which ensures there will be no files present on it.

dajoRH:
*CFME QE Bot*

cfme/common/vm_console.py
kedark3: I will update the comment. can you please explain your take on making the resolution configurable?

dajoRH:
Lint report for commit 948fc486feadbc29740f59edcfbcd230325c58d3:

`cfme/infrastructure/provider/virtualcenter.py`:
- :bangbang: Line 77:5: **E301** *expected 1 blank line, found 0*

**No commit flaws detected**.

*CFME QE Bot*

cfme/common/vm_console.py
kedark3: I added the comment below on why resolution change is required. If it looks good, kindly resolve comment.

cfme/common/vm.py
```diff
@@ -237,6 +239,25 @@ def compliant(self):
...
+        a canvas widget with a specific ID
+        '''
+        browser = self.appliance.browser.widgetastic
+        appliance_handle = browser.window_handle
+        cur_handles = browser.selenium.window_handles
+        logger.info("Current Window Handles:  {}".format(cur_handles))
+
+        for handle in cur_handles:
+            if handle != appliance_handle:
+                # XXX: Add code to verify the tab has the correct widget
```
mshriver: **Required**

Is this too complex to include in the scope here? Needs to be a FIXME, ideally at least some bare minimum check implemented.

cfme/common/vm.py
```diff
@@ -347,6 +368,44 @@ def get_detail(self, properties=None, icon_href=False):
...
+            raise NotImplementedError('Not supported console type: {}'.format(console))
+
+        view = navigate_to(self, 'Details')
+
+        # Click console button given by type
+        view.toolbar.access.item_select(console)
+
+        # Get the consoles window handle, and then create a VMConsole object, and store
+        # the VMConsole object aside.
+        console_handle = self.console_handle
```
mshriver: **Required**
Basic handling of when console_handle returns none here. raise an exception.

cfme/tests/cloud_infra_common/test_html5_vm_console.py
mshriver: **Required**

I don't see an update to implement this yet?

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -0,0 +1,188 @@
...
+    """
+    Create a VM on the provider with the given template, and return the vm_obj.
+
+    Also, it will remove VM from provider using nested function _delete_vm
+    after the test is completed.
+
+    """
+    vm_obj = VM.factory(vm_name, provider, template_name=console_template)
+
+    @request.addfinalizer
```
mshriver: **Optional**
Consider using yield fixture like `configure_websocket`

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -0,0 +1,188 @@
...
+        - Using ssh we will check that the command worked (i.e. that the file
+          was created.
+    """
+    console_vm_username = credentials[provider.data.templates.get('console_template')
+                            ['creds']].get('username')
+    console_vm_password = credentials[provider.data.templates.get('console_template')
+                            ['creds']].get('password')
+
+    vm_obj.open_console(console='VM Console')
+    assert vm_obj.vm_console is not None, 'VMConsole object should be created'
```
mshriver: `is not None` is redundant here

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -0,0 +1,188 @@
...
+    vm_console = vm_obj.vm_console
+    try:
+        # If the banner/connection-status element exists we can get
+        # the connection status text and if the console is healthy, it should connect.
+        assert vm_console.wait_for_connect() is True, "VM Console did not reach 'connected' state"
```
mshriver: `is True` is redundant

cfme/tests/cloud_infra_common/test_html5_vm_console.py
kedark3: This is not working out with the cirros image. I will need to find way to get it work. Will submit another PR for that. Currently the way it is implemented is consistently working.

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -0,0 +1,188 @@
...
+
+        # Get the login screen image, and make sure it is a jpeg file:
+        screen = vm_console.get_screen()
+        assert imghdr.what('', screen) == 'jpeg'
+
+        assert vm_console.wait_for_text(text_to_find="login:", timeout=200) is not None,\
```
mshriver: `is not None` redundant

mshriver: Also prefer the statement wrapped in `()` instead of line continuation with `\`

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -0,0 +1,188 @@
...
+            "VM Console didn't prompt for Login"
+
+        # Enter Username:
+        vm_console.send_keys(console_vm_username)
+
+        assert vm_console.wait_for_text(text_to_find="Password", timeout=200) is not None,\
```
mshriver: `is not None` redundant

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -0,0 +1,188 @@
...
+            "VM Console didn't prompt for Password"
+        # Enter Password:
+        vm_console.send_keys("{}\n".format(console_vm_password))
+
+        time.sleep(5)  # wait for login to complete
```
mshriver: Can you call `vm_console.wait_for_text()` here instead of just sleeping?

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -0,0 +1,188 @@
...
+            # last line of the console will contain one of the following words:
+            # [login, password, incorrect]
+            # if so, regex_for_login_password will find it and result will not be []
+            # .split('\n')[-1] splits the console text on '\n' & picks last item of resulting list
+            result = regex_for_login_password.findall(vm_console.get_screen_text().split('\n')[-1])
+            return result == []
+
+        # if _try_login() returns True, it means we did not find any of words
+        # [login, password, incorrect] on last line of console text, which implies login success
+        wait_for(func=lambda: _try_login(), timeout=300, delay=5)
```
mshriver: **Required**

Remove lambda here, just pass the method without calling it

```
 wait_for(func=_try_login)

File: cfme/tests/cloud_infra_common/test_html5_vm_console.py [hunk 39]
```diff
@@ -0,0 +1,188 @@
...
+        vm_console.send_keys("{}\n".format(console_vm_password))
+
+        time.sleep(5)  # wait for login to complete
+
+        # This regex can find if there is a word 'login','password','incorrect' present in
+        # text, irrespective of its case
+        regex_for_login_password = re.compile(r'\blogin\b | \bpassword\b| \bincorrect\b',
+         flags=re.I | re.X)
+
+        def _try_login():
```
mshriver: **Required**

This name isn't accurate, this isn't trying to login, this is validating login was successful.

mshriver: **Optional**

Since this is just used inline, consider the decorator implementation of wait_for.

https://github.com/RedHatQE/wait_for/blob/master/wait_for/__init__.py#L187

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -0,0 +1,188 @@
...
+            result = regex_for_login_password.findall(vm_console.get_screen_text().split('\n')[-1])
+            return result == []
+
+        # if _try_login() returns True, it means we did not find any of words
+        # [login, password, incorrect] on last line of console text, which implies login success
+        wait_for(func=lambda: _try_login(), timeout=300, delay=5)
+
+        logger.info("Wait to get the '$' prompt")
+        if provider.one_of(VMwareProvider):
+            vm_console.wait_for_text(text_to_find="root@photon", timeout=200)
```
mshriver: **Required**

Don't hardcode this text, pull it from conf.

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -0,0 +1,188 @@
...
+            "VM Console didn't prompt for Login"
+
+        with ssh.SSHClient(hostname=vm_obj.ip_address, username=console_vm_username,
+                password=console_vm_password) as ssh_client:
+            # if file was created in previous steps it will be removed here
+            # we will get instance of SSHResult
+            command_result = ssh_client.run_command("rm bla*", ensure_user=True)
+            assert command_result
+    finally:
+        vm_console.close_console_window()
```
mshriver: :+1: good call putting this here

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -0,0 +1,188 @@
...
+
+        # create file on system
+        vm_console.send_keys("touch blather\n")
+        logger.info("Wait 5 sec for file creation")
+        time.sleep(5)  # Wait for file creation
+
+        # Test pressing ctrl-alt-delete...we should be able to get a new login prompt:
+        vm_console.send_ctrl_alt_delete()
+        vm_console.wait_for_text(text_to_find="login:", timeout=200, to_disappear=True)
+        assert vm_console.wait_for_text(text_to_find="login:", timeout=200) is not None,\
```
mshriver: `is not None` is redundant

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -0,0 +1,188 @@
...
+            "VM Console didn't prompt for Login"
+
+        with ssh.SSHClient(hostname=vm_obj.ip_address, username=console_vm_username,
+                password=console_vm_password) as ssh_client:
+            # if file was created in previous steps it will be removed here
+            # we will get instance of SSHResult
+            command_result = ssh_client.run_command("rm bla*", ensure_user=True)
```
mshriver: **Required**

You need to be more specific here, and DRY this and the `touch` call on L171

cfme/tests/cloud_infra_common/test_html5_vm_console.py
kedark3: After username and password is entered, sometimes it takes a short while before prompting:

- Either login was successful so I will see command prompt
- Or it failed and it will ask for login.

_try_login() will then be invoked to check if it was successful and if not, attempt for login.  wait_for_text on line 166 will find out if login was successful by trying to read prompt. But only for VMware PhotonOS, for cirros, there is only a "$" sign on prompt line and tesseract can't read it almost always.

cfme/cloud/provider/openstack.py
```diff
@@ -85,3 +89,18 @@ def from_config(cls, prov_config, prov_key, appliance=None):
                    tenant_mapping=prov_config.get('tenant_mapping', False),
                    infra_provider=infra_provider,
                    appliance=appliance)
+
+    # Following methods will only work if the remote console window is open
+    # and if selenium focused on it. These will not work if the selenium is
+    # focused on Appliance window.
```
mshriver: **Required**

Because of this, you need to wrap these calls in a `try` and call a consistent exception. Check out what's in `cfme.exceptions`, and import or add a new one if needed.

Allows callers to consistently handle how this fails when the browser tab selection hasn't worked correctly.

cfme/common/vm.py
jamesooden: XXX is the old traditional way of denoting FIXME, or TODO.   It's all over the RPM and anaconda code.

jamesooden: But probably what we should do is add  Jira card.

cfme/tests/cloud_infra_common/test_html5_vm_console.py
jamesooden: A jira card to track the needed work would be good.

cfme/common/vm.py
kedark3: Currently, we have routine that closes the browser window/tab containing console after each test, if I ran it against 3 different providers back to back, irrespective of pass or fail on one of the providers, the next provider will not see any window other than appliance window. Hence this logic is currently not implemented. It is consistently working so far. 
I will put it FIXME for now.

cfme/tests/cloud_infra_common/test_html5_vm_console.py
jamesooden: The reason it is not specific here is because presently the OpenStack and RHV VM consoles sometimes drop characters.   We have a Jira card to explore fixes for this:

https://projects.engineering.redhat.com/browse/RHCFQE-3592

Of course the dropping of characters sent is technically a bug, but though I don't know if they are dropped locally or remotely.

cfme/tests/cloud_infra_common/test_html5_vm_console.py
kedark3: it is not in conf, it is just how the prompt looks like after login. :/

cfme/tests/cloud_infra_common/test_html5_vm_console.py
mshriver: Thanks for explanation @jamesooden.

@kedark3 Please add comment detailing the failure here and that this is a workaround

**Consider**

Generating a random filename with a prefix, and `rm <prefix>*` here so that it can be unique test-to-test.

cfme/tests/cloud_infra_common/test_html5_vm_console.py
mshriver: Add it to conf, this is why we added templates as an array instead of simple maps.

cfme/cloud/provider/openstack.py
kedark3: All these methods will throw NoSuchElementException consistently. Do we still need to make it explicit in the code with try block?

cfme/tests/cloud_infra_common/test_html5_vm_console.py
kedark3: Gotcha :)

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -0,0 +1,188 @@
...
+            return result == []
+
+        # if _validate_login() returns True, it means we did not find any of words
+        # [login, password, incorrect] on last line of console text, which implies login success
+        wait_for(func=_validate_login, timeout=300, delay=5)
+
+        logger.info("Wait to get the '$' prompt")
+        if provider.one_of(VMwareProvider):
+            vm_console.wait_for_text(text_to_find=provider.data.templates.get('console_template')
+                            ['prompt_text'], timeout=200)
```
kedark3: Please merge GitLab PR for this to work : https://gitlab.cee.redhat.com/cfme-qe/cfme-qe-yamls/merge_requests/261

cfme/cloud/provider/openstack.py
kedark3: Added below ItemNotFound Exception. Please review and comment or resolve.

cfme/common/vm.py
kedark3: Done. Please review and resolve.

cfme/tests/cloud_infra_common/test_html5_vm_console.py
kedark3: @mshriver  comment is added.  For every test a fresh VM is created. So even though same name is used, we do not face any issues. because there is no file present on Home directory before and 'blather' becomes the first file.

cfme/common/vm_console.py
```diff
@@ -0,0 +1,203 @@
...
+        # due to an exception being thrown.
+        try:
+            text = self.provider.get_console_connection_status()
+            logger.info('Read following text from console banner: {}'.format(text))
+        except NoSuchElementException as e:
+            logger.info('Could not find banner element.')
+            logger.info('Exception: {}'.format(e))
+            return None
+
+        self.switch_to_appliance()
```
mshriver: This is a bad pattern, won't switch back to appliance handle on exception

mshriver:
Waiting on PRT results, as its limited to the providers that currently support the testing and should pass.

jamesooden:
It failed in PRT, but the cause is known and a Jira card exists for it.   Basically, it fails to get connect from the console, but this happens the first time you access the console on a new provider almost every time.   Here is the console output:

   http://10.16.4.32/trackerbot/pr/run/14654#nTRloeKq

The error is:

   >           assert vm_console.wait_for_connect(), "VM Console did not reach 'connected' state"

The Jira card talking about this is here:  

    https://projects.engineering.redhat.com/browse/RHCFQE-3577

mshriver:
PRT Run #14654 is false positive for downstream-57z.

@jamesooden both tests failed against downstream-58z.

mshriver:
@psav Please weigh in here, I'm ready to merge with the knowledge that there is followup already in-progress to improve and expand support here.

PR #4999, opened 2017-07-18
[1LP][RFR] Adding tests for archived/orphaned VMs

lkhomenk:
Purpose or Intent
=================
__Extending__ tests coverage for archived/orphaned VMs because we don't have any
{{pytest: --use-provider vsphere65-nested --long-running -k "archived or orphaned"}}

### DIFF
--- a/cfme/tests/infrastructure/test_vm_power_control.py
+++ b/cfme/tests/infrastructure/test_vm_power_control.py
@@ -332,7 +332,7 @@ def test_no_power_controls_on_archived_vm(testing_vm, archive_vm, soft_assert):
         * Verify the Power toolbar button is not visible
     """
     testing_vm.load_details()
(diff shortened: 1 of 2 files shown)


cfme/tests/infrastructure/test_vm_power_control.py
```diff
@@ -54,6 +57,14 @@ def archive_vm(provider, testing_vm):
 
 
 @pytest.fixture(scope="function")
+def orphane_vm(provider, testing_vm):
```
mkoura: I believe "orphane" is misspelled, correct form is "orphan".

psav: I believe it should actually be ```orphaned``` as it is supposed to provide an orphaned vm

lkhomenk: @mkoura @psav - updated

PR #5016, opened 2017-07-21
[RFR][NOTEST] Generate test case names that contain class names

mkoura:
We use test case names with class names in Polarion (e.g.
``TestProvidersRESTAPI.test_create_arbitration_profiles`` vs ``test_create_arbitration_profiles``).

This was not reflected in the XML files genereted by the xunit_tools.

PRT results are not relevant here.


mkoura:
wrt testing, I already used this for uploading new test cases to Polarion

PR #5020, opened 2017-07-23
[1LP][RFR] Added paginator to ContainerObjectAllBaseView + fixed test_tables_sort

gshefer:
- Changed ContainerObjectAllBaseView to inherit ProvidersView.
- fixed test_tables_sort

{{ pytest: cfme/tests/containers/test_tables_sort.py --use-provider cm-env2 }}

### DIFF
--- a/cfme/containers/provider/__init__.py
+++ b/cfme/containers/provider/__init__.py
@@ -10,7 +10,6 @@ from widgetastic_patternfly import (SelectorDropdown, Dropdown, BootstrapSelect,
                                     Input, Button, Tab)
 from widgetastic.widget import Text, View, TextInput
 from wrapanapi.utils import eval_strings
-from widgetastic_manageiq import Table
 from widgetastic.xpath import quote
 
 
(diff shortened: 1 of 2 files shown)


gshefer:
@izapolsk Cool, I modified accordingly hope it's better now.

cfme/common/provider_views.py
```diff
@@ -313,6 +314,7 @@ class ContainersProvidersView(ProvidersView):
      represents Main view displaying all Containers providers
     """
     table = Table(locator="//div[@id='list_grid']//table")
+    paginator = View.nested(PaginationPane)
```
izapolsk: EntitiesView which is used in ProvidersView is a bit sophisticated. So, you probably didn't notice that paginator is already there. Nevertheless,  paginator is already present there. You don't need to add a new one.

https://github.com/izapolsk/integration_tests/blob/master/widgetastic_manageiq/__init__.py#L1980

cfme/common/provider_views.py
```diff
@@ -17,7 +17,8 @@
                                   BaseEntitiesView,
                                   DynaTree,
                                   BootstrapTreeview,
-                                  ProviderEntity)
+                                  ProviderEntity,
+                                  PaginationPane)
```
izapolsk: you don't need this import

gshefer:
@izapolsk Done.

PR #5024, opened 2017-07-24
[1LP][RFR] Added wait_for to solve NoSuchElementException

vatsalparekh:
Purpose or Intent
=================
tests in `cfme/infrastructure/test_infta_quota.py` were failing with the error `NoSuchElementException`, but the locator seemed correct; So adding a `wait_for` to let the page load properly, and then go for further execution.

{{pytest: cfme/tests/infrastructure/test_infra_quota.py -k storage -vvv}}


vatsalparekh:
{{pytest: cfme/tests/infrastructure/test_infra_quota.py -vvv}}

PR #5031, opened 2017-07-25
[RFR] fix for case when OpenStack provider uses Oscilomenter instead of AMPQ

### LINKED ISSUE #5028: Can't add any Openstack provider with Ceilometer event source
Can't setup Openstack (both Infra and Cloud) provider using framework with `event_stream: Ceilometer`, because it tries to click on Validate button which doesn't exist for Ceilometer event source.

izapolsk:
Quick fix of Events endpoint for OpenStack provider.
Honestly, the best fix would be to turn this view into ConditionalView but I'm not sure that ConditionalView can be used along with View.include. Otherwise,  fill scheme has to be changed.

Closes #5028

{{pytest: cfme/tests/cloud/test_providers.py cfme/tests/infrastructure/test_providers.py}}

### DIFF
--- a/cfme/common/provider.py
+++ b/cfme/common/provider.py
@@ -198,7 +198,8 @@ class BaseProvider(Taggable, Updateable, SummaryMixin, Navigatable):
                             endp_view.fill(endpoint.credentials.view_value_mapping)
                         # sometimes we have cases that we need to validate even though
                         # there is no credentials, such as Hawkular endpoint
-                        if validate_credentials and hasattr(endp_view, 'validate'):
+                        if (validate_credentials and hasattr(endp_view, 'validate') and
+                                endp_view.validate.is_displayed):
                             # there are some endpoints which don't demand validation like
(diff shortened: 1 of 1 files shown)


cfme/common/provider.py
```diff
@@ -309,7 +310,8 @@ def update(self, updates, cancel=False, validate_credentials=True):
                             endp_view.fill(endpoint.credentials.view_value_mapping)
                     # sometimes we have cases that we need to validate even though
                     # there is no credentials, such as Hawkular endpoint
-                    if validate_credentials and hasattr(endp_view, 'validate'):
+                    if (validate_credentials and hasattr(endp_view, 'validate') and
+                       endp_view.validate.is_displayed):
```
mfalesni: these need to be indented differently:
```
if (very long expression
........that continues here):
....code goes here
```

izapolsk: yep, pycharm also adviced me this indention.

PR #5032, opened 2017-07-25
[RFR][NOTEST] Script for listing outdated test cases

mkoura:
The script creates a CSV file with test cases that are present in Polarion but are missing in pytest.

It's handy when you want to find out what test cases should be deactivated/deleted in Polarion.

### DIFF
--- a/scripts/list_outdated_testcases.py
+++ b/scripts/list_outdated_testcases.py
@@ -1,0 +1,204 @@
+#!/usr/bin/env python
+
+# -*- coding: utf-8 -*-
+# pylint: disable=logging-format-interpolation
+"""
+Compare output of 'pytest --collect-only' with data in Polarion.
+
+The outcome is a CSV file with test cases that might need to be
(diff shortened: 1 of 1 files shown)


scripts/list_outdated_testcases.py
```diff
@@ -0,0 +1,204 @@
...
+
+from dump2polarion.csvtools import get_imported_data
+
+
+# pylint: disable=invalid-name
+logger = logging.getLogger(__name__)
+
+
+class CompareException(Exception):
+    """testcases exception."""
```
izapolsk: there should be at least pass inside

mkoura: Why? Exception declaration with just docstring is perfectly functional and often recommended, see e.g. https://stackoverflow.com/a/26938914

izapolsk: I wouldn't say this is recommended. But I agree I missed that point that docstring could be enough in such case.

PR #5036, opened 2017-07-26
[WIP]Test autotagging providers

pavelzag:
WIP at the moment

PR #5037, opened 2017-07-26
[1LP][RFR]map_tags navigation and crud fixes

rlbabyuk:
Purpose or Intent
=================
Fixed map tags navigation, update and delete methods needed for new tests(https://github.com/ManageIQ/integration_tests/pull/5036)

### DIFF
--- a/cfme/configure/configuration/region_settings.py
+++ b/cfme/configure/configuration/region_settings.py
@@ -355,7 +355,7 @@ class MapTags(Navigatable, Pretty, Updateable):
             flash_message = 'Add of new Container Label Tag Mapping was cancelled by the user'
         else:
             view.add_button.click()
-            flash_message = 'Container Label Tag Mapping "{}" was saved'.format(self.label)
+            flash_message = 'Container Label Tag Mapping "{}" was added'.format(self.label)
 
         view = self.create_view(MapTagsAllView)
         view.flash.assert_success_message(flash_message)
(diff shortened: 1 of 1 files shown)


cfme/configure/configuration/region_settings.py
```diff
@@ -355,7 +355,7 @@ def create(self, cancel=False):
             flash_message = 'Add of new Container Label Tag Mapping was cancelled by the user'
         else:
             view.add_button.click()
-            flash_message = 'Container Label Tag Mapping "{}" was saved'.format(self.label)
+            flash_message = 'Container Label Tag Mapping "{}" was added'.format(self.label)
```
mkoura: Is this the same for 5.8+ and 5.7?

cfme/configure/configuration/region_settings.py
```diff
@@ -368,27 +368,32 @@ def update(self, updates, cancel=False):
...
+            )
         else:
             view.save_button.click()
             flash_message = 'Container Label Tag Mapping "{}" was saved'.format(self.label)
 
         view = self.create_view(MapTagsAllView, override=updates)
         view.flash.assert_success_message(flash_message)
 
-    def delete(self, cancel=True):
+    def delete(self, cancel=False):
```
mkoura: Will this change of default value affect existing tests?

cfme/configure/configuration/region_settings.py
rlbabyuk: yes the same for both versions, my typo correction

cfme/configure/configuration/region_settings.py
rlbabyuk: no, it won't affect, there are no existing tests for map tags now. But they are just in progress in https://github.com/ManageIQ/integration_tests/pull/5036 PR

PR #5041, opened 2017-07-27
[RFR][1LP] Fixing RHEV Chargeback failures

nachandr:
PRT results
---------------
5.8 
All tests passed

5.7
Tests skipped because of a BZ

### DIFF
--- a/cfme/common/provider.py
+++ b/cfme/common/provider.py
@@ -102,6 +102,10 @@ class BaseProvider(Taggable, Updateable, SummaryMixin, Navigatable):
         return self.type_name
 
     @property
+    def id(self):
+        return self.get_provider_id()
+
+    @property
     def version(self):
         return self.data['version']
 
--- a/cfme/common/provider.py
+++ b/cfme/common/provider.py
@@ -106,12 +106,7 @@ class BaseProvider(Taggable, Updateable, SummaryMixin, Navigatable):
         """"
         Return the ID associated with the specified provider name
         """
-        # TODO: Get Provider object from ProviderCollection.find, then use Provider.id to get the id
-        logger.debug('Retrieving the ID for provider: {}'.format(self.name))
-        for provider_id in self.get_all_provider_ids():
-            details = self.get_provider_details(provider_id)
-            if details['name'] == self.name:
-                return provider_id
+        return self.appliance.rest_api.collections.providers.find_by(name=self.name)[0].id
 
     @property
     def version(self):
--- a/cfme/common/provider.py
+++ b/cfme/common/provider.py
@@ -103,7 +103,15 @@ class BaseProvider(Taggable, Updateable, SummaryMixin, Navigatable):
 
     @property
     def id(self):
-        return self.get_provider_id()
+        """"
+        Return the ID associated with the specified provider name
+        """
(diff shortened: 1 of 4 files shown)


cfme/tests/intelligence/reports/test_validate_chargeback_report.py
```diff
@@ -213,21 +213,54 @@ def resource_usage(vm_ownership, appliance, provider):
...
+        # Verify that rollups are present in the metric_rollups table.
+        vm_name = provider.data['cap_and_util']['chargeback_vm']
+
+        ems = appliance.db.client['ext_management_systems']
+        metrics = appliance.db.client['metrics']
+
+        rc, out = appliance.ssh_client.run_rails_command(
+            "\"vm = Vm.where(:ems_id => {}).where(:name => {})[0];\
+            vm.perf_capture('realtime', 1.hour.ago.utc, Time.now.utc)\""
+            .format(provider_id, repr(vm_name)))
```
izapolsk: you use provider_id variable from outside. Please add "id" property to BaseProvider and use provider instead.

dajoRH:
*CFME QE Bot*

nachandr:
Closed by mistake.Reopening this.

cfme/common/provider.py
```diff
@@ -100,6 +100,10 @@ def type(self):
         return self.type_name
 
     @property
+    def id(self):
```
psav: **REQUIRED:** Two things here
1) The ```get_provider_id()``` method used requires a ```provider_name``` which you would probably ask why? I agree it shouldn't need a provider name if it is a class method.
2) The ```get_provider_id()``` method isn't used at all in the codebase. So I propose we switch the code into the ```id``` property and drop the requirement for the arg.  It's a required arg to, so any calls to this I would assume would fail.

psav: I verified it does fail....

dajoRH:
Lint report for commit d7527152f5a95f0b4409785056f21303ddb949c2:

`cfme/common/provider.py`:
- :red_circle: Line 2:1: **F401** *'pytest' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

cfme/common/provider.py
nachandr: @psav, Done

psav: If we have this implemented above using REST, why do we need to do it again here via the UI?

nachandr: @psav, I'm not quite sure why we don't have an id property in BaseProvider class in common/provider.py  in the first place, but have one in the InfraProvider class cfme/infrastructure/provider/__init__.py.Do let me know if I'm missing something.Anyway, I placed the id property in  the BaseProvider class in common/provider.py and removed it from cfme/infrastructure/provider/__init__.py since that would be just a duplicate override.Both InfraProvider and CloudProvider are child classes of CloudInfraProvider, which in turn is a child class of  BaseProvider class.I tested the code for both cloud and infra providers and it works fine.

Also, that new id property in BaseProvider class uses REST to fetch provider ID.If you are okay with this, I could push my code changes.

nachandr: @psav, I've made changes as per my previous comment.I've also updated the PRT results in the comment section.

Could you review and merge the PR?

PR #4908, opened 2017-06-22
[1LP][RFR] Convert AnalysisProfile to Widgetastic

mshriver:
Separate AnalysisProfile into its own module
Widgetastic conversion of the pages and CRUD

### Included tests
All tests in `cfme/tests/configure/test_analysis_profiles.py` should pass.

NOT Including test_compliance.py and test_instance_analysis.py because I slightly changed the yield_fixture there for an AnalysisProfile object, and the tests pretty much completely failed for unrelated reasons.

{{ pytest: -v --long-running cfme/tests/configure/test_analysis_profiles.py }}

PR #5048, opened 2017-07-27
[1LP][RFR] Update wt_manageiq, add DynamicTable

mshriver:
Add an miq implementation for dynamic table, covers those with Add/Save buttons in the right-hand column.

Note this may not be everything DynamicTable eventually needs, but satisfies the required actions for #4908, AnalysisProfile implementation. We'll certainly need a row_delete method as well.

Needed by #4908

Depends on ~~https://github.com/RedHatQE/widgetastic.core/pull/52~~  <- merged, need a release with it.

### DIFF
--- a/widgetastic_manageiq/__init__.py
+++ b/widgetastic_manageiq/__init__.py
@@ -41,6 +41,11 @@ from widgetastic_patternfly import (
 from cfme.exceptions import ItemNotFound, ManyEntitiesFound
 
 
+class DynamicTableAddError(Exception):
+    """Raised when an attempt to add a row to a dynamictable fails"""
+    pass
+
+
 class DynaTree(Widget):
     """ A class directed at CFME Tree elements
 
@@ -1970,8 +1975,8 @@ class BaseTileIconEntity(ParametrizedView):
     @property
     def is_displayed(self):
         try:
-            return super(BaseTileIconEntity, self).is_displayed and \
-                   self.browser.is_displayed(self.LIST)
+            return (super(BaseTileIconEntity, self).is_displayed and
+                    self.browser.is_displayed(self.LIST))
         except NoSuchElementException:
             return False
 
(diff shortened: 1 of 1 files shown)


widgetastic_manageiq/__init__.py
```diff
@@ -1947,8 +1952,8 @@ def fill(self, values):
     @property
     def is_displayed(self):
         try:
-            return super(BaseTileIconEntity, self).is_displayed and \
-                   self.browser.is_displayed(self.LIST)
+            return (super(BaseTileIconEntity, self).is_displayed and
```
mshriver: lint was complaining about this

widgetastic_manageiq/__init__.py
```diff
@@ -2174,3 +2179,40 @@ class ProviderEntity(BaseEntity):
...
+        super(DynamicTable, self).__init__(*args, **kwargs)
+
+    def row_add(self):
+        """Use the action-cell column widget to add a row
+
+        Clicks on the row directly, not the action button
+        """
+        try:
+            self[self.action_row].click()
+            return self.action_row
```
mfalesni: In the case action_row is negative, then the negative index is relative to the end of the table, which can move in this case. I recommend that in the case action_row is negative, then before doing the click convert the negative index into positive so it is absolute. Then access the row with this value and also return it.

mshriver: @mfalesni I believe I understand the concern here, but I don't think that I need to specifically handle that case here.

Since I'm using `Table[row_index]` access method here, its calling `widgetastic.Table.__getitem__()`, which is calling `widgetastic.Table._process_negative_index()` and handling the negative index appropriately.

Returning this same index allows the caller to further access the row using the same __getitem__ code path.

Does this structure not handle negative indexes in some case? Shouldn't a -1 index work regardless of the table end moving, since it is relative?

mshriver: ping @mfalesni, thoughts on this already being handled by `__getitem__`?

mfalesni: Hmm, I will look.

mfalesni: So, yes, getitem handles it, but only for itself, so if action_row is negative, getitem will correctly handle it, but then you return action_row again, which is always pointing to the "add row" when negative. So my concern is that when new row is added, and the number is negative, it will return the wrong row unless the row turns itself into the editing mode, and by returning the negative, the machinery that uses the result of this method's call will not have a stable anchor point since if anything changes in the table, it will suddenly operate on a different row. So I would pre-process the action_row with the _process_negative_index here as well to have a positive index :)

mshriver: Understood.

I guess this hinges on what row_add should return, if anything, with your intent being to return the index of the row that was just added. That wasn't my intent with returning action_row, but I see how in widgetastic.widget.Table.fill the return from row_add is used to index the table.

Will do.

dajoRH:
*CFME QE Bot*

mshriver:
@mfalesni requesting re-review now that we have widgetastic 0.10.0 with some Table updates.

#5129

widgetastic_manageiq/__init__.py
```diff
@@ -2322,3 +2327,58 @@ def fill(self, values):
...
+    def row_add(self):
+        """Use the action-cell column widget to add a row
+
+        Clicks on the row directly, not the action button
+
+        Returns:
+            int positive row index of the action row where the new widgets should be displayed
+        """
+        # convert action_row into a positive index
+        pos_action_index = (self.action_row
```
izapolsk: please use ordinary if instead. ternary operator should fit one line.

widgetastic_manageiq/__init__.py
```diff
@@ -2322,3 +2327,58 @@ def fill(self, values):
 
     def read(self):
         return self.all_shortcuts
+
+
+class DynamicTable(VanillaTable):
```
mshriver: In this case the table doesn't support the other methods provided by manageiq.Table - there are no check_all or row checkboxes, nor are the editable tables sortable.

For these reasons I'd like to keep them separate, though I do understand your want to minimize the class sprawl. If you feel strongly about minimizing the classes I could be convinced to combine them.

mfalesni: If it is true that this table is likely to not have the checkboxes, I am for having a separate class.

mfalesni: Remember, people will ask.

PR #5053, opened 2017-07-28
Fix cleanup_old_vms, logging typo

mshriver:


mshriver:
Missed one logging formatter change.

PR #5060, opened 2017-07-31
[1LP][RFR] fix NameError in genealogy

apagac:
This is a quick fix for existing test. The test was failing with NameError exception, so I introduced a try+except block, sent the exception to log and the test is now failing with pytest.fail.

### DIFF
--- a/cfme/tests/cloud_infra_common/test_genealogy.py
+++ b/cfme/tests/cloud_infra_common/test_genealogy.py
@@ -8,6 +8,7 @@ from cfme.cloud.provider import CloudProvider
 from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.infrastructure.provider.scvmm import SCVMMProvider
 from utils.generators import random_vm_name
+from utils.log import logger
 
 pytestmark = [
     pytest.mark.usefixtures('uses_infra_providers', 'uses_cloud_providers', 'provider'),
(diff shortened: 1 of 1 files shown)


cfme/tests/cloud_infra_common/test_genealogy.py
```diff
@@ -87,6 +88,10 @@ def _():
             parent = pytest.sel.text(opt).strip()
         assert parent.startswith(small_template), "The parent template not detected!"
     else:
-        vm_crud_ancestors = vm_crud.genealogy.ancestors
+        try:
+            vm_crud_ancestors = vm_crud.genealogy.ancestors
+        except NameError:
+            logger.exception("The parent template not detected!")
```
jkrocil: REQUIRED: If we are going to log it, let's log the original exception, not our custom message.

cfme/tests/cloud_infra_common/test_genealogy.py
```diff
@@ -87,6 +88,10 @@ def _():
...
+            raise pytest.fail("The parent template not detected!")
```
jkrocil: OPTIONAL: Would it make sense to log the name of the small template here as well? I'm not 100% sure about the context here - would it make it easier to debug?

cfme/tests/cloud_infra_common/test_genealogy.py
apagac: logger.exception will send original exception to log. The custom message is redundant.

cfme/tests/cloud_infra_common/test_genealogy.py
apagac: This is a parent template/vm of a current vm, in genealogy context. If NameError occurs, no parent has been detected. The error message displayed reflects this, I believe.

cfme/tests/cloud_infra_common/test_genealogy.py
jkrocil: Ah, gotcha! I didn't realize it will log both the exception and the message. :smile: Just learned something new :+1:

cfme/tests/cloud_infra_common/test_genealogy.py
jkrocil: Ok

PR #5065, opened 2017-07-31
[RFR][1LP] Adding tesseract dependency in quickstart

kedark3:
Purpose or Intent
=================

__Extending__ quickstart to include the "tesseract" package. As  @RonnyPfannschmidt pointed it out that PRT may start using quickstart to build containers for testing, I would need those to contain the package mentioned above in order to get PR #4998 tested on PRT before it is merged.


kedark3:
@RonnyPfannschmidt Yes. This package can be installed on Fedora 23-26. RHEL is also supposed to have it but might need subscription for that.

PR #5076, opened 2017-08-02
[1LP][RFR] refactoring ProviderDetailsView

izapolsk:
1. According to hhovsepy, this view didn't work well with Middleware providers
2. this view had to be refactored since CondtionalView widget was developed

{{pytest: cfme/tests/infrastructure/test_providers.py cfme/tests/cloud/test_providers.py}}


izapolsk:
tests failed only for rhv41 provider, which has wrong credentials in yaml

PR #5085, opened 2017-08-02
[WIP] Adding test case for testing of landing page for rbac users

anurag03:
This test case automates the checks for the setting of landing page for new custom role based users.

{{pytest: cfme/tests/configure/test_start_page.py -v }}

### DIFF
--- a/cfme/tests/configure/test_start_page.py
+++ b/cfme/tests/configure/test_start_page.py
@@ -1,0 +1,70 @@
+import pytest
+import fauxfactory
+
+from cfme.base.credential import Credential
+from cfme.configure.settings import Visual
+from cfme.configure.access_control import Role, Group, User
+from cfme.utils.appliance.implementations.ui import navigate_to
+
+
+faux_value = fauxfactory.gen_alphanumeric()
+pytestmark = pytest.mark.usefixtures('browser')
+
+
+def create_role():
+    role = Role(name='test{}'.format(faux_value))
+    role.create()
+
+
+def create_group():
+    role = 'test' + faux_value
+    group = Group(description='test{}'.format(faux_value), role=role)
+    group.create()
+
+
+def create_user():
+    group_user = Group('test{}'.format(faux_value))
+    username = 'test{}'.format(faux_value.lower())
+    password = 'test'
+    cred = Credential(principal=username, secret=password)
+    user = User(name='test{}'.format(faux_value),
+                credential=cred, group=group_user)
+
+    user.create()
+    return user
+
+
+def user_login(user, appliance):
+    login_page = navigate_to(appliance.server, 'LoginScreen')
+    login_page.log_in(user)
+
+
+def set_landing_page(user, values, appliance):
+    # This list contains the list of pages which show some error or alerts after login.
+    # TODO remove all these pages when BZ is closed.
+    page_list = ['Bottlenecks', 'Automate Log']
+    for value in values:
+        view = navigate_to(Visual, 'All')
+        if (view.visualstartpage.show_at_login.fill(value.text) and
+                not any(substring in value.text for substring in page_list)):
+            view.save.click()
+            user_login(user, appliance)
+        logged_in_page = navigate_to(appliance.server, 'LoggedIn')
+        if not logged_in_page.is_displayed:
+            return False
+    return True
+
+
+def test_start_page_rbac(appliance):
+    """
+        This test checks the functioning of the landing page; 'Start at Login'
+        option on 'Visual' tab of setting page for custom role based users.
+    """
+    create_role()
+    create_group()
+    user = create_user()
+    appliance.server.logout()
+    user_login(user, appliance)
+    view = navigate_to(Visual, 'All')
+    values = view.visualstartpage.show_at_login.all_options
+    assert set_landing_page(user, values, appliance)
--- a/cfme/tests/configure/test_start_page.py
+++ b/cfme/tests/configure/test_start_page.py
@@ -11,27 +11,40 @@ faux_value = fauxfactory.gen_alphanumeric()
 pytestmark = pytest.mark.usefixtures('browser')
 
 
-def create_role():
+def get_landing_page():
+    view = navigate_to(Visual, 'All')
+    landing_pages = view.visualstartpage.show_at_login.all_options
+    start_page = []
+    for landing_page in landing_pages:
+        start_page.append(landing_page.text)
+    return start_page
+
+
+@pytest.yield_fixture(scope="module")
+def rbac_user(appliance):
     role = Role(name='test{}'.format(faux_value))
     role.create()
-
-
-def create_group():
-    role = 'test' + faux_value
-    group = Group(description='test{}'.format(faux_value), role=role)
+    role_name = 'test' + faux_value
+    group = Group(description='test{}'.format(faux_value), role=role_name)
     group.create()
-
-
-def create_user():
     group_user = Group('test{}'.format(faux_value))
     username = 'test{}'.format(faux_value.lower())
     password = 'test'
     cred = Credential(principal=username, secret=password)
     user = User(name='test{}'.format(faux_value),
                 credential=cred, group=group_user)
-
     user.create()
-    return user
+    appliance.server.logout()
+    user_login(user, appliance)
+    view = navigate_to(Visual, 'All')
+    values = view.visualstartpage.show_at_login.all_options
+    yield user, values
+    appliance.server.logout()
+    login_page = navigate_to(appliance.server, 'LoginScreen')
+    login_page.login_admin()
+    user.delete()
+    group.delete()
+    role.delete()
 
 
 def user_login(user, appliance):
@@ -39,7 +52,7 @@ def user_login(user, appliance):
     login_page.log_in(user)
 
 
-def set_landing_page(user, values, appliance):
+def set_landing_page_rbac(user, values, appliance):
     # This list contains the list of pages which show some error or alerts after login.
     # TODO remove all these pages when BZ is closed.
     page_list = ['Bottlenecks', 'Automate Log', 'Compute / Containers / Containers']
@@ -55,16 +68,37 @@ def set_landing_page(user, values, appliance):
     return True
 
 
-def test_start_page_rbac(appliance):
+def set_landing_page_admin(value, appliance):
+    # This list contains the list of pages which show some error or alerts after login.
+    # TODO remove all these pages when BZ is closed.
+    page_list = ['Bottlenecks', 'Automate Log', 'Compute / Containers / Containers']
+    view = navigate_to(Visual, 'All')
+    if (view.visualstartpage.show_at_login.fill(value) and
+            not any(substring in value for substring in page_list)):
+        view.save.click()
+        login_page = navigate_to(appliance.server, 'LoginScreen')
+        login_page.login_admin()
+        logged_in_page = navigate_to(appliance.server, 'LoggedIn')
+        if not logged_in_page.is_displayed:
+            return False
+    return True
+
+
+def test_start_page_rbac(appliance, rbac_user):
     """
         This test checks the functioning of the landing page; 'Start at Login'
-        option on 'Visual' tab of setting page for custom role based users.
+        option on 'Visual' tab of setting page for custom role based users. This test case doesn't
+        check the exact page but verifies that all the landing page options works properly.
     """
-    create_role()
-    create_group()
-    user = create_user()
-    appliance.server.logout()
-    user_login(user, appliance)
-    view = navigate_to(Visual, 'All')
-    values = view.visualstartpage.show_at_login.all_options
-    assert set_landing_page(user, values, appliance)
+    user, start_page = rbac_user
+    assert set_landing_page_rbac(user, start_page, appliance)
+
+
+@pytest.mark.parametrize('start_page', get_landing_page(), scope="module")
+def test_strat_page_admin(start_page, appliance):
+    """
+        This test checks the functioning of the landing page; 'Start at Login'
+        option on 'Visual' tab of setting page for administrator. This test case doesn't
+        check the exact page but verifies that all the landing page options works properly.
+    """
+    assert set_landing_page_admin(start_page, appliance)
--- a/cfme/tests/configure/test_start_page.py
+++ b/cfme/tests/configure/test_start_page.py
@@ -3,7 +3,8 @@ import fauxfactory
 
 from cfme.base.credential import Credential
 from cfme.configure.settings import Visual
-from cfme.configure.access_control import Role, Group, User
+from cfme.fixtures.tag import role  #noqa
+from cfme.configure.access_control import Group, User
 from cfme.utils.appliance.implementations.ui import navigate_to
 
 
@@ -21,11 +22,9 @@ def get_landing_page():
 
 
 @pytest.yield_fixture(scope="module")
-def rbac_user(appliance):
-    role = Role(name='test{}'.format(faux_value))
-    role.create()
-    role_name = 'test' + faux_value
-    group = Group(description='test{}'.format(faux_value), role=role_name)
+def rbac_user(appliance, role):
+    role = role
+    group = Group(description='test{}'.format(faux_value), role=role.name)
     group.create()
     group_user = Group('test{}'.format(faux_value))
     username = 'test{}'.format(faux_value.lower())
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit 977c2f14e199a103fec853719f0c967d4841ba6e:

`cfme/tests/configure/test_start_page.py`:
- :bangbang: Line 27:1: **E305** *expected 2 blank lines after class or function definition, found 0*
- :bangbang: Line 31:28: **E226** *missing whitespace around arithmetic operator*
- :bangbang: Line 37:18: **E226** *missing whitespace around arithmetic operator*
- :bangbang: Line 38:37: **E226** *missing whitespace around arithmetic operator*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/configure/test_start_page.py
```diff
@@ -0,0 +1,70 @@
...
+pytestmark = pytest.mark.usefixtures('browser')
+
+
+def create_role():
+    role = Role(name='test{}'.format(faux_value))
+    role.create()
+
+
+def create_group():
+    role = 'test' + faux_value
```
quarckster: Please use `format`

cfme/tests/configure/test_start_page.py
```diff
@@ -0,0 +1,70 @@
...
+from cfme.configure.settings import Visual
+from cfme.configure.access_control import Role, Group, User
+from cfme.utils.appliance.implementations.ui import navigate_to
+
+
+faux_value = fauxfactory.gen_alphanumeric()
+pytestmark = pytest.mark.usefixtures('browser')
+
+
+def create_role():
```
quarckster: This should be a fixture.

cfme/tests/configure/test_start_page.py
```diff
@@ -0,0 +1,70 @@
...
+    role = Role(name='test{}'.format(faux_value))
+    role.create()
+
+
+def create_group():
```
quarckster: And this

cfme/tests/configure/test_start_page.py
```diff
@@ -0,0 +1,70 @@
...
+    role = 'test' + faux_value
+    group = Group(description='test{}'.format(faux_value), role=role)
+    group.create()
+
+
+def create_user():
```
quarckster: And this

cfme/tests/configure/test_start_page.py
```diff
@@ -0,0 +1,104 @@
...
+    landing_pages = view.visualstartpage.show_at_login.all_options
+    start_page = []
+    for landing_page in landing_pages:
+        start_page.append(landing_page.text)
+    return start_page
+
+
+@pytest.yield_fixture(scope="module")
+def rbac_user(appliance):
+    role = Role(name='test{}'.format(faux_value))
```
rlbabyuk: **Required:** Use role fixture here (cfme/fixtures/tag.py)

cfme/tests/configure/test_start_page.py
```diff
@@ -0,0 +1,104 @@
...
+    role.create()
+    role_name = 'test' + faux_value
+    group = Group(description='test{}'.format(faux_value), role=role_name)
+    group.create()
+    group_user = Group('test{}'.format(faux_value))
+    username = 'test{}'.format(faux_value.lower())
+    password = 'test'
+    cred = Credential(principal=username, secret=password)
+    user = User(name='test{}'.format(faux_value),
```
rlbabyuk: **Suggestion:**  Also can restricted_user fixture used here if group with tag doesn't affect the tests result

cfme/tests/configure/test_start_page.py
```diff
@@ -0,0 +1,104 @@
...
+    yield user, values
+    appliance.server.logout()
+    login_page = navigate_to(appliance.server, 'LoginScreen')
+    login_page.login_admin()
+    user.delete()
+    group.delete()
+    role.delete()
+
+
+def user_login(user, appliance):
```
rlbabyuk: **Required:** this function can be replaced with:
with user:
       user_steps_here
Here a user will be logged in, will do the steps and log out

cfme/tests/configure/test_start_page.py
```diff
@@ -0,0 +1,104 @@
...
+    login_page = navigate_to(appliance.server, 'LoginScreen')
+    login_page.log_in(user)
+
+
+def set_landing_page_rbac(user, values, appliance):
```
rlbabyuk: **Required:**  This and set_landing_page_admin can be one function, they are almost the same

cfme/tests/configure/test_start_page.py
```diff
@@ -0,0 +1,104 @@
...
+        view.save.click()
+        login_page = navigate_to(appliance.server, 'LoginScreen')
+        login_page.login_admin()
+        logged_in_page = navigate_to(appliance.server, 'LoggedIn')
+        if not logged_in_page.is_displayed:
+            return False
+    return True
+
+
+def test_start_page_rbac(appliance, rbac_user):
```
rlbabyuk: If use one set_landing_page function, this tests can be parametrize with users to use

dajoRH:
Lint report for commit f13c463c712cef6290c31c6e49968b169e18de74:

`cfme/tests/configure/test_start_page.py`:
- :red_circle: Line 6:1: **F401** *'cfme.fixtures.tag.role' imported but unused*
- :bangbang: Line 6:37: **E262** *inline comment should start with '# '*
- :red_circle: Line 24:1: **F811** *redefinition of unused 'role' from line 6*

**No commit flaws detected**.

*CFME QE Bot*

anurag03:
Not able to resolve lint error, and the failure in 5.8z is because of ssh issue, so changing state from [WIPTEST] to [RFR].

cfme/tests/configure/test_start_page.py
```diff
@@ -0,0 +1,88 @@
+import pytest
+import fauxfactory
+
+from cfme.base.credential import Credential
+from cfme.configure.settings import Visual
+from cfme.fixtures.tag import role  #noqa
```
quarckster: **REQUIRED**
You don't need to import a fixture to use it in the module.

anurag03: Okay

cfme/tests/configure/test_start_page.py
```diff
@@ -0,0 +1,87 @@
...
+    username = 'test{}'.format(faux_value.lower())
+    password = 'test'
+    cred = Credential(principal=username, secret=password)
+    user = User(name='test{}'.format(faux_value),
+                credential=cred, group=group_user)
+    user.create()
+    login_page = navigate_to(appliance.server, 'LoginScreen')
+    login_page.log_in(user)
+    view = navigate_to(Visual, 'All')
+    landing_pages = view.visualstartpage.show_at_login.all_options
```
mshriver: **Question/Required**
why not use `get_landing_page` here, it's the same block of code.

cfme/tests/configure/test_start_page.py
```diff
@@ -0,0 +1,87 @@
...
+    for landing_page in landing_pages:
+        start_page.append(landing_page.text)
+    yield user, landing_pages
+    login_page = navigate_to(appliance.server, 'LoginScreen')
+    login_page.login_admin()
+    user.delete()
+    group.delete()
+
+
+def set_landing_page(value, appliance, user=None):
```
mshriver: Please use more descriptive parameter names here, consider `page_name` instead of `value`

cfme/tests/configure/test_start_page.py
```diff
@@ -0,0 +1,87 @@
...
+    # This list contains the list of pages which show some error or alerts after login.
+    # TODO remove all these pages when BZ is closed.
+    page_list = ['Bottlenecks', 'Automate Log', 'Compute / Containers / Containers']
+    view = navigate_to(Visual, 'All')
+    if (view.visualstartpage.show_at_login.fill(value) and
```
mshriver: **Required**
Switch the order of these statements in the `if`, so that the test doesn't spend time filling forms that it won't save anyway because the page is excluded.

cfme/tests/configure/test_start_page.py
```diff
@@ -0,0 +1,87 @@
...
+            not any(substring in value for substring in page_list)):
+        view.save.click()
+        login_page = navigate_to(appliance.server, 'LoginScreen')
```
mshriver: **Required**
Return here, and move the actual validation into the test.

cfme/tests/configure/test_start_page.py
```diff
@@ -0,0 +1,87 @@
...
+        check the exact page but verifies that all the landing page options works properly.
+    """
+    user, start_pages = rbac_user
+    values = [(user, start_page[0]) for start_page in start_pages]
+    for value in values:
+        assert set_landing_page(value[1], appliance, value[0])
+
+
+@pytest.mark.parametrize('start_page', get_landing_page(), scope="module")
+def test_strat_page_admin(start_page, appliance):
```
psav: **REQUIRED:** Typo in the test name

cfme/tests/configure/test_start_page.py
```diff
@@ -0,0 +1,87 @@
...
+    view = navigate_to(Visual, 'All')
+    if (view.visualstartpage.show_at_login.fill(value) and
+            not any(substring in value for substring in page_list)):
+        view.save.click()
+        login_page = navigate_to(appliance.server, 'LoginScreen')
+        if user is None:
+            login_page.login_admin()
+        else:
+            login_page.log_in(user)
+        logged_in_page = navigate_to(appliance.server, 'LoggedIn')
```
mshriver: **Required**
This is logging in as appliance.user, completely negating the `login_admin` or `log_in` calls that you made on line 59 or 61.

This needs to be removed, and you need to create a BaseLoggedInPage view object in order to check if the login was successful.

Please checkout how the LoggedIn step works, and consider reworking the test to actually set appliance.user and just use the navigation destination.

cfme/tests/configure/test_start_page.py
```diff
@@ -0,0 +1,87 @@
...
+        if not logged_in_page.is_displayed:
+            return False
+    return True
+
+
+def test_start_page_rbac(appliance, rbac_user):
```
psav: **REQUIRED:** This test needs to be parametrized, and rbac_user should not return the landing pages. This is hiding the contents of a fixture/parameter inside another fixture and needs to be fixed before merging. This test looks almost identical to the last one and should be structured similarly.

cfme/tests/configure/test_start_page.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def set_landing_page(value, appliance, user=None):
+    # This list contains the list of pages which show some error or alerts after login.
+    # TODO remove all these pages when BZ is closed.
+    page_list = ['Bottlenecks', 'Automate Log', 'Compute / Containers / Containers']
+    view = navigate_to(Visual, 'All')
+    if (view.visualstartpage.show_at_login.fill(value) and
+            not any(substring in value for substring in page_list)):
+        view.save.click()
```
mshriver: **Required**
I think you should use the `Visual.login_page` setter here to fill the form. Taking into account my comment above about the order of the statements in the `if`, you can just check the page_list and then use the setter.

cfme/tests/configure/test_start_page.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def test_start_page_rbac(appliance, rbac_user):
+    """
+        This test checks the functioning of the landing page; 'Start at Login'
+        option on 'Visual' tab of setting page for custom role based users. This test case doesn't
+        check the exact page but verifies that all the landing page options works properly.
+    """
+    user, start_pages = rbac_user
+    values = [(user, start_page[0]) for start_page in start_pages]
```
psav: **REQUIRED:** This seems very difficult to read. You have ```user```, which is static as it's come from ```rbac_user```,  which isn't parametrized. which is then rolled into a tuple, and placed in a list, before being unpacked again by accessing an index. What started as a nicely named variable called ```user```, became ```value[0]``` which is a lot more difficult to track.

The below example illustrates that we can completely remove the list comprehension, and the insertion of ```user``` into the tuple.

```python
    user, start_pages = rbac_user
    for start_page in start_pages:
        assert set_landing_page(start_page, appliance, user)
```

cfme/tests/configure/test_start_page.py
```diff
@@ -0,0 +1,87 @@
...
+from cfme.configure.settings import Visual
+from cfme.configure.access_control import Group, User
+from cfme.utils.appliance.implementations.ui import navigate_to
+
+
+faux_value = fauxfactory.gen_alphanumeric()
+pytestmark = pytest.mark.usefixtures('browser')
+
+
+def get_landing_page():
```
psav: **REQUIRED:** Though it's clever, I'm not a fan of dynamic parametrization. The only "dynamic" parameter we have right now is ```provider``` and as we have discussed in the team calls, we are going to get rid of it in favour of statically defined parameters.

There are reasons for this
1) Parameters link to test cases, and so making them dynamic means you are not sure what your test plan really is
2) They can be screwed up and tricked, leading to bad test cases being generated, particularly for the last case where these are parametrized.
3) This requires that the landing pages are read at collection time. This is a **huge** no no. It means that we MUST have an appliance in order to do collection which is simply not allowed. This should be switched to a static list of landing pages.

cfme/tests/configure/test_start_page.py
```diff
@@ -0,0 +1,87 @@
+import pytest
+import fauxfactory
+
+from cfme.base.credential import Credential
+from cfme.configure.settings import Visual
+from cfme.configure.access_control import Group, User
+from cfme.utils.appliance.implementations.ui import navigate_to
+
+
+faux_value = fauxfactory.gen_alphanumeric()
```
psav: **REQUIRED:** There is no need for this to be at the modular level, your group is module scoped so that call to the function can just be inside there.

cfme/tests/configure/test_start_page.py
```diff
@@ -0,0 +1,87 @@
...
+pytestmark = pytest.mark.usefixtures('browser')
+
+
+def get_landing_page():
+    view = navigate_to(Visual, 'All')
+    landing_pages = view.visualstartpage.show_at_login.all_options
+    start_page = []
```
psav: **OPTIONAL:** This is actuall a great use for list_comprehension
```python
return [landing_page.text for landing_page in landing_pages]
```

cfme/tests/configure/test_start_page.py
```diff
@@ -0,0 +1,87 @@
...
+        start_page.append(landing_page.text)
+    return start_page
+
+
+@pytest.yield_fixture(scope="module")
+def rbac_user(appliance, role):
+    role = role
+    group = Group(description='test{}'.format(faux_value), role=role.name)
+    group.create()
+    group_user = Group('test{}'.format(faux_value))
```
psav: **REQUIRED:** Where is ```group``` used? We create ```group```, but then we pass ```group_user``` into the ```User``` object. We should be able to pass ```group``` to ```User```. Any reason not to?

cfme/tests/configure/test_start_page.py
psav: **REQUIRED:** We already have methods for setting the landing page I think, why are we rewriting them?

cfme/tests/configure/test_start_page.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+@pytest.mark.parametrize('start_page', get_landing_page(), scope="module")
+def test_strat_page_admin(start_page, appliance):
+    """
+        This test checks the functioning of the landing page; 'Start at Login'
+        option on 'Visual' tab of setting page for administrator. This test case doesn't
+        check the exact page but verifies that all the landing page options works properly.
+    """
+    assert set_landing_page(start_page, appliance)
```
psav: **REQUIRED:** This test feels like it's too magic. What are we testing? There is a little too much abstraction here.

mshriver:
I had more comments - but it looks like @psav started reviewing right after me and had a lot of the same thoughts :)

@anurag03 I'm also available to help you out with these comments, and with reworking this test.

anurag03:
Yes will make required changes @psav @mshriver

cfme/tests/configure/test_start_page.py
anurag03: When we set the start up page many times we face issues like: 
startup_page

When this issue arises we are not able to access any page at all, so wanted a test case to keep a track on this issue.

anurag03: Will change the name of the test case

psav: We have a bug for this?

anurag03:
Will be raising a new PR for this test case addition.

PR #5088, opened 2017-08-03
[WIP] Use pre-created cmqe singer cert for SSL

jaryn:
The CMQE is using pre-created static cert for signing services deployed by us. This cert needs to be injected into CFME system (/etc/pki/ca-trust/anchors) and `update-ca-trust` before evmserverd is started to make it trust the endpoints.

Previous cert was not the correct one to use. It was valid only for one of our deployments.

It might be better to have the cert defined in the credentials.yaml, but I will leave this for the future PR


utils/appliance/__init__.py
```diff
@@ -1088,6 +1089,24 @@ def deploy_merkyl(self, start=False, log_callback=None):
             log_callback("Setting it to start after reboot")
             client.run_command("chkconfig merkyl on")
 
+    @logger_wrap("Adding ca certs: {}")
+    def add_ca_certs(self, log_callback=None):
```
quarckster: **REQUIRED**
This method deploys only one certificate and has a name `add_ca_certs`. Please add `certificates` argument and rewrite the code.

dajoRH:
*CFME QE Bot*

PR #5095, opened 2017-08-03
[1LP][RFR] This PR contains collection in keypairs.

sbulage:
Shriver adding prt line:
{{ pytest: cfme/tests/cloud/test_keypairs.py -v --long-running --use-provider rhos7-ga --use-provider ec2west }}

### DIFF
--- a/cfme/cloud/keypairs.py
+++ b/cfme/cloud/keypairs.py
@@ -4,10 +4,8 @@ from widgetastic.widget import View
 from widgetastic.utils import VersionPick, Version
 from widgetastic.exceptions import NoSuchElementException
 from widgetastic_patternfly import Dropdown, Button, FlashMessages
-
-from cfme.base.ui import BaseLoggedInPage
-from cfme.common import SummaryMixin, Taggable
 from cfme.exceptions import KeyPairNotFound
+from cfme.base.ui import BaseLoggedInPage
 from cfme.web_ui import match_location, mixins
 from utils.appliance.implementations.ui import navigate_to, navigator, CFMENavigateStep
 from utils.appliance import Navigatable
@@ -53,6 +51,7 @@ class KeyPairEntities(View):
 class KeyPairDetailsEntities(View):
     breadcrumb = BreadCrumb()
     title = Text('//div[@id="main-content"]//h1')
+    table = Table("//div[@id='list_grid']//table")
     properties = SummaryTable(title='Properties')
     relationships = SummaryTable(title='Relationships')
     smart_management = SummaryTable(title='Smart Management')
@@ -105,6 +104,7 @@ class KeyPairDetailsView(KeyPairView):
     toolbar = View.nested(KeyPairDetailsToolbar)
     sidebar = View.nested(KeyPairDetailsAccordion)
     entities = View.nested(KeyPairDetailsEntities)
+    paginator = View.nested(PaginationPane)
 
 
 class KeyPairAddView(KeyPairView):
@@ -119,7 +119,49 @@ class KeyPairAddView(KeyPairView):
     form = View.nested(KeyPairAddForm)
 
 
-class KeyPair(Taggable, SummaryMixin, Navigatable):
+class KeyPairCollection(Navigatable):
+    """ Collection object for the :py:class: `cfme.cloud.KeyPair`. """
+
+    def instantiate(self, name, provider, public_key=None):
+        return KeyPair(name, provider, public_key=public_key or "", collection=self)
+
+    def create(self, name, provider, public_key=None, cancel=False):
+        """Create new keyPair.
+
+        Args:
+            name (str): name of the KeyPair
+            public_key (str): RSA Key if present
+            provider (str): Cloud Provider
+            cancel (boolean): Cancel Keypair creation
+        """
+
+        view = navigate_to(self, 'Add')
+        changed = view.form.fill({'name': name,
+                                  'public_key': public_key,
+                                  'provider': provider.name
+                                  })
+        if cancel and not changed:
+            view.form.cancel.click()
+            flash_message = 'Add of new Key Pair was cancelled by the user'
+        else:
+            view.form.add.click()
+            flash_message = VersionPick({
+                Version.lowest(): 'Creating Key Pair {}'.format(name),
+                '5.8': 'Key Pair "{}" created'.format(name)}).pick(self.appliance.version)
+
+        # add/cancel should redirect, new view
+        view = self.create_view(KeyPairAllView)
+        # TODO BZ 1444520 causing ridiculous redirection times after submitting the form
+        wait_for(lambda: view.is_displayed, num_sec=120, delay=3,
+                 fail_func=view.flush_widget_cache, handle_exception=True)
+        view = self.create_view(KeyPairAllView)
+        assert view.is_displayed
+        view.entities.flash.assert_success_message(flash_message)
+        keypair = self.instantiate(name, provider, public_key=public_key)
+        return keypair
+
+
+class KeyPair(Navigatable):
     """ Automate Model page of KeyPairs
 
     Args:
@@ -127,16 +169,17 @@ class KeyPair(Taggable, SummaryMixin, Navigatable):
     """
     _param_name = "KeyPair"
 
-    def __init__(self, name, provider, public_key=None, appliance=None):
-        Navigatable.__init__(self, appliance=appliance)
+    def __init__(self, name, provider, public_key=None, appliance=None, collection=None):
+        self.collection = collection or KeyPairCollection(appliance=appliance)
+        Navigatable.__init__(self, appliance=self.collection.appliance)
         self.name = name
         self.provider = provider
-        self.public_key = public_key
+        self.public_key = public_key or ""
 
     def delete(self, cancel=False, wait=False):
         view = navigate_to(self, 'Details')
         view.toolbar.configuration.item_select('Remove this Key Pair',
-                                               handle_alert=not cancel)
+                                               handle_alert=(not cancel))
         # cancel doesn't redirect, confirmation does
         view.flush_widget_cache()
         if cancel:
@@ -151,53 +194,22 @@ class KeyPair(Taggable, SummaryMixin, Navigatable):
             view.entities.flash.assert_success_message('Delete initiated for 1 Key Pair')
 
         if wait:
-            self.wait_for_delete()
+            def refresh():
+                self.provider.refresh_provider_relationships()
+                view.browser.refresh()
+                view.flush_widget_cache()
 
-    def wait_for_delete(self):
-        def refresh():
-            self.provider.refresh_provider_relationships()
-            view.browser.selenium.refresh()
-            view.flush_widget_cache()
-        view = navigate_to(self, 'All')
-        # look for the row, call is_displayed on it to get boolean
-        # find_row_on_pages is going to raise NoSuchElement when the row disappears
-        try:
-            wait_for(lambda: view.paginator.find_row_on_pages(view.entities.table,
-                                                              name=self.name).is_displayed,
-                     fail_condition=True, message="Wait keypairs to disappear", num_sec=500,
-                     delay=5, fail_func=refresh)
-        except NoSuchElementException:
-            # The row doesn't exist anymore, success
-            pass
-
-    def create(self, cancel=False):
-        """Create new keypair"""
-        def refresh():
-            self.provider.refresh_provider_relationships()
-            view.browser.selenium.refresh()
-            view.flush_widget_cache()
-
-        view = navigate_to(self, 'Add')
-        view.form.fill({'name': self.name,
-                        'public_key': self.public_key,
-                        'provider': self.provider.name})
-        if cancel:
-            view.form.cancel.click()
-            flash_message = 'Add of new Key Pair was cancelled by the user'
-        else:
-            view.form.add.click()
-            flash_message = VersionPick({
-                Version.lowest(): 'Creating Key Pair {}'.format(self.name),
-                '5.8': 'Key Pair "{}" created'.format(self.name)}).pick(self.appliance.version)
-
-        # add/cancel should redirect, new view
-        view = self.create_view(KeyPairAllView)
-        # TODO BZ 1444520 causing ridiculous redirection times after submitting the form
-        wait_for(lambda: view.is_displayed, fail_condition=False, num_sec=120, delay=3,
-                 fail_func=lambda: view.flush_widget_cache(), handle_exception=True)
-        view = self.create_view(KeyPairAllView)
-        view.entities.flash.assert_no_error()
-        view.entities.flash.assert_success_message(flash_message)
+                # look for the row, call is_displayed on it to get boolean
+                # find_row_on_pages is going to raise NoSuchElement when the row disappears
+            wait_for(
+                lambda: not self.exists,
+                message="Wait keypairs to disappear",
+                fail_condition=False,
+                num_sec=300,
+                timeout=1000,
+                delay=20,
+                fail_func=refresh
+            )
 
     def add_tag(self, tag, **kwargs):
         """Tags the Keypair by given tag"""
@@ -211,12 +223,16 @@ class KeyPair(Taggable, SummaryMixin, Navigatable):
 
     @property
     def exists(self):
-        view = navigate_to(self, 'Details')
-        return view.is_displayed
+        try:
+            navigate_to(self, 'Details')
+        except KeyPairNotFound:
+            return False
+        else:
+            return True
 
 
-@navigator.register(KeyPair, 'All')
-class All(CFMENavigateStep):
+@navigator.register(KeyPairCollection, 'All')
+class CloudKeyPairs(CFMENavigateStep):
     VIEW = KeyPairAllView
     prerequisite = NavigateToAttribute('appliance.server', 'LoggedIn')
 
@@ -227,7 +243,7 @@ class All(CFMENavigateStep):
 @navigator.register(KeyPair, 'Details')
 class Details(CFMENavigateStep):
     VIEW = KeyPairDetailsView
-    prerequisite = NavigateToSibling('All')
+    prerequisite = NavigateToAttribute('collection', 'All')
 
     def step(self, *args, **kwargs):
         self.prerequisite_view.toolbar.view_selector.select('List View')
@@ -242,10 +258,10 @@ class Details(CFMENavigateStep):
         row.click()
 
 
-@navigator.register(KeyPair, 'Add')
+@navigator.register(KeyPairCollection, 'Add')
 class Add(CFMENavigateStep):
     VIEW = KeyPairAddView
-    prerequisite = NavigateToSibling('All')
+    prerequisite = NavigateToSibling("All")
 
     def step(self, *args, **kwargs):
         """Raises DropdownItemDisabled from widgetastic_patternfly if no RHOS provider present"""
--- a/cfme/tests/cloud/test_keypairs.py
+++ b/cfme/tests/cloud/test_keypairs.py
@@ -2,13 +2,12 @@ import fauxfactory
 import pytest
 from Crypto.PublicKey import RSA
 
-from cfme.cloud.keypairs import KeyPair
+from cfme.cloud.keypairs import KeyPairCollection
 from cfme.cloud.provider.openstack import OpenStackProvider
 from cfme.exceptions import KeyPairNotFound
 from cfme.web_ui import mixins
 from utils import testgen
 from utils.blockers import BZ
-from utils.appliance.implementations.ui import navigate_to
 from utils.wait import TimedOutError
 
 pytestmark = [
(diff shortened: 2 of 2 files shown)


dajoRH:
Lint report for commit 2fc4b4fc25b60345042ed1dc3909659d0aaf3c6d:

`cfme/tests/cloud/test_keypairs.py`:
- :red_circle: Line 86:5: **F841** *local variable 'keypair' is assigned to but never used*

**No commit flaws detected**.

*CFME QE Bot*

cfme/cloud/keypairs.py
```diff
@@ -119,19 +121,58 @@ def is_displayed(self):
...
+        """Create new keyPair.
+
+        Args:
+            name (str): name of the KeyPair
+            Public Key (str): RSA Key if present
+            Provider (str): Cloud Provider
+        """
+
+        view = navigate_to(self, 'Add')
+        view.form.fill({'name': name,
```
quarckster: **REQUIRED**
```python
changed = view.form.fill(...)
if not changed:
    cancel = True
```

cfme/cloud/keypairs.py
```diff
@@ -119,19 +121,58 @@ def is_displayed(self):
...
+        else:
+            view.form.add.click()
+            flash_message = VersionPick({
+                Version.lowest(): 'Creating Key Pair {}'.format(name),
+                '5.8': 'Key Pair "{}" created'.format(name)}).pick(self.appliance.version)
+
+        # add/cancel should redirect, new view
+        view = self.create_view(KeyPairAllView)
+        # TODO BZ 1444520 causing ridiculous redirection times after submitting the form
+        wait_for(lambda: view.is_displayed, fail_condition=False, num_sec=120, delay=3,
```
quarckster: `fail_condition=False` is not needed

cfme/cloud/keypairs.py
```diff
@@ -5,9 +5,9 @@
 from widgetastic.exceptions import NoSuchElementException
 from widgetastic_patternfly import Dropdown, Button, FlashMessages
 
+from cfme.base import Server
 from cfme.base.ui import BaseLoggedInPage
-from cfme.common import SummaryMixin, Taggable
-from cfme.exceptions import KeyPairNotFound
+# from cfme.exceptions import KeyPairNotFound
```
mshriver: remove if not needed, don't comment out.

cfme/cloud/keypairs.py
```diff
@@ -119,19 +121,58 @@ def is_displayed(self):
...
+                '5.8': 'Key Pair "{}" created'.format(name)}).pick(self.appliance.version)
+
+        # add/cancel should redirect, new view
+        view = self.create_view(KeyPairAllView)
+        # TODO BZ 1444520 causing ridiculous redirection times after submitting the form
+        wait_for(lambda: view.is_displayed, fail_condition=False, num_sec=120, delay=3,
+                 fail_func=lambda: view.flush_widget_cache(), handle_exception=True)
+        view = self.create_view(KeyPairAllView)
+        assert view.is_displayed
+        view.entities.flash.assert_no_error()
```
quarckster: **REQUIRED**
Unnecessary, you already have below  `view.entities.flash.assert_success_message(flash_message)`

cfme/cloud/keypairs.py
```diff
@@ -158,46 +199,18 @@ def refresh():
...
-        view = navigate_to(self, 'All')
+        view = navigate_to(self, 'Details')
         # look for the row, call is_displayed on it to get boolean
         # find_row_on_pages is going to raise NoSuchElement when the row disappears
         try:
             wait_for(lambda: view.paginator.find_row_on_pages(view.entities.table,
                                                               name=self.name).is_displayed,
-                     fail_condition=True, message="Wait keypairs to disappear", num_sec=500,
+                     fail_condition=True, message="Wait keypairs to disappear", num_sec=600,
                      delay=5, fail_func=refresh)
```
quarckster:
```python
wait_for(
    lambda: not view.paginator.find_row_on_pages(
        view.entities.table, name=self.name).is_displayed,
    message="Wait keypairs to disappear",
    num_sec=600,
    delay=5,
    fail_func=refresh
)
```

cfme/cloud/keypairs.py
sbulage: Done

sbulage: Done.

cfme/cloud/keypairs.py
```diff
@@ -119,19 +121,58 @@ def is_displayed(self):
...
-class KeyPair(Taggable, SummaryMixin, Navigatable):
+class KeyPairCollection(Navigatable):
+    """ Collection object for the :py:class: `cfme.cloud.KeyPair`. """
+
+    def create(self, name, provider, public_key=None, cancel=False):
+        """Create new keyPair.
+
+        Args:
+            name (str): name of the KeyPair
+            Public Key (str): RSA Key if present
```
sbulage: Ok :)

cfme/cloud/keypairs.py
sbulage: Removed `fail_condition=False`.

cfme/cloud/keypairs.py
sbulage: Kept only `view.entities.flash.assert_success_message(flash_message)`

cfme/cloud/keypairs.py
sbulage: Updated.

cfme/cloud/keypairs.py
sbulage: @quarckster can you please explain me in detail...?

sbulage: Ok I updated it.

dajoRH:
Lint report for commit 1de4812c4d70edecf46f5989a8d8cf915b29f85f:

`cfme/cloud/keypairs.py`:
- :bangbang: Line 136:35: **E127** *continuation line over-indented for visual indent*

**No commit flaws detected**.

*CFME QE Bot*

cfme/cloud/keypairs.py
```diff
@@ -53,6 +51,7 @@ class KeyPairEntities(View):
 class KeyPairDetailsEntities(View):
     breadcrumb = BreadCrumb()
     title = Text('//div[@id="main-content"]//h1')
+    table = Table("//div[@id='list_grid']//table")
```
mshriver: It's about time to convert this to use the new Entities conditional view that @izapolsk implemented, which will get you table + tiles + quadicons in a conditional view.

I don't think it needs to be done in the scope of the PR, but @sbulage could you write a card/issue so we don't miss it please?

cfme/cloud/keypairs.py
```diff
@@ -119,19 +119,54 @@ def is_displayed(self):
...
+            view.form.add.click()
+            flash_message = VersionPick({
+                Version.lowest(): 'Creating Key Pair {}'.format(name),
+                '5.8': 'Key Pair "{}" created'.format(name)}).pick(self.appliance.version)
+
+        # add/cancel should redirect, new view
+        view = self.create_view(KeyPairAllView)
+        # TODO BZ 1444520 causing ridiculous redirection times after submitting the form
+        wait_for(lambda: view.is_displayed, num_sec=120, delay=3,
+                 fail_func=lambda: view.flush_widget_cache(), handle_exception=True)
```
mshriver: Don't need a lambda for `fail_func` since you're passing a bound method, you can pass `view.flush_widget_cache`

cfme/cloud/keypairs.py
sbulage: ok I will create card for it and will assign to you.

cfme/cloud/keypairs.py
sbulage: ok I will update it accordingly.

cfme/cloud/keypairs.py
```diff
@@ -158,46 +193,22 @@ def refresh():
             self.provider.refresh_provider_relationships()
             view.browser.selenium.refresh()
             view.flush_widget_cache()
-        view = navigate_to(self, 'All')
+        view = navigate_to(self, 'Details')
```
mshriver: **Optional**

Doesn't really have to do with Collection, but consider removing this convenience method and making an enhancement to KeyPair.exists to catch navigation exception and return false.

Then callers can directly call `wait_for(lambda: keypair.exists)`

cfme/cloud/keypairs.py
```diff
@@ -215,19 +226,19 @@ def exists(self):
         return view.is_displayed
 
 
-@navigator.register(KeyPair, 'All')
-class All(CFMENavigateStep):
+@navigator.register(Server)
+class CloudKeyPair(CFMENavigateStep):
```
mshriver: **Required**
This class name implies a single keypair, I'd like to see something like `CloudKeyPairs` or `KeyPairsCollection`, something along those lines that makes it more clear you're navigating to 'All'.

Additionally, consider keeping 'All' in the registration as the name, so that we can keep consistent navigation destinations, and I can say import KeyPair, Volume, and Instance, and navigate to 'All' against any of the classes.

cfme/cloud/keypairs.py
```diff
@@ -215,19 +226,19 @@ def exists(self):
...
     def step(self, *args, **kwargs):
         self.prerequisite_view.navigation.select('Compute', 'Clouds', 'Key Pairs')
 
 
-@navigator.register(KeyPair, 'Details')
+@navigator.register(KeyPair)
 class Details(CFMENavigateStep):
     VIEW = KeyPairDetailsView
-    prerequisite = NavigateToSibling('All')
+    prerequisite = NavigateToAttribute("appliance.server", "CloudKeyPair")
```
mshriver: Update to match name change from required comment above.

cfme/cloud/keypairs.py
```diff
@@ -237,15 +248,15 @@ def step(self, *args, **kwargs):
                 name=self.obj.name
             )
         except NoSuchElementException:
-            raise KeyPairNotFound
-
-        row.click()
+            pass
```
mshriver: Why are we burying this exception now?

cfme/tests/cloud/test_keypairs.py
```diff
@@ -46,9 +44,6 @@ def test_keypair_crud(openstack_provider):
         openstack_provider.mgmt.api.keypairs.delete(keypair.name)
         pytest.fail('Timed out deleting keypair')
 
-    with pytest.raises(KeyPairNotFound):
```
mshriver: This test assertion needs to be replaced somehow if the exception is no longer raised. I think the exception should still be raised and this block can stay as-is if so.

cfme/tests/cloud/test_keypairs.py
```diff
@@ -79,9 +72,6 @@ def test_keypair_crud_with_key(openstack_provider):
...
```
mshriver: Again this test assertion needs to be replaced somehow, ideally by keeping the exception on navigation failure.

cfme/tests/cloud/test_keypairs.py
```diff
@@ -136,6 +124,3 @@ def test_keypair_add_and_remove_tag(openstack_provider):
     except TimedOutError:
         openstack_provider.mgmt.api.keypairs.delete(keypair.name)
         pytest.fail('Timed out deleting keypair')
-
-    with pytest.raises(KeyPairNotFound):
```
mshriver: Again, test assertion needs to be replaced not removed.

cfme/tests/cloud/test_keypairs.py
quarckster: Maybe `assert not keypair.exists` will be enough.

dajoRH:
Lint report for commit aefd0d660b16f663d4cc7d2cdfc7d669ad197160:

`cfme/tests/cloud/test_keypairs.py`:
- :red_circle: Line 9:1: **F401** *'utils.appliance.implementations.ui.navigate_to' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/cloud/test_keypairs.py
sbulage: @quarckster @mshriver It is throwing error DID NOT RAISE KEYPAIR not found.

cfme/cloud/keypairs.py
```diff
@@ -119,19 +120,54 @@ def is_displayed(self):
...
+class KeyPairCollection(Navigatable):
+    """ Collection object for the :py:class: `cfme.cloud.KeyPair`. """
+
+    def create(self, name, provider, public_key=None, cancel=False):
+        """Create new keyPair.
+
+        Args:
+            name (str): name of the KeyPair
+            public_key (str): RSA Key if present
+            provider (str): Cloud Provider
```
sbulage: ok adding` Args: cancel(Boolean)`

cfme/tests/cloud/test_keypairs.py
sbulage: Added back with `assert`.

cfme/tests/cloud/test_keypairs.py
sbulage: Added it.

cfme/tests/cloud/test_keypairs.py
sbulage: Fixed it.

dajoRH:
Lint report for commit 6db837bbc0944377bcc1fb6f35653562b6f33aee:

`cfme/cloud/keypairs.py`:
- :bangbang: Line 126:28: **E231** *missing whitespace after ','*

**No commit flaws detected**.

*CFME QE Bot*

cfme/cloud/keypairs.py
```diff
@@ -119,24 +119,67 @@ def is_displayed(self):
     form = View.nested(KeyPairAddForm)
 
 
-class KeyPair(Taggable, SummaryMixin, Navigatable):
+class KeyPairCollection(Navigatable):
+    """ Collection object for the :py:class: `cfme.cloud.KeyPair`. """
+
+    def instantiate(self, name, provider, public_key):
```
quarckster: `public_key=None`

cfme/cloud/keypairs.py
```diff
@@ -119,24 +119,67 @@ def is_displayed(self):
...
+            view.form.add.click()
+            flash_message = VersionPick({
+                Version.lowest(): 'Creating Key Pair {}'.format(name),
+                '5.8': 'Key Pair "{}" created'.format(name)}).pick(self.appliance.version)
+
+        # add/cancel should redirect, new view
+        view = self.create_view(KeyPairAllView)
+        # TODO BZ 1444520 causing ridiculous redirection times after submitting the form
+        wait_for(lambda: view.is_displayed, num_sec=120, delay=3,
+                 fail_func=view.flush_widget_cache(), handle_exception=True)
```
quarckster: **REQUIRED**
```python
wait_for(lambda: view.is_displayed, num_sec=120, delay=3, fail_func=view.flush_widget_cache, handle_exception=True)
```

cfme/cloud/keypairs.py
```diff
@@ -119,24 +119,67 @@ def is_displayed(self):
...
+        view = self.create_view(KeyPairAllView)
+        assert view.is_displayed
+        view.entities.flash.assert_success_message(flash_message)
+        keypair = self.instantiate(name, provider, public_key=public_key or "")
```
quarckster: **REQUIRED**
```python
keypair = self.instantiate(name, provider, public_key=public_key)
```

cfme/cloud/keypairs.py
```diff
@@ -154,51 +197,22 @@ def delete(self, cancel=False, wait=False):
...
+        self.provider.refresh_provider_relationships()
         # look for the row, call is_displayed on it to get boolean
         # find_row_on_pages is going to raise NoSuchElement when the row disappears
         try:
-            wait_for(lambda: view.paginator.find_row_on_pages(view.entities.table,
-                                                              name=self.name).is_displayed,
-                     fail_condition=True, message="Wait keypairs to disappear", num_sec=500,
-                     delay=5, fail_func=refresh)
+            wait_for(
+                lambda: self.exists,
```
quarckster: **REQUIRED**
```python
lambda: not self.exists,
```
and remove `fail_condition=True,`

cfme/tests/cloud/test_keypairs.py
```diff
@@ -92,11 +87,12 @@ def test_keypair_create_cancel(openstack_provider):
...
-    keypair.create(cancel=True)
-
-    with pytest.raises(KeyPairNotFound):
-        navigate_to(keypair, 'Details')
+    keypairs = KeyPairCollection()
+    keypair = keypairs.create(name="", provider=openstack_provider, cancel=True)
+    try:
+        assert not keypair.exists
+    except KeyPairNotFound:
+        pass
```
quarckster: **REQUIRED**
`KeyPairNotFound` will never caught, just leave `assert not keypair.exists`

cfme/cloud/keypairs.py
sbulage: Done.

cfme/cloud/keypairs.py
sbulage: @quarckster what changes need to be done here ..?

cfme/cloud/keypairs.py
sbulage: Done.

cfme/tests/cloud/test_keypairs.py
sbulage: Remove `try-except` blcok

cfme/cloud/keypairs.py
quarckster: @sbulage you put `view.flush_widget_cache()` instead of `view.flush_widget_cache`

sbulage: ok

cfme/cloud/keypairs.py
sbulage: Yeah removed `fail_condition=True,` and added `lambda: not self.exists,`.

quarckster:
@sbulage please add some tests for PRT.

mshriver:
PRT run 15269 had passing results for cfme/tests/cloud/test_keypairs.py

cfme/cloud/keypairs.py
```diff
@@ -119,24 +119,67 @@ def is_displayed(self):
...
         self.name = name
         self.provider = provider
-        self.public_key = public_key
+        self.public_key = public_key or ""
 
     def delete(self, cancel=False, wait=False):
         view = navigate_to(self, 'Details')
         view.toolbar.configuration.item_select('Remove this Key Pair',
-                                               handle_alert=not cancel)
+                                               handle_alert=True)
```
mshriver: **Required**
`cancel` arg is not actually used here to handle the alert, meaning if cancel is passed as True, this method will still delete the keypair.

cfme/cloud/keypairs.py
```diff
@@ -154,51 +197,21 @@ def delete(self, cancel=False, wait=False):
...
-        view = navigate_to(self, 'All')
+        self.provider.refresh_provider_relationships()
         # look for the row, call is_displayed on it to get boolean
         # find_row_on_pages is going to raise NoSuchElement when the row disappears
         try:
-            wait_for(lambda: view.paginator.find_row_on_pages(view.entities.table,
-                                                              name=self.name).is_displayed,
-                     fail_condition=True, message="Wait keypairs to disappear", num_sec=500,
-                     delay=5, fail_func=refresh)
+            wait_for(
```
mshriver: This is causing problems for me in testing.

The 'refresh' method exists because a browser refresh or a provider refresh alone isn't sufficient. I'm seeing fast timeouts when calling delete that suggest that the fail_condition check isn't working like you intend.

Also you're passing num_sec and timeout, which are the same.

mshriver: Really this entire `wait_for_delete()` method should be removed anyway.

cfme/cloud/keypairs.py
sbulage: @mshriver do I need to add `not cancel` back to code ..?

cfme/cloud/keypairs.py
sbulage: @mshriver if I remove `wait_for_delete()` method then `self.exists` will return error after it tries to navigate to keypair details page and assert will fail.

cfme/cloud/keypairs.py
mshriver: Looks like it, the differences for `cancel` look like they're still handled elsewhere.

Where I had to implement this, I wrote it with () so it was a bit more clear on reading:
`handle_alert=(not cancel)`

sbulage: I will update like this `handle_alert=(not cancel)`.

cfme/cloud/keypairs.py
mshriver: The `exists` method for any object should return Boolean, not raise an error, unless something unexpected happens. For example we'll know if we try to navigate and it doesn't exist we'll get `CandidateNotFound` when trying to navigate using accordion - so the `exists` method catches CandidateNotFound and returns false. If some _other_ exception happens, then it won't be caught.

sbulage: yes that is true, but after initiate delete keypair took some time to disappear from UI.
`wait_for_delete()` is required there.

PR #5109, opened 2017-08-04
[1LP][RFR] Correct recognition of entities in list view 

izapolsk:
When BaseEntitiesView is switched to list view, it cannot correctly recognize entities in it due to missing entity_names method.
This PR fixes that issue.


hhovsepy:
Thanks for the fix @izapolsk looking forward to see it merged

PR #5110, opened 2017-08-06
[WIPTEST] Skip unsupported rest reset vm test for RHV provider

istein1:
The REST reset VM test is running for the RHV provider, though the power reset option is not supported there.
Thus, need to skip it for now.


jkrocil:
The change is good, checked PRT, failures are unrelated. :+1:

PR #5111, opened 2017-08-07
[1LP][RFR] RHV Verify TLS test fix uncollection condition

slaykovsky:
Fixing test uncollectif condition because it is currently incorrect because provider.data does not have 'verify_tls' key. It should be provider.endpoints['default'].verify_tls.

Thanks.

{{pytest: -v --long-running cfme/tests/infrastructure/test_providers.py -k test_provider_rhv_create_delete_tls --use-provider rhv41}}

### DIFF
--- a/cfme/tests/infrastructure/test_providers.py
+++ b/cfme/tests/infrastructure/test_providers.py
@@ -228,7 +228,7 @@ def test_provider_crud(provider):
 @test_requirements.provider_discovery
 @pytest.mark.parametrize('verify_tls', [False, True], ids=['no_tls', 'tls'])
 @pytest.mark.uncollectif(lambda provider:
-    not provider.one_of(RHEVMProvider) or not provider.endpoints.get('default').verify_tls)
+    not (provider.one_of(RHEVMProvider) and provider.endpoints.get('default').verify_tls))
 def test_provider_rhv_create_delete_tls(request, provider, verify_tls):
     """Tests RHV provider creation with and without TLS encryption
 
--- a/cfme/tests/infrastructure/test_providers.py
+++ b/cfme/tests/infrastructure/test_providers.py
@@ -228,7 +228,8 @@ def test_provider_crud(provider):
 @test_requirements.provider_discovery
 @pytest.mark.parametrize('verify_tls', [False, True], ids=['no_tls', 'tls'])
 @pytest.mark.uncollectif(lambda provider:
-    not (provider.one_of(RHEVMProvider) and provider.endpoints.get('default').verify_tls))
+    not (provider.one_of(RHEVMProvider) and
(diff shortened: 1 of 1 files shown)


slaykovsky:
It skipped at CI because env with rhv41 provider doesn't have verify_tls value as True. To have it enabled it's necessary to get cacerts from specific engine (http://engine/ovirt-engine/services/pki-resource?resource=ca-certificate&FORMAT=X509-PEM-CA) and have it either in yaml (runtime patching) or update provider.endpoints object.

On internal env with rhv provider which has verify_tls enabled and also has cacert in yaml it passes.

Thanks.

cfme/tests/infrastructure/test_providers.py
```diff
@@ -228,7 +228,7 @@ def test_provider_crud(provider):
 @test_requirements.provider_discovery
 @pytest.mark.parametrize('verify_tls', [False, True], ids=['no_tls', 'tls'])
 @pytest.mark.uncollectif(lambda provider:
-    not provider.one_of(RHEVMProvider) and not provider.data.get('verify_tls'))
+    not (provider.one_of(RHEVMProvider) and provider.endpoints.get('default').verify_tls))
```
jkrocil: @slaykovsky Alexey, what will happen if you run collection with a provider that has no `verify_tls` under default endpoint? :wink: 

`AttributeError`.

I would strongly suggest running any rhev-only tests against some other provider as well. Maybe not locally, but certainly so on PRT. If you invoke `py.test` without `--use-provider`, it will run the `default` group which contains some VMWare providers. That will be sufficient.

slaykovsky: Yeah, never thought about others providers :) Will fix, thanks!

cfme/tests/infrastructure/test_providers.py
```diff
@@ -228,7 +228,8 @@ def test_provider_crud(provider):
 @test_requirements.provider_discovery
 @pytest.mark.parametrize('verify_tls', [False, True], ids=['no_tls', 'tls'])
 @pytest.mark.uncollectif(lambda provider:
-    not provider.one_of(RHEVMProvider) and not provider.data.get('verify_tls'))
+    not (provider.one_of(RHEVMProvider) and
+         provider.endpoints.get('default').__dict__.get('verify_tls')))
```
jkrocil: `provider.endpoints.get('default', {}).get('verify_tls')`

jkrocil: I should have mentioned that before. Sorry!

slaykovsky: But default endpoints exist anywhere... No problem, will change.

jkrocil: Ignore me, I was thinking about something else there! Merged.

PR #5112, opened 2017-08-07
[1LP][RFR] Fixing test_select_attributes

mkoura:
The BZ 1437201 affecting the tests will be fixed in 5.9. Skipping in
older versions.
Cleanups of unsupported versions.
Skipping all test_select_attributes tests because of https://github.com/ManageIQ/manageiq/issues/15754

{{pytest: -v -k "test_select_attributes or test_query_simple_collections" cfme/tests/test_rest.py}}

### DIFF
--- a/cfme/tests/test_rest.py
+++ b/cfme/tests/test_rest.py
@@ -70,6 +70,11 @@ COLLECTIONS_ADDED_IN_58 = {
 }
 
 
+COLLECTIONS_REMOVED_IN_59 = {
+    "arbitration_settings", "arbitration_profiles", "virtual_templates", "arbitration_rules",
+}
+
+
 COLLECTIONS_ALL = {
     "actions", "alert_definitions", "alerts", "arbitration_profiles",
     "arbitration_rules", "arbitration_settings", "authentications", "automate",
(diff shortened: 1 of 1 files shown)


cfme/tests/test_rest.py
```diff
@@ -130,17 +129,18 @@ def test_query_simple_collections(appliance, collection_name):
...
 )
+@pytest.mark.meta(blockers=['GH#ManageIQ/manageiq:15754'])
 def test_select_attributes(appliance, collection_name):
     """Tests that it's possible to limit returned attributes.
 
     Metadata:
         test_flag: rest
     """
+    if collection_name in COLLECTIONS_BUGGY_ATTRS and current_version() < '5.9':
+        pytest.skip("Affected by BZ 1437201, cannot test.")
```
quarckster: As I know skipped tests affect on successful pass score in Jenkins, maybe it better to uncollect them here?

mkoura: I don't know about Jenkins, but this is more or less the same as if the tests were marked with blocker (which I can't do because it would mark all test variants).
These tests are in Polarion and will remain unexecuted there if they are just uncollected. Then I would need to manually check why they didn't run and mark them skipped.
When the tests are skipped with BZ number, they will be automatically marked as skipped in Polarion (which saves my time :)

PR #4872, opened 2017-06-16
[1LP][RFR] cloud.instance widgetastic conversion (Part 2)

mshriver:
Convert cloud.instances model to WT views and widgets.

Implemented common views and widgets for instances and VM's in cfme.common.vm_views. These are cases where the pages are exactly the same. Any pages with slight differences should be implemented in their respective modules.

Moved ProvisioningForm class to cfme.common.vm_views.ProvisionView. ProvisionView.form nested view class implements `before_fill()` to select the provisioning image from the table, the first half of the UI form. This gives the syntax `view.form.fill({<tab keyed fill dict>})` to complete the entire provisioning form fill process, through both pages.

## PRT Results
Skipping RESTAPI tests, not relevant here.

### PRT run 14802
* 58z: 70+% overall, failures in test_provisioning due to VM's not provisioning, no page model/widget/view exceptions.
* 57z: 85%+ overall, same provisioning failures as 58z.

{{ pytest: -v cfme/tests/cloud/test_instance_power_control.py cfme/tests/cloud/test_provisioning.py --use-provider gce_central --use-provider rhos7-ga --use-provider azure --use-provider ec2weset --long-running -k "not TestInstanceRESTAPI"}}

PR #5113, opened 2017-08-07
[1LP][RFR] Wigdetastic in cfme.infrastructure.virtual_machines_Pt_1

lkhomenk:
Purpose or Intent
=================
Part 1 - Navigation part updates. splitting PR into few parts
Shriver: ~~forked from #4872~~

{{ pytest: -v cfme/tests/infrastructure/test_vm_power_control.py cfme/tests/infrastructure/test_provisioning.py --use-provider vsphere65-nested --use-provider scvmm --use-provider vsphere6-nested  --long-running -k "not RESTAPI"}}

### DIFF
--- a/cfme/common/vm_views.py
+++ b/cfme/common/vm_views.py
@@ -64,15 +64,15 @@ def InstanceEntity():  # noqa
     })
 
 
-class NonEditableTable(Table):
+class SelectTable(Table):
     """Wigdet for non-editable table. used for selecting value"""
     def fill(self, values):
         """Clicks on item - fill by selecting required value"""
         value = values.get('name', '<None>')
         changed = False
-        if value != self.read():
+        if value != self.currently_selected:
             changed = True
-        self.row(name=value).click()
+            self.row(name=value).click()
         return changed
 
     @property
@@ -93,34 +93,7 @@ class NonEditableTable(Table):
 
     def read_content(self):
         """This is a default Table.read() method for those who will need table content"""
-        rows = list(self)
-        # Cut the unwanted rows if necessary
-        if self.rows_ignore_top is not None:
-            rows = rows[self.rows_ignore_top:]
-        if self.rows_ignore_bottom is not None and self.rows_ignore_bottom > 0:
-            rows = rows[:-self.rows_ignore_bottom]
-        if self.assoc_column_position is None:
-            return [row.read() for row in rows]
-        else:
-            result = {}
-            for row in rows:
-                row_read = row.read()
-                try:
-                    key = row_read.pop(self.header_index_mapping[self.assoc_column_position])
-                except KeyError:
-                    try:
-                        key = row_read.pop(self.assoc_column_position)
-                    except KeyError:
-                        try:
-                            key = row_read.pop(self.assoc_column)
-                        except KeyError:
-                            raise ValueError(
-                                'The assoc_column={!r} could not be retrieved'.format(
-                                    self.assoc_column))
-                if key in result:
-                    raise ValueError('Duplicate value for {}={!r}'.format(key, result[key]))
-                result[key] = row_read
-            return result
+        return super(SelectTable, self).read()
 
 
 class VMToolbar(View):
@@ -192,12 +165,12 @@ class BasicProvisionFormView(View):
         vm_description = Input(name='service__vm_description')
         vm_filter = BootstrapSelect('service__vm_filter')
         num_vms = BootstrapSelect('service__number_of_vms')
-        catalog_name = NonEditableTable('//div[@id="prov_vm_div"]/table')
+        catalog_name = SelectTable('//div[@id="prov_vm_div"]/table')
         provision_type = BootstrapSelect('service__provision_type')
         linked_clone = Input(name='service__linked_clone')
         pxe_server = BootstrapSelect('service__pxe_server_id')
-        pxe_image = NonEditableTable('//div[@id="prov_pxe_img_div"]/table')
-        iso_file = NonEditableTable('//div[@id="prov_iso_img_div"]/table')
+        pxe_image = SelectTable('//div[@id="prov_pxe_img_div"]/table')
+        iso_file = SelectTable('//div[@id="prov_iso_img_div"]/table')
 
     @View.nested
     class environment(Tab):  # noqa
@@ -217,10 +190,10 @@ class BasicProvisionFormView(View):
         resource_pool = BootstrapSelect('environment__placement_rp_name')
         folder = BootstrapSelect('environment__placement_folder_name')
         host_filter = BootstrapSelect('environment__host_filter')
-        host_name = NonEditableTable('//div[@id="prov_host_div"]/table')
+        host_name = SelectTable('//div[@id="prov_host_div"]/table')
         datastore_create = Input('environment__new_datastore_create')
         datastore_filter = BootstrapSelect('environment__ds_filter')
-        datastore_name = NonEditableTable('//div[@id="prov_ds_div"]/table')
+        datastore_name = SelectTable('//div[@id="prov_ds_div"]/table')
 
     @View.nested
     class hardware(Tab):  # noqa
@@ -266,7 +239,7 @@ class BasicProvisionFormView(View):
             ip_address = Input(name='customize__ip_addr')
             subnet_mask = Input(name='customize__subnet_mask')
             gateway = Input(name='customize__gateway')
-            custom_template = NonEditableTable('//div[@id="prov_template_div"]/table')
+            custom_template = SelectTable('//div[@id="prov_template_div"]/table')
             hostname = Input(name='customize__hostname')
 
     @View.nested
--- a/cfme/common/vm_views.py
+++ b/cfme/common/vm_views.py
@@ -64,6 +64,65 @@ def InstanceEntity():  # noqa
     })
 
 
+class NonEditableTable(Table):
+    """Wigdet for non-editable table. used for selecting value"""
+    def fill(self, values):
+        """Clicks on item - fill by selecting required value"""
+        value = values.get('name', '<None>')
+        changed = False
+        if value != self.read():
+            changed = True
+        self.row(name=value).click()
+        return changed
+
+    @property
+    def currently_selected(self):
+        """Return Name of the selected row"""
+        selected = self.browser.elements(
+            ".//tr[@class='selected']/td[1]",
+            parent=self)
+        result = map(self.browser.text, selected)
+        if len(result) == 0:
+            self.logger.info('Nothing is currently selected')
+            return None
+        else:
+            return result[0]
+
+    def read(self):
+        return self.currently_selected
+
+    def read_content(self):
+        """This is a default Table.read() method for those who will need table content"""
+        rows = list(self)
+        # Cut the unwanted rows if necessary
+        if self.rows_ignore_top is not None:
+            rows = rows[self.rows_ignore_top:]
+        if self.rows_ignore_bottom is not None and self.rows_ignore_bottom > 0:
+            rows = rows[:-self.rows_ignore_bottom]
+        if self.assoc_column_position is None:
+            return [row.read() for row in rows]
+        else:
+            result = {}
+            for row in rows:
+                row_read = row.read()
+                try:
+                    key = row_read.pop(self.header_index_mapping[self.assoc_column_position])
+                except KeyError:
+                    try:
+                        key = row_read.pop(self.assoc_column_position)
+                    except KeyError:
+                        try:
+                            key = row_read.pop(self.assoc_column)
+                        except KeyError:
+                            raise ValueError(
+                                'The assoc_column={!r} could not be retrieved'.format(
+                                    self.assoc_column))
+                if key in result:
+                    raise ValueError('Duplicate value for {}={!r}'.format(key, result[key]))
+                result[key] = row_read
+            return result
+
+
 class VMToolbar(View):
     """
     Toolbar view for vms/instances collection destinations
@@ -133,12 +192,12 @@ class BasicProvisionFormView(View):
         vm_description = Input(name='service__vm_description')
         vm_filter = BootstrapSelect('service__vm_filter')
         num_vms = BootstrapSelect('service__number_of_vms')
-        catalog_name = Table('//div[@id="prov_vm_div"]/table')
+        catalog_name = NonEditableTable('//div[@id="prov_vm_div"]/table')
         provision_type = BootstrapSelect('service__provision_type')
         linked_clone = Input(name='service__linked_clone')
         pxe_server = BootstrapSelect('service__pxe_server_id')
-        pxe_image = Table('//div[@id="prov_pxe_img_div"]/table')
-        iso_file = Table('//div[@id="prov_iso_img_div"]/table')
+        pxe_image = NonEditableTable('//div[@id="prov_pxe_img_div"]/table')
+        iso_file = NonEditableTable('//div[@id="prov_iso_img_div"]/table')
 
     @View.nested
     class environment(Tab):  # noqa
@@ -158,10 +217,10 @@ class BasicProvisionFormView(View):
         resource_pool = BootstrapSelect('environment__placement_rp_name')
         folder = BootstrapSelect('environment__placement_folder_name')
         host_filter = BootstrapSelect('environment__host_filter')
-        host_name = Table('//div[@id="prov_host_div"]/table')
+        host_name = NonEditableTable('//div[@id="prov_host_div"]/table')
         datastore_create = Input('environment__new_datastore_create')
         datastore_filter = BootstrapSelect('environment__ds_filter')
-        datastore_name = Table('//div[@id="prov_ds_div"]/table')
+        datastore_name = NonEditableTable('//div[@id="prov_ds_div"]/table')
 
     @View.nested
     class hardware(Tab):  # noqa
@@ -192,23 +251,23 @@ class BasicProvisionFormView(View):
         # GCE
         is_preemtible = Input(name='hardware__is_preemptible')
 
-    @View.nested
-    class customize(Tab):  # noqa
-        TAB_NAME = 'Customize'
-        # Common
-        dns_servers = Input(name='customize__dns_servers')
-        dns_suffixes = Input(name='customize__dns_suffixes')
-        customize_type = BootstrapSelect('customize__sysprep_enabled')
-        specification_name = Table('//div[@id="prov_vc_div"]/table')
-        admin_username = Input(name='customize__root_username')
-        admin_password = Input(name='customize__root_password')
-        linux_host_name = Input(name='customize__linux_host_name')
-        linux_domain_name = Input(name='customize__linux_domain_name')
-        ip_address = Input(name='customize__ip_addr')
-        subnet_mask = Input(name='customize__subnet_mask')
-        gateway = Input(name='customize__gateway')
-        custom_template = Table('//div[@id="prov_template_div"]/table')
-        hostname = Input(name='customize__hostname')
+        @View.nested
+        class customize(Tab):  # noqa
+            TAB_NAME = 'Customize'
+            # Common
+            dns_servers = Input(name='customize__dns_servers')
+            dns_suffixes = Input(name='customize__dns_suffixes')
+            customize_type = BootstrapSelect('customize__sysprep_enabled')
+            specification_name = Table('//div[@id="prov_vc_div"]/table')
+            admin_username = Input(name='customize__root_username')
+            root_password = Input(name='customize__root_password')
+            linux_host_name = Input(name='customize__linux_host_name')
+            linux_domain_name = Input(name='customize__linux_domain_name')
+            ip_address = Input(name='customize__ip_addr')
+            subnet_mask = Input(name='customize__subnet_mask')
+            gateway = Input(name='customize__gateway')
+            custom_template = NonEditableTable('//div[@id="prov_template_div"]/table')
+            hostname = Input(name='customize__hostname')
 
     @View.nested
     class schedule(Tab):  # noqa
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -10,34 +10,36 @@ import re
 from selenium.common.exceptions import NoSuchElementException
 
 from navmazing import NavigateToSibling, NavigateToAttribute
+from widgetastic.widget import Text, View
+from widgetastic_patternfly import (
+    Button, BootstrapSelect, BootstrapSwitch, Dropdown, Input as WInput, Tab)
+from widgetastic_manageiq import (
+    TimelinesView, Accordion, ManageIQTree, SummaryTable, ConditionalSwitchableView)
+from widgetastic_manageiq.vm_reconfigure import DisksTable
 
 from cfme.base.login import BaseLoggedInPage
 from cfme.common.vm import VM, Template as BaseTemplate
-from cfme.common.vm_views import ManagementEngineView
-from cfme.exceptions import (CandidateNotFound, VmNotFound, OptionNotAvailable,
-                             DestinationNotFound, TemplateNotFound)
+from cfme.common.vm_views import (
+    ManagementEngineView, ProvisionView, EditView, RetirementView, VMDetailsEntities, VMToolbar,
+    VMEntities)
+from cfme.exceptions import (
+    CandidateNotFound, VmNotFound, OptionNotAvailable, DestinationNotFound, ItemNotFound,
+    VmOrInstanceNotFound)
 from cfme.fixtures import pytest_selenium as sel
 from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.services.requests import Request
 import cfme.web_ui.toolbar as tb
 from cfme.web_ui import (
-    CheckboxTree, Form, InfoBlock, Region, Quadicon, Tree, accordion, fill, flash, form_buttons,
-    match_location, Table, search, toolbar, Calendar, Select, Input, CheckboxTable,
-    summary_title, BootstrapTreeview, AngularSelect
-)
-from cfme.web_ui.search import search_box
+    CheckboxTree, Form, InfoBlock, Region, Quadicon, Tree, fill, flash, form_buttons,
+    match_location, Table, toolbar, Calendar, Select, Input, CheckboxTable,
+    summary_title, BootstrapTreeview, AngularSelect)
 from utils.appliance.implementations.ui import navigator, CFMENavigateStep, navigate_to
 from utils.conf import cfme_data
 from utils.log import logger
 from utils.pretty import Pretty
 from utils.wait import wait_for
 from utils import version, deferred_verpick
-from widgetastic.widget import Text, View
-from widgetastic_patternfly import (
-    Button, BootstrapSelect, BootstrapSwitch, Dropdown, Input as WInput
-)
-from widgetastic_manageiq import TimelinesView
-from widgetastic_manageiq.vm_reconfigure import DisksTable
+
 
 # for provider specific vm/template page
 QUADICON_TITLE_LOCATOR = ("//div[@id='quadicon']/../../../tr/td/a[contains(@href,'vm_infra/x_show')"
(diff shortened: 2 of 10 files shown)


dajoRH:
Lint report for commit d08b6f470449a4f4dfa155ff61d47c4bf712b766:

- :bangbang: Line 142:101: **E501** *line too long (107 > 100 characters)*
- :bangbang: Line 153:101: **E501** *line too long (105 > 100 characters)*
- :bangbang: Line 155:101: **E501** *line too long (107 > 100 characters)*
- :bangbang: Line 1119:101: **E501** *line too long (107 > 100 characters)*
- :bangbang: Line 1127:13: **E265** *block comment should start with '# '*

Hmmm
:frowning: You might like to check the lint of your code :point_up: 


**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
Lint report for commit 91f3d0e62eef94209bc40a06a159a34f30fca299:

- :bangbang: Line 170:101: **E501** *line too long (105 > 100 characters)*
- :bangbang: Line 172:101: **E501** *line too long (101 > 100 characters)*
- :bangbang: Line 1135:101: **E501** *line too long (107 > 100 characters)*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

dajoRH:
Lint report for commit 190046cfb0f98be808b0f259c206b1b563937d1b:

- :heavy_exclamation_mark: Line 138:1: **W293** *blank line contains whitespace*
- :bangbang: Line 186:101: **E501** *line too long (105 > 100 characters)*
- :bangbang: Line 188:101: **E501** *line too long (101 > 100 characters)*
- :bangbang: Line 1152:101: **E501** *line too long (107 > 100 characters)*
- :bangbang: Line 1240:16: **E126** *continuation line over-indented for hanging indent*

Hmmm
:frowning: You might like to check the lint of your code :point_up: 


**No commit flaws detected**.

*CFME QE Bot*

cfme/common/vm_views.py
```diff
@@ -0,0 +1,366 @@
+# -*- coding: utf-8 -*-
```
mshriver: Stealing all my code I see! :D

cfme/cloud/instance/__init__.py
```diff
@@ -404,25 +453,46 @@ def step(self, *args, **kwargs):
...
+
+
+@navigator.register(Instance, 'AttachVolume')
+class AttachVolume(CFMENavigateStep):
+    VIEW = AttachVolumeView
+
+    def prerequisite(self, *args, **kwargs):
+        if not self.obj.provider.one_of(OpenStackProvider):
+            raise DestinationNotFound('Attach Volume only available for Openstack instances')
+        elif self.obj.appliance.version < '5.7':
```
quarckster: **REQUIRED**
Not needed anymore

cfme/cloud/instance/__init__.py
```diff
@@ -431,61 +501,104 @@ class Details(CFMENavigateStep):
...
 
 
+@navigator.register(Instance, 'DetachVolume')
+class DetachVolume(CFMENavigateStep):
+    VIEW = DetachVolumeView
+
+    def prerequisite(self, *args, **kwargs):
+        if not self.obj.provider.one_of(OpenStackProvider):
+            raise DestinationNotFound('Detach Volume only available for Openstack instances')
+        elif self.obj.appliance.version < '5.7':
```
quarckster: **REQUIRED**
Not needed anymore

cfme/cloud/instance/__init__.py
```diff
@@ -521,42 +651,46 @@ def step(self):
 
 @navigator.register(Instance, 'RemoveFloatingIP')
 class RemoveFloatingIP(CFMENavigateStep):
-    prerequisite = NavigateToSibling('Details')
+    VIEW = RemoveFloatingIPView
+
+    def prerequisite(self, *args, **kwargs):
+        if not self.obj.provider.one_of(OpenStackProvider):
+            raise DestinationNotFound('Remove Floating IP only available for Openstack instances')
+        elif self.obj.appliance.version < '5.7':
```
quarckster: **REQUIRED**
Not needed anymore

cfme/common/vm_views.py
```diff
@@ -0,0 +1,366 @@
...
+            last_name = Input(name='requester__owner_last_name')
+            notes = Input(name='requester__request_notes')
+            manager_name = Input(name='requester__owner_manager')
+
+        @View.nested
+        class purpose(Tab):  # noqa
+            TAB_NAME = 'Purpose'
+            apply_tags = VersionPick({
+                Version.lowest(): CheckboxSelect('//div[@id="all_tags_treebox"]//ul'),
+                '5.7': BootstrapTreeview('all_tags_treebox')})
```
quarckster: **REQUIRED**
You can keep `BootstrapTreeview('all_tags_treebox')` only

cfme/common/vm_views.py
```diff
@@ -0,0 +1,366 @@
...
+        @View.nested
+        class properties(Tab):  # noqa
+            TAB_NAME = 'Properties'
+            instance_type = BootstrapSelect('hardware__instance_type')
+            guest_keypair = BootstrapSelect('hardware__guest_access_key_pair')
+            hardware_monitoring = BootstrapSelect('hardware__monitoring')
+            boot_disk_size = BootstrapSelect('hardware__boot_disk_size')
+            # GCE
+            is_preemtible = VersionPick({
+                Version.lowest(): None, '5.7': Input(name='hardware__is_preemptible')})
```
quarckster: **REQUIRED**
You should keep only 5.7 and higher.

cfme/common/vm_views.py
```diff
@@ -0,0 +1,366 @@
...
+
+    @View.nested
+    class form(View):  # noqa
+        """The form portion of the view"""
+        custom_identifier = TextInput(id='custom_1')
+        description = TextInput(id='description')
+        parent_vm = BootstrapSelect(id='chosen_parent')
+        # MultiBoxSelect element only has table ID in CFME 5.8+
+        # https://bugzilla.redhat.com/show_bug.cgi?id=1463265
+        chile_vms = MultiBoxSelect(id='child-vm-select')
```
quarckster: **REQURIED**
`MultiBoxSelect` doesn't have `id` argument. https://github.com/ManageIQ/integration_tests/blob/master/widgetastic_manageiq/__init__.py#L441

cfme/infrastructure/virtual_machines.py
```diff
@@ -85,36 +86,113 @@
...
+        expected_name = self.context['object'].name
+        expected_provider = self.context['object'].provider.name
+        try:
+            relationship_provider_name = self.entities.relationships.get_text_of('Infrastructure '
+                                                                                 'Provider')
+        except NameError:
+            if self.sidebar.vmstemplates.tree.currently_selected[-1] in ['<Archived>',
+                                                                         '<Orphaned>']:
+                return self.in_infra_vms and\
+                    self.entities.title.text == 'VM and Instance "{}"'.format(expected_name)
```
quarckster: **REQUIRED**
Don't use `\` to break lines, please parenthesis.

cfme/infrastructure/virtual_machines.py
```diff
@@ -1014,10 +1138,11 @@ def step(self, *args, **kwargs):
                 datacenter = self.obj.datacenter
         else:
             raise DestinationNotFound("the destination isn't found")
-        accordion.tree('VMs & Templates', 'All VMs & Templates', provider, datacenter)
+        self.view.sidebar.vmstemplates.tree.click_path('VMs & Templates',
+                                                       'All VMs & Templates', provider, datacenter)
 
     def resetter(self, *args, **kwargs):
-        reset_page()
+        self.view.reset_page()
```
quarckster: **REQUIRED**
I believe this is not needed. Use`BaseEntity` to get entities.

cfme/infrastructure/virtual_machines.py
```diff
@@ -954,22 +1074,24 @@ def get_number_of_vms(do_not_navigate=False):
...
+        self.view.sidebar.vmstemplates.tree.click_path('All VMs & Templates')
 
     def resetter(self, *args, **kwargs):
-        reset_page()
+        self.view.reset_page()
 
-    def am_i_here(self, *args, **kwargs):
-        return match_page(summary='All VMs & Templates')
+    def am_i_here(self):
+        return self.view.is_displayed and self.view.sidebar.vmstemplates.is_opened
```
quarckster: **REQUIRED**
Here and below `am_i_here` is not needed if the view has `is_displayed`

cfme/common/vm_views.py
lkhomenk: one more way to get your PR reviewed ;)

cfme/infrastructure/virtual_machines.py
lkhomenk: a sort of. main view check page. sub-views verify that we are in required sidebar - 3 different Accordions. this is a try to limit useless navigation

lkhomenk:
this PR was forked from @mshriver 's one. His review is a must ;)

cfme/common/vm_views.py
izapolsk: @mshriver, all your code will anyway be merged as yours. Look at commits tab ;)

mshriver: oh yeah, no worries here, just having some fun :)

dajoRH:
*CFME QE Bot*

cfme/infrastructure/virtual_machines.py
```diff
@@ -9,23 +9,23 @@
...
 
-from navmazing import NavigateToSibling, NavigateToAttribute
+from navmazing import NavigateToSibling, NavigateToObject
 
 from cfme.base.login import BaseLoggedInPage
+from cfme.base import Server
 from cfme.common.vm import VM, Template as BaseTemplate
-from cfme.common.vm_views import ManagementEngineView
+from cfme.common.vm_views import (ManagementEngineView, ProvisionView, EditView, RetirementView,
+                                  VMDetailsEntities, VMToolbar, VMEntities)
```
mshriver: Drop this import group onto a new line so they're all the same style:

```python
from cfme.common.vm_views import (
    View1, View2, View3, etc)

File: cfme/infrastructure/virtual_machines.py [hunk 12]
```diff
@@ -34,9 +34,9 @@
 from utils import version, deferred_verpick
 from widgetastic.widget import Text, View
```
mshriver: Can you move these up so the imports are correctly organized into the 3 groups (no comments, just import groups)

```python
# Import built-ins

# Import installed modules

# Import local code

File: cfme/infrastructure/virtual_machines.py [hunk 13]
```diff
@@ -86,46 +86,110 @@
...
 
-class InfraVmDetailsToolbar(View):
-    configuration = Dropdown("Configuration")
-    policy = Dropdown("Policy")
-    lifecycle = Dropdown("Lifecycle")
+class InfraVmDetailsToolbar(VMToolbar):
     monitoring = Dropdown("Monitoring")
-    power = Dropdown("Power")
     access = Dropdown("Access")
+    reload = Button(title='Reload current display')
```
mshriver: `reload` is already in common.vm_views.VMToolbar

cfme/infrastructure/virtual_machines.py
```diff
@@ -86,46 +86,110 @@
...
 
 
-class InfraVmDetailsToolbar(View):
-    configuration = Dropdown("Configuration")
-    policy = Dropdown("Policy")
-    lifecycle = Dropdown("Lifecycle")
+class InfraVmDetailsToolbar(VMToolbar):
     monitoring = Dropdown("Monitoring")
-    power = Dropdown("Power")
     access = Dropdown("Access")
```
mshriver: It looks like there is a slightly different toolbar for vm templates vs vms, where templates toolbars don't have the Access or Power dropdowns

cfme/infrastructure/virtual_machines.py
```diff
@@ -86,46 +86,110 @@
...
 
 
 drift_table = CheckboxTable("//th[normalize-space(.)='Timestamp']/ancestor::table[1]")
 
 
-class InfraVmDetailsToolbar(View):
-    configuration = Dropdown("Configuration")
-    policy = Dropdown("Policy")
-    lifecycle = Dropdown("Lifecycle")
+class InfraVmDetailsToolbar(VMToolbar):
```
mshriver: VMToolbar is written for the 'All' view, and has a dropdown for the download button.

For the details toolbar, you'll want to either add a common view for details toolbar or add all the widgets here and build it off of View.

cloud.instance.InstanceDetailsToolbar inherits View directly and implements all its widgets, mostly because the 'Power' button title is different between instances and infra VM's.

cfme/infrastructure/virtual_machines.py
```diff
@@ -86,46 +86,110 @@
...
+        ACCORDION_NAME = 'Templates'
+        tree = ManageIQTree()
+
 
+class InfraVMView(BaseLoggedInPage):
+    """Base view for header/nav check, inherit for navigatable views"""
 
-class InfraVmDetailsView(BaseLoggedInPage):
-    # TODO this is only minimal implementation for toolbar access through widgetastic
+    including_entities = View.include(VMEntities, use_parent=True)
```
mshriver: You shouldn't include the VMEntities view here, since this view is inherited for both the 'All' page and the 'Details' page below.

I think this class should only have the `in_infra_vms` property

cfme/infrastructure/virtual_machines.py
```diff
@@ -86,46 +86,110 @@
...
+                match_location(controller='vm_infra', title='Virtual Machines'))
+
+
+class VmsTemplatesAllView(InfraVMView):
+    """
+    The collection page for instances
+    """
+    @property
+    def is_displayed(self):
+        return self.in_infra_vms
```
mshriver: I would additionally check the title in entities, and possibly what accordion is open in the sidebar.

cfme/infrastructure/virtual_machines.py
```diff
@@ -86,46 +86,110 @@
...
+
 
+class InfraVMView(BaseLoggedInPage):
+    """Base view for header/nav check, inherit for navigatable views"""
 
-class InfraVmDetailsView(BaseLoggedInPage):
-    # TODO this is only minimal implementation for toolbar access through widgetastic
+    including_entities = View.include(VMEntities, use_parent=True)
+
+    @View.nested
```
mshriver: Why are you redefining the accordion here? Shouldn't be needed as you already have a view above that you can include/nest as needed.

cfme/infrastructure/virtual_machines.py
```diff
@@ -86,46 +86,110 @@
...
+
+    toolbar = View.nested(VMToolbar)
+    sidebar = View.nested(VmsTemplatesAccordion)
+    including_entities = View.include(VMEntities, use_parent=True)
+
+
+class InfraVmDetailsView(InfraVMView):
     toolbar = View.nested(InfraVmDetailsToolbar)
+    sidebar = View.nested(VmsTemplatesAccordion)
+    entities = View.nested(VMDetailsEntities)
```
mshriver: It looks like some infra VM's will have additional `SummaryTable` widgets in their details view.

I'm seeing 'Datastore Allocation Summary' and 'Datastore Actual Usage Summary' when looking at vsphere VM's.

cfme/infrastructure/virtual_machines.py
```diff
@@ -86,46 +86,110 @@
...
+        tree = ManageIQTree()
+
+    @View.nested
+    class templates(Accordion):  # noqa
+        ACCORDION_NAME = 'Templates'
+        tree = ManageIQTree()
+
+    @property
+    def in_infra_vms(self):
+        return (self.logged_in_as_current_user and
```
mshriver: You have two different styles in this PR for returns with multiple statements. Let's use single style, as such:

```python
return (
    first_statement and
    second_thing == ['some', 'check'] and
    third_statement)

File: cfme/infrastructure/virtual_machines.py [hunk 21]
```diff
@@ -151,6 +215,47 @@ class InfraVmReconfigureView(BaseLoggedInPage):
     is_displayed = False
 
 
+class MigrateView(BaseLoggedInPage):
+    title = Text('#explorer_title_text')
+
+    @View.nested
+    class form(View):  # noqa
+        submit_button = Button('Submit')
```
mshriver: just `submit` and `cancel` for these widget names

cfme/infrastructure/virtual_machines.py
```diff
@@ -151,6 +215,47 @@ class InfraVmReconfigureView(BaseLoggedInPage):
...
+            folder = BootstrapSelect('environment__placement_folder_name')
+            host_filter = BootstrapSelect('environment__host_filter')
+            host_name = Table('//div[@id="prov_host_div"]/table')
+            datastore_filter = BootstrapSelect('environment__ds_filter')
+            datastore_name = Table('//div[@id="prov_ds_div"]/table')
+
+        @View.nested
+        class schedule(Tab):  # noqa
+            TAB_NAME = 'Schedule'
+            # TODO radio widget # schedule_type = Radio('schedule__schedule_type')
```
mshriver: https://github.com/ManageIQ/integration_tests/blob/master/widgetastic_manageiq/__init__.py#L1416

Aware of widgetastic_manageiq.RadioGroup and just postponing implementation?

cfme/infrastructure/virtual_machines.py
```diff
@@ -151,6 +215,47 @@ class InfraVmReconfigureView(BaseLoggedInPage):
...
+
+    @property
+    def is_displayed(self):
+        # Nothing is shown
+        return False
```
mshriver: :+1:

cfme/infrastructure/virtual_machines.py
mshriver: lkhomenk CFMENavigateStep.am_i_here calls self.view.is_displayed by default. These checks should be moved into the view's is_displayed method.

cfme/infrastructure/virtual_machines.py
```diff
@@ -960,22 +1066,21 @@ def get_number_of_vms(do_not_navigate=False):
 @navigator.register(Template, 'All')
 @navigator.register(Vm, 'All')
 class VmAllWithTemplates(CFMENavigateStep):
-    prerequisite = NavigateToAttribute('appliance.server', 'LoggedIn')
-
-    def step(self, *args, **kwargs):
-        self.prerequisite_view.navigation.select('Compute', 'Infrastructure', '/vm_infra/explorer')
-        accordion.tree('VMs & Templates', 'All VMs & Templates')
+    VIEW = VmsTemplatesAllView
+    prerequisite = NavigateToObject(Server, 'LoggedIn')
```
mshriver: Keep `NavigateToAttribute` here so that we're specfically logging into appliance.server, not just Server object.

cfme/infrastructure/virtual_machines.py
```diff
@@ -960,22 +1066,21 @@ def get_number_of_vms(do_not_navigate=False):
...
-    def am_i_here(self, *args, **kwargs):
-        return match_page(summary='All VMs & Templates')
+    def am_i_here(self):
+        return self.view.is_displayed and self.view.sidebar.vmstemplates.is_opened
 
 
 @navigator.register(Template, 'AllForProvider')
 @navigator.register(Vm, 'AllForProvider')
 class VmAllWithTemplatesForProvider(CFMENavigateStep):
+    VIEW = VmsTemplatesAllView
```
mshriver: I think you actually need a separate view here, with the check for sidebar selection in that view's is_displayed method.

Reason being - although `am_i_here()` covers the difference between the views during navigation, if a caller uses `view.is_displayed` apart from navigation, the results could be a false-positive.

cloud.instances has separate `InstanceAllView` and `InstanceProviderAllView` for this reason.

cfme/infrastructure/virtual_machines.py
```diff
@@ -997,17 +1099,19 @@ def am_i_here(self, *args, **kwargs):
             provider = self.obj.provider.name
         else:
             raise DestinationNotFound("the destination isn't found")
-        return match_page(summary='VM or Templates under Provider "{}"'.format(provider))
+        return (self.view.sidebar.vmstemplates.is_opened and self.view.entities.title.text ==
```
mshriver: I think this check needs to move into `view.is_displayed`, and you should call it directly here as `self.view.is_displayed`, so that `view.is_displayed` is accurate when called outside of navigation context as well.

cfme/infrastructure/virtual_machines.py
```diff
@@ -997,17 +1099,19 @@ def am_i_here(self, *args, **kwargs):
...
+                'VM or Templates under Provider "{}"'.format(provider))
 
 
 @navigator.register(Template, 'AllForDatacenter')
 @navigator.register(Vm, 'AllForDatacenter')
 class VmAllWithTemplatesForDatacenter(CFMENavigateStep):
+    VIEW = VmsTemplatesAllView
```
mshriver: Same as Provider view above, I think this needs to be a separate view with `is_displayed` containing the checks for the title and accordion selection.

cfme/infrastructure/virtual_machines.py
```diff
@@ -1020,10 +1124,8 @@ def step(self, *args, **kwargs):
                 datacenter = self.obj.datacenter
         else:
             raise DestinationNotFound("the destination isn't found")
-        accordion.tree('VMs & Templates', 'All VMs & Templates', provider, datacenter)
-
-    def resetter(self, *args, **kwargs):
-        reset_page()
+        self.view.sidebar.vmstemplates.tree.click_path('VMs & Templates',
```
mshriver: This click_path looks redundant, given that you're calling it from sidebar.vmstemplates.tree - which should be the tree in the 'VMs & Templates' accordion already.

Note how in cloud.instances.AllForProvider.step, the first argument to 'sidebar.instances_by_provider.tree.click_path' is 'Instances by Provider', which is the folder label within the accordion. 

Believe your first argument here should be 'All VMs & Templates`, since its being given against a specific accordion .

cfme/infrastructure/virtual_machines.py
```diff
@@ -1071,113 +1174,118 @@ class VmAllWithTemplatesDetails(CFMENavigateStep):
     VIEW = InfraVmDetailsView
```
mshriver: Toolbar is different between template and VM view, you'll need a different view here

cfme/infrastructure/virtual_machines.py
```diff
@@ -1071,113 +1174,118 @@ class VmAllWithTemplatesDetails(CFMENavigateStep):
...
     prerequisite = NavigateToSibling('All')
 
-    def step(self, *args, **kwargs):
-        sel.click(self.obj.find_quadicon(do_not_navigate=True))
+    def step(self):
+        self.prerequisite_view.sidebar.vmstemplates.tree.click_path('All VMs & Templates')
+        try:
+            row = self.prerequisite_view.entities.get_first_entity(by_name=self.obj.name)
```
mshriver: I wouldn't call it 'row' here as it may be a quadicon or tile, you never selected Grid View specifically.

cfme/infrastructure/virtual_machines.py
```diff
@@ -1071,113 +1174,118 @@ class VmAllWithTemplatesDetails(CFMENavigateStep):
     VIEW = InfraVmDetailsView
     prerequisite = NavigateToSibling('All')
 
-    def step(self, *args, **kwargs):
-        sel.click(self.obj.find_quadicon(do_not_navigate=True))
+    def step(self):
+        self.prerequisite_view.sidebar.vmstemplates.tree.click_path('All VMs & Templates')
```
mshriver: Why not go to 'Templates' accordion so that there are fewer entities to search through?

cfme/infrastructure/virtual_machines.py
lkhomenk: right - that's why I override some buttons in InfraVmDetailsToolbar. and I'll update "Download" too

mshriver: But the VM Details page doesn't have 'Lifecycle' dropdown, so it needs a new view. We don't want to inherit from views that have 'extra' widgets.

cfme/infrastructure/virtual_machines.py
lkhomenk: it was missed. forgot to remove

cfme/infrastructure/virtual_machines.py
lkhomenk: @mshriver but that leads to other independent views like AllArchived/AllOrphaned/AllForDatastore. maybe it would be better to add method inside VmsTemplatesAllView that would navigate thru accordion? this way we would get correct is_displayed from VmsTemplatesAllView. I tried to work with it creating am_i_here's in navigation steps - which were using is_displayed from VmsTemplatesAllView + additional checks as we were getting  "sub-views"

cfme/infrastructure/virtual_machines.py
lkhomenk: right. Ievgen helped with entities. before I was using List View + row.

cfme/infrastructure/virtual_machines.py
lkhomenk: got it - contitionalview will help

cfme/infrastructure/virtual_machines.py
lkhomenk: this should be in TODO - override VmEntities to provide - provider and type / search not only by name. 
Here it is used that way to have 1 navigation step for both VM and Templates. with has some gaps - we do not check if name is unique

cfme/infrastructure/virtual_machines.py
lkhomenk: that one just doesn't work. not sure - maybe someone else is actively working on RadioGroup widget or not

cfme/infrastructure/virtual_machines.py
lkhomenk: mb we should file RFE. there's really not a single chance to guest which VM is migrating

mshriver: I think its worth a BZ or GH issue to improve this page title.

cfme/infrastructure/virtual_machines.py
mshriver: @izapolsk @mfalesni @psav Tagging some other widgetastic dev's since this is an important consideration for many of our modules.

I don't think the 'All' view for any page model should have methods to navigate through a given widget to get to some detailed selection, as this will confuse page models with navigation destinations.

If these pages have different UI elements, then we need separate view classes to accurately represent them. The assertion for whether a page is currently displayed needs to be collapsed down into the view's `is_displayed` method otherwise it will be very unreliable to use outside of trying to navigate again.  We need to be able to create a view object and reliably assert `view.is_displayed` - we can't try to call some navigation class's `am_i_here` directly to assert what page we're at.

I don't think we necessarily need to provide navigation destinations for Archived/Orphaned and other folders within the accordion menus, especially if their views are not unique and share toolbar/entities with the view that is used for the other folders within the specific accordion.  This may create a situation where I navigate to 'Vms & Templates' accordion, and then use the returned view to call `view.tree.click_path('Archived')` to get to the archived folder.  Now - if I call view.is_displayed() at this point, should it return False since I'm not in the 'All Vms & Templates' folder, or should it return True since its still in 'Vms & Templates'?  I believe it should return False - though the view can still be used to interact with the toolbar/entities after selecting a different option in the tree.

I do think that we need to model separate views for these pages when there are differences in their toolbars/entities, and add a navigation destination that returns this view if a test needs it.
In cloud.instances, there are navigation destinations for the 4 accordion blocks, but not to the Archived/Orphaned sections within the block.

We haven't provided navigation destinations to the filter folders within the 'All' views anywhere, for example.

cfme/infrastructure/virtual_machines.py
mshriver: That's definitely a needed enhancement for VMEntities.

What I'm getting at here is that we can very easily eliminate a big part of this gap by having separate navigation for VM details and Template details, that use their respective accordion menu to narrow down the entities that we're searching for in the first place.

Additionally the details view for a template is different than that for a VM, and the navigation destination needs to return an accurate view.  This can be done with a ConditionalView though that's based on the title, which has 'Template' or 'Virtual Machine' in it.  The navigation destination returns the wrapper view that collapses to either VM view or Template view.

cfme/infrastructure/virtual_machines.py
mshriver: can you capture why/how its not working and write an issue against the widget? We need to get it either fixed or expanded to support the radio on this form.

dajoRH:
*CFME QE Bot*

dajoRH:
Lint report for commit 3f2b19af4a34cd3d917a21edf8cdf5fef4a795c0:

`cfme/tests/infrastructure/test_pxe_provisioning.py`:
- :bangbang: Line 117:13: **E123** *closing bracket does not match indentation of opening bracket's line*

**No commit flaws detected**.

*CFME QE Bot*

mshriver:
Modified PR description for PRT providers, adding vsphere6-nested, rhv41, and scvmm. Provisioned VM's weren't showing up in PRT run 15221 on vsphere65-nested.   Providers are getting cleaned up to ensure adequate resources for provisioning.

cfme/infrastructure/virtual_machines.py
```diff
@@ -86,46 +87,161 @@
...
+class InfraVMView(BaseLoggedInPage):
+    """Base view for header/nav check, inherit for navigatable views"""
+
+    @property
+    def in_infra_vms(self):
+        return (
+            self.logged_in_as_current_user and
+            self.navigation.currently_selected == ['Compute', 'Infrastructure',
+                                                   'Virtual Machines'] and
+            match_location(controller='vm_infra', title='Virtual Machines'))
```
quarckster: **REQUIRED**
Old web_ui code should be removed from widgetastic.

cfme/infrastructure/virtual_machines.py
```diff
@@ -151,6 +267,48 @@ class InfraVmReconfigureView(BaseLoggedInPage):
...
+    title = Text('#explorer_title_text')
+
+    @View.nested
+    class form(View):  # noqa
+        submit = Button('Submit')
+        cancel = Button('Cancel')
+
+        @View.nested
+        class request(Tab):  # noqa
+            TAB_NAME = 'Request'
```
quarckster: **REQUIRED**
`TAB NAME` is only needed when a class name cannot be converted to a tab name with capitalize method.

lkhomenk: it was used all over provisioning view - so just sticking to common practice

cfme/infrastructure/virtual_machines.py
lkhomenk: @quarckster further updates will remove match_location totally from this module.

cfme/common/vm_views.py
```diff
@@ -12,6 +12,14 @@
     CheckableManageIQTree, VersionPick, Version, CheckboxSelect, BaseEntitiesView)
 
 
+class NonEditableTable(Table):
+    def fill(self, values):
+        """Clicks on item - fill by selecting required value"""
+
+        value = values.get('name', '<None>')
+        self.row(name=value).click()
```
mfalesni: Protocol of `fill` is that if it did any change, it must return True, otherwise False. Also, if this is the "selector" table type, it should probably also reimplement `read` to give you the name of the thing selected :)

cfme/provisioning.py
```diff
@@ -170,18 +170,16 @@ def do_vm_provisioning(template_name, provider, vm_name, provisioning_data, requ
...
-    fill(provisioning_form, provisioning_data,
-         action=provisioning_form.submit_button)
+        'request': {
+            'email': 'template_provisioner@example.com',
+            'first_name': 'Template',
+            'last_name': 'Provisioner',
+            'notes': note}})
+    view = navigate_to(vm, 'Provision')
+    view.form.fill(provisioning_data)
+    view.form.submit_button.click()
```
mfalesni: Use `view.form.fill_with(provisioning_data, on_change=view.form.submit_button)`

cfme/infrastructure/virtual_machines.py
```diff
@@ -86,46 +87,169 @@
...
+class VmsTemplatesAllView(InfraVMView):
+    """
+    The collection page for instances
+    """
+    @property
+    def is_displayed(self):
+        return (
+            self.in_infra_vms and
+            self.sidebar.vmstemplates.tree.currently_selected == 'All VMs & Templates' and
+            self.entities.title.text == 'All VMs & Templates')
```
mfalesni: Put is_displayed after widget definitions, this does not look good :)

mfalesni: (applies for all such occurences)

cfme/infrastructure/virtual_machines.py
```diff
@@ -86,46 +87,169 @@
...
+    toolbar = ConditionalSwitchableView(reference='entities.title')
+
+    @toolbar.register(lambda title: "VM and Instance" in title)
+    class VmsToolbar(InfraVmDetailsToolbar):
+        pass
+
+    @toolbar.register(lambda title: "VM Template and Image" in title)
+    class TemplatesToolbar(InfraVmTemplatesGenericDetailsToolbar):
+        pass
+    sidebar = View.nested(VmsTemplatesAccordion)
```
mfalesni: Leave a blank line between 219 and 220

cfme/infrastructure/virtual_machines.py
```diff
@@ -86,46 +87,169 @@
...
+    @property
+    def is_displayed(self):
+        expected_name = self.context['object'].name
+        expected_provider = self.context['object'].provider.name
+        try:
+            relationship_provider_name = self.entities.relationships.get_text_of('Infrastructure '
+                                                                                 'Provider')
+        except NameError:
+            if self.sidebar.vmstemplates.tree.currently_selected[-1] in ['<Archived>',
+                                                                         '<Orphaned>']:
```
mfalesni: pull the currently_selected value into an intermediate variable to remove the need of multilining the if, it will look better.

cfme/infrastructure/virtual_machines.py
```diff
@@ -86,46 +87,169 @@
...
+                return (
+                    self.in_infra_vms and
+                    self.entities.title.text == 'VM and Instance "{}"'.format(expected_name))
+            logger.warning('No "Infrastructure Provider" Relationship, VM details view not '
```
mfalesni: Use self.logger in widgetastic widgets.

cfme/infrastructure/virtual_machines.py
```diff
@@ -86,46 +87,169 @@
...
 class InfraVmTimelinesView(TimelinesView, BaseLoggedInPage):
     @property
     def is_displayed(self):
-        return self.logged_in_as_current_user and \
+        return (
+            self.logged_in_as_current_user and
             self.navigation.currently_selected == ['Compute', 'Infrastructure',
-                                                   '/vm_infra/explorer'] and \
-            super(TimelinesView, self).is_displayed
+                                                   '/vm_infra/explorer'] and
```
mfalesni: Hmm, I believe that the comparison will not work with URL matching. Pity is that so far is_displayed in navmazing gets muffled.

cfme/infrastructure/virtual_machines.py
mfalesni: Or hide the VM's id somewhere in a data attribute in html :)

cfme/infrastructure/virtual_machines.py
lkhomenk: wasn't checking this view. just was fixing it's string formating to meet all others

cfme/infrastructure/virtual_machines.py
lkhomenk: done

cfme/common/vm_views.py
```diff
@@ -12,6 +12,65 @@
     CheckableManageIQTree, VersionPick, Version, CheckboxSelect, BaseEntitiesView)
 
 
+class NonEditableTable(Table):
+    """Wigdet for non-editable table. used for selecting value"""
+    def fill(self, values):
```
lkhomenk: @mfalesni is such implementation of fill/read fine? I've also left read_content to get table values - just in case.

dajoRH:
Lint report for commit 18d86e14ee8f490919b04711cdc21682fa6fd49a:

`cfme/common/vm_views.py`:
- :bangbang: Line 65:1: **E302** *expected 2 blank lines, found 1*

**No commit flaws detected**.

*CFME QE Bot*

cfme/common/vm_views.py
lkhomenk: @mshriver please add you comment on this too if there're any

dajoRH:
*CFME QE Bot*

lkhomenk:
latest run 15476:
5.7 74% success:
test_provisioning.py - 4/12 NoSuchElement - due to missing partial match in WT.Dropdown
test_vm_power_control.py - issue with tinytemplate which supposed to have VMTools. and others like other_div_displayed which are not related to these code changes
5.8 81%:
same issue with Dropdown and others not related to code changes.
I'll check/fix test_power_options and all other provider related failures saparatly

cfme/common/vm_views.py
```diff
@@ -64,6 +64,65 @@ def InstanceEntity():  # noqa
...
+            self.logger.info('Nothing is currently selected')
+            return None
+        else:
+            return result[0]
+
+    def read(self):
+        return self.currently_selected
+
+    def read_content(self):
+        """This is a default Table.read() method for those who will need table content"""
```
mfalesni: Is this copied from the original table class? You can do just `return super(NonEditableTable, self).read()` in this method to bridge this behaviour into this method.

lkhomenk: my bad. fixed

cfme/common/vm_views.py
```diff
@@ -64,6 +64,40 @@ def InstanceEntity():  # noqa
...
+class NonEditableTable(Table):
+    """Wigdet for non-editable table. used for selecting value"""
+    def fill(self, values):
+        """Clicks on item - fill by selecting required value"""
+        value = values.get('name', '<None>')
+        changed = False
+        if value != self.currently_selected:
+            changed = True
+            self.row(name=value).click()
+            return changed
```
mshriver: **Required**

Collapse these two return statements, no need for the `else` splitting them.

cfme/common/vm_views.py
```diff
@@ -64,6 +64,40 @@ def InstanceEntity():  # noqa
     })
 
 
+class NonEditableTable(Table):
```
mshriver: I don't think the name 'NonEditable' is accurate, I would expect more method overrides in such a table class.

What do you think about 'ClickableTable' or 'SelectTable', something along those lines?

I think this class may be more appropriate defined in WT_manageiq and imported here, because I think there may be other places in the UI where we'll see or use it.

cfme/infrastructure/virtual_machines.py
```diff
@@ -10,34 +10,36 @@
...
 from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.services.requests import Request
 import cfme.web_ui.toolbar as tb
 from cfme.web_ui import (
-    CheckboxTree, Form, InfoBlock, Region, Quadicon, Tree, accordion, fill, flash, form_buttons,
-    match_location, Table, search, toolbar, Calendar, Select, Input, CheckboxTable,
-    summary_title, BootstrapTreeview, AngularSelect
-)
-from cfme.web_ui.search import search_box
+    CheckboxTree, Form, InfoBlock, Region, Quadicon, Tree, fill, flash, form_buttons,
```
mshriver: :( sad to still see these here, but they need to be removed from common at the same time as here... another day.

cfme/infrastructure/virtual_machines.py
```diff
@@ -86,46 +88,174 @@
...
 
 
 drift_table = CheckboxTable("//th[normalize-space(.)='Timestamp']/ancestor::table[1]")
 
 
-class InfraVmDetailsToolbar(View):
-    configuration = Dropdown("Configuration")
-    policy = Dropdown("Policy")
-    lifecycle = Dropdown("Lifecycle")
+class InfraVmTemplatesGenericDetailsToolbar(View):
```
mshriver: **Optional**
Consider making these class names a bit shorter, I think `VmTemplatesDetailsToolbar` is fine here, and just `Vm` in place of `InfraVm` in the class names below.

At least all consistent capitalization - there's an `InfraVM` on L132

mshriver: On second thought, it looks like there are only slight differences between this and VMToolbar, I believe a chunk of this might get updated again with the common.vm refactor.

cfme/common/vm_views.py
lkhomenk: I was thinking that this Table won't be popular, but if @mfalesni don't mind it can be moved to WT_manageiq.

cfme/infrastructure/virtual_machines.py
lkhomenk: now they are just in Vm class. I decided not to add it in this PR.

cfme/infrastructure/virtual_machines.py
lkhomenk: I had to make Generic toolbar for templates and vms. when we'll be merging it with instances it can be refactored

cfme/infrastructure/virtual_machines.py
```diff
@@ -86,46 +88,174 @@
...
+    toolbar = View.nested(VMToolbar)
+    sidebar = View.nested(VmsTemplatesAccordion)
+    including_entities = View.include(VMEntities, use_parent=True)
+
+    @property
+    def is_displayed(self):
+        return (
+            self.in_infra_vms and
+            str(self.entities.title.text) ==
+            'VM or Templates under Provider \"{}\"'.format(self.context['object'].provider.name))
```
psav: The slashes here are superfluous, just sayin

PR #5120, opened 2017-08-08
[1LP][RFR]Enable Container object testing

shalomnaim1:
{{pytest: cfme/tests/containers/test_smart_management.py -v --use-provider cm-env2}}

Add Container object to smart management tests


shalomnaim1:
Thanks!!!

PR #5122, opened 2017-08-08
[WIPTEST] JsPaginator for upstream

izapolsk:
This PR adds updated paginator based on js api for upstream/5.9.


izapolsk:
I decided to keep both Paginator and QuadIcons in one PR

PR #5124, opened 2017-08-09
[1LP][RFR] test snapshot operations on suspended vm

apagac:
This is a new test for testing basic snapshot operations on a suspended vm.

First, I create a snapshot when the vm is running, then I suspend it and create a second snapshot [1]. While the vm is suspended, I try to revert to first snapshot and check the status of the vm (off). I then revert back to the second snapshot, check the status of the vm (suspended) and try to delete all the snapshots. ~~This is a bit tricky since I can't mask the NameError with wait_for, so I used try/except blocks.~~ Update: The NameError is now caught within Snapshot.exists.

[1] we can't create two snapshots when the vm is suspended: https://bugzilla.redhat.com/show_bug.cgi?id=1419872

### DIFF
--- a/cfme/tests/infrastructure/test_snapshot.py
+++ b/cfme/tests/infrastructure/test_snapshot.py
@@ -174,6 +174,52 @@ def test_verify_revert_snapshot(full_test_vm, provider, soft_assert, register_ev
 
 
 @pytest.mark.uncollectif(lambda provider: not provider.one_of(VMwareProvider))
+def test_operations_suspended_vm(small_test_vm, provider, soft_assert):
+    """Tests snapshot operations on suspended vm
+
+    Metadata:
+        test_flag: snapshot, provision
+    """
+    # Create first snapshot when VM is running
+    snapshot1 = new_snapshot(small_test_vm)
+    snapshot1.create()
+    wait_for(lambda: snapshot1.active, num_sec=300, delay=20, fail_func=snapshot1.refresh)
+    # Suspend the VM
+    small_test_vm.power_control_from_cfme(option=small_test_vm.SUSPEND, cancel=False)
+    small_test_vm.wait_for_vm_state_change(desired_state=small_test_vm.STATE_SUSPENDED)
+    current_state = small_test_vm.find_quadicon().state
+    soft_assert(current_state.startswith('currentstate-suspended'),
+                "Quadicon state is {}".format(current_state))
+    # Create second snapshot when VM is suspended
+    snapshot2 = new_snapshot(small_test_vm)
+    snapshot2.create()
+    wait_for(lambda: snapshot2.active, num_sec=300, delay=20, fail_func=snapshot2.refresh)
+    # Try to revert to first snapshot while the VM is suspended
+    snapshot1.revert_to()
+    wait_for(lambda: snapshot1.active, num_sec=300, delay=20, fail_func=snapshot1.refresh)
+    # Check VM state, VM should be off
+    current_state = small_test_vm.find_quadicon().state
+    soft_assert(current_state.startswith('currentstate-off'),
+                "Quadicon state is {}".format(current_state))
+    assert small_test_vm.provider.mgmt.is_vm_stopped(small_test_vm.name)
+    # Revert back to second snapshot
+    snapshot2.revert_to()
+    wait_for(lambda: snapshot2.active, num_sec=300, delay=20, fail_func=snapshot2.refresh)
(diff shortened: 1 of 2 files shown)


apagac:
PR #5105 needs to be merged for this to work.

apagac:
PRT is failing because there is a known problem with a provider. On my local setup, this runs fine.

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -170,6 +170,47 @@ def test_verify_revert_snapshot(test_vm, provider, soft_assert, register_event,
...
 @pytest.mark.uncollectif(lambda provider: not provider.one_of(VMwareProvider))
+def test_operations_suspended_vm(test_vm, provider, soft_assert):
+    """Tests snapshot operations on suspended vm
+
+    Metadata:
+        test_flag: snapshot, provision
+    """
+    snapshot1 = new_snapshot(test_vm)
+    snapshot1.create()
+    wait_for(lambda: snapshot1.active, num_sec=300, delay=20, fail_func=sel.refresh)
```
quarckster: **REQUIRED**
I recommend to avoid using `pytest_selenium` in the new code. Please use widgetastic.

apagac:
This is continuously failing PRT with 5.7., but I can run this with 5.7.3. on my local setup just fine and can't reproduce the failure.
Moving to RFR.

dajoRH:
*CFME QE Bot*

apagac:
Dependent PRs have been merged, this PR has been rebased and is now ready to review.

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -174,6 +174,47 @@ def test_verify_revert_snapshot(full_test_vm, provider, soft_assert, register_ev
...
+    assert small_test_vm.provider.mgmt.is_vm_stopped(small_test_vm.name)
+    snapshot2.revert_to()
+    wait_for(lambda: snapshot2.active, num_sec=300, delay=20, fail_func=snapshot2.refresh)
+    current_state = small_test_vm.find_quadicon().state
+    soft_assert(current_state.startswith('currentstate-suspended'),
+                "Quadicon state is {}".format(current_state))
+    assert small_test_vm.provider.mgmt.is_vm_suspended(small_test_vm.name)
+    snapshot2.delete_all()
+    try:
+        wait_for(lambda: not snapshot1.exists, num_sec=300, delay=20, fail_func=snapshot1.refresh)
```
mkoura: You can use ``fail_condition=True`` to avoid lambda

mkoura: I didn't realized that it's a property, not function, so disregard my comment :)

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -174,6 +174,56 @@ def test_verify_revert_snapshot(full_test_vm, provider, soft_assert, register_ev
...
+    current_state = small_test_vm.find_quadicon().state
+    soft_assert(current_state.startswith('currentstate-suspended'),
+                "Quadicon state is {}".format(current_state))
+    assert small_test_vm.provider.mgmt.is_vm_suspended(small_test_vm.name)
+    # Try to delete both snapshots while the VM is suspended
+    snapshot2.delete_all()
+    # Wait for both snapshots to disappear
+    try:
+        wait_for(lambda: not snapshot1.exists, num_sec=300, delay=20, fail_func=snapshot1.refresh)
+    except NameError:
```
quarckster: @apagac  I don't understand why `NameError` is caught here? Should it be `TimeOutError`?

apagac: So this wait_for is used as a check if the snapshot is really deleted from the VM. If all goes well, the snapshot we are looking for is not there, and snapshot1.exists throws NameError exception.

mshriver: Snapshot.exists shouldn't throw NameError exception, it should catch that and return `False`, allowing for much better flow here.

I'd like to scope that change to this PR to avoid implementing this flow.

apagac:
Now also catching NameError in Snapshot.exists.

apagac:
PRT is now failing on unrelated test and I can't even reproduce it on my setup.

I recommend we merge this even with the Snapshot.refresh method, and when the widgetastic conversion gets merged, I can refactor.

mshriver:
Confirmed PRT failure on 57z in run 15451 isn't in this test.

PR #5003, opened 2017-07-19
[1LP][RFR] test vm state after revert snapshot

apagac:
New automation for snapshot area. Tests check the state of a VM after snapshot revert.

 - when reverting from snapshot with memory created on running VM, state should be powered on
 - when reverting from snapshot without memory created on running VM, state should be powered off
 - when reverting from snapshot created on stopped VM, state should be powered off

PR #5126, opened 2017-08-09
[1LP][RFR] modify snapshot testing

apagac:
This modifies some core components used in snapshot testing. Features:

 - testing VM gets always deployed with a different name [1]
 - able to choose a template for testing VM [2]
 - able to choose to create a snapshot with/without memory

[1] This is needed due to a known (?) issue with pytest scopes. Solves the intermittent error that sometimes appears when destroying and then again deploying a VM with the same name.

[2] Some of the snapshot tests need a testing VM that has network configured, sshd running and automation can connect to it. This is not possible when the VM is deployed from the smallest template. On the other hand, most of the snapshot tests do not need bigger template, so it makes no sense to run them all with default template. Thus, we can now choose what template to use in each test.

### DIFF
--- a/cfme/tests/infrastructure/test_snapshot.py
+++ b/cfme/tests/infrastructure/test_snapshot.py
@@ -13,6 +13,7 @@ from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.infrastructure.virtual_machines import Vm  # For Vm.Snapshot
 from utils import testgen
 from utils.conf import credentials
+from utils.generators import random_vm_name
 from utils.log import logger
 from utils.path import data_path
 from utils.ssh import SSHClient
@@ -30,21 +31,18 @@ pytest_generate_tests = testgen.generate([InfraProvider], scope="module")
 
 
 @pytest.fixture(scope="module")
-def vm_name():
-    return "test_snpsht_" + fauxfactory.gen_alphanumeric()
-
-
-@pytest.fixture(scope="module")
 def domain(request):
     dom = DomainCollection().create(name=fauxfactory.gen_alpha(), enabled=True)
     request.addfinalizer(dom.delete_if_exists)
     return dom
 
 
-@pytest.fixture(scope="module")
-def test_vm(setup_provider_modscope, provider, vm_name, request):
-    """Fixture to provision appliance to the provider being tested if necessary"""
-    vm = VM.factory(vm_name, provider, template_name=provider.data['full_template']['name'])
+def provision_vm(provider, request, small_template=True):
+    vm_name = random_vm_name(context="snapshot")
+    if small_template:
+        vm = VM.factory(vm_name, provider, template_name=provider.data['small_template'])
+    else:
+        vm = VM.factory(vm_name, provider, template_name=provider.data['full_template']['name'])
 
     if not provider.mgmt.does_vm_exist(vm_name):
         vm.create_on_provider(find_in_cfme=True, allow_skip="default")
@@ -52,16 +50,26 @@ def test_vm(setup_provider_modscope, provider, vm_name, request):
     return vm
 
 
-def new_snapshot(test_vm, has_name=True):
+@pytest.fixture(scope="module")
+def small_test_vm(setup_provider_modscope, provider, request):
+    return provision_vm(provider, request)
+
+
+@pytest.fixture(scope="module")
+def full_test_vm(setup_provider_modscope, provider, request):
+    return provision_vm(provider, request, small_template=False)
+
+
+def new_snapshot(test_vm, has_name=True, memory=False):
     if has_name:
         new_snapshot = Vm.Snapshot(
-            name="snpshot_" + fauxfactory.gen_alphanumeric(8),
-            description="snapshot", memory=False, parent_vm=test_vm
+            name="snpshot_{}".format(fauxfactory.gen_alphanumeric(8)),
+            description="snapshot", memory=memory, parent_vm=test_vm
         )
     else:
         new_snapshot = Vm.Snapshot(
-            description="snapshot_" + fauxfactory.gen_alphanumeric(8),
-            memory=False, parent_vm=test_vm
+            description="snapshot_{}".format(fauxfactory.gen_alphanumeric(8)),
+            memory=memory, parent_vm=test_vm
         )
     return new_snapshot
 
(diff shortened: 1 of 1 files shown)


cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -43,25 +45,31 @@ def domain(request):
...
+        if small_template:
+            vm = VM.factory(name, provider, template_name=provider.data['small_template'])
+        else:
+            vm = VM.factory(name, provider, template_name=provider.data['full_template']['name'])
+
+        if not provider.mgmt.does_vm_exist(name):
+            vm.create_on_provider(find_in_cfme=True, allow_skip="default")
+            request.addfinalizer(vm.delete_from_provider)
+        return vm
+    return _test_vm
```
quarckster: **REQUIRED**
I would make two separate fixtures: `small_test_vm` and `full_test_vm`:
```python
def provision_vm(request, name, provider, small_template=True):
    if small_template:
        vm = VM.factory(name, provider, template_name=provider.data['small_template'])
    else: 
        vm = VM.factory(name, provider, template_name=provider.data['full_template']['name'])
    if not provider.mgmt.does_vm_exist(name):
        vm.create_on_provider(find_in_cfme=True, allow_skip="default")
        request.addfinalizer(vm.delete_from_provider)
    return vm

def small_test_vm(request, setup_provider_modscope, provider, vm_name):
    return provision_vm(request, vm_name(), provider)

def full_test_vm(request, setup_provider_modscope, provider, vm_name):
    return provision_vm(request, vm_name(), provider, small_template=False)
```

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -41,40 +43,52 @@ def domain(request):
...
     not provider.one_of(RHEVMProvider, VMwareProvider) or
     (provider.one_of(RHEVMProvider) and provider.version < 4))
-def test_snapshot_crud(test_vm, provider):
+def test_snapshot_crud(small_test_vm, provider):
     """Tests snapshot crud
 
     Metadata:
         test_flag: snapshot, provision
     """
+    test_vm = small_test_vm
```
apagac: I can refactor if requested.

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -31,7 +31,9 @@
 
 @pytest.fixture(scope="module")
 def vm_name():
-    return "test_snpsht_" + fauxfactory.gen_alphanumeric()
+    def _vm_name():
+        return "test_snpsht_" + fauxfactory.gen_alphanumeric()
```
quarckster: **REQUIRED**
Please use `format`

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -31,7 +31,9 @@
 
 @pytest.fixture(scope="module")
 def vm_name():
-    return "test_snpsht_" + fauxfactory.gen_alphanumeric()
```
psav: **REQUIRED**: If I understand correctly, you just need to change the scope of the fixture to function to get a new vm name every time right?

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -41,40 +43,52 @@ def domain(request):
     return dom
 
 
-@pytest.fixture(scope="module")
-def test_vm(setup_provider_modscope, provider, vm_name, request):
-    """Fixture to provision appliance to the provider being tested if necessary"""
-    vm = VM.factory(vm_name, provider, template_name=provider.data['full_template']['name'])
+def provision_vm(provider, vm_name, request, small_template=True):
+    if small_template:
+        vm = VM.factory(vm_name, provider, template_name=provider.data['small_template'])
```
psav: ```small_template``` doesn't follow the same format and have a ```name``` attribute? That seems poor. (not blaming you for this). Would be nice to fix this up so it looks like
```python
template_source = 'small_template' if small_template else 'full_template':
vm = VM.factory(vm_name, provider, template_name=provider.data[template_source]['name'])
```

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -41,40 +43,52 @@ def domain(request):
...
+            memory=memory, parent_vm=test_vm
         )
     return new_snapshot
 
 
 @pytest.mark.uncollectif(lambda provider:
     not provider.one_of(RHEVMProvider, VMwareProvider) or
     (provider.one_of(RHEVMProvider) and provider.version < 4))
-def test_snapshot_crud(test_vm, provider):
+def test_snapshot_crud(small_test_vm, provider):
```
psav: This obviously won't get a new vm name each time, I assume this is OK?

cfme/tests/infrastructure/test_snapshot.py
psav: Would be nice...

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -100,23 +115,22 @@ def test_delete_all_snapshots(test_vm, provider):
...
+    test_vm = full_test_vm
     if provider.one_of(RHEVMProvider):
         snapshot1 = new_snapshot(test_vm, has_name=False)
     else:
         snapshot1 = new_snapshot(test_vm)
-
     ssh_kwargs = {
-        'hostname': snapshot1.vm.provider.mgmt.get_ip_address(snapshot1.vm.name),
         'username': credentials[provider.data['full_template']['creds']]['username'],
-        'password': credentials[provider.data['full_template']['creds']]['password']
```
psav: Any reason for the reorder?

cfme/tests/infrastructure/test_snapshot.py
apagac: I do need to change the scope, yes. I did try, it didn't work. What you see here is just a workaround, the underlying problem is a bit different.

There's a PR with new snapshot test #5003 . The test is parametrized and it's using module-scoped test_vm. Either I'm not getting the scopes at all, or it doesn't work, because instead of creating one VM for the test with all it's parametrization, it creates a new VM each time the test is run with different parameter. This means that the VM gets created, the test is run once with one parameter, then the VM is destroyed and immediately created again, the test is run the second time, and so on. Because of this, there were issues with multiple VMs in CFME with the same name, some of which were orphaned, and the automation was choosing the already destroyed one.

This of course isn't enough trouble, while running this I was hitting and issue with virtualcenter.delete_vm() https://github.com/ManageIQ/wrapanapi/issues/154 . Can't reproduce that locally, it does happen with PRT. Can't run PRT, because I need this PR to be merged first.

My solution is what you are looking at now. I turned my back on scopes, changed fixtures so that the VMs have always different name, problem went away.

cfme/tests/infrastructure/test_snapshot.py
apagac: This requires change in the structure of yaml file if I understand your request correctly. Probably affecting a lot of stuff. It is not my ambition to change lot of stuff in this PR, I just need to change bits of snapshot testing.

cfme/tests/infrastructure/test_snapshot.py
apagac: Yep, this works just fine.

cfme/tests/infrastructure/test_snapshot.py
apagac: Although I personally think this is a) faster and b) nicer, I did expect that I  would have to refactor :) On it.

cfme/tests/infrastructure/test_snapshot.py
apagac: None whatsoever. I thought dictionaries are unordered.

cfme/tests/infrastructure/test_snapshot.py
mshriver: Please use the existing `utils.generators.random_vm_name()` for this

cfme/tests/infrastructure/test_snapshot.py
mshriver: @psav @apagac There is inconsistency in the template naming throughout providers in cfme_data.yaml, which has been recorded in an issue, and that I have an open card to address. I'll be adding a 'templates' section to each provider, with consistent structure for `small_template`, `full_template` arrays that define `name` and `creds` for each template.

At that time I'll be auditing tests to update their access of the templates in the yaml, and can collapse this to a consistent access path at that time.

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -30,93 +31,98 @@
...
     request.addfinalizer(dom.delete_if_exists)
     return dom
 
 
-@pytest.fixture(scope="module")
-def test_vm(setup_provider_modscope, provider, vm_name, request):
-    """Fixture to provision appliance to the provider being tested if necessary"""
-    vm = VM.factory(vm_name, provider, template_name=provider.data['full_template']['name'])
+def provision_vm(provider, request, small_template=True):
+    vm_name = random_vm_name(context="snapshot")
```
mshriver: :+1:

PR #5137, opened 2017-08-10
[1LP][RFR] RHV5 fixes and lingering PR review fixes

jamesooden:
- searching for canvas element is now using a wait_for loop and
  a try-catch to avoid having an initial miss in finding the
  canvas element not cause the test to fail.
- Fixed the exceptions that were being caught from the provider
  in vm_console.py.  Previously we had been catching the selenium
  exception, but now the provider code throws its own exception, so
  now vm_console.py catches that.
- Make running of HTML5 Console test be based on a test_flag,
  html5_console, in the cfme_data.yaml file rather than provider                                                           
  type.

The problem with failing on the first time you access a VM console on a provider failing because we never see the status element (really its having a problem connecting to the websocket) still exists, so PRT may very well fail.

### DIFF
--- a/cfme/common/vm_console.py
+++ b/cfme/common/vm_console.py
@@ -7,13 +7,13 @@ import base64
 import re
 import tempfile
 
+from cfme.exceptions import ItemNotFound
 from PIL import Image, ImageFilter
 from pytesseract import image_to_string
+from selenium.webdriver.common.keys import Keys
 from utils.log import logger
 from utils.pretty import Pretty
 from wait_for import wait_for, TimedOutError
-from selenium.common.exceptions import NoSuchElementException
-from selenium.webdriver.common.keys import Keys
 
 
 class VMConsole(Pretty):
@@ -39,7 +39,7 @@ class VMConsole(Pretty):
         # due to an exception being thrown.
         try:
             text = self.provider.get_console_connection_status()
-        except NoSuchElementException:
+        except ItemNotFound:
             logger.exception('Could not find banner element.')
             return None
         finally:
@@ -48,7 +48,7 @@ class VMConsole(Pretty):
         logger.info('Read following text from console banner: %s', text)
         return text
 
-    def get_screen(self):
+    def get_screen(self, timeout=15):
         """
         Retrieve the bit map from the canvas widget that represents the console screen.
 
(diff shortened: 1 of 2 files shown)


dajoRH:
Lint report for commit caf990b244972a1dcb76c7a2c32ec584ce626945:

`cfme/common/vm_console.py`:
- :bangbang: Line 52:5: **E303** *too many blank lines (2)*

**No commit flaws detected**.

*CFME QE Bot*

jamesooden:
The most recent update of the commit is now failing as expected (i.e. where the first time it uses a provider on an appliance it fails to see the status element, and never gets connected status).   The first commit was failing due to a completely different problem that was fixed today, and I just merged in.

cfme/common/vm_console.py
```diff
@@ -62,10 +62,27 @@ def get_screen(self):
...
+                logger.info('Exception: {}'.format(e))
+                return False
+
+            return canvas
+
         self.switch_to_console()
 
         # Get the canvas element
-        canvas = self.provider.get_remote_console_canvas()
+        canvas, wait = wait_for(func=lambda: _get_canvas_element(self.provider),
```
mshriver: drop lambda, pass func_args:
https://github.com/RedHatQE/wait_for/blob/master/wait_for/__init__.py#L92

cfme/common/vm_console.py
```diff
@@ -62,10 +62,27 @@ def get_screen(self):
...
         """
+        # Internal function to use in wait_for().   We need to try to get the
+        # canvas element within a try-catch, that is in within a wait_for() so
+        # we can handle it not showing up right away as it is wont to do on
+        # at least RHV providers.
+        def _get_canvas_element(provider):
+            try:
+                canvas = provider.get_remote_console_canvas()
+            except ItemNotFound as e:
+                logger.info('Could not find canvas element.')
```
mshriver: If you use logger.exception here, you don't have to capture the exception (` as e`), and it'll automatically include the exception information with your given log message

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -14,12 +15,14 @@
...
 from utils.conf import credentials
+from utils.providers import ProviderFilter
 from wait_for import wait_for
 
 pytestmark = pytest.mark.usefixtures('setup_provider')
 
 pytest_generate_tests = testgen.generate(
-    [OpenStackProvider, VMwareProvider],
+    gen_func=testgen.providers,
+    filters=[ProviderFilter(classes=[CloudInfraProvider], required_flags=['html5_console'])],
```
mshriver: :+1:

cfme/common/vm_console.py
jamesooden: ✔️

cfme/common/vm_console.py
jamesooden: ✔️

PR #5146, opened 2017-08-11
[RFR] Update template_upload_rhevm, use block_upload

mshriver:
Added yaml key to mgmt_system[provider]['template_upload'] to allow for
YAML to specify blocking upload to rhevm providers


mshriver:
Ninja-merging this, tested locally and is breaking our test env providers at the moment. Does not impact the framework or tests.

PR #5157, opened 2017-08-15
[1LP][READTOMERGE] utils/providers.py Fix for test flags.

slaykovsky:
Now test flags are expecting string type in CFME yaml, but unfortunately it could be list in some environments. 

In this patch I've just added a check for a type so tests will not fail at string splitting and will proceed successfully in environments where test flags are defined as list.

Thanks.

### DIFF
--- a/utils/providers.py
+++ b/utils/providers.py
@@ -128,12 +128,12 @@ class ProviderFilter(object):
             test_flags = [flag.strip() for flag in self.required_flags]
 
             defined_flags = conf.cfme_data.get('test_flags', '')
-            if isinstance(defined_flags, str):
+            if isinstance(defined_flags, six.string_types):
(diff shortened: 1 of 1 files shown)


utils/providers.py
```diff
@@ -127,10 +127,14 @@ def _filter_required_flags(self, provider):
         if self.required_flags:
             test_flags = [flag.strip() for flag in self.required_flags]
 
-            defined_flags = conf.cfme_data.get('test_flags', '').split(',')
+            defined_flags = conf.cfme_data.get('test_flags', '')
+            if isinstance(defined_flags, str):
```
quarckster: **REQUIRED**
```python
import six
...
if isinstance(defined_flags, six.string_types):
...
```

slaykovsky: Done.

PR #5162, opened 2017-08-16
[RFR] Fixed PaginationPane wrapper

quarckster:
Purpose
=================

PaginationPane wrapper always returns VersionPick object, also wrapper doesn't have any arguments.

### DIFF
--- a/widgetastic_manageiq/__init__.py
+++ b/widgetastic_manageiq/__init__.py
@@ -1356,11 +1356,13 @@ class NonJSPaginationPane(View):
                                          .format(kwargs, table))
 
 
-def PaginationPane():  # noqa
-    return VersionPick({
-        Version.lowest(): NonJSPaginationPane(),
-        '5.9': JSPaginationPane(),
+def PaginationPane(*args, **kwargs):  # noqa
+    parent = kwargs.get("parent")
+    verpick_obj = VersionPick({
+        Version.lowest(): NonJSPaginationPane(*args, **kwargs),
+        '5.9': JSPaginationPane(*args, **kwargs),
     })
+    return verpick_obj.pick(parent.browser.product_version) if parent else verpick_obj
 
 
 class Stepper(View):


widgetastic_manageiq/__init__.py
```diff
@@ -1356,11 +1356,12 @@ def find_row_on_pages(self, table, *args, **kwargs):
                                          .format(kwargs, table))
 
 
-def PaginationPane():  # noqa
+def PaginationPane(*args, **kwargs):  # noqa
```
izapolsk: we don't use PaginationPane directly. It is always included into some view. 
VersionPick will be correctly resolved by widgetastic. So, it will become widget and will get parent using ordinary way.

You change makes sense only in one case when we want to use PaginationPage directly from ipython. I'll merge your changes, If you think we may have such use case in future.

izapolsk: oh your right. forgive me my forgetfulness. I've just added Pagination pane to web_ui.
So, it really makes sense.

widgetastic_manageiq/__init__.py
```diff
@@ -1356,11 +1356,13 @@ def find_row_on_pages(self, table, *args, **kwargs):
...
-    return VersionPick({
-        Version.lowest(): NonJSPaginationPane(),
-        '5.9': JSPaginationPane(),
+def PaginationPane(*args, **kwargs):  # noqa
+    parent = kwargs.get("parent")
+    verpick_obj = VersionPick({
+        Version.lowest(): NonJSPaginationPane(*args, **kwargs),
+        '5.9': JSPaginationPane(*args, **kwargs),
     })
+    return verpick_obj.pick(parent.product_version) if parent else verpick_obj
```
mfalesni: `parent.browser.product_version`

izapolsk: I managed to get product_version in any view as well

mfalesni: Of course, if you have a simple view that is not nested in any view then parent will always be a browser. But I would like it to be universal :)

izapolsk:
I tested it locally. didn't run into any issues. merging.

PR #5168, opened 2017-08-16
[1LP][RFR] Adding extended timeout to avoid RHV Console failure due to TimedOutError

kedark3:
Purpose or Intent
=================
__Extending__ Timeout for the wait_for_connect() function call because RHV HTML5 console takes longer time to open on first launch. OSP does not require that much time and it was working with default (30 sec) timeout with the function as well. This PR is connected to **RHCFQE-3577** on our sprint board. See comments on that card for additional details. Also moved the fixture configure_websocket before vm_obj, so that there is more time between Enabling Websocket Role and Opening of Remote Console.

{{ pytest: cfme/tests/cloud_infra_common/test_html5_vm_console.py --use-provider vsphere6-nested --use-provider default }}

### DIFF
--- a/cfme/tests/cloud_infra_common/test_html5_vm_console.py
+++ b/cfme/tests/cloud_infra_common/test_html5_vm_console.py
@@ -86,7 +86,7 @@ def configure_websocket():
 
 
 @pytest.mark.uncollectif(lambda: version.current_version() < '5.8', reason='Only valid for >= 5.8')
-def test_html5_vm_console(appliance, provider, vm_obj, configure_websocket,
+def test_html5_vm_console(appliance, provider, configure_websocket, vm_obj,
         configure_vmware_console_for_test):
     """
     Test the HTML5 console support for a particular provider.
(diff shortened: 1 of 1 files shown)


cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -191,3 +191,4 @@ def _validate_login():
             assert command_result
     finally:
         vm_console.close_console_window()
+        appliance.server.logout()
```
psav: Why the logout here?

kedark3: Hey @psav, logout was required because when running the Test back 2 back against RHV and VMware Providers, I would face the following issue:
If I just finished test for RHV, code would proceed to adding VMware Provider and once it is added, when it will navigate to Infrastructure -> Virtual Machines Page, I would get "Page Does not Exists" Error, because the browser will try to go to the VM details page of RHV VM which is already deleted and test would fail.Logging out would get rid of this issue.

mfalesni: Can you mention that in the code?

dajoRH:
*CFME QE Bot*

kedark3:
downstram-58z failing on ToggleFullScreen button test, which was merged in #5181  and it is a known bug in bugzilla.

PR #5171, opened 2017-08-17
[1LP][WIPTEST] Integrate wrapanapi containers

gshefer:
{{ pytest: cfme/tests/containers/test_labels.py --use-provider cm-env2 }}

- Added mgmt property to containers objects for API access.
- Getting bearer token and use it for the mgmt system.
- + get_bearer_token function  …
- Match Labalable class to use mgmt
- Removed redundant exceptions

### DIFF
--- a/cfme/common/provider.py
+++ b/cfme/common/provider.py
@@ -127,11 +127,19 @@ class BaseProvider(Taggable, Updateable, SummaryMixin, Navigatable):
         """
         # gotta stash this in here to prevent circular imports
         from utils.providers import get_mgmt
+        from cfme.containers.provider.openshift import OpenshiftProvider
+
+        if isinstance(self, OpenshiftProvider):
+            credentials = conf.cfme_data['management_systems'][self.key]['endpoints']['default']['credentials']
+            credentials = conf.credentials[credentials]
+            credentials['token'] = self.cli.run_command('oc whoami -t').output
+        else:
+            credentials = None
 
         if self.key:
-            return get_mgmt(self.key)
+            return get_mgmt(self.key, credentials=credentials)
         elif getattr(self, 'provider_data', None):
-            return get_mgmt(self.provider_data)
+            return get_mgmt(self.provider_data, credentials=credentials)
         else:
             raise ProviderHasNoKey(
                 'Provider {} has no key, so cannot get mgmt system'.format(self.name))
--- a/cfme/common/provider.py
+++ b/cfme/common/provider.py
@@ -127,19 +127,11 @@ class BaseProvider(Taggable, Updateable, SummaryMixin, Navigatable):
         """
         # gotta stash this in here to prevent circular imports
         from utils.providers import get_mgmt
-        from cfme.containers.provider.openshift import OpenshiftProvider
-
-        if isinstance(self, OpenshiftProvider):
-            credentials = conf.cfme_data['management_systems'][self.key]['endpoints']['default']['credentials']
-            credentials = conf.credentials[credentials]
-            credentials['token'] = self.cli.run_command('oc whoami -t').output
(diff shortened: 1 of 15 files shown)


gshefer:
@psav @mshriver @RonnyPfannschmidt - for some reason although test_labels_remove is marked with blocker the test is not being skipped. maybe because the status is "POST"? anyway it should be skipped.
You are welcome to review as well :-)

cfme/common/provider.py
```diff
@@ -127,11 +127,20 @@ def get_mgmt_system(self):
...
         from utils.providers import get_mgmt
+        from cfme.containers.provider.openshift import OpenshiftProvider
+
+        if isinstance(self, OpenshiftProvider):
+            credentials = conf.cfme_data['management_systems'].get(
+                self.key)['endpoints']['default']['credentials']
+            credentials = conf.credentials[credentials]
+            credentials['token'] = self.get_bearer_token()
+        else:
+            credentials = None
```
quarckster: **REQUIRED**
Please override `get_mgmt_system()` method in `OpenshiftProvider`.

cfme/containers/provider/__init__.py
```diff
@@ -601,24 +540,15 @@ def remove_label(self, name, silent_failure=False):
...
             :py:class:`LabelNotFoundException`.
         """
-        json_content = self._get_json()
-        if name not in json_content['metadata'].get('labels', {}).keys():
-            failure_signature = 'Could not find label "{}", labels: {}' \
-                .format(name, json_content['metadata']['labels'])
+        try:
+            self.mgmt.delete_label(name)
+            return True
+        except:
```
quarckster: **REQUIRED**
It should be at least `except Exception:` but it's better to catch the exact exception.

gshefer: Unfortunately the exact exception is Exception - so I'll change to that.

cfme/containers/provider/openshift.py
```diff
@@ -70,6 +72,37 @@ def __init__(self, name=None, key=None, zone=None,
...
+        from utils.providers import get_mgmt
+
+        credentials = conf.cfme_data['management_systems'].get(
+            self.key)['endpoints']['default']['credentials']
+        credentials = conf.credentials[credentials]
+        credentials['token'] = self.get_bearer_token()
+
+        if self.key:
+            return get_mgmt(self.key, credentials=credentials)
+        elif getattr(self, 'provider_data', None):
```
Probably you would want:
```python
elif hasattr(self, 'provider_data'):
```

cfme/containers/provider/openshift.py
```diff
@@ -70,6 +72,37 @@ def __init__(self, name=None, key=None, zone=None,
...
+        credentials['token'] = self.get_bearer_token()
+
+        if self.key:
+            return get_mgmt(self.key, credentials=credentials)
+        elif hasattr(self, 'provider_data'):
+            return get_mgmt(self.provider_data, credentials=credentials)
+        else:
+            raise ProviderHasNoKey(
+                'Provider {} has no key, so cannot get mgmt system'.format(self.name))
+
```
jkrocil: Can you please explain what is going on here? You can use `prov.endpoints['default'].token` to get token and `prov.mgmt` to get management system. Why do we need these two methods here?

gshefer: Since for the API we should use the Bearer token and not the full.

dajoRH:
Lint report for commit 7b773b05b97b2db457f51e0aedd13ecfaf43b1d3:

`cfme/containers/provider/openshift.py`:
- :red_circle: Line 4:1: **F401** *'utils.conf' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

PR #5191, opened 2017-08-18
[1LP][RFR] Making sure 'touch blather' command is sent to console correctly for RHOS and RHV

kedark3:
Purpose or Intent
=================

__Fixing__ File creation command 'touch blather' to make sure that RHV and RHOS HTML5 Remote consoles are receiving all characters correctly, otherwise sometimes RHOS and RHV Providers drop some characters resulting in the file not being created.

{{ pytest: cfme/tests/cloud_infra_common/test_html5_vm_console.py --use-provider vsphere6-nested --use-provider default }}

### DIFF
--- a/cfme/common/vm_console.py
+++ b/cfme/common/vm_console.py
@@ -6,6 +6,7 @@ Module containing classes with common behaviour for consoles of both VMs and Ins
 import base64
 import re
 import tempfile
+import time
 
 from cfme.exceptions import ItemNotFound
 from PIL import Image, ImageFilter
@@ -140,9 +141,13 @@ class VMConsole(Pretty):
         """Send text to the console."""
         self.switch_to_console()
         canvas = self.provider.get_remote_console_canvas()
-        canvas.send_keys(text)
-        canvas.send_keys(Keys.ENTER)
         logger.info("Sending following Keys to Console {}".format(text))
+        for character in text:
+            canvas.send_keys(character)
+            # time.sleep() is used as a short delay between two keystrokes.
+            # If keys are sent to canvas any faster, canvas fails to receive them.
+            time.sleep(0.3)
+        canvas.send_keys(Keys.ENTER)
         self.switch_to_appliance()
 
     def send_ctrl_alt_delete(self):
--- a/cfme/tests/cloud_infra_common/test_html5_vm_console.py
+++ b/cfme/tests/cloud_infra_common/test_html5_vm_console.py
@@ -171,9 +171,7 @@ def test_html5_vm_console(appliance, provider, configure_websocket, vm_obj,
             time.sleep(15)
 
         # create file on system
-        vm_console.send_keys("touch blather\n")
-        logger.info("Wait 5 sec for file creation")
-        time.sleep(5)  # Wait for file creation
+        vm_console.send_keys("touch blather")
 
         # Test pressing ctrl-alt-delete...we should be able to get a new login prompt:
         vm_console.send_ctrl_alt_delete()
@@ -191,7 +189,7 @@ def test_html5_vm_console(appliance, provider, configure_websocket, vm_obj,
             # we will get instance of SSHResult
             # Sometimes Openstack drops characters from word 'blather' hence try to remove
             # file using partial file name. Known issue, being worked on.
-            command_result = ssh_client.run_command("rm bla*", ensure_user=True)
+            command_result = ssh_client.run_command("rm blather", ensure_user=True)
             assert command_result
     finally:
         vm_console.close_console_window()


kedark3:
Currently failing due to errors that should be fixed by merging #5168

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -171,9 +171,15 @@ def _validate_login():
...
-        time.sleep(5)  # Wait for file creation
+        def _touch_file():
+            vm_console.send_keys("\n\ntouch blather\n")
+            logger.info("Wait 5 sec for file creation")
+            time.sleep(5)  # Wait for file creation
+            logger.info("vm_console.find_text_on_screen(text_to_find='touch blather') is True {}"
+                .format(vm_console.find_text_on_screen(text_to_find='touch blather') is True))
+            return vm_console.find_text_on_screen(text_to_find='touch blather') is True
+
+        wait_for(func=_touch_file, timeout=180, delay=5)
```
mshriver: you're waiting 5s for the file creation within `_touch_file`, I would drop this delay to .1

cfme/tests/cloud_infra_common/test_html5_vm_console.py
```diff
@@ -171,9 +171,15 @@ def _validate_login():
             time.sleep(15)
 
         # create file on system
-        vm_console.send_keys("touch blather\n")
-        logger.info("Wait 5 sec for file creation")
-        time.sleep(5)  # Wait for file creation
+        def _touch_file():
+            vm_console.send_keys("\n\ntouch blather\n")
```
mshriver: check whether or not this was sent before waiting 5s

kedark3: I think I should reword the log entry, actually I am not waiting for the file creation, but giving time for send_keys to do the work and then run the check using find_text_on_screen, because if we take image of console too early, it might not see all characters being sent, even if they were sent and received correctly by the console.

kedark3:
Tests are failing for ToggleFullScreen button which is a known bug.

kedark3:
@mshriver Even-though we discussed the elegant solutions, I did some more testing and found out that it wasn't the console dropping the characters, but it was selenium webdriver which was sending the characters too fast. Slowing it down with slight delay made the send_keys completely reliable. I tested multiple times, and it did not drop characters even once. Please review and comment.

cfme/common/vm_console.py
```diff
@@ -140,9 +141,11 @@ def send_keys(self, text):
         """Send text to the console."""
         self.switch_to_console()
         canvas = self.provider.get_remote_console_canvas()
-        canvas.send_keys(text)
-        canvas.send_keys(Keys.ENTER)
         logger.info("Sending following Keys to Console {}".format(text))
+        for character in text:
+            canvas.send_keys(character)
+            time.sleep(0.3)
```
quarckster: **REQUIRED**
Please use `utils.wait.wait_for` instead `time.sleep`.

cfme/common/vm_console.py
```diff
@@ -140,9 +140,11 @@ def send_keys(self, text):
         """Send text to the console."""
         self.switch_to_console()
         canvas = self.provider.get_remote_console_canvas()
-        canvas.send_keys(text)
-        canvas.send_keys(Keys.ENTER)
         logger.info("Sending following Keys to Console {}".format(text))
+        for character in text:
+            canvas.send_keys(character)
+            wait_for(lambda: False, delay=0.3, num_sec=0.3, silent_failure=True)
```
mfalesni: `wait_for` makes sense only when there is a condition to check, otherwise time.sleep is a perfectly valid solution. We usually have something to check for. Here, it probably isn't since it is sending keys to a canvas, so you can revert to using time.sleep, with a comment why it is used here, since the condition in wait_for is usually self-explaining. time.sleep with a comment why the sleep is needed will be much more readable than wait for false.

I know @quarckster suggested using wait_for, we do that when we see time.sleep, but we then sometimes have people come back and tell us there is nothing specific to check, so we likely do allow the use then.

kedark3: @mfalesni , okay. Yes I did replace time.sleep because @quarckster  suggested it, but what you are saying is perfectly making sense. I will change it back to original. Thanks for review.

cfme/common/vm_console.py
```diff
@@ -140,9 +141,13 @@ def send_keys(self, text):
         """Send text to the console."""
         self.switch_to_console()
         canvas = self.provider.get_remote_console_canvas()
-        canvas.send_keys(text)
-        canvas.send_keys(Keys.ENTER)
         logger.info("Sending following Keys to Console {}".format(text))
+        for character in text:
```
mshriver: :+1:

PR #5192, opened 2017-08-19
[1LP][RFR] Adding NEW TEST: VMware WebMKS Console

kedark3:
Purpose or Intent
=================

__Adding tests__ for feature VMware WebMKS Console which should work with vsphere6 and vsphere6.5.

{{ pytest: cfme/tests/infrastructure/test_webmks_console.py --use-provider vsphere6-nested --use-provider vsphere65-nested}}

### DIFF
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -374,7 +374,8 @@ class BaseVM(Pretty, Updateable, PolicyProfileAssignable, Taggable, SummaryMixin
         view = navigate_to(self, 'Details')
 
         # Click console button given by type
-        view.toolbar.access.item_select(console)
+        view.toolbar.access.item_select(console, handle_alert=None
+            if invokes_alert is False else True)
 
         # Get the consoles window handle, and then create a VMConsole object, and store
         # the VMConsole object aside.
--- a/cfme/tests/infrastructure/test_webmks_console.py
+++ b/cfme/tests/infrastructure/test_webmks_console.py
@@ -1,0 +1,152 @@
+# -*- coding: utf-8 -*-
+"""Test for WebMKS Remote Consoles of VMware Providers."""
+import imghdr
+import pytest
+import re
+
+from cfme.common.vm import VM
+from cfme.configure.configuration import VMwareConsoleSupport
+from cfme.infrastructure.provider import InfraProvider
+from utils import testgen, version, ssh
+from utils.appliance.implementations.ui import navigate_to
+from utils.conf import credentials
+from utils.log import logger
+from utils.providers import ProviderFilter
+from wait_for import wait_for
+
+pytestmark = pytest.mark.usefixtures('setup_provider')
+
+pytest_generate_tests = testgen.generate(
+    gen_func=testgen.providers,
+    filters=[ProviderFilter(classes=[InfraProvider], required_flags=['webmks_console'])],
+    scope='module'
+)
+
+
+@pytest.yield_fixture(scope="function")
+def vm_obj(request, provider, setup_provider, console_template, vm_name):
+    """VM creation/deletion fixture.
+
+    Create a VM on the provider with the given template, and return the vm_obj.
+    Also, remove VM from provider using nested function _delete_vm
+    after the test is completed.
+    """
+    vm_obj = VM.factory(vm_name, provider, template_name=console_template)
+    vm_obj.create_on_provider(timeout=2400, find_in_cfme=True, allow_skip="default")
+    yield vm_obj
+    try:
+        vm_obj.delete_from_provider()
+    except Exception:
+        logger.warning("Failed to delete vm `{}`.".format(vm_obj.name))
+
+
+@pytest.fixture(scope="module")
+def configure_vmware_console_for_test(appliance, provider):
+    """Configure VMware Console to use VNC which is what is required for the HTML5 console."""
+    navigate_to(appliance.server, 'Server')
+
+    settings_pg = VMwareConsoleSupport(
+        appliance=appliance,
+        console_type='VMware WebMKS',
+    )
+    settings_pg.update()
+
+
+@pytest.mark.uncollectif(lambda: version.current_version() < '5.8', reason='Only valid for >= 5.8')
+def test_webmks_vm_console(request, appliance, provider, vm_obj,
+        configure_vmware_console_for_test):
+    """Test the VMware WebMKS console support for a particular provider.
+
+    The supported providers are:
+
+        VMware vSphere6 and vSphere6.5
+
+    For a given provider, and a given VM, the console will be opened, and then:
+
+        - The console's status will be checked.
+        - A command that creates a file will be sent through the console.
+        - Using ssh we will check that the command worked (i.e. that the file
+          was created.)
+    """
+    console_vm_username = credentials[provider.data.templates.get('console_template')
+                            ['creds']].get('username')
+    console_vm_password = credentials[provider.data.templates.get('console_template')
+                            ['creds']].get('password')
+
+    vm_obj.open_console(console='VM Console', invokes_alert=True)
+    assert vm_obj.vm_console, 'VMConsole object should be created'
+    vm_console = vm_obj.vm_console
+
+    request.addfinalizer(vm_console.close_console_window)
+    request.addfinalizer(appliance.server.logout)
+
+    # Get the login screen image, and make sure it is a jpeg file:
+    screen = vm_console.get_screen(180)
+    assert imghdr.what('', screen) == 'jpeg'
+
+    with ssh.SSHClient(hostname=vm_obj.ip_address, username=console_vm_username,
+            password=console_vm_password) as vm_ssh_client:
+        assert vm_console.wait_for_text(text_to_find="login:", timeout=200),\
+            "VM Console didn't prompt for Login"
+
+        result_before_login = vm_ssh_client.run_command("who --count", ensure_user=True)
+        # Enter Username:
+        vm_console.send_keys(console_vm_username)
+
+        assert vm_console.wait_for_text(text_to_find="Password", timeout=200),\
+            "VM Console didn't prompt for Password"
+        # Enter Password:
+        vm_console.send_keys("{}\n".format(console_vm_password))
+
+        result_after_login = vm_ssh_client.run_command("who --count", ensure_user=True)
+        # Number of users before login would be 0 and after login would be 180
+        # If below assertion would fail result_after_login is also 0, denoting login failed
+        assert (result_before_login.output.split('=')[-1].strip() <
+             result_after_login.output.split('=')[-1].strip()), "Login Failed"
+
+    # This regex can find if there is a word 'login','password','incorrect' present in
+    # text, irrespective of its case
+    regex_for_login_password = re.compile(r'\blogin\b | \bpassword\b| \bincorrect\b',
+     flags=re.I | re.X)
+
+    def _validate_login():
+        """
+        Try to read what is on present on the last line in console.
+
+        If it is word 'login', enter username, if 'password' enter password, in order
+        to make the login successful
+        """
+        if vm_console.find_text_on_screen(text_to_find='login', current_line=True):
+            vm_console.send_keys(console_vm_username)
+
+        if vm_console.find_text_on_screen(text_to_find='Password', current_line=True):
+            vm_console.send_keys("{}\n".format(console_vm_password))
+        # if the login attempt failed for some reason (happens with RHOS-cirros),
+        # last line of the console will contain one of the following words:
+        # [login, password, incorrect]
+        # if so, regex_for_login_password will find it and result will not be []
+        # .split('\n')[-1] splits the console text on '\n' & picks last item of resulting list
+        result = regex_for_login_password.findall(vm_console.get_screen_text().split('\n')[-1])
+        return result == []
+
+    # if _validate_login() returns True, it means we did not find any of words
+    # [login, password, incorrect] on last line of console text, which implies login success
+    wait_for(func=_validate_login, timeout=300, delay=5)
+
+    logger.info("Wait to get the '$' prompt")
+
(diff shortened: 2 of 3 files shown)


cfme/tests/infrastructure/test_webmks_console.py
```diff
@@ -0,0 +1,178 @@
...
+
+        - The console's status will be checked.
+        - A command that creates a file will be sent through the console.
+        - Using ssh we will check that the command worked (i.e. that the file
+          was created.
+    """
+    console_vm_username = credentials[provider.data.templates.get('console_template')
+                            ['creds']].get('username')
+    console_vm_password = credentials[provider.data.templates.get('console_template')
+                            ['creds']].get('password')
```
mkoura: Is the 'console_template' key guaranteed to be always there? This will fail with TypeError if not.

cfme/tests/infrastructure/test_webmks_console.py
```diff
@@ -0,0 +1,178 @@
...
+
+    @request.addfinalizer
+    def _delete_vm():
+        try:
+            vm_obj.delete_from_provider()
+        except Exception:
+            logger.warning("Failed to delete vm `{}`.".format(vm_obj.name))
+
+    vm_obj.create_on_provider(timeout=2400, find_in_cfme=True, allow_skip="default")
+    return vm_obj
```
quarckster: **OPTIONAL**
You could make yield_fixture instead.

cfme/tests/infrastructure/test_webmks_console.py
```diff
@@ -0,0 +1,178 @@
...
+    navigate_to(appliance.server, 'Server')
+
+    settings_pg = VMwareConsoleSupport(
+        appliance=appliance,
+        console_type='VMware WebMKS',
+    )
+    settings_pg.update()
+
+
+@pytest.fixture(scope="session")
```
quarckster: **REQUIRED**
```python
@pytest.yield_fixture(scope="session")
```

cfme/tests/infrastructure/test_webmks_console.py
```diff
@@ -0,0 +1,178 @@
...
+        with ssh.SSHClient(hostname=vm_obj.ip_address, username=console_vm_username,
+                password=console_vm_password) as ssh_client:
+            # if file was created in previous steps it will be removed here
+            # we will get instance of SSHResult
+            # Sometimes Openstack drops characters from word 'blather' hence try to remove
+            # file using partial file name. Known issue, being worked on.
+            command_result = ssh_client.run_command("rm blather", ensure_user=True)
+            assert command_result
+    finally:
+        vm_console.close_console_window()
```
mkoura: The file is more-or-less the same as cfme/tests/cloud_infra_common/test_html5_vm_console.py
What are the plans for refactoring wrt reusing shared functionality?

cfme/tests/infrastructure/test_webmks_console.py
```diff
@@ -0,0 +1,178 @@
...
+    """
+    console_vm_username = credentials[provider.data.templates.get('console_template')
+                            ['creds']].get('username')
+    console_vm_password = credentials[provider.data.templates.get('console_template')
+                            ['creds']].get('password')
+
+    vm_obj.open_console(console='VM Console', invokes_alert=True)
+    assert vm_obj.vm_console, 'VMConsole object should be created'
+    vm_console = vm_obj.vm_console
+    try:
```
quarckster: **REQUIRED**
Why is this required? You could just do:
```python
request.addfinalizer(vm_console.close_console_window)
```

cfme/tests/infrastructure/test_webmks_console.py
```diff
@@ -0,0 +1,178 @@
...
+    """
+    roles = configuration.get_server_roles()
+    if 'websocket' in roles and not roles['websocket']:
+        logger.info('Enabling the websocket role to allow console connections')
+        roles['websocket'] = True
+        configuration.set_server_roles(**roles)
+        yield
+    roles['websocket'] = False
+    logger.info('Disabling the websocket role to avoid intrusive popups')
+    configuration.set_server_roles(**roles)
```
kedark3: Yes, websocket will be disabled if it was already enabled before, because @psav and I had a discussion when I was creating these tests and I found out that provider setup fixture was disabling the websocket role, I asked pete if I could stop that from happening, he suggested me to create this yield fixture that would enable the websocket which is disabled by the provider fixture and then disable it back when my tests are complete.

cfme/tests/infrastructure/test_webmks_console.py
kedark3: Yes, console_template key is guaranteed to be there on all the providers which support these tests, and hence it should not fail. We have a "webmks_console" flag to filter which providers are supported in the cfme_data.yaml and those providers which are supported will always fulfill the requirement, unless somebody deletes that key from cfme_data.yaml.

cfme/tests/infrastructure/test_webmks_console.py
kedark3: I agree its similar file. But haven't given much thought to it. Primary aim is to get the test scenario covered in this prototype. And once this is done, I can proceed to reducing the duplication. Suggestions welcome. :smile:

cfme/tests/infrastructure/test_webmks_console.py
mkoura: But why not to revert it back to the state where it was before the fixture execution? Why always disabling it, even in situation when it was enabled?

mkoura: @kedark3 explained to me the reasons behind this in detail offline and I agreed than in current situation (other fixture is setting it to disabled) it's safer to not touch it during teardown.

cfme/tests/infrastructure/test_webmks_console.py
```diff
@@ -0,0 +1,169 @@
...
+
+    # Enter Username:
+    vm_console.send_keys(console_vm_username)
+
+    assert vm_console.wait_for_text(text_to_find="Password", timeout=200),\
+        "VM Console didn't prompt for Password"
+    # Enter Password:
+    vm_console.send_keys("{}\n".format(console_vm_password))
+
+    time.sleep(5)  # wait for login to complete
```
quarckster: Is there way how to know  when login is completed?

cfme/tests/infrastructure/test_webmks_console.py
```diff
@@ -0,0 +1,169 @@
...
+
+    logger.info("Wait to get the '$' prompt")
+
+    vm_console.wait_for_text(text_to_find=provider.data.templates.get('console_template')
+                    ['prompt_text'], timeout=200)
+
+    # create file on system
+    vm_console.send_keys("touch blather\n")
+    logger.info("Wait 5 sec for file creation")
+    time.sleep(5)  # Wait for file creation
```
quarckster: **REQUIRED**
This must be replaced by `wait_for`

cfme/tests/infrastructure/test_webmks_console.py
kedark3: @quarckster  There is not really a way to know this, time.sleep is needed because after I enter username and password, I need to wait at least a sec or two before I get either command prompt or "Login Incorrect" message. So I can not think of anything besides time.sleep in this situation. You might say I can wait_for _command prompt_ or _"Login Incorrect"_ message. But for Image to Text recognition finding out command prompt correctly is a challenge. And if I detect _"Login Incorrect"_ I will then need to attempt to login again. Hence I just replaced all that login with validate_login which validates if login went through and if not, attempt to re-login. Any suggestions to make it even better are welcome. Thanks :)

cfme/tests/infrastructure/test_webmks_console.py
kedark3: This is simple blind sleep before SSH Client can go in and try to remove the file. I want some buffer to make sure command had enough time to create a file. Would you still suggest changing this with something like 
`wait_for(vm_console.send_keys("touch blather"), num_sec=5, delay=5)` ?

quarckster: You should check if the file is created in the vm via ssh_client

cfme/tests/infrastructure/test_webmks_console.py
quarckster: I believe it can be checked via `ssh_client` and `who` command. I would investigate such approach.

cfme/tests/infrastructure/test_webmks_console.py
kedark3: I think removing the file (as done on  line 168 below) makes sure the file was created. That is the test here, i.e. if console is functioning correctly and capable of taking the input, file should be created(within couple of seconds) after I sent touch command, and when I run "rm blather" via SSH, if the console was functioning correctly, I will have the file present on VM and "rm blather" will work correctly. Otherwise it will fail to find the file and remove it causing failure. Making sure the file is created via ssh_client seems extra work which IMHO not required. Please provide your thoughts.

cfme/common/vm.py
```diff
@@ -374,7 +374,10 @@ def open_console(self, console='VM Console', invokes_alert=False, cancel=False):
         view = navigate_to(self, 'Details')
 
         # Click console button given by type
-        view.toolbar.access.item_select(console)
+        if invokes_alert:
```
psav: **REQUIRED**: Given this is ```False``` by default and I'm thinking ```True``` if not, why can't we just do

```view.toolbar.access.item_select(console, handle_alert=invokes+alert)```

File: cfme/tests/infrastructure/test_webmks_console.py [hunk 10]
```diff
@@ -0,0 +1,169 @@
...
+@pytest.fixture(scope="module")
+def configure_vmware_console_for_test(appliance, provider):
+    """Configure VMware Console to use VNC which is what is required for the HTML5 console."""
+    navigate_to(appliance.server, 'Server')
+
+    settings_pg = VMwareConsoleSupport(
+        appliance=appliance,
+        console_type='VMware WebMKS',
+    )
+    settings_pg.update()
```
psav: **REQUIRED**: Should we unset this afterwards?

cfme/tests/infrastructure/test_webmks_console.py
```diff
@@ -0,0 +1,169 @@
...
+
+@pytest.yield_fixture(scope="session")
+def configure_websocket():
+    """Enable websocket role if it is disabled.
+
+    Currently the fixture cfme/fixtures/base.py,
+    disables the websocket role to avoid intrusive popups. It is enabled
+    here before test begins and disabled during teardown.
+    """
+    roles = configuration.get_server_roles()
```
psav: **QUESTION**: Doesn't the set_roles check with the get_roles anyway? I think there is more code here than there needs to be

cfme/tests/infrastructure/test_webmks_console.py
psav: **QUESTION**: I was going to ask just what you mentioned @kedark3 about checking for the two strings. Can't we just wait_for the prompt and if it doesn't show up after a certain amount of time then we know it failed?

cfme/tests/infrastructure/test_webmks_console.py
psav: It is required, arbitrary sleeps are not immune to networking problems. The problem is without wait_for any sleep we put in is permanent and everytime we need to wait, we need to wait that long. Since boxes can be under load, or connections take a hwile to connect etc, I prefer the wait_for, I odn't think it's that much extra work, and meas we can retry every 10 seconds, whilst having a maximum timeout of 2minutes if we wanted. 99% of the time it'll be done within 10 seconds, but if it has to wait longer it's not big deal.

cfme/tests/infrastructure/test_webmks_console.py
```diff
@@ -0,0 +1,169 @@
...
+    vm_console.wait_for_text(text_to_find=provider.data.templates.get('console_template')
+                    ['prompt_text'], timeout=200)
+
+    # create file on system
+    vm_console.send_keys("touch blather\n")
+    logger.info("Wait 5 sec for file creation")
+    time.sleep(5)  # Wait for file creation
+
+    with ssh.SSHClient(hostname=vm_obj.ip_address, username=console_vm_username,
+            password=console_vm_password) as ssh_client:
```
psav: **REQUIRED**: let's call this ```vm_ssh_client``` to differentiate it from the normal ```ssh_client``` which is usually used for the appliance itself

cfme/tests/infrastructure/test_webmks_console.py
kedark3: Pete, unfortunately, I was not able to success tesseract to parse the image correctly and identify the prompt, but @quarckster's method seems like a good option. I think I should give it a try. Any inputs welcome.

cfme/tests/infrastructure/test_webmks_console.py
kedark3: Unset to what? It defaults to "VMware VMRC Plugin" when the appliance is created. Should I set it back to that? And why is it required?

kedark3:
Currently failing with "Create_on_provider" function since there is some change in the pagination with new build 5.8.2.0. http://10.16.45.124/y7ibwMNb/artifacts/cfme/tests/infrastructure/test_webmks_console.py/test_webmks_vm_console[vsphere65-nested]/filedump-traceback.log

cfme/tests/infrastructure/test_webmks_console.py
```diff
@@ -0,0 +1,152 @@
...
+        return result == []
+
+    # if _validate_login() returns True, it means we did not find any of words
+    # [login, password, incorrect] on last line of console text, which implies login success
+    wait_for(func=_validate_login, timeout=300, delay=5)
+
+    logger.info("Wait to get the '$' prompt")
+
+    vm_console.wait_for_text(text_to_find=provider.data.templates.get('console_template')
+                    ['prompt_text'], timeout=200)
```
mkoura: Strange indentation / line break.

PR #5193, opened 2017-08-19
[1LP][RFR] Service manual approval test

sshveta:
{{pytest: -v --long-running --use-provider vsphere55 cfme/tests/services/test_service_manual_approval.py }}

### DIFF
--- a/cfme/tests/services/test_service_manual_approval.py
+++ b/cfme/tests/services/test_service_manual_approval.py
@@ -1,0 +1,78 @@
+# -*- coding: utf-8 -*-
+import fauxfactory
+import pytest
+
+from cfme.common.provider import cleanup_vm
+from cfme.infrastructure.provider import InfraProvider
+from cfme.services.catalogs.service_catalogs import ServiceCatalogs
+from cfme.automate.explorer.domain import DomainCollection
+from cfme.services import requests
+from cfme import test_requirements
+from utils.log import logger
+from utils.update import update
+from utils import testgen
+
+
+pytestmark = [
+    pytest.mark.meta(server_roles="+automate"),
+    pytest.mark.usefixtures('vm_name', 'catalog_item', 'uses_infra_providers'),
+    test_requirements.service,
+    pytest.mark.long_running
+]
+
+
+pytest_generate_tests = testgen.generate([InfraProvider], required_fields=[
+    ['provisioning', 'template'],
+    ['provisioning', 'host'],
+    ['provisioning', 'datastore']
+], scope="module")
+
+
+@pytest.fixture(scope="function")
+def create_domain(request):
+    """Create new domain and copy instance from ManageIQ to this domain"""
+
+    dc = DomainCollection()
+    new_domain = dc.create(name=fauxfactory.gen_alphanumeric(), enabled=True)
+    request.addfinalizer(new_domain.delete_if_exists)
+    instance = dc.instantiate(name='ManageIQ')\
+        .namespaces.instantiate(name='Service')\
+        .namespaces.instantiate(name='Provisioning')\
+        .namespaces.instantiate(name='StateMachines')\
+        .classes.instantiate(name='ServiceProvisionRequestApproval')\
+        .instances.instantiate(name='Default')
+    instance.copy_to(new_domain)
+    return new_domain
+
+
+@pytest.fixture(scope="function")
+def modify_instance(create_domain):
+    """Modify the instance in new domain to change it to manual approval instead of auto"""
+
+    instance = create_domain.namespaces.instantiate(name='Service')\
+        .namespaces.instantiate(name='Provisioning')\
+        .namespaces.instantiate(name='StateMachines')\
+        .classes.instantiate(name='ServiceProvisionRequestApproval')\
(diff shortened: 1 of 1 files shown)


sshveta:
Tests not running on PRT

cfme/services/requests.py
```diff
@@ -180,7 +181,7 @@ def wait_for_request(cells, partial_check=False):
          The matching :py:class:`cfme.web_ui.Table.Row` if found, ``False`` otherwise.
     """
     row = find_request(cells, partial_check)
-    if row.request_state.text in REQUEST_FINISHED_STATES:
+    if row.request_state.text in REQUEST_FINISHED_STATES or REQUEST_PENDING_STATE:
```
quarckster: **REQUIRED**
It doesn't work like that :)
```python
if row.request_state.text in REQUEST_FINISHED_STATES or row.request_state.text in REQUEST_PENDING_STATE:
```
Bu I think it's a bit better will be:
```python
if row.request_state.text in list(REQUEST_PENDING_STATE) + list(REQUEST_FINISHED_STATES):
```

cfme/tests/services/test_service_manual_approval.py
```diff
@@ -0,0 +1,77 @@
...
+        .namespaces.instantiate(name='StateMachines')\
+        .classes.instantiate(name='ServiceProvisionRequestApproval')\
+        .instances.instantiate(name='Default')
+    instance.copy_to(domain)
+    return domain
+
+
+@pytest.fixture(scope="function")
+def modify_instance(domain):
+    dc = DomainCollection()
```
quarckster: **REQUIRED**
You already have `DomainCollection` in `domain` fixture. Use it to get required instance.

dajoRH:
Lint report for commit 6c3843807bf183b93a8024ae2c3c0a0ca8efe20b:

`cfme/tests/services/test_service_manual_approval.py`:
- :red_circle: Line 13:1: **F401** *'utils.wait.wait_for' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/services/test_service_manual_approval.py
```diff
@@ -0,0 +1,75 @@
...
+
+@pytest.fixture(scope="function")
+def modify_instance(domain):
+    dc = domain.parent
+    instance = dc.instantiate(name=domain.name)\
+        .namespaces.instantiate(name='Service')\
+        .namespaces.instantiate(name='Provisioning')\
+        .namespaces.instantiate(name='StateMachines')\
+        .classes.instantiate(name='ServiceProvisionRequestApproval')\
+        .instances.instantiate(name='Default')
```
quarckster: I would implement all this stuff in `copy_to()`method. So when you make `instance.copy_to(domain)`, it would return you a new instance.

cfme/tests/services/test_service_manual_approval.py
```diff
@@ -0,0 +1,75 @@
...
+    instance.copy_to(domain)
+    return domain
+
+
+@pytest.fixture(scope="function")
+def modify_instance(domain):
+    dc = domain.parent
+    instance = dc.instantiate(name=domain.name)\
```
quarckster: **REQUIRED**
```python
instance = domain\
```

sshveta: Changed . 
Both instance are from different domain so one method cannot be used .

cfme/tests/services/test_service_manual_approval.py
```diff
@@ -0,0 +1,75 @@
...
+        .namespaces.instantiate(name='StateMachines')\
+        .classes.instantiate(name='ServiceProvisionRequestApproval')\
+        .instances.instantiate(name='Default')
+    instance.copy_to(domain)
+    return domain
+
+
+@pytest.fixture(scope="function")
+def modify_instance(domain):
+    instance = domain.parent.instantiate(name=domain.name)\
```
quarckster: **REQUIRED**
I don't understand why do you instantiate another domain with the same name?

sshveta: **def domain(request):** 
method is creating a new domain and the copying the instance from ManageIQ domain to the new domain .
 
def modify_instance(domain): 
this method is then modifying  the above instance in new domain to the change the approval to manual.
So the instance in both method are different , one belongs to "ManageIQ" and the other to new_domain

quarckster: Yes, you already have that new domain object, so why do you create another object with the same name in `domain.parent.instantiate(name=domain.name)`?

cfme/tests/services/test_service_manual_approval.py
```diff
@@ -0,0 +1,78 @@
...
+        .instances.instantiate(name='Default')
+    instance.copy_to(new_domain)
+    return new_domain
+
+
+@pytest.fixture(scope="function")
+def modify_instance(create_domain):
+    """Modify the instance in new domain to change it to manual approval instead of auto"""
+
+    instance = create_domain.namespaces.instantiate(name='Service')\
```
sshveta: Now got what you are saying :)
Fixed :)

PR #4891, opened 2017-06-20
[1LP][RFR] Add Chargeback support for cloud providers

nachandr:
{{pytest: cfme/tests/intelligence/reports/test_validate_chargeback_report.py --use-provider complete}}

PR #5198, opened 2017-08-21
[1LP][RFR] Add cloud.instance.set_ownership

### LINKED ISSUE #5038: Unable to set and unset ownership for Cloud instances
https://paste.fedoraproject.org/paste/3TABZclj2Buyh8p5kHH2wg

This issue happens with Cloud instances only and not infra VMs.

mshriver:
Override BaseVM.set_ownership until subclasses support widgetastic

FIXES #5038
FIXES RHCFQE-3855

## PRT Results
Scoping PRT for `cloud_infra_common/test_vm_ownership` for standard ownership tests, and `intelligence/reports/test_validate_chargeback_report` for where the failure was originally observed.

test_vm_ownership is only supported by ec2 according to YAML configuration.

The original issue was written against intelligence/reports/test_validate_chargeback_report, which at the moment does not support cloud instances for testing. @nachandr Is adding cloud instance support to the test via #4891, and will be independently running that test against this commit.

{{ pytest: -v --long-running --use-provider ec2west cfme/tests/cloud_infra_common/test_vm_ownership.py }}

Seeing passing results in run 15099 that have called set_ownership, which is as much as I wanted to see out of PRT. There may be test failures but the results at least demonstrate this method working.

### DIFF
--- a/cfme/cloud/instance/__init__.py
+++ b/cfme/cloud/instance/__init__.py
@@ -423,6 +423,65 @@ class Instance(VM, Navigatable):
         view.toolbar.power.item_select(kwargs.get('option'),
                                        handle_alert=not kwargs.get('cancel', False))
 
+    def set_ownership(self, user=None, group=None, click_cancel=False, click_reset=False):
+        """Set instance ownership
+
+        TODO: collapse this back to common.vm after both subclasses converted to widgetastic
+        Args:
+            user (str): username for ownership
+            group (str): groupname for ownership
+            click_cancel (bool): Whether to cancel form submission
+            click_reset (bool): Whether to reset form after filling
+        """
+        view = navigate_to(self, 'SetOwnership')
+        fill_result = view.form.fill({
+            'user_name': user,
+            'group_name': group})
+        if not fill_result:
+            view.flash.assert_no_error()
+            view.form.cancel_button.click()
+            view = self.create_view(InstanceDetailsView)
+            view.flash.assert_success_message('Set Ownership was cancelled by the user')
+            view.flash.assert_no_error()
+            return
+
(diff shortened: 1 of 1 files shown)


cfme/cloud/instance/__init__.py
```diff
@@ -423,6 +423,65 @@ def power_control_from_cfme(self, *args, **kwargs):
...
 
+    def set_ownership(self, user=None, group=None, click_cancel=False, click_reset=False):
+        """Set instance ownership
+
+        TODO: collapse this back to common.vm after both subclasses converted to widgetastic
+        Args:
+            user (str): username for ownership
+            group (str): groupname for ownership
+            click_cancel (bool): Whether to cancel form submission
+            click_reset (bool): Whether to reset form after filling
```
jkrocil: **REQUIRED**
@mshriver I understand you based this on `cfme.common.vm.BaseVM.set_ownership` which has `click_cancel` and `click_reset` params, but it's not correct. We don't prefix `cancel` and `reset` params with `click_` anywhere else, and since this is a good opportunity to get rid of it in `set_ownership` as well, let's do exactly that, please. :blush:

mshriver: @jkrocil I completely agree, and wanted to change this now - but it really should wait.

For now, this is callable both on infra.vm and cloud.instance in the same way, with the same arguments. I want to maintain this consistency until #5113 is merged and this method gets moved up into and overwrites cfme.common.vm.BaseVM.set_ownership - at which time I'll modify the arguments to remove 'click_'

jkrocil: @mshriver Sounds great - merging!

PR #5201, opened 2017-08-21
[1LP][WIP] Non Interactive Quadicons support + BaseNonInteractiveEntitiesView

izapolsk:
We have a bunch of views which contain quadicons like Edit Tags/Manage Policies/etc. Those quadicons provide some info and can be clicked. 
This PR allows to use BaseEntity widget for interaction with such QuadIcons. Besides that, it has a BaseNonInteractiveEntitiesView which is added to mentioned above views and allows to get all quadicons there and play with them.

### DIFF
--- a/cfme/common/provider_views.py
+++ b/cfme/common/provider_views.py
@@ -17,7 +17,8 @@ from widgetastic_manageiq import (BreadCrumb,
                                   BaseEntitiesView,
                                   DynaTree,
                                   BootstrapTreeview,
-                                  ProviderEntity)
+                                  ProviderEntity,
+                                  BaseNonInteractiveEntitiesView)
 
 
 class ProviderDetailsToolBar(View):
@@ -176,6 +177,13 @@ class ProvidersManagePoliciesView(BaseLoggedInPage):
     """
     policies = VersionPick({Version.lowest(): DynaTree('protect_treebox'),
                             '5.7': BootstrapTreeview('protectbox')})
+
+    @View.nested
+    class entities(BaseNonInteractiveEntitiesView):  # noqa
+        @property
+        def entity_class(self):
+            return ProviderEntity().pick(self.browser.product_version)
+
     save = Button('Save')
     reset = Button('Reset')
     cancel = Button('Cancel')
@@ -192,6 +200,12 @@ class ProvidersEditTagsView(BaseLoggedInPage):
     tag_category = BootstrapSelect('tag_cat')
     tag = BootstrapSelect('tag_add')
     chosen_tags = Table(locator='//div[@id="assignments_div"]/table')
+
+    @View.nested
+    class entities(BaseNonInteractiveEntitiesView):  # noqa
+        @property
+        def entity_class(self):
+            return ProviderEntity().pick(self.browser.product_version)
 
     save = Button('Save')
     reset = Button('Reset')
--- a/widgetastic_manageiq/__init__.py
+++ b/widgetastic_manageiq/__init__.py
@@ -2092,9 +2092,11 @@ class BaseQuadIconEntity(ParametrizedView, ClickableMixin):
 
     """
     PARAMETERS = ('name',)
-    ROOT = ParametrizedLocator('.//table[./tbody/tr/td/a[contains(@title, {name|quote})]]')
+    ROOT = ParametrizedLocator('.//table[./tbody/tr/td/*[(self::a or self::span) and '
+                               'contains(@title, {name|quote})]]')
     LIST = '//dl[contains(@class, "tile")]/*[self::dt or self::dd]'
-    label = Text(locator=ParametrizedLocator('./tbody/tr/td/a[contains(@title, {name|quote})]'))
+    label = Text(locator=ParametrizedLocator('./tbody/tr/td/*[(self::a or self::span) and '
+                                             'contains(@title, {name|quote})]'))
     checkbox = Checkbox(locator='./tbody/tr/td/input[@type="checkbox"]')
     QUADRANT = './/div[@class="flobj {pos}72"]/*[self::p or self::img or self::div]'
 
@@ -2137,8 +2139,8 @@ class BaseTileIconEntity(ParametrizedView):
 
     """
     PARAMETERS = ('name',)
-    ROOT = ParametrizedLocator('.//table[.//table[./tbody/tr/td/a[contains(@title, '
-                               '{name|quote})]]]')
+    ROOT = ParametrizedLocator('.//table[.//table[./tbody/tr/td/*[(self::a or self::span) and '
+                               'contains(@title, {name|quote})]]]')
     LIST = '//dl[contains(@class, "tile")]/*[self::dt or self::dd]'
     quad_icon = ParametrizedView.nested(BaseQuadIconEntity)
 
@@ -2317,13 +2319,7 @@ class EntitiesConditionalView(View, ReportDataControllerMixin):
 
         Returns: all current page entities
         """
-        current_version = VersionPick({
-            Version.lowest(): 'old',
-            '5.9': 'new',
-            'upstream': 'new'
-        }).pick(self.browser.product_version)
-
-        if current_version == 'old':
+        if self.browser.product_version < '5.9':
             br = self.browser
             return [br.get_attribute('title', el) for el in br.elements(self.elements)]
         else:
@@ -2346,10 +2342,10 @@ class EntitiesConditionalView(View, ReportDataControllerMixin):
                                 for name in self.entity_names])
             return entities
 
-    def get_entities(self, by_name=None, surf_pages=False):
+    def get_entities(self, by_partial_name, surf_pages=False):
         """ obtains all matched entities like QuadIcon displayed by view
         Args:
-            by_name (str): only entities which match to by_name will be returned
+            by_partial_name (str): only entities which match to by_name will be returned
             surf_pages (bool): current page entities if False, all entities otherwise
 
         Returns: all matched entities (QuadIcon/etc.) displayed by view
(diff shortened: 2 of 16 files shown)


dajoRH:
*CFME QE Bot*

dajoRH:
Lint report for commit 3e46d4d7ec277c6263f13251c36bcee3162b5501:

- :red_circle: Line 182:11: **N801** *class names should use CapWords convention*
- :red_circle: Line 205:11: **N801** *class names should use CapWords convention*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

widgetastic_manageiq/__init__.py
```diff
@@ -2820,3 +2816,83 @@ def fill(self, value):
...
+        return [self.entity_class(parent=self, name=name) for name in self.entity_names]
+
+    def get_entities(self, by_name=None):
+        """ obtains all matched entities like QuadIcon displayed by view
+        Args:
+            by_name (str): only entities which match to by_name will be returned
+
+        Returns: all matched entities (QuadIcon/etc.) displayed by view
+        """
+        entities = self.get_all()
```
mshriver: Consider sticking this call in-line with `for entity in self.get_all()` and calling `remaining_entites` something like `matched_entities`

widgetastic_manageiq/__init__.py
```diff
@@ -2820,3 +2816,83 @@ def fill(self, value):
...
+    def get_entity(self, by_name=None):
+        """ obtains one entity matched to by_name
+        raises exception if no entities or several entities were found
+        Args:
+            by_name (str): only entity which match to by_name will be returned
+
+        Returns: matched entities (QuadIcon/etc.)
+        """
+        entities = self.get_entities(by_name=by_name)
+        if len(entities) == 0:
```
mshriver: Consider moving this into `get_entities` so its also raised by that method when called directly and nothing matches.

widgetastic_manageiq/__init__.py
```diff
@@ -2820,3 +2816,83 @@ def fill(self, value):
...
+        """ obtains all entities like QuadIcon displayed by view
+
+        Returns: all entities (QuadIcon/etc.) displayed by view
+        """
+        return [self.entity_class(parent=self, name=name) for name in self.entity_names]
+
+    def get_entities(self, by_name=None):
+        """ obtains all matched entities like QuadIcon displayed by view
+        Args:
+            by_name (str): only entities which match to by_name will be returned
```
mshriver: **Required**

I don't think `by_name` should be an optional argument here, since if its not passed this method will always return an empty list.

widgetastic_manageiq/__init__.py
```diff
@@ -2820,3 +2816,83 @@ def fill(self, value):
...
+
+        Returns: all matched entities (QuadIcon/etc.) displayed by view
+        """
+        entities = self.get_all()
+        remaining_entities = []
+        for entity in entities:
+            if by_name and by_name in entity.name:
```
mshriver: If `by_name` is changed to not be optional as I've suggested above, this check isn't needed. currently this check is making the method return an empty list with the default args.

widgetastic_manageiq/__init__.py
```diff
@@ -2820,3 +2816,83 @@ def fill(self, value):
...
+        entities = self.get_entities(by_name=by_name)
+        if len(entities) == 0:
+            raise ItemNotFound("Entity {name} isn't found on this page".format(name=by_name))
+        elif len(entities) > 1:
+            raise ManyEntitiesFound("Several entities with {name} were found".format(name=by_name))
+        return entities[0]
+
+    def get_first_entity(self, by_name=None):
+        """ obtains one entity matched to by_name and stops on that page
+        raises exception if no entity or several entities were found
```
mshriver: copypasta, doesn't raise exception on multiple entities found

widgetastic_manageiq/__init__.py
```diff
@@ -2820,3 +2816,83 @@ def fill(self, value):
...
+        Args:
+            by_name (str): only entity which match to by_name will be returned
+
+        Returns: matched entity (QuadIcon/etc.)
+        """
+        found_entities = [self.entity_class(parent=self, name=name)
+                          for name in self.entity_names if by_name == name]
```
mshriver: Here I question by_name default behavior again - if I call `get_first_entity()`, I get an exception because of the check against `by_name == name`.

I think this method should return the first entity, regardless of name, when no name is supplied - and possibly not even take a by_name argument at all, since a caller could simply call `view.get_entities(by_name='test')[0]` to get the first.

mshriver:
Leaving as RFR for @quarckster review too

widgetastic_manageiq/__init__.py
izapolsk: I had probably not slept well the day before I created these methods :)
I've refactored all get_some_entity methods. 

Thanks very helpful objection.

widgetastic_manageiq/__init__.py
izapolsk: I have refactored it

widgetastic_manageiq/__init__.py
```diff
@@ -2357,42 +2353,42 @@ def get_entities(self, by_name=None, surf_pages=False):
...
                 return found_entities[0]
+            elif len(found_entities) > 1:
+                raise ManyEntitiesFound("Several entities with "
+                                        "{name} were found".format(name=by_name))
+            if not surf_pages:
+                raise ItemNotFound("Entity {name} isn't found on this page".format(name=by_name))
+        else:
+            raise ItemNotFound("Entity {name} isn't found on this page".format(name=by_name))
+
+    def get_first_entity(self, ):
```
quarckster: I believe it's a typo.

izapolsk: it doesn't influence anything but you're right this is typo. fixed

widgetastic_manageiq/__init__.py
```diff
@@ -2820,3 +2816,77 @@ def fill(self, value):
...
+        elif len(entities) > 1:
+            raise ManyEntitiesFound("Several entities with {name} were found".format(name=by_name))
+        return entities[0]
+
+    def get_first_entity(self):
+        """ obtains first entity
+
+        Returns: matched entity (QuadIcon/etc.)
+        """
+        for name in self.entity_names:
```
mshriver: I love this simple flow here.

widgetastic_manageiq/__init__.py
izapolsk: I lied give me a minute to fix this

widgetastic_manageiq/__init__.py
```diff
@@ -2357,42 +2353,45 @@ def get_entities(self, by_name=None, surf_pages=False):
...
         """
         for _ in self.paginator.pages():
             found_entities = [self.parent.entity_class(parent=self, name=name)
                               for name in self.entity_names if by_name == name]
-            if found_entities:
+            if len(found_entities) == 1:
                 return found_entities[0]
+            elif len(found_entities) > 1:
+                raise ManyEntitiesFound("Several entities with "
+                                        "{name} were found".format(name=by_name))
```
mshriver: This doesn't cover when its empty, need an `else`

widgetastic_manageiq/__init__.py
```diff
@@ -2820,3 +2819,79 @@ def fill(self, value):
...
+        """ obtains all matched entities like QuadIcon displayed by view
+        raises exception if no entities found
+        Args:
+            by_name (str): only entities which match to by_name will be returned
+
+        Returns: all matched entities (QuadIcon/etc.) displayed by view
+        """
+        remaining_entities = []
+        for entity in self.get_all():
+            if by_name in entity.name:
```
mshriver: `in` makes this a partial matching method, consider changing 'by_name' to 'by_partial_name' or turning this into an exact match.

widgetastic_manageiq/__init__.py
izapolsk: Mike, nope it is covered by  for ... else statement.

mshriver: I see the flow now, the for's else catches. on no return, and found_entities isn't used. looks safe.

cfme/common/provider_views.py
```diff
@@ -176,6 +177,13 @@ class ProvidersManagePoliciesView(BaseLoggedInPage):
     """
     policies = VersionPick({Version.lowest(): DynaTree('protect_treebox'),
                             '5.7': BootstrapTreeview('protectbox')})
+
+    @View.nested
+    class entities(BaseNonInteractiveEntitiesView):  # noqa
+        @property
+        def entity_class(self):
+            return ProviderEntity().pick(self.browser.product_version)
```
mfalesni: I think this property and verpick malarkey will be resolved by https://github.com/RedHatQE/widgetastic.core/issues/59

widgetastic_manageiq/__init__.py
mfalesni: Though:
```python
try:
    return self.entity_class(parent=self, name=self.entity_names[0])
except IndexError:
    raise ItemNotFound("No Entities found on this page")
```

would be more pythonic :stuck_out_tongue:

mfalesni:
@izapolsk 
```
Uncollection Stats:
 manual: 0
 uncollectif: 3
 0 tests left after all uncollections

========================= no tests ran in 0.51 seconds =========================
```

Can you please specify tests to run?

cfme/common/provider_views.py
izapolsk: good idea. I'll update code later when this change get merged

izapolsk:
I've tested it locally and added some small enhancement. Any issues haven't been found.

PR #5205, opened 2017-08-21
[WIP]Upgrade test and script

shalomnaim1:
# summary
The following PR will contain upgrade test. 

## Upgrade automation testing strategy
The upgrade itself executing by ansible playbook, the automation in this PR have to validate the consistency of the upgrade and detect  errors on the system after the upgrade finished such db corruption etc

## Actual testing plan:
The validation performed in two phases:

### Before running the upgrade:
Backup system's status to yaml file, the file will contain few samples for each object that regarding to containers providers and all the info the GUI provide about it 

**For example:** 
For Pod I will have the following structure:
```
  Pod:
    main_key: Name
    values:
      docker-registry-1-hx9fc: # <-- The name of the pod
        Containers: 1/1
        DNS Policy: ClusterFirst
        Phase: Running
        Project Name: default
        Provider: CM ENV2
        Ready: 'True'
        Restart Policy: Always
```
### After the upgrade
Validate that the data on the CFME don't corrupt after db migration.
For each object on the yaml file i will validate that the values are identical to the CFME

### DIFF
--- a/cfme/tests/containers/test_upgrade_cfme.py
+++ b/cfme/tests/containers/test_upgrade_cfme.py
@@ -1,0 +1,53 @@
+import pytest
+from utils import testgen
+
+from utils.version import current_version
+
+from cfme.configure.configuration import Tag
+from cfme.containers.provider import ContainersProvider
+from cfme.containers.image import Image
+from cfme.containers.project import Project
+from cfme.containers.node import Node
+from cfme.containers.image_registry import ImageRegistry
+from cfme.containers.pod import Pod
+from cfme.containers.template import Template
+from cfme.containers.container import Container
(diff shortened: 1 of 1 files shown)


cfme/tests/containers/test_upgrade_cfme.py
```diff
@@ -0,0 +1,62 @@
...
+
+        # collect all instance values
+        all_instance["values"][curr_instance_name] = \
+            {filed: str(instance.columns[headers_mapping[filed]].text) for
+             filed in set(headers_mapping) - set([main_key])}
+
+    return all_instance
+
+
+def test_pre_upgrade_status_save(provider):
```
gshefer: Please don't forget to mark with polarion marker.

Issue #5180, opened 2017-08-17
Issue with WT Paginator on Services/Requests page


lkhomenk:
This page has simplified paginator and which doesn't have 'check_all_items'. which leads to failure in paginator.pages() method. 
To check it's paginator - provision anything - VM/Instance/Service/Migrate/etc - after that Request's page will be opened.**there's an open BZ also related to this - https://bugzilla.redhat.com/show_bug.cgi?id=1450002** - Missing Paginator on the same page if opened manually.
paginator.is_displayed() should be updated to fit this particular page too.

hint: old web_ui paginator was yielding None if paginator wasn't found - which was working for Requests's page just fine ;)

PR #5207, opened 2017-08-22
[RFR]adding BZ blocker and workaround for Requests page

lkhomenk:
Purpose or Intent
=================

__Fixing__  GH issue #5180. When we are using paginator, we expect some page result. such workaround would give us a try to go thru items at page we currently have. as that happened with Requests page - no pagination controls at all if we open them manually


lkhomenk:
another way or idea would be - give a try to iterate items even if None was returned by paginator.pages() + add some logging. or raise an error on paginator not found. for now it just silently dies. @izapolsk I would have your review here please

PR #5220, opened 2017-08-23
[1LP][RFR] Add support of keystone v3 to Openstack Cloud provider

okolisny:
Provide a possibility of adding OSP cloud provider on keystone v3 api
PRT run is just to ensure that old configs still work
{{pytest: cfme/tests/openstack/cloud --use-provider rhos7-ga}}

### DIFF
--- a/cfme/cloud/provider/openstack.py
+++ b/cfme/cloud/provider/openstack.py
@@ -84,11 +84,16 @@ class OpenStackProvider(CloudProvider):
         from cfme.utils.providers import get_crud
         infra_prov_key = prov_config.get('infra_provider_key')
         infra_provider = get_crud(infra_prov_key, appliance=appliance) if infra_prov_key else None
+        api_version = prov_config.get('api_version', None)
+
+        if not api_version:
+            api_version = 'Keystone v2'
+
         return cls(name=prov_config['name'],
                    hostname=prov_config['hostname'],
                    ip_address=prov_config['ipaddress'],
                    api_port=prov_config['port'],
-                   api_version=prov_config.get('api_version', 'Keystone v2'),
+                   api_version=api_version,
                    endpoints=endpoints,
                    zone=prov_config['server_zone'],
                    key=prov_key,
--- a/cfme/cloud/provider/openstack.py
+++ b/cfme/cloud/provider/openstack.py
@@ -21,13 +21,16 @@ class OpenStackProvider(CloudProvider):
     _ctrl_alt_del_xpath = '//*[@id="sendCtrlAltDelButton"]'
 
     def __init__(self, name=None, endpoints=None, zone=None, key=None, hostname=None,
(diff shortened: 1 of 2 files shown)


cfme/cloud/provider/openstack.py
```diff
@@ -83,9 +88,11 @@ def from_config(cls, prov_config, prov_key, appliance=None):
                    hostname=prov_config['hostname'],
                    ip_address=prov_config['ipaddress'],
                    api_port=prov_config['port'],
+                   api_version=prov_config.get('api_version', None),
```
mkoura: ``None`` is redundant, it's returned by default.

cfme/cloud/provider/openstack.py
```diff
@@ -83,9 +88,11 @@ def from_config(cls, prov_config, prov_key, appliance=None):
...
                    endpoints=endpoints,
                    zone=prov_config['server_zone'],
                    key=prov_key,
+                   keystone_v3_domain_id=prov_config.get('domain_id', None),
```
mkoura: ``None`` is redundant, it's returned by default.

okolisny:
@mkoura those tests in PRT failed because of race conditions - couldn't wait for instance to appear. That's because OSP deployment feels not so well.
Anyway, the purpose of this run is to show that it is still possible to add OSP cloud provider with keystone v2 (i.e. nothing has been broken). It is better to run PRT with provider on keystone v3, but as I know there is no appropriate providers in yamls.

okolisny:
Waiting for https://github.com/ManageIQ/wrapanapi/pull/174 to get merged

psav:
Can we squash this and try it against rhos11 provider?

okolisny:
@psav rhos11 configured to work over v2.0... Also, I can see it is tagged as 'disabled' - is it okay ?

cfme/cloud/provider/openstack.py
```diff
@@ -83,9 +88,11 @@ def from_config(cls, prov_config, prov_key, appliance=None):
                    hostname=prov_config['hostname'],
                    ip_address=prov_config['ipaddress'],
                    api_port=prov_config['port'],
+                   api_version=prov_config.get('api_version', 'Keystone v2'),
```
mshriver: if api_version is None, this won't default to 'Keystone v2', it will be None. I think you probably want to default to 'Keystone v2' when api_version is None, not only when api_version isn't set.

okolisny: Got it, done.

mshriver:
Everything looks fine to me but @okolisny can you adjust your {{pytest}} line in your description, its collecting/uncollectif on 11 cases so nothing is actually running.

Will be waiting until Wed Sept 27th to merge at earliest - in sprint 13.

okolisny:
@mshriver lots of tests were failed by reasons which are not connected with these changes.. The main point - provider was added successfully

PR #5231, opened 2017-08-24
[WIPTEST]Ansible Fixtures first commit

pavelzag:
Purpose or Intent
=================
{{pytest: cfme/tests/containers/test_containers_summary.py -v --use-provider cm-env2 }}

This PR allows usage of Ansible based fixtures to add and remove providers


mfalesni:
Hmm, Ansible uses REST, and we will use REST once Sentaku goes in. Directly from the framework. So I don't believe this is something we want now.

@psav ? ^

PR #5232, opened 2017-08-24
[RFR] Removed the grabbing of bearer token

gshefer:
{{ pytest: cfme/tests/containers/test_labels.py --use-provider cm-env2 }}
We don't have to use the session token and we can just use the token from the config.

### DIFF
--- a/cfme/containers/provider/openshift.py
+++ b/cfme/containers/provider/openshift.py
@@ -71,34 +71,6 @@ class OpenshiftProvider(ContainersProvider):
     def cli(self):
         return OcpCli(self)
 
-    def get_bearer_token(self):
-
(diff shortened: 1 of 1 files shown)


jkrocil:
But you just added this - is it not needed?

gshefer:
Yes, but we have found a way to use the full, you just have to use the correct service account and have the right roles. I guess @jaryn can explain further.

jaryn:
I think it should not be needed if the provider is configured right way (with correct token). I didn't understand reason for adding these functions.

jkrocil:
Cool.

jkrocil:
Tested manually; works.

PR #5238, opened 2017-08-25
[1LP][RFR] Adding version pick ability to pagination

kedark3:
Purpose or Intent
=================

__Extending__ set_items_per_page pagination function to accommodate different CFME versions. 

{{ pytest: cfme/tests/infrastructure/test_webmks_console.py --use-provider vsphere6-nested --use-provider vsphere65-nested}}

### DIFF
--- a/cfme/intelligence/reports/reports.py
+++ b/cfme/intelligence/reports/reports.py
@@ -353,7 +353,7 @@ class CustomSavedReport(Updateable, Pretty, Navigatable):
         Returns: :py:class:`SavedReportData`.
         """
         view = navigate_to(self, "Details")
-        view.paginator.set_items_per_page("1000 items")
+        view.paginator.set_items_per_page(1000)
         try:
             headers = tuple([hdr.encode("utf-8") for hdr in view.table.headers])
             body = []
--- a/widgetastic_manageiq/__init__.py
+++ b/widgetastic_manageiq/__init__.py
@@ -1316,13 +1316,17 @@ class NonJSPaginationPane(View):
         """Selects number of items to be displayed on page.
 
         Args:
-            value: Ideally a str of the format 'x items', x could be {10,20,50,..,1000}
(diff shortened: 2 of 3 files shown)


widgetastic_manageiq/__init__.py
```diff
@@ -1322,7 +1323,11 @@ def set_items_per_page(self, value):
         if not isinstance(value, (int, six.string_types)):
             raise TypeError("Value should either be of format either e.g. 10 or a string"
                 " e.g. '10 items' not {}".format(value))
-        items_text = value if 'items' in str(value) else '{} items'.format(value)
+        items_text = version.pick({
```
mshriver: Discussed offline using widgetastic.utils.VersionPick for this instead.

cfme/intelligence/reports/reports.py
```diff
@@ -353,7 +353,7 @@ def data(self):
         Returns: :py:class:`SavedReportData`.
         """
         view = navigate_to(self, "Details")
-        view.paginator.set_items_per_page("1000 items")
+        view.paginator.set_items_per_page(1000)
```
mshriver: :+1:

widgetastic_manageiq/__init__.py
```diff
@@ -1316,13 +1316,19 @@ def set_items_per_page(self, value):
...
+        try:
+            value = int(value)
+        except ValueError:
+            raise ValueError("Value should be integer and not {}".format(value))
+
+        items_text = VersionPick({
+            Version.lowest(): str(value),
+            '5.7.3': value if 'items' in str(value) else '{} items'.format(value),
+            '5.8.1.5': value if 'items' in str(value) else '{} items'.format(value),
+            '5.8.2': value.split(' ')[0] if 'items' in str(value) else str(value),
```
mshriver: This can be simplified now that you're only accepting ints.

PR #5244, opened 2017-08-28
[RFR] Fix test_relationships

gshefer:
{{ pytest: cfme/tests/containers/test_relationships.py --use-provider cm-env2 }}


gshefer:
Dependency: https://github.com/ManageIQ/integration_tests/pull/5241

PR #5247, opened 2017-08-29
[RFR][NOTEST] Make 1 provider the default limit

jkrocil:
Parallelizer is not enough to take care of this; jenkins simply has to run limited to 1 provider and since there is really no reason not to have this as a default for non-parallelized (local) runs either, let's set it to 1.


jkrocil:
@mfalesni @psav @RonnyPfannschmidt

PR #5274, opened 2017-09-01
[1LP][RFR] Update cloud.keypairs, use BaseEntitiesView

mshriver:
Original conversion done before BaseEntitiesView was written, convert it over. So pretty.

Use BaseEntitiesView via common.vm_views.VMEntities, since it handles the version picking already.  Maybe those entities views are a bit more common than vm's.

Tests are fine on RHOS running locally. The CFME QE RHOS provider that was available for PRT has been unstable and removed, so we won't get PRT results.

## PRT Results
Run 15737:
* 57z pass
* 58z failures from slow UI updates.

{{ pytest: -v --long-running --use-provider rhos7-ga cfme/tests/cloud/test_keypairs.py }}

### DIFF
--- a/cfme/cloud/keypairs.py
+++ b/cfme/cloud/keypairs.py
@@ -2,17 +2,19 @@
 from navmazing import NavigateToAttribute, NavigateToSibling
 from widgetastic.widget import View
 from widgetastic.utils import VersionPick, Version
-from widgetastic.exceptions import NoSuchElementException
 from widgetastic_patternfly import Dropdown, Button, FlashMessages
-from cfme.exceptions import KeyPairNotFound
+from widgetastic_manageiq import (
+    ItemsToolBarViewSelector, Text, TextInput, Accordion, ManageIQTree, BreadCrumb,
+    SummaryTable, BootstrapSelect, ItemNotFound)
+
 from cfme.base.ui import BaseLoggedInPage
+from cfme.common.vm_views import VMEntities
+from cfme.exceptions import KeyPairNotFound
+
 from cfme.web_ui import match_location, mixins
 from utils.appliance.implementations.ui import navigate_to, navigator, CFMENavigateStep
 from utils.appliance import Navigatable
 from utils.wait import wait_for
-from widgetastic_manageiq import (
-    ItemsToolBarViewSelector, Text, TextInput, Table, Search, PaginationPane, Accordion,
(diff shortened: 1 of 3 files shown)


dajoRH:
*CFME QE Bot*

cfme/cloud/keypairs.py
```diff
@@ -89,7 +87,6 @@ def is_displayed(self):
 
     toolbar = View.nested(KeyPairToolbar)
     entities = View.nested(KeyPairEntities)
```
izapolsk: please use include  here

mshriver: I ended up removing empty KeyPairEntities and used View.include for VMEntities directly here.

mshriver:
@izapolsk included entities view instead of nesting, please re-review.  PRT will not be sane because of lacking RHOS provider.

cfme/cloud/keypairs.py
```diff
@@ -88,8 +82,7 @@ def is_displayed(self):
             self.entities.title.text == 'Key Pairs')
 
     toolbar = View.nested(KeyPairToolbar)
-    entities = View.nested(KeyPairEntities)
-    paginator = PaginationPane()
+    including_entities = View.include(VMEntities, use_parent=True)
```
izapolsk: @mshriver, why do you use VMEntities, not BaseEntitiesView here ?

mshriver: Discussed in depth offline, but the core of it was I misunderstood the need to version pick the entity class with the new entities - this is only necessary when implementing data property, which isn't needed here.

Testing with BaseEntitiesView included.

jkrocil:
@mshriver Are the current 5.8 failures on PRT relevant?

mshriver:
@jkrocil I do *not* believe they're relevant for the scope of the work in this PR.  The tests need to be refactored a bit to correctly use the collection to instantiate and create keypair objects. The failures in 5.8 were from the keypair not appearing in the UI in time, which is mostly related to the slow RHOS updates in CFME. 

I can spend more time with this PR, updating all of the tests to be compliant with collection design, but I think we'd be better off waiting until collection base class is standardized. I can write GH issue + card to fix tests and we can get the FA owner to address them?

jkrocil:
@mshriver Sounds good to me. Merging.

PR #5280, opened 2017-09-04
[1LP][RFR] replacing vm/instance find_quadicon with widgetastic implementation

izapolsk:
vm/instance.find_quadicon method is widely used and doesn't allow to get rid of web_ui.QuadIcon. 
This PR intends to replace find_quadicon with widgetastic calls and updates all tests to use widgetastic Entity instead

### DIFF
--- a/cfme/infrastructure/datastore.py
+++ b/cfme/infrastructure/datastore.py
@@ -118,8 +118,7 @@ class DatastoreCollection(Navigatable):
     """Collection class for `cfme.infrastructure.datastore.Datastore`"""
 
     def __init__(self, appliance=None):
-        self.appliance = appliance
-        Navigatable.__init__(self, appliance=self.appliance)
+        Navigatable.__init__(self, appliance=appliance)
 
     def instantiate(self, name, provider, type=None):
         return Datastore(name, provider, type=type, collection=self)
@@ -210,7 +209,7 @@ class Datastore(Pretty, Navigatable):
         view = navigate_to(self, 'Details')
         view.contents.relationships.click_at('Hosts')
         hosts_view = view.browser.create_view(RegisteredHostsView)
-        return [h.name for h in hosts_view.entities.get_all()]
+        return hosts_view.entities.get_all()
 
     def get_vms(self):
         """ Returns names of VMs (from quadicons) that use this datastore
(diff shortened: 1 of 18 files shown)


dajoRH:
Lint report for commit d058a1625d1e37bd81d5643e2300a5b54ddb4355:

`cfme/cloud/instance/__init__.py`:
- :red_circle: Line 15:1: **F401** *'cfme.web_ui.Quadicon' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

dajoRH:
*CFME QE Bot*

izapolsk:
PRT fail due to different issues with rhos7 and rhos11

cfme/infrastructure/datastore.py
```diff
@@ -118,8 +118,7 @@ class DatastoreCollection(Navigatable):
     """Collection class for `cfme.infrastructure.datastore.Datastore`"""
 
     def __init__(self, appliance=None):
-        self.appliance = appliance
-        Navigatable.__init__(self, appliance=self.appliance)
+        Navigatable.__init__(self, appliance=appliance)
```
quarckster: **REQURIED**
According last email from @psav it should be like this:
```python
class DatastoreCollection(NavigatableMixin):
     """Collection class for `cfme.infrastructure.datastore.Datastore`"""
     def __init__(self, appliance):
         self.appliance = appliance
```

izapolsk: @quarckster, I agree but I'm not planning to fix collections in my PR. You see this change in my PR only because there was error before and @rbabyuk and me fixed it at the same time.  Her PR was merged first. But she didn't remove self.appliance definition. So, I fixed this small issue.

cfme/infrastructure/datastore.py
```diff
@@ -210,7 +209,7 @@ def get_hosts(self):
         view = navigate_to(self, 'Details')
         view.contents.relationships.click_at('Hosts')
         hosts_view = view.browser.create_view(RegisteredHostsView)
-        return [h.name for h in hosts_view.entities.get_all()]
+        return hosts_view.entities.get_all()
```
mshriver: This is changing from returning list of host names to list of host entitiy objects - I think you want to call `all_entity_names()` here.

izapolsk: :+1: yes, good idea. this PR was finished before you added that method.

mshriver: Though it looks like you've changed callers to handle getting entity objects back, not just names. Forget previous comment, because I think its _better_ to return entity objects here, not just the names.

PR #5288, opened 2017-09-05
[1LP][RFR] Bump wt-pf to 0.0.14 everywhere

jkrocil:
Fixes bootstrap switch when used as a widget in a table, among other things. That's why the vm reconfigure test below.

{{pytest: cfme/tests/infrastructure/test_vm_reconfigure.py -v --use-provider vsphere65-nested --long-running}}


jkrocil:
If at least one of the tests passes in 5.7 and 5.8, we are good. And right now it's 4 passes per stream (except upstream).

PR #5294, opened 2017-09-06
[RFR] fix for test tag objects

izapolsk:


izapolsk:
Despite of many test skips, PRT results look good because there were failures caused by introducing collections.
Besides that, I've started PRT once more.

izapolsk:
such fixes are already done in #5268

PR #5296, opened 2017-09-06
[1LP][RFR] Update WidgetasticTaggable, standard views/widgets

mshriver:
Standardize naming for view.form and view.entities
Maintain support for current implementation for any classes that are
using it already

## PRT Results

Local tests for catalog_item tagging passed, waiting for PRT.
### Run 15734
- 57z: passed
- 58z: **Unrelated** failure during test teardown when deleting catalog

{{ pytest: -v --long-running cfme/tests/services/test_catalog_item.py  -k "test_edit_tags or test_tagvis_catalog_items" }}

### DIFF
--- a/cfme/common/__init__.py
+++ b/cfme/common/__init__.py
@@ -199,11 +199,11 @@ class WidgetasticTaggable(object):
         - 'Details' navigation
         - 'Details' view should have entities.smart_management SummaryTable widget
         - 'EditTags' navigation
-        - 'EditTags' view should have nested 'form' view with 'tags' table widget (TagPageView)
+        - 'EditTags' view should have nested 'form' view with 'tags' table widget
+            - Suggest using class cfme.common.TagPageView as view for 'EditTags' nav
 
-    This class should be used for already converted to widgetastic classes
-
-    It provides functionality to assign and unassigned tags.
+    This class provides functionality to assign and unassigned tags for page models with
+    standardized widgetastic views
     """
 
     def add_tag(self, category=None, tag=None, cancel=False, reset=False):
--- a/cfme/common/__init__.py
+++ b/cfme/common/__init__.py
@@ -2,9 +2,10 @@
 from functools import partial
 import re
 from urlparse import urlparse
-from widgetastic.exceptions import NoSuchElementException
+from widgetastic.exceptions import NoSuchElementException, RowNotFound
 from widgetastic_patternfly import BootstrapSelect, Button
-from widgetastic.widget import Table, Text
+from widgetastic.widget import Table, Text, View
+from widgetastic_manageiq import BaseNonInteractiveEntitiesView
 
 from cached_property import cached_property
 from cfme.base.login import BaseLoggedInPage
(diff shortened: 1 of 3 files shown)


cfme/common/__init__.py
```diff
@@ -213,13 +213,14 @@ def add_tag(self, category=None, tag=None, cancel=False, reset=False):
         if isinstance(tag, Tag):
             category = tag.category.display_name
             tag = tag.display_name
+        form = view.form if getattr(view, 'form', None) else view
```
quarckster:
```python
form = view.form if hasattr(view, 'form') else view
```

cfme/common/__init__.py
```diff
@@ -286,7 +288,9 @@ def get_tags(self, tenant="My Company Tags"):
             List of tags in format "Tag_category: Tag_name"
         """
         view = navigate_to(self, 'Details')
-        tags = view.tag.read()
+        # look for simple 'tag' widget first, then standard entities.smart_management
+        tag_table = view.tag if getattr(view, 'tag', None) else view.entities.smart_management
```
quarckster:
```python
tag_table = getattr(view, 'tag', view.entities.smart_management)
```

mshriver: Doing this above as well

cfme/common/__init__.py
```diff
@@ -191,9 +191,11 @@ def is_displayed(self):
...
     """
     This class can be inherited by any class that honors tagging.
-    Class should have 'EditTags' navigate step registered,
-    also 'Details' view should be present with tag attribute(widget with tag
-    information on details page)
+    Class should have following:
+        - 'Details' navigation
+        - 'Details' view should have entities.smart_management SummaryTable widget
+        - 'EditTags' navigation
+        - 'EditTags' view should have nested view.form for add and view.tags table widget for remove
```
rlbabyuk: it is not mandatory to have 'EditTags' view with a form, we can use common TagPageView in EditTags navigation step

mshriver: I see this class now - there's one in common/vm_views as well. I'll revise this a bit, but I'm thinking the common view should have a nested form as we've done that on other pages. the title/buttons are not in the nested form, but in the top-level view. Only form widgets are in the nested 'form' view.

dajoRH:
Lint report for commit 42d348f01a45818f2e2a32b2909de033d72cbb59:

`cfme/common/__init__.py`:
- :red_circle: Line 178:11: **N801** *class names should use CapWords convention*

**No commit flaws detected**.

*CFME QE Bot*

cfme/common/__init__.py
```diff
@@ -172,28 +173,33 @@ class TagPageView(BaseLoggedInPage):
...
     """
     This class can be inherited by any class that honors tagging.
-    Class should have 'EditTags' navigate step registered,
-    also 'Details' view should be present with tag attribute(widget with tag
-    information on details page)
+    Class should have following:
+        - 'Details' navigation
+        - 'Details' view should have entities.smart_management SummaryTable widget
+        - 'EditTags' navigation
+        - 'EditTags' view should have nested 'form' view with 'tags' table widget (TagPageView)
```
mshriver: @rlbabyuk thoughts on this wording, and the changes to TagPageView to match?

rlbabyuk: As an option: 'TagPageView should be used in 'EditTags' navigation step.'

mshriver: yep, good point, I just put the class name in (), but it would be much more clear stated that way.

mshriver:
Requested a few reviews, not necessarily asking all to review, just whoever is most relevant for 2nd level :+1:

PR #5299, opened 2017-09-06
[1LP][RFR] Use BootstrapSelect instead of BootstrapSelectByLocator

mfalesni:
Fixes an exception

{{pytest: -v cfme/tests/automate/test_domain.py -k git --long-running }}


mfalesni:
This is what I call a fast turnaround :champagne:

PR #5305, opened 2017-09-07
[1LP][RFR] Fix generate report tests

niyazRedhat:
Fixed test_generate_report.py
{{pytest: cfme/tests/intelligence/reports/test_generate_report.py -v --use-provider vsphere55}}

### DIFF
--- a/cfme/intelligence/reports/reports.py
+++ b/cfme/intelligence/reports/reports.py
@@ -14,7 +14,7 @@ from widgetastic.utils import ParametrizedLocator
 from widgetastic.widget import Text, Checkbox, View, ParametrizedView, Table as VanillaTable
 from widgetastic.exceptions import NoSuchElementException
 from widgetastic_manageiq import PaginationPane, Table
-from widgetastic_patternfly import Button, Input, BootstrapSelect, Tab
+from widgetastic_patternfly import Button, Input, BootstrapSelect, Tab, CandidateNotFound
 from cfme.web_ui.expression_editor_widgetastic import ExpressionEditor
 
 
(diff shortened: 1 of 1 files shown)


cfme/intelligence/reports/reports.py
```diff
@@ -539,7 +539,6 @@ def delete(self, cancel=False):
             view.flash.assert_no_error()
         else:
             view.flash.assert_no_error()
-            view.flash.assert_message("Successfully deleted Saved Report from the CFME Database")
```
quarckster: **REQUIRED**
Why it's deleted?

cfme/tests/intelligence/reports/test_generate_report.py
```diff
@@ -46,4 +46,4 @@ def test_reports_generate_report(request, path):
     """
 
     report = CannedSavedReport.new(path)
-    request.addfinalizer(report.delete_if_exists)
+    request.addfinalizer(report.delete)
```
quarckster: **REQUIRED**
If it not exists, exception will be raised.

cfme/intelligence/reports/reports.py
niyazRedhat: Hey when we delete the saved report it doesn't display the flash message. Its bug.I logged that.This is reproducible in 5.7 and 5.8 version. so i think its not required. 
Just have a look ---> https://bugzilla.redhat.com/show_bug.cgi?id=1489387

quarckster: Don't delete this line, just comment it and add TODO about this bug

niyazRedhat: Yes right. I will comment.

mshriver:
Tests passing in PRT, results just haven't updated here yet.

PR #5309, opened 2017-09-07
[1LP][RFR] Fixed BootstrapSwitchSelect view

quarckster:
Purpose
=================
BootstrapSwitchSelect stopped to work after https://github.com/RedHatQE/widgetastic.patternfly/pull/20 merging.

{{pytest: -v -k "test_assign_two_random_events_to_control_policy" --long-running}}

### DIFF
--- a/widgetastic_manageiq/__init__.py
+++ b/widgetastic_manageiq/__init__.py
@@ -674,59 +674,67 @@ class CheckboxSelect(Widget):
         return [cb for cb in self.checkboxes if cb.selected]
 
 
-class BootstrapSwitchSelect(CheckboxSelect):
+class BootstrapSwitchSelect(View):
     """BootstrapSwitchSelect view.
 
-    This view is very similar to parent CheckboxSelect view. BootstrapSwitches used instead of
-    usual Checkboxes. It can be found in the same policy's events assignment screen since
-    CFME 5.8.1.
+    This view is very similar to CheckboxSelect view. BootstrapSwitches used instead of
+    usual Checkboxes. It can be found in policy's events assignment screen since CFME 5.8.1.
 
     """
 
-    BS_TEXT = '/../../following-sibling::text()[1]'
+    ROOT = ParametrizedLocator(".//div[@id={@search_root|quote}]")
+    BS_LOCATOR = ".//text()[normalize-space(.)='{}']/preceding-sibling::div[1]//input"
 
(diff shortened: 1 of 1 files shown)


widgetastic_manageiq/__init__.py
```diff
@@ -674,59 +674,68 @@ def read(self):
...
 
-    def checkbox_by_id(self, id):
-        """Finds bootstrap switch by id."""
-        return BootstrapSwitch(self, id=id)
+    def switch_by_text(self, text):
+        id_attr = self._get_attr_from_text("id", text)
+        return self._bootstrap_switch(id_attr).switch
 
-    @property
+    @cached_property
```
mshriver: Why is this cached? I see you're deleting it on fill which is good, but I'm not seeing the advantage to caching it

quarckster: Because reading of it takes long time

widgetastic_manageiq/__init__.py
```diff
@@ -674,59 +674,68 @@ def read(self):
...
+    def _get_attr_from_text(self, attr, text):
+        return self.browser.element(self.BS_LOCATOR.format(text)).get_attribute(attr)
 
-        Returns: str
+    @cached_property
+    def all_labels(self):
+        elements = self.browser.elements(".//div[@class='form-horizontal']")
+        labels = []
+        for el in elements:
+            labels.extend(self.browser.text(el).split(" Yes No ")[1:])
```
mfalesni: Man, this is sooo crude :D

quarckster: I didn't find better way to get these strings 👅

widgetastic_manageiq/__init__.py
```diff
@@ -674,59 +674,68 @@ def read(self):
...
+    def switch_by_text(self, text):
+        id_attr = self._get_attr_from_text("id", text)
+        return self._bootstrap_switch(id_attr).switch
 
-    @property
+    @cached_property
     def selected_text(self):
-        """Only selected bootstrap switches' text descriptions."""
-        return {self._get_bs_description(bs) for bs in self.selected_checkboxes}
+        return [label for label in self.all_labels if self.switch_by_text(label).read()]
```
mfalesni: Can you replace the read with the specific property/method? I don't want the read method's signature cluttering the logs in this case :)

PR #5310, opened 2017-09-07
[RFR] fix Host's wait_for_host_state_change

okolisny:
Fixed lost old-style entity usage
Sorry, no PRT. There are no OSP Infra providers available


mshriver:
So this doesn't need any change to HostQuadIconEntity to support the `data()` call?

okolisny:
@mshriver It appeared to work fine. I've just tested it on old master version before:(

izapolsk:
Mike data property  is already there.

PR #5316, opened 2017-09-08
[1LP][RFR] small fix for test_events

izapolsk:
Often that test fails because Add button doesn't get active quick enough.
This PR should sort out that issue.

{{pytest: --use-provider=scvmm2016 cfme/tests/cloud_infra_common/test_events.py}}

### DIFF
--- a/cfme/control/explorer/actions.py
+++ b/cfme/control/explorer/actions.py
@@ -160,7 +160,7 @@ class Action(Updateable, Navigatable, Pretty):
             return self._alerts_to_evaluate
 
     def create(self):
-        "Create this Action in UI."
+        """Create this Action in UI."""
         view = navigate_to(self, "Add")
         view.fill({
             "description": self.description,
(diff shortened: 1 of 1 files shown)


cfme/control/explorer/actions.py
```diff
@@ -181,6 +181,7 @@ def create(self):
             "remove_tag": self.remove_tag,
             "run_ansible_playbook": self.run_ansible_playbook
         })
+        self.browser.plugin.ensure_page_safe()
```
psav: Shouldn't the .click() ensure the page is safe? @mfalesni???

mfalesni: Widgetastic's click calls ensure_page_safe at the end.

jkrocil: So... does this mean we might want to add it at the beginning as well? Before clicking that is.

izapolsk: @psav, added appropriate todo

izapolsk:
PRT is green

cfme/control/explorer/actions.py
mfalesni: @jkrocil Element lookups do check as well, so we check before and after.

jkrocil: So why is the `self.browser.plugin.ensure_page_safe()` required?

izapolsk: I suspect it added some delay.

PR #5321, opened 2017-09-08
[1LP][WIP] Fix test_compliance.py

quarckster:
Purpose
=================

__Fixing__ test_compliance.py after analysis profile widgetastic conversion. 

{{pytest: -v -k "test_compliance" --long-running --use-provider vsphere6-nested}}

### DIFF
--- a/cfme/tests/control/test_compliance.py
+++ b/cfme/tests/control/test_compliance.py
@@ -75,12 +75,9 @@ def vddk_url(provider):
 
 
 @pytest.yield_fixture(scope="module")
-def configure_fleecing(appliance, has_no_providers_modscope, provider, setup_provider_modscope,
-        vddk_url):
+def configure_fleecing(appliance, provider, setup_provider_modscope, vddk_url):
     setup_providers_hosts_credentials(provider.key)
-    appliance.install_vddk(reboot=True, vddk_url=vddk_url)
-    appliance.reboot()
-    appliance.browser.quit_browser()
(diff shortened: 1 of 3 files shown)


cfme/tests/control/test_compliance.py
```diff
@@ -75,12 +75,9 @@ def vddk_url(provider):
...
 
 @pytest.yield_fixture(scope="module")
-def configure_fleecing(appliance, has_no_providers_modscope, provider, setup_provider_modscope,
-        vddk_url):
+def configure_fleecing(appliance, provider, setup_provider_modscope, vddk_url):
     setup_providers_hosts_credentials(provider.key)
-    appliance.install_vddk(reboot=True, vddk_url=vddk_url)
-    appliance.reboot()
-    appliance.browser.quit_browser()
+    appliance.install_vddk(reboot=False, vddk_url=vddk_url)
```
jkrocil: How does this work if we never reboot the appliance after installing VDDK?

quarckster: It's working

jkrocil: So that's 2 issues right there that we need to take care of.
1. `.install_vddk` shouldn't have a `reboot` parameter if it's not necessary
2. Does this mean we __must not__ reboot the appliance once it has vddk installed on it? That sounds like a serious issue!

More investigation is necessary here.

PR #5338, opened 2017-09-12
[1LP][RFR] Refactored test_service_ansible_playbook_order_pass_extra_vars

quarckster:
Intent
=================

`test_service_ansible_playbook_order_pass_extra_vars` renamed to `test_service_ansible_playbook_pass_extra_vars`. Also from now it has two parameters: provisioning and retirement.

{{pytest: -v -k "test_service_ansible_playbook_pass_extra_vars or test_service_ansible_playbook_plays_table" --long-running}}

### DIFF
--- a/cfme/services/catalogs/__init__.py
+++ b/cfme/services/catalogs/__init__.py
@@ -9,6 +9,8 @@ from utils.appliance.implementations.ui import navigator, CFMENavigateStep
 
 
 class ServicesCatalogView(BaseLoggedInPage):
+
+    @property
     def in_explorer(self):
         return (
             self.logged_in_as_current_user and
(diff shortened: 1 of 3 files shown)


cfme/services/catalogs/__init__.py
```diff
@@ -9,6 +9,8 @@
 
 
 class ServicesCatalogView(BaseLoggedInPage):
+
+    @property
```
psav: **REQUIRED:** Check usage against https://github.com/ManageIQ/integration_tests/blob/master/cfme/services/catalogs/ansible_catalog_item.py#L155

quarckster: @psav fixed and checked, everything works fine.

PR #5349, opened 2017-09-13
[RFR] AutomateExplorer loaded server base

psav:
* This is NOT what we want to do, but is happening because the
  registration is in a different file Not seeing much of a way around
  this, but it is most definitely not ideal

{{pytest: -k test_pull_splitter_persistence[Server-AutomateExplorer]}}


izapolsk:
PRT have been restarted

izapolsk:
PRT failed again.

PR #5354, opened 2017-09-13
[1LP][RFR][NOTEST] Added pdf downloading check into test_widgets_operation

quarckster:
Purpose
=================

Added pdf downloading check for dashboard widgets https://bugzilla.redhat.com/show_bug.cgi?id=1461008. The test cannot be ran in PRT because it has blocker meta.


psav:
What do you mean blocking meta?

PR #5360, opened 2017-09-13
[1LP][RFR]Fixing test case failure for quadicon

anurag03:
Purpose or Intent
=================
Fixed failure related to quadicon in test_visual_cloud.py and test_visual_cloud.py

{{pytest: -v -k 'test_cloudprovider_noquads or test_infraprovider_noquads or test_host_noquads' }}

### DIFF
--- a/cfme/tests/configure/test_visual_cloud.py
+++ b/cfme/tests/configure/test_visual_cloud.py
@@ -158,5 +158,5 @@ def test_cloud_start_page(request, appliance, start_page):
 
 
 def test_cloudprovider_noquads(request, set_cloud_provider_quad):
-    navigate_to(CloudProvider, 'All')
-    assert visual.check_image_exists, "Image View Failed!"
+    view = navigate_to(CloudProvider, 'All')
+    assert view.entities.get_all()[0].data is None
--- a/cfme/tests/configure/test_visual_cloud.py
+++ b/cfme/tests/configure/test_visual_cloud.py
@@ -159,4 +159,5 @@ def test_cloud_start_page(request, appliance, start_page):
 
 def test_cloudprovider_noquads(request, set_cloud_provider_quad):
     view = navigate_to(CloudProvider, 'All')
-    assert view.entities.get_all()[0].data is None
+    view.toolbar.view_selector.select("Grid View")
+    assert view.entities.get_first_entity().data is None
--- a/cfme/tests/configure/test_visual_cloud.py
+++ b/cfme/tests/configure/test_visual_cloud.py
@@ -160,4 +160,5 @@ def test_cloud_start_page(request, appliance, start_page):
 def test_cloudprovider_noquads(request, set_cloud_provider_quad):
     view = navigate_to(CloudProvider, 'All')
     view.toolbar.view_selector.select("Grid View")
+    # Here get_first_entity() method will return None when the Quadrants option is deactivated.
     assert view.entities.get_first_entity().data is None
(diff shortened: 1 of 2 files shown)


cfme/tests/configure/test_visual_cloud.py
```diff
@@ -158,5 +158,5 @@ def test_cloud_start_page(request, appliance, start_page):
 
 
 def test_cloudprovider_noquads(request, set_cloud_provider_quad):
-    navigate_to(CloudProvider, 'All')
-    assert visual.check_image_exists, "Image View Failed!"
+    view = navigate_to(CloudProvider, 'All')
+    assert view.entities.get_all()[0].data is None
```
izapolsk: REQUIRED: If I understand this test correctly, it checks that cloud providers view displays not quad icons.
In such case you can check this either 
1. by checking what is chosen in view_selector. It should be List View or Tile View
or
2. using get_first_entity() instead and isinstance(BaseListEntity, BaseTileIconEntity). 
Note: second way won't work in upstream/5.9 with JSBaseEntity. It should have special method to check its type.
Please create a card about that and assign it to me.

cfme/tests/configure/test_visual_infra.py
```diff
@@ -213,13 +213,13 @@ def test_infra_start_page(request, appliance, start_page):
 
 
 def test_infraprovider_noquads(request, set_infra_provider_quad):
-    navigate_to(InfraProvider, 'All')
-    assert visual.check_image_exists, "Image View Failed!"
+    view = navigate_to(InfraProvider, 'All')
+    assert view.entities.get_all()[0].data is None
```
izapolsk: ^^

cfme/tests/configure/test_visual_infra.py
```diff
@@ -213,13 +213,13 @@ def test_infra_start_page(request, appliance, start_page):
...
 
 
 def test_host_noquads(request, set_host_quad):
-    navigate_to(Host, 'All')
-    assert visual.check_image_exists, "Image View Failed!"
+    view = navigate_to(Host, 'All')
```
izapolsk: ^^

cfme/tests/configure/test_visual_cloud.py
anurag03: This method is working, here I just want to check that quad is present or not. If I get None value when I use view.entities.get_all()[0].data then the quad icon is not displayed.
Correct me if I am missing something.

izapolsk: 1. this method will raise exception if there are no any entities
2. data isn't None if there are List or Tile entities
3. entity.data  returns empty dict instead of None
4. using get_all() instead of get_first_entity isn't optimal
5. etc

anurag03: I have made the changes based on your suggestions
Firstly by using view_selector it is switching to Grid view and then the check using get_first_entity is done.

izapolsk: okay, now it looks much better but you also have to take care of two more things:
1. there will be exception if that view doesn't contain any entities. please check whether this is ok or it should be handled somehow.
2. entity returns empty dict not None

mshriver: @izapolsk I think he's safe on the no entities, because if that's the case the provider for the test wasn't setup and it should fail.

mshriver:
Requested some in-code comments on the entity quadicon `data` check.

mshriver:
PRT results haven't collected yet, but run 16116 was successful for 57z and 58z

PR #5362, opened 2017-09-13
[WIP] Fix splitter usage

psav:
This will probably become a new WT type if one isn't already there
that tries to reset the splitter in the case that it isn't already in
the right place

Splitter tests and usage, do not fail correctly, we should actually take the burden off the tests and instead make the widget intelligent so that if the trees are not visible, we at least TRY to open the pslitter back to reasonable levels and THEN FAIL


dajoRH:
Lint report for commit b0c912b29b059242d0089b0c6e9db83f2f12418c:

- :bangbang: Line 133:101: **E501** *line too long (108 > 100 characters)*

**No commit flaws detected**.

*CFME QE Bot*

PR #5366, opened 2017-09-14
[1LP][RFR] test_timelines against 5.7.4 fixes

izapolsk:
{{pytest: cfme/tests/infrastructure/test_timelines.py --use-provider=vsphere6-nested}}

### DIFF
--- a/cfme/infrastructure/cluster.py
+++ b/cfme/infrastructure/cluster.py
@@ -317,7 +317,10 @@ class Details(CFMENavigateStep):
         """Navigate to the correct view"""
         self.prerequisite_view.toolbar.view_selector.select('List View')
         version = self.obj.appliance.version
-        cluster_name = self.obj.short_name if version >= '5.8.1.2' else self.obj.name
+        if (version >= '5.7.4' and version < '5.8') or version >= '5.8.1.2':
+            cluster_name = self.obj.short_name
+        else:
+            cluster_name = self.obj.name
         try:
             row = self.prerequisite_view.paginator.find_row_on_pages(
                 self.prerequisite_view.entities.table,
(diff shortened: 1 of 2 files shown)


cfme/infrastructure/cluster.py
```diff
@@ -317,7 +317,10 @@ def step(self, *args, **kwargs):
         """Navigate to the correct view"""
         self.prerequisite_view.toolbar.view_selector.select('List View')
         version = self.obj.appliance.version
-        cluster_name = self.obj.short_name if version >= '5.8.1.2' else self.obj.name
+        if (version >= '5.7.4' and version < '5.8') or version >= '5.8.1.2':
```
mshriver: Gross. Not on your part though @izapolsk.  Hope we don't run into too many of these.

izapolsk: I also hope so. but as you can see they merged some of 5.8 changes into 5.7.4

widgetastic_manageiq/__init__.py
```diff
@@ -1340,7 +1340,7 @@ def set_items_per_page(self, value):
 
         items_text = VersionPick({
             Version.lowest(): '{} items'.format(value),
-            '5.8.2': str(value),
+            '5.7.4': str(value),
```
jkrocil: What about `5.8.0` and `5.8.1` here - shouldn't it be `'{} items`? @izapolsk

izapolsk: @jkrocil, I assume we won't test older 5.8 versions. However, I can use if statement there if you believe those versions can be tested.

jkrocil: Yeh, I'd say it doesn't really matter too. I mainly asked because you did take `5.8.0` and `5.8.1` into consideration above.

PR #5371, opened 2017-09-14
[1LP][RFR] Cloud provision with tag test

rlbabyuk:
Purpose or Intent
=================
Cloud provivioning tests with selected tag

{{pytest: cfme/tests/cloud/test_provisioning.py -v -k 'test_provision_from_template or test_cloud_provision_with_tag' --long-running }}

### DIFF
--- a/cfme/tests/cloud/test_provisioning.py
+++ b/cfme/tests/cloud/test_provisioning.py
@@ -158,7 +158,7 @@ def provision_check(request, provider):
 
 
 @pytest.mark.parametrize('testing_instance', [True, False], ids=["Auto", "Manual"], indirect=True)
-def test_provision_from_template(appliance, provider, testing_instance, soft_assert):
+def test_provision_from_template(provider, testing_instance, soft_assert, provision_check):
     """ Tests instance provision from template
 
     Metadata:
@@ -647,4 +647,3 @@ def test_cloud_provision_with_tag(testing_instance, provision_check, soft_assert
     tags = instance.get_tags()
     assert any(tag.category.display_name == "Service Level" and tag.display_name == "Gold"
                for tag in tags), "Service Level: Gold not in tags ({})".format(str(tags))
-
--- a/cfme/tests/cloud/test_provisioning.py
+++ b/cfme/tests/cloud/test_provisioning.py
@@ -33,7 +33,7 @@ pytest_generate_tests = testgen.generate(
 
 
 @pytest.yield_fixture(scope="function")
-def testing_instance(request, setup_provider, provider, provisioning, vm_name):
+def testing_instance(request, setup_provider, provider, provisioning, vm_name, tag):
     """ Fixture to prepare instance parameters for provisioning
     """
     image = provisioning['image']['name']
@@ -55,11 +55,18 @@ def testing_instance(request, setup_provider, provider, provisioning, vm_name):
     recursive_update(inst_args, {'catalog': {'vm_name': vm_name}})
 
     # Check whether auto-selection of environment is passed
+    auto = False  # By default provisioning will be manual
     try:
-        auto = request.param
+        parameter = request.param
+        if parameter == 'tag':
+            inst_args['purpose'] = {
+                'apply_tags': ('{} *'.format(tag.category.display_name), tag.display_name)
+            }
+        else:
+            auto = parameter
     except AttributeError:
         # in case nothing was passed just skip
-        auto = False
+        pass
 
     # All providers other than Azure
     if not provider.one_of(AzureProvider):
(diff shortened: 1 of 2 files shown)


dajoRH:
Lint report for commit 93362d721cbcbd17d789ff243ba93929d43e6a59:

`cfme/tests/cloud/test_provisioning.py`:
- :heavy_exclamation_mark: Line 650:1: **W391** *blank line at end of file*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/cloud/test_provisioning.py
```diff
@@ -125,35 +125,47 @@ def vm_name(request):
     return random_vm_name('prov')
 
 
+@pytest.fixture(scope="function")
+def provision_check(request, provider):
```
mkoura: I don't see where ``request`` is being used.

cfme/tests/cloud/test_provisioning.py
```diff
@@ -615,3 +627,23 @@ def _finish_method():
         if "volume_id" in locals():  # To handle the case of 1st or 2nd assert
             if provider.mgmt.volume_exists(volume_id):
                 provider.mgmt.delete_volume(volume_id)
+
+
+def test_cloud_provision_with_tag(testing_instance, provision_check, soft_assert):
+    """ Tests tagging instance using provisioning dialogs.
```
mkoura: Add blank line after summary line.

cfme/tests/cloud/test_provisioning.py
rlbabyuk: corrected

cfme/tests/cloud/test_provisioning.py
rlbabyuk: done

cfme/tests/cloud/test_provisioning.py
```diff
@@ -125,35 +125,47 @@ def vm_name(request):
     return random_vm_name('prov')
 
 
+@pytest.fixture(scope="function")
+def provision_check(provider):
```
quarckster: Why do you need this fixture? Why not to  create a regular function `provision_check(provider, instance, inst_args, image)`?

rlbabyuk: don't have any good reason to use fixture here, so will convert it to function

rlbabyuk: Done

cfme/tests/cloud/test_provisioning.py
```diff
@@ -125,25 +125,21 @@ def vm_name(request):
...
 
-@pytest.mark.parametrize('testing_instance', [True, False], ids=["Auto", "Manual"], indirect=True)
-def test_provision_from_template(provider, testing_instance, soft_assert):
-    """ Tests instance provision from template
-
-    Metadata:
-        test_flag: provision
-    """
-    instance, inst_args, image = testing_instance
+def provision_check(provider, instance, inst_args, image):
```
quarckster: **REQUIRED**
After looking more deeply into this module I think this could be a fixture, but with name let's say `provisioned_instance`. Also it should take `testing_instance` fixture as a argument:
```python
@pytest.fixture
def provisioned_instance(provider, testing_instance):
...
```
After that in your tests you can do something like that:
```python
@pytest.mark.parametrize('testing_instance', [True, False], ids=["Auto", "Manual"], indirect=True)
def test_provision_from_template(provider, provisioned_instance, soft_assert):
    """ Tests instance provision from template

    Metadata:
        test_flag: provision
    """
    instance, inst_args, image = testing_instance
    instance = provision_check(provider, instance, inst_args, image)
```
What do you think?

rlbabyuk: Sounds good)  This will perfectly work for provisioning from a template, as for tagging, I updated testing_instance fixture to handle tag option. @quarckster, could you please check it, if it is ok?

dajoRH:
*CFME QE Bot*

cfme/tests/cloud/test_provisioning.py
```diff
@@ -154,7 +159,17 @@ def test_provision_from_template(appliance, provider, testing_instance, soft_ass
...
+
+
+@pytest.mark.parametrize('testing_instance', [True, False], ids=["Auto", "Manual"], indirect=True)
+def test_provision_from_template(provider, provisioned_instance, soft_assert):
+    """ Tests instance provision from template
+
+    Metadata:
+        test_flag: provision
+    """
+    soft_assert(provisioned_instance.does_vm_exist_on_provider(), "Instance wasn't provisioned")
```
izapolsk: OPTIONAL: it doesn't make sense to use soft_assert here because it is necessary only in case when you don't want to stop test even if this intermediate condition is False.
I'd replace it with regular assert

cfme/tests/cloud/test_provisioning.py
```diff
@@ -615,3 +630,23 @@ def _finish_method():
...
+
+    Steps:
+        * Open the provisioning dialog.
+        * Apart from the usual provisioning settings, pick a tag.
+        * Submit the provisioning request and wait for it to finish.
+        * Visit instance page, it should display the selected tags
+    Metadata:
+        test_flag: provision
+    """
+    soft_assert(provisioned_instance.does_vm_exist_on_provider(), "Instance wasn't provisioned")
```
izapolsk: same here

PR #5372, opened 2017-09-14
[1LP][RFR] Check for badness rework

psav:
We are running things in the browser without first checking for modal dialog, I decided it may be better to put all the precheck stuff inside the try block. This is a PoC really, and I think check_for_badness really needs to be redesigned.

### DIFF
--- a/cfme/utils/appliance/implementations/ui.py
+++ b/cfme/utils/appliance/implementations/ui.py
@@ -196,18 +196,7 @@ class CFMENavigateStep(NavigateStep):
         except (AttributeError, NoSuchElementException):
             return False
 
-    def check_for_badness(self, fn, _tries, nav_args, *args, **kwargs):
-        if getattr(fn, '_can_skip_badness_test', False):
-            # self.log_message('Op is a Nop! ({})'.format(fn.__name__))
-            return
-
-        # TODO: Uncomment after resolving the issue in widgetastic. Shouldn't be needed though :)
-        # if self.VIEW:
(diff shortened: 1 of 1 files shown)


dajoRH:
*CFME QE Bot*

mshriver:
I believe this is the root cause of a large number of failures in the 17.26.0-rc test run. Once an UnexpectedAlertPresentException is encountered in a browser session, subsequent tests on that session will fail to navigate.  I think we should merge this for the 17.26.0 release.

cfme/utils/appliance/implementations/ui.py
```diff
@@ -287,6 +276,20 @@ def check_for_badness(self, fn, _tries, nav_args, *args, **kwargs):
             self.go(_tries, *args, **go_kwargs)
             # If there is a rails error past this point, something is really awful
 
+    def check_for_badness(self, fn, _tries, nav_args, *args, **kwargs):
+        if getattr(fn, '_can_skip_badness_test', False):
```
jamesooden: **QUESTION:** What is fn?

mshriver: fn is 'function', check_for_badness is a wrapper for step/am_i_here and other NavigateStep methods.

psav: It's all a little tiny, insignificant;y small bit of a mess :)

mshriver:
PRT looks good, local testing looks good - I verified that with an unexpected modal/dialog present, navigation dismisses it.

PR #5373, opened 2017-09-14
[1LP][RFR] Added new smoke test for embedded ansible

lcouzens:
Added new smoke test for enabling embedded ansible.

### DIFF
--- a/cfme/utils/appliance/__init__.py
+++ b/cfme/utils/appliance/__init__.py
@@ -1548,6 +1548,16 @@ class IPAppliance(object):
         return output.success
 
     @property
+    def is_nginx_running(self):
+        output = self.ssh_client.run_command("systemctl status nginx")
+        return output.success
+
+    @property
+    def is_rabbitmq_running(self):
+        output = self.ssh_client.run_command("systemctl status rabbitmq-server")
+        return output.success
+
+    @property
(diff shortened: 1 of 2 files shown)


cfme/utils/appliance/__init__.py
```diff
@@ -1697,6 +1707,19 @@ def server_roles(self, roles):
         self.set_yaml_config(yaml)
         wait_for(lambda: self.server_roles == roles, num_sec=300, delay=15)
 
+    def enable_embedded_ansible_role(self):
```
quarckster: **REQUIRED**
I don't think it's a good idea to have this in appliance. We already have [`server_roles_enabled`](https://github.com/ManageIQ/integration_tests/blob/master/cfme/configure/configuration/__init__.py#L1326) function.

psav: The reason for this is because whilst the other roles seem to start up very quickly, embedded ansible takes a LOT longer, it was thought unfair and potentially performance degrading to increase the wait for timeout, plus as it is a setter, getter, not very easy to do. We consulted with ronny too on this and we thought the helper function for ansible made sense.

quarckster:
@lcouzens please sign your commits in order to run them in PRT

quarckster:
PRT failed.

PR #5375, opened 2017-09-14
[RFR] Fix utils import on post_jenkins_result

mshriver:


mshriver:
Not sure why my master wasn't showing this update, big old wtf.

PR #5378, opened 2017-09-15
[1LP][RFR] Remove some web_ui methods

psav:
Remove some web_ui calls


cfme/middleware/server.py
```diff
@@ -286,10 +286,12 @@ def shutdown_server(self, timeout=10, cancel=False):
     def restart_server(self):
         view = self.load_details(refresh=True)
         view.toolbar.power.item_select('Restart Server', handle_alert=True)
+        view.flash.assert_message('Restart initiated for selected server(s)')
```
mshriver: assert_success_message for tighter assertion matching what was in the tests?

PR #5384, opened 2017-09-15
[1LP][RFR] Update ansible.repositories, NavigatableMixin

mshriver:
Update for proper collections+appliance use

Fix tests where needed.

## PRT Results
Not expecting any failures, that's kinda the point here :)

### DIFF
--- a/cfme/ansible/credentials.py
+++ b/cfme/ansible/credentials.py
@@ -10,7 +10,7 @@ from widgetastic_patternfly import BootstrapSelect, Button, Dropdown, FlashMessa
 from cfme.base import Server
 from cfme.base.login import BaseLoggedInPage
 from cfme.exceptions import ItemNotFound
-from cfme.utils.appliance import Navigatable
+from cfme.utils.appliance import NavigatableMixin
 from cfme.utils.appliance.implementations.ui import navigator, navigate_to, CFMENavigateStep
 from cfme.utils.wait import wait_for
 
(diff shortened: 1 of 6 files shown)


cfme/tests/ansible/test_embedded_ansible_basic.py
```diff
@@ -69,25 +69,27 @@ def wait_for_ansible(appliance):
     appliance.wait_for_embedded_ansible()
 
 
-@pytest.yield_fixture(scope="module")
-def repository(wait_for_ansible):
-    repositories = RepositoryCollection()
+@pytest.yield_fixture(scope='function')
```
quarckster: Scope mismatch is here, it should be kept as `module`. You cannot use this fixture in "module" scoped fixture catalog_item.

cfme/ansible/credentials.py
```diff
@@ -217,11 +222,11 @@ def _wait_until_appeared():
         return Credential(name, credential_type, appliance=self.appliance, **credentials)
 
 
-class Credential(Navigatable):
+class Credential(NavigatableMixin):
     """A class representing one Embedded Ansible credential in the UI."""
-    def __init__(self, name, credential_type, collection=None, appliance=None, **credentials):
-        self.collection = collection or CredentialsCollection(appliance=appliance)
-        Navigatable.__init__(self, appliance=self.collection.appliance)
+    def __init__(self, name, credential_type, collection, appliance=None, **credentials):
```
quarckster: **REQUIRED**
`appliance` argument is not used here anymore, so it should be removed.

PR #5385, opened 2017-09-18
[1LP][RFR] Moved version to package so it can be used elsewhere

psav:
Version and some supporting tools are now in a new pacakge


psav:
This will break miq and miq-runtest withotu a pip-update so we need to email out and give notice. I also want mfalesni/jkrocil to approve it before we run with it

dajoRH:
*CFME QE Bot*

PR #5389, opened 2017-09-18
[1LP][RFR] Get rid of web_ui in infra provider

izapolsk:
{{ pytest: cfme/tests/infrastructure/test_providers.py  cfme/tests/infrastructure/test_timelines.py cfme/tests/infrastructure/test_cluster_analysis.py }}

1. This PR intends to remove web_ui traces from infra provider. This is necessary for successful testing 5.9/upstream.
2. It refactors clusters in order to use BaseEntitiesView instead of table.

### DIFF
--- a/cfme/infrastructure/cluster.py
+++ b/cfme/infrastructure/cluster.py
@@ -196,7 +196,7 @@ class Cluster(Pretty, NavigatableMixin, WidgetasticTaggable):
         col = self.appliance.rest_api.collections
         self._id = [
             cl.id
-            for cl in col.clusters.all
+            for cl in col.clusters
             if cl.name in (self._short_name, self.name) and cl.ems_id == self.provider.id
         ][-1]
 
--- a/cfme/infrastructure/cluster.py
+++ b/cfme/infrastructure/cluster.py
@@ -90,20 +90,6 @@ class ClusterAllView(ClusterView):
     including_entities = View.include(BaseEntitiesView, use_parent=True)
 
 
-class ClusterAllFromProviderView(ClusterView):
-    """The all view page for clusters open from provider detail page"""
-    @property
-    def is_displayed(self):
-        """Determine if this page is currently being displayed"""
-        return (
-            self.logged_in_as_current_user and
-            self.entities.title.text == '{p}(All Clusters)'.format(p=self.context['object'].name))
-
-    toolbar = View.nested(ClusterToolbar)
-    breadcrumb = BreadCrumb()
-    including_entities = View.include(BaseEntitiesView, use_parent=True)
-
-
 class ClusterDetailsView(ClusterView):
     """The details page of a cluster"""
     @property
@@ -340,13 +326,3 @@ class EditTagsFromDetails(CFMENavigateStep):
 
     def step(self):
         self.prerequisite_view.toolbar.policy.item_select('Edit Tags')
-
-# TODO: This doesn't seem to be used, and needs to be migrated to Widgetastic
-# @navigator.register(Cluster, 'DetailsFromProvider')
-# class DetailsFromProvider(CFMENavigateStep):
-    # def step(self, *args, **kwargs):
-        # """Navigate to the correct view"""
-        # navigate_to(self.obj.provider, 'Details')
-        # list_acc.select('Relationships', 'Show all managed Clusters', by_title=True,
-        #                 partial=False)
-        # sel.click(Quadicon(self.obj.name, self.obj.quad_name))
--- a/cfme/infrastructure/cluster.py
+++ b/cfme/infrastructure/cluster.py
@@ -288,8 +288,9 @@ class All(CFMENavigateStep):
 
     def resetter(self):
         """Reset the view"""
-        self.view.entities.paginator.check_all()
-        self.view.entities.paginator.uncheck_all()
+        if self.view.entities.paginator.exists:
+            self.view.entities.paginator.check_all()
+            self.view.entities.paginator.uncheck_all()
 
 
 @navigator.register(Cluster, 'Details')
--- a/cfme/infrastructure/cluster.py
+++ b/cfme/infrastructure/cluster.py
@@ -1,20 +1,16 @@
 """ A model of an Infrastructure Cluster in CFME
 
-
-:var page: A :py:class:`cfme.web_ui.Region` object describing common elements on the
-           Cluster pages.
 """
 from navmazing import NavigateToSibling, NavigateToAttribute
 from widgetastic.exceptions import NoSuchElementException
 from widgetastic.widget import View
 from widgetastic_manageiq import (Accordion, BreadCrumb, ItemsToolBarViewSelector, ManageIQTree,
-                                  PaginationPane, Search, SummaryTable, Table, Text, TimelinesView)
+                                  SummaryTable, Text, TimelinesView, BaseEntitiesView)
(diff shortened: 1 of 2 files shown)


dajoRH:
*CFME QE Bot*

dajoRH:
Lint report for commit 8ee0092f8a840ebdaf5f671a6923085de78bfe19:

`cfme/infrastructure/cluster.py`:
- :red_circle: Line 13:1: **F401** *'cfme.exceptions.ClusterNotFound' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

cfme/infrastructure/cluster.py
```diff
@@ -104,8 +87,21 @@ def is_displayed(self):
             self.entities.title.text == 'Clusters')
 
     toolbar = View.nested(ClusterToolbar)
-    entities = View.nested(ClusterEntities)
-    paginator = PaginationPane()
+    including_entities = View.include(BaseEntitiesView, use_parent=True)
+
+
+class ClusterAllFromProviderView(ClusterView):
```
mshriver: **Optional**
Consider adding a navigation destination to match, even though it has to go through provider details to get there.

cfme/infrastructure/cluster.py
```diff
@@ -104,8 +87,21 @@ def is_displayed(self):
...
+    """The all view page for clusters open from provider detail page"""
+    @property
+    def is_displayed(self):
+        """Determine if this page is currently being displayed"""
+        return (
+            self.logged_in_as_current_user and
+            self.entities.title.text == '{p}(All Clusters)'.format(p=self.context['object'].name))
```
mshriver: Shouldn't this be the provider name, not the cluster name? After clicking the clusters on the provider details page, I see the title:

"vSphere 6.5 (nested) (All Clusters)"

cfme/infrastructure/provider/__init__.py
```diff
@@ -191,14 +186,12 @@ def hosts(self):
 
     def get_clusters(self):
         """returns the list of clusters belonging to the provider"""
-        web_clusters = []
+        collection = ClusterCollection()
```
mshriver: provide appliance arg to ClusterCollection

izapolsk: there is @psav PR which fixes ClusterCollection and etc.

cfme/infrastructure/cluster.py
izapolsk: actually, this view is used in only one place -  provider.get_clusters() and it obtains provider as a context['object']. However, it will have to be fixed if I turn it into nav destination

cfme/infrastructure/cluster.py
izapolsk: okay, I'll add new nav destination

dajoRH:
*CFME QE Bot*

cfme/infrastructure/cluster.py
mshriver: Noted - I think having views defined in a module which assume the object for the view is a class from a different module is a tricky precedent to set.  We have other classes with 'AllFromProvider' destinations that are still registered against the local classes, and access the provider through the context object, they don't assume that the context object _is_ the provider.

cfme/infrastructure/cluster.py
mshriver: Thanks @izapolsk, will keep things consistent with other modules that have 'AllForProvider' destinations.

mshriver:
@izapolsk other than the `AllForProvider` nav destination to match the view, the PR and the PRT results look great! ping me once you've updated and I'll merge.

cfme/infrastructure/cluster.py
izapolsk: @mshriver, creating and using such view will be problematic due to collections. Collections know nothing about provider. So, I moved this view and new nav destination to infra provider.

cfme/infrastructure/cluster.py
izapolsk: in fact,  this is not actual now because this view was moved under infra provider

cfme/infrastructure/provider/__init__.py
izapolsk: psavs Pr was merged. So, this is not relevant any longer

izapolsk:
@mshriver, I've changed that AllForProvider view. Please take a look.

cfme/infrastructure/cluster.py
```diff
@@ -307,9 +288,8 @@ def step(self, *args, **kwargs):
 
     def resetter(self):
         """Reset the view"""
-        self.view.toolbar.view_selector.select('Grid View')
-        self.view.paginator.check_all()
-        self.view.paginator.uncheck_all()
+        self.view.entities.paginator.check_all()
```
psav: **REQUIRED:** I've seen issues here where we fall over if there are no datastores, so we should perhaps make it a conditional on the fact that the paginator is displayed?

cfme/infrastructure/cluster.py
```diff
@@ -319,19 +299,14 @@ class Details(CFMENavigateStep):
 
     def step(self, *args, **kwargs):
         """Navigate to the correct view"""
-        self.prerequisite_view.toolbar.view_selector.select('List View')
-        version = self.obj.appliance.version
-        if (version >= '5.7.4' and version < '5.8') or version >= '5.8.1.2':
-            cluster_name = self.obj.short_name
-        else:
-            cluster_name = self.obj.name
+        # todo: figure out why the same cfme version shows clusters with short and long name
```
psav: **QUESTION:** It randomly chooses a name?

cfme/infrastructure/provider/__init__.py
```diff
@@ -184,15 +194,10 @@ def hosts(self):
...
-        view = navigate_to(self, 'Details')
-        # todo: create nav location + view later
-        view.contents.relationships.click_at('Clusters')
-        icons = Quadicon.all(qtype='cluster')
+        view = navigate_to(self, 'Clusters')
         cluster_col = self.appliance.get(ClusterCollection)
-        for icon in icons:
-            web_clusters.append(cluster_col.instantiate(icon.name, self))
-        return web_clusters
+        # todo: to handle clusters on multiple pages
```
psav: right, good todo here, ```view.entities.get_all()``` should be modified to do this I think

cfme/infrastructure/cluster.py
izapolsk: clusters appear with either short name f.e. "Default" or long name "Default in Default". 
I thought it depends on MiQ version but I ran into cases when appliances of the same version showed either short names or long names. So, this piece of code tries to use long name if cluster with short name was not found.

Honestly, I think this needs more investigation but I'd like to get around this issue for investigation time.

cfme/infrastructure/provider/__init__.py
izapolsk: get_all(surf_pages=True) already does this. I can't recall why I added this todo instead of adding surf_pages param.

cfme/infrastructure/provider/__init__.py
mshriver: :+1: everywhere! :)

cfme/infrastructure/cluster.py
mshriver: I'm okay with that for now with the TODO.

PR #5397, opened 2017-09-19
[1LP][RFR] Fixed Collections for container Node

psav:
Fixed Collections for container Node

### DIFF
--- a/cfme/containers/node.py
+++ b/cfme/containers/node.py
@@ -18,7 +18,7 @@ from cfme.containers.provider import ContainersProvider, Labelable,\
 from cfme.exceptions import NodeNotFound
 from cfme.fixtures import pytest_selenium as sel
 from cfme.web_ui import CheckboxTable, toolbar as tb, InfoBlock, match_location
-from cfme.utils.appliance import BaseCollection, BaseInstance
+from cfme.utils.appliance import BaseCollection, BaseEntity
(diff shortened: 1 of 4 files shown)


cfme/containers/node.py
```diff
@@ -81,17 +84,15 @@ def is_displayed(self):
     paginator = PaginationPane()
 
 
-class Node(Taggable, Labelable, SummaryMixin, Navigatable):
+class Node(Taggable, Labelable, SummaryMixin, NavigatableMixin):
 
     PLURAL = 'Nodes'
 
     def __init__(self, name, provider, collection=None, appliance=None):
```
gshefer: I think we can remove ```appliance=None``` argument right? since we get this from the collection.
or maybe we can do instead:
```python
self.appliance = self.collection.appliance or appliance
```

cfme/tests/containers/test_table_views.py
```diff
@@ -66,8 +66,10 @@ def test_default_views(random_default_views):
 
 
 @pytest.mark.polarion('CMP-10570')
-def test_table_views():
+def test_table_views(appliance):
     for obj in objects_mapping.keys():
+        if "Collection" in obj.__name__:
```
gshefer: I would create some AbstractCollection class, inherit this in the collection classes and then check with isinstance. But anyway don't think it's critical. up to you.

psav: I think you are right, we spoke about doing it before, I will do it

PR #5399, opened 2017-09-19
[1LP][RFR]Ansible Tower request fix

sshveta:
Fix for the recent widgetastic changes in request.py


sshveta:
The failures are the ansible tower providers that failed credential validation . 
Test passes for ansible tower 310.

PR #5314, opened 2017-09-07
[1LP][RFR] Update Tests for Templates Change in Yaml

mshriver:
This goes with a CFME QE YAML merge request (# 297) that implements a 'templates' section for all providers, with 'small_template', 'big_template', etc, as relevant for the different providers.

The pairing to that is updating the `_get_template` method that supplies all of of the `small_template`, `big_template`, etc fixtures, and updating all of the tests to handle these fixtures providing a template AttrDict instead of just the template name.

## PRT Results
I am not addressing test failures directly here. I've created PRs #5400, #5401, #5429, #5413, #5384 to fix tests/modules that I found had bugs during testing of this PR.

{{pytest: --long-running --use-provider complete}}

PR #5400, opened 2017-09-19
[1LP][RFR] Update common.vm.wait_for_appear, provider+browser refresh

mshriver:
Pulled this out of #5314 where I found it was causing test failures.  refresh provider relationships and the UI browser in `BaseVM.wait_to_appear()`

## PRT Results
58z pass, true positive
57z pass, true positive
upstream waiting

{{pytest: --use-provider ec2west -k "test_delete_instance_appear_after_refresh" cfme/tests/cloud/test_delete_cloud_object.py }}

### DIFF
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -181,7 +181,7 @@ class BaseVM(Pretty, Updateable, PolicyProfileAssignable, Taggable, SummaryMixin
     #
     def check_compliance(self, timeout=240):
         """Initiates compliance check and waits for it to finish.
-        
+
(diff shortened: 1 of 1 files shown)


izapolsk:
I'm not sure that default tests use this function. so, I added one PRT test which uses it. PRT is restarted

psav:
Looks like this tested against vsphere55 which is being decommissioned, so it could be that the vm no longer exists on it. We should try against another provider maybe?

izapolsk:
added --use-provider and restarted PRT

PR #5401, opened 2017-09-19
[1LP][RFR] Update MyService.ui, services.Request, and settings.DefaultView, small fixes

mshriver:
Removed some originally committed code to add a web_ui.Quadicon.click() method, screw that.

Small tweaks to services/myservice, biggest being the use of RequestsView for redirect validation.

## PRT Results
* 57z: 75% passing, test_remove_template_provisioning is failing when requests succeed, working with @sshveta on finding/writing BZ, not really related 
* 58z: 88% passing, same test failure, but only on one provider here
* upstream: uncollected, stream ignored.

### DIFF
--- a/cfme/services/myservice/__init__.py
+++ b/cfme/services/myservice/__init__.py
@@ -2,9 +2,10 @@ import sentaku
 
 from cfme.utils.appliance import Navigatable
 from cfme.common import WidgetasticTaggable
+from cfme.utils.update import Updateable
 
 
-class MyService(Navigatable, WidgetasticTaggable, sentaku.modeling.ElementMixin):
+class MyService(Updateable, Navigatable, WidgetasticTaggable, sentaku.modeling.ElementMixin):
     """
         My Service main class to context switch between ui
         and ssui. All the below methods are implemented in both ui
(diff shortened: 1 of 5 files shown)


cfme/services/requests.py
```diff
@@ -221,9 +221,14 @@ def edit_request(self, values, cancel=False):
         view.flash.assert_no_error()
 
 
+class RequestsToolbar(View):
```
izapolsk: :+1:

cfme/services/myservice/__init__.py
```diff
@@ -2,9 +2,10 @@
 
 from cfme.utils.appliance import Navigatable
 from cfme.common import WidgetasticTaggable
+from cfme.utils.update import Updateable
 
 
-class MyService(Navigatable, WidgetasticTaggable, sentaku.modeling.ElementMixin):
+class MyService(Updateable, Navigatable, WidgetasticTaggable, sentaku.modeling.ElementMixin):
```
sshveta: 👍

cfme/services/requests.py
mshriver: MyService kinda needs this too, but I was trying to keep scope small. Only added for details, since there were 5 other widgets to change for the collection view toolbar.

mshriver:
I'd like to follow this PR with creating a `MyServicesToobar` to nest in `MyServicesView`, moving the dropdown/button widgets and removing their '_btn' suffixes.

PR #5396, opened 2017-09-19
[1LP][RFR] Fixed service template collection

psav:
Fixed, this one needs some looking through for usage of methods I changed

PR #5409, opened 2017-09-20
[1LP][RFR] Automate collections remediate

psav:
Automate collections remediate

### DIFF
--- a/cfme/automate/service_dialogs.py
+++ b/cfme/automate/service_dialogs.py
@@ -122,7 +122,7 @@ class Dialog(BaseEntity, Fillable):
     @cached_property
     def tabs(self, ):
         from .dialog_tab import TabCollection
-        return TabCollection(self)
+        return TabCollection(self.appliance, self)
 
     @property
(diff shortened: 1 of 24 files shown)


cfme/automate/service_dialogs.py
```diff
@@ -73,17 +73,16 @@ def is_displayed(self):
         )
 
 
-class DialogCollection(Navigatable):
+class DialogCollection(NavigatableMixin):
```
izapolsk: please make sure, that this is fixed in another PR

psav: I is already in another pr #5396

mshriver:
Moved to WIPTEST because of TypeErrors on the __init__ calls for mismatched args.

Most may be taken care of by the combination of this PR and #5396 but I'd like to see some justification for the failures.

dajoRH:
*CFME QE Bot*

mshriver:
PRT results in 16259 are good, failing tests aren't related.

PR #5411, opened 2017-09-20
[1LP][RFR] Enforce Good Collections

psav:
A proof of concept to enforce good collection objects

### DIFF
--- a/cfme/utils/appliance/__init__.py
+++ b/cfme/utils/appliance/__init__.py
@@ -1,3 +1,4 @@
+import inspect
 import json
 import logging
 import os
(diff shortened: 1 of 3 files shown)


cfme/utils/appliance/__init__.py
```diff
@@ -2625,8 +2626,24 @@ def __init__(self, appliance=None):
 
 
 class BaseCollection(NavigatableMixin):
-    pass
+    def __new__(cls, *args, **kwargs):
+        if inspect.getargspec(cls.__init__).args[1] != 'appliance':
```
izapolsk: what if appliance isn't passed at all?

mshriver: This is the IndexError that I saw that only comes from especially 'dumb' implementation/use

cfme/utils/appliance/__init__.py
```diff
@@ -2668,8 +2727,40 @@ def __init__(self, appliance=None):
 
 
 class BaseCollection(NavigatableMixin):
-    pass
+    """Class for helping create consistent Collections
+
+    The BaseCollection class is responsible for ensuring two things:
+
+    1) That the API consistently has the first argument passed to it
+    2) That that first argument is an appliance
```
mshriver: '...argument is an appliance instance'

cfme/utils/appliance/__init__.py
```diff
@@ -2668,8 +2727,40 @@ def __init__(self, appliance=None):
...
 
 
 class BaseEntity(NavigatableMixin):
-    pass
+    """Class for helping create consistent entitys
+
+    The BaseEntity class is responsible for ensuring two things:
+
+    1) That the API consistently has the first argument passed to it
+    2) That that first argument is a collection
```
mshriver: '...argument is a collection instance'

PR #5414, opened 2017-09-21
[RFR] Adds new tests for /api/service_orders/cart

mkoura:
{{pytest: -v --long-running -k TestServiceOrderCart cfme/tests/services/test_rest_services.py}}

### DIFF
--- a/cfme/tests/services/test_rest_services.py
+++ b/cfme/tests/services/test_rest_services.py
@@ -179,11 +179,8 @@ def cart(appliance):
 
     yield cart
 
-    try:
+    if cart.exists:
         cart.action.delete()
-    except Exception:
-        # can be deleted by test
-        pass
 
 
 def unassign_templates(templates):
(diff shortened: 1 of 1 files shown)


cfme/tests/services/test_rest_services.py
```diff
@@ -171,6 +171,21 @@ def vm(request, a_provider, appliance):
...
+@pytest.yield_fixture(scope="function")
+def cart(appliance):
+    cart = appliance.rest_api.collections.service_orders.action.create(name="cart")
+    assert_response(appliance)
+    cart = cart[0]
+
+    yield cart
+
+    try:
+        cart.action.delete()
```
psav: **REQUIRED**: We need to add a finalizer to the relevant tests, else the ones that don't may fail during delete and we'd not know

mkoura: @psav I changed it so the exception handling doesn't mask delete failures. For the majority of tests simply deleting the cart when it's present is enough. In tests where ``delete`` actions are tested there's checking for status codes etc. to make sure the delete actions behaves as expected.

PR #5415, opened 2017-09-21
[RFR] missed click in common/provider

izapolsk:
{{pytest: --long-running --use-provider=rhv41 -k "test_alert_vm_turned_on_more_than_twice_in_past_15_minutes" cfme/tests/control/test_alerts.py }}


izapolsk:
PRT failed against 5.7 correctly, there is a bug that EditEventView cannot be open in 5.7

PR #5421, opened 2017-09-21
[1LP][RFR] Fix Snapshot.delete for RHV provider

apagac:
This is a quick fix for `Snapshot.delete`.
After creating a VM on a RHV provider, there already is one RHV-created snapshot, the default one. When we create another snapshot using automation, it does get created, but it is NOT active and it is NOT highlighted. This behaviour differs from what we can see when working with vmware providers, where most of snapshot testcases are runnning, and is causing failures.
The fix is simple, I make sure we are currently on the snapshot we created and not on the snapshot that gets created by RHV. This way, it works with RHV as well as vsphere providers.

PS: this is only a quick remedy for jenkins failures. For the future, more complex refactoring in snapshot area will be needed. I need to change the way we interact with snapshots so that these kind of problems won't appear anymore.

{{pytest: -v --long-running --use-provider vsphere65-nested --use-provider rhv41 -k test_snapshot_crud}}

### DIFF
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -533,9 +533,10 @@ class Vm(VM):
         @property
         def exists(self):
             self._nav_to_snapshot_mgmt()
+            title = self.description if self.vm.provider.one_of(RHEVMProvider) else self.name
             try:
                 self.snapshot_tree.find_path_to(
-                    re.compile(r"{}.*?".format(self.name or self.description)))
+                    re.compile(r"{}.*?".format(title)))
                 return True
             except CandidateNotFound:
                 return False
@@ -564,9 +565,10 @@ class Vm(VM):
                 bool: True if snapshot is active, False otherwise.
             """
             self._nav_to_snapshot_mgmt()
+            title = self.description if self.vm.provider.one_of(RHEVMProvider) else self.name
             try:
-                self._click_tree_path(self.name or self.description)
-                if sel.is_displayed_text("{} (Active)".format(self.name or self.description)):
+                self._click_tree_path(title)
+                if sel.is_displayed_text("{} (Active)".format(title)):
                     return True
             except CandidateNotFound:
                 return False
@@ -590,6 +592,10 @@ class Vm(VM):
 
         def delete(self, cancel=False):
             self._nav_to_snapshot_mgmt()
+
+            title = self.description if self.vm.provider.one_of(RHEVMProvider) else self.name
+            self._click_tree_path(title)
+
             toolbar.select('Delete Snapshots', 'Delete Selected Snapshot', invokes_alert=True)
             sel.handle_alert(cancel=cancel)
             if not cancel:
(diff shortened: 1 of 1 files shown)


apagac:
PRT consistently fails with 5.7z on test_black_console. I don't think the change introduced in this PR has anything to do with that test. Moving to RFR.

cfme/infrastructure/virtual_machines.py
```diff
@@ -590,6 +590,9 @@ def create(self, force_check_memory=False):
 
         def delete(self, cancel=False):
             self._nav_to_snapshot_mgmt()
+
+            self._click_tree_path(self.name or self.description)
```
mshriver: Is name/description display version dependent? randomly selected by UI?

apagac: The snapshot in the tree is named differently when run against different providers. When running against vmware provider, the snapshot name is based on the name we gave it when it was created. When run against RHV provider, the snapshot does _not_ have a name, it has only description, and the snapshot name in the tree is based on that description.

apagac:
Per discussion with @mshriver, I'm going to change this to something like:
`title = self.description if self.vm.provider.one_of(RHEVMProvider) else self.name`
`self._click_tree_path(title)`

cfme/infrastructure/virtual_machines.py
mshriver: **Required**

Gotcha, thanks for explanation @apagac.

I'd like to see a simple change here to select `self.name` or `self.description` based on the type of provider found by `self.vm.provider` - using Provider.one_of() method to verify type.

That way we are selecting the appropriate field based on the provider type, and have tighter assertion over the display and behavior of the UI.  This potentially hides bugs or changes in the UI between versions.

Its done this way elsewhere in the file (actions property) but I don't think that's a good example, and would argue it should be selected based on provider type as well. The change to actions doesn't need to be made here.

apagac:
Refactored all occurrences of `self.name or self.description`. Manually tested all changed methods with vsphere55 and rhv41, all PASS.

cfme/infrastructure/virtual_machines.py
```diff
@@ -564,9 +565,11 @@ def active(self):
                 bool: True if snapshot is active, False otherwise.
             """
             self._nav_to_snapshot_mgmt()
+            title = self.description if self.vm.provider.one_of(RHEVMProvider) else self.name
+            self._click_tree_path(title)
```
mshriver: This is a duplicate of the call that's inside the try block, think you want to remove this one.

mshriver:
I updated PR description pytest line to include rhv41 provider

cfme/infrastructure/virtual_machines.py
apagac: You're right, missed that.

mshriver:
PRT results look good actually, upstream failure is a widgetastic.patternfly bug (#23), and the 57z failure was during teardown.

cfme/infrastructure/virtual_machines.py
```diff
@@ -533,9 +533,10 @@ def _nav_to_snapshot_mgmt(self):
         @property
         def exists(self):
             self._nav_to_snapshot_mgmt()
+            title = self.description if self.vm.provider.one_of(RHEVMProvider) else self.name
```
psav: This should be a bug imho

PR #5422, opened 2017-09-21
[1LP][WIP] Fix test_snapshot_crud for RHV

apagac:
This is a quick fix for `test_snapshot_crud` mainly for RHV provider.
The `Snapshot.delete` method does not wait for the deletion process to complete. When the test is run against RHV provider, this causes error at the end of the test, when automation is trying to delete the test VM. The VM can't be deleted, because there is pending snapshot operation.
Proposed fix is to wait for the snapshot deletion inside the test.

### DIFF
--- a/cfme/tests/infrastructure/test_snapshot.py
+++ b/cfme/tests/infrastructure/test_snapshot.py
@@ -118,6 +118,11 @@ def test_snapshot_crud(small_test_vm, provider):
         snapshot = new_snapshot(small_test_vm)
     snapshot.create()
     snapshot.delete()
+    # The Snapshot.delete method does not wait for snapshot deletion, it just initializes
+    # the process. We however need to wait for the process to complete, otherwise there
+    # may be a failure when trying to delete small_test_vm from the provider. The VM can't be
+    # deleted, because there is a pending snapshot operation. Does occur on RHV provider.
+    wait_for(lambda: not snapshot.exists, num_sec=300, delay=20, fail_func=snapshot.refresh)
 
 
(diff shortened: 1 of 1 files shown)


apagac:
Depends on PR #5421

apagac:
None of the PRT failures are relevant to test_snapshot_crud, this test passes. Moving to RFR.

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -118,6 +118,11 @@ def test_snapshot_crud(small_test_vm, provider):
         snapshot = new_snapshot(small_test_vm)
     snapshot.create()
     snapshot.delete()
+    # The Snapshot.delete method does not wait for snapshot deletion, it just initializes
+    # the process. We however need to wait for the process to complete, otherwise there
+    # may be a failure when trying to delete small_test_vm from the provider. The VM can't be
+    # deleted, because there is a pending snapshot operation. Does occur on RHV provider.
+    wait_for(lambda: not snapshot.exists, num_sec=300, delay=20, fail_func=snapshot.refresh)
```
psav: This wait should be inside the delete, with the option to not_wait, but to wait by default

apagac: This change requires review of all existing snapshot tests and possibly changing the way we check for snapshot deletion.

apagac:
Opened PR #5442 that addresses also the problem with RHV. I recommend we close this PR.

apagac:
Closing in favour of #5442

PR #5423, opened 2017-09-21
[RFR] test_snapshot.py test_snapshot_crud RHV Workaround

slaykovsky:
Hi!
When there's RHV as a provider the newly created snapshot is not getting selected after creation.
So this changes are only to select this snapshot to delete it then.


slaykovsky:
https://github.com/ManageIQ/integration_tests/pull/5421
https://github.com/ManageIQ/integration_tests/pull/5422

PR #5428, opened 2017-09-21
[1LP][RFR] Speeds up initial test start by a lot (REST conversion)

psav:
...at least in remote locations

### DIFF
--- a/cfme/utils/appliance/__init__.py
+++ b/cfme/utils/appliance/__init__.py
@@ -1740,6 +1740,26 @@ class IPAppliance(object):
             server id, server zone id)``
         """
         try:
+            servers = self.rest_api.collections.servers.all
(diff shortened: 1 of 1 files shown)


cfme/utils/appliance/__init__.py
```diff
@@ -1740,6 +1740,26 @@ def configuration_details(self):
             server id, server zone id)``
         """
         try:
+            servers = self.rest_api.collections.servers.all
```
mkoura: I would rather use ``list(self.rest_api.collections.servers)`` to make sure the data is up-to-date (``.all`` will not reload the data).

psav: Right but it's in a cached property anyway, so we are only doing it once anyway at the very beginning of the run

PR #5436, opened 2017-09-22
[1LP][RFR] Minor fixes which will help failing tests pass

kedark3:
Purpose or Intent
=================

__Extending__ vm_console class to show some logging and send "\n" if the screen_text is coming out to be blank, which happens with RHOS7-GA Cirros VMs, they go into screensaver mode causing a blank "Black" screen to be displayed and test fails even if console was displayed. Sending "\n' causes screen to wake up.

{{ pytest: cfme/tests/cloud_infra_common/test_html5_vm_console.py --use-provider rhos7-ga --use-provider vsphere6-nested --use-provider rhv41}}

### DIFF
--- a/cfme/common/vm_console.py
+++ b/cfme/common/vm_console.py
@@ -168,6 +168,8 @@ class VMConsole(Pretty):
         after_height = self.selenium.get_window_size()['height']
         fullscreen_btn.click()
         self.switch_to_console()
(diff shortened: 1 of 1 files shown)


kedark3:
Dowstream failure reason as follows: 
RHV41 - This is linked to card I have in Jira about updating the Console_template on RHV41( RHCFQE-4563)
VMware- https://bugzilla.redhat.com/1492246

PR #5442, opened 2017-09-25
[1LP][WIP] Add wait to Snapshot.delete

apagac:
This adds waiting for completion of delete action for snapshots. The `Snapshot.delete` method did not wait for the snapshot to be actually deleted, it just initiated the action. This required changes in some of the snapshot tests that were checking for successful deletion.

I've decided not to add waiting to `Snapshot.delete_all`, since this method can not wait for deletion of all the snapshots that are currently on a test VM. It could wait for the snapshot that it has been called on, but this would not ensure the deletion of all snapshots. I modified the tests accordingly.

This solves the problem that PR #5422 has been originally created for.

PRT results of changed tests:
 - test_snapshot_crud: PASS for both vsphere and RHV providers
 - test_delete_all_snapshots: PASS
 - test_operations_suspended_vm: PASS
 - test_operations_powered_off_vm: 5.8 PASS, 5.7 FAIL at creating first snapshot (unrelated)
   - PASS after rerun

{{pytest: -v --long-running --use-provider vsphere65-nested -k test_operations_powered_off_vm}}

### DIFF
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -602,6 +602,7 @@ class Vm(VM):
             if not cancel:
                 flash.assert_message_match('Remove Snapshot initiated for 1 '
                                            'VM and Instance from the CFME Database')
+            wait_for(lambda: not self.exists, num_sec=300, delay=20, fail_func=sel.refresh)
 
         def delete_all(self, cancel=False):
             self._nav_to_snapshot_mgmt()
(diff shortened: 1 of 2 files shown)


jkrocil:
`...this method can not wait for deletion of all the snapshots that are currently on a test VM.` Why not? We can discuss this offline but I don't like the difference in behavior between `delete` and `delete_all` and if there is anything we can do to keep the behavior the same, we should. Please, ping me tomorrow so we can discuss it.

PR #5443, opened 2017-09-25
[RFR] fix for delete clusters method

izapolsk:
This PR is intended to fix errors like 

> AttributeError'ListView' object has no attribute 'table'

{{pytest: --long-running --use-provider=vsphere55 cfme/tests/infrastructure/test_delete_infra_object.py}}


izapolsk:
PRT errors aren't relevant to the fix

PR #5425, opened 2017-09-21
[1LP][RFR] get_tag() multiple tags handling + vm, template classes update

rlbabyuk:
Purpose or Intent
=================
Update SummaruTable  and SummaryForm widgets to handle multiple tags on details page.
Update VM, Template classes to use WidgetasticTaggable

{{pytest: -v -k 'test_control_action_run_ansible_playbook_in_requests or test_service_ansible_playbook_tagging or test_keypair_add_and_remove_tag or test_tag_vis_vm or test_tagvis_infra_object' --long-running}}
PRT: some of test_tagvis_infra_object tests can fail but those will be refactored this sprint to be more stable

PR #5444, opened 2017-09-25
[1LP][RFR] Fix for test_tag_vis_vm

rlbabyuk:
Purpose or Intent
=================
Fixing test_tag_vis_vm test

{{pytest: -v -k test_tag_vis_vm --long-running}}


rlbabyuk:
test_tag_vis_vm[rhv41]  is failing as duplication vms was found, this will be fixed in scope https://github.com/ManageIQ/integration_tests/pull/5425, (vm, template classes will be updated to use widgetastic taggable + new navigation step will be added)

PR #5446, opened 2017-09-25
[1LP][RFR] This PR will fix host_drift_analysis.py

sbulage:


cfme/infrastructure/host.py
```diff
@@ -403,6 +402,14 @@ def _select_rows(indexes):
             drift_analysis_view.toolbar.all_attributes.click()
         return drift_analysis_view.drift_analysis(drift_section).is_changed
 
+    def tag(self, tag):
+        navigate_to(self, 'EditTags')
+        self.add_tag(tag)
+
+    def untag(self, tag):
+        navigate_to(self, 'EditTags')
+        self.remove_tag(tag)
```
quarckster: This methods aren't needed anymore. `Host` object is `WidgetasticTaggable` and it inherits `add_tag` and `remove_tag`. You should refactor `test_host_drift_analysis.py` instead of adding extra methods.

sbulage: ok @quarckster updating file accordingly.

sbulage:
Add/remove tag is working fine on Host locally (miq shell).

PR #5448, opened 2017-09-25
[WIPTEST] Add pytest params to access controls tests when testing edit/delete operations on default groups and roles

LandoCalrizzian:
Add pytest parameters to validate against all default RBAC groups and roles any test that attempts to delete OR edit default groups and roles

{{pytest: -v -k 'test_delete_default_group or test_edit_default_group or test_delete_default_roles or test_edit_default_roles'}}

### DIFF
--- a/cfme/configure/access_control.py
+++ b/cfme/configure/access_control.py
@@ -840,6 +840,27 @@ class Group(Updateable, Pretty, Navigatable):
         except CandidateNotFound:
             return False
 
+    def get_default_group_names(self):
+        if self.appliance.version < '5.8':
+            group_names = [
+                "EvmGroup-administrator", "EvmGroup-approver", "EvmGroup-auditor",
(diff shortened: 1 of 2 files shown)


cfme/configure/access_control.py
```diff
@@ -840,6 +840,27 @@ def exists(self):
         except CandidateNotFound:
             return False
 
+    def get_default_group_names(self):
```
quarckster: **REQUIRED**
This method should be a property and probably be a part of `GroupCollection`. Until it's not implemented you can just put this into the tests.

cfme/configure/access_control.py
```diff
@@ -1129,6 +1150,27 @@ def set_role_product_features(self, view, product_features):
             feature_update = True
         return feature_update
 
+    def get_default_role_names(self):
```
quarckster: **REQUIRED**
Here as well.

PR #5450, opened 2017-09-25
[WIP] Fix broken tests

shalomnaim1:
{{pytest: cfme/tests/containers/test_breadcrumbs.py  -v --use-provider cm-env2 }}

During the migration to collection some tests fail when using node object due to broken infra code.
In this PR I fixed the broken code

### DIFF
--- a/cfme/containers/node.py
+++ b/cfme/containers/node.py
@@ -117,7 +117,8 @@ class Node(Taggable, Labelable, SummaryMixin, BaseEntity, Navigatable):
         """Generating random instances."""
         node_list = provider.mgmt.list_node()
         random.shuffle(node_list)
-        return [cls(obj.name, provider, collection=NodeCollection(appliance or current_appliance()))
+        curr_collection = NodeCollection(appliance or current_appliance())
+        return [curr_collection.instantiate(obj.name, provider)
                 for obj in itertools.islice(node_list, count)]
 
 
--- a/cfme/containers/node.py
+++ b/cfme/containers/node.py
@@ -117,7 +117,7 @@ class Node(Taggable, Labelable, SummaryMixin, BaseEntity):
         """Generating random instances."""
         node_list = provider.mgmt.list_node()
         random.shuffle(node_list)
-        return [cls(obj.name, provider, appliance=appliance)
+        return [cls(obj.name, provider, collection=NodeCollection(appliance))
                 for obj in itertools.islice(node_list, count)]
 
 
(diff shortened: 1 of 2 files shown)


cfme/containers/provider/__init__.py
```diff
@@ -326,13 +326,6 @@ def report_kibana_failure():
         return logging_url
 
 
-class ProviderDetailsView(BaseLoggedInPage, LoggingableView):
-
-    @property
-    def is_displayed(self):
-        return match_page(summary="{} (Summary)".format(self.obj.name))
-
-
```
shalomnaim1: This view already exists on providers_view so I deleted it to avoid duplication

cfme/containers/node.py
```diff
@@ -117,7 +117,7 @@ def get_random_instances(cls, provider, count=1, appliance=None):
         """Generating random instances."""
         node_list = provider.mgmt.list_node()
         random.shuffle(node_list)
-        return [cls(obj.name, provider, appliance=appliance)
+        return [cls(obj.name, provider, collection=NodeCollection(appliance or current_appliance()))
```
quarckster: **REQUIRED**
There should be one common collection, but you create a separate collection for each instance. Moreover I think this method `get_random_instances` should be a part of `NodeCollection`.

shalomnaim1: Is this good solution?

`    @classmethod
    def get_random_instances(cls, provider, count=1, appliance=None):
        """Generating random instances."""
        node_list = provider.mgmt.list_node()
        random.shuffle(node_list)
        curr_collection=NodeCollection(appliance or current_appliance())
        return [cls(obj.name, provider, collection=curr_collection)
                for obj in itertools.islice(node_list, count)]
`

I agree that we have to move `get_random_instances` to `NodeCollection` but i will do it in a separated PR, this is massive change and which will take a lot of time

quarckster: It looks slightly better, but nevertheless I wouldn't approve such collections using. Let's cast other contributors @jkrocil, @psav, @mfalesni.

shalomnaim1: More precisely
 
`
        curr_collection = NodeCollection(appliance or current_appliance())
        return [curr_collection.instantiate(obj.name, provider) 
                for obj in itertools.islice(node_list, count)]
`

dajoRH:
*CFME QE Bot*

dajoRH:
*CFME QE Bot*

PR #4974, opened 2017-07-11
[RFR][1LP] Use widgetastic automation for access control tests and resolve test failures on CFME default users/groups/roles

LandoCalrizzian:
This PR updates these tests so that they all call the widgetastic methods and adds a new exception (RBACOperationBlocked) that will be thrown when an action(create, delete, edit) is blocked due to operating on an object that is immutable (default user, group, role).  Tests can now catch RBACOperationBlocked without having to rely on checking for a specific flash message to see if RBAC allowed an operation.  The state of the page will not be modified so a test can check the page elements after the exception

{{pytest: -v -k 'test_delete_default_group or test_delete_default_roles or test_delete_default_user or test_delete_group_with_assigned_user or test_delete_roles_with_assigned_group or test_edit_default_group or test_edit_default_roles or test_edit_sequence_usergroups or test_group_duplicate_name or test_rolename_duplicate_validation or test_user_duplicate_name or test_current_user_login_delete' }}

PR #5451, opened 2017-09-25
[1LP][RFR] Update test_delete_group_with_assigned_user to catch RBACOperationBlocked exception

LandoCalrizzian:
PR #4974 added an RBACOperationBlocked exception that will be thrown whenever an operation is performed but blocked due to permissions. This test was mistakenly left out of that update. This moves the flash message check out of the test and adds Group.delete() where it will be checked against a list of messages that specify that the delete operation was blocked

{{pytest: -v -k 'test_delete_group_with_assigned_user or test_delete_default_group or test_group_crud'}}

### DIFF
--- a/cfme/configure/access_control.py
+++ b/cfme/configure/access_control.py
@@ -717,9 +717,13 @@ class Group(Updateable, Pretty, Navigatable):
                 for currently selected group
         """
         flash_success_msg = 'EVM Group "{}": Delete successful'.format(self.description)
-        flash_blocked_msg = (
-            "EVM Group \"{}\": "
-            "Error during delete: A read only group cannot be deleted.".format(self.description))
+        flash_blocked_msg_list = [
+            ('EVM Group "{}": '
+                'Error during delete: A read only group cannot be deleted.'.format(
+                    self.description)),
+            ('EVM Group "{}": Error during delete: '
+                'The group has users assigned that do not '
+                'belong to any other group'.format(self.description))]
         delete_group_txt = 'Delete this Group'
 
         view = navigate_to(self, 'Details')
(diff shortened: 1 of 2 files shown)


cfme/configure/access_control.py
```diff
@@ -716,9 +716,12 @@ def delete(self, cancel=True):
...
         flash_success_msg = 'EVM Group "{}": Delete successful'.format(self.description)
-        flash_blocked_msg = (
-            "EVM Group \"{}\": "
-            "Error during delete: A read only group cannot be deleted.".format(self.description))
+        flash_blocked_msg_list = [
+            ("EVM Group \"{}\": "
+            "Error during delete: A read only group cannot be deleted.".format(self.description)),
+            ("EVM Group \"{}\": Error during delete: "
+                "The group has users assigned that do not "
+                "belong to any other group".format(self.description))]
```
mkoura: REQUIRED: unify the indentation

cfme/configure/access_control.py
```diff
@@ -716,9 +716,13 @@ def delete(self, cancel=True):
                 for currently selected group
         """
         flash_success_msg = 'EVM Group "{}": Delete successful'.format(self.description)
-        flash_blocked_msg = (
-            "EVM Group \"{}\": "
-            "Error during delete: A read only group cannot be deleted.".format(self.description))
+        flash_blocked_msg_list = [
+            ("EVM Group \"{}\": "
```
mshriver: **Required**

Use single quotes so you don't have to escape the double quotes.

```
'EVM Group "{}": "Error during delete: A read only group cannot be deleted.'

LandoCalrizzian: Good catch. updated

PR #5453, opened 2017-09-25
[1LP][RFR] Getting rid of Quadicons

gshefer:
{{ pytest: cfme/tests/containers/test_provider_configuration_menu.py cfme/tests/containers/test_provider_configuration_menu.py --use-provider cm-env2 }}

- Get rid of Quadicons and use widgetastic.
- test_provider_configuration_menu has been updated accordingly.

### DIFF
--- a/cfme/containers/provider/__init__.py
+++ b/cfme/containers/provider/__init__.py
@@ -22,7 +22,7 @@ from cfme.common.provider_views import BeforeFillMixin,\
     ContainersProviderEditView, ProvidersView, ProviderDetailsView
 from cfme.base.credential import TokenCredential
 from cfme.web_ui import (
-    Quadicon, toolbar as tb, InfoBlock, Region, match_location, PagedTable)
+    toolbar as tb, InfoBlock, Region, match_location, PagedTable)
 from cfme.utils import version
 from cfme.utils.appliance import Navigatable
 from cfme.utils.appliance.implementations.ui import navigator, CFMENavigateStep, navigate_to
@@ -326,7 +326,7 @@ class LoggingableView(View):
         return logging_url
 
 
-class ProviderDetailsView(BaseLoggedInPage, LoggingableView):
+class ContainersProviderDetailsView(ProviderDetailsView, LoggingableView):
 
     @property
     def is_displayed(self):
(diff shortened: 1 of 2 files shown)


cfme/common/provider_views.py
```diff
@@ -270,11 +270,19 @@ def is_displayed(self):
...
+    """
+    pass
+
+
 class ContainersProvidersView(ProvidersView):
     """
      represents Main view displaying all Containers providers
     """
     table = Table(locator="//div[@id='list_grid']//table")
+    including_entities = View.include(ContainersProviderEntities, use_parent=True)
```
izapolsk: REQUIRED: you don't need this, since ProvidersView already includes ProviderEntities

cfme/containers/provider/__init__.py
```diff
@@ -351,7 +352,8 @@ class Edit(CFMENavigateStep):
     prerequisite = NavigateToSibling('All')
 
     def step(self):
-        sel.check(Quadicon(self.obj.name, self.obj.quad_name).checkbox())
+        self.prerequisite_view.entities.get_entity(self.obj.name,
```
izapolsk: I guess it would be better to call Details and click appropriate configuration item in step
`prerequisite = NavigateToSibling('Details')`
and
`view.configuration.item_select('blabla')`

cfme/containers/provider/__init__.py
```diff
@@ -368,7 +370,8 @@ class EditTags(CFMENavigateStep):
...
```
izapolsk: ^^

cfme/tests/containers/test_provider_configuration_menu.py
```diff
@@ -17,10 +19,16 @@
...
+    If no such item found - return None."""
+    view = navigate_to(ContainersProvider, 'All')
     tb.select('Grid View')
-    Quadicon.select_first_quad()
-    return Quadicon.get_first_quad_title()
+    try:
+        entity = view.entities.get_first_entity()
+    except ItemNotFound:
+        return
+    entity.check()
```
izapolsk: I don't recall whether selection is restored after navigation.
Please do paginator.check_all()/uncheck_all() if items are selected after navigation.

cfme/containers/provider/__init__.py
gshefer: about `view.configuration.item_select('blabla')`  -  I would like this change in different PR.

cfme/tests/containers/test_provider_configuration_menu.py
gshefer: This is already done here - https://github.com/ManageIQ/integration_tests/blob/master/cfme/containers/provider/__init__.py#L274.

cfme/containers/provider/__init__.py
```diff
@@ -348,10 +349,9 @@ def resetter(self):
 @navigator.register(ContainersProvider, 'Edit')
 class Edit(CFMENavigateStep):
     VIEW = ContainersProviderEditView
-    prerequisite = NavigateToSibling('All')
+    prerequisite = NavigateToSibling('Details')
 
     def step(self):
-        sel.check(Quadicon(self.obj.name, self.obj.quad_name).checkbox())
         cfg_btn('Edit Selected Containers Provider')
```
izapolsk: usually such items have different names on Details and All pages.
Please double check that this item name is correct

PR #5459, opened 2017-09-26
[1LP][RFR] Fix value error

slaykovsky:
Fixing `check_power_option` function's Value Error as `vm.GUEST_RESTART` does not exist in `must_be_available` dictionary.


cfme/tests/infrastructure/test_vm_power_control.py
```diff
@@ -88,19 +88,20 @@ def if_scvmm_refresh_provider(provider):
 
 
 def check_power_options(provider, soft_assert, vm, power_state):
-    must_be_available = {'on': [vm.POWER_OFF, vm.SUSPEND, vm.RESET],
```
izapolsk: @slaykovsky, please don't reformat old code without real need in future.

PR #5463, opened 2017-09-27
[1LP][RFR]Fix to resolve test case failure.

anurag03:
Purpose or Intent
=================
Made few changes to resolve jenkin job failures.

{{pytest: -v -k 'test_cloud_start_page'}}

### DIFF
--- a/cfme/tests/configure/test_visual_cloud.py
+++ b/cfme/tests/configure/test_visual_cloud.py
@@ -11,7 +11,7 @@ from cfme.cloud.instance import Instance
 from cfme.cloud.keypairs import KeyPairCollection
 from cfme.cloud.stack import StackCollection
 from cfme.cloud.tenant import TenantCollection
-from cfme.web_ui import toolbar as tb, match_location
+from cfme.web_ui import toolbar as tb
 from cfme.utils.appliance import BaseCollection
 from cfme.utils.appliance.implementations.ui import navigate_to
 
--- a/cfme/tests/configure/test_visual_cloud.py
+++ b/cfme/tests/configure/test_visual_cloud.py
@@ -3,7 +3,7 @@ import pytest
 
 from cfme import test_requirements
 from cfme.configure.settings import visual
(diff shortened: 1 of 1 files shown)


cfme/tests/configure/test_visual_cloud.py
```diff
@@ -151,17 +151,20 @@ def test_cloud_list_page_per_item(request, page, set_list, appliance):
...
+    start = "" if appliance.version < '5.8' else "Compute / "
+    start_page = "{}{}".format(start, start_page)
     request.addfinalizer(set_default_page)
     visual.login_page = start_page
     appliance.server.logout()
     appliance.server.login_admin()
-    match_args = landing_pages[start_page]
-    assert match_location(**match_args), "Landing Page Failed"
+    logged_in_page = navigate_to(appliance.server, 'LoggedIn')
+    assert logged_in_page.is_displayed
```
rlbabyuk: Maybe I don't fully understand the test, but we need to navigate to different cloud-related pages and check is it is a correct page, correct me if I'm wrong. But here we every time navigate to login page and check login page is opened

anurag03: Yes, with this test we are checking if there is an issue with the landing page or not.

cfme/tests/configure/test_visual_cloud.py
```diff
@@ -1,18 +1,16 @@
 # -*- coding: utf-8 -*-
 import pytest
 
-
 from cfme import test_requirements
 from cfme.configure.settings import visual
-from cfme.cloud.availability_zone import AvailabilityZone
-from cfme.cloud.provider import CloudProvider
-from cfme.cloud.flavor import Flavor
+from cfme.cloud.availability_zone import AvailabilityZone, AvailabilityZoneView
```
rlbabyuk: **Required:** here should be AvailabilityZoneAllView as AvailabilityZoneView matches only header

anurag03: @rlbabyuk I have made the changes

PR #5467, opened 2017-09-27
[WIP] Skip ansible provider tests

shalomnaim1:
{{pytest: cfme/tests/containers/test_manageiq_ansible_providers.py}}

We have to skip the following test for now
In this PR, I marked all of them as skiped

### DIFF
--- a/cfme/tests/containers/test_manageiq_ansible_providers.py
+++ b/cfme/tests/containers/test_manageiq_ansible_providers.py
@@ -29,6 +29,7 @@ def ansible_providers():
     remove_tmp_files()
 
 
+@pytest.mark.skip(reason="Needs pzagal review")
 @pytest.mark.polarion('CMP-xxx')
 @pytest.mark.usefixtures('has_no_containers_providers')
 def test_manageiq_ansible_add_provider_ssl(ansible_providers, provider):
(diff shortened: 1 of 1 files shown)


cfme/tests/containers/test_manageiq_ansible_providers.py
```diff
@@ -29,6 +29,7 @@ def ansible_providers():
     remove_tmp_files()
 
 
+@pytest.mark.skip(reason="Needs pzagal review")
```
quarckster: This shouldn't be in the code

shalomnaim1: This the goal of this PR, skip broken tests.

shalomnaim1: I don't understand what changes you expect to see in this PR, can you be clearer?

quarckster: I understand your point, but it should be uncollected and add TODO with that reason.

shalomnaim1: @quarckster This code is not relevant, i redefine the reason, can you take a look?

PR #5469, opened 2017-09-27
[1LP][RFR]Quadicon related fix

anurag03:
Purpose or Intent
=================
3 test case related to quadicon were failing after conversion of setting page to widgetastic. In this PR I have added entities for datastore and fixed other failing test cases.

{{pytest: -v -k 'test_datastore_noquads or test_vm_noquads or test_template_noquads' }}


lkhomenk:
please add some Docstrings - it's confusing what whould "no_quad" mean. Like if we are testing quadrants or quadicons.  but in general everything is just fine

PR #5476, opened 2017-09-28
[1LP][RFR] Add ansible.repository.as_fill_value method

mshriver:
Add method that's used in a test in catalog_item creation.

Works fine locally

## PRT Results
Run 16506: Dropdown was disabled in UI for adding ansible repository. I saw this locally, and it cleared when I re-ran the test. There may be a UI bug here.  Continuing to fail on re-runs of the test.

{{pytest: -v --use-provider complete --long-running cfme/tests/ansible/test_embedded_ansible_basic.py::test_control_add_ansible_playbook_action_invalid_address}}

### DIFF
--- a/cfme/tests/ansible/test_embedded_ansible_basic.py
+++ b/cfme/tests/ansible/test_embedded_ansible_basic.py
@@ -89,7 +89,7 @@ def catalog_item(ansible_repository):
         fauxfactory.gen_alphanumeric(),
         fauxfactory.gen_alphanumeric(),
         provisioning={
-            "repository": ansible_repository.as_fill_value(),
+            "repository": ansible_repository,
             "playbook": "dump_all_variables.yml",
             "machine_credential": "CFME Default Credential",
             "create_new": True,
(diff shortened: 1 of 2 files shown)


cfme/tests/ansible/test_embedded_ansible_basic.py
```diff
@@ -89,7 +89,7 @@ def catalog_item(ansible_repository):
         fauxfactory.gen_alphanumeric(),
         fauxfactory.gen_alphanumeric(),
         provisioning={
-            "repository": ansible_repository.as_fill_value,
+            "repository": ansible_repository.as_fill_value(),
```
izapolsk: you don't need to call as_fill_value() if you pass, this to fill further. you can just pass object itself and widgetastic will call as_fill_value.

mshriver: Doesn't seem that the fill call is picking this up as an object and calling as_fill_value on it, it's getting passed as an object resulting in an error when a replace method is called (expecting string in fill)

izapolsk: Repository is not inherited from Fillable. it should start to work if you add it

mshriver: @izapolsk pointed out offline that I hadn't added Fillable inheritence.

PR #5478, opened 2017-09-28
[1LP][RFR] removed nesty provider endpoint warning

izapolsk:


cfme/common/provider.py
```diff
@@ -954,8 +954,8 @@ def __init__(self, **kwargs):
             setattr(self, key, val)
 
         if not hasattr(self, 'credentials'):
-            logger.warn("credentials weren't passed "
-                        "for endpoint {}".format(self.__class__.__name__))
+            logger.debug("credentials weren't passed to"
+                         " endpoint {}".format(self.__class__.__name__))
```
mshriver: **Optional**

Maybe required... Add some more context, currently this results in many of the exact same messages in the logs, with no indication of which provider object they're tied to.

izapolsk: I can't add context w/o analyzing mro or using inspect. So, I decided to just remove this warning.

PR #5479, opened 2017-09-28
[1LP][RFR] Fix node collection order

psav:
{{pytest: cfme/tests/containers/test_table_views.py}}

### DIFF
--- a/cfme/containers/node.py
+++ b/cfme/containers/node.py
@@ -55,7 +55,7 @@ class NodeCollection(BaseCollection):
         self.appliance = appliance
 
     def instantiate(self, name, provider):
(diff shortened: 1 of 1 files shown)


gshefer:
[get_random_instances](https://github.com/psav/cfme_tests/blob/5ec2b5f40fffae44b71fb261f9031dd0366b65fa/cfme/containers/node.py#L119) should be updated as well. maybe there are few more instantiations that should be updated. let me check and I'll come with results soon.

psav:
Merging this, @gshefer already commented that it's OK if it breaks get_random_instances and he will fix it up later

gshefer:
Fix: https://github.com/ManageIQ/integration_tests/pull/5494

PR #5484, opened 2017-09-29
[RFR] PoC for filtered collections

psav:
PoC for filtered collections
---------------------------------

So we got talking about how just instantiating a collection wasn't always enough and indeed there were some collection objects that already take some secondary arguments which they use for filtering. This would allow you to have a collection object for VirtualMachines, that only listed virtual machines for a particular provider.

The issue is the in order to get the BaseCollection, we had to make all of the constructors the same and since we wanted to get it via the appliance,  we had introduced this kind of methodology

```python
appliance.collections.virtualmachines.all
```

Which would return for you a list/iterator of all VirtualMachine objects that the appliance has access to. However, even though the Collection may look like this:

```python
class VirtualMachineCollection(BaseCollection):
    def __init__(self, provider=None):
        self.provider = provider

    def all(self):
        # pseudocode
        list = []
        for virt_machine in list_of_all_virt_machines:
            if virt_machine.provider == self.provider
                list.append(virt_machine)
        return
```

It would mean that everytime you used ```all```, or any other function that needed to be filtered, we'd somehow need to convert all the functions to take all the filter parameters. This seemed clunky, and counter to the good practice that people seemed to already have, which was to create "filtered" collection objects.

So we did some thinking and came up with this

```python
appliance.collections.virtualmachines.filter(provider=ProviderObject).all()
```

In here, the collection has a filter method which returns a new collection which would set the filtering on it. Note that now though, each constructor for the Collection would need to be different, ie, which args do they take. We didn't like this, so we changed it slightly.

```python
appliance.collections.virtualmachines.filter({'provider': ProviderObject}).all()
```

So now, we are only passing a single argument which is a dict called ```filter``` internally. This is nice, and leads us to now have identical constructors for all collection objects, but with the benefit of being able to pass in some filter dict which can be used inside the collection for filtering ```all``` and any other methods we require.

Now we have this though, a single homogenous way of creating collections with and without filtering. It also means we can link ```Entity``` objects to other collections. As an example a provider can now have a collection object on it, which details all of the virtualmachines which are associated with it. This could be created like so.

```python
class Provider(BaseEntity):
    def __init__(self, collection, **kwargs):
        self.collection = collection
        self.appliance = self.collection.appliance

    def vms(self):
        return self.appliance.collections.virtualmachines.filter({'provider': self})
```

Now we can access the virtualmachines on a provider like so

```python
for vm in provider.vms.all()
```

Swanky eh? But we decided to try to take it one further. Since a Provider could have multiple collections associated with it, we implemented a helper scheme

```python
class Provider(BaseEntity):

    _collections = {'vms': VirtualMachineCollection}
    _collections_obj = None
```

This will, thanks to ```BaseEntity```, create a collections attribute which points to a special ```ObjectCollections``` object, which automatically loads all collections in the _collections dict, and instantiates them with a ```parent``` filter. This looks like this ```{'parent': self}```. We don't think there is actually too many other objects needed for filtering but the collections can also be specified like this.

```python
class Provider(BaseEntity):

    _collections = {'vms': (VirtualMachineCollection, {'on': True})}
    _collections_obj = None
```

Which passes an additional filter. This allows you now to access ```provider.collections.vms.all()``` without writing any special ```vms``` property on the provider object. Nifty eh?

~~Once last point that isn't handled yet. When you filter, we don't yet chain the filters, it is a replacement. I expect chainable filtering to land in about 45mins.~~
OK it took a little longer, but you can now do

```python
    col = appliance.collections.virtualmachines.filter({'provider': provider})
    col.filter(({'on': True})
```

This would apply both filters to the new collection.

As an additional benefit, we have now, since we introduced the ```ENTITY``` attribute on the collections class, can do away with the normal ```instantiate``` method and delegate it to the base class. This means no one has to write one of these, unless there is something special. The method is basically a wrapper around having to pass the collection to the entity.

This rework also introduces a new ```init``` method. Why? well it means that we can delegate the handling of ```appliance``` and ```collection``` to the base class ```__init__``` meaning no one needs to handle them anymore, unless they are doing something fancy. Just remember to use ```init``` and not ```__init__``` when writing constructors for Collection and Entity objects. As a side not ```__init__``` calls ```init```.

### DIFF
--- a/cfme/ansible/playbooks.py
+++ b/cfme/ansible/playbooks.py
@@ -103,9 +103,6 @@ class PlaybooksCollection(BaseCollection):
     def init(self):
         self.parent = self.filters.get('parent', None) if self.filters else None
 
-    def instantiate(self, name, repository):
-        return Playbook(self, name, repository)
-
     def all(self):
         view = navigate_to(Server, "AnsiblePlaybooks")
         playbooks = []
--- a/cfme/ansible/playbooks.py
+++ b/cfme/ansible/playbooks.py
@@ -100,9 +100,8 @@ class PlaybooksView(PlaybookBaseView):
 class PlaybooksCollection(BaseCollection):
     """Collection object for the :py:class:`Playbook`."""
 
-    def __init__(self, appliance, filter=None):
-        self.appliance = appliance
-        self.parent = filter.get('parent', None) if filter else None
+    def init(self):
+        self.parent = self.filters.get('parent', None) if self.filters else None
 
     def instantiate(self, name, repository):
         return Playbook(self, name, repository)
@@ -119,9 +118,7 @@ class PlaybooksCollection(BaseCollection):
 class Playbook(BaseEntity):
     """A class representing one Embedded Ansible playbook in the UI."""
 
-    def __init__(self, collection, name, repository):
-        self.collection = collection
-        self.appliance = self.collection.appliance
+    def init(self, name, repository):
         self.name = name
         self.repository = repository
 
--- a/cfme/ansible/playbooks.py
+++ b/cfme/ansible/playbooks.py
@@ -102,7 +102,7 @@ class PlaybooksCollection(BaseCollection):
 
     def __init__(self, appliance, filter=None):
         self.appliance = appliance
-        self.parent = filter.get('parent_repository', None) if filter else None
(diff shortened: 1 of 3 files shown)


dajoRH:
*CFME QE Bot*

PR #5488, opened 2017-09-29
[1LP][RFR] This PR contains removal of "reboot" from install_vddk() method.

sbulage:


cfme/configure/configuration/analysis_profile.py
```diff
@@ -233,8 +233,9 @@ def create(self, cancel=False):
         else:
             view.add.click()
 
-        view = self.create_view(AnalysisProfileAllView)
         view.flush_widget_cache()
```
mfalesni: TODO: This may get removed as well since this view is no longer used

jkrocil:
(discussed offline)

PR #5490, opened 2017-09-29
[1LP][RFR]Correcting configuration for VMware Console

kedark3:
Purpose or Intent
=================

__Fixing__ Configuration because it is currently doing looking for name in CONSOLE_TYPES which is incorrect and causing Console Test Failures. It should be looking for Value in CONSOLE_TYPES. 


kedark3:
Failed in PRT due to https://bugzilla.redhat.com/1492246

PR #5492, opened 2017-10-02
[1LP][RFR] Collection conversion of Tenant & Project

vatsalparekh:
Purpose or Intent
=================

Refactored and converted Tenant to Collection + test cases updated
{{ pytest: -v -k tenant cfme/tests/infrastructure/test_infra_quota.py cfme/tests/configure/test_access_control.py }}

### DIFF
--- a/cfme/configure/access_control.py
+++ b/cfme/configure/access_control.py
@@ -1,5 +1,7 @@
+import attr
 from navmazing import NavigateToSibling, NavigateToAttribute
-from widgetastic_manageiq import UpDownSelect, PaginationPane, SummaryFormItem, Table
+from widgetastic_manageiq import (
+    UpDownSelect, PaginationPane, SummaryFormItem, Table, BaseListEntity)
 from widgetastic_patternfly import (
     BootstrapSelect, Button, Input, Tab, CheckableBootstrapTreeview,
     BootstrapSwitch, CandidateNotFound, Dropdown)
@@ -10,6 +12,7 @@ from cfme.base.credential import Credential
 from cfme.base.ui import ConfigurationView
 from cfme.exceptions import OptionNotAvailable, RBACOperationBlocked
 from cfme.utils.appliance import Navigatable
+from cfme.modeling.base import BaseCollection, BaseEntity
 from cfme.utils.appliance.implementations.ui import navigator, CFMENavigateStep, navigate_to
 from cfme.utils.log import logger
 from cfme.utils.pretty import Pretty
@@ -1192,8 +1195,12 @@ class TenantForm(ConfigurationView):
     """ Tenant Form """
     name = Input(name='name')
     description = Input(name='description')
-
+    add_button = Button('Add')
     cancel_button = Button('Cancel')
+
+
+class ListEntity(BaseListEntity):
+    pass
 
 
 class TenantQuotaForm(View):
@@ -1228,6 +1235,7 @@ class TenantQuotaView(ConfigurationView):
 class AllTenantView(ConfigurationView):
     """ All Tenants View """
     toolbar = View.nested(AccessControlToolbar)
+    table = Table('//*[@id="fieldset"]/table')
 
     @property
     def is_displayed(self):
@@ -1237,9 +1245,9 @@ class AllTenantView(ConfigurationView):
         )
 
 
-class AddTenantView(TenantForm):
+class AddTenantView(View):
     """ Add Tenant View """
-    add_button = Button('Add')
+    form = View.nested(TenantForm)
 
     @property
     def is_displayed(self):
@@ -1252,6 +1260,10 @@ class AddTenantView(TenantForm):
 class DetailsTenantView(ConfigurationView):
     """ Details Tenant View """
     toolbar = View.nested(AccessControlToolbar)
+    name = Text('Name')
+    description = Text('Description')
+    parent = Text('Parent')
+    table = Table('//*[@id="fieldset"]/table')
 
     @property
     def is_displayed(self):
@@ -1271,20 +1283,24 @@ class ParentDetailsTenantView(DetailsTenantView):
         )
 
 
-class EditTenantView(TenantForm):
+class EditTenantView(View):
     """ Edit Tenant View """
+    form = View.nested(TenantForm)
     save_button = Button('Save')
     reset_button = Button('Reset')
 
     @property
     def is_displayed(self):
         return (
-            self.accordions.accesscontrol.is_opened and
-            self.title.text == 'Editing Tenant "{}"'.format(self.context['object'].name)
+            self.form.accordions.accesscontrol.is_opened and
+            any([self.form.title.text == 'Editing Tenant "{}"'.format(self.context['object'].name),
+                self.form.title.text == 'Editing Project "{}"'.format(self.context['object'].name)])
         )
+        # used any() because `project` inherits `tenant` and edit page can have any one in name
 
 
-class Tenant(Updateable, Pretty, Navigatable):
+@attr.s
+class Tenant(Updateable, BaseEntity):
     """ Class representing CFME tenants in the UI.
     * Kudos to mfalesni *
 
@@ -1296,95 +1312,10 @@ class Tenant(Updateable, Pretty, Navigatable):
         description: Description of the tenant
         parent_tenant: Parent tenant, can be None, can be passed as string or object
     """
-    pretty_attrs = ["name", "description"]
-
-    @classmethod
-    def get_root_tenant(cls):
-        return cls(name="My Company", _default=True)
-
-    def __init__(self, name=None, description=None, parent_tenant=None, _default=False,
-                 appliance=None):
-        Navigatable.__init__(self, appliance=appliance)
-        self.name = name
-        self.description = description
-        self.parent_tenant = parent_tenant
-        self._default = _default
-
-    @property
-    def parent_tenant(self):
-        if self._default:
-            return None
-        if self._parent_tenant:
-            return self._parent_tenant
-        return self.get_root_tenant()
-
-    @parent_tenant.setter
-    def parent_tenant(self, tenant):
-        if tenant is not None and isinstance(tenant, Project):
-            # If we try to
-            raise ValueError("Project cannot be a parent object.")
-        if isinstance(tenant, basestring):
-            # If parent tenant is passed as string,
-            # we assume that tenant name was passed instead of object
-            tenant = Tenant(tenant)
-        self._parent_tenant = tenant
-
-    def __eq__(self, other):
-        if not isinstance(other, type(self)):
-            return False
-        else:
-            return self.name == other.name
-
-    @property
-    def exists(self):
-        try:
-            navigate_to(self, 'Details')
-            return True
-        except CandidateNotFound:
-            return False
-
-    @property
-    def tree_path(self):
-        if self._default:
-            return [self.name]
-        else:
-            return self.parent_tenant.tree_path + [self.name]
-
-    @property
-    def parent_path(self):
-        return self.tree_path[:-1]
-
-    def create(self, cancel=False):
-        """ Create role method
-
-        Args:
-            cancel: True - if you want to cancel role creation,
-                    by defaul(False), role will be created
-        """
-        if self._default:
-            raise ValueError("Cannot create the root tenant {}".format(self.name))
-
-        view = navigate_to(self, 'Add')
-        view.fill({'name': self.name,
-                   'description': self.description})
-        if cancel:
-            view.cancel_button.click()
-            tenant_flash_message = 'Add of new Tenant was cancelled by the user'
-            project_flash_message = 'Add of new Project was cancelled by the user'
-        else:
-            view.add_button.click()
-            tenant_flash_message = 'Tenant "{}" was saved'.format(self.name)
-            project_flash_message = 'Project "{}" was saved'.format(self.name)
-        view = self.create_view(ParentDetailsTenantView)
-        if isinstance(self, Tenant):
-            view.flash.assert_success_message(tenant_flash_message)
-        elif isinstance(self, Project):
-            view.flash.assert_success_message(project_flash_message)
-        else:
-            raise TypeError(
-                'No Tenant or Project class passed to create method{}'.format(
-                    type(self).__name__))
-        assert view.is_displayed
+    name = attr.ib()
+    description = attr.ib(default="")
+    parent_tenant = attr.ib(default=None)
+    _default = attr.ib(default=False)
 
     def update(self, updates):
         """ Update tenant/project method
@@ -1395,8 +1326,8 @@ class Tenant(Updateable, Pretty, Navigatable):
         Note: In case updates is the same as original tenant/project data, update will be canceled,
             as 'Save' button will not be active
         """
-        view = navigate_to(self, 'Edit')
-        changed = view.fill(updates)
+        view = navigate_to(self, 'Edit', wait_for_view=True)
+        changed = view.form.fill(updates)
         if changed:
             view.save_button.click()
             flash_message = 'Project "{}" was saved'.format(updates.get('name', self.name))
@@ -1406,7 +1337,6 @@ class Tenant(Updateable, Pretty, Navigatable):
                 updates.get('name', self.name))
         view = self.create_view(DetailsTenantView, override=updates)
         view.flash.assert_message(flash_message)
-        assert view.is_displayed
 
     def delete(self, cancel=True):
         """ Delete existing role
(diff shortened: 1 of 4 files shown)


cfme/configure/access_control.py
```diff
@@ -1,5 +1,6 @@
 from navmazing import NavigateToSibling, NavigateToAttribute
-from widgetastic_manageiq import UpDownSelect, PaginationPane, SummaryFormItem, Table
+from widgetastic_manageiq import UpDownSelect, PaginationPane, SummaryFormItem, Table, \
+    BaseListEntity
```
quarckster: **REQUIRED**
Please follow the style guide

cfme/configure/access_control.py
```diff
@@ -1213,6 +1218,7 @@ def is_displayed(self):
 class AllTenantView(ConfigurationView):
     """ All Tenants View """
     toolbar = View.nested(AccessControlToolbar)
+    table = ListEntity
```
quarckster: **REQURIED**
It shouldn't be named as table, but `list_entity`.

cfme/configure/access_control.py
vatsalparekh: Fixed

cfme/configure/access_control.py
vatsalparekh: Fixed

cfme/configure/access_control.py
```diff
@@ -1237,6 +1243,10 @@ def is_displayed(self):
 class DetailsTenantView(ConfigurationView):
     """ Details Tenant View """
     toolbar = View.nested(AccessControlToolbar)
+    name = Text('Name')
+    description = Text('Description')
+    parent = Text('Parent')
+    list_entity = ListEntity
```
quarckster: **REQURIED**
Where is `list_entity` used? Also you should initilaize it.

vatsalparekh: Initialized

cfme/configure/access_control.py
```diff
@@ -1177,10 +1178,14 @@ class TenantForm(ConfigurationView):
     """ Tenant Form """
     name = Input(name='name')
     description = Input(name='description')
-
+    add_button = Button('Add')
     cancel_button = Button('Cancel')
 
 
+class ListEntity(BaseListEntity):
```
psav: **REQUIRED:** Why are we subclassing this if we are not adding anything to it? line 1221 can just be 
```list_entity = BaseListEntity```

File: cfme/configure/access_control.py [hunk 5]
```diff
@@ -1444,32 +1451,13 @@ def step(self):
...
     VIEW = DetailsTenantView
-    prerequisite = NavigateToSibling('All')
+    prerequisite = NavigateToAttribute('collection', 'All')
 
     def step(self):
         self.prerequisite_view.accordions.accesscontrol.tree.click_path(
             self.obj.appliance.server_region_string(), 'Tenants', *self.obj.tree_path)
 
 
-@navigator.register(Tenant, 'Add')
```
psav: **REQUIRED:** Where did this go? Is it already on ```TenantCollection```?

cfme/tests/configure/test_access_control.py
```diff
@@ -790,8 +790,19 @@ def test_user_change_password(appliance, request):
 
 
 # Tenant/Project test cases
+
+@pytest.fixture(scope="module")
```
psav: **REQUIRED:** These are not necessary as it's just a one liner in the test that will soon go away....with the updated Collections PR

cfme/configure/access_control.py
vatsalparekh: I used it that way itself, @quarckster required to use it like this
@psav what's the final change?

cfme/configure/access_control.py
vatsalparekh: To add a new Tenant/Project, we first need to navigate to the parent, under which we want to add it. 
And as per `Collection` design `create` method should be on `TenantCollection`. 

So writing `prerequisite = NavigateToAttribute(Tenant, 'Detail')` in `Add` navigator registered with collection would need the `Tenant` object which we want to navigate to. And passing args to navigation didn't seem good design (and would probably yell if done?) 

So I've removed the `Add` and doing the some steps of navigation inside `create` method on collection. Its just like navigating first to `Details` and then going to `Add` page.               
`create` func [code](https://github.com/vatsalparekh/integration_tests/blob/1b394e02e25c23176f6806d43948df50e22c1b9f/cfme/configure/access_control.py#L1305)

cfme/tests/configure/test_access_control.py
vatsalparekh: Updated.

dajoRH:
Lint report for commit 664299e7dff1e235fd04df8a7b2295912982c16a:

- :bangbang: Line 953:101: **E501** *line too long (101 > 100 characters)*
- :bangbang: Line 958:101: **E501** *line too long (102 > 100 characters)*

**No commit flaws detected**.

*CFME QE Bot*

cfme/configure/access_control.py
mshriver: The prerequisite for adding a tenant should be the 'All' view registered against TenantCollection, not the details view.

vatsalparekh: @mshriver but then that `Add` view need to be registered with `Tenant` to have `parent_tenant` accessible, and we're calling it from Collection so kind of contradictory in design
@psav help please in deciding the final change

cfme/configure/access_control.py
izapolsk: @vatsalparekh, you shouldn't use ListEntity directly. Use BaseEntitiesView instead and switch view mode to List View.
So, get_entity() and other methods will return you ListEntity()

quarckster: I proposed to do that as it done [here](https://github.com/ManageIQ/integration_tests/blob/master/cfme/common/host_views.py#L73), I was thinking that there are other kinds of entities like grid or tile. Also should we directly use Base entities?

izapolsk: @quarckster, we can but I wouldn't recommend to do that.  I'll take a look what we can do there and come up with something in a bit.

psav:
Please squash

cfme/configure/access_control.py
izapolsk: it is interesting idea to reuse ListEntity here. However, unfortunately, there are several reasons why it won't suit here.
I'd recommend to either use regular Table with checkboxes here or create some TenantEntity class working on top of Table's Row and providing appropriate methods.

cfme/configure/access_control.py
mshriver: So, within `TenantCollection.create()` can we not do:

```python
# pseudocode
tenant_entity = self.instantiate(**kwargs)
view=navigate_to(tenant_entity, 'add')
view.fill
return tenant_entity

File: cfme/configure/access_control.py (hunk 4 above)
Comment (vatsalparekh) [reply cfme/configure/access_control.py]: simply used regular `Table`

File: cfme/configure/access_control.py [hunk 7]
```diff
@@ -1256,20 +1266,73 @@ def is_displayed(self):
...
+
+    def __init__(self, appliance):
+        self.appliance = appliance
+
+    def instantiate(self, name, description=None, parent=None, default=False):
+        return Tenant(collection=self, name=name, description=description, parent_tenant=parent,
+                      _default=default)
+
+    def get_root_tenant(self):
+        return self.instantiate('My Company', default=True)
```
mfalesni: **REQUIRED**: Check if you can get the "My Company" value directly from the appliance and use it. That will save headaches later. `My Company` can be changed, that is why I am requesting that.

cfme/configure/access_control.py
```diff
@@ -1256,20 +1266,73 @@ def is_displayed(self):
...
+
+    def create(self, name, description, parent):
+
+        if isinstance(parent, Project):
+            raise ValueError("Projects cannot have children.")
```
mfalesni: **REQUIRED**: Please review this - I don't think the test framework should restrain you too much from doing things to the UI. Let it fail on the dropdown item lookup, so you can actually test this case later.

cfme/configure/access_control.py
```diff
@@ -1256,20 +1266,73 @@ def is_displayed(self):
...
+
+        tenant = self.instantiate(name, description, parent)
```
mfalesni: **REQUIRED**: `create` should always end with `return self.instantiate(...)`.

cfme/configure/access_control.py
```diff
@@ -1256,20 +1266,73 @@ def is_displayed(self):
...
+        else:
+            view.form.cancel_button.click()
+
+        view = self.create_view(ParentDetailsTenantView)
+        view.flash.assert_success_message('Tenant "{}" was saved'.format(name))
+
+        return tenant
+
+    def delete(self, *tenants):
+        tenants = list(tenants)
```
mfalesni: **REQUIRED**: Because you don't mutate `tenants`, this line is not needed.

cfme/configure/access_control.py
```diff
@@ -1256,20 +1266,73 @@ def is_displayed(self):
...
+
+        view = navigate_to(self, 'All')
+
+        for tenant in tenants:
+            try:
+                view.table.row(name=tenant.name).check()
+            except Exception:
+                break
```
mfalesni: **REQUIRED**: This should *at least* log the exception.

cfme/configure/access_control.py
```diff
@@ -1391,7 +1374,6 @@ def update(self, updates):
                 updates.get('name', self.name))
         view = self.create_view(DetailsTenantView, override=updates)
         view.flash.assert_message(flash_message)
-        assert view.is_displayed
```
mfalesni: **QUESTION**: Why no longer asserting htat you are on correct place? Problematic UI? Maybe `is_displayed` needs checking?

cfme/configure/access_control.py
```diff
@@ -1430,8 +1412,33 @@ def set_quota(self, **kwargs):
...
+    @property
+    def tree_path(self):
+        if self._default:
+            return [self.name]
+        else:
+            return self.parent_tenant.tree_path + [self.name]
+
+    @property
+    def parent_path(self):
+        return self.tree_path[:-1]
```
mfalesni: **OPTIONAL**: Wouldn't it be nicer to use something like `self.parent.tree_path`?

cfme/configure/access_control.py
```diff
@@ -1495,6 +1483,40 @@ def step(self):
...
+    def instantiate(self, name, description=None, parent=None, default=False):
+        return Project(collection=self, name=name, description=description, parent_tenant=parent)
+
+    def get_root_tenant(self):
+        # returning Tenant directly because 'My Company' needs to be treated like Tenant object,
+        # to be able to make child tenant/project under it
+        return Tenant(collection=self, name='My Company', _default=True)
+
+    def create(self, name, description, parent):
+        project = self.instantiate(name, description, parent)
```
mfalesni: **REQUIRED**: As I said before.

cfme/configure/access_control.py
```diff
@@ -1495,6 +1483,40 @@ def step(self):
...
+class ProjectCollection(TenantCollection):
+    """Collection class for Projects under Tenants"""
+
+    def instantiate(self, name, description=None, parent=None, default=False):
+        return Project(collection=self, name=name, description=description, parent_tenant=parent)
+
+    def get_root_tenant(self):
+        # returning Tenant directly because 'My Company' needs to be treated like Tenant object,
+        # to be able to make child tenant/project under it
+        return Tenant(collection=self, name='My Company', _default=True)
```
mfalesni: **REQUIRED**: As mentioned before.

cfme/configure/access_control.py
```diff
@@ -1495,6 +1483,40 @@ def step(self):
...
+        changed = view.form.fill({'name': name,
+                   'description': description})
+        if changed:
+            view.form.add_button.click()
+        else:
+            view.form.cancel_button.click()
+
+        view = self.create_view(ParentDetailsTenantView)
+        view.flash.assert_success_message('Project "{}" was saved'.format(name))
+
```
mfalesni: **QUESTION**: You should probably also assert the view displayed here. Remember you can pass the parent tenant object to the create_view as well, so `is_displayed` works as expected.

cfme/tests/configure/test_access_control.py
```diff
@@ -803,16 +804,17 @@ def test_superadmin_tenant_crud(request):
         * Update name of tenat
         * Delete tenant
     """
-    tenant = Tenant(
+    tc = TenantCollection(appliance=appliance)
+    tenant = tc.create(
         name='tenant1' + fauxfactory.gen_alphanumeric(),
-        description='tenant1 description')
+        description='tenant1 description',
+        parent=tc.get_root_tenant())
```
mfalesni: **QUESTION**: This doesn't look right to me structurally. I would expect something like this:

```python
tc = TenantCollection(...)
rt = tc.root_tenant()
t = rt.tenants.create(...)   # without parent
# or in case of projects under tenant
p = rt.projects.create(...)
```

Your opinion here @psav @mshriver ?

cfme/tests/configure/test_access_control.py
```diff
@@ -836,22 +838,24 @@ def test_superadmin_tenant_project_crud(request):
         * Delete project
         * Delete tenant
     """
-    tenant = Tenant(
+    tc = TenantCollection(appliance=appliance)
+    pc = ProjectCollection(appliance=appliance)
```
mfalesni: Same here, this seems to be going away fro ma nice tree structure.

cfme/configure/access_control.py
vatsalparekh: Makes sense, removed.

cfme/configure/access_control.py
vatsalparekh: `Add` navigation on `Tenant` didn't seem a good design, so here I first initialized the tenant to navigate to its `Details` and then filled the details for create

cfme/configure/access_control.py
vatsalparekh: removed.

cfme/configure/access_control.py
vatsalparekh: Added.

cfme/configure/access_control.py
vatsalparekh: I removed it thinking that its already asserting the flash message, so `view.is_displayed` won't be necessary. Is it still required @mfalesni ?

cfme/configure/access_control.py
vatsalparekh: Oh yes, corrected.

cfme/configure/access_control.py
vatsalparekh: As I commented before.

cfme/configure/access_control.py
vatsalparekh: Again, is it necessary to explicitly check `is_displayed` after asserting flash message? @mfalesni

cfme/tests/configure/test_access_control.py
vatsalparekh: Ah yes, this looks better way to do, changing

cfme/tests/configure/test_access_control.py
vatsalparekh: Changed

cfme/configure/access_control.py
vatsalparekh: Changed it to use `appliance.rest_api.collections.tenants[0].name`

psav:
This needs to be converted to CEMv3 (Collections/Entity Model)

dajoRH:
*CFME QE Bot*

dajoRH:
Lint report for commit e7ec6f7cca433ba3a354e381901a7922b942b736:

`cfme/configure/access_control.py`:
- :bangbang: Line 1512:101: **E501** *line too long (135 > 100 characters)*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
Lint report for commit 34929ab8de7878eb496029c6b308f5c305181ef7:

`cfme/configure/access_control.py`:
- :bangbang: Line 1512:101: **E501** *line too long (135 > 100 characters)*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
Lint report for commit d676c3700434751afc043284d6afe3807f7d3fee:

- :red_circle: Line 6:1: **F401** *'cfme.configure.access_control.ProjectCollection' imported but unused*
- :red_circle: Line 6:1: **F401** *'cfme.configure.access_control.TenantCollection' imported but unused*

`cfme/tests/infrastructure/test_infra_quota.py`:
- :red_circle: Line 7:1: **F401** *'cfme.configure.access_control.TenantCollection' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

cfme/configure/access_control.py
```diff
@@ -1261,20 +1274,24 @@ def is_displayed(self):
...
+            self.form.accordions.accesscontrol.is_opened and
+            any([self.form.title.text == 'Editing Tenant "{}"'.format(self.context['object'].name),
+                self.form.title.text == 'Editing Project "{}"'.format(self.context['object'].name)])
         )
+        # used any() because `project` inherits `tenant` and edit page can have any one in name
 
 
-class Tenant(Updateable, Pretty, Navigatable):
+@attr.s
+class Tenant(BaseEntity):
```
psav: **REQUIRED:** Is the Tenant no longer Updateable?

cfme/configure/access_control.py
```diff
@@ -1435,8 +1367,83 @@ def set_quota(self, **kwargs):
...
+
+    @property
+    def exists(self):
+        try:
+            navigate_to(self, 'Details')
+            return True
+        except CandidateNotFound:
+            return False
+
+    @cached_property
```
psav: **REQUIRED:** I'm not sure this makes sense here. One would assume that a ```tenants``` property on a tenant object would return a list of sub tenants. This doens't do that it actually returns ```self.parent```. If they were intended to be filtered collections, then they should be implemented properly. The same goes for projects below.

cfme/configure/access_control.py
psav: **REQUIRED:** This needs to be responded to

cfme/tests/configure/test_access_control.py
```diff
@@ -799,16 +800,18 @@ def test_superadmin_tenant_crud(request):
         * Update name of tenat
         * Delete tenant
     """
-    tenant = Tenant(
+    roottenant = appliance.collections.tenants.get_root_tenant()
+    tenant = roottenant.tenants.create(
```
psav: **REQUIRED:** There is actually no reason why we need to do this from ```roottenant.tenants``` here, we could have done it from ```appliance.tenants```. Had tenants been special, it may have been so.

cfme/tests/configure/test_access_control.py
```diff
@@ -882,12 +885,11 @@ def _delete_tenants():
                 tenant.delete()
 
     for i in range(1, number_of_childrens + 1):
-        new_tenant = Tenant(
+        new_tenant = tenant.tenants.create(
```
psav: **REQUIRED:** Again here and in other places

cfme/configure/access_control.py
vatsalparekh: It is Updateable, added.

cfme/configure/access_control.py
vatsalparekh: Corrected

cfme/tests/configure/test_access_control.py
vatsalparekh: Changed

cfme/tests/configure/test_access_control.py
vatsalparekh: Changed at all places

cfme/configure/access_control.py
vatsalparekh: @mshriver 
1) The functionality to add a tenant can be accessed from the `Detail` view of a `tenant`, so it can't be registered with `All` as prerequisite
2) with the pseudocode given, issue is in prerequisite of `add` view; how would we navigate to `details` when we don't have the instantiated `tenant_entity` accessible inside the view?

dajoRH:
Lint report for commit e37b18c089efe0d1759a2fb58b633bb54eacb482:

`cfme/configure/access_control.py`:
- :red_circle: Line 2:1: **F401** *'cached_property.cached_property' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

cfme/configure/access_control.py
psav: I understand now, thanks

cfme/configure/access_control.py
```diff
@@ -1386,7 +1317,8 @@ def update(self, updates):
             as 'Save' button will not be active
         """
         view = navigate_to(self, 'Edit')
-        changed = view.fill(updates)
+        wait_for(lambda: view.is_displayed, num_sec=5)
```
psav: @mfalesni this shouldn't be necessary should it?

vatsalparekh: It was failing with the error 
` 'MoveTargetOutOfBoundsException: Message: Despite all the workarounds, scrolling to Input(locator=u'.//*[(self::input or self::textarea) and @name="description"]') was unsuccessful'
`
it was filling the form quick, so added a wait_for

PR #5504, opened 2017-10-03
[1LP][RFR] Small fixes for OSP tests

okolisny:
Couple of changes

### DIFF
--- a/cfme/tests/openstack/infrastructure/test_relationships.py
+++ b/cfme/tests/openstack/infrastructure/test_relationships.py
@@ -6,7 +6,6 @@ results of the relationships
 import pytest
 
 import cfme.fixtures.pytest_selenium as sel
-from selenium.common.exceptions import NoSuchElementException
 from cfme.infrastructure.provider.openstack_infra import OpenstackInfraProvider
 from cfme.web_ui import InfoBlock, Table
 from cfme.utils import testgen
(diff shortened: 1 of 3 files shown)


cfme/tests/openstack/infrastructure/test_relationships.py
```diff
@@ -24,7 +23,7 @@ def test_assigned_roles(provider):
     navigate_to(provider, 'Details')
     try:
         res = provider.get_detail('Relationships', 'Deployment Roles')
-    except NoSuchElementException:
+    except NameError:
```
quarckster: Why is it changed to this exception? [This line](https://github.com/ManageIQ/integration_tests/blob/master/widgetastic_manageiq/__init__.py#L491) raises `NoSuchElementException` if entry not found.

okolisny: I can see it behaves differently
https://rhos-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/Director/view/CFME_RHOS_Integration/job/rhos9_cfme58_integration/26/TEST_REPORT/

quarckster: Oh, it's `SummaryTable`, not `SummaryForm`.

PR #5506, opened 2017-10-03
[RFR] update image streams along with new cfme build

izapolsk:
upload script makes opeshift to pull latest image stream changes now

### DIFF
--- a/scripts/template_upload_openshift.py
+++ b/scripts/template_upload_openshift.py
@@ -89,15 +89,18 @@ def upload_template(hostname, username, password, provider, url, name, provider_
         if not check_kwargs(**kwargs):
             return False
 
+        logger.info("checking whether this template is already present in provider env")
         if name not in list_templates(hostname, username, password, upload_folder):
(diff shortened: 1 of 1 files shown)


izapolsk:
script run results:
http://pastebin.test.redhat.com/521766

scripts/template_upload_openshift.py
```diff
@@ -77,8 +77,8 @@ def list_templates(hostname, username, password, upload_folder):
             return []
 
 
-def upload_template(hostname, username, password,
-                    provider, url, name, provider_data, stream, upload_folder):
+def upload_template(hostname, username, password, provider, url, name, provider_data,
+                    stream, upload_folder, oc_username, oc_password):
```
mshriver: Can we pull these from `cfme.utils.conf` + `provider.data.<some key for creds>` instead of passing them?

mshriver: I see below where you are before calling this, nvm.

PR #5091, opened 2017-08-03
[1LP][RFR] Server Tab: Server roles, updated tests(Part2)

rlbabyuk:
Purpose or Intent
=================
Convert Server tab for server settings to widgetastic.(Part2). PR contains only server roles changes
Also updated all places where previous classes were used

{{pytest: -v -k 'test_metrics_collection'  --long-running}}

PRT:
Tests failures are not related to PR changes

PR #5507, opened 2017-10-03
[1LP][RFR] Adding Websocket fixture to enable websocket server role

kedark3:
Purpose or Intent
=================

__Extending__ test_ssui_myservice Adding Websocket fixture to enable websocket server role, for vm conole tests that are to be added soon. Adding the role to pytest.mark.meta, next to +automate would not work because the fixture adding provider will disable the websocket role after pytest has already enabled the role, hence need separate fixture to re-enable websocket role.

{{pytest: -v --long-running cfme/tests/ssui/test_ssui_myservice.py  }}

### DIFF
--- a/cfme/tests/ssui/test_ssui_myservice.py
+++ b/cfme/tests/ssui/test_ssui_myservice.py
@@ -1,13 +1,14 @@
 # -*- coding: utf-8 -*-
 import pytest
 
+from cfme.configure.configuration import server_roles_enabled
 from cfme.infrastructure.provider import InfraProvider
 from cfme.services.myservice import MyService
 from cfme import test_requirements
 from cfme.utils import testgen
-from cfme.utils.version import current_version
 from cfme.utils.appliance import get_or_create_current_appliance, ViaSSUI
-
+from cfme.utils.log import logger
+from cfme.utils.version import current_version
 
 pytestmark = [
     pytest.mark.meta(server_roles="+automate"),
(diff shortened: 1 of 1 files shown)


cfme/tests/ssui/test_ssui_myservice.py
```diff
@@ -23,6 +24,20 @@
 ], scope="module")
 
 
+@pytest.fixture(scope="session")
```
quarckster: This fixture enables websocket role for whole session, maybe it's better to use "module" scope?

rlbabyuk:
@kedark3 please be aware of changes in set up servers roles usage https://github.com/ManageIQ/integration_tests/pull/5091

quarckster:
@rlbabyuk by default server roles are changed via db.

cfme/tests/ssui/test_ssui_myservice.py
```diff
@@ -23,6 +24,22 @@
...
+
+    Currently the fixture cfme/fixtures/base.py:27,
+    disables the websocket role to avoid intrusive popups.
+    """
+    logger.info('Enabling the websocket role to allow console connections')
+    server_information = server_settings.ServerInformation(appliance)
+    server_information.enable_server_roles('websocket')
+    yield
+    logger.info('Disabling the websocket role to avoid intrusive popups')
+    server_information.disable_server_roles('websocket')
```
quarckster: **REQUIRED**
It can be done easier:
```python
from cfme.configure.configuration import server_roles_enabled
...
def configure_websocket(appliance):
    with server_roles_enabled('websocket'):
        yield
```

kedark3: @quarckster Done :)

PR #5511, opened 2017-10-04
[1LP][RFR] Fix to test_appliance_console smoke test

lcouzens:
Smoke test was randomly failing, changed the collection method of command output to fix the issue.
{{pytest: -v test_appliance_console.py::test_black_console}}


quarckster:
@lcouzens what about PRT?

lcouzens:
This fix is just for the test_black_console smoke test, the failures are for other tests outside this pr's scope.

PR #5516, opened 2017-10-04
[1LP][RFR]REST dialog change

sshveta:
Service provisioning generates service with names that contains timestamp of provisioning .
It is difficult to assert for service names containing these random timestamps .
So as a work around the dialog in use should have element name = "service_name" 
so that service name can be overridden by the value provided while ordering the service .

Hence some changes in dialog data .
@mkoura : Please check once

### DIFF
--- a/cfme/rest/gen_data.py
+++ b/cfme/rest/gen_data.py
@@ -84,15 +84,18 @@ def dialog_ui(appliance):
     """Creates service dialog using UI."""
     service_dialogs = appliance.get(DialogCollection)
     uid = fauxfactory.gen_alphanumeric()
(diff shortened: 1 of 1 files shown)


cfme/rest/gen_data.py
```diff
@@ -114,7 +117,7 @@ def dialog_rest(request, rest_api):
     uid = fauxfactory.gen_alphanumeric()
     data = {
         "description": "my dialog {}".format(uid),
-        "label": "dialog_{}".format(uid),
+        "label": uid,
         "buttons": "submit,cancel",
```
mfalesni: devs' string typing ftw :rofl:

PR #5524, opened 2017-10-05
[1LP][RFR] Various Fixes, increased some timeouts, Added missing navigation

hhovsepy:
{{pytest: cfme/tests/middleware/test_middleware_provider.py::test_password_mismatch_validation -v --use-provider hawkular}}

### DIFF
--- a/cfme/tests/middleware/deployment_methods.py
+++ b/cfme/tests/middleware/deployment_methods.py
@@ -18,8 +18,8 @@ RESOURCE_WAR_CONTENT_NEW = 'Original JSP 2'
 
 DEPLOYMENT_URL = 'http://{}:8180/{}'
 WAR_EXT = '.war'
-DELAY = 10
-NUM_SECS = 120
+DELAY = 30
+NUM_SECS = 600
 
 
 def deploy(provider, server, archive_name, runtime_name=None, enabled=True,
--- a/cfme/tests/middleware/jdr_report_methods.py
+++ b/cfme/tests/middleware/jdr_report_methods.py
@@ -1,29 +1,32 @@
 from cfme.utils.wait import wait_for
 
+DELAY = 30
+NUM_SECS = 300
+
 
 def verify_report_queued(jdr_rc, date_after):
     wait_for(lambda: jdr_rc.report_queued(date_after=date_after),
-             delay=3, num_sec=120,
+             delay=DELAY, num_sec=NUM_SECS,
              message='JDR Report must be found Queued after {}'
              .format(date_after))
 
 
 def verify_report_ready(jdr_rc, date_after):
     wait_for(lambda: jdr_rc.report_ready(date_after=date_after),
-             delay=3, num_sec=120,
+             delay=DELAY, num_sec=NUM_SECS,
              message='JDR Report must be found Ready after {}'
              .format(date_after))
 
 
 def verify_report_running(jdr_rc, date_after):
     wait_for(lambda: jdr_rc.report_running(date_after=date_after),
-             delay=3, num_sec=120,
+             delay=DELAY, num_sec=NUM_SECS,
              message='JDR Report must be found Running after {}'
              .format(date_after))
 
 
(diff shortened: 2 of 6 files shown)


cfme/tests/middleware/test_middleware_provider.py
```diff
@@ -54,7 +54,7 @@ def test_password_mismatch_validation(provider, soft_assert):
         verify_secret=fauxfactory.gen_alphanumeric(6)
     )
     prov.endpoints = endpoints
-    with error.expected('Credential validation was successful'):
+    with error.expected('Credential validation was not successful: Invalid credentials'):
```
psav: **REQUIRED:** Does this need version picking?

cfme/tests/middleware/deployment_methods.py
```diff
@@ -18,8 +18,8 @@
 
 DEPLOYMENT_URL = 'http://{}:8180/{}'
 WAR_EXT = '.war'
-DELAY = 10
-NUM_SECS = 120
+DELAY = 30
```
mshriver: **Required**
Keep the delay down to 10 seconds unless there is a strong reason to not query more often in `wait_for()`.

cfme/tests/middleware/deployment_methods.py
```diff
@@ -18,8 +18,8 @@
...
+NUM_SECS = 600
```
mshriver: **Question**
No issue increasing overall timeout, but would like to hear some justification for the magnitude of the increase.  

Is there a potential bug or RFE here, to improve MIQ/CFME performance?  What used to work in two minutes now may take 10?

cfme/tests/middleware/jdr_report_methods.py
```diff
@@ -1,29 +1,32 @@
 from cfme.utils.wait import wait_for
 
+DELAY = 30
```
mshriver: Same as comments above, would like to hear justification for the increase in delay and timeout.

cfme/tests/middleware/test_middleware_provider.py
```diff
@@ -54,7 +54,10 @@ def test_password_mismatch_validation(provider, soft_assert):
         verify_secret=fauxfactory.gen_alphanumeric(6)
     )
     prov.endpoints = endpoints
-    with error.expected('Credential validation was successful'):
+    message = 'Credential validation was successful'
+    if current_version() >= '5.9':
```
mshriver: **Optional**
VERY optional, more a FYI

This isn't a new import, but in the future, it is a better approach to take appliance as a fixture for your test, and use its version attribute instead of the current_version() method.

mshriver:
Please address the 5.7.4 failure in PRT run 17497, a NoSuchElementException

mshriver:
I restarted PRT because the 59z result had a failed appliance.

cfme/tests/middleware/deployment_methods.py
hhovsepy: I set it to query less often (30 secs) and increased the timeout to 10 minutes. This is done because when deploying a lot of archives on the server, it starts inventorying the deployments slower, and tests were failing.

cfme/tests/middleware/deployment_methods.py
hhovsepy: the answer is in above comment ^^

cfme/tests/middleware/deployment_methods.py
hhovsepy: I think the performance issue is not in CFME/MM side, but on EAP side itself

cfme/tests/middleware/test_middleware_provider.py
hhovsepy: I see, I did not find any other better examples implemented, so made in the old way

hhovsepy:
Upstream failure is known blocker issue.

hhovsepy:
Disabled for 5.7, as we do not support that version.

PR #5531, opened 2017-10-06
[1LP][RFR]Remove old set/get_server_roles functions

rlbabyuk:
Purpose or Intent
=================
Remove old old set/get_server_roles functions + one fixtute update
{{pytest: -v --long-running cfme/tests/ssui/test_ssui_myservice.py }}


cfme/configure/configuration/__init__.py
```diff
@@ -1365,33 +803,6 @@ def get_workers_list(do_not_navigate=False, refresh=True):
     return workers
 
 
-def setup_authmode_database():
```
psav: **REQUIRED** This is still used in cfme/fixtures/setup_auth_mode.py

PR #5534, opened 2017-10-06
[1LP][RFR] Update request, get description from cells

mshriver:
This is kinda hackish, but @mfalesni identified a failure where after editing a request navigation failed.

I believe the core failure is not in updating the description as the request is edited to change some values (number of vms), but because the description was never set because only `cells` was passed to the Request constructor.

RHCFQE-4868

## PRT Results

* 57z: webdriver exceptions during smtp setup, not related to these change, test_operations passed
* 58z: success
* upstream: Requests table isn't recognized/found in xpath, larger bug at play.

{{pytest: cfme/tests/infrastructure/test_provisioning.py cfme/tests/services/test_operations.py --use-provider vsphere6-nested --use-provider ec2west --long-running -v }}

### DIFF
--- a/cfme/services/requests.py
+++ b/cfme/services/requests.py
@@ -42,7 +42,7 @@ class Request(BaseEntity):
         """
         self.collection = collection
         self.appliance = self.collection.appliance
-        self.description = description
+        self.description = description or cells.get('Description')
         self.partial_check = partial_check
         self.cells = cells or {'Description': self.description}
         self.row = None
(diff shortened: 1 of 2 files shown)


dajoRH:
*CFME QE Bot*

izapolsk:
dup to #5525

mshriver:
@izapolsk changes to test_operations were a dup of #5525, but I've removed those. I don't think the change to `Request.__init__` is a dup. Thoughts?

izapolsk:
sure, agreed. I also made a bunch of changes to Request and thought those are similar to my ones.

PR #5537, opened 2017-10-09
[1LP][RFR] fixes for tests which don't correctly restore appliance when fail

izapolsk:
There are tests which kill podified applience by stopping evmserverd, etc. This happens because those don't add finalizers.

### DIFF
--- a/cfme/tests/cli/test_evmserverd.py
+++ b/cfme/tests/cli/test_evmserverd.py
@@ -19,7 +19,7 @@ pytestmark = [pytest.mark.usefixtures("start_evmserverd_after_module")]
 
 
 @pytest.mark.tier(1)
-def test_evmserverd_stop(appliance):
+def test_evmserverd_stop(appliance, request):
     """Tests whether stopping the evmserverd really stops the CFME server processes.
 
     Steps:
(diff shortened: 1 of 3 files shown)


cfme/tests/cli/test_evmserverd.py
```diff
@@ -38,6 +38,7 @@ def test_evmserverd_stop(appliance):
     })
 
     server_names = {server[server_name_key] for server in appliance.ssh_client.status["servers"]}
+    request.addfinalizer(lambda: appliance.ssh_client.run_command("systemctl start evmserverd"))
```
jkrocil: https://github.com/ManageIQ/integration_tests/blob/master/cfme/utils/appliance/__init__.py#L1325
`<jkrocil> we shouldnt really do it through ssh_client directly`

Same with the other occurances.

cfme/utils/ssh.py
```diff
@@ -527,10 +527,7 @@ def status(self):
                 'Please use .* instead',
                 'key :terminate is duplicated and overwritten',
             ]))
-        if version.current_version() < "5.5":
-            data = self.run_command("systemctl status evmserverd")
-        else:
-            data = self.run_rake_command("evm:status")
+        data = self.run_rake_command("evm:status")
```
jkrocil: I believe this whole method should exist under `IPAppliance` object, not here, but that's (understandably) out of scope for this PR.

cfme/tests/cli/test_evmserverd.py
```diff
@@ -8,18 +8,18 @@
 
 @pytest.yield_fixture(scope="module")
 def start_evmserverd_after_module(appliance):
-    appliance.ssh_client.run_command("systemctl start evmserverd")
+    appliance.start_evm_service()
```
mshriver: :+1:

PR #5544, opened 2017-10-09
[1LP][RFR] pod upload scripts have to be updated to support 5.9

izapolsk:


scripts/template_upload_all.py
```diff
@@ -294,7 +294,10 @@ def browse_directory(dir_url):
         if key == 'template_upload_openshift':
             # this is necessary because headers don't contain last-modified date for folders
             # todo: remove this along with refactoring script
-            url = name_dict[key] + 'cfme-template.yaml'
+            if '5.8' in name_dict[key]:
+                url = name_dict[key] + 'cfme-template.yaml'
+            else:
+                url = name_dict[key] + 'templates/' + 'cfme-template.yaml'
```
rlbabyuk: **Question:** This covers also 5.7 or just 5.9?

izapolsk: podified cfme exists from 5.8+

mfalesni: **REQUIRED**: `'templates/cfme-template.yaml'`

izapolsk:
@mfalesni , all those scripts are written in such style. url.join should be used there instead. However, I don't want to rewrite all such code occurrences. I just want to follow scripts' style.
Are you sure it makes sense to change one row ?

btw, I have a card to rewrite all those scripts.

izapolsk:
@mfalesni, as you noticed about I don't share your opinion on that but I fixed what you want.

mfalesni:
@izapolsk hehe, I didn't actually mean using urljoin, just not having to join two string constants, but I guess this is good as well :)

PR #5550, opened 2017-10-09
[RFR] Fix test_alerts creation of alert_collection

mshriver:
RHCFQE-4898

Little regression in creating an alert through the alert_collection object.


quarckster:
Already fixed in #5549

PR #5553, opened 2017-10-09
[1LP][RFR] test_vm_retire_extend: Add check to verify extended retirement date

LandoCalrizzian:
After extending the VM retirement date, verify that the WebUI is updated with the correct date
{{pytest: -v -k 'test_vm_retire_extend'}}

### DIFF
--- a/cfme/tests/automate/test_common_methods.py
+++ b/cfme/tests/automate/test_common_methods.py
@@ -49,6 +49,9 @@ def testing_vm(request, vm_name, setup_provider, provider, provisioning):
 
 @pytest.fixture(scope="function")
 def retire_extend_button(request):
+    """
+    Add an automate button that will extend the vm retirement date
+    """
     grp_name = "grp_{}".format(fauxfactory.gen_alphanumeric())
     grp = ButtonGroup(
         text=grp_name,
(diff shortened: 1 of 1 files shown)


cfme/tests/automate/test_common_methods.py
```diff
@@ -49,6 +49,9 @@ def _finalize():
 
 @pytest.fixture(scope="function")
 def retire_extend_button(request):
+    """
+    Add an automate button that will extend the vm retirement date
```
mshriver: :+1:

cfme/tests/automate/test_common_methods.py
```diff
@@ -105,14 +108,19 @@ def test_vm_retire_extend(request, testing_vm, soft_assert, retire_extend_button
...
     retire_extend_button()
 
-    # dajo - 20140920 - this fails because its not turning the calendar to the next month?
-    # wait_for(
-    #     lambda: testing_vm.retirement_date >= current_retirement_date + timedelta(days=14),
-    #     num_sec=60,
-    #     message="extend the retirement date by 14 days"
-    # )
+    # CFME automate vm_retire_extend method defaults to extending the date by 14 days
+    extend_duration_days = 14
```
mshriver: I'd much rather pull this from appliance configuration or something like that, since it could easily be changed from the default. This is fine for now though, later enhancements to the test (set button to non-default interval and make sure the right interval is applied)

mshriver:
59z PRT failure is due to missing search box that breaks find_quadicon, unrelated to this.

PR #5555, opened 2017-10-09
[1LP][RFR] Update services/test_rest_services, service names and collection fixtures

mshriver:
test is trying to get ordered service name from request message, but was
adding in the template name, which isn't present in 58z

As @mkoura suggested, get the dialog name from the request options, this is the service name.

RHCFQE-4905

## PRT Results
* **58z 75% passing**
There are 2 failures in 58z that I'd like to leave to @mkoura if he's alright with that. This PR fixes ~10 failures in the rest test in the latest builds because of UI changes and framework collections changes.

Limiting PRT to test cases that were modified
{{pytest: cfme/tests/services/test_rest_services.py --long-running -v -k "test_order_single_catalog_item or test_order_multiple_catalog_items or test_order_catalog_bundle or test_order_manual_approval or test_user_item_order or test_order_cart or test_power_service" }}

### DIFF
--- a/cfme/tests/services/test_rest_services.py
+++ b/cfme/tests/services/test_rest_services.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 import datetime
-import re
 
 import fauxfactory
 import pytest
@@ -8,7 +7,6 @@ import pytest
 from manageiq_client.api import ManageIQClient as MiqApi
 
 from cfme import test_requirements
-from cfme.automate.explorer.domain import DomainCollection
 from cfme.infrastructure.provider import InfraProvider
 from cfme.rest.gen_data import (
     _creating_skeleton,
@@ -39,6 +37,9 @@ pytestmark = [
     test_requirements.service,
     pytest.mark.tier(2)
 ]
+
+
+NUM_BUNDLE_ITEMS = 4
 
 
 @pytest.fixture(scope="module")
@@ -107,7 +108,7 @@ def catalog_bundle(request, dialog, service_catalog_obj, appliance, a_provider):
         service_dialog=dialog,
         service_catalog=service_catalog_obj,
         a_provider=a_provider,
-        num=4)
+        num=NUM_BUNDLE_ITEMS)
 
     uid = fauxfactory.gen_alphanumeric()
     bundle_name = 'test_rest_bundle_{}'.format(uid)
@@ -820,8 +821,8 @@ class TestServiceCatalogsRESTAPI(object):
 
         wait_for(_order_finished, num_sec=180, delay=10)
 
-        service_name = re.search(
-            r'\[({}[0-9-]*)\] '.format(template.name), service_request.message).group(1)
+        service_name = str(service_request.options['dialog']['dialog_service_name'])
+        assert '[{}]'.format(service_name) in service_request.message
         # this fails if the service with the `service_name` doesn't exist
         new_service = appliance.rest_api.collections.services.get(name=service_name)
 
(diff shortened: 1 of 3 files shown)


cfme/tests/services/test_rest_services.py
```diff
@@ -820,8 +820,12 @@ def _order_finished():
...
         wait_for(_order_finished, num_sec=180, delay=10)
 
-        service_name = re.search(
-            r'\[({}[0-9-]*)\] '.format(template.name), service_request.message).group(1)
+        # Regex to parse "[EVM] Service [NAME I CARE ABOUT] Provisioned Successfully
+        # group 'name', non-greedy .*? to get the ending ]
+        # (?!{<appliance.server_name>}) this is to filter out [EVM] part from matches
+        regex_string = '\[(?P<name>(?!{})*?)\]'.format(appliance.server_name())
+
+        service_name = re.search(regex_string, service_request.message).group('name')
```
mkoura: Due to https://github.com/ManageIQ/integration_tests/pull/5516 the service names no longer contain timestamp and ``template.name``. The name of the service is now the same as the label of the dialog that corresponds to the service template.
The correct approach is to get the dialog label instead of using regex, IMHO.

mshriver: Thanks Martin - I was definitely still playing with things here, as a similar pattern is used in other places in the test. I'll check the available objects to get this name in a better way - possibly changing this to assert the expected name that we already have is in the `service_request.message`.

cfme/tests/services/test_rest_services.py
```diff
@@ -1092,8 +1093,10 @@ def _order_approved():
 
         wait_for(_order_approved, num_sec=180, delay=10)
 
-        service_name = re.search(
-            r' \[([^\]]+)\] ', pending_request.message).group(1)
+        # service_name = re.search(
```
mshriver: Going to remove these commented lines, don't want to disrupt current PRT run though.

cfme/tests/services/test_rest_services.py
```diff
@@ -863,16 +865,22 @@ def _order_finished(service_request):
...
         def _finished():
-            appliance.rest_api.collections.services.action.delete(*new_services)
+            try:
+                appliance.rest_api.collections.services.action.delete(*new_services)
+            except APIException as ex:
+                if 'NotFound' in ex.message:
+                    logger.warning('APIException for service "NotFound" on delete, passing')
+                    pass
+                else:
+                    raise
```
mkoura: The services should exist at this point and delete should succeed. If it is not the case, something is wrong and this would only mask the issue.

mshriver: Noted @mkoura.

I added this because the test was failing in teardown with this API Exception.  I'll remove this change and let you handle this failure.

mshriver:
@mkoura I'm suspecting you meant to set 1LP+WIP and not just WIP.  I don't think removing that finalizer should require a new 1st level review, do you think so? I'll have to ping Dmitri for a re-review if so.

mkoura:
@mshriver You're right with the WIP, that's what I meant, sorry. Now you have 2 1LP approvals anyway :)

cfme/tests/services/test_rest_services.py
```diff
@@ -897,8 +897,8 @@ def _order_finished():
 
         wait_for(_order_finished, num_sec=2000, delay=10)
 
-        service_name = re.search(
-            r'\[({}[0-9-]*)\] '.format(catalog_bundle.name), service_request.message).group(1)
```
izapolsk: that a good idea to get rid of regexps :+1:

PR #5556, opened 2017-10-09
[1LP][RFR] Verify RBAC User/Group/Role is highlighted in accordion when Details view is displayed

LandoCalrizzian:
See https://bugzilla.redhat.com/show_bug.cgi?id=1498075

{{pytest: -v -k 'test_group_crud or test_user_crud or test_role_crud or test_delete_default_group or test_delete_default_roles or test_edit_default_group or test_edit_default_roles'}}

### DIFF
--- a/cfme/configure/access_control.py
+++ b/cfme/configure/access_control.py
@@ -77,7 +77,9 @@ class DetailsUserView(ConfigurationView):
     def is_displayed(self):
         return (
             self.title.text == 'EVM User "{}"'.format(self.context['object'].name) and
-            self.accordions.accesscontrol.is_opened
+            self.accordions.accesscontrol.is_opened and
+            (self.accordions.accesscontrol.tree.selected_item.text ==
+                self.context['object'].name)
         )
 
 
@@ -475,7 +477,9 @@ class DetailsGroupView(ConfigurationView):
     def is_displayed(self):
         return (
             self.accordions.accesscontrol.is_opened and
-            self.title.text == 'EVM Group "{}"'.format(self.context['object'].description)
+            self.title.text == 'EVM Group "{}"'.format(self.context['object'].description) and
+            (self.accordions.accesscontrol.tree.selected_item.text ==
+                self.context['object'].description)
         )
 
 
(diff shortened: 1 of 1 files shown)


cfme/configure/access_control.py
```diff
@@ -77,7 +77,9 @@ class DetailsUserView(ConfigurationView):
     def is_displayed(self):
         return (
             self.title.text == 'EVM User "{}"'.format(self.context['object'].name) and
-            self.accordions.accesscontrol.is_opened
+            self.accordions.accesscontrol.is_opened and
+            (self.accordions.accesscontrol.tree.selected_item.text ==
```
mkoura: No need for extra parentheses, indentation is enough to visually connect the expression on two lines.
The ``return`` statement already has parentheses so additional ones are not needed for line continuation.

cfme/configure/access_control.py
```diff
@@ -955,7 +959,9 @@ class DetailsRoleView(RoleForm):
     def is_displayed(self):
         return (
             self.accordions.accesscontrol.is_opened and
-            self.title.text == 'Role "{}"'.format(self.context['object'].name)
+            self.title.text == 'Role "{}"'.format(self.context['object'].name) and
+            (self.accordions.accesscontrol.tree.selected_item.text ==
+                self.context['object'].name)
```
mfalesni: **REQUIRED**: I would probably replace `.tree.selected_item.text` with `.tree.currently_selected[-1]` as by doing `.text` you bypass widgetastic browser. `currently_selected` returns a list or None if nothing selected.

mfalesni: And you could actually check the whole tree path in there to check you indeed are where you should be, but that is optional in this case.

cfme/configure/access_control.py
LandoCalrizzian: Updated to remove the parenthesis. It was originally copy/pasted from another line of code that was longer than 100 chars which used them to wrap it.

cfme/configure/access_control.py
LandoCalrizzian: @mfalesni Updated to use tree.currently_selected w/ a comment to explain the use of the last item index.

PR #5577, opened 2017-10-12
[RFR] Updated RequestCollection calls according to collection v3 model

izapolsk:


izapolsk:
some PRT results are red but this isn't related to my changes.

PR #5579, opened 2017-10-12
[1LP][RFR] Automate conversion collv3

psav:
Converted Automate and removed collections v2 objects

{{pytest: cfme/tests/automate/}}


dajoRH:
Lint report for commit 2779b03f2884372e2d781f2ca1a8e922a83450a0:

`cfme/automate/explorer/namespace.py`:
- :red_circle: Line 12:1: **F401** *'cfme.modeling.base.parent_of_type' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

cfme/automate/explorer/domain.py
```diff
@@ -88,8 +88,10 @@ class Domain(BaseEntity, Fillable):
     """A class representing one Domain in the UI."""
 
     def __init__(
-            self, collection, name, description, enabled=None, locked=None,
+            self, collection, name, description=None, enabled=None, locked=None,
```
izapolsk: description is not necessary here ?

cfme/automate/explorer/instance.py
```diff
@@ -336,6 +249,89 @@ def delete_if_exists(self):
...
+                'display_name': display_name,
+                'description': description,
+                'fields': fields,
+            }.items()
+            if v is not None}
+        add_page.fill(fill_dict)
+        if cancel:
+            add_page.cancel_button.click()
+            add_page.flash.assert_no_error()
+            add_page.flash.assert_message('Add of new Automate Instance was cancelled by the user')
```
mshriver: **Optional**
Consider `flash.assert_success_message()`, which calls `assert_no_error` and `assert_message`

cfme/automate/explorer/instance.py
```diff
@@ -336,6 +249,89 @@ def delete_if_exists(self):
...
+        for instance in instances:
+            parents.add(instance.parent)
+        if len(parents) > 1:
+            raise ValueError('You passed instances that are not under one class.')
+
+        checked_instances = []
+        if not all_page.instances.table.is_displayed:
+            raise ValueError('No instance found!')
+        all_page.instances.table.uncheck_all()
+        for row in all_page.instances.table:
```
mshriver: This doesn't account for multiple pages, I think we're going to need a generic method to handle this so we don't end up repeating this loop in multiple collection delete methods.

PR #5582, opened 2017-10-12
[1LP][RFR] Fix template_upload_vsphere, wrapanapi return

mshriver:
add_disk_to_vm was returning tuple, unpack it.


lkhomenk:
LGTM

PR #5587, opened 2017-10-12
[1LP][RFR] Handling vm console opening in SSUI and returning console handle to the test

kedark3:
Purpose or Intent
=================

__Updating test and other files__ to be able to handle the VM Console opening for SSUI. I have tried to use existing HTML5 console code by slight modification to achieve DRY.

{{pytest: -v --long-running cfme/tests/ssui/test_ssui_myservice.py -k 'test_vm_console' }}

### DIFF
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -243,6 +243,26 @@ class BaseVM(Pretty, Updateable, PolicyProfileAssignable, WidgetasticTaggable,
                 #      for a console tab.
                 return handle
 
+    @property
+    def vm_console(self):
+        """Get the consoles window handle, and then create a VMConsole object, and store
+        the VMConsole object aside.
+        """
+        console_handle = self.console_handle
+
+        if console_handle is None:
+            raise TypeError("Console handle should not be None")
+
+        appliance_handle = self.appliance.browser.widgetastic.window_handle
+        logger.info("Creating VMConsole:")
+        logger.info("   appliance_handle: {}".format(appliance_handle))
+        logger.info("     console_handle: {}".format(console_handle))
+        logger.info("               name: {}".format(self.name))
+
+        return VMConsole(appliance_handle=appliance_handle,
+                console_handle=console_handle,
+                vm=self)
+
     def delete(self, cancel=False, from_details=False):
         """Deletes the VM/Instance from the VMDB.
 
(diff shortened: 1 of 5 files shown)


dajoRH:
Lint report for commit 2dab0897f77fd3c18f6f2d56e1d61bc7d5383453:

`cfme/services/myservice/ssui.py`:
- :bangbang: Line 89:19: **E261** *at least two spaces before inline comment*

**No commit flaws detected**.

*CFME QE Bot*

cfme/services/myservice/ssui.py
```diff
@@ -80,8 +81,13 @@ def update(self, updates):
 
 
 @MyService.launch_vm_console.external_implementation_for(ViaSSUI)
-def launch_vm_console(self):
+def launch_vm_console(self, catalog_item):
     navigate_to(self, 'VM Console')
+    vm_obj = VM.factory(catalog_item.provisioning_data.get('vm_name') + '001',
+     catalog_item.provider, template_name=catalog_item.catalog_name)
```
quarckster: **REQUIRED**
Here is the indentation error.

cfme/services/myservice/ssui.py
```diff
@@ -80,8 +81,13 @@ def update(self, updates):
...
+    import time
+    time.sleep(10)  # used temporarily, will be replaced by wait_for appropriately.
```
quarckster: What kind of problems are here? Why you can't use `wait_for`?

kedark3:
Upstream failure is due to change in SSUI's html elements, and framework update will be required to make it work by changing the xpath selectors.

cfme/fixtures/service_fixtures.py
```diff
@@ -80,4 +80,4 @@ def order_catalog_item_in_ops_ui(appliance, provider, provisioning, vm_name, dia
                                                                  partial_check=True)
     provision_request.wait_for_request()
     assert provision_request.is_finished()
-    return catalog_item.name
+    return catalog_item
```
quarckster: Are you sure this fixture is not used in other tests except `test_myservice_crud` and `test_vm_console`?

kedark3: Hey @quarckster, yes I made that sure by talking to @sshveta. She mentioned that it shouldn't be a problem.

cfme/common/vm.py
```diff
@@ -353,25 +373,8 @@ def open_console(self, console='VM Console', invokes_alert=False, cancel=False):
         # Click console button given by type
         view.toolbar.access.item_select(console, handle_alert=None
             if invokes_alert is False else True)
+        self.vm_console
```
izapolsk: Q: can you please explain why vm_console is property ? and why you don't care of returned class ?

cfme/services/myservice/ssui.py
```diff
@@ -80,8 +81,15 @@ def update(self, updates):
 
 
 @MyService.launch_vm_console.external_implementation_for(ViaSSUI)
-def launch_vm_console(self):
+def launch_vm_console(self, catalog_item):
     navigate_to(self, 'VM Console')
+    vm_obj = VM.factory(catalog_item.provisioning_data.get('vm_name') + '001',
```
izapolsk: Q: this is unclear to me why you use such vm name. could you please explain ?

kedark3: So when we provision a VM through a service, suppose name of vm is given as test-scat-xRjy, then at the end of provisioning the service will create a VM with name test-scat-xRjy001. If I reran the same service, next VM that will be created will have name test-scat-xRjy002, and so on. In case of this test, I am only creating a single VM by running the service only once, and hence I need to append that 001 to VM name.

cfme/common/vm.py
kedark3: the vm_console property of the vm class was designed by James Oden. He is currently on short term leave. but I remember he told me why he did that so and I don't remember that exactly. I think what he told me was as follows( if I recall correctly): that each VM will have only one console open at any given instance and hence it makes sense to make it as a property.

cfme/services/myservice/ssui.py
izapolsk: I woudn't worry about this statement if you added this to test.
However, you are adding this statement to the part of framework.

kedark3:
Waiting for merge of #5629

PR #5597, opened 2017-10-13
[1LP][RFR] Workaround and blockers for BZ 1501941 on 5.9 and upstream

mkoura:
{{pytest: -v cfme/tests/cloud_infra_common/test_rest_providers.py}}

### DIFF
--- a/cfme/tests/cloud_infra_common/test_rest_providers.py
+++ b/cfme/tests/cloud_infra_common/test_rest_providers.py
@@ -5,8 +5,9 @@ from cfme import test_requirements
 from cfme.cloud.provider.azure import AzureProvider
 from cfme.common.provider import CloudInfraProvider
 from cfme.utils import error, testgen
+from cfme.utils.blockers import BZ
 from cfme.utils.rest import assert_response
-from cfme.utils.version import pick
+from cfme.utils.version import current_version, pick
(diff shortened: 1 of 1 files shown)


cfme/tests/cloud_infra_common/test_rest_providers.py
```diff
@@ -22,8 +23,20 @@ def delete_provider(appliance, name):
         return
 
     prov = provs[0]
-    prov.action.delete()
-    prov.wait_not_exists(num_sec=30)
+
+    # workaround for BZ1501941
+    if current_version() >= '5.9' and BZ(1501941, forced_streams=['5.9', 'upstream']).blocks:
```
psav: **REQUIRED:** ```appliance.version``` here is more preferable, ```current_version()``` will one day be gone

cfme/tests/cloud_infra_common/test_rest_providers.py
```diff
@@ -22,8 +23,20 @@ def delete_provider(appliance, name):
...
+        prov.action.edit(enabled=False)
+        for __ in range(3):
```
psav: **REQUIRED:** Instead of another loop here, why not use ```waitfor``` with a tiny func?

PR #5598, opened 2017-10-13
[1LP][RFR] Check V2_key perms

OpCode1300:
Adding tests for file permissions on appliance


OpCode1300:
The failure in 58z is expected as the change has not been back-ported.
https://bugzilla.redhat.com/show_bug.cgi?id=1489556

cfme/tests/security/test_file_permissions.py
```diff
@@ -0,0 +1,10 @@
+import pytest
+from cfme.utils import version
+
+
+@pytest.mark.uncollectif(lambda: version.current_version() < '5.6', reason='Only valid for >5.7')
```
psav: **REQUIRED:** This line can go, the framework is not valid for 5.6 and below

cfme/tests/security/test_file_permissions.py
```diff
@@ -0,0 +1,10 @@
...
+def test_v2_key_permissions(appliance):
+    """Verifies that the v2_key has proper permissions"""
+    stdout = appliance.ssh_client.run_command(
```
psav: You should be able to do
```python
stdout = appliance.ssh_client.run_command(........).output
assert int(stdout) == 400
```

Issue #5599, opened 2017-10-15
Navigation to Add Provider menu stopped working on v5.9 


pavelzag:
This part seems to bring me to the Containers Providers main menu and not to the Add Provider screen as it should and as it still works on older versions:

add_view = navigate_to(self, 'Add')
in 
common/provider.py

gshefer:
This failure expected.
The reason for this is that the setup provider UI has been (dramatically) changed.
We should update the widgetastic objects accordingly.

mshriver:
I believe this is being fixed in #5604 by @jawatts

PR #5604, opened 2017-10-16
[1LP][RFR] Initial fixes for ContainerProvider form for 5.9

jawatts:
In 5.9, there were recent changes that renamed the 'Hawkular' endpoint from "Hawkular" to "Metrics" and added a new drop down widget called Metrics. The changes in this PR address these updates and allow the creation of a ContainerProvider in 5.9 

{{ pytest: --use-provider cm-env2 --long-running cfme/tests/containers/test_start_page.py cfme/tests/containers/test_ssl_providers.py}}

### DIFF
--- a/cfme/common/provider_views.py
+++ b/cfme/common/provider_views.py
@@ -114,6 +114,16 @@ class MiddlewareProviderDetailsView(ProviderDetailsView):
                 self.navigation.currently_selected == ['Middleware', 'Providers'])
 
 
+class ContainerProviderDetailsView(ProviderDetailsView):
+    """
+     Container Details page
+    """
+    @property
+    def is_displayed(self):
+        return (super(ContainerProviderDetailsView, self).is_displayed and
+                self.navigation.currently_selected == ['Compute', 'Containers', 'Providers'])
+
+
 class ProviderTimelinesView(TimelinesView, BaseLoggedInPage):
     """
      represents Timelines page
@@ -274,7 +284,7 @@ class ProvidersView(BaseLoggedInPage):
     including_entities = View.include(ProviderEntitiesView, use_parent=True)
 
 
-class ContainersProvidersView(ProvidersView):
+class ContainerProvidersView(ProvidersView):
     """
      represents Main view displaying all Containers providers
     """
@@ -287,7 +297,7 @@ class ContainersProvidersView(ProvidersView):
 
     @property
     def is_displayed(self):
-        return (super(ContainersProvidersView, self).is_displayed and
+        return (super(ContainerProvidersView, self).is_displayed and
                 self.navigation.currently_selected == ['Compute', 'Containers', 'Providers'] and
                 self.entities.title.text == self.SUMMARY_TEXT)
 
@@ -393,9 +403,9 @@ class CloudProviderAddView(ProviderAddView):
                 self.title.text == 'Add New Cloud Provider')
 
 
-class ContainersProviderAddView(ProviderAddView):
+class ContainerProviderAddView(ProviderAddView):
     """
-     represents Containers Provider Add View
+     represents Container Provider Add View
     """
     prov_type = BootstrapSelect(id='ems_type')
 
@@ -404,6 +414,15 @@ class ContainersProviderAddView(ProviderAddView):
         return (super(ProviderAddView, self).is_displayed and
                 self.navigation.currently_selected == ['Compute', 'Containers', 'Providers'] and
                 self.title.text == 'Add New Containers Provider')
+
+
+class ContainerProviderAddViewUpdated(ContainerProviderAddView):
+    """
+     Additional widgets for builds 5.9 and up
+    """
+
+    metrics_type = BootstrapSelect(id='metrics_selection')
+    alerts_type = BootstrapSelect(id='alerts_selection')
 
 
 class MiddlewareProviderAddView(ProviderAddView):
@@ -461,15 +480,24 @@ class CloudProviderEditView(ProviderEditView):
                 self.title.text == 'Edit Cloud Provider')
 
 
-class ContainersProviderEditView(ProviderEditView):
+class ContainerProviderEditView(ProviderEditView):
     """
-     represents Containers Provider Edit View
+     represents Container Provider Edit View
     """
     @property
     def is_displayed(self):
         return (super(ProviderEditView, self).is_displayed and
                 self.navigation.currently_selected == ['Compute', 'Containers', 'Providers'] and
                 'Edit Containers Provider' in self.title.text)
+
+
+class ContainerProviderEditViewUpdated(ContainerProviderEditView):
+    """
+     Additional widgets for builds 5.9 and up
+    """
+
+    metrics_type = BootstrapSelect(id='metrics_selection')
+    alerts_type = BootstrapSelect(id='alerts_selection')
 
 
 class MiddlewareProviderEditView(ProviderEditView):
--- a/cfme/containers/provider/__init__.py
+++ b/cfme/containers/provider/__init__.py
@@ -3,23 +3,31 @@ import random
 from random import sample
 from traceback import format_exc
 
-
 from navmazing import NavigateToSibling, NavigateToAttribute
-from widgetastic_patternfly import (SelectorDropdown, Dropdown, BootstrapSelect,
-                                    Input, Button, Tab, FlashMessages)
+
+from widgetastic_patternfly import (
+    SelectorDropdown, Dropdown, BootstrapSelect, Input, Button, Tab, FlashMessages
+)
+from widgetastic.utils import VersionPick, Version
 from widgetastic.widget import Text, View, TextInput
+from widgetastic_manageiq import (
+    SummaryTable, BreadCrumb, Accordion, ManageIQTree
+)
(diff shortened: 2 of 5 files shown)


mshriver:
Can you try adding the metrics type selection widget, and the data in cfme_data.yaml, so that the field is selected (with 'Hawkular') for the CFME container providers? We may be able to fully hack together container provider add in the shorter term before figuring a better way to dynamically tie the endpoints fill to a reference widget.

dajoRH:
Lint report for commit 763e72941a00194c33ff6702e848c367a29e325d:

`cfme/containers/provider/__init__.py`:
- :red_circle: Line 334:9: **N802** *function name should be lowercase*
- :red_circle: Line 420:9: **N802** *function name should be lowercase*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

jawatts:
Manual Testing with miq shell:
Both create and edit have been tested against upstream 5.9, downstream 5.9, and downstream 5.8.

cfme/common/provider_views.py
```diff
@@ -114,6 +114,16 @@ def is_displayed(self):
                 self.navigation.currently_selected == ['Middleware', 'Providers'])
 
 
+class ContainerProviderDetailsView(ProviderDetailsView):
```
mshriver: **Required**

Keep the pluralization consistent here, `ContainerProvider(...)` here and `ContainersProvider(...)` class names below.

cfme/containers/provider/__init__.py
```diff
@@ -310,14 +346,23 @@ def resetter(self):
 
 @navigator.register(ContainersProvider, 'Edit')
 class Edit(CFMENavigateStep):
-    VIEW = ContainersProviderEditView
+
+    def pickcontainerprovidereditview(self):
```
mshriver: use `_` here to make this a bit more readable, consider `container_provider_edit_view_class`.

Calling it `pick(...)` is misleading since it doesn't actually pick, it returns a VersionPick object. The method could take a version argument and call pick before returning, where the caller passes appliance.version

cfme/containers/provider/__init__.py
```diff
@@ -310,14 +346,23 @@ def resetter(self):
...
+        return self.pickcontainerprovidereditview().pick(self.obj.appliance.version)
     prerequisite = NavigateToSibling('All')
 
     def step(self):
-        self.prerequisite_view.entities.get_entity(self.obj.name,
-                                                   surf_pages=True).check()
+        self.prerequisite_view.entities.get_entity(by_name=self.obj.name, surf_pages=True).check()
         self.prerequisite_view.toolbar.configuration.item_select(
-            'Edit Selected Containers Provider')
+            'Edit Selected Containers Provider'
```
mshriver: keep old style here

cfme/containers/provider/openshift.py
```diff
@@ -1,15 +1,21 @@
 from cached_property import cached_property
+from os import path
 
 from . import ContainersProvider
-from cfme.utils.varmeth import variable
+
+from cfme.containers.provider import (
+    ContainersProviderDefaultEndpoint, ContainersProviderEndpointsForm
+)
+
```
mshriver: drop these newlines, single block for local imports

cfme/containers/provider/openshift.py
```diff
@@ -76,12 +82,20 @@ def href(self):
...
-        return {
+
+        mapping = {
             'name': self.name,
-            'prov_type': 'OpenShift Container Platform',
             'zone': self.zone,
+            'metrics_type': self.metrics_type
         }
 
+        if appliance_is_downstream():
```
mshriver: **Required**
This is using store.current_appliance, instead reference `self.appliance.is_downstream` to specifically query the appliance associated with this provider.

cfme/common/provider_views.py
jawatts: :+1:

cfme/containers/provider/openshift.py
```diff
@@ -76,12 +82,20 @@ def href(self):
...
+            mapping['prov_type'] = 'OpenShift Container Platform'
+        else:
+            mapping['prov_type'] = 'OpenShift'
+
```
mshriver: **Optional**
Style - consider turning this into a ternary, though this example is 102 chars so it needs to be split

```
mapping['prov_type'] = 
    'OpenShift Container Platform' 
    if self.appliance.is_downstream 
    else 'OpenShift'

File: cfme/containers/provider/openshift.py [hunk 7]
```diff
@@ -109,16 +123,24 @@ def from_config(cls, prov_config, prov_key, appliance=None):
                 endpoints[endp] = OpenshiftDefaultEndpoint(**prov_config['endpoints'][endp])
             elif HawkularEndpoint.name == endp:
                 endpoints[endp] = HawkularEndpoint(**prov_config['endpoints'][endp])
+            # TODO Add Prometheus and logic for having to select or the other based on metrcis_type
             else:
                 raise Exception('Unsupported endpoint type "{}".'.format(endp))
 
+        if current_version() >= '5.9':
```
mshriver: **Required**
Use self.appliance.version here

mshriver: Though this requires accounting for when appliance isn't passed.

cfme/containers/provider/__init__.py
jawatts: :+1:

cfme/containers/provider/__init__.py
jawatts: :+1:

cfme/containers/provider/openshift.py
jawatts: :+1:

mshriver:
Should fix #5599

cfme/containers/provider/openshift.py
mshriver: Ignore this comment, we'll allow this for now.

cfme/containers/provider/openshift.py
```diff
@@ -1,15 +1,21 @@
...
+from cfme.common.provider import DefaultEndpoint
+
+from cfme.utils.ocp_cli import OcpCli
 from cfme.utils.path import data_path
-from os import path
+from cfme.utils.varmeth import variable
+from cfme.utils.version import current_version, appliance_is_downstream
+
+
 from wrapanapi.containers.providers.openshift import Openshift
```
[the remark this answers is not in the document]
jawatts: :+1:

cfme/containers/provider/openshift.py
jawatts: :+1:

cfme/containers/provider/openshift.py
mshriver: @jkrocil Suggested a better approach to this within your PR scope.

Don't try to version pick metrics_type here, just pass what's in the yaml (or None) via `prov_config.get()`.

Then in `OpenshiftProvider.__init__`, version pick based on the appliance version there.

Currently OpenshiftProvider has an optional appliance arg, but this will be a required option when converted to CEMv3.

cfme/common/provider_views.py
mshriver: Can you confirm that you audited for uses of the plural class names and updated any existing imports/uses of those names in the rest of the framework?

cfme/containers/provider/__init__.py
```diff
@@ -28,10 +22,21 @@
 from cfme.utils.varmeth import variable
 from cfme.utils.log import logger
 from cfme.utils.wait import wait_for
-from widgetastic_manageiq import SummaryTable, BreadCrumb, Accordion,\
-    ManageIQTree
+
+
+from navmazing import NavigateToSibling, NavigateToAttribute
+
+from widgetastic_patternfly import (
```
mshriver: These should all be above the local imports.

```
<python builtin imports>

<pip installed imports>

<local imports>

File: cfme/containers/provider/__init__.py [hunk 10]
```diff
@@ -142,21 +162,23 @@ class ContainersProvider(BaseProvider, Pretty):
     quad_name = None
     db_types = ["ContainerManager"]
     endpoints_form = ContainersProviderEndpointsForm
-    all_view = ContainersProvidersView
-    details_view = ContainersProviderDetailsView
+    all_view = ContainerProvidersView
```
mshriver: Hopefully everything trying to use these views was doing so though these attributes

cfme/containers/provider/__init__.py
```diff
@@ -142,21 +162,23 @@ class ContainersProvider(BaseProvider, Pretty):
...
             zone=None,
+            metrics_type=None,
             endpoints=None,
             provider_data=None,
             appliance=None):
         Navigatable.__init__(self, appliance=appliance)
         self.name = name
         self.key = key
         self.zone = zone
+        self.metrics_type = metrics_type
```
mshriver: Here's where you can use appliance.version() to set this to None for older versions. Appliance ultimately should not be an optional parameter here

mshriver:
PRT 17648 is false negative, need to add a containers provider to test against via {{pytest: --use-provider <key>}}

PR #5607, opened 2017-10-17
[1LP][RFR] add Openstack cloud volumes tests

okolisny:
Adding create, edit and delete tests for Openstack Cloud provider

{{pytest: cfme/tests/openstack/cloud/test_volumes.py --use-provider rhos7-ga}}

### DIFF
--- a/cfme/storage/volume.py
+++ b/cfme/storage/volume.py
@@ -251,10 +251,10 @@ class VolumeCollection(BaseCollection):
     def create(self, name, storage_manager, tenant, size, provider):
         """Create new storage volume"""
         view = navigate_to(self, 'Add')
-        view.form.storage_manager.fill(storage_manager)
-        view.form.tenant.fill(tenant)
-        view.form.volume_name.fill(name)
-        view.form.size.fill(size)
+        view.form.fill({'storage_manager': storage_manager,
+                        'tenant': tenant,
+                        'volume_name': name,
+                        'size': size})
         view.form.add.click()
         view.flash.assert_success_message('Cloud Volume "{}" created'.format(name))
 
--- a/cfme/storage/volume.py
+++ b/cfme/storage/volume.py
@@ -8,12 +8,12 @@ from widgetastic_manageiq import (
     BaseListEntity,
     BaseQuadIconEntity,
     BaseTileIconEntity,
+    BootstrapSelect,
     BreadCrumb,
     ItemsToolBarViewSelector,
     JSBaseEntity,
     ManageIQTree,
     NonJSBaseEntity,
-    Select,
     SummaryTable,
     TextInput,
     Version,
@@ -147,10 +147,10 @@ class VolumeAddEntities(View):
 
 
 class VolumeAddForm(View):
-    storage_manager = Select(name='storage_manager_id')
+    storage_manager = BootstrapSelect(name='storage_manager_id')
     volume_name = TextInput(name='name')
     size = TextInput(name='size')
-    tenant = Select(name='cloud_tenant_id')
+    tenant = BootstrapSelect(name='cloud_tenant_id')
     add = Button('Add')
     cancel = Button('Cancel')
 
--- a/cfme/storage/volume.py
+++ b/cfme/storage/volume.py
@@ -171,6 +171,17 @@ class VolumeAddView(VolumeView):
     form = View.nested(VolumeAddForm)
 
 
+class VolumeEditView(VolumeView):
+    @property
+    def is_displayed(self):
+        return False
+
+    flash = FlashMessages('.//div[@id="flash_msg_div"]/div[@id="flash_text_div" or '
+                          'contains(@class, "flash_text_div")]')
+    volume_name = TextInput(name='name')
+    save = Button('Save')
+
+
 @attr.s
 class Volume(BaseEntity):
     # Navigation menu option
@@ -195,6 +206,14 @@ class Volume(BaseEntity):
         except TimedOutError:
             logger.error('Timed out waiting for Volume to disappear, continuing')
 
+    def edit(self, name):
+        """Edit cloud volume"""
+        view = navigate_to(self, 'Edit')
+        view.volume_name.fill(name)
+        view.save.click()
+        view.flash.assert_success_message('Cloud Volume "{}" updated'.format(name))
+        self.name = name
+
     def delete(self, wait=True):
         """Delete the Volume"""
 
@@ -214,6 +233,16 @@ class Volume(BaseEntity):
             return True
         except ItemNotFound:
             return False
+
+    @property
+    def size(self):
+        view = navigate_to(self, 'Details')
+        return view.entities.properties.get_text_of('Size')
+
+    @property
+    def tenant(self):
+        view = navigate_to(self, 'Details')
+        return view.entities.relationships.get_text_of('Cloud Tenants')
 
 
 @attr.s
@@ -294,3 +323,12 @@ class VolumeAdd(CFMENavigateStep):
 
     def step(self, *args, **kwargs):
         self.prerequisite_view.toolbar.configuration.item_select('Add a new Cloud Volume')
+
+
+@navigator.register(Volume, 'Edit')
+class VolumeEdit(CFMENavigateStep):
+    VIEW = VolumeEditView
+    prerequisite = NavigateToSibling('Details')
+
+    def step(self, *args, **kwargs):
+        self.prerequisite_view.toolbar.configuration.item_select('Edit this Cloud Volume')
--- a/cfme/storage/volume.py
+++ b/cfme/storage/volume.py
@@ -164,8 +164,6 @@ class VolumeAddView(VolumeView):
             self.entities.title.text == expected_title and
             self.entities.breadcrumb.active_location == expected_title)
 
-    flash = FlashMessages('.//div[@id="flash_msg_div"]/div[@id="flash_text_div" or '
-                          'contains(@class, "flash_text_div")]')
     entities = View.nested(VolumeAddEntities)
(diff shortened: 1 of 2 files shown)


cfme/storage/volume.py
```diff
@@ -204,19 +226,44 @@ def delete(self, wait=True):
...
 
 @attr.s
 class VolumeCollection(BaseCollection):
     """Collection object for the :py:class:'cfme.storage.volume.Volume'. """
     ENTITY = Volume
 
+    def create(self, name, storage_manager, tenant, size, provider):
+        """Create new storage volume"""
+        view = navigate_to(self, 'Add')
+        view.form.storage_manager.fill(storage_manager)
```
psav: **REQUIRED:** We should use the form fill with a dict of data here.

cfme/tests/openstack/cloud/test_volumes.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+pytest_generate_tests = testgen.generate([OpenStackProvider], scope='module')
+
+pytestmark = [pytest.mark.usefixtures("setup_provider_modscope")]
+
+
+@pytest.fixture(scope='module')
+def volume_size():
+    return 1
```
psav: **QUESTION:** Is there a reason for this to be a fixture?

okolisny: Just wanted to keep this value in one place..

cfme/storage/volume.py
```diff
@@ -191,6 +205,14 @@ def refresh():
         except TimedOutError:
             logger.error('Timed out waiting for Volume to disappear, continuing')
 
+    def edit(self, name):
```
rlbabyuk: **Question:**  Do we have only name editable for volume?

okolisny: Yes, at least as I can see it for volume of Openstack provider.

okolisny:
@psav ready

cfme/storage/volume.py
```diff
@@ -163,10 +164,23 @@ def is_displayed(self):
             self.entities.title.text == expected_title and
             self.entities.breadcrumb.active_location == expected_title)
 
+    flash = FlashMessages('.//div[@id="flash_msg_div"]/div[@id="flash_text_div" or '
```
mshriver: This should be redundant, as it should be available through inheritence of VolumeView -> BaseLoggedInPage  (where flash is defined)

cfme/storage/volume.py
```diff
@@ -163,10 +164,23 @@ def is_displayed(self):
...
     entities = View.nested(VolumeAddEntities)
     form = View.nested(VolumeAddForm)
 
 
+class VolumeEditView(VolumeView):
+    @property
+    def is_displayed(self):
+        return False
+
+    flash = FlashMessages('.//div[@id="flash_msg_div"]/div[@id="flash_text_div" or '
```
mshriver: Again, should be available through inheritance.

cfme/storage/volume.py
```diff
@@ -204,19 +226,44 @@ def delete(self, wait=True):
...
+            navigate_to(self, 'Details')
             return True
-        except ItemNotFound:
+        except VolumeNotFound:
             return False
 
+    @property
+    def size(self):
+        view = navigate_to(self, 'Details')
+        return view.entities.properties.get_text_of('Size')
```
mshriver: :+1:

cfme/tests/openstack/cloud/test_volumes.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+@pytest.yield_fixture(scope='function')
+def volume(appliance, provider, volume_size):
+    collection = appliance.collections.volumes
+    volume = collection.create(name=fauxfactory.gen_alpha(),
+                               storage_manager='{} Cinder Manager'.format(provider.name),
+                               tenant=provider.get_yaml_data()['tenant'],
```
mshriver: **Optional**

Consider using `data` property here instead of calling get_yaml_data() directly.

```
provider.data.get('tenant')

File: cfme/tests/openstack/cloud/test_volumes.py (hunk 2 above)
Comment (mshriver) [reply cfme/tests/openstack/cloud/test_volumes.py]: You can just set a constant in the module and reference it in the tests.

File: cfme/tests/openstack/cloud/test_volumes.py [hunk 8]
```diff
@@ -0,0 +1,53 @@
...
+
+
+def test_create_volume(volume, volume_size, provider):
+    assert volume.exists
+    assert volume.size == '{} GB'.format(volume_size)
+    assert volume.tenant == provider.get_yaml_data()['tenant']
+
+
+def test_edit_volume(volume):
+    volume.edit(fauxfactory.gen_alpha())
```
mshriver: **Optional**

I'd rather see this name stored and then asserted against after the edit+assert exists

cfme/tests/openstack/cloud/test_volumes.py
mshriver: https://github.com/ManageIQ/integration_tests/pull/5555/files#diff-9ee724665011348e6753f3fd66392437R42

here's an example of one of my open PR's with such a change. I didn't like '4' showing up in the test methods

cfme/tests/openstack/cloud/test_volumes.py
okolisny: Hm, I'm not following you.. New name is recorded inside `edit` method. And then goes assert `volume.exists`

mshriver: if the edit method doesn't work, and doesn't raise exception, you'll never catch it because the name didn't change and it can still be navigated to.

okolisny: Ah.. Well, name is being changed after the flash message assertion, and it is like 99% success. However, you are right

cfme/storage/volume.py
```diff
@@ -147,9 +147,10 @@ class VolumeAddEntities(View):
 
 
 class VolumeAddForm(View):
+    storage_manager = Select(name='storage_manager_id')
```
mshriver: This widget is only present in CFME 58z+.

cfme/tests/openstack/cloud/test_volumes.py
```diff
@@ -0,0 +1,53 @@
...
+
+VOLUME_SIZE = 1
+
+
+@pytest.yield_fixture(scope='function')
+def volume(appliance, provider):
+    collection = appliance.collections.volumes
+    volume = collection.create(name=fauxfactory.gen_alpha(),
+                               storage_manager='{} Cinder Manager'.format(provider.name),
+                               tenant=provider.data.get('tenant'),
```
mshriver: This value is actually under `provisioning` in the yamls, or should be, so this is 'None' in testing, and the form fails to validate (in 58z/upstream)

cfme/tests/openstack/cloud/test_volumes.py
```diff
@@ -0,0 +1,53 @@
...
+pytestmark = [pytest.mark.usefixtures("setup_provider_modscope")]
+
+VOLUME_SIZE = 1
+
+
+@pytest.yield_fixture(scope='function')
+def volume(appliance, provider):
+    collection = appliance.collections.volumes
+    volume = collection.create(name=fauxfactory.gen_alpha(),
+                               storage_manager='{} Cinder Manager'.format(provider.name),
```
mshriver: The storage_manager is only a field value in 58z, and is getting passed as a fill value. This raises NoSuchElement exceptions in 57z, because the form widget doesn't exist.

cfme/storage/volume.py
```diff
@@ -147,9 +147,10 @@ class VolumeAddEntities(View):
...
     volume_name = TextInput(name='name')
     size = TextInput(name='size')
-    tenant = BootstrapSelect(id='cloud_tenant_id')
+    tenant = Select(name='cloud_tenant_id')
```
mshriver: I'm seeing 'bootstrap-select' classes in these div's, I think these need to stay as BootstrapSelect, not Select.

cfme/storage/volume.py
mshriver: use BootstrapSelect

cfme/tests/openstack/cloud/test_volumes.py
okolisny: Added version.pick workaround for 5.7 here

cfme/storage/volume.py
okolisny: Done.

cfme/tests/openstack/cloud/test_volumes.py
okolisny: Added tenant get from provisioning

cfme/storage/volume.py
okolisny: done

cfme/storage/volume.py
```diff
@@ -204,19 +222,44 @@ def delete(self, wait=True):
...
 
+    def create(self, name, storage_manager, tenant, size, provider):
+        """Create new storage volume"""
+        view = navigate_to(self, 'Add')
+        view.form.fill({'storage_manager': storage_manager,
+                        'tenant': tenant,
+                        'volume_name': name,
+                        'size': size})
+        view.form.add.click()
+        view.flash.assert_success_message('Cloud Volume "{}" created'.format(name))
```
mshriver: This flash message is for CFME 58z+, and needs to be version picked for 57z:
' Creating Cloud Volume "<name>"'

otsuman:
Looks good, but pls take a look on PRT, it fails when trying to wait for a volume after creation

PR #5617, opened 2017-10-17
[WIPTEST] Remove match_location from some files

rsnyman:
Start removing the `match_location()` function. In these cases, we probably don't need to replace it with anything, as these are from base views which are then inherited by child views.


dajoRH:
*CFME QE Bot*

psav:
This is the conflict I spoke of :( sorry man

PR #5625, opened 2017-10-18
[1LP][RFR] - Fixed reset db test

lcouzens:
Fix to test_black_console_db_reset


izapolsk:
@lcouzens, please pay attention on PRT results, there are 3+ tests failed on int convertion of incorrect value.

PR #5626, opened 2017-10-18
[1LP][RFR] removed old provisioning form from test_host_provisioning

izapolsk:
It turned out that template tester fails due to absent web_ui provisioning form. 
This PR should sort that issue out.


mshriver:
tests were uncollected in PRT, going to merge as the collection shows that the import is fixed. potential test failures be damned ;)

PR #5627, opened 2017-10-18
[RFR] parent service tests changes after BZ 1441412 fix

mkoura:
* fixing test_retire_parent_service_now
* new test test_add_child_resource

{{pytest: -v --long-running -k "test_add_child_resource or test_retire_parent_service_now" cfme/tests/services/test_rest_services.py}}

### DIFF
--- a/cfme/tests/services/test_rest_services.py
+++ b/cfme/tests/services/test_rest_services.py
@@ -24,6 +24,7 @@ from cfme.rest.gen_data import (
 )
 from cfme.services.catalogs.catalog_item import CatalogBundle
 from fixtures.provider import setup_one_or_skip
+from fixtures.pytest_store import store
 from cfme.utils import error, version
 from cfme.utils.blockers import BZ
 from cfme.utils.providers import ProviderFilter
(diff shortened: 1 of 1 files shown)


cfme/tests/services/test_rest_services.py
```diff
@@ -522,7 +523,24 @@ def test_add_service_parent(self, request, appliance):
         parent = collection.action.create(service_body())[0]
         request.addfinalizer(parent.action.delete)
         child = collection.action.create(service_body())[0]
-        child.action.edit(ancestry=str(parent.id))
+        child.action.edit(parent_service=parent._ref_repr())
+        assert_response(appliance)
+        child.reload()
+        assert child.ancestry == str(parent.id)
+
+    @pytest.mark.uncollectif(lambda: store.current_appliance.version < '5.8')
```
rlbabyuk: Use version.current_version() here

cfme/tests/services/test_rest_services.py
```diff
@@ -556,25 +574,26 @@ def _action_and_check(action, resulting_state):
         _action_and_check('suspend', 'suspended')
         _action_and_check('start', 'on')
 
-    @pytest.mark.meta(blockers=[BZ(1441412, forced_streams=['5.8', 'upstream'])])
-    def test_retire_parent_service_now(self, appliance, service_data):
+    @pytest.mark.uncollectif(lambda: store.current_appliance.version < '5.8')
```
rlbabyuk: the same with version.current_version()

cfme/tests/services/test_rest_services.py
mkoura: My understanding is that ``current_version()`` is being deprecated and will slowly go away. ``appliance.version`` is the preferred way now and since it cannot be used in decorator, ``store.current_appliance.version`` is the closest thing.

mkoura: See https://github.com/ManageIQ/integration_tests/pull/5597#discussion_r144657163

rlbabyuk: I'm also not really sure if it is good to use store.current_appliance.version. Maybe better will be get_or_create_current_appliance().version?
@izapolsk @psav what do you think?

PR #5651, opened 2017-10-20
[1LP][RFR] Fixed Ansible playbook catalog item creation

quarckster:
{{pytest: -v -k "test_control_crud_ansible_playbook_action" --long-running}}


cfme/tests/ansible/test_embedded_ansible_basic.py
```diff
@@ -91,7 +91,7 @@ def catalog_item(ansible_repository):
         fauxfactory.gen_alphanumeric(),
         fauxfactory.gen_alphanumeric(),
         provisioning={
-            "repository": ansible_repository,
+            "repository": ansible_repository.name,
```
mshriver: fillable+as_fill_value isn't providing this?

quarckster: It seems no

PR #5663, opened 2017-10-23
[1LP][RFR] DB setup reacts to the output of appliance_console_cli

mfalesni:
Appliance DB setup happily ignores the error messages coming out of appliance_console_cli. This recitfies the issue by looking at the RC and the output of the CLI script. Unfortunately, with the current issue it does not change the RC so I also added a check for "failed" in the output.


dajoRH:
*CFME QE Bot*

PR #5665, opened 2017-10-23
[1LP][WIPTEST] Include certificate when creating provider via REST

jan-zmeskal:
__Extending__ cfme/tests/cloud_infra_common/test_rest_providers.py because it was not possible to set up provider via REST with SSL certificate.

{{pytest: cfme/tests/cloud_infra_common/test_rest_providers.py --use-provider rhv41 -v}}

### DIFF
--- a/cfme/tests/cloud_infra_common/test_rest_providers.py
+++ b/cfme/tests/cloud_infra_common/test_rest_providers.py
@@ -91,8 +91,13 @@ def provider_rest(request, appliance, provider):
     else:
         pytest.skip("No credentials info found for provider {}.".format(provider.name))
 
-    if hasattr(endpoint_default, "verify_tls") and not endpoint_default.verify_tls:
-        default_connection["endpoint"]["verify_ssl"] = 0
+    cert = getattr(endpoint_default, "ca_certs", None)
+    if cert:
+        default_connection["endpoint"]["certificate_authority"] = cert
+        con_config_to_include.append(default_connection)
+
+    if hasattr(endpoint_default, "verify_tls"):
+        default_connection["endpoint"]["verify_ssl"] = 1 if endpoint_default.verify_tls else 0
         con_config_to_include.append(default_connection)
     if hasattr(endpoint_default, "api_port") and endpoint_default.api_port:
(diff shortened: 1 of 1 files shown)


dgur:
Hi Jan, did you run with this fix? Do you have  where to run it?

jan-zmeskal:
@dgur Yes, I verified it on our TLV machine (10.35.71.64)

jan-zmeskal:
@psav Could you please take a look at result of test run against 5.9? It failed but it seems to me that it has not much to do with this PR. Maybe some REST API calls have changed in 5.9? Anyway we do not really run our tests against 5.9 as of now, since we decided to focus our efforts on stabilizing RHV 4.1 - CFME 5.8. However I am still afraid I might break things for other people. What is your opinion on that?

cfme/tests/cloud_infra_common/test_rest_providers.py
```diff
@@ -91,8 +91,13 @@ def provider_rest(request, appliance, provider):
...
 
-    if hasattr(endpoint_default, "verify_tls") and not endpoint_default.verify_tls:
-        default_connection["endpoint"]["verify_ssl"] = 0
+    cert = getattr(endpoint_default, "ca_certs", None)
+    if cert:
+        default_connection["endpoint"]["certificate_authority"] = cert
+        con_config_to_include.append(default_connection)
+
+    if hasattr(endpoint_default, "verify_tls"):
+        default_connection["endpoint"]["verify_ssl"] = 1 if endpoint_default.verify_tls else 0
```
mshriver: **Required**
Use True/False instead of 1/0 if possible here.

mshriver: In fact I'm pretty sure you can just write this as:

```python
default_connection['endpoint']['verify_ssl'] = endpoint_default.verify_tls

mkoura: @mshriver, these are not form values but dictionary representing JSON of REST request. The values needs to be 0 or 1, not True or False.

mshriver: _fine_ ;)

PR #5669, opened 2017-10-23
[1LP][RFR] Fix rc 17.14.0

psav:
Some fixes applied to problems on RC over weekend

### DIFF
--- a/cfme/infrastructure/host.py
+++ b/cfme/infrastructure/host.py
@@ -56,7 +56,7 @@ class Host(BaseEntity, Updateable, Pretty, PolicyProfileAssignable, WidgetasticT
     pretty_attrs = ['name', 'hostname', 'ip_address', 'custom_ident']
 
     name = attr.ib()
-    provider = attr.ib()
+    provider = attr.ib(default=None)
     hostname = attr.ib(default=None)
     ip_address = attr.ib(default=None)
     custom_ident = attr.ib(default=None)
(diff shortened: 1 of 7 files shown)


cfme/infrastructure/host.py
```diff
@@ -56,7 +56,7 @@ class Host(BaseEntity, Updateable, Pretty, PolicyProfileAssignable, WidgetasticT
     pretty_attrs = ['name', 'hostname', 'ip_address', 'custom_ident']
 
     name = attr.ib()
-    provider = attr.ib()
+    provider = attr.ib(default=None)
```
mshriver: A host can have no provider?

cfme/tests/configure/test_ntp_server.py
```diff
@@ -98,7 +98,7 @@ def test_ntp_conf_file_update_check(request, appliance, empty_ntp_dict, ntp_serv
 
 @pytest.mark.tier(3)
 def test_ntp_server_check(request, appliance, ntp_servers_keys):
-    request.addfinalizer(partial(clear_ntp_settings, appliance))
+    request.addfinalizer(lambda: appliance.server.settings.update_ntp_servers(empty_ntp_dict))
```
mshriver: @izapolsk has a fix for the exception this causes.

cfme/infrastructure/host.py
izapolsk: I asked psav  the same yestarday :). Yes it can have no provider.

cfme/tests/configure/test_ntp_server.py
izapolsk: yep

PR #5680, opened 2017-10-24
[RFR] fixes to cli tests to use dbdisk option

lcouzens:
Fix to cli tests that configure database using manditory dbdisk option


lkhomenk:
sorry Luke - I assume the only test that uses dbdisk is test_configure_appliance_internal_fetch_key - failed on 5.7 and 5.9, any ideas why?

jkrocil:
@lkhomenk Leo, you can push PR a state back (from RFR to WIP) when you have a question.

lcouzens:
So the failures are because of the current RHOS templates. This test will only work against rhv templates as they have an additional disk for usage when creating the database.

PR #5681, opened 2017-10-25
[WIPTEST] Update appliance.db enable_internal for dbdata

mshriver:
CFME 5.9 requires a separate disk for setup, which we can pass via
--dbdata to appliance_console_cli

Following is copied from JIRA card notes:
```
Note my PR has several lines commented out that were accessing appliance properties that triggered running ssh or rails commands that would fail without the database setup.
Got to the point where the appliance_console_cli command is running with dbdisk, and I'm seeing the following output, with a 0 exit code!
So if you run appliance.configure or scripts/clone_template.py on my branch, it will fail starting evmserverd, because the db setup call returns 0 after failing.


u'create encryption key
configuring internal database
Initialize postgresql disk starting
Initialize postgresql disk failed with error - /sbin/parted exit code: 1.
MiqSignalError
Failed to configure internal database
TERM environment variable not set.
'
```

### DIFF
--- a/cfme/utils/appliance/db.py
+++ b/cfme/utils/appliance/db.py
@@ -139,16 +139,17 @@ class ApplianceDB(AppliancePlugin):
 
         """
         self.logger.info('Starting DB setup')
-        if self.appliance.is_downstream:
+        # TODO is this check even needed?
+        # if self.appliance.is_downstream:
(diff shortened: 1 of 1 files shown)


mshriver:
@mfalesni pulled these commits into a branch to continue work on this, closing. #5684

PR #5682, opened 2017-10-25
[1LP][RFR] service tests fix

izapolsk:
The issue is caused by Tab click. That click was added in order to resolve the issue when cfme didn't see some changes made in provisioning dialog. 
So, one of ways to fix that issue could be to fix that root issue.

{{pytest: --long-running cfme/tests/services/test_service_catalogs.py cfme/tests/infrastructure/test_provisioning_dialog.py}}

### DIFF
--- a/cfme/common/vm_views.py
+++ b/cfme/common/vm_views.py
@@ -4,7 +4,7 @@ from time import sleep
 
 from widgetastic.widget import View, Text, TextInput, Checkbox, ParametrizedView
 from widgetastic_patternfly import (
-    Dropdown, BootstrapSelect, FlashMessages, Tab, Input, CheckableBootstrapTreeview)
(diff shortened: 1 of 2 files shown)


izapolsk:
@lkhomenk, your thoughts ?

izapolsk:
I've restarted PRT in order to check issues with filling/clearing input fields really disappeared.

mshriver:
some PRT results in for 58z and 59z in run 17257.

jkrocil:
Merging, providing we find a proper solution in then next 4 weeks. If not, we need to revisit this. @izapolsk

lkhomenk:
proper solution would be - provision forms migration to Angular as per developer's comment on this =)

izapolsk:
@lkhomenk, have you heard about ETA ?

PR #5691, opened 2017-10-25
[WIPTEST] Adding different path to Details view for archived VM

jan-zmeskal:
__Extending__ VM model. The reason is that if you want to access VM's details, navmazing path to it contains view with VM of provider. However, archived VM is no longer visible under provider. That's why VM model now first determines whether the VM is archived or not and chooses navmazing path according to that.

{{pytest: cfme/tests/infrastructure/test_vm_power_control.py -k "not test_no_template_power_control" --use-provider rhv41 -v}}

### DIFF
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -305,6 +305,24 @@ class BaseVM(Pretty, Updateable, PolicyProfileAssignable, WidgetasticTaggable,
         else:
             return False
 
+    def get_state_from_quadicon(self):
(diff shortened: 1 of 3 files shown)


dgur:
Are we merging it?

jan-zmeskal:
It was actually merged in this [PR](https://github.com/ManageIQ/integration_tests/pull/5696l). I am sorry, maybe I forgot to update the PR link in our sheet.

PR #5696, opened 2017-10-26
[1LP][RFR] Adding different path to Details view for archived/orphaned VM

jan-zmeskal:
__Extending__ VM model. The reason is that if you want to access VM's details, navmazing path to it contains view with VMs of a provider. However, archived VM is no longer visible under provider. That's why you can now specify that you want to load details independently of provider.

{{pytest: cfme/tests/infrastructure/test_vm_power_control.py::test_no_power_controls_on_archived_vm --use-provider rhv41 -v --long-running}}

### DIFF
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -1222,6 +1222,12 @@ class VmAllWithTemplatesDetails(CFMENavigateStep):
 @navigator.register(Template, 'AnyProviderDetails')
 @navigator.register(Vm, 'AnyProviderDetails')
 class VmAllWithTemplatesDetailsAnyProvider(VmAllWithTemplatesDetails):
+    """
+    Page with details for VM or template.
+    This is required in case you want to get details about archived/orphaned VM/template.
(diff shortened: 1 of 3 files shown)


cfme/infrastructure/virtual_machines.py
```diff
@@ -1219,6 +1219,12 @@ def resetter(self, *args, **kwargs):
         self.view.toolbar.reload.click()
 
 
+@navigator.register(Template, 'AnyProviderDetails')
+@navigator.register(Vm, 'AnyProviderDetails')
+class VmAllWithTemplatesDetailsAnyProvider(VmAllWithTemplatesDetails):
```
izapolsk: I doubt this is good idea. Inheritance will add some restrictions and can cause issues in navmazing.
@psav, can you share your opinion ?

izapolsk: [12:05:22] 18<psav> I think it's OK
[12:05:28] 18<psav> maybe needs a docstring

jan-zmeskal: Docstring added as requested

PR #5705, opened 2017-10-27
[1LP][RFR] storage volume backups model page[ WT and collection]

digitronik:
- Its supporting model page for cfme.storage.volume.py for volume backup operations.
- from version >5.7 separations of backups and snapshot page came.
- backup creation going to be add using volume.py.
- added a simple tag test using available backups only. as create method add in volume.py its possible to add and remove a tag on newly created backup.  [RHCFQE-5459](https://projects.engineering.redhat.com/browse/RHCFQE-5459)

{{pytest: cfme/tests/storage/test_volume_backup.py --use-provider rhos7-ga  -v}}

### DIFF
--- a/cfme/storage/volume.py
+++ b/cfme/storage/volume.py
@@ -9,6 +9,7 @@ from widgetastic_manageiq import (
     BaseQuadIconEntity,
     BaseTileIconEntity,
     BootstrapSelect,
+    BootstrapSwitch,
     BreadCrumb,
     ItemsToolBarViewSelector,
     JSBaseEntity,
@@ -175,6 +176,21 @@ class VolumeEditView(VolumeView):
     save = Button('Save')
 
 
+class VolumeBackupView(VolumeView):
+    @property
+    def is_displayed(self):
+        return False
+
+    backup_name = TextInput(name='backup_name')
+    # options
+    incremental = BootstrapSwitch(name='incremental')
+    force = BootstrapSwitch(name='force')
+
+    save = Button('Save')
+    reset = Button('Reset')
+    cancel = Button('Cancel')
+
+
 @attr.s
 class Volume(BaseEntity):
     # Navigation menu option
@@ -227,6 +243,18 @@ class Volume(BaseEntity):
         self.provider.refresh_provider_relationships()
         self.browser.refresh()
 
+    def create_backup(self, name, incremental=None, force=None):
+        """create backup of cloud volume"""
+        view = navigate_to(self, 'Backup')
+        view.backup_name.fill(name)
+        view.incremental.fill(incremental)
+        view.force.fill(force)
+
+        view.save.click()
+        view.flash.assert_success_message('Backup for Cloud Volume "{}" created'.format(self.name))
+
+        wait_for(lambda: int(self.backups) > 0, delay=20, timeout=1000, fail_func=self.refresh)
+
     @property
     def exists(self):
         try:
@@ -244,6 +272,11 @@ class Volume(BaseEntity):
     def tenant(self):
         view = navigate_to(self, 'Details')
         return view.entities.relationships.get_text_of('Cloud Tenants')
+
+    @property
+    def backups(self):
+        view = navigate_to(self, 'Details')
+        return view.entities.relationships.get_text_of('Cloud Volume Backups')
 
 
 @attr.s
@@ -347,3 +380,13 @@ class VolumeEdit(CFMENavigateStep):
 
     def step(self, *args, **kwargs):
         self.prerequisite_view.toolbar.configuration.item_select('Edit this Cloud Volume')
+
+
+@navigator.register(Volume, 'Backup')
+class VolumeBackup(CFMENavigateStep):
+    VIEW = VolumeBackupView
+    prerequisite = NavigateToSibling('Details')
+
+    def step(self, *args, **kwargs):
+        self.prerequisite_view.toolbar.configuration.item_select('Create a Backup of this Cloud '
+                                                                 'Volume')
--- a/cfme/storage/volume.py
+++ b/cfme/storage/volume.py
@@ -24,7 +24,7 @@ from widgetastic_patternfly import Button, Dropdown, FlashMessages
 from widgetastic.widget import View, Text, ParametrizedView
 
 from cfme.base.ui import BaseLoggedInPage
-from cfme.exceptions import VolumeNotFound, ItemNotFound
+from cfme.exceptions import VolumeNotFoundError, ItemNotFound
 from cfme.utils.appliance.implementations.ui import CFMENavigateStep, navigator, navigate_to
 from cfme.modeling.base import BaseCollection, BaseEntity
 from cfme.utils.log import logger
@@ -253,30 +253,45 @@ class Volume(BaseEntity):
         view.save.click()
         view.flash.assert_success_message('Backup for Cloud Volume "{}" created'.format(self.name))
 
-        wait_for(lambda: int(self.backups) > 0, delay=20, timeout=1000, fail_func=self.refresh)
+        wait_for(lambda: self.backups > 0, delay=20, timeout=1000, fail_func=self.refresh)
 
     @property
     def exists(self):
         try:
             navigate_to(self, 'Details')
             return True
-        except VolumeNotFound:
+        except VolumeNotFoundError:
             return False
 
     @property
     def size(self):
+        """ size of storage cloud volume.
+
+        Returns:
+            :py:class:`int' size of volume.
+        """
         view = navigate_to(self, 'Details')
         return view.entities.properties.get_text_of('Size')
 
     @property
     def tenant(self):
+        """ cloud tenants for volume.
+
+        Returns:
+            :py:class:`str' respective tenants.
+        """
         view = navigate_to(self, 'Details')
         return view.entities.relationships.get_text_of('Cloud Tenants')
 
     @property
     def backups(self):
+        """ number of available backups for volume.
+
+        Returns:
+            :py:class:`int' backup count.
+        """
         view = navigate_to(self, 'Details')
-        return view.entities.relationships.get_text_of('Cloud Volume Backups')
+        return int(view.entities.relationships.get_text_of('Cloud Volume Backups'))
 
 
 @attr.s
@@ -328,7 +343,7 @@ class VolumeCollection(BaseCollection):
                 try:
                     view.entities.get_entity(volume.name).check()
                 except ItemNotFound:
-                    raise VolumeNotFound("Volume {} not found".format(volume.name))
+                    raise VolumeNotFoundError("Volume {} not found".format(volume.name))
 
             view.toolbar.configuration.item_select('Delete selected Cloud Volumes',
                                                    handle_alert=True)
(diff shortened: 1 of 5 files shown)


cfme/storage/volume_backup.py
```diff
@@ -0,0 +1,249 @@
...
+    name = attr.ib()
+    provider = attr.ib()
+
+    def restore(self, name):
+        """Restore the volume backup. this feature included in 5.9 and above.
+
+        Args:
+            name: volume name
+        """
+        view = navigate_to(self, 'restore')
```
izapolsk: REQUIRED: navmazing destination names start from upper letter.

cfme/storage/volume_backup.py
```diff
@@ -0,0 +1,249 @@
...
+                try:
+                    view.entities.get_entity(backup.name).check()
+                except ItemNotFound:
+                    raise BackupNotFound("Volume backup {} not found".format(backup.name))
+
+            view.toolbar.configuration.item_select('Delete selected Backups', handle_alert=True)
+
+            for backup in backups:
+                wait_for(
+                    lambda: backup.name not in view.entities.all_entity_names,
```
izapolsk: it makes sense to check all names together instead of one by one.

cfme/tests/storage/test_volume_backup.py
```diff
@@ -0,0 +1,31 @@
...
+import random
+
+from cfme.cloud.provider.openstack import OpenStackProvider
+from cfme.utils import testgen, version
+
+
+pytestmark = pytest.mark.usefixtures("setup_provider")
+
+
+pytest_generate_tests = testgen.generate([OpenStackProvider], scope="module")
```
izapolsk: REQUIRED: please use pytest.mark.provider instead. you can find many examples in our code.

cfme/tests/storage/test_volume_backup.py
```diff
@@ -0,0 +1,31 @@
...
+    # add tag with category Department and tag communication
+    backup.add_tag('Department', 'Communication')
+    tag_available = backup.get_tags()
+    assert tag_available[0].display_name == 'Communication'
+    assert tag_available[0].category.display_name == 'Department'
+
+    # remove assigned tag
+    backup.remove_tag('Department', 'Communication')
+    tag_available = backup.get_tags()
+    assert tag_available == []
```
izapolsk: REQUIRED: please use len(tags_available) == 0 instead

quarckster:
```python
assert not tag_available
```
will be enough

cfme/storage/volume_backup.py
digitronik: @izapolsy I think still we need compress list using list compression as we are looking for back.name. Please, can you any other idea for comparison?

cfme/tests/storage/test_volume_backup.py
digitronik: yup, it's a good idea.

dajoRH:
Lint report for commit 009faff56b16e81c735382afec3a16d59dd64fc9:

`cfme/tests/storage/test_volume_backup.py`:
- :red_circle: Line 6:1: **F401** *'cfme.utils.testgen' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

cfme/storage/volume_backup.py
izapolsk: I guess you can use something like:
bool(set([backup.name for backup in backups]) & set(view.entities.all_entity_names))

cfme/storage/volume_backup.py
```diff
@@ -0,0 +1,252 @@
...
+
+class VolumeBackupView(BaseLoggedInPage):
+    """A base view for all the Volume Backup pages"""
+    title = Text('.//div[@id="center_div" or @id="main-content"]//h1')
+    flash = FlashMessages(
+        './/div[@id="flash_msg_div"]/div[@id="flash_text_div" or '
+        'contains(@class, "flash_text_div")]')
+
+    @property
+    def in_volume_backup(self):
```
psav: **REQUIRED:** Why is this not ```is_displayed```?

psav: Ahh I see why, we could always use a super in the view below...... @mfalesni what you think?

cfme/storage/volume_backup.py
```diff
@@ -0,0 +1,252 @@
...
+
+@attr.s
+class VolumeBackupCollection(BaseCollection):
+    """Collection object for :py:class:'cfme.storage.volume_backups.VolumeBackup' """
+    ENTITY = VolumeBackup
+
+    def all(self):
+        """returning all backup objects"""
+        view = navigate_to(self, 'All')
+        backups = [self.instantiate(name=item, provider=self.filters.get('provider'))
```
psav: **REQUIRED:** This doesn't filter by provider from what I can see. ```self.filters.get('provider')``` will return to you the provider that the collection is filtered with, but you would then need a conditional on the end of the list comprehension to only supply the backups that are related to that provider.

cfme/tests/storage/test_volume_backup.py
```diff
@@ -0,0 +1,30 @@
...
+              pytest.mark.usefixtures('setup_provider'),
+              pytest.mark.provider([OpenStackProvider], scope='module')]
+
+
+@pytest.mark.tier(3)
+@pytest.mark.uncollectif(lambda: version.current_version() < '5.8')
+def test_storage_volume_backup_edit_tag_from_detail(appliance, provider):
+    collection = appliance.collections.volume_backups.filter({'provider': provider})
+    backups = collection.all()
+    backup = random.choice(backups)
```
psav: **REQUIRED:** This assumes that a backup is already present on the provider. Are these a constant thing, or should we have code that _adds_ a backup. Could make more sense that we also would be checking inventory at the same time. My fear is that this will just fail if there are none available.

digitronik: @psav pointed right, I mention that fear in PR introduction only. I have a card for this RHCFQE-5459. but going add it in this PR only.

dajoRH:
Lint report for commit e4c6c6938a506f5538543cae81e63e927c2793bf:

`cfme/storage/volume.py`:
- :bangbang: Line 256:43: **E225** *missing whitespace around operator*
- :heavy_exclamation_mark: Line 392:75: **W292** *no newline at end of file*

`cfme/tests/storage/test_volume.py`:
- :red_circle: Line 11:1: **F401** *'cfme.utils.log.logger' imported but unused*

`cfme/tests/storage/test_volume_backup.py`:
- :bangbang: Line 28:44: **E251** *unexpected spaces around keyword / parameter equals*

**No commit flaws detected**.

*CFME QE Bot*

cfme/storage/volume_backup.py
mfalesni: @psav We try to avoid super to mitigate is_displayed from multiple-inheritance views which can sometimes happen and bite one's a$$ as happened quite recently, so it is better to create these `in_something` methods and use them in `is_displayed` in the child classes.

mfalesni: But here should be at least is_displayed returning in_volume_backup

digitronik: @mfalesni added 'is_displayed' method returning 'in_volume_backups' but as per my understanding >  we required is_displayed for the repetitive view; this is BaseView going to inherited by other views mainly 'All'. All other main views have their owned 'is_displayed' method. It's just for checking navigation to backups.

digitronik:
@psav, @izapolsk added suggested change.

cfme/storage/volume_backup.py
```diff
@@ -0,0 +1,266 @@
...
+    """Collection object for :py:class:'cfme.storage.volume_backups.VolumeBackup' """
+    ENTITY = VolumeBackup
+
+    def all(self):
+        """returning all backup objects for respective storage manager type"""
+        view = navigate_to(self, 'All')
+
+        backups = [self.instantiate(name=item['Name'], provider=self.filters.get('provider'))
+                   for item in view.entities.elements.read()
+                   if self.filters.get('provider').name in item['Cloud Provider']]
```
izapolsk: REQUIRED: according to pep 8 list comprehensions shouldn't take several lines or be complex.  Please convert this to regular for statement

cfme/storage/volume_backup.py
```diff
@@ -0,0 +1,266 @@
...
+        Args:
+            One or Multiple 'cfme.storage.volume_backup.VolumeBackup' objects
+        """
+
+        view = navigate_to(self, 'All')
+
+        if view.entities.get_all():
+            for backup in backups:
+                try:
+                    view.entities.get_entity(backup.name).check()
```
izapolsk: please note, this statement may get broken after PR #5571. 
you need to use name=backup.name instead after that PR.

cfme/storage/volume_backup.py
```diff
@@ -0,0 +1,266 @@
...
+
+
+@navigator.register(VolumeBackup, 'Details')
+class Details(CFMENavigateStep):
+    VIEW = VolumeBackupDetailsView
+    prerequisite = NavigateToAttribute('parent', 'All')
+
+    def step(self, *args, **kwargs):
+        try:
+            self.prerequisite_view.entities.get_entity(self.obj.name,
```
izapolsk: ^^

cfme/tests/storage/test_volume_backup.py
```diff
@@ -0,0 +1,78 @@
...
+    backup = backup_collection.instantiate(backup_name, provider)
+
+    yield backup
+
+    try:
+        if backup.exists:
+            backup_collection.delete(backup)
+        if volume.exists:
+            volume.delete(wait=False)
+    except Exception:
```
izapolsk: it isn't good idea to use so broad except statement

cfme/storage/volume_backup.py
digitronik: yup, waiting for that to merge; I will too happy 🥇

cfme/storage/volume.py
```diff
@@ -227,24 +243,56 @@ def refresh(self):
...
 
     @property
     def size(self):
+        """ size of storage cloud volume.
+
+        Returns:
+            :py:class:`int' size of volume.
+        """
         view = navigate_to(self, 'Details')
         return view.entities.properties.get_text_of('Size')
```
izapolsk: REQUIRED: you mention int class but still return text

digitronik: my mistake; I don't want touch volume size property. Maybe due to this other test fails. changing doc string to string. actually, it returning '1 GB'

cfme/storage/volume_backup.py
```diff
@@ -0,0 +1,270 @@
...
+
+    @property
+    def size(self):
+        """ size of cloud volume backup.
+
+        Returns:
+            :py:class:`int' size of volume backup in GB.
+        """
+        view = navigate_to(self, 'Details')
+        return int(view.entities.properties.get_text_of('Size')[:-2])
```
izapolsk: OPTIONAL: it should be accepted by int() by I suspect you forgot to remove space

PR #5706, opened 2017-10-27
[1LP][RFR] quadicons/paginator stuff should wait for data load in 5.9

izapolsk:


lkhomenk:
@izapolsk  add some testing for PRT - any test which uses Requests page  would fit

cfme/utils/appliance/implementations/ui.py
```diff
@@ -90,9 +90,11 @@ def ensure_page_safe(self, timeout='10s'):
         # THIS ONE SHOULD ALWAYS USE JAVASCRIPT ONLY, NO OTHER SELENIUM INTERACTION
 
         def _check():
-            result = self.browser.execute_script(self.ENSURE_PAGE_SAFE, silent=True)
```
psav: Can we get this added to cfme/utils/appliance/implementations/ui.py in the ENSURE_PAGE_SAFE function

izapolsk: I have to check appliance version in JS then. Could you advice how I could check appliance version?

izapolsk:
I can't get green results because there is some issue with PRT.

cfme/utils/appliance/implementations/ui.py
mfalesni: @izapolsk you can just check if the thing is undefined, remember that adding more javascript is basically "free" for us. So you would do `if(typeof window.ManageIQ.gtl.loading !== "undefined"){...}`. Or something like that. If we keep it as it is now in the PR, we will get a small slowdown.

PR #5732, opened 2017-10-31
[1LP][RFR]Update clone_template, imports and logging

mshriver:
- log to console using utils.logger.add_stdout_handler
- remove bulk cfme.utils import, only import what's needed for script

### DIFF
--- a/scripts/clone_template.py
+++ b/scripts/clone_template.py
@@ -5,14 +5,16 @@ Where possible, defaults will come from cfme_data"""
 import argparse
 import sys
 
-import cfme.utils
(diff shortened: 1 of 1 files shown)


scripts/clone_template.py
```diff
@@ -86,7 +88,6 @@ def main(**kwargs):
     if kwargs.get('deploy'):
         kwargs['configure'] = True
         kwargs['outfile'] = 'appliance_ip_address_1'
-        provider_data = cfme.utils.conf.provider_data
```
mshriver: This is covered by the direct import now, which without the --deploy option is never accessed and runs fine without the corresponding file.

PR #5736, opened 2017-11-01
[RFR] DB setup does not raise Exception if internal DB exists

jan-zmeskal:


jan-zmeskal:
@mfalesni What do you think about this? I know you have changed this part of script recently and it breaks our automation now. The reason is that we are trying to configure the DB in every Jenkins run and the fact that it is already configured is not really an issue.

dgur:
Please make sure to merge it for Sprint 15 downstream stable as it seems it is blocking us. @mfalesni @psav

PR #5738, opened 2017-11-01
[1LP][RFR] Fixing EditCatalogItemView

jan-zmeskal:
__Fixing__ is_displayed() method of EditCatalogItemView since test_edit_catalog_after_deleting_provider was failing on that.

{{pytest: cfme/tests/services/test_service_catalogs.py::test_edit_catalog_after_deleting_provider --use-provider rhv41 -v --long-running}}

### DIFF
--- a/cfme/services/catalogs/catalog_item.py
+++ b/cfme/services/catalogs/catalog_item.py
@@ -105,8 +105,12 @@ class EditCatalogItemView(BasicInfoForm):
 
     @property
     def is_displayed(self):
-        return self.in_explorer and self.catalog_items.is_opened and \
-            self.title.text == "Editing Service Catalog Item {}".format(self.context['object'].name)
+        return (
+            self.in_explorer and
+            self.catalog_items.is_opened and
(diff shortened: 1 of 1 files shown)


cfme/services/catalogs/catalog_item.py
```diff
@@ -106,7 +106,8 @@ class EditCatalogItemView(BasicInfoForm):
     @property
     def is_displayed(self):
         return self.in_explorer and self.catalog_items.is_opened and \
-            self.title.text == "Editing Service Catalog Item {}".format(self.context['object'].name)
+            self.title.text == 'Editing Service Catalog Item "{}"'.format(
+                self.context['object'].name)
```
quarckster: **REQUIRED**
```python
@property
def is_displayed(self):
    return (
        self.in_explorer and
        self.catalog_items.is_opened and
        self.title.text == 'Editing Service Catalog Item "{}"'.format(self.context['object'].name)
    )
```

jan-zmeskal:
@quarckster Hope I got it right this time :-)

PR #5745, opened 2017-11-02
[1LP][RFR] Image conversion to collection

brahmanim:
{{pytest: cfme/tests/infrastructure/test_host_analysis.py -v  --use-provider vsphere65-nested --long-running }}

**Shriver: Expanded PRT to include a test that also uses is_task_finished because of the change to using regex compiled expressions
cfme/tests/containers/test_containers_smartstate_analysis.py  (removed for now)
**

Image conversion to collection
=================

In this PR @shalomnaim1 and I will convert Container Image object to compatible with collection as preparation for SSA test
Also make some updates to support CFME 5.9 UI changes

### DIFF
--- a/cfme/configure/tasks.py
+++ b/cfme/configure/tasks.py
@@ -2,6 +2,7 @@
 
 """ Module dealing with Configure/Tasks section.
 """
+import re
 from navmazing import NavigateToAttribute
 from widgetastic.utils import Version, VersionPick
 from widgetastic.widget import View
@@ -41,6 +42,12 @@ def delete_all_tasks(destination):
 def is_task_finished(destination, task_name, expected_status, clear_tasks_after_success=True):
     view = navigate_to(Tasks, destination)
     tab_view = getattr(view.tabs, destination.lower())
+
+    # task_name change from str to support also regular expression pattern
+    task_name = re.compile(task_name)
+    # expected_status change from str to support also regular expression pattern
+    expected_status = re.compile(expected_status, re.IGNORECASE)
+
     try:
         row = tab_view.table.row(task_name=task_name, state=expected_status)
     except IndexError:
@@ -128,7 +135,8 @@ class TasksView(BaseLoggedInPage):
 
         @View.nested
         class mytasks(Tab):  # noqa
-            TAB_NAME = "My VM and Container Analysis Tasks"
+            TAB_NAME = VersionPick({Version.lowest(): 'My VM and Container Analysis Tasks',
+                                    '5.9': 'My Tasks'})
             table = Table(table_loc)
 
         @View.nested
--- a/cfme/containers/image.py
+++ b/cfme/containers/image.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
-import random
+import attr
 import itertools
+import random
 from cached_property import cached_property
 
 from navmazing import NavigateToSibling, NavigateToAttribute
@@ -10,10 +11,11 @@ from cfme.common import (WidgetasticTaggable, PolicyProfileAssignable,
                          TagPageView)
 from cfme.containers.provider import (Labelable, navigate_and_get_rows, ContainerObjectAllBaseView,
                                       ContainerObjectDetailsBaseView, LoadDetailsMixin,
-                                      refresh_and_navigate, ContainerObjectDetailsEntities)
+                                      refresh_and_navigate, ContainerObjectDetailsEntities,
+                                      ContainersProvider)
 from cfme.utils.appliance.implementations.ui import CFMENavigateStep, navigator, navigate_to
-from cfme.utils.appliance import Navigatable
 from cfme.configure import tasks
+from cfme.modeling.base import BaseCollection, BaseEntity
 from cfme.utils.wait import wait_for, TimedOutError
 from widgetastic_manageiq import SummaryTable, BaseEntitiesView
 from widgetastic.widget import View
@@ -33,17 +35,16 @@ class ImageDetailsView(ContainerObjectDetailsBaseView):
         compliance = SummaryTable(title='Compliance')
 
 
-class Image(WidgetasticTaggable, Labelable, Navigatable, LoadDetailsMixin, PolicyProfileAssignable):
+@attr.s
+class Image(BaseEntity, WidgetasticTaggable, Labelable, LoadDetailsMixin, PolicyProfileAssignable):
 
     PLURAL = 'Container Images'
     all_view = ImageAllView
     details_view = ImageDetailsView
 
-    def __init__(self, name, image_id, provider, appliance=None):
-        self.name = name
-        self.id = image_id
-        self.provider = provider
-        Navigatable.__init__(self, appliance=appliance)
+    name = attr.ib()
+    id = attr.ib()
+    provider = attr.ib()
 
     @cached_property
     def mgmt(self):
@@ -56,6 +57,9 @@ class Image(WidgetasticTaggable, Labelable, Navigatable, LoadDetailsMixin, Polic
     def perform_smartstate_analysis(self, wait_for_finish=False, timeout='7M'):
         """Performing SmartState Analysis on this Image
         """
+        # task_name change from str to regular expression pattern following Bugzilla Bug 1483590
+        # task name will contain also Image name
+        # the str compile on tasks module
         view = navigate_to(self, 'Details')
         view.toolbar.configuration.item_select('Perform SmartState Analysis', handle_alert=True)
         # TODO: Modify accordingly once there is FlashMessages.assert_massage_contains()
@@ -63,7 +67,7 @@ class Image(WidgetasticTaggable, Labelable, Navigatable, LoadDetailsMixin, Polic
         if wait_for_finish:
             try:
                 wait_for(tasks.is_analysis_finished,
-                         func_kwargs={'task_name': 'Container image analysis',
+                         func_kwargs={'name': '(?i)(Container Image.*)',
                                       'task_type': 'container'},
                          timeout=timeout,
                          fail_func=self.appliance.server.browser.refresh)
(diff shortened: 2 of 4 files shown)


dajoRH:
*CFME QE Bot*

cfme/containers/image.py
```diff
@@ -116,7 +114,33 @@ def compliant(self):
...
+            images.append(self.instantiate(name=name, id=id,
+                provider=ContainersProvider(name=provider_name, appliance=self.appliance)))
+        return images
+
+    def get_random_instances(self, count=1):
+
+        return random.sample(self.all(), count)
+
+
 @navigator.register(Image, 'All')
```
jawatts: Remove this. It should only be registered to the ImageCollection

brahmanim: We spook with Pete, he say that it is OK for now, so we leave it as is.

dajoRH:
*CFME QE Bot*

dajoRH:
Lint report for commit 0aae405279981b63e8fff974cfbccc420833f748:

`cfme/configure/tasks.py`:
- :bangbang: Line 116:5: **E303** *too many blank lines (2)*

**No commit flaws detected**.

*CFME QE Bot*

brahmanim:
PRT failed on ci/downstream-59z , due to BZ 1515775 Perform SSA on container image cause Task view get error 500 Internal Server Error , since that the test could find tasks table.

dajoRH:
*CFME QE Bot*

brahmanim:
All changes have been implement.
Again PRT failed on ci/downstream-59z , due to BZ 1515775 Perform SSA on container image cause Task view get error 500 Internal Server Error , since that the test could find tasks table.

dajoRH:
*CFME QE Bot*

dajoRH:
*CFME QE Bot*

cfme/configure/tasks.py
```diff
@@ -128,7 +135,8 @@ class tabs(View):  # noqa
 
         @View.nested
         class mytasks(Tab):  # noqa
-            TAB_NAME = "My VM and Container Analysis Tasks"
+            TAB_NAME = VersionPick({'5.9': 'My Tasks',
```
mshriver: **Required**

This duplicates the `TAB_NAME` from L144 in `myothertasks` class below.

cfme/configure/tasks.py
```diff
@@ -102,6 +106,12 @@ def delete_all_tasks(destination):
 def is_task_finished(destination, task_name, expected_status, clear_tasks_after_success=True):
     view = navigate_to(Tasks, destination)
     tab_view = getattr(view.tabs, destination.lower())
+
+    # task_name change from str to support also regular expression pattern
+    task_name = re.compile(task_name)
```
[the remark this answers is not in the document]
mshriver: @brahmanim  Looks like only half of this change was implemented, with `{?i)Finished` passed below.

cfme/containers/image.py
mshriver: **Required**

I'd like to revisit the allowance of this during the transition to collections - I think that we need to enforce registering the 'All' destination against the collection only. It is trivial to navigate against the entity's parent in order to get to the 'All' destination.

cfme/containers/image.py
```diff
@@ -122,7 +126,33 @@ def compliant(self):
...
+            image_table.name, image_table.image_ref, ems_table.name).join(
+            ems_table, image_table.ems_id == ems_table.id)
+        images = []
+
+        for name, id, provider_name in image_query.all():
+            images.append(self.instantiate(name=name, id=id,
+                provider=ContainersProvider(name=provider_name, appliance=self.appliance)))
+        return images
+
+    def get_random_instances(self, count=1):
```
mshriver: **Required**

If needed, this should be a test fixture, and not a convenience method on the collection class.

Please move this into the test modules.

cfme/containers/image.py
```diff
@@ -133,23 +163,24 @@ def step(self):
     def resetter(self):
         # Reset view and selection
         self.view.toolbar.view_selector.select("List View")
-        self.view.paginator.reset_selection()
+        self.view.paginator.set_items_per_page(1000)
```
mshriver: **Required/Question**

Is there a reason for increasing this limit every time the page is visited?

cfme/containers/image.py
```diff
@@ -133,23 +163,24 @@ def step(self):
...
+
         self.prerequisite_view.entities.get_entity(provider=self.obj.provider.name,
                                                    id=self.obj.id).click()
 
 
 @navigator.register(Image, 'EditTags')
 class ImageRegistryEditTags(CFMENavigateStep):
     VIEW = TagPageView
-    prerequisite = NavigateToSibling('Details')
+    prerequisite = NavigateToAttribute("parent", 'Details')
```
mshriver: **Required**

This navigation call doesn't look right to me - 'Details' is registered against the `Image` class, as is 'EditTags', making them sibling destinations.  Shouldn't be a need for a change to this prerequisite

setup.py
```diff
@@ -119,6 +119,7 @@
             'object_store_objects = cfme.storage.object_store_object:ObjectStoreObjectCollection',
             ('object_store_containers = '
                 'cfme.storage.object_store_container:ObjectStoreContainerCollection'),
+            'images = cfme.containers.image:ImageCollection',
```
mshriver: **Required**

Use something more specific, like `container_images`, since there are other types of image collections.

cfme/containers/image.py
```diff
@@ -133,23 +163,24 @@ def step(self):
...
 
 
 @navigator.register(Image, 'Details')
 class Details(CFMENavigateStep):
     VIEW = ImageDetailsView
-    prerequisite = NavigateToSibling('All')
+    prerequisite = NavigateToAttribute("parent", 'All')
 
     def step(self):
+
```
mshriver: remove empty line

cfme/containers/image.py
```diff
@@ -122,7 +126,33 @@ def compliant(self):
...
 
 
+@attr.s
+class ImageCollection(BaseCollection):
+    """Collection object for :py:class:`Image`."""
+
+    ENTITY = Image
+
+    def all(self):
+
```
mshriver: remove this empty line or preferably write docblock

cfme/configure/tasks.py
brahmanim: Its the same for 5.9 , but different on Version.lowest !
What is your suggestion here?

cfme/configure/tasks.py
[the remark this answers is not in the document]
brahmanim: As Gilad suggested , I change only the expected_status ("Finished") to ignore case.
On the task_name , i leave it for the caller to decide whether to use ignore case or not.
For example for container images i call this function with '(?i)(Container Image.*)'

cfme/containers/image.py
brahmanim: There was a test fixture on test modules that call this method.
According to your suggestion i move this implementation to test fixture into  test modules.

cfme/containers/image.py
brahmanim: I remove this line according to your suggestion,
Now it registering the 'All' destination against the collection only.

cfme/containers/image.py
brahmanim: We have problem with very slow navigation to Container Image Details ,
According to Pete and Ievgen the problem related to uses a lot of selenium calls because of older Table usage in 5.8 etc .
One of Ievgen suggestion , for try to improve it little beat by set item per page to 1000 , in order to save the moving from page to page in order to get to the right image.

cfme/containers/image.py
brahmanim: I am sorry , Can you please elaborate exactly what are you suggest me to change?
I changed this prerequisite according to Pete suggestion that way it use in collection implementation!

setup.py
brahmanim: Done

cfme/containers/image.py
brahmanim: Done

cfme/containers/image.py
brahmanim: Done

brahmanim:
Please see my reply on each of your comments,
Most of then updated according to you suggestion.
On some of them I need you info\advice.

cfme/configure/tasks.py
mfalesni: **REQUIRED**: A nitpick, but please keep the order oldest-newest, top to bottom.

brahmanim: Hi Milan, what is nitpick?,
I can't understand what you want me to change, 
Since i change it as the other classes like mothertasks and alltasks below are written!
Can you please let me know exactly what should i change?
thank you

jawatts: @mfalesni is asking for:
TAB_NAME = VersionPick({
            Version.lowest(): 'My VM and Container Analysis Tasks',
            '5.9': 'My Tasks'
        })

brahmanim:
Milan i change this , flip the version.lowest with 5.9, thank you!

cfme/containers/image.py
jawatts: @brahmanim: EditTags is on the same object as 'Details'. So the navigate step should be prerequeisite = NavigateToSibiling('Details'). This is what @mshriver  means by sibling destinations. What you have here is equivalent to navigate_to(self.parent,'Details').

brahmanim: Thank you! I change it back to prerequisite = NavigateToSibling('Details')

cfme/configure/tasks.py
brahmanim: Thank you! I change it to oldest-newest, top to bottom

cfme/containers/image.py
mshriver: There is a 'use_search' boolean for `entities.get_entity` now, but at the moment it only supports name search, which doesn't look like it will work for details.

I think you should move this setting into the Details navigation step, above L169, so that the limit is only changed when actually navigating to 'Details', and 'All' navigation isn't slowed by this.

cfme/tests/containers/test_containers_smartstate_analysis.py
```diff
@@ -43,15 +43,18 @@
     )
 )
 
+NUM_SELECTED_IMAGES = 1
```
mshriver: :+1:

cfme/tests/containers/test_containers_smartstate_analysis.py
```diff
@@ -43,15 +43,18 @@
...
 
 @pytest.fixture(scope='function')
 def delete_all_container_tasks():
     delete_all_tasks('AllTasks')
 
 
 @pytest.fixture(scope='function')
-def random_image_instance(provider, appliance):
-    return Image.get_random_instances(provider, 1, appliance).pop()
+def random_image_instance(appliance):
```
mshriver: :+1:

mshriver:
Changes to container images for collection look fine.

The change to tasks concerns me, and requires far more testing. I've pulled the container tests out of PRT to only run test_host_analysis, I couldn't get test_instance_analysis to collect (which looks like a problem with that test's generation).

cfme/configure/tasks.py
```diff
@@ -128,7 +135,8 @@ class tabs(View):  # noqa
 
         @View.nested
         class mytasks(Tab):  # noqa
-            TAB_NAME = "My VM and Container Analysis Tasks"
+            TAB_NAME = VersionPick({Version.lowest(): 'My VM and Container Analysis Tasks',
```
mshriver: On second consideration, until this module is refactored appropriately to get rid of the module methods, this is actually allow-able and covers tests/framework where `mytasks` is used instead of `myothertasks`, where in 59z they both should direct to 'My Tasks' named tab.

Issue #5754, opened 2017-11-02
Save/Reset widgets in server_settings.ServerAuthenticationView


mshriver:
The `cfme.configure.configuration.server_settings.ServerAuthenticationView` class includes two widgets for `save` and `reset`, using the button text 'Save' and 'Reset'.

The DOM for the Authentication page actually includes two complete sets of buttons for save/reset, using the same text in them:

```html
<div id="buttons_on" style="display: none;">
<button name="button" type="submit" class="btn btn-primary" alt="Save Changes" title="Save Changes" onclick="miqAjaxButton('/ops/settings_update/authentication?button=save', false);">Save</button>
<button name="button" type="submit" class="btn btn-default" alt="Reset Changes" title="Reset Changes" onclick="miqAjaxButton('/ops/settings_update/authentication?button=reset');">Reset</button>
</div>
```

What is happening is that on a fill of this form, changed is set to true even when the form isn't updated (I suspect because of the obfuscated password field).

The save/reset buttons aren't enabled in this case, and it tries to click Save and fails.

I believe we will need a new widget wrapper for Button that adds XPATH for `div id='buttons_on'` and/or `div id='buttons_off'` as a parent element of the actual buttons to correctly handle this.

The disabled class is only added to the buttons_off set of buttons, and is never added to the buttons_on set.

PR #5755, opened 2017-11-02
[1LP][RFR] Update auth fixtures, views, widgets, and tests

mshriver:
Related to #5754, and includes a hack to work around it.

# REVIEWERS PLEASE NOTE

There will be test failures in PRT, I'm expecting them and am engaging with dev. I will either write many specific RBAC bugs, or start to lump some of the permission mismatches into 2 BZ's that were written while writing these tests.

PRT failures in run 19290 will all be BZ's.

### DIFF
--- a/cfme/roles.py
+++ b/cfme/roles.py
@@ -279,3 +279,312 @@ group_data = {
         'about'
     ]
 }
+
+# Matches structure/string format of VerticalNavigation output for tree, not UI access control tree
+# TODO include non-vertical nav RBAC to settings, help
+# TODO RBAC goes deeper than veritcal nav, into accordions. example cloud intel -> Reports
+role_access_ui_59z = {
+    'evmgroup-super_administrator': {
+        'Automation': {
+            'Ansible': ['Credentials', 'Repositories', 'Playbooks'],
+            'Ansible Tower': ['Jobs', 'Explorer'],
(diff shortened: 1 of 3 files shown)


dajoRH:
*CFME QE Bot*

dajoRH:
*CFME QE Bot*

cfme/roles.py
```diff
@@ -279,3 +279,179 @@ def _remove_from_all(roles, r_page):
         'about'
     ]
 }
+
+# Matches structure/string format of VerticalNavigation output for tree, not UI access control tree
+# TODO include non-vertical nav RBAC to settings, help
+# TODO RBAC goes deeper than veritcal nav, into accordions. example cloud intel -> Reports
```
mshriver: This should probably be in a yaml file, I'd prefer to leave that change to a follow-up PR.

mshriver:
I'm removing the logic to modify the role_access data structure based on version within the test. The conditionals are unweildy, and have to take into account Z-stream versions, nested dictionaries, and differences between users. Ultimately I think it's a better implementation to have a separate `role_access` data structure for each z-stream version to avoid confusion and have a more maintainable test - at the cost of very similar data structures with small differences between them.

mshriver:
@psav thoughts on merging this automation re-write of the test? I can skip just based on the 2 BZ's I've written so far, but I'll be writing more BZ's tomorrow morning.

PR #5761, opened 2017-11-03
[1LP][RFR] Added Lenovo Provider base views etc

psav:
Mostly everything is there for the Lenovo provider to be created now, but the string_name is broken in the latest build and I don't want to screw with create method unless I have to

### DIFF
--- a/cfme/common/provider_views.py
+++ b/cfme/common/provider_views.py
@@ -365,6 +365,18 @@ class MiddlewareProvidersView(ProvidersView):
                 self.entities.title.text == 'Middleware Providers')
 
 
+class PhysicalProvidersView(ProvidersView):
+    """
+     represents Main view displaying all Infra providers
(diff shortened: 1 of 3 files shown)


cfme/common/provider_views.py
```diff
@@ -325,6 +325,18 @@ def is_displayed(self):
                 self.entities.title.text == 'Middleware Providers')
 
 
+class PhysicalProvidersView(ProvidersView):
+    """
+     represents Main view displaying all Infra providers
+    """
+    @property
+    def is_displayed(self):
+        return (super(PhysicalProvidersView, self).is_displayed and
```
mfalesni: I was gonna eek here on the super, but since we have a single inheritance, I can close one eye :)

dajoRH:
*CFME QE Bot*

PR #5768, opened 2017-11-06
[RFR] Snapshot: remove soft asserts

apagac:
**WAS**: Fix wait_for_vm_state_change for 5.9

~~This is a quick fix to enable `wait_for_vm_state_change` usage for 5.9. There was a change in dictionary that is returned from `find_quadicon().data`, in 5.9 we must be looking for key 'quadicon' instead of 'state' which is not there any more.~~

~~`find_quadicon().data['quadicon']` returns a rather nasty and long line of html code, which thankfully contains also 'currentstate-*', so there is no need to change anything else but the item key.~~

Also removed all occurrences of `find_quadicon` from snapshot tests, since it is exactly what `wait_for_vm_state_change` does, so `find_quadicon` doesn't bring any added value to the test.

{{pytest: -v --long-running --use-provider vsphere65-nested -k "test_operations_suspended_vm"}}

### DIFF
--- a/cfme/tests/infrastructure/test_snapshot.py
+++ b/cfme/tests/infrastructure/test_snapshot.py
@@ -184,9 +184,6 @@ def test_verify_revert_snapshot(full_test_vm, provider, soft_assert, register_ev
     full_test_vm.wait_for_vm_state_change(desired_state=full_test_vm.STATE_OFF, timeout=720)
     full_test_vm.power_control_from_cfme(option=full_test_vm.POWER_ON, cancel=False)
     full_test_vm.wait_for_vm_state_change(desired_state=full_test_vm.STATE_ON, timeout=900)
-    current_state = full_test_vm.find_quadicon().data['state']
-    soft_assert(current_state.startswith('currentstate-on'),
-                "Quadicon state is {}".format(current_state))
     soft_assert(full_test_vm.provider.mgmt.is_vm_running(full_test_vm.name), "vm not running")
     wait_for(lambda: ssh_client.run_command('test -e snapshot1.txt').rc == 0,
              num_sec=400, delay=20, handle_exception=True, fail_func=ssh_client.close())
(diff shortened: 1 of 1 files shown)


apagac:
Upstream failure is unrelated to changes introduced by this PR and is fixed in PR #5767 .

cfme/common/vm.py
```diff
@@ -641,8 +641,11 @@ def _looking_for_state_change():
                 self.load_details(refresh=True)
                 return self.get_detail(properties=detail_t) == desired_state
             else:
+                key = version.pick({
```
rlbabyuk: **Required** Use self.appliance.version here instead

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -184,9 +184,6 @@ def test_verify_revert_snapshot(full_test_vm, provider, soft_assert, register_ev
     full_test_vm.wait_for_vm_state_change(desired_state=full_test_vm.STATE_OFF, timeout=720)
     full_test_vm.power_control_from_cfme(option=full_test_vm.POWER_ON, cancel=False)
     full_test_vm.wait_for_vm_state_change(desired_state=full_test_vm.STATE_ON, timeout=900)
-    current_state = full_test_vm.find_quadicon().data['state']
-    soft_assert(current_state.startswith('currentstate-on'),
```
rlbabyuk: **Question**: Shouldn't we check if the vm state is really changed?

cfme/common/vm.py
apagac: This fix is no longer needed, "state" will work with 5.9.0.9.
I used version.pick() because it is used across the file and I wanted  to be consistent. self.appliance.version is not used in this file.

cfme/tests/infrastructure/test_snapshot.py
apagac: wait_for_vm_state_change() does the checking. If the state is anything other than what we expect, the function fails at wait_for(). It makes no sense to have the soft asserts here, since it will always be True, and if wait_for_vm_state_change() fails, it will never get to them.

apagac:
I need to remove the wait_for_vm_state_change() fix and make this PR just delete the soft asserts which are not needed.

apagac:
The current failure on 5.9 is due to a known UI bug.

PR #5774, opened 2017-11-06
[1LP][RFR] Service Dialogs now use new UI which is not modeled in FW

psav:
Service Dialogs now use new UI which is not modeled in FW

{{pytest: cfme/tests/services/test_service_catalogs.py}}

### DIFF
--- a/cfme/automate/service_dialogs.py
+++ b/cfme/automate/service_dialogs.py
@@ -143,6 +143,10 @@ class Add(CFMENavigateStep):
     prerequisite = NavigateToSibling('All')
 
     def step(self):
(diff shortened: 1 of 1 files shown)


cfme/automate/service_dialogs.py
```diff
@@ -143,6 +143,10 @@ class Add(CFMENavigateStep):
     prerequisite = NavigateToSibling('All')
 
     def step(self):
+        if self.obj.appliance.version > '5.9':
+            raise Exception(
```
mshriver: `NotImplemented` ?

PR #5781, opened 2017-11-07
[RFR][1LP] New miq upgrade script

lcouzens:
New script to allow people to quickly upgrade a single appliance for additional testing.

### DIFF
--- a/cfme/scripting/appliance.py
+++ b/cfme/scripting/appliance.py
@@ -33,18 +33,22 @@ def main():
     pass
 
 
-@main.command('upgrade', help='Upgrades an appliance to new version')
+@main.command('upgrade', help='Upgrades an appliance to latest Z-stream')
 @click.argument('appliance-ip', default=None, required=False)
 @click.option('--cfme-only', is_flag=True, help='Upgrade cfme packages only')
-@click.option('--update-to', default='59', help='Must be 58 or (59 is default)')
+@click.option('--update-to', default='5.9.z', help='Must be 5.8.z or (5.9.z is default)')
 def upgrade_appliance(appliance_ip, cfme_only, update_to):
     """Upgrades an appliance"""
+    supported_version_repo_map = {'5.8.z': 'update_url_58', '5.9.z': 'update_url_59'}
+    assert update_to in supported_version_repo_map, "{} is not a supported version".format(
+        update_to
+    )
+    update_url = supported_version_repo_map[update_to]
     if appliance_ip:
         print('Connecting to {}'.format(appliance_ip))
     else:
         print('Fetching appliance from env.local.yaml')
     app = get_appliance(appliance_ip)
-    update_url = 'update_url_{}'.format(update_to)
     print('Extending appliance partitions')
     app.db.extend_partition()
     urls = process_url(cfme_data['basic_info'][update_url])
--- a/cfme/scripting/appliance.py
+++ b/cfme/scripting/appliance.py
@@ -8,9 +8,13 @@ Usage:
    scripts/encrypt_conf.py credentials
 """
 import click
+import tempfile
 from cached_property import cached_property
+from cfme.utils import os
+from cfme.utils.conf import cfme_data
 from functools import partial
 from .setup_ansible import setup_ansible
+from scripts.repo_gen import process_url, build_file
 
 
 def get_appliance(appliance_ip):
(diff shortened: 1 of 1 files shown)


cfme/scripting/appliance.py
```diff
@@ -29,6 +33,57 @@ def main():
...
 
 
+@main.command('upgrade', help='Upgrades an appliance to new version')
+@click.argument('appliance-ip', default=None, required=False)
+@click.option('--cfme-only', is_flag=True, help='Upgrade cfme packages only')
+@click.option('--update-to', default='59', help='Must be 58 or (59 is default)')
+def upgrade_appliance(appliance_ip, cfme_only, update_to):
+    """Upgrades an appliance"""
+    if appliance_ip:
+        print 'Connecting to {}'.format(appliance_ip)
```
quarckster: **REQUIRED**
Here and below. Please make it python3 compatible:
```python
print('Connecting to {}'.format(appliance_ip))
```

cfme/scripting/appliance.py
```diff
@@ -29,6 +33,57 @@ def main():
...
+        os.fsync(f.fileno())
+        app.ssh_client.put_file(
+            f.name, '/etc/yum.repos.d/update.repo')
+    ver = '95' if app.version >= '5.8' else '94'
+    cfme = '-y'
+    if cfme_only:
+        cfme = 'cfme -y'
+    print 'Stopping EVM'
+    app.evmserverd.stop()
+    with app.ssh_client as ssh:
```
quarckster: **REQUIRED**
Context manager is not needed for `ssh_client`. Just use `app.ssh_client.run_command`.

cfme/scripting/appliance.py
```diff
@@ -29,6 +33,57 @@ def main():
...
+        print 'Running yum update'
+        rc, out = ssh.run_command('yum update {}'.format(cfme), timeout=3600)
+        assert rc == 0, "update failed {}".format(out)
```
quarckster: **REQUIRED**
You can extend `run_command` function from [here](https://github.com/ManageIQ/integration_tests/blob/master/cfme/scripting/setup_ansible.py#L43) and use one line instead of three ones.

cfme/scripting/appliance.py
```diff
@@ -29,6 +33,56 @@ def main():
     pass
 
 
+@main.command('upgrade', help='Upgrades an appliance to new version')
+@click.argument('appliance-ip', default=None, required=False)
+@click.option('--cfme-only', is_flag=True, help='Upgrade cfme packages only')
+@click.option('--update-to', default='59', help='Must be 58 or (59 is default)')
```
jkrocil: **REQUIRED** This will have to accept version, not just '58' or '59' string. The moment you show this to the public, they will be asking `I have this BZ that requires DB migrated to 5.8.3.1, how do I do that?` which means we will have to support that.
Let's start by changing these `58` and `59` strings to `5.8.z` and `5.9.z` so that it's clear we update to latest of that stream.

Next step (future PR) would be to be able to `--list-supported-versions` to get a list of versions the given appliance (based on it's version) can be updated to. This also depends on our yamls supporting that by having proper repo files available, not just the latest builds.

cfme/scripting/appliance.py
```diff
@@ -29,6 +33,56 @@ def main():
...
+        'db:migrate:status 2>/dev/null | grep "^\s*down"', timeout=30)
+    assert rc != 0, "Migration failed; migrations in 'down' state found: {}".format(out)
+    print('Restarting postgres service')
+    rc, out = app.ssh_client.run_command('systemctl restart rh-postgresql{}-postgresql'.format(ver))
+    assert rc == 0, "Failed to restart postgres: {}".format(out)
+    print('Starting EVM')
+    app.start_evm_service()
+    print('Waiting for webui')
+    app.wait_for_web_ui()
+    print('Appliance upgrade completed')
```
jkrocil: We should have this as a method of appliance. All of this, with the ability to reuse it in tests as well (whatever shared portion is valid for tests as well).

jkrocil: Again, future PR. ^

mfalesni: Also for future, the references to streams and postgres versions should come directly from the yamls, so there is no need to change the code unless there are some process changes.

PR #5796, opened 2017-11-07
[1LP][RFR] test_clear_datastore_filter_results fix

izapolsk:
{{pytest: cfme/tests/infrastructure/test_set_default_filter.py::test_clear_datastore_filter_results}}

### DIFF
--- a/cfme/tests/infrastructure/test_set_default_filter.py
+++ b/cfme/tests/infrastructure/test_set_default_filter.py
@@ -73,7 +73,7 @@ def test_clear_datastore_filter_results(appliance):
     """ Test for clearing filter results for datastores."""
     dc = DatastoreCollection(appliance)
     view = navigate_to(dc, 'All')
-    view.sidebar.datastores.tree.click_path('Datastores', 'All Datastores', 'Global Filters',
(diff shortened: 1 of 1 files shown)


izapolsk:
My change fixed the issue.
However,  PRT fails because this test has to be redesigned. I think this has to be done by test owner.

cfme/tests/infrastructure/test_set_default_filter.py
```diff
@@ -73,7 +73,7 @@ def test_clear_datastore_filter_results(appliance):
     """ Test for clearing filter results for datastores."""
     dc = DatastoreCollection(appliance)
     view = navigate_to(dc, 'All')
-    view.sidebar.datastores.tree.click_path('Datastores', 'All Datastores', 'Global Filters',
+    view.sidebar.datastores.tree.click_path('All Datastores', 'Global Filters',
```
quarckster: Is it correct for all versions of CFME?

izapolsk: yes, this is bug created during web_ui->widgetastic conversion

mshriver:
@izapolsk You're correct, test makes it past this call now but still fails. FA owner should revise test.

PR #5802, opened 2017-11-08
[1LP][RFR] Fixing test_edit_catalog_after_deleting_provider with WA

jan-zmeskal:
__Fixing__ test_edit_catalog_after_deleting_provider with this workaround after consulting with @mfalesni. The reason is that framework sometimes attempted to click "Save" button on Catalog Item edit sceen before it was actually clickable.

{{pytest: cfme/tests/services/test_service_catalogs.py::test_edit_catalog_after_deleting_provider --use-provider rhv41 --long-running -vv}}

### DIFF
--- a/cfme/services/catalogs/catalog_item.py
+++ b/cfme/services/catalogs/catalog_item.py
@@ -14,6 +14,7 @@ from cfme.utils.pretty import Pretty
 from cfme.utils.appliance import Navigatable
 from cfme.utils.appliance.implementations.ui import navigator, CFMENavigateStep, navigate_to
 from cfme.utils import version
+from cfme.utils.wait import wait_for
 from . import ServicesCatalogView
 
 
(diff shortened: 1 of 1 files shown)


jan-zmeskal:
@mfalesni I tried to run it four times in sequence. Three times it was successful but once it time out. Which is quite strange since 10 seconds should be all the time in the world for this purpose. 
[Jenkins output](https://rhv-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/cfme-5.8-rhv-4.1-integration-test-jan/48/console)

mfalesni:
@jan-zmeskal more leverage to ramble at the UI developers :) though 3 out of 4 is a better score than before ...

jan-zmeskal:
@mfalesni Are you sure this is UI thing? If it is the case, I can open BZ and wrap the test. However I tried it manually like 10 times and the Save button always became active in like 1 second.

jan-zmeskal:
@mfalesni Please advise what to do with this PR. Since I am not sure about the real cause of the test failure (although it definitely looks like an UI issue), I don't know if I should push it to [RFR]

mfalesni:
@jan-zmeskal It can be caused by automation behaving a bit differently than the real user, moving mouse differently, therefore some JS can be suppressed and similar things ...

jan-zmeskal:
@izapolsk I think you can merge. On 5.9 it fails because of UI change that is known and @rhrazdil is working on fix

PR #5805, opened 2017-11-08
[1LP][RFR] Adding Support for Alerts Endpoint and Settings for OpenShift Provider

jawatts:
Adding Support for

Alerts:
New dropdown widget called "Alerts", that when selected, creates a new Endpoint nested tab.

Settings:
A new form called "Settings" with two nested tabs, "Proxy" and "Advanced", was added to the "Add New Containers Provider" view. 

{{ pytest: --use-provider cm-env2 --long-running  cfme/tests/containers/test_ssl_providers.py }}

### DIFF
--- a/cfme/common/provider_views.py
+++ b/cfme/common/provider_views.py
@@ -1,9 +1,10 @@
 # -*- coding: utf-8 -*-
 from lxml.html import document_fromstring
+
 from widgetastic.exceptions import NoSuchElementException
 from widgetastic.utils import VersionPick, Version
 from widgetastic.widget import View, Text, ConditionalSwitchableView, ParametrizedView
-from widgetastic_patternfly import Dropdown, BootstrapSelect
+from widgetastic_patternfly import Dropdown, BootstrapSelect, Tab
 
 from cfme.base.login import BaseLoggedInPage
 from cfme.common.host_views import HostEntitiesView
@@ -469,6 +470,30 @@ class CloudProviderAddView(ProviderAddView):
                 self.title.text == 'Add New Cloud Provider')
 
 
+class ContainerProviderSettingView(ProvidersView):
+    """
+       Settings view for builds 5.9 and up
+      """
+
+    @View.nested
+    class proxy(Tab, BeforeFillMixin):  # NOQA
+
+        TAB_NAME = 'Proxy'
+        http_proxy = Input('provider_options_proxy_settings_http_proxy')
+
+    @View.nested
+    class advanced(Tab, BeforeFillMixin):  # NOQA
+
+        TAB_NAME = 'Advanced'
+        adv_http = Input('provider_options_image_inspector_options_http_proxy')
+        adv_https = Input('provider_options_image_inspector_options_https_proxy')
+        no_proxy = Input('provider_options_image_inspector_options_no_proxy')
+        image_repo = Input('provider_options_image_inspector_options_repository')
+        image_reg = Input('provider_options_image_inspector_options_registry')
+        image_tag = Input('provider_options_image_inspector_options_image_tag')
+        cve_loc = Input('provider_options_image_inspector_options_cve_url')
+
+
 class ContainerProviderAddView(ProviderAddView):
     """
      represents Container Provider Add View
@@ -482,13 +507,18 @@ class ContainerProviderAddView(ProviderAddView):
                 self.title.text == 'Add New Containers Provider')
 
 
-class ContainerProviderAddViewUpdated(ContainerProviderAddView):
+class ContainerProviderAddViewUpdated(ContainerProviderAddView, ContainerProviderSettingView):
     """
      Additional widgets for builds 5.9 and up
     """
+    COND_WIDGETS = ['prov_type', 'metrics_type', 'alerts_type']
 
     metrics_type = BootstrapSelect(id='metrics_selection')
     alerts_type = BootstrapSelect(id='alerts_selection')
+
+    def before_fill(self, values):
+        for widget in self.COND_WIDGETS:
+            getattr(self, widget).fill(values.get(widget))
 
 
 class MiddlewareProviderAddView(ProviderAddView):
@@ -567,13 +597,19 @@ class ContainerProviderEditView(ProviderEditView):
                 'Edit Containers Provider' in self.title.text)
 
 
-class ContainerProviderEditViewUpdated(ContainerProviderEditView):
+class ContainerProviderEditViewUpdated(ContainerProviderEditView, ContainerProviderSettingView):
     """
      Additional widgets for builds 5.9 and up
     """
 
+    COND_WIDGETS = ['prov_type', 'metrics_type', 'alerts_type']
+
     metrics_type = BootstrapSelect(id='metrics_selection')
     alerts_type = BootstrapSelect(id='alerts_selection')
+
+    def before_fill(self, values):
+        for widget in self.COND_WIDGETS:
+            getattr(self, widget).fill(values.get(widget))
 
 
 class MiddlewareProviderEditView(ProviderEditView):
--- a/cfme/containers/provider/__init__.py
+++ b/cfme/containers/provider/__init__.py
@@ -48,8 +48,9 @@ class ContainersProviderDefaultEndpoint(DefaultEndpoint):
         if self.sec_protocol.lower() == 'ssl trusting custom ca' and hasattr(self, 'get_ca_cert'):
             out['trusted_ca_certificates'] = self.get_ca_cert()
 
-        if version.current_version() < '5.9':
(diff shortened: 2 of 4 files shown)


cfme/containers/provider/openshift.py
```diff
@@ -63,11 +83,45 @@ class OpenshiftProvider(ContainersProvider):
...
+            adv_https=None,
+            no_proxy=None,
+            image_repo=None,
+            image_reg=None,
+            image_tag=None,
+            cve_loc=None):
+
+        gt_59 = True if self.appliance.version >= '5.9' else False
+
+        self.http_proxy = http_proxy if gt_59 else None
```
jaryn: I don't like that setting of those attributes depends on which appliance version is being used. It just isn't a behaviour I would expect. Usually the initialiser function (__init__) doesn't contain such logic. It ususaly just is blindly setting the state of the object.

jaryn: What more, instead of just ignoring the arg, you should at least log some warning or raise exception. Silent ignoration is often bad for people trying to debug.

jawatts: Dicussing with @mshriver where to move it to. It use to be in from_config but I was asked to move it from there.

jawatts: Moved to view_value_mapping

jawatts:
@gshefer I just want to put this on your radar for next sprint

jawatts:
Test failures are not caused by these changes. Tests need to be updated (out of scope for this PR)

cfme/common/provider_views.py
```diff
@@ -554,6 +573,24 @@ class ContainerProviderEditViewUpdated(ContainerProviderEditView):
     metrics_type = BootstrapSelect(id='metrics_selection')
     alerts_type = BootstrapSelect(id='alerts_selection')
 
+    @View.nested
+    class proxy(Tab, BeforeFillMixin):  # NOQA
```
gshefer: __OPTIONAL__: Instead of duplicating ```proxy``` and `advanced` (in `ContainerProviderAddViewUpdated` and `MiddlewareProviderAddView`) - why not creating some `ProviderAddViewSettings` add inherit it?

jawatts: Done

jawatts:
@gshefer Requested change has been made

cfme/containers/provider/__init__.py
```diff
@@ -187,17 +201,26 @@ def __init__(
...
             'prov_type': self.type,
-            'zone': self.zone,
-            'metrics_type': self.metrics_type
+            'zone': self.zone
         }
 
+        if self.appliance.version >= '5.9':
+            mapping['metrics_type'] = self.metrics_type
+            mapping['alerts_type'] = self.alerts_type
+        else:
```
jaryn: This is not too important, but why not to initialize the fiels to `None`, so no `else` would be needed

```
        mapping = {
             'name': self.name,
            ...
            'metrics_type': None
            'alerts_type': None
            ...
            'zone': self.zone
         }
```

cfme/containers/provider/openshift.py
```diff
@@ -55,6 +55,26 @@ def view_value_mapping(self):
...
+    name = 'alerts'
+
+    @property
+    def view_value_mapping(self):
+        if current_version() > '5.9':
+            out = {
+                'hostname': self.hostname,
+                'api_port': self.api_port
+            }
+            if current_version() >= '5.8':
```
jaryn: Why is this `if` nested? It is bit confusing to my brain.

I'm not sure what is the logic here but something is wrong - if you visit this line `current_version() >= '5.8'` is True for sure since `current_version() > '5.9'`. It looks like a mistake - please fix this

cfme/containers/provider/openshift.py
```diff
@@ -120,18 +192,38 @@ def from_config(cls, prov_config, prov_key, appliance=None):
...
                 raise Exception('Unsupported endpoint type "{}".'.format(endp))
 
+        http_proxy = prov_config.get('settings', {}).get('proxy', {}).get('http_proxy')
+        adv_http = prov_config.get('settings', {}).get('advanced', {}).get('adv_http')
+        adv_https = prov_config.get('settings', {}).get('advanced', {}).get('adv_https')
+        no_proxy = prov_config.get('settings', {}).get('advanced', {}).get('no_proxy')
+        image_repo = prov_config.get('settings', {}).get('advanced', {}).get('image_repo')
+        image_reg = prov_config.get('settings', {}).get('advanced', {}).get('image_reg')
+        image_tag = prov_config.get('settings', {}).get('advanced', {}).get('image_tag')
+        cve_loc = prov_config.get('settings', {}).get('advanced', {}).get('cve_loc')
```
gshefer: __OPTIONAL__
Change all the above 8 lines to:
```python
settings = prov_config.get('settings', {})
advanced = settings.get('advanced', {})
http_proxy = settings.get('proxy', {}).get('http_proxy')
adv_https, adv_https, no_proxy, image_repo, image_reg, image_tag, cve_loc = [
    advanced.get(field) for field in
    ('adv_https', 'adv_https', 'no_proxy', 'image_repo', 'image_reg', 'image_tag', 'cve_loc')
]
```
Note that this is just my approach for lines count reduction but probably it's too complicated and the current is good enough :-) up to you.

cfme/containers/provider/openshift.py
jawatts: Fixed, good catch

cfme/containers/provider/openshift.py
jawatts: :+1: This is a great suggestion

dajoRH:
*CFME QE Bot*

cfme/containers/provider/__init__.py
jawatts: I removed this from  __init__.  Metrics and Alters should be in the OpenShiftProvider init

cfme/containers/provider/openshift.py
```diff
@@ -55,6 +55,25 @@ def view_value_mapping(self):
         return out
 
 
+class AlertsEndpoint(DefaultEndpoint):
+    """Represents Alerts Endpoint"""
+    name = 'alerts'
+
+    @property
+    def view_value_mapping(self):
+        if current_version() > '5.9':
```
`current_version()` is deprecated, please use `appliance.version` or `self.browser.product_version` in views.

jawatts: Done

cfme/common/provider_views.py
```diff
@@ -464,14 +489,19 @@ def is_displayed(self):
...
      Additional widgets for builds 5.9 and up
     """
+    COND_WIDGETS = ['prov_type', 'metrics_type', 'alerts_type']
 
     metrics_type = BootstrapSelect(id='metrics_selection')
     alerts_type = BootstrapSelect(id='alerts_selection')
 
+    def before_fill(self, values):
+        for widget in self.COND_WIDGETS:
+            getattr(self, widget).fill(values.get(widget))
```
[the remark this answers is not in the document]
jawatts: So what was happening was the Setting form was being filled before the selection of the provider type. Without the provider type selected, the Settings form does not appear on the page. So here, I select all the types before filling out the rest of the form.

cfme/containers/provider/__init__.py
```diff
@@ -48,7 +48,7 @@ def view_value_mapping(self):
         if self.sec_protocol.lower() == 'ssl trusting custom ca' and hasattr(self, 'get_ca_cert'):
             out['trusted_ca_certificates'] = self.get_ca_cert()
 
-        if version.current_version() < '5.9':
+        if current_appliance.version < '5.9':
```
jawatts: @quarckster I removed the current_version() call here as well

cfme/containers/provider/openshift.py
```diff
@@ -48,10 +48,39 @@ def view_value_mapping(self):
...
             'api_port': self.api_port
         }
-        if current_version() >= '5.8':
-            out['sec_protocol'] = self.sec_protocol
-            if self.sec_protocol.lower() == 'ssl trusting custom ca':
-                out['trusted_ca_certificates'] = OpenshiftDefaultEndpoint.get_ca_cert()
+
+        out['sec_protocol'] = version.pick({
+            version.LOWEST: None,
+            '5.8': self.sec_protocol})
```
5.7 is not supported anymore
```python
out['sec_protocol'] = self.sec_protocol
```

jawatts: Fixed

mshriver:
Please go through PRT results and triage in description or update with fixes.

dajoRH:
*CFME QE Bot*

jawatts:
Test are failures are not due to the changes in this PR

PR #5806, opened 2017-11-08
[RFR] Update template_upload_*, use yaml block_upload

mshriver:
No framework impacts, PRT irrelevant.


mshriver:
Merging, I've tested all of these locally and want to get them into CI ASAP.

PR #5811, opened 2017-11-09
[RFR] Tag objects refactoring

rlbabyuk:
Purpose or Intent
=================
Fix/refactor tag objects tests


cfme/cloud/tenant.py
```diff
@@ -49,7 +49,7 @@ class relationships(Accordion):     # noqa
 
 class TenantEntities(BaseEntitiesView):
     """The entities on the main list page"""
-    table = Table('//div[@id="list_grid"]//table')
+    table = Table('//div[@id="gtl_div"]//table')
```
quarckster: **REQURIED**
What about 5.8? Is it changed there as well.

cfme/tests/cloud_infra_common/test_tag_objects.py
```diff
@@ -1,128 +1,156 @@
...
-from cfme.infrastructure.cluster import ClusterCollection
-from cfme.infrastructure.datastore import DatastoreCollection
-from cfme.infrastructure.host import HostCollection
+from cfme.cloud.instance.image import Image
+from cfme.cloud.keypairs import KeyPair
 from cfme.infrastructure.provider import InfraProvider
 from cfme.infrastructure.virtual_machines import Vm, Template
-from cfme.web_ui import mixins
-from cfme.modeling.base import BaseCollection
+from fixtures.provider import setup_one_or_skip
```
quarckster: **REQUIRED**
Please sort imports alphabetically.

cfme/tests/cloud_infra_common/test_tag_objects.py
```diff
@@ -1,128 +1,156 @@
...
-        * Select ``Policy/Edit Tags`` and assign the tag to it.
-        * Click on the quadicon and verify the tag is assigned. (TODO)
-        * Go back to the quadicon view and select ``Policy/Edit Tags`` and remove the tag.
-        * Click on the quadicon and verify the tag is not present. (TODO)
-    """
-    _navigate_and_check(location, appliance)
+@pytest.fixture(scope='module')
+def infra_provider(request):
+    prov_filter = ProviderFilter(classes=[InfraProvider])
+    return setup_one_or_skip(request, filters=[prov_filter])
```
quarckster: **REQUIRED**
This fixture is not needed. You can use `@pytest.mark.provider` and `provider` fixture.

cfme/tests/cloud_infra_common/test_tag_objects.py
```diff
@@ -1,128 +1,156 @@
...
-        mixins.remove_tag(tag)
-    request.addfinalizer(lambda: diaper(_delete))
-    mixins.add_tag(tag)
-    _delete()
 
+@pytest.fixture(scope='module')
+def cloud_provider(request):
+    prov_filter = ProviderFilter(classes=[CloudProvider],
+                                 required_fields=[['provisioning', 'stacks']])
+    return setup_one_or_skip(request, filters=[prov_filter])
```
quarckster: **REQUIRED**
Here as well

cfme/cloud/tenant.py
rlbabyuk: This locator covers all 3 versions

cfme/tests/cloud_infra_common/test_tag_objects.py
rlbabyuk: I've restored cloud/infra provider fixtures, I think it is better to use them here, as it selects any provider, not just OpenStack or Vmware as I set before, second point if use pytest.mark.provider it adds provider info to name tests case, I would like to omit that(will be easier to track in polarion), one more thing in my case while running test cases to many time is taken by provider add/delete(for every second test case), test run lasts 1h longer.  

I would like to leave local cloud/infra provider fixtures if such usage is not deprecated, however, if there is another better way to do that I'll be happy to add updates
@quarckster  what do think?

quarckster: I'm ok with that, but I also would like to know what @mshriver, @jkrocil  think about it.

cfme/cloud/availability_zone.py
```diff
@@ -32,7 +33,7 @@ class AvailabilityZoneDetailsToolBar(View):
     view_selector = View.nested(ItemsToolBarViewSelector)
 
 
-class AvailabilityZoneEntities(View):
+class AvailabilityZoneEntities(BaseEntitiesView):
     """View containing the widgets for the main content pane"""
     title = Text('//div[@id="main-content"]//h1')
     table = Table("//div[@id='gtl_div']//table")
```
mshriver: I think you should likely expand the scope of the refactor a bit here.

If using BaseEntitiesView, there shouldn't be a need for these table or title widgets (unless they're unique).

This means re-writing the navigation step below on L138 to use the entity functions and used `surf_pages`.

cfme/common/__init__.py
```diff
@@ -258,6 +265,26 @@ def step(self):
...
+        self.prerequisite_view.entities.get_entity(surf_pages=True, name=self.obj.name).check()
+        self.prerequisite_view.toolbar.policy.item_select('Edit Tags')
+
+
+@navigator.register(WidgetasticTaggable, 'EditTagsCollection')
+class EditTagsFromListCollection(CFMENavigateStep):
+    VIEW = TagPageView
+    prerequisite = NavigateToAttribute('parent', 'All')
+
+    def step(self):
```
mshriver: **Question**

Thought about this design for a while. I do kind of like registering the nav destinations against the inheritable class.
I think that we need to have a kwarg passed to this destination's step though, a list of entities in the collection.

I'm considering a single navigation class, with a prerequisite that uses `isinstance(self.obj, BaseCollection)` to select the prerequisite call.  `step()` taking `**kwargs` would allow for optionally passing a list of entities to check.

cfme/tests/cloud_infra_common/test_tag_objects.py
jkrocil: Yes, this is fine. It could only be an issue if one of the fixtures was trying to set itself up while the other was still being used (since it would remove the other provider), but that is not the case here.

rlbabyuk:
I've created separate PR for BaseEntitiesView updates. https://github.com/ManageIQ/integration_tests/pull/5908

dajoRH:
Lint report for commit ebdd7675c7beadbe8f5503400a728d74e4b42b11:

`cfme/common/__init__.py`:
- :red_circle: Line 12:1: **F401** *'cfme.modeling.base.BaseCollection' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

dajoRH:
Lint report for commit db950d32a81b51afffb00c28aef626fbcc2b374a:

`cfme/common/__init__.py`:
- :red_circle: Line 2:1: **F401** *'navmazing.NavigateToAttribute' imported but unused*
- :red_circle: Line 11:1: **F401** *'cfme.modeling.base.BaseEntity' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

dajoRH:
Lint report for commit 6a9e99ccef5c15d7c77929988806f339e09499ff:

`cfme/fixtures/tag.py`:
- :heavy_exclamation_mark: Line 125:34: **W292** *no newline at end of file*

**No commit flaws detected**.

*CFME QE Bot*

rlbabyuk:
Some of 5.9 test can fail due to https://bugzilla.redhat.com/show_bug.cgi?id=1524933, and didn't block the tests as this issue will be fixed in next 5.9 build

PR #5820, opened 2017-11-10
[RFR] Mark both test_provider_crud methods as tier(1)

mshriver:
These tests should be the same tier, infra was already set to 1

CFME QE wants to include these in smoke testing as well, so add smoke mark.


dajoRH:
*CFME QE Bot*

PR #5834, opened 2017-11-13
[RFR] Hack had invalid logic and invalid bz_blocker

psav:
The logic was always returning False for the BZ, and should have been NOT'ing the blocks

{{pytest: cfme/tests/infrastructure/test_provider_discovery.py -k test_discover_infra}}

### DIFF
--- a/cfme/common/provider.py
+++ b/cfme/common/provider.py
@@ -3,7 +3,7 @@ from collections import Iterable
 
 from manageiq_client.api import APIException
 from widgetastic.widget import View, Text
-from widgetastic_patternfly import Input, Button
+from widgetastic_patternfly import Input, Button, DropdownItemDisabled
 
 from cfme.base.credential import (
     Credential, EventsCredential, TokenCredential, SSHCredential, CANDUCredential)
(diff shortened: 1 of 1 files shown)


psav:
This needs to be validated for 5.9 especially

cfme/common/provider.py
```diff
@@ -618,7 +618,8 @@ def clear_providers(cls):
         from cfme.utils.appliance import current_appliance as app
         app.rest_api.collections.providers.reload()
         # cfme 5.9 doesn't allow to remove provider thru api
-        if app.version < '5.9' or (app.version >= '5.9' and BZ(1501941).blocks):
+        bz_blocked = BZ(1501941, forced_streams=['5.9']).blocks
+        if app.version < '5.9' or (app.version >= '5.9' and not bz_blocked):
```
izapolsk: yeah, my bad. I slightly updated if statement after testing code.

izapolsk:
@psav, providers aren't removed in 5.8 and 5.7. Maybe provider.all issue ?

psav:
This is a much better piece of code now :)

izapolsk:
@psav, I agree it looks better. However, I added that logic with grouping provider types in order to get rid of navmazing delays.

PR #5833, opened 2017-11-13
[1LP][RFR] Added test test_custom_button_ansible_credential_list

quarckster:
{{pytest: -v -k "test_custom_button_ansible_credential_list" --long-running}}

PR #5837, opened 2017-11-14
[1LP][RFR] updating locator for 5.9

lkhomenk:
updating locator for Requests table
{{pytest:: cfme/tests/infrastructure/test_provisioning.py -k test_provision_ --long-running}}

### DIFF
--- a/cfme/services/requests.py
+++ b/cfme/services/requests.py
@@ -247,7 +247,7 @@ class RequestBasicView(BaseLoggedInPage):
 
 
 class RequestsView(RequestBasicView):
-    table = Table(locator='//*[@id="list_grid"]/table')
+    table = Table(locator='//div[@id="gtl_div"]//table')
     paginator = PaginationPane()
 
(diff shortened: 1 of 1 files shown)


cfme/services/requests.py
```diff
@@ -247,7 +247,7 @@ def in_requests(self):
 
 
 class RequestsView(RequestBasicView):
-    table = Table(locator='//*[@id="list_grid"]/table')
+    table = Table(locator='//div[@id="gtl_div"]//table')
```
quarckster: **REQUIRED**
Fixed here https://github.com/ManageIQ/integration_tests/pull/5833/files#diff-cbcd422f5444a1c4d39c88ca3f3aa7bfR251

lkhomenk: but this locator covers both situations  - no version picking and it's better to have it separately - it's a fix not related to other code @quarckster

quarckster: Cool, the locator indeed is more versatile. Thanks.

PR #5838, opened 2017-11-14
[1LP][RFR] adding forgotten name property to quadicons

izapolsk:
{{ pytest: --long-running -v --use-provider rhv41 -k test_infra_cluster_event }}


izapolsk:
errors for 5.9/upstream aren't related this fix

PR #5839, opened 2017-11-14
[RFR] Fix template_upload_ec2, block_upload check

mshriver:
This check was failing if template_upload wasn't in yaml, which was not the intent.


mshriver:
Tested locally and in CI, merging.

PR #5842, opened 2017-11-14
[1LP][RFR] Adds Generic Objects collection and CRUD test

mkoura:
and minor enhancements related to Generic Object Definitions

{{pytest: -v cfme/tests/generic_objects/test_instances.py cfme/tests/generic_objects/test_definitions.py}}

### DIFF
--- a/cfme/generic_objects/instance/__init__.py
+++ b/cfme/generic_objects/instance/__init__.py
@@ -9,8 +9,8 @@ from cfme.utils.update import Updateable
 
 
 @attr.s
-class GenericObjectDefinition(BaseEntity, Updateable, sentaku.modeling.ElementMixin):
-    """Generic Objects Definition class to context switch between UI and REST.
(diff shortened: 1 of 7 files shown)


cfme/generic_objects/instance/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+    delete = sentaku.ContextualMethod()
+    exists = sentaku.ContextualProperty()
+
+    name = attr.ib(default=None)
+    definition = attr.ib(default=None)  # generic object definition
+    attributes = attr.ib(default=None)  # e.g. {'address': 'Test Address'}
+    associations = attr.ib(default=None)  # e.g. {'services': [myservice1, myservice2]}
+
+    def __attrs_post_init__(self):
+        self.rest_response = None
```
psav: **REQUIRED:** Why do we need to use ```__attrs_post_init__``` here? If you don't want it to be in the constructor, attrs has an option for that

PR #5844, opened 2017-11-15
[RFR] Kill smtp_collector process correctly on rhel-based systems

jkrocil:
@lkhomenk Please test this locally, if it works for you as well.

{{pytest: cfme/tests/control/test_alerts.py -k test_alert_vm_turned_on_more_than_twice_in_past_15_minutes --use-provider vsphere65-nested --long-running}}


jkrocil:
Tested locally on fedora, works.

PR #5850, opened 2017-11-15
[1LP][RFR] added simple entity name caching mechanizm

izapolsk:
It turned out that some quadicons cannot provide their name when not displayed after their primary key has been replaced from name to id.
These changes fix this issue because name should be always available.


sbulage:
@izapolsk I tested it locally working fine for me.

PR #5851, opened 2017-11-15
[1LP][RFR] Add WT pages for cloud.provider Instance/Image rel

mshriver:
create views for the cloud provider relationship pages for instances and
images.

Navigation destinations already existed but were using web_ui and
pytest_selenium, this removes that use

FIXES RHCFQE-5777

I've tested navigation and is_displayed locally with success.

### DIFF
--- a/cfme/cloud/provider/__init__.py
+++ b/cfme/cloud/provider/__init__.py
@@ -44,8 +44,8 @@ class CloudProviderInstancesView(BaseLoggedInPage):
 
 class CloudProviderImagesToolbar(View):
     """
-        Toolbar view for cloud provider images relationships
-        """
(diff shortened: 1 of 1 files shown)


cfme/cloud/provider/__init__.py
```diff
@@ -30,6 +27,47 @@ def is_displayed(self):
...
+
+    breadcrumb = BreadCrumb()
+    toolbar = View.nested(VMToolbar)
+    including_entities = View.include(VMEntities, use_parent=True)
+
+
+class CloudProviderImagesToolbar(View):
+    """
+        Toolbar view for cloud provider images relationships
+        """
```
quarckster: **REQUIRED**
Extra indent

PR #5856, opened 2017-11-16
[WIP] add initial physical_server views

rodneyhbrown7:
Adds the initial physical server views for testing server pages in the UI.

PR #5868, opened 2017-11-17
[1LP][RFR] Adds physical server views

rodneyhbrown7:
Add the views for testing the physical server provider functions of the Lenovo Provider for ManageIQ
Replaces  https://github.com/ManageIQ/integration_tests/pull/5856

### DIFF
--- a/cfme/common/physical_server_views.py
+++ b/cfme/common/physical_server_views.py
@@ -42,7 +42,7 @@ class ComputePhysicalInfrastructureServersView(BaseLoggedInPage):
     def in_compute_physical_infrastructure_servers(self):
         def _server_page(title):
             return self.navigation.currently_selected == ["Compute", "Physical Infrastructure",
-                title]
+                "Servers"]
 
         return (
             self.logged_in_as_current_user and (_server_page("Physical Servers"))
--- a/cfme/common/physical_server_views.py
+++ b/cfme/common/physical_server_views.py
@@ -1,0 +1,195 @@
+# -*- coding: utf-8 -*-
+from lxml.html import document_fromstring
+from widgetastic.utils import (
+    Parameter,
+    Version,
+    VersionPick
+)
+from widgetastic.widget import ParametrizedView, Text, View
+from widgetastic_manageiq import (
+    BaseEntitiesView,
+    NonJSBaseEntity,
+    JSBaseEntity,
+    BaseListEntity,
+    BaseQuadIconEntity,
+    BaseTileIconEntity,
+    BootstrapTreeview,
+    BreadCrumb,
+    Button,
+    ItemsToolBarViewSelector,
+    PaginationPane,
+    SummaryTable,
+    TimelinesView,
+    BaseNonInteractiveEntitiesView
+)
+from widgetastic_patternfly import (
+    Dropdown,
+    FlashMessages,
+)
+
+from cfme.base.login import BaseLoggedInPage
+
+
+class ComputePhysicalInfrastructureServersView(BaseLoggedInPage):
+    """Common parts for server views."""
+    title = Text('.//div[@id="center_div" or @id="main-content"]//h1')
+    flash = FlashMessages(
+        './/div[@id="flash_msg_div"]/div[@id="flash_text_div" or '
+        'contains(@class, "flash_text_div")]'
+    )
+
+    @property
+    def in_compute_physical_infrastructure_servers(self):
+        def _server_page(title):
+            return self.navigation.currently_selected == ["Compute", "Physical Infrastructure",
+                title]
+
+        return (
+            self.logged_in_as_current_user and (_server_page("Physical Servers"))
+        )
+
+
+class PhysicalServerQuadIconEntity(BaseQuadIconEntity):
+    @property
+    def data(self):
+        return {
+            'no_host': int(self.browser.text(self.QUADRANT.format(pos="a"))),
+            'state': self.browser.get_attribute("style", self.QUADRANT.format(pos="b")),
+            'vendor': self.browser.get_attribute("alt", self.QUADRANT.format(pos="c")),
+            'creds': self.browser.get_attribute("alt", self.QUADRANT.format(pos="d"))
+        }
+
+
+class PhysicalServerTileIconEntity(BaseTileIconEntity):
+    quad_icon = ParametrizedView.nested(PhysicalServerQuadIconEntity)
+
+
+class PhysicalServerListEntity(BaseListEntity):
+    pass
+
+
+class NonJSPhysicalServerEntity(NonJSBaseEntity):
+    quad_entity = PhysicalServerQuadIconEntity
+    list_entity = PhysicalServerListEntity
+    tile_entity = PhysicalServerTileIconEntity
+
+
+class JSPhysicalServerEntity(JSBaseEntity):
+    @property
+    def data(self):
+        data_dict = super(JSPhysicalServerEntity, self).data
+        if 'quadicon' in data_dict and data_dict['quadicon']:
+            quad_data = document_fromstring(data_dict['quadicon'])
+            data_dict['no_host'] = int(quad_data.xpath(self.QUADRANT.format(pos="a"))[0].text)
+            data_dict['state'] = quad_data.xpath(self.QUADRANT.format(pos="b"))[0].get('style')
+            data_dict['vendor'] = quad_data.xpath(self.QUADRANT.format(pos="c"))[0].get('alt')
+            data_dict['creds'] = quad_data.xpath(self.QUADRANT.format(pos="d"))[0].get('alt')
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit 64a2e5a41231f23865374e8cb8a0bdcc5e95f262:

- :bangbang: Line 182:1: **E302** *expected 2 blank lines, found 1*
- :bangbang: Line 287:1: **E302** *expected 2 blank lines, found 0*
- :bangbang: Line 553:90: **E231** *missing whitespace after ','*
- :bangbang: Line 553:101: **E501** *line too long (106 > 100 characters)*

`cfme/physical/provider/__init__.py`:
- :red_circle: Line 3:1: **F401** *'navmazing.NavigateToObject' imported but unused*
- :red_circle: Line 8:1: **F401** *'cfme.common.provider_views.PhysicalProvidersView' imported but unused*
- :bangbang: Line 111:101: **E501** *line too long (101 > 100 characters)*

Hmmm


**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

dajoRH:
Lint report for commit a39fb538368e25f6e3a12380a8e94085bd0b61a8:

- :bangbang: Line 190:1: **E302** *expected 2 blank lines, found 1*
- :bangbang: Line 295:1: **E302** *expected 2 blank lines, found 0*

**No commit flaws detected**.

*CFME QE Bot*

skovic:
@psav @mshriver Could you take a look at this PR when you get a chance? Thanks

cfme/common/physical_server_views.py
```diff
@@ -0,0 +1,197 @@
...
+    FlashMessages,
+)
+
+from cfme.base.login import BaseLoggedInPage
+
+
+class ComputePhysicalInfrastructureServersView(BaseLoggedInPage):
+    """Common parts for server views."""
+    title = Text('.//div[@id="center_div" or @id="main-content"]//h1')
+    flash = FlashMessages(
```
mshriver: flash is included in BaseLoggedInPage, and doesn't need to be redefined here.  Don't have to change any calls for it, widgets in inherited views are 'flat'.

cfme/common/physical_server_views.py
```diff
@@ -0,0 +1,197 @@
...
+        './/div[@id="flash_msg_div"]/div[@id="flash_text_div" or '
+        'contains(@class, "flash_text_div")]'
+    )
+
+    @property
+    def in_compute_physical_infrastructure_servers(self):
+        def _server_page(title):
```
mshriver: title isn't used within the method, and I'd prefer to not have a method here if possible, following the style convention of the below example.

There are cases where some of the checks are version dependent, or require more complex statements than fit within the context of a return statement. These we pull out into a check before the return statement (saved and included within return), but a local method implementation would be good too.

In this case I don't think its needed, and adds unnecessary complexity.

```python
return(
    self.logged_in_as_current_user and
    self.navigation.currently_selected == ['Compute',
                                                                'Physical Infrastructure',
                                                                'Servers'])
```

Here's an example from an open PR of mine where the title check is dependant on a BZ, and I pulled the if/else block out of the return statement:
https://github.com/ManageIQ/integration_tests/pull/6354/files#diff-682d6837dd0d81e338a6e7a93b13eb94R35

dajoRH:
Lint report for commit f6b86a2cdd4bd0b285cd69385be9395b72940dc1:

`cfme/common/physical_server_views.py`:
- :red_circle: Line 25:1: **F401** *'widgetastic_patternfly.FlashMessages' imported but unused*
- :bangbang: Line 40:101: **E501** *line too long (104 > 100 characters)*

**No commit flaws detected**.

*CFME QE Bot*

skovic:
@mshriver I believe that I have addressed all of your comments. Please let me know if I missed anything. Thanks

cfme/common/physical_server_views.py
```diff
@@ -0,0 +1,188 @@
...
+    def is_displayed(self):
+        return (self.in_compute_physical_infrastructure_servers and
+                super(TimelinesView, self).is_displayed)
+
+
+class PhysicalServerManagePoliciesView(BaseLoggedInPage):
+    """PhysicalServer's Manage Policies view."""
+    policies = BootstrapTreeview("protectbox")
+    entities = View.nested(BaseNonInteractiveEntitiesView)
+    save_button = Button("Save")
```
mshriver: **Required**

We're trying to get rid of implementations with `_button` suffix, one of those cases where some code is a good example and some isn't. :)

Just `save`, `reset`, `cancel` names for these button widgets, please.

cfme/common/physical_server_views.py
```diff
@@ -0,0 +1,188 @@
...
+    lifecycle = Dropdown(text="Lifecycle")
+    monitoring = Dropdown(text="Monitoring")
+    power = Dropdown(text="Power")
+    identify = Dropdown(text="Identify")
+    view_selector = View.nested(ItemsToolBarViewSelector)
+
+
+class PhysicalServerSideBar(View):
+    """Represents left side bar. It usually contains navigation, filters, etc."""
+    pass
```
mshriver: **Required**

Since this is empty, I think I'd prefer it left out of the views for now, with a #TODO to add a sidebar with an accordion for the filter - or just add the accordion widget now from widgetastic.patternfly.

cfme/common/physical_server_views.py
```diff
@@ -0,0 +1,188 @@
...
+    """Represents the view with different items like hosts."""
+    @property
+    def entity_class(self):
+        return PhysicalServerEntity().pick(self.browser.product_version)
+
+
+class PhysicalServersView(ComputePhysicalInfrastructureServersView):
+    toolbar = View.nested(PhysicalServersToolbar)
+    sidebar = View.nested(PhysicalServerSideBar)
+    paginator = PaginationPane()
```
mshriver: So the views here are highly nested and conditional when using BaseEntitiesView, and it actually already includes the widget definitions for paginator and title - so you can drop this widget definition, it just has to be accessed through `view.entities.paginator`.

https://github.com/ManageIQ/integration_tests/blob/4addd2881a759c29abcab6f1d0f799d9f56bc94e/widgetastic_manageiq/__init__.py#L2899

cfme/common/physical_server_views.py
```diff
@@ -0,0 +1,188 @@
...
+    including_entities = View.include(PhysicalServerEntitiesView, use_parent=True)
+
+    @property
+    def is_displayed(self):
+        return (self.in_compute_physical_infrastructure_servers and
+                self.title.text in "Physical Servers")
```
mshriver: **Required**

This title check should be an `==` comparison, since filtered views also include the 'Physical Servers' string in their title.
Adding the `Filters` accordion widget to the sidebar view (previous comment) would allow for an additional check that the `ALL (Default)` option is selected and the view is unfiltered.

cfme/common/physical_server_views.py
skovic: @mshriver Just making sure that I understand. Do you mean that the _PhysicalServerSideBar_ class should be derived from _Accordion_ instead of _View_?  Maybe similar to the _filters_ class in _cfme/common/host_views.py_ ?

mshriver: Ah sorry for confusion - no, the sidebar should still inherit from View - but it should nest accordion views inside of it. In this case just the one filter accordion, on most pages there are a few accordions within the sidebar. 

Example:
In cfme.cloud.keypairs, there is a `KeyPairDetailsAccordion` class inheriting `View`, and that class is `View.nested()` in `KeyPairDetailsView` as the `sidebar` widget:
https://github.com/ManageIQ/integration_tests/blob/master/cfme/cloud/keypairs.py#L34
https://github.com/ManageIQ/integration_tests/blob/master/cfme/cloud/keypairs.py#L97

You can directly nest the view with a decorator, instead of creating a separate class, if you'd like. Since its only a single accordion it would be short.

dajoRH:
Lint report for commit d6fec0735ce92cda694915d487d4832707a5464e:

`cfme/common/physical_server_views.py`:
- :red_circle: Line 9:1: **F401** *'widgetastic_manageiq.PaginationPane' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

cfme/common/physical_server_views.py
```diff
@@ -0,0 +1,191 @@
...
+    power = Dropdown(text="Power")
+    identify = Dropdown(text="Identify")
+    view_selector = View.nested(ItemsToolBarViewSelector)
+
+
+class PhysicalServerSideBar(View):
+    """Represents left side bar. It usually contains navigation, filters, etc."""
+
+    @View.nested
+    class filters(Accordion): # noqa
```
mshriver: nice! :+1:

PR #5871, opened 2017-11-17
[1LP][RFR] data property shouldn't raise exception  when quads are disabled

izapolsk:
It turned out that quads can be disabled for QuadIcons. Quadicon.data doesn't work in such case. This fix inhibits exceptions and return empty dict for such case


dajoRH:
Lint report for commit 6fb6f0e51087be397c451fc1017a8bb3f5c80a17:

`cfme/common/host_views.py`:
- :red_circle: Line 99:28: **F841** *local variable 'TypeError' is assigned to but never used*

- :red_circle: Line 70:28: **F841** *local variable 'TypeError' is assigned to but never used*

**No commit flaws detected**.

*CFME QE Bot*

PR #5872, opened 2017-11-17
[1LP][RFR]PXE fixes for 5.9

lkhomenk:
fixes for PXE tests in 5.9.
{{pytest: cfme/tests/infrastructure/test_customization_template.py cfme/tests/infrastructure/test_pxe.py cfme/tests/infrastructure/test_pxe_provisioning.py  cfme/tests/services/test_pxe_service_catalogs.py  --long-running --use-provider vsphere65-nested}}

### DIFF
--- a/cfme/infrastructure/pxe.py
+++ b/cfme/infrastructure/pxe.py
@@ -501,18 +501,10 @@ class CustomizationTemplate(Updateable, Pretty, BaseEntity):
         view = navigate_to(self, 'Edit')
         view.fill(updates)
         main_view = self.create_view(PXECustomizationTemplatesView, override=updates)
-        name = updates.get('name') or self.name
-        description = updates.get('description') or self.description
-
         if cancel:
             view.cancel.click()
-            # msg = 'Edit of Customization Template "{}" was cancelled by the user'.format(name)
         else:
             view.save.click()
-            # if self.appliance.version < 5.9:
-            #     msg = 'Customization Template "{}" was saved'.format(name)
-            # else:
-            #     msg = 'Customization Template "{}" was saved'.format(description)
         main_view.flash.assert_no_error()
 
 
@@ -547,13 +539,8 @@ class CustomizationTemplateCollection(BaseCollection):
         main_view = self.create_view(PXECustomizationTemplatesView)
         if cancel:
             view.cancel.click()
-            # msg = 'Add of new Customization Template was cancelled by the user'
         else:
             view.add.click()
-            # if self.appliance.version < 5.9:
-            #     msg = 'Customization Template "{}" was saved'.format(name)
-            # else:
-            #     msg = 'Customization Template "{}" was saved'.format(description)
         main_view.flash.assert_no_error()
         return customization_templates
 
@@ -572,8 +559,7 @@ class CustomizationTemplateCollection(BaseCollection):
                                                    handle_alert=not cancel)
             if not cancel:
                 main_view = ct_obj.create_view(PXECustomizationTemplatesView)
-                msg = 'Customization Template "{}": Delete successful'.format(ct_obj.description)
-                main_view.flash.assert_success_message(msg)
+                main_view.flash.assert_no_error()
             else:
                 navigate_to(ct_obj, 'Details')
 
--- a/cfme/infrastructure/pxe.py
+++ b/cfme/infrastructure/pxe.py
@@ -530,18 +530,19 @@ class CustomizationTemplateCollection(BaseCollection):
 
         customization_templates = self.instantiate(name, description, script_data,
                                                    image_type, script_type)
-        view = navigate_to(self, 'Add')
-        view.fill({'name': name,
-                   'description': description,
-                   'image_type': image_type,
-                   'type': script_type,
-                   'script': script_data})
-        main_view = self.create_view(PXECustomizationTemplatesView)
-        if cancel:
-            view.cancel.click()
-        else:
-            view.add.click()
-        main_view.flash.assert_no_error()
+        if not customization_templates.exists():
+            view = navigate_to(self, 'Add')
+            view.fill({'name': name,
+                       'description': description,
+                       'image_type': image_type,
+                       'type': script_type,
+                       'script': script_data})
+            main_view = self.create_view(PXECustomizationTemplatesView)
+            if cancel:
+                view.cancel.click()
+            else:
+                view.add.click()
+            main_view.flash.assert_no_error()
         return customization_templates
 
     def delete(self, cancel=False, *ct_objs):
@@ -964,7 +965,7 @@ class PXEMainPage(CFMENavigateStep):
 
 def get_template_from_config(template_config_name):
     """
-    Convenience function to grab the details for a template from the yamls.
+    Convenience function to grab the details for a template from the yamls and create template.
     """
 
     template_config = conf.cfme_data.get('customization_templates', {})[template_config_name]
@@ -975,11 +976,11 @@ def get_template_from_config(template_config_name):
     script_data = script_data.read()
     appliance = get_or_create_current_appliance()
     collection = appliance.collections.customization_templates
-    return collection.instantiate(name=template_config['name'],
-                                  description=template_config['description'],
-                                  image_type=template_config['image_type'],
-                                  script_type=template_config['script_type'],
-                                  script_data=script_data)
+    return collection.create(name=template_config['name'],
+                             description=template_config['description'],
+                             image_type=template_config['image_type'],
+                             script_type=template_config['script_type'],
+                             script_data=script_data)
 
 
 def get_pxe_server_from_config(pxe_config_name):
--- a/cfme/infrastructure/pxe.py
+++ b/cfme/infrastructure/pxe.py
@@ -506,14 +506,14 @@ class CustomizationTemplate(Updateable, Pretty, BaseEntity):
 
         if cancel:
             view.cancel.click()
-            msg = 'Edit of Customization Template "{}" was cancelled by the user'.format(name)
+            # msg = 'Edit of Customization Template "{}" was cancelled by the user'.format(name)
         else:
             view.save.click()
-            if self.appliance.version < 5.9:
-                msg = 'Customization Template "{}" was saved'.format(name)
-            else:
-                msg = 'Customization Template "{}" was saved'.format(description)
-        main_view.flash.assert_success_message(msg)
+            # if self.appliance.version < 5.9:
+            #     msg = 'Customization Template "{}" was saved'.format(name)
+            # else:
+            #     msg = 'Customization Template "{}" was saved'.format(description)
+        main_view.flash.assert_no_error()
 
 
 @attr.s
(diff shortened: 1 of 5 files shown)


cfme/infrastructure/pxe.py
```diff
@@ -232,7 +233,7 @@ def create(self, cancel=False, refresh=True, refresh_timeout=120):
                                                    'was cancelled by the user')
         else:
             view.add.click()
-            main_view.flash.assert_success_message('PXE Server "{}" was added'.format(self.name))
+            main_view.flash.assert_no_error()
```
lkhomenk: we just don't need errors.

cfme/tests/infrastructure/test_pxe_provisioning.py
```diff
@@ -122,4 +122,4 @@ def test_pxe_provision_from_template(appliance, provider, vm_name, smtp_test, se
             'vlan': pxe_vlan}}
 
     do_vm_provisioning(appliance, pxe_template, provider, vm_name, provisioning_data, request,
-                       smtp_test, num_sec=2100)
+                       smtp_test, num_sec=2400)
```
lkhomenk: PXE provision runs up to 35-40min

dajoRH:
Lint report for commit ef7c10d8b02c98206fa250fcac679fdeda11db2f:

`cfme/infrastructure/pxe.py`:
- :bangbang: Line 296:48: **E126** *continuation line over-indented for hanging indent*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/infrastructure/test_customization_template.py
```diff
@@ -88,14 +89,19 @@ def test_duplicate_name_error_validation():
 @pytest.mark.xfail(message='http://cfme-tests.readthedocs.org/guides/gotchas.html#'
     'selenium-is-not-clicking-on-the-element-it-says-it-is')
 def test_name_max_character_validation():
-    """Test to validate name with maximum characters in customization templates."""
+    """Test to validate name with maximum characters in customization templates.
+       Max length is controlled by UI elements - we are not allowed to input more than we should
+       Opens template details to verify that extra symbols were cut
+    """
```
lkhomenk: do we need this at all?

lkhomenk:
pxe service provision fails due to https://bugzilla.redhat.com/show_bug.cgi?id=1515841
rhv41 fails due to yamls gap - we are trying to pick up vlan matching whole name, no partial name matching here

cfme/tests/infrastructure/test_pxe_provisioning.py
```diff
@@ -122,4 +122,4 @@ def test_pxe_provision_from_template(appliance, provider, vm_name, smtp_test, se
             'vlan': pxe_vlan}}
 
     do_vm_provisioning(appliance, pxe_template, provider, vm_name, provisioning_data, request,
-                       smtp_test, num_sec=2100)
+                       smtp_test, num_sec=3600)
```
lkhomenk: post provision script takes extremely long time to finish. not sure we need that stuff as it's not related to this tests in particular

cfme/tests/infrastructure/test_customization_template.py
```diff
@@ -88,14 +89,19 @@ def test_duplicate_name_error_validation():
...
     template_name = pxe.CustomizationTemplate(
         name=fauxfactory.gen_alphanumeric(256),
         description=fauxfactory.gen_alphanumeric(16),
         image_type='RHEL-6',
         script_type='Kickstart',
         script_data='Testing the script')
-
-    with error.expected('Name is required'):
```
lkhomenk: logic here was changed - we are able to input only 255 symbols, everything else gets cutted => open details of template by slice and check it's lengths

cfme/infrastructure/pxe.py
rlbabyuk: **Question:** I think we still need to check messages for older version than 5.9. And isn't this a bz that we don't have success messages for 5.9?

lkhomenk: I assume for checking content of messages there're should be separate test. these assertions are not even in tests - but are blocking template creation

lkhomenk: i'll file BZ on missing messages - maybe there was some error and we should get those messages

dajoRH:
*CFME QE Bot*

cfme/tests/infrastructure/test_customization_template.py
```diff
@@ -63,7 +65,12 @@ def test_pxe_image_type_required_error_validation(collection):
             script_data='Testing the script')
 
 
-@pytest.mark.uncollectif(lambda: version.current_version() < '5.9')
+@pytest.mark.meta(
+    blockers=[
+        BZ(1092951, ignore_bugs=[1083198]),
+        BZ(1450927, forced_streams=['5.8']),
+    ]
+)
```
lkhomenk: there's a BZ 1450927 for this test case. it was created against 5.8 but got fixed in 5.9. It has just got 5.8z flag - fix will be copied into 5.8. Also 5.7 wasn't affected but issue with duplicates - taking all this into account - better to leave this until in gets fixed in 5.8 too

cfme/tests/infrastructure/test_customization_template.py
```diff
@@ -63,7 +65,7 @@ def test_pxe_image_type_required_error_validation(collection):
             script_data='Testing the script')
 
 
-@pytest.mark.uncollectif(lambda: version.current_version() < '5.9')
```
lkhomenk: there's a BZ 1449116 for this test case. it was created against 5.8 but got fixed in 5.9. It has just got 5.8z flag - fix will be copied into 5.8. Also 5.7 wasn't affected but issue with duplicates - taking all this into account - better to leave this until in gets fixed in 5.8 too

lkhomenk:
added BZ blocker and Customization Template fixes to pass PRT

lkhomenk:
Previous test run was almost good except failures on Service order - BZ blockers were added. @izapolsk could you please recheck

cfme/services/service_catalogs/ui.py
```diff
@@ -1,3 +1,5 @@
+import pytest
```
psav: **REQUIRED** We shouldn't ever import pytest into a model. We can raise a NotImplemented error which I think is now caught and turned in to a skip

cfme/tests/infrastructure/test_customization_template.py
```diff
@@ -86,16 +88,19 @@ def test_duplicate_name_error_validation(collection):
...
-            description=fauxfactory.gen_alphanumeric(16),
-            image_type='RHEL-6',
-            script_type='Kickstart',
-            script_data='Testing the script')
-    collection.delete(False, template_name)
+    """Test to validate name with maximum characters in customization templates.
+       Max length is controlled by UI elements - we are not allowed to input more than we should
+       Opens template details to verify that extra symbols were cut
+    """
+    template_name = collection.create(
```
psav: **REQUIRED**: The collection should instantiate and return the object. Not just a name. I'm not sure why we need to create a second object here.

cfme/tests/services/test_pxe_service_catalogs.py
```diff
@@ -67,8 +67,6 @@ def setup_pxe_servers_vm_prov(pxe_server, pxe_cust_template, provisioning):
     if not pxe_server.exists():
         pxe_server.create()
     pxe_server.set_pxe_image_type(provisioning['pxe_image'], provisioning['pxe_image_type'])
-    if not pxe_cust_template.exists():
```
psav: **REQUIRED**: Why do we no longer assert that the template is there?

cfme/tests/infrastructure/test_customization_template.py
lkhomenk: @psav  : we create an object in UI. but UI cuts template name to 255 symbols only. so for convenience - I instantiate another object with correct name. otherwise it would be view.sidebar.tree.click_path() + no entity + delete clicking UI elements as we.
I'd rather remove this test at all =) - max length is under JS control right now

cfme/tests/services/test_pxe_service_catalogs.py
lkhomenk: @psav - calling pxe_cust_template - executes 'get_template_from_config' which  already creates an UI object - no need to check it once again

cfme/tests/infrastructure/test_customization_template.py
psav: so instead of recreating the object, just reassign the name?
```template_name.name = template_name[:255]``` after the create?

Comment (lkhomenk) [reply cfme/tests/infrastructure/test_customization_template.py]: @psav done

File: cfme/tests/infrastructure/test_customization_template.py [hunk 11]
```diff
@@ -63,7 +65,7 @@ def test_pxe_image_type_required_error_validation(collection):
             script_data='Testing the script')
 
 
-@pytest.mark.uncollectif(lambda: version.current_version() < '5.9')
+@pytest.mark.uncollectif(BZ(1449116, forced_streams=['5.7', '5.8']).blocks, reason='BZ 1449116')
```
mfalesni: **REQUIRED**: This introduces a module-import-time external access. Please make it a proper blocker or put it into lambda.

izapolsk:
@lkhomenk, please check PRT. it fails.

lkhomenk:
reason why 5.9 failed - it occurred that few test creating CustomizationTemplates were executed on the same appliance -  and CustomizationTemplates lacking additional verification/cleaning. that should be added in separate PR. IMHO.

lkhomenk:
rright now all PXE test are blocked by some unknown PXE servers which is connected to "VM Network". all VMware providers are affected

cfme/services/service_catalogs/ui.py
```diff
@@ -128,16 +128,9 @@ def order(self):
         view.fill(self.dialog_values)
     if self.ansible_dialog_values:
         view.fill(self.ansible_dialog_values)
-    if self.appliance.version < "5.9":
-        msg = "Order Request was Submitted"
-        msg_type = "success"
-    else:
-        msg = "Dialog submitted successfully!"
-        msg_type = "info"
```
lkhomenk: IMHO - we don't need this at all. it just gives no value. we have only these tests which are testing text from flash messages - https://github.com/lkhomenk/integration_tests/blob/d54791a439584da3a912b483e6bb417f4c60f84d/cfme/tests/cloud/test_providers.py#L33. other than that - we just need to order service. assert_no_error is enough

sshveta: @lkhomenk : this message in 5.9 "msg = "Dialog submitted successfully!" is a bug and should go in next build may be .https://bugzilla.redhat.com/show_bug.cgi?id=1518430

lkhomenk: @sshveta yes. that's true. but such cases should be separate tests like in the link I've pasted - separate test to check if flash message is correct. otherwise all other test  provide no value/metrics if they are failed on this - from jenkins side it will be failed "test_pxe_servicecatalog". I won't even put soft_assert there - all assertions supposed to be inside test, but not in helper functions etc.

dajoRH:
Lint report for commit c8bf9e9f788f66095f27221f671f3a652608c02c:

`cfme/infrastructure/pxe.py`:
- :bangbang: Line 993:1: **E303** *too many blank lines (3)*

**No commit flaws detected**.

*CFME QE Bot*

lkhomenk:
Latest PRT failed only on  PXE provisions as we still have renegate PXE/DHCP server

cfme/infrastructure/pxe.py
```diff
@@ -986,11 +975,18 @@ def get_template_from_config(template_config_name):
...
-                                  description=template_config['description'],
-                                  image_type=template_config['image_type'],
-                                  script_type=template_config['script_type'],
-                                  script_data=script_data)
+    custimization_template = collection.instantiate(name=template_config['name'],
+                                                    description=template_config['description'],
+                                                    image_type=template_config['image_type'],
+                                                    script_type=template_config['script_type'],
+                                                    script_data=script_data)
+    if not custimization_template.exists():
```
mfalesni: There is a typo in here.

PR #5873, opened 2017-11-17
[WIPTEST] Implement SSUINavigateStep.go wait_for_view arg

mshriver:
Taking a crack at implementing wait_for_view for SSUI navigation

### DIFF
--- a/cfme/utils/appliance/implementations/ssui.py
+++ b/cfme/utils/appliance/implementations/ssui.py
@@ -1,3 +1,5 @@
+import os
+
 from jsmin import jsmin
 from navmazing import Navigate, NavigateStep
 from selenium.common.exceptions import NoSuchElementException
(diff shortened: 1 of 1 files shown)


cfme/utils/appliance/implementations/ssui.py
```diff
@@ -127,10 +133,16 @@ def go(self, _tries=0):
             self.prerequisite_view = self.prerequisite()
             logger.debug("SSUI-NAVIGATE: Heading to destination {}".format(self._name))
             self.do_nav(_tries)
-        self.resetter()
+        if nav_args['use_resetter']:
+            self.resetter()
         self.post_navigate(_tries)
-        if self.VIEW is not None:
```
mshriver: So if this is false, the method returns None anyway, so might as well return self.view if its None too.

mshriver:
nvm this is dup of 5866

PR #5878, opened 2017-11-18
[1LP][RFR] SSUI retire test automation

sshveta:
{{pytest: -v --long-running --use-provider vsphere cfme/tests/ssui/test_ssui_myservice.py }}

### DIFF
--- a/cfme/services/myservice/ssui.py
+++ b/cfme/services/myservice/ssui.py
@@ -6,7 +6,7 @@ from widgetastic_manageiq import (
     Notification,
     SSUIAppendToBodyDropdown,
     SSUIConfigDropdown)
-from widgetastic_patternfly import Input, Button
+from widgetastic_patternfly import Input, Button, Dropdown
 from widgetastic.utils import VersionPick, Version
 
 from cfme.base.ssui import SSUIBaseLoggedInPage
@@ -25,7 +25,7 @@ import time
 
 class MyServicesView(SSUIBaseLoggedInPage):
     title = Text(locator='//li[@class="active"]')
-    service = SSUIlist(list_name='serviceList')
+    service = SSUIlist()
     notification = Notification()
 
     @property
@@ -51,14 +51,16 @@ class DetailsMyServiceView(MyServicesView):
                 self.title.text in {self.context['object'].name, 'Service Details'})
 
     notification = Notification()
-    policy_btn = SSUIDropdown('Policy')
-    lifecycle_btn = SSUIDropdown('Lifecycle')
+    policy = SSUIDropdown('Policy')
     power_operations = SSUIDropdown('Power Operations')
     access_dropdown = SSUIAppendToBodyDropdown('Access')
     remove_service = Button("Remove Service")
     configuration = VersionPick({
         Version.lowest(): SSUIConfigDropdown("dropdownKebabRight"),
         '5.8': SSUIDropdown('Configuration')})
+    lifecycle = VersionPick({
+        Version.lowest(): Dropdown("Retire"),
+        '5.8': SSUIDropdown('Lifecycle')})
     console_button = Button(tooltip="HTML5 console", classes=['open-console-button'])
 
 
@@ -137,6 +139,19 @@ class RemoveServiceView(MyServicesView):
             self.title.text == 'Remove Service')
 
 
+class RetireServiceView(MyServicesView):
+    title = Text(locator='//h4[@id="myModalLabel"]')
+
+    retire = Button('Yes, Retire Service Now')
+    cancel = Button('Cancel')
+
+    @property
+    def is_displayed(self):
+        return (
+            self.in_myservices and
+            self.title.text == 'Retire Service Now')
+
+
 @MyService.update.external_implementation_for(ViaSSUI)
 def update(self, updates):
     view = navigate_to(self, 'Edit')
@@ -147,7 +162,7 @@ def update(self, updates):
         lambda: view.is_displayed, delay=15, num_sec=300,
         message="waiting for view to be displayed"
     )
-    # TODO - remove sleep when BZ 1496233 is fixed
+    # TODO - remove sleep when BZ 1518954 is fixed
     time.sleep(10)
     assert view.notification.assert_message(
         "{} was edited.".format(self.name))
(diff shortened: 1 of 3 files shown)


widgetastic_manageiq/__init__.py
```diff
@@ -1211,27 +1211,26 @@ class SSUIlist(Widget, ClickableMixin):
...
+            ParametrizedLocator('.//div/span/*[normalize-space(.)={item_name|quote}]'))
 
         def list_click(self):
             """Clicks the list item with this name."""
 
             return self.list_item.click()
 
-    def __init__(self, parent, list_name, logger=None):
+    def __init__(self, parent, logger=None):
         Widget.__init__(self, parent, logger=logger)
```
`__init__` is not needed in that case.

cfme/services/myservice/ssui.py
```diff
@@ -59,6 +59,9 @@ def is_displayed(self):
     configuration = VersionPick({
         Version.lowest(): SSUIConfigDropdown("dropdownKebabRight"),
         '5.8': SSUIDropdown('Configuration')})
+    lifecycle_btn = VersionPick({
```
mshriver: **Required**

This duplicates the widget on L55, and please take this opportunity to change to `lifecycle` instead of `lifecycle_btn` - especially since it isn't a button.  We want to make this change to all the toolbar widgets, removing `_btn` suffix.

cfme/services/myservice/ssui.py
```diff
@@ -225,6 +241,20 @@ def launch_vm_console(self, catalog_item):
     return vm_obj
 
 
+@MyService.retire.external_implementation_for(ViaSSUI)
+def retire(self):
+    view = navigate_to(self, 'Retire')
+    view.retire.click()
+    view = self.create_view(MyServicesView)
+    wait_for(
+        lambda: view.is_displayed, delay=15, num_sec=300,
```
mshriver: **Optional**
Consider dropping this delay to 3s so we're not waiting so long to refresh

cfme/services/myservice/ssui.py
```diff
@@ -225,6 +241,20 @@ def launch_vm_console(self, catalog_item):
...
+        message="waiting for view to be displayed"
+    )
+    assert view.is_displayed
+    # TODO - remove sleep when BZ 1496233 is fixed
```
mshriver: Remove this sleep, the BZ is verified.

cfme/services/myservice/ssui.py
```diff
@@ -225,6 +241,20 @@ def launch_vm_console(self, catalog_item):
...
+@MyService.retire.external_implementation_for(ViaSSUI)
+def retire(self):
+    view = navigate_to(self, 'Retire')
+    view.retire.click()
+    view = self.create_view(MyServicesView)
+    wait_for(
+        lambda: view.is_displayed, delay=15, num_sec=300,
+        message="waiting for view to be displayed"
+    )
+    assert view.is_displayed
```
mshriver: **Optional**

Consider moving this assert to L249, and assert the return from wait_for, since it'll be the result of the last call to `view.is_displayed`.
https://github.com/RedHatQE/wait_for/blob/master/wait_for/__init__.py#L116

cfme/services/myservice/ssui.py
```diff
@@ -290,3 +320,20 @@ class MyServiceEditTags(SSUINavigateStep):
...
+
+
+@navigator.register(MyService, 'Retire')
+class MyServiceRetire(SSUINavigateStep):
+    VIEW = RetireServiceView
+
+    prerequisite = NavigateToSibling('Details')
+
+    def step(self):
+        wait_for(
```
mshriver: I'm on the fence if this should be required or not, but I'm leaning toward yes.

@psav recently made some updates to `SSUINavigateStep`, which now supports `wait_for_view` boolean arg.

I think you should drop this wait_for call, and expand your prerequisite to be a method, using `navigate_to(self.context['object'], 'Details', wait_for_view=True)`

cfme/tests/ssui/test_ssui_myservice.py
```diff
@@ -56,6 +58,17 @@ def test_myservice_crud(appliance, setup_provider, context, order_catalog_item_i
         my_service.delete()
 
 
+@pytest.mark.meta(blockers=[BZ(1518056, forced_streams=["5.7"])])
+@pytest.mark.parametrize('context', [ViaSSUI])
+def test_retire_service(appliance, setup_provider, context, order_catalog_item_in_ops_ui):
+    """Tests retire service"""
+    service_name = order_catalog_item_in_ops_ui.name
+    with appliance.context.use(context):
+        appliance.server.login()
```
mshriver: **Question**
Is this `login` necessary, the navigation in `my_service.retire()` should handle navigation within the context?

cfme/services/myservice/ssui.py
sshveta: This bug is fixed for 5.9 not for 5.8 , BZ - 1518954

cfme/services/myservice/ssui.py
sshveta: assert for view.is_displayed cannot be moved before wait as it fails without wait.

mshriver: No, I mean that wait_for returns what view.is_displayed returned - so you can just assert against that wait for statement. Not put the call before, make it one line.

cfme/services/myservice/ssui.py
mshriver: Please add both bz numbers to comment

cfme/services/myservice/ssui.py
sshveta: Done

cfme/services/myservice/ssui.py
sshveta: Done

cfme/services/myservice/ssui.py
sshveta: Done

cfme/services/myservice/ssui.py
sshveta: done

mshriver:
Can't change my review after merge, but nice work @sshveta!!

PR #5885, opened 2017-11-20
[1LP][RFR] Wait for embedded ansible only when enabling it

jkrocil:
Fix for 16+ tests

### DIFF
--- a/cfme/utils/appliance/__init__.py
+++ b/cfme/utils/appliance/__init__.py
@@ -1726,13 +1726,16 @@ class IPAppliance(object):
         if self.server_roles == roles:
             self.log.debug(' Roles already match, returning...')
             return
-        ansible = roles.get('embedded_ansible', False)
+        ansible_old = self.server_roles.get('embedded_ansible', False)
+        ansible_new = roles.get('embedded_ansible', False)
+        enabling_ansible = ansible_old is False and ansible_new is True
+
         yaml = self.get_yaml_config()
         yaml['server']['role'] = ','.join([role for role, boolean in roles.iteritems() if boolean])
         self.set_yaml_config(yaml)
-        timeout = 600 if ansible else 300
+        timeout = 600 if enabling_ansible else 300
         wait_for(lambda: self.server_roles == roles, num_sec=timeout, delay=15)
-        if ansible:
+        if enabling_ansible:
             self.wait_for_embedded_ansible()
 
     def enable_embedded_ansible_role(self):


cfme/utils/appliance/__init__.py
```diff
@@ -1726,13 +1726,18 @@ def server_roles(self, roles):
         if self.server_roles == roles:
             self.log.debug(' Roles already match, returning...')
             return
-        ansible = roles.get('embedded_ansible', False)
+        ansible_old = self.server_roles('embedded_ansible', False)
+        ansible_new = roles.get('embedded_ansible', False)
+        enabling_ansible = False
+        if ansible_old is False and ansible_new is True:
```
mshriver: consider just setting `enabling_ansible` to the result of the boolean statement.

```python
enabling = (old is False and new is True)

**Comment by jkrocil** [comment]
@quarckster

File: cfme/utils/appliance/__init__.py [hunk 2]
```diff
@@ -1726,13 +1726,18 @@ def server_roles(self, roles):
...
+            enabling_ansible = True
+
         yaml = self.get_yaml_config()
         yaml['server']['role'] = ','.join([role for role, boolean in roles.iteritems() if boolean])
         self.set_yaml_config(yaml)
-        timeout = 600 if ansible else 300
+        timeout = 900 if enabling_ansible else 300
```
mshriver: So if I understand correctly, the UI role is enabled almost immediately, but the workers for the role are what take several minutes to start up.

If my understanding is correct, can't we leave this timeout to be 300 for the UI update, and apply the longer timeout to `self.wait_for_embedded_ansible` call on L1741

jkrocil: Good catch! Yes, I agree with this, but Pete previously increased this to 600 and after talking to @quarckster I thought even 900 might be needed. But your are correct that 900 is really needed for `wait_for_embedded_ansible`, not for this. And `wait_for_embedded_ansible` is set to 900 by default so... I'll change this one back to 600, assuming Pete had a better reason to increase it to 600 than I have for the 900.

mshriver: @psav can you confirm behavior here, UI doesn't actually need 600s to update roles, just the workers?  Ultimately if that's the case the timeout will never actually come into play anyway :)

jkrocil:
None of the failures on PRT are related to this PR.

jkrocil:
Tested locally; works.

jkrocil:
2 approvals, appears to be working; merging to master.

PR #5888, opened 2017-11-20
[RFR] Update tasks and dependent modules, remove web_ui

mshriver:
Remove web_ui/pytest_selenium code from configure/tasks.

Remove wait_for_task_finished, was only being used in one place.

## TODO find good PRT mix to test this.  As far as I can tell none of it was used anywhere, other than the single containers.image call which I replaced.


dajoRH:
*CFME QE Bot*

PR #5903, opened 2017-11-22
[1LP][RFR][NOTEST] Corrected table xpath and added quotas property

vatsalparekh:
Purpose or Intent
=================

Currently there are no tests checking this table or the property so I've put [NOTEST]

### DIFF
--- a/cfme/configure/access_control.py
+++ b/cfme/configure/access_control.py
@@ -1289,7 +1289,9 @@ class DetailsTenantView(ConfigurationView):
     name = Text('Name')
     description = Text('Description')
     parent = Text('Parent')
-    table = Table('//*[@id="fieldset"]/table')
+    table = Table(VersionPick(
+        {Version.lowest(): '//*[@id="fieldset"]/table',
+         '5.9': '//*[@id="tenant_info"]/div[2]/div/fieldset/table'}))
 
     @property
     def is_displayed(self):
@@ -1404,6 +1406,28 @@ class Tenant(Updateable, BaseEntity):
         view.flash.assert_success_message('Quotas for {} "{}" were saved'.format(
             self.obj_type, self.name))
         assert view.is_displayed
+
+    @property
+    def quota(self):
+        view = navigate_to(self, 'Details')
+        quotas = {
+            'cpu': 'Allocated Virtual CPUs',
+            'memory': 'Allocated Memory in GB',
+            'storage': 'Allocated Storage in GB',
+            'num_vms': 'Allocated Number of Virtual Machines',
+            'templates': 'Allocated Number of Templates'
+        }
+
+        for field in quotas.keys():
+            item = view.table.row(name=quotas[field])
+            quotas[field] = {
+                'total': item.total_quota.text,
+                'in_use': item.in_use.text,
+                'allocated': item.allocated.text,
+                'available': item.available.text
+            }
+
+        return quotas
 
     def __eq__(self, other):
         if not isinstance(other, type(self)):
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit 4f92d13f454b99e20777e9538303367669a85258:

`cfme/configure/access_control.py`:
- :bangbang: Line 1414:18: **E203** *whitespace before ':'*
- :bangbang: Line 1425:29: **E126** *continuation line over-indented for hanging indent*
- :bangbang: Line 1429:25: **E121** *continuation line under-indented for hanging indent*

**No commit flaws detected**.

*CFME QE Bot*

cfme/configure/access_control.py
```diff
@@ -1405,6 +1407,29 @@ def set_quota(self, **kwargs):
...
+
+        for field in quotas.keys():
+            for item in view.table.rows():
+                if item.name.text == quotas[field]:
+                    quotas[field] = {
+                        'total': item.total_quota.text,
+                        'in_use': item.in_use.text,
+                        'allocated': item.allocated.text,
+                        'available': item.available.text
+                    }
```
mfalesni: I think you can replace the for + if with this:
```python
item = view.table.row(name=quotas[field])
```
It should be faster and more obvious.

vatsalparekh: Changed.

mfalesni: Why the `for` is still there? The code hints that it should only process one row, why?

vatsalparekh: @mfalesni Corrected.

cfme/configure/access_control.py
```diff
@@ -1289,7 +1289,9 @@ class DetailsTenantView(ConfigurationView):
     name = Text('Name')
     description = Text('Description')
     parent = Text('Parent')
-    table = Table('//*[@id="fieldset"]/table')
+    table = Table(VersionPick(
+        {Version.lowest(): '//*[@id="fieldset"]/table',
```
izapolsk: REQUIRED: I believe you can get rid of VersionPick here by using something like
```
//*[self::fieldset or @id="fieldset"]/table
```

cfme/configure/access_control.py
```diff
@@ -1405,6 +1407,28 @@ def set_quota(self, **kwargs):
...
+        view = navigate_to(self, 'Details')
+        quotas = {
+            'cpu': 'Allocated Virtual CPUs',
+            'memory': 'Allocated Memory in GB',
+            'storage': 'Allocated Storage in GB',
+            'num_vms': 'Allocated Number of Virtual Machines',
+            'templates': 'Allocated Number of Templates'
+        }
+
+        for field in quotas.keys():
```
izapolsk: OPTIONAL: when you iterate over dict, python uses keys() be default.
So,
for key in some_dict.keys()  equal to for key in some_dict

cfme/configure/access_control.py
mfalesni: Depends how many things match that xpath, in the case of fieldset tag I would not be so sure about uniqueness, I prefer an ID in the path :)

izapolsk: @mfalesni,  as far as I see old div with fieldset id turned into fieldset class. So, I doubt there are several fieldset classes on the page.
One more note, such long xpath with indexes and etc doesn't look robust.

vatsalparekh: I checked on all z-streams, this table is the only one having `fieldset` on the page

PR #5909, opened 2017-11-22
[1LP][RFR] Some work on Automate and its tests

mfalesni:
{{pytest: -v cfme/tests/automate/ -k 'not (_dialog_ or _dialogs_ or paginator or common_methods or _dialog)'}}


mfalesni:
Ignoring the `__init__` error now for now and I will take a look at the lock_unlock for 5.9 in some subsequent PR, so results-wise I think this should be good to go if hte code is good :)

PR #5912, opened 2017-11-22
[1LP][RFR]fix for ec2 cloudinit

lkhomenk:
Depends on https://github.com/ManageIQ/integration_tests/pull/5872 
Right now EC2 gives us VM IP much faster than VM is created which leads to ssh failure

### DIFF
--- a/cfme/tests/cloud/test_cloud_init_provisioning.py
+++ b/cfme/tests/cloud/test_cloud_init_provisioning.py
@@ -77,9 +77,15 @@ def test_provision_cloud_init(request, setup_provider, provider, provisioning,
     logger.info('Instance args: {}'.format(inst_args))
 
     instance.create(**inst_args)
-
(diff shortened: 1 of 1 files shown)


cfme/tests/cloud/test_cloud_init_provisioning.py
```diff
@@ -77,9 +77,15 @@ def test_provision_cloud_init(request, setup_provider, provider, provisioning,
...
-    connect_ip, tc = wait_for(mgmt_system.get_ip_address, [vm_name], num_sec=300,
-                              handle_exception=True)
+    provision_request = provider.appliance.collections.requests.instantiate(vm_name,
+                                                                   partial_check=True)
+    try:
+        provision_request.wait_for_request()
+    except Exception as e:
+        logger.info(
+            "Provision failed {}: {}".format(e, provision_request.request_state))
+        raise e
```
mfalesni: **REQUIRED**: Just `raise` so we can see the original stack trace in this case.

mshriver:
PRT failed because of missing template

PR #5915, opened 2017-11-23
[1LP][RFR] Changing test_datastores_summary to account for multiple providers

jan-zmeskal:
__Updating test__ test_datastores_summary. This is needed in case you have two providers and each one of them has datastore with the same name but different attributes. In the old way, datastores were compared if their name was matching. However if you had two datastores with the same name but e.g. different free space, it might have happened that they were compared incorrectly. 

This approach basically parses both outputs (from DB and from report) into the list with the same format and then compares if both lists are equal. 

If they are not, AssertionError from this is also quite readable and informative IMHO (see attached screenshot)
assert

{{pytest: cfme/tests/intelligence/reports/test_canned_corresponds.py::test_datastores_summary --use-provider rhv41 -vv}}

### DIFF
--- a/cfme/tests/intelligence/reports/test_canned_corresponds.py
+++ b/cfme/tests/intelligence/reports/test_canned_corresponds.py
@@ -126,7 +126,6 @@ def test_operations_vm_on(soft_assert, appliance):
 def test_datastores_summary(soft_assert, appliance):
     """Checks Datastores Summary report with DB data. Checks all data in report, even rounded
     storage sizes."""
-
     adb = appliance.db.client
     storages = adb['storages']
     vms = adb['vms']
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit 4b0b73a5b02eb60040e0dd22a3977551f7906e53:

`cfme/tests/intelligence/reports/test_canned_corresponds.py`:
- :red_circle: Line 186:5: **F401** *'time.sleep' imported but unused*
- :bangbang: Line 186:27: **E702** *multiple statements on one line (semicolon)*
- :red_circle: Line 186:29: **F821** *undefined name 'time'*

**No commit flaws detected**.

*CFME QE Bot*

PR #5916, opened 2017-11-23
[RFR] Web ui removals

psav:
{{pytest: --collect-only}}


cfme/provisioning.py
```diff
@@ -19,7 +18,7 @@ def do_vm_provisioning(appliance, template_name, provider, vm_name, provisioning
             'notes': note}})
     view = navigate_to(vm, 'Provision')
     view.form.fill_with(provisioning_data, on_change=view.form.submit_button)
-    flash.assert_no_errors()
+    view.flash.assert_no_error()
```
izapolsk: I doubt this is so easy. this procedure is called from different views.

PR #5940, opened 2017-11-24
[1LP][RFR] More complete provider.create_rest

mkoura:
Adds support for all infra and cloud providers.

{{pytest: -v -k 'test_provider_refresh and not gce_central' cfme/tests/cloud_infra_common/test_rest_providers.py --use-provider complete}}

### DIFF
--- a/cfme/common/provider.py
+++ b/cfme/common/provider.py
@@ -132,7 +132,6 @@ class BaseProvider(WidgetasticTaggable, Updateable, SummaryMixin, Navigatable):
             check_existing (boolean): Check if this provider already exists, skip if it does
             validate_inventory (boolean): Whether or not to block until the provider stats in CFME
                 match the stats gleaned from the backend management system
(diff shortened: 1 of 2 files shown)


cfme/common/provider.py
```diff
@@ -207,21 +206,130 @@ def create(self, cancel=False, validate_credentials=True, check_existing=False,
...
+        response = self.appliance.rest_api.response
+        if not response:
+            raise AssertionError("Provider wasn't added, status code {}".format(
+                response.status_code))
+
+        if validate_inventory:
+            self.validate()
+
+        self.appliance.rest_api.response = response
+        return True
```
quarckster: **REQUIRED**
This method doesn't look good. It's too big. I would like to see it divided into several parts.

mkoura:
@otsuman please re-review as there was a massive code shuffle. thanks

PR #5951, opened 2017-11-27
[1LP][RFR] Fix tenancy and project flash messages

jkrocil:
5+ tests, SSIA

User, Role and Group are not affected. Dev only made this flash message change for Tenant and Project... :face_with_head_bandage: 

{{pytest: -k test_unique_project_name_on_parent_level cfme/tests/configure/test_access_control.py}}


otsuman:
Failed in PRT for 5.9 - didn't create a new tenant. Impossible to check this fix because of that

PR #5956, opened 2017-11-27
[RFR] Removed old timelines mixin

psav:
{{pytest: --collect-only}}


izapolsk:
@psav, I believe you can remove web_ui.timelines as well.

PR #5976, opened 2017-11-28
[1LP][RFR] Fix automate nav by using icons

mfalesni:
{{pytest: -v cfme/tests/automate/test_instance.py cfme/tests/automate/test_method.py -k crud --long-running}}

### DIFF
--- a/cfme/automate/explorer/instance.py
+++ b/cfme/automate/explorer/instance.py
@@ -143,6 +143,8 @@ class InstanceEditView(AutomateExplorerView):
 
 
 class Instance(BaseEntity, Copiable):
+    ICON_NAME = 'fa-file-text-o'
+
     def __init__(self, collection, name, display_name=None, description=None, fields=None):
         super(Instance, self).__init__(collection)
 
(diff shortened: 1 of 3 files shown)


mfalesni:
Almost there, just needs the tree path matching in is_displayed to count with the icon ...

dajoRH:
*CFME QE Bot*

cfme/automate/explorer/instance.py
```diff
@@ -196,10 +196,15 @@ def domain(self):
 
     @property
     def tree_path(self):
+        if self.appliance.version < '5.8':
```
otsuman: No need for this, as I understood we don't support 5.7 now

cfme/automate/explorer/method.py
```diff
@@ -157,10 +157,17 @@ def domain(self):
 
     @property
     def tree_path(self):
+        if self.appliance.version < '5.8':
```
otsuman: same here

cfme/automate/explorer/instance.py
mfalesni: Yeah, I will remove it

PR #5988, opened 2017-11-28
[1LP][RFR] web_ui removal from reports test

psav:
web_ui removal from reports test

### DIFF
--- a/widgetastic_manageiq/__init__.py
+++ b/widgetastic_manageiq/__init__.py
@@ -1816,6 +1816,49 @@ class CompareToolBarActionsView(View):
             yield self.compressed_button
 
 
+class ReportToolBarViewSelector(View):
+    """ represents toolbar's view selector control
+
+    .. code-block:: python
+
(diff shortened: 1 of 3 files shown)


widgetastic_manageiq/__init__.py
```diff
@@ -1816,6 +1816,50 @@ def _view_buttons(self):
...
+    """ represents toolbar's view selector control
+        it is present on pages with items like Infra or Cloud Providers pages
+
+    .. code-block:: python
+
+        view_selector = View.nested(ItemsToolBarViewSelector)
+
+        view_selector.select('Tile View')
+        view_selector.selected
+    """
```
Seems like copy paste from `ItemsToolBarViewSelector` 😄. Maybe it's better to inherit `ItemsToolBarViewSelector` and redefine ROOT and buttons?

psav: I will fix the comment, but I think I'll need to pull out into a common View, which I will do in another PR, but thank you for this

PR #5989, opened 2017-11-28
[1LP][RFR] Add ability to verpick collections classes

mfalesni:


mfalesni:
Tests don't fail with this, which is a good sign.

mshriver:
PRT failure unrelated

PR #6020, opened 2017-11-30
[1LP][RFR] Reverting sentaku'd properties

psav:
* Keeping everything as commented out for now so it can be reverted
* Have the idea of a _fundamental_ that doesn't need to be picked
  properly

Just in case we can't resolve the problem another way

**Shriver adding PRT**
{{ pytest: cfme/tests/configure/test_zones.py --long-running -v --use-provider default }}

### DIFF
--- a/cfme/base/__init__.py
+++ b/cfme/base/__init__.py
@@ -1,6 +1,5 @@
 import attr
 import importscan
-
 import sentaku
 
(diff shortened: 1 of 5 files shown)


mfalesni:
@psav `__init__() got an unexpected keyword argument 'region'` ... is that expected? 28% pass rate

PR #6022, opened 2017-11-30
[1LP][RFR] Fix service ordering flash msg

jkrocil:
7 tests, SSIA

I don't believe it will pass on PRT. I think it will trigger the NotImplementedError and therefore skip.

{{pytest: cfme/tests/services/test_service_manual_approval.py -k test_service_manual_approval --long-running --use-provider vsphere6-nested}}

### DIFF
--- a/cfme/services/service_catalogs/ui.py
+++ b/cfme/services/service_catalogs/ui.py
@@ -132,7 +132,7 @@ def order(self):
         msg = "Order Request was Submitted"
         msg_type = "success"
     else:
-        msg = "Service ordered successfully!"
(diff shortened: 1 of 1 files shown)


cfme/services/service_catalogs/ui.py
```diff
@@ -132,7 +132,7 @@ def order(self):
         msg = "Order Request was Submitted"
         msg_type = "success"
     else:
-        msg = "Service ordered successfully!"
+        msg = "Dialog submitted successfully!"
```
mfalesni: :angry: but why they had to change it :godmode:

PR #6029, opened 2017-11-30
Update category/tag flash assers

nzr-zts:
Purpose or Intent
=================
Update category/tag flash assers while crud

{{pytest: -v -k test_tagvis_catalog}}

### DIFF
--- a/cfme/configure/configuration/region_settings.py
+++ b/cfme/configure/configuration/region_settings.py
@@ -155,7 +155,7 @@ class Category(Pretty, Navigatable, Updateable):
         row.actions.click()
         view.browser.handle_alert(cancel=cancel)
         if not cancel:
-            if not BZ(1510473, forced_streams=['5.9']).blocks:
+            if BZ(1510473, forced_streams=['5.9']).blocks:
                 view.flash.assert_success_message(
                     'Category "{}": Delete successful'.format(self.name))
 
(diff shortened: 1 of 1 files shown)


cfme/configure/configuration/region_settings.py
```diff
@@ -115,14 +115,11 @@ def create(self, cancel=False):
 
         if cancel:
             view.cancel_button.click()
-            flash_message = 'Add of new Category was cancelled by the user'
```
quarckster: **QUESTION**
Why did you remove all flash messages?

nzr-zts: As it was discussed before, it's better to check if there is no error than checking the text of the flash message. Flash messages could be checked in a separate test case to decrease the amount of failed test cases by this reason.

quarckster: I disagree, @mfalesni , @mshriver, @psav  your opinions.

mshriver: Not sure who/where this was discussed, but I would disagree with this opinion and change.

Though changing flash messages cause us some headache and churn for text locators, simply removing these assertions from the framework greatly lowers the quality of existing tests with no replacement for this lost test coverage.

I would consider accepting this if it also included new tests for all of the flash messages, but even then I would be resistant to the change. 

If these flash messages have changed, please submit a PR with version picking for the expected message.

dajoRH:
*CFME QE Bot*

PR #6035, opened 2017-12-01
[1LP][RFR] Rude evm restart is now less rude but still thorough

jkrocil:
`--ui-coverage` uses rude evm restart and it kills ~40% appliances per test run; I wasn't able to reproduce it locally, but this should take care of the issue.

If not, I'll revisit after next jenkins run.

Tested locally, works fine.

### DIFF
--- a/cfme/utils/appliance/__init__.py
+++ b/cfme/utils/appliance/__init__.py
@@ -1365,10 +1365,24 @@ class IPAppliance(object):
         store.terminalreporter.write_line('evmserverd is being restarted, be patient please')
         with self.ssh_client as ssh:
             if rude:
-                log_callback('restarting evm service by killing processes')
+                self._evm_service_command(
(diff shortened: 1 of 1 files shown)


jkrocil:
@mfalesni

cfme/utils/appliance/__init__.py
```diff
@@ -1365,10 +1365,24 @@ def restart_evm_service(self, rude=False, log_callback=None):
         store.terminalreporter.write_line('evmserverd is being restarted, be patient please')
         with self.ssh_client as ssh:
             if rude:
-                log_callback('restarting evm service by killing processes')
+                self._evm_service_command(
+                    "stop", expected_exit_code=0, log_callback=log_callback)
+                log_callback('Waiting for evm service to stop')
```
mfalesni: `self.evmserverd.stop()` maybe?

otsuman:
PRT failures not related to current PR

PR #6036, opened 2017-12-01
[1LP][RFR] Storage Volume Snapshot Model page [WT + Collection]

digitronik:
Purpose or Intent
=================

- added a new model page for storage volume snapshot
- added a new test for snapshot
- collection and widgetastic

{{pytest: cfme/tests/storage/test_volume_snapshot.py -v --long-running }}

### DIFF
--- a/cfme/storage/volume.py
+++ b/cfme/storage/volume.py
@@ -147,6 +147,18 @@ class VolumeBackupView(VolumeView):
     cancel = Button('Cancel')
 
 
+class VolumeSnapshotView(VolumeView):
+    @property
+    def is_displayed(self):
+        return False
+
+    snapshot_name = TextInput(name='snapshot_name')
+
+    save = Button('Save')
+    reset = Button('Reset')
+    cancel = Button('Cancel')
+
+
 @attr.s
 class Volume(BaseEntity):
 
@@ -206,6 +218,17 @@ class Volume(BaseEntity):
                  timeout=1000,
                  fail_func=self.refresh)
 
+    def create_snapshot(self, name):
+        """create snapshot of cloud volume"""
+        view = navigate_to(self, 'Snapshot')
+        view.snapshot_name.fill(name)
+
+        view.save.click()
+        view.flash.assert_success_message('Snapshot for Cloud Volume "{}" '
+                                          'created'.format(self.name))
+
+        wait_for(lambda: self.snapshots > 0, delay=20, timeout=1000, fail_func=self.refresh)
+
     @property
     def exists(self):
         try:
@@ -256,6 +279,16 @@ class Volume(BaseEntity):
         """
         view = navigate_to(self, 'Details')
         return int(view.entities.relationships.get_text_of('Cloud Volume Backups'))
+
+    @property
+    def snapshots(self):
+        """ number of available snapshots for volume.
+
+        Returns:
+            :py:class:`int' snapshot count.
+        """
+        view = navigate_to(self, 'Details')
+        return int(view.entities.relationships.get_text_of('Cloud Volume Snapshots'))
 
 
 @attr.s
@@ -365,4 +398,14 @@ class VolumeBackup(CFMENavigateStep):
 
     def step(self, *args, **kwargs):
         self.prerequisite_view.toolbar.configuration.item_select('Create a Backup of this Cloud '
+                                                                 'Volume')
+
+
+@navigator.register(Volume, 'Snapshot')
+class VolumeSnapshot(CFMENavigateStep):
+    VIEW = VolumeSnapshotView
+    prerequisite = NavigateToSibling('Details')
+
+    def step(self, *args, **kwargs):
+        self.prerequisite_view.toolbar.configuration.item_select('Create a Snapshot of this Cloud '
                                                                  'Volume')
--- a/cfme/storage/volume.py
+++ b/cfme/storage/volume.py
@@ -218,7 +218,7 @@ class Volume(BaseEntity):
                  timeout=1000,
                  fail_func=self.refresh)
 
-    def create_snapshot(self, name, cancle=False, reset=False):
+    def create_snapshot(self, name, cancel=False, reset=False):
         """create snapshot of cloud volume"""
         initial_snapshot_count = self.snapshots_count
         view = navigate_to(self, 'Snapshot')
@@ -231,7 +231,7 @@ class Volume(BaseEntity):
                 view.reset.click()
                 view.flash.assert_message('All changes have been reset')
 
-            elif cancle:
+            elif cancel:
                 view.cancel.click()
                 view = self.create_view(VolumeDetailsView)
                 wait_for(lambda: view.is_displayed, delay=2, timeout=10, message='waiting for view')
@@ -343,7 +343,7 @@ class VolumeCollection(BaseCollection):
         view.flash.assert_success_message(base_message.format(name))
 
         volume = self.instantiate(name, provider)
-        wait_for(lambda: volume.exists, delay=20, timeout=500, fail_func=volume.refresh)
+        wait_for(lambda: volume.exists, delay=50, timeout=1500, fail_func=volume.refresh)
 
         return volume
 
--- a/cfme/storage/volume.py
+++ b/cfme/storage/volume.py
@@ -218,24 +218,39 @@ class Volume(BaseEntity):
                  timeout=1000,
                  fail_func=self.refresh)
 
-    def create_snapshot(self, name):
+    def create_snapshot(self, name, cancle=False, reset=False):
         """create snapshot of cloud volume"""
         initial_snapshot_count = self.snapshots_count
         view = navigate_to(self, 'Snapshot')
 
-        view.snapshot_name.fill(name)
-        view.save.click()
+        changed = view.snapshot_name.fill(name)
 
-        view = self.create_view(VolumeDetailsView)
-        wait_for(lambda: view.is_displayed, delay=5, timeout=50, message='waiting for view')
+        # For changes only Save and Reset button activate
+        if changed:
+            if reset:
+                view.reset.click()
+                view.flash.assert_message('All changes have been reset')
 
-        view.flash.assert_success_message('Snapshot for Cloud Volume "{}" '
-                                          'created'.format(self.name))
+            elif cancle:
+                view.cancel.click()
+                view = self.create_view(VolumeDetailsView)
+                wait_for(lambda: view.is_displayed, delay=2, timeout=10, message='waiting for view')
+                view.flash.assert_message(
+                    'Snapshot of Cloud Volume "{}" was cancelled by the user'.format(self.name))
 
-        wait_for(lambda: self.snapshots_count > initial_snapshot_count,
-                 delay=20,
-                 timeout=1000,
-                 fail_func=self.refresh)
+            else:
+                view.save.click()
+
+                view = self.create_view(VolumeDetailsView)
+                wait_for(lambda: view.is_displayed, delay=2, timeout=10, message='waiting for view')
+
+                view.flash.assert_success_message('Snapshot for Cloud Volume "{}" '
+                                                  'created'.format(self.name))
+
+                wait_for(lambda: self.snapshots_count > initial_snapshot_count,
+                         delay=20,
+                         timeout=1000,
+                         fail_func=self.refresh)
 
     @property
     def exists(self):
--- a/cfme/storage/volume.py
+++ b/cfme/storage/volume.py
@@ -220,14 +220,18 @@ class Volume(BaseEntity):
 
     def create_snapshot(self, name):
         """create snapshot of cloud volume"""
+        initial_snapshot_count = self.snapshots_count
         view = navigate_to(self, 'Snapshot')
-        view.snapshot_name.fill(name)
 
+        view.snapshot_name.fill(name)
         view.save.click()
         view.flash.assert_success_message('Snapshot for Cloud Volume "{}" '
                                           'created'.format(self.name))
 
-        wait_for(lambda: self.snapshots > 0, delay=20, timeout=1000, fail_func=self.refresh)
+        wait_for(lambda: self.snapshots_count > initial_snapshot_count,
+                 delay=20,
+                 timeout=1000,
+                 fail_func=self.refresh)
 
     @property
     def exists(self):
@@ -255,7 +259,7 @@ class Volume(BaseEntity):
         """ size of storage cloud volume.
 
         Returns:
-            :py:class:`str' size of volume.
+            :py:class:`str` size of volume.
         """
         view = navigate_to(self, 'Details')
         return view.entities.properties.get_text_of('Size')
@@ -265,7 +269,7 @@ class Volume(BaseEntity):
         """ cloud tenants for volume.
 
         Returns:
-            :py:class:`str' respective tenants.
+            :py:class:`str` respective tenants.
         """
         view = navigate_to(self, 'Details')
         return view.entities.relationships.get_text_of('Cloud Tenants')
@@ -275,17 +279,17 @@ class Volume(BaseEntity):
         """ number of available backups for volume.
 
         Returns:
-            :py:class:`int' backup count.
+            :py:class:`int` backup count.
         """
         view = navigate_to(self, 'Details')
         return int(view.entities.relationships.get_text_of('Cloud Volume Backups'))
 
     @property
-    def snapshots(self):
+    def snapshots_count(self):
         """ number of available snapshots for volume.
 
         Returns:
-            :py:class:`int' snapshot count.
+            :py:class:`int` snapshot count.
         """
         view = navigate_to(self, 'Details')
         return int(view.entities.relationships.get_text_of('Cloud Volume Snapshots'))
--- a/cfme/storage/volume.py
+++ b/cfme/storage/volume.py
@@ -234,7 +234,7 @@ class Volume(BaseEntity):
             elif cancel:
                 view.cancel.click()
                 view = self.create_view(VolumeDetailsView)
-                wait_for(lambda: view.is_displayed, delay=2, timeout=10, message='waiting for view')
+                view.wait_displayed(timeout='10s')
                 view.flash.assert_message(
                     'Snapshot of Cloud Volume "{}" was cancelled by the user'.format(self.name))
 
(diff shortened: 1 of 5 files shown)


cfme/tests/storage/test_volume_snapshot.py
```diff
@@ -0,0 +1,81 @@
...
+    pytest.mark.provider([OpenStackProvider], scope='module')
+]
+
+STORAGE_SIZE = 1
+
+
+@pytest.yield_fixture(scope='module')
+def snapshot(appliance, provider):
+    volume_collection = appliance.collections.volumes
+    storage_manager = version.pick({'5.8': '{} Cinder Manager'.format(provider.name),
```
quarckster: **REQUIRED**
Please use `appliance.version` instead of `version.pick`.

cfme/storage/volume.py
```diff
@@ -293,6 +316,16 @@ def backups(self):
         view = navigate_to(self, 'Details')
         return int(view.entities.relationships.get_text_of('Cloud Volume Backups'))
 
+    @property
+    def snapshots(self):
```
mfalesni: **REQUIRED**: Wrong property naming. One would expect this to be a link to snapshot collection for this volume. Rename this to ``snapshot_count`` or something like that.

cfme/storage/volume.py
```diff
@@ -255,6 +267,17 @@ def create_backup(self, name, incremental=None, force=None):
...
+    def create_snapshot(self, name):
+        """create snapshot of cloud volume"""
+        view = navigate_to(self, 'Snapshot')
+        view.snapshot_name.fill(name)
+
+        view.save.click()
+        view.flash.assert_success_message('Snapshot for Cloud Volume "{}" '
+                                          'created'.format(self.name))
+
+        wait_for(lambda: self.snapshots > 0, delay=20, timeout=1000, fail_func=self.refresh)
```
mfalesni: **REQUIRED**: This should probably check the snapshot count first and then check against the previous snapshot count.

cfme/storage/volume_snapshot.py
```diff
@@ -0,0 +1,238 @@
...
+
+class VolumeSnapshotView(BaseLoggedInPage):
+    """A base view for all the Volume Snapshot pages"""
+    title = Text('.//div[@id="center_div" or @id="main-content"]//h1')
+    flash = FlashMessages(
+        './/div[@id="flash_msg_div"]/div[@id="flash_text_div" or '
+        'contains(@class, "flash_text_div")]')
+
+    @property
+    def in_volume_snapshot(self):
```
mfalesni: **REQUIRED**: Plural please, so we are consistent with the UI.

cfme/storage/volume_snapshot.py
```diff
@@ -0,0 +1,238 @@
...
+class VolumeSnapshotAllView(VolumeSnapshotView):
+    """The all Volume Snapshot page"""
+    toolbar = View.nested(VolumeSnapshotToolbar)
+    including_entities = View.include(BaseEntitiesView, use_parent=True)
+
+    @property
+    def is_displayed(self):
+        return (
+            self.in_volume_snapshot and
+            self.title.text == 'Cloud Volume Snapshots')
```
mfalesni: **REQUIRED**: If this is listing snapshots for a particular volume, then the is_displayed should check also the name of the volume from the UI. In case that is not possible (which would qualify for a bug against CFME anyway I guess), then return false.

cfme/storage/volume_snapshot.py
```diff
@@ -0,0 +1,238 @@
...
+        """ volume name of snapshot.
+
+        Returns:
+            :py:class:`str' respective volume name.
+        """
+        view = navigate_to(self, 'Details')
+        return view.entities.relationships.get_text_of('Cloud Volume')
+
+    @property
+    def tenant(self):
```
mfalesni: **REQUIRED**: Since this does not return the tenant object but just the name, ``tenant_name`` will be a more appropriate name for this property.

cfme/storage/volume_snapshot.py
```diff
@@ -0,0 +1,238 @@
...
+        """ size of cloud volume snapshot.
+
+        Returns:
+            :py:class:`int' size of volume snapshot in GB.
+        """
+        view = navigate_to(self, 'Details')
+        return int(view.entities.properties.get_text_of('Size').split()[0])
+
+    @property
+    def volume(self):
```
mfalesni: Same here with the name.

cfme/storage/volume_snapshot.py
digitronik: @mfalesni this is_displayed method for normal All page [Storage>Block Storage> Volume Snapshots] showing all available snapshots for different volumes. Not for a particular volume.

cfme/storage/volume.py
digitronik: my bad; I thought we are creatiing new volume for test so snapshot cout always zero; but good to compair wth initial snapshot count for famework flexibility.

cfme/storage/volume.py
```diff
@@ -268,7 +295,7 @@ def size(self):
         """ size of storage cloud volume.
 
         Returns:
-            :py:class:`str' size of volume.
+            :py:class:`str` size of volume.
```
mshriver: :+1:

cfme/storage/volume_snapshot.py
```diff
@@ -0,0 +1,238 @@
...
+
+    @View.nested
+    class relationships(Accordion):  # noqa
+        tree = ManageIQTree()
+
+
+class VolumeSnapshotView(BaseLoggedInPage):
+    """A base view for all the Volume Snapshot pages"""
+    title = Text('.//div[@id="center_div" or @id="main-content"]//h1')
+    flash = FlashMessages(
```
mshriver: **Required**
This is provided by inherited `BaseLoggedInPage` view, remove duplicate widget here.

cfme/storage/volume_snapshot.py
```diff
@@ -0,0 +1,238 @@
...
+    def delete(self):
+        """Delete snapshot """
+
+        view = navigate_to(self, 'Details')
+        view.toolbar.configuration.item_select('Delete Cloud Volume Snapshot')
+        view.flash.assert_success_message('Delete initiated for 1 Cloud Volume Snapshot.')
+
+        wait_for(
+            lambda: not self.exists,
+            message="Wait backups to disappear",
```
mshriver: **Required**
copypasta, 'Wait for snapshot to disappear' instead of 'backups'

cfme/tests/storage/test_volume_snapshot.py
```diff
@@ -0,0 +1,81 @@
...
+
+    yield snapshot
+
+    try:
+        if snapshot.exists:
+            snapshot_collection.delete(snapshot)
+        if volume.exists:
+            volume.delete(wait=False)
+    except Exception:
+        logger.warning('Exception during volume deletion - skipping..')
```
mshriver: **Required**
Split these into separate try blocks - if the snapshot delete fails, it'll never try the volume delete.

cfme/tests/storage/test_volume_snapshot.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+@pytest.mark.tier(3)
+@pytest.mark.uncollectif(lambda: version.current_version() < '5.8')
```
mshriver: You can drop this since we set the 5.7-breakpoint tag.

cfme/storage/volume_snapshot.py
digitronik: right; flash widget removing. @mshriver its good to add title also in `BaseLoggedInPage` View as it's common for all pages; what you say?

cfme/tests/storage/test_volume_snapshot.py
digitronik: Yup; Thank's to correct me; @izapolsk also suggested in `volume_backup.py`. coming with corrections to backup also. I need to implement other learned suggestions.

cfme/storage/volume_snapshot.py
mshriver: ah! glad you mentioned it.

Title is actually defined as part of the entities section of the page, in BaseEntitiesView.
https://github.com/ManageIQ/integration_tests/blob/master/widgetastic_manageiq/__init__.py#L2819

cfme/storage/volume_snapshot.py
```diff
@@ -0,0 +1,235 @@
...
+        tree = ManageIQTree()
+
+    @View.nested
+    class relationships(Accordion):  # noqa
+        tree = ManageIQTree()
+
+
+class VolumeSnapshotView(BaseLoggedInPage):
+    """A base view for all the Volume Snapshot pages"""
+    title = Text('.//div[@id="center_div" or @id="main-content"]//h1')
```
mshriver: Because of commits previous discussion has been hidden.

@digitronik raised a good point about title definition having to be repeated.

BaseEntitiesView includes a title widget via its conditionalview:
https://github.com/ManageIQ/integration_tests/blob/master/widgetastic_manageiq/__init__.py#L2819

So I think you can drop title from this generic top-level view. There's no need to define it in `VolumeSnapshotAllView` because `BaseEntitiesView` is included there. There is a need to define a `title` widget for the `VolumeSnapshotDetailsEntities`, as this is nested in `VolumeSnapshotDetailsView`.

This will provide a consistent path to the title widget via `view.entities.title` - regardless of what type the `view` object is.

digitronik: right @mshriver; going to add changes waiting for #6085. Due to this snapshot failing on PRT.

cfme/tests/storage/test_volume_snapshot.py
```diff
@@ -0,0 +1,81 @@
...
+        if snapshot.exists:
+            snapshot_collection.delete(snapshot)
+    except Exception:
+        logger.warning('Exception during snapshot deletion - skipping..')
+
+    try:
+        if volume.exists:
+            volume.delete(wait=False)
+    except Exception:
+        logger.warning('Exception during volume deletion - skipping..')
```
mfalesni: **REQUIRED**: Can you replace these warnings with `exception` or at least print out the exception type and message.

dajoRH:
Lint report for commit 7cbf18c3438663dcbbd07ee067e147d24d37292b:

`cfme/tests/storage/test_volume_snapshot.py`:
- :bangbang: Line 1:1: **E902** *TokenError: EOF in multi-line statement*
- :bangbang: Line 50:7: **E999** *SyntaxError: invalid syntax*

**No commit flaws detected**.

*CFME QE Bot*

cfme/storage/volume_snapshot.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+@navigator.register(VolumeSnapshotCollection, 'All')
+class All(CFMENavigateStep):
+    VIEW = VolumeSnapshotAllView
+    prerequisite = NavigateToAttribute('appliance.server', 'LoggedIn')
+
+    def step(self):
+            self.prerequisite_view.navigation.select(
+                'Storage', 'Block Storage', 'Volume Snapshots')
```
mfalesni: **REQUIRED**: These two lines are indented one more indent than necessary.

cfme/tests/storage/test_volume_snapshot.py
```diff
@@ -0,0 +1,85 @@
...
+    test_requirements.storage,
+    pytest.mark.ignore_stream("upstream"),
+    pytest.mark.usefixtures('setup_provider'),
+    pytest.mark.provider([OpenStackProvider], scope='module')
+]
+
+STORAGE_SIZE = 1
+
+
+@pytest.yield_fixture(scope='module')
```
mfalesni: **REQUIRED**: This should be function scoped as there are tests that delete it.

cfme/tests/storage/test_volume_snapshot.py
```diff
@@ -0,0 +1,85 @@
...
+            volume.delete(wait=False)
+    except Exception as e:
+        logger.warning("{name}:{msg} Volume deletion - skipping...".format(
+            name=type(e).__name__,
+            msg=str(e)))
+
+
+@pytest.mark.tier(3)
+def test_storage_volume_snapshot_create(snapshot):
+    wait_for(lambda: snapshot.status == 'available',
```
mfalesni: **REQUIRED**: Where does this test create the snapshot? Whatever specific is tested should be in the test and not in the fixture, so you should put the `.create` in the test.

cfme/storage/volume_snapshot.py
```diff
@@ -0,0 +1,234 @@
...
+    ENTITY = VolumeSnapshot
+
+    def all(self):
+        """returning all Snapshot objects for respective storage manager type"""
+        view = navigate_to(self, 'All')
+        view.toolbar.view_selector.select("List View")
+        snapshots = []
+
+        for item in view.entities.elements.read():
+            if self.filters.get('provider').name in item['Cloud Provider']:
```
mshriver: This needs to handle the AttributeException that occurs when there is no `provider` filter. 

Same behavior as: https://github.com/ManageIQ/integration_tests/issues/6123

mshriver:
Outstanding comments from me an Milan, WIPTEST

cfme/storage/volume_snapshot.py
digitronik: @mshriver can you please review #6132, so I can implement here also .has_key('provider') as well in issue you assign by you.

digitronik:
@mshriver @mfalesni added changes,
@mshriver all() method improved 
@mfalesni  Crud test improved as per your suggestion, creation and deletion in the same test. Created two fixture volume with scope `module` and snapshot with scope `function`.

cfme/storage/volume.py
```diff
@@ -206,6 +218,40 @@ def create_backup(self, name, incremental=None, force=None):
                  timeout=1000,
                  fail_func=self.refresh)
 
+    def create_snapshot(self, name, cancle=False, reset=False):
```
quarckster: **REQUIRED**
typo

cfme/storage/volume_snapshot.py
```diff
@@ -0,0 +1,250 @@
...
+        return snapshots
+
+
+@navigator.register(VolumeSnapshotCollection, 'All')
+class All(CFMENavigateStep):
+    VIEW = VolumeSnapshotAllView
+    prerequisite = NavigateToAttribute('appliance.server', 'LoggedIn')
+
+    def step(self):
+            self.prerequisite_view.navigation.select('Storage', 'Block Storage', 'Volume Snapshots')
```
quarckster: **REQUIRED**
Overindentation

cfme/storage/volume.py
```diff
@@ -206,6 +218,40 @@ def create_backup(self, name, incremental=None, force=None):
...
+        # For changes only Save and Reset button activate
+        if changed:
+            if reset:
+                view.reset.click()
+                view.flash.assert_message('All changes have been reset')
+
+            elif cancel:
+                view.cancel.click()
+                view = self.create_view(VolumeDetailsView)
+                wait_for(lambda: view.is_displayed, delay=2, timeout=10, message='waiting for view')
```
mfalesni: **REQUIRED**: Use `view.wait_displayed(timeout='10s')` here

cfme/storage/volume_snapshot.py
mfalesni: **REQUIRED**: Still indented too much.

cfme/storage/volume.py
digitronik: @mfalesni added changes.

psav:
Still two comments to address please

cfme/tests/storage/test_volume_snapshot.py
digitronik: @psav: now  `volume` fixture scope  `module` is right as it's for complete test not going to delete in any test. and I created new fixture l.e. `snapshot` having scope `function`. As per @mfalesni and IRC discussion fixture should be used for `setup` and `teardown` so I am not including `snapshot` fixture in `crud test`.

cfme/storage/volume.py
```diff
@@ -206,6 +218,39 @@ def create_backup(self, name, incremental=None, force=None):
...
+                view.wait_displayed(timeout='10s')
+                view.flash.assert_message(
+                    'Snapshot of Cloud Volume "{}" was cancelled by the user'.format(self.name))
+
+            else:
+                view.save.click()
+
+                view = self.create_view(VolumeDetailsView)
+                view.wait_displayed(timeout='10s')
+                view.flash.assert_success_message('Snapshot for Cloud Volume "{}" '
```
mshriver: **Required**

Move this flash message assertion to the test. This prevents a changed flash message from blocking tests that need to create one for further testing.

Since you've got a new test here with a fixture, it should be simple to create a view (BaseLoggedInPage has flash and would be appropriate) and assert this success message there.

Here, just use view.flash.assert_no_error() to at least make sure that the action was successful - just not asserting the specific flash message text.

digitronik: @mshriver: I agreed with your suggestion l.e. changes in the flash message will avoid blocking further testing. But while creating `snapshot` or `backup` for `storage volume` I need a check for `snapshot count`. If I assert this message in test it will disappear after `wait_for` specific count.

mshriver: Noted, and I think that wait can also be removed from this CRUD method and moved to the test as well.

I think a good way to do this would be for this method to instantiate and return a VolumeSnapshot instance on success (None on cancel/failure).

Then any framework components using it further, or tests, can simply use `Volume.snapshots_count` before/after create to validate counts, and can use the returned instance to assert it exists.

cfme/tests/storage/test_volume_snapshot.py
```diff
@@ -0,0 +1,174 @@
...
+    view = volume.create_view(VolumeDetailsView)
+    view.wait_displayed(timeout='10s')
+    view.flash.assert_success_message(
+        'Snapshot for Cloud Volume "{}" created'.format(volume.name))
+
+    # check for volume relationship tables snapshot count
+    try:
+        wait_for(lambda: volume.snapshots_count > initial_snapshot_count,
+                 delay=20, timeout=1000, fail_func=volume.refresh)
+    except TimeoutException:
```
mshriver: **Required**
This isn't the exception you want, this is `selenium.common.exceptions.TimeoutException`, which is specific to selenium timeouts.

`wait_for` raises a `TimedOutError`, which you can import from `cfme.utils.wait`.

cfme/tests/storage/test_volume_snapshot.py
```diff
@@ -0,0 +1,174 @@
+# -*- coding: utf-8 -*-
+import fauxfactory
+import pytest
+from selenium.common.exceptions import TimeoutException
```
mshriver: **Required**

Remove this import, you actually want `cfme.utils.wait.TimedOutError`

cfme/tests/storage/test_volume_snapshot.py
digitronik: @mshriver added changes

PR #6053, opened 2017-12-04
[1LP][RFR] Remove version.pick (part 1)

ohiliazo:
__Updating tests__ to remove support of versions lower than 5.8.

### DIFF
--- a/cfme/infrastructure/config_management.py
+++ b/cfme/infrastructure/config_management.py
@@ -360,9 +360,7 @@ class ConfigManager(Updateable, Pretty, Navigatable):
 
     @property
     def _refresh_flash_msg(self):
-        return version.pick({'5.7': 'Refresh Provider initiated for 1 provider ({})'.
-                                    format(self.type),
-                             '5.8': 'Refresh Provider initiated for 1 provider'})
+        return 'Refresh Provider initiated for 1 provider'
 
     @property
     def exists(self):
--- a/cfme/infrastructure/config_management.py
+++ b/cfme/infrastructure/config_management.py
@@ -236,6 +236,7 @@ class ConfigManager(Updateable, Pretty, Navigatable):
 
     pretty_attr = ['name', 'url']
     type = None
+    refresh_flash_msg = 'Refresh Provider initiated for 1 provider'
 
     def __init__(self, name=None, url=None, ssl=None, credentials=None, key=None, appliance=None):
         Navigatable.__init__(self, appliance=appliance)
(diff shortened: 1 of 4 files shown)


cfme/infrastructure/provider/__init__.py
```diff
@@ -86,9 +86,7 @@ def __init__(
 
     @cached_property
     def vm_name(self):
-        return version.pick({
-            version.LOWEST: "VMs and Instances",
-            '5.7.1': "Virtual Machines"})
+        return "Virtual Machines"
```
mfalesni: If there are no other instances of `def vm_name`, can you turn this into a class attribute?

cfme/infrastructure/config_management.py
```diff
@@ -360,9 +360,7 @@ def delete(self, cancel=False, wait_deleted=True, force=False):
 
     @property
     def _refresh_flash_msg(self):
-        return version.pick({'5.7': 'Refresh Provider initiated for 1 provider ({})'.
-                                    format(self.type),
-                             '5.8': 'Refresh Provider initiated for 1 provider'})
+        return 'Refresh Provider initiated for 1 provider'
```
mfalesni: This can probably just be a class attribute.

ohiliazo: OK

cfme/infrastructure/provider/__init__.py
ohiliazo: OK, I'll look and turn it if everything is ok

ohiliazo: I haven't wound any other instance of `def vm_name` in infrastructure/provider.

cfme/infrastructure/config_management.py
ohiliazo: Done.

cfme/infrastructure/provider/rhevm.py
```diff
@@ -71,10 +68,7 @@ def __init__(self, name=None, endpoints=None, zone=None, key=None, hostname=None
     def view_value_mapping(self):
         return {
             'name': self.name,
-            'prov_type': version.pick({version.LOWEST: 'Red Hat Enterprise Virtualization Manager',
-                                       '5.7.1': 'Red Hat Virtualization Manager',
-                                       '5.7.3': 'Red Hat Virtualization',
-                                       '5.8': 'Red Hat Virtualization Manager',
+            'prov_type': version.pick({version.LOWEST: 'Red Hat Virtualization Manager',
```
vatsalparekh: @ohiliazo in all 5.7z, 5.8z & 5.9z, I see `Red Hat Virtualization` in the provider list, my tests related to RHV failed for 5.7 so got here, maybe you can update version from `5.8.0.10` to latest 5.8z
@psav I think this got merged without running PRT, so this got in without failing.

PR #6057, opened 2017-12-05
[1LP][RFR] Removed current_version() for unsupported versions (part 1)

ohiliazo:
__Updating tests__ to remove support of versions lower than 5.8.

### DIFF
--- a/cfme/rest/gen_data.py
+++ b/cfme/rest/gen_data.py
@@ -77,6 +77,7 @@ def tags(request, rest_api, categories):
 
 def dialog_ui(appliance):
     """Creates service dialog using UI."""
+    # TODO: this function is not used with supported versions
     service_dialogs = appliance.collections.service_dialogs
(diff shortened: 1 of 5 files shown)


cfme/rest/gen_data.py
```diff
@@ -77,6 +77,7 @@ def tags(request, rest_api, categories):
 
 def dialog_ui(appliance):
```
ohiliazo: Not used anymore but not sure if we need to remove this.

cfme/rest/gen_data.py
```diff
@@ -384,8 +377,9 @@ def service_templates_rest(request, appliance, service_dialog=None, service_cata
 
 
 def service_templates(request, appliance, service_dialog=None, service_catalog=None, num=4):
```
ohiliazo: There are two tests using this function

PR #6061, opened 2017-12-05
[1LP][RFR] Image container reduce row selection time

brahmanim:
{{pytest: cfme/tests/containers/test_containers_smartstate_analysis.py -v --use-provider cm-env2 }}

Fixing navigation to Container Image Details because it is currently run very slow on appliance 5.8 searching row one by one which is very slow.
add search image by Image name , before get_entity by Image ID.

### DIFF
--- a/cfme/containers/image.py
+++ b/cfme/containers/image.py
@@ -156,8 +156,8 @@ class All(CFMENavigateStep):
 
     def resetter(self):
         # Reset view and selection
+        self.view.entities.search.clear_simple_search()
         self.view.toolbar.view_selector.select("List View")
-        self.view.paginator.set_items_per_page(1000)
 
 
 @navigator.register(Image, 'Details')
@@ -167,6 +167,7 @@ class Details(CFMENavigateStep):
 
     def step(self):
         self.prerequisite_view.entities.get_entity(provider=self.obj.provider.name,
+                                                   use_search=True, name=self.obj.name,
                                                    id=self.obj.id).click()
 
 


cfme/containers/image.py
```diff
@@ -166,6 +166,8 @@ class Details(CFMENavigateStep):
     prerequisite = NavigateToAttribute("parent", 'All')
 
     def step(self):
+        self.prerequisite_view.entities.get_entity(provider=self.obj.provider.name,
```
mshriver: This should replace the call on L171, shouldn't it?  If so it needs the click call as well. Right now this call returns an entity, which isn't captured, and then calls get_entity again, not using search, and calls click().

brahmanim: On Container Images table , the image name is not a unique key.
There for we use the search by image name (the only possibility that we have to use search by name),In order to reduce the number of lines on table , and get view with all entity's that includes this name.
Then we use the get_entity with Image ID that is the unique key , in order to select the specific image and click on it.
So the get_entity is not duplicated in this special condition that we have.

izapolsk: @brahmanim, did you try to pass both name and id ? It should work. Please let me know if it doesn't work. I'll fix this.

brahmanim:
@izapolsk , when i try to use the search box manually , not related to automation, it let me to search only by name , and not by id.
I can try it on automation , if you think that it will work!

brahmanim:
@izapolsk you are right!, i check it with name & id. the search was done by name and selection by ID.
I will upload the changes , and move this to review again.

brahmanim:
Test fail on appliance 5.8 for real reason, can't scan image that not based on RHEL, like image created from docker.io registry.
BZ 1448323 https://bugzilla.redhat.com/show_bug.cgi?id=1448323 correct this on appliance 5.9

PR #6067, opened 2017-12-05
[1LP][RFR] Removed current_version() from tests (part 2)

ohiliazo:
__Updating tests__ to remove support of versions <5.8.
__Deleting tests__ with uncollectif always True for versions 5.8+.

### DIFF
--- a/cfme/tests/infrastructure/test_config_management.py
+++ b/cfme/tests/infrastructure/test_config_management.py
@@ -9,8 +9,7 @@ from cfme.utils.blockers import BZ
 
 pytest_generate_tests = generate(gen_func=config_managers)
 pytestmark = [pytest.mark.uncollectif(lambda config_manager_obj:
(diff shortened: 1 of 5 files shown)


cfme/tests/infrastructure/test_config_management.py
```diff
@@ -48,21 +47,11 @@ def tag(category):
 
 
 @pytest.mark.tier(3)
-@pytest.mark.meta(
-    blockers=[BZ(1388928, unblock=lambda config_manager_obj: (
-        config_manager_obj.type == "Ansible Tower" and version.current_version() > "5.7.0.9") or
-        config_manager_obj.type != "Ansible Tower")]
-)
```
ohiliazo:
```
lambda x: (x.type == 'some_type' and True) or x.type != 'some_type' -->
lambda x: x.type == 'some_type or x.type != 'some_type'
lambda x: True
```
:)

PR #6070, opened 2017-12-05
[1LP][RFR] Fixes to broken automation

lcouzens:
Fixes to test_rh_updates to work with 5.9 packages and timeout extension to test_rh_register as it takes a longer period to register appliances.


lcouzens:
Current sat6 failures are due to sat6 environment not currently configured correctly.

PR #6080, opened 2017-12-06
[RFR]Implementing fix for timeprofile test cases.

anurag03:
Purpose or Intent
=================
Implementing fix related to timeprofile test case for 5.9.

{{pytest: cfme/tests/configure/test_timeprofile.py -v }}

### DIFF
--- a/cfme/configure/settings.py
+++ b/cfme/configure/settings.py
@@ -20,6 +20,7 @@ class TimeProfileAddForm(View):
     days = BootstrapSwitch(name='all_days')
     hours = BootstrapSwitch(name='all_hours')
     save_button = Button('Save')
+    add = Button('Add')
     configuration = Dropdown('Configuration')
     table = Table("//div[@id='main_div']//table")
     save_edit_button = Button('Save')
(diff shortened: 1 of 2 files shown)


cfme/configure/settings.py
```diff
@@ -20,6 +20,7 @@ class TimeProfileAddForm(View):
     days = BootstrapSwitch(name='all_days')
     hours = BootstrapSwitch(name='all_hours')
     save_button = Button('Save')
+    add = Button('Add')
```
izapolsk: REQUIRED: please follow widget naming approach used in this view

cfme/configure/settings.py
```diff
@@ -20,6 +20,7 @@ class TimeProfileAddForm(View):
...
     configuration = Dropdown('Configuration')
     table = Table("//div[@id='main_div']//table")
     save_edit_button = Button('Save')
```
izapolsk: REQUIRED: please check why does this view have two Save buttons

cfme/configure/settings.py
izapolsk: this is wrong way to add one more widget because title has been changed. use VersionPick for titles instead

cfme/configure/settings.py
```diff
@@ -59,9 +60,11 @@ def create(self, cancel=False):
             'timezone': self.timezone,
         })
         if not cancel:
-            view.timeprofile_form.save_button.click()
-            end = "saved" if self.appliance.version > '5.7' else "added"
-            view.flash.assert_message('Time Profile "{}" was {}'.format(self.description, end))
+            if self.appliance.version < '5.9':
```
izapolsk: ^^

cfme/configure/settings.py
anurag03: The second save button is always save on edit page.

izapolsk: as far as I see both Button widgets have Save title. So, I don't see any difference between them.

PR #6081, opened 2017-12-06
[RFR] Removed support for versions <5.8

ohiliazo:
__Updating tests__ to disable support of versions lower than 5.8.

### DIFF
--- a/fixtures/ui_coverage.py
+++ b/fixtures/ui_coverage.py
@@ -181,9 +181,9 @@ class CoverageManager(object):
         def _bundle_install():
             self.ipapp.ssh_client.run_command('yum -y install git')
             self.ipapp.ssh_client.run_command('cd {}; bundle'.format(rails_root))
+
         version.pick({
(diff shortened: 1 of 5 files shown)


fixtures/ui_coverage.py
```diff
@@ -181,9 +181,9 @@ def _gem_install():
         def _bundle_install():
             self.ipapp.ssh_client.run_command('yum -y install git')
             self.ipapp.ssh_client.run_command('cd {}; bundle'.format(rails_root))
+
         version.pick({
-            version.LOWEST: _bundle_install,
-            '5.4': _gem_install,
+            version.LOWEST: _gem_install,
```
izapolsk: Q: please double check I suspect this version pick can be removed at all.

ohiliazo: Depends on which version after 5.4 started to use _bundle_install.
This version pick can be removed if _bundle_install was used from 5.8 version.
Since I don't have this information, I've left this as it is.

PR #6085, opened 2017-12-06
[1LP][RFR] fix for changes in VirsionPick

digitronik:
Purpose or Intent
=================
- fixed changes in `volume.py`
- removed VirsionPick complatly from `manager.py`

{{pytest: cfme/tests/openstack/cloud/test_volumes.py --use-provider rhos7-ga -v}}

### DIFF
--- a/cfme/storage/volume.py
+++ b/cfme/storage/volume.py
@@ -102,10 +102,9 @@ class VolumeView(BaseLoggedInPage):
     """Base class for header and nav check"""
     @property
     def in_volume(self):
-        nav = Volume.nav.pick(self.context['object'].appliance.version)
         return (
             self.logged_in_as_current_user and
-            self.navigation.currently_selected == nav)
+            self.navigation.currently_selected == VolumeCollection.nav)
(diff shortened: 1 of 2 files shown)


cfme/storage/volume.py
```diff
@@ -297,6 +294,9 @@ class VolumeCollection(BaseCollection):
     """Collection object for the :py:class:'cfme.storage.volume.Volume'. """
     ENTITY = Volume
 
+    # Navigation menu option
+    nav = ['Storage', 'Block Storage', 'Volumes']
```
psav: **REQUIRED:** As nice as this is, right now no one else is doing it, we can propose it somewhere but I don't want to include it in this PR please

psav: Looking again I see that this has been in for a while, I would ask you to please remove it as part of this PR till we decide to do it as a standard, Thanks

digitronik: Ok @psav going to change in our standard format. Thanks

PR #6090, opened 2017-12-06
[RFR] Removed more web_ui

psav:
Removed more web_ui

### DIFF
--- a/cfme/tests/infrastructure/test_provisioning_dialog.py
+++ b/cfme/tests/infrastructure/test_provisioning_dialog.py
@@ -1,25 +1,24 @@
 # -*- coding: utf-8 -*-
 """This module tests various ways how to set up the provisioning using the provisioning dialog."""
-import re
-import fauxfactory
-import pytest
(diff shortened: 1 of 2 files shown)


cfme/tests/infrastructure/test_provisioning_dialog.py
```diff
@@ -74,7 +73,8 @@ def _provisioner(template, provisioning_data, delayed=None):
         vm = Vm(name=vm_name, provider=provider, template_name=template)
         view = navigate_to(vm, 'Provision')
         view.form.fill_with(provisioning_data, on_change=view.form.submit_button)
-        flash.assert_no_errors()
+        base_view = vm.appliance.browser.create_view(BaseLoggedInPage)
```
rlbabyuk: **Question**: Maybe better use RequestsView here, as user after submitting redirects to Requests page

psav: Not if there is an error

PR #6097, opened 2017-12-07
[1LP][RFR] Added support of Compliances for Middleware Servers. Added new tests.

hhovsepy:
{{pytest: cfme/tests/middleware/test_middleware_server_compliance.py -v --use-provider hawkular}}

### DIFF
--- a/cfme/middleware/server.py
+++ b/cfme/middleware/server.py
@@ -5,7 +5,11 @@ from navmazing import NavigateToSibling, NavigateToAttribute
 from selenium.common.exceptions import NoSuchElementException
 from wrapanapi.hawkular import CanonicalPath
 
-from cfme.common import WidgetasticTaggable, UtilizationMixin
+from cfme.common import (
+    WidgetasticTaggable,
+    PolicyProfileAssignable,
+    UtilizationMixin
+)
 from cfme.exceptions import MiddlewareServerNotFound, \
     MiddlewareServerGroupNotFound
 from cfme.middleware.domain import MiddlewareDomain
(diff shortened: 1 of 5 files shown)


quarckster:
One test failed in 5.9

hhovsepy:
@quarckster this failure is because of middleware environment issue, runs on my local.

cfme/middleware/server.py
```diff
@@ -275,6 +280,32 @@ def is_stopped(self):
         self.load_details(refresh=True)
         return self.get_detail("Properties", "Server State").lower() == 'stopped'
 
+    @variable(alias='ui')
```
izapolsk: it makes sense to implement this using sentaku

cfme/tests/middleware/test_middleware_server_compliance.py
```diff
@@ -0,0 +1,107 @@
...
+    get_hawkular_server,
+    get_domain_server,
+    verify_server_compliant
+)
+
+
+pytestmark = [
+    pytest.mark.usefixtures('setup_provider'),
+    pytest.mark.uncollectif(lambda: current_version() < '5.9'),
+    pytest.mark.provider([HawkularProvider], scope="function"),
```
izapolsk: Q: are you sure you really need to add a provider for each new test ?

cfme/middleware/server.py
psav: It does and we will, but right now middleware has a lot of variant and we'll convert them next year

cfme/tests/middleware/test_middleware_server_compliance.py
psav: Unless it gets removed it won't get re-added

PR #6099, opened 2017-12-07
[1LP][RFR]Changing the way of selecting instances for test_group_ownership

nzr-zts:
Purpose or Intent
=================
Changing user_ownership_vm.exists to group_ownership_vm.find_quadicon(from_any_provider=True) for restricted user, due to error "Could not find the item Instances by Provider/ec2-west in Boostrap tree instances_tree" because restricted user doesn't have permission to see Instances by Provider/ec2-west, another way is to search through All Instances.

{{pytest: -v cfme/tests/cloud_infra_common/test_vm_ownership.py -k 'test_user_ownership_crud or test_group_ownership'}}

### DIFF
--- a/cfme/tests/cloud_infra_common/test_vm_ownership.py
+++ b/cfme/tests/cloud_infra_common/test_vm_ownership.py
@@ -6,11 +6,7 @@ from cfme import test_requirements
 from cfme.base.credential import Credential
 from cfme.common.provider import BaseProvider
 from cfme.common.vm import VM
-<<<<<<< 68a63981a292aaee5095f0925b359f35cf05c84d
-=======
 from cfme.exceptions import VmOrInstanceNotFound
-from cfme.utils import testgen
->>>>>>> Changing user_ownership_vm.exists to group_ownership_vm.find_quadicon(from_any_provider=True) for restricted user
 from cfme.utils.blockers import BZ
 from cfme.utils.log import logger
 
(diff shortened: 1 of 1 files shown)


cfme/tests/cloud_infra_common/test_vm_ownership.py
```diff
@@ -126,15 +128,23 @@ def test_user_ownership_crud(request, user1, setup_provider, provider):
         assert user_ownership_vm.exists, "vm not found"
     user_ownership_vm.unset_ownership()
     with user1:
-        assert not user_ownership_vm.exists, "vm exists"
+        try:
```
rlbabyuk: **Required**: We have a code duplication here. Better to move it to function with return value True/False. And then assert the result of this function in the test.

nzr-zts:
Checks failed due to BZ [1524933](https://bugzilla.redhat.com/show_bug.cgi?id=1524933)

cfme/tests/cloud_infra_common/test_vm_ownership.py
```diff
@@ -104,6 +105,19 @@ def new_user(group_only_user_owned):
     return user
 
 
+def check_vm_exists(vm_ownership):
+    """
+        Checks if VM exists through All Instances tab.
+        Args:
+            vm_ownership: VM object for ownership test
+    """
```
mfalesni:
```python
""" Checks if VM exists through All Instances tab.

Args:
    vm_ownership: VM object for ownership test

Returns:
    :py:class:`bool`
"""
```

nzr-zts:
downstream-59z and upstream failed because of BZ [1524933](https://bugzilla.redhat.com/show_bug.cgi?id=1524933)

PR #6101, opened 2017-12-07
[1LP][RFR] Workaround: use internal disk on RHOS appliances

patchkez:
Purpose or Intent
=================
Currently we have problems running CFME appliances in RHOS environment. 
The issue is that before CFME appliance image is used by Sprout system, we start the image on each provider and do pre-configuration of that image. Pre-configuration involves setup of the VMDB database. appliance_console script requires that you have new block device available for the DB. Currently we start appliances on Openstack with ephemeral disk which is then visible in the system as vdb device. Once we are done with pre-configuration we shut down the instance down and make snapshots of it. Then this snapshot is ready for use. But the problem is that ephemeral disks are deleted on instance stop and snaphots cannot be done for them. Until we come with proper solution for openstack (heatstack templates/use volume and not ephemeral disk), this PR should provide workaround so we can use appliances on RHOS providers with just one disk.

This fix do these actions during image pre-configuration phase:

- allocates 5GB file into /var mountpoint
- maps this file as loopback interface
- adds udev rule which creates symlink for loop0, loop0p1 devices to vdb, vdb1
- appliance_console script the recognize vdb block device and configures DB on it
- adds systemd unit file which makes sure that loop0 is mapped during boot, it is recognized by LVM and all this is done after /var is mounted

There will be another PR to yamls repo to change the flavor to not use ephemeral disk.

### DIFF
--- a/cfme/utils/appliance/__init__.py
+++ b/cfme/utils/appliance/__init__.py
@@ -540,12 +540,11 @@ class IPAppliance(object):
 
             logging.info("Creating loopback script, unit file and udev rule")
             for path, content in CONTENT:
-                # client.run_command("service auditd restart", ensure_host=True)
                 client.run_command("cat > {path} << {content}".format(path=path, content=content))
 
             for command in COMMANDS_TO_RUN:
                 logging.info("Running command: {}".format(command))
-                client.run_command("cat > {path} << {content}".format(path=path, content=content))
+                client.run_command(command)
 
     # TODO: this method eventually needs to be moved to provider class..
     @logger_wrap("Configure GCE IPAppliance: {}")
--- a/cfme/utils/appliance/__init__.py
+++ b/cfme/utils/appliance/__init__.py
@@ -475,8 +475,8 @@ class IPAppliance(object):
         prov_data = conf.cfme_data.get('management_systems', {})['how to pass provider name here?']
         if prov_data['type'] == 'openstack':
 
-            LOOPBACK_SCRIPT_PATH = "/usr/local/sbin/loopbacks"
-            LOOPBACK_SCRIPT_CONTENT = """EOF
+            loopback_script_path = "/usr/local/sbin/loopbacks"
+            loopback_script_content = """EOF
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit dbcf38011d735356ceb15b1cc0e7420bc35a739f:

- :red_circle: Line 478:13: **N806** *variable in function should be lowercase*
- :red_circle: Line 479:13: **N806** *variable in function should be lowercase*
- :red_circle: Line 499:13: **N806** *variable in function should be lowercase*
- :red_circle: Line 500:13: **N806** *variable in function should be lowercase*
- :red_circle: Line 519:13: **N806** *variable in function should be lowercase*
- :red_circle: Line 520:13: **N806** *variable in function should be lowercase*
- :red_circle: Line 525:13: **N806** *variable in function should be lowercase*
- :red_circle: Line 533:13: **N806** *variable in function should be lowercase*

Hmmm


**No commit flaws detected**.

*CFME QE Bot*

cfme/utils/appliance/__init__.py
```diff
@@ -468,6 +474,78 @@ def configure(self, log_callback=None, **kwargs):
                 self.restart_evm_service(log_callback=log_callback)
             self.wait_for_web_ui(timeout=1800, log_callback=log_callback)
 
+    def configure_rhos_db_disk(self):
+        loopback_script_path = "/usr/local/sbin/loopbacks"
+        loopback_script_content = """EOF
```
mfalesni: **REQUIRED**: Please use `textwrap.dedent` on these strings just to be on the safe side with whitespace :)

```python
from textwrap import dedent

        loopback_script_content = dedent("""EOF
        #!/bin/bash
        ...
        EOF"""
```

mfalesni:
Okay, I am going to give it a shot and if it works, I will merge it and start working on sprout incorporation.

mshriver:
I've merged the sprout-yamls change in cfme-qe-yamls MR 391 that goes with this.

PR #6108, opened 2017-12-07
[RFR] Update sync_template_tracker, limit gce to private

mshriver:
GCE was collecting public images, we only want private

FIXES RHCFQE-6182


mshriver:
Tested locally fine, returned 28 images instead of 1021.

Its still returning images in all regions/zones for API interfaces for each zone, so each region/zone shows 28 images.

mshriver:
Merging, as CI will continue to run this and pull 1021 images every hour. No framework or test impacts.

PR #6128, opened 2017-12-12
[RFR] Removed Flash Messages and reordered imports

psav:
Implementation of Flash Singularity Proposal

### DIFF
--- a/cfme/base/ui.py
+++ b/cfme/base/ui.py
@@ -4,7 +4,6 @@ import time
 
 import re
 from navmazing import NavigateToSibling, NavigateToAttribute
-
 from selenium.webdriver.common.keys import Keys
 from widgetastic.widget import View, Table, Text, Image, FileInput
 from widgetastic_patternfly import (Accordion, Input, Button, Dropdown,
@@ -270,8 +269,6 @@ class LoggedIn(CFMENavigateStep):
 
 
 class ConfigurationView(BaseLoggedInPage):
-    flash = FlashMessages(
-        './/div[starts-with(@id, "flash_text_div") or starts-with(@class, "flash_text_div")]')
     title = Text('#explorer_title_text')
 
     @View.nested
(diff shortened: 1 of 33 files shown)


dajoRH:
Lint report for commit 032531d01b10e83ee30deb613a05e17b5f75723d:

`cfme/middleware/provider/middleware_views.py`:
- :bangbang: Line 349:42: **E999** *SyntaxError: invalid syntax*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

mshriver:
Travis build failed with a FlashMessage missing import

psav:
Updated

cfme/base/ui.py
```diff
@@ -1257,7 +1254,6 @@ def step(self):
 
 
 class AutomateImportExportBaseView(BaseLoggedInPage):
-    flash = FlashMessages('div.import-flash-message')
```
mshriver: This flash div is actually unique from the flash widget in BaseLoggedInPage, and should be kept.

I think it actually needs to be a unique name, like `import_flash`, because the page has the regular flash message div and this import div.  Trying to import an invalid file puts a message in the import-flash div, not the standard flash div.

cfme/base/ui.py
```diff
@@ -1257,6 +1254,8 @@ def step(self):
 
 
 class AutomateImportExportBaseView(BaseLoggedInPage):
+    # TODO This is currently overiding the base flash and should be renamed and efforts made
```
mshriver: :heart:

cfme/infrastructure/config_management.py
```diff
@@ -122,10 +122,6 @@ class ConfigManagementAddEntities(View):
     add = Button('Add')
     cancel = Button('Cancel')
 
-    # element attributes changed from id to class in upstream-fine+, capture both with locator
-    flash = FlashMessages('.//div[@id="flash_msg_div"]'
```
mshriver: These are defined under `entities` in the actual view, so there are calls to view.entities.flash.

https://github.com/psav/cfme_tests/blob/86a2aa54102ec46dc9d437d7f710e18511647ae2/cfme/infrastructure/config_management.py#L287

The ConfigManagementView inherits BaseLoggedInPage, so I think the calls to `view.entities.flash` just need to be changed to `view.flash`

mshriver: ** Fixed by updates below to `view.entities.flash` **

PR #6132, opened 2017-12-12
[1LP][RFR] Object: fix nav; changed tag method

digitronik:
Purpose or Intent
=================
- Fixed navigation: attribute change 'key';
- improve all method with filter
- added provider attribute; i wast go with standard; If create method came we need provider as attri
- new changes in tag
{{pytest: cfme/tests/storage/test_object_store_object.py --use-provider rhos7-ga -v}}

### DIFF
--- a/cfme/storage/object_store_object.py
+++ b/cfme/storage/object_store_object.py
@@ -9,6 +9,7 @@ from cfme.common import TagPageView, WidgetasticTaggable
 from cfme.exceptions import ItemNotFound
 from cfme.modeling.base import BaseCollection, BaseEntity
 from cfme.utils.appliance.implementations.ui import CFMENavigateStep, navigator, navigate_to
+from cfme.utils.providers import get_crud_by_name
 from widgetastic_manageiq import (
     Accordion,
     BaseEntitiesView,
@@ -100,6 +101,7 @@ class ObjectStoreObject(BaseEntity, WidgetasticTaggable):
         provider: provider
     """
     key = attr.ib()
+    provider = attr.ib()
 
 
 @attr.s
@@ -111,10 +113,24 @@ class ObjectStoreObjectCollection(BaseCollection):
     def all(self):
         """returning all Object Store Objects"""
         view = navigate_to(self, 'All')
-        view.toolbar.view_selector.select("Grid View")
-        objects = [self.instantiate(key=item)
-                   for item in view.entities.entity_names]
-        return objects
+        view.entities.paginator.set_items_per_page(500)
+        objects = []
+
+        try:
+            if 'provider'in self.filters:
+                for item in view.entities.elements.read():
+                    if self.filters['provider'].name in item['Cloud Provider']:
+                        objects.append(self.instantiate(key=item['Key'],
+                                                        provider=self.filters['provider']))
+            else:
+                for item in view.entities.elements.read():
+                    provider_name = item['Cloud Provider'].split()[0]
+                    provider = get_crud_by_name(provider_name)
+                    objects.append(self.instantiate(key=item['Key'], provider=provider))
+            return objects
+
+        except NoSuchElementException:
+            return None
 
     def delete(self, *objects):
         # TODO: capture flash message after BZ 1497113 resolve.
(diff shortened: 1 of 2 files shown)


cfme/storage/object_store_object.py
```diff
@@ -115,9 +116,14 @@ class ObjectStoreObjectCollection(BaseCollection):
     def all(self):
         """returning all Object Store Objects"""
         view = navigate_to(self, 'All')
-        view.toolbar.view_selector.select("Grid View")
-        objects = [self.instantiate(key=item)
-                   for item in view.entities.entity_names]
+
+        objects = []
+
+        # ToDo: use all_entity_names method as JS API issue (#2898) resolve.
```
psav: **REQUIRED:** This assumes that you will always want the provider filtered view and to return nothing, in the even that there is no provider filter. I don't think this is right. If there is no filter, we should return ALL of them. Otherwise, we shuldn't make the collection available on the appliance at all.

digitronik: @psav Its filtering objects as per provider. I am reading all objects and filter them with key attribute Cloud Provider. and return all objects of respective provider.

I used .filter() to provide provider in all() method as it's part of the collection class. all() method returning filtered object as per provider.

Maybe this wrong way of using .filter() method. please, can you suggest me exact use?

psav: Right, but the problem with the approach is that this is the ONLY supported way of using it. Meaning that if you don't supply the provider in the filter, you get an empty list.
It is expected that if you run a collection without any filters, that it should return ALL objects on the appliance.

digitronik: Adding exception so it will return ALL object without filter...

dajoRH:
Lint report for commit e1538bf35db1ed343af9f25a4b6560e3511bf253:

`cfme/storage/object_store_object.py`:
- :heavy_exclamation_mark: Line 123:24: **W601** *.has_key() is deprecated, use 'in'*

**No commit flaws detected**.

*CFME QE Bot*

cfme/storage/object_store_object.py
```diff
@@ -115,9 +116,18 @@ class ObjectStoreObjectCollection(BaseCollection):
...
+        if self.filters.has_key('provider'):                                # noqa
```
mshriver: this might be GitHub weird formatting, but only two spaces before in-line comments please.

Formatting aside, why is this line noqa'd? doesn't look like it needs it?

```python
if some.statement.true:  # noqa

Comment (digitronik) [reply cfme/storage/object_store_object.py]: @mshriver added more pythonic way: `.has_key()` support removed in py3 so best to go with in method.

File: cfme/storage/object_store_object.py [hunk 3]
```diff
@@ -104,6 +104,7 @@ class ObjectStoreObject(BaseEntity, WidgetasticTaggable):
         provider: provider
     """
     key = attr.ib()
+    provider = attr.ib(default=None)
```
mshriver: **Question/Required**
Can two object store objects on different providers have the same key?

If so, provider needs to be required, and all() will have to create a provider instance as well.

digitronik: Yup; I got a good idea about it: Using `get_crud_by_name(provider_name)` to create provider and `provider_name` generating from `Cloud Provider Name` or `Storage Manager Name` fetching from table only. It will similar to above filter only difference will return all objects with respective provider. Now no need to set `provider=None`. Improving @mshriver

dajoRH:
*CFME QE Bot*

digitronik:
I think; Its PRT failure; Able to run locally.

cfme/storage/object_store_object.py
```diff
@@ -111,10 +113,24 @@ class ObjectStoreObjectCollection(BaseCollection):
...
-        objects = [self.instantiate(key=item)
-                   for item in view.entities.entity_names]
-        return objects
+        view.entities.paginator.set_items_per_page(500)
+        objects = []
+
+        try:
+            if 'provider'in self.filters:
+                for item in view.entities.elements.read():
+                    if self.filters.get('provider').name in item['Cloud Provider']:
```
mfalesni: **REQUIRED**: Since you ensured `provider` indeed is in the `self.filters`, therefore you can use `[]`, `.get` is not necessary.

digitronik:
@mfalesni remove.get() for provider

mshriver:
Waiting on a commit squash to verified commits only.

PR #6133, opened 2017-12-12
[1LP][RFR] Fix appliance server and server name test, remove caching

jkrocil:
8+ tests

`appliance.server_name()` would get invalidated but `appliance.server.name` wouldn't. This takes care of that issue.

I'm leaving the deprecated methods in, just in case. I believe they could be just removed, but I'd prefer to do it the nice way, as planned by whoever added the deprecation messages.

Edit:
Removed all caching. Also removed `zone_description` in favor of `zone.description`. There is no mention of `zone_description` in the whole codebase, except the 2 changes I've made, so I believe it is safe to just remove, without deprecating it first.

{pytest: cfme/tests/configure/test_server_name.py}}

### DIFF
--- a/cfme/base/ui.py
+++ b/cfme/base/ui.py
@@ -493,7 +493,7 @@ class Details(CFMENavigateStep):
         self.prerequisite_view.accordions.settings.tree.click_path(
             self.obj.zone.region.settings_string,
             "Zones",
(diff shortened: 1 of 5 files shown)


dajoRH:
*CFME QE Bot*

PR #6135, opened 2017-12-12
[RFR] Removed web_ui toolbar from test_default_views_cloud

ohiliazo:
__Updating views__ to have them toolbar included.
__Updating tests__ to remove web_ui toolbars.

### DIFF
--- a/cfme/tests/configure/test_default_views_cloud.py
+++ b/cfme/tests/configure/test_default_views_cloud.py
@@ -52,13 +52,13 @@ def test_cloud_grid_defaultview(request, key):
     set_and_test_default_view(key, 'Grid View', gtl_params[key])
 
 
-def set_and_test_view(group_name, view):
+def set_and_test_view(group_name, view, selector_type='views_selector'):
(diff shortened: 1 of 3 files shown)


cfme/tests/configure/test_default_views_cloud.py
```diff
@@ -72,9 +71,20 @@ def test_cloud_compressed_view(request):
     set_and_test_view('Compare', 'Compressed View')
 
 
-def test_cloud_details_view(request):
-    set_and_test_view('Compare Mode', 'Details Mode')
+def set_and_test_mode(group_name, mode):
```
rlbabyuk: **Required**: set_and_test_view function can be updated to use also mode_selector, as too much code duplication. Add a parameter for mode_selector, views_selector, as an option:
def set_and_test_mode(group_name, mode, selector_type='views_selector'):
     selected_view = getattr(inst_view.actions, selector_type).selected

ohiliazo: @rlbabyuk ok, that's wise

PR #6141, opened 2017-12-12
[RFR] Fix up issues left over from VM conversion

rsnyman:
Fix up some issues highlighted in my previous PR, as well as completely remove references to web_ui in virtual_machines.py


rsnyman:
The failures are the following error, which doesn't seem to be related to my changes:

EndpointNotFound
Could not find requested endpoint in Service Catalog.

PR #6151, opened 2017-12-13
[RFR] Removed Flash Messages and reordered imports

psav:
Removed Flash Messages and reordered imports


cfme/infrastructure/virtual_machines.py
```diff
@@ -26,12 +22,18 @@
...
     VmOrInstanceNotFound)
 from cfme.infrastructure.provider.rhevm import RHEVMProvider
+from cfme.utils import version
 from cfme.utils.appliance.implementations.ui import navigator, CFMENavigateStep, navigate_to
 from cfme.utils.conf import cfme_data
 from cfme.utils.log import logger
 from cfme.utils.pretty import Pretty
 from cfme.utils.wait import wait_for
-from cfme.utils import version
+from cfme.web_ui import (
```
mshriver: This should be wt_manageiq Table import

psav: It hsould indeed - merge issue

cfme/web_ui/mixins.py
```diff
@@ -1,8 +1,9 @@
 # -*- coding: utf-8 -*-
-from cfme.fixtures import pytest_selenium as sel
-from cfme.web_ui import fill, Form, AngularSelect, Table, toolbar, form_buttons, flash
 from xml.sax.saxutils import quoteattr
+
+from cfme.fixtures import pytest_selenium as sel
 from cfme.utils.version import current_version
+from cfme.web_ui import fill, Form, AngularSelect, Table, toolbar, form_buttons
```
mshriver: web_ui.Form was removed in #6150

mshriver: mixin is used in one test module, cloud_infra_common.test_tag_objects.

Need to update that test module, and then this mixin can be removed

mshriver: This is in #6154 actually

PR #5701, opened 2017-10-26
[1LP][RFR]  Adding more events to the test timeline

t-fischer:
^ Make sure to prefix the Title with [WIP]/[WIPTEST]

Purpose or Intent
=================

__Updating tests__ Adding more events to the test timeline
‎ {{pytest: -v -k "test_timeline_events" --use-provider=vsphere6-nested}}

PR #6153, opened 2017-12-13
[RFR] fix bug in is_displayed

t-fischer:

fix bug in is_displayed
=================
Fixing a bug in vm view is displayed. it was returning false. So navigating through all vms was taking more time. Bug was noticed in the #5701 .This PR can be tested once #5701 will be merged


dajoRH:
*CFME QE Bot*

mfalesni:
I believe this is being taken care of in a different PR ... and besides, we should match the whole list, so wrapping hte string in a list would make more sense.

PR #6157, opened 2017-12-13
[1LP][RFR] Added test_custom_button_order_ansible_playbook_service

quarckster:
Intent
=================

__Adding test__  for ordering ansible service via a custom button. That feature implemented in version 5.9.

{{pytest: -v -k "test_custom_button_order_ansible_playbook_service" --long-running}}

### DIFF
--- a/cfme/ansible/repositories.py
+++ b/cfme/ansible/repositories.py
@@ -3,7 +3,8 @@
 import attr
 from navmazing import NavigateToAttribute, NavigateToSibling
 from widgetastic.exceptions import NoSuchElementException
-from widgetastic.widget import Text, Checkbox, Fillable
+from widgetastic.widget import Text, Checkbox, Fillable, View
+from widgetastic_manageiq import Table, PaginationPane, SummaryTable
 from widgetastic_patternfly import Dropdown, Button, Input
 
 from cfme.base.login import BaseLoggedInPage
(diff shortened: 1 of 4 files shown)


dajoRH:
Lint report for commit b08848536faa7ddf58f8edd91d1fd35d34fdb4de:

`cfme/automate/buttons.py`:
- :red_circle: Line 280:15: **N801** *class names should use CapWords convention*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

cfme/ansible/repositories.py
```diff
@@ -110,6 +117,12 @@ def db_object(self):
     def playbooks(self):
         return self.collections.playbooks
 
+    def get_detail(self, title, field, refresh=False):
```
mshriver: **Required/Question**

@psav @mfalesni @jkrocil Would like your opinion here.
It was my understanding part of the effort to convert to WT was removing `load_detail`/`get_detail` methods such as this, since with WT this can instead be achieved with a single line `view.entities.repository_options.get_text_of(field)`.

Looks like this is only used in a wait_for, where the refresh click can be passed as fail_func.

I'd prefer to not have `get_detail` methods sporadically through the API.

jkrocil: @mshriver Agreed.

This issues exists in a few other places of our codebase already, though:
https://github.com/ManageIQ/integration_tests/blob/master/cfme/infrastructure/host.py#L202
https://github.com/ManageIQ/integration_tests/blob/master/cfme/containers/provider/__init__.py#L572
https://github.com/ManageIQ/integration_tests/blob/master/cfme/common/vm.py#L277
...

They should be fixed all at once, but in a separate PR. I'm in favor of a bit of tech-debt for the sake of more stable RC.

PR #6174, opened 2017-12-14
[1LP][RFR] update reload buttons 

izapolsk:
reload button title was changed in latest 5.9.


ohiliazo:
It seems one button is left: __cfme/configure/tasks.py: 112__

izapolsk:
@ohiliazo, I fixed tasks.py. It has "the" in the title. Please check that it is correct.
config_management was already fixed by someone else.

PR #6175, opened 2017-12-14
[1LP][RFR] BZ blocker tries to handle cfme-future correctly

mfalesni:


cfme/utils/bz.py
```diff
@@ -160,13 +161,20 @@ def resolve_blocker(self, blocker, version=None, ignore_bugs=None, force_block_s
                 filtered.clear()
                 filtered.add(variant)
                 break
+            elif (
+                    variant.release_flag is not None and variant.release_flag == 'future' and
```
otsuman: **REQUIRED:** `variant.release_flag is not None` statement is redundant if you are using `variant.release_flag == 'future' ` as well

mfalesni: Done.

PR #6186, opened 2017-12-15
[1LP][RFR] Checks Password field empty or not after invalid login attempt

niyazRedhat:
This test will fail in 5.8 due to this bz 1089786.This bz is only fixed in 5.9. 
{{pytest: -v cfme/tests/test_login.py -k test_bad_password}}

### DIFF
--- a/cfme/tests/test_login.py
+++ b/cfme/tests/test_login.py
@@ -2,6 +2,7 @@ import pytest
 
 from cfme import test_requirements
 from cfme.base.credential import Credential
+from cfme.base.ui import LoginPage
 from cfme.utils.appliance import ViaSSUI, ViaUI
 from cfme.utils import conf, error, version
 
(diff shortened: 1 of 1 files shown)


cfme/tests/test_login.py
```diff
@@ -51,3 +52,5 @@ def test_bad_password(context, request, appliance):
     with appliance.context.use(context):
         with error.expected(error_message):
             appliance.server.login(user)
+        view = appliance.browser.create_view(LoginPage)
```
rlbabyuk: **Question**: Will this issue be fixed in 5.8? If not, please add version check here not to fail 5.8 test

vatsalparekh: @niyazRedhat `if version.current_version() < '5.9':`

niyazRedhat: @rlbabyuk Yes sure thanks. I am adding version check.

niyazRedhat: @vatsalparekh thanks

PR #6195, opened 2017-12-18
[1LP][RFR] .pick() is not needed as descriptor resolves it

psav:
* Was part of the WT conversion

### DIFF
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -1,10 +1,8 @@
 # -*- coding: utf-8 -*-
 """Module containing classes with common behaviour for both VMs and Instances of all types."""
 from datetime import datetime, date, timedelta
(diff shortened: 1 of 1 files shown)


cfme/common/vm.py
```diff
@@ -218,7 +215,7 @@ def delete(self, cancel=False, from_details=False):
 
         if from_details:
             view = navigate_to(self, 'Details')
-            view.toolbar.configuration.item_select(self.REMOVE_SINGLE.pick(self.appliance.version),
+            view.toolbar.configuration.item_select(self.REMOVE_SINGLE,
                                                    handle_alert=not cancel)
```
mfalesni: You can probably put the handle_alert on the same line.

PR #6197, opened 2017-12-18
[WIPTEST] Fixed BaseVM.compliance_status in cfme/common/vm.py

ohiliazo:
__Fixing__ BaseVM.compliance_status because its return statement currently raises AttributeError.

{{pytest: --long-running cfme/tests/control/test_actions.py -k test_action_check_compliance}}


quarckster:
Already done in #6191

PR #6201, opened 2017-12-18
[RFR] Do not flush - workaround

mfalesni:
Though I noticed that people use it in various places :(

### DIFF
--- a/cfme/utils/appliance/implementations/ui.py
+++ b/cfme/utils/appliance/implementations/ui.py
@@ -201,8 +201,9 @@ class CFMENavigateStep(NavigateStep):
             # self.log_message('Op is a Nop! ({})'.format(fn.__name__))
             return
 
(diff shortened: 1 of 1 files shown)


cfme/utils/appliance/implementations/ui.py
```diff
@@ -201,8 +201,9 @@ def check_for_badness(self, fn, _tries, nav_args, *args, **kwargs):
             # self.log_message('Op is a Nop! ({})'.format(fn.__name__))
             return
 
-        if self.VIEW:
-            self.view.flush_widget_cache()
+        # TODO: Uncomment after resolving the issue in widgetastic. Shouldn't be needed though :)
```
mshriver: might want to include link to WT PR if this actually gets merged: 
https://github.com/RedHatQE/widgetastic.core/pull/67

PR #6212, opened 2017-12-19
[1LP][RFR] Remove the Middleware Provider.

hhovsepy:


dajoRH:
Lint report for commit 9f0a5709ebed91d1ba89642aa397a210b5d9696f:

- :heavy_exclamation_mark: Line 531:2: **W291** *trailing whitespace*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

dajoRH:
*CFME QE Bot*

izapolsk:
@hhovsepy, please explain what is it for ?

quarckster:
@izapolsk I guess this required because Hawkular will be not included in CFME 5.9

hhovsepy:
@izapolsk @quarckster right, this is because Middleware Manager integration is cancelled as of 5.9 relase.

quarckster:
@hhovsepy  Honestly, I don't like the idea to just comment out everything. I would propose two options:
1. Leave everything as is and just disable tests related to Middleware.
2. Delete completely entire Middleware related code and keep it in your personal branch locally.
When Middleware will be integrated again you can enable the tests or push back your branch.

izapolsk:
I also thought of that. I think we have only choice - remove unnecessary code. We can always restore code later if it becomes necessary.

hhovsepy:
@quarckster  The reason why I did not leave middleware code, is because when it remains, then it should be maintained with the framework changes.
We discussed with @jkrocil that better to disable the code instead of deleting it at all is that there is a chance that hawkular will come within the Openshift Provider, so it will be easier to return it back by FIXME Middleware keyword.

But if it is critical to remove the unused code, I could delete it.

jkrocil:
@hhovsepy Since everyone would prefer to see it removed, ignore my initial suggestion to just comment it out, please, and let's just delete it. It's true that we can always recover it if necessary sometimes in the near future.

dajoRH:
*CFME QE Bot*

PR #6149, opened 2017-12-13
[WIPTEST]Timeline addditional events

t-fischer:
{pytest: -v -k "test_timeline_events" --use-provider=vsphere6-nested
Purpose or Intent
=================
Adding more events to test the timeline of infrastructure. 
Divide the test in PR #5701 to smaller ones.
Also fix navigation issue with in clone_vm and migrate_vm, ( due to legacy web_ui and wrong page displayed in ProvisionView.
Some tests like rename/clone might fail because there is no such event displayed yet. Bugs have been raised for this issue.

PR #6216, opened 2017-12-19
[1LP][RFR] added the reload button in InfraVmDetailsToolbar

t-fischer:
Fix navigation Issue in Upstream
==========================
{{pytest: cfme/tests/infrastructure/test_timelines.py -k "test_timeline_events" --use-provider=vsphere6-nested}}
It was impossible to navigate to the VM's Details/Timeline pages and other menu in the toolbar in Upstream. This is because it was because it was not finding the button "Reload" In the toolbar. 
PR like #6149 were failing in upstream.

### DIFF
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -90,6 +90,7 @@ class InfraVmDetailsToolbar(InfraGenericDetailsToolbar):
     """
     access = Dropdown("Access")
     power = Dropdown('VM Power Functions')
+    reload = Button(title='Reload current display')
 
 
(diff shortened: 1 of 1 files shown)


cfme/infrastructure/virtual_machines.py
```diff
@@ -90,6 +90,7 @@ class InfraVmDetailsToolbar(InfraGenericDetailsToolbar):
     """
     access = Dropdown("Access")
     power = Dropdown('VM Power Functions')
+    reload = Button(title='Reload current display')
```
izapolsk: REQUIRED: please update VersionPick in InfraGenericDetailsToolbar instead

PR #6232, opened 2017-12-20
[RFR] make azure clean up script remove stale resources from any subscription and any resource groups

izapolsk:
Script should now remove pips, nics from all subscriptions and resource groups.
slightly changed logging mechanizm.

requires ManageIQ/wrapanapi/202


mshriver:
This doesn't work in the CI context, because we don't have an appliance when running cleanup. The call to providers.get_crud fails because of this.

izapolsk:
Bad, I tested it only locally and didn't know about Jenkins job specifics.

jkrocil:
I'll keep that in mind for next time as well :wink:

PR #6243, opened 2017-12-21
[WIP] Implementing fix for "Name can't be blank issue".

anurag03:
Purpose or Intent
=================
To implement fix for test_quota_tagging which is failing due to "name can't be blank issue."
It's a selenium issue, resolving with workaround.

{{pytest: cfme/tests/infrastructure/test_quota_tagging.py -v }}

### DIFF
--- a/cfme/automate/explorer/domain.py
+++ b/cfme/automate/explorer/domain.py
@@ -289,6 +289,8 @@ class DomainCollection(BaseCollection):
             k: v
             for k, v in {'name': name, 'description': description, 'enabled': enabled}.items()
             if v is not None}
+        #   Workaround for solving automation issue.
+        add_page.name.fill(name)
         add_page.fill(fill_dict)
         if cancel:
             add_page.cancel_button.click()


cfme/automate/explorer/domain.py
```diff
@@ -289,6 +289,8 @@ def create(self, name=None, description=None, enabled=None, cancel=False):
             k: v
             for k, v in {'name': name, 'description': description, 'enabled': enabled}.items()
             if v is not None}
+        #   Workaround for solving automation issue.
```
izapolsk: @mfalesni, what's your opinion on this fix ?

anurag03: @izapolsk actually there is a change in UI, now we don't need to load the saved advance search option to delete it so there was a test case failure.

anurag03: @izapolsk : In the above case the name field doesn't get filled so I have to perform the fill action again.

izapolsk: @anurag03, I understand but I'm not sure that this is the best way to fix this issue.

btw, please add this form  to google doc where we collect pages with such old forms.

jkrocil: @psav This issue has to do with old forms. We need consensus on workarounds. I'm for making workarounds such as this with a `# FIXME FORM` comment nearby for easy identification later. Thoughts?

mshriver: Consider putting this in `before_fill()` or after_fill()` for the impacted views instead of a separate fill call. Then any tests directly filling will automatically include it.

anurag03:
@mshriver @izapolsk : should I close this PR?

mshriver:
@anurag03  I don't think you have to close the PR, especially if the workaround is still needed for stable automation. 

Can you explore the approach of doing this in the view's `before_fill` method (make sure you inherit BeforeFIll too), and adding `#FIXME FORM` comments as @jkrocil requested.  Doing it there will apply it to all fill calls to that form, not just the `create()` methods.

There are several examples of `before_fill` use in the framework, let me know if you want some assistance working through it.

anurag03:
This bug is resolved and working fine now.

PR #6245, opened 2017-12-21
[1LP][RFR] Fixed failing test_set_default_host_filter issue

Yadnyawalkya:
{{pytest: -v cfme/tests/infrastructure/test_set_default_filter.py -k test_set_default_host_filter}}

**Description:** 
list[0] = ALL
list[-2] = 'Status / Running (Default)

**Screenshot:**

screenshot from 2017-12-21 20-06-47

### DIFF
--- a/cfme/tests/infrastructure/test_set_default_filter.py
+++ b/cfme/tests/infrastructure/test_set_default_filter.py
@@ -25,7 +25,7 @@ def test_set_default_host_filter(request, appliance):
     appliance.server.logout()
     appliance.server.login_admin()
     navigate_to(host_collection, 'All')
-    assert view.filters.navigation.currently_selected == ['Status / Running (Default)']
+    assert view.filters.navigation.currently_selected[0] == 'Status / Running (Default)'
 
 
 def test_clear_host_filter_results(appliance):
--- a/cfme/tests/infrastructure/test_set_default_filter.py
+++ b/cfme/tests/infrastructure/test_set_default_filter.py
@@ -25,7 +25,7 @@ def test_set_default_host_filter(request, appliance):
     appliance.server.logout()
     appliance.server.login_admin()
     navigate_to(host_collection, 'All')
-    assert view.filters.navigation.currently_selected[-2] == 'Status / Running (Default)'
+    assert view.filters.navigation.currently_selected == ['Status / Running (Default)']
 
 
 def test_clear_host_filter_results(appliance):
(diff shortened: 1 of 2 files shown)


cfme/tests/infrastructure/test_set_default_filter.py
```diff
@@ -25,7 +25,7 @@ def unset_default_host_filter():
     appliance.server.logout()
     appliance.server.login_admin()
     navigate_to(host_collection, 'All')
-    assert view.filters.navigation.currently_selected[0] == 'Status / Running (Default)'
+    assert view.filters.navigation.currently_selected[-2] == 'Status / Running (Default)'
```
izapolsk: REQUIRED : If I got you right, currently_selected returns list of all filters. This is wrong. So, currently_selected should be fixed along with test.

Yadnyawalkya: You are right, currently_selected should return single selected value not list. Changing it! :) 👍

Yadnyawalkya:
@izapolsk: thanks for review! :)
Changes has been made, current locator works fine, please have a look!

cfme/tests/infrastructure/test_set_default_filter.py
```diff
@@ -25,7 +25,7 @@ def unset_default_host_filter():
     appliance.server.logout()
     appliance.server.login_admin()
     navigate_to(host_collection, 'All')
-    assert view.filters.navigation.currently_selected[0] == 'Status / Running (Default)'
+    assert view.filters.navigation.currently_selected == ['Status / Running (Default)']
```
izapolsk: REQUIRED: hehe, you still compare lists when there should be comparison of strings

Yadnyawalkya: Oh, Thanks @izapolsk! Earlier it was giving list of all element since wrong locator was there, now it is giving single string in list, you were expecting this right?

Yadnyawalkya:
Thanks @izapolsk and @rlbabyuk!

@izapolsk: Oh! Earlier it was giving list of all element since wrong locator was there, now it is giving single string in list, you were expecting this right?

izapolsk:
@Yadnyawalkya, it was returning list because there was list of elements. This widget always has only one active element. So, I think it should return one string instead of list.

NB: please double check that your changes won't break other tests which use views with this widget.

Yadnyawalkya:
Thank you @izapolsk: Earlier locator was returning all 'filter elements' in list, so after changing this, now it is returning 'single' active element in a list (as a string). We should not change returning list to string as method used here (currently_selected) also been used to check vertical navigation in widgetastic_patternfly, so this implementation is the correct one that I feel. 

Yes, I had checked other tests too and this change will not cause any damage at all.

izapolsk:
@Yadnyawalkya, vertical navigation can have several selected elements. However, this widget always has only one active element. So, I doubt this is correct. However, this remark isn't critical. So, I'm not going to insist.
@mfalesni, what do you think ?

PR #6253, opened 2017-12-22
[1LP][RFR] Fix up some outstanding issues after the VM merge

rsnyman:
Picked up a couple of small issues that caused some of the snapshot and genealogy tests to fail.


rsnyman:
Failure is due to https://bugzilla.redhat.com/show_bug.cgi?id=1509941

PR #6254, opened 2017-12-22
[1LP][RFR] Fixed failing tests due to string match issue

vatsalparekh:
Purpose or Intent
=================
{{ pytest: -vvv cfme/tests/cloud/test_provisioning.py -k 'test_provision_with_additional_volume or test_provision_with_boot_volume' }}

### DIFF
--- a/cfme/tests/cloud/test_provisioning.py
+++ b/cfme/tests/cloud/test_provisioning.py
@@ -6,6 +6,7 @@ import pytest
 
 from riggerlib import recursive_update
 from textwrap import dedent
+from widgetastic.utils import partial_match
 
 from cfme import test_requirements
 from cfme.automate.explorer.domain import DomainCollection
@@ -15,6 +16,7 @@ from cfme.cloud.provider.azure import AzureProvider
 from cfme.cloud.provider.gce import GCEProvider
 from cfme.cloud.provider.ec2 import EC2Provider
 from cfme.cloud.provider.openstack import OpenStackProvider
+from cfme.utils import error
 from cfme.utils.conf import credentials
 from cfme.utils.rest import assert_response
 from cfme.utils.generators import random_vm_name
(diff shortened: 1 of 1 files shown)


vatsalparekh:
in `5.9` the flavors are showing with full details with the resources of the flavor, this would make tests fail with are using just `m1.tiny`, so used `partial_match` to fix it.

izapolsk:
@vatsalparekh, all PRT failed please check and fix or explain errors.

cfme/tests/cloud/test_provisioning.py
```diff
@@ -33,7 +35,12 @@
 ]
 
 
-@pytest.yield_fixture(scope="function")
+@pytest.fixture()
+def vm_name():
+    return random_vm_name(context='prov')
+
+
+@pytest.fixture()
```
lkhomenk: this should be left as yield_fixture

vatsalparekh: for `pytest` > version 3, `yield_fixture` and just `fixture` would work as just same, just doing `yield` would make fixture yield fixture. 
@lkhomenk any reasons why it should be left like this?

lkhomenk: no. you are right. it's deprecated. the only reason to keep it might be some readability, but i'm not sure it's needed here

cfme/tests/cloud/test_provisioning.py
```diff
@@ -626,7 +630,19 @@ def _finish_method():
 
     instance.create(**inst_args)
 
-    prov_instance = provider.mgmt._find_instance_by_name(vm_name)
+    request_description = 'Provision from [{}] to [{}]'.format(small_template.name, instance.name)
```
vatsalparekh: @izapolsk `instance.note` is there for this

izapolsk:
I'm still trying to fix automation scripts

cfme/tests/cloud/test_provisioning.py
```diff
@@ -473,7 +475,9 @@ def original_request_class(appliance):
 
 @pytest.fixture(scope="module")
 def modified_request_class(request, domain, original_request_class):
-    original_request_class.copy_to(domain)
+    with error.handler('error: Error during \'Automate Class copy\''):
```
izapolsk: REQUIRED: please use different quote marks in this case

PR #6259, opened 2017-12-22
[1LP][WIP]Fix for test_permissions_catalog_add.

anurag03:
Purpose or Intent
=================
There were issues with the role creation and unnecessary argument passing to catalog.create(), which is fixed in this PR.

{{pytest: -v -k 'test_permissions_catalog_add or test_permissions'}}

### DIFF
--- a/cfme/tests/services/test_catalog.py
+++ b/cfme/tests/services/test_catalog.py
@@ -3,10 +3,13 @@ import fauxfactory
 import pytest
 
 import cfme.tests.configure.test_access_control as tac
-from cfme.utils import error
+
 from cfme import test_requirements
 from cfme.services.catalogs.catalog import Catalog
+from cfme.utils import error
+from cfme.utils.blockers import BZ
 from cfme.utils.update import update
+from cfme.utils.version import current_version
 
 pytestmark = [test_requirements.service, pytest.mark.tier(2)]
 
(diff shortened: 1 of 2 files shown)


cfme/tests/services/test_catalog.py
```diff
@@ -31,6 +34,9 @@ def test_catalog_duplicate_name():
     cat.delete()
 
 
+# Un-collecting for 5.8 because of BZ: 1486041
+@pytest.mark.uncollectif(lambda: current_version() < '5.9')
```
izapolsk: REQUIRED: please use BZ instead

NB: please don't mix several fixes in one PR

anurag03: @izapolsk  actually this is a fix for 5.9, but in 5.8 this test case fails due to above BZ.
The problem with the above BZ is that it's marked as verified, so here BZ can't be used.

izapolsk: @anurag03, 
1. there should be clone of this BZ for 5.8 in Bugzilla then. please ask Simaishi about clone.
2. replace current_version with  lambda: store.current_appliance.version

anurag03:
@izapolsk : Closing this PR as the fix is already provided by some other PR.

PR #6269, opened 2017-12-24
[1LP][RFR] Fixed expression editor

quarckster:
Purpose
=================

__Fixing__ expression editor. Due to the issue editing an entity's scope was always failed. It affected test_action_canel clone.

{{pytest: -v -k test_action_cancel_clone --long-running --use-provider vsphere6-nested}}

### DIFF
--- a/cfme/tests/control/test_actions.py
+++ b/cfme/tests/control/test_actions.py
@@ -215,14 +215,16 @@ def action_collection(appliance):
     return appliance.collections.actions
 
 
-@pytest.fixture(scope="module")
+@pytest.yield_fixture(scope="module")
 def compliance_condition(appliance):
     condition_collection = appliance.collections.conditions
-    return condition_collection.create(
+    _compliance_condition = condition_collection.create(
         conditions.VMCondition,
         fauxfactory.gen_alpha(),
         expression="fill_tag(VM and Instance.My Company Tags : Service Level, Gold)"
     )
+    yield _compliance_condition
+    _compliance_condition.delete()
 
 
 @pytest.fixture(scope="module")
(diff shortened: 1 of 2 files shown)


cfme/tests/control/test_actions.py
```diff
@@ -215,14 +215,16 @@ def action_collection(appliance):
     return appliance.collections.actions
 
 
-@pytest.fixture(scope="module")
+@pytest.yield_fixture(scope="module")
 def compliance_condition(appliance):
     condition_collection = appliance.collections.conditions
-    return condition_collection.create(
+    compliance_condition_ = condition_collection.create(
```
izapolsk: REQUIRED: Please use leading underscore instead. pep8 states:
_single_leading_underscore: weak "internal use" indicator. E.g. from M import * does not import objects whose name starts with an underscore.
single_trailing_underscore_: used by convention to avoid conflicts with Python keyword, e.g.
Tkinter.Toplevel(master, class_='ClassName')

dajoRH:
Lint report for commit e4f6022260fa22ce21cff5ae022e065023f34631:

`cfme/tests/control/test_actions.py`:
- :red_circle: Line 227:5: **F821** *undefined name '_compliance_condition_'*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/control/test_actions.py
```diff
@@ -215,14 +215,16 @@ def action_collection(appliance):
     return appliance.collections.actions
 
 
-@pytest.fixture(scope="module")
+@pytest.yield_fixture(scope="module")
```
izapolsk: this change is not necessary in pytest 3

quarckster: @izapolsk agree, but @mshriver insists to keep it.

PR #6270, opened 2017-12-25
[RFR] Fixed alert creation

ohiliazo:
__Fixing__ creation of alert. Now action instantiated before filling the form.

{{pytest: --target test_actions.py::test_action_create_snapshots_and_delete_them}}

### DIFF
--- a/cfme/control/explorer/actions.py
+++ b/cfme/control/explorer/actions.py
@@ -227,31 +227,30 @@ class ActionCollection(BaseCollection):
 
     def create(self, description, action_type, action_values=None):
         """Create an Action in the UI."""
-        action_values = action_values or {}
+        action = self.instantiate(description, action_type, action_values=action_values)
         view = navigate_to(self, "Add")
         view.fill({
-            "description": description,
-            "action_type": action_type,
-            "snapshot_name": action_values.get("snapshot_name"),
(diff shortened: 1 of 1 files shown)


cfme/control/explorer/actions.py
```diff
@@ -227,31 +227,30 @@ class ActionCollection(BaseCollection):
 
     def create(self, description, action_type, action_values=None):
         """Create an Action in the UI."""
-        action_values = action_values or {}
+        action = self.instantiate(description, action_type, action_values=action_values)
```
quarckster: **QUESTION**
@mfalesni  @psav should we instantiate an object before its creation in the UI? Does it matter?

ohiliazo: @quarckster in this case it matters, because ```alerts_to_evaluate``` are converted from list of Alerts to strings.
Without this it throws TypeError: unhashable type: 'Alert'.
I've looked at this for reference: [link](https://github.com/ManageIQ/integration_tests/blob/master/cfme/control/explorer/alert_profiles.py#L211)

quarckster:
Fixed in #6274

quarckster:
Obsoleted by #6274

jkrocil:
@ohiliazo Can we close this one?

ohiliazo:
@jkrocil I think yes

PR #6279, opened 2017-12-28
[RFR] Fix for change in id

vatsalparekh:
Purpose or Intent
=================
changed `name` to `id`, because in `5.9` only `id` is there in these `input`
{{ pytest: -vvv cfme/tests/services/test_provision_stack.py }}

### DIFF
--- a/cfme/services/service_catalogs/ui.py
+++ b/cfme/services/service_catalogs/ui.py
@@ -79,10 +79,10 @@ class OrderForm(ServicesCatalogView):
     role_arn = Input(locator='.//input[../../../div[contains(normalize-space(.), "Role ARN")]]')
     ssh_location = Input(name="param_SSHLocation")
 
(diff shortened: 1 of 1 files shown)


vatsalparekh:
PRT should fail at [here](https://github.com/ManageIQ/integration_tests/blob/master/cfme/services/service_catalogs/ui.py#L163)

izapolsk:
@vatsalparekh, this fix is not enough. almost all tests fail against 5.9 and 50% fail against 5.8.
role_arn you can find here - https://github.com/vatsalparekh/integration_tests/blob/9d79610e7e48886ca39ee5e839fb747a474bb157/cfme/services/service_catalogs/ui.py#L79

vatsalparekh:
@izapolsk I'm not finding `role_arn` related entity in the UI, or its use in the tests, need to check it with @sshveta 
The changes in this PR are required anyway

PR #6280, opened 2017-12-28
[RFR] Fixed test_user_duplicate_username to check duplicate Username instead of Full Name

ohiliazo:
__Fixing__ test_user_duplicate_username because it is currently testing identical Full Name field instead of Username.
__Adding__ test to check that it is allowed to create users with same full name.

{{pytest: -v cfme/tests/configure/test_access_control.py::test_user_duplicate_username cfme/tests/configure/test_access_control.py::test_user_allow_duplicate_name}}

### DIFF
--- a/cfme/tests/configure/test_access_control.py
+++ b/cfme/tests/configure/test_access_control.py
@@ -114,15 +114,13 @@ def test_user_login(appliance, group_collection):
 
 @pytest.mark.tier(3)
 def test_user_duplicate_username(appliance, group_collection):
-    """
-    Tests that creating user with existing username is forbidden.
+    """ Tests that creating user with existing username is forbidden.
 
     Steps:
-        1. Generate credential
-        2. Create a user with this credential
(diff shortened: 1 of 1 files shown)


quarckster:
@ohiliazo Please provide tests for PRT that affected by this PR. You can do it like this:
{{pytest: -v -k test_user_duplicate_username}}

cfme/tests/configure/test_access_control.py
```diff
@@ -112,21 +113,54 @@ def test_user_login(appliance, group_collection):
...
+def test_user_duplicate_username(appliance, group_collection):
+    """
+    Tests that creating user with existing username is forbidden.
+
+    Steps:
+        1. Generate credential
+        2. Create a user with this credential
+        3. Create another user with same credential
+
+    Should raise an exception."""
```
izapolsk: REQUIRED: please make docstring same like everywhere

PR #6293, opened 2017-12-29
[1LP][RFR]Azure cleanup

lkhomenk:
Right now  cleanup is broken on jenkins. ~~to loop though subscriptions we ought to re-instantiate some azure components. - added method for that~~. also extended all current cleanup methods to use Resource group.
Our cleanup will loop though subscription:resource_groups:storage_account:containers.
Logging was left as we have right now - no reporting, only stdout log with what we have removed 
needs - https://github.com/ManageIQ/wrapanapi/pull/204

### DIFF
--- a/scripts/azure_cleanup.py
+++ b/scripts/azure_cleanup.py
@@ -36,68 +36,32 @@ def azure_cleanup(nic_template, pip_template, days_old):
                 mgmt = get_mgmt(prov_key)
                 mgmt.logger = logger
                 for name, scr_id in mgmt.list_subscriptions():
-                    logger.info("subscription {s} is chosen".format(s=name))
+                    logger.info("Subscription '{s}' is chosen".format(s=name))
                     setattr(mgmt, 'subscription_id', scr_id)
-                    # removing stale nics
-                    removed_nics = mgmt.remove_nics_by_search(nic_template)
-                    if removed_nics:
-                        logger.info('following nics were removed:')
-                        for nic in removed_nics:
-                            logger.info(nic[0])
-                    else:
-                        logger.info("No '{}' NICs were found".format(nic_template))
+                    for resource_group in mgmt.list_resource_groups():
+                        mgmt.logger.info('Checking "{}" resource group:'.format(resource_group))
 
-                    # removing public ips
-                    removed_pips = mgmt.remove_pips_by_search(pip_template)
-                    if removed_pips:
-                        logger.info('following pips were removed:')
-                        for pip in removed_pips:
-                            logger.info(pip[0])
-                    else:
-                        logger.info("No '{}' Public IPs were found".format(pip_template))
+                        # removing stale nics
+                        mgmt.remove_nics_by_search(nic_template, resource_group)
 
-                # removing stale stacks
-                stack_list = mgmt.list_stack(days_old=days_old)
-                if stack_list:
-                    logger.info("Removing empty Stacks:")
-                    removed_stacks = []
-                    for stack in stack_list:
-                        if mgmt.is_stack_empty(stack):
-                            removed_stacks.append(stack)
-                            mgmt.delete_stack(stack)
+                        # removing public ips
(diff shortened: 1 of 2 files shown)


mshriver:
Sorry to break - merged a small fix for the script, shouldn't impact your other changes to additional cleanup, only the creation of the `AzureSystem` object.

dajoRH:
*CFME QE Bot*

scripts/azure_cleanup.py
```diff
@@ -36,68 +36,32 @@ def azure_cleanup(nic_template, pip_template, days_old):
                 mgmt = get_mgmt(prov_key)
                 mgmt.logger = logger
                 for name, scr_id in mgmt.list_subscriptions():
-                    logger.info("subscription {s} is chosen".format(s=name))
+                    logger.info("Subscription '{s}' is chosen".format(s=name))
```
mshriver: Pass options to logger instead of using format.

`logger.info('Subscription "%s" is chosen', name)`

scripts/azure_cleanup.py
```diff
@@ -36,68 +36,32 @@ def azure_cleanup(nic_template, pip_template, days_old):
...
+
+                        # removing stale stacks
+                        stack_list = mgmt.list_stack(resource_group=resource_group,
+                                                     days_old=days_old)
+                        if stack_list:
+                            removed_stacks = []
+                            for stack in stack_list:
+                                if mgmt.is_stack_empty(stack, resource_group=resource_group):
+                                    removed_stacks.append(stack)
+                                    mgmt.delete_stack(stack, resource_group)
```
mshriver: **Required**

Try/except around this delete_stack call to continue the loop and try to remove all the stacks.  Without if a stack delete fails it'll stop the whole cleanup.

scripts/azure_cleanup.py
```diff
@@ -36,68 +36,32 @@ def azure_cleanup(nic_template, pip_template, days_old):
...
+
+                            if not removed_stacks:
+                                logger.info("No empty stacks older '{}' days were found".format(
+                                    days_old))
+                        mgmt.remove_unused_blobs(resource_group)
```
mshriver: **Question**

Why remove all the logging around this call that was there before?

What about the delete_container call from the removed block, I don't see it called anywhere anymore.

scripts/azure_cleanup.py
```diff
@@ -36,68 +36,32 @@ def azure_cleanup(nic_template, pip_template, days_old):
...
-                    logger.info(removed_disks['Managed'])
-
-                if len(removed_disks['Unmanaged']) > 0:
-                    logger.info('Unmanaged blobs:')
-                    logger.info(removed_disks['Unmanaged'])
+                    for resource_group in mgmt.list_resource_groups():
+                        mgmt.logger.info('Checking "{}" resource group:'.format(resource_group))
+
+                        # removing stale nics
+                        mgmt.remove_nics_by_search(nic_template, resource_group)
```
mshriver: **Optional**

Suggest (almost required?) that you capture the returns from these calls, like was done before, and log what was removed.

mshriver:
# Note
This PR contains an update to frozen.txt for a not-yet-released version of wrapanapi. It MUST NOT BE MERGED until wrapanapi #204 is merged and released.

scripts/azure_cleanup.py
lkhomenk: delete_container was used for removal of bootdiag* containers. we had separate method for that. and I added it's usage into remove_unused_blobs method

lkhomenk:
wrapanapi part got merged.

scripts/azure_cleanup.py
mshriver: Addressed, try/except blocks added.

mshriver:
Script has been run/tested with a local install of wrapanapi v2.5.0.  This version isn't available on pypi yet, so I can't merge until it is.

PR #6302, opened 2018-01-02
[1LP][RFR] Adding wait_for_delete() to Container SSL Tests

jawatts:
The test was deleting and attempting to readd the provider before CFME had fully deleted the previous provider, which would cause a failure in adding the new provider. Adding wait_for_delete() method to prevent hitting this case.

  Failures in PRT are not caused by these changes

### DIFF
--- a/cfme/tests/containers/test_ssl_providers.py
+++ b/cfme/tests/containers/test_ssl_providers.py
@@ -72,7 +72,9 @@ def test_add_provider_naming_conventions(provider, appliance, soft_assert):
                 'Containers Providers "' + provider_name + '" was saved')
         except AssertionError:
             soft_assert(False, provider_name + ' wasn\'t added successfully')
-        ContainersProvider.clear_providers()
+        else:
+            new_provider.delete(cancel=False)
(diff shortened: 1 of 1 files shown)


cfme/tests/containers/test_ssl_providers.py
```diff
@@ -72,7 +72,9 @@ def test_add_provider_naming_conventions(provider, appliance, soft_assert):
                 'Containers Providers "' + provider_name + '" was saved')
         except AssertionError:
             soft_assert(False, provider_name + ' wasn\'t added successfully')
-        ContainersProvider.clear_providers()
```
jawatts: Removed clear_providers() in favor of just deleting the provider that was added by the test.

jawatts:
@gshefer

cfme/tests/containers/test_ssl_providers.py
psav: **QUESTION** I think this is OK, but I wonder if we should add a wait to the clear_providers too......

PR #6322, opened 2018-01-03
[1LP][RFR] Added assertionError in crud method

sshveta:
For duplicate dialog test ,error is expected .
SO removed assert_no_error.

### DIFF
--- a/cfme/automate/dialogs/dialog_element.py
+++ b/cfme/automate/dialogs/dialog_element.py
@@ -181,9 +181,10 @@ class ElementCollection(BaseCollection):
             view.fill(element)
             view.ele_save_button.click()
         view.save_button.click()
-        view.flash.assert_no_error()
-        view.flash.assert_message('{} was saved'.
-                                  format(self.parent.tab.dialog.label))
+        try:
+            view.flash.assert_no_error()
+        except AssertionError:
+            raise
         return self.instantiate(element_data=element_data)
 
     def set_element_type(self, view, element):
--- a/cfme/tests/automate/test_service_dialog.py
+++ b/cfme/tests/automate/test_service_dialog.py
@@ -4,7 +4,8 @@ import fauxfactory
 
 from cfme import test_requirements
 from cfme.utils.update import update
-from cfme.utils import version, error
+from cfme.utils.appliance.implementations.ui import navigator
+from cfme.utils import version
 
 pytestmark = [
     pytest.mark.ignore_stream("upstream"),
@@ -22,8 +23,8 @@ def create_dialog(appliance, element_data, label=None):
                          tab_desc="my tab desc")
     box = tab.boxes.create(box_label='box_' + fauxfactory.gen_alphanumeric(),
                            box_desc="my box desc")
-    box.elements.create(element_data=[element_data])
-    return sd
+    element = box.elements.create(element_data=[element_data])
+    return sd, element
 
 
 def test_crud_service_dialog(appliance):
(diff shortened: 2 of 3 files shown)


sshveta:
One failure in 5.8 is sporadic timeout waiting failure .

cfme/automate/dialogs/dialog_element.py
```diff
@@ -181,9 +182,17 @@ def create(self, element_data):
             view.fill(element)
             view.ele_save_button.click()
         view.save_button.click()
-        view.flash.assert_no_error()
-        view.flash.assert_message('{} was saved'.
-                                  format(self.parent.tab.dialog.label))
+        error_message = version.pick({
```
quarckster: **REQUIRED**
```python
if self.appliance.version < "5.9":
    error_message = ...
else:
    error_message = ...
```

cfme/tests/automate/test_service_dialog.py
```diff
@@ -59,12 +58,7 @@ def test_service_dialog_duplicate_name(appliance, request):
     label = 'duplicate_' + fauxfactory.gen_alphanumeric()
     dialog = create_dialog(appliance, element_data, label=label)
     request.addfinalizer(dialog.delete_if_exists)
-    error_message = version.pick({
-        '5.8': 'Validation failed: Label is not unique within region 0',
-        '5.9': 'There was an error editing this dialog: '
-               'Failed to create a new dialog - Validation failed: '
-               'Name is not unique within region 0'})
-    with error.expected(error_message):
+    with pytest.raises(AssertionError):
```
mshriver: @sshveta This is sufficient - but consider actually asserting the flash message here. You know what view is expected, and can create that view object to assert against its flash message widget.

cfme/automate/dialogs/dialog_element.py
mshriver: I'm not seeing the benefit of including the expected flash message text in the `AssertionError` - if anything I would think the content of the UI flash message should be included, so whatever called create() that resulted in a flash error message could assert against the content of the exception.

cfme/automate/dialogs/dialog_element.py
```diff
@@ -181,9 +182,17 @@ def create(self, element_data):
...
+            '5.8': 'Validation failed: Label is not unique within region 0',
+            '5.9': 'There was an error editing this dialog: '
+                   'Failed to create a new dialog - Validation failed: '
+                   'Name is not unique within region 0'})
+        if view.is_displayed:
```
mshriver: Isn't view the 'Add' destination view class here, and after a save it would be redirected to a different screen?  I'm thinking the flash message asserts should be in a `try|except` clause instead of dependent on the view.

If save was successful, the flash message assertions will not raise exception and the method can continue. If not, the flash message assert will raise the AssertionError anyway.

dajoRH:
Lint report for commit e8513da6b1694358fc2d5a835a45261d5590e2f6:

`cfme/tests/automate/test_service_dialog.py`:
- :bangbang: Line 49:1: **E303** *too many blank lines (3)*

**No commit flaws detected**.

*CFME QE Bot*

cfme/automate/dialogs/dialog_element.py
```diff
@@ -181,9 +181,12 @@ def create(self, element_data):
             view.fill(element)
             view.ele_save_button.click()
         view.save_button.click()
-        view.flash.assert_no_error()
-        view.flash.assert_message('{} was saved'.
```
psav: **REQUIRED** We can remove the specific message match here as you have it in the test below right?

mshriver: @psav had signed off, talked with @sshveta about this, she's adding the positive case flash message assertion to the existing CRUD test, and removing here. Also fixing instances where create_dialog returns a tuple and the caller doesn't unpack.

cfme/tests/automate/test_service_dialog.py
```diff
@@ -38,7 +39,10 @@ def test_crud_service_dialog(appliance):
             'default_text_box': "Default text"
         }
     }
-    dialog = create_dialog(appliance, element_data)
+
+    dialog, element = create_dialog(appliance, element_data)
```
mshriver: :+1:

cfme/automate/dialogs/dialog_element.py
mshriver: :+1:

mshriver:
New updates look good

PR #6327, opened 2018-01-04
[1LP][RFR] Fixing deploy_template to get small_template name properly

jan-zmeskal:
__Fixing__ deploy_template method so it is consistent with new way of defining templates:

templates:
 - small_template
    - name: my_template
    - creds: my_creds

The old way is kept for backwards compatibility.

I am getting `INTERNALERROR> Exception: SSH is unavailable` for 5.9 in PRT. However, I tested this locally and the change is pretty straightforward.

{{pytest: cfme/tests/infrastructure/test_vm_power_control.py::test_no_power_controls_on_archived_vm --use-provider rhv41-vv --long-running}}

### DIFF
--- a/cfme/utils/virtual_machines.py
+++ b/cfme/utils/virtual_machines.py
@@ -58,11 +58,8 @@ def deploy_template(provider_key, vm_name, template_name=None, timeout=900, **de
         try:
             deploy_args.update(template=provider_crud.data['templates']['small_template']['name'])
         except KeyError:
-            try:
-                deploy_args.update(template=provider_crud.data['small_template'])
-            except KeyError:
-                raise ValueError('small_template not defined for Provider {} in cfme_data.yaml'
-                    .format(provider_key))
+            raise KeyError('small_template not defined for Provider {} in cfme_data.yaml'
+                .format(provider_key))
     else:
         deploy_args.update(template=template_name)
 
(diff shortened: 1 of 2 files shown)


cfme/utils/virtual_machines.py
```diff
@@ -56,10 +56,13 @@ def deploy_template(provider_key, vm_name, template_name=None, timeout=900, **de
...
-            deploy_args.update(template=provider_crud.data['small_template'])
+            deploy_args.update(template=provider_crud.data['templates']['small_template']['name'])
         except KeyError:
-            raise ValueError('small_template not defined for Provider {} in cfme_data.yaml'.format(
-                provider_key))
+            try:
+                deploy_args.update(template=provider_crud.data['small_template'])
+            except KeyError:
+                raise ValueError('small_template not defined for Provider {} in cfme_data.yaml'
+                    .format(provider_key))
```
quarckster: **REQUIRED**
It's better to fix yaml than the code.

mshriver: I think we can drop it, and fix the KeyError either in yaml or in the specific test.

mshriver: @jan-zmeskal Please drop this 2nd try, and just raise the ValueError directly.

PR #6328, opened 2018-01-04
[RFR] Fixed test_alert_timeline_cpu

quarckster:
{{pytest: -v -k test_alert_timeline_cpu --long-running --use-provider vsphere6-nested}}

### DIFF
--- a/cfme/tests/control/test_alerts.py
+++ b/cfme/tests/control/test_alerts.py
@@ -341,7 +341,7 @@ def test_alert_timeline_cpu(request, vm, set_performance_capture_threshold, prov
     timeline = vm.open_timelines()
     timeline.filter.fill({
         "event_category": "Alarm/Status Change/Errors",
-        "time_range": "Days",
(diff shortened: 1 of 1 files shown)


cfme/tests/control/test_alerts.py
```diff
@@ -341,7 +341,7 @@ def test_alert_timeline_cpu(request, vm, set_performance_capture_threshold, prov
     timeline = vm.open_timelines()
     timeline.filter.fill({
         "event_category": "Alarm/Status Change/Errors",
-        "time_range": "Days",
+        "time_range": "Weeks",
```
rlbabyuk: **Question**:  Why we need to change it to 'Weeks'?

quarckster: @rlbabyuk No idea, when I was refactoring this test last time it worked with "Days". I guess timelines behavior has changed and now in this test the event can be found only with "Weeks".

PR #6292, opened 2017-12-29
[1LP][RFR] Fix some views elements

rlbabyuk:
Purpose or Intent
=================
{{pytest: -k 'test_add_button or test_filter_save_and_load' --use-provider complete}}

PR #6333, opened 2018-01-04
[RFR]Move button name into class

rlbabyuk:
{{pytest: -k 'test_add_button'}}


mshriver:
Followup to #6292 to move VersionPick definition out of module

PR #6336, opened 2018-01-05
[1LP][RFR] Fixed test_permission_edit

nzr-zts:
Purpose 
=================
Refactored test, added new action to view VMs insted of provisioning VM as there is  a separate test for provisioning.

{{pytest: -v cfme/tests/configure/test_access_control.py -k test_permission_edit}}

### DIFF
--- a/cfme/tests/configure/test_access_control.py
+++ b/cfme/tests/configure/test_access_control.py
@@ -35,7 +35,12 @@ def a_provider(request):
 
 
 def new_credential():
-    return Credential(principal='uid{}'.format(fauxfactory.gen_alphanumeric()), secret='redhat')
+    if BZ(1487199, forced_streams=['5.8']).blocks:
+        return Credential(principal='uid{}'.format(fauxfactory.gen_alphanumeric().lower()),
+                          secret='redhat')
+    else:
+        return Credential(principal='uid{}'.format(fauxfactory.gen_alphanumeric()),
+                          secret='redhat')
 
 
 def new_user(appliance, group, name=None, credential=None):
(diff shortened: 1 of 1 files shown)


cfme/tests/configure/test_access_control.py
```diff
@@ -578,11 +586,9 @@ def _test_vm_removal():
...
 @pytest.mark.parametrize(
     'product_features, action',
-    [({version.LOWEST: [
-        ['Everything', 'Compute', 'Infrastructure', 'Virtual Machines', 'Accordions'],
-        ['Everything', 'Access Rules for all Virtual Machines', 'VM Access Rules', 'Modify',
-         'Provision VMs']], },
-      _test_vm_provision)])
+    [([['Everything', 'Access Rules for all Virtual Machines', 'VM Access Rules', 'View'],
+       ['Everything', 'Compute', 'Infrastructure', 'Virtual Machines', 'Accordions']],
+      _test_vm_view)])
```
mshriver: Since this action is the same method in all cases, why parametrize it? Can't you call it directly instead of this obfuscated `action`?

cfme/tests/configure/test_access_control.py
```diff
@@ -558,6 +562,10 @@ def _test_vm_provision(appliance):
     vms.lcl_btn("Provision VMs")
 
 
+def _test_vm_view(appliance):
+    logger.info("Checking for vms view access")
```
psav: **REQUIRED**: ```appliance``` isn't used in this function. Are we requiring it because soon we will need to use vms collections which will require an appliance?

nzr-zts: It wasn't required, updated to use navigate_to in test instead of external function.

cfme/tests/configure/test_access_control.py
nzr-zts: Updated to use directly function instead of parametrization

PR #6350, opened 2018-01-05
[RFR] Fixing bug in manage_provider skipper

psav:
* During dev, a line was misintroduced, it should have been the one
  here.

### DIFF
--- a/fixtures/log.py
+++ b/fixtures/log.py
@@ -37,7 +37,7 @@ def pytest_runtest_logreport(report):
     #      test_tracking['test_name']['teardown'] = 'failed'
     yield
     test_tracking[_format_nodeid(report.nodeid, False)][report.when] = report.outcome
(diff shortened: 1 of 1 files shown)


fixtures/log.py
```diff
@@ -37,7 +37,7 @@ def pytest_runtest_logreport(report):
     #      test_tracking['test_name']['teardown'] = 'failed'
     yield
     test_tracking[_format_nodeid(report.nodeid, False)][report.when] = report.outcome
-    if report.when == 'teardown' and pytest.store.parallelizer_role == 'master':
+    if report.when == 'teardown' and pytest.store._maybe_get_plugin('parallel_session') is None:
```
mshriver: https://github.com/ManageIQ/integration_tests/blob/master/fixtures/pytest_store.py#L135

property available

PR #6353, opened 2018-01-05
[1LP][RFR] Implement physical server navigation steps

skovic:
This PR implements the navigation steps for physical server views.
 
This PR depends on ManageIQ/integration_tests/pull/5868


quarckster:
@skovic  Is it in progress [WIP] or ready for review [RFR]?

skovic:
@psav @mshriver When you get a chance, could you also take a look at this PR? Thanks

jkrocil:
@skovic Hi, please read this: https://github.com/ManageIQ/integration_tests/wiki/PR-Process

Thank you

skovic:
I believe that the Travis tests are failing because ManageIQ/integration_tests/pull/5868 has not been merged yet.

mshriver:
WIP until the dependent views are merged.

mshriver:
@skovic Likely won't be able to review today, self-assigned though so it'll stay visible for me.

cfme/physical/physical_server.py
```diff
@@ -0,0 +1,238 @@
...
+from navmazing import NavigateToSibling, NavigateToAttribute
+
+from cfme.common import PolicyProfileAssignable, WidgetasticTaggable
+from cfme.common.physical_server_views import (
+    PhysicalServerDetailsView,
+    PhysicalServerManagePoliciesView,
+    PhysicalServersView,
+    PhysicalServerTimelinesView
+)
+
```
quarckster: **REQUIRED**
This blank line is not needed.

cfme/physical/physical_server.py
```diff
@@ -0,0 +1,238 @@
...
+from cfme.utils.appliance.implementations.ui import CFMENavigateStep, navigate_to, navigator
+from cfme.utils.log import logger
+from cfme.utils.pretty import Pretty
+from cfme.utils.update import Updateable
+from cfme.utils.wait import wait_for
+
+
+@attr.s
+class PhysicalServer(BaseEntity, Updateable, Pretty, PolicyProfileAssignable, WidgetasticTaggable):
+    """Model of an Physical Server in cfme.
```
quarckster: Trolling is a art 😃

skovic:
@quarckster @mshriver I've made the recommended changes. Please let me know if anything else needs to be changed. Thanks

mshriver:
@skovic thanks for making updates, that's great. The title still containing WIP is what's keeping the PR from moving - WIPTEST at least so that PRT runs its checks, RFR for when you're ready for re-review.  Updated the title this time for you :)

cfme/physical/physical_server.py
```diff
@@ -0,0 +1,232 @@
...
+        view = navigate_to(self.parent, "All")
+        try:
+            view.entities.get_entity(name=self.name, surf_pages=True)
+        except ItemNotFound:
+            return False
+        else:
+            return True
+
+    def get_db_id(self):
+        if self.db_id is None:
```
mshriver: This definitely looks like it would benefit from use of cached_property.

cfme/physical/physical_server.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+@attr.s
+class PhysicalServerCollection(BaseCollection):
+    """Collection object for the :py:class:`cfme.infrastructure.host.PhysicalServer`."""
+
+    ENTITY = PhysicalServer
+
+    def check_physical_servers(self, physical_servers):
+        physical_servers = list(physical_servers)
```
mshriver: **Required**
docblock

cfme/physical/physical_server.py
```diff
@@ -0,0 +1,232 @@
...
+        checked_physical_servers = list()
+        view = navigate_to(self, 'All')
+
+        for physical_server in physical_servers:
+            try:
+                view.entities.get_entity(name=physical_server.name, surf_pages=True).check()
+                checked_physical_servers.append(physical_server)
+            except ItemNotFound:
+                raise (ItemNotFound(
+                       'Could not find physical server {} in the UI'.format(physical_server.name)))
```
mshriver: **Question**
Why re-raise the same exception here?  ItemNotFound as raised by `get_entity` contains:
"Entity {keys} isn't found on this page" already, where keys is the passed dictionary of `name=physical_server.name`.  Doesn't appear to add any value to the exception, and the try/except is unnecessary.

cfme/physical/physical_server.py
```diff
@@ -0,0 +1,232 @@
...
+        )
+
+
+@attr.s
+class PhysicalServerCollection(BaseCollection):
+    """Collection object for the :py:class:`cfme.infrastructure.host.PhysicalServer`."""
+
+    ENTITY = PhysicalServer
+
+    def check_physical_servers(self, physical_servers):
```
mshriver: I think we need to consider a consistent collection API here if we're going to include convenience methods like this.

It should be named more abstractly, like `select_entity_rows`.  @psav thoughts on entity selection methods?

cfme/physical/physical_server.py
```diff
@@ -0,0 +1,232 @@
...
+                view.entities.get_entity(name=physical_server.name, surf_pages=True).check()
+                checked_physical_servers.append(physical_server)
+            except ItemNotFound:
+                raise (ItemNotFound(
+                       'Could not find physical server {} in the UI'.format(physical_server.name)))
+        return view
+
+    def all(self, provider):
+        """returning all physical_servers objects"""
+        view = navigate_to(self, 'All')
```
mshriver: This is taking a provider and using it for instantiation of entity objects - but there's nothing that is filtering the view based on that provider. We could have multiple providers configured, and be creating entity objects with the wrong provider assigned.

There is a `get_crud_by_name` method that can be used to fetch a provider object based on its name. The provider name should be pulled from the table along with the item name and passed to this method to create the correct provider object for the given row.

https://github.com/ManageIQ/integration_tests/blob/master/cfme/utils/providers.py#L322

cfme/physical/physical_server.py
```diff
@@ -0,0 +1,232 @@
...
+class PolicyAssignment(CFMENavigateStep):
+    VIEW = PhysicalServerManagePoliciesView
+    prerequisite = NavigateToSibling("Details")
+
+    def step(self):
+        self.prerequisite_view.toolbar.policy.item_select("Manage Policies")
+
+
+@navigator.register(PhysicalServer)
+class Provision(CFMENavigateStep):
```
mshriver: I think we should omit this nav destination until there's a view for it.

cfme/physical/physical_server.py
```diff
@@ -0,0 +1,232 @@
...
+class Details(CFMENavigateStep):
+    VIEW = PhysicalServerDetailsView
+    prerequisite = NavigateToAttribute("parent", "All")
+
+    def step(self):
+        self.prerequisite_view.entities.get_entity(name=self.obj.name, surf_pages=True).click()
+
+
+@navigator.register(PhysicalServer)
+class PolicyAssignment(CFMENavigateStep):
```
mshriver: **Optional**
We don't quite have a consistent definition for this yet, but I'd prefer 'ManagePolicies' for this destination name.

There is some work to be done to PolicyProfileAssignable class, which has a 'ManagePolicies' destination already registered, in order for it to support navigation for classes inheriting it. Its step uses 'All' as a prerequisite, without using `parent` as the target object.

cfme/physical/physical_server.py
skovic: @mshriver Could you give an example of where cached_property is used? Thanks

mshriver: Absolutely:
It's a pip project with pretty good simple examples:
https://pypi.python.org/pypi/cached-property

We use it throughout the codebase, with varying degrees of complexity:
https://github.com/ManageIQ/integration_tests/blob/067ec3d197a2afa2bca8ff4a06d4aeec2d9c415d/cfme/infrastructure/config_management.py#L540

The idea being its a property that doesn't really change, but can be invalidated when need be.

cfme/physical/physical_server.py
skovic: @mshriver So I just need to rename this class to 'ManagePolicies'?

mshriver: If you're not too attached to `PolicyAssignment` - easier now while its not in-use anywhere

dajoRH:
Lint report for commit 3b92aa36ef43601d00858898da532eb7d3db90ae:

`cfme/physical/physical_server.py`:
- :bangbang: Line 195:5: **E303** *too many blank lines (2)*

**No commit flaws detected**.

*CFME QE Bot*

skovic:
@mshriver I believe I addressed all of your review comments. Please let me know if I missed anything. Thanks

mshriver:
@psav you have a physical server provider to test on?

PR #6354, opened 2018-01-06
[1LP][RFR] Update Chargeback Rates, views, navigation, and CRUD tests

mshriver:
Some refactoring of the chargeback rates views and CRUD methods, moving flash message assertions into the tests.

Only parametrized one test to highlight a BZ blocker, @nachandr will be following up with more parametrization of the test case, reducing the separate create/edit/delete tests to a single function.

{{ pytest: cfme/tests/intelligence/test_chargeback.py --use-provider vsphere65-nested --long-running }} 
  
# PRT Results
The test_validate_chargeback_reports is currently blocked from BZ's.

Test skips from blockers, passing otherwise.

### DIFF
--- a/cfme/intelligence/chargeback/rates.py
+++ b/cfme/intelligence/chargeback/rates.py
@@ -1,51 +1,68 @@
 # -*- coding: utf-8 -*-
 # Page model for Intel->Chargeback->Rates.
+from wait_for import TimedOutError
 
 from cached_property import cached_property
 from navmazing import NavigateToSibling, NavigateToAttribute
 from widgetastic.utils import ParametrizedLocator, ParametrizedString
-from widgetastic.widget import Text, ParametrizedView
+from widgetastic.widget import Text, ParametrizedView, View
 from widgetastic_patternfly import Button, Input, Dropdown
 
+from cfme.exceptions import ChargebackRateNotFound
 from cfme.utils import ParamClassName
 from cfme.utils.appliance import Navigatable
 from cfme.utils.appliance.implementations.ui import navigator, navigate_to, CFMENavigateStep
+from cfme.utils.blockers import BZ
 from cfme.utils.pretty import Pretty
 from cfme.utils.update import Updateable
-from widgetastic_manageiq import Select
+from widgetastic_manageiq import Select, Table
 from . import ChargebackView
 
 
 class RatesView(ChargebackView):
     title = Text("#explorer_title_text")
+    table = Table(".//div[@id='records_div' or @class='miq-data-table']/table")
+
+    @property
+    def in_rates(self):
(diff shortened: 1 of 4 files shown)


cfme/intelligence/chargeback/rates.py
```diff
@@ -198,8 +198,13 @@ def update(self, updates):
...
     def delete(self):
         # Delete a rate in UI
-        view = navigate_to(self, 'Details')
-        view.configuration.item_select('Remove from the VMDB', handle_alert=True)
+        view = navigate_to(self, 'Details', wait_for_view=True)
+        assert view.is_displayed
+        view.configuration.item_select(
+            VersionPick({
+                Version.lowest(): 'Remove from the VMDB',
+                '5.9': 'Remove selected Chargeback Rates from the VMDB'}),
```
mshriver: This may not be relevant - the configuration dropdown wasn't where it should be for the test, and it should be on details

mshriver:
@nachandr We can, but that changes a few things that I didn't want to address within the scope of this PR, and that should be done with the CRUD test refactor to a single parametrized function.

I would prefer that the fixture not call the create method, so that the CRUD test is executing the create during test execution, and not during test setup.

Your suggested fixture structure would setup for a test where a single fixture is providing the two different chargeback rate objects. This means the test isn't parametrized on rate types, and gets both storage and compute objects.  
This creates dependencies within the test between the success of the CRUD actions on the two objects.  We would have to `soft_assert()` most calls so that a failure in the first create call doesn't block the second create.  
After this all of the `update()`, `copy()`, and `delete()` calls would have to be dependent on if the object actually exists, or would have to be diapered in a try block for the various exceptions that could occur during those calls. If we're not having to `soft_assert()` because we're only dealing with a single rate type, then an exception would halt the test phase, and start cleanup - with no need to catch it.

Check it out as you simplify the test to a single CRUD function, and see how the failures are reported when something goes wrong, and how those failures would interact with each other in a  single function scope.

nachandr:
@mshriver, I'd like to clarify that I was asking you to make the suggested changes to the new_compute_rate() fixture in test_validate_chargeback_report.py. The fixture does create 2 objects.I  agree on the fact that it creates dependencies within the test between the success of the CRUD actions on the two objects. But, as discussed, I do intend to  refactor the CRUD tests in test_chargeback.py  to a single parametrized function. 

Since the CRUD tests are in test_chargeback.py and test_validate_chargeback_report.py is more about validating Chargeback costs than about testing CRUD operations on Chargeback rates, I thought it would be good to unblock 4 of the custom tests and block just  test_validate_custom_rate_storage_usage_cost .

I'll submit a PR for unblocking all test_validate_custom_rate* tests except  test_validate_custom_rate_storage_usage_cost.

PR #6357, opened 2018-01-07
[1LP][RFR] Fix test_relationships

gshefer:
{{ pytest: cfme/tests/containers/test_relationships.py --use-provider ocp-v1 }}
- Fix test_relationships
- Fix has_persistent_volume fixture to create PV only if such doesn't exist

### DIFF
--- a/cfme/tests/containers/test_relationships.py
+++ b/cfme/tests/containers/test_relationships.py
@@ -7,7 +7,6 @@ from cfme.containers.provider import ContainersProvider, ContainersTestItem
 from cfme.containers.service import Service, ServiceCollection
 from cfme.containers.replicator import Replicator, ReplicatorCollection
 from cfme.containers.image import Image, ImageCollection
-from cfme.containers.project import Project, ProjectCollection
 from cfme.containers.template import Template, TemplateCollection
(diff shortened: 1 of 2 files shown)


cfme/tests/containers/test_relationships.py
```diff
@@ -70,18 +72,20 @@ def test_relationships_tables(provider, has_persistent_volume, appliance, test_i
     # Check the relationships linking & data integrity
     view = navigate_to(instance, 'Details')
     relations = [key for key, val in view.entities.relationships.read().items() if val != '0']
+    if current_version().is_in_series([5, 8]):  # Treat BZ#1518862
+        relations = filter(lambda rel: rel != 'Containers', relations)
```
quarckster: **REQUIRED**
```python
def test_relationships_tables(appliance, provider, has_persistent_volume, appliance, test_item):
...
    if appliance.version.is_in_series([5, 8]):  # Treat BZ#1518862
        relations = filter(lambda rel: rel != 'Containers', relations)
```

gshefer: What is required?

gshefer: Oh yeah got it :-)

PR #6365, opened 2018-01-08
[1LP][RFR] Fixed test_add_vm_to_service and removed vm hardcoding

Yadnyawalkya:
{{pytest: -v cfme/tests/services/test_add_remove_vm_to_service.py -k 'test_add_vm_to_service' --long-running --use-provider vsphere55}}
 
**What was the issue?**
VM name `auto_test_services` hardcoded in earlier code which was requiring VM to be present on every infra provider, I am creating and deleting new VM instead of that.

@sshveta: What say?

### DIFF
--- a/cfme/tests/services/test_add_remove_vm_to_service.py
+++ b/cfme/tests/services/test_add_remove_vm_to_service.py
@@ -5,6 +5,7 @@ from cfme import test_requirements
 from cfme.automate.explorer.domain import DomainCollection
 from cfme.automate.simulation import simulate
 from cfme.common.provider import cleanup_vm
+from cfme.common.vm import VM
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.services.service_catalogs import ServiceCatalogs
 from cfme.services.myservice import MyService
@@ -22,6 +23,18 @@ pytestmark = [
     pytest.mark.tier(3),
     pytest.mark.provider([VMwareProvider], scope="module"),
 ]
+
+
+@pytest.yield_fixture(scope='function')
+def new_vm(provider, setup_provider, small_template_modscope):
+    """Fixture to provision and delete vm on the provider"""
+    vm_name = 'test_service_{}'.format(fauxfactory.gen_alphanumeric())
+    vm = VM.factory(vm_name, provider, small_template_modscope.name)
+    vm.create_on_provider(find_in_cfme=True, timeout=700, allow_skip="default")
+    yield vm
+    if provider.mgmt.does_vm_exist(vm.name):
+        provider.mgmt.delete_vm(vm.name)
+    provider.refresh_provider_relationships()
 
 
 @pytest.fixture(scope="function")
@@ -59,7 +72,7 @@ def myservice(appliance, setup_provider, provider, catalog_item, request):
 
 
 @pytest.mark.ignore_stream("upstream")
-def test_add_vm_to_service(myservice, request, copy_domain):
+def test_add_vm_to_service(myservice, request, copy_domain, new_vm):
     """Tests adding vm to service
 
     Metadata:
@@ -95,7 +108,7 @@ def test_add_vm_to_service(myservice, request, copy_domain):
         message="create",
         request=method.name,
         target_type='VM and Instance',
-        target_object="auto_test_services",
+        target_object=new_vm.name,
         execute_methods=True
     )
-    myservice.check_vm_add("auto_test_services")
+    myservice.check_vm_add(new_vm.name)


cfme/tests/services/test_add_remove_vm_to_service.py
```diff
@@ -90,12 +92,16 @@ def add_to_service
         .methods.create(name='InspectMe', location='inline', script=method_torso)
 
     request.addfinalizer(method.delete_if_exists)
+    vm = VM.factory(random_vm_name(), provider, small_template_modscope.name)
+    vm.create_on_provider(find_in_cfme=True, timeout=700, allow_skip="default")
```
sshveta: 👍 LGTM

widgetastic_manageiq/__init__.py
```diff
@@ -2652,9 +2652,9 @@ class BaseQuadIconEntity(ParametrizedView, ClickableMixin):
 
     """
     PARAMETERS = ('entity_id',)
-    ROOT = ParametrizedLocator('.//table[./tbody/tr/td/*[(self::a or self::span) and '
-                               'substring(@href, string-length(@href)'
-                               '-string-length("/{entity_id}")+1)="/{entity_id}"]]')
+    ROOT = ParametrizedLocator(
+        './/table[./tbody/tr/td/*[(self::a or self::span) and substring(@href, string-length(@href)'
+        '-string-length("/{entity_id}?link=true")+1)="/{entity_id}?link=true"]]')
```
izapolsk: Q: are you sure that ?link=true is always present ?

Yadnyawalkya: URL for 5.8 changed to `https://10.10.10.10/vm_infra/show/86?link=true` whereas 5.9 not handle URL implementation in that way. You are correct, should I add this xpath in `or` then? @izapolsk 

 screenshot from 2018-02-12 14-23-28

dajoRH:
*CFME QE Bot*

cfme/tests/services/test_add_remove_vm_to_service.py
```diff
@@ -90,12 +92,16 @@ def add_to_service
...
+    request.addfinalizer(vm.delete_from_provider)
```
quarckster: **REQUIRED**
I would move vm creation to a fixture.

widgetastic_manageiq/__init__.py
```diff
@@ -3576,7 +3576,7 @@ def _current_page_elements(self):
         if self.browser.product_version < '5.9':
             br = self.browser
             for el in br.elements(self.elements):
-                el_id = int(br.get_attribute('href', el).split('/')[-1])
+                el_id = br.get_attribute('href', el).split('/')[-1]
```
quarckster: **QUESTION**
Why is it considered as a string?

izapolsk: because all entities have string id in non 0 regions. It looks like 1r0001

cfme/tests/services/test_add_remove_vm_to_service.py
Yadnyawalkya: Okay let me try this one

izapolsk: I'm not sure this is required since this file has only one test

widgetastic_manageiq/__init__.py
Yadnyawalkya: @quarckster: 5.8's URL pattern have change in for quadicon links, `?link=true` at last part, check this (**Reference:** https://github.com/ManageIQ/integration_tests/commit/73f4b629884265d36c4bae5c5879fd75aa4d41b9)

screenshot from 2018-02-16 14-23-21

cfme/tests/services/test_add_remove_vm_to_service.py
quarckster: I believe that in the test we should only test, and other prerequisite stuff should moved to fixtures.

widgetastic_manageiq/__init__.py
izapolsk: hm, but ?link isn't part of id. everything after id has to be removed then

Yadnyawalkya: If we only wanted id(number) there, we have to change
`el_id = int(br.get_attribute('href', el).split('/')[-1]).split('?')[0])`
Note `.split('?')[0])` there. but then it fails in `if item(entity_id=self.entity_id).is_displayed:` 

This is where it should failed
```
class NonJSBaseEntity(View):
    quad_entity = BaseQuadIconEntity
  
    def __init__(self, parent, entity_id, name=None, logger=None):
        View.__init__(self, parent, logger=logger)
        self.entity_id = entity_id
        self._name = name

    def _get_existing_entity(self):
        for item in (self.quad_entity, self.tile_entity, self.list_entity):
        >>  if item(entity_id=self.entity_id).is_displayed:
```

This is b'cause `ROOT` of `BaseQuadIconEntity` need to changed so that it should parse `?link=true` from `href` and if you change `ROOT` to new one i.e.
```
    ROOT = ParametrizedLocator('.//table[./tbody/tr/td/*[(self::a or self::span) and '
                               'substring(@href, string-length(@href)'
                               '-string-length("/{entity_id}")+1)="/{entity_id}"]]' or
                           >    './/table[./tbody/tr/td/*[(self::a or self::span) and '
                           >    'substring(@href, string-length(@href)'
                           >   '-string-length("/{entity_id}?link=true")+1)="/{entity_id}?link=true"]]')
```
This all workout okay (I have done this before)! but I personally feel removal of `int()` if better approach than things mentioned above as I see https://github.com/ManageIQ/integration_tests/commit/73f4b629884265d36c4bae5c5879fd75aa4d41b9 what say @quarckster @izapolsk?

izapolsk: I'm looking into this issue and will add my commit with fix in a bit. Please update PR according to @quarckster 's remarks.

izapolsk: @Yadnyawalkya, please use  there
el_id = br.get_attribute('href', el).split('?')[0].split('/')[-1]

dajoRH:
*CFME QE Bot*

dajoRH:
*CFME QE Bot*

cfme/tests/services/test_add_remove_vm_to_service.py
```diff
@@ -24,6 +25,19 @@
 ]
 
 
+@pytest.fixture
+def new_vm(request, provider, setup_provider, small_template_modscope):
```
izapolsk: I believe this fixture will look much better if you turn it into yield fixture

cfme/tests/services/test_add_remove_vm_to_service.py
```diff
@@ -24,6 +25,18 @@
...
+@pytest.yield_fixture(scope='function')
+def new_vm(provider, setup_provider, small_template_modscope):
+    """Fixture to provision and delete vm on the provider"""
+    vm_name = 'test_service_{}'.format(fauxfactory.gen_alphanumeric())
+    vm = VM.factory(vm_name, provider, small_template_modscope.name)
+    if not provider.mgmt.does_vm_exist(vm.name):
+        logger.info("deploying %s on provider %s", vm.name, provider.key)
+        vm.create_on_provider(find_in_cfme=True, timeout=700, allow_skip="default")
+    yield vm
+    vm.delete_from_provider
```
quarckster: **REQUIRED**
```python
yield vm
if provider.mgmt.does_vm_exist(vm.name):
    provider.mgmt.delete_vm(vm.name)
```

cfme/tests/services/test_add_remove_vm_to_service.py
```diff
@@ -24,6 +25,18 @@
 ]
 
 
+@pytest.yield_fixture(scope='function')
+def new_vm(provider, setup_provider, small_template_modscope):
+    """Fixture to provision and delete vm on the provider"""
+    vm_name = 'test_service_{}'.format(fauxfactory.gen_alphanumeric())
+    vm = VM.factory(vm_name, provider, small_template_modscope.name)
+    if not provider.mgmt.does_vm_exist(vm.name):
```
quarckster: **REQUIRED**
I don't think this check is required.

izapolsk: why do you check this ?

PR #6368, opened 2018-01-08
[1LP][RFR] Fix to rhsm/sat6 registration to overcome bug issue

lcouzens:
Required wait before running register.


lcouzens:
time.sleep was added due to a backend issue we need to wait on.

PR #6373, opened 2018-01-08
[1LP][RFR] A few fixes for Sauce run

psav:
Some fixes


quarckster:
PRT failed on test_server_name

cfme/tests/configure/test_session_timeout.py
```diff
@@ -26,7 +26,7 @@ def _finalize():
     # Try getting timeout
     # I had to use wait_for because on 5.4 and upstream builds it made weird errors
     wait_for(
-        lambda: appliance.browser.widgetastic.selenium.elements(
```
jamesooden: Did the api change here?   I just tried to look up the old call and could not find it in the current documentation.   I'm just curious.   Clearly the new call does exactly what it says, so its even more readable.

PR #6374, opened 2018-01-09
[1LP][RFR] Added partial match for updated values in 5.9

sshveta:
{{pytest: -v --long-running --use-provider ec2 cfme/tests/services/test_cloud_service_catalogs.py }} 

While provisioning values in dropdown of security_group and instance_type are different for 5.8 and 5.8.
Ex: for 5.8 instance_type shows value "m1.tiny" and for 5.9 value displayed is "m1.tiny :1 CPU ..." 
So by adding partial match to the value only m1.tiny is checked in the dropdown value and then selected.

### DIFF
--- a/cfme/tests/services/test_cloud_service_catalogs.py
+++ b/cfme/tests/services/test_cloud_service_catalogs.py
@@ -2,6 +2,8 @@
 import fauxfactory
 import pytest
 
+from widgetastic.utils import partial_match
+
 from cfme.common.provider import cleanup_vm
 from cfme.cloud.provider import CloudProvider
 from cfme.services.catalogs.catalog_item import CatalogItem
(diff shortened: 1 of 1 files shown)


quarckster:
PRT hasn't passed.

jamesooden:
Could you add to the commit description why we wanted to make this change?    It looks straight forward, I am just not aware of what problem it solved.   Thanks

cfme/tests/services/test_cloud_service_catalogs.py
```diff
@@ -36,27 +39,31 @@ def test_cloud_catalog_item(appliance, setup_provider, provider, dialog, catalog
     provisioning_data = {
         'catalog': {'vm_name': vm_name,
                     },
-        'properties': {'instance_type': provisioning['instance_type'],
+        'properties': {'instance_type': partial_match(provisioning['instance_type']),
                        },
-        'environment': {'security_groups': [provisioning['security_group']],
-                        },
+        'environment': {'security_groups': partial_match(provisioning['security_group']),
```
vatsalparekh: I think gce was not checked here, which is breaking currently, I'm not seeing `security_group` option in service catalog with `gce_central` provider

mshriver: There was no change here to which providers this applies to, so if it was breaking for GCE on including security_group it must have been failing before this PR as well.

PR #6379, opened 2018-01-09
[1LP][RFR] Buttons access AVP correctly

mfalesni:


cfme/automate/buttons.py
```diff
@@ -281,6 +281,10 @@ class attribute(ParametrizedView):  # noqa
             key = Input(name=ParametrizedString('attribute_{number}'))
             value = Input(name=ParametrizedString('value_{number}'))
 
+            @classmethod
+            def all(cls, browser):
```
jamesooden: Should there be a comment here describing what the use of this function is.   It looks like it just returns a list of numbers 1 through 6.    I'm sure I'm missing something here, but looking at it I can't see why I would want to call it.

mfalesni: Look into https://pypi.python.org/pypi/widgetastic.core#parametrized-views

mfalesni: Also, here is documentation of the method itself: https://github.com/RedHatQE/widgetastic.core/blob/master/src/widgetastic/widget.py#L805

mfalesni: There are always and only 5 Attribute-Value field pairs, numbered from 1, so therefore it generates `[(1, ), (2, ), (3, ), (4, ), (5, )]` as values for PARAMETERS so it can be accessed as a list or read.

PR #6380, opened 2018-01-09
[1LP][RFR] Fix wrong widgetastic browser reference

mfalesni:


mfalesni:
now I have no idea why it does not work in 5.9. Looks they changed it so it no longer works this way? Anayway the issue this fixes is fixed, you can see it works on 5.8 fine.

Issue #6409, opened 2018-01-11
Updates Required For cfme.tests.containers.test_ssl_providers


jawatts:
cfme.tests.containers.test_ssl_providers are failing due to not having the correct certificate on the CFME appliance. A solution to put the OCP certificate on the CFME appliance is needed. 

@shalomnaim1 is taking the lead to implemented this solution.

shalomnaim1:
I wrote a fixture for sync the certifications, you welcome to take a look at it --> https://github.com/ManageIQ/integration_tests/pull/6432

shalomnaim1:
@jawatts I don't think this necessary because I am removing this line in my PR, I think that enough for PRT to run

PR #6410, opened 2018-01-11
[1LP][RFR] Blocking test_ssl_providers

jawatts:
Block test_ssl_providers until https://github.com/ManageIQ/integration_tests/issues/6409 is resolved. 


jawatts:
@mshriver

PR #6418, opened 2018-01-12
[1LP][RFR][NOTEST] Removed test_drag_and_drop_widget_to_the_bottom_of_another_column

quarckster:
Purpose
=================

This test never worked. After closing [BZ1316134](https://bugzilla.redhat.com/show_bug.cgi?id=1316134) another one was opened [BZ1485637](https://bugzilla.redhat.com/show_bug.cgi?id=1485637). But it does not fix the problem, so this will always fail.

### DIFF
--- a/cfme/tests/intelligence/test_dashboard.py
+++ b/cfme/tests/intelligence/test_dashboard.py
@@ -8,7 +8,6 @@ from random import sample
 
 from cfme import test_requirements
 from cfme.intelligence.reports.dashboards import Dashboard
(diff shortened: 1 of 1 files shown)


rlbabyuk:
**Question**:  1511524 seems to be fixed for 5.9, Is it possible to update test after this fix?

quarckster:
@rlbabyuk it's not a fix, but some sort of workaround. The test doesn't work anyway.

PR #6432, opened 2018-01-14
[1LP][RFR] Add a fixture for sync SSL certificate between CFME and OCP

### LINKED ISSUE #6409: Updates Required For cfme.tests.containers.test_ssl_providers

shalomnaim1:
{{ pytest: cfme/tests/containers/test_ssl_providers.py::test_add_provider_ssl --use-provider ocp-v1 }}

## summery
In the following PR I will create a new fixture for sync SSL certification between OCP and CFME

**3 steps required for sync the certificates:**
1. Get the certificate from OCP
1. Deploy the certificate to CFME
1. Run "update-ca-trust" to actually install the new certificate

FIXES #6409

### DIFF
--- a/cfme/containers/provider/openshift.py
+++ b/cfme/containers/provider/openshift.py
@@ -8,11 +8,11 @@ from . import ContainersProvider
 from cfme.containers.provider import (
     ContainersProviderDefaultEndpoint, ContainersProviderEndpointsForm
 )
+
 from cfme.common.provider import DefaultEndpoint
 from cfme.utils.ocp_cli import OcpCli
-from cfme.utils.path import data_path
 from cfme.utils.varmeth import variable
-from cfme.utils import version
+from cfme.utils import version, ssh
 
 
 class CustomAttribute(object):
@@ -25,17 +25,19 @@ class CustomAttribute(object):
 
 class OpenshiftDefaultEndpoint(ContainersProviderDefaultEndpoint):
     """Represents Openshift default endpoint"""
+
     @staticmethod
-    def get_ca_cert():
+    def get_ca_cert(connection_info):
         """Getting OpenShift's certificate from the master machine.
         Args:
-           No args.
+            connection_info (dict): username, password and hostname for OCP
         returns:
             certificate's content.
         """
-        cert_file_path = path.join(str(data_path), 'cert-auths', 'cmqe-tests-openshift-signer.crt')
-        with open(cert_file_path) as f:
-            return f.read()
+
+        with ssh.SSHClient(**connection_info) as provider_ssh:
+            _, stdout, _ = provider_ssh.exec_command("cat /etc/origin/master/ca.crt")
+            return str("".join(stdout.readlines()))
 
 
 class HawkularEndpoint(DefaultEndpoint):
@@ -51,7 +53,10 @@ class HawkularEndpoint(DefaultEndpoint):
         }
 
         if out['sec_protocol'] and self.sec_protocol.lower() == 'ssl trusting custom ca':
-            out['trusted_ca_certificates'] = OpenshiftDefaultEndpoint.get_ca_cert()
+            out['trusted_ca_certificates'] = OpenshiftDefaultEndpoint.get_ca_cert(
+                {"username": "root",
+                 "password": self.ssh_password,
+                 "hostname": self.master_hostname})
 
         return out
 
@@ -76,7 +81,10 @@ class AlertsEndpoint(DefaultEndpoint):
             '5.9': self.sec_protocol})
 
         if out['sec_protocol'] and self.sec_protocol.lower() == 'ssl trusting custom ca':
-            out['trusted_ca_certificates'] = OpenshiftDefaultEndpoint.get_ca_cert()
+            out['trusted_ca_certificates'] = OpenshiftDefaultEndpoint.get_ca_cert(
+                {"username": "root",
+                 "password": self.ssh_password,
+                 "hostname": self.master_hostname})
 
         return out
 
@@ -192,7 +200,17 @@ class OpenshiftProvider(ContainersProvider):
 
         endpoints = {}
         token_creds = cls.process_credential_yaml_key(prov_config['credentials'], cred_type='token')
+
+        master_hostname = prov_config['endpoints']['default'].hostname
+        ssh_password = cls.process_credential_yaml_key(prov_config['credentials'],
+                                                       cred_type='ssh').secret
+
         for endp in prov_config['endpoints']:
+            # Add ssh_password for each endpoint, so get_ca_cert
+            # will be able to get SSL cert form OCP for each endpoint
+            setattr(prov_config['endpoints'][endp], "master_hostname", master_hostname)
+            setattr(prov_config['endpoints'][endp], "ssh_password", ssh_password)
+
             if OpenshiftDefaultEndpoint.name == endp:
                 prov_config['endpoints'][endp]['token'] = token_creds.token
                 endpoints[endp] = OpenshiftDefaultEndpoint(**prov_config['endpoints'][endp])
--- a/cfme/containers/provider/openshift.py
+++ b/cfme/containers/provider/openshift.py
@@ -54,8 +54,8 @@ class HawkularEndpoint(DefaultEndpoint):
 
         if out['sec_protocol'] and self.sec_protocol.lower() == 'ssl trusting custom ca':
             out['trusted_ca_certificates'] = OpenshiftDefaultEndpoint.get_ca_cert(
-                {"username": "root",
-                 "password": self.ssh_password,
+                {"username": self.ssh_creds.principal,
+                 "password": self.ssh_creds.secret,
                  "hostname": self.master_hostname})
 
         return out
@@ -82,8 +82,8 @@ class AlertsEndpoint(DefaultEndpoint):
 
         if out['sec_protocol'] and self.sec_protocol.lower() == 'ssl trusting custom ca':
             out['trusted_ca_certificates'] = OpenshiftDefaultEndpoint.get_ca_cert(
-                {"username": "root",
-                 "password": self.ssh_password,
+                {"username": self.ssh_creds.principal,
+                 "password": self.ssh_creds.secret,
                  "hostname": self.master_hostname})
 
         return out
@@ -202,14 +202,13 @@ class OpenshiftProvider(ContainersProvider):
         token_creds = cls.process_credential_yaml_key(prov_config['credentials'], cred_type='token')
 
         master_hostname = prov_config['endpoints']['default'].hostname
-        ssh_password = cls.process_credential_yaml_key(prov_config['credentials'],
-                                                       cred_type='ssh').secret
+        ssh_creds = cls.process_credential_yaml_key(prov_config['ssh_creds'])
 
         for endp in prov_config['endpoints']:
             # Add ssh_password for each endpoint, so get_ca_cert
             # will be able to get SSL cert form OCP for each endpoint
             setattr(prov_config['endpoints'][endp], "master_hostname", master_hostname)
-            setattr(prov_config['endpoints'][endp], "ssh_password", ssh_password)
+            setattr(prov_config['endpoints'][endp], "ssh_creds", ssh_creds)
 
             if OpenshiftDefaultEndpoint.name == endp:
                 prov_config['endpoints'][endp]['token'] = token_creds.token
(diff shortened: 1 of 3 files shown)


jawatts:
I would remove this line from the test https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/containers/test_ssl_providers.py#L14 and run the run test_ssl_providers so we can see the results in PRT.

cfme/tests/containers/test_ssl_providers.py
```diff
@@ -4,14 +4,13 @@
 from fauxfactory import gen_alphanumeric, gen_integer
 import pytest
 
+from cfme.utils import ssh
```
jawatts: Could you put the imports in alphabetical order.

cfme/tests/containers/test_ssl_providers.py
```diff
@@ -49,9 +48,38 @@
 )
 
 
+@pytest.fixture(scope="module")
+def sync_ssl_certificate(provider, appliance):
+    """
+    fixture which sync SSL certificate between CFME and OCP
+    :param provider:  OCP provider object
+    :param appliance: CFME appliance object
+    """
```
jawatts: The correct docblock format is found here: https://pypi.python.org/pypi/sphinxcontrib-napoleon

cfme/tests/containers/test_ssl_providers.py
```diff
@@ -49,9 +48,38 @@
...
+
+    provider_ip = provider.data["ipaddress"]
+    provider_cred_name = provider.data["credentials"]
```
jawatts: Thoughts on this instead since `provider_cred_name` is not used anywhere else:
`provider_cred_password = provider.get_credentials_from_config(provider_cred_name).secret`

and you can set `password` below to `provider_cred_password'

jawatts:
General comment: Rebase the commit.

jawatts:
Check out the results of PRT, your tests are failing to collect. It is because of the pytest command up top, you want:
`pytest: cfme/tests/containers/test_ssl_providers.py --use-provider ocp-v2`

PRT checks out master, merges your changes in, then runs.

cfme/tests/containers/test_ssl_providers.py
shalomnaim1: Done

cfme/tests/containers/test_ssl_providers.py
shalomnaim1: I hope that good now, please take a look at it.

cfme/tests/containers/test_ssl_providers.py
shalomnaim1: I don't understand this comment, can you explain again, please?

jawatts: Sure, happy to explain. So I see that you are setting `provider_cred_name = provider.data["credentials"]`, but it is only used once (on line 65). So I was thinking, to shorten line 65, you could set `provider_cred_password = provider.get_credentials_from_config(provider_cred_name).secret` instead of `provider_cred_name = provider.data["credentials"]`. Then on line 65, set `"password": provider_cred_password,`. This is only a suggestion.

mshriver:
@shalomnaim1 I added 'FIXES' line to your description to properly link this PR to the related issue.

cfme/containers/provider/openshift.py
```diff
@@ -25,17 +25,19 @@ def __init__(self, name, value, field_type=None, href=None):
...
-           No args.
+            connection_info (dict): username, password and hostname for OCP
         returns:
             certificate's content.
         """
-        cert_file_path = path.join(str(data_path), 'cert-auths', 'cmqe-tests-openshift-signer.crt')
-        with open(cert_file_path) as f:
-            return f.read()
+
+        with ssh.SSHClient(**connection_info) as provider_ssh:
```
There is already [ssh connection in the provider](https://github.com/ManageIQ/integration_tests/blob/master/cfme/containers/provider/openshift.py#L132), Please insert the provider into the function instead and perform this with cli.run_command (https://github.com/ManageIQ/integration_tests/blob/master/cfme/utils/ocp_cli.py#L31). it should be something like:
```python
def get_ca_cert(provider):
    return ''.join(provider.cli.run_command('cat /etc/origin/master/ca.crt')[1].readlines()))
```

cfme/containers/provider/openshift.py
```diff
@@ -25,17 +25,19 @@ def __init__(self, name, value, field_type=None, href=None):
...
+            _, stdout, _ = provider_ssh.exec_command("cat /etc/origin/master/ca.crt")
+            return str("".join(stdout.readlines()))
```
gshefer: __OPTIONAL__
I think "".join(stdout.readlines()) is enough right?

cfme/containers/provider/openshift.py
```diff
@@ -51,7 +53,11 @@ def view_value_mapping(self):
         }
 
         if out['sec_protocol'] and self.sec_protocol.lower() == 'ssl trusting custom ca':
-            out['trusted_ca_certificates'] = OpenshiftDefaultEndpoint.get_ca_cert()
+            out['trusted_ca_certificates'] = \
+                OpenshiftDefaultEndpoint.get_ca_cert(
```
gshefer: With the above comment, do: ```OpenshiftDefaultEndpoint.get_ca_cert(provider)```

gshefer:
@mshriver @jawatts @psav I think that although my changes request we can merge this and fix next week, because this fix is working and very important to our automation (blocking almost all the tests because we cannot add provider with SSL without it). what do you think?

cfme/containers/provider/openshift.py
shalomnaim1: I wasn't sure if the appliance approve the cert in Unicode so I forced conversion to ASCII

cfme/tests/containers/test_ssl_providers.py
```diff
@@ -49,9 +48,60 @@
...
+    connections_info = {"username": "root",
+                        "password": provider.get_credentials_from_config(provider_cred_name).secret,
+                        "hostname": provider_ip}
+
+    provider_ssh = ssh.SSHClient(**connections_info)
+    appliance_ssh = appliance.ssh_client()
+
+    # Connection to the applince in case of dead connection
+    if not appliance_ssh.connected:
+        appliance_ssh.connect()
```
jawatts: Is it possible that this connection could be dead as well?

gshefer: It's probably not connected by default?

shalomnaim1: @jawatts @gshefer yes, this possible to get a not-connected ssh object

shalomnaim1: This is a demo of getting an SSH client from an appliance

``` python
In [1]: import cfme.utils.appliance as appl
   ...: appl.CREATE_IS_PEDANTIC = False
   ...: app = appl.get_or_create_current_appliance()
   ...: app_ssh = app.ssh_client()
   ...: print app_ssh.connected
   ...: 
[WARNING] cfme/utils/appliance/__init__.py:2762: DeprecationWarning: Using class 'CurrentAppliance' (either directly or via inheritance) is deprecated: The CurrentAppliance descriptor is being phased outin favour of collections.
  appliance = CurrentAppliance()
 (/home/snaim/virtEns/myDevEnv/lib/python2.7/site-packages/debtcollector/_utils.py:61)
[WARNING] cfme/utils/appliance/__init__.py:2712: ApplianceSummoningWarning: magical creation of appliance objects has been deprecated, please obtain a appliance object directly
  category=ApplianceSummoningWarning,
 (cfme/utils/appliance/__init__.py:2712)
[WARNING] cfme/utils/appliance/__init__.py:2700: DeprecationWarning: Your appliance specification has old-style base_url, please change
  return load_appliances(appliances, global_kwargs)
 (cfme/utils/appliance/__init__.py:2619)
[WARNING] /home/snaim/virtEns/myDevEnv/lib/python2.7/site-packages/yaycl_crypt/__init__.py:148: YayclCryptWarning: yaml "./conf/credentials.yaml" and eyaml present for "./conf/credentials.eyaml" config. Ignoring encrypted yaml.
  warnings.warn(warn_msg, YayclCryptWarning)
 (/home/snaim/virtEns/myDevEnv/lib/python2.7/site-packages/yaycl_crypt/__init__.py:148)
[WARNING] /home/snaim/virtEns/myDevEnv/lib/python2.7/site-packages/oslo_i18n/_message.py:21: ImportWarning: Not importing directory '/home/snaim/virtEns/myDevEnv/lib/python2.7/site-packages/oslo_i18n/locale': missing __init__.py
  import locale
 (/home/snaim/virtEns/myDevEnv/lib/python2.7/site-packages/oslo_i18n/_message.py:21)
[WARNING] /usr/lib64/python2.7/pkgutil.py:186: ImportWarning: Not importing directory './data': missing __init__.py
  file, filename, etc = imp.find_module(subname, path)
 (/usr/lib64/python2.7/pkgutil.py:186)
None

In [2]: app_ssh.connected

In [3]: print  app_ssh.connected
None

In [4]: app_ssh.connect()

In [5]: print  app_ssh.connected
True

In [6]: 
```

As you can see the SSH connection is not connected by defualt

cfme/containers/provider/openshift.py
shalomnaim1: I tried to implement this kind of solution but OpenShit endpoint doesn't have a provider attribute, and I am not sure that is the right solution for us, due to the size of this object.

I can add an add the provider in [1] instead of ssh_password

[1] https://github.com/shalomnaim1/integration_tests/blob/a232af749a7eee66bb54d245909be5a368623c87/cfme/containers/provider/openshift.py#L206

shalomnaim1: I thought about it, and the solution I suggested here in the previous comment is not valid since ```from_config``` is a class method, so I don't have any access to ```self``` and I don't want to create a dummy provider object to overload ```prov_config```

gshefer: I understand, so this is make sense.

cfme/containers/provider/openshift.py
gshefer: Ignore this.

shalomnaim1: see my response to the above comment

shalomnaim1:
I changed the main idea of the PR, instead of getting the CA cert for each endpoint by executing the same ssh command, I got the cert once and saved it for each endpoint.

BTW, this is the default configuration of OCP, all the endpoint share the same certificate unless a specific configuration is done for some endpoint

cfme/tests/containers/test_ssl_providers.py
```diff
@@ -49,9 +48,60 @@
...
+
+    # Do in case of failure (return code is not 0)
+    if stdout.readline().replace('\n', "") != "0":
+        cert_name = "{provider_name}.ca.crt".format(
+            provider_name=provider.provider_data.hostname.split(".")[0])
+
+        # Copy certificate to the appliance
+        provider_ssh.get_file("/etc/origin/master/ca.crt", "/tmp/ca.crt")
+        appliance_ssh.put_file("/tmp/ca.crt",
+                               "/etc/pki/ca-trust/source/anchors/{crt}".format(crt=cert_name))
```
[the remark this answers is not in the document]
shalomnaim1: it will add OCP correct certificate and add it to the appliance, it won't remove ant existing certificate

cfme/containers/provider/openshift.py
```diff
@@ -192,7 +182,21 @@ def from_config(cls, prov_config, prov_key, appliance=None):
 
         endpoints = {}
         token_creds = cls.process_credential_yaml_key(prov_config['credentials'], cred_type='token')
+
+        with ssh.SSHClient(**{"username": "root",
```
psav: **REQUIRED**: I really don't like calling out to SSH during what is essentially the init. I'm not sure what @mfalesni or @jkrocil say, but I think we should be able to instantiate a provider object with calling out to SSH. In fact ding this would mean there would currently be SSH calls during collection. This is usually a big no no.

shalomnaim1: So what do you suggest doing instead, in the previous implementation I overload the ssh_password for each endpoint, and when this certificate was required, it got the certificate via ssh, is this sound better?

I can revert the code to the previous version, but before I do please take a look at it 
 --> https://github.com/ManageIQ/integration_tests/pull/6432/commits/a232af749a7eee66bb54d245909be5a368623c87

shalomnaim1:
@psav the code changed according to your comments, I would like you to review the new code in your free time, in addition, I rebase the code as required, many thanks ;)

shalomnaim1:
@psav, any news?

PR #6436, opened 2018-01-15
[WIPTEST] Waiting for view in Tenant.set_quota method

jan-zmeskal:
__Adding__ additional check for `was_change` value in `TenantQuotaForm.fill` method. It's all pretty much explainged in the docstring of `check_fill_request_is_same_as_current_form` method.

{{pytest: cfme/tests/infrastructure/test_tenant_quota.py -vv --use-provider rhv41}}

### DIFF
--- a/cfme/configure/access_control/__init__.py
+++ b/cfme/configure/access_control/__init__.py
@@ -1290,6 +1290,29 @@ class TenantQuotaForm(View):
     vm_txt = Input(id='id_vms_allocated')
     template_txt = Input(id='id_templates_allocated')
 
+    @staticmethod
+    def check_fill_request_is_same_as_current_form(request, current):
+        """Description"""
+        for r_key, r_val in request.items():
+            if r_key.endswith('_cb') and bool(r_val) != current[r_key]:
+                return False
+            if r_key.endswith('_txt'):
+                if r_val is None and current[r_key] == '':
+                    pass
+                elif str(r_val) == current[r_key]:
+                    pass
+                else:
+                    return False
+        return True
+
+    def fill(self, values):
+        current = self.read()
+        if self.check_fill_request_is_same_as_current_form(values, current):
+            was_change = False
+        else:
+            was_change = super(TenantQuotaForm, self).fill(values)
+        return was_change
+
 
 class TenantQuotaView(ConfigurationView):
     """ Tenant Quota View """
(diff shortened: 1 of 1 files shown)


vatsalparekh:
@jan-zmeskal please change pytest to run it for all the related tests
{{pytest: cfme/tests/infrastructure/test_tenant_quota.py -vvv}}

cfme/configure/access_control/__init__.py
```diff
@@ -1290,6 +1290,40 @@ class TenantQuotaForm(View):
...
+            if request_key.endswith('_cb') and bool(request_value) != current[request_key]:
+                return False
+            if request_key.endswith('_txt'):
+                if request_value is None and current[request_key] == '':
+                    pass
+                elif str(request_value) == current[request_key]:
+                    pass
+                else:
+                    return False
+        return True
```
quarckster: **REQUIRED**
I don't think we should workaround the issue such way. The real reason should be found. Why does fill method return a wrong value?
сс @mfalesni

jan-zmeskal: @quarckster Honestly I did not think it's a workaround. I thought that fill is sometimes so specific that you need to overload it to suit your needs. But if we find more universal solution for this, I am of course happy,

mfalesni: We shouldn't be using the suffixes in first place :stuck_out_tongue: 
If it returns True even if nothing was changed, can you look in the logs which widget returned True so fill returned true?

jan-zmeskal: @mfalesni Here is relevant piece of [cfme.log](https://up1-vrutkovs.int.open.paas.redhat.com/#0sZrfyiIxYkRQ7xUVMLoOg). If fails right after this, because of course no success flash message is not displayed if Save button was not clicked. I mean it probably was clicked, but it did nothing since it is not active.

jan-zmeskal: @mfalesni What do you think about it? Should we merge this and consider it as a WA? Or can you create an actual fix? Or maybe point me in the right direction? I am asking because it would be really great if the fix made it to the next downstream-stable.

quarckster: @jan-zmeskal I cannot reproduce the issue, I ran tenant_quota_test with `set_quota` code:
```python
def set_quota(self, **kwargs):
    """ Sets tenant quotas """
    view = navigate_to(self, 'ManageQuotas', wait_for_view=True)
    changed = view.form.fill({'cpu_cb': kwargs.get('cpu_cb'),
                              'cpu_txt': kwargs.get('cpu'),
                              'memory_cb': kwargs.get('memory_cb'),
                              'memory_txt': kwargs.get('memory'),
                              'storage_cb': kwargs.get('storage_cb'),
                              'storage_txt': kwargs.get('storage'),
                              'vm_cb': kwargs.get('vm_cb'),
                              'vm_txt': kwargs.get('vm'),
                              'template_cb': kwargs.get('template_cb'),
                              'template_txt': kwargs.get('template')})
    if changed:
        view.save_button.click()
        view = self.create_view(DetailsTenantView)
        view.flash.assert_success_message('Quotas for {} "{}" were saved'.format(
            self.obj_type, self.name))
    else:
        view.cancel_button.click()
        view = self.create_view(DetailsTenantView)
        view.flash.assert_success_message('Manage quotas for {} "{}" was cancelled by the user'
            .format(self.obj_type, self.name))
    assert view.is_displayed
```
I didn't changed the view at all and it works fine.

jan-zmeskal: @quarckster Try running it on an appliance where all the values that are going to be set by test are manually set before.

jan-zmeskal: @mfalesni, @quarckster Sorry to be bothering you again, but I need your ideas on this. If you want me to rework this, then ok, but at least give me some pointers. However right now `view.form.fill` returns `True` even if the desired values were in the form even before filling and that's a problem. So I cannot just leave it as it is.

jan-zmeskal:
Conversation with Milan:

> <@mfalesni> jzmeskal: I don't like the fact the suffixes are used for detection, where they really should not be used at all (it is because of the names of the fields were taken from older code)
> <jzmeskal> mfalesni: I guess I could change that but that's definitely not the main point of the PR
> <@mfalesni> it is a pattern we do not want in the code. Once in teh code, forever copypasted :)
> <mshriver> jzmeskal: I've seen behavior like this on forms with obfuscated passwords (like auth configuration) - the fill() returns true because the password is re-entered, but CFME disables the save button because the value is unchanged
> <@mfalesni> this doesnt look like such case though
> <mshriver> gotcha
> <@mfalesni> jzmeskal: try not filling integers
> <@mfalesni> fill strings
> <@mfalesni> that will fix it
> <@mfalesni> (unfortunately)
> <@mfalesni> I think that may be it
> <@mfalesni> since the filling thing converts everything to string, but 2 != '2'
> <jzmeskal> mfalesni: okay, thank for the hint, I'll try it tomorrow. Do you mind if I paste our conversation to PR so I dont' forget what to do?
> <@mfalesni> sure
> <@mfalesni> just make sure you don't fill integers to inputs
> <@mfalesni> the views filling do not like it
> <@mfalesni> I may try updating the view fill method to account for integers though
> <@mfalesni> but not sure how that would fit generically :)

cfme/configure/access_control/__init__.py
mfalesni: So I pointed out @jan-zmeskal fills integers into inputs, which breaks the fill logic. (2 != '2' -> fill it, and it works because we convert all fills into strings)

PR #6438, opened 2018-01-15
[1LP][RFR] Check provider for available load balancers

rlbabyuk:
Purpose or Intent
=================
Added a check for available load balancers for provider, it not found will skip the test

{{pytest: cfme/tests/networks/test_sdn_balancers.py -vvv --use-provider complete}}

### DIFF
--- a/cfme/tests/networks/test_sdn_balancers.py
+++ b/cfme/tests/networks/test_sdn_balancers.py
@@ -23,9 +23,10 @@ def network_prov_with_load_balancers(appliance):
         try:
             sum_all = len(prov.balancers.all())
         except DestinationNotFound:
-            pytest.skip("No available load balancers for current provider")
+            continue
         available_prov[prov] = sum_all
-    return available_prov
+    return available_prov if available_prov else pytest.skip(
+        "No available load balancers for current providers")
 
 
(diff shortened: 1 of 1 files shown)


cfme/tests/networks/test_sdn_balancers.py
```diff
@@ -12,26 +14,35 @@
...
-def test_balances_detail(provider, appliance):
+def test_balances_detail(provider, network_prov_with_load_balancers):
     """ Test of getting attribute from balancer object """
-    collection = appliance.collections.network_providers
-    providers = collection.all()
-    for prov in providers:
+    for prov, _ in network_prov_with_load_balancers.items():
         for balancer in prov.balancers.all():
             check = balancer.health_checks
             assert check is not None
```
psav: **REQUIRED:** Just to ask why are we deleting the provider below here? You should also apply an else on your for loop to skip if no providers were found else the test will Falsely pass

rlbabyuk: Not sure why this provider delete was added, but yes it is unnecessary here.
And as for else, I'm not sure I understand as we add provider before test run, and test will skip if provider doesn't have load balancers. Or did I missed something?

PR #6441, opened 2018-01-15
[WIP] Fix vm reconfiguration timedOutError

niyazRedhat:
{{ pytest: cfme/tests/infrastructure/test_vm_reconfigure.py -k test_vm_reconfig_add_remove_disk_cold -vvv }}

### DIFF
--- a/cfme/tests/infrastructure/test_vm_reconfigure.py
+++ b/cfme/tests/infrastructure/test_vm_reconfigure.py
@@ -90,12 +90,12 @@ def test_vm_reconfig_add_remove_disk_cold(
 
     small_vm.reconfigure(new_config)
     wait_for(
-        lambda: small_vm.configuration == new_config, timeout=360, delay=45,
+        lambda: small_vm.configuration.num_disks == new_config.num_disks, timeout=360, delay=45,
         fail_func=small_vm.refresh_relationships,
         message="confirm that disk was added")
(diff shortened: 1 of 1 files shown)


cfme/tests/infrastructure/test_vm_reconfigure.py
```diff
@@ -90,12 +90,12 @@ def test_vm_reconfig_add_remove_disk_cold(
 
     small_vm.reconfigure(new_config)
     wait_for(
-        lambda: small_vm.configuration == new_config, timeout=360, delay=45,
+        lambda: small_vm.configuration.num_disks == new_config.num_disks, timeout=360, delay=45,
```
jkrocil: QUESTION
Using `small_vm.configuration.disks` instead of `small._vm.configuration.num_disks` should be preferred here because it will actually compare the disks as well (types, sizes, etc). Can you try if that works, please?

niyazRedhat: Yes sure I will try that. Thanks

psav: Can we merge as is to make the tests pass, this is potentially a +18 difference, which is large.

niyazRedhat:
Its been fixed. Closing this one.

PR #6447, opened 2018-01-15
[WIPTEST]added a separate template for rhos11

sshveta:
Network id and key pair is different for RHOS 11 , so added a separate template .


sshveta:
One sporadic failure in 5.8 .

cfme/tests/services/test_provision_stack.py
```diff
@@ -38,8 +38,10 @@ def template(provider, provisioning, setup_provider):
 
     if provider.type == "ec2":
         data_file = load_data_file(str(orchestration_path.join('aws_vm_template.json')))
-    elif provider.type == "openstack":
+    elif provider.name == "RHOS7-GA":
```
psav: **REQUIRED:** There should never be private data in a test file. If there is a required data file it can be referenced in the yaml, but we shuldn't ever put keys in the test. This means the test would only work if these keys were present. Which may not always be true and certainly won't be true if dev start running our tests.

PR #6451, opened 2018-01-16
[1LP][RFR] Ownership: Remove dependency on existing vm

apagac:
This is a refactor of existing ownership tests. The point of this PR is to remove the dependency on a vm set in the cfme_data.yaml.

Before, the ownership tests required a pre-provisioned VM with name set in the cfme_data file in every provider we wanted to test, like so:
"ownership_vm: <name_of_the_vm>"
This was cumbersome, the data repeated across the cfme_data.yaml file and it caused problems when trying to run PRT or automation, because there was no entry "ownership_vm" in the cfme_data.yaml.

Now, the module works with a yield fixture which provisions the VM beforehand, as seen in many other tests that are currently running in automation.

{{ pytest: -v --long-running --use-provider vsphere65-nested cfme/tests/cloud_infra_common/test_vm_ownership.py }}

### DIFF
--- a/cfme/tests/cloud_infra_common/test_vm_ownership.py
+++ b/cfme/tests/cloud_infra_common/test_vm_ownership.py
@@ -4,18 +4,32 @@ import pytest
 import cfme.configure.access_control as ac
 from cfme import test_requirements
 from cfme.base.credential import Credential
-from cfme.common.provider import BaseProvider
+from cfme.common.provider import CloudInfraProvider
 from cfme.common.vm import VM
 from cfme.exceptions import VmOrInstanceNotFound
 from cfme.utils.blockers import BZ
+from cfme.utils.generators import random_vm_name
+from cfme.utils.log import logger
 
 
 pytestmark = [
     test_requirements.ownership,
     pytest.mark.meta(blockers=[BZ(1380781, forced_streams=["5.7"])]),
     pytest.mark.tier(3),
-    pytest.mark.provider([BaseProvider], required_fields=['ownership_vm'], scope='module')
+    pytest.mark.provider([CloudInfraProvider], scope='module')
 ]
+
+
+@pytest.yield_fixture(scope="module")
+def vm_crud(provider):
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit 02fcd72e1241afffb6c10d589639ec70148463fe:

`cfme/tests/cloud_infra_common/test_vm_ownership.py`:
- :red_circle: Line 31:9: **F821** *undefined name 'logger'*

**No commit flaws detected**.

*CFME QE Bot*

apagac:
5.9 failure occurs due to a known bug 1533391

cfme/tests/cloud_infra_common/test_vm_ownership.py
```diff
@@ -8,16 +8,30 @@
...
 
+@pytest.yield_fixture(scope="module")
+def vm_crud(provider):
+    vm = VM.factory(random_vm_name(context='ownrs'), provider)
+    vm.create_on_provider(find_in_cfme=True, allow_skip="default")
+    yield vm
+
+    try:
+        vm.delete_from_provider()
+    except:
```
rlbabyuk: **Required**: Specify exception type, at least Exception should be used here

cfme/tests/cloud_infra_common/test_vm_ownership.py
```diff
@@ -8,16 +8,30 @@
...
+from cfme.utils.generators import random_vm_name
+from cfme.utils.log import logger
 
 
 pytestmark = [
     test_requirements.ownership,
     pytest.mark.meta(blockers=[BZ(1380781, forced_streams=["5.7"])]),
     pytest.mark.tier(3),
-    pytest.mark.provider([BaseProvider], required_fields=['ownership_vm'], scope='module')
+    pytest.mark.provider([BaseProvider], scope='module')
```
mshriver: **Required**

Restrict this to CloudInfraProvider, I don't think you'll want to generate against network providers, physical infra providers, or container providers - yet.

PR #6452, opened 2018-01-16
[1LP][RFR] Workaround for test_security_group create

rlbabyuk:
Purpose or Intent
=================
One more case with MoveTargetOutOfBoundsException: Message. This time I've tried to close flash message, as some action before open configuration in toolbar(BZ 1520669).

{{pytest: cfme/tests/cloud/test_security_groups.py -vvv}}

### DIFF
--- a/cfme/cloud/security_groups.py
+++ b/cfme/cloud/security_groups.py
@@ -239,4 +239,7 @@ class Add(CFMENavigateStep):
     def step(self, *args, **kwargs):
         """Raises DropdownItemDisabled from widgetastic_patternfly
         if no RHOS Network manager present"""
+        # Todo remove when fixed 1520669
+        if self.prerequisite_view.flash.messages:
+            self.prerequisite_view.flash.dismiss()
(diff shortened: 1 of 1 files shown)


cfme/cloud/security_groups.py
```diff
@@ -239,4 +239,7 @@ class Add(CFMENavigateStep):
     def step(self, *args, **kwargs):
         """Raises DropdownItemDisabled from widgetastic_patternfly
         if no RHOS Network manager present"""
+        # Todo remove when fixed 1520669
+        if self.prerequisite_view.flash.messages:
+            self.prerequisite_view.flash.dismiss()
```
quarckster: **REQUIRED**
```python
if BZ(blockers=[1520669], forced_streams='5.9').blocks and self.prerequisite_view.flash.messages:
   self.prerequisite_view.flash.dismiss()
```

rlbabyuk: done

PR #6460, opened 2018-01-16
[1LP][RFR][NOTEST]Sprout: Appliance did not pass on_openstack. Also added a condition for only downstreams creating such disk

mfalesni:


mfalesni:
We are investigating issue where the appliance_console and appliance_console_cli refuse to use our workaround because /dev/vdb is a symlink to /dev/loop0 which is the only one displayed in fdisk.

mshriver:
PRT failures in smoke test are unrelated.

PR #6461, opened 2018-01-16
[1LP][RFR] fix: test_delete_resource_pool_appear_after_refresh 

digitronik:
Purpose or Intent
=================
- fix test `test_delete_resource_pool_appear_after_refresh`
- Removed table for entities as `BaseEntitiesView` going to work on page
- Fixed All page `is_displayed()`
- Removed view resets
- Improve `exist()` method 
{{pytest: cfme/tests/infrastructure/test_delete_infra_object.py -k 'test_delete_resource_pool_appear_after_refresh' --use-provider vsphere55 -v}}

### DIFF
--- a/cfme/infrastructure/resource_pool.py
+++ b/cfme/infrastructure/resource_pool.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 import attr
 from navmazing import NavigateToAttribute
-from widgetastic.exceptions import NoSuchElementException
+
 from widgetastic.widget import View
 from widgetastic_patternfly import Button, Dropdown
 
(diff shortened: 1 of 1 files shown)


cfme/infrastructure/resource_pool.py
```diff
@@ -151,7 +150,7 @@ def delete(self, cancel=False, wait=False):
             def refresh():
                 if self.provider:
                     self.provider.refresh_provider_relationships()
-                view.browser.selenium.refresh()
+                view.browser.refresh()
```
quarckster: **QUESTION**
What's the point of this change?

cfme/infrastructure/resource_pool.py
```diff
@@ -195,12 +194,7 @@ def get_detail(self, *ident):
     @property
     def exists(self):
         view = navigate_to(self.parent, 'All')
-        try:
-            view.toolbar.view_selector.select('List View')
-            view.entities.paginator.find_row_on_pages(view.entities.table, name=self.name)
-            return True
-        except NoSuchElementException:
-            return False
+        return self.name in view.entities.entity_names
```
quarckster: **QUESTION**
Does `view.entities.entity_names` return entities not only from current page? Alternatively, you can check the existence like that:
```python
try:
    navigate_to(self, 'Details')
    return False
except ItemNotFound:
    return False
```

digitronik: @quarckster: while `navigate_to(self, 'Details')`, we are locating the entity and try navigate to `Detail Page`. Rather than changing page with extra navigation task. I thought its good to check the 'All' page with `view.entities.entities_name`; It will return all entities present on `All page` at that time so we can check our entity l.e. `self.name` is present in returning list or not. Its will fast and effective check for `exists()`.

cfme/infrastructure/resource_pool.py
digitronik: both work same to hide `selenium` @quarckster

PR #6463, opened 2018-01-17
[1LP][RFR] Fixed step issue

psav:
Issue: Imagine we had landed on Cloud Providers page, and then via REST
       we add a cloud provider. If we now try to navigate to the Details
       page of that cloud provider it will fail because the am_i_here is
       True, but the provider is not visible on the page as it was done
       via REST.

* Added a try/except around step to retry the step operation after a
  refresh. This _could_ make some things fail, but then they were going
  to fail anyway.

### DIFF
--- a/cfme/utils/appliance/implementations/ui.py
+++ b/cfme/utils/appliance/implementations/ui.py
@@ -446,7 +446,15 @@ class CFMENavigateStep(NavigateStep):
         if not here:
             self.log_message("Prerequisite Needed")
             self.prerequisite_view = self.prerequisite()
-            self.check_for_badness(self.step, _tries, nav_args, *args, **kwargs)
+            try:
+                self.check_for_badness(self.step, _tries, nav_args, *args, **kwargs)
+            except Exception as e:
+                self.log_message(
+                    "Exception raised [{}] whilst running step, trying refresh".format(e),
+                    level="error"
+                )
+                self.appliance.browser.widgetastic.refresh()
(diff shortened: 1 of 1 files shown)


cfme/utils/appliance/implementations/ui.py
```diff
@@ -446,7 +446,15 @@ def go(self, _tries=0, *args, **kwargs):
         if not here:
             self.log_message("Prerequisite Needed")
             self.prerequisite_view = self.prerequisite()
-            self.check_for_badness(self.step, _tries, nav_args, *args, **kwargs)
+            try:
+                self.check_for_badness(self.step, _tries, nav_args, *args, **kwargs)
+            except Exception as e:
+                self.log_message(
+                    "Exception raised [{}] whilst running step, trying refresh".format(e),
```
mshriver: There's some potential here to break navigation when prerequisite and step are closely coupled - consider if prerequisite for a navigation destination is registered against a collection, and is navigating to 'All' and checking table rows before the `step` method clicks the toolbar dropdown.  In this case, the refresh would clear the selection made by prerequisite.

We don't have anything doing this now, just a thought for how navigation for collections might get implemented. Closest we have is infra/datastores, where it navigates to datastores filtered by provider, but this isn't impacted by refresh.

This also locks navigation into refreshing anytime there is an exception during `step`, and doesn't allow us control over where we may want to ignore it, and where we really want or expect to see that exception without refresh.

PR #6467, opened 2018-01-17
[1LP][RFR] Block test_cfme_auth ext_openldap on GH issue

mshriver:
need new server deployed

{{ pytest: cfme/tests/integration/test_cfme_auth.py -k ext_openldap }}


mshriver:
There will be failures in the other tests, all I'm doing with this PR is blocking ext_openldap because we need a new server.  In-progress on fixing the other tests.

PR #6470, opened 2018-01-17
[WIP] Fixed "TimedOutError" for test_discovery.py

sbulage:
{{ pytest: cfme/tests/cloud_infra_common/test_discovery.py -k test_vm_discovery }}


vatsalparekh:
I would suggest to run PRT for other providers as well, not just GCE.

sbulage:
I don't think so running it on every provider make any sense, since it is failing on GCE provider only.

sbulage:
PR https://github.com/ManageIQ/wrapanapi/pull/207 merged which has same fix. So closing this PR.

PR #6471, opened 2018-01-17
[1LP][RFR] Resolving Timeout Error caused by handle_alert

kedark3:
Purpose or Intent
=================

__Fixing__ item_select in open_console which was causing timeout error as given in RHCFQE-6568. It was occurring because it was taking False as the default value and it would wait for 10 sec to find an alert and timeout, it should rather take none, but during WT conversion someone changed that logic as it seems based on git history. 

Removed the cancel parameter from the function since it was not being used at all and I do not see a use case for it at this moment. 

### DIFF
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -287,7 +287,7 @@ class BaseVM(Pretty, Updateable, PolicyProfileAssignable, WidgetasticTaggable, N
         return getattr(view.entities, properties[0].lower().replace(' ', '_')).get_text_of(
             properties[1])
 
-    def open_console(self, console='VM Console', invokes_alert=False, cancel=False):
+    def open_console(self, console='VM Console', invokes_alert=None):
         """
         Initiates the opening of one of the console types supported by the Access
         button.   Presently we only support VM Console, which is the HTML5 Console.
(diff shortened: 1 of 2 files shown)


sshveta:
👍 LGTM

cfme/common/vm.py
```diff
@@ -287,7 +287,7 @@ def get_detail(self, properties=None):
         return getattr(view.entities, properties[0].lower().replace(' ', '_')).get_text_of(
             properties[1])
 
-    def open_console(self, console='VM Console', invokes_alert=False, cancel=False):
+    def open_console(self, console='VM Console', invokes_alert=None):
```
mshriver: Pretty sure you want to default this to True, not None - None means no handling.  True means it'll confirm the alert that's raised when the console is opened.  If there is no alert during item_selection, the arg should be removed completely as the default for `item_select()` is `handle_alert=None`

https://github.com/RedHatQE/widgetastic.patternfly/blob/master/src/widgetastic_patternfly/__init__.py#L1489

mshriver: @kedark3 talked to me offline, and this is needed for webmks, and doesn't have alerts otherwise. Preferable to taking `**kwargs` too.

PR #6472, opened 2018-01-17
[WIPTEST] Add CFMENavigateStep.force_refresh

mshriver:
Set true for CloudProvider and InfraProvider 'All' destinations.

This is a prototype for a very simple 'fix' to an issue raised by @psav for provider navigation.

Providers are added via REST often during tests, and navigation will fail because no refresh is performed on the 'All' providers page when no new navigation occurs.

This attribute can be set for navigation classes that should force a refresh when `am_i_here` is true - and thus 'step' isn't called.

Removes need for navigation classes to include a `resetter` function that refreshes.

Still prototype, needs testing and some cases considered - like a `skip_refresh` nav_arg for overriding the class setting in a test.


mshriver:
Related to #6463 @ManageIQ/admins-qe please review both for consideration of how to handle auto-refresh during navigation.

dajoRH:
*CFME QE Bot*

mshriver:
Closing, chose #6463 instead

PR #6476, opened 2018-01-17
[1LP][WIP] create_rest RHOS enhancements/fixes

mkoura:
* fix Keystone API setting (so far defaults were used)
* add AMQP endpoint support

{{pytest: -v --use-provider complete -k test_provider_refresh cfme/tests/cloud_infra_common/test_rest_providers.py}}

### DIFF
--- a/cfme/common/provider.py
+++ b/cfme/common/provider.py
@@ -219,6 +219,17 @@ class BaseProvider(WidgetasticTaggable, Updateable, Navigatable):
         if getattr(self, "project", None):
             provider_attributes["project"] = self.project
 
+        # Keystone API
+        api_version = getattr(self, "api_version", None)
+        if api_version and "Keystone" in api_version:
+            api_version_trimmed = self.api_version.replace("Keystone ", "")
+            provider_attributes["api_version"] = api_version_trimmed
+        if api_version == "Keystone v3" and getattr(self, "keystone_v3_domain_id", None):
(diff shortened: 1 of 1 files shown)


cfme/common/provider.py
```diff
@@ -219,6 +219,17 @@ def _fill_provider_attributes(self, provider_attributes):
         if getattr(self, "project", None):
             provider_attributes["project"] = self.project
 
+        # Keystone API
```
psav: **REQUIRED:** We shouldn't have any provider specific information in here. If we need to do anything special for a provider, we should create a method on the provider wit ha default on the base so that we can do provider specific operations. We already do this for other things.

psav: PS. I know we have done this and allowed it in the past, but we need to make providers totally modular so there is no core contamination.

dajoRH:
*CFME QE Bot*

mshriver:
@mkoura I think this has otherwise been implemented at this point - was not aware of this PR when I was reviewing a more recent contribution. Pete has some very good points about some of the logic polluting the common provider class with specific checks for certain provider types.

If you are still interested in working on this, I would love to discuss a strategy for removing the provider-type specific checks that are in the common provider class.

I'm going to close this PR for now though, we can re-open with newer changes if you are interested.

Thanks!

mkoura:
Sure @mshriver, actually I pointed @valaparthvi to this PR some time ago and her recent changes were partially based on it, so it already served its purpose.

PR #6477, opened 2018-01-18
[1LP][RFR] Remove store usage

psav:
Remove store usage


mshriver:
PRT failures unrelated.

PR #6490, opened 2018-01-19
[WIPTEST] Do not assume "Provider" noun

miha-plesko:
Currently, all Providers contain `string_name` attribute (e.g. `"Cloud"`) which is then concatenated with a hard-coded string `" Provider"`, which results in names like "Cloud Provider", "Infrastructure Provider" etc. As of Nuage provider this is no longer valid since Nuage uses following name: "Network Manager".

With this commit we update all providers to contain both parts in its `string_name`. In other words, instead:

```
string_name = "Cloud"
```

we now use

```
string_name = "Cloud Provider"
```

which makes it possible to use "Network Manager" in case of Nuage provider.

### DIFF
--- a/cfme/cloud/provider/__init__.py
+++ b/cfme/cloud/provider/__init__.py
@@ -96,7 +96,7 @@ class CloudProvider(Pretty, CloudInfraProvider):
     category = "cloud"
     pretty_attrs = ['name', 'credentials', 'zone', 'key']
     STATS_TO_MATCH = ['num_template', 'num_vm']
-    string_name = "Cloud"
+    string_name = "Cloud Provider"
     templates_destination_name = "Images"
     vm_name = "Instances"
     template_name = "Images"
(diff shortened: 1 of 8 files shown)


miha-plesko:
@psav I'm unable to see any error message regarding why this is failing downstream 58z and 59z, is there perhaps a log somewhere?

mfalesni:
@miha-plesko The PR testing is available only for us redhatters so far - because we need providers, we need to run it internally.

miha-plesko:
I'm not sure about this one: should I simply override methods that are different for NetworkProvider (e.g. `create`, `delete`) instead of updating the code like in this PR? Problem that I'm solving is that popup text differs between CloudProvider and NetworkManager.

I kindly ask you guys for advice!

miha-plesko:
Closing. I'll go with the second option and open up a new PR.

PR #6491, opened 2018-01-19
[WIPTEST] First test for Nuage provider

miha-plesko:
With this commit we implement the very first test for Nuage provider. It just opens up a form for creating new Nuage provider, clicks "Cancel" and then asserts specific popup content.

Depends on: https://github.com/ManageIQ/integration_tests/pull/6490


miha-plesko:
Closing in favor of https://github.com/ManageIQ/integration_tests/pull/6499

PR #6498, opened 2018-01-19
[WIP] Fixing catalog_item_type for RHV

jan-zmeskal:
__Fixing__ catalog_item_type for RHV provider. So it seems that since 5.9.0.17 catalog item type is "Red Hat Virtualization" as opposed to previous "RHEV". Honestly I do not know if the proposed soluting is the best one. We need a way to keep the old name, because we will still be running automation on 5.8. And I don't think it's a good idea to change this on test case level. You would need to changed in in each test case and authors of new test cases might not be aware of that.

{{pytest: -vv cfme/tests/services/test_service_catalogs.py --use-provider rhv41 --long-running}}

5.9 will be failing until yamls repo in GitLab is updated. However, I cannot do that now, since it is unreachable. The change needed RHV provider is this:
`provisioning[catalog_item_type]` from "RHEV" to "Red Hat Virtualization".

### DIFF
--- a/cfme/fixtures/service_fixtures.py
+++ b/cfme/fixtures/service_fixtures.py
@@ -4,10 +4,12 @@ import pytest
 from widgetastic.utils import partial_match
 
 from cfme.common.provider import cleanup_vm
+from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.rest.gen_data import dialog as _dialog
 from cfme.rest.gen_data import service_catalog_obj as _catalog
 from cfme.services.catalogs.catalog_item import CatalogItem
 from cfme.services.service_catalogs import ServiceCatalogs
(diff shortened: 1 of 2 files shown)


psav:
I'd prefer this one #6510 as it solves the old problem of having provisioning_data hold the catalog_type which was always wrong.

jan-zmeskal:
@psav Yes, of course. It solves the problem in more systemic way. Therefore closing this PR in favor of #6510.

jan-zmeskal:
@psav Your solution is more systematic, so definitely let's go with it. Anyway the removed of the old hack in test case is still necessary.

dajoRH:
*CFME QE Bot*

jan-zmeskal:
Closing in favor of #6510

PR #6501, opened 2018-01-22
[WIPTEST] Block test_add_hawkular_provider_ssl due BZ 1536828

jawatts:
Blocking test_add_hawkular_provider_ssl due to BZ 1536828. 


jawatts:
@shalomnaim1

PR #6502, opened 2018-01-23
[1LP][RFR] Fixed test_restart_workers

nzr-zts:
Purpose or Intent
=================
Updated to use checkboxes on 5.9 instead of clicking on row. 

{{pytest: -v cfme/tests/configure/test_workers.py -k test_restart_workers}}

### DIFF
--- a/cfme/configure/configuration/diagnostics_settings.py
+++ b/cfme/configure/configuration/diagnostics_settings.py
@@ -2,8 +2,8 @@ import attr
 
 from navmazing import NavigateToAttribute, NavigateToSibling
 from widgetastic_patternfly import Input, BootstrapSelect, Button, Dropdown
-from widgetastic_manageiq import SummaryFormItem
-from widgetastic.widget import View, Table, RowNotFound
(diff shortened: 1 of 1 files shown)


nzr-zts:
It fails on 5.9.0.17 due to [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1533548) which is in POST and will be fixed in next build, but checkboxes work correctly due to this fix.

izapolsk:
@nzr-zts, please add appropriate comment about BZ to your fix and expain PRT errors.

nzr-zts:
This PR fixes the way of selecting worker, previously it was working by clicking on the appropriate row, but now it requires to use checkbox. PRT on 5.9 failed because of [BZ Workers: Infinite loading while selecting worker from the list](https://bugzilla.redhat.com/show_bug.cgi?id=1533548), after this fix it selects them (we can see that on [screenshots](http://10.16.45.124/XgULebH2/artifacts/cfme/tests/configure/test_workers.py/test_restart_workers/filedump-exception_screenshot.png) from report)  but not restarted that's why PRT fails.

PR #6510, opened 2018-01-23
[1LP][RFR] Fixed catalog_name selection

psav:
* This value used to be obtained via the yaml files as part of the
  provisioning data.
* As the catalog_name is provider specific it should be stored in the
  provider module.
* This includes a fix to make 5.9 work properly as the string changed

These values can now be removed from the cfme_data.yaml

### DIFF
--- a/cfme/tests/services/test_pxe_service_catalogs.py
+++ b/cfme/tests/services/test_pxe_service_catalogs.py
@@ -1,18 +1,17 @@
 # -*- coding: utf-8 -*-
 import fauxfactory
 import pytest
-
 from widgetastic.utils import partial_match
 
+from cfme import test_requirements
 from cfme.common.provider import cleanup_vm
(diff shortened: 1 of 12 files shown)


cfme/tests/services/test_pxe_service_catalogs.py
```diff
@@ -111,7 +110,7 @@ def catalog_item(provider, vm_name, dialog, catalog, provisioning, setup_pxe_ser
     }
 
     item_name = fauxfactory.gen_alphanumeric()
-    catalog_item = CatalogItem(item_type=catalog_item_type, name=item_name,
+    catalog_item = CatalogItem(item_type=provider.catalog_type, name=item_name,
```
mkoura: ``provider.catalog_name``?

psav: Fixed

PR #6515, opened 2018-01-24
[1LP][WIPTEST][NOMERGE] Chargeback reports, provider setup update

rlbabyuk:
Purpose or Intent
=================
Replaced setup fixture for pytest.mark.usefixture('setup_provider_modscope'), as previously provider was not added
{{pytest: cfme/tests/intelligence/reports/test_validate_chargeback_report.py --long-running --use-provider complete}}

### DIFF
--- a/cfme/tests/intelligence/reports/test_validate_chargeback_report.py
+++ b/cfme/tests/intelligence/reports/test_validate_chargeback_report.py
@@ -13,7 +13,6 @@ from cfme import test_requirements
 from cfme.base.credential import Credential
 from cfme.cloud.provider.azure import AzureProvider
 from cfme.cloud.provider.gce import GCEProvider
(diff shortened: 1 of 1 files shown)


rlbabyuk:
PRT for 5.9 doesn't look good. However, I would like first to merge this small fix. And continue fixing other cases in others PR.

rlbabyuk:
The previous fixture worked in module scope, so I replaced it with setup_provider fixture also in module scope, and as I am aware we can have 1 provider set up for 1 appliance in other cases all other providers will be removed. @psav could you please take a look, if this will work this way?

psav:
We are going to wait to see if this is fixed by @mkoura PR

rlbabyuk:
Looks like this fix isn't needed anymore. Tests are working fine. Closing

PR #6516, opened 2018-01-24
[1LP][RFR] Fix get_random_instances by limit the count with the count of instances

gshefer:
{{ pytest: cfme/tests/containers/test_properties.py --use-provider cm-env2 }}
- Fix get_random_instances by limit the count with the count of instances
- Added description


gshefer:
The Failures are expected.

PR #6523, opened 2018-01-25
[1LP][RFR] Moving VM Console fixture to separate file

kedark3:
Purpose or Intent
=================

__Fixing__ VM Console Fixtures by de-duplicating code from WebMKS,HTML5,SSUI VM Console tests and aggregating it as fixture.

{{pytest: -v --long-running cfme/tests/ssui/test_ssui_myservice.py -k 'test_vm_console' --use-provider vsphere6-nested}}

### DIFF
--- a/cfme/fixtures/vm_console.py
+++ b/cfme/fixtures/vm_console.py
@@ -6,14 +6,14 @@ from cfme.utils.log import logger
 
 
 @pytest.fixture(scope="module")
-def configure_console_vnc(appliance, provider):
+def configure_console_vnc(appliance):
     """Configure VMware Console to use VMware VNC."""
     logger.info("Changing VMware console suppport configuration to VNC")
     appliance.server.settings.update_vmware_console({'console_type': 'VNC'})
 
 
 @pytest.fixture(scope="module")
-def configure_console_webmks(appliance, provider):
+def configure_console_webmks(appliance):
     """Configure VMware Console to use VMware WebMKS."""
     logger.info("Changing VMware console suppport configuration to VMware WebMKS")
     appliance.server.settings.update_vmware_console({'console_type': 'VMware WebMKS'})
--- a/cfme/fixtures/vm_console.py
+++ b/cfme/fixtures/vm_console.py
@@ -1,0 +1,30 @@
+# -*- coding: utf-8 -*-
+"""Fixtures that VMware console tests use to configure VM Console type and start websocket."""
+import pytest
+
+from cfme.utils.log import logger
+
+
+@pytest.fixture(scope="module")
+def configure_console_vnc(appliance, provider):
+    """Configure VMware Console to use VMware VNC."""
+    logger.info("Changing VMware console suppport configuration to VNC")
+    appliance.server.settings.update_vmware_console({'console_type': 'VNC'})
+
+
+@pytest.fixture(scope="module")
+def configure_console_webmks(appliance, provider):
+    """Configure VMware Console to use VMware WebMKS."""
+    logger.info("Changing VMware console suppport configuration to VMware WebMKS")
+    appliance.server.settings.update_vmware_console({'console_type': 'VMware WebMKS'})
+
+
+@pytest.yield_fixture(scope="module")
+def configure_websocket(appliance):
+    """Enable websocket role if it is disabled.
+
+    Currently the fixture cfme/fixtures/base.py:27,
+    disables the websocket role to avoid intrusive popups.
+    """
+    logger.info('Enabling the websocket role to allow console connections')
+    appliance.server.settings.enable_server_roles('websocket')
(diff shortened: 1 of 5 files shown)


cfme/fixtures/vm_console.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+@pytest.yield_fixture(scope="module")
+def configure_websocket(appliance):
+    """Enable websocket role if it is disabled.
+
+    Currently the fixture cfme/fixtures/base.py:27,
+    disables the websocket role to avoid intrusive popups.
+    """
+    appliance.server.settings.enable_server_roles('websocket')
```
jamesooden: Is there anyway to make this code smarter to enable it only it hasn't been enabled, and later disable it only it hasn't been disabled?

cfme/fixtures/vm_console.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
```
jamesooden: Just for my understanding, why make this fixture code as opposed to vm_console class code?

jamesooden: Never mind they have always been fixtures...but I don't remember why (-;

jamesooden:
This all looks really good to me.  I'm not an official reviewer though (-:

cfme/fixtures/vm_console.py
kedark3: I think that is what we were doing earlier and then we changed it in this PR, don't recall why. 😸  https://github.com/kedark3/integration_tests/commit/485e3e58d275049fa1c19bd5c40d0cd7f2c8501a#diff-8dc02060f6e8056874c3e03630e11ff3

cfme/fixtures/vm_console.py
kedark3: I believe this is because these are pre-requisites for the test and vm_console class is instantiated much later in the test.

cfme/fixtures/vm_console.py
jamesooden: It probably doesn't matter and the code is simple this way.   I only worry that one day it being on may be the default and turning it off would be the wrong thing.

kedark3: Probably. But the primary reason we need to enable the role is "Add Provider" Fixture disables it and @psav had told me it will be best to disable it after the test, as most(almost all, correct me psav) run with "websocket" disabled, except few special cases.

jamesooden: The absolute right thing to do is to return it to the state it was in.   Presently this is always disabled.   Someday it may not be.   Still its not a big deal, but its the kind of thing I worry about biting us later.   However in this case there are so many mitigations to it biting us, we may just be fine.

quarckster: **REQUIRED**
I agree with @jamesooden. You need to get the role status before it will be changed and revert it that original state in tear down.

dajoRH:
*CFME QE Bot*

cfme/fixtures/vm_console.py
psav: So, I totally get what @jamesooden is saying. I did ask @kedark3 to make it turn it back off again, because it was so vital that it be off for testing. That being said, it actually turns out that turning it off takes a while, and rather than waiting for it we have decided to take a different approach and we temporarily just dismiss ALL notifications. So in all actuality we can probably just leave it on now :) Ironically this is how these things sometimes go.

kedark3: Leaving it ON for now. Not disabling anymore.

cfme/fixtures/vm_console.py
```diff
@@ -0,0 +1,24 @@
...
+import pytest
+
+from cfme.utils.log import logger
+
+
+@pytest.fixture(scope="module")
+def configure_vmware_console_for_test(appliance, provider, request):
+    """Configure VMware Console to use console type as required by test(like VNC, VMware WebMKS)."""
+    logger.info("Changing VMware console suppport configuration to {} "
+            .format(request.param['console_type']))
```
mfalesni: **REQUIRED**: [Bad logger format.](https://docs.python.org/2/library/logging.html#logging.Logger.debug)

mshriver:
I like it in general, but the indirect fixture use with singular parametrization creates a pretty funky test case name:
`test_vm_console[vsphere6-nested-order_catalog_item_in_ops_ui0-configure_vmware_console_for_test0-ViaSSUI] `

kedark3:
@mshriver if I get consensus from you, @psav  and @jamesooden  I can go ahead and remove the configure_vmware_console_for_test fixture and move it under each test as a single function call(getting rid of fixture entirely).

dajoRH:
*CFME QE Bot*

PR #6524, opened 2018-01-25
[1LP][RFR] Use v0.4.0 of manageiq-client

mkoura:
{{pytest: -v -k test_query_simple_collections cfme/tests/test_rest.py}}


mkoura:
@quarckster I changed it back to RFR since new version was released so I changed that one line :)

PR #6526, opened 2018-01-25
[1LP][RFR] Update BZ blockers for test_aws_iam_auth_and_roles

mshriver:
These BZ's weren't tagged for the specific streams, so I'm forcing the skip.


cfme/tests/integration/test_aws_iam_auth_and_roles.py
```diff
@@ -7,8 +7,6 @@
 from cfme.utils.conf import cfme_data, credentials
 from cfme.roles import role_access_ui_58z, role_access_ui_59z, role_access_ssui
 
-ONLY_IN_59 = ['Physical Infrastructure', 'Object Storage']
```
mshriver: Unused

PR #6528, opened 2018-01-25
[RFR] Require numpy 1.14.0

bsquizz:
This gets us around RunTimeError's seen while pytest imports plugins. I ran into this after setting up a fresh test environment on fedora27

```
RuntimeError: module compiled against API version 0xc but this version of numpy is 0xb
```
```
ImportError: Error importing plugin "fixtures.version_info": numpy.core.multiarray failed to import
```

### DIFF
--- a/requirements/frozen.txt
+++ b/requirements/frozen.txt
@@ -90,7 +90,7 @@ netaddr==0.7.19
 netifaces==0.10.6
 notebook==5.0.0
 ntlm-auth==1.0.4
-numpy==1.13.1
+numpy==1.14.0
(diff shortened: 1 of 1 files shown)


mfalesni:
@psav Shouldn't these changes be put in the nonfrozen requirements as well? To make sure we are on right versions when we regenerate our frozen requirements.

bsquizz:
Will open another PR to update frozen_docs and template_docs -- similar to this older numpy version bump PR https://github.com/ManageIQ/integration_tests/pull/5194/files

PR #6531, opened 2018-01-26
[1LP][RFR] Add messages to wait_for for snapshot tests

apagac:
This can be considered as fix for all of the existing snapshot tests.

The problem in the Automation Issues spreadsheet is that there are multiple failures like this:
"TimedOutError: Could not do function <lambda>()"
This makes it harder to detect why exactly the failure occurred. It's caused by missing attribute 'message' in function wait_for. This PR adds those messages to wait_for functions with regards to snapshot testing.

There are other snapshot issues currently being fixed, the PRT might not look good for this PR. Reviewers, please keep in mind that this PR only adds text messages to wait_for, nothing else changed.

{{ pytest: -v --long-running --use-provider vsphere55 cfme/tests/infrastructure/test_snapshot.py }}

### DIFF
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -594,7 +594,8 @@ class Vm(VM):
             view.create.click()
             list_view = self.vm.create_view(InfraVmSnapshotView)
             wait_for(lambda: self.exists, num_sec=300, delay=20,
-                     fail_func=list_view.toolbar.reload.click, handle_exception=True)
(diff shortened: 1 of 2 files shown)


apagac:
Only failure is VmOrInstanceNotFound when deleting VM - unrelated to changes introduced in this PR.

jamesooden:
I love this strategy.   I think its so important to always leave enough information to know what is going specifically when a problem occurs, and this is doing just that.

PR #6534, opened 2018-01-26
[WIPTEST] Upping the snapshot create timeout

apagac:
I'm trying to reproduce two snapshot issues from the Automation Issues spreadsheet.

First, there is the issue when waiting for ssh connection times out.
Second, creation of second snapshot in course of the test times out.

I can't reproduce any of this locally but it gives trouble to PRT and jenkins. I have upped the create timeout for all snapshots a bit and will let PRT deal with it and see if I can reproduce any of the two issues.

RESULTS:
 - no error when the create timeout is 400 and provider vsphere55
 - no error when the create timeout is 300 and provider vsphere65-nested
 - no error when the create timeout is 300 and provider rhv41

{{ pytest: -v --long-running --use-provider rhv41 -k test_verify_revert_snapshot }}

### DIFF
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -593,7 +593,7 @@ class Vm(VM):
             view.fill(snapshot_dict)
             view.create.click()
             list_view = self.vm.create_view(InfraVmSnapshotView)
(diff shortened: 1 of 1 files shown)


apagac:
Issue can't be reproduced. Closing this PR

PR #6542, opened 2018-01-29
[1LP][RFR] Fix coverage merger; Log nonzero exit code of ssh commands

mfalesni:


cfme/utils/ssh.py
```diff
@@ -324,6 +324,8 @@ def run_command(
                 if session.exit_status_ready():
                     break
             exit_status = session.recv_exit_status()
+            if exit_status != 0:
+                logger.warning('Exit code %d!', exit_status)
```
jamesooden: I wonder if this should just be a log entry rather than a warning.   Not all non-zero exist codes are flagging a bad thing, or in some cases the bad thing is expected, so in that context its not bad.

mfalesni: This is a log entry, just with the [W] prefix instead of [D] or [I] :)

PR #6538, opened 2018-01-26
[WIPTEST] Refactor freeipa ext auth tests

mshriver:
Refactoring of the authentication automation, including fixtures and tests.

PRT results should be clean, aside from what is skipped.

{{ pytest: cfme/tests/integration/test_cfme_auth.py cfme/tests/integration/test_aws_iam_auth_and_roles.py --long-running }}

PR #6546, opened 2018-01-29
[RFR] Fix collection on auth tests, auth_data.yaml

mshriver:
Fix collection for auth tests, a change to cfme_data.yaml (moving things to auth_data.yaml) broke collection.

I have #6538 open to actually do some auth test refactoring, this is fix for collection.

Only looking for collect-only to be fixed - there likely will be auth test failures, see #6538 for any of those fixes.


mshriver:
Test collection passed for PRT #20664

mshriver:
As this is a PRT blocker, I'm merging as-is. Collection is fixed, some tests are failing with key_error because the credentials were moved out of the unencrypted yaml. #6538 will address these test failures.

PR #6549, opened 2018-01-30
[WIP] pxe_servers gen_func to return list of argvalues

jan-zmeskal:


cfme/utils/testgen.py
```diff
@@ -335,5 +335,4 @@ def pxe_servers(metafunc):
     argnames = ['pxe_name']
 
     items = cfme_data.get('pxe_servers', {}).keys()
-
-    return argnames, [items], items
+    return argnames, items, items
```
lkhomenk: @jan-zmeskal check all others functions - we expect `argvalues` (in pxe_servers case it's `items`)to be a list.

jan-zmeskal: @RonnyPfannschmidt What do you think about this? I think this should not break anything since `cfme_data.get()` returns a list even if it contains only one element and the number of `argnames` is hardcoded.

jan-zmeskal: This change causes problems with setup of PXE tests, looking into it

dajoRH:
Lint report for commit 2a010dc7bd2241f12ea17d01dae6771b3cfa1b03:

`cfme/infrastructure/pxe.py`:
- :bangbang: Line 1066:101: **E501** *line too long (102 > 100 characters)*

**No commit flaws detected**.

*CFME QE Bot*

cfme/utils/testgen.py
lkhomenk: we should either align all code in this module or change the way we work with it `return argnames, argvalues, idlist` further. anyway - don't edit `pxe_servers` only

jan-zmeskal: @lkhomenk This is not actually something I eventually want to merge. I keep it here more as a reference to @RonnyPfannschmidt

jan-zmeskal:
Closing in favor of #6666

PR #6558, opened 2018-01-30
[1LP][RFR] Fixed provision stack key

psav:
IDs had changed

{{pytest: -k test_provision_stack --long-running --use-provider complete}}


psav:
Failures are not related, this fix improves certain tests scores

Issue #6484, opened 2018-01-18
cfme.tests.containers.test_manageiq_ansible_* in need of updates


jawatts:
The cfme.tests.containers.test_manageiq_ansible_* tests are poorly written, which creates inconsistency in the results. These tests need to either be updates or removed.

jawatts:
It was decided by Einat to remove these tests

PR #6564, opened 2018-01-31
[RFR] Updates to cfme.tests.containers.test_manageiq_ansible_*

jawatts:
https://github.com/ManageIQ/integration_tests/issues/6484

{{ pytest: cfme/tests/containers/test_tags.py --use-provider cm-env2 --long-running }}


shalomnaim1:
@jawatts according to the discussion we had with @psav, I think you need to add this test ```@pytest.mark.manual``` decoration and not remove them.

@psav please correct me if I am wrong.

cfme/tests/containers/test_tags.py
```diff
@@ -0,0 +1,98 @@
...
+        d = {}
+        ref = {'name': getattr(selected_entity, 'name', None),
+               'project_name': getattr(selected_entity, 'project_name', None),
+               'host': getattr(selected_entity, 'host', None),
+               'id': getattr(selected_entity, 'id', None),
+               'provider': provider}
+        for arg in ['name', 'project_name', 'host', 'id', 'provider']:
+            if arg in [att.name for att in klass.ENTITY.__attrs_attrs__]:
+                d.update({arg: ref[arg]})
+        return item_collection.instantiate(**d)
```
gshefer: __REQUIRED__
You can change all the above lines to:
```python
d = {}
for arg in ('name', 'project_name', 'host', 'id', 'provider'):
    if arg in [att.name for att in klass.ENTITY.__attrs_attrs__]:
        d[arg] = getattr(selected_entity, arg, None)})
return item_collection.instantiate(**d)
```

cfme/tests/containers/test_tags.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def _tag_cleanup(test_item, tag):
```
gshefer: REQUIRED
Please add a short docstring that describes this function

cfme/tests/containers/test_tags.py
```diff
@@ -0,0 +1,98 @@
...
+    tags = test_item.get_tags()
+    if tags:
+        result = (
+            not object_tags.category.display_name == tag.category.display_name and
+            not object_tags.display_name == tag.display_name for object_tags in tags
+        )
+        if not result:
```
gshefer: __REQUIRED__
Is it possible that you forgot `any` or `all` in the `result`? now `result` is generator and always `True`
```

gshefer: result is always `True`, reason (x for x in <something>) is generator, but probably you wanted to do some operation on it, like `all` or `any`.

jawatts:
@shalomnaim1 Einat wanted the tests completely removed.

mshriver:
@gshefer comments have been addressed, passing tests, merging.

PR #6567, opened 2018-02-01
[1LP][RFR] Increased timeout for wait scvmm to retire

ohiliazo:
__Fixing__ test_resume_retired_instance to wait a bit longer for SCVMM to retire, because it takes more than 10 min.

#### Shriver: 
without a line to specify what provider to use, it will collect against 'default' providers, which doesn't include scvmm.

Leaving this here for you to see and know, but your PR's small timeout change doesn't warrant needing to verify all the results. 
{{ pytest: cfme/tests/cloud_infra_common/test_retirement.py -k test_retirement_now --use-provider scvmm --long-running }}


mshriver:
We might want to refresh relationships via rest and refresh the ui in `fail_func` of the wait, but this is still a reasonable timeout.

PR #6569, opened 2018-02-01
[1LP][RFR] added support of custom data in trackerbot templates

izapolsk:


cfme/utils/trackerbot.py
```diff
@@ -344,13 +347,16 @@ def __init__(self, key):
...
     """dict subclass to help serialize templates as JSON"""
-    def __init__(self, name, group=None, datestamp=None):
+    def __init__(self, name, group=None, datestamp=None, custom_data=None):
         self['name'] = name
         if group is not None:
             self['group'] = group
         if datestamp is not None:
             self['datestamp'] = datestamp.strftime('%Y-%m-%d')
 
+        if custom_data is not None:
```
mshriver: `if custom_data` is preferred.

izapolsk: @mshriver, in this case it is important because custom_data may have {} which is valid.

PR #6575, opened 2018-02-02
[RFR] Firefox focus fix

mfalesni:
{{pytest: cfme/tests/configure/test_access_control.py -v --long-running}}

### DIFF
--- a/cfme/utils/appliance/implementations/ui.py
+++ b/cfme/utils/appliance/implementations/ui.py
@@ -12,7 +12,7 @@ from selenium.common.exceptions import (
     ErrorInResponseException, InvalidSwitchToTargetException,
     InvalidElementStateException, WebDriverException, UnexpectedAlertPresentException,
     NoSuchElementException, StaleElementReferenceException)
-from widgetastic.browser import Browser, DefaultPlugin
+from widgetastic.browser import Browser, DefaultPlugin, BrowserParentWrapper
 from widgetastic.utils import VersionPick
 from widgetastic.widget import Text, View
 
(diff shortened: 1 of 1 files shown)


cfme/utils/appliance/implementations/ui.py
```diff
@@ -184,6 +184,32 @@ def create_view(self, *args, **kwargs):
...
+        else:
+            browser = self
+        result = super(MiqBrowser, browser).handle_alert(*args, **kwargs)
+        if self.browser_type != 'firefox':
+            return result
+
+        # A workaround because of selenium on firefox being a PITA
+        # Open a new window, and close it to regain focus
+        # Workaround by @psav
+        self.logger.debug('Fixing firefox alert focus mess by opening and closing a new window')
```
jamesooden: Is there anyway this might interact badly with our console tests?   In those tests a new tab or window is opened for the console, such that if this code ever happened during the test there would be three windows/tabs open, and you might accidentally close the console tab if somehow this code got called during any of the console tests.  It might be good to change the algorithm to get all the current windows (which in most cases is going to be one) and only close the window that is not in that list.

psav: You know I did that originally, I opted for simpler approach, but I'm sure mfalensi could whip that up real quick :) I just need to get it merged for tonight.

jamesooden: I don't feel really strongly about this...I guess I'm just warning there are other cases where more than one window may already be open and up till very recently that would be unexpected.

psav: Yup

mfalesni: I sincerely hope we can get rid of this nastiness by updating things at some point.

PR #6577, opened 2018-02-02
[WIPTEST] Extending test_change_cpu_ram to check emails

jan-zmeskal:
__Updating test__ to check that there were not any undesirable emails. This is based on [BZ1422208](https://bugzilla.redhat.com/show_bug.cgi?id=1422208).

{{ pytest: cfme/tests/infrastructure/test_provisioning_dialog.py -vv --use-provider rhv41 --long-running }}

### DIFF
--- a/cfme/tests/infrastructure/test_provisioning_dialog.py
+++ b/cfme/tests/infrastructure/test_provisioning_dialog.py
@@ -105,7 +105,7 @@ def provisioner(appliance, request, setup_provider, provider, vm_name):
     return _provisioner
 
 
-def test_change_cpu_ram(provisioner, soft_assert, provider, prov_data, vm_name):
+def test_change_cpu_ram(provisioner, soft_assert, provider, prov_data, vm_name, smtp_test):
     """ Tests change RAM and CPU in provisioning dialog.
 
     Prerequisities:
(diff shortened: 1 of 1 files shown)


lkhomenk:
wow. please DON"T =) extend this test - add separate if you need email verification. it's planned to make those separate anyway. especially if there's a BZ/separate polarion test case

jan-zmeskal:
@lkhomenk Ok, I will make it into separate test. Could you however look at trackerbot and help my solving why the mails are not received? When I run it locally, I receive both mails much sooner than in 90 secs.

jan-zmeskal:
Actually, could I know your opinion on this, @psav? I was thinking about this during weeknd. And I know it is generally good idea to create one test case for one issue. But that new test case would literally be the same as test_change_cpu_ram. It would be only checking received emails on top of that.

cfme/tests/infrastructure/test_provisioning_dialog.py
```diff
@@ -143,6 +152,8 @@ def test_change_cpu_ram(provisioner, soft_assert, provider, prov_data, vm_name):
         raise ValueError("Could not parse string {}".format(repr(data)))
     soft_assert(num_cpus == "4", "num_cpus should be {}, is {}".format("4", num_cpus))
     soft_assert(memory == "2048 MB", "memory should be {}, is {}".format("2048 MB", memory))
+    wait_for(check_approval_mail_received, num_sec=90, delay=5)
+    wait_for(check_completed_mail_received, num_sec=90, delay=5)
```
psav: **QUESTION**

Was asked to put my opinion on this code here:

```wait_for(lambda: smtp_test.get_emails(subject_like="%%Your Virtual Machine configuration was
             Approved%%"), num_sec=90, delay=5, message="Wait for approval mail")
```

lkhomenk:
if we fail on email verification that would mean we failed on VM provision, which is not true. Vm provision functionality is crucial, I would expect to not overload this tests. your test needs VM provision as kinda fixture. you could also use REST provision to save time navigating etc.

jan-zmeskal:
@lkhomenk Ok, the point with REST provision and fixture approach is a fair one, I'll look into that. Anyway, do you know why the emails have not been received in PRT run for this PR? Is that some know issue I have to work around? Since I had no problem with receiving those emails while running locally.

PR #6581, opened 2018-02-02
[1LP][RFR] Fixed test_control_identical_descriptions

quarckster:
{{pytest: -v -k test_control_identical_descriptions}}


cfme/tests/control/test_bugs.py
```diff
@@ -272,8 +274,11 @@ def test_control_identical_descriptions(request, create_function, collections):
         The item shouldn't be created.
     """
     args, kwargs = create_function.fn(request, collections[create_function.name])
-    with pytest.raises(AssertionError):
```
quarckster: After merging #6479 `TimedOutError` can be raised.

PR #6587, opened 2018-02-05
[WIPTEST] Fix for create_on_provider

lkhomenk:
Purpose or Intent
=================

__Fixing__  `create_on_provider` right not it can cause test setup failure and leave VM on the environment
All test will be using it same way but additionally if it fails - it will delete vm
{{pytest: -v --long-running cfme/cloud/test_provisioning.py cfme/tests/cloud/test_instance_power_control.py --use provider complete}}


cfme/common/vm.py
```diff
@@ -632,17 +635,23 @@ def delete_from_provider(self):
         else:
             return True
 
-    def create_on_provider(self, timeout=900, find_in_cfme=False, **kwargs):
+    def create_on_provider(self, timeout=900, find_in_cfme=False, delete_on_failure=True, **kwargs):
         """Create the VM on the provider
 
         Args:
             timeout: Number of seconds to wait for the VM to appear in CFME
                      Will not wait at all, if set to 0 (Defaults to ``900``)
```
jkrocil: **REQUIRED**
Please add comments for `find_in_cfme` and `delete_on_failure`.

cfme/common/vm.py
```diff
@@ -632,17 +635,23 @@ def delete_from_provider(self):
...
-            self.wait_to_appear(timeout=timeout, load_details=False)
+        try:
+            if find_in_cfme:
+                self.wait_to_appear(timeout=timeout, load_details=False)
+        except Exception as e:
+            logger.warn("Couldn't find VM or Instance in CMFE")
+            if delete_on_failure:
+                logger.info("Removing Vm from mgmt system")
+                self.provider.mgmt.delete_vm(self.name)
+            raise e
```
jkrocil: **REQUIRED**
So when we don't attempt to "find_in_cfme", there is no deletion, even if the provisioning failed. We gotta mention that in the comment above.

lkhomenk: `deploy_template` can take care of VMs provisioned in Error/Failure state. and this `try..except` handles any UI issues we might face during UI verification of VM

PR #6589, opened 2018-02-05
[1LP][RFR] Add BZ blocker for category in fixtures/tag.py

nzr-zts:
Purpose
=================
In case if Category name starts from digit it will be on the top of the list and selected by default as the first one. Because of [BZ#1517285](https://bugzilla.redhat.com/show_bug.cgi?id=1517285), it will have other Tags than preconfigured for this Category.

{{pytest: -v cfme/tests/cloud_infra_common/test_tag_objects.py -k test_tag_infra_objects}}

### DIFF
--- a/cfme/fixtures/tag.py
+++ b/cfme/fixtures/tag.py
@@ -15,7 +15,8 @@ def category():
         Object can be used in all test run session
     """
     if BZ(1517285, forced_streams='5.9').blocks:
-        display_name = 'test-{}'.format(fauxfactory.gen_alphanumeric(length=32))
+        display_name = 'test-{}'.format(fauxfactory.gen_alphanumeric(length=27))
+    # display_name should be with max length of 32
     else:
         display_name = fauxfactory.gen_alphanumeric(length=32)
     cg = Category(name=fauxfactory.gen_alpha(8).lower(),
--- a/cfme/fixtures/tag.py
+++ b/cfme/fixtures/tag.py
@@ -15,13 +15,12 @@ def category():
         Object can be used in all test run session
     """
     if BZ(1517285, forced_streams='5.9').blocks:
-        cg = Category(name=fauxfactory.gen_alpha(8).lower(),
-                      description=fauxfactory.gen_alphanumeric(length=32),
-                      display_name='test-{}'.format(fauxfactory.gen_alphanumeric(length=32)))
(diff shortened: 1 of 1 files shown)


cfme/fixtures/tag.py
```diff
@@ -13,9 +14,14 @@ def category():
...
-                  description=fauxfactory.gen_alphanumeric(length=32),
-                  display_name=fauxfactory.gen_alphanumeric(length=32))
+    if BZ(1517285, forced_streams='5.9').blocks:
+        cg = Category(name=fauxfactory.gen_alpha(8).lower(),
+                      description=fauxfactory.gen_alphanumeric(length=32),
+                      display_name='test-{}'.format(fauxfactory.gen_alphanumeric(length=32)))
+    else:
+        cg = Category(name=fauxfactory.gen_alpha(8).lower(),
+                      description=fauxfactory.gen_alphanumeric(length=32),
+                      display_name=fauxfactory.gen_alphanumeric(length=32))
```
quarckster: **OPTIONAL**
```python
if BZ(1517285, forced_streams='5.9').blocks:
    display_name = 'test-{}'.format(fauxfactory.gen_alphanumeric(length=32))
else:
    display_name = fauxfactory.gen_alphanumeric(length=32)
cg = Category(name=fauxfactory.gen_alpha(8).lower(),
                       description=fauxfactory.gen_alphanumeric(length=32),
                       display_name=fauxfactory.gen_alphanumeric(length=32))
```

cfme/fixtures/tag.py
```diff
@@ -13,9 +14,13 @@ def category():
         Returns random created category object
         Object can be used in all test run session
     """
+    if BZ(1517285, forced_streams='5.9').blocks:
+        display_name = 'test-{}'.format(fauxfactory.gen_alphanumeric(length=32))
```
mshriver: **Required**
Drop the length passed to aphanumeric from `32` to `27` to account for the characters in 'test-' and in both cases use a 32 character name for testing.

jamesooden: I would go a step further and make that number 27 be calculated, and leave a comment about the field it is going into being a max length of 32.  This makes the number 27 completely unmysterious, and answers any questions someone reading the code would have.

PR #6592, opened 2018-02-06
[WIPTEST] Snapshots: reproducing issue from spreadsheet

apagac:
I need to run PRT with one of the snapshot tests and see if I can reproduce the issue that occurs in jenkins. Can't reproduce locally.

{{ pytest: -v --long-running --use-provider vsphere6-nested -k test_operations_suspended_vm }}


apagac:
The issue can't be reproduced. Closing the PR.

PR #6597, opened 2018-02-06
[1LP][RFR] Fixed test_tagdialog_catalog_item

quarckster:
{{pytest: -v -k test_tagdialog_catalog_item --long-running --use-provider vsphere6-nested}}

PR #6598, opened 2018-02-06
[RFR] Added select field which got removed in recent changes

sshveta:
{{pytest: -v --use-provider vsphere cfme/tests/services/test_different_dialogs_in_catalogs.py}}


quarckster:
Already fixed in #6597

PR #6599, opened 2018-02-06
[WIPTEST] Jenkins failure Name_is_required fix

sshveta:
After filling the orchestration template form , the name field becomes empty sometimes and throws assertion error "Name is required" .
This PR is to try and fix that failure .


jamesooden:
You've added code to try to alleviate this problem but no comments in the code to explain the problem you are trying to avoid.   These comments could be helpful to someone else evaluating the necessity of the approach you have chosen at a later time.

sshveta:
@jamesooden : Yeah it's still WIP , working on this PR , will update it in sometime.

PR #6604, opened 2018-02-07
[1LP][RFR] Glitch master hack

psav:
Hack to help with the FF issue


cfme/utils/appliance/implementations/ui.py
```diff
@@ -111,9 +111,28 @@ class MiqBrowserPlugin(DefaultPlugin):
     )
     DEFAULT_WAIT = .8
 
+    def make_document_focused(self):
```
jamesooden: Could you add a comment explaining what this is fixing and how it helps?

PR #6612, opened 2018-02-08
[1LP][RFR] Add auth provider classes

mshriver:
Auth provider classes for the MIQ and external auth providers
registered with entrypoints for dynamic type fetching
get_auth_crud method for easy from_config() access

To support auth test refactoring in #6538 

PRT irrelevant

### DIFF
--- a/cfme/utils/auth/__init__.py
+++ b/cfme/utils/auth/__init__.py
@@ -1,0 +1,178 @@
+from six import iteritems
+import attr
+
+from cfme.configure.configuration.server_settings import (
+    AmazonAuthenticationView, LdapAuthenticationView, LdapsAuthenticationView
+)
+from cfme.exceptions import UnknownProviderType
+from cfme.utils.conf import credentials, auth_data
+
+auth_prov_data = auth_data.get("auth_providers", {})  # setup on module import
+
+USER_TYPES = {
+    'principal': 'User Principal Name',
+    'email': 'E-mail Address',
+    'dn-cn': 'Distinguished Name (CN=<user>)',
+    'dn-uid': 'Distinguished Name (UID=<user>)',
+    'sam': 'SAM Account Name'
+}
+LDAP_PORT = 389
+LDAPS_PORT = 636
+
+
+def auth_provider_types():
+    """Fetch the registered classes from entry_points manageiq.auth_provider_categories"""
+    from pkg_resources import iter_entry_points
+    return {
+        ep.name: ep.resolve()
+        for ep in iter_entry_points('manageiq.auth_provider_types')
+    }
+
+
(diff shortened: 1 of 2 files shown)


cfme/utils/auth/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+
+@attr.s
+class MIQAuthProvider(BaseAuthProvider):
+    """base class for miq auth providers (ldap/ldaps modes in UI)
+    Intended to be used for freeipa, AD, openldap and openldaps type providers
+    """
+    auth_type = None
+    host1 = attr.ib()
+    bind_password = attr.ib()  # Ordered to adhere to mandatory attrs sequence
+    host2 = attr.ib(default=None)
```
izapolsk: maybe hosts = [] ?

cfme/utils/auth/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+    cert_filepath = attr.ib(default=None)
+    ipaddress = attr.ib(default=None)
+    ldap_conf = attr.ib(default=None)
+    sssd_conf = attr.ib(default=None)
+
+    # TODO as_external_value method
+
+
+@attr.s
+class OpenldapAuthProvider(MIQAuthProvider):
```
izapolsk: I'd make LDAP/IPA upper case since those are accronims

cfme/utils/auth/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+    return klass.from_config(auth_prov_config, auth_prov_key)
+
+
+@attr.s
+class BaseAuthProvider(object):
+    """Base class for authentication provider objects
+
+    Might not be needed
+    """
+    auth_type = None
```
izapolsk: attr.ib(default=None) ?

cfme/utils/auth/__init__.py
mshriver: The older auth config actually kind of had a list of hosts - but it created complexities otherwise that made me prefer specifying the hosts as 1/2/3 - this also allows controlling their configuration order a little easier.

cfme/utils/auth/__init__.py
mshriver: ah you're right, LDAP is capitalized in openLDAP's brand materials. :+1:

cfme/utils/auth/__init__.py
mshriver: I could use attr.ib for this - but I don't want this passed during instantiation so I'd need `init=False` option for attr.ib.  I'll check it out.

mshriver:
@izapolsk addressed comments, any other thoughts?

PR #6613, opened 2018-02-08
[1LP][RFR] Update appliance openldap config method

mshriver:
Update appliance.configure_openldap
To support more consistent auth automation and setup for RBAC tests

update auth_data/cfme_data queries in utils/testgen

To support auth test refactoring in #6538, will be rebasing and following up anything from here in that PR.

PRT failures are pseudo-random smoke test failures from timeouts, not related

### DIFF
--- a/cfme/utils/appliance/__init__.py
+++ b/cfme/utils/appliance/__init__.py
@@ -179,21 +179,19 @@ class ApplianceConsoleCli(object):
             " --dbname {dbname} --verbose --dbdisk {dbdisk} --key --standalone".format(
                 region=region, username=username, password=password, dbname=dbname, dbdisk=dbdisk))
 
-    def configure_ipa(self, ipaserver, username, password, domain, realm):
-        self._run("--ipaserver {ipaserver} --ipaprincipal {username} --ipapassword {password}"
-            " --ipadomain {domain} --iparealm {realm}".format(
-                ipaserver=ipaserver, username=username, password=password, domain=domain,
-                realm=realm))
-        assert self.appliance.ssh_client.run_command("systemctl status sssd | grep running")
-        return_code, output = self.appliance.ssh_client.run_command(
-            "cat /etc/ipa/default.conf | grep 'enable_ra = True'")
-        assert return_code == 0
+    def configure_ipa(self, ipaserver, username, password, domain=None, realm=None):
+        assert self._run("--ipaserver {ipaserver} --ipaprincipal {username} "
+                         "--ipapassword {password} {domain} {realm}"
+                         .format(ipaserver=ipaserver, username=username, password=password,
+                                 domain='--ipadomain {}'.format(domain) if domain else '',
+                                 realm='--iparealm {}'.format(realm) if realm else ''))
+        assert self.appliance.sssd.running
+        assert self.appliance.ssh_client.run_command("cat /etc/ipa/default.conf "
+                                                     "| grep 'enable_ra = True'")
 
     def uninstall_ipa_client(self):
-        self._run("--uninstall-ipa")
-        return_code, output = self.appliance.ssh_client.run_command(
-            "cat /etc/ipa/default.conf")
-        assert return_code != 0
+        assert self._run("--uninstall-ipa")
+        assert not self.appliance.ssh_client.run_command("cat /etc/ipa/default.conf")
 
 
 class IPAppliance(object):
@@ -219,6 +217,8 @@ class IPAppliance(object):
     _nav_steps = {}
 
     evmserverd = SystemdService.declare(unit_name='evmserverd')
+    httpd = SystemdService.declare(unit_name='httpd')
+    sssd = SystemdService.declare(unit_name='sssd')
     db = ApplianceDB.declare()
 
     CONFIG_MAPPING = {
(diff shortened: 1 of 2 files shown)


cfme/utils/appliance/__init__.py
```diff
@@ -2137,45 +2146,116 @@ def check_domain_enabled(self, domain):
...
+        # ext auth ipa requires NTP sync
+        if ipaserver not in self.server.settings.ntp_servers_values:
+            self.server.settings.update_ntp_servers({'ntp_server_1': ipaserver})
+            # sleep(120)
+
+        # backend appliance configuration of ext auth provider
+        self.appliance_console_cli.configure_ipa(
+            ipaserver, principal, password,
+            kwargs.pop('ipadomain') if 'ipadomain' in kwargs else '',  # pop, not in ui form
+            kwargs.pop('iparealm') if 'iparealm' in kwargs else ''
```
quarckster: **QUESTION**
Do you really need to pop here? Otherwise, you can do like this:
```python
kwargs.get('ipadomain', ''),
kwargs.get('iparealm', '')
```

mshriver: I used pop here because the ipadomain/iparealm aren't used in `authentication.configure_auth(auth_mode='external', **kwargs)` below.

This definitely has room for improvement, and I plan on that in #6538. These methods should use the new auth provider objects (#6612), which provide `as_fill_value()` and `as_external_value()` methods.  I split some of these changes into separate PRs for easier initial reviews. Would you rather see some different syntax here in the meantime?

mshriver:
@quarckster you can re-review if you feel the need - I dropped the freeipa methods from this PR as they're not actually in use yet.

cfme/utils/appliance/__init__.py
```diff
@@ -179,21 +179,19 @@ def configure_appliance_dedicated_db(self, region, username, password, dbname, d
...
-    def configure_ipa(self, ipaserver, username, password, domain, realm):
-        self._run("--ipaserver {ipaserver} --ipaprincipal {username} --ipapassword {password}"
-            " --ipadomain {domain} --iparealm {realm}".format(
-                ipaserver=ipaserver, username=username, password=password, domain=domain,
-                realm=realm))
-        assert self.appliance.ssh_client.run_command("systemctl status sssd | grep running")
-        return_code, output = self.appliance.ssh_client.run_command(
-            "cat /etc/ipa/default.conf | grep 'enable_ra = True'")
-        assert return_code == 0
+    def configure_ipa(self, ipaserver, username, password, domain=None, realm=None):
```
izapolsk: Mike wouldn't it be better to move all auth related stuff to appliance plugin ?

mshriver: Really like this idea, discussed offline, **optional** for this PR - but we'll be moving things later.

cfme/utils/appliance/__init__.py
```diff
@@ -2135,45 +2144,71 @@ def check_domain_enabled(self, domain):
...
-
+            Args:
+                ipaddress: ip of the openldap server
+                hostname: hostname of the openldap server
+                cert_filepath: optional path for placing TLS/SSL cert on appliance
+                cert_filename: optional name of the cert file in conf_path
+                kwargs: additional args for UI auth configuration (3 booleans)
         """
-        ldap_yaml = conf.cfme_data.auth_modes.ext_openldap
+        # TODO refactor using passed auth provider object instead of passed fields
```
izapolsk: :+1:

cfme/utils/appliance/__init__.py
```diff
@@ -2135,45 +2144,71 @@ def check_domain_enabled(self, domain):
...
         # write /etc/hosts entry for ldap hostname  TODO DNS
-        self.ssh_client.run_command('echo "{}    {}" >> /etc/hosts'
-                                    .format(ldap_yaml.ipaddress, ldap_yaml.hostname))
+        self.ssh_client.run_command('echo "{}  {}" >> /etc/hosts'
```
izapolsk: is Matt working on adding our ldap to dns ?

cfme/utils/appliance/__init__.py
```diff
@@ -2135,45 +2144,71 @@ def check_domain_enabled(self, domain):
...
+                                    .format(ipaddress, hostname))
         # place cert from local conf directory on ldap server
-        self.ssh_client.put_file(local_file=conf_path.join(ldap_yaml.cert_filename).strpath,
-                                 remote_file=ldap_yaml.cert_filepath)
+        self.ssh_client.put_file(local_file=conf_path.join(cert_filename).strpath,
```
izapolsk: what's wrong with certs ?

cfme/utils/appliance/__init__.py
```diff
@@ -2135,45 +2144,71 @@ def check_domain_enabled(self, domain):
...
-            'cp {template_dir}/etc/httpd/conf.d/manageiq-external-auth.conf.erb {manageiq_ext_auth}'
-            .format(template_dir=template_dir,
-                    manageiq_ext_auth=manageiq_ext_auth))
-        self.ssh_client.run_command('setenforce 0 && '
-                                    'systemctl restart sssd && '
-                                    'systemctl restart httpd')
+            'cp {t}{c}.erb {c}'
+            .format(t=template_dir, c=self.CONF_FILES['httpd_ext_auth'])
+        )
+        assert self.ssh_client.run_command('setenforce 0')
```
izapolsk: I would rather check selinux state instead. I doubt it will work in containers.

cfme/utils/appliance/__init__.py
mshriver: Not sure exactly what you're asking here, can you clarify please?

cfme/utils/appliance/__init__.py
mshriver: This comes straight from the MIQ documentation on LDAP configuration:

http://manageiq.org/docs/reference/latest/auth/ldap

cfme/utils/appliance/__init__.py
mshriver: Yes - new auth servers that we deploy will be in DNS. This is an additional configuration to ensure that the given host is resolvable even if not in DNS. We could dynamically decide to add this based on whether or not hostname is resolvable on the appliance though.

cfme/utils/appliance/__init__.py
izapolsk: okay, lets leave it until we run into some issue

cfme/utils/appliance/__init__.py
izapolsk: As far as I see you append additional cert to appliance. 
It may happen because 
1. cert is used for authentication 
or 
2. LDAP server's certificates aren't configured propertly

PR #6615, opened 2018-02-09
[1LP][RFR] Using default vNIC profile in REST VM provision

jan-zmeskal:
This PR has three parts:
1) __Fixing__ provisioning data in `test_provisioning_rest.py` to be compliant with new definition of vlan attribute: https://bugzilla.redhat.com/show_bug.cgi?id=1541036#c4

2) __Moving__ some helper functions from test_provision to module level, since I will need them for future tests as well and don't want to duplicate code.

3) __Removing__ BZ wrapper. The referenced BZ is still not solved, but it became clear that it is not actually a blocker.

Note: This still does not work for vsphere. I tried to do a quick fix but it didn't work. Unfortunately I don't really have the resources for fixing vsphere.

### DIFF
--- a/cfme/tests/infrastructure/test_provisioning_rest.py
+++ b/cfme/tests/infrastructure/test_provisioning_rest.py
@@ -79,6 +79,13 @@ def clean_vm(provider, vm_name):
     provider.mgmt.delete_vm(vm_name)
 
 
+def request_finished(provision_request):
+    provision_request.reload()
+    if "error" in provision_request.status.lower():
+        pytest.fail("Error when provisioning: `{}`".format(provision_request.message))
+    return provision_request.request_state.lower() in ("finished", "provisioned")
+
+
 # Here also available the ability to create multiple provision request, but used the save
 # href and method, so it doesn't make any sense actually
 def test_provision(request, provision_data, provider, appliance):
@@ -100,13 +107,9 @@ def test_provision(request, provision_data, provider, appliance):
     assert appliance.rest_api.response.status_code == 200
     provision_request = response[0]
 
-    def _finished():
-        provision_request.reload()
-        if "error" in provision_request.status.lower():
-            pytest.fail("Error when provisioning: `{}`".format(provision_request.message))
-        return provision_request.request_state.lower() in ("finished", "provisioned")
+    wait_for(request_finished, func_args=[provision_request], num_sec=800, delay=5,
+        message="REST provisioning finishes")
 
-    wait_for(_finished, num_sec=800, delay=5, message="REST provisioning finishes")
     assert provider.mgmt.does_vm_exist(vm_name), "The VM {} does not exist!".format(vm_name)
 
 
--- a/cfme/tests/infrastructure/test_provisioning_rest.py
+++ b/cfme/tests/infrastructure/test_provisioning_rest.py
@@ -15,6 +15,9 @@ pytestmark = [
     pytest.mark.meta(server_roles="+automate"),
     pytest.mark.usefixtures("setup_provider"),
     pytest.mark.provider([VMwareProvider, RHEVMProvider], scope="module"),
+    pytest.mark.meta(blockers=[BZ(1541036,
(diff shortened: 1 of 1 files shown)


cfme/tests/infrastructure/test_provisioning_rest.py
```diff
@@ -73,6 +74,20 @@ def provision_data(appliance, provider, small_template_modscope):
     return get_provision_data(appliance.rest_api, provider, small_template_modscope.name)
 
 
+def clean_vm(provider, vm_name):
+    if not provider.mgmt.is_vm_stopped(vm_name):
```
lkhomenk: https://github.com/lkhomenk/integration_tests/blob/1bba00780339341dc1c259f9d13db0119d893686/cfme/common/vm.py#L611 - we already have this. It might be useful

cfme/tests/infrastructure/test_provisioning_rest.py
```diff
@@ -73,6 +74,20 @@ def provision_data(appliance, provider, small_template_modscope):
...
+        provider.mgmt.stop_vm(vm_name)
+        provider.mgmt.wait_vm_stopped(vm_name)
+    provider.mgmt.delete_vm(vm_name)
+
+
+def request_finished(provision_request):
```
lkhomenk: wait_for + https://github.com/lkhomenk/integration_tests/blob/ffac0136cf5bc87b457a334094dbd29d838ee800/cfme/services/requests.py#L186 also might be useful. it also uses REST by default

cfme/tests/infrastructure/test_provisioning_rest.py
```diff
@@ -57,14 +54,18 @@ def get_provision_data(rest_api, provider, template_name, auto_approve=True):
...
+            "placement_auto": "true"
         },
         "ems_custom_attributes": {},
         "miq_custom_attributes": {}
     }
 
     if provider.one_of(RHEVMProvider):
         result["vm_fields"]["provision_type"] = "native_clone"
+        if current_version() > '5.9.0.16':
+            result["vm_fields"]["vlan"] = "<Template>"
```
lkhomenk: **QUESTION**: does this affect infra providers only?

mshriver: **Required**
Regardless, please use provider.appliance.version instead of current_version, even though its not a new import here.

cfme/tests/infrastructure/test_provisioning_rest.py
mshriver: I'd agree with @lkhomenk here unless there's something preventing use of `delete_from_provider()`  please use that (or modify it to support your use case).

cfme/tests/infrastructure/test_provisioning_rest.py
mshriver: This is also checking the task state, but that is missing from this test anyway from the looks of it.  @jan-zmeskal please investigate and consider this change to using Requests methods.

cfme/tests/infrastructure/test_provisioning_rest.py
```diff
@@ -73,6 +74,18 @@ def provision_data(appliance, provider, small_template_modscope):
...
+def clean_vm(vm_name, provider):
+    vm_obj = VM.factory(vm_name=vm_name, provider=provider)
+    vm_obj.delete_from_provider()
+
+
+def request_finished(provision_request):
+    provision_request.reload()
+    if "error" in provision_request.status.lower():
+        pytest.fail("Error when provisioning: `{}`".format(provision_request.message))
+    return provision_request.request_state.lower() in ("finished", "provisioned")
```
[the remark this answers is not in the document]
lkhomenk:
```
    request_description = 'Provision from [{}] to [{}]'.format(image, instance.name)
    provision_request = appliance.collections.requests.instantiate(request_description)
    provision_request.wait_for_request()
```
you can also use `partial_check` during `instantiate` - but request should have unique description

PR #6628, opened 2018-02-11
[1LP][RFR] Cloud provider discovery test

digitronik:
Purpose or Intent
=================
- Added Cloud provider discovery test for `Azure` and `AWS`; supported for `version  < 5.9`
- Small change in `Credential` Class for Azure support
- Fix in `discover_dis()`; changed key to `confirm_password` as per `CloudProvidersDiscoverView()`

{{py.test: -v cfme/tests/cloud/test_providers.py --use-provider complete}}

### DIFF
--- a/cfme/common/provider_views.py
+++ b/cfme/common/provider_views.py
@@ -16,7 +16,6 @@ from widgetastic_manageiq import (BreadCrumb,
                                   ItemsToolBarViewSelector,
                                   Checkbox,
                                   Input,
-                                  Table,
                                   BaseEntitiesView,
                                   PaginationPane,
                                   BaseTileIconEntity,
@@ -314,6 +313,7 @@ class ProviderVmsView(ProviderVmsTemplatesView):
         return (self.logged_in_as_current_user and
                 self.navigation.currently_selected == ['Compute', 'Infrastructure', 'Providers'] and
                 self.title.text == title)
+
 
 class ProviderToolBar(View):
     """
--- a/cfme/tests/cloud/test_providers.py
+++ b/cfme/tests/cloud/test_providers.py
@@ -96,6 +96,38 @@ def test_providers_discovery_amazon(appliance):
     wait_for_a_provider()
 
 
+@pytest.mark.uncollectif(lambda provider: (store.current_appliance.version >= '5.9' or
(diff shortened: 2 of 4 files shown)


digitronik:
As per my observation, PRT failures are not due to changes. 
5.9 most of failures are due to  'MoveTargetOutOfBoundsException'. 
5.8 ec2 crud failure.
Moving to RFR. If my bad, let me know as changes in the common file.

cfme/common/provider_views.py
```diff
@@ -16,7 +16,6 @@
                                   ItemsToolBarViewSelector,
                                   Checkbox,
                                   Input,
-                                  Table,
```
quarckster: **QUESTION**
Why is it removed?

cfme/tests/cloud/test_providers.py
```diff
@@ -96,6 +96,38 @@ def test_providers_discovery_amazon(appliance):
     wait_for_a_provider()
 
 
+@pytest.mark.uncollectif(lambda provider: (store.current_appliance.version >= '5.9' or
```
quarckster: **QUESTION**
Why do we need a 5.8 only test?

cfme/common/provider_views.py
digitronik: It's due to my flake8 git hook.
I had two option while fixing `discover_dict()` method of ec2:
1. change key of `discover_dict` from `password_verify` to `confirm_password`
2. Add changes in `CloudProvidersDiscoverView` and change `confirm_password` to `password_verify`

First I follow the second method and push changes. but found common practice is to used `confirm_password`. So revert changes but due to lint problem (l.e. table is imported but not used); I have to go with lint correction now.

cfme/tests/cloud/test_providers.py
digitronik: @quarckster :
The test is for cloud provider discovery. The discovery feature is support for 5.8 but not in 5.9. They just removed discovery option for cloud provider l.e. both Azure and AWS.

mshriver: **Required**

Use `provider.appliance.version` instead of store.current_appliance

PR #6630, opened 2018-02-12
[1LP][RFR] Change login method for FF52+

rlbabyuk:
Purpose or Intent
=================
Updated login default method to use Enter button instead of running js, as as js login was not successful for FF versions 52+ and selenium3


quarckster:
Will it work with Chrome/Chromium?

rlbabyuk:
Yes it is, also tested for chrome

mfalesni:
I think we shouldn't've been using the js method anyway :)

PR #6633, opened 2018-02-12
[1LP][RFR] Blocking test_vm_reconfig_add_remove_hw_cold for RHV

jan-zmeskal:
__Blocking__ test_vm_reconfig_add_remove_hw_cold. I changed the test case to *decrease* memory instead of increase few weeks ago. I did so because I thought that RHV does not support increasing VM's memory over the max size. Well, it turns out that it does and this is indeed a [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1534520). Therefore reverting the test case to its original state and wrapping it with BZ.

{{pytest: cfme/tests/infrastructure/test_vm_reconfigure.py::test_vm_reconfig_add_remove_hw_cold -vv --use-provider rhv41 --long-running}}


jan-zmeskal:
@izapolsk Done

PR #6637, opened 2018-02-12
[1LP][RFR] Enabled test for 5.9

sshveta:


cfme/services/service_catalogs/ssui.py
```diff
@@ -86,10 +86,7 @@ class ShoppingCartView(DetailsServiceCatalogsView):
 
     @property
     def is_displayed(self):
-        return (
-            self.in_service_catalogs and
```
quarckster: **QUESTION**
Why is it removed?

sshveta: when the shopping cart button is clicked , a window opens on the top of service catalog page and hence , in_service_catalog becomes false

PR #6642, opened 2018-02-13
[RFR] test name change for cloud and infra respectively 

digitronik:
Purpose or Intent
=================
The old names make confusion in polarion. separating cloud and infra tests.

### DIFF
--- a/cfme/tests/cloud/test_providers.py
+++ b/cfme/tests/cloud/test_providers.py
@@ -33,7 +33,7 @@ pytestmark = [pytest.mark.provider([CloudProvider], scope="function")]
 @test_requirements.discovery
 @pytest.mark.uncollectif(lambda: store.current_appliance.version >= '5.9',
                          reason='no more support for cloud provider discovery')
-def test_empty_discovery_form_validation(appliance):
+def test_empty_discovery_form_validation_cloud(appliance):
     """ Tests that the flash message is correct when discovery form is empty."""
     discover(None, AzureProvider)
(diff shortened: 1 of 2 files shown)


cfme/tests/cloud/test_providers.py
```diff
@@ -70,7 +70,7 @@ def test_add_cancelled_validation(request):
 @pytest.mark.tier(3)
 @pytest.mark.uncollectif(lambda: store.current_appliance.version >= '5.9',
                          reason='no more support for cloud provider discovery')
-def test_password_mismatch_validation(appliance):
+def test_discovery_password_mismatch_validation(appliance):
```
rlbabyuk: **Question**: Doesn't 'cloud' also needed here?

quarckster:
What about Polarion? You need also change the names and respectful fields there as well.

cfme/tests/cloud/test_providers.py
digitronik: @rlbabyuk: I think good to go with discovery only,  as for infra provider discovery not need of password verification. Only cloud provider discovery we need it; specific AWS only. Azure does not support this.

mshriver:
@digitronik who is assigned these cases in Polarion? Have you engaged with them to assign these new cases as they come in, and disable the old?

digitronik:
@mshriver : All test cases related to Discovery and provider add screen, so I have some and remaining owning from @niyazRedhat

PR #6645, opened 2018-02-13
[1LP][RFR] Fix scripts imports

mshriver:
The new openshift* package has a 'scripts' package that is getting picked for package/module resolution on imports. This breaks several of our imports `from scripts (...)`

Moving only the impacted modules out of scripts and into cfme/utils so they can be reliably/absolutely imported.


setup.py
```diff
@@ -22,7 +22,7 @@
             'miq-artifactor-server = artifactor.__main__:main',
```
jkrocil: **REQUIRED**
L#21 ^ `'miq-release = scripts.release:main'`

PR #6647, opened 2018-02-14
[1LP][RFR] fix provider api port test

jadhaj:
Purpose or Intent
=================
Replaced access to provider api port through details view instead of provider.summary.properties.api_port.value 

{{ pytest: cfme/tests/openstack/infrastructure/test_provider.py::test_api_port }}

### DIFF
--- a/cfme/tests/openstack/infrastructure/test_provider.py
+++ b/cfme/tests/openstack/infrastructure/test_provider.py
@@ -13,8 +13,9 @@ pytestmark = [
 
 
 def test_api_port(provider):
+    view_details = navigate_to(provider, 'Details')
     port = provider.data['endpoints']['default']['api_port']
-    assert provider.summary.properties.api_port.value == port, 'Invalid API Port'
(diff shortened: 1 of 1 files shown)


jamesooden:
Quick question.   Why did we make this change?   Was it because the old way was broken, the new way is better for some reason, or something else?    Whatever the answer is, that would a good thing to have in the description of this PR and its commit.

mfalesni:
The problem was that when @izapolsk first wrote it, he let the scavenged string be converted into an int. Which made sense, as if you have Region 0, all ids start with 1. Now if you configure CF with region id > 0, the sequential id is calculated as region_id * 10^12 + seq_id, and they started to contract these long numbers as eg. (if region id == 1) `1r1`, `1r2`, ...

Therefore, when running against such appliance, these errors happened. The solution was simple, just don't cast it, since at no point of the entity_id's existence nothing cares about its type assuming it can just be put as a string format.

PR #6648, opened 2018-02-14
[1LP][RFR] Add services custom attributes tests

mkoura:
{{pytest: -v --long-running -k services cfme/tests/cloud_infra_common/test_custom_attributes_rest.py}}

### DIFF
--- a/cfme/tests/cloud_infra_common/test_custom_attributes_rest.py
+++ b/cfme/tests/cloud_infra_common/test_custom_attributes_rest.py
@@ -21,8 +21,9 @@ pytestmark = [
     test_requirements.rest
 ]
 
-COLLECTIONS = ['providers', 'instances', 'vms']
-COLLECTIONS_ADDED_IN_59 = ['instances']
+COLLECTIONS = ['providers', 'vms']
+COLLECTIONS_ADDED_IN_59 = ['instances', 'services']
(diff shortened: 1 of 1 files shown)


cfme/tests/cloud_infra_common/test_custom_attributes_rest.py
```diff
@@ -63,12 +64,37 @@ def _get_vm():
...
+        service = appliance.rest_api.collections.services.find_by(name=name)
+        if not service:
+            body = {
+                'name': name,
+                'description': 'Test REST Service {}'.format(uid),
+            }
+            service = appliance.rest_api.collections.services.action.create(body)
+        return service[0]
+
+    yield _get_service
```
jamesooden: Why do we yield here?

mshriver: @jamesooden So we're yielding the function here, and its called later down on L 145: 
https://github.com/ManageIQ/integration_tests/pull/6648/files#diff-82f0d8e84509575e71ce820205c41e29R145

The yield allows for cleanup after the test scope is completed. In this case, the fixture is module scoped, so after all of the test functions (parametrized) have executed, the module will close out by returning to here and executing the try/except block below to delete the service.

PR #6650, opened 2018-02-14
[1LP][RFR] Ability to isolate browser sessions

mfalesni:
Tested locally.

### DIFF
--- a/cfme/test_framework/browser_isolation.py
+++ b/cfme/test_framework/browser_isolation.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+"""Plugin enabling us to isolate browser sessions per test."""
+import pytest
+
+
+def pytest_addoption(parser):
+    parser.addoption(
+        '--browser-isolation',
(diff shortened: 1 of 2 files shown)


cfme/test_framework/browser_isolation.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+"""Plugin enabling us to isolate browser sessions per test."""
+import pytest
+
+
+def pytest_addoption(parser):
```
jamesooden: I'm not sure what it means to isolate a browser session.   Perhaps you could add a comment explaining what you mean by that.

mfalesni: Sure I can. Basically it is ensuring that each test will start a fresh browser when needed. If we have a browser screwup again, this is one of the things that may help reducing the failure propagation further (like what happened recently).

jamesooden: Sweet, makes sense.

PR #6652, opened 2018-02-14
[1LP][RFR] Check for order view

rlbabyuk:
Purpose or Intent
=================
Added additional check for submiting orver view
{{pytest: cfme/tests/services/test_provision_stack.py --long-running}}
PRT: Failures are not caused by these changes

### DIFF
--- a/cfme/services/service_catalogs/ui.py
+++ b/cfme/services/service_catalogs/ui.py
@@ -176,6 +176,8 @@ def order(self):
     if BZ(1513541, forced_streams=['5.9']).blocks:
         raise NotImplementedError("Service Order is broken - check BZ 1513541")
     view.submit_button.click()
+    # Additional check if submit is done, to cover case when last filled field value is not taken
+    # Can appear every 2nd case when whole module is run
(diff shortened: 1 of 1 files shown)


cfme/services/service_catalogs/ui.py
```diff
@@ -176,6 +176,8 @@ def order(self):
     if BZ(1513541, forced_streams=['5.9']).blocks:
         raise NotImplementedError("Service Order is broken - check BZ 1513541")
     view.submit_button.click()
+    if view.is_displayed and view.submit_button.is_displayed:
+        view.submit_button.click()
```
quarckster: **QUESTION**
Why are these checks needed?

rlbabyuk: While running the whole module, there are some cases in which submit clicked too fast, and we get an error that last field is not filled, but it is. So I tried some check for field value is set and so on, this check will cover if submit is not successful

quarckster: Please add this comment to the code.

rlbabyuk: Done

PR #6654, opened 2018-02-14
[1LP][RFR] fix get element id to support any region

jadhaj:
Purpose or Intent
=================
Removed integer casting from get element id because it supports only region 0
{{ pytest: cfme/tests/openstack/infrastructure/test_hosts.py::test_host_memory }}

### DIFF
--- a/widgetastic_manageiq/__init__.py
+++ b/widgetastic_manageiq/__init__.py
@@ -2934,7 +2934,7 @@ class EntitiesConditionalView(View, ReportDataControllerMixin):
         if self.browser.product_version < '5.9':
             br = self.browser
             for el in br.elements(self.elements):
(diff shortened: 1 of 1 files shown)


widgetastic_manageiq/__init__.py
```diff
@@ -2934,7 +2934,7 @@ def _current_page_elements(self):
         if self.browser.product_version < '5.9':
             br = self.browser
             for el in br.elements(self.elements):
-                el_id = int(br.get_attribute('href', el).split('/')[-1])
+                el_id = (br.get_attribute('href', el).split('/')[-1])
```
mfalesni: **REQUIRED**: Remove the parentheses as well, they are not necessary now.

PR #6663, opened 2018-02-15
[DOC][1LP][RFR] Introducing pudb

quarckster:
Intent
=================

pytest has a cool feature to enable a debugger if an exception occurred. By default, it calls `pdb`, but it isn't so convenient. There is another debugger called `pudb`. It requires only two packages to be installed and then you can use it in such way:

`pytest some_test --pudb`

https://github.com/inducer/pudb
https://github.com/wronglink/pytest-pudb

### DIFF
--- a/docs/guides/debugging.rst
+++ b/docs/guides/debugging.rst
@@ -1,0 +1,24 @@
+Debugging
+=========
+
+Pytest has a cool feature to enable a debugger on failures. Just provide an additional
+command line option::
+
+    pytest --pdb
+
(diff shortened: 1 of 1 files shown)


psav:
I'm not sure we want to add more dependencies seeing as this isn't required to be used. Many people use pycharm instead.

quarckster:
Makes sense, it might be better to mention pudb in our documentation.

lkhomenk:
https://github.com/alexmojaki/birdseye ;) can be used in PyCharm too

PR #6668, opened 2018-02-16
[WIPTEST] Remove unneeded BZ blockers from custom attributes tests

mkoura:
The BZ is mentioned in a comment (few lines above) so the information is not lost.

{{pytest: -v --long-running -k test_bad_section_ cfme/tests/cloud_infra_common/test_custom_attributes_rest.py}}


mshriver:
This is way old, but has no merge conflicts and the BZ is indeed closed.

PR #6672, opened 2018-02-16
[1LP][RFR]Check for smart_management attr

rlbabyuk:
Purpose or Intent
=================
Fix get_tags after get_details changes
{{pytest: cfme/tests/infrastructure/test_provisioning_dialog.py cfme/tests/configure/test_access_control.py -k 'test_tag or test_tagvis_user' --long-running}}

### DIFF
--- a/cfme/common/__init__.py
+++ b/cfme/common/__init__.py
@@ -349,7 +349,11 @@ class WidgetasticTaggable(object):
         """
         view = navigate_to(self, 'Details')
         tags = []
-        tag_table = view.entities.smart_management
+        entities = view.entities
+        if getattr(entities, 'smart_management', False):
+            tag_table = entities.smart_management
+        else:
+            tag_table = entities.summary('Smart Management')
         tags_text = tag_table.get_text_of(tenant)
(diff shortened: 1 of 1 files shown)


cfme/common/__init__.py
```diff
@@ -349,7 +349,11 @@ def get_tags(self, tenant="My Company Tags"):
         """
         view = navigate_to(self, 'Details')
         tags = []
-        tag_table = view.entities.smart_management
+        entities = view.entities
+        if getattr(entities, 'smart_management', False):
+            tag_table = entities.smart_management
+        else:
+            tag_table = entities.summary('Smart Management')
```
quarckster: **REQUIRED**
```python
if hasattr(entities, 'smart_management', False):
    'smart_management', False):
else:
    tag_table = entities.summary('Smart Management')
```

quarckster: It will go, when `get_detail` will be removed from hosts and providers.

mshriver: There's no third option for `hasattr()`, its wrapping `getattr()` and returning false, so just 
```
if hasattr(entities, 'smart_maangement')

PR #6691, opened 2018-02-19
[1LP][RFR] Wrap test_rhev_iso_service_catalog and test_iso_provision_from template

rhrazdil:
Wrapping test_rhev_iso_service_catalog and test_iso_provision_from template tests, because our environment isn't prepared for these tests yet (these tests were previously excluded by provision flag, that flag is now removed).

Removed BZ 1358069 wrapper as it is not needed anymore.

### DIFF
--- a/cfme/tests/infrastructure/test_iso_provisioning.py
+++ b/cfme/tests/infrastructure/test_iso_provisioning.py
@@ -2,12 +2,14 @@
 import fauxfactory
 import pytest
 
-from cfme.utils.conf import cfme_data
 from cfme.common.provider import cleanup_vm
 from cfme.infrastructure.provider import InfraProvider
+from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.infrastructure.pxe import get_template_from_config, ISODatastore
 from cfme.provisioning import do_vm_provisioning
 from cfme.utils import testgen
+from cfme.utils.blockers import GH
+from cfme.utils.conf import cfme_data
 
 pytestmark = [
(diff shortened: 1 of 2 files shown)


cfme/tests/infrastructure/test_iso_provisioning.py
```diff
@@ -2,12 +2,14 @@
...
 
-from cfme.utils.conf import cfme_data
 from cfme.common.provider import cleanup_vm
+from cfme.utils import testgen
+from cfme.utils.blockers import GH
+from cfme.utils.conf import cfme_data
 from cfme.infrastructure.provider import InfraProvider
+from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.infrastructure.pxe import get_template_from_config, ISODatastore
 from cfme.provisioning import do_vm_provisioning
```
quarckster: **REQUIRED**
Please sort imports alphabetically

rhrazdil:
Changed imports order to be alphabetical. Additionally, only our 'rhv_cfme_integration' provider is skipped so other test runs are not affected.

rhrazdil:
Since the tests fail on PRT as well, there is no reason to skip our specific provider, changed back to skip all rhevm.

jkrocil:
It seems that our RHEVs are likely not ready for this test either:
```
Could not find the item All ISO Datastores/rhv41/ISO Images/cfme-provisioning-rhel6.8.iso in Boostrap tree iso_datastores_tree
```

PR #6695, opened 2018-02-19
[1LP][RFR] fix for quadicons in 5.8

izapolsk:
Quadicons have been broken recently by my changes. 
This change should fix the issue. 
To be honest, the best way would be to get rid of ParametrizedLocator in QuadIcon classes since xpath gets more and more sophisticated. However, I believe Quadicons 5.8 will go away soon along with 5.8.
So, I left ParametrizedLocator with that compex xpath

In addition, I bumped into issue with click in TileIcon and fixed it toghether with the rest of stuff.

I tested it locally.


izapolsk:
PRT failures aren't related to my changes.

PR #6698, opened 2018-02-19
[1LP][RFR] Added ansible playbook custom button

quarckster:
Intent
=================

Ansible Playbook button was introduced in CFME 5.9

screenshot_20180219_171452

{{pytest: -v -k "test_ansible_playbook_button_crud or test_custom_button_order_ansible_playbook_service or test_custom_button_ansible_credential_list or test_button_crud or test_button_on_host or test_vm_retire_extend or test_vmware_vimapi_hotadd_disk " --long-running --use-provider vsphere6-nested}}

### DIFF
--- a/cfme/automate/buttons.py
+++ b/cfme/automate/buttons.py
@@ -5,8 +5,10 @@ import re
 from navmazing import NavigateToSibling, NavigateToAttribute
 
 from widgetastic.utils import ParametrizedString, VersionPick
-from widgetastic.widget import Text, Checkbox, ParametrizedView
-from widgetastic_manageiq import SummaryFormItem, FonticonPicker, PotentiallyInvisibleTab
+from widgetastic.xpath import quote
+from widgetastic.widget import Checkbox, ConditionalSwitchableView, ParametrizedView, Text, View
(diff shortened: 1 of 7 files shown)


quarckster:
Failed tests require #6700

cfme/automate/buttons.py
```diff
@@ -29,15 +52,27 @@ def is_displayed(self):
 class ButtonFormCommon(AutomateCustomizationView):
 
     class options(PotentiallyInvisibleTab):  # noqa
-        type = BootstrapSelect('button_type')
-        # TODO: Add the Ansible fields.
+        form = ConditionalSwitchableView(reference="type")
+        type = VersionPick({
+            Version.lowest(): FakeWidget(read_value='Default'),
```
psav: **REQUIRED:** Can you please explain how this Fake Widget works?

quarckster: @psav There is an implementation of it in widgetastic_mangeiq.py. This widget can return any value which provided to `read_value` argument. I need such widget here because `type` is absent in CFME5.8, but `ConditionalSwitchableView` requires that referenced widget provide some value.

PR #6702, opened 2018-02-20
[RFR] fixing sprout for openshift and other providers

izapolsk:
It turned out that I missed some of dependencies in previous PR. I hope this PR will fix the rest of missed stuff


sprout/appliances/models.py
```diff
@@ -906,7 +906,7 @@ def is_visible_only_in_group(self, group):
 
     @property
     def containerized(self):
-        return self.template.container is not None
+        return self.template.container is not None and self.provider.type != 'openshift'
```
mfalesni: Hehe, I think this condition will puzzle people.

sprout/appliances/models.py
```diff
@@ -1171,7 +1171,10 @@ def clone(self, num_appliances=None, time_leased=60, owner=None):
     def create(cls, owner, group, version=None, date=None, provider=None, num_appliances=1,
             time_leased=60, preconfigured=True, yum_update=False, container=False, ram=None,
             cpu=None, provider_type=None):
-        container_q = ~Q(container=None) if container else Q(container=None)
+        if container:
+            container_q = ~Q(container=None) & ~Q(provider_type='openshift')
```
mfalesni: This will not work as expected, because provider_type is not an obligatory parameter.

sprout/appliances/tasks.py
```diff
@@ -305,6 +306,9 @@ def poke_trackerbot(self):
                         name=template_name, preconfigured=False, date=date, custom_data=custom_data,
                         version=template_version, ready=True, exists=True, usable=True)
                     tpl.save()
+                    if provider.type == 'openshift':
```
mfalesni: Is this the Sprout provider model? In that case it would be the `.provider_type`.

sprout/appliances/tasks.py
```diff
@@ -1956,6 +1960,9 @@ def process_docker_images_from_url_group(self, group_id, version, docker_version
         if provider.remaining_configuring_slots < 1:
             # Will do it later ...
             continue
+        if provider.type == 'openshift':
```
mfalesni: Seems like so, so `.provider_type` it is.

sprout/appliances/tasks.py
izapolsk: fixed

sprout/appliances/models.py
izapolsk: provider_type is mandatory now. thanks @mfalesni

dajoRH:
Lint report for commit 9a21a83f7dc0d551f883103e63de6142f7da14fc:

`sprout/appliances/migrations/0043_provider_provider_type.py`:
- :red_circle: Line 11:5: **N806** *variable in function should be lowercase*

**No commit flaws detected**.

*CFME QE Bot*

PR #6707, opened 2018-02-20
[WIP][RFR] Add pre-setup step for openshift provider with alert endpoint

shalomnaim1:
In this PR I added a pre-step for openshift provider setup.

In case an OpenShift provider has Prometheus alerts endpoint pre-setup steps are required.

I added the assignment of alert's profile (both node and provider) to "The Enterprise", so CFME will not ignore alerts that raise after adding the endpoint.

This PR can't run in PRT because it requires yamls change see https://gitlab.cee.redhat.com/cfme-qe/cfme-qe-yamls/merge_requests/466

### DIFF
--- a/cfme/common/provider.py
+++ b/cfme/common/provider.py
@@ -135,7 +135,6 @@ class BaseProvider(WidgetasticTaggable, Updateable, Navigatable):
             created = False
         else:
             created = True
-
             logger.info('Setting up Provider: %s', self.key)
             add_view = navigate_to(self, 'Add')
 
--- a/cfme/containers/provider/__init__.py
+++ b/cfme/containers/provider/__init__.py
@@ -70,7 +70,7 @@ class ContainersProviderEndpointsForm(View):
         api_port = Input('default_api_port')
 
     @View.nested
-    class hawkular(Tab, BeforeFillMixin):  # NOQA
+    class metrics(Tab, BeforeFillMixin):  # NOQA
         TAB_NAME = VersionPick({
             Version.lowest(): 'Hawkular',
             '5.9': 'Metrics'
@@ -91,15 +91,18 @@ class ContainersProviderEndpointsForm(View):
             Version.lowest(): Input('hawkular_api_port'),
             '5.9': Input('metrics_api_port')
         })
+
         validate = Button('Validate')
 
     @View.nested
     class alerts(Tab, BeforeFillMixin):  # NOQA
         TAB_NAME = 'Alerts'
+
         sec_protocol = BootstrapSelect(id='prometheus_alerts_security_protocol')
         trusted_ca_certificates = TextInput('prometheus_alerts_tls_ca_certs')
         hostname = Input('prometheus_alerts_hostname')
         api_port = Input('prometheus_alerts_api_port')
+
         validate = Button('Validate')
 
 
--- a/cfme/containers/provider/openshift.py
+++ b/cfme/containers/provider/openshift.py
@@ -1,20 +1,15 @@
-from cached_property import cached_property
 from os import path
+
+from cached_property import cached_property
 from wrapanapi.containers.providers.rhopenshift import Openshift
-
-from . import ContainersProvider
-
-from cfme.containers.provider import (
-    ContainersProviderDefaultEndpoint, ContainersProviderEndpointsForm
-)
 
 from cfme.common.provider import DefaultEndpoint
 from cfme.control.explorer.alert_profiles import ProviderAlertProfile, NodeAlertProfile
+from cfme.utils import ssh
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.ocp_cli import OcpCli
 from cfme.utils.varmeth import variable
-from cfme.utils import ssh
-from cfme.utils.wait import TimedOutError
+from . import ContainersProvider, ContainersProviderDefaultEndpoint, ContainersProviderEndpointsForm
 
 
 class CustomAttribute(object):
--- a/cfme/containers/provider/openshift.py
+++ b/cfme/containers/provider/openshift.py
@@ -1,6 +1,5 @@
 from cached_property import cached_property
 from os import path
-
 from wrapanapi.containers.providers.rhopenshift import Openshift
 
 from . import ContainersProvider
@@ -10,10 +9,12 @@ from cfme.containers.provider import (
 )
 
 from cfme.common.provider import DefaultEndpoint
+from cfme.control.explorer.alert_profiles import ProviderAlertProfile, NodeAlertProfile
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.ocp_cli import OcpCli
 from cfme.utils.varmeth import variable
-from cfme.utils import version, ssh
+from cfme.utils import ssh
+from cfme.utils.wait import TimedOutError
 
 
 class CustomAttribute(object):
@@ -41,17 +42,13 @@ class OpenshiftDefaultEndpoint(ContainersProviderDefaultEndpoint):
             return str("".join(stdout.readlines()))
 
 
-class HawkularEndpoint(DefaultEndpoint):
-    """Represents Hawkular Endpoint"""
-    name = 'hawkular'
+class ServiceBasedEndpoint(DefaultEndpoint):
 
     @property
     def view_value_mapping(self):
-        out = {
-            'hostname': self.hostname,
-            'api_port': self.api_port,
-            'sec_protocol': self.sec_protocol
-        }
+        out = {'hostname': self.hostname,
+               'api_port': self.api_port,
+               'sec_protocol': self.sec_protocol}
 
         if out['sec_protocol'] and self.sec_protocol.lower() == 'ssl trusting custom ca':
             out['trusted_ca_certificates'] = OpenshiftDefaultEndpoint.get_ca_cert(
@@ -62,32 +59,14 @@ class HawkularEndpoint(DefaultEndpoint):
         return out
 
 
-class AlertsEndpoint(DefaultEndpoint):
+class MetricsEndpoint(ServiceBasedEndpoint):
+    """Represents metrics Endpoint"""
+    name = 'metrics'
+
+
+class AlertsEndpoint(ServiceBasedEndpoint):
     """Represents Alerts Endpoint"""
     name = 'alerts'
-
-    @property
-    def view_value_mapping(self):
-
-        out = {}
-
-        out['hostname'] = version.pick({
-            version.LOWEST: None,
-            '5.9': self.hostname})
-        out['api_port'] = version.pick({
-            version.LOWEST: None,
-            '5.9': self.api_port})
-        out['sec_protocol'] = version.pick({
-            version.LOWEST: None,
-            '5.9': self.sec_protocol})
-
-        if out['sec_protocol'] and self.sec_protocol.lower() == 'ssl trusting custom ca':
-            out['trusted_ca_certificates'] = OpenshiftDefaultEndpoint.get_ca_cert(
-                {"username": self.ssh_creds.principal,
-                 "password": self.ssh_creds.secret,
-                 "hostname": self.master_hostname})
-
-        return out
 
 
 class OpenshiftProvider(ContainersProvider):
@@ -136,6 +115,23 @@ class OpenshiftProvider(ContainersProvider):
             alerts_type=alerts_type,
             endpoints=endpoints,
             appliance=appliance)
+
+    def create(self, **kwargs):
+
+        # Enable alerts collection before adding the provider to avoid missing active
+        # alert after adding the provider
+        # For more info: https://bugzilla.redhat.com/show_bug.cgi?id=1514950
+        if self.appliance.version >= '5.9' and getattr(self, "alerts_type") == "Prometheus":
+            alert_profiles = self.appliance.collections.alert_profiles
+            provider_profile = alert_profiles.instantiate(ProviderAlertProfile,
+                                                          "Prometheus Provider Profile")
+            node_profile = alert_profiles.instantiate(NodeAlertProfile,
+                                                      "Prometheus node Profile")
+
+            for profile in [provider_profile, node_profile]:
+                profile.assign_to("The Enterprise")
+
+        super(OpenshiftProvider, self).create(**kwargs)
 
     @cached_property
     def cli(self):
@@ -217,11 +213,10 @@ class OpenshiftProvider(ContainersProvider):
             if OpenshiftDefaultEndpoint.name == endp:
                 prov_config['endpoints'][endp]['token'] = token_creds.token
                 endpoints[endp] = OpenshiftDefaultEndpoint(**prov_config['endpoints'][endp])
-            elif HawkularEndpoint.name == endp:
-                endpoints[endp] = HawkularEndpoint(**prov_config['endpoints'][endp])
+            elif MetricsEndpoint.name == endp:
+                endpoints[endp] = MetricsEndpoint(**prov_config['endpoints'][endp])
             elif AlertsEndpoint.name == endp:
                 endpoints[endp] = AlertsEndpoint(**prov_config['endpoints'][endp])
-            # TODO Add Prometheus and logic for having to select or the other based on metrcis_type
             else:
                 raise Exception('Unsupported endpoint type "{}".'.format(endp))
 
--- a/cfme/control/explorer/alert_profiles.py
+++ b/cfme/control/explorer/alert_profiles.py
@@ -304,3 +304,8 @@ class ServerAlertProfile(BaseAlertProfile):
 class VMInstanceAlertProfile(BaseAlertProfile):
 
     TYPE = "VM and Instance"
+
+
+class NodeAlertProfile(BaseAlertProfile):
+
+    TYPE = "Node"
--- a/cfme/tests/containers/test_alerts.py
+++ b/cfme/tests/containers/test_alerts.py
@@ -1,0 +1,20 @@
+import pytest
+
+from cfme.containers.provider import ContainersProvider
+from cfme.utils.version import current_version
+
+
+pytestmark = [
+    pytest.mark.uncollectif(lambda: current_version() < "5.9"),
+    pytest.mark.provider([ContainersProvider], scope='module')
+]
+
+
+@pytest.fixture
+def ensure_alerts(provider):
+    if 'alerts' not in provider.endpoints:
+        pytest.skip('No alerts endpoint found for this provider')
+
+
+def test_add_alerts_provider(ensure_alerts, provider):
+    provider.setup()


cfme/common/provider.py
```diff
@@ -136,6 +135,16 @@ def create(self, cancel=False, validate_credentials=True, check_existing=False,
         else:
             created = True
 
+            # Enable alerts collection before adding the provider to avoid missing active
```
jamesooden: You're only enabling alerts if the alert type is Prometheus.   This comment should say something about that.   The question that I feel needs answered by the comment is why do we need to enable alerts collection when it is the alert type Prometheus, and why we don't need to do that when it is not.

dajoRH:
*CFME QE Bot*

cfme/common/provider.py
shalomnaim1: @jamesooden Thanks for your question, 

The only protocol that supports alerts for openshift is Prometheus. When we adding an alerts endpoint on CFME, CFME automatically starts to listen for OpenShift's alerts but if the profiles are not enabled CFME ignore any alerts. I opened a BZ regarding this illogic behavior but for now, this is the implementation (https://bugzilla.redhat.com/show_bug.cgi?id=1514950).

About the "why only for Prometheus":
It just want to make sure that the value of ```alerts_type``` is not ```Disable```

I will add a reference to the CFME docs about this limitation

dajoRH:
Lint report for commit f59216e30aa35b0938bbf0a8d2f1e74625d47e49:

* f59216e30aa35b0938bbf0a8d2f1e74625d47e49

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

dajoRH:
Lint report for commit 1a810b3d20ca911df438c944b63e736af30808ab:

`cfme/containers/provider/__init__.py`:
- :red_circle: Line 97:11: **N801** *class names should use CapWords convention*

**No commit flaws detected**.

*CFME QE Bot*

cfme/common/provider.py
```diff
@@ -8,7 +8,8 @@
 from cfme.base.credential import (
     Credential, EventsCredential, TokenCredential, SSHCredential, CANDUCredential)
 from cfme.common import WidgetasticTaggable
-from cfme.exceptions import ProviderHasNoKey, HostStatsNotContains, ProviderHasNoProperty
+from cfme.exceptions import (
```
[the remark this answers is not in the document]
shalomnaim1: fixed

cfme/common/provider.py
```diff
@@ -134,6 +135,16 @@ def create(self, cancel=False, validate_credentials=True, check_existing=False,
         else:
             created = True
 
+            # Enable alerts collection before adding the provider to avoid missing active
+            # alert after adding the provider
+            # For more info: https://bugzilla.redhat.com/show_bug.cgi?id=1514950
+
+            from cfme.containers.provider.openshift import OpenshiftProvider
+            if isinstance(self, OpenshiftProvider) and \
```
mshriver: **Required**

Use self.one_of(OpenshiftProvider) here, not isinstance()

shalomnaim1: fixed

dajoRH:
Lint report for commit f0b95e2c64ff875f37c71d051abde70357c49bf7:

`cfme/containers/provider/__init__.py`:
- :red_circle: Line 97:11: **N801** *class names should use CapWords convention*

**No commit flaws detected**.

*CFME QE Bot*

cfme/common/provider.py
mshriver: Beyond the use of `provider.one_of`, there is a more fundamental change required here @shalomnaim1.

This is very provider specific, and while below there is use of `self.one_of(VmwareProvider)` that is a bad example that should not be here either.

This should be in a new `OpenshiftProvider.create()` method, which makes this configuration change and then calls `super(OpenshiftProvider, self).create()` in order to call this method from the 'parent' class.  No need to completely override the create method.

Happy to discuss on IRC/BlueJeans if its not clear how this would be done.

cfme/common/provider.py
```diff
@@ -134,6 +134,16 @@ def create(self, cancel=False, validate_credentials=True, check_existing=False,
         else:
             created = True
 
+            # Enable alerts collection before adding the provider to avoid missing active
+            # alert after adding the provider
+            # For more info: https://bugzilla.redhat.com/show_bug.cgi?id=1514950
+
+            from cfme.containers.provider.openshift import OpenshiftProvider
+            if self.one_of(OpenshiftProvider) and \
```
mshriver: **Required**

Re-commenting since first post was hidden on `self.one_of()` update:


Beyond the use of provider.one_of, there is a more fundamental change required here @shalomnaim1.

This is very provider specific, and while below there is use of self.one_of(VmwareProvider) that is a bad example that should not be here either.

This should be in a new OpenshiftProvider.create() method, which makes this configuration change and then calls super(OpenshiftProvider, self).create() in order to call this method from the 'parent' class. No need to completely override the create method.

Happy to discuss on IRC/BlueJeans if its not clear how this would be done.

cfme/containers/provider/__init__.py
```diff
@@ -93,6 +93,16 @@ class hawkular(Tab, BeforeFillMixin):  # NOQA
         })
         validate = Button('Validate')
 
+    @View.nested
+    class prometheus(Tab, BeforeFillMixin): # noqa
```
mshriver: please stay consistent with the rest of the module and use `# NOQA`. In-line comments also require 2 spaces before the `#`, see L73 above.

shalomnaim1: Ok, done

cfme/containers/provider/openshift.py
```diff
@@ -137,6 +158,13 @@ def __init__(
             endpoints=endpoints,
             appliance=appliance)
 
+    def create(self, **kwargs):
+
+        if getattr(self, "alerts_type", None) == "Prometheus":
+            self.enable_all_alerts_collection()
+
+        super(OpenshiftProvider, self).create(**kwargs)
```
mshriver: Yep, this is exactly what I was thinking with moving this out of the common provider class.  Let me know how testing goes, if its working the same as it was when it was in the common class's method (it should!)

shalomnaim1: That's works the same way...

cfme/containers/provider/openshift.py
```diff
@@ -412,3 +441,31 @@ def is_refreshed(self, refresh_timer=None, refresh_delta=600):
...
+
+        provider_profile_view = navigate_to(provider_profile, 'Edit assignments')
+        if provider_profile_view.assign_to.selected_option != "The Enterprise":
+            provider_profile_view.assign_to.select_by_visible_text("The Enterprise")
+            provider_profile_view.save_button.click()
+        else:
+            provider_profile_view.cancel_button.click()
+
+    def enable_node_alerts_collection(self):
+        node_profile = \
```
mshriver: Please don't use `\` line continuation when it can be avoided.  Consider shortening `node_profile` to just `profile` since its the only one in this methods scope.

```python
        node_profile = self.appliance.collections.alert_profiles.instantiate(
            NodeAlertProfile, "Prometheus node Profile")
        node_profile = self.appliance.collections.alert_profiles.instantiate(NodeAlertProfile,
                                                                                                                "Prometheus node Profile")

File: cfme/containers/provider/openshift.py [hunk 8]
```diff
@@ -412,3 +441,31 @@ def is_refreshed(self, refresh_timer=None, refresh_delta=600):
             )
         finally:
             return valid
+
+    def enable_provider_alerts_collection(self):
```
mshriver: These two methods (enable_provider_alerts_collection and enable_node_alerts_collection)are exactly the same, I think they should be a single method that takes an argument, or even better a list of arguments, for what type of alert profiles to assign.

shalomnaim1: @mshriver I added a third function which getting a profile and assigns it to 'The Enterprise'.

This function supports both single and multiple profiles, I didn't use the multiple version because I enable each profile in various function, and I want to keep this structure.

shalomnaim1: https://github.com/shalomnaim1/integration_tests/blob/a933c585d2530f31569b7f14b44f7d8040bbc6a2/cfme/containers/provider/openshift.py#L448-L458

mshriver:
PRT is a false positive, no tests are running.

shalomnaim1:
@mshriver I walked about this yesterday, this PR can't run on the PRT since it requires the YAML update.

cfme/containers/provider/openshift.py
```diff
@@ -137,6 +158,16 @@ def __init__(
             endpoints=endpoints,
             appliance=appliance)
 
+    def create(self, **kwargs):
+
+        # Enable alerts collection before adding the provider to avoid missing active
+        # alert after adding the provider
+        # For more info: https://bugzilla.redhat.com/show_bug.cgi?id=1514950
+        if getattr(self, "alerts_type", None) == "Prometheus":
+            self.enable_all_alerts_collection()
```
mshriver: Discussed this offline, wanted to record comment here.

I think this is ultimately driving a bit of an anti-pattern, in an effort to work around a bug, and is resulting in a bunch of layered convenience methods that make a lot of assumptions, and repeat the functionality provided by `BaseAlertProfile.assign_to()`

I think the new methods should be collapsed into a few lines here instantiating the alert profiles for node and provider, and then using their `assign_to()` methods, passing 'The Enterprise' during provider configuration.

Consider wrapping this behavior in an `if BZ(1514950).blocks()` block, so that we stop doing this workaround as soon as the bug is fixed.

mshriver:
@shalomnaim1 Although this can't run against a prometheus enabled provider, I think you should still collect and run the container provider crud tests, to show that this doesn't break any existing tests on non-alert providers.

cfme/containers/provider/openshift.py
```diff
@@ -137,6 +158,23 @@ def __init__(
             endpoints=endpoints,
             appliance=appliance)
 
+    def create(self, **kwargs):
+
+        # Enable alerts collection before adding the provider to avoid missing active
+        # alert after adding the provider
+        # For more info: https://bugzilla.redhat.com/show_bug.cgi?id=1514950
+        if getattr(self, "alerts_type", None) == "Prometheus":
+            alert_profiles = self.appliance.collections.alert_profiles
```
mshriver: Thanks for being open and flexible about this, I prefer this much more to the layered methods that were in use before.  This makes it very clear what is being configured, and why.

shalomnaim1: sure ;)

shalomnaim1:
@mshriver Do you see any other changes in this PR?

cfme/containers/provider/openshift.py
```diff
@@ -137,6 +158,23 @@ def __init__(
             endpoints=endpoints,
             appliance=appliance)
 
+    def create(self, **kwargs):
+
+        # Enable alerts collection before adding the provider to avoid missing active
+        # alert after adding the provider
+        # For more info: https://bugzilla.redhat.com/show_bug.cgi?id=1514950
+        if getattr(self, "alerts_type", None) == "Prometheus":
```
mshriver: **Required**

I believe this _has_ to be version picked, and even if it will work without failure in assigning the profiles, prometheus is only available in MIQ Gaprindashvili / CFME 5.9+.

The way that OpenshiftProvider.from_config() is implemented, the provider object will have an alerts type set if its in the yaml, regardless of version.

shalomnaim1: Is this really necessary? 
you want me to change it to something like that?

``` python
getattr(self, "alerts_type") == "Prometheus" 
```

mshriver: I believe so, yes. At a minimum it keeps configuration consistent, where you would only assign enterprise profiles for a 5.9/Gaprindashvili appliance when configuring Prometheus.  Discussed with @jawatts a bit too.

The getattr() with a default use was correct, as that prevents an AttributeException - but I think since that attribute is included in the constructor with a default of None, you don't have to wrap that.

```python
if self.alerts_type.lower() == PrometheusEndpoint.name and self.appliance.version >= '5.9':
```

shalomnaim1:
@mshriver I will be happy to get all the comments at once, so the review will be more efficient and the code will be merged asap.

Thanks

cfme/containers/provider/openshift.py
shalomnaim1: I don't think that ignoring yaml corruption is right, think that if someone forgot to add this field to the YAML the entire test suite will fail, and will not run for a period of time (until the yaml be fixed and merged to the yamls repo)

I do agree that MIQ version check has to be added to this if statement

mshriver: I'm not concerned with the yaml being incorrect here - Its that a provider in the yaml with this prometheus endpoint would have these alert profiles configured, regardless of what version of the appliance is being tested.  The two OCP providers without prometheus don't have that endpoint configured, and their alerts_type will be None.

shalomnaim1: @mshriver I thought about the 5,9 version validation and I think it needless because Prometheus is available only in version 5.9.

According to the direction of thought presented above, I have to trust the user

mshriver: @shalomnaim1 I don't think you're quite following. This isn't about 'trusting the user', and this isn't optional.  This will _not work_ with CFME/MIQ versions that don't support Prometheus, if the provider is fetched from a yaml configuration.

http://pastebin.test.redhat.com/560050

cfme/containers/provider/__init__.py
```diff
@@ -93,6 +93,16 @@ class hawkular(Tab, BeforeFillMixin):  # NOQA
         })
         validate = Button('Validate')
 
+    @View.nested
+    class prometheus(Tab, BeforeFillMixin):  # NOQA
+        TAB_NAME = 'Metrics'
```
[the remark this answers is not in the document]
mshriver: So there's a bit of confusing interplay between the yaml and these tabs, and it stems from them being called hawkular instead of generically 'metrics'. I think a good update to these modules would be to update the endpoints and these tabs to use metrics, instead of hawkular, so that in the yaml the endpoint can be 'metrics' generically.

If this is removed, the label for the endpoint in the yaml has to be 'hawkular', matching the others.  This would be fixed with updating the nested 'hawkular' view and calling all of the endpoints in the yaml 'metrics'.

Ultimately this new class is being forced in, just to call it 'prometheus', when in reality there is no difference in the CFME UI between the types, only what endpoint is pointed to and how the endpoint is configured/validation.

Unless you would like to take on modifying the other providers, and refactoring this structure a bit to use 'metrics' more generically for the tabs, I think this should be removed. Its not needed to actually support the new endpoint.

cfme/containers/provider/openshift.py
```diff
@@ -137,6 +158,23 @@ def __init__(
             endpoints=endpoints,
             appliance=appliance)
 
+    def create(self, **kwargs):
+
+        # Enable alerts collection before adding the provider to avoid missing active
+        # alert after adding the provider
+        # For more info: https://bugzilla.redhat.com/show_bug.cgi?id=1514950
+        if self.appliance.version >= '5.9' and getattr(self, "alerts_type") == "Prometheus":
```
mshriver: **Question**

Is this required when alerts_type OR metrics_type are prometheus? Or do these alert profiles only need to be set with Prometheus alerts, but not with Prometheus metrics?

cfme/containers/provider/openshift.py
```diff
@@ -63,6 +63,27 @@ def view_value_mapping(self):
         return out
 
 
+class PrometheusEndpoint(DefaultEndpoint):
+    """Represents prometheus Endpoint"""
+    name = 'prometheus'
+
+    @property
+    def view_value_mapping(self):
+        out = {
```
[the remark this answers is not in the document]
shalomnaim1: I created a new class for Prometheus based endpoints.
Both Prometheus Alerts and Prometheus matrics classes to inherit for that base class, the base class contains a view_value_mapping function.

since both endpoints exist only from 5.9, I used version pick in the base class.

@mshriver What do you think? Is this solution is good enough?

cfme/containers/provider/__init__.py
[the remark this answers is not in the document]
shalomnaim1: Well, basicly you right. But I think that fix is out of the scope of this PR.

The main focus of this PR is to allow the FW work with Prometheus metrics as a metrics provider and supply correct configuration before adding an alert endpoint.

I think the solution for the issue you describe here should be in a different PR.

cfme/containers/provider/openshift.py
```diff
@@ -63,6 +63,27 @@ def view_value_mapping(self):
         return out
 
 
+class PrometheusEndpoint(DefaultEndpoint):
```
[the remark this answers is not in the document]
shalomnaim1: So I change Prometheus Metrics endpoint to work with version pick as requested

``` Python 
    @property
    def view_value_mapping(self):

        out = {}

        out['hostname'] = version.pick({
            version.LOWEST: None,
            '5.9': self.hostname})
        out['api_port'] = version.pick({
            version.LOWEST: None,
            '5.9': self.api_port})
        out['sec_protocol'] = version.pick({
            version.LOWEST: None,
            '5.9': self.sec_protocol})

        if out['sec_protocol'] and self.sec_protocol.lower() == 'ssl trusting custom ca':
            out['trusted_ca_certificates'] = OpenshiftDefaultEndpoint.get_ca_cert(
                {"username": self.ssh_creds.principal,
                 "password": self.ssh_creds.secret,
                 "hostname": self.master_hostname})

        return out
```

I don't think Prometheus Metrics and Hawkular are the same, Prometheus not supported in >5.9, therefore it returns None for each field in the view in a version older than 5.9.

Both Prometheus alerts and Prometheus metrics have the same structure, so I created a ```PrometheusBasedEndpoint``` class so ```PrometheusMetricsEndpoint``` and ```PrometheusAlertsEndpoint``` inheriting from.

cfme/containers/provider/openshift.py
shalomnaim1: Assigning alert profiles is required only for alerts, 

By assigning the profiles you allow MIQ to pay attention to Prometheus alerts, If an alert endpoint is added without assigning both profiles, MIQ will collect the alerts from Prometheus and ignore them.

cfme/containers/provider/openshift.py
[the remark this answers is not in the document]
shalomnaim1: See the comment above[1]

[1] https://github.com/ManageIQ/integration_tests/pull/6707#discussion_r172028869

cfme/containers/provider/__init__.py
```diff
@@ -91,15 +91,41 @@ class hawkular(Tab, BeforeFillMixin):  # NOQA
...
+        TAB_NAME = VersionPick({
+            '5.9': 'Metrics'})
+        sec_protocol = VersionPick({
+            '5.9': BootstrapSelect(id='metrics_security_protocol')})
+        trusted_ca_certificates = VersionPick({
+            '5.9': TextInput('metrics_tls_ca_certs')})
+        hostname = VersionPick({
+            '5.9': Input('metrics_hostname')})
+        api_port = VersionPick({
+            '5.9': Input('metrics_api_port')})
```
mshriver: @shalomnaim1 I still do not believe this is the correct approach to modeling these endpoint tabs.  I pulled your ocp-v3 provider yaml, and this branch, in order to demonstrate what this creates in the views.

This is a completely duplicate view of the hawkular endpoint, which should just be converted to a metrics tab, where the name is version picked for 'Hawkular' in 5.8 and 'Metrics' in 5.9.

```python
In []: ocp = providers.get_crud('ocp-v3')

In []: app = store.current_appliance

In []: ocp.endpoints.keys()
Out[]: ['default', 'prometheus', 'alerts']
```

Above isn't correct in my view, the openshift endpoints should be `default`, `metrics`, and `alerts` for providers in 5.9, and `default` and `metrics` in 5.8. Prometheus is simply a type for the metrics endpoint. This is ultimately driven by the yaml, and it matching the endpoint class name.

The changing tab name needing to match the endpoint name in the yaml can be solved with a version picked name for the common metrics endpoint class that would replace the hawkular endpoints class 

```python
view = navigate_to(ocp, 'Add')
prom_view = getattr(ocp.endpoints_form(parent=view), 'prometheus')

prom_view.__class__()
Out[]: prometheus()

prom_view.cls_widget_names()
Out[]: ('sec_protocol', 'trusted_ca_certificates', 'hostname', 'api_port', 'validate')
```
Above is how endpoints are filled in BaseProvider.create(), through the endpoint_form attribute and the endpoint key from the yaml.

Note the widget names for the prometheus view tab, for ocp-v3 with 'prometheus' in yaml

Look below at how ocp-v2, with 'hawkular' in yaml has exactly the same widgets, but is a different class

```
ocp2 = providers.get_crud('ocp-v2')
ocp2.endpoints.keys()
Out[]: ['default', 'hawkular']

hawk_view.__class__()
Out[]: hawkular()

hawk_view = getattr(ocp.endpoints_form(parent=view), 'hawkular')
hawk_view.cls_widget_names()
Out[]: ('sec_protocol', 'trusted_ca_certificates', 'hostname', 'api_port', 'validate')
```

This demonstrates the duplicate classes that you've created here in this PR.

There may be some small complications from the class simplification when it comes to having to version pick some things, but I don't believe that a simplified class structure for a common 'metrics' endpoint is complex or unreasonable to require at such a core level of the framework.

shalomnaim1: Take a look at this now please, I updated the yaml as well so all OCP providers have metrics endpoint instead of Hawkular/Prometheus

cfme/containers/provider/openshift.py
```diff
@@ -91,6 +68,16 @@ def view_value_mapping(self):
         return out
 
 
+class MetricsEndpoint(ServiceBasedEndpoint):
+    """Represents Hawkular Endpoint"""
```
jawatts: This doc string will need to be updated

cfme/containers/provider/openshift.py
```diff
@@ -138,6 +125,23 @@ def __init__(
             endpoints=endpoints,
             appliance=appliance)
 
+    def create(self, **kwargs):
+
+        # Enable alerts collection before adding the provider to avoid missing active
+        # alert after adding the provider
+        # For more info: https://bugzilla.redhat.com/show_bug.cgi?id=1514950
+        if self.appliance.version >= '5.9' and getattr(self, "alerts_type") == "Prometheus":
```
If you don't specify a default in the ```getattr``` and `alert_type` is missing it'll raise an `AttributeError`. please add a default, i.e.
```python
if self.appliance.version >= '5.9' and getattr(self, "alerts_type", None) == "Prometheus":
```

cfme/containers/provider/__init__.py
```diff
@@ -91,15 +91,26 @@ class hawkular(Tab, BeforeFillMixin):  # NOQA
...
         validate = Button('Validate')
 
     @View.nested
     class alerts(Tab, BeforeFillMixin):  # NOQA
         TAB_NAME = 'Alerts'
-        sec_protocol = BootstrapSelect(id='prometheus_alerts_security_protocol')
-        trusted_ca_certificates = TextInput('prometheus_alerts_tls_ca_certs')
-        hostname = Input('prometheus_alerts_hostname')
-        api_port = Input('prometheus_alerts_api_port')
+
```
mshriver: I don't think these version picks are necessary, as there is only one value anyway.

If we wanted to model that this tab isn't in the view on 59z, the entire view would have to be extended and version picked, which I do not think is necessary in this case.

I think you can leave this class as-is.

cfme/containers/provider/openshift.py
```diff
@@ -42,30 +42,7 @@ def get_ca_cert(connection_info):
...
-                 "password": self.ssh_creds.secret,
-                 "hostname": self.master_hostname})
-
-        return out
-
-
-class AlertsEndpoint(DefaultEndpoint):
-    """Represents Alerts Endpoint"""
-    name = 'alerts'
+class ServiceBasedEndpoint(DefaultEndpoint):
```
mshriver: I need to do some local testing to confirm, but this does not look like it will support configuration of the metrics endpoints (hawkular tab) in 58z, since the view_value_mapping will return nothing.

How has your local testing of 58z and ocp-v1/v2 worked with their hawkular endpoints?  Too bad PRT is in a deadlock and we can't merge the yamls PR first because it will break collection :/

cfme/containers/provider/openshift.py
shalomnaim1: If "alert_type"  missing then the statement return False, I don't see why an exception is required here.

cfme/containers/provider/openshift.py
shalomnaim1: ```view_value_mapping``` returns a ```dict``` object with all the endpoint info.

you can see this at https://github.com/shalomnaim1/integration_tests/blob/59a1ae7fbd83251214e2c54bb4e5add9043766ac/cfme/containers/provider/openshift.py#L68

jawatts: ocp-v2 is not supported in 5.8 - I think (documentation for 4.5 kinda stinks). But ocp-v1 should be checked.

cfme/containers/provider/__init__.py
shalomnaim1: I will revert this code.

cfme/containers/provider/openshift.py
shalomnaim1: And default is ```None``` in this case

shalomnaim1: I added a default before but @mshriver asked me to delete it for some reason.

cfme/containers/provider/openshift.py
shalomnaim1: @jawatts I was able to add ```ocp-v1``` to 5.8 appliances. about ```ocp-v2``` I didn't check that.

cfme/containers/provider/openshift.py
shalomnaim1: @mshriver Do you want me to add back the default value?

shalomnaim1: As far as I remember the reason for default removal is to get this exception, so we will fix the yamls accordingly

dajoRH:
*CFME QE Bot*

dajoRH:
Lint report for commit 65c913b6358c7defb0761aab3c23a36dd4f2ac81:

`cfme/containers/provider/openshift.py`:
- :red_circle: Line 17:1: **F401** *'cfme.utils.wait.TimedOutError' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

PR #6711, opened 2018-02-21
[1LP][RFR] Fixing table of tasks in Tasks view

jan-zmeskal:
_Fixing_ test_host_drift_analysis.py. I recently re-activated this module in RHV-CFME integration suite and found out it failes seemingly randomly. After some digging I discovered that the reason for failure was this:
-  FW runs SSA on host
- Then goes to Administration -> Tasks view
- Then it tries to check the task of SSA was completed successfully

However, at this point it sometimes happen that the task is not even displayed yet in the table of tasks. Therefore any attempt to operate on it resulted `NoSuchElementException`. The only way AFAIK to get task's updated status is to refresh the whole view. I achieved that by overloading `wait_displayed` method of Table, but I don't know if this is the right way. Please share your ideas with me. 

Addendum:
I also discovered that test collection for test_host_analysis was not working if a provider had more than one host, so I fixed that. I think it should not break anything (I tried collection with more providers), but honestly pytest/FW test parametrization is black magic to me.

{{pytest: cfme/tests/infrastructure/test_host_drift_analysis.py cfme/tests/infrastructure/test_host_analysis.py -vv}}

### DIFF
--- a/cfme/configure/tasks.py
+++ b/cfme/configure/tasks.py
@@ -47,6 +47,12 @@ TABS_DATA_PER_PROVIDER = {
 }
 
 
+class TasksTable(Table):
+
+    def wait_displayed(self, timeout=10, delay=2, fail_func=None):
+        wait_for(lambda: self.is_displayed, timeout=timeout, delay=delay, fail_func=fail_func)
+
+
 def is_vm_analysis_finished(name, **kwargs):
     return is_analysis_finished(name=name, task_type='vm', **kwargs)
 
@@ -125,6 +131,7 @@ def check_tasks_have_no_errors(task_name, task_type, expected_num_of_tasks, sile
     # expected_status change from str to support also regular expression pattern
     expected_status = re.compile(expected_status, re.IGNORECASE)
 
+    tab_view.table.wait_displayed(fail_func=view.reload.click)
     try:
         rows = list(tab_view.table.rows(task_name=task_name, state=expected_status))
     except IndexError:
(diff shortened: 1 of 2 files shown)


cfme/configure/tasks.py
```diff
@@ -47,6 +47,12 @@
 }
 
 
+class TasksTable(Table):
+
+    def wait_displayed(self, timeout=10, delay=2, fail_func=None):
+        wait_for(lambda: self.is_displayed, timeout=timeout, delay=delay, fail_func=fail_func)
```
quarckster: **REQUIRED**
I don't like the idea to bring new things without necessity.

cfme/configure/tasks.py
```diff
@@ -125,6 +131,7 @@ def check_tasks_have_no_errors(task_name, task_type, expected_num_of_tasks, sile
     # expected_status change from str to support also regular expression pattern
     expected_status = re.compile(expected_status, re.IGNORECASE)
 
+    tab_view.table.wait_displayed(fail_func=view.reload.click)
```
quarckster: **REQUIRED**
```python
wait_for(
    lambda: tab_view.table.is_displayed,
    timeout=10,
    delay=2,
    fail_func=view.reload.click
)
```

jan-zmeskal: @quarckster Yeah, that sound like much better idea. :+1:

jan-zmeskal: @quarckster Fixed

PR #6715, opened 2018-02-21
[1LP][RFR] Metering Reports: Add support for storage metrics

nachandr:
{{pytest: cfme/tests/intelligence/reports/test_metering_report.py --use-provider complete}}

PRT results:
59 : Passed on RHV, Provider skips - v6, v65
58 : Tests uncollected like expected since Metering reports have been introduced in 59

### DIFF
--- a/cfme/tests/intelligence/reports/test_metering_report.py
+++ b/cfme/tests/intelligence/reports/test_metering_report.py
@@ -269,7 +269,7 @@ def test_validate_cpu_usage(resource_usage, metering_report):
             usage_from_report = groups["CPU Used"]
             if 'GHz' in usage_from_report:
                 estimated_cpu_usage = estimated_cpu_usage * math.pow(2, -10)
-            usage = re.sub(r'[MHz, GHz,]', r'', usage_from_report)
+            usage = re.sub(r'(MHz|GHz|,)', r'', usage_from_report)
             assert estimated_cpu_usage - DEVIATION <= float(usage) \
                 <= estimated_cpu_usage + DEVIATION, 'Estimated cost and report cost do not match'
             break
@@ -283,7 +283,7 @@ def test_validate_memory_usage(resource_usage, metering_report):
             usage_from_report = groups["Memory Used"]
             if 'GB' in usage_from_report:
                 estimated_memory_usage = estimated_memory_usage * math.pow(2, -10)
-            usage = re.sub(r'[MB, GB,]', r'', usage_from_report)
+            usage = re.sub(r'(MB|GB|,)', r'', usage_from_report)
             assert estimated_memory_usage - DEVIATION <= float(usage) \
                 <= estimated_memory_usage + DEVIATION, 'Estimated cost and report cost do not match'
             break
@@ -295,7 +295,7 @@ def test_validate_network_usage(resource_usage, metering_report):
         if groups["Network I/O Used"]:
             estimated_network_usage = resource_usage['network_io']
             usage_from_report = groups["Network I/O Used"]
-            usage = re.sub(r'[KBps,]', r'', usage_from_report)
+            usage = re.sub(r'(KBps|,)', r'', usage_from_report)
             assert estimated_network_usage - DEVIATION <= float(usage) \
                 <= estimated_network_usage + DEVIATION,\
                 'Estimated cost and report cost do not match'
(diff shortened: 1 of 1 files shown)


cfme/tests/intelligence/reports/test_metering_report.py
```diff
@@ -304,3 +312,16 @@ def test_validate_disk_usage(resource_usage, metering_report):
...
             break
+
+
+def test_validate_storage_usage(resource_usage, metering_report):
+    """Test to validate storage usage."""
+    for groups in metering_report:
+        if groups["Storage Used"]:
+            estimated_storage_usage = resource_usage['storage_used']
+            usage_from_report = groups["Storage Used"]
+            usage = re.sub(r'[MB, GB,]', r'', usage_from_report)
```
mkoura: REQUIRED: What is the purpose of this regex? To strip 'MB' or 'GB' from the string? If that's the case, it should be ``r'(MB|GB)'`` (or ``r' ?(MB|GB)'`` if you there can be a space that you want to strip as well).
What you wrote is equivalent to ``r'[GMB, ]'`` and will strip any G, M, B, comma or space character. If it works in this case it's just coincidental.
It should be fixed in other places as well.

cfme/tests/intelligence/reports/test_metering_report.py
```diff
@@ -275,32 +283,45 @@ def test_validate_memory_usage(resource_usage, metering_report):
...
 
 def test_validate_network_usage(resource_usage, metering_report):
-    """Test to validate network usage cost."""
+    """Test to validate network usage."""
     for groups in metering_report:
         if groups["Network I/O Used"]:
             estimated_network_usage = resource_usage['network_io']
             usage_from_report = groups["Network I/O Used"]
-            usage = re.sub(r'[KBps,]', r'', usage_from_report)
+            usage = re.sub(r'(KBps)', r'', usage_from_report)
```
mkoura: OPTIONAL: ``usage_from_report.replace('KBps', '')`` is enough here

cfme/tests/intelligence/reports/test_metering_report.py
```diff
@@ -275,32 +283,45 @@ def test_validate_memory_usage(resource_usage, metering_report):
...
 
 def test_validate_disk_usage(resource_usage, metering_report):
-    """Test to validate disk usage cost."""
+    """Test to validate disk usage."""
     for groups in metering_report:
         if groups["Disk I/O Used"]:
             estimated_disk_usage = resource_usage['disk_io_used']
             usage_from_report = groups["Disk I/O Used"]
-            usage = re.sub(r'[KBps,]', r'', usage_from_report)
+            usage = re.sub(r'(KBps)', r'', usage_from_report)
```
mkoura: OPTIONAL: ``usage_from_report.replace('KBps', '')`` is enough here

cfme/tests/intelligence/reports/test_metering_report.py
nachandr: @mkoura , Fixed in all the places

cfme/tests/intelligence/reports/test_metering_report.py
nachandr: I have changed it to this , since the comma also needs to be stripped.
usage = re.sub(r'(KBps|,)', r'', usage_from_report)

cfme/tests/intelligence/reports/test_metering_report.py
```diff
@@ -137,9 +138,16 @@ def resource_usage(vm_ownership, appliance, provider):
     appliance.db.client.session.query(metrics).delete()
     appliance.db.client.session.query(rollups).delete()
 
-    # Chargeback reporting is done on hourly and daily rollup values and not real-time values.So, we
+    # Metering reports are done on hourly and daily rollup values and not real-time values.So, we
     # are capturing C&U data and forcing hourly rollups by running these commands through
     # the Rails console.
+    #
+    # Metering reports differ from Chargeback reports in that Metering reports 1)report only
```
mshriver: :+1:

mshriver:
Nothing jumping out at me, but I want to wait for PRT results (it's being worked on) due to the changes in the `re.sub` calls, and the exception that occurred in PRT run 21704.

```
    def test_validate_network_usage(resource_usage, metering_report):
        """Test to validate network usage."""
        for groups in metering_report:
            if groups["Network I/O Used"]:
                estimated_network_usage = resource_usage['network_io']
                usage_from_report = groups["Network I/O Used"]
                usage = re.sub(r'(KBps)', r'', usage_from_report)
>               assert estimated_network_usage - DEVIATION <= float(usage) \
                    <= estimated_network_usage + DEVIATION,\
                    'Estimated cost and report cost do not match'
E               ValueError: invalid literal for float(): 1,915.36

cfme/tests/intelligence/reports/test_metering_report.py:299: ValueError

nachandr:
@mshriver, That's correct. The 043792b commit added the ',' to the regex to fix the error from your comment.

Thanks for reviewing and merging the PR.

PR #6728, opened 2018-02-22
[1LP][RFR] New tests and enhancements to service dialogs API testing

mkoura:
* removed unused ``dialog_ui`` function
* better checking of data returned by create and edit actions
* new test ``test_edit_service_dialogs``

{{pytest: -v --long-running -k TestServiceDialogsRESTAPI cfme/tests/services/test_rest_services.py}}

### DIFF
--- a/cfme/tests/services/test_rest_services.py
+++ b/cfme/tests/services/test_rest_services.py
@@ -627,6 +627,9 @@ class TestServiceRESTAPI(object):
 
 class TestServiceDialogsRESTAPI(object):
     def check_returned_dialog(self, appliance):
+        # full dialog is returned only in >= 5.9
+        if appliance.version < '5.9':
+            return
         returned = appliance.rest_api.response.json()
         if 'results' in returned:
             results = returned['results']
--- a/cfme/tests/services/test_rest_services.py
+++ b/cfme/tests/services/test_rest_services.py
@@ -13,6 +13,7 @@ from cfme.rest.gen_data import (
     blueprints as _blueprints,
     copy_role,
     dialog as _dialog,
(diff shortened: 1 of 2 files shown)


cfme/tests/services/test_rest_services.py
```diff
@@ -97,6 +98,12 @@ def dialog(request, appliance):
     return _dialog(request, appliance)
 
 
+@pytest.fixture(scope="function")
+def service_dialogs(request, appliance, num=3):
```
quarckster: **QUESTION**
Why did you add `num` argument here?

mkoura: I could either add it inside the function as a variable or as a function argument. I chose argument because I use the same approach for other fixtures in this test file. On some places I call fixtures as normal functions, esp. when I need just single resource (``somefixture(..., num=1)``). This one is not used this way (yet), but since both possibilities to declare the variable are more-or-less equivalent, I haven't considered it as needlesly future-proofing :)

cfme/tests/services/test_rest_services.py
```diff
@@ -619,44 +626,106 @@ def test_retire_parent_service_now(self, request, appliance):
...
+        # full dialog is returned only in >= 5.9
+        if appliance.version < '5.9':
+            return
+        returned = appliance.rest_api.response.json()
+        if 'results' in returned:
+            results = returned['results']
+        else:
+            results = [returned]
+        for result in results:
+            dialog_tabs, = result['dialog_tabs']
```
izapolsk: Q: please explain me what's going on here. I'm a bit confused about this syntax

PR #6729, opened 2018-02-22
[1LP][RFR] Refactored test_provision_stack.py

quarckster:
Fixed incorrect teardown in test_provision_stack.py. We should remove orchestration templates via UI. Besides some refactoring has been made. All prerequisite stuff moved to fixtures.

{{pytest: -v cfme/tests/services/test_provision_stack.py --use-provider ec2west}}

### DIFF
--- a/cfme/tests/services/test_provision_stack.py
+++ b/cfme/tests/services/test_provision_stack.py
@@ -4,8 +4,7 @@ import pytest
 from cfme import test_requirements
 from cfme.cloud.provider import CloudProvider
 from cfme.cloud.provider.azure import AzureProvider
-from cfme.cloud.stack import StackCollection
-from cfme.configure.settings import DefaultView
+from cfme.cloud.provider.openstack import OpenStackProvider
 from cfme.services.catalogs.catalog import Catalog
 from cfme.services.catalogs.catalog_item import CatalogItem
 from cfme.services.catalogs.orchestration_template import OrchestrationTemplate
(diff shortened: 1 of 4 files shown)


dajoRH:
Lint report for commit f108968b3cfd47ee97fed015a04957631db1c689:

- :red_circle: Line 11:1: **F401** *'cfme.utils.wait.wait_for' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
Lint report for commit aed9aaf91653cc934444b7f1efb2c6e500b9a2aa:

`cfme/tests/services/test_provision_stack.py`:
- :heavy_exclamation_mark: Line 102:76: **W291** *trailing whitespace*

**No commit flaws detected**.

*CFME QE Bot*

quarckster:
`test_remove_template_provisioning` failed because an orchestration template cannot be deleted until it's in use. @sshveta, please clarify how is this test supposed to work?

sshveta:
When running locally , this test passes
cfme/tests/services/test_provision_stack.py::test_remove_template_provisioning[ec2west] PASSED

cfme/tests/services/test_provision_stack.py
```diff
@@ -121,129 +72,148 @@ def stack_data(appliance, provider, provisioning):
...
+    template.create_service_dialog_from_template(dialog_name, template.template_name)
+    yield template
+    if stack.exists:
+        stack.retire_stack()
+    if template.exists:
+        template.delete()
+
+
+@pytest.yield_fixture
+def catalog():
```
sshveta: Fixture catalog can be used here 
https://github.com/ManageIQ/integration_tests/blob/master/cfme/fixtures/service_fixtures.py#L21

cfme/tests/services/test_provision_stack.py
```diff
@@ -121,129 +72,148 @@ def stack_data(appliance, provider, provisioning):
...
-                             stack_data):
+
+def test_provision_stack(provision_request, stack):
     """Tests stack provisioning
 
     Metadata:
         test_flag: provision
     """
-    catalog_item, template = catalog_item
+    assert provision_request.is_succeeded()
```
psav: **REQUIRED:** I'm not 100% sure about this. It seems that the majority of what happens here happens in fixtures. The test body should contain what is being tested. In this case the test body is only testing that the response was successful.

psav: Some of the tests below use the fixtures correctly but I believe this one is an oversimplification, probbaly so that the fixtures can be reused.

quarckster: @psav I don't think it's an oversimplification. The test does exactly what it describes. It checks if the stack was successfully provisioned and if the stack exists. Other things are not related to the test itself. We don't test service ordering or catalog item creation or template creation.

psav: I think we need someone else to come in and resolve this one @jkrocil perhaps
To me, the test is called ```test_provision_stack```. Whilst I agree that the catalog item creation and the template creation shouldn't be part of the test. I **do** think that the act of ordering the stack **is** something that should form part of the test. In reality it could be that this is the **only** place we should ever see this inside a test. I would be perfectly happy for anything that use a stack in the test to provision it as part of a fixture. However I do firmly believe that this should be part of the test in this case. I am of course sometimes wrong, may-hap often. So I will let @jkrocil give the final decision on it.

PR #6730, opened 2018-02-22
Update appliance configuration on RHOS

mshriver:
Set hostname from reverse-lookup of IP for 'on_openstack' appliances
within Appliance._custom_configure()


cfme/utils/appliance/__init__.py
```diff
@@ -2346,6 +2346,18 @@ def _custom_configure(self, **kwargs):
             self.restart_evm_service(log_callback=log_callback)
             self.wait_for_web_ui(log_callback=log_callback)
 
+        # Set fqdn for openstack appliance
+        if on_openstack:
+            fqdn = None
+            host = self.ssh_client.run_command(
+                'host {}'.format(self.hostname))  # hostname is IP
```
mshriver: I need to assert this RC, to make sure it actually resolved.

PR #6731, opened 2018-02-22
[1LP][RFR] Fixed SSH buffering

psav:
Fixed SSH buffering

### DIFF
--- a/cfme/utils/ssh.py
+++ b/cfme/utils/ssh.py
@@ -1,29 +1,26 @@
 # -*- coding: utf-8 -*-
-import fauxfactory
-import iso8601
-import re
 import socket
 import sys
 from collections import namedtuple
-from os import path as os_path
 from subprocess import check_call
-from urlparse import urlparse
 
-
-from cached_property import cached_property
-import paramiko
-from scp import SCPClient
 import diaper
+import fauxfactory
+import iso8601
+import paramiko
+import re
+from cached_property import cached_property
+from os import path as os_path
+from scp import SCPClient
 
 from cfme.utils import conf, ports, version
 from cfme.utils.log import logger
 from cfme.utils.net import net_check
-from cfme.utils.version import Version
-from fixtures.pytest_store import store
 from cfme.utils.path import project_path
 from cfme.utils.quote import quote
 from cfme.utils.timeutil import parsetime
-
+from cfme.utils.version import Version
+from fixtures.pytest_store import store
 
 # Default blocking time before giving up on an ssh command execution,
 # in seconds (float)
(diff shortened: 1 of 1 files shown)


cfme/utils/ssh.py
```diff
@@ -309,18 +306,34 @@ def run_command(
...
-                        output.append(line)
-                        if self._streaming:
-                            self.f_stderr.write(line)
+                    line = stderr.next()
+                    output.append(line)
+                    if self._streaming:
+                        self.f_stderr.write(line)
 
                 if session.exit_status_ready():
+                    while session.recv_ready:
```
jamesooden: A comment on why this is needed will help future folks who look at this code (could even be future me and you).

cfme/utils/ssh.py
```diff
@@ -309,18 +306,34 @@ def run_command(
             stderr = session.makefile_stderr()
             while True:
```
jamesooden: Might be good to have a comment on why we are doing it this way, so no one unwittingly simplifies it later.

jamesooden:
Could we add a description to the commit log that explains what we were seeing before that this fixes.

jamesooden:
And thank you for this (-:

cfme/utils/ssh.py
```diff
@@ -307,20 +304,39 @@ def run_command(
...
+                        write_output(line, self.f_stderr)
+                    except StopIteration:
+                        pass
 
                 if session.exit_status_ready():
+                    # When the program finishes, we need to grab the rest of the output that is left
+                    # Though it's possible, we should have read enough of the buffers so that we can
+                    # just dump the rest.
+                    if session.recv_ready:
+                        for line in stdout:
```
jamesooden: What happens if the last line does not send an EOL character?

jamesooden: I mean if the last bit of output from the remote program does not have an EOL ending it.

cfme/utils/ssh.py
```diff
@@ -307,20 +304,39 @@ def run_command(
...
+                            write_output(line, self.f_stdout)
+                    if session.recv_stderr_ready:
+                        for line in stderr:
```
jamesooden: Same as line 335.

cfme/utils/ssh.py
mshriver: @jamesooden So stdout is a paramiko ChannelFile, which is a paramiko BufferedFile that behaves like a python file.

https://github.com/paramiko/paramiko/blob/master/paramiko/channel.py#L669

Here, in the recv method, which is used for ChannelFile.read(), the docblock includes the description of the behavior when zero length string is received.

It looks to me like if there is no EOL, it won't necessarily matter as it will just handle it as the file has been closed.

Some interesting points in paramiko: 
https://github.com/paramiko/paramiko/blob/master/paramiko/file.py#L291
https://github.com/paramiko/paramiko/blob/master/paramiko/file.py#L119

mshriver:
Talked with @jamesooden offline, and he ran some additional testing.  PRT failures are unrelated, and I believe this addresses the original issues and should handle missing EOL.

PR #6735, opened 2018-02-23
[WIPTEST] Standard template naming

mshriver:
Re-opening this since @ohiliazo is going to be taking a look at reworking the template_upload scripts to be object oriented provider actions.

There is additional impact to sprout regex, and to existing trackerbot records using the current naming convention. This may require re-creation of existing MIQ/CFME templates on RH trackerbot servers.

Can be merged into template_upload_all, but I think template naming could/should be pulled into its own utility module.

### DIFF
--- a/scripts/template_upload_all.py
+++ b/scripts/template_upload_all.py
@@ -30,8 +30,8 @@ from cfme.utils import path, trackerbot
 from cfme.utils.conf import cfme_data
 from cfme.utils.log import logger, add_stdout_handler
 
-CFME_BREW_ID = "cfme"
-NIGHTLY_MIQ_ID = "manageiq"
+CFME_ID = "cfme"
+MIQ_ID = "manageiq"
 
 add_stdout_handler(logger)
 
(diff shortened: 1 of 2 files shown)


scripts/template_upload_all.py
```diff
@@ -62,106 +62,124 @@ def parse_cmd_line():
...
+    """
+
     pattern = re.compile(r'.*/(.*)')
-    image_name = pattern.findall(image_link)[0]
-    image_name = image_name.lower()
-    image_dt = get_last_modified(checksum_link)
-    # CFME brew builds
-    if CFME_BREW_ID in image_name:
+    image_name = pattern.findall(image_link)[0].lower()
+    formatted_timestamp = '{year}{month}{day}'.format(year=str(timestamp.year),
```
jamesooden: Consider using time.strftime() to do your formatting of time:

https://docs.python.org/2/library/time.html#time.strftime

jamesooden: This baby comes from C and ends up in almost all other languages.   Your needs are real simple but it does some really really cools stuff, and makes time formatting for humans really easy.

mshriver:
Broke TemplateName class into a separate branch, closing

PR #6736, opened 2018-02-23
[1LP][RFR] Add TemplateName class to cfme/utils/trackerbot

mshriver:
Class for generating trackerbot/sprout friendly template names from the CFME build server URLs.

This is intended as a replacement for what is currently in template_upload_all generating template names for us.  The current naming scheme is inconsistent, and uses different date timestamps sometimes for the same build of MIQ/CFME.  Timestamps also were not consistent across streams.

To be used in new template_upload flow that @ohiliazo is working on.

Testing against several CFME/MIQ build URLs, including the URLs that until now have not been able to run through template_upload_all:

http://pastebin.test.redhat.com/558699

### DIFF
--- a/cfme/utils/trackerbot.py
+++ b/cfme/utils/trackerbot.py
@@ -28,20 +28,32 @@ stream_matchers = (
     (get_stream('5.4'), r'^cfme-54.*-(?P<month>\d{2})(?P<day>\d{2})'),
     (get_stream('5.5'), r'^cfme-55.*-(?P<month>\d{2})(?P<day>\d{2})'),
     (get_stream('5.6'), r'^cfme-56.*-(?P<month>\d{2})(?P<day>\d{2})'),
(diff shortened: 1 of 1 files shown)


cfme/utils/trackerbot.py
```diff
@@ -1,16 +1,19 @@
...
 from collections import defaultdict, namedtuple
-from datetime import date
-import urllib
+from datetime import date, datetime
 
+import attr
 import slumber
 import requests
-import time
+from bs4 import BeautifulSoup
```
quarckster: **QUESTION**
I don't see `bs4` package in `frozen.txt`. Should it be added? If yes, please consider to use `lxml` instead.

mshriver: Odd, I don't recall actually installing it in my venv, but I have it? Its actually in `pip freeze` as `beautifulsoup4==4.6.0`, but is imported through bs4.  Will evaluate lxml and update either frozen.txt or my implementation.

PR #6737, opened 2018-02-25
[1LP][RFR] Override is_refreshed method in openshift provider

brahmanim:
Add is_refreshed method in openshift provider,
Overrides cfme.common.provider.BaseProvider#is_refreshed.
Add validate_status check after provider creation,
In order to make sure that openShift provider data refresh completed.

{{pytest: cfme/tests/containers/test_provider_openscap_policy_attached.py -v --use-provider ocp-v1}}

### DIFF
--- a/cfme/containers/provider/openshift.py
+++ b/cfme/containers/provider/openshift.py
@@ -14,6 +14,7 @@ from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.ocp_cli import OcpCli
 from cfme.utils.varmeth import variable
 from cfme.utils import version, ssh
+from cfme.utils.wait import TimedOutError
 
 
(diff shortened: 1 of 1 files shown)


cfme/containers/provider/openshift.py
```diff
@@ -384,3 +385,30 @@ def sync_ssl_certificate(self):
...
+        valid = False
+        # check refresh time
+        if not super(OpenshiftProvider, self).is_refreshed():
+            return valid
+
+        # check the detail page matches the Providers information
+        try:
+            self.validate_stats(ui=False)
+            valid = True
+        except TimedOutError:
```
gshefer: __QUESTION__
I think you can remove the `except` because it's already raise this TimeOutError if it fails. Or you just wanted more informative exception?

PR #6749, opened 2018-02-26
[1LP][RFR] Add support for adding a user to multiple RBAC groups

LandoCalrizzian:
Add support for adding a user to multiple RBAC groups. Supported in version >= 5.9

access_control.UserCollection.create will allow you to create a user that is a member of multiple groups.
You can now check the currently assigned group and all assigned groups properly in 5.9

{{pytest: -v -k 'test_user_group_test_user_crud or test_user_assign_multiple_groups or test_user_group_error_validation or test_group_ownership_on_user_or_group_role or test_group_ownership_on_user_only_role'}}

### DIFF
--- a/cfme/base/login.py
+++ b/cfme/base/login.py
@@ -15,6 +15,9 @@ class BaseLoggedInPage(View):
     )
     help = NavDropdown('.//li[./a[@id="dropdownMenu1"]]|.//li[./a[@id="help-menu"]]')
     settings = NavDropdown('.//li[./a[@id="dropdownMenu2"]]')
+    # 5.9 Locator for Settings item that replaces current group name when user has multiple groups
+    group_list_locator = (
+        './/ul/li[contains(@class, "dropdown-submenu") and contains(., "Change Group")]')
     navigation = VerticalNavigation('#maintab')
 
     @property
@@ -44,7 +47,33 @@ class BaseLoggedInPage(View):
 
     @property
     def current_groupname(self):
-        return self.settings.items[1].strip()
+        # 5.9 Locators for finding current group when user has multiple groups
+        current_group_locator = '{}/ul/li/a[@title="Currently Selected Group"]'.format(
+            self.group_list_locator)
+        current_group_marker = ' (Current Group)'
+        try:
+            current_group = self.browser.element(current_group_locator)
+            return self.browser.text(current_group).replace(current_group_marker, '')
+        except NoSuchElementException:
+            return self.settings.items[1].strip()
+
+    @property
+    def group_names(self):
+        """ Return a list of the logged in user's assigned groups.
+
+        Returns:
+            Version >= 5.9 - list off all groups the logged in user is assigned to
+            Version < 5.9 - single item list containing the user's current group
+        """
+        group_list_locator = '{}/ul/li'.format(self.group_list_locator)
+        current_group_marker = ' (Current Group)'
+
+        group_list = self.browser.elements(group_list_locator)
+        if group_list:
+            return [
+                self.browser.text(group).replace(current_group_marker, '') for group in group_list]
+        else:
+            return [self.current_groupname]
 
     @property
     def logged_in(self):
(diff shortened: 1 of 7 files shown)


dajoRH:
Lint report for commit 8090e45bbdfeceff399633248d370d005060b0c0:

`cfme/base/login.py`:
- :heavy_exclamation_mark: Line 57:39: **W291** *trailing whitespace*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/configure/test_access_control.py
```diff
@@ -103,6 +104,32 @@ def test_user_crud(appliance, group_collection):
     user.delete()
 
 
+@pytest.mark.tier(2)
+@pytest.mark.uncollectif(lambda: version.current_version() < '5.9')
```
quarckster: **REQUIRED**
```python
@pytest.mark.uncollectif(lambda appliance: appliance.version < '5.9')
```

cfme/configure/access_control/__init__.py
```diff
@@ -362,17 +366,27 @@ def create(self, name=None, credential=None, email=None, group=None, cost_center
 
         user = self.instantiate(
             name=name, credential=credential, email=email, group=group, cost_center=cost_center,
-            value_assign=value_assign
+            value_assign=value_assign, groups=groups
         )
 
+        if user.groups and self.appliance.version < "5.9":
+            raise Exception(
```
quarckster: **REQUIRED**
Please raise `CFMEException`

cfme/tests/configure/test_access_control.py
```diff
@@ -13,7 +13,7 @@
 from cfme.infrastructure import virtual_machines as vms
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.services.myservice import MyService
-from cfme.utils import error
+from cfme.utils import error, version
```
quarckster: **REQUIRED**
`version` is not needed.

cfme/configure/access_control/__init__.py
```diff
@@ -133,6 +135,7 @@ class User(Updateable, Pretty, BaseEntity):
     credential = attr.ib(default=None)
     email = attr.ib(default=None)
     group = attr.ib(default=None)
+    groups = attr.ib(default=None)
```
psav: **REQUIRED:** I really think we should change ```group``` to be ```groups```. Yes we will need to change it in more places, but to me it makes much more sense to pass in either a single item and for it to be transformed into a 1 len list, rather than have two params, that's just confusing.

LandoCalrizzian: Updated so that UserCollection.create accepts a single 'groups' args that will accept a list of groups --OR-- a single group obj for backwards compatibility

cfme/base/login.py
```diff
@@ -15,6 +15,9 @@ class BaseLoggedInPage(View):
     )
     help = NavDropdown('.//li[./a[@id="dropdownMenu1"]]|.//li[./a[@id="help-menu"]]')
     settings = NavDropdown('.//li[./a[@id="dropdownMenu2"]]')
+    # 5.9 Locator for Settings item that replaces current group name when user has multiple groups
+    group_list_locator = (
```
mshriver: I really think this, and the code in the methods below using it, should be turning into a manageiq_widgetastic widget.  That would move the browser.element use out of this view model, and into the widget model where I believe it should be.

That would allow for defining properties or methods for the actions taken below, like:
* get current group when there are multiple
* determine if there are multiple via is_displayed instead of a NoSuchElementException on a browser.element call
* a widget.read() method that does the current_group_marker replacement
* definition of current_group_marker as a widget attribute that's usable outside of this View class property
* returning list of available groups in the widget

mshriver: This widget would implement `fill()`, and could be used to switch groups as well as reading.

mshriver: @LandoCalrizzian I see TODO added, acceptable to me!

PR #6751, opened 2018-02-27
[WIP] Refactor template upload

ohiliazo:
Purpose
=================
Refactoring template upload to be more object-oriented, readable and extendable.


cfme/utils/template/base.py
```diff
@@ -0,0 +1,161 @@
...
+
+DEFAULT_MGMT_SYS = cfme_data.management_systems
+DEFAULT_HOST_USERNAME = credentials['host_default']['username']
+DEFAULT_HOST_PASSWORD = credentials['host_default']['password']
+
+TEMPLATE_UPLOAD_DATA = cfme_data['template_upload']
+lock = Lock()
+
+
+class BaseProviderClient(object):
```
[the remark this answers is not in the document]
ohiliazo: @izapolsk, yeah, I'm already removing this

cfme/utils/template/base.py
```diff
@@ -0,0 +1,245 @@
...
+        :var image_patters: regex to be matched when stream URL is used
+    """
+    provider_type = None
+    log_name = None
+    image_pattern = None
+
+    def __init__(self, stream=None, provider=None, template_name=None,
+                 cmd_line_args=None, **kwargs):
+        """
+        Required parameters:
```
mshriver: Conflicting docblock here, where stream, provider, and template_name have `None` defaults and thus aren't required.

Keep to google style docblocks as well, with 'Args' header

cfme/utils/template/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+    --stream-url CUSTOM_URL_TO_STREAM_DIRECTORY
+        REQUIRES --stream
+
+    --image-url URL_PATH_TO_IMAGE_FILE
+
+    --template-name CUSTOM_TEMPLATE_NAME
+        if template name is used -> template will be formatted as {template}-{stream}
+    
+    --provider-data  # PATH TO LOCAL YAML FILE WITH LOCAL PROVIDER DATA
```
mshriver: I think we can greatly simplify a  bunch of the logic/methods throughout if we drop support for provider-data.

With support of `--provider`, its trivial to have your data defined in cfme_data.local.yaml or cfme_data.yaml.

Was there any driver that you found to keep support for `--provider-data`?

cfme/utils/template/base.py
```diff
@@ -0,0 +1,245 @@
...
+
+    @property
+    def provider_data(self):
+        """ Returns AttrDict from cfme_data[management_systems][provider]."""
+        if not self._provider_data:
+            return cfme_data.management_systems[self.provider]
+        return self._provider_data
+
+    @property
+    def template_upload_data(self):
```
mshriver: Not a fan of naming here, since there is a yaml configuration block for template_upload, and this is a BaseTemplateUpload class.

I think my confusion ultimately stems from the class name though, given the class's modeling.

What do you think about calling this class `ProviderTemplateUpload` instead of 'Base', since its in effect modeling a trackerbot provider-template record.

cfme/utils/template/ec2.py
```diff
@@ -0,0 +1,138 @@
...
+from cfme.utils.template.base import BaseTemplateUpload, log_wrap
+from cfme.utils.wait import wait_for
+
+
+class EC2TemplateUpload(BaseTemplateUpload):
+    log_name = 'EC2'
+    provider_type = 'ec2'
+    image_pattern = re.compile(r'<a href="?\'?([^"\']*ec2[^"\'>]*)')
+
+    def get_creds(self, creds_type=None, **kwargs):
```
mshriver: Looks like this is unused here, and there's no more need for the aws_cli_tool_client query because the file is downloaded locally and uploaded through the provider's wrapanapi class.

cfme/utils/template/README.md
ohiliazo: No, there was no specific reason for keeping this support. So if we can just through it out, I'll do it.

cfme/utils/template/ec2.py
```diff
@@ -0,0 +1,138 @@
...
+import urllib2
+
+from cfme.utils.log import logger
+from cfme.utils.template.base import BaseTemplateUpload, log_wrap
+from cfme.utils.wait import wait_for
+
+
+class EC2TemplateUpload(BaseTemplateUpload):
+    log_name = 'EC2'
+    provider_type = 'ec2'
```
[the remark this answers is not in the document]
ohiliazo: The reason is only for logging purpose, as well as log_name.

cfme/utils/template/ec2.py
ohiliazo: So I can change script to download file locally instead of aws_client_tool?

cfme/utils/template/README.md
mshriver: go for it

cfme/utils/template/ec2.py
mshriver: `download_image` method is already pulling the file to a local directory :)

But yep, no more need for aws_client_tool since this is strictly using the wrapanapi mgmt class to interact with the API.

PR #6755, opened 2018-02-27
[RFR] ContainersProvider update num_container to support changes on CFME 5.9 DB

brahmanim:
Fixing ContainersProvider num_container, adding a query that supports changes on CFME 5.9 DB.
The container_definitions table doesn't exist anymore on CFME 5.9,
The data on this table merged into containers table.

{{pytest: cfme/tests/containers/test_provider_openscap_policy_attached.py -v --use-provider ocp-v1}}


mshriver:
PRT for 21649 is good, shows provider completing configuration, which was being blocked by the failed DB query.

PR #6756, opened 2018-02-27
[WIPTEST] Move alert profile assignment flash message into test

mshriver:
There was a change to the flash message between CFME 5.8/5.9, @shalomnaim1 found in PR #6707 

Move the flash message assertion into a test instead of in the assign_to
method


cfme/tests/control/test_basic.py
```diff
@@ -543,14 +544,23 @@ def test_alert_profile_crud(request, alert_profile_class, alert_collection,
 
 
 @pytest.mark.tier(2)
-@pytest.mark.meta(blockers=[BZ(1416311, forced_streams=["5.7"])])
```
mshriver: BZ closed/verified in 5.8

cfme/tests/control/test_basic.py
```diff
@@ -543,14 +544,23 @@ def test_alert_profile_crud(request, alert_profile_class, alert_collection,
...
-def test_alert_profile_assigning(alert_profile):
+def test_alert_profile_assigning(alert_profile, appliance):
+    view = appliance.browser.create_view(AlertProfileDetailsView)
     if isinstance(alert_profile, alert_profiles.ServerAlertProfile):
-        if BZ(1489697, forced_streams=["5.8"]).blocks:
```
mshriver: BZ closed/verified in 5.8

mshriver:
PRT failures are due to vsphere65-nested outage during the run. 95% passing (95/101) on both versions with test_basic.py passing 100%

cfme/tests/control/test_alerts.py
```diff
@@ -128,7 +129,7 @@ def configure_fleecing(appliance, provider, vm, vddk_url):
 
 @pytest.fixture
 def setup_for_alerts(alert_profile_collection, action_collection, policy_collection,
```
jamesooden: Could we have some sort of a function header here.   I know its not your code, but somethings that explains what this things is supposed to accomplish with a little more detail than what is in the name.

mshriver: I'll add something in

mshriver:
Merging with two first-level reviews because this is a blocker to container provider configuration.

PR #6763, opened 2018-02-28
[1LP][RFR] fixed non 0 region issue for list views in 5.8

izapolsk:
{{pytest: --long-running cfme/tests/openstack/cloud/test_networks.py }}


psav:
Are failures related?

quarckster:
@psav no, our infra is down

izapolsk:
@psav, I tested my change locally. it is ok. Tests fail because of different env and framework issues not related to my change.

PR #6767, opened 2018-02-28
[RFR] Dev trial

psav:
This branch enables the running of the suite against a rails server

### DIFF
--- a/fixtures/templateloader.py
+++ b/fixtures/templateloader.py
@@ -2,9 +2,11 @@
 """Preloads all templates on all providers that were selected for testing. Useful for test collect.
 """
 import pytest
-from fixtures.pytest_store import store
+
 from cfme.utils import trackerbot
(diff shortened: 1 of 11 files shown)


fixtures/templateloader.py
```diff
@@ -2,22 +2,32 @@
...
+        default=False, help="Use a cached version of the templates and not redownload them"
+    )
 
 
 def pytest_configure(config):
-    if store.parallelizer_role == 'master' or trackerbot.conf.get('url') is None:
+    is_dev = False
+    if 'appliances' in env:
+        for appliance in env.appliances:
+            if appliance.get('is_dev', False):
```
jamesooden: **QUESTION:**  It looks like you are seitting the appliance to be a dev appliance if it wasn't set that way.   Why do that?   Probably the answer is worthy of being codified in a comment.

PR #6773, opened 2018-02-28
[1LP][RFR] Added cloud support for service catalogs and some refactoring

sshveta:
Refactored some code and added cloud support for service catalogs in SSUI

### DIFF
--- a/cfme/fixtures/service_fixtures.py
+++ b/cfme/fixtures/service_fixtures.py
@@ -59,30 +59,31 @@ def create_catalog_item(provider, provisioning, vm_name, dialog, catalog, consol
         description="my catalog", display_in=True, catalog=catalog,
         dialog=dialog, catalog_name=catalog_name,
         provider=provider, prov_data=provisioning_data)
+    catalog_item.create()
     return catalog_item
 
 
-@pytest.fixture(scope="function")
-def order_catalog_item_in_ops_ui(appliance, provider, provisioning, vm_name, dialog, catalog,
-        console_template, request):
-    """
-        Fixture for SSUI tests.
-        Orders catalog item in OPS UI.
-    """
+@pytest.fixture
+def order_service(appliance, provider, provisioning, vm_name,
+                  dialog, catalog, console_template, request):
+    """ Orders service once the catalog item is created"""
     if hasattr(request, 'param'):
         catalog_item = create_catalog_item(provider, provisioning, vm_name, dialog, catalog,
-            console_template if 'console_test' in request.param else None)
+                                           console_template if 'console_test' in request.param else None)
     else:
         catalog_item = create_catalog_item(provider, provisioning, vm_name, dialog, catalog)
-    vm_name = catalog_item.provisioning_data['catalog']["vm_name"]
-    request.addfinalizer(lambda: cleanup_vm("{}0001".format(vm_name), provider))
-    catalog_item.create()
-    service_catalogs = ServiceCatalogs(appliance, catalog_item.catalog, catalog_item.name)
+    service_catalogs = ServiceCatalogs(appliance, catalog_item.catalog,
+                                       catalog_item.name)
     service_catalogs.order()
-    logger.info("Waiting for cfme provision request for service {}".format(catalog_item.name))
-    request_description = catalog_item.name
-    provision_request = appliance.collections.requests.instantiate(request_description,
-                                                                   partial_check=True)
-    provision_request.wait_for_request()
-    assert provision_request.is_finished()
     return catalog_item
+
+
+@pytest.fixture
+def provision_request(appliance, order_service):
+    """Checks if the service request is completed and provisioned"""
+    catalog_item = order_service
+    provision_request = appliance.collections.requests.instantiate(catalog_item.name,
+                                                                   partial_check=True)
+    provision_request.wait_for_request(method='ui')
+    return catalog_item, provision_request
+# TODO - remove request finally
--- a/cfme/fixtures/vm_name.py
+++ b/cfme/fixtures/vm_name.py
@@ -13,7 +13,11 @@ def vm_name(provider):
     yield vm_name
     try:
         logger.info('Cleaning up VM %s on provider %s', vm_name, provider.key)
-        provider.mgmt.delete_vm(vm_name)
+        scat_vm = "{}0001".format(vm_name)
+        if scat_vm in provider.mgmt.list_vm():
+            provider.mgmt.delete_vm("{}0001".format(vm_name))
+        else:
+            provider.mgmt.delete_vm(vm_name)
     except:
         # The mgmt_sys classes raise Exception :\
         logger.warning('Failed to clean up VM %s on provider %s', vm_name, provider.key)
--- a/cfme/services/service_catalogs/ssui.py
+++ b/cfme/services/service_catalogs/ssui.py
@@ -5,7 +5,7 @@ from widgetastic.widget import Text
 from widgetastic_patternfly import Input, Button, BootstrapSelect
 
 from cfme.base.ssui import SSUIBaseLoggedInPage
-from cfme.services.service_catalogs import ServiceCatalogs
+from cfme.services.service_catalogs import ServiceCatalogs, BaseOrderForm
 from cfme.utils.appliance import MiqImplementationContext
 from cfme.utils.appliance.implementations.ssui import (
     navigator,
@@ -14,7 +14,7 @@ from cfme.utils.appliance.implementations.ssui import (
     ViaSSUI
 )
 from cfme.utils.wait import wait_for
-from widgetastic_manageiq import SSUIServiceCatalogcard, SSUIInput, Notification
+from widgetastic_manageiq import SSUIServiceCatalogcard, Notification
 
 
 class ServiceCatalogsView(SSUIBaseLoggedInPage):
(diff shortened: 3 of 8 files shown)


cfme/fixtures/service_fixtures.py
```diff
@@ -59,30 +58,30 @@ def create_catalog_item(provider, provisioning, vm_name, dialog, catalog, consol
...
 
 
-@pytest.fixture(scope="function")
-def order_catalog_item_in_ops_ui(appliance, provider, provisioning, vm_name, dialog, catalog,
-        console_template, request):
-    """
-        Fixture for SSUI tests.
-        Orders catalog item in OPS UI.
-    """
+@pytest.yield_fixture
```
quarckster: **REQUIRED**
What's the point to use yield fixture here?

cfme/fixtures/service_fixtures.py
```diff
@@ -59,30 +58,30 @@ def create_catalog_item(provider, provisioning, vm_name, dialog, catalog, consol
...
+def provision_request(appliance, order_service):
+    catalog_item = order_service
+    provision_request = appliance.collections.requests.instantiate(catalog_item.name,
                                                                    partial_check=True)
-    provision_request.wait_for_request()
-    assert provision_request.is_finished()
-    return catalog_item
+    provision_request.wait_for_request(method='ui')
+    yield catalog_item, provision_request
+    if provision_request.exists:
```
quarckster: **REQUIRED**
```python
if provision_request.exists():
```

cfme/fixtures/service_fixtures.py
```diff
@@ -59,30 +58,29 @@ def create_catalog_item(provider, provisioning, vm_name, dialog, catalog, consol
...
     return catalog_item
+
+
+@pytest.fixture
+def provision_request(appliance, order_service):
+    catalog_item = order_service
+    provision_request = appliance.collections.requests.instantiate(catalog_item.name,
+                                                                   partial_check=True)
+    provision_request.wait_for_request(method='ui')
+    return catalog_item, provision_request
```
sshveta: Cannot remove request once it is provisioned as ssui orders one request with same name and hence the test fails with error " Multiple requests with matching "desc" found - be more specific! .

Removed it for now , until I come up with a workaround .

cfme/services/service_catalogs/ssui.py
```diff
@@ -35,32 +38,63 @@ def is_displayed(self):
 class OrderForm(ServiceCatalogsView):
```
quarckster: **REQUIRED**
What's the point to double the code? You can just import this view from cfme/services/service_catalogs/ui.py

cfme/tests/ssui/test_ssui_service_catalogs.py
```diff
@@ -1,31 +1,36 @@
...
+def test_service_catalog_crud(appliance, setup_provider,
+                              context, provision_request):
     """Tests Service Catalog in SSUI."""
 
-    service_name = order_catalog_item_in_ops_ui.name
+    catalog_item, provision_request = provision_request
     with appliance.context.use(context):
-        appliance.server.login()
-        service = ServiceCatalogs(appliance, name=service_name)
+        dialog_values = {'service_name': "ssui_" + fauxfactory.gen_alphanumeric()}
```
quarckster: **REQUIRED**
Please use `format`.

cfme/services/service_catalogs/ssui.py
sshveta: Cannot use same code as both have different views.

cfme/fixtures/vm_name.py
```diff
@@ -13,7 +13,7 @@ def vm_name(provider):
     yield vm_name
     try:
         logger.info('Cleaning up VM %s on provider %s', vm_name, provider.key)
-        provider.mgmt.delete_vm(vm_name)
+        provider.mgmt.delete_vm("{}0001".format(vm_name))
```
mshriver: I think this is too specific to service catalog provisioning requests for such a generic fixture.  Should check for existence of the '0001' vm

TODO above the fixture states cleanup shouldn't even happen here, but I don't quite agree with that.

cfme/services/service_catalogs/ssui.py
```diff
@@ -35,32 +38,63 @@ def is_displayed(self):
...
+        def read(self):
+            return self.visible_widget.read()
+
+        def fill(self, value):
+            return self.visible_widget.fill(value)
+
+    def fill(self, fill_data):
+        values = deflatten_dict(fill_data)
+        was_change = False
+        self.before_fill(values)
```
mshriver: this should modify was_change, there's an open issue against wt.core for this behavior as well.

cfme/services/service_catalogs/ssui.py
quarckster: @sshveta I don't see the difference. Please point me. `OrderForm` in ssui.py is the same as `OrderForm` in ui.py.

cfme/services/service_catalogs/ssui.py
```diff
@@ -35,32 +38,57 @@ def is_displayed(self):
...
+        values = deflatten_dict(fill_data)
+        for key, value in values.items():
+            widget = self.fields(key)
+            if value is None:
+                self.logger.debug('Skipping fill of %r because value was None', key)
+                continue
+            try:
+                widget.fill(value)
+            except NotImplementedError:
+                continue
```
quarckster: **REQUIRED**
`fill()` must return boolean.

cfme/services/service_catalogs/ssui.py
sshveta: @quarckster : form is same but the servicecatalog view is different for both of them 
OrderForm(ServiceCatalogsView):

quarckster: **REQUIRED**
In order to avoid code duplication, you may create a base `OrderForm` view and place it in cfme/services/service_catalogs/__init__.py. Then you can inherit it.

cfme/fixtures/vm_name.py
```diff
@@ -13,7 +13,11 @@ def vm_name(provider):
     yield vm_name
     try:
         logger.info('Cleaning up VM %s on provider %s', vm_name, provider.key)
-        provider.mgmt.delete_vm(vm_name)
+        scat_vm = "{}0001".format(vm_name)
```
mshriver: :+1: that var name hints at why the '001' is being checked too, nice.

cfme/services/service_catalogs/ssui.py
sshveta: Tried to add a common view for both UI's.
Using one orderForm now.

sshveta:
Monthly charge failure is not related to this PR and 
test_service_catalog_crud fails for one provider and passes for others .
Also , test passes locally .

cfme/services/service_catalogs/ui.py
```diff
@@ -39,7 +39,7 @@ class service_catalogs(Accordion):  # noqa
         tree = ManageIQTree()
 
 
-class OrderForm(ServicesCatalogView):
+class OrderForm(OrderFormView):
```
quarckster: **REQUIRED**
This view should be in `cfme/services/service_catalogs/__init__.py` and should be shared between ui.py and ssui.py. Each of these files have their own `ServiceCatalogsView`. Resulted OrderFrom view should inherit base `OrderForm` from `__init__.py` and `ServiceCatalogsView`
`__init__.py`
```python
class BaseOrderForm(View):
    """Represents the order form of a service.
    This form doesn't have a static set of elements apart from titles and buttons. In the most cases
    the fields can be either regular inputs or dropdowns. Their locators depend on field names. In
    order to find and fill required fields a parametrized view is used here. The keys of a fill
    dictionary should match ids of the fields. For instance there is a field with such html
    <input id="some_key"></input>, so a fill dictionary should look like that:
    {"some_key": "some_value"}
    """
    dialog_title = Text(
        VersionPick({
            Version.lowest(): ".//div[@id='main_div']//h3",
            "5.9": ".//div[@id='main_div']//h2"
        })
    )

    @ParametrizedView.nested
    class fields(ParametrizedView):  # noqa
        PARAMETERS = ("key",)
        input = Input(id=Parameter("key"))
        select = Select(id=Parameter("key"))
        param_input = Input(id=ParametrizedString("param_{key}"))
        dropdown = VersionPick({
            Version.lowest(): BootstrapSelect(Parameter("key")),
            "5.9": BootstrapSelect(locator=ParametrizedString(
                ".//div[contains(@class, 'bootstrap-select') and "
                "select[@id={key|quote}]]"))
        })
        param_dropdown = VersionPick({
            Version.lowest(): BootstrapSelect(ParametrizedString("param_{key}")),
            "5.9": BootstrapSelect(locator=ParametrizedString(
                ".//div[contains(@class, 'bootstrap-select') and "
                "select[@id='param_{key}']]"))
        })

        @property
        def visible_widget(self):
            if self.input.is_displayed:
                return self.input
            elif self.dropdown.is_displayed:
                return self.dropdown
            elif self.param_input.is_displayed:
                return self.param_input
            elif self.param_dropdown.is_displayed:
                return self.param_dropdown
            elif self.select.is_displayed:
                return self.select
            else:
                raise ItemNotFound("Visible widget is not found")

        def read(self):
            return self.visible_widget.read()

        def fill(self, value):
            return self.visible_widget.fill(value)

    def fill(self, fill_data):
        values = deflatten_dict(fill_data)
        was_change = False
        self.before_fill(values)
        for key, value in values.items():
            widget = self.fields(key)
            if value is None:
                self.logger.debug('Skipping fill of %r because value was None', key)
                continue
            try:
                if widget.fill(value):
                    was_change = True
            except NotImplementedError:
                continue

        self.after_fill(was_change)
        return was_change
```
`ui.py`
```python
class OrderForm(ServicesCatalogView, BaseOrderFrom):
     pass
```
`ssui.py`
```python
class OrderForm(ServicesCatalogView, BaseOrderFrom):
     pass
```

mshriver:
@sshveta Please comment on PRT results frp, 21970, there are failures in visible_widget where none of the checked widgets are visible on a few tests.

sshveta:
1 failure in 5.8 about the console test , not related to the changes.
(KK owns it )

PR #6778, opened 2018-03-01
[1LP][RFR] Adding hostname information in cfme_data in lenovo section

douglasgabriel:
Purpose or Intent
=================

__Updating `cfme_data.yaml.template`__ adding hostname information in `lenovo` section.

__Updating__ the way that Physical Providers access the `hostname` and `ip_address` information to follow the standard on project

### DIFF
--- a/cfme/physical/provider/__init__.py
+++ b/cfme/physical/provider/__init__.py
@@ -16,6 +16,7 @@ from cfme.utils.pretty import Pretty
 from cfme.utils.varmeth import variable
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.log import logger
+from cfme.utils.net import resolve_hostname
 
 
 class PhysicalProvider(Pretty, BaseProvider, Fillable):
@@ -36,6 +37,14 @@ class PhysicalProvider(Pretty, BaseProvider, Fillable):
         self.endpoints = self._prepare_endpoints(endpoints)
         self.name = name
         self.key = key
+
+    @property
+    def hostname(self):
+        return getattr(self.default_endpoint, "hostname", None)
+
+    @property
+    def ip_address(self):
+        return getattr(self.default_endpoint, "ipaddress", resolve_hostname(str(self.hostname)))
 
     @variable(alias='ui')
     def num_server(self):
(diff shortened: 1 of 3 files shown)


douglasgabriel:
failures are not caused by this PR changes

conf/cfme_data.yaml.template
```diff
@@ -170,6 +170,7 @@ management_systems:
             default:
                 credentials: lenovo
                 ipaddress: 10.0.0.1
+                hostname: 10.243.6.103
```
mkoura: QUESTION: why do the addresses of `hostname` and `ipaddress` differ?

mkoura: QUESTION: Why IP address and not hostname? Or why different IP address?

psav: **REQUIRED:** Yeh please set to 10.0.0.1

douglasgabriel: About the difference between `ipaddress` and `hostname`, it was a typo, that I will fix. I've added this new property because I couldn't figure out how to access the `ipaddress` data on the `LenovoProvider` constructor, but I was able to access the `hostname` prop, so I go on and add this new property even it is duplicating data. (see [lenovo.py](https://github.com/ManageIQ/integration_tests/blob/master/cfme/physical/provider/lenovo.py#L34))

Do you know some way to prevent this duplication of data? Maybe a solution to access the `ipaddress` prop into `LenovoProvider`

douglasgabriel:
@mkoura ?

mkoura:
Looks good, is it still in WIP for a reason? Please change to RFR if it's ready.

douglasgabriel:
I adapted the way that Physical Providers access the `ip_address` and `hostname` information, to look more like what is done in other places, like you oriented me :+1:

cfme/physical/provider/__init__.py
```diff
@@ -37,6 +38,14 @@ def __init__(
         self.name = name
         self.key = key
 
+    @property
+    def hostname(self):
+        return getattr(self.default_endpoint, "hostname", None)
+
+    @property
```
psav: **REQUIRED:** I'm not sure I like this, it means do a DNS resolution during collection which at the moment could be considered suboptimal.

douglasgabriel: Do you mean use **host name (ex: integration.tests)** instead an **IP address (ex: 10.0.0.1)**? @psav

PR #6779, opened 2018-03-01
[1LP][RFR] Extending clean_appliance.py

jan-zmeskal:
__Extending__ clean_appliance.py utility script.
1) `-q` argument for quiet
2) Waiting for closing all DB connections. If you try to call evm:db:reset and some connections are still active, the call will fail
3) I noticed that evm:db:reset sometimes leaves appliance in a state from where it cannot be started using `systemctl start evmserverd`. AFAIK this only happens with 5.8. Therefore handling this with condition.

### DIFF
--- a/scripts/clean_appliance.py
+++ b/scripts/clean_appliance.py
@@ -18,6 +18,7 @@ import sys
 from cfme.utils.conf import credentials
 from cfme.utils.path import scripts_path
 from cfme.utils.ssh import SSHClient
+from cfme.utils.wait import wait_for_decorator
 
 
 def main():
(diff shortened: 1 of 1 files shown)


scripts/clean_appliance.py
```diff
@@ -38,14 +42,30 @@ def main():
...
+    with SSHClient(stream_output=not args.quiet, **ssh_kwargs) as ssh_client:
         # Graceful stop is done here even though it is slower than killing ruby processes.
         # Problem with killing ruby is that evm_watchdog gets spawned right after being killed
         # and then prevents you from destroying the DB.
         # Graceful stop makes sure there are no active connections to the DB left.
         print('Stopping evmserverd...')
         ssh_client.run_command('systemctl stop evmserverd')
+
+        @wait_for_decorator(num_sec=60, delay=5)
+        def check_no_db_connections():
```
jamesooden: I don't see this subroutine being called anywhere? Is it being called anywhere?

jan-zmeskal: It is called thanks to the decorator. If you use that decorator, below function will get called wrapped in `wait_for`.

jamesooden: Cool, thanks I understand now.

PR #6783, opened 2018-03-01
[1LP][RFR] Stabilization fix to test_db_restore

lcouzens:
Changed fixture usage for test to improve test passing rate due to template issues.


jkrocil:
The PRT failure is not related.

PR #6785, opened 2018-03-01
[1LP][RFR] Adding Physical Server API tests

douglasgabriel:
Purpose or Intent
=================

__Adding tests__ for the Physical Server API:

- GET
  -  Existent physical server;
  -  Nonexistent physical server.
- Power actions:
  -  invalid_action;
  -  refresh;
  -  power_on;
  -  power_off;
  -  power_off_now;
  -  restart;
  -  restart_now.
- Led actions:
  -  blink_loc_led;
  -  turn_on_loc_led;
  -  turn_off_loc_led.

Implementation details
=================
Once the tests are covering the entire cycle of *power operations* it needs to wait the following steps to be conclude:
1. Send request to ManageIQ;
1. Send request to Physical Servers Provider;
1. The operation be concluded on Physical Server;
1. The refresh cycle of Physical Server Provider to identify the power state of Physical Server change;
1. The refresh cycle of ManageIQ to get the new state of Physical Server;
1. The Integration Tests verify the state change.

Due to this the **timeout** of **each action tests** was set to **5 minutes** and the delay between the state change verification was set to **1 minute**.

### DIFF
--- a/cfme/tests/physical_infrastructure/api/test_physical_server_api.py
+++ b/cfme/tests/physical_infrastructure/api/test_physical_server_api.py
@@ -1,0 +1,88 @@
+# -*- coding: utf-8 -*-
+import pytest
+
+from cfme.physical.provider.lenovo import LenovoProvider
+from cfme.utils import error
+from cfme.utils.rest import assert_response
+from cfme.utils.wait import wait_for
+
+pytestmark = [
+    pytest.mark.tier(3),
+    pytest.mark.provider([LenovoProvider], scope='module')
+]
+
+# The max time that a single test has to perform the entire cycle of action
+# in seconds (5 min)
+TIMEOUT = 600
+
+# The time interval to perform a verification of state change
+# in seconds (1 min)
+DELAY = 60
+
+
+@pytest.fixture(scope="module")
+def physical_server(setup_provider_modscope, appliance):
+    physical_server = appliance.rest_api.collections.physical_servers[0]
+    return physical_server
+
+
+def get_server_attr(attr_name, provider, physical_server):
+    provider.refresh_provider_relationships()
+    physical_server.reload()
+    return physical_server[attr_name]
+
+
+def test_get_physical_server(physical_server, appliance):
+    existent_server = appliance.rest_api.get_entity('physical_servers', physical_server.id)
+    existent_server.reload
+    assert_response(appliance)
+
+
+def test_get_nonexistent_physical_server(appliance):
+    nonexistent = appliance.rest_api.get_entity('physical_servers', 999999)
+    with error.expected('ActiveRecord::RecordNotFound'):
+        nonexistent.reload()
+    assert_response(appliance, http_status=404)
+
+
+def test_invalid_action(physical_server, appliance):
+    payload = {
+        "action" : "invalid_action"
+    }
+    with error.expected('Api::BadRequestError'):
+        appliance.rest_api.post(physical_server.href, **payload)
+
+
+def test_refresh_physical_server(appliance, physical_server):
+    assert getattr(physical_server.action, "refresh")()
+    assert_response(appliance)
+
+
+actions = [
+    ("power_off", "power_state", "off"),
+    ("power_on", "power_state", "on"),
+    ("power_off_now", "power_state", "off"),
+    ("restart", "power_state", "on"),
+    ("restart_now", "power_state", "on"),
+    ("blink_loc_led", "location_led_state", "Blinking"),
+    ("turn_on_loc_led", "location_led_state", "On"),
+    ("turn_off_loc_led", "location_led_state", "Off")
+]
+@pytest.mark.parametrize("action, verification_attr, desired_state", actions, ids=[action[0] for action in actions])
+def test_server_actions(physical_server, appliance, provider, action, verification_attr, desired_state):
+    """ Test the physical server actions sending the action request, waiting the task be complete on MiQ
+        and then waiting the state of some attribute of physical server be change
+    Params:
+        * action:            the action to be perform against the Physical Server
+        * verification_attr: the physical server attribute that will be change by the action
+        * desired_state:     the value of the attribute after the action execution
+    Metadata:
+        test_flag: rest
+    """
+    assert getattr(physical_server.action, action)()
+    assert_response(appliance)
+    wait_for(
+        lambda: get_server_attr(verification_attr, provider, physical_server).lower() == desired_state.lower(),
+        num_sec=TIMEOUT,
+        delay=DELAY
+    )


douglasgabriel:
failures are not caused by this PR changes

cfme/tests/physical_infrastructure/api/test_physical_server_api.py
```diff
@@ -0,0 +1,122 @@
...
+]
+
+"""
+The max time that a single test has to perform the entire cycle of action
+"""
+TIMEOUT = 300 # in seconds (5 min)
+"""
+The time interval to perform a verification of state change
+"""
+DELAY   = 60 # in seconds (1 min)
```
quarckster: **REQUIRED**
Please follow pep8. Replace double quotes by `#` here. Please avoid using inline comments.

cfme/tests/physical_infrastructure/api/test_physical_server_api.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def verify_attribute(server, attribute, value):
+    server.reload()
+    return server[attribute] == value
+
+
+def test_get_physical_server(physical_server, appliance):
+  existent_server = appliance.rest_api.get_entity('physical_servers', physical_server.id)
+  assert existent_server is not None
```
quarckster: **REQUIRED**
Here and further the indentation should be 4 spaces.

cfme/tests/physical_infrastructure/api/test_physical_server_api.py
```diff
@@ -0,0 +1,122 @@
...
+
+@pytest.mark.parametrize("action, desired_state",[
+    ("blink_loc_led", "Blinking"),
+    ("turn_on_loc_led", "On"),
+    ("turn_off_loc_led", "Off")
+])
+def test_server_led_actions(physical_server, appliance, provider, action, desired_state):
+    assert perform_physical_server_action(action, appliance, physical_server)
+    wait_for_loc_led_state_change(desired_state, appliance, provider, physical_server)
+    assert verify_attribute(physical_server, "location_led_state" ,desired_state)
```
quarckster: **REQUIRED**
Missed a blank line in the end.

cfme/tests/physical_infrastructure/api/test_physical_server_api.py
```diff
@@ -0,0 +1,122 @@
...
+@pytest.mark.parametrize("action, desired_state",[
+    ("power_off", "off"),
+    ("power_on", "on"),
+    ("power_off_now", "off"),
+    ("restart", "on"),
+    ("restart_now", "on")
+])
+def test_server_actions(physical_server, appliance, provider, action, desired_state):
+    assert perform_physical_server_action(action, appliance, physical_server)
+    wait_for_power_state_change(desired_state, appliance, provider, physical_server)
```
quarckster: **REQUIRED**
This wrapper function doesn't make sense.

cfme/tests/physical_infrastructure/api/test_physical_server_api.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+@pytest.mark.parametrize("action, desired_state",[
+    ("blink_loc_led", "Blinking"),
+    ("turn_on_loc_led", "On"),
+    ("turn_off_loc_led", "Off")
+])
+def test_server_led_actions(physical_server, appliance, provider, action, desired_state):
+    assert perform_physical_server_action(action, appliance, physical_server)
+    wait_for_loc_led_state_change(desired_state, appliance, provider, physical_server)
```
quarckster: **REQUIRED**
This wrapper function doesn't make sense.

cfme/tests/physical_infrastructure/api/test_physical_server_api.py
```diff
@@ -0,0 +1,113 @@
...
+    Perform an action againist a single physical server
+    """
+    payload = {
+        "action" : action
+    }
+    return do_post_physical_server(payload, appliance, physical_server)
+
+
+def get_server_attr(attr_name, appliance, provider, physical_server):
+    provider.refresh_provider_relationships()
```
mkoura: QUESTION: Is this really necessary? You are using this function in ``wait_for`` meaning the refresh is repeated over and over again and it's quite time-consuming operation.
If not also decrease ``DELAY`` so we don't need to wait whole minute.

cfme/tests/physical_infrastructure/api/test_physical_server_api.py
```diff
@@ -0,0 +1,113 @@
...
+def unexistent_physical_server(setup_provider_modscope, appliance):
+    return appliance.rest_api.get_entity('physical_servers', 999999)
+
+
+def do_post_physical_server(params, appliance, physical_server):
+    """
+    Do post againist a single physical server
+    """
+    appliance.rest_api.post(physical_server.href, **params)
+    return appliance.rest_api.response.ok
```
mkoura: REQUIRED: please use ``assert_response``

cfme/tests/physical_infrastructure/api/test_physical_server_api.py
```diff
@@ -0,0 +1,113 @@
...
+
+@pytest.mark.parametrize("action, desired_state",[
+    ("power_off", "off"),
+    ("power_on", "on"),
+    ("power_off_now", "off"),
+    ("restart", "on"),
+    ("restart_now", "on")
+])
+def test_server_actions(physical_server, appliance, provider, action, desired_state):
+    assert perform_physical_server_action(action, appliance, physical_server)
```
mkoura: REQUIRED: you can use ``resource.action.action_name()`` in our python API client. Please replace the ``perform_physical_server_action`` calls. In this case it would be smtg like
```python
assert getattr(physical_server.action, action)()
```
This way you can test that the action works AND that it is listed in available actions.

cfme/tests/physical_infrastructure/api/test_physical_server_api.py
```diff
@@ -0,0 +1,113 @@
...
+    payload = {
+        "action" : action
+    }
+    return do_post_physical_server(payload, appliance, physical_server)
+
+
+def get_server_attr(attr_name, appliance, provider, physical_server):
+    provider.refresh_provider_relationships()
+    physical_server.reload()
+    return physical_server[attr_name]
```
mkoura: QUESTION: are you sure it's accessible this way? Usually it's ``physical_server.attr_name``, or ``physical_server._data[attr_name]`` when you really want to access the dictionary.
The ``physical_server.attr_name`` is preferable.

cfme/tests/physical_infrastructure/api/test_physical_server_api.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def verify_attribute(server, attribute, value):
+    server.reload()
+    return server[attribute] == value
```
mkoura: QUESTION: same as above (wrt the attributes)

cfme/tests/physical_infrastructure/api/test_physical_server_api.py
```diff
@@ -0,0 +1,113 @@
...
+def test_get_physical_server(physical_server, appliance):
+    existent_server = appliance.rest_api.get_entity('physical_servers', physical_server.id)
+    assert existent_server is not None
+
+
+def test_get_nonexistent_physical_server(unexistent_physical_server, appliance):
+    with pytest.raises(APIException) as exception:
+      unexistent_physical_server.name
+    assert exception.match("Couldn't find PhysicalServer with")
+    assert appliance.rest_api.response.status_code == 404
```
mkoura: REQUIRED: smtg like:
```python
def test_get_nonexistent_physical_server(appliance):
    nonexistent = appliance.rest_api.get_entity('physical_servers', 999999)
    with error.expected('ActiveRecord::RecordNotFound'):
        nonexistent.reload()
    assert_response(appliance, http_status=404)
```

cfme/tests/physical_infrastructure/api/test_physical_server_api.py
```diff
@@ -0,0 +1,113 @@
...
+    ("power_off", "off"),
+    ("power_on", "on"),
+    ("power_off_now", "off"),
+    ("restart", "on"),
+    ("restart_now", "on")
+])
+def test_server_actions(physical_server, appliance, provider, action, desired_state):
+    assert perform_physical_server_action(action, appliance, physical_server)
+    wait_for(
+        lambda: get_server_attr("power_state", appliance, provider, server) == state, num_sec=TIMEOUT, delay=DELAY
```
mkoura: REQUIRED: I can't see where you get the ``state`` from. In any case the ``desired_state`` should be checked in the ``wait_for`` and there's no need for the ``assert verify_attribute`` afterwards (the ``wait_for`` does the checking and fails if the condition is not met).
smtg like:
```python
wait_for(
    lambda: physical_server.power_state.lower() == desired_state,
    # change accordingly if refresh_provider_relationships is really necessary
    fail_func=physical_server.reload,
    num_sec=TIMEOUT,
    delay=DELAY
)
```

cfme/tests/physical_infrastructure/api/test_physical_server_api.py
```diff
@@ -0,0 +1,113 @@
...
+    ("blink_loc_led", "Blinking"),
+    ("turn_on_loc_led", "On"),
+    ("turn_off_loc_led", "Off")
+])
+def test_server_led_actions(physical_server, appliance, provider, action, desired_state):
+    assert perform_physical_server_action(action, appliance, physical_server)
+    wait_for(
+        lambda: get_server_attr("location_led_state", appliance, provider,server) == state, num_sec=TIMEOUT, delay=DELAY
+    )
+    assert verify_attribute(physical_server, "location_led_state" ,desired_state)
```
mkoura: REQUIRED: same as above, also I don't recommend ``desired_state`` to be case sensitive (use ``.lower()`` where necessary)

cfme/tests/physical_infrastructure/api/test_physical_server_api.py
douglasgabriel: Great!

cfme/tests/physical_infrastructure/api/test_physical_server_api.py
douglasgabriel: Yeah, it is! I'm using this approach to I can pass the attribute name as parameter and reuse this code for many tests.

image

cfme/tests/physical_infrastructure/api/test_physical_server_api.py
douglasgabriel: :thinking: About the `disired_state`, was a typo, sorry about that :thinking: and about the `assert verify_attribute` I just want to make more clear what I'm really testing, but yeah, I can remove it

cfme/tests/physical_infrastructure/api/test_physical_server_api.py
douglasgabriel: Yes, It is necessary. The `refresh` of the MiQ must be done after Physical Provider conclude the entire operation (the points 2, 3 and 4 described on this PR description). The problem is that we can't assume a time to Physical Provider finish the operation, so we can't know when to execute the `refresh` to get the new state of Physical Provider. If we doesn't force the `refresh` and wait the "natural" refresh of the MiQ the tests could last for 30 min (according to my tests). So, even the `refresh` being a onerous task, it is really necessary.

cfme/tests/physical_infrastructure/api/test_physical_server_api.py
mkoura: The ``wait_for`` I suggested does the same check. If the `power_state` doesn't correspond to the `desired_state`, ``wait_for`` fails and the assert is not reached anyway.

douglasgabriel: Yeah, I got it :+1:

cfme/tests/physical_infrastructure/api/test_physical_server_api.py
```diff
@@ -0,0 +1,78 @@
...
+
+def get_server_attr(attr_name, provider, physical_server):
+    provider.refresh_provider_relationships()
+    physical_server.reload()
+    return physical_server[attr_name]
+
+
+def test_get_physical_server(physical_server, appliance):
+    existent_server = appliance.rest_api.get_entity('physical_servers', physical_server.id)
+    assert existent_server is not None
```
mkoura: REQUIRED: this will not fail even when the resource with id of ``physical_server.id`` doesn't exist. The ``existent_server`` will be an ``Entity`` instance, not ``None``. You need to do the same thing as you do in the ``test_get_nonexistent_physical_server``:
```python
existent_server.reload()  # it will fail here if the resource doesn't exist
assert_response(appliance)
```

cfme/tests/physical_infrastructure/api/test_physical_server_api.py
```diff
@@ -0,0 +1,78 @@
+# -*- coding: utf-8 -*-
+import pytest
+
+from cfme.physical.provider.lenovo import LenovoProvider
+from cfme.utils import error
+from cfme.utils.rest import assert_response
+from cfme.utils.wait import wait_for
+from manageiq_client.api import APIException
```
mkoura: REQUIRED: unused import

cfme/tests/physical_infrastructure/api/test_physical_server_api.py
douglasgabriel: Hmm, good point! Thank you @mkoura

cfme/tests/physical_infrastructure/api/test_physical_server_api.py
```diff
@@ -0,0 +1,87 @@
...
+def test_get_physical_server(physical_server, appliance):
+    existent_server = appliance.rest_api.get_entity('physical_servers', physical_server.id)
+    existent_server.reload
+    assert_response(appliance)
+
+
+def test_get_nonexistent_physical_server(appliance):
+    nonexistent = appliance.rest_api.get_entity('physical_servers', 999999)
+    with error.expected('ActiveRecord::RecordNotFound'):
+      nonexistent.reload()
```
quarckster: **REQUIRED**
Plase, keep four spaces for the indentation.

douglasgabriel: Sorry about that, VSCode issues :disappointed:

douglasgabriel:
Hey @mkoura I think that I've already address all your concerns. Please let me know if there is some more thing to care about.

cfme/tests/physical_infrastructure/api/test_physical_server_api.py
```diff
@@ -0,0 +1,87 @@
...
+    with error.expected('Api::BadRequestError'):
+        appliance.rest_api.post(physical_server.href, **payload)
+
+
+def test_refresh_physical_server(appliance, physical_server):
+    assert getattr(physical_server.action, "refresh")()
+    assert_response(appliance)
+
+
+@pytest.mark.parametrize("action, verification_attr, desired_state",[
```
psav: **REQUIRED:** Can we please add an ```ids``` parameter that just has the action name so that we don't have super long test ids. We can either just create a new list. Or do something like this

```python
actions = [
    ("power_off", "power_state", "off"),
    ("power_on", "power_state", "on"),
    ("power_off_now", "power_state", "off"),
    ("restart", "power_state", "on"),
    ("restart_now", "power_state", "on"),
    ("blink_loc_led", "location_led_state", "Blinking"),
    ("turn_on_loc_led", "location_led_state", "On"),
    ("turn_off_loc_led", "location_led_state", "Off")
]
@pytest.mark.parametrize("action, verification_attr, desired_state", actions, ids=[action[0] for action in actions])
......
```

douglasgabriel:
It's done, @psav 😄

PR #6786, opened 2018-03-01
[1LP][WIPTEST]Add widget and tests for changing the RBAC group for the currently logged in user via the webui

LandoCalrizzian:
Enables the currently logged in user to change its RBAC group via the webui via a call to BaseLoggedInPage.change_group. CFME 5.9 added support for a user being assigned to multiple groups. 

Follow up to #6749 

{{ pytest: -v cfme/tests/configure/test_access_control.py -k 'test_user_change_groups or test_user_group_test_user_crud or test_user_assign_multiple_groups or test_user_group_error_validation or test_group_ownership_on_user_or_group_role or test_group_ownership_on_user_only_role' }}

### DIFF
--- a/cfme/base/login.py
+++ b/cfme/base/login.py
@@ -1,6 +1,9 @@
 from widgetastic.exceptions import NoSuchElementException
 from widgetastic.widget import View
+from widgetastic_manageiq import SettingsNavDropdown
 from widgetastic_patternfly import NavDropdown, VerticalNavigation, FlashMessages
+
+from cfme.exceptions import CFMEException
 
 
 class BaseLoggedInPage(View):
@@ -14,10 +17,7 @@ class BaseLoggedInPage(View):
         './/div[starts-with(@class, "flash_text_div") or @id="flash_text_div"]'
     )
     help = NavDropdown('.//li[./a[@id="dropdownMenu1"]]|.//li[./a[@id="help-menu"]]')
-    settings = NavDropdown('.//li[./a[@id="dropdownMenu2"]]')
-    # 5.9 Locator for Settings item that replaces current group name when user has multiple groups
-    group_list_locator = (
-        './/ul/li[contains(@class, "dropdown-submenu") and contains(., "Change Group")]')
+    settings = SettingsNavDropdown('.//li[./a[@id="dropdownMenu2"]]')
     navigation = VerticalNavigation('#maintab')
 
     @property
@@ -29,6 +29,26 @@ class BaseLoggedInPage(View):
             return False
 
         return user.name == self.current_fullname
+
+    def change_group(self, group_name):
+        """ From the settings menu change to the group specified by 'group_name'
+            Only available in versions >= 5.9
+
+            User is required to be currently logged in
+        """
+        if self.extra.appliance.version < '5.9':
+            raise CFMEException("Changing the user group is not supported in versions < 5.9")
+
+        if not self.logged_in_as_user:
+            raise CFMEException("Unable to change group when a user is not logged in")
+
+        if group_name not in self.group_names:
+            raise CFMEException("{} is not an assigned group for {}".format(
+                group_name, self.current_username))
+
+        self.settings.groups.select_item(group_name)
+
+        return True
 
     @property
     def logged_in_as_current_user(self):
@@ -51,35 +71,31 @@ class BaseLoggedInPage(View):
 
     @property
     def current_groupname(self):
-        # TODO: Move these locators and accessors to a widget "Group Dropdown" widget
-        # 5.9 Locators for finding current group when user has multiple groups
-        current_group_locator = '{}/ul/li/a[@title="Currently Selected Group"]'.format(
-            self.group_list_locator)
-        current_group_marker = ' (Current Group)'
-        try:
-            current_group = self.browser.element(current_group_locator)
-            return self.browser.text(current_group).replace(current_group_marker, '')
-        except NoSuchElementException:
-            return self.settings.items[1].strip()
+        current_groups = self.settings.groups.items
+
+        # User is only assigned to one group
+        if len(current_groups) == 1:
+            return current_groups[0]
+
+        for group in current_groups:
+            if self.settings.groups.SELECTED_GROUP_MARKER in group:
+                return group.replace(self.settings.groups.SELECTED_GROUP_MARKER, '')
+        else:
+            # Handle some weird case where we don't detect a current group
+            raise CFMEException("User is not currently assigned to a group")
 
     @property
     def group_names(self):
         """ Return a list of the logged in user's assigned groups.
 
         Returns:
-            Version >= 5.9 - list of all groups the logged in user is assigned to
+            Version >= 5.9 - list containing all groups the logged in user is assigned to
             Version < 5.9 - single item list containing the user's current group
         """
-        # TODO: Move these locators and accessors to a widget "Group Dropdown" widget
-        group_list_locator = '{}/ul/li'.format(self.group_list_locator)
-        current_group_marker = ' (Current Group)'
 
-        group_list = self.browser.elements(group_list_locator)
-        if group_list:
-            return [
-                self.browser.text(group).replace(current_group_marker, '') for group in group_list]
-        else:
-            return [self.current_groupname]
+        return [
+            group.replace(self.settings.groups.SELECTED_GROUP_MARKER, '')
+            for group in self.settings.groups.items]
 
     @property
     def logged_in(self):
--- a/cfme/tests/configure/test_access_control.py
+++ b/cfme/tests/configure/test_access_control.py
@@ -115,7 +115,7 @@ def test_user_assign_multiple_groups(appliance, request, group_collection):
     group_names = [
         'EvmGroup-user', 'EvmGroup-administrator', 'EvmGroup-user_self_service', 'EvmGroup-desktop']
 
-    group_list = [group_collection.instantiate(description=names) for names in group_names]
+    group_list = [group_collection.instantiate(description=name) for name in group_names]
 
     user = new_user(appliance, groups=group_list)
 
@@ -128,6 +128,35 @@ def test_user_assign_multiple_groups(appliance, request, group_collection):
         assert set(assigned_groups) == set(group_names), (
             "User {} assigned groups {} are different from expected groups {}"
             .format(user, view.group_names, group_names))
+
+
+@pytest.mark.tier(2)
+@pytest.mark.uncollectif(lambda appliance: appliance.version < '5.9')
+def test_user_change_groups(appliance, group_collection):
+    """Assign a user to multiple groups and confirm that the user can successfully change groups
+    """
+    group_names = [
+        'EvmGroup-super_administrator', 'EvmGroup-administrator', 'EvmGroup-approver',
+        'EvmGroup-auditor', 'EvmGroup-desktop', 'EvmGroup-operator',
+        'EvmGroup-security', 'EvmGroup-user', 'EvmGroup-vm_user', ]
+
+    group_list = [group_collection.instantiate(description=name) for name in group_names]
+
+    test_user = new_user(appliance, groups=group_list)
+    with test_user:
+        view = navigate_to(appliance.server, 'LoggedIn')
+
+        orig_group = view.current_groupname
+
+        # Set the group list order so that we change to the original group last
+        group_test_list = [name for name in group_names if name != orig_group] + [orig_group]
+
+        for group in group_test_list:
+            view.change_group(group)
+
+            assert group == view.current_groupname, (
+                "User failed to change current group from {} to {}".format(
+                    view.current_groupname, group))
 
 
 # @pytest.mark.meta(blockers=[1035399]) # work around instead of skip
--- a/widgetastic_manageiq/__init__.py
+++ b/widgetastic_manageiq/__init__.py
@@ -15,7 +15,7 @@ from jsmin import jsmin
 from lxml.html import document_fromstring
 from selenium.common.exceptions import WebDriverException
 from wait_for import TimedOutError, wait_for
-from widgetastic.exceptions import NoSuchElementException
+from widgetastic.exceptions import NoSuchElementException, WidgetOperationFailed
 from widgetastic.log import logged
 from widgetastic.utils import ParametrizedLocator, Parameter, ParametrizedString, attributize_string
 from widgetastic.utils import VersionPick, Version
@@ -37,7 +37,7 @@ from widgetastic.widget import (
 from widgetastic.xpath import quote
 from widgetastic_patternfly import (
     Accordion as PFAccordion, BootstrapSwitch, BootstrapTreeview,
-    BootstrapSelect, Button, Dropdown, Input, VerticalNavigation, Tab)
+    BootstrapSelect, Button, Dropdown, Input, NavDropdown, VerticalNavigation, Tab)
 
 from cfme.exceptions import ItemNotFound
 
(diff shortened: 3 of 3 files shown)


dajoRH:
*CFME QE Bot*

widgetastic_manageiq/__init__.py
```diff
@@ -825,6 +825,115 @@ def workaround_save_issue(self):
...
+    @property
+    def expanded(self):
+        if not self.expandable:
+            return False
+        return self.browser.element(self.GROUP_SUBMENU, parent=self).is_displayed()
+
+    def expand(self):
+        """ Expand the Settings->Change Group submenu by hovering. """
+        if not self.expandable:
+            raise ValueError('{} is not expandable'.format(self.locator))
```
mshriver: Let's use a more specific exception here, I'd suggest:
https://github.com/RedHatQE/widgetastic.core/blob/master/src/widgetastic/exceptions.py#L17

widgetastic_manageiq/__init__.py
```diff
@@ -825,6 +825,115 @@ def workaround_save_issue(self):
...
+
+        # Make sure the Settings menu is expanded
+        if not self.parent.expanded:
+            self.parent.expand()
+
+        if not self.expanded:
+            # Hover over 'Change Group'
+            self.move_to()
+            if not self.expanded:
+                raise Exception('Could not expand {}'.format(self.locator))
```
mshriver: Again I'd suggest using `WidgetOperationFailed` exception here.

widgetastic_manageiq/__init__.py
```diff
@@ -825,6 +825,115 @@ def workaround_save_issue(self):
...
+    def expand(self):
+        """ Expand the Settings->Change Group submenu by hovering. """
+        if not self.expandable:
+            raise ValueError('{} is not expandable'.format(self.locator))
+
+        # Make sure the Settings menu is expanded
+        if not self.parent.expanded:
+            self.parent.expand()
+
+        if not self.expanded:
```
mshriver: This check feels unnecessary since its also checked after `self.move_to()` - if hovering while its already expanded doesn't close, I'd recommend skipping this first check and doing the `self.move_to()` in all cases, then checking expanded status once.

widgetastic_manageiq/__init__.py
```diff
@@ -825,6 +825,115 @@ def workaround_save_issue(self):
...
 
 
+class SettingsGroupSubmenu(NavDropdown):
+    """
+    Widget for the Group 'Dropdown' menu that allows a user assigned to multiple groups
+    to change their active group dynamically Multiple group functionality added in 5.9
+    """
+    GROUP_SUBMENU = './/ul[contains(@class, "dropdown-menu scrollable-menu")]'
+
+    def __init__(self, parent, locator, logger=None):
```
mshriver: All of these methods should have docblocks

widgetastic_manageiq/__init__.py
```diff
@@ -825,6 +825,115 @@ def workaround_save_issue(self):
...
+            self.browser.text(element)
+            for element
+            in self.browser.elements('{}/li'.format(self.GROUP_SUBMENU), parent=self)]
+
+    def has_item(self, item):
+        return item in self.items
+
+    def item_enabled(self, item):
+        if not self.has_item(item):
+            raise ValueError('There is no such item {}'.format(item))
```
mshriver: Include `self.items` in exception message so we can see from a failure whether it was something very small like incorrect capitalization.

Something like 'Available SettingsGroupSubmenu items are {}`.format(self.items) - you'll notice I've stuck this into the dropdown widgets for the same reason.

cfme/base/login.py
```diff
@@ -53,33 +75,28 @@ def current_fullname(self):
     def current_groupname(self):
         # TODO: Move these locators and accessors to a widget "Group Dropdown" widget
```
mshriver: Can remove this TODO :)

cfme/tests/configure/test_access_control.py
```diff
@@ -128,6 +128,31 @@ def test_user_assign_multiple_groups(appliance, request, group_collection):
...
 
+@pytest.mark.tier(2)
+@pytest.mark.uncollectif(lambda appliance: appliance.version < '5.9')
+def test_user_change_groups(appliance, request, group_collection):
+    """Assign a user to multiple groups and confirm that the user can successfully change groups
+    """
+    group_names = [
+        'EvmGroup-user', 'EvmGroup-user_self_service', 'EvmGroup-desktop']
+
+    group_list = [group_collection.instantiate(description=names) for names in group_names]
```
mshriver: nit-pick on plural vs singular of `names`

widgetastic_manageiq/__init__.py
LandoCalrizzian: Will do. Most of the code for this class was copied from the parent class with a few changes to account for the fact that the group dropdown uses hover to expand instead of a click and xpath locator updates

widgetastic_manageiq/__init__.py
```diff
@@ -825,6 +825,115 @@ def workaround_save_issue(self):
...
+            if not self.expanded:
+                raise Exception('Could not expand {}'.format(self.locator))
+            else:
+                self.logger.info('expanded')
+
+    def collapse(self):
+        if not (self.parent.is_displayed or self.expandable):
+            return
+
+        if self.expanded:
```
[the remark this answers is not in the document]
LandoCalrizzian: Good catch. It is an unnecessary since we check to see if the parent element is displayed first thing

widgetastic_manageiq/__init__.py
LandoCalrizzian: Good catch.

dajoRH:
Lint report for commit a37e0f70e6e435ad378cd16f5cdb263a551795d5:

- :bangbang: Line 940:21: **E126** *continuation line over-indented for hanging indent*

**No commit flaws detected**.

*CFME QE Bot*

cfme/base/login.py
```diff
@@ -30,6 +29,29 @@ def logged_in_as_user(self, user):
...
+            raise CFMEException("Changing the user group is not supported in versions < 5.8")
+
+        if not self.logged_in_as_user:
+            raise CFMEException("Unable to change group when a user is not logged in")
+
+        if group_name not in self.group_names:
+            raise CFMEException("{} is not an assigned group for {}".format(
+                group_name, self.current_username))
+
+        if self.current_groupname == group_name:
```
jamesooden: **QUESTION:** Normally in just a library to enable someone to do something, and not sitting in a test framework, I would do exactly this.   However I wonder if there is a situation where you want to test setting the group to the same group you are in, to make sure no awfulness happens when that is done?

cfme/base/login.py
```diff
@@ -51,35 +73,29 @@ def current_fullname(self):
...
         """ Return a list of the logged in user's assigned groups.
 
         Returns:
-            Version >= 5.9 - list of all groups the logged in user is assigned to
+            Version >= 5.9 - list containing all groups the logged in user is assigned to
             Version < 5.9 - single item list containing the user's current group
         """
-        # TODO: Move these locators and accessors to a widget "Group Dropdown" widget
-        group_list_locator = '{}/ul/li'.format(self.group_list_locator)
         current_group_marker = ' (Current Group)'
```
jamesooden: **REQUIRED:** This should be a constant at the top of the module.   It's used at least twice in different functions.

cfme/tests/configure/test_access_control.py
```diff
@@ -128,6 +128,31 @@ def test_user_assign_multiple_groups(appliance, request, group_collection):
...
+
+    curr_group = view.current_groupname
+
+    for group in [name for name in group_names if name != curr_group]:
+        view.change_group(group)
+        assert group == view.current_groupname, (
+            "User failed to change current group from {} to {}".format(
+                view.current_groupname, group))
+
+
```
jamesooden: **QUESTION:** Do we want to add a change back to the original group (which is held in curr_group presently).

cfme/base/login.py
```diff
@@ -30,6 +29,29 @@ def logged_in_as_user(self, user):
...
         return user.name == self.current_fullname
 
+    def change_group(self, group_name):
+        """ From the settings menu change to the group specified by 'group_name'
+            Only available in versions >= 5.9
+
+            User is required to be currently logged in
+        """
+        if self.extra.appliance.version < '5.9':
+            raise CFMEException("Changing the user group is not supported in versions < 5.8")
```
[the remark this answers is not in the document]
LandoCalrizzian: Good catch.

cfme/tests/configure/test_access_control.py
LandoCalrizzian: I don't think so since we creating a user specifically for this test and deleting it at the end.

cfme/base/login.py
```diff
@@ -51,35 +73,29 @@ def current_fullname(self):
...
-            current_group = self.browser.element(current_group_locator)
-            return self.browser.text(current_group).replace(current_group_marker, '')
-        except NoSuchElementException:
-            return self.settings.items[1].strip()
+        current_groups = self.settings.groups.items
+
+        for group in current_groups:
+            if current_group_marker in group:
+                return group.replace(current_group_marker, '')
+        else:
```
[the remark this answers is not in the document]
LandoCalrizzian: It is correct. In this context it's probably unnecessary since it's functions the same as replacing with a same return statement at the end of the method. I reworked it to make it clearer that we return the current group when only one group is assigned

cfme/tests/configure/test_access_control.py
jamesooden: It's not to leave the system in a sane state, its to make sure nothing weird happens when you change back to the same group (i.e. for completeness).

LandoCalrizzian: That's a valid point and I've updated the test so that it cycles back around to the original group. 
This test was designed to test basic webui functionality that the user can change groups. A true test for completeness would be implemented in a future test that would verify the users ability to iterate over all the possible permutations of default group transitions while validating that the navigation ui is updated accordingly after each transition.

cfme/base/login.py
LandoCalrizzian: That's a valid point. The current_groupname check was added for ease of use so the caller can blindly set the group w/o concern that it's already active since trying to change to the currently assigned group will throw a WidgetOperationFailed exception.

cfme/tests/configure/test_access_control.py
LandoCalrizzian: done

cfme/base/login.py
LandoCalrizzian: I moved it to the SettingsGroupSubMenu class

cfme/tests/configure/test_access_control.py
```diff
@@ -136,20 +136,26 @@ def test_user_change_groups(appliance, request, group_collection):
     """Assign a user to multiple groups and confirm that the user can successfully change groups
     """
     group_names = [
-        'EvmGroup-user', 'EvmGroup-user_self_service', 'EvmGroup-desktop']
+        'EvmGroup-super_administrator', 'EvmGroup-administrator', 'EvmGroup-approver',
+        'EvmGroup-auditor', 'EvmGroup-desktop', 'EvmGroup-operator',
+        'EvmGroup-security', 'EvmGroup-user', 'EvmGroup-vm_user', ]
 
-    group_list = [group_collection.instantiate(description=names) for names in group_names]
+    group_list = [group_collection.instantiate(description=name) for name in group_names]
```
jamesooden: I like this approach.   I would have simply tried to change the group back as a separate line of code after the loop.   This is much more elegant.

cfme/base/login.py
```diff
@@ -1,5 +1,7 @@
+from cfme.exceptions import CFMEException
```
rlbabyuk: Required: move it down as this intermal import from cfme.exceptions import CFMEException:

from widgetastic.exceptions import NoSuchElementException
from widgetastic.widget import View
from widgetastic_manageiq import SettingsNavDropdown
from widgetastic_patternfly import NavDropdown, VerticalNavigation, FlashMessages	 

from cfme.exceptions import CFMEException

cfme/tests/configure/test_access_control.py
```diff
@@ -130,6 +130,37 @@ def test_user_assign_multiple_groups(appliance, request, group_collection):
...
+    user = new_user(appliance, groups=group_list)
+    request.addfinalizer(user.delete)
+    request.addfinalizer(user.appliance.server.login_admin)
+
+    view = appliance.server.login(user)
+
+    orig_group = view.current_groupname
+
+    # Set the group list order so that we change to the original group last
+    group_test_list = [name for name in group_names if name != orig_group] + [orig_group]
```
rlbabyuk: Optional: .extend([orig_group]) can be used here

cfme/base/login.py
LandoCalrizzian: Updated

cfme/base/login.py
```diff
@@ -51,35 +70,31 @@ def current_fullname(self):
...
+
+        # User is only assigned to one group
+        if len(current_groups) == 1:
+            return current_groups[0]
+
+        for group in current_groups:
+            if self.settings.groups.SELECTED_GROUP_MARKER in group:
+                return group.replace(self.settings.groups.SELECTED_GROUP_MARKER, '')
+
+        # Handle some weird case where we don't detect a current group
```
mshriver: **Required**

Let's move this into an `else` block on the for loop, so the behavior here is a bit more explicit and this raise isn't just hanging off the end.

```python
for group in current_groups:
    if check:
        return group.replace.yadayada
else:
    # This block executes when return isn't hit in the for loop
    raise ShitsBroke('no active group')

File: cfme/tests/configure/test_access_control.py [hunk 18]
```diff
@@ -130,6 +130,37 @@ def test_user_assign_multiple_groups(appliance, request, group_collection):
...
+        'EvmGroup-auditor', 'EvmGroup-desktop', 'EvmGroup-operator',
+        'EvmGroup-security', 'EvmGroup-user', 'EvmGroup-vm_user', ]
+
+    group_list = [group_collection.instantiate(description=name) for name in group_names]
+
+    user = new_user(appliance, groups=group_list)
+    request.addfinalizer(user.delete)
+    request.addfinalizer(user.appliance.server.login_admin)
+
+    view = appliance.server.login(user)
```
mshriver: I think this should be reworked to use a context here, which will control the login and ensure a return to default appliance user (admin) when the context exits.

```python
with custom_user:
    # testing done while user is logged in
    # changing groups, etc

# out of the context
# now I can login_admin and do more test - or NOT, and the next test won't accidentally be logged in as 'custom_user'
```

Used in remote auth tests: https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/integration/test_cfme_auth.py#L120

mshriver: Here are the methods that control the context manager: 
https://github.com/ManageIQ/integration_tests/blob/c57379fe6213dda6deebcce2b7f5c1f88aceabc5/cfme/configure/access_control/__init__.py#L140

So you can logout/login, do whatever testing you want on the appliance, and within that context it will always use the custom_user instead of admin. Once the contextual code block is complete, `__exit__` will restore the previously assigned user (admin) as the default.

mshriver:
@LandoCalrizzian This looks great, ping me today and let's talk about the user context, it should be small but I'd accept it as an Optional and do it in a followup when more tests are written.

cfme/base/login.py
LandoCalrizzian: Updated

cfme/tests/configure/test_access_control.py
LandoCalrizzian: Updated to use context. I can't remember why I used request over context other than deleting the user at the end which isn't really required for testing at the moment

dajoRH:
Lint report for commit 2206f3263d3d27c4341d3ef76ae42230662ef4a5:

- :red_circle: Line 9:1: **F401** *'cfme.base.login.BaseLoggedInPage' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

mshriver:
changes look good, keeping an eye on PRT results - won't merge until lint is fixed though ;)

PR #6787, opened 2018-03-02
[1LP][NOTEST][RFR] Adding Saved reports All Navigation View

niyazRedhat:
Detail page navigation failing because Navigation step was missing for All page.


rlbabyuk:
@niyazRedhat Do we have some test to check this navigation?

niyazRedhat:
@rlbabyuk There is no test case for that But i want to write test case which will deletes all saved reports.

mshriver:
PRT smoke test failures unrelated.

PR #6788, opened 2018-03-02
[1LP][RFR] Removed vm/host_rising in test_utilization_metrics.

rhrazdil:
Removed  vm/host_rising in test_utilization_metrics because it was necessary for host and vm to report metrics within 15 seconds interval. It should be sufficient to wait until non-zero number of metrics are reported by host and vm.


rhrazdil:
2 tests that failed also fail in trackerbot because of 0 values, but no tests failed on collecting metrics.

PR #6790, opened 2018-03-02
[RFR] Fixing test_providers_summary

jan-zmeskal:
__Fixing__ test_providers_summary. The test used to failed with: 

> AttributeError: 'ProviderDetailsSummaryView' object has no attribute 'properties'

I am also excluding redhat_network provider from the test.

{{pytest: cfme/tests/intelligence/reports/test_canned_corresponds.py::test_providers_summary -vv}}

### DIFF
--- a/cfme/tests/intelligence/reports/test_canned_corresponds.py
+++ b/cfme/tests/intelligence/reports/test_canned_corresponds.py
@@ -32,11 +32,13 @@ def test_providers_summary(soft_assert):
     presence of units and rounding."""
     path = ["Configuration Management", "Providers", "Providers Summary"]
     report = CannedSavedReport.new(path)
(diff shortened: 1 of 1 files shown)


cfme/tests/intelligence/reports/test_canned_corresponds.py
```diff
@@ -32,11 +32,12 @@ def test_providers_summary(soft_assert):
     presence of units and rounding."""
     path = ["Configuration Management", "Providers", "Providers Summary"]
     report = CannedSavedReport.new(path)
+    skipped_providers = {"ec2", "openstack", "redhat_network"}  # Skip cloud and network
```
jamesooden: Would you add the reason these are being skipped to the comment.   That will help future maintainers if they are evaluating fi they should still be skipped.   Might need to put the comment in front of the statement then.

PR #6794, opened 2018-03-02
[WIPTEST] Turn the other two recv_ready()'s into functions

jamesooden:
I got two of the recv_ready calls in ssh.py but missed the other
two.  This fixes that.


mshriver:
Small PRs are good, but you can do 4 lines in one PR :P

mshriver:
Reverting this as it triggered recv failures while now running these methods.  @psav and @jamesooden did debug, we will reinstate these method calls with other fixes included.

PR #6796, opened 2018-03-02
Revert "[WIPTEST] Turn the other two recv_ready()'s into functions"

mshriver:
Reverts ManageIQ/integration_tests#6794


mshriver:
doesn't have correct signature for PRT/automation, re-submitting as my own commit.

PR #6799, opened 2018-03-05
[1LP][RFR] Add Nuage to the RECOGNIZED_BY_CREDS list

miha-plesko:
With this commit we add Nuage provder to the RECOGNIZED_BY_CREDS in order to make fixtures work again. Doing this it turns out that Nuage provider is not hashable, which causes problems when adding it to a set(), which we also fix in this commit.

### DIFF
--- a/cfme/networks/provider/__init__.py
+++ b/cfme/networks/provider/__init__.py
@@ -28,7 +28,7 @@ from cfme.utils.appliance.implementations.ui import navigator, CFMENavigateStep,
 from cfme.utils.log import logger
 
 
-@attr.s
+@attr.s(hash=False)
 class NetworkProvider(BaseProvider, WidgetasticTaggable, BaseEntity):
     """ Class representing network provider in sdn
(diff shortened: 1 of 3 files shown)


cfme/networks/provider/__init__.py
```diff
@@ -28,7 +28,7 @@
 from cfme.utils.log import logger
 
 
-@attr.s
+@attr.s(hash=False)
```
quarckster: **QUESTION**
Is it necessary to change it here?

miha-plesko: Yes, otherwise python-attrs library forces own `__hash__` function on `NetworkProvider` hence the one from `BaseProvider` (which we actually want)  gets overwritten. I'm quite certain, because it didn't work until I've added the `False` here as well.

psav: We have seen this in the past - I think it's OK here

PR #6713, opened 2018-02-21
[1LP][RFR] Replacement get_detail() for providers

quarckster:
Intent
=================

Continuation of replacement get_detail() method by calling ParametrizedSummaryTable in respectful views.

{{pytest: -v -k "test_infra_provider_relationships or test_assigned_roles" --long-running}}

PR #6802, opened 2018-03-05
[1LP][RFR] Make `summary("Relationships")` great again for NetworkManagers

miha-plesko:
This is a followup commit for a PR where we introduced ParametrizedSummaryTable for all providers at once, but NetworkProviders were somehow only partially refactored and hence left in broken state.

The PR we're fixing: https://github.com/ManageIQ/integration_tests/pull/6713

### DIFF
--- a/cfme/networks/views.py
+++ b/cfme/networks/views.py
@@ -5,7 +5,7 @@ from widgetastic_patternfly import (Dropdown, Accordion, Button, TextInput,
 from cfme.base.login import BaseLoggedInPage
 from cfme.common.provider_views import ProviderAddView, ProviderEditView
 from widgetastic_manageiq import (ManageIQTree, SummaryTable, ItemsToolBarViewSelector,
-                                  BaseEntitiesView)
+                                  BaseEntitiesView, ParametrizedSummaryTable)
 
 
 class NetworkProviderToolBar(View):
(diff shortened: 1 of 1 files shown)


cfme/networks/views.py
```diff
@@ -82,11 +82,7 @@ class NetworkProviderDetailsView(BaseLoggedInPage):
     @View.nested
     class entities(View):  # noqa
         """ Represents details page when it's switched to Summary/Table view """
-        properties = SummaryTable(title="Properties")
```
rlbabyuk: **Required**: Please update deleted properties usages in tests to new summary property

miha-plesko:
Hi @rlbabyuk, good point, you're right. TBH I actually thought that Nuage is the only provider that uses NetworkProviderDetailsView page, but apparently it isn't!

Trying to refactor as you suggest I gave up after an hour since it's just too difficult for me to know what occurences of `entities.properties, entities.relationships, ...` to replace all over the project. In other words, I don't have the nuts to do it, especially because there's no way for me to test for other providers than Nuage.

Eventually I decided to only add the new `ParametrizedSummaryTable` feature without removing anything. Can you please see if it's OK this way?

izapolsk:
nice title :thumbsup:

miha-plesko:
@rlbabyuk any chance we push this one forward? It's quite annoying that I need to cherry-pick it everytime I want to run Nuage tests 😇

chargio:
@miha-plesko 
Could you please address the changes requested?


Adding @loicavenel

miha-plesko:
@chargio changes have already been addressed 3 days ago, I think we're good to merge.

cfme/networks/views.py
```diff
@@ -82,12 +82,15 @@ class NetworkProviderDetailsView(BaseLoggedInPage):
...
     class entities(View):  # noqa
         """ Represents details page when it's switched to Summary/Table view """
+        # TODO: remove all SummaryTables in favor of ParametrizedSummaryTable when safe
         properties = SummaryTable(title="Properties")
         relationships = SummaryTable(title="Relationships")
         status = SummaryTable(title="Status")
         overview = SummaryTable(title="Overview")
         smart_management = SummaryTable(title="Smart Management")
 
+        summary = ParametrizedView.nested(ParametrizedSummaryTable)
```
mshriver: Why is this a nested ParametrizedView?  I think just setting it as an instance of `ParametrizedSummaryTable()` is the correct thing to do here, see existing uses of it in the codebase.

miha-plesko: You're right, fixed. It works either way so I didn't even notice 😄

PR #6339, opened 2018-01-05
[WIP] fix for test_canned_corresponds.py->test_datastores_summary

izapolsk:
{{pytest: cfme/tests/intelligence/reports/test_canned_corresponds.py -k test_datastores_summary}}

PR #6803, opened 2018-03-05
[WIPTEST]Updating test_datastores_summary

nzr-zts:
Purpose
=================
Updating _test_datastores_summary_ to decrease the delay between creating a report via UI and querying db. Because time to time while running regression we are able to see different values in a report and in the database which causes failure of the test. It could be caused by other slaves which provision VM at that time.

{{pytest: -vvv cfme/tests/intelligence/reports/test_canned_corresponds.py -k test_datastores_summary --use-provider complete --long-running}}

### DIFF
--- a/cfme/tests/intelligence/reports/test_canned_corresponds.py
+++ b/cfme/tests/intelligence/reports/test_canned_corresponds.py
@@ -142,20 +142,11 @@ def test_datastores_summary(soft_assert, appliance, request):
     vms = adb['vms']
     host_storages = adb['host_storages']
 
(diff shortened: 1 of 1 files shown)


izapolsk:
@nzr-zts, man this won't help. I played around with this test and report for a while.
https://github.com/ManageIQ/integration_tests/pull/6339
I tried to decrease timeout and even add some kind of threshold. it didn't help.

It seems to me this is bug in 5.9. Maybe something related to caching mechanizm. It especially visible when you kick off this test against several providers. Data in tables and data in report are different.
Please investigate and submit BZ.

PR #6805, opened 2018-03-05
[1LP][RFR] New db ui test

rlbabyuk:
Purpose or Intent
=================
Added Summary View for ui Database settings
Added New test to check last db start time, test for db garbage collection
{{pytest: -vv -k 'test_configure_vmdb_last_start_time or test_configuration_database_garbage_collection'}}

### DIFF
--- a/cfme/tests/configure/test_database_ui.py
+++ b/cfme/tests/configure/test_database_ui.py
@@ -5,9 +5,9 @@ from dateutil import parser
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.log_validator import LogValidator
 
+
 @pytest.mark.tier(2)
 @pytest.mark.uncollectif(lambda appliance: appliance.version < '5.9')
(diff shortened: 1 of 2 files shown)


cfme/tests/configure/test_database_ui.py
```diff
@@ -0,0 +1,47 @@
+import pytest
+import pytz
+from dateutil import parser
+
+from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.log_validator import LogValidator
+
+@pytest.mark.tier(2)
+@pytest.mark.uncollectif(lambda appliance: appliance.version < '5.9')
+# this is work starting 5.9
```
psav: **REQUIRED**: I don't like the comment being here, can we have it at the end of the decorator line, one could argue it's not even needed as the decorator tells you that.

rlbabyuk: comment is removed

PR #6806, opened 2018-03-05
[1LP][RFR] Fix sdn_add_topology

t-fischer:
Add more providers to the test
And fix it ( it was only disabliing the items otherwise)
 {{pytest: -v -q cfme/tests/networks/test_sdn_topology.py --use-provider=complete}}

### DIFF
--- a/cfme/common/topology.py
+++ b/cfme/common/topology.py
@@ -58,7 +58,8 @@ class BaseTopologyView(BaseLoggedInPage):
 
         @property
         def is_enabled(self):
-            return 'active' in self.browser.get_attribute('class', self.el)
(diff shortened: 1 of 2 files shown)


quarckster:
:+1:

cfme/common/topology.py
```diff
@@ -58,7 +58,8 @@ class legends(ParametrizedView):  # noqa
 
         @property
         def is_enabled(self):
-            return 'active' in self.browser.get_attribute('class', self.el)
+            el = self.browser.element('./ancestor::kubernetes-topology-icon[1]', self.el)
```
psav: **REQUIRED:** This seems specific to kubernetes, what about other topologies?

quarckster: @psav It seems it's common to all providers

PR #6809, opened 2018-03-05
[DOC][RFR][NOTEST] Sprout - Remove misleading files and make a handy deployment documentation

mfalesni:


izapolsk:
1. please mention that HUBBER_URL should be set
2. users should belong to some group, providers should also belong to that group.

PR #5933, opened 2017-11-23
[WIP] Projects Dashboard Tests

pavelzag:
# Purpose or Intent

{{pytest: cfme/tests/containers/test_projects_dashboard.py -v --use-provider cm-env2 }}

This test is checking the data integrity of the Objects counter box and the Pods Table in the Containers Projects' Dashboard view.
It's depending on https://github.com/ManageIQ/wrapanapi/pull/193 PR that allows fetching Pods and Services object per selected project. It also allows getting Images per specific project
https://bugzilla.redhat.com/show_bug.cgi?id=1461618

PR #6810, opened 2018-03-05
[WIP]Test projects dashboard

brahmanim:
Purpose or Intent
=================
This PR based on pavelzag PR [ManageIQ/integration_tests#5933](https://github.com/ManageIQ/integration_tests/pull/5933) .
This test is checking the data integrity of the Objects counter box and the Pods Table in the Containers Projects' Dashboard view.
It's depending on ManageIQ/wrapanapi#193 PR that allows fetching Pods and Services object per selected project. It also allows getting Images per specific project
https://bugzilla.redhat.com/show_bug.cgi?id=1461618

{{pytest: cfme/tests/containers/test_projects_dashboard.py -v --use-provider ocp-v2 }}


cfme/tests/containers/test_projects_dashboard.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def get_api_object_counts(appliance, project_name, provider):
+    """ Fetches amount of Containers/Services/Images from the API per selected project name"""
+    out = {
+        Container: 0,
+        Service: 0,
+        Image: 0
+    }
+    if isinstance(provider, ContainersProvider):
```
gshefer: __REQUIRED__
You should be sure that `provider` is instance of `ContainersProvider`. Is there any alternative "`provider`"?
I'd change this entire function to the following:
```python
assert isinstance(provider, ContainersProvider)
return {
    Container: len(provider.mgmt.list_container(project_name=project_name))
    Service: len(provider.mgmt.list_service(project_name=project_name))
    Image: len(get_container_images_amt(provider, project_name))
}
```

cfme/tests/containers/test_projects_dashboard.py
```diff
@@ -0,0 +1,96 @@
...
+
+pytestmark = [
+    pytest.mark.uncollectif(lambda provider: current_version() < "5.9"),
+    pytest.mark.usefixtures('setup_provider'),
+    pytest.mark.tier(1),
+    pytest.mark.provider([ContainersProvider], scope='function')
+]
+
+tested_objects = [Container, Image, Service]
+project_name = 'test-project-dashboard'  # Predefine project for this test
```
gshefer: __OPTIONAL__
It's common habit in Python to define constants as UPPERCASE.

PavelZagalsky-Via:
LGTM

cfme/tests/containers/test_projects_dashboard.py
```diff
@@ -0,0 +1,96 @@
...
+    Steps:
+        * Go to Projects / Dashboard View
+        * Compare the data in the status boxes to API data forz
+        Containers/Images/Services numbers
+    """
+    api_values = get_api_object_counts(appliance, project_name, provider)
+    collection_object = appliance.collections.container_projects
+    inst = collection_object.instantiate(name=project_name, provider=provider)
+    view = navigate_to(inst, 'Dashboard')
+    for object in tested_objects:
```
gshefer: __REQUIRED__
You are overriding the built-in `object` here, change to `obj` or something.

cfme/tests/containers/test_projects_dashboard.py
brahmanim: Change the function, you are right it looks better like that.

cfme/tests/containers/test_projects_dashboard.py
brahmanim: change constants to UPPERCASE.

cfme/tests/containers/test_projects_dashboard.py
brahmanim: change to obj

cfme/tests/containers/test_projects_dashboard.py
```diff
@@ -0,0 +1,92 @@
...
+from cfme.containers.container import Container
+from cfme.containers.image import Image
+from cfme.containers.service import Service
+
+from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.version import current_version
+
+
+pytestmark = [
+    pytest.mark.uncollectif(lambda provider: current_version() < "5.9"),
```
mshriver: **Required**

Do not use `current_version()`, use appliance fixture's version attribute.

That aside, I think you would be better off using `pytest.mark.ignore_stream('5.8')` here.

cfme/tests/containers/test_projects_dashboard.py
```diff
@@ -0,0 +1,92 @@
+import pytest
+
+from cfme.containers.provider import ContainersProvider
+from cfme.containers.container import Container
+from cfme.containers.image import Image
+from cfme.containers.service import Service
+
```
mshriver: remove empty line, all of these imports are local.

cfme/containers/project.py
```diff
@@ -100,6 +110,22 @@ def resetter(self):
             self.view.toolbar.view_selector.select("Summary View")
 
 
+@navigator.register(Project, 'Dashboard')
+class Dashboard(CFMENavigateStep):
+    VIEW = ProjectDetailsView
```
mshriver: Shouldn't this be `ProjectDashboardView`, the new view class created above?

cfme/containers/project.py
```diff
@@ -22,12 +22,17 @@ class ProjectDetailsView(ContainerObjectDetailsBaseView):
...
+    pass
+
+
 @attr.s
 class Project(BaseEntity, Taggable, Labelable):
 
     PLURAL = 'Projects'
     all_view = ProjectAllView
     details_view = ProjectDetailsView
+    dashboard_view = ProjectDashboardView
```
mshriver: I would recommend _not_ defining this here, as it can be easily queried from the navigation destination class's `VIEW` attribute.

There's a method in the Navigate object:
https://github.com/RedHatQE/navmazing/blob/master/navmazing/__init__.py#L77

And examples in the codebase of the query:
https://github.com/ManageIQ/integration_tests/blob/master/cfme/common/provider.py#L185

cfme/containers/project.py
```diff
@@ -36,6 +41,11 @@ class Project(BaseEntity, Taggable, Labelable):
     def mgmt(self):
         return ApiProject(self.provider.mgmt, self.name)
 
+    @classmethod
+    def get_project_entity(cls, provider, provider_name):
+        selected_provider = provider.mgmt.get_project_by_name(provider_name)
+        return cls(selected_provider.name, provider)
```
mshriver: **Required**

Remove this, its serving the same function as the `ProjectCollection.instantiate` method provided by `BaseCollection`.  Maybe there's some other purpose or need that `instantiate` isn't solving, but I don't think this is the correct way to instantiate project objects.  They will have no appliance attribute.

Also this function isn't used at all in this PR, what's the intended purpose?  The function should have a docblock that explains this, if it even is necessary.

cfme/tests/containers/test_projects_dashboard.py
```diff
@@ -0,0 +1,92 @@
...
+def test_projects_dashboard_pods(provider, appliance, soft_assert):
+    """Tests data integrity of Pods names in Pods status box in Projects Dashboard.
+    Steps:
+        * Go to Projects / Dashboard View
+        * Compare the data in the Pods status box to API data for
+        Pods names
+    """
+    api_pod_names = get_api_pods_names(provider)
+    collection_object = appliance.collections.container_projects
+    inst = collection_object.instantiate(name=PROJECT_NAME, provider=provider)
```
mshriver: **Required**

Use a fixture to generate the project instance, eliminating duplicate code and moving this functionality into test setup.

cfme/tests/containers/test_projects_dashboard.py
```diff
@@ -0,0 +1,92 @@
...
+        * Go to Projects / Dashboard View
+        * Compare the data in the status boxes to API data forz
+        Containers/Images/Services numbers
+    """
+    api_values = get_api_object_counts(appliance, PROJECT_NAME, provider)
+    collection_object = appliance.collections.container_projects
+    inst = collection_object.instantiate(name=PROJECT_NAME, provider=provider)
+    view = navigate_to(inst, 'Dashboard')
+    for obj in TESTED_OBJECTS:
+        statusbox_value = getattr(view, obj.PLURAL.split(' ')[-1].lower()).value
```
mshriver: document in-line what you're trying to get from `obj.PLURAL.split()[-1].lower.value, as its not at all clear what's being parsed.

cfme/tests/containers/test_projects_dashboard.py
```diff
@@ -0,0 +1,92 @@
...
+    Steps:
+        * Go to Projects / Dashboard View
+        * Compare the data in the status boxes to API data forz
+        Containers/Images/Services numbers
+    """
+    api_values = get_api_object_counts(appliance, PROJECT_NAME, provider)
+    collection_object = appliance.collections.container_projects
+    inst = collection_object.instantiate(name=PROJECT_NAME, provider=provider)
+    view = navigate_to(inst, 'Dashboard')
+    for obj in TESTED_OBJECTS:
```
mshriver: ambiguous naming here, these aren't objects, they're classes.

cfme/tests/containers/test_projects_dashboard.py
```diff
@@ -0,0 +1,92 @@
...
+
+TESTED_OBJECTS = [Container, Image, Service]
+PROJECT_NAME = 'test-project-dashboard'  # Predefine project for this test
+
+
+def get_api_object_counts(appliance, project_name, provider):
+    """ Fetches amount of Containers/Services/Images from the API per selected project name"""
+    assert isinstance(provider, ContainersProvider)
+    return {
+        Container: len(provider.mgmt.list_container(project_name=project_name)),
```
mshriver: I think if this return is going to be keyed on the classes, those class keys should be used in the test instead of TESTED_OBJECTS list.

```python
for cls in api_values.keys():
    <check UI value of cls.__name__ / cls.PLURAL etc>

Comment (mshriver) [reply cfme/tests/containers/test_projects_dashboard.py]: Basically the idea is to link the data sets somehow - right now this method generates a dictionary keyed on classes that are hard coded here.  If TESTED_OBJECTS is changed, but a new class isn't added here to match, then the test will have a mismatch it what its trying to compare.  The test should just use the class keys that are in the dictionary returned by this method, not some other arbitrary list of classes.

**Comment by mshriver** [comment]
Please squash commit history as well.

File: cfme/tests/containers/test_projects_dashboard.py (hunk 4 above)
Comment (brahmanim) [reply cfme/tests/containers/test_projects_dashboard.py]: Use pytest.mark.ignore_stream('5.8') as you suggest.
It definitely looks better.

File: cfme/tests/containers/test_projects_dashboard.py (hunk 5 above)
Comment (brahmanim) [reply cfme/tests/containers/test_projects_dashboard.py]: remove the empty line.

File: cfme/containers/project.py (hunk 6 above)
Comment (brahmanim) [reply cfme/containers/project.py]: Yes, you are right.
Change it to ProjectDashboardView.

File: cfme/containers/project.py (hunk 7 above)
Comment (brahmanim) [reply cfme/containers/project.py]: Agree, remove the dashboard_view definition.

File: cfme/containers/project.py (hunk 8 above)
Comment (brahmanim) [reply cfme/containers/project.py]: Sorry, I didn't pay attention to that.
This function is leftover from origin Pavel PR were there was no collection implementation yet.
Indeed I use the ProjectCollection.instantiate on my test instead.
I removed this function.

File: cfme/tests/containers/test_projects_dashboard.py (hunk 9 above)
Comment (brahmanim) [reply cfme/tests/containers/test_projects_dashboard.py]: Create new fixture container_project_instance, use it instead of duplication code.

File: cfme/tests/containers/test_projects_dashboard.py (hunk 12 above)
Comment (brahmanim) [reply cfme/tests/containers/test_projects_dashboard.py]: Remove TESTED_OBJECTS, use api_values.keys() instead.

**Comment by brahmanim** [comment]
@mshriver 
Hi Mike,

PRT fail on appliance 5.8, that test shouldn't be running on it at all!
I used pytest.mark.ignore_stream('5.8') as your suggestion. 

For some reason, the appliance IP from sprout is 0.0.0.0,
and the error message:
2018-04-12 15:39:33,824 [E] (0.0.0.0) appliance.version could not be retrieved from REST, falling back (cfme/utils/appliance/__init__.py:844)

I run the PRT twice and get the same problem.
You can see logs on http://10.16.4.32/trackerbot/pr/run/23153.

Please your advice about it!

**Comment by mshriver** [comment]
@brahmanim the 58z PRT/sprout issue has been noted and won't be held against you here ;)

File: cfme/tests/containers/test_projects_dashboard.py [hunk 13]
```diff
@@ -0,0 +1,95 @@
...
+    pytest.mark.tier(1),
+    pytest.mark.provider([ContainersProvider], scope='function')
+]
+
+PROJECT_NAME = 'test-project-dashboard'  # Predefine project for this test
+
+
+@pytest.fixture(scope='function')
+def container_project_instance(appliance, provider):
+    collection_object = appliance.collections.container_projects
```
mshriver: :+1:

cfme/tests/containers/test_projects_dashboard.py
```diff
@@ -0,0 +1,95 @@
...
+        Container: len(provider.mgmt.list_container(project_name=project_name)),
+        Service: len(provider.mgmt.list_service(project_name=project_name)),
+        Image: len(get_container_images_amt(provider, project_name))
+    }
+
+
+def get_container_images_amt(provider, project_name=None):
+    """ Fetches images amount from the API per selected project name"""
+    images = provider.mgmt.list_image()
+    project_images = []
```
mshriver: For future reference, I think this could have been a generator in-line above. I've expanded it to 4 lines here for readability.

```python
project_images = [
    img for img 
    in provider.mgmt.list_image() 
    if img.image_project_name == project_name
]

mshriver:
Modified description to run against more providers in PRT

brahmanim:
ocp-v2 has known problem with some pods creation using NFS storage (PV).
That problem also causes another problem with deleting\recreat a project.
I expect these test failures on ocp-v2.
We are trying to solve this problem on this provider, but it will take some time.

cfme/tests/containers/test_projects_dashboard.py
brahmanim: change the code, indeed that is more readable.

mshriver:
@brahmanim That provider will be collected against these tests, given how the test is written.

There are a few options to handle this:
1. ocp-v2 needs to be disabled
2. ProviderFilter with required_tags or required_flags: 
    https://github.com/ManageIQ/integration_tests/blob/master/cfme/utils/providers.py#L42

PR #6811, opened 2018-03-05
[1LP][RFR] Turn eventing ON for Nuage provider

miha-plesko:
With this commit we extend Nuage provider so that it now inputs AMQP credentials upon creation as well. Until now we were only inputing the Default credentials.

Also, with this commit we update test that was only testing
- CREATE
- RETRIEVE
- DELETE

of Nuage provider to now also validate

- UPDATE

which gives us a complete CRUD test in the end.

### DIFF
--- a/cfme/networks/provider/__init__.py
+++ b/cfme/networks/provider/__init__.py
@@ -1,6 +1,7 @@
 import attr
 from cached_property import cached_property
 from navmazing import NavigateToSibling, NavigateToAttribute
+from widgetastic.exceptions import MoveTargetOutOfBoundsException
 
 from cfme.exceptions import DestinationNotFound
 from cfme.common import WidgetasticTaggable
@@ -21,7 +22,8 @@ from cfme.networks.views import (
     OneProviderNetworkPortView,
     OneProviderNetworkRouterView,
     OneProviderSecurityGroupView,
-    OneProviderSubnetView
+    OneProviderSubnetView,
+    NetworkProviderEditView
 )
 from cfme.utils import version
 from cfme.utils.appliance.implementations.ui import navigator, CFMENavigateStep, navigate_to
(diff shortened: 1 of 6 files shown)


cfme/networks/provider/__init__.py
```diff
@@ -144,6 +146,76 @@ def create(self, cancel=False, validate_credentials=True, validate_inventory=Fal
...
+                            NotImplementedError(
+                                "Such endpoint doesn't have change password/key button")
+
+                        endp_view.fill(endpoint.credentials.view_value_mapping)
+                if (validate_credentials and hasattr(endp_view, 'validate') and
+                        endp_view.validate.is_displayed):
+                    endp_view.validate.click()
+                    self._post_validate_checks(edit_view)
+
+        details_view = self.create_view(navigator.get_class(self, 'Details').VIEW)
```
jamesooden: These two views are not used on cancel.   you should probably limit there scope to only where they are used.

cfme/networks/provider/__init__.py
```diff
@@ -144,6 +146,76 @@ def create(self, cancel=False, validate_credentials=True, validate_inventory=Fal
...
+        main_view = self.create_view(navigator.get_class(self, 'All').VIEW)
+
+        if cancel:
+            edit_view.cancel.click()
```
jamesooden: Why not just have a _post_cance() routine and include the cancel click with the checks.   It seems kind of odd the separation.

cfme/networks/provider/__init__.py
```diff
@@ -144,6 +146,76 @@ def create(self, cancel=False, validate_credentials=True, validate_inventory=Fal
...
+            self._post_cancel_edit_checks(main_view)
+        else:
+            edit_view.save.click()
```
jamesooden: Sames as 207.

jamesooden: Though this has more logic to it due to the endpoints, but I think the argument still holds.

cfme/networks/provider/__init__.py
```diff
@@ -232,6 +321,20 @@ def step(self):
...
+    VIEW = NetworkProviderEditView
+    prerequisite = NavigateToSibling('All')
+
+    def step(self):
+        self.prerequisite_view.entities.get_entity(name=self.obj.name, surf_pages=True).check()
+        try:
+            self.prerequisite_view.toolbar.configuration.item_select('Edit Selected Network Provider')
+        except MoveTargetOutOfBoundsException:
+            # TODO: Remove once fixed 1475303
+            self.prerequisite_view.toolbar.configuration.item_select('Edit Selected Network Provider')
```
jamesooden: Perhaps a comment on why just retrying the operation is acceptable.

cfme/networks/provider/nuage.py
```diff
@@ -69,9 +83,14 @@ def mgmt(self):
 
     @classmethod
     def from_config(cls, prov_config, prov_key, appliance=None):
```
jamesooden: Would you add a docstring here.   Every little bit helps.

conf/cfme_data.yaml.template
```diff
@@ -180,6 +180,12 @@ management_systems:
                 credentials: nuage
```
jamesooden: **REQUIRED:** These changes need to be done in the cfme_yaml repository.

conf/credentials.yaml.template
```diff
@@ -32,6 +32,9 @@ lenovo:
 nuage:
   username: admin
   password: smartvm
+nuage_events:
```
jamesooden: **REQUIRED:** These changes need to be done in the cfme_yaml repository.

jamesooden:
If you look at the PRT run it's got the following error:

    ERROR: Bad signature. Please sign your commits!

I'm guess you have not gotten your gpg key stuff setup for your git commits.   If you go to the [CFME QE New Hire page](https://mojo.redhat.com/docs/DOC-1067199), and to the section "Setting up your personal GPG key." there are instructions on how to do this.   Let me know if you need any help.

cfme/networks/provider/nuage.py
miha-plesko: Good point, thanks, done.

cfme/tests/networks/test_network_providers.py
```diff
@@ -1,8 +1,10 @@
 # -*- coding: utf-8 -*-
 import pytest
+import uuid
```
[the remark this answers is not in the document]
miha-plesko: Not where I copied this [from](https://github.com/manageiq/integration_tests/blob/master/cfme/tests/cloud/test_providers.py#L184) 😉 , but done. I too prefer shorter random strings over full-blown UUID.

cfme/networks/provider/__init__.py
mfalesni: We had broken toolbars at one point when JS decided to be bad on us. Try if it works without the retry code, please :)

mfalesni:
@miha-plesko I reckon that you as an external person can't have access to our internal providers, so this may get a bit tricky reviewing.

cfme/networks/provider/__init__.py
jamesooden: Sure, I was hoping you knew why (-:   If not no need for a comment.   I can't fault you for those before you not leaving the reason they were dong this.

cfme/networks/provider/nuage.py
jamesooden: sweet

jamesooden:
@miha-plesko All your changes look good to me.   @psav  is going to handle the commit signing issue.   I'm going approve what I have seen, but that still leaves the [1LP] review.

miha-plesko:
@jamesooden thank you.
@mfalesni you're right, it works now even without the ugly try-catch. I've removed it and repushed.
@psav from what I understand, this PR is now on you.

PR #6815, opened 2018-03-05
[1LP][RFR] Update CMQE tests to use the new parametized summary table

jawatts:
During examination of the RC jobs, the tests were failing with the following attribute error:
AttributeError
'ProviderDetailsSummaryView' object has no attribute 'custom_attributes'

Updating tests to use the new parameritized table

{{ pytest: cfme/tests/containers/test_containers_summary.py cfme/tests/containers/test_static_custom_attributes.py --use-provider ocp-v2 --long-running }}

### DIFF
--- a/cfme/tests/containers/test_containers_summary.py
+++ b/cfme/tests/containers/test_containers_summary.py
@@ -12,7 +12,6 @@ from cfme.containers.image_registry import ImageRegistry
 from cfme.containers.node import Node
 
 from cfme.utils.appliance.implementations.ui import navigate_to
-from cfme.utils.blockers import BZ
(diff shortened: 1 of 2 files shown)


jawatts:
Waiting on https://github.com/ManageIQ/integration_tests/pull/6835 to be merged before moving this RFR

jawatts:
I think we should combine test_add_static_custom_attributes, test_edit_static_custom_attributes,  test_delete_static_custom_attributes into a single CRUD test. What do you think @gshefer ? This would be handled in another PR. We would have to create a new CMP ID and remove the old ones.

PR #6822, opened 2018-03-06
[1LP][RFR] make ip fields bigger in sprout in order to fit openshift hostnames

izapolsk:
It turned out that Appliance.ip_address fields cannot fit openshift hostname in prod sprout whilst it worked fine locally. This is fix for this issue.


sprout/appliances/models.py
```diff
@@ -748,9 +748,9 @@ class Power(object):
     appliance_pool = models.ForeignKey("AppliancePool", null=True, on_delete=models.CASCADE,
         help_text="Which appliance pool this appliance belongs to.")
     name = models.CharField(max_length=64, help_text="Appliance's name as it is in the provider.")
-    ip_address = models.CharField(max_length=45, null=True, help_text="Appliance's IP address")
+    ip_address = models.CharField(max_length=64, null=True, help_text="Appliance's IP address")
```
mshriver: Should be 253 (or 256 for roundness) to account for max DNS length allowed by RFC1034

izapolsk: you're  right but openshift and some other stuff doesn't  allow names longer than 64

izapolsk: even more I had to shorten names to approximately 50 chars because 58-60 caused different issues.

PR #6823, opened 2018-03-06
[1LP][RFR]fixes for vm_reconfigure

lkhomenk:
Purpose or Intent
=================

__Fixing__  vm_reconfigure. CPU&RAM test were broken after last commit
Current fix would also work for various combinations of CPU-RAM-DISK reconfigurations
Consideration: default values for `sockets/cores_per_socket/mem_size/mem_size_unit` - `reconfigure()` doesn't know anything about `old_config`

### DIFF
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -874,6 +874,15 @@ class Vm(VM):
         fill_data = {k: v for k, v in changes.iteritems() if k != 'disks'}
         vm_recfg.fill(fill_data)
 
+        # Helpers for VM Reconfigure request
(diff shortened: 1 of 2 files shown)


lkhomenk:
@jan-zmeskal @mshriver ^ please take a look

cfme/infrastructure/virtual_machines.py
```diff
@@ -874,6 +874,15 @@ def reconfigure(self, new_configuration=None, changes=None, cancel=False):
         fill_data = {k: v for k, v in changes.iteritems() if k != 'disks'}
         vm_recfg.fill(fill_data)
 
+        # Helpers for VM Reconfigure request
```
mshriver: :+1:

PR #6824, opened 2018-03-07
[1LP][RFR] Uncollecting test_vm_console for RHV-CFME integration

jan-zmeskal:
__Uncollecting__ test_vm_console for RHV-CFME integration. The reason is quite pragmatic. The console tests are very much browser-dependent. In order to achieve stability, we wound need to emulate CFME QE testing environment. Now in the context of scaling down of resources dedicated to this effort, we cannot really do that for the purpose of this test case.

{{pytest:  cfme/tests/ssui/test_ssui_myservice.py::test_vm_console}}

### DIFF
--- a/cfme/tests/ssui/test_ssui_myservice.py
+++ b/cfme/tests/ssui/test_ssui_myservice.py
@@ -54,9 +54,10 @@ def test_retire_service(appliance, setup_provider, context, provision_request):
 @pytest.mark.meta(blockers=[BZ(1544535, forced_streams=['5.9'])])
 @pytest.mark.parametrize('context', [ViaSSUI])
 @pytest.mark.parametrize('provision_request', [['console_test']], indirect=True)
-@pytest.mark.uncollectif(lambda provider: provider.one_of(VMwareProvider) and
-                         provider.version >= 6.5,
-                         'VNC consoles are unsupported on VMware ESXi 6.5 and later')
+@pytest.mark.uncollectif(lambda provider:
+    provider.one_of(VMwareProvider) and provider.version >= 6.5 or
+    'html5_console' in provider.data.excluded_test_flags,
+    'VNC consoles are unsupported on VMware ESXi 6.5 and later')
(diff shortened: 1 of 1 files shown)


cfme/tests/ssui/test_ssui_myservice.py
```diff
@@ -55,8 +55,8 @@ def test_retire_service(appliance, setup_provider, context, order_catalog_item_i
 @pytest.mark.meta(blockers=[BZ(1544535, forced_streams=['5.9'])])
 @pytest.mark.parametrize('context', [ViaSSUI])
 @pytest.mark.parametrize('order_catalog_item_in_ops_ui', [['console_test']], indirect=True)
-@pytest.mark.uncollectif(lambda provider: provider.one_of(VMwareProvider) and
-                         provider.version >= 6.5,
+@pytest.mark.uncollectif(lambda provider: (provider.one_of(VMwareProvider) and
+                         provider.version >= 6.5) or provider.name == 'rhv_cfme_integration',
```
mshriver: **Required**
I don't think this is a good way to uncollect a test, since this is based only on what the provider name is.

Providers have 'excluded_test_flags' in their data, and I think uncollecting based on the inclusion of the `html5_console` flag in the yaml is a much better way of uncollecting this test.


```
@pytest.mark.uncollectif(lambda provider: (provider.one_of(VMwareProvider) and provider.version >= 6.5) or
                                                                     'html5_console' in provider.data.excluded_test_flags)
```

This could also be done with a provider mark that includes this flag in the provider filter. This module is marked at the top for provider parametrization, but you can separately decorate individual functions for unique generation:

```
@pytest.mark.provider(gen_func=providers,
                                      filters=[ProviderFilter(classes=[InfraProvider, CloudProvider],
                                                  required_fields=['provisioning'],
                                                  required_flags=['html5_console']])])
```


Lastly this could be done with test metadata specifying the required_flag, but we're trying to deprecate and phase out this implementation type.

dajoRH:
*CFME QE Bot*

PR #6835, opened 2018-03-07
[1LP][RFR] ParametrizedSummaryTable.is_displayed support

izapolsk:
I see two ways to fix this:
1. add __locator__
2. copy SummaryTable.BASELOC to ParametrizedSummaryTable.ROOT + ParametrizedString

I chose first way.

### DIFF
--- a/widgetastic_manageiq/__init__.py
+++ b/widgetastic_manageiq/__init__.py
@@ -603,6 +603,9 @@ class ParametrizedSummaryTable(ParametrizedView):
     PARAMETERS = ("title", )
     _table = SummaryTable(title=Parameter("title"))
 
+    def __locator__(self):
+        return self._table.ROOT
+
     @property
     def fields(self):
         return self._table.fields


mshriver:
I definitely prefer this solution to option \#2.  I also think directly implementing `ParametrizedSummaryTable.is_displayed` (returning self._table.is_displayed) would be a good approach since this class is a ParametrizedView, not a widget, and doesn't really have its own locator.  All of its methods are just wrapping calls to SummaryTable.<method> through `self._table`.

widgetastic_manageiq/__init__.py
```diff
@@ -603,6 +603,9 @@ class ParametrizedSummaryTable(ParametrizedView):
     PARAMETERS = ("title", )
     _table = SummaryTable(title=Parameter("title"))
 
+    def __locator__(self):
```
mshriver: PRT failures look directly related, as they're failing to find the Relationships table on provider details for stat validation.

jawatts: Tested the changes locally and observed the same failures. I tried a different fix by implementing is_displayed within ParametrizedSummaryTable and 

```
@property
    def is_displayed(self):
        return self._table.is_displayed
```
Failures were not observed with this fix. Maybe an alternate fix?

izapolsk:
@mfalesni, honestly, I think we have to remove that private widget and move code to primary table instead. But this change requires some efforts. so, I don't consider it atm :)

PR #6838, opened 2018-03-07
[1LP][RFR] image_upload_glance : Add image location url

nachandr:
Note: Another file has crept into this PR.I will take care of it.

### DIFF
--- a/scripts/image_upload_glance.py
+++ b/scripts/image_upload_glance.py
@@ -11,8 +11,11 @@ Usage:
 2.scripts/image_upload_glance.py --image cfme-rhevm-5.8.2.1-1.x86_64.qcow2 \
                                  --image_name_in_glance cfme-rhevm-5.8.2.1-1.x86_64.qcow2 \
                                  --provider glance11 \
+3.scripts/image_upload_glance.py --image_name_in_glance cfme-5922.qcow2 \
+                                 --provider glance11-server \
+                                 --url http://user:pswd@xyz:8080/cfme-rhevm-5.9.0.22-1.x86_64.qcow2
 
-If disk_format is not passed, it defaults to qcow2.
+Note : If disk_format is not passed, it defaults to qcow2.
 """
 import argparse
 import sys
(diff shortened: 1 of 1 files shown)


scripts/image_upload_glance.py
```diff
@@ -27,16 +27,17 @@ def parse_cmd_line():
     parser = argparse.ArgumentParser(description=__doc__,
                                      formatter_class=argparse.RawDescriptionHelpFormatter)
 
-    parser.add_argument('--image', help='Image to be uploaded to Glance', required=True)
+    parser.add_argument('--image', help='Image to be uploaded to Glance', required=False)
     parser.add_argument('--image_name_in_glance', help='Image name in Glance', required=True)
     parser.add_argument('--provider', help='Glance provider key in cfme_data', required=True)
     parser.add_argument('--disk_format', help='Disk format of image', default='qcow2')
+    parser.add_argument('--url', help='Location of image to be uploaded to Glance')
```
jamesooden: It may be be better to use URL rather than Location in the help, as one could thing of a file path as giving  location too.   With the context they should get that, but I think URL is more specific.

cfme/tests/intelligence/reports/test_metering_report.py
```diff
@@ -1,4 +1,8 @@
 # -*- coding: utf-8 -*-
+# All providers that support C&U support Metering Reports.SCVMM doesn't support C&U.
```
mshriver: @nachandr I don't think you want this commit in this PR :)

mshriver:
Need to cleanup the metrics file that got included, otherwise looks good, glad it was a simple addition!

nachandr:
@mshriver, done.

cfme/tests/intelligence/reports/test_metering_report.py
nachandr: @mshriver , done

scripts/image_upload_glance.py
nachandr: @jamesooden , done

scripts/image_upload_glance.py
```diff
@@ -27,16 +30,17 @@ def parse_cmd_line():
     parser = argparse.ArgumentParser(description=__doc__,
                                      formatter_class=argparse.RawDescriptionHelpFormatter)
 
-    parser.add_argument('--image', help='Image to be uploaded to Glance', required=True)
+    parser.add_argument('--image', help='Image to be uploaded to Glance', required=False)
```
izapolsk: REQUIRED: As far as I understand one of parameters --image or --url is required. Please add a check that at least one of them is present.

nachandr: @izapolsk , done

PR #6839, opened 2018-03-08
[WIPTEST]Ansible service catalog in SSUI

sshveta:
Created a fixture file for ansible .
These fixtures can be used in other ansible tests too .


cfme/fixtures/ansible_fixtures.py
```diff
@@ -0,0 +1,106 @@
...
+    ansible_service_catalog.order()
+    cat_item_name = ansible_catalog_item.name
+    request_descr = "Provisioning Service [{0}] from [{0}]".format(cat_item_name)
+    service_request = appliance.collections.requests.instantiate(description=request_descr)
+    if service_request.exists():
+        service_request.wait_for_request()
+        service_request.remove_request()
+    yield cat_item_name
+
+    @request.addfinalizer
```
quarckster: **REQUIRED**
You don't need `addfinalizer` in a yield fixture.

quarckster:
Why the fixtures have only `function` scope. What if it's required keeping an ansible catalog item across the whole module?

sshveta:
Changed all fixtures to module scoped .

PR #6842, opened 2018-03-08
[1LP][RFR] New test added for checking repos

lcouzens:
Added new test to check default repos of appliance. Also checks what repos are set after clicking default button.
{{pytest: -v -k "test_rhsm_registration_check_repo_names"}}

### DIFF
--- a/cfme/tests/configure/test_register_appliance.py
+++ b/cfme/tests/configure/test_register_appliance.py
@@ -178,6 +178,7 @@ def test_rh_registration(appliance, request, reg_method, reg_data, proxy_url, pr
 def test_rhsm_registration_check_repo_names(temp_appliance_preconfig_funcscope, soft_assert):
     """ Checks default rpm repos on a fresh appliance """
     ver = temp_appliance_preconfig_funcscope.version.series()
+    # TODO We need the context manager here is that RedHatUpdates doesn't yet support Collections.
     with temp_appliance_preconfig_funcscope:
         view = navigate_to(RedHatUpdates, 'Edit')
         soft_assert(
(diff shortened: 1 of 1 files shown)


cfme/tests/configure/test_register_appliance.py
```diff
@@ -174,6 +175,20 @@ def test_rh_registration(appliance, request, reg_method, reg_data, proxy_url, pr
     request.addfinalizer(appliance.unregister)
 
 
+def test_rhsm_registration_check_repo_names(temp_appliance_preconfig_funcscope, soft_assert):
+    """ Checks default repo on a fresh appliance and checks """
```
jamesooden: Would you change "Checks default repo" to "Checks default rpm repo".   That wil make it clearer what type of repo it is checking.

cfme/tests/configure/test_register_appliance.py
```diff
@@ -174,6 +175,20 @@ def test_rh_registration(appliance, request, reg_method, reg_data, proxy_url, pr
     request.addfinalizer(appliance.unregister)
 
 
+def test_rhsm_registration_check_repo_names(temp_appliance_preconfig_funcscope, soft_assert):
+    """ Checks default rpm repos on a fresh appliance """
+    ver = temp_appliance_preconfig_funcscope.version.series()
+    with temp_appliance_preconfig_funcscope:
```
psav: **REQUIRED:** I think we should add a TODO here, **TODO:** We shouldn't have to do this. The only reason we need the context manager here is that ```RedHatUpdates``` doesn't yet support Collections.

PR #6844, opened 2018-03-08
[1LP][RFR] Refactor get_yaml_config/set_yaml_config for REST

mshriver:
Implemented for Server, Zone, and Region

Appliance set_yaml_config runs against the master server

Repetitive code for get/set yaml_config, but each uses a different 'id' attribute. But as a single method, each class would still need a method wrapping them, and I don't think this is appropriate to implement in BaseEntity, thus, code duplication.

Requesting lots of reviews, I want to get a lot of eyes on this.

I've tested locally, would love to get some feedback regardless of PRT results at this time, given the impact.

### DIFF
--- a/cfme/base/__init__.py
+++ b/cfme/base/__init__.py
@@ -62,12 +62,12 @@ class Server(BaseEntity, sentaku.modeling.ElementMixin):
         return self.zone.collections.servers.filter({'slave': True}).all()
 
     @property
-    def api_settings_url(self):
+    def _api_settings_url(self):
         return '/'.join([self.appliance.rest_api.collections.servers._href, self.sid, 'settings'])
 
     def get_yaml_config(self):
         """GET servers/:id/settings api endpoint to query server configuration"""
-        return self.appliance.rest_api.get(url=self.api_settings_url)
+        return self.appliance.rest_api.get(url=self._api_settings_url)
 
     def set_yaml_config(self, data_dict):
         """PATCH settings from the server's api/server/:id/settings endpoint
@@ -82,7 +82,7 @@ class Server(BaseEntity, sentaku.modeling.ElementMixin):
         # Failing with some data_dict, like 'authentication'
         # https://bugzilla.redhat.com/show_bug.cgi?id=1553394
         result = self.appliance.rest_api._session.patch(
-            url=self.api_settings_url,
+            url=self._api_settings_url,
             data=json.dumps(data_dict)
         )
         assert result.ok
@@ -154,12 +154,12 @@ class Zone(Pretty, BaseEntity, sentaku.modeling.ElementMixin):
         return region_obj
 
     @property
-    def api_settings_url(self):
+    def _api_settings_url(self):
         return '/'.join([self.appliance.rest_api.collections.zones._href, self.id, 'settings'])
 
     def get_yaml_config(self):
         """"GET zones/:id/settings api endpoint to query zone configuration"""
-        return self.appliance.rest_api.get(self.api_settings_url)
+        return self.appliance.rest_api.get(self._api_settings_url)
 
     def set_yaml_config(self, data_dict):
         """PATCH settings from the zone's api/zones/:id/settings endpoint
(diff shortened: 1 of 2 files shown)


cfme/base/__init__.py
```diff
@@ -59,6 +61,32 @@ def zone(self):
     def slave_servers(self):
         return self.zone.collections.servers.filter({'slave': True}).all()
 
+    @property
```
jamesooden: Anyway these could not be duplicated?   That is it really looks like the same code in both cases.

mshriver: Server uses `sid`, Zone uses `id`, and Region uses `number` attributes, and the collection referenced is different.  Possibly a settingsmixin class, but each would still need a property to know what attribute to reference for the `id` field.  Doesn't make sense in the context of BaseEntity to put it there.

cfme/utils/appliance/__init__.py
```diff
@@ -1959,44 +1959,22 @@ def is_storage_enabled(self):
...
-        if not result:
-            raise Exception('Unable to set config: {!r}:{!r}'.format(result.rc, result.output))
+        Args:
+            data_dict: dictionary of the changes to be made to the yaml configuration
+                       JSON dumps data_dict to pass as raw hash data to rest_api session
+        Raises:
+            ApplianceException when server_id isn't set
+        """
+        # Can only modify through server ID, raise if that's not set yet
+        if self.server_id() is None:
```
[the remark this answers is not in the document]
mshriver: In the Server class, `sid` has a default and thus will never be None and it won't happen when called directly against a server.

The appliance on the other hand, could have no master server set, thus `self.server` would raise an IndexError in ServerCollection.get_master().  `self.server_id()` catches this occurrence and returns None.

I could switch this to try `self.server.set_yaml_config()` and catch the IndexError that would happen if `self.server` fails.

cfme/base/__init__.py
mshriver: It does indeed work, and they are in fact different settings dicts: http://pastebin.test.redhat.com/563692

cfme/base/__init__.py
```diff
@@ -59,6 +61,32 @@ def zone(self):
...
+    def _api_settings_url(self):
+        return '/'.join([self.appliance.rest_api.collections.servers._href, self.sid, 'settings'])
+
+    def get_yaml_config(self):
+        """GET servers/:id/settings api endpoint to query server configuration"""
+        return self.appliance.rest_api.get(url=self._api_settings_url)
+
+    def set_yaml_config(self, data_dict):
```
quarckster: **QUESTION**
Why are `get_yaml_config` and `set_yaml_config` not a property?

```python
@property
def yaml_config(self):
    ...

@yaml_config.setter
def yaml_config(self, data_dict):
    ...
```

mshriver: Mostly because of existing appliance methods not being implemented as properties.  Don't see a reason why they couldn't be, would just require updating all the current uses.  Good suggestion!

@psav thoughts on expanding the scope here a bit to include changing these to properties?

quarckster: I think it's out of the scope. It should be done in another PR anyway.

mshriver: then we agree on all sides! :)

dajoRH:
Lint report for commit 9d249891c2b0001d85859a1419362c8cada2f0c7:

- :red_circle: Line 21:1: **F401** *'yaml' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

mshriver:
@mfalesni absolutely agree on passing only the desired changes to set_yaml_config, planned on doing so when I refactor for property+setter use instead of `get_/set_` as @quarckster suggested.

mshriver:
@psav @mfalesni cleaned up lint with last commit.

mshriver:
PRT results in 22117 are 'random' smoke test failures, not associated with any yaml get/set.

mfalesni:
Hoping for a good merge ...

PR #6846, opened 2018-03-09
[WIPTEST]Implementing new test case related to quota parent child.

anurag03:
Purpose or Intent
=================
Implementing new test cases related to quota to improve test coverage related to parent child.

{{pytest: -v -k 'test_setting_child_quota_more_than_parent'}}

### DIFF
--- a/cfme/tests/infrastructure/test_tenant_quota.py
+++ b/cfme/tests/infrastructure/test_tenant_quota.py
@@ -9,6 +9,7 @@ from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.provisioning import do_vm_provisioning
 from cfme.services.catalogs.catalog_item import CatalogItem
 from cfme.services.service_catalogs import ServiceCatalogs
+from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.generators import random_vm_name
 
 from widgetastic.utils import partial_match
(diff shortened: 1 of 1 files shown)


cfme/tests/infrastructure/test_tenant_quota.py
```diff
@@ -171,3 +186,31 @@ def test_tenant_quota_vm_reconfigure(appliance, provider, setup_provider, set_ro
...
+        [('storage', '1'), ('storage', '2'), 'Storage'],
+        [('vm', '1'), ('vm', '2'), 'Vms'],
+        [('template', '1'), ('template', '2'), 'Templates']
+    ],
+    ids=['cpu', 'memory', 'storage', 'vm', 'template']
+)
+def test_setting_child_quota_more_than_parent(tenants_setup, parent_quota, child_quota,
+                                              flash_text):
+    test_parent, test_child = tenants_setup
+    view = navigate_to(test_parent, 'ManageQuotas', wait_for_view=True)
```
psav: **REQUIRED:** Is there a ManageQuota method we can use instead of trying to interact with the forms directly.

anurag03: @psav :No, there is a method "set_quota" which takes kwargs as argument but here we need to pass the kargs with different variable name to set different parameters, so with parameterization it's not possible.

PR #6850, opened 2018-03-09
[1LP][RFR][NOTEST]adding stack removal for ec2

lkhomenk:
Purpose or Intent
=================
Removes EC2 Cloudformations - ManageIQ EC2 Stack objects

- Added Condition 1 day old stack to avoid interrupting other tests.
- Added `sleep(5)` to avoid "Rate exceeded"-issue. which probably won't ever happen unless we are not deleting 10+ stacks at a time
- Generates reports like:

```
Date: 2018-03-09 13:52:06.076403
| Provider Key   | Stack Name   |
|----------------+--------------|
| ec2-west       | testlkhom19  |
| ec2-west       | testlkhom18  |
| ec2-west       | testlkhom17  |
| ec2-west       | testlkhom16  |
| ec2-west       | testlkhom15  |
| ec2-west       | testlkhom14  |
| ec2-west       | testlkhom13  |
| ec2-west       | testlkhom12  |
| ec2-west       | testlkhom11  |
| ec2-west       | testlkhom10  |
| ec2-west       | testlkhom9   |
| ec2-west       | testlkhom8   |
| ec2-west       | testlkhom7   |
| ec2-west       | testlkhom6   |
| ec2-west       | testlkhom5   |
| ec2-west       | testlkhom4   |
| ec2-west       | testlkhom3   |
| ec2-west       | ec2          |
```

### DIFF
--- a/scripts/ec2cleanup.py
+++ b/scripts/ec2cleanup.py
@@ -2,7 +2,6 @@ import argparse
 import sys
 from datetime import datetime, timedelta
 from tabulate import tabulate
-from time import sleep
 
 from cfme.utils.path import log_path
 from cfme.utils.providers import list_provider_keys, get_mgmt
@@ -23,7 +22,7 @@ def parse_cmd_line():
     parser.add_argument('--exclude-enis', nargs='+',
                         help='List of ENIs, which should be excluded. ENI ID is allowed.')
     parser.add_argument('--exclude_stacks', nargs='+',
-                       help='List of Stacks, which should be exclude')
+                       help='List of Stacks, which should be excluded')
     parser.add_argument("--output", dest="output", help="target file name, default "
                                                         "'cleanup_ec2.log' in utils.path.log_path",
                         default=log_path.join('cleanup_ec2.log').strpath)
@@ -146,8 +145,6 @@ def delete_stacks(provider_mgmt, excluded_stacks, output):
                 if stack.creation_time < some_date:
                     stack_list.append([provider_name, stack.stack_name])
                     provider_mgmt.delete_stack(stack.stack_name)
-                    # TODO: Remove once we have reliable Trottling-resistant aproach
-                    sleep(5)
         logger.info("  Deleted CloudFormation Stacks: %r", stack_list)
         with open(output, 'a+') as report:
             if stack_list:
@@ -191,4 +188,4 @@ def ec2cleanup(exclude_volumes, exclude_eips, exclude_elbs, exclude_enis, exclud
 if __name__ == "__main__":
     args = parse_cmd_line()
     sys.exit(ec2cleanup(args.exclude_volumes, args.exclude_eips, args.exclude_elbs,
-                        args.exclude_stacks, args.exclude_enis, args.output))
+                        args.exclude_enis, args.exclude_stacks, args.output))
--- a/scripts/ec2cleanup.py
+++ b/scripts/ec2cleanup.py
@@ -148,7 +148,11 @@ def delete_stacks(provider_mgmt, excluded_stacks, stack_template, output):
                 some_date = today - timedelta(days=1)
                 if stack.creation_time < some_date:
                     stack_list.append([provider_name, stack.stack_name])
-                    provider_mgmt.delete_stack(stack.stack_name)
+                    try:
+                        provider_mgmt.delete_stack(stack.stack_name)
+                    except Exception as e:
+                        logger.error(e)
+                        continue
         logger.info("  Deleted CloudFormation Stacks: %r", stack_list)
         with open(output, 'a+') as report:
             if stack_list:
--- a/scripts/ec2cleanup.py
+++ b/scripts/ec2cleanup.py
@@ -23,6 +23,8 @@ def parse_cmd_line():
                         help='List of ENIs, which should be excluded. ENI ID is allowed.')
     parser.add_argument('--exclude_stacks', nargs='+',
                        help='List of Stacks, which should be excluded')
+    parser.add_argument('--stack-template',
+                        help='Stack name template to be removed', default="test", type=str)
     parser.add_argument("--output", dest="output", help="target file name, default "
                                                         "'cleanup_ec2.log' in utils.path.log_path",
                         default=log_path.join('cleanup_ec2.log').strpath)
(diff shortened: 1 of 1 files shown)


lkhomenk:
as no parametrization for names of objects existed - I wasn't adding any either.
But this is a concern - which might lead to full module refactoring

scripts/ec2cleanup.py
```diff
@@ -21,6 +22,8 @@ def parse_cmd_line():
                         help='List of ELBs, which should be excluded.')
     parser.add_argument('--exclude-enis', nargs='+',
                         help='List of ENIs, which should be excluded. ENI ID is allowed.')
+    parser.add_argument('--exclude_stacks', nargs='+',
+                       help='List of Stacks, which should be exclude')
```
izapolsk: OPTIONAL: typo

scripts/ec2cleanup.py
```diff
@@ -129,7 +132,34 @@ def delete_unused_network_interfaces(provider_mgmt, excluded_enis, output):
...
 
 
-def ec2cleanup(exclude_volumes, exclude_eips, exclude_elbs, exclude_enis, output):
+def delete_stacks(provider_mgmt, excluded_stacks, output):
+    stack_list = []
+    provider_name = provider_mgmt.kwargs['name']
+    try:
+        for stack in provider_mgmt.list_stacks():
+            if excluded_stacks and stack.stack_name in excluded_stacks:
+                logger.info("  Excluding Stack name: %r", stack.stack_name)
```
izapolsk: OPTIONAL: preceding space

scripts/ec2cleanup.py
```diff
@@ -129,7 +132,34 @@ def delete_unused_network_interfaces(provider_mgmt, excluded_enis, output):
...
+                continue
+            else:
+                today = datetime.utcnow().replace(tzinfo=None)
+                some_date = today - timedelta(days=1)
+                if stack.creation_time < some_date:
+                    stack_list.append([provider_name, stack.stack_name])
+                    provider_mgmt.delete_stack(stack.stack_name)
+                    # TODO: Remove once we have reliable Trottling-resistant aproach
```
izapolsk: Q: could you please explain what's the issue you run into ?

scripts/ec2cleanup.py
```diff
@@ -129,7 +132,34 @@ def delete_unused_network_interfaces(provider_mgmt, excluded_enis, output):
...
+                    sleep(5)
+        logger.info("  Deleted CloudFormation Stacks: %r", stack_list)
```
izapolsk: ^^

scripts/ec2cleanup.py
```diff
@@ -129,7 +132,34 @@ def delete_unused_network_interfaces(provider_mgmt, excluded_enis, output):
...
+        with open(output, 'a+') as report:
+            if stack_list:
+                report.write(tabulate(tabular_data=stack_list,
```
izapolsk: Q: are you sure this report is necessary at all ? I'd rather use logger instead.

scripts/ec2cleanup.py
psav: I think there are intentional for formatting

psav:
**REQUIRED:** It seems like it deletes ALL stacks, not just ones name ```test*``` is that right? And is that what we want?

scripts/ec2cleanup.py
lkhomenk: yes. that's it

scripts/ec2cleanup.py
lkhomenk: Purpose of this PR add stack removal. I guess, we might add GH issue to rework this module in the future to use logger

lkhomenk:
@psav yes. I also considered that and added a condition to delete at least 1 day old stacks.
@mmojzis Do we need name filtering for EC2 objects in `ec2cleanup.py `?

mmojzis:
@lkhomenk 
You mean like for excluding? I think we do. We exclude some names from deleting like cu-24x7 and so on.

lkhomenk:
@psav added stack names filtering and arg for it.

lkhomenk:
```
Deleting old stacks...
 Terminating EC2 stack testSvSAd56sTX
  Excluding Stack name: u'newstack'
  Excluding Stack name: u'ec2'
  Deleted CloudFormation Stacks: [['ec2-west', u'testSvSAd56sTX']]
```

lkhomenk:
removed `sleep` as with https://github.com/ManageIQ/wrapanapi/pull/230 we don't need it

lkhomenk:
waiting for https://github.com/ManageIQ/wrapanapi/pull/228 to be merged

lkhomenk:
@izapolsk could you please check this once again?

scripts/ec2cleanup.py
```diff
@@ -129,7 +133,35 @@ def delete_unused_network_interfaces(provider_mgmt, excluded_enis, output):
         logger.exception('Exception in %r', delete_unused_network_interfaces.__name__)
 
 
-def ec2cleanup(exclude_volumes, exclude_eips, exclude_elbs, exclude_enis, output):
+def delete_stacks(provider_mgmt, excluded_stacks, stack_template, output):
+    stack_list = []
+    provider_name = provider_mgmt.kwargs['name']
+    try:
+        for stack in provider_mgmt.list_stack():
+            if (excluded_stacks and
```
izapolsk: you don't need braces here

scripts/ec2cleanup.py
```diff
@@ -129,7 +133,35 @@ def delete_unused_network_interfaces(provider_mgmt, excluded_enis, output):
...
+                    stack.stack_name in excluded_stacks or
+                    not stack.stack_name.startswith(stack_template)):
+                logger.info("  Excluding Stack name: %r", stack.stack_name)
+                continue
+            else:
+                today = datetime.utcnow().replace(tzinfo=None)
+                some_date = today - timedelta(days=1)
+                if stack.creation_time < some_date:
```
izapolsk: are you sure creation_data doesn't have any offset ?

scripts/ec2cleanup.py
```diff
@@ -129,7 +133,35 @@ def delete_unused_network_interfaces(provider_mgmt, excluded_enis, output):
...
+                    stack_list.append([provider_name, stack.stack_name])
+                    provider_mgmt.delete_stack(stack.stack_name)
```
izapolsk: REQUIRED: if  some exception happens during stack deletion, the rest of stacks won't be removed. 
I believe you need to wrap try/catch around delete_stack operation instead of whole method.

scripts/ec2cleanup.py
lkhomenk: yes. it has 00

psav:
By default we don't have any exclusions, so this will delete everything.....right?

scripts/ec2cleanup.py
lkhomenk: it's for formatting purpose. conditions don't feet in 100 - each condition from new line for readability.

lkhomenk:
@psav it will delete all "test*" stacks. additionally we can add exclusions

scripts/ec2cleanup.py
izapolsk: ah, you're right

PR #6851, opened 2018-03-09
[1LP][RFR] New db maintenance and time/date tests

lcouzens:
Automating a few more tests, added date/time test and db maintenance tests.

### DIFF
--- a/cfme/tests/cli/test_appliance_console.py
+++ b/cfme/tests/cli/test_appliance_console.py
@@ -28,8 +28,8 @@ tzs = [
 
 
 @pytest.fixture(scope="function")
-def temp_appliance_time_crud(temp_appliance_preconfig_funcscope):
-    """Grab fresh appliance and set time and date prior to running tests"""
+def appliance_with_preset_time(temp_appliance_preconfig_funcscope):
+    """Grabs fresh appliance and sets time and date prior to running tests"""
     command_set = ('ap', '', '3', 'y', '2020-10-20', '09:58:00', 'y', '')
     temp_appliance_preconfig_funcscope.appliance_console.run_commands(command_set)
     return temp_appliance_preconfig_funcscope
@@ -82,21 +82,19 @@ def test_appliance_console_datetime(temp_appliance_preconfig_funcscope):
     app.appliance_console.run_commands(command_set)
 
     def date_changed():
-        return bool(app.ssh_client.run_command("date | grep 'Tue Oct 20 10:00'").success)
-
-    assert bool(app.ssh_client.run_command("date | grep 2020").success)
+        return app.ssh_client.run_command("date +%F-%T | grep 2020-10-20-10:00").success
     wait_for(date_changed)
 
 
-def test_appliance_console_db_maintenance_hourly(temp_appliance_time_crud):
+def test_appliance_console_db_maintenance_hourly(appliance_with_preset_time):
     """Test database hourly re-indexing through appliance console"""
-    app = temp_appliance_time_crud
+    app = appliance_with_preset_time
     command_set = ('ap', '', '7', 'y', 'n', '')
     app.appliance_console.run_commands(command_set)
 
     def maintenance_run():
-        return bool(app.ssh_client.run_command(
-            "grep REINDEX /var/www/miq/vmdb/log/hourly_continuous_pg_maint_stdout.log").success)
+        return app.ssh_client.run_command(
+            "grep REINDEX /var/www/miq/vmdb/log/hourly_continuous_pg_maint_stdout.log").success
 
     wait_for(maintenance_run, timeout=300)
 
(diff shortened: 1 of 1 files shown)


cfme/tests/cli/test_appliance_console.py
```diff
@@ -27,6 +27,14 @@
 ]
 
 
+@pytest.fixture(scope="function")
+def temp_appliance_time_crud(temp_appliance_preconfig_funcscope):
```
jamesooden: **Required:** This function name is not very descriptive I think.   Maybe instead of temp_appliance_time_crud call it preconfig_appliance_time_crud or preconfig_appliance_time.   I'm also not sure what this has to do with crud in and of itself.   Not saying my names are the best, but the function name should very clearly state what the function is doing on some level.

cfme/tests/cli/test_appliance_console.py
```diff
@@ -67,6 +75,54 @@ def test_appliance_console_set_timezone(timezone, temp_appliance_preconfig_modsc
...
 
 
+def test_appliance_console_datetime(temp_appliance_preconfig_funcscope):
+    """Grab fresh appliance and set time and date through appliance_console and check result"""
+    app = temp_appliance_preconfig_funcscope
+    command_set = ('ap', '', '3', 'y', '2020-10-20', '09:58:00', 'y', '')
+    app.appliance_console.run_commands(command_set)
+
+    def date_changed():
+        return bool(app.ssh_client.run_command("date | grep 'Tue Oct 20 10:00'").success)
```
jamesooden: Consider providing a specific date format to date so that you are not relying on its default behavior which could change.

cfme/tests/cli/test_appliance_console.py
```diff
@@ -67,6 +75,54 @@ def test_appliance_console_set_timezone(timezone, temp_appliance_preconfig_modsc
...
+
+    assert bool(app.ssh_client.run_command("date | grep 2020").success)
```
jamesooden: Same as 85.

cfme/tests/cli/test_appliance_console.py
jamesooden: Oh, I'm beginning to understand what this function is all about.   You are returning a temp appliance with its time pre-configured.   Hard to find a good name for that in my head.   crud throws me because it makes me think you are going through all those operations.   Perhaps temp_appliance_with_base_time is better?

cfme/tests/cli/test_appliance_console.py
```diff
@@ -67,6 +75,54 @@ def test_appliance_console_set_timezone(timezone, temp_appliance_preconfig_modsc
...
+], ids=['hour', 'day', 'week', 'month'])
+def test_appliance_console_db_maintenance_periodic(period, temp_appliance_time_crud):
+    """Tests full vacuums on database through appliance console"""
+    app = temp_appliance_time_crud
+    command_set = ['ap', '', '7', 'n', 'y']
+    command_set.extend(period)
+    app.appliance_console.run_commands(command_set)
+
+    def maintenance_run():
+        return bool(app.ssh_client.run_command(
```
jamesooden: I don't think you need to wrap that in bool, since success returns a boolean value:

    @property
    def success(self):
        return self.rc == 0

cfme/tests/cli/test_appliance_console.py
```diff
@@ -67,6 +75,54 @@ def test_appliance_console_set_timezone(timezone, temp_appliance_preconfig_modsc
...
+
+
+def test_appliance_console_db_maintenance_hourly(temp_appliance_time_crud):
+    """Test database hourly re-indexing through appliance console"""
+    app = temp_appliance_time_crud
+    command_set = ('ap', '', '7', 'y', 'n', '')
+    app.appliance_console.run_commands(command_set)
+
+    def maintenance_run():
+        return bool(app.ssh_client.run_command(
```
jamesooden: I don't think you need to wrap that in bool, since success returns a boolean value:

@property
def success(self):
    return self.rc == 0

jamesooden:
All the changes look good, but PRT is failing on your test, test_appliance_console_datetime:

    cfme/tests/cli/test_appliance_console.py::test_appliance_console_datetime FAILED

Here is the link:

    http://10.16.4.32/trackerbot/pr/run/22030

cfme/tests/cli/test_appliance_console.py
```diff
@@ -67,6 +75,52 @@ def test_appliance_console_set_timezone(timezone, temp_appliance_preconfig_modsc
...
+
+def test_appliance_console_db_maintenance_hourly(appliance_with_preset_time):
+    """Test database hourly re-indexing through appliance console"""
+    app = appliance_with_preset_time
+    command_set = ('ap', '', '7', 'y', 'n', '')
+    app.appliance_console.run_commands(command_set)
+
+    def maintenance_run():
+        return app.ssh_client.run_command(
+            "grep REINDEX /var/www/miq/vmdb/log/hourly_continuous_pg_maint_stdout.log").success
```
quarckster: **OPTIONAL**
"grep -q REINDEX /var/www/miq/vmdb/log/hourly_continuous_pg_maint_stdout.log"

cfme/tests/cli/test_appliance_console.py
```diff
@@ -67,6 +75,52 @@ def test_appliance_console_set_timezone(timezone, temp_appliance_preconfig_modsc
...
 
+def test_appliance_console_datetime(temp_appliance_preconfig_funcscope):
+    """Grab fresh appliance and set time and date through appliance_console and check result"""
+    app = temp_appliance_preconfig_funcscope
+    command_set = ('ap', '', '3', 'y', '2020-10-20', '09:58:00', 'y', '')
+    app.appliance_console.run_commands(command_set)
+
+    def date_changed():
+        return app.ssh_client.run_command("date +%F-%T | grep 2020-10-20-10:00").success
+    wait_for(date_changed)
```
quarckster: **OPTIONAL**
```python
@wait_for_decorator
def date_changed():
    return app.ssh_client.run_command("date +%F-%T | grep 2020-10-20-10:00").success
```

PR #6865, opened 2018-03-12
[1LP][RFR] Fix publish_to_template; add ownership test

apagac:
Two things happening here:
1. Fix for publish_to_template method in cfme.infrastructure.virtual_machines.Vm
2. Add template ownership test which also tests the above fix

For the first point, publish_to_template navigated to a wrong view, resulting in error when trying to fill a from that wasn't in that view. So instead of navigating to 'Details', I created a new view 'PublishVmView' in cfme.common.vm_views. This view has a form from 'BasicProvisionFormView' that was already implemented and contains all the things publish_to_template needs.
I then created new navigation step 'Publish' to enable navigation to my newly created view and added that to publish_to_template method.

For the second point, I wrote a very simple ownership test that just tries to set an ownership to a template. This automates BZ 1446801. It uses the publish_to_template method to create a template from a VM.

As a bonus, there's a tiny one-line fix for cfme.infrastructure.virtual_machines.Template making the assertion for flash messages correctly work when working with templates.

Currently limited to a vmware type provider, need to test and see if I can add more providers.

{{ pytest: -v --long-running --use-provider vsphere55  cfme/tests/cloud_infra_common/test_vm_ownership.py -k test_template_set_ownership }}

### DIFF
--- a/cfme/common/vm_views.py
+++ b/cfme/common/vm_views.py
@@ -439,6 +439,19 @@ class MigrateVmView(BaseLoggedInPage):
         return False
 
 
+class PublishVmView(BaseLoggedInPage):
+    title = Text('#explorer_title_text')
+
+    @View.nested
+    class form(BasicProvisionFormView):  # noqa
+        submit_button = Button('Submit')
+        cancel_button = Button('Cancel')
+
+    @property
+    def is_displayed(self):
+        return False
+
+
 class RetirementView(BaseLoggedInPage):
     """
     Set Retirement date view for vms/instances
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -17,8 +17,9 @@ from widgetastic_patternfly import (
 from cfme.base.login import BaseLoggedInPage
 from cfme.common.vm import VM, Template as BaseTemplate
 from cfme.common.vm_views import (
-    ManagementEngineView, CloneVmView, MigrateVmView, ProvisionView, EditView, RetirementView,
-    RetirementViewWithOffset, VMDetailsEntities, VMToolbar, VMEntities, SetOwnershipView)
+    ManagementEngineView, CloneVmView, MigrateVmView, ProvisionView, EditView, PublishVmView,
+    RetirementView, RetirementViewWithOffset, VMDetailsEntities, VMToolbar, VMEntities,
+    SetOwnershipView)
 from cfme.exceptions import (
     VmNotFound, OptionNotAvailable, DestinationNotFound, ItemNotFound,
     VmOrInstanceNotFound)
@@ -749,8 +750,7 @@ class Vm(VM):
         view.form.fill_with(provisioning_data, on_change=view.form.submit_button)
 
     def publish_to_template(self, template_name, email=None, first_name=None, last_name=None):
-        view = navigate_to(self, 'Details', use_resetter=False)
-        view.toolbar.lifecycle.item_select("Publish this VM to a Template")
+        view = navigate_to(self, 'Publish')
         first_name = first_name or fauxfactory.gen_alphanumeric()
         last_name = last_name or fauxfactory.gen_alphanumeric()
         email = email or "{}@{}.test".format(first_name, last_name)
@@ -948,6 +948,7 @@ class Vm(VM):
 
 class Template(BaseTemplate):
     REMOVE_MULTI = "Remove Templates from the VMDB"
+    VM_TYPE = "Template"
 
     @property
     def genealogy(self):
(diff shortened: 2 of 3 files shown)


dajoRH:
Lint report for commit 5e347a6a0fc5e3d8ec436e2e7d8688acaff65af5:

`cfme/common/vm_views.py`:
- :red_circle: Line 440:11: **N801** *class names should use CapWords convention*

`cfme/tests/cloud_infra_common/test_vm_ownership.py`:
- :red_circle: Line 190:63: **F821** *undefined name 'VMwareProvider'*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
Lint report for commit ddfef2ba879c4d243920af673205c663a2d1c8be:

`cfme/common/vm_views.py`:
- :red_circle: Line 440:11: **N801** *class names should use CapWords convention*

**No commit flaws detected**.

*CFME QE Bot*

apagac:
I can't do that, lint bot. Class 'form' in views has to be lower-case, at least I believe so at this point.

cfme/common/vm_views.py
```diff
@@ -433,6 +433,19 @@ def is_displayed(self):
         return False
 
 
+class PublishVmView(BaseLoggedInPage):
+    title = Text('#explorer_title_text')
+
+    @View.nested
+    class form(BasicProvisionFormView):
```
rlbabyuk: Required: Add # noqa in the end of the line to fix lint error

cfme/common/vm_views.py
```diff
@@ -433,6 +433,19 @@ def is_displayed(self):
...
+        submit_button = Button('Submit')
+        cancel_button = Button('Cancel')
+
+    @property
+    def is_displayed(self):
+        return False
```
rlbabyuk: Question: Any check here?

cfme/infrastructure/virtual_machines.py
```diff
@@ -1360,6 +1361,15 @@ def step(self, *args, **kwargs):
         self.prerequisite_view.toolbar.lifecycle.item_select("Migrate this VM")
 
 
+@navigator.register(Vm, 'Publish')
+class VmPublish(CFMENavigateStep):
+    VIEW = PublishVmView
+    prerequisite = NavigateToSibling('Details')
+
+    def step(self, *args, **kwargs):
```
rlbabyuk: Required: we don't need *args, **kwargs here, as we are not using them further in the step

cfme/tests/cloud_infra_common/test_vm_ownership.py
```diff
@@ -182,6 +183,20 @@ def test_group_ownership_on_user_or_group_role(
         assert not check_vm_exists(vm_crud), "vm exists! but shouldn't exist"
 
 
+# Sets ownership to an infra template.
```
rlbabyuk: Required: Move those comments to test docs as test description

cfme/common/vm_views.py
apagac: I found that without is_displayed returning False the navigation skips some steps and the test fails. I proceeded in a manner you can see a few lines above in MigrateVmView and CloneVmView.
Apart from the title, these views probably can't be checked by is_displayed. They are essentially just filtered options from BasicProvisionFormView. No identification of the view, apart from the title, can be checked to determine that the view is displayed.

cfme/tests/cloud_infra_common/test_vm_ownership.py
```diff
@@ -182,6 +183,22 @@ def test_group_ownership_on_user_or_group_role(
         assert not check_vm_exists(vm_crud), "vm exists! but shouldn't exist"
 
 
+@pytest.mark.uncollectif(lambda provider: not provider.one_of(VMwareProvider),
+                         'Not VMware provider')
```
quarckster: **REQUIRED**
```python
@pytest.mark.provider([VMwareProvider])
```

apagac: So this is a tricky one.

Every test in this module is currently marked by pytest.mark.provider([CloudInfraProvider], scope='module'), so I can't just mark this one with VMwareProvider, that doesn't work.

What I would have to do is mark each test individually with CloudInfraProvider except this one, which I'll mark with VMwareProvider, but every mark will have to be scope='module', because the fixtures that are used in the test are marked for module. So if I wanted to mark.provider with scope='function', it would produce a scope mismatch error. This seems to me a bit weird, marking each test inside a module with marker with scope='module'.

Another solution I can think of is just leave the uncollectif there, which solves all above issues.

What do you think?

cfme/tests/cloud_infra_common/test_vm_ownership.py
```diff
@@ -185,6 +186,21 @@ def test_group_ownership_on_user_or_group_role(
         assert not check_vm_exists(vm_crud), "vm exists! but shouldn't exist"
 
 
+@pytest.mark.provider([VMwareProvider], override=True, scope="module")
```
psav: :+1:

PR #6883, opened 2018-03-15
[1LP][RFR] Refactor Volume tests

digitronik:
Purpose or Intent
=================
- Storage pages improved now. It's time to add test cases
- `test_volume_navigation` test no more needed. All view going to use in different upcoming tests.
- `test_volume_collective_crud` test was not good approach [ Initial time it was good means test with model page] as it's unable to delete volume having status `in-use` better to go with `test_storage_volume_crud` now.
- changes added for future purpose as I am trying to automate all tests related to storage.

{{pytest: cfme/tests/storage/test_volume.py -v}}

### DIFF
--- a/cfme/storage/volume.py
+++ b/cfme/storage/volume.py
@@ -306,7 +306,7 @@ class VolumeCollection(BaseCollection):
     """Collection object for the :py:class:'cfme.storage.volume.Volume'. """
     ENTITY = Volume
 
-    def create(self, name, storage_manager, tenant, size, provider):
+    def create(self, name, storage_manager, tenant, provider, size=1, cancel=False):
         """Create new storage volume
 
         Args:
@@ -315,24 +315,29 @@ class VolumeCollection(BaseCollection):
             tenant: tenant name
             size: volume size in GB
             provider: provider
+            cancel: bool
 
         Returns:
             object for the :py:class: cfme.storage.volume.Volume
         """
 
         view = navigate_to(self, 'Add')
-        view.form.fill({'storage_manager': storage_manager,
-                        'tenant': tenant,
-                        'volume_name': name,
-                        'size': size})
-        view.form.add.click()
-        base_message = 'Cloud Volume "{}" created'
-        view.flash.assert_success_message(base_message.format(name))
 
-        volume = self.instantiate(name, provider)
-        wait_for(lambda: volume.exists, delay=50, timeout=1500, fail_func=volume.refresh)
+        if not cancel:
+            view.form.fill({'storage_manager': storage_manager,
+                            'tenant': tenant,
+                            'volume_name': name,
+                            'size': size})
+            view.form.add.click()
+            base_message = 'Cloud Volume "{}" created'
+            view.flash.assert_success_message(base_message.format(name))
 
-        return volume
+            volume = self.instantiate(name, provider)
+            wait_for(lambda: volume.exists, delay=50, timeout=1500, fail_func=volume.refresh)
+            return volume
+        else:
+            view.form.cancel.click()
+            return None
 
     def delete(self, *volumes):
         """Delete one or more Volumes from list of Volumes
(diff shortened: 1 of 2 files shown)


cfme/tests/storage/test_volume.py
```diff
@@ -1,54 +1,68 @@
...
+                             size=STORAGE_SIZE,
+                             provider=provider,
+                             cancel=True)
+
+    view = volume_collection.create_view(VolumeAllView)
+    view.flash.assert_message('Add of new Cloud Volume was cancelled by the user')
+
+
+@pytest.mark.tier(1)
+def test_storage_volume_crud(appliance, provider):
```
jamesooden: **QUESTION:** Since this is a a CRUD test, shouldn't we have an UPDATE component (i.e. we change an existing entity, perhaps the one we created)?

digitronik: @jamesooden right; we have `edit` method for volume (yes it support only change in name) but nice suggestion going to push

cfme/storage/volume.py
```diff
@@ -272,24 +272,29 @@ def create(self, name, storage_manager, tenant, size, provider):
...
+            view.form.add.click()
+            base_message = 'Cloud Volume "{}" created'
+            view.flash.assert_success_message(base_message.format(name))
+
+            volume = self.instantiate(name, provider)
+            wait_for(lambda: volume.exists, delay=20, timeout=500, fail_func=volume.refresh)
+            return volume
+        else:
+            view.form.cancel.click()
+            return None
```
quarckster: `return None` is not mandatory.

digitronik: @quarckster right not mandatory but as we are returning `volume` object for `add` so better to return something for cancel well.

cfme/tests/storage/test_volume.py
```diff
@@ -1,54 +1,72 @@
...
+    volume = volume_collection.create(name=fauxfactory.gen_alpha(),
+                                      storage_manager=manager_name,
+                                      tenant=provider.data['provisioning']['cloud_tenant'],
+                                      size=STORAGE_SIZE,
+                                      provider=provider)
+    assert volume.exists
+
+    # update volume
+    new_volume_name = fauxfactory.gen_alpha()
+    volume.edit(new_volume_name)
```
quarckster: **REQUIRED**
Please use a context manager here:
```python
with update(volume):
    volume.name = new_volume_name
```
`Volume` should inherit `Updateable` mixin and implement `update` method, not `edit`.

digitronik: @quarckster added update method

quarckster:
@digitronik please make `Volume` class Updateable

dajoRH:
*CFME QE Bot*

digitronik:
@quarckster : volume class now `Updatable`; rebase also done.

cfme/storage/volume.py
```diff
@@ -315,24 +315,29 @@ def create(self, name, storage_manager, tenant, size, provider):
...
+            view.form.add.click()
+            base_message = 'Cloud Volume "{}" created'
+            view.flash.assert_success_message(base_message.format(name))
+
+            volume = self.instantiate(name, provider)
+            wait_for(lambda: volume.exists, delay=50, timeout=1500, fail_func=volume.refresh)
+            return volume
+        else:
+            view.form.cancel.click()
+            return None
```
quarckster: **REQUIRED**
`return None` is not needed

cfme/tests/storage/test_volume.py
```diff
@@ -1,54 +1,76 @@
...
+                                      storage_manager=manager_name,
+                                      tenant=provider.data['provisioning']['cloud_tenant'],
+                                      size=STORAGE_SIZE,
+                                      provider=provider)
+    assert volume.exists
+
+    # update volume
+    old_name = volume.name
+    with update(volume):
+        volume.name = fauxfactory.gen_alpha()
```
quarckster: **REQUIRED**
Please add an assertion, we should be assured that the value has changed in the UI.

digitronik: As per IRC discussion no need of it

PR #6889, opened 2018-03-15
[1LP][RFR] New Provider: KubeVirt

vatsalparekh:
Purpose or Intent
=================
Adding the new KubeVirt Provider 
{{ pytest: -vvvv cfme/tests/infrastructure/test_kubevirt.py --use-provider kubevirt }}

### DIFF
--- a/cfme/common/provider_views.py
+++ b/cfme/common/provider_views.py
@@ -523,10 +523,11 @@ class ContainerProviderAddViewUpdated(ContainerProviderAddView, ContainerProvide
     """
      Additional widgets for builds 5.9 and up
     """
-    COND_WIDGETS = ['prov_type', 'metrics_type', 'alerts_type']
+    COND_WIDGETS = ['prov_type', 'metrics_type', 'alerts_type', 'virt_type']
 
     metrics_type = BootstrapSelect(id='metrics_selection')
     alerts_type = BootstrapSelect(id='alerts_selection')
+    virt_type = BootstrapSelect(id='virtualization_selection')
 
     def before_fill(self, values):
         for widget in self.COND_WIDGETS:
@@ -620,6 +621,7 @@ class ContainerProviderEditViewUpdated(ContainerProviderEditView, ContainerProvi
 
     metrics_type = BootstrapSelect(id='metrics_selection')
     alerts_type = BootstrapSelect(id='alerts_selection')
+    virt_type = BootstrapSelect(id='virtualization_selection')
 
     def before_fill(self, values):
         for widget in self.COND_WIDGETS:
--- a/cfme/containers/provider/__init__.py
+++ b/cfme/containers/provider/__init__.py
@@ -70,6 +70,12 @@ class ContainersProviderEndpointsForm(View):
         api_port = Input('default_api_port')
 
     @View.nested
+    class virtualization(Tab, BeforeFillMixin):  # NOQA
+        TAB_NAME = 'Virtualization'
+        kubevirt_token = Input('kubevirt_password')
+        validate = Button('Validate')
+
+    @View.nested
     class metrics(Tab, BeforeFillMixin):  # NOQA
         TAB_NAME = VersionPick({
             Version.lowest(): 'Hawkular',
--- a/cfme/containers/provider/openshift.py
+++ b/cfme/containers/provider/openshift.py
@@ -54,6 +54,17 @@ class ServiceBasedEndpoint(DefaultEndpoint):
         return out
 
 
+class VirtualizationEndpoint(ServiceBasedEndpoint):
+    """Represents virtualization Endpoint"""
+    name = 'virtualization'
+
+    @property
+    def view_value_mapping(self):
+        # values like host, port are taken from Default endpoint
+        # and not editable in Virtualization endpoint, only token can be added
+        return {'kubevirt_token': self.token}
+
+
 class MetricsEndpoint(ServiceBasedEndpoint):
     """Represents metrics Endpoint"""
     name = 'metrics'
@@ -90,7 +101,8 @@ class OpenshiftProvider(ContainersProvider):
             image_repo=None,
             image_reg=None,
             image_tag=None,
-            cve_loc=None):
+            cve_loc=None,
+            virt_type=None):
 
         self.http_proxy = http_proxy
         self.adv_http = adv_http
@@ -100,6 +112,7 @@ class OpenshiftProvider(ContainersProvider):
         self.image_reg = image_reg
         self.image_tag = image_tag
         self.cve_loc = cve_loc
+        self.virt_type = virt_type
 
         super(OpenshiftProvider, self).__init__(
             name=name,
@@ -169,6 +182,9 @@ class OpenshiftProvider(ContainersProvider):
             mapping['alerts_type'] = None
             mapping['proxy'] = None
             mapping['advanced'] = None
+
+        if self.virt_type and not self.appliance.is_downstream:
+            mapping['virt_type'] = self.virt_type
 
         return mapping
 
@@ -240,7 +256,8 @@ class OpenshiftProvider(ContainersProvider):
             image_repo=image_repo,
             image_reg=image_reg,
             image_tag=image_tag,
-            cve_loc=cve_loc
+            cve_loc=cve_loc,
+            virt_type=prov_config.get('virt_type')
         )
 
     def custom_attributes(self):
--- a/cfme/infrastructure/provider/kubevirt.py
+++ b/cfme/infrastructure/provider/kubevirt.py
@@ -17,9 +17,10 @@ class KubeVirtProvider(InfraProvider):
             appliance=kwargs.get('appliance'))
 
     def create(self, *args, **kwargs):
-        # Override the standard behaviour to actually create the parent container provider first.
-        created = self.parent_provider.create()
-        return created
+        # KubeVirt infra provider is automatically added in appliance when adding a container provider,
+        # with selecting the virtualization option
+        # so Override the standard behaviour to actually create the parent container provider first
+        return self.parent_provider.create()
 
     @classmethod
     def from_config(cls, prov_config, prov_key, appliance=None):
--- a/cfme/infrastructure/provider/kubevirt.py
+++ b/cfme/infrastructure/provider/kubevirt.py
@@ -17,8 +17,8 @@ class KubeVirtProvider(InfraProvider):
             appliance=kwargs.get('appliance'))
 
     def create(self, *args, **kwargs):
-        # KubeVirt infra provider is automatically added in appliance when adding a container provider,
-        # with selecting the virtualization option
+        # KubeVirt infra provider is automatically added in appliance when adding
+        # a container provider, with selecting the virtualization option
         # so Override the standard behaviour to actually create the parent container provider first
         return self.parent_provider.create()
 
--- a/cfme/infrastructure/provider/kubevirt.py
+++ b/cfme/infrastructure/provider/kubevirt.py
@@ -1,0 +1,47 @@
+from cfme.utils.providers import get_crud
+from cfme.containers.provider.openshift import VirtualizationEndpoint
+from . import InfraProvider
+
+
+class KubeVirtProvider(InfraProvider):
+
+    def __init__(self, parent_provider, **kwargs):
+
+        self.parent_provider = parent_provider
+
+        super(KubeVirtProvider, self).__init__(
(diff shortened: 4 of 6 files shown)


dajoRH:
*CFME QE Bot*

vatsalparekh:
Conf and Creds update PR for this: [#472](https://gitlab.cee.redhat.com/cfme-qe/cfme-qe-yamls/merge_requests/472)

cfme/containers/provider/__init__.py
```diff
@@ -69,6 +69,11 @@ class default(Tab, DefaultEndpointForm, BeforeFillMixin):  # NOQA
         trusted_ca_certificates = TextInput('default_tls_ca_certs')
         api_port = Input('default_api_port')
 
+    @View.nested
+    class Virtualization(default):
```
mfalesni: **REQUIRED**: Use `snake_case` for nested views together with `# NOQA` so linter does not complain.

vatsalparekh: changed.

mfalesni:
@vatsalparekh I also noticed you exposed an internal URL in your comment, so I removed it.

dajoRH:
Lint report for commit 1abd437600143963a78fd94d0f117d5d31da4c2b:

`cfme/infrastructure/provider/kubevirt.py`:
- :red_circle: Line 2:1: **F401** *'cfme.containers.provider.openshift.OpenshiftDefaultEndpoint' imported but unused*
- :bangbang: Line 8:101: **E501** *line too long (118 > 100 characters)*

**No commit flaws detected**.

*CFME QE Bot*

mshriver:
This needs to be merged before the CFME QE team's yaml is updated with any kubevirt providers so that test collection doesn't break on the new type.   cfme-qe-yamls MR 477

cfme/common/provider_views.py
```diff
@@ -523,10 +523,11 @@ class ContainerProviderAddViewUpdated(ContainerProviderAddView, ContainerProvide
     """
      Additional widgets for builds 5.9 and up
     """
-    COND_WIDGETS = ['prov_type', 'metrics_type', 'alerts_type']
+    COND_WIDGETS = ['prov_type', 'metrics_type', 'alerts_type', 'virtualization']
 
     metrics_type = BootstrapSelect(id='metrics_selection')
     alerts_type = BootstrapSelect(id='alerts_selection')
+    virtualization = BootstrapSelect(id='virtualization_selection')
```
mshriver: **Required**

Consistent naming with the other endpoint type selections, let's call this `virt_type` or `virtualization_type`.

Update the maps in the yaml accordingly

cfme/containers/provider/openshift.py
```diff
@@ -54,6 +54,15 @@ def view_value_mapping(self):
         return out
 
 
+class VirtualizationEndpoint(ServiceBasedEndpoint):
+    """Represents virtualization Endpoint"""
+    name = 'virtualization'
+
+    @property
+    def view_value_mapping(self):
+        return {'kubevirt_token': self.token}
```
mshriver: Please put a comment in-line explaining how this endpoint takes its hostname + port + sec_protocol from the default endpoint and only token can be entered.

vatsalparekh: added.

cfme/containers/provider/openshift.py
```diff
@@ -170,6 +181,9 @@ def view_value_mapping(self):
             mapping['proxy'] = None
             mapping['advanced'] = None
 
+        if self.virtualization:
```
mshriver: This needs to be version picked into the code blocks above, I'm thinking:

```
if not self.appliance.is_downstream:  # no is_upstream method
    mapping['virtualization'] = self.virtualization
elif self.appliance.version>= '5.9':
    <existing block of mapping for 5.9>
else:
    <existing block of mapping to None>
    mapping['virtualization'] = self.virtualization

File: cfme/containers/provider/openshift.py [hunk 5]
```diff
@@ -240,7 +254,8 @@ def from_config(cls, prov_config, prov_key, appliance=None):
             image_repo=image_repo,
             image_reg=image_reg,
             image_tag=image_tag,
-            cve_loc=cve_loc
+            cve_loc=cve_loc,
+            virtualization=prov_config.get('virtualization', 'Disabled')
```
mshriver: I think this should default to None like metrics_type and alerts_type above, that way its just ignored/not filled when its not in the YAML.

cfme/infrastructure/provider/kubevirt.py
```diff
@@ -0,0 +1,45 @@
...
+        super(KubeVirtProvider, self).__init__(
+            name=name,
+            endpoints=endpoints,
+            key=key,
+            provider_data=provider_data,
+            appliance=appliance)
+
+    def create(self, *args, **kwargs):
+        # Override the standard behaviour to actually create the parent container provider first.
+        created = self.parent_provider.create()
```
mshriver: This blows up with AttributeError when parent_provider is `None`

cfme/infrastructure/provider/kubevirt.py
```diff
@@ -0,0 +1,45 @@
+from cfme.utils.providers import get_crud
+from cfme.containers.provider.openshift import VirtualizationEndpoint
+from . import InfraProvider
+
+
+class KubeVirtProvider(InfraProvider):
+
+    def __init__(self, name=None, endpoints=None, key=None, provider_data=None,
+                 parent_provider=None, appliance=None):
```
mshriver: parent_provider isn't optional is it? The only way to get one of these providers in CFME is through adding a parent provider.

mshriver: Only specify parent_provider here, and use `**kwargs` to capture the rest of the args and pass them to `InfraProvider.__init__` so that the parent class's kwargs can change without breaking this.

cfme/infrastructure/provider/kubevirt.py
mshriver: Pass `**kwargs` on to the create() method so that the `BaseProvider.create()` kwargs can be passed through this `cancel=False, validate_credentials=True, check_existing=False, validate_inventory=False`

cfme/containers/provider/openshift.py
vatsalparekh: I don't think its needed, KubeVirt is upstream only for now
Added just `is_downstream` check with `if self.virtualization`

cfme/infrastructure/provider/kubevirt.py
```diff
@@ -0,0 +1,45 @@
...
+    def from_config(cls, prov_config, prov_key, appliance=None):
+
+        endpoints = {}
+        token_creds = cls.process_credential_yaml_key(prov_config['credentials'], cred_type='token')
+        for endp in prov_config['endpoints']:
+            if VirtualizationEndpoint.name == endp:
+                prov_config['endpoints'][endp]['token'] = token_creds.token
+                endpoints[endp] = VirtualizationEndpoint(**prov_config['endpoints'][endp])
+
+        parent_provider = get_crud('ocp-kubevirt')
```
mshriver: absolutely cannot hardcode this provider name here.

cfme/containers/provider/openshift.py
vatsalparekh: Changed.

cfme/infrastructure/provider/kubevirt.py
mshriver: Your yaml for this provider will need to include a map of the parent provider key, which you'll call `get_crud` against here, and then pass as `parent_provider` to the constructor.

cfme/infrastructure/provider/kubevirt.py
vatsalparekh: I feel `AttributeError` would be okay here as parent_provider must be there, or can just handle is with parent_provider check and raise an exception

cfme/infrastructure/provider/kubevirt.py
vatsalparekh: suggestion for more robust way?

cfme/infrastructure/provider/kubevirt.py
mshriver: If you made the parent_provider a required kwarg, then its a non-issue :)

cfme/infrastructure/provider/kubevirt.py
mshriver: **Required**

parent_provider is a required kwarg.

cfme/infrastructure/provider/kubevirt.py
mshriver: yaml must have a parent_provider map, raise exception if it doesn't.  Use that map ( parent_provider: "ocp-kubevirt" ) to call `providers.get_crud(prov_config['parent_provider'])`

cfme/containers/provider/openshift.py
mshriver: So here's the use-case that I'm thinking about when I consider changes like this - The yaml restricts where the provider is collected via the 'since_version', which means if you're running tests the provider won't be collected when you're testing downstream, and there's no problem.

Outside of running tests, the yaml 'since_version' field doesn't block me from getting a provider crud object and trying to configure it against my appliance, which could be any version.  If I get the crud object for an openshift provider, the blocks above make sure that when I try to fill in the UI form for this provider, only the configuration options relevant for that appliance version are returned. This means that a `provider.create()` call won't attempt to fill widgets that don't exist, and will only fill the form values available for that provider type on that appliance version. The provider may still fail to configure correctly, but it will fail because of CFME validation or endpoint incompatibility - not with an AttributeError or NoSuchElementException type of error.

A separate `if` statement is just fine here, my suggestion was just that it may read better by incorporating it into the existing `if/else` structure.

cfme/infrastructure/provider/kubevirt.py
vatsalparekh: yup made it a required kwarg

cfme/infrastructure/provider/kubevirt.py
vatsalparekh: corrected.

cfme/infrastructure/provider/kubevirt.py
```diff
@@ -0,0 +1,47 @@
...
+            name=kwargs.get('name'),
+            endpoints=kwargs.get('endpoints'),
+            key=kwargs.get('key'),
+            provider_data=kwargs.get('provider_data'),
+            appliance=kwargs.get('appliance'))
+
+    def create(self, *args, **kwargs):
+        # Override the standard behaviour to actually create the parent container provider first.
+        created = self.parent_provider.create()
+        return created
```
mfalesni: **QUESTION**: If you say "create parent provider first" then where the actual provider gets created? Isn't there a super call missing? By the way you don't need an intermediate variable in here if this is all it gets.

vatsalparekh: @mfalesni the `KubeVirt(InfraProvider)` gets automatically created on the CF side when we add the ocp with `KubeVirt` virtualization

mshriver: @vatsalparekh I think a more descriptive comment in the code avoid potential confusion here.  Instead of saying create the provider first - state the KubeVirt provider is added automatically when creating the parent Openshift Provider, and no further action is needed to see this provider.

cfme/tests/infrastructure/test_kubevirt.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+import fauxfactory
+import pytest
+
+from cfme.infrastructure.provider.kubevirt import KubeVirtProvider
+from cfme.utils.update import update
+
+
+pytestmark = [
+    pytest.mark.provider([KubeVirtProvider], scope="module")
```
izapolsk: as far as I understand this provider is available only in upstream. Where is restriction that this test has to be run only against upstream ? in yaml ?

dajoRH:
Lint report for commit f5ae66d89896c4124a03fab66baa03218b50d75c:

`cfme/infrastructure/provider/kubevirt.py`:
- :bangbang: Line 20:101: **E501** *line too long (103 > 100 characters)*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/infrastructure/test_kubevirt.py
vatsalparekh: yes @izapolsk , yaml has it

mshriver:
PRT green because no downstream tests are run.

PR #6890, opened 2018-03-15
[WIPTEST] Added new policy types Provider and Physiacl Infrastructure

quarckster:
{{pytest: -v cfme/tests/control/test_basic --long-running}}


cfme/tests/control/test_basic.py
```diff
@@ -463,13 +467,35 @@ def test_control_assign_actions_to_event(request, policy, action):
...
+                         ids=lambda item: item.name)
+def test_assign_condition_to_control_policy(request, policy_and_condition, condition_collection,
+                                            policy_collection):
     """This test checks if a condition is assigned to a control policy.
     Steps:
         * Create a control policy.
         * Assign a condition to the created policy.
     """
-    policy, condition = policy_and_condition
+    condition_class = policy_and_condition.condition
```
psav: **REQUIRED:** The policy and condition here shouldn't be created inside the test. These are fixtures I believe and should be created as such outside. I believe we already have tests that test creating conditions and policies.

quarckster: Agree, but there is a problem with uncollecting. Here is the problematic place https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/control/test_basic.py#L368. I can't create a Physical Infrastructure policy in CFME 5.8.

psav: I think I would prefer to do it in this way, put those into a fixture and skip if we get those values, now, I KNOW, that we shouldn't do that, in that a skip is for a transient failure due to some other factor, but I think having the test body represent what it _should_ actually test, is more important right now.

dajoRH:
*CFME QE Bot*

quarckster:
PRT failures are not related to the changes.

cfme/tests/control/test_basic.py
```diff
@@ -256,13 +266,18 @@ def alert_profile_collection(appliance):
 
 
 @pytest.yield_fixture
-def two_random_policies(policy_collection):
+def two_random_policies(appliance, policy_collection):
+    # Physical Infrastucture policies excluded
+    if appliance.version < "5.9":
+        policies = COMPLIANCE_POLICIES[:-1] + CONTROL_POLICIES[:-1]
```
mshriver: **Required**

Make this remove the correct policy dynamically. If another policy is added to either COMPLIANCE_POLICIES or CONTROL_POLICIES this will stop working correctly.

quarckster: @mshriver I'm pretty sure no new policies will be add :)

PR #6891, opened 2018-03-15
[RFR] Remove is_refreshed method from OpenShiftProvider class

jawatts:
validate_stats is to specific of a check for checking if the provider is refreshed. This is blocking adding OCP providers.  This is undoing https://github.com/ManageIQ/integration_tests/pull/6737
{{ pytest: cfme/tests/containers/test_static_custom_attributes.py --use-provider ocp-v1 --long-running }}


jawatts:
PRT was a false positive, it failed to add the provider. I was able to run it locally fine, trying it again.

PR #6897, opened 2018-03-16
[1LP][RFR] Mitigate the bug in CFME where symbols are converted to strings by REST

psav:
* Symbols are converted to strings in REST API leading to broken config
  for OpenShift.
* Previously we would get_yaml_config, modify it and then push the
  entire config back. The getting of the yaml via the REST is what
  breaks the strings.
* We now only do full patching for 5.8. This will need to be merged
  alongside the other PR for yaml updates

### DIFF
--- a/cfme/utils/appliance/__init__.py
+++ b/cfme/utils/appliance/__init__.py
@@ -4,6 +4,7 @@ import socket
 import traceback
 from copy import copy
 from datetime import datetime
+from tempfile import NamedTemporaryFile
 from textwrap import dedent
 from time import sleep, time
 
@@ -25,6 +26,7 @@ from werkzeug.local import LocalStack, LocalProxy
 
 from cfme.utils import clear_property_cache
 from cfme.utils import conf, ssh, ports
+from cfme.utils.datafile import load_data_file
 from cfme.utils.log import logger, create_sublogger, logger_wrap
 from cfme.utils.net import net_check
 from cfme.utils.path import data_path, patches_path, scripts_path, conf_path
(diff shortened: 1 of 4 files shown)


mfalesni:
I wonder why ruby went with mutable strings and immutable symbols. Just makes things complicated :)

psav:
Needs another review, had to add the old set_yaml back in for 5.8

cfme/utils/appliance/__init__.py
```diff
@@ -2063,9 +2066,32 @@ def set_yaml_config(self, data_dict):
             ApplianceException when server_id isn't set
         """
         # Can only modify through server ID, raise if that's not set yet
-        if self.server_id() is None:
-            raise ApplianceException('No server id is set, cannot modify yaml config via REST')
-        return self.server.set_yaml_config(data_dict)
+        if self.version < '5.9':
```
mshriver: :sigh:

cfme/utils/appliance/__init__.py
```diff
@@ -2154,25 +2179,16 @@ def set_rubyrep_replication(self, host, port=5432, database='vmdb_production',
...
-            dest = yaml['workers']['worker_base'][':replication_worker'][':replication'][
-                ':destination']
-            dest[':database'] = database
-            dest[':username'] = username
-            dest[':password'] = password
-            dest[':port'] = port
-            dest[':host'] = host
-            logger.debug('Dest: {}'.format(dest))
-        self.set_yaml_config(yaml)
+        dest = {'workers': {'worker_base': {'replication_worker': {'replication': {
```
mshriver: Before `dest` was the content of `yaml['workers']['worker_base']['replication_worker']['replication']['destination']`, and then 'database', 'username', 'password', 'port', and 'host' were set in that level of the dict.

Here `dest` is the dictionary containing what `dest` was before, and the statements below will set 'database' etc at the top level of `dest`.

Before: 
```python
self.set_yaml_config({'workers': {'worker_base': {'replication_worker': {'repliaction': {'destination': {
    'database': database,
    'username': username,
    'password': password,
    'port': port,
    'host': host}}}}}})
```

Now:
```python
self.set_yaml_config({{'workers': {'worker_base': {'replication_worker': {'repliaction': {'destination': {}}}}},
    'database': database,
    'username': username,
    'password': password,
    'port': port,
    'host': host})
```

PR #6900, opened 2018-03-16
[1LP][RFR] Fixed configure_fleecing fixture in test_alerts.py

quarckster:
Fixing broken fixture  after merging #6820


quarckster:
@izapolsk I rebased the PR, but PRT failures are not related the changes.

PR #6904, opened 2018-03-17
[1LP][RFR] openshift support in pytest sprout options

izapolsk:
Opeshift appliances can be requested only when preconfigured=False and template_type = 'openshift_pod'. This PR is going to make pytest's sprout plugin to request such appliances

REQUIRES: https://github.com/ManageIQ/integration_tests/pull/6905

### DIFF
--- a/cfme/test_framework/sprout/plugin.py
+++ b/cfme/test_framework/sprout/plugin.py
@@ -81,8 +81,7 @@ def mangle_in_sprout_appliances(config):
     log.info("Appliances were provided:")
     for appliance in requested_appliances:
 
-        url = "https://{}/".format(appliance["ip_address"])
-        appliance_args = {'hostname': url}
+        appliance_args = {'hostname': appliance['url']}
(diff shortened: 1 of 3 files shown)


izapolsk:
@mfalesni , @mshriver  I tested it locally against both regular appliance and openshift one.  No issues so far.

cfme/test_framework/sprout/plugin.py
```diff
@@ -73,8 +80,29 @@ def mangle_in_sprout_appliances(config):
     appliances = config.option.appliances
     log.info("Appliances were provided:")
     for appliance in requested_appliances:
+
         url = "https://{}/".format(appliance["ip_address"])
```
mfalesni: **REQUIRED**: Can you change this to pull the URL off appliance using the `.url` property?

izapolsk: honestly, it isn't related to this pr but I moved it to Appliance class

PR #6910, opened 2018-03-19
[1LP][RFR] Update SSL test to support any metrics type

shalomnaim1:
Shriver: modified pytest line below, needs double brackets, was only singular bracket so no tests collected.
{{ pytest: cfme/tests/containers/test_ssl_providers.py --use-provider ocp-v3 }}

Update SSL tests to support Prometheus metrics in addition to Hawkular

### DIFF
--- a/cfme/tests/containers/test_ssl_providers.py
+++ b/cfme/tests/containers/test_ssl_providers.py
@@ -24,7 +24,7 @@ DEFAULT_SEC_PROTOCOLS = (
     pytest.mark.polarion('CMP-10599')('SSL')
 )
 
-checked_item = namedtuple('TestItem', ['default_sec_protocol', 'hawkular_sec_protocol'])
+checked_item = namedtuple('TestItem', ['default_sec_protocol', 'metrics_sec_protocol'])
 TEST_ITEMS = (
     pytest.mark.polarion('CMP-10593')
     (checked_item('SSL trusting custom CA', 'SSL trusting custom CA')),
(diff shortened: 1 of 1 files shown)


cfme/tests/containers/test_ssl_providers.py
```diff
@@ -139,7 +140,7 @@ def test_add_hawkular_provider_ssl(provider, appliance, test_item, soft_assert,
     except AssertionError:
         soft_assert(False, provider.name + ' wasn\'t added successfully using ' +
                     test_item.default_sec_protocol + ' security protocol and ' +
-                    test_item.hawkular_sec_protocol + ' hawkular security protocol')
+                    test_item.hawkular_sec_protocol + ' metrics security protocol')
```
quarckster: **REQUIRED**
Please use `format` to concatenate strings.

shalomnaim1: Done

cfme/tests/containers/test_ssl_providers.py
```diff
@@ -130,16 +131,19 @@ def test_add_hawkular_provider_ssl(provider, appliance, test_item, soft_assert,
...
         new_provider.setup()
         view = appliance.browser.create_view(ContainerProvidersView)
         view.flash.assert_success_message(
             'Containers Providers "' + provider.name + '" was saved')
     except AssertionError:
-        soft_assert(False, provider.name + ' wasn\'t added successfully using ' +
-                    test_item.default_sec_protocol + ' security protocol and ' +
-                    test_item.hawkular_sec_protocol + ' hawkular security protocol')
+        soft_assert(False,
+                    "provider_name wasn\'t added successfully using {default_sec_protocol} "
```
quarckster: **REQUIRED**
If you use double quotes then no need to escape single quote.

shalomnaim1: Could you be clearer? I don't use and single quote here

quarckster: Instead of `"provider_name wasn\'t added successfully using {default_sec_protocol} "`, just write `"provider_name wasn't added successfully using {default_sec_protocol} "`

PR #6911, opened 2018-03-19
[1LP][WIP] Enable C&U while adding openshift provider with metrics endpoint

shalomnaim1:
{{ pytest: cfme/tests/containers/test_alerts.py::test_add_alerts_provider --use-provider ocp-v3 }}
Enable C&U while adding openshift provider with metrics endpoint

### DIFF
--- a/cfme/containers/provider/openshift.py
+++ b/cfme/containers/provider/openshift.py
@@ -114,7 +114,7 @@ class OpenshiftProvider(ContainersProvider):
     def create(self, **kwargs):
 
         # Enable C&U for any provider with metrics provider
-        if getattr(self, "metrics_type") != "Disabled":
+        if self.metrics_type != "Disabled":
             self.appliance.server.settings.enable_server_roles(
                 'ems_metrics_coordinator', 'ems_metrics_collector', 'ems_metrics_processor')
 
--- a/cfme/containers/provider/openshift.py
+++ b/cfme/containers/provider/openshift.py
@@ -113,10 +113,10 @@ class OpenshiftProvider(ContainersProvider):
 
     def create(self, **kwargs):
 
+        # Enable C&U for any provider with metrics provider
         if getattr(self, "metrics_type") != "Disabled":
-            server = self.appliance.server.settings
-            server.enable_server_roles('ems_metrics_coordinator', 'ems_metrics_collector',
-                                       'ems_metrics_processor')
+            self.appliance.server.settings.enable_server_roles(
+                'ems_metrics_coordinator', 'ems_metrics_collector', 'ems_metrics_processor')
 
         # Enable alerts collection before adding the provider to avoid missing active
         # alert after adding the provider
--- a/cfme/containers/provider/openshift.py
+++ b/cfme/containers/provider/openshift.py
@@ -113,6 +113,11 @@ class OpenshiftProvider(ContainersProvider):
 
     def create(self, **kwargs):
 
+        if getattr(self, "metrics_type") != "Disabled":
+            server = self.appliance.server.settings
+            server.enable_server_roles('ems_metrics_coordinator', 'ems_metrics_collector',
+                                       'ems_metrics_processor')
+
         # Enable alerts collection before adding the provider to avoid missing active
         # alert after adding the provider
         # For more info: https://bugzilla.redhat.com/show_bug.cgi?id=1514950


jamesooden:
I'm waiting on your PRT results before I change to 1LP.

cfme/containers/provider/openshift.py
```diff
@@ -113,6 +113,12 @@ def __init__(
 
     def create(self, **kwargs):
 
+        # Enable C&U for any provider with metrics provider
+        if getattr(self, "metrics_type") != "Disabled":
```
if a provider doesn't have "metrics_type" attribute  `AttributeError` will be raised.

shalomnaim1: I had this discussion with @mshriver last week, we agreed that all the OpenShift Providers MUST have this attribute and if it missing then we have to fix the YAML accordingly, therefore I didn't set a default value to ```getattr```

mshriver: @quarckster there's a default of None for metrics_type attribute, so will always be there.

cfme/containers/provider/openshift.py
```diff
@@ -113,6 +113,12 @@ def __init__(
...
+            server = self.appliance.server.settings
```
mshriver: `server` is a bit of a mis-name here, I'd rather just see a direct call to `self.appliance.server.settings.enable_server_roles()` since its not used for any further calls.

shalomnaim1: fixed....

cfme/containers/provider/openshift.py
quarckster: @shalomnaim1 @mshriver  In that case `getattr` doesn't make sense, use `self.metrics_type` instead.

shalomnaim1: @quarckster Ok, that's make seanse

cfme/containers/provider/openshift.py
```diff
@@ -113,6 +113,11 @@ def __init__(
 
     def create(self, **kwargs):
 
+        # Enable C&U for any provider with metrics provider
+        if self.metrics_type != "Disabled":
```
mshriver: Is there a reason for automatically enabling these roles, and not just enabling them when they're needed for test?

quarckster: I changed to WIP for now

shalomnaim1: @mshriver Basically, this three roles activating the metrics collection for openshift provider (and maybe for other providers as well), so adding a metrics endpoint without enabling is not relevant at all. Since we don't have any automated test for metrics I testing these features manually and adding the provider using the automation.

quarckster: Honestly, I don't like it. If you remove openshift provider these roles will be active still.

mshriver: I agree @quarckster. I'm opposed to this change, these roles should only be enabled when needed for testing, and disabled at the end of a test (fixture setup+teardown).

@shalomnaim1 If you need these roles enabled for your manual testing, having them enabled when the provider is configured via automation is not the right solution.  You can use `miq shell` and these methods to configure the roles for yourself when testing.

This just needlessly increases the appliance worker load on all tests using this provider.

shalomnaim1: @mshriver As long as the metrics endpoint exists, this role has to be enabled.

mfalesni: @shalomnaim1 Do the tests fail without that role enabled?

shalomnaim1: @mfalesni sure, without this roles no metrics test can't be run since CFME will not collect any metrics. In addition, if someone adding an openshift provider for some reason without enabling this role, no metrics will be collected

mshriver: That's the expected behavior with all providers, and fixtures within the tests should enable the roles on the appliance.  There are, as far as I know, no container metrics tests in automation yet. When added, they should enable the roles in a fixture as well.

shalomnaim1: IMO, when someone adds a metrics endpoint he wants metrics to be collected, so it has to be a part of the setup and not part of the test. In the same way, you can say that if I want to test metrics on OCP I will have to create a fixture of setting up the metric endpoint, for me it just not making any sense.

quarckster: @shalomnaim1 How do you propose to disable roles when an openshift provider was removed?

shalomnaim1: @quarckster I think it can be done by adding code which disables these roles while removing OCP provider with configured metrics endpoint or just metrics endpoint removal.

PR #6912, opened 2018-03-19
[1LP][RFR]Implementing fix landing_page related failures

anurag03:
Purpose or Intent
=================
There is a change in the setting model page which caused some failure in the landing page test case, providing a fix for that.

{{pytest: -v -k 'test_landing_page_admin'}}

### DIFF
--- a/cfme/tests/configure/test_landing_page.py
+++ b/cfme/tests/configure/test_landing_page.py
@@ -1,6 +1,5 @@
 import pytest
 
-from cfme.configure.settings import Visual
 from cfme.utils.appliance.implementations.ui import navigate_to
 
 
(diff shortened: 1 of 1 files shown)


cfme/tests/configure/test_landing_page.py
```diff
@@ -105,13 +102,18 @@
     'Red Hat Access Insights']
 
 
-def set_landing_page(value, appliance):
+@pytest.yield_fixture(scope='module')
```
izapolsk: Q: why yield_fixture ?

And I doubt you need such function at all. You can just add such definition to test or to both fixtures.

anurag03: It was repetitive so I though of using it, moreover if I add some more test case related to rbac user I can directly use it.

izapolsk: @anurag03, honestly I don't see any benefit of wrapping one row in function/fixture. It doesn't bring any value to you. you just put another fixture into test definition instead of defining that variable in test.

PR #6915, opened 2018-03-19
[1LP][RFR] Moved the setup_provider first in the fixture list

sshveta:
Moved the setup_provider fixture first in the list else the catalog item was created before setting up provider .


sshveta:
Two test failures  are fixed in PR https://github.com/ManageIQ/integration_tests/pull/6919. 
 test_retire_service_on_date and test_crud_set_ownership_and_edit_tags.

Rest service failures are not related to this PR .

PR #6927, opened 2018-03-20
[1LP][RFR] Wrap appliance log tarring, catch errors

mshriver:
Tested locally with files that DNE on the appliance, tar command does not error out and what files are present get included in the archive.

Added `try:except` block to capture command assertion error to stop pytest from exiting with the exception.

### DIFF
--- a/cfme/test_framework/appliance_log_collector.py
+++ b/cfme/test_framework/appliance_log_collector.py
@@ -59,9 +59,16 @@ def pytest_unconfigure(config):
                     app.hostname)
                 logger.debug('Creating tar file on app %s:%s with log files %s',
                              app, tar_file, ' '.join(log_files))
-                tar_cmd = 'tar -czvf {tar} {files}'.format(tar=tar_file,
-                                                           files=' '.join(log_files))
-                assert ssh_client.run_command(tar_cmd)
+                # wrap the files in ls, redirecting stderr, to ignore files that don't exist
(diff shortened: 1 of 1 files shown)


cfme/test_framework/appliance_log_collector.py
```diff
@@ -59,9 +59,16 @@ def pytest_unconfigure(config):
                     app.hostname)
                 logger.debug('Creating tar file on app %s:%s with log files %s',
                              app, tar_file, ' '.join(log_files))
-                tar_cmd = 'tar -czvf {tar} {files}'.format(tar=tar_file,
-                                                           files=' '.join(log_files))
-                assert ssh_client.run_command(tar_cmd)
+                # wrap the files in ls, redirecting stderr, to ignore files that don't exist
+                tar_result = ssh_client.run_command(
+                    'tar -czvf {tar} $(ls {files} 2>/dev/null)'
```
izapolsk: -j could be better and you probably 1>&2 2>/dev/null

mshriver: I don't want `1>&2` because I'm relying on the stdout here to populate the tar command for files that do exist.

mshriver: Definitely worth considering bzip archives here if/when we see space issues or timing problems with pulling the archives.

PR #6928, opened 2018-03-20
[1LP][RFR] Filter azure out of test_reset_vm, azure doesn't support hard reset

bsquizz:
You can't do a 'hard reboot' from CFME UI on an azure instance, since that is unsupported by azure. So reset VM via REST shouldn't be expected to work either.


jamesooden:
Looks good to me.   Let me know when the PRT is done and I'll approve to 1LP.

bsquizz:
@jamesooden just double checked, correct, those were failures in test_suspend on rhos, not related to this test.

PR #6929, opened 2018-03-20
[1LP][RFR] Tweak base login and ui for exception handling

mshriver:
Was producing an exception because the view isn't actually displayed.

```python
>               appliance.server.login(user, method='click_on_login')

cfme/tests/integration/test_cfme_auth.py:183: 
/usr/lib/python2.7/site-packages/sentaku/context.py:132: in __call__
    return bound_method(*k, **kw)
cfme/base/ui.py:189: in login
    user.name = logged_in_view.current_fullname

self = <cfme.base.login.BaseLoggedInPage object at 0x7fcd42f2f610>

    @property
    def current_fullname(self):
>       return self.settings.text.strip().split('|', 1)[0].strip()
E       AttributeError: 'NoneType' object has no attribute 'strip'

cfme/base/login.py:46: AttributeError
```

### DIFF
--- a/cfme/base/login.py
+++ b/cfme/base/login.py
@@ -44,6 +44,7 @@ class BaseLoggedInPage(View):
     @property
     def current_fullname(self):
         try:
+            # When the view isn't displayed self.settings.text is None, resulting in AttributeError
             return self.settings.text.strip().split('|', 1)[0].strip()
         except AttributeError:
             return None
--- a/cfme/base/login.py
+++ b/cfme/base/login.py
@@ -43,7 +43,10 @@ class BaseLoggedInPage(View):
 
     @property
     def current_fullname(self):
-        return self.settings.text.strip().split('|', 1)[0].strip()
+        try:
+            return self.settings.text.strip().split('|', 1)[0].strip()
+        except AttributeError:
+            return None
 
     @property
     def current_groupname(self):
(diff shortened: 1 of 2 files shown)


cfme/base/ui.py
```diff
@@ -186,10 +186,10 @@ def login(self, user=None, method=LOGIN_METHODS[1]):
 
         login_view.log_in(user, method=method)
         logged_in_view.flush_widget_cache()
-        user.name = logged_in_view.current_fullname
```
mshriver: Moving this until after the `view.is_displayed` so that raises an exception before getting `None` for this property and modifying user.

cfme/base/login.py
```diff
@@ -43,7 +43,10 @@ def current_username(self):
 
     @property
     def current_fullname(self):
-        return self.settings.text.strip().split('|', 1)[0].strip()
+        try:
+            return self.settings.text.strip().split('|', 1)[0].strip()
+        except AttributeError:
```
jamesooden: Is there anything that could be said in a comment about why we would get that attribute error, or the named exception enough info?   My thinking is you saw this error under a specific condition and hence commenting on that condition may be helpful for future you or me.

mshriver: As in the methods above, it ends up happening because there's no check that the view is actually displayed.  An AttributeError is raised because `self.settings.text` is None, which has no `strip()` attr. Will put in comment.

PR #6930, opened 2018-03-20
[1LP][RFR] Automation Fixes for CMQE

jawatts:
Failures in the last RC job required some test updates. 

{{ pytest: cfme/tests/containers/test_basic_metrics.py  cfme/tests/containers/test_manageiq_ansible_custom_attributes.py cfme/tests/containers/test_ad_hoc_metrics.py --use-provider ocp-v2 }}

### DIFF
--- a/cfme/tests/containers/test_ad_hoc_metrics.py
+++ b/cfme/tests/containers/test_ad_hoc_metrics.py
@@ -14,13 +14,20 @@ pytestmark = [
 @pytest.fixture(scope="function")
 def metrics_up_and_running(provider):
 
-    router = [router for router in provider.mgmt.o_api.get('route')[1]['items'] if
-              router["metadata"]["name"] == "hawkular-metrics"].pop()
-    hawkular_url = router["status"]["ingress"][0]["host"]
-    response = requests.get("https://{url}:443".format(url=hawkular_url), verify=False)
+    try:
+        router = [router for router in provider.mgmt.o_api.get('route')[1]['items']
+                  if router["metadata"]["name"] == 'hawkular-metrics' or
+                  router["metadata"]["name"] == 'prometheus'].pop()
+    except IndexError:
+        pytest.skip('No Metrics Route available for {}'.format(provider.key))
 
-    assert response.ok, "hawkular failed to start!"
-    logger.info("hawkular started successfully")
+    metrics_url = router["status"]["ingress"][0]["host"]
+    creds = provider.get_credentials_from_config(provider.key, cred_type='token')
+    header = {"Authorization": "Bearer {token}".format(token=creds.token)}
+    response = requests.get("https://{url}:443".format(url=metrics_url), headers=header,
+                            verify=False)
+    assert response.ok, "{metrics} failed to start!".format(metrics=router["metadata"]["name"])
+    logger.info("{metrics} started successfully".format(metrics=router["metadata"]["name"]))
 
 
 def is_ad_hoc_greyed(provider_object):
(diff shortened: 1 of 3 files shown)


cfme/tests/containers/test_basic_metrics.py
```diff
@@ -18,7 +21,7 @@ def test_basic_metrics(provider):
     routes = provider.mgmt.o_api.get('route')[1]['items']
     metrics_hostname = [route for route in routes
                         if route['metadata']['name'] == 'hawkular-metrics']
-    assert metrics_hostname, 'Could not find route for hawkular-metrics'
+    assert metrics_hostname, 'Could not find route for metrics'
     metrics_hostname = metrics_hostname.pop()['spec']['host']
     host_url = 'https://' + metrics_hostname + '/hawkular/metrics/'
```
shalomnaim1: This test requires a massive code refactoring, IMO instead of using the ```curl``` command, we can use python's ```request``` module. In addition, to validate Prometheus using ```curl``` do like the following example:
```curl -H "Authorization: Bearer <OCP token>" -k https://<Prometheus metrics route>/graph```.

File: cfme/tests/containers/test_basic_metrics.py [hunk 2]
```diff
@@ -8,8 +8,11 @@
     pytest.mark.tier(1),
     pytest.mark.provider([ContainersProvider], scope='function')]
 
+# TODO This test needs to be reevaluated. This is not testing anyting in CFME.
+
 
 @pytest.mark.polarion('CMP-10205')
+@pytest.mark.uncollectif(lambda provider: provider.metrics_type == 'Prometheus')
```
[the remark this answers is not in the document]
shalomnaim1: We can consider parameterization on this test so it would support any provider type, but on failure, it would raise an exception on specific endpoint type.

cfme/tests/containers/test_basic_metrics.py
shalomnaim1: Validating Prometheus endpoint is done by sending a request to ```https://<Promethues metrics route/graph``` and it requires add of ```Authorization``` header to the request e.g: ```curl -H "Authorization: Bearer <OCP Token>" -k https:/<Prometheus metrics route>/graph```

shalomnaim1: In addition, I think using ```curl``` command here is less appropriate, I would prefer using  python's ```requests``` module

jawatts:
@shalomnaim1  I agree with you that this test does need a refactor, see the TODO I added on line 11 I had added already, but I do not believe this PR is the correct place to do it. For now, I am cleaning up our test run. A future PR is needed to do a complete overhaul of this test.

cfme/tests/containers/test_ad_hoc_metrics.py
```diff
@@ -14,13 +14,16 @@
...
-    router = [router for router in provider.mgmt.o_api.get('route')[1]['items'] if
-              router["metadata"]["name"] == "hawkular-metrics"].pop()
-    hawkular_url = router["status"]["ingress"][0]["host"]
-    response = requests.get("https://{url}:443".format(url=hawkular_url), verify=False)
-
-    assert response.ok, "hawkular failed to start!"
-    logger.info("hawkular started successfully")
+    router = [router for router in provider.mgmt.o_api.get('route')[1]['items']
+              if router["metadata"]["name"] == 'hawkular-metrics' or
+              router["metadata"]["name"] == 'prometheus'].pop()
```
psav: **REQUIRED:** This could result in an IndexError if there are no routers - We should handle this and skip the test if so.

PR #6933, opened 2018-03-21
[WIP] Adding a new test to enable and disable reports schedule

niyazRedhat:


cfme/intelligence/reports/schedules.py
```diff
@@ -168,6 +168,16 @@ def queue(self):
...
 
+    @property
+    def status(self):
+        view = navigate_to(self.parent, "All")
+        for item in view.schedules_table.read():
+            if item['Name'] == self.name:
+                if item['Active'] == 'True':
+                    return True
+                else:
+                    return False
```
digitronik: I think best to go with property name `is_enable()` rather than `status`. it will give meaning to this property for checking `schedule` is `enable` or `disable`. 

`schedule.is_enable` will return `True` if its enable else `False`.

jamesooden: I like that idea.   I looked at this earlier and I could understand it and it was pretty straightforward, however your suggestion would reduce the line noise and make it even more readable.

mshriver: Great call guys, but I'd prefer `enabled` in this context. Like our use of `self.exists` elsewhere in the framework.

mshriver: There's no need to split the return out here, you can return the result of the statement that your `if` is working off of.

```python
for item in view.schedules_table.read():
    if item['Name'] == self.name 
        return item['Active'] == 'True'
```

cfme/tests/intelligence/reports/test_crud.py
```diff
@@ -80,6 +80,19 @@ def test_schedule_crud(schedule_data, appliance):
     schedule.delete()
 
 
+@pytest.mark.sauce
+@pytest.mark.tier(3)
+@test_requirements.report
+def test_reports_disable_enable_schedule(schedule_data, appliance):
+    schedules = ScheduleCollection(appliance)
```
mshriver: Collection items should be retrieved through the collections registered to the appliance.

```python
schedules = appliance.collections.schedules
```

PR #6941, opened 2018-03-21
[1LP][RFR] Changed tag name because of a recent change

sshveta:
In a recent change all tags are suffixed by * ,
so removing it from the test here .

### DIFF
--- a/cfme/tests/services/test_myservice.py
+++ b/cfme/tests/services/test_myservice.py
@@ -101,7 +101,7 @@ def test_crud_set_ownership_and_edit_tags(appliance, context, myservice):
     with appliance.context.use(context):
         myservice = MyService(appliance, name=service_name, vm_name=vm_name)
         myservice.set_ownership("Administrator", "EvmGroup-administrator")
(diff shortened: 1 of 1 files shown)


cfme/tests/services/test_myservice.py
```diff
@@ -101,7 +101,7 @@ def test_crud_set_ownership_and_edit_tags(appliance, context, myservice):
     with appliance.context.use(context):
         myservice = MyService(appliance, name=service_name, vm_name=vm_name)
         myservice.set_ownership("Administrator", "EvmGroup-administrator")
-        myservice.add_tag("Cost Center *", "Cost Center 001")
+        myservice.add_tag("Cost Center", "Cost Center 001")
```
otsuman: QUESTION: Is it for both 5.8 and 5.9?

sshveta: Yes on both

sshveta:
File download tests failure on PRT are sporadic , tests pass locally .

PR #6943, opened 2018-03-22
[1LP][RFR] storage Manager provider cleanup

digitronik:
Purpose or Intent
=================
 - I found that in Jenkins run for `storage`. It first run `test_manager.py` in that `test_storage_manager_delete` causes deletion of `child providers`. Its continue with that brocken provider for other files causes `NoSuchElementException`. 
- best to clear that `CloudProvider`

{{pytest: cfme/tests/storage/ -v}}

### DIFF
--- a/cfme/tests/storage/test_manager.py
+++ b/cfme/tests/storage/test_manager.py
@@ -2,7 +2,6 @@
 import pytest
 
 from cfme import test_requirements
-from cfme.cloud.provider import CloudProvider
 from cfme.cloud.provider.openstack import OpenStackProvider
 from cfme.storage.manager import StorageManagerDetailsView
 from cfme.utils.appliance.implementations.ui import navigate_to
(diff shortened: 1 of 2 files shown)


digitronik:
Errors are due to #6947

cfme/tests/storage/test_manager.py
```diff
@@ -11,13 +12,19 @@
...
     pytest.mark.usefixtures('setup_provider'),
-    pytest.mark.provider([OpenStackProvider])
+    pytest.mark.provider([OpenStackProvider], scope='module')
 ]
 
 
 MANAGER_TYPE = ['Swift Manager', 'Cinder Manager']
 
 
+@pytest.yield_fixture(scope='module')
```
psav: **REQUIRED:** I would prefer to see the ```openstack_provider``` object being used in the fixture rather than clearing all CloudProviders. Is that possible?

digitronik: @psav why not. thank's for pointing this.

digitronik:
@psav, I tried with `provider.delete()` `#run-22783` found that deletion with UI took more time and its continue with a broken provider.
 
As its part of `tear down`. I choose deletion with rest `provider.delete_rest()` and use method `provider.wait_for_delete()`. Internally most our provider cleanup method used rest.

PR #6948, opened 2018-03-22
[1LP][RFR]updating Vm verification

lkhomenk:
Purpose or Intent
=================

__Fixing__ VM verification because atm module-scoped test being parallelized to different appliances might provision duplicated VMs - EC2 allows duplicated names.
1. Verify that VM with such name exists on backend
2. Check it in CFME - refresh if it's not

This leads to the EC2 cleanup issue - we had duplicated Instances which we couldn't delete

{{pytest: cfme/tests/cloud_infra_common/test_custom_attributes_rest.py --long-running --use-provider ec2 --use-sprout --sprout-group downstream-59z  --sprout-appliance 5}}

### DIFF
--- a/cfme/tests/cloud_infra_common/test_custom_attributes_rest.py
+++ b/cfme/tests/cloud_infra_common/test_custom_attributes_rest.py
@@ -53,13 +53,8 @@ def get_vm(appliance, provider, vm_obj):
 
     def _get_vm():
         if not provider.mgmt.does_vm_exist(vm_obj.name):
-            vms = collection.find_by(name=vm_obj.name)
-            if not vms:
-                vm_obj.create_on_provider(timeout=2400, find_in_cfme=True, allow_skip='default')
+            vm_obj.create_on_provider(timeout=2400, find_in_cfme=True, allow_skip='default')
         vms = collection.find_by(name=vm_obj.name)
-        if not vms:
-            provider.refresh_provider_relationships()
-            vms = collection.find_by(name=vm_obj.name)
         return vms[0]
 
     return _get_vm
--- a/cfme/tests/cloud_infra_common/test_custom_attributes_rest.py
+++ b/cfme/tests/cloud_infra_common/test_custom_attributes_rest.py
@@ -52,9 +52,13 @@ def get_vm(appliance, provider, vm_obj):
         collection = appliance.rest_api.collections.instances
 
     def _get_vm():
+        if not provider.mgmt.does_vm_exist(vm_obj.name):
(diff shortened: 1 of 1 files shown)


cfme/tests/cloud_infra_common/test_custom_attributes_rest.py
```diff
@@ -56,9 +56,13 @@ def get_vm(appliance, provider, vm_obj):
         collection = appliance.rest_api.collections.instances
 
     def _get_vm():
+        if not provider.mgmt.does_vm_exist(vm_obj.name):
+            vms = collection.find_by(name=vm_obj.name)
```
mkoura: QUESTION: why is this necessary? If we are in this branch, the VM doesn't exist so ``find_by`` will not return anything, right?

lkhomenk: collection checks for VM in CFME. this test being parallelized and module-scoped - is executed simultaneously with the same `vm_name` on different appliances. which would lead to `create_on_provider` if vm not found on the appliance

lkhomenk:
making fixtures `function-scoped` has a huge downside - to many new VMs will be created
BTW - to testing puspose we don't even need VM on the backend side. test passes if we use any archived VM/Instance too - hence it would be nice to limit created vms

lkhomenk:
@mkoura @mmojzis ^ any objections?

cfme/tests/cloud_infra_common/test_custom_attributes_rest.py
mkoura: @lkhomenk Yes, but I don't see a point in calling ``collection.find_by(..)`` when we already checked that the VM doesn't exist on provider. Could this part be just:
```python
if not provider.mgmt.does_vm_exist(vm_obj.name):
    vm_obj.create_on_provider(timeout=2400, find_in_cfme=True, allow_skip='default')
```
?

lkhomenk: @mkoura if VM exists on the provider but missing in CFME - we'll fail getting VM's REST representation - `vms = collection.find_by(name=vm_obj.name)`

mkoura: @lkhomenk that's my point. If the vm doesn't exist on provider, there's no point checking if it exists in CFME - it will always fail. Checking if it exists on provider and creating it if doesn't is enough.

mshriver:
Positive PRT results, tests were collected.

PR #6949, opened 2018-03-22
[1LP][WIP] Test updating password from the login screen

apagac:
This adds a new test and some new implementation that goes with it.

Test creates new user, then tries to change user's password while logging in, then checks if user can't use the old password.

I opted for creating new user instead of just using the Administrator user. Changing the password for the Administrator user results in weird automation behaviour and also Sprout is not happy with it.

### DIFF
--- a/cfme/tests/test_login.py
+++ b/cfme/tests/test_login.py
@@ -1,3 +1,4 @@
+import fauxfactory
 import pytest
 
 from cfme import test_requirements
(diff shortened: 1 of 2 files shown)


dajoRH:
Lint report for commit dd9a1e12be6f611bc9eefe856c4fb061a0581dda:

`cfme/base/ui.py`:
- :bangbang: Line 119:5: **E303** *too many blank lines (2)*
- :heavy_exclamation_mark: Line 168:1: **W293** *blank line contains whitespace*

`cfme/tests/test_login.py`:
- :bangbang: Line 71:20: **E251** *unexpected spaces around keyword / parameter equals*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/test_login.py
```diff
@@ -55,3 +55,36 @@ def test_bad_password(context, request, appliance):
...
             view = appliance.browser.create_view(LoginPage)
             assert view.password.read() == '' and view.username.read() == ''
+
+
+@pytest.mark.parametrize('context', [ViaUI])
+def test_update_password(context, request, appliance):
+    """ Test updating password from the login screen. """
+
+    # First, create a temporary new user
+    import fauxfactory
```
quarckster: **REQUIRED**
Please place it at the top of the module.

cfme/tests/test_login.py
```diff
@@ -55,3 +56,35 @@ def test_bad_password(context, request, appliance):
...
+
+    # Now try to login while changing the password for the user
+    changed_pass_page = appliance.server.update_password(new_password='changeme', user=user)
+    assert changed_pass_page.is_displayed
+    changed_pass_page.logout()
+
+    # Try to login with the user with old password
+    error_message = "Incorrect username or password"
+    with error.expected(error_message):
+        appliance.server.login(user)
```
psav: **REQUIRED** Where are the credentials being updated on the user's credential object?

apagac: As far as I understand it, the user object is not being updated. The process of updating a password with "update_password" does change the password in the appliance but not in the user object used in this test.

And it's not entirely a bad thing in this case. When I'm trying to log-in with an old password, I just use the one the user object has stored. This should produce an error, because the password in the appliance is different than we have stored in user object.

Considering that I don't need the user object after the password changes, I had no motivation to go deeper or to update the password for the user manually.

psav: OK I understand.

PR #6951, opened 2018-03-22
[1LP][RFR] updated custom data isn't saved

izapolsk:
When there is new openshift template in trackerbot, sprout doesn't update custom data correctly.
It happens because only some certain fields are saved but custom_data isn't. 
This PR fixes that issue.

### DIFF
--- a/sprout/appliances/tasks.py
+++ b/sprout/appliances/tasks.py
@@ -270,8 +270,13 @@ def poke_trackerbot(self):
         template_name = template["template"]["name"]
         ga_released = template['template']['ga_released']
         date = parse_template(template_name).datestamp
+
+        # nasty trackerbot slightly corrupts json data and it is parsed in wrong way
+        # as a result
         custom_data = template['template'].get('custom_data', "{}")
         processed_custom_data = custom_data.replace("u'", '"').replace("'", '"')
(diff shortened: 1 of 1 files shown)


jamesooden:
This looks good to me.   Waiting for PRT results.

sprout/appliances/tasks.py
```diff
@@ -311,14 +311,15 @@ def poke_trackerbot(self):
                 with transaction.atomic():
                     tpl = Template(
                         provider=provider, template_group=group, original_name=template_name,
-                        name=template_name, preconfigured=False, date=date, custom_data=custom_data,
+                        name=template_name, preconfigured=False, date=date,
+                        custom_data=processed_custom_data,
                         version=template_version, ready=True, exists=True, usable=True)
                     tpl.save()
                     if provider.provider_type == 'openshift':
                         tpl.custom_data = yaml.safe_load(processed_custom_data)
```
mfalesni: **QUESTION**: Here is yaml load, but on line 315 it is set directly. I don't understand it, it can digest it like this?

psav: Tagging as **REQUIRED**

izapolsk: @mfalesni , @psav, optimized yaml calls.

PR #6953, opened 2018-03-23
[1LP][RFR] infra provider discovery screen test

digitronik:
Purpose or Intent
=================
- Automate Infra provider discovery screen test by @jistefan 
https://polarion.engineering.redhat.com/polarion/#/project/RHCF3/workitem?id=RHCF3-10050
- Added `OpenStack Intra` Discovery support in `DiscoveryView`
{{pytest: cfme/tests/infrastructure/test_providers.py::test_infra_discovery_screen -v}}

### DIFF
--- a/cfme/tests/infrastructure/test_providers.py
+++ b/cfme/tests/infrastructure/test_providers.py
@@ -27,20 +27,20 @@ pytestmark = [
 ]
 
 discovery_ips = [
-    {'from': ['120', '120', '120', '120'], 'to': '119',
+    {'from': ['10', '120', '120', '120'], 'to': '119',
      'msg': "Infrastructure Providers Discovery returned: "
             "Ending address must be greater than starting address"},
     {'from': ['333', '120', '120', '120'], 'to': '120',
      'msg': "Infrastructure Providers Discovery returned: IP address octets must be 0 to 255"},
-    {'from': ['120', '333', '120', '120'], 'to': '120',
+    {'from': ['10', '333', '120', '120'], 'to': '120',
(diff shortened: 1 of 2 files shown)


cfme/tests/infrastructure/test_providers.py
```diff
@@ -25,6 +26,24 @@
     pytest.mark.provider([InfraProvider], scope="function"),
 ]
 
+discovery_ips = [
+    {'from': ['120', '120', '120', '120'], 'to': '119',
```
psav: **REQUIRED:** These addresses are potentially valid, and though we expect to fail, we should use a private address range so that it doesn't potentially hit an external IP.
```10.0.0.0 – 10.255.255.255```
```172.16.0.0 – 172.31.255.255```
```192.168.0.0 – 192.168.255.255```
The 333 ones are OK because even if it didn't catch it was invalid, 333 IS invalid and wouldn't communicate out.

File: cfme/tests/infrastructure/test_providers.py [hunk 2]
```diff
@@ -234,3 +253,46 @@ def test_infrastructure_add_provider_trailing_whitespaces():
...
+    view.scvmm.click()
+    assert not view.scvmm.selected
+
+    view.rhevm.click()
+    assert view.rhevm.selected
+    view.rhevm.click()
+    assert not view.rhevm.selected
+
+    if appliance.version >= '5.9.2':
+        view.osp_infra.click()
```
psav: **REQUIRED:** We should assert ```is_displayed``` like we do on the others, inside this conditional.

PR #6963, opened 2018-03-26
[1LP][RFR] Fixed navigation for Node Alert Profile

quarckster:
{{pytest: -v -k test_alert_profile_crud}}

### DIFF
--- a/cfme/control/explorer/alert_profiles.py
+++ b/cfme/control/explorer/alert_profiles.py
@@ -308,4 +308,9 @@ class VMInstanceAlertProfile(BaseAlertProfile):
 
 class NodeAlertProfile(BaseAlertProfile):
 
-    TYPE = "Node"
+    @property
+    def TYPE(self):  # noqa
+        if self.appliance.version < "5.9.2":
+            return "Node"
+        else:
+            return "Container Node"


cfme/control/explorer/alert_profiles.py
```diff
@@ -308,4 +308,9 @@ class VMInstanceAlertProfile(BaseAlertProfile):
 
 class NodeAlertProfile(BaseAlertProfile):
 
-    TYPE = "Node"
+    @property
+    def TYPE(self):  # noqa
+        if self.appliance.version < "5.9.2":
```
CharlleDaniel: `<` doesn't compares strings as expected in this case. Try "5.10.2". It will return True.

mshriver: @CharlleDaniel This appears to be behaving correctly to me:

```
In [5]: node
Out[5]: NodeAlertProfile(description='test-profile', alerts=None, notes=None)

In [6]: node.TYPE
Out[6]: 'Container Node'

In [7]: app.version
Out[7]: Version('5.9.2.0')
```

```
In [5]: node = profiles.instantiate(NodeAlertProfile, description='test-profile')

In [6]: node.TYPE
Out[6]: 'Node'

In [7]: app.version
Out[7]: Version('5.9.1.1')
```

quarckster: `appliance.version` refers to `miq_version.Version`. Feel free to study it.

CharlleDaniel: @quarckster  We still have a string comparison. Here is a running example:
image

Feel free to run [https://trinket.io/python/6edc87ab97](https://trinket.io/python/6edc87ab97)

quarckster: @CharlleDaniel your example shows strings comparison, but in my PR I compare string and instance of `miq_version.Version` class https://github.com/RedHatQE/miq_version/blob/master/miq_version/__init__.py. It has implemented `__eq__`, `__lt__` and `__cmp__`. Please study the framework and its classes.

CharlleDaniel: @quarckster I didn't have noticed that, thank you.

PR #6966, opened 2018-03-26
[1LP][RFR] Update appliance get/set_yaml_config to property

mshriver:
Update appliance `get_yaml_config`/`set_yaml_config` methods, rewrite to be an `advanced_settings` property with a setter.

Slight enhancement to appliance's yaml configuration methods, a property is a better means to represent the appliance configuration.

Updating all uses of the old methods to instead use correct property+setter syntax.

{{ pytest: cfme/tests/configure/test_ntp_server.py cfme/tests/control/test_alerts.py cfme/tests/configure/test_server_roles.py --long-running -v}}

### DIFF
--- a/cfme/base/__init__.py
+++ b/cfme/base/__init__.py
@@ -68,25 +68,26 @@ class Server(BaseEntity, sentaku.modeling.ElementMixin):
             [self.appliance.rest_api.collections.servers._href, str(self.sid), 'settings']
         )
 
-    def get_yaml_config(self):
+    @property
+    def advanced_settings(self):
         """GET servers/:id/settings api endpoint to query server configuration"""
         return self.appliance.rest_api.get(url=self._api_settings_url)
 
-    def set_yaml_config(self, data_dict):
+    def update_advanced_settings(self, settings_dict):
         """PATCH settings from the server's api/server/:id/settings endpoint
 
         Args:
-            data_dict: dictionary of the changes to be made to the yaml configuration
-                       JSON dumps data_dict to pass as raw hash data to rest_api session
+            settings_dict: dictionary of the changes to be made to the yaml configuration
+                       JSON dumps settings_dict to pass as raw hash data to rest_api session
         Raises:
             AssertionError: On an http result >=400 (RequestsResponse.ok)
         """
-        # Calling the _session patch method because the core patch will wrap data_dict in a list
-        # Failing with some data_dict, like 'authentication'
+        # Calling the _session patch method because the core patch will wrap settings_dict in a list
+        # Failing with some settings_dict, like 'authentication'
         # https://bugzilla.redhat.com/show_bug.cgi?id=1553394
         result = self.appliance.rest_api._session.patch(
             url=self._api_settings_url,
-            data=json.dumps(data_dict)
+            data=json.dumps(settings_dict)
         )
         assert result.ok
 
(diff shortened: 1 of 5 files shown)


jamesooden:
How did you come to the name of advanced_settings?   Is that how this yaml config is referred to in the CFME docs?

mshriver:
@jamesooden I chose advanced_settings because in the CFME UI the tab on the settings page is called 'Advanced'.  Over the REST API its JSON data being transmitted, though the data is written/loaded from disk as yaml and represented in YAML in the UI for easier editing (I guess).  I'm not that married to the name, but figured if there was a time to update it, doing so with the property switch would be the time.

Any suggestions for better names?

jamesooden:
No, now that I know how you got it, that makes sense.   Might make sense to leave that in a comment somewhere..

cfme/utils/appliance/__init__.py
```diff
@@ -2519,7 +2527,7 @@ def configure_fleecing(self, log_callback=None):
                 self.add_rhev_direct_lun_disk()
 
             log_callback('Enabling smart proxy role...')
-            roles = self.server.settings.server_roles_db
+            roles = self.server.settingghos.server_roles_db
```
rlbabyuk: a typo?

mshriver: Literally saw this earlier and thought I had fixed, thanks @rlbabyuk

cfme/utils/appliance/__init__.py
```diff
@@ -2522,13 +2530,6 @@ def configure_fleecing(self, log_callback=None):
             roles = self.server.settings.server_roles_db
             if not roles["smartproxy"]:
                 self.server.settings.enable_server_roles("smartproxy")
-                # web ui crashes
```
mshriver: Old block for CFME 5.2 or 5.5

psav:
The only concern I have is around the fact that a setter/getter feels like a complete replacement, yet we are patching in later appliances. thoughts?

mfalesni:
@psav I think having a property and a method would represent the interface the best. Property for getting all configuration and for setting all configuration ... and a method for submitting a patch for the configuration.

mshriver:
@mfalesni I think you have hit on the best way to approach this. As I was making these updates, I repeated the same few lines to modify the existing top-level settings dict key, instead of completely setting it as was happening before.

I'm going to rework this to do just that, completely pull/write the yaml config for the 'advanced_setting' property, and keep a 'set_yaml_config()' method for modifying configuration.

I'll likely use riggerlib's recursive_update in `set_yaml_config()` so that a partial dictionary tree can be passed and will only modify the intended values.

mfalesni:
@mshriver Cool, though I would rather call the method like `update_yaml_config` or something like that to not confuse its behaviour. :)

mshriver:
Requested review from @izapolsk  and @rlbabyuk as they had approved earlier. Due to the scope of changes in latest commits, bumping myself back down to 1st level.

dajoRH:
*CFME QE Bot*

cfme/base/__init__.py
```diff
@@ -247,23 +249,24 @@ def _api_settings_url(self):
...
                          'settings'])
 
-    def get_yaml_config(self):
+    @property
+    def advanced_settings(self):
         """"GET zones/:id/settings api endpoint to query region configuration"""
         return self.appliance.rest_api.get(self._api_settings_url)
 
-    def set_yaml_config(self, data_dict):
+    def update_advanced_settings(self, settings_dict):
```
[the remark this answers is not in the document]
mshriver: @izapolsk The commit history has been squashed now - but I actually did implement with a setter at first.

This raised the issue of the setter implying that when called, a complete advanced settings tree would be passed. This is never actually the case in any of our uses of advanced settings - we're only ever updating some specific parts of the config tree.

So after discussion with mfalesni/psav above, we agreed on this 'hybrid' kind of implementation, and adding a setter is trivial when a use-case arises for setting a complete new configuration. Or that use-case could just use the existing method, passing a complete settings_dict.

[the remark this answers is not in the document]
izapolsk: gotcha, ty

PR #6974, opened 2018-03-27
[RFR] Removed setup_ansible from miq appliance

quarckster:
Purpose
=================

This code is not needed anymore. Upstream uses AWX which doesn't require any configuring in order to enable embedded ansible role.

http://talk.manageiq.org/t/embedded-ansible-using-awx/3057


jamesooden:
The prt failures appear to be due to your changes.

PR #6976, opened 2018-03-27
[1LP][RFR]Fix some ui tests

rlbabyuk:
Purpose or Intent
=================
1. Corrected timezone format, got incorrest one
2. Updated entities check as floating ips have only address

{{pytest: -k 'test_display_network_topology or test_configure_vmdb_last_start_time'}}

### DIFF
--- a/cfme/tests/configure/test_database_ui.py
+++ b/cfme/tests/configure/test_database_ui.py
@@ -23,6 +23,8 @@ def test_configure_vmdb_last_start_time(appliance):
                 "journalctl -u {}-postgresql.service  --boot=0 | sed '4!d'".format(item))
 
     ui_last_start_time = parser.parse(view.summary('Properties').get_text_of('Last Start Time'))
+    # timedatectl is used here as we will get full timezone name, like 'US/Eastern',
+    #  which is easier and safer(to omit UnknownTimeZoneError) to use later
     tz = pytz.timezone(appliance.ssh_client.run_command("timedatectl | grep 'Time zone'")
                        .output.strip().split(' ')[2])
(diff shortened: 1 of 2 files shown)


cfme/tests/configure/test_database_ui.py
```diff
@@ -23,8 +23,8 @@ def test_configure_vmdb_last_start_time(appliance):
                 "journalctl -u {}-postgresql.service  --boot=0 | sed '4!d'".format(item))
 
     ui_last_start_time = parser.parse(view.summary('Properties').get_text_of('Last Start Time'))
-
-    tz = pytz.timezone(appliance.ssh_client.run_command("date +'%Z'").output.strip())
+    tz = pytz.timezone(appliance.ssh_client.run_command("timedatectl | grep 'Time zone'")
```
jamesooden: **QUESTION:** What was wrong with the original method?   I would have thought that would have been correct (obviously it wasn't because you are changing it...just trying to understand the exact nature of the error).   This might be a place to add a comment about why use timedatectl above other utilities.

rlbabyuk: yup original method worked well in prt run, but after jenkins run we got UnknownTimeZoneError, as it turned out timezone name that we got not worked well for object creation in some cases, with the help of timedatectl we can get full timezone name that much better and safely can be used here

jamesooden: *REQUIRED:* Please add a comment describing what you just described.

rlbabyuk: Added a comment here

PR #6977, opened 2018-03-27
[1LP][RFR] moving cleanup_vm def to common.vm from common.provider

izapolsk:
Someone put cleanup_vm def into common.provider instead of common.vm.

### DIFF
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -957,3 +957,12 @@ class Template(BaseVM, _TemplateMixin):
     def does_template_exist_on_provider(self):
         """Check if template exists on provider itself"""
         return self.provider.mgmt.does_template_exist(self.name)
+
+
+def cleanup_vm(vm_name, provider):
+    try:
+        logger.info('Cleaning up VM %s on provider %s', vm_name, provider.key)
+        provider.mgmt.delete_vm(vm_name)
+    except:
+        # The mgmt_sys classes raise Exception :\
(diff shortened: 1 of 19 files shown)


cfme/common/vm.py
```diff
@@ -957,3 +957,12 @@ def __init__(self, name, provider, template_name=None):
...
         return self.provider.mgmt.does_template_exist(self.name)
+
+
+def cleanup_vm(vm_name, provider):
+    try:
+        logger.info('Cleaning up VM %s on provider %s', vm_name, provider.key)
+        provider.mgmt.delete_vm(vm_name)
+    except:
+        # The mgmt_sys classes raise Exception :\
+        logger.warning('Failed to clean up VM %s on provider %s', vm_name, provider.key)
```
sshveta: Optional: This code can be used in https://github.com/ManageIQ/integration_tests/blob/master/cfme/fixtures/vm_name.py#L14
also .

izapolsk: thanks but  this is small part of huge task and I don't want to deviate because each change requires much time for testing.

bsquizz:
I already have a PR addressing cleanup VM stuff. #6866

cfme/common/vm.py
sshveta: :+1:

izapolsk:
@bsquizz, thanks, I close my PR then.

PR #6978, opened 2018-03-27
[1LP][RFR] Update miq sprout options

mshriver:
Add required options for SproutProvisioningRequest

Log commands that are run by dockerbot setup script

### DIFF
--- a/cfme/scripting/sprout.py
+++ b/cfme/scripting/sprout.py
@@ -41,14 +41,22 @@ def main():
 @click.option('--populate-yaml', is_flag=True, default=False,
               help="Populate the yaml with the appliance")
 @click.option('--provider', default=None, help="Which provider to use")
+@click.option('--provider-type', 'provider_type', default=None,
+              help='A provider type to select from')
+@click.option('--template-type', 'template_type', default=None, help='A template type')
+@click.option('--preconfigured/--notconfigured', default=True,
+              help='Whether the appliance is configured')
 def checkout(appliances, timeout, provision_timeout, group, version, date, desc,
-             override_ram, override_cpu, populate_yaml, provider):
+             override_ram, override_cpu, populate_yaml, provider, provider_type, template_type,
+             preconfigured):
     """checks out a sprout provisioning request, and returns it on exit"""
     override_cpu = override_cpu or None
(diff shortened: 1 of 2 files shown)


cfme/scripting/sprout.py
```diff
@@ -41,14 +41,20 @@ def main():
 @click.option('--populate-yaml', is_flag=True, default=False,
               help="Populate the yaml with the appliance")
 @click.option('--provider', default=None, help="Which provider to use")
+@click.option('--provider_type', default=None, help='A provider type to select from')
```
jamesooden: here you use an underbar as a delimiter and for --template-type you use a -.    Just need to be consistent.   BTW, if you want --provider-type to use the variable provider_type, then your line would look like:

    `@click.option('--provider-type', 'provider_type', default=None, help='A provider type to select from')`

cfme/scripting/sprout.py
```diff
@@ -41,14 +41,20 @@ def main():
...
+@click.option('--template-type', default=None, help='A template type')
```
jamesooden: This should look like:

`@click.option('--template-type', 'template_type', default=None, help='A template type')`

Based on line 48.

cfme/scripting/sprout.py
jamesooden: This is described here:  http://click.pocoo.org/5/parameters/#parameter-names

cfme/scripting/sprout.py
```diff
@@ -41,14 +41,21 @@ def main():
 @click.option('--populate-yaml', is_flag=True, default=False,
               help="Populate the yaml with the appliance")
 @click.option('--provider', default=None, help="Which provider to use")
+@click.option('--provider-type', 'provider_type', default=None,
+              help='A provider type to select from')
+@click.option('--template-type', 'template_type', default=None, help='A template type')
+@click.option('--preconfigured', default=True, help='Whether the appliance is configured')
```
izapolsk: Question: will this option have False if it is used ?

mshriver: @izapolsk Good call, I'll actually add a --notconfigured false boolean, so its an explicit flag.

PR #6988, opened 2018-03-28
[1LP][RFR] Change user testing via REST following BZ fixes

mkoura:
{{pytest: -v -k TestUsersViaREST cfme/tests/configure/test_rest_access_control.py}}

**PRT**
The ``AttributeError: catch_log_handler`` on 5.8 is not related to changes in this PR.


lkhomenk:
one day we might start adding `reason` for uncollected items

psav:
@lkhomenk You bring up a good point. I think from this point on we require reasons.

psav:
Please add reasons @mkoura

PR #5984, opened 2017-11-28
[1LP][RFR] Added test_service_rbac.py testcase

Yadnyawalkya:
{{pytest: cfme/tests/services/test_service_rbac.py --use-template-cache --use-provider vsphere67-nested -vvvv}}

PR #6992, opened 2018-03-28
[1LP][RFR] Disable view.is_display check at the end of Role.update. Fix for BZ156198

LandoCalrizzian:
After saving a role that has it's name updated, the accordion may not update to
show the role under the new name.

[BZ 1561698](https://bugzilla.redhat.com/show_bug.cgi?id=1561698) and fix for #5984 PRT failure

{{ pytest: -vvvv cfme/tests/configure/test_access_control.py -k 'role_crud or test_edit_default_roles' }}

### DIFF
--- a/cfme/configure/access_control/__init__.py
+++ b/cfme/configure/access_control/__init__.py
@@ -1006,12 +1006,13 @@ class Role(Updateable, Pretty, BaseEntity):
         view = self.create_view(DetailsRoleView, override=updates)
         view.flash.assert_message(flash_message)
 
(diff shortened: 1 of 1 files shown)


psav:
Would it not be better to put the page refresh after the Role create?

LandoCalrizzian:
@psav As a workaround, I removed the manual refresh and use a BZ check to disable the assert view.is_displayed. I didn't want to remove the view.is_displayed completely since it's a good check for web ui bugs

PR #7004, opened 2018-03-29
[1LP][RFR] Unifies delete from details REST tests

mkoura:
So far only tests in test_rest.py were converted, other tests will follow in separate PRs.

{{pytest: -v -k test_delete_ cfme/tests/test_rest.py}}

### DIFF
--- a/cfme/tests/test_rest.py
+++ b/cfme/tests/test_rest.py
@@ -18,6 +18,7 @@ from cfme.utils.providers import ProviderFilter
 from cfme.utils.rest import (
     assert_response,
     delete_resources_from_collection,
+    delete_resources_from_detail,
     query_resource_attributes,
 )
 from cfme.utils.wait import wait_for, TimedOutError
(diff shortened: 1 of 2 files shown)


cfme/tests/test_rest.py
```diff
@@ -848,14 +849,7 @@ def test_delete_arbitration_settings_from_detail(self, appliance, arbitration_se
...
         """
-        for setting in arbitration_settings:
-            del_action = getattr(setting.action.delete, method.upper())
-            del_action()
-            assert_response(appliance)
-
-            with pytest.raises(Exception, match='ActiveRecord::RecordNotFound'):
-                del_action()
-            assert_response(appliance, http_status=404)
+        delete_resources_from_detail(arbitration_settings, method)
```
mshriver: **Required**

You need to pass `not_found=True` here (and below) to actually keep the same behavior. The default of `False` means this isn't trying to delete twice anymore, and isn't matching the 'RecordNotFound' exception.

mkoura:
@mshriver :+1: Actually ``not_found`` is left-over that I forgot to remove. We want to always check what the behavior is when deleting non-existent resource. I removed it, thanks for spotting!

PR #7022, opened 2018-04-04
[1LP][RFR] Update dev_guide doc with RC tag/release

mshriver:
First draft.

Documenting RC tagging and release process

### DIFF
--- a/docs/guides/dev_guide.rst
+++ b/docs/guides/dev_guide.rst
@@ -1,5 +1,5 @@
-Style Guide
-===========
+Contributors Guide
+==================
 
 General Guidelines
 ------------------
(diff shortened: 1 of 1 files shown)


docs/guides/dev_guide.rst
```diff
@@ -69,6 +69,47 @@ may, permissions allowing, merge the commit him/herself.
...
+
+The MIQ/integration_tests maintainers will use a two week release schedule, with a release candidate
+(RC) commit tagged on the last Friday in the cycle.  When this RC tag is set, new PRs are not
+accepted for merging unless fixing things that were broken in the current release cycle.
+This 2-3 day period is commonly called the 'dev-freeze'.
+
+Release tags will be created on the following Tuesday, and the downstream-stable branch updated
+to the release tag commit.
+
+Releases are tagged with a version number in the format ``\d{2}\.\d{1,2}\.\d{1}``,
```
jamesooden: **SUGESTION:**  I would just put a + beside the \d because although it is unlikely the first or 2nd digit climb to three digits there is no reason to prevent it programatically, and the first one could also go to two digits.   I think the thing that is important is to show that its three integers separated by periods.   This not actually being in code makes it less worrisome that we were more restrictive than need be in the regex definition, hence I don't feel too strongly about this.

docs/guides/dev_guide.rst
```diff
@@ -69,6 +69,47 @@ may, permissions allowing, merge the commit him/herself.
...
+for example ``17.25.0``.
+
+The release candidate commits will be tagged on Friday with a ``downstream-stable-rc`` tag,
+and a version numbered tag that will match the version number of the *next* release.
+For example, the Friday before ``17.25.0`` is released we create an RC tag ``17.25.0-rc``.
+
+This means we have a tag, ``downstream-stable-rc`` that moves each time an RC commit is selected,
```
jamesooden: I find this statement very helpful.

PR #7027, opened 2018-04-04
[1LP][RFR] Remove 'with pytest.raises' statement in test_permission_edit

LandoCalrizzian:
The initial navigate_to() statement is wrapped in a 'with pytest.raises'
incorrectly (#6994) since the statement is expected to succeed. Added comments
to explain that the navigation should succeed on the initial attempt
with valid permissions

{{ pytest: -v cfme/tests/configure/test_access_control.py -k 'test_permission_edit' }}

### DIFF
--- a/cfme/tests/configure/test_access_control.py
+++ b/cfme/tests/configure/test_access_control.py
@@ -627,7 +627,9 @@ def _test_vm_removal():
          ['Everything', 'Compute', 'Infrastructure', 'Virtual Machines', 'Accordions']]])
 def test_permission_edit(appliance, request, product_features):
     """
-    Ensures that changes in permissions are enforced on next login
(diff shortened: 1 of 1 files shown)


cfme/tests/configure/test_access_control.py
```diff
@@ -627,7 +627,9 @@ def _test_vm_removal():
          ['Everything', 'Compute', 'Infrastructure', 'Virtual Machines', 'Accordions']]])
 def test_permission_edit(appliance, request, product_features):
     """
-    Ensures that changes in permissions are enforced on next login
+    Ensures that changes in permissions are enforced on next login by attempting to navigate to
```
jamesooden: Nice clarification.

PR #7029, opened 2018-04-05
[1LP][RFR] RHV tier analyzer

jan-zmeskal:
This little __utility script__ is to be used by RHV-CFME integration team to compare
-  All the tests run for our provider with
- Tests that are marked with RHV tier 1, 2 or 3.

It's quite useful for us since from time to time we need to assign tiers to new tests. I am not sure if you're okay with having this in your repository. If not, we can store it somewhere else. But I thought this would be the most intuitive location.

PRT: There is SSH exception once again for downstream-58z

Example output:

************************************************************
There are 40 test cases not marked with "-m rhv1 or rhv2 or rhv3":

test_action_create_snapshot_and_delete_last[rhv_cfme_integration]
test_action_create_snapshots_and_delete_them[rhv_cfme_integration]
test_advanced_search_registry_element[rhv_cfme_integration]
test_cockpit_server_role[rhv_cfme_integration-disabled]
test_cockpit_server_role[rhv_cfme_integration-enabled]
test_create_snapshot[rhv_cfme_integration]
test_create_without_description[rhv_cfme_integration]
test_delete_snapshot_from_collection[rhv_cfme_integration]
test_delete_snapshot_from_detail[rhv_cfme_integration-DELETE]
test_delete_snapshot_from_detail[rhv_cfme_integration-POST]
test_delete_snapshot_race[rhv_cfme_integration]
test_drift_analysis[rhv_cfme_integration-rhel-xfs]
test_host_relationships[rhv_cfme_integration-Cluster]
test_host_relationships[rhv_cfme_integration-Datastores]
test_host_relationships[rhv_cfme_integration-Infrastructure Provider]
test_host_relationships[rhv_cfme_integration-Templates]
test_host_relationships[rhv_cfme_integration-VMs]
test_infra_provider_relationships[rhv_cfme_integration-Clusters]
test_infra_provider_relationships[rhv_cfme_integration-Datastores]
test_infra_provider_relationships[rhv_cfme_integration-Hosts]
test_infra_provider_relationships[rhv_cfme_integration-Templates]
test_infra_provider_relationships[rhv_cfme_integration-Virtual Machines]
test_monthly_charges[rhv_cfme_integration-ViaSSUI]
test_myservice_crud[rhv_cfme_integration-ViaSSUI]
test_provider_refresh_relationship[rhv_cfme_integration]
test_reconfig_vm_negative_cancel[rhv_cfme_integration]
test_retire_service[rhv_cfme_integration-ViaSSUI]
test_revert_on_running_vm[rhv_cfme_integration]
test_revert_snapshot[rhv_cfme_integration]
test_service_catalog_crud[rhv_cfme_integration-ViaSSUI]
test_service_start[rhv_cfme_integration-ViaSSUI]
test_snapshot_history_btn[rhv_cfme_integration]
test_ssa_files[rhv_cfme_integration-rhel-xfs]
test_ssa_groups[rhv_cfme_integration-rhel-xfs]
test_ssa_packages[rhv_cfme_integration-rhel-xfs]
test_ssa_schedule[rhv_cfme_integration-rhel-xfs]
test_ssa_template[rhv_cfme_integration-rhel-xfs]
test_ssa_users[rhv_cfme_integration-rhel-xfs]
test_total_services[rhv_cfme_integration-ViaSSUI]
test_vm_clone_neg[rhv_cfme_integration]
************************************************************

{{pytest: cfme/tests --collect-only}}

### DIFF
--- a/scripts/tier_analyzer.py
+++ b/scripts/tier_analyzer.py
@@ -1,5 +1,9 @@
 #!/usr/bin/env python
 
+"""This simple script lists all tests generated for the given provider. Then it lists all the tests
+marked with given tier marker(s). In the end it simply compares those two list, showing you tests
+that are generated for provider but NOT marked with the tier(s)."""
+
 import argparse
 import re
 import subprocess
@@ -7,15 +11,22 @@ import sys
 
 
 def check_virtualenv():
-    """Check if we are in virtualenv and if not, exit."""
+    """Check if we are in virtualenv and if not, raise an error."""
     if not hasattr(sys, 'real_prefix'):
-        print('You must activate CFME virtualenv in oder to run this script.')
-        sys.exit(1)
+        raise EnvironmentError('You must activate CFME virtualenv in oder to run this script.')
 
 
-def get_pytest_output(args, use_tier_marker=False):
-    """Get string output of pytest command."""
-    print('Getting pytest output {} tier marker.'.format('with' if use_tier_marker else 'without'))
+def get_pytest_collect_only_output(args, use_tier_marker=False):
+    """
+    Get a string that is returned by pytest after invoking it with --collect-only argument.
+    If use_tier_marker is set to False (default), all the tests for given provider are collected.
+    If use_tier_marker is set to True, only the marked tests are collected.
+    """
+    print('Getting pytest --collect-only output for provider {} {} tier marker {}.'
+        .format(
+            args.provider,
+            'with' if use_tier_marker else 'without',
+            args.tier_marker))
     pytest_args = ['pytest',
         args.test_path,
         '--collect-only',
@@ -31,12 +42,13 @@ def get_pytest_output(args, use_tier_marker=False):
     return subprocess.check_output(pytest_args)
 
 
-def parse_pytest_output(output):
-    """Parse raw pytest output to get list of parametrized test cases."""
+def get_testcases_from_pytest_output(output):
+    """Parse raw pytest --collect-only output to get list of parametrized test cases."""
     test_cases = []
-    tc_regex = re.compile(r"\s*<Function '(?P<test_case_name>test_\w+)(?P<parameters>\[.+\])?'>")
+    testcase_regex = \
+        re.compile(r"\s*<Function '(?P<test_case_name>test_\w+)(?P<parameters>\[.+\])?'>")
     for line in output.splitlines():
-        match = re.match(pattern=tc_regex, string=line)
+        match = re.match(pattern=testcase_regex, string=line)
         if match:
             test_cases.append('{}{}'.format(
                 match.group('test_case_name'),
@@ -45,7 +57,7 @@ def parse_pytest_output(output):
     return test_cases
 
 
-def compare_parsed_outputs(all, tiers):
+def get_diff_from_lists(all, tiers):
     """Get sorted list of test cases that are not marked with tiers."""
     diff = set(all).difference(set(tiers))
     return sorted(diff)
@@ -75,10 +87,7 @@ def print_message(args, diff, all_parsed, all_count, tiers_parsed, tiers_count):
 
 if __name__ == '__main__':
 
-    parser = argparse.ArgumentParser(description="""This simple script lists all tests generated
-    for the given provider. Then it lists all the tests makred with given tier marker(s).
-    In the end it simply compares thhose two list, showing you tests that are generated for provider
-    but NOT marked with the tier(s).""")
+    parser = argparse.ArgumentParser(description=__doc__)
     parser.add_argument('-v', '--verbose', action='store_true')
     parser.add_argument('-p', '--provider', action='store', default='rhv_cfme_integration')
     parser.add_argument('-t', '--test-path', action='store', default='cfme/tests')
@@ -87,14 +96,14 @@ if __name__ == '__main__':
 
     check_virtualenv()
 
-    output_with_tiers = get_pytest_output(args, use_tier_marker=True)
-    output_without_tiers = get_pytest_output(args, use_tier_marker=False)
-    tiers_parsed = parse_pytest_output(output_with_tiers)
-    all_parsed = parse_pytest_output(output_without_tiers)
+    output_with_tiers = get_pytest_collect_only_output(args, use_tier_marker=True)
+    output_without_tiers = get_pytest_collect_only_output(args, use_tier_marker=False)
+    tiers_parsed = get_testcases_from_pytest_output(output_with_tiers)
+    all_parsed = get_testcases_from_pytest_output(output_without_tiers)
 
     tiers_count = len(tiers_parsed)
     all_count = len(all_parsed)
 
-    final_diff = compare_parsed_outputs(all_parsed, tiers_parsed)
+    final_diff = get_diff_from_lists(all_parsed, tiers_parsed)
 
     print_message(args, final_diff, all_parsed, all_count, tiers_parsed, tiers_count)
--- a/scripts/tier_analyzer.py
+++ b/scripts/tier_analyzer.py
@@ -1,0 +1,100 @@
+#!/usr/bin/env python
+
+import argparse
+import re
+import subprocess
+import sys
+
+
+def check_virtualenv():
+    """Check if we are in virtualenv and if not, exit."""
+    if not hasattr(sys, 'real_prefix'):
+        print('You must activate CFME virtualenv in oder to run this script.')
+        sys.exit(1)
+
+
+def get_pytest_output(args, use_tier_marker=False):
+    """Get string output of pytest command."""
+    print('Getting pytest output {} tier marker.'.format('with' if use_tier_marker else 'without'))
+    pytest_args = ['pytest',
+        args.test_path,
+        '--collect-only',
+        '--long-running',
+        '--use-provider',
+        args.provider,
+        '-m']
+    if use_tier_marker:
+        pytest_args.append(args.tier_marker)
+    else:
+        pytest_args.append('uses_testgen')
+
+    return subprocess.check_output(pytest_args)
+
+
+def parse_pytest_output(output):
+    """Parse raw pytest output to get list of parametrized test cases."""
+    test_cases = []
+    tc_regex = re.compile(r"\s*<Function '(?P<test_case_name>test_\w+)(?P<parameters>\[.+\])?'>")
+    for line in output.splitlines():
+        match = re.match(pattern=tc_regex, string=line)
+        if match:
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit 7e0353eb77846235eb817c8bfbb138d52548b190:

`scripts/rhv_tier_analyzer.py`:

**No commit flaws detected**.

*CFME QE Bot*

scripts/rhv_tier_analyzer.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def parse_pytest_output(output):
+    """Parse raw pytest output to get list of parametrized test cases."""
+    list_of_lines = output.splitlines()
+    list_of_functions = [l.strip() for l in list_of_lines if l.strip().startswith('<Function ')]
+    integration_only = [l for l in list_of_functions if PROVIDER in l]
+
+    test_cases = []
+    for func in integration_only:
```
mshriver: **Optional**

Consider using a regex expression here with a named matching group to pull out the test case string here.

scripts/rhv_tier_analyzer.py
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env python
+
+import argparse
+import subprocess
+import sys
+
+PROVIDER = 'rhv_cfme_integration'
```
mshriver: **Required**

Make these CLI args, so that the script has some more flexibility/function to it - we can then run it against generic tier markers, and any provider we want.

Since PROVIDER is passed to `pytest` call, it can be passed on CLI as anything that matches for that use  - a provider_key or a tag.

scripts/rhv_tier_analyzer.py
```diff
@@ -0,0 +1,92 @@
...
+    """Check if we are in virtualenv and if not, exit."""
+    if not hasattr(sys, 'real_prefix'):
+        print('You must activate CFME virtualenv in oder to run this script.')
+        sys.exit(1)
+
+
+def get_pytest_output(use_tier_marker=False):
+    """Get string output of pytest command."""
+    print('Getting pytest output {} tier marker.'.format('with' if use_tier_marker else 'without'))
+    args = ['pytest', TEST_PATH, '--collect-only', '--long-running', '--use-provider', PROVIDER]
```
mshriver: One way to further limit the collection here is to concatenate '-m uses_testgen' into the markers list, so that only test cases parameterized on providers are collected. This may also reduce how many functions are parsed out in the function below, or eliminate the need altogether to first filter by PROVIDER in the line.

mshriver:
Very close to approving, but figured let's make it more flexible :) Well done!

dajoRH:
Lint report for commit 5c949723502dcb471a651e63e29d41e6a566665d:

`scripts/rhv_tier_analyzer.py`:
- :bangbang: Line 44:17: **E123** *closing bracket does not match indentation of opening bracket's line*
- :bangbang: Line 80:50: **E251** *unexpected spaces around keyword / parameter equals*

**No commit flaws detected**.

*CFME QE Bot*

mshriver:
@jan-zmeskal Thanks - absolutely, it wasn't my intention to have you rework it to be library-level, just some tweaks to set us up so that its easy to adapt, as you've done.

scripts/rhv_tier_analyzer.py
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env python
```
jamesooden: **REQUIRED:** Add a docstring describing this script.   No implementation details just at a high level what it is supposed to do and why.

scripts/tier_analyzer.py
```diff
@@ -0,0 +1,100 @@
...
+    sep()
+    print('There are {} test cases not marked with "{}":\n'.format(len(diff), args.tier_marker))
+    for test_case in diff:
+        print(test_case)
+    sep()
+
+
+if __name__ == '__main__':
+
+    parser = argparse.ArgumentParser(description="""This simple script lists all tests generated
```
jan-zmeskal: @jamesooden There is some simple description. Is that sufficient?

scripts/tier_analyzer.py
```diff
@@ -0,0 +1,100 @@
...
+    for the given provider. Then it lists all the tests makred with given tier marker(s).
```
jamesooden: **REQUIRED:** mispelled marked (see makred).

scripts/tier_analyzer.py
```diff
@@ -0,0 +1,100 @@
...
+    In the end it simply compares thhose two list, showing you tests that are generated for provider
```
jamesooden: **REQUIRED:** those misspelled (see thhose)

scripts/tier_analyzer.py
```diff
@@ -0,0 +1,100 @@
...
+import re
+import subprocess
+import sys
+
+
+def check_virtualenv():
+    """Check if we are in virtualenv and if not, exit."""
+    if not hasattr(sys, 'real_prefix'):
+        print('You must activate CFME virtualenv in oder to run this script.')
+        sys.exit(1)
```
jamesooden: **OPTIONAL:** Instead of exiting consider throwing an exception (which uncaught will exit but with more diagnostic information).   See scripts/coverage_report_jenkins.py as an example.

scripts/tier_analyzer.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def get_pytest_output(args, use_tier_marker=False):
```
jamesooden: **REQUIRED:** This function is not as generic as the name implies.   Change the name to something more specific such as get_pytest_collect_only_output().

scripts/tier_analyzer.py
```diff
@@ -0,0 +1,100 @@
...
+    """Get string output of pytest command."""
```
jamesooden: **REQUIRED:** The docstring needs to be more specific this is not just generic output of just any pytest command.   It is pytest --collect-only output...and specifically for the purpose of finding out what tests would run for the specified provider.   All this information need to be included in the

scripts/tier_analyzer.py
```diff
@@ -0,0 +1,100 @@
...
+    print('Getting pytest output {} tier marker.'.format('with' if use_tier_marker else 'without'))
```
jamesooden: **REQUIRED:** Change "pytest output" to "pytest collect-only output".   Also add to the end "for provider  {}"

scripts/tier_analyzer.py
```diff
@@ -0,0 +1,100 @@
...
+        '-m']
+    if use_tier_marker:
+        pytest_args.append(args.tier_marker)
+    else:
+        pytest_args.append('uses_testgen')
+
+    return subprocess.check_output(pytest_args)
+
+
+def parse_pytest_output(output):
```
jamesooden: **REQUIRED:** This is not generically parsing pytest output.   It is parsing out the parameterized test cases.   The name should change to reflect that, such as parse_testcases_from_pytest_output, or get_testcases_from_pytest_output.

Just something to think about, you could even change to get_pytest_testcases_for_provider(), and then call the function that runs pytest and gets the output from here.   That way your main function just calls this function.   It's fine the way it is, but it would make the main function show less knowledge about the details and only at a high level what it is doing, which is getting the test cases for a particular provider with and without filtering on a tier.

scripts/tier_analyzer.py
```diff
@@ -0,0 +1,100 @@
...
+    """Parse raw pytest output to get list of parametrized test cases."""
+    test_cases = []
+    tc_regex = re.compile(r"\s*<Function '(?P<test_case_name>test_\w+)(?P<parameters>\[.+\])?'>")
```
jamesooden: **OPTIONAL:** Consider changing tc_regex to testcase_regex.   It will be much more apparent what it is and it won't make any lines too long.

scripts/tier_analyzer.py
jamesooden: And you could also argue that your current way is simpler.   I'm just offering a different perspective of simple.

scripts/tier_analyzer.py
```diff
@@ -0,0 +1,100 @@
...
+    for line in output.splitlines():
+        match = re.match(pattern=tc_regex, string=line)
+        if match:
+            test_cases.append('{}{}'.format(
```
jamesooden: **QUESTION:** Do we want a delimiter between the test case name and its parameters for easy of seeing which is which?

scripts/tier_analyzer.py
```diff
@@ -0,0 +1,100 @@
...
+                match.group('test_case_name'),
+                match.group('parameters') if match.group('parameters') else ''))
+
+    return test_cases
+
+
+def compare_parsed_outputs(all, tiers):
```
jamesooden: **REQUIRED:** This really is just a straight diff of two lists.   It should have a more generic name, like compare_lists().

jan-zmeskal:
Thank you for insightful comments, @jamesooden, I will improve the script along those lines

scripts/tier_analyzer.py
jamesooden: Well, my preference is that all files start with a docstring at the top to explain what they are there for.   This servers two purposes.   The first is, the first question that comes to mind when you open a file is answered, which is what is this thing?   The second is docstrings can be picked up as actual documentation, whereas your argparser description will not provide for that.

scripts/tier_analyzer.py
jan-zmeskal: @jamesooden I understand your argument and I don't even want to make the case that my current implementation is better. I just think that it is not very defensible from the ROI point of view to put so much effort into design of such a narrow utility script. That's why I'd like to go with what I have. Not because it is necessarily better, but because it's not really incorrect and gets the job done.

scripts/tier_analyzer.py
jan-zmeskal: @jamesooden I don't think so. The advantage of current output  is that you can easily just copy&paste it and then search for it.

jan-zmeskal:
@jamesooden, @mshriver Pls review once again - I implemented most of the stuff suggested by James

scripts/tier_analyzer.py
jamesooden: I mainly wanted to pass down a design principle of keeping implementation details hidden and providing abstractions that keep the user of the abstraction, and in this case the reader of the code focused on the problem to be solved rather than the details of how it is being solved.    So mission accomplished.

scripts/tier_analyzer.py
jamesooden: Got it.

PR #7030, opened 2018-04-05
[1LP][RFR] Marking TestRESTSnapshots with RHV tiers

jan-zmeskal:
PRT: SSH exception

{{pytest: cfme/tests/cloud_infra_common/test_snapshots_rest.py::TestRESTSnapshots --long-running --use-provider rhv41 -m "rhv1 or rhv2 or rhv3"}}


istein1:
Not sure why the REST delete snapshot is set to Tier3, while revert and create are set to Tier2,
Seems to me all those should be included under same Tier (2 or 3).

jan-zmeskal:
@istein1 Well my idea behind this was that being unable to delete snapshot would have much lower impact on the user as opposed to the inability to create or revert to it

PR #7031, opened 2018-04-05
[1LP][RFR] Add rest bind_timeout test for BZ

mshriver:
modify advanced settings authentication:bind_timeout value via REST, for BZ 1553394
currently BZ active and blocking

## PRT
58z will be ignored
59z will be BZ skip
Failures are as expected in 22951

I've run the test locally without the BZ skip, and it fails as expected due to the active BZ.

{{ pytest: cfme/tests/configure/test_session_timeout.py -k test_bind_timeout_rest }}

### DIFF
--- a/cfme/tests/configure/test_session_timeout.py
+++ b/cfme/tests/configure/test_session_timeout.py
@@ -3,6 +3,7 @@ import time
 
 import pytest
 
+from cfme.utils.blockers import BZ
 from cfme.utils.browser import ensure_browser_open, quit
 from cfme.utils.wait import wait_for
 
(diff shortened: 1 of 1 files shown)


cfme/tests/configure/test_session_timeout.py
```diff
@@ -35,3 +36,22 @@ def _finalize():
...
         )
     )
+
+
+@pytest.mark.tier(3)
+@pytest.mark.ignore_stream('5.8')  # Modifying settings via rest in 5.9+
+@pytest.mark.meta(blockers=[BZ(1553394)])
+def test_bind_timeout_rest(appliance, request):
+    """Sets the session timeout via REST
+        Written for BZ 1553394
```
mkoura: OPTIONAL: strange indentation, there should be a blank line between summary line and the rest of the docstring. See https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

mshriver: @quarckster pointed this out in another PR, I had recently re-installed and it seems I need to give pycharm settings a beatdown.

PR #7032, opened 2018-04-05
[1LP][RFR] Remove thing_toucher.rb  and its use.

jamesooden:
thing_toucher.rb was designed to ensure that all source files that
comprised CFME would show up in the coverage reports every time even
if they were not used.   At some point it failed to work because
it was including a library that went into a debugger when you included
it (lib/vmdb/gettext/debug.rb).   After excluding that file it was found that few
libraries would successfully get included such that there really
was no point in running thing_toucher.rb.

This commit removes the use of thing_toucher.rb.   A future
commit may come at the problem in a different way (e.g. surgical edits
of the .resultset.json file on result set merger).


jamesooden:
PRT errors not related to these changes.

mshriver:
20180406165742-a31e0b5bdb gif-2-mp4 com

PR #7033, opened 2018-04-05
[1LP][RFR] Marikng test_cockpit_server_role with RHV tier

jan-zmeskal:
{{pytest: cfme/tests/cloud_infra_common/test_cockpit_server.py --use-provider rhv41 -m "rhv3" -vv}}


istein1:
Ack for the tier marking. 
Is the cockpit test stable for RHV? If I am not mistaken it requires cockpit package installed on the template. thanks.

jan-zmeskal:
@istein1 I was running it locally and it looked okay. Anyway if I understand this test case correct, I does not actually tests the cockpit itself, just the button.

PR #7034, opened 2018-04-05
[1LP][RFR] Marking test_relationships with RHV tiers

jan-zmeskal:
{{pytest: cfme/tests/cloud_infra_common/test_relationships.py --use-provider rhv41 -m "rhv1 or rhv2 or rhv3" -vv}}


jan-zmeskal:
@istein1 Pls provide your opinion on the tiers

PR #7038, opened 2018-04-05
[1LP][RFR] Fix failure in test_edit_sequence_usergroups for 5.9.2

LandoCalrizzian:
Do a version check for specifying the column name in the RBAC Group checklist changed from "Name" to "Description".

{{ pytest: -v cfme/tests/configure/test_access_control.py -k 'test_edit_sequence_usergroups' }}

### DIFF
--- a/cfme/configure/access_control/__init__.py
+++ b/cfme/configure/access_control/__init__.py
@@ -700,7 +700,11 @@ class Group(BaseEntity, Taggable):
         Args:
             updated_order: group order list
         """
-        name_column = "Name"
+        if self.appliance.version < "5.9.2":
+            name_column = "Name"
+        else:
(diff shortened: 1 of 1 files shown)


cfme/configure/access_control/__init__.py
```diff
@@ -700,7 +700,11 @@ def set_group_order(self, updated_order):
         Args:
             updated_order: group order list
         """
-        name_column = "Name"
+        if self.appliance.version < "5.9.2":
```
mshriver: **Optional**

Express as ternary, might be over 100 chars at this indent though.

```python
name_column = 'Name' if self.appliance.version < '5.9.2' else 'Description'
```

PR #7048, opened 2018-04-09
[1LP][RFR] cleanup_edomain_templates, handle API exception

mshriver:
There is an exception raised during creation of the API object when the credentials are invalid, this was occuring when cleanup_templates was called with provider_mgmt.api passed.

Make this call in the try block that's already supposed to be checking for API handler.

Move the print statement for the provider start to before we actually start trying to connect to the provider.

## PRT
PRT results are void here unless I've managed to make a colossal error.

The script is currently in-use by CFME QE Jenkins, and runs successfully, logging the API failure to a provider with incorrect credentials.

### DIFF
--- a/scripts/cleanup_edomain_templates.py
+++ b/scripts/cleanup_edomain_templates.py
@@ -294,25 +294,31 @@ def run(**kwargs):
         if cli_provider and cli_provider != provider:
             continue
 
+        print("\n--------Start of {}--------".format(provider))
+
         provider_mgmt = get_mgmt(provider)
 
-        if not net.is_pingable(provider_mgmt.kwargs.get('ipaddress')):
-            continue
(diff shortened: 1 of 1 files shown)


scripts/cleanup_edomain_templates.py
```diff
@@ -294,25 +294,31 @@ def run(**kwargs):
         if cli_provider and cli_provider != provider:
             continue
 
-        provider_mgmt = get_mgmt(provider)
+        print("\n--------Start of {}--------".format(provider))
```
jamesooden: **OPTIONAL:** Make this message more specific to what its doing so something like:   Start of {} template cleanup

jamesooden: The principle here is to give a reasonable amount of context so the reader will of the output will know what is going on at a high level.

PR #7050, opened 2018-04-10
[1LP][RFR] Added new ansible credentials types

quarckster:
Purpose
=================

CFME 5.9.2.1 has added new types of ansible credentials: vault and red hat virtualization.

{{pytest: -v -k test_embedded_ansible_credential_crud --long-running}}


mshriver:
Adding tech-debt label as this does not include a parameter for the vault credential type.

PR #7059, opened 2018-04-10
[1LP][RFR] Fix provider create for RHV and CFME < 5.8.0.8

jan-zmeskal:
The ability to provide TLS certificate for RHV provider was added to CFME only in version 5.8.0.8. That's why creation of provider will fail on older versions -> the framework is not able to fill in the provider creation form. This PR __fixes__ that by removing TLS information from endpoint if appliance is < 5.8.0.8.

PRT fail: SSH exception

{{pytest: cfme/tests/infrastructure/test_providers.py::test_infra_provider_crud --use-provider rhv41}}

### DIFF
--- a/cfme/infrastructure/provider/rhevm.py
+++ b/cfme/infrastructure/provider/rhevm.py
@@ -18,6 +18,12 @@ class RHEVMEndpoint(DefaultEndpoint):
                 'verify_tls': getattr(self, 'verify_tls', None),
                 'ca_certs': getattr(self, 'ca_certs', None)}
 
+    def delete_tls_info(self):
+        if getattr(self, 'ca_certs', None):
+            delattr(self, 'ca_certs')
+        if getattr(self, 'verify_tls', None):
+            delattr(self, 'verify_tls')
+
 
 class RHEVMEndpointForm(View):
     @View.nested
@@ -55,6 +61,7 @@ class RHEVMProvider(InfraProvider):
     _ctrl_alt_del_xpath = '//*[@id="ctrlaltdel"]'
     _fullscreen_xpath = '//*[@id="fullscreen"]'
     bad_credentials_error_msg = "Credential validation was not successful"
+    tls_since_version = '5.8.0.8'
 
     ems_events = [
         ('vm_create', {'event_type': 'USER_ADD_VM_FINISHED_SUCCESS', 'vm_or_template_id': None}),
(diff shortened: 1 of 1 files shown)


cfme/infrastructure/provider/rhevm.py
```diff
@@ -18,6 +18,12 @@ def view_value_mapping(self):
                 'verify_tls': getattr(self, 'verify_tls', None),
                 'ca_certs': getattr(self, 'ca_certs', None)}
 
+    def delete_tls_info(self):
```
mshriver: First impression its a little odd looking, but I much prefer this to having provider-specific modifications in the `BaseProvider.create()` method. This means a provider instance will have a default endpoint with key/value pairs relevant for the appliance version, and view_value_mapping includes them when they're there.

cfme/infrastructure/provider/rhevm.py
```diff
@@ -73,13 +80,14 @@ def __init__(self, name=None, endpoints=None, zone=None, key=None, hostname=None
...
         if ip_address:
             self.ip_address = ip_address
+        if self.appliance.version < self.tls_since_version:
+            self.endpoints['default'].delete_tls_info()
 
     @property
     def view_value_mapping(self):
         return {
             'name': self.name,
-            'prov_type': version.pick({version.LOWEST: 'Red Hat Virtualization Manager',
```
jamesooden: **QUESTION:** Are we no longer supporting versions that said 'Red Hat Virtualization Manager'?

jan-zmeskal: I was recently testing 5.7.4.3 and there is 'Red Hat Virtualization' so I suppose it was backported. Because 5.7 is as old as it gets.

mshriver: Note the framework is only being written to support 58z and 59z  (and of course MIQ) at this time, so if 58z+ has `Red Hat Virtualization` then its fine to remove the version_pick.

In this case it looks like the change was made in a specific version of CFME - 5.8.0.10, and thus 5.8.0.9 would still use `Red Hat Virtualization Manager`.

I think this version pick should be left as-is for now, as its not required for what you're trying to resolve here with TLS configuration.


Note in general this fix, in the master branch, is really relevant for testing CFME 5.8.0.0+, and thus this fix will only have an effect when running against these older versions _still greater than 5.8.0.0_.  If your intent is to test 5.7z, and this fix is being added to support that, its slightly misguided and will need to be backported to the `5.8-breakpoint-patched` branch, which is run when testing 57z.

Since the version you trigger off of is in the 58z stream, the fix is still relevant here - it just won't be used (from master branch) when testing 57z.

PR #7060, opened 2018-04-10
[1LP][RFR] Update manageiq-client to 0.4.1

mkoura:
Fixes issue with ``None`` attributes - https://github.com/ManageIQ/manageiq-api-client-python/pull/39

{{pytest: -v -k test_attributes_present cfme/tests/test_rest.py}}


mkoura:
@vatsalparekh ^^

PR #7062, opened 2018-04-11
[RFR][1LP] Added new version picking to scripts

lcouzens:
Upgrade script extension to allow users to upgrade appliances from 5.7+ to a specific version they require.

### DIFF
--- a/cfme/scripting/appliance.py
+++ b/cfme/scripting/appliance.py
@@ -35,10 +35,15 @@ def main():
 @main.command('upgrade', help='Upgrades an appliance to latest Z-stream')
 @click.argument('appliance-ip', default=None, required=False)
 @click.option('--cfme-only', is_flag=True, help='Upgrade cfme packages only')
-@click.option('--update-to', default='5.9.z', help='Must be 5.8.z or (5.9.z is default)')
+@click.option('--update-to', default='5.9.z', help='Supported versions 5.8.1, 5.8.2, 5.8.z,'
+              ' 5.9.1, 5.9.2, 5.9.z (.z means latest and default is 5.9.z)')
 def upgrade_appliance(appliance_ip, cfme_only, update_to):
     """Upgrades an appliance"""
-    supported_version_repo_map = {'5.8.z': 'update_url_58', '5.9.z': 'update_url_59'}
+    supported_version_repo_map = {
+        '5.8.z': 'update_url_58', '5.8.0': 'update_url_580', '5.8.1': 'update_url_581',
+        '5.8.2': 'update_url_582', '5.9.z': 'update_url_59', '5.9.0': 'update_url_590',
+        '5.9.1': 'update_url_591', '5.9.2': 'update_url_592'
+    }
     assert update_to in supported_version_repo_map, "{} is not a supported version".format(
         update_to
     )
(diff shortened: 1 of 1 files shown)


cfme/scripting/appliance.py
```diff
@@ -35,10 +35,15 @@ def main():
...
 @click.argument('appliance-ip', default=None, required=False)
 @click.option('--cfme-only', is_flag=True, help='Upgrade cfme packages only')
-@click.option('--update-to', default='5.9.z', help='Must be 5.8.z or (5.9.z is default)')
+@click.option('--update-to', default='5.9.z', help='Supported versions 5.8.1, 5.8.2, 5.8.z,'
+              ' 5.9.1, 5.9.2, 5.9.z (.z means latest and default is 5.9.z)')
 def upgrade_appliance(appliance_ip, cfme_only, update_to):
     """Upgrades an appliance"""
-    supported_version_repo_map = {'5.8.z': 'update_url_58', '5.9.z': 'update_url_59'}
+    supported_version_repo_map = {
+        '5.8.z': 'update_url_58', '5.8.0': 'update_url_580', '5.8.1': 'update_url_581',
```
jamesooden: **SUGESTION:**  This information and the information on 38 could probably go into cfme_data.yaml.   It's information that will change often and probably should be in a config file rather than hard coded here.

cfme/scripting/appliance.py
```diff
@@ -48,6 +53,7 @@ def upgrade_appliance(appliance_ip, cfme_only, update_to):
     else:
         print('Fetching appliance from env.local.yaml')
     app = get_appliance(appliance_ip)
+    assert app.version > '5.7', "{} is not supported, must be 5.7 or higher".format(app.version)
```
jamesooden: **SUGESTION:** This too could go into a config, and should not be hard coded.   Also, I would even argue that the Appliance object itself should know what its minimally supported version is.

cfme/scripting/appliance.py
lcouzens: What exactly do you suggest for this? I have created sections for current/future releases here, https://gitlab.cee.redhat.com/cfme-qe/cfme-qe-yamls/merge_requests/483/diffs

In terms of it changing often is not the problem exactly, the whole migration procedure could change for each build so that is going to potentially need updating regardless of the build versions.

cfme/scripting/appliance.py
lcouzens: Again suggestions exactly what you mean would help, but basically its currently set to 5.7 as that had a major postgres upgrade, older versions need a lot more TLC to be coded, just easier to have a cut off currently. (May change in the future) In terms of the 'Appliance object' knowing its supported versions, this can again change depending on the builds and what _we_ want to actually support/implement, I have just added things that people always request my help with.

cfme/scripting/appliance.py
jamesooden: The actual versions should be in the config such that the whole supported_version_repo_map is simply read in from config.   Does that make sense?

cfme/scripting/appliance.py
jamesooden: Not objecting to a cut off, but suggesting there be some minimimal_cfme_version in the cfme_data.yaml.    It would be best if the Appliance class itself had its own minimal_supported_version() class method that read this value from the config...but just having the minimally supported CFME version in the config rather than the code is much better.

cfme/scripting/appliance.py
psav: This does make sense, I think though it can be a secondary PR

cfme/scripting/appliance.py
psav: This version specificity should be made in the framework and not in the yaml. If it is in the yaml, it means someone can alter the "supported" version. We don't want that. Again, until we have that this specified in the Appliance object, we can do this work later, but it is a very good suggestion.

PR #7065, opened 2018-04-11
[1LP][RFR] RHV tiers for several tests

jan-zmeskal:
RHV tier markers for several test cases

### DIFF
--- a/cfme/tests/infrastructure/test_snapshot.py
+++ b/cfme/tests/infrastructure/test_snapshot.py
@@ -243,6 +243,7 @@ def test_revert_active_snapshot(full_test_vm, provider, soft_assert, register_ev
                            active_snapshot=True)
 
 
+@pytest.mark.rhv3
 @pytest.mark.uncollectif(lambda provider: not provider.one_of(RHEVMProvider))
 @pytest.mark.meta(automates=[BZ(1375544)])
 def test_revert_on_running_vm(small_test_vm):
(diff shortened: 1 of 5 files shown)


cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -243,6 +243,7 @@ def test_revert_active_snapshot(full_test_vm, provider, soft_assert, register_ev
                            active_snapshot=True)
 
 
+@pytest.mark.rhv3
```
istein1: I think this should be tier 2

cfme/tests/infrastructure/test_vm_clone.py
```diff
@@ -57,6 +57,7 @@ def test_vm_clone(appliance, provider, clone_vm_name, create_vm):
     assert request_row.is_succeeded(method='ui'), msg
 
 
+@pytest.mark.rhv3
```
istein1: The functionality of clone VM (create a VM from existing one) exist on RHV, but do not exist on CFME.
Not sure why this test is run for RHV provider.

cfme/tests/infrastructure/test_snapshot.py
jan-zmeskal: @istein1 Actually here I have to disagree. If you check the description, it just checks one button for being clickable. I cannot imagine such test anywhere out of tier3.

cfme/tests/infrastructure/test_vm_clone.py
istein1: Great, thanks for explaining

cfme/tests/infrastructure/test_snapshot.py
istein1: After discussion on what this test do offline, I agree it should be tier 3.

PR #7071, opened 2018-04-12
[1LP][RFR] Provisioning dialogs collections conversion

quarckster:
{{pytest: -v cfme/tests/automate/test_provisioning_dialogs.py}}

### DIFF
--- a/cfme/automate/provisioning_dialogs.py
+++ b/cfme/automate/provisioning_dialogs.py
@@ -1,4 +1,6 @@
 # -*- coding: utf-8 -*-
+import attr
+
 from navmazing import NavigateToSibling, NavigateToAttribute
 from widgetastic.widget import View, Text, TextInput
 from widgetastic_patternfly import Dropdown, BootstrapSelect, CandidateNotFound
@@ -6,7 +8,7 @@ from widgetastic_manageiq import Button, Table, PaginationPane, SummaryForm, Scr
 
 from cfme.base.login import BaseLoggedInPage
 from cfme.base.ui import automate_menu_name
-from cfme.utils.appliance import Navigatable
+from cfme.modeling.base import BaseCollection, BaseEntity
 from cfme.utils.appliance.implementations.ui import navigator, navigate_to, CFMENavigateStep
 from cfme.utils.pretty import Pretty
 from cfme.utils.update import Updateable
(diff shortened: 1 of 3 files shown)


cfme/automate/provisioning_dialogs.py
```diff
@@ -182,7 +156,7 @@ def update(self, updates, cancel=False, reset=False):
             btn = view.form.save
 
         btn.click()
-        view = self.create_view(ProvDiagDetailsView)
+        view = self.create_view(ProvDiagDetailsView, override=updates)
         # TODO use override in create_view in order to assert view.is_displayed
```
jamesooden: **REQUIRED:** This is no longer a todo.   Might want to still have a comment on why you wanted to override updates.

cfme/automate/provisioning_dialogs.py
```diff
@@ -118,25 +120,14 @@ class form(ProvDiagForm):  # noqa
...
+@attr.s
+class ProvisioningDialog(Updateable, Pretty, BaseEntity):
     pretty_attrs = ['name', 'description', 'diag_type', 'content']
 
-    def __init__(self, diag_type, name=None, description=None, content=None, appliance=None):
-        Navigatable.__init__(self, appliance=appliance)
-        self.name = name
-        self.description = description
-        self.content = content
-        if diag_type in self.ALLOWED_TYPES:
```
jamesooden: **REQUIRED:** I do not believe we should use this validation.   It is happening in the right place (within the class that contains the attribute, and it provides useful information if someone used something other than one of these ALLOWED_TYPES.   This shows that you can define a validator for an attribute, so you might be able to use that somehow:   http://www.attrs.org/en/stable/api.html#attr.Attribute

quarckster: Done

cfme/automate/provisioning_dialogs.py
```diff
@@ -180,11 +159,9 @@ def update(self, updates, cancel=False, reset=False):
             flash_msg = ('Dialog "{}" was saved'.format(updates.get('description') or
                                                         self.description))
             btn = view.form.save
-
+            view = self.create_view(ProvDiagDetailsView, override=updates)
```
mshriver: **Required**

Two test failures in PRT from this - override should only be passed if `cancel == False`, otherwise the is_displayed check fails because the updates were not saved.

quarckster: Fixed.

dajoRH:
*CFME QE Bot*

PR #7074, opened 2018-04-12
[1LP][RFR] Remove version check from RHEVMProvider constructor

mshriver:
was forcing appliance version check on providers.get_crud() which breaks scripts that don't actually need an appliance to interact with providers.

Moved ca_cert and tls_verify picks into the view_value_mapping

Local testing on 5.7.4.3 and 5.9.2.2 was good.

## PRT Results
Run 23135: 
* 58z did not receive appliance from sprout
* 59z some smoke test failures, on scvmm providers, not related.         
Restarted with only rhv41 provider to get cleaner results that are all relevant to changes.

{{ pytest: -k test_infra_provider_crud --use-provider rhv41 }}

### DIFF
--- a/cfme/infrastructure/provider/rhevm.py
+++ b/cfme/infrastructure/provider/rhevm.py
@@ -13,16 +13,16 @@ from . import InfraProvider
 class RHEVMEndpoint(DefaultEndpoint):
     @property
     def view_value_mapping(self):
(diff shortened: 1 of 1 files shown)


mshriver:
Added 1LP since I've got some approvals already.

PR #7088, opened 2018-04-13
[RFR] gce doesn't support vms with underscore in name

izapolsk:
{{pytest: --use-provider=gce_central cfme/tests/cloud_infra_common/test_events.py}}


mshriver:
failure in PRT run 23200 is unrelated

PR #7090, opened 2018-04-13
[WIPTEST] First pass at removing management_systems from cfme_data

mshriver:
Creating new provider_data.yaml to hold management_systems, eventually configuration_managers as well and anything else that's defining a CFME provider

Touches a whole bunch of modules across the framework and tests.

Did some slight refactoring as I went, in places where the cfme_data file should not have been queried
Removed provider_data support from template_upload_all as this effectively replaces that. local yaml files should be used to get the correct behavior from the script, as with all other yaml files

{{ pytest: --long-running -v --use-provider complete }}


scripts/cleanup_edomain_templates.py
```diff
@@ -12,7 +12,7 @@
 from threading import Lock, Thread
 
 from cfme.utils import net
-from cfme.utils.conf import cfme_data, credentials
+from cfme.utils.conf import cfme_data, credentials, provider_data as provider_yaml
```
jamesooden: **QUESTION:** why as provider_yaml?

mshriver: ah good call - I had done this in a few places where provider_data was in use, not necessary here. but I would import as a different name to avoid using the same thing as a name already in use in the script.

jamesooden: **SUGGESTION:** Add a comment explaining that.

It's all beautiful though.

mshriver: I dropped this, and have done a bit of logging refactor in the script, as well as just importing the whole conf module.

dajoRH:
*CFME QE Bot*

PR #7097, opened 2018-04-17
[1LP][RFR] Update tests using custom users

mshriver:
We are seeing test failures that surface as a ValueError, when trying to navigate (primarily with the 'Configuration' dropdown).

Debugging with @psav we found that some tests were specifically logging in with an `appliance.server.login()` command, both within and outside of user contexts.

This was causing interaction between tests, where some tests wouldn't "cleanup" and return to superadmin login for the appliance. Subsequent tests would be on a non-admin user, and would be missing permissions for some action required for the test.

This PR addresses a few actions across a few different tests:
1. Replace uses of direct login calls with a specific user with a `navigate_to()` call
2. Have a context manager for any custom users

## PRT
running...

{{ pytest: --long-running -k 'test_user_assign_multiple_groups or test_permissions or test_user_change_password or test_group_roles or test_login_evm_group or test_login_retrieve_group' }}

### DIFF
--- a/cfme/configure/access_control/__init__.py
+++ b/cfme/configure/access_control/__init__.py
@@ -95,10 +95,7 @@ class DetailsUserView(ConfigurationView):
     def is_displayed(self):
         return (
             self.title.text == 'EVM User "{}"'.format(self.context['object'].name) and
-            self.accordions.accesscontrol.is_opened and
-            # tree.currently_selected returns a list of strings with each item being the text of
-            # each level of the accordion. Last element should be the User's name
-            self.accordions.accesscontrol.tree.currently_selected[-1] == self.context['object'].name
+            self.accordions.accesscontrol.is_opened
         )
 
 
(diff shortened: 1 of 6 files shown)


cfme/configure/access_control/__init__.py
```diff
@@ -95,10 +95,7 @@ class DetailsUserView(ConfigurationView):
     def is_displayed(self):
         return (
             self.title.text == 'EVM User "{}"'.format(self.context['object'].name) and
-            self.accordions.accesscontrol.is_opened and
-            # tree.currently_selected returns a list of strings with each item being the text of
-            # each level of the accordion. Last element should be the User's name
-            self.accordions.accesscontrol.tree.currently_selected[-1] == self.context['object'].name
+            self.accordions.accesscontrol.is_opened
```
LandoCalrizzian: You'll want to remove similar checks in DetailsRoleView and DetailsGroupView to fix the same type of failures

mshriver: Done, good catch, thanks @LandoCalrizzian !

PR #7105, opened 2018-04-18
[1LP][RFR] Rename Infrastructure.virtual_machines.Vm to InfraVm

mshriver:
There are a ton of tests collected due to the scope of this change. I'm not aiming to fix any test issues here, just an enhancement against the `virtual_machines.Vm` class name.

## PRT Results

PRT run 23306 executed about 1100 tests, with ~90% passing.  No failures look related to this rename.

{{ pytest: --use-provider scvmm --use-provider onprem }}

### DIFF
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -544,12 +544,13 @@ class VMConfiguration(Pretty):
         return changes
 
 
-class Vm(VM):
-    """Represents a VM in CFME
+class InfraVm(VM):
+    """Represents an infrastructure provider's virtual machine in CFME
 
+    Note the args are defined at common.BaseVM|VM class
     Args:
-        name: Name of the VM
-        provider_crud: :py:class:`cfme.cloud.provider.Provider` object
+        name: Name of the virtual machine resource
+        provider: :py:class:`cfme.infrastructure.provider.InfraProvider` object
         template_name: Name of the template to use for provisioning
     """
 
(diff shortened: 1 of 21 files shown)


cfme/infrastructure/virtual_machines.py
```diff
@@ -544,7 +544,7 @@ def get_changes_to_fill(self, other_configuration):
         return changes
 
 
-class Vm(VM):
+class InfraVm(VM):
     """Represents a VM in CFME
```
jamesooden: **QUESTION:** Shouldn't the comment change too?

jamesooden:
**QUESTION:** What is the reasoning behind this change?   I personally like the idea that we can refer to VM's from infrastructure and cloud providers as VM's as their are many operations that are similar between the two.   And I know you and others see that similarity too, so what problem are we trying to solve by making it more clear in the code that we are dealing with an infrastructure VM versus a Cloud VM?

mshriver:
@jamesooden I'm making this change because we currently have `BaseVM`, `VM`, `Instance`, and `Vm` classes.

`BaseVM` and `VM` are in the common module, for modeling things applicable to both cloud and infrastructure virtual machines.

The cloud modules use an `Instance` class as that's what CFME calls cloud VM's.

This module was using `Vm`, which can easily cause confusion since its only capitlized differently than the common `VM` class. This makes it more explicit, throughout, when we're using the Infrastructure specific VM class.

I don't think this is really solving any bugs in the code, its just improving the name and making it more explicit for readability, hence 'enhancement' label only.

cfme/infrastructure/virtual_machines.py
mshriver: It can, but I don't think its necessary for all comment uses to specify `InfraVm` unless specifically referring to the class, here its just a generic description.  I'll update it if you say **Required** :)

jamesooden: I just want it to say represents an Infrastructure VM and maybe spell out how they are different than the rest of their VM kind...not every where just here at the class description.

PR #7110, opened 2018-04-19
[1LP][RFR] Fix cockpit server role tests for cloud providers

niyazRedhat:
{{ pytest: cfme/tests/cloud_infra_common/test_cockpit_server.py --use-provider complete }}

### DIFF
--- a/cfme/tests/cloud_infra_common/test_cockpit_server.py
+++ b/cfme/tests/cloud_infra_common/test_cockpit_server.py
@@ -3,8 +3,6 @@ import pytest
 
 from cfme.cloud.instance import Instance
 from cfme.cloud.provider import CloudProvider
-from cfme.cloud.provider.ec2 import EC2Provider
-from cfme.cloud.provider.azure import AzureProvider
 from cfme.cloud.provider.gce import GCEProvider
 from cfme.common.provider import CloudInfraProvider
 from cfme.infrastructure.virtual_machines import Vm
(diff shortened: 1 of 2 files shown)


cfme/tests/cloud_infra_common/test_cockpit_server.py
```diff
@@ -42,14 +40,7 @@ def test_cockpit_server_role(appliance, provider, setup_provider, new_vm, enable
 
     view = navigate_to(new_vm, 'Details')
```
psav: **REQUIRED** I believe the conditional can be replaced with this
```python
    view = navigate_to(new_vm, 'Details')
    assert view.toolbar.access.item_enabled('Web Console') == enable
```

niyazRedhat: @psav It's a parametrized test with checking both `True` and `False` condition for cockpit, means if its not enabled, check its actually off, and on if its actually enabled.

mshriver: @niyazRedhat the assertion psav is suggesting takes that into account - here you've just asserted the boolean returned by item_enabled within if blocks controlled by `enable`, instead of asserting directly against that boolean compared to `enable`.  Functionally its the same, psav's suggestion is just simpler overall syntax.

Since you take an action when `enable` is `True` anyway, that's not taken when false, the if block will be required anyway, and I don't see any harm in asserting against the boolean returned by `item_enabled`.

PR #7117, opened 2018-04-19
[1LP][RFR] Hack to allow upstream testing to work again

psav:
* There is currently a bug in MiQ upstream where the servername field is
  not populated in the UI, consequently, saving settings clears the name
  which is required for creating the Server object. We now use a default
  in the case that the Attribute can't be accessed. This is because null
  attributes are not returned in the REST API.

### DIFF
--- a/cfme/base/__init__.py
+++ b/cfme/base/__init__.py
@@ -6,7 +6,9 @@ import sentaku
 
 from cfme.modeling.base import BaseCollection, BaseEntity
 from cfme.utils import ParamClassName
+from cfme.utils.log import logger
 from cfme.utils.pretty import Pretty
+from cfme.utils.version import LATEST
 
 
 @attr.s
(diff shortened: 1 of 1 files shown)


cfme/base/__init__.py
```diff
@@ -121,7 +123,14 @@ def get_master(self):
             server = server_collection.all[0]
         else:
             server = server_collection.find_by(is_master=True)[0]
-        return self.instantiate(name=server.name, sid=server.id)
+
+        try:
+            name = server.name
+        except AttributeError:
+            logger.error('The EVM has no name, setting it to EVM')
+            name = 'EVM' if self.appliance.version == LATEST else server.name
```
mshriver: **Optional**

Use `not self.appliance.is_downstream` instead of importing `LATEST` and checking version against it.

**Question**
If server.name just produced an AttributeError, if you're just expecting it to reraise, consider breaking up the ternary and setting name if upstream, else explicitly reraise.

PR #7150, opened 2018-04-25
[1LP][RFR] Activating test_guest_os_reset for RHV

jan-zmeskal:
Also wrapping `wait_for_last_boot_timestamp_refresh` with `soft_assert`. If that is not the case, the test case can succeed even though last boot time was not updated.

Regarding PRT outcome: As you can see, tests for VMWare failed. That might be an actual bug. I've filled this [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1571830) for RHV but maybe it's actually the same one.

{{pytest: cfme/tests/infrastructure/test_vm_power_control.py::test_guest_os_reset --long-running}}

### DIFF
--- a/cfme/tests/infrastructure/test_vm_power_control.py
+++ b/cfme/tests/infrastructure/test_vm_power_control.py
@@ -131,8 +131,17 @@ def wait_for_last_boot_timestamp_refresh(vm, boot_time, timeout=300):
     try:
         wait_for(_wait_for_timestamp_refresh, num_sec=timeout, delay=30,
                  fail_func=view.toolbar.reload.click)
+        return True
     except TimedOutError:
         return False
+
+
+def ensure_state_changed_on_unchanged(vm, state_changed_on):
+    """Returns True if current value of State Changed On in the Power Management
+    is the same as the supplied (original) value."""
+    view = navigate_to(vm, "Details")
+    new_state_changed_on = view.entities.summary("Power Management").get_text_of("State Changed On")
+    return state_changed_on == new_state_changed_on
 
 
 def wait_for_vm_tools(vm, timeout=300):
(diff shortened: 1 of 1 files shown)


cfme/tests/infrastructure/test_vm_power_control.py
```diff
@@ -425,7 +425,9 @@ def test_vm_power_options_from_off(provider, soft_assert, testing_vm, ensure_vm_
     check_power_options(provider, soft_assert, testing_vm, testing_vm.STATE_OFF)
 
 
-@pytest.mark.uncollectif(lambda provider: not provider.one_of(VMwareProvider))
+@pytest.mark.uncollectif(lambda provider: not provider.one_of((VMwareProvider, RHEVMProvider)))
+@pytest.mark.meta(blockers=[BZ(1571830, forced_streams=['5.8', '5.9'],
```
bsquizz: Just wondering if it'd be easier to read this if this was taken care of in 1 lambda:

```python
@pytest.mark.uncollectif(
     lambda provider: (provider.one_of(RHEVMProvider) and BZ(1571830, forced_streams=['5.8', '5.9']).blocks)
         or not provider.one_of((RHEVMProvider, VMwareProvider))
)
```

rlbabyuk: Better to have separate BZ blocker, as we split the reasons why we skip the test

cfme/tests/infrastructure/test_vm_power_control.py
```diff
@@ -425,7 +425,9 @@ def test_vm_power_options_from_off(provider, soft_assert, testing_vm, ensure_vm_
     check_power_options(provider, soft_assert, testing_vm, testing_vm.STATE_OFF)
 
 
-@pytest.mark.uncollectif(lambda provider: not provider.one_of(VMwareProvider))
+@pytest.mark.uncollectif(lambda provider: not provider.one_of((VMwareProvider, RHEVMProvider)))
```
quarckster: **REQUIRED**
```python
@pytest.mark.provider([VMwareProvider, RHEVMProvider])
```

jan-zmeskal:
@mshriver, @quarckster Okay, so I did the reproduction on vSphere Client version 6.5.0.10000 and the behaviour described in BZ is not there

cfme/tests/infrastructure/test_vm_power_control.py
```diff
@@ -439,9 +441,12 @@ def test_guest_os_reset(appliance, testing_vm_tools, ensure_vm_running, soft_ass
 
     testing_vm_tools.wait_for_vm_state_change(
```
bsquizz: I think this check for to wait for the VM to be on is not needed, because at least in VMWare the VM state never powers off. This is probably the case in RHEV too. If this is the way it works in RHEV as well, it would probably be a better test to check that the text of 'Power Management' -> 'State Changed On' has NOT changed once we see that 'Power Management' -> 'Last Boot Time' has changed. This would "prove" that the VM was not hard rebooted

jan-zmeskal:
@bsquizz Thank you very much for looking into this. I'll try to implement your suggestions!

cfme/tests/infrastructure/test_vm_power_control.py
jan-zmeskal: @bsquizz I realized one more thing. This test method uses `ensure_vm_running` fixture. Which means that at this point the VM will be ON anyway - otherwise the fixture would fail.

bsquizz: Yes exactly. So the wait_for_running at the very top of the test is redundant too :)

PR #7153, opened 2018-04-25
[WIP] Fixing CMQE Automation Tests

jawatts:
Updates to tests that are failing. 

{{ pytest: cfme/tests/containers/test_cockpit.py cfme/tests/containers/test_containers_summary.py --use-provider ocp-v3 }}


cfme/containers/provider/__init__.py
```diff
@@ -564,6 +572,14 @@ class ContainerObjectAllBaseView(ProvidersView):
     policy = Dropdown('Policy')
     download = Dropdown('Download')
     toolbar = View.nested(ProviderToolBar)
+    SUMMARY_TEXT = None
+
+    @property
+    def summary_text(self):
+        if isinstance(self.SUMMARY_TEXT, string_types):
```
mshriver: Handle when its None cleanly (unlikely since this is a base view, but possible) by:

`if isinstance(self.SUMMARY_TEXT, (string_types, None))`

cfme/containers/provider/__init__.py
```diff
@@ -616,6 +633,13 @@ def is_displayed(self):
             '{} (Summary)'.format(self.context['object'].name) in self.breadcrumb.active_location
         )
 
+    @property
+    def summary_text(self):
+        if isinstance(self.SUMMARY_TEXT, string_types):
```
mshriver: same as above, handle None cleanly

cfme/tests/containers/test_cockpit.py
```diff
@@ -11,25 +12,35 @@
...
-
-    port_num = ':9090/system'
-    url_cockpit = provider.hostname + port_num
-    view.browser.url = url_cockpit
-    title_pg = view.browser.selenium.title()
-    assert title_pg == 'Cockpit', 'Cockpit main page failed to open'
+    if cockpit:
+        appliance.server.browser.refresh()
+        assert not view.toolbar.web_console.disabled
+        view.toolbar.web_console.click()
```
mshriver: I don't like completely losing the verification that at least the new console tab was opened, and I think this creates some bad assumptions for validation - if the `click()` did open a new tab, the flash widget won't be displayed to call `assert_no_error` against it.

I'm thinking a flow like the following pseudo-code would be a more deterministic verification:
```
view.toolbar.web_console.click()
if provider.version < 3.6:
    assert view.is_displayed  # shouldn't have changed
    view.flash.assert_message('Feature not available/supported')
else:
    assert not view.is_displayed  # browser should be on a different tab now
    assert ':9090/system' in view.browser.url
    assert provider.hostname in view.browser.url  # may not work if the IP is used
    # close the console tab
    assert view.is_displayed
    view.flash.assert_no_error()
```

cfme/common/vm_console.py
```diff
@@ -17,6 +17,51 @@
 from wait_for import wait_for, TimedOutError
 
 
+class ConsoleMixin(object):
+    '''
```
quarckster: **REQUIRED**
Please use double quotes for docstrings.

PR #7155, opened 2018-04-26
[1LP][RFR] Added support for other providers [VM graph]

digitronik:
Purpose or Intent
=================
- Added support of VM graph test for other provider.
  * Azure
  * AWS
  * Vsphere6
  * RHV
- Added changes in Views as per Provider
- One test uncollected due to our infra problem; we are not getting proper metrics for EC2 disk activity. 

{{pytest: cfme/tests/candu/ -v --use-provider complete --long-running}}

### DIFF
--- a/cfme/cloud/provider/azure.py
+++ b/cfme/cloud/provider/azure.py
@@ -3,6 +3,7 @@ import attr
 from wrapanapi.msazure import AzureSystem
 
 from cfme.common.provider import DefaultEndpoint, DefaultEndpointForm
+from cfme.infrastructure.provider.rhevm import RHEVMVMUtilizationView
 from cfme.services.catalogs.catalog_items import AzureCatalogItem
 from cfme.utils.version import pick
 from . import CloudProvider
@@ -24,6 +25,11 @@ class AzureEndpointForm(DefaultEndpointForm):
     pass
 
 
+class AzureInstanceUtilizationView(RHEVMVMUtilizationView):
+    """A VM Utilization view for Azure providers"""
+    pass
+
+
 @attr.s(hash=False)
 class AzureProvider(CloudProvider):
     """
@@ -31,6 +37,7 @@ class AzureProvider(CloudProvider):
      represents CFME provider and operations available in UI
     """
     catalog_item_type = AzureCatalogItem
+    vm_utilization_view = AzureInstanceUtilizationView
     type_name = "azure"
     mgmt_class = AzureSystem
     db_types = ["Azure::CloudManager"]
--- a/cfme/common/candu_views.py
+++ b/cfme/common/candu_views.py
@@ -18,16 +18,6 @@ class VMUtilizationView(View):
     title = Text(".//div[@id='main-content']//h1")
     options = View.nested(OptionForm)
 
-    vm_cpu = LineChart(id='miq_chart_parent_candu_0')
-    vm_cpu_state = LineChart(id='miq_chart_parent_candu_1')
-    vm_memory = LineChart(id='miq_chart_parent_candu_2')
-    vm_disk = LineChart(id='miq_chart_parent_candu_3')
-    vm_network = LineChart(id='miq_chart_parent_candu_4')
-
-
-class VMUtilizationAllView(VMUtilizationView):
-    """A All view without select compare to option"""
-
     @property
     def is_displayed(self):
         if self.options.compare_to.is_displayed:
--- a/cfme/common/candu_views.py
+++ b/cfme/common/candu_views.py
@@ -1,5 +1,6 @@
-from widgetastic_manageiq import LineChart
 from widgetastic.widget import ConditionalSwitchableView, Text, TextInput, View
+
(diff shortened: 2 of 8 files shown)


cfme/tests/candu/test_vm_graph.py
```diff
@@ -2,21 +2,32 @@
...
     pytest.mark.usefixtures('setup_provider'),
-    pytest.mark.provider([VMwareProvider],
+    pytest.mark.provider([VMwareProvider, RHEVMProvider, EC2Provider, AzureProvider],
                          required_fields=[(['cap_and_util', 'capandu_vm'], 'cu-24x7')])
 ]
 
 VM_GRAPHS = ['vm_cpu', 'vm_cpu_state', 'vm_memory', 'vm_disk', 'vm_network']
 
 
+# ToDo: Currently disk activity for EC2 not collecting due to infra issue.
```
jamesooden: **QUESTION:** Is there some sort of upstream bug/issue we could reference in this comment?

cfme/tests/candu/test_vm_graph.py
```diff
@@ -43,18 +54,29 @@ def test_vm_most_recent_hour_graph_screen(graph_type, provider, enable_candu):
...
+
+    graph_data = 0
     for leg in graph.all_legends:
         graph.hide_legends(leg)
         assert not graph.legend_is_displayed(leg)
         graph.display_legends(leg)
         assert graph.legend_is_displayed(leg)
 
-        leg_data = 0
         for data in graph.data_for_legends(leg).values():
```
jamesooden: **OPTIONAL:** Could you add a comment explaining what you are trying to achieve with this part. Also, perhaps explain why the resulting sum should be greater than zero.

cfme/tests/candu/test_vm_graph.py
digitronik: No, we are working on it. Its volume issue basically. 
https://forums.aws.amazon.com/thread.jspa?threadID=108613

cfme/infrastructure/virtual_machines.py
```diff
@@ -1473,7 +1473,18 @@ def step(self, *args, **kwargs):
 
 @navigator.register(Vm, 'candu')
 class VmUtilization(CFMENavigateStep):
-    VIEW = VMUtilizationAllView
+    @property
+    def get_view(self):
```
mshriver: This should just be `def VIEW(self):` property method, there's no need to wrap the method here just to change the name.

When is `VIEW=None` a valid use case that shouldn't raise an exception?

digitronik: @mshriver: :+1: right, It's my bad as `CFMENavigateStep` has `VIEW=None` so I can overwrite it with the same name property.

dajoRH:
Lint report for commit 54f4eb48084263939b1ebee6491b22b5396b6870:

- :red_circle: Line 1477:9: **N802** *function name should be lowercase*

**No commit flaws detected**.

*CFME QE Bot*

nachandr:
@digitronik , Could you add this comment?
#TODO: Add support for GCE once https://github.com/ManageIQ/integration_tests/pull/7155 is fixed

digitronik:
@nachandr @mshriver: changes added

cfme/common/candu_views.py
```diff
@@ -37,3 +27,32 @@ def is_displayed(self):
             )
         else:
             return "Capacity & Utilization data for Virtual Machine" in self.title.text
+
+
+class VirtualcenterVMUtilizationView(VMUtilizationView):
```
psav: **REQUIRED**: This is a no-no I'm afraid. It's a real shame that these are not linking the integer with the particular graph. However, what we need to do, is move these Views to the provider modules. They are too specific to be here.

cfme/infrastructure/virtual_machines.py
```diff
@@ -1473,7 +1473,15 @@ def step(self, *args, **kwargs):
 
 @navigator.register(Vm, 'candu')
 class VmUtilization(CFMENavigateStep):
-    VIEW = VMUtilizationAllView
+    @property
+    def VIEW(self):     # noqa
+        """Property returning VM Utilization Views as per provider type"""
+        if self.obj.provider.type is 'virtualcenter':
```
psav: **REQUIRED**: I'm also not keen on this - what we should do instead is have a property on the provider class that links to the view instead. If the views are inside the provider modules, this actually becomes pretty easy.

dajoRH:
*CFME QE Bot*

dajoRH:
*CFME QE Bot*

cfme/cloud/provider/azure.py
```diff
@@ -81,3 +87,7 @@ def discover_dict(credential):
             'tenant_id': getattr(credential, 'tenant_id', None),
             'subscription': getattr(credential, 'subscription_id', None)
         }
+
+    @property
```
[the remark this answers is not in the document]
digitronik: @psav: Done

PR #7171, opened 2018-04-30
[1LP][RFR] Host Utilization for recent hr

digitronik:
Purpose or Intent
=================
- Added test for recent hour host graph
- Supported Providers
  * Vsphere
  * RHV
- Added Views for support
- Added Navigation and wait for Utilization enable with reference to host object.

{{pytest: cfme/tests/candu/test_host_graph.py -v --use-provider complete}}

### DIFF
--- a/cfme/tests/candu/test_host_graph.py
+++ b/cfme/tests/candu/test_host_graph.py
@@ -1,0 +1,84 @@
+import pytest
+
+from cfme import test_requirements
+from cfme.infrastructure.provider.rhevm import RHEVMProvider
+from cfme.infrastructure.provider.virtualcenter import VMwareProvider
+from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.wait import wait_for
+
+
+pytestmark = [
+    pytest.mark.tier(3),
+    test_requirements.c_and_u,
+    pytest.mark.usefixtures('setup_provider'),
+    pytest.mark.provider([VMwareProvider, RHEVMProvider],
+                         required_fields=[(['cap_and_util', 'capandu_vm'], 'cu-24x7')])
+]
+
+
+HOST_RECENT_HR_GRAPHS = ['host_cpu',
+                         'host_memory',
+                         'host_disk',
+                         'host_network']
+
+
+@pytest.fixture(scope='function')
+def host(appliance, provider):
+    collection = appliance.collections.hosts
+    for test_host in provider.data['hosts']:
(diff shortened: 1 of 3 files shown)


cfme/tests/candu/test_host_graph.py
```diff
@@ -0,0 +1,84 @@
...
+             delay=5, timeout=600, fail_func=refresh)
+
+    graph_data = 0
+    for leg in graph.all_legends:
+        graph.hide_legends(leg)
+        assert not graph.legend_is_displayed(leg)
+        graph.display_legends(leg)
+        assert graph.legend_is_displayed(leg)
+
+        for data in graph.data_for_legends(leg).values():
```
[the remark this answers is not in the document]
digitronik: @jamesooden: I added the last step with the short description in [doc steps]( https://github.com/ManageIQ/integration_tests/pull/7171/files#diff-db180b74ac3901aac7ad21913712960eR48). I think it will be sufficient.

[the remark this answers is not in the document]
jamesooden: I personally am not understanding exactly what the summing is about, even, with the description in the doc steps.

cfme/tests/candu/test_host_graph.py
```diff
@@ -0,0 +1,90 @@
...
+    view = navigate_to(host, 'candu')
+    view.options.interval.fill('Most Recent Hour')
+
+    if graph_type == 'host_cpu':
+        graph = view.interval_type.host_cpu
+    elif graph_type == 'host_memory':
+        graph = view.interval_type.host_memory
+    elif graph_type == 'host_disk':
+        graph = view.interval_type.host_disk
+    else:
```
psav: **OPTIONAL**: Personally, I'd like to see either a ```host_network``` inside an ```elif``` and an ```else``` declaring that it's not a valid type, just in case. In all honesty, I'd love to see this being a dictionary. I know it's difficult because we need to have the ```view``` before we can get them. We could use a chain of ```getattr``` with the dict value. However I am happy to merge this without. Please let me know.

digitronik: I like to try suggestions. Please wait.

digitronik: @psav: I removed total conditions. just used 
` graph = getattr(view.interval_type, graph_type)`

I found no need of passing `dict`. As we are passing same `str` which are attributes of `view.interval`. Let me know, your comment on it.

PR #7174, opened 2018-04-30
[1LP][RFR] Fixed a bug with some network provider collections

psav:


mshriver:
I think this may be missing one use of instantiate: 
https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/networks/test_tag_tagvis.py#L50

izapolsk:
my bad, thanks for fixing this.

PR #7179, opened 2018-05-01
[1LP][RFR] Move FQDN setting into Appliance.set_resolvable_hostname

mshriver:
Allows for calling in custom_configure as-is currently done, and for calling separately during deployment of a snapshotted configured image in sprout

### DIFF
--- a/cfme/utils/appliance/__init__.py
+++ b/cfme/utils/appliance/__init__.py
@@ -5,7 +5,6 @@ import traceback
 from copy import copy
 from datetime import datetime
 from tempfile import NamedTemporaryFile
-from textwrap import dedent
 from time import sleep, time
 
 import attr
@@ -140,7 +139,7 @@ class ApplianceConsoleCli(object):
             "appliance_console_cli {}".format(appliance_console_cli_command))
 
     def set_hostname(self, hostname):
-        self._run("--host {host}".format(host=hostname))
+        return self._run("--host {host}".format(host=hostname))
 
     def configure_appliance_external_join(self, dbhostname,
             username, password, dbname, fetch_key, sshlogin, sshpass):
@@ -507,6 +506,8 @@ class IPAppliance(object):
             # 'self.unpartitioned_disks' should exist and therefore this won't run.
             if self.is_downstream and not self.unpartitioned_disks:
                 self.db.create_db_lvm()
+            if on_openstack:
+                self.set_resolvable_hostname(log_callback=log_callback)
 
             self.db.setup(region=region, key_address=key_address,
                           db_address=db_address, is_pod=self.is_pod)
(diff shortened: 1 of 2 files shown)


cfme/utils/appliance/__init__.py
```diff
@@ -2480,6 +2480,42 @@ def is_ip_available():
         appliance.name = name or cls._default_name
         return appliance
 
+    def set_resolvable_hostname(self):
+        """Try to lookup the hostname based on self.hostname, which is generally an IP
+
+        If a hostname is resolved for the IP, its used to set the hostname via appliance_cli
+        If an IP is resolved for self.hostname, no action is taken
```
izapolsk: please add a note that it is relevant to appliances hosted on openstack.

mshriver: done

sprout/appliances/tasks.py
```diff
@@ -1627,6 +1628,12 @@ def appliance_yum_update(self, appliance_id):
     appliance.ipapp.update_rhel(reboot=False)
 
 
+@singleton_task()
+def appliance_set_hostname(appliance_id):
+    appliance = Appliance.objects.get(id=appliance_id)
+    appliance.ipapp.set_resolvable_hostname(reboot=False)
```
izapolsk: REQUIRED: reboot=False ?

sprout/appliances/tasks.py
```diff
@@ -1627,6 +1628,12 @@ def appliance_yum_update(self, appliance_id):
     appliance.ipapp.update_rhel(reboot=False)
 
 
+@singleton_task()
+def appliance_set_hostname(appliance_id):
+    appliance = Appliance.objects.get(id=appliance_id)
```
izapolsk: REQUIRED: I believe you need to to check whether this appliance is hosted on openstack before doing changes.

sprout/appliances/tasks.py
mshriver: copypasta, can't believe I missed that.

sprout/appliances/tasks.py
mshriver: done.

sprout/appliances/tasks.py
```diff
@@ -1627,6 +1628,13 @@ def appliance_yum_update(self, appliance_id):
...
+    if appliance.provider.provider_data.get('type', None) == 'openstack':
+        appliance.ipapp.set_resolvable_hostname()
```
izapolsk: REQUIRED: are you sure it will work well w/o callback ?

mshriver: it will now that I wrapped `set_resolvable_hostname` with `logger_wrap` decorator. :)

dajoRH:
*CFME QE Bot*

mshriver:
@izapolsk after testing this locally (thanks for your help in setup!!!) I think this is a working solution to setting hostnames.

I found that appliance_console_cli isn't cleaning up hostnames, so there is either a BZ to be fixed or workaround to be implemented in seal_for_templatizing. I don't think that's a blocker to merging this as-is and updating production sprout so that openstack instances will have resolvable hostnames.

PR #7181, opened 2018-05-01
[1LP][RFR] preconfigured openshift sprout providers

izapolsk:
This PR contains a bunch of sprout changes which are necessary for turning on preconfigured openshift appliances.
In addition, it restores mfalesni's sprout.sh script which is convenient to kick off sprout locally for testing/development purpose.
Besides that, there is .gitignore file for sprout which is convenient for development.

### DIFF
--- a/cfme/utils/appliance/__init__.py
+++ b/cfme/utils/appliance/__init__.py
@@ -2445,10 +2445,7 @@ class Appliance(IPAppliance):
             def is_ip_available():
                 try:
                     ip = provider.mgmt.get_ip_address(vm_name)
-                    if ip is None:
-                        return False
-                    else:
-                        return ip
+                    return ip or False  # get_ip_address might return None
                 except AttributeError:
                     return False
             ec, tc = wait_for(is_ip_available,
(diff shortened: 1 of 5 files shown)


izapolsk:
it isn't big. Major part of this PR is mfalesni's script which don't need to be reviewed.

cfme/utils/appliance/__init__.py
```diff
@@ -2434,30 +2434,48 @@ def from_provider(cls, provider_key, vm_name, name=None, **kwargs):
...
+        app_kwargs = kwargs.copy()
 
-        if 'hostname' in kwargs:
-            hostname = kwargs.pop('hostname')
-        else:
+        if 'hostname' not in app_kwargs:
+            def is_ip_available():
+                try:
+                    ip = provider.mgmt.get_ip_address(vm_name)
+                    if ip is None:
```
mshriver: **Optional**

```python
return ip or False
```

cfme/utils/appliance/__init__.py
```diff
@@ -2434,30 +2434,48 @@ def from_provider(cls, provider_key, vm_name, name=None, **kwargs):
...
+                        return False
+                    else:
+                        return ip
+                except AttributeError:
```
mshriver: **Optional**

Consider using `wait_for(handle_exception=True)`, though it isn't limited to only AttributeError exception skipping and may be too broad.

cfme/utils/appliance/__init__.py
mshriver: Went ahead and added a commit for this.

PR #7186, opened 2018-05-02
[RFR] add logging and retry for enable/disable provider check in sprout

izapolsk:
it should help to find out why some providers become disabled for some time


sprout/appliances/tasks.py
```diff
@@ -1165,9 +1165,11 @@ def check_templates_in_provider(self, provider_id):
     # Get templates and update metadata
     try:
         templates = map(str, provider.api.list_template())
-    except Exception:
+    except Exception as err:
+        self.logger.warning("Provider will be marked as not working because of {e}".format(e=err))
```
mshriver: use args with logger instead of `.format()`.

PR #7221, opened 2018-05-07
[1LP][WIP] Add stats for the number of network and storage devices

skovic:
This PR adds stats for the number of network and storage devices associated with a physical server.

Depends on PR https://github.com/ManageIQ/wrapanapi/pull/252


mshriver:
Marking WIP since it depends on an unmerged/released wrapanapi PR.

mshriver:
It looks like this code was included in #7253, we can merge it all under that PR, please close this if that's the case.

PR #7222, opened 2018-05-07
[RFR][1LP] removing retry from check template task because it produces many failed tasks for invalid providers

izapolsk:
REQUIRES: https://github.com/ManageIQ/wrapanapi/pull/253


mshriver:
PRT not relevant, waiting on wrapanapi PR merge+release

PR #7230, opened 2018-05-08
Put the default passwords in the template

bdunne:
Purpose or Intent
=================
Fixing default credentials because they are currently incorrect.


bdunne:
@psav Please review

PR #7233, opened 2018-05-09
[WIP] DatePicker Widget

digitronik:
Purpose or Intent
=================
- DatePicker widget for selection of date for read-only and read-write date box.
- Supporting two method `select` and `fill`

### DIFF
--- a/widgetastic_manageiq/__init__.py
+++ b/widgetastic_manageiq/__init__.py
@@ -3,7 +3,7 @@ import atexit
 import json
 import math
 from collections import namedtuple
-from datetime import date
+from datetime import date, datetime
 from math import ceil
 from tempfile import NamedTemporaryFile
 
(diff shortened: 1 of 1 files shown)


mshriver:
I believe this is a patternfly widget, are there any MIQ specific changes to the widget - if not, it should go into widgetastic.patternfly library, not here.

http://www.patternfly.org/pattern-library/forms-and-controls/datepicker/

widgetastic_manageiq/__init__.py
```diff
@@ -4051,3 +4051,117 @@ def data_for_timestamp(self, timestamp):
...
+        if (date.strftime("%B") not in sw_value) or (year not in sw_value):
+            self.browser.click(self.DATE_PICKER_SWITH)
+            if year not in self.date_switch_value:
+                self.browser.click(self.DATE_PICKER_SWITH)
+                self._select_year(year)
+            self._select_month(month)
+        self._select_date(day)
+        return True
+
+    def fill(self, date):
```
lkhomenk: QUESTION: not sure that this widget can be filled in this way. tried with the one from gif - we can't edit/send_keys to this input.
also - your `select` should be your `fill`. that way we could call `form.fill` to fill all widgets on the form

PR #7235, opened 2018-05-10
[1LP][RFR] Use a cache for returning provider mgmt instances

bsquizz:
Creating a new instance of wrapanapi System classes is inefficient, the most glaring example involves vCenter where every new instance of the class needs to call a method that takes 5min or more -- "service_instance.RetrieveContent()" -- this is a cached property in the class so if we return the same provider mgmt instance when asked to get mgmt again for the same provider key, we save a ton of time.

### DIFF
--- a/cfme/utils/providers.py
+++ b/cfme/utils/providers.py
@@ -23,6 +23,10 @@ providers_data = conf.cfme_data.get("management_systems", {})
 # Dict of active provider filters {name: ProviderFilter}
 global_filters = {}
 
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit be579b3b208754a076694cf660bdf9cfc889fb43:

`cfme/utils/providers.py`:
- :bangbang: Line 390:101: **E501** *line too long (116 > 100 characters)*

**No commit flaws detected**.

*CFME QE Bot*

PR #7236, opened 2018-05-11
[1LP][RFR] V2v infra mapping wizard

kedark3:
Purpose or Intent
=================

_Adding_ new Widgets required for V2V infrastructure mapping wizard, also, the wizard itself.

**UPSTREAM ONLY**

{{pytest: cfme/tests/v2v/test_v2v.py -v }}


dajoRH:
Lint report for commit 5d01bf4cb1db5bf945af689690aba387d0edd973:

- :bangbang: Line 4352:101: **E501** *line too long (102 > 100 characters)*

**No commit flaws detected**.

*CFME QE Bot*

widgetastic_manageiq/__init__.py
```diff
@@ -4051,3 +4051,461 @@ def data_for_timestamp(self, timestamp):
         for row in self.tooltip.rows():
             tooltip_data.update({row[0].text: row[1].text})
         return tooltip_data
+
+# Widget and its form views for Infra Planning Wizard
+
+
+class MultiSelectBox(Widget):
```
We already have `MultiBoxSelect`. Please use another name to avoid mistakes.

widgetastic_manageiq/__init__.py
```diff
@@ -4051,3 +4051,461 @@ def data_for_timestamp(self, timestamp):
...
+    """Multi Select Box to be used in Infra Planning Wizard for v2v.
+
+    Args:
+        div_id (str): id of div element that contains multiselectbox
+    Usage:
+        e.g. target_clusters = MultiSelectBox('target_clusters')
+    """
+    ITEM_LOCATOR = ParametrizedLocator('.//div[contains(@id,{@div_id|quote})]/div/div/div/span'
```
quarckster: Widget should have `ROOT` attribute.

widgetastic_manageiq/__init__.py
```diff
@@ -4051,3 +4051,461 @@ def data_for_timestamp(self, timestamp):
...
+    SELECTED_ITEM_LOCATOR = ParametrizedLocator('.//div[contains(@id,{@div_id|quote})]/'
+        'div[contains(@class,"selected")]/div/div/span'
+    '[contains(@class,"dual-pane-mapper-item-container")]')
+
+    def __init__(self, parent, div_id, logger=None):
+        Widget.__init__(self, parent, logger=logger)
+        self.div_id = div_id
+
+    @property
+    def is_displayed(self):
```
Not needed if `ROOT` is defined.

widgetastic_manageiq/__init__.py
```diff
@@ -4051,3 +4051,461 @@ def data_for_timestamp(self, timestamp):
...
+        """Method to read elements that are currently selected."""
+        return self.list_selected_items
+
+    def fill(self, values, partial_match=False):
+        """Click/Select multiple the items in the multiselectbox
+
+        Args:
+            items: 'list' of items of type 'str' to be matched
+            partial_match: if you want to use partial text matching. Default:False
+        Returns: True if all values were found and clicked. False if key_error_count != 0.
```
`fill` must return `False` if `values == self.read()` and `True` if filling finished successfully.

widgetastic_manageiq/__init__.py
```diff
@@ -4051,3 +4051,461 @@ def data_for_timestamp(self, timestamp):
...
+    def list_all_items_locators(self):
+        """Returns list of item locators for each of the items."""
+        return self.browser.elements(self.ITEM_LOCATOR.locator)
+
+    @property
+    def list_selected_items_locators(self):
+        """Returns list of item locators for currently selected items."""
+        return self.browser.elements(self.SELECTED_ITEM_LOCATOR.locator)
+
+    def _get_all_options(self):
```
This should be a property.

widgetastic_manageiq/__init__.py
```diff
@@ -4051,3 +4051,461 @@ def data_for_timestamp(self, timestamp):
...
+            raise ItemNotFound("{} was not found in {}".format(text, self.list_items))
+
+
+class InfraMappingTreeView(Widget):
+    """Widget to capture/interact with the InfraMappingTreeView in Infra Mapping Wizard for v2v.
+
+    Args:
+        tree_class(str): Class name of the div representing this TreeView
+    """
+    ROOT_ITEMS = ParametrizedLocator('//div[contains(@class,{@tree_class|quote})]/ul/li/'
```
`ROOT` should be defined

widgetastic_manageiq/__init__.py
```diff
@@ -4051,3 +4051,461 @@ def data_for_timestamp(self, timestamp):
...
+                try:
+                    self.all_items_locators[item].click()
+                except KeyError:
+                    key_error_count += 1
+                    self.logger.debug("Could not find {} in {}"
+                        .format(item, self.all_items_locators.keys()))
+        return (key_error_count == 0 and len(values) > 0)
+
+
+class InfraMappingCommonButtonsSet1(View):
```
All these views below should go in `cfme/v2v`

widgetastic_manageiq/__init__.py
```diff
@@ -4051,3 +4051,461 @@ def data_for_timestamp(self, timestamp):
...
+                    partial_match=partial_match)
+                self.add_mapping.click()
+
+
+class InfraMappingWizardResultsViw(InfraMappingCommonButtonsSet1):
+    next_btn = Button("Close")
+    continue_to_plan_wizard_btn = Button("Continue to the plan wizard")
+
+
+class InfraMappingWizard(Widget):
```
This should be a view.

widgetastic_manageiq/__init__.py
```diff
@@ -4051,3 +4051,461 @@ def data_for_timestamp(self, timestamp):
...
+        if isinstance(sources, list) and sources and target:
+            for item in sources:
+                self.logger.info("Mapping Source Cluster {} to Target Cluster {}"
+                    .format(item, target))
+                self.source_clusters.select_by_text(item, partial_match=partial_match)
+                self.target_clusters.select_by_text(target, partial_match=partial_match)
+                self.add_mapping.click()
+
+
+class InfraMappingWizardDatastoresViw(InfraMappingCommonButtonsSet1, InfraMappingCommonButtonsSet2):
```
1. Typo is here. 
2. Don't inherit `InfraMappingCommonButtonsSet1` and `InfraMappingCommonButtonsSet2`, but include them using `View.include`

widgetastic_manageiq/__init__.py
```diff
@@ -4051,3 +4051,461 @@ def data_for_timestamp(self, timestamp):
...
+        """
+        if isinstance(sources, list) and sources and target:
+            for item in sources:
+                self.source_networks.select_by_text(item, partial_match)
+                self.target_networks.select_by_text(target.split("(")[0],
+                    partial_match=partial_match)
+                self.add_mapping.click()
+
+
+class InfraMappingWizardResultsViw(InfraMappingCommonButtonsSet1):
```
Typo

widgetastic_manageiq/__init__.py
kedark3: I believe @mshriver wanted me to put it in widgetastic_manageiq

widgetastic_manageiq/__init__.py
kedark3: @mshriver I believe we discussed this to be a widget

widgetastic_manageiq/__init__.py
quarckster: From my perspective only widely ised widgets and views should go in `widgetastic_manageiq`. These view are specific for v2v and should go in to a respectful module.

widgetastic_manageiq/__init__.py
quarckster: Nope, it should be a view.

widgetastic_manageiq/__init__.py
quarckster: This part `.//div[contains(@id,{@div_id|quote})]` could be `ROOT`

kedark3:
@quarckster Thanks for all reviews. I will update the PR.

dajoRH:
Lint report for commit 1e7d33ad83194140ff961cde3be944f01030a7f6:

- :bangbang: Line 4316:5: **E303** *too many blank lines (2)*

**No commit flaws detected**.

*CFME QE Bot*

widgetastic_manageiq/__init__.py
mshriver: @kedark3 don't recall recommending that, but regardless, these aren't MIQ widgets, they're specific views for v2v model, so yeah I'd put them in `cfme/v2v`

cfme/v2v/migrations.py
```diff
@@ -0,0 +1,130 @@
...
+from cfme.modeling.base import BaseCollection, BaseEntity
+from cfme.utils.appliance.implementations.ui import navigator, CFMENavigateStep, navigate_to
+
+from navmazing import NavigateToAttribute, NavigateToSibling
+from widgetastic.widget import View
+from widgetastic_patternfly import Text, TextInput, Button
+from widgetastic_manageiq import InfraMappingWizard
+
+
+# class InfrastructureMappingWizard(View):
```
mshriver: remove?

cfme/v2v/migrations.py
```diff
@@ -0,0 +1,130 @@
+import attr
+
+from cfme.base.login import BaseLoggedInPage
+from cfme.modeling.base import BaseCollection, BaseEntity
+from cfme.utils.appliance.implementations.ui import navigator, CFMENavigateStep, navigate_to
+
+from navmazing import NavigateToAttribute, NavigateToSibling
```
mshriver: imports are reversed, local should be last.

cfme/v2v/migrations.py
```diff
@@ -0,0 +1,130 @@
...
+
+# Collections Entities
+
+
+@attr.s
+class InfrastructureMapping(BaseEntity):
+    """Class representing v2v infrastructure mappings"""
+    name = attr.ib()
+    description = attr.ib()
+    clusters_view = attr.ib(default=dict())
```
mshriver: It isn't clear what these are, are they view classes? Why default to dict for something that is `<name>_view` named?

setup.py
```diff
@@ -128,8 +128,10 @@
             'generic_objects = cfme.generic_objects.instance:GenericObjectInstanceCollection',
             'groups = cfme.configure.access_control:GroupCollection',
             'infra_providers = cfme.infrastructure.provider:InfraProviderCollection',
+            'infrastructure_mapping = cfme.v2v.migrations:InfrastructureMappingCollection',
```
mshriver: Consider names like `v2v_mapping` and `v2v_plan`

widgetastic_manageiq/__init__.py
mshriver: @kedark3 maybe a name change for `list_items` would help, like `all_items`, since we commonly use `all_<component>` for other widget methods.

widgetastic_manageiq/__init__.py
mshriver: I'd agree @quarckster, InfraMappingCommonButtonsSet1/2 being base classes for other V2V widgets creates a bit of an odd structure. `View.include` is perfect here.

widgetastic_manageiq/__init__.py
mshriver: @kedark3 was there a reason for using `Widget` instead of `View` here?  `View` definitely makes more sense as its a modal that's a collection of widgets.

widgetastic_manageiq/__init__.py
kedark3: done

cfme/v2v/migrations.py
kedark3: For each of these x_view=attr.ib(default=dict()), they are basically pages in the wizard/form and I pass data to fill those forms as dictionary. Do you want me to name it differently? Maybe, clusters_form_data or clusters_view_data?

mshriver: I think its a bit of an odd pattern - The attributes of InfrastructureMapping should detail items that are part of the model - specific elements of the clusters view.  Things like `source_cluster`, `target_cluster`, etc.

An `as_fill_value()` method would then return the dictionary for filling the form, built from the object's attributes.  This could be split to `clusters_fill_value`, but I think it would be better to have a single method that returns a single dict matching the form class's View+Widget structure.

This leaves navigation and creation to simple `view = navigate_to(collection, 'Add')`, `view.form.fill(infra_mapping_instance.as_fill_value())`

kedark3: the reason I did not break it down to things like source_clusters, target_clusters because it is difficult to predict those attributes upfront. What I am doing is as follows: 
 https://github.com/ManageIQ/integration_tests/pull/7236/files#diff-264354a78406ae4ca6d836bd7388d376R4269
 I need to pass in a "Mapping":{'sources':[], 'target': } and there could be more than one mappings. So best way I to depict that in the entity was to let user pass in entire dict for each view/page in the wizard, that includes all the data as follows: https://github.com/ManageIQ/integration_tests/pull/7236/files#diff-3a77e1415d5ddb90c31c2f74d6d32536R15
 and Form_data in the entity then constructs that from input here: https://github.com/ManageIQ/integration_tests/pull/7236/files#diff-3a77e1415d5ddb90c31c2f74d6d32536R15 and that can be used to fill in the values in the form
the form_data from test_v2v.py maybe removed and added to YAML as for a particular test I will need to have particular set of inputs. 
happy to discuss further. I tried to discuss this with landon and that's the best we could come up with i.e. getting data from YAML

cfme/v2v/migrations.py
```diff
@@ -0,0 +1,425 @@
...
+    include_buttons = View.include(InfraMappingCommonButtonsSet1)
+
+    def fill(self, values):
+        try:
+            self.name.fill(values['name'])
+            self.description.fill(values['description'])
+            return True
+        except Exception as e:
+            self.logger.debug("Following Exception occured while trying to fill:{}".format(e))
+            return False
```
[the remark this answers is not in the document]
kedark3: I wasn't aware of this being not required. I will remove it.

cfme/v2v/migrations.py
```diff
@@ -0,0 +1,425 @@
...
+                    },
+                    ...
+                }
+        """
+        try:
+            for mapping in values:
+                self._add_cluster_mapping(values[mapping]['sources'], values[mapping]['target'],
+                    partial_match=partial_match)
+            return True
+        except Exception as e:
```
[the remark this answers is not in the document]
kedark3: could be a KeyError if values dict is malformed. Also, select_by_text method from MultiSelectList may raise ItemNotFound.

cfme/v2v/migrations.py
```diff
@@ -0,0 +1,425 @@
...
+    include_buttons_set1 = View.include(InfraMappingCommonButtonsSet1)
+    include_buttons_set2 = View.include(InfraMappingCommonButtonsSet2)
+    source_networks = MultiSelectList('source_networks')
+    target_networks = MultiSelectList('target_networks')
+    next_btn = Button("Create")  # overriding, since 'Next' is called 'Create' in this form
+    cluster_selector = BootstrapSelect(locator='.//div[contains(@class,"cluster_select_select")]')
+
+    @property
+    def is_displayed(self):
+        return self.source_networks.item_count > 0 or self.target_networks.item_count > 0
```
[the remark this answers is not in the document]
kedark3: Removed

widgetastic_manageiq/__init__.py
```diff
@@ -4051,3 +4051,169 @@ def data_for_timestamp(self, timestamp):
...
+
+    @property
+    def list_selected_items(self):
+        """Returns list of items that are currently selected."""
+        return [item.text for item in self.list_selected_items_locators]
+
+    @property
+    def list_all_items_locators(self):
+        """Returns list of item locators for each of the items."""
+        return self.root_browser.elements(self.ROOT.locator + self.ITEM_LOCATOR)
```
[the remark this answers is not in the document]
kedark3: Initially, I was trying to use self.browser but it did not work, self.root_browser did. I suspect it is because I didn't make locator relative. I will check again.

widgetastic_manageiq/__init__.py
```diff
@@ -4051,3 +4051,169 @@ def data_for_timestamp(self, timestamp):
...
+
+    @property
+    def list_selected_items_locators(self):
+        """Returns list of item locators for currently selected items."""
+        return self.root_browser.elements(self.ROOT.locator + self.SELECTED_ITEM_LOCATOR)
```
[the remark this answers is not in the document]
kedark3: same as above

widgetastic_manageiq/__init__.py
```diff
@@ -4051,3 +4051,169 @@ def data_for_timestamp(self, timestamp):
...
+            return False
+        key_error_count = 0
+        if values:
+            for item in values:
+                try:
+                    self.select_by_text(item, partial_match=partial_match)
+                except KeyError:
+                    key_error_count += 1
+                    self.logger.debug("Could not find {} in {}".format(item, self.list_items))
+        return key_error_count == 0  # Meaning filling was successful
```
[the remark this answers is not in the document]
kedark3: should return False. updated code.

cfme/v2v/migrations.py
```diff
@@ -0,0 +1,425 @@
...
+
+class InfraMappingWizardClustersView(View):
+    include_buttons_set1 = View.include(InfraMappingCommonButtonsSet1)
+    include_buttons_set2 = View.include(InfraMappingCommonButtonsSet2)
+    source_clusters = MultiSelectList('source_clusters')
+    target_clusters = MultiSelectList('target_clusters')
+
+    @property
+    def is_displayed(self):
+        return self.source_clusters.item_count > 0 or self.target_clusters.item_count > 0
```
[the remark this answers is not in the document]
kedark3: Correction: I will not remove it as I am calling wait_displayed which uses is_displayed on my views.

widgetastic_manageiq/__init__.py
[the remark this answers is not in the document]
quarckster: **REQUIRED**
Please consider using the same logic as it was implemented here https://github.com/ManageIQ/integration_tests/blob/master/widgetastic_manageiq/__init__.py#L180

widgetastic_manageiq/__init__.py
```diff
@@ -4051,3 +4051,165 @@ def data_for_timestamp(self, timestamp):
...
+
+
+class InfraMappingTreeView(Widget):
+    """Widget to capture/interact with the InfraMappingTreeView in Infra Mapping Wizard for v2v.
+
+    Args:
+        tree_class(str): Class name of the div representing this TreeView
+    """
+    ROOT = ParametrizedLocator('.//div[contains(@class,{@tree_class|quote})]/ul/li')
+    ROOT_ITEMS = ('/span[contains(@class,"expand-icon")]/..')
```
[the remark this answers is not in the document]
kedark3: I was able to use relative path for ROOT_ITEMS, updating code shortly, but for ITEMS_LOCATOR, I couldn't find way to use appropriate locator that can be relative and yield correct output. discussed this with @mshriver on rocketchat and will try to debug that with him on Monday.

widgetastic_manageiq/__init__.py
```diff
@@ -4051,3 +4051,166 @@ def data_for_timestamp(self, timestamp):
...
+    def fill(self, values):
+        """Click on each value from list of values."""
+        if values == self.selected_items:
+            return False
+        key_error_count = 0
+        if values:
+            for item in values:
+                try:
+                    self.all_items_locators[item].click()
+                except KeyError:
```
Don't catch exceptions here, let them raise.

widgetastic_manageiq/__init__.py
```diff
@@ -4051,3 +4051,166 @@ def data_for_timestamp(self, timestamp):
...
+        Returns: True if all values were found and clicked. False if key_error_count != 0.
+         """
+        if values == self.list_selected_items or values == []:
+            return False
+        key_error_count = 0
+        if values:
+            for item in values:
+                try:
+                    self.select_by_text(item, partial_match=partial_match)
+                except KeyError:
```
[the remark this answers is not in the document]
kedark3: will use set() here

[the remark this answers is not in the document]
kedark3: Okay will remove it.

widgetastic_manageiq/__init__.py
```diff
@@ -4051,3 +4051,166 @@ def data_for_timestamp(self, timestamp):
...
+
+class InfraMappingTreeView(Widget):
+    """Widget to capture/interact with the InfraMappingTreeView in Infra Mapping Wizard for v2v.
+
+    Args:
+        tree_class(str): Class name of the div representing this TreeView
+    """
+    ROOT = ParametrizedLocator('.//div[contains(@class,{@tree_class|quote})]/ul/li')
+    ROOT_ITEMS = '//span[contains(@class,"expand-icon")]/..'
+    ITEMS_LOCATOR = ('/span/../text()[not(contains(.,"Default")) '
```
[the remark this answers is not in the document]
kedark3: Because I could not get it to work that way. :( I will try to debug it with Mike today. self.browser.elements without parent=self was not yielding correct output, and removing self.root_browser was only getting me 1 element instead of the list.

cfme/v2v/migrations.py
```diff
@@ -0,0 +1,417 @@
...
+    include_buttons_set2 = View.include(InfraMappingCommonButtonsSet2)
+    source_datastores = MultiSelectList('source_datastores')
+    target_datastores = MultiSelectList('target_datastores')
+    cluster_selector = BootstrapSelect(locator='.//div[contains(@class,"cluster_select_select")]')
+
+    @property
+    def is_displayed(self):
+        return self.source_datastores.item_count > 0 or self.target_datastores.item_count > 0
+
+    def fill(self, values, partial_match=False):
```
quarckster: **REQUIRED**
Please implement in the same as in `InfraMappingWizardClustersView`

cfme/v2v/migrations.py
```diff
@@ -0,0 +1,417 @@
...
+            target: str containing target item
+        """
+        sources = sources or []
+        if isinstance(sources, list) and sources and target:
+            for item in sources:
+                self.source_datastores.select_by_text(item, partial_match)
+                self.target_datastores.select_by_text(target.split("(")[0],
+                    partial_match=partial_match)
+                if handle_alert and self.browser.alert_present:
+                    # TODO: Investigate why alert_present is False
```
quarckster: Because it's not a browser alert, but a flash message.

cfme/v2v/migrations.py
```diff
@@ -0,0 +1,417 @@
...
+                }
+        """
+        try:
+            for mapping in values:
+                self._add_datastore_mapping(values[mapping]['sources'],
+                    values[mapping]['target'], partial_match=partial_match)
+            return True
+        except Exception as e:
+            self.logger.debug("Following Exception occured while trying to fill:%s", e)
+            return False
```
quarckster: **REQUIRED**
```python
source_datastores_filled = []
target_clusters_filled = []
for mapping in values:
    source_datastores_filled.append(self.source_datastores.fill(values[mapping]['sources'], partial_match=partial_match))
    target_datastores_filled.append(self.target_datastores.fill(values[mapping]['target'], partial_match=partial_match))
    self.add_mapping.click()
    # Here should be check of existence of a flash message with text 'source datastores exceeds the available space'. I guess you will need to create a new widget for it.
return any(source_datastores_filled) and any(target_datastores_filled)

File: cfme/v2v/migrations.py (hunk 21 above)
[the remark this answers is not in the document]
Comment (quarckster) [reply cfme/v2v/migrations.py]: **REQUIRED**
I believe it can be removed if you have `self.browser.plugin.ensure_page_safe()` in your fill method, please test it.

File: cfme/v2v/migrations.py [hunk 29]
```diff
@@ -0,0 +1,417 @@
...
+    def fill(self, values):
+        self.general_view.fill(values['general_view'])
+        self.general_view.next_btn.click()
+
+        self.clusters_view.wait_displayed()
+        self.clusters_view.fill(values['clusters_view'])
+        self.clusters_view.next_btn.click()
+
+        self.datastores_view.wait_displayed()
+        self.datastores_view.cluster_selector.select_by_visible_text(
```
quarckster: **REQUIRED**
Please use `fill` method of the widget.

cfme/v2v/migrations.py
```diff
@@ -0,0 +1,417 @@
...
+            values['datastores_view']['cluster'])
+        self.datastores_view.fill(values['datastores_view']['mappings'], partial_match=True)
+        self.datastores_view.next_btn.click()
+
+        self.networks_view.wait_displayed()
+        self.networks_view.cluster_selector.select_by_visible_text(
```
quarckster: **REQUIRED**
Please use `fill` method of the widget.

widgetastic_manageiq/__init__.py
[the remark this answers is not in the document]
quarckster: **REQUIRED**
Please change your xpaths, it's the reason. I strongly recommend you to look implementation of `BootstrapTreeview`. Your tree is very similar. You can even try to inherit it and override some methods and attributes.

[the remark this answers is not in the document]
kedark3: I tried that briefly when I began to implement it, but I could not see a way to implement off of BootstrapTreeview but I will reconsider.

cfme/v2v/migrations.py
kedark3: How do we validate presence of flash?

quarckster: `widgetastic.patternfly` has `FlashMessage` class, but it doesn't work here, so you can create a special widget for flash messages in this wizard.

widgetastic_manageiq/__init__.py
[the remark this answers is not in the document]
quarckster: At least look on xpaths of `BootstrapTreeview`, except `ROOT` they are should be the same.

cfme/v2v/migrations.py
kedark3: @quarckster thanks, I will keep this as TODO for this PR, since it is a corner case and I will take care of not letting my test case hit this. Even if I did hit, it won't fail test case. If that's alright.

widgetastic_manageiq/__init__.py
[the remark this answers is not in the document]
kedark3: @quarckster can you point me to a use case in manageiq where I can find other tree view? so I can compare the locators in those and my treeview!

cfme/v2v/migrations.py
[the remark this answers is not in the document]
kedark3: @quarckster view.browser.plugin.ensure_page_safe() this is returning None for me.

widgetastic_manageiq/__init__.py
```diff
@@ -4051,3 +4051,169 @@ def data_for_timestamp(self, timestamp):
...
+            tree_elements[element.text] = element
+
+        for element in self.root_browser.elements(self.ROOT.locator + self.ROOT_ITEMS):
+            tree_elements[element.text] = element
+        return tree_elements
+
+    def read(self):
+        return self.selected_items
+
+    def fill(self, values):
```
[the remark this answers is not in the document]
kedark3: this is done as well. will update PR with new commit shortly.

widgetastic_manageiq/__init__.py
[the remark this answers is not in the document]
kedark3: Got it to work as 
```
self.browser.elements(self.ROOT_ITEMS)
self.browser.elements(self.ITEMS_LOCATOR)
```

quarckster:
@kedark3 I pushed a commit to your branch, now it works fine.

dajoRH:
Lint report for commit ea273964e96999ecdf52d6a5736647dcd10a4944:

`cfme/tests/v2v/test_v2v.py`:
- :bangbang: Line 18:23: **E231** *missing whitespace after ':'*
- :bangbang: Line 19:30: **E231** *missing whitespace after ':'*
- :bangbang: Line 19:68: **E231** *missing whitespace after ','*
- :bangbang: Line 25:34: **E231** *missing whitespace after ':'*
- :bangbang: Line 30:26: **E231** *missing whitespace after ':'*
- :bangbang: Line 32:16: **E121** *continuation line under-indented for hanging indent*
- :bangbang: Line 32:26: **E231** *missing whitespace after ':'*
- :bangbang: Line 33:21: **E126** *continuation line over-indented for hanging indent*
- :bangbang: Line 34:34: **E231** *missing whitespace after ':'*
- :bangbang: Line 34:53: **E231** *missing whitespace after ','*
- :bangbang: Line 38:34: **E231** *missing whitespace after ':'*
- :bangbang: Line 41:17: **E121** *continuation line under-indented for hanging indent*
- :bangbang: Line 42:12: **E122** *continuation line missing indentation or outdented*
- :bangbang: Line 43:9: **E122** *continuation line missing indentation or outdented*
- :bangbang: Line 44:24: **E231** *missing whitespace after ':'*
- :bangbang: Line 48:34: **E231** *missing whitespace after ':'*
- :bangbang: Line 48:48: **E231** *missing whitespace after ','*
- :bangbang: Line 52:34: **E231** *missing whitespace after ':'*
- :bangbang: Line 57:9: **E122** *continuation line missing indentation or outdented*

Hmmm


**No commit flaws detected**.

*CFME QE Bot*

widgetastic_manageiq/__init__.py
```diff
@@ -4051,3 +4051,149 @@ def data_for_timestamp(self, timestamp):
...
+        """Returns list of Mapping targets or in other words, leaf nodes."""
+        return [self.browser.text(el) for el in self.browser.elements(self.ROOT_ITEMS)]
+
+    @property
+    def mapping_sources(self):
+        """Returns list of Mapping sources or in other words, parent nodes to leaf nodes."""
+        return [self.browser.text(el) for el in self.browser.elements(self.ITEMS_LOCATOR)]
+
+    @property
+    def selected_item(self):
```
kedark3: @quarckster this is returning NoSuchElement if nothing is selected. Do you want to handle that or let it throw?

widgetastic_manageiq/__init__.py
```diff
@@ -4051,3 +4051,149 @@ def data_for_timestamp(self, timestamp):
...
+            tree_elements[self.browser.text(element)] = element
+        return tree_elements
+
+    def read(self):
+        return self.selected_item
+
+    def fill(self, value):
+        if value == self.selected_item:
+            return False
+        self.all_items_locators[value or self.selected_item].click()
```
kedark3: Why is this value or self.selected_item?

quarckster: if value is empty list or None it will click on selected item again

dajoRH:
*CFME QE Bot*

cfme/tests/v2v/test_v2v.py
```diff
@@ -0,0 +1,64 @@
+"""Test to validate basic navigations.Later to be replaced with End-to-End functional testing."""
+import fauxfactory
+import pytest
+
+from cfme.utils.log import logger
+
+pytest.mark.ignore_stream('5.8', '5.9')
```
mshriver: **Required**

This is still getting collected against 5.8 in PRT.  This either needs to be included in a pytestmark list, or decorated on the individual test case.

cfme/tests/v2v/test_v2v.py
```diff
@@ -0,0 +1,64 @@
...
+        'datastores_view':
+            {
+                'Cluster (Default)': {
+                    'mappings': [
+                        {
+                            'sources': ['NFS_Datastore_1'],
+                            'target': ['hosted_storage']
+                        },
+                        {
+                            'sources': ['h02-Local_Datastore-8GB', 'h01-Local_Datastore-8GB'],
```
mshriver: **Required**

These values cannot be hardcoded and need to be compiled form the cfme_data yaml file.

cfme/tests/v2v/test_v2v.py
```diff
@@ -0,0 +1,64 @@
...
+                            'target': ['Storage - VLAN 33']
+                        }
+                    ]
+                }
+            }
+    }
+
+    infrastructure_mapping_collection = appliance.collections.v2v_mappings
+    add_infrastucture_mapping = infrastructure_mapping_collection.create(form_data)
+    logger.info(add_infrastucture_mapping)
```
mshriver: **Question / Required**

Flash message assertion on the view after create?

cfme/v2v/migrations.py
```diff
@@ -0,0 +1,329 @@
...
+
+from cfme.base.login import BaseLoggedInPage
+from cfme.modeling.base import BaseCollection, BaseEntity
+from cfme.utils.appliance.implementations.ui import navigator, CFMENavigateStep, navigate_to
+
+
+# Views
+
+
+class InfraMappingCommonButtonsSet1(View):
```
mshriver: **Required**

`Set1` and `Set2` are really ambiguous names, let's find something more descriptive and applicable to what these buttons are.

Maybe something like `InfraMappingFormControlButtons` and `InfraMappingWizardCommon` - the second view isn't just buttons, it includes the tree view.

cfme/v2v/migrations.py
```diff
@@ -0,0 +1,329 @@
...
+
+class InfraMappingCommonButtonsSet2(View):
+
+    add_mapping = Button('Add Mapping')
+    remove_mapping = Button('Remove Selected')
+    remove_all_mappings = Button('Remove All')
+    mappings_tree = InfraMappingTreeView(tree_class='treeview')
+
+
+class InfraMappingWizardGeneralView(View):
```
mshriver: I think this can be simply `InfraMappingWizardView`

cfme/v2v/migrations.py
```diff
@@ -0,0 +1,329 @@
...
+                'datastores_view':{
+                    'Cluster (Default)': {
+                       'mappings':[
+                            {
+                                'sources':['NFS_Datastore_1','iSCSI_Datastore_1'],
+                                'target':['hosted_storage']
+                            },
+                            {
+                                'sources':['h02-Local_Datastore-8GB', 'h01-Local_Datastore-8GB'],
+                                'target':['env-rhv41-01-nfs-iso']
```
mshriver: Don't use CFME QE specific examples :)

cfme/v2v/migrations.py
```diff
@@ -0,0 +1,329 @@
...
+                self.add_mapping.click()
+        was_change = any(source_networks_filled) and any(target_networks_filled)
+        if was_change:
+            self.next_btn.click()
+        return was_change
+
+
+class InfraMappingWizardResultsView(View):
+    include_buttons_set1 = View.include(InfraMappingCommonButtonsSet1)
+    next_btn = Button("Close")
```
mshriver: Why `next_btn` and not `close`?

cfme/v2v/migrations.py
```diff
@@ -0,0 +1,329 @@
...
+                                'sources':['DPortGroup'],
+                                'target':['Storage VLAN 33']
+                            }
+                        ]
+                    }
+                }
+            }
+    """
+    title = Text(locator='.//h4[contains(@class,"modal-title")]')
+    general_view = View.nested(InfraMappingWizardGeneralView)
```
mshriver: Sorry to harp on naming here but I think these can all drop the `_view` suffix and be simply `general`, `clusters`, `datastores`, `networks`, `results`

cfme/v2v/migrations.py
```diff
@@ -0,0 +1,329 @@
...
+        '[text()="Create Infrastructure Mapping"]')
+    create_migration_plan = Text(locator='(//a|//button)[text()="Create Migration Plan"]')
+
+    @property
+    def is_displayed(self):
+        return self.navigation.currently_selected == ['Compute', 'Migration']
+
+
+class AddInfrastructureMappingView(View):
+    wizard_form = InfraMappingWizard()
```
mshriver: Consider just using `form` here, for some consistency in our Views with forms and their widget access. That way our implementations are more consistently `view.form.fill()`

cfme/v2v/migrations.py
```diff
@@ -0,0 +1,329 @@
...
+
+@attr.s
+class InfrastructureMappingCollection(BaseCollection):
+    """Collection object for Migration mapping object"""
+    ENTITY = InfrastructureMapping
+
+    def create(self, form_data):
+        infra_map = self.instantiate(
+            name=form_data['general_view']['name'],
+            description=form_data['general_view']['description'],
```
mshriver: I think this needs to be a `.get()` call since description is optional and may not be included in the form_data dictionary.

cfme/v2v/migrations.py
```diff
@@ -0,0 +1,329 @@
...
+# TODO: Next Entity and Collection classes are to be filled by Yadnyawalk(ytale),
+# which he will submit PR for once my PR merged.
+
+
+@attr.s
+class MigrationPlan(BaseEntity):
+    """Class representing v2v Migration Plan"""
+    category = 'migrationplan'
+    string_name = 'Migration Plan'
+
```
mshriver: No `name` or `description` attributes for the entity?

cfme/v2v/migrations.py
```diff
@@ -0,0 +1,329 @@
...
+
+@attr.s
+class MigrationPlanCollection(BaseCollection):
+    """Collection object for Migration Plan object"""
+    ENTITY = MigrationPlan
+
```
mshriver: No create method to use the 'Add' nav destination and view that was written?

setup.py
```diff
@@ -137,6 +137,10 @@
             'infra_templates = cfme.infrastructure.virtual_machines:InfraTemplateCollection',
             'infra_vms = cfme.infrastructure.virtual_machines:InfraVmCollection',
             'hosts = cfme.infrastructure.host:HostCollection',
+            'infrastructure_mapping = cfme.v2v.migrations:InfrastructureMappingCollection',
```
mshriver: alphabetize these collection names into the rest of the list.

mshriver: All collection names must be plural.

widgetastic_manageiq/__init__.py
```diff
@@ -4207,3 +4207,153 @@ def data_for_timestamp(self, timestamp):
...
+            partial_match: if you want to use partial text matching. Default:False
+
+        Returns:
+            True if all values were found and clicked.
+        """
+        # TODO self.browser.plugin.ensure_page_safe() should be here
+        wait_for(lambda: "loading" not in self.browser.classes(self), timeout=5)
+        if values == self.list_selected_items or values == []:
+            return False
+        if values:
```
mshriver: I don't think this logic is right, as if `values` is None, this fill method will return True without doing anything.

I think the if statement above should also check `or values is None`, and then there's no need for this additional if statement.

widgetastic_manageiq/__init__.py
```diff
@@ -4207,3 +4207,153 @@ def data_for_timestamp(self, timestamp):
...
+
+    def select_by_text(self, text, partial_match=False):
+        """Select/click option using text
+
+        Args:
+            text(str): text to be matched
+            partial_match: if you want to use partial text matching. Default:False
+        """
+        try:
+            if text and not partial_match:
```
mshriver: **Required**

I think this should integrate with widgetastic.core.utils.partial_match and a contains xpath, like widgetastic.patternfly.BootstrapNav.

This might be too much to scope for including today though :/

cfme/tests/v2v/test_v2v.py
kedark3: There is no flash message to assert. I maybe able to add a text widget which can read the text "successfully mapped..." to assert the success?

cfme/v2v/migrations.py
kedark3: This will move out to yaml, isn't it ?

cfme/v2v/migrations.py
kedark3: because there is no next_btn in this view and including the button_set1, brings next_btn by default, so I override it to point to correct button. If you say, we can have close button pointing to next button and both will be basically pointing to Button("Close").  Thoughts?

mshriver: Sounds like it may be better to not include the CommonButtonsSet1 and just directly instantiate the buttons that are present on this view.

cfme/v2v/migrations.py
kedark3: added # TODO as we discussed on irc

cfme/tests/v2v/test_v2v_copy.py
```diff
@@ -0,0 +1,115 @@
...
+        'general':
+            {
+                'name': 'infra_map_{}'.format(fauxfactory.gen_alphanumeric()),
+                'description': fauxfactory.gen_string("alphanumeric", length=50)
+            },
+        'cluster':
+            {
+                'mappings': [
+                    {
+                        'sources': ['source_Datacenter_name \ source_Cluster_name'],
```
mshriver: much better with the generic values here!

I think you can remove the 2nd item in the lists for datastore and network, doesn't need to be that verbose

PR #7250, opened 2018-05-15
[WIP] handle alerts during execute_script

izapolsk:
NEEDS: https://github.com/RedHatQE/widgetastic.core/pull/94/files


dajoRH:
Lint report for commit e47e3c61c2a3151cb3f504ab6115ebf73c78c7f6:

`cfme/utils/appliance/implementations/ui.py`:
- :heavy_exclamation_mark: Line 127:26: **W291** *trailing whitespace*

**No commit flaws detected**.

*CFME QE Bot*

cfme/utils/appliance/implementations/ui.py
```diff
@@ -133,11 +161,15 @@ def make_document_focused(self):
 
     def ensure_page_safe(self, timeout='20s'):
         # THIS ONE SHOULD ALWAYS USE JAVASCRIPT ONLY, NO OTHER SELENIUM INTERACTION
+        if self.browser.page_dirty:
+                if (self.browser.alert_present and
```
mshriver: This indentation looks too far?

mshriver: Consider implementing this as single set of statements:

```
if (self.browser.page_dirty and
       self.browser.alert_present
       self.browser.get_alert().text == 'Abandon changes?'):
    self.browser.handle_alert()
```

izapolsk: it is already implemented

cfme/utils/appliance/implementations/ui.py
```diff
@@ -174,6 +203,13 @@ def before_keyboard_input(self, element, keyboard_input):
         sleep(0.3)
         self.make_document_focused()
 
+    def before_click(self, element):
+        # this is necessary in order to handle unexpected alerts like "Abandon Changes"
+        self.browser.page_dirty = self.page_has_changes
+
+    def after_click(self, element):
+        self.browser.page_dirty = None
```
mshriver: Why set to `None` instead of call self.page_has_changes again? Its possible the click only changed part of a page.

psav: I would imagine to not have to call it again - but it's a good point

izapolsk: it is necessary to check page for changes only before doing something now. Because we need to close "Abandon changes" alert. 

There is a reason to set page_dirty to None or False in after_click. If it was left True, all next calls of ensure_page_safe would check for alert presence. whereas alert methods are enormously slow in selenium. 
btw, I can set it to None/False immediately after closing alert if you wish.
btw, I can also update page_dirty after any manipulation in selenium but it will slow down our tests.

mshriver: Please add some inline comments concerning what you've found @izapolsk, thanks!

PR #7259, opened 2018-05-16
[1LP][RFR] Raw metric test wait_for and relocate 

digitronik:
Purpose or Intent
=================

- Add proper `wait_for`
- moved test file to `candu` dir

{{pytest: cfme/tests/candu/test_utilization_metrics.py -v --use-provider complete --long-running }}


digitronik:
, Network test will fail. As found no iperfserver not running.

rlbabyuk:
@digitronik could you please check and explain PRT, there are some failures there
otherwise looks good

digitronik:
@psav Yes, we have to enable candu. We have a global fixture for that now. https://github.com/ManageIQ/integration_tests/blob/master/cfme/fixtures/candu.py#L7

PR #7266, opened 2018-05-17
[1LP][RFR]Added new collection for taging api testing

rlbabyuk:
Purpose or Intent
=================
New collections for REST taggign testing
{{pytest: cfme/tests/configure/test_tag.py}}

### DIFF
--- a/cfme/tests/configure/test_tag.py
+++ b/cfme/tests/configure/test_tag.py
@@ -200,7 +200,8 @@ class TestTagsViaREST(object):
     @pytest.mark.meta(blockers=[BZ(1451025, forced_streams=['5.7'])])
     @pytest.mark.parametrize(
         "collection_name", ["clusters", "hosts", "data_stores", "providers", "resource_pools",
-        "services", "service_templates", "tenants", "vms"])
+        "services", "service_templates", "tenants", "vms", "availability_zones", "cloud_networks",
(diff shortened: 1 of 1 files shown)


cfme/tests/configure/test_tag.py
```diff
@@ -198,9 +198,16 @@ def test_create_tag_with_wrong_arguments(self, appliance):
 
     @pytest.mark.tier(3)
     @pytest.mark.meta(blockers=[BZ(1451025, forced_streams=['5.7'])])
+    @pytest.mark.uncollectif(
+        lambda appliance, collection_name: appliance.version < '5.9' and
+        collection_name in ["availability_zones", "cloud_networks", "cloud_networks",
+                            "cloud_subnets", "flavors", "network_routers", "security_groups"],
+        reason="5.8 version doesn't support all available collection for tagging"
```
bsquizz: Can you add which collection name we're skipping in the reason so the reason for uncollecting is very clear? For example: 
```python
reason="appliances <5.9 do not support tagging with collection '{}'".format(collection_name)
```

rlbabyuk: collection_name is not accessible for the reason in uncollectif

bsquizz: True, just noticed that's coming from the lambda ;)

psav: We could expose that.....but its probably a black hole

PR #7277, opened 2018-05-22
[1LP][RFR][NOTEST] Adding cloud-init test for RHV

jan-zmeskal:
This PR __activates__ `test_provision_cloud_init` for RHV-CFME integration. 
So what's it about?
- Fixing `provisioning_data` supplied to VM provision request.
- Fixing `required_fiels` of provider.
- Providing new customization template. This probably needs some explanation. The [customization template ](https://github.com/ManageIQ/integration_tests/blob/master/data/basiccloudinit.cfg)that is already presentein integration_tests could not be reused. The reason is that no user is specified there. In such case, cloud-init changes password of the default user. That is fedora for Fedora, ubuntu for Ubuntu etc. However, RHEL (which we use as VM's OS) does not have such user. Therefore I had to provide customization template that changes root password. 

I am not providing any `{ pytest }` clause here, since I don't really know how to test it in PRT. By looking into Ostriz, it seems to me this test case is not being executed at all as of now. However, it passes in our CI, so I can provide verification link on demand.

### DIFF
--- a/cfme/tests/infrastructure/test_cloud_init_provisioning.py
+++ b/cfme/tests/infrastructure/test_cloud_init_provisioning.py
@@ -2,6 +2,8 @@
 import fauxfactory
 import pytest
 
+from widgetastic.utils import partial_match
+
 from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.provisioning import do_vm_provisioning
 from cfme.infrastructure.pxe import get_template_from_config
@@ -16,15 +18,15 @@ pytestmark = [
                          required_fields=[['provisioning', 'ci-template'],
                                           ['provisioning', 'ci-username'],
                                           ['provisioning', 'ci-pass'],
-                                          ['provisioning', 'image'],
+                                          ['provisioning', 'ci-image'],
                                           ['provisioning', 'vlan']],
                          scope='module'),
 ]
 
 
 @pytest.fixture(scope="module")
-def setup_ci_template(provisioning, appliance):
-    cloud_init_template_name = provisioning['ci-template']
+def setup_ci_template(provider, appliance):
+    cloud_init_template_name = provider.data['provisioning']['ci-template']
     get_template_from_config(cloud_init_template_name, create=True, appliance=appliance)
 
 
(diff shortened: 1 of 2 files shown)


lkhomenk:
we have rhv41 provider. which doesn't have 'ci-*' requirements atm.

jan-zmeskal:
@lkhomenk Yes, I am aware of that

cfme/tests/infrastructure/test_cloud_init_provisioning.py
```diff
@@ -57,9 +59,10 @@ def test_provision_cloud_init(appliance, setup_provider, provider, setup_ci_temp
             'host_name': {'name': host},
             'datastore_name': {'name': datastore}},
         'network': {
-            'vlan': vlan},
+            'vlan': partial_match(vlan)},
```
bsquizz: What is `partial_match` needed for here?

cfme/tests/infrastructure/test_cloud_init_provisioning.py
```diff
@@ -17,15 +19,15 @@
...
                                           ['provisioning', 'vlan']],
                          scope='module'),
 ]
 
 
 @pytest.fixture(scope="module")
-def setup_ci_template(provisioning, appliance):
-    cloud_init_template_name = provisioning['ci-template']
+def setup_ci_template(provider, appliance):
+    cloud_init_template_name = provider.data.get('provisioning').get('ci-template')
```
bsquizz: cloud_init_template_name can't be `None` when passed into `get_template_from_config`, it'll cause things in that method to fail. I think it'd be better to not use `.get()` here and instead hit the `KeyError` if those keys are not found

cfme/tests/infrastructure/test_cloud_init_provisioning.py
jan-zmeskal: @bsquizz That's because the actual vlan name is 'ovirtmgmt' but provisioning dialog displays it as 'ovirtmgmgt (ovirtmgmt)' . The displayed name is  formatted as '<vlan> (<vNIC>)' if I remember correctly.

cfme/tests/infrastructure/test_cloud_init_provisioning.py
jan-zmeskal: @bsquizz Ok, will do

dajoRH:
*CFME QE Bot*

bsquizz:
OK, let's get the changes to yamls merged and see if we can try a PRT run with this :)

jan-zmeskal:
@bsquizz Actually there is not a PR to cfme-qe-yamls that's associated with it. In RHV-CFME integration project, we have custom RHV provider (rhv_cfme_integration). Its configuration is mainainted in our gerrit repo.

PR #7282, opened 2018-05-23
[RFR] Fix locator in SettingsGroupSubmenu

mshriver:
Locator forcing the `expandable` method to return False, resulting in `items` returning 'Change Groups' when the UI was actually expandable, instead of the list of group names.

Tested locally, setting to RFR since its so small.

{{ pytest: cfme/tests/configure/test_access_control.py -k 'test_user_assign_multiple_groups or test_user_change_groups' }}


LandoCalrizzian:
Looks good. This is a more complete XPath locator for finding the group submenu

PR #7290, opened 2018-05-24
wrapanapi 2.9.3 for rhevm.mark_as_template fix

mshriver:


mshriver:
travis will probably fail because its not quite in pypi yet.

PR #7300, opened 2018-05-28
[RFR] Adding commit signature information in the docs

saulotoledo:
Purpose or Intent
=================

__Updating__ the documentation to have commit signature information because users should sign their commits before creating pull requests.


docs/guides/dev_guide.rst
```diff
@@ -234,7 +241,7 @@ We also do a few things that aren't explicitly called out in PEP 8:
             Nothing at all.
 
         Raises:
-	    CertainQuestionsError: Raises certain questions about the authors sanity.
+      CertainQuestionsError: Raises certain questions about the authors sanity.
```
psav: **REQUIRED:** This formatting is not correct

docs/guides/dev_guide.rst
```diff
@@ -310,7 +317,7 @@ Layout
     fixture that exposes the utility to the tests. Modules in this directory
     will be auto loaded.
 
-  	* ``tests/`` Unit tests for utils
+    * ``tests/`` Unit tests for utils
```
psav: **REQUIRED:** This also is at the right level

docs/guides/dev_guide.rst
```diff
@@ -35,13 +35,20 @@ Contributing
...
 * All pull requests should be squashed down to logical blocks of distinctive
-  functionality that work by themselves and do not result in brokenness of master
+  functionality that work by themselves and do not result in brokenness of master.
 
   * As an example, if you were working on a test which required new pages,
     utilities and tests, it would be OK to split the page, utility and test
     changes into separate requests or commits, providing they were in the correct
     order of dependency.
 
+* All commits should be signed.
```
psav: **REQUIRED:** GPG signing is good practice, however the only thing it is required for is PRT testing.

docs/guides/dev_guide.rst
saulotoledo: These lines were using tab for indent. I replaced 1 tab for 2 spaces and did not notice that. I will fix it, thanks!

docs/guides/dev_guide.rst
saulotoledo: I am not sure about this one. It is already one level from the previous indent. I replaced 1 tab with 2 spaces here too.

docs/guides/dev_guide.rst
saulotoledo: I added as "should" because we will always have those 2 errors (ci/downstream-58z and ci/downstream-59z). They are also difficult to understand and sound like a problem. Are you sure I should change this text? I was thinking about update the text to mention the 2 errors somehow.

saulotoledo: I noticed that the checkings were removed. Is this just a temporary change?

saulotoledo: Done. But I added more information about the checks. Could you please review again? I can remove the new info if you prefer.

saulotoledo:
@psav Could you review again?

docs/guides/dev_guide.rst
saulotoledo: As we talked on Gitter, that should be ok, right?
Are you ok with the text changes too? Thanks in advance!

PR #7302, opened 2018-05-29
[1LP][RFR] Gap Collection C&U

digitronik:
Purpose or Intent
=================
- Gap collection tests for VM and Host

{{pytest: cfme/tests/candu/test_gap_collection.py -v --use-provider complete}}

### DIFF
--- a/cfme/tests/candu/test_gap_collection.py
+++ b/cfme/tests/candu/test_gap_collection.py
@@ -1,0 +1,88 @@
+import pytest
+from datetime import datetime, timedelta
+
+from cfme import test_requirements
+from cfme.infrastructure.provider.virtualcenter import VMwareProvider
+from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.wait import wait_for
+
+
+pytestmark = [
+    pytest.mark.tier(3),
+    test_requirements.c_and_u,
+    pytest.mark.usefixtures('setup_provider'),
+    pytest.mark.provider([VMwareProvider],
+    scope='module',
+    required_fields=[(['cap_and_util', 'capandu_vm'], 'cu-24x7')])
+]
+
+
+ELEMENTS = ['vm', 'host']
+GRAPH_TYPE = ['hourly', 'daily']
+
+
+@pytest.fixture(scope='module')
+def order_data(appliance, provider, enable_candu):
+    # Order two day back gap collection data for testing
+    end_date = datetime.now()
+    start_date = end_date - timedelta(days=2)
+
+    view = navigate_to(appliance.server.zone, 'CANDUGapCollection')
+    view.candugapcollection.end_date.fill(end_date)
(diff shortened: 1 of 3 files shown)


dajoRH:
*CFME QE Bot*

dajoRH:
Lint report for commit cf911b6b6c4487eb9104da9a91ee39d76929de22:

- :red_circle: Line 6:1: **F401** *'datetime.datetime' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/candu/test_gap_collection.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+@pytest.fixture(scope='module')
+def order_data(appliance, provider, enable_candu):
+    # Order two day back gap collection data for testing
+    end_date = datetime.now()
+    start_date = end_date - timedelta(days=2)
+
+    view = navigate_to(appliance.server.zone, 'CANDUGapCollection')
+    view.candugapcollection.end_date.fill(end_date)
```
rlbabyuk: Required: 
view.candugapcollection.fill{
'end_date': end_date,
'start_date': start_date
}

cfme/tests/candu/test_gap_collection.py
```diff
@@ -0,0 +1,88 @@
...
+        * C&U enabled appliance
+
+    Steps:
+        * Navigate to Configuration > Diagnostics > Zone Gap Collection Page
+        * Order old data
+        * Navigate to VM or Host Utilization page
+        * Check for Hourly data
+        * Check for Daily data
+    """
+    if element == 'host':
```
rlbabyuk: Required: This looks like a setup, Suggest moving it to parametrize fixture which will return you a correct object
line 51-59

cfme/tests/candu/test_gap_collection.py
digitronik: Yes, Its possibly right solution but I am adding it as workaround means first fill `end_date` and after that `start_date`. As per my knowledge; passing `dict` not fallow order. Please, correct me If I am wrong. As I am playing with `all Vsphere providers`. I need to execute it for each and `calendar` will block selection, if I select first `start_date` then end date will `disable` as gap collection already executed for `first provider`.

rlbabyuk: They are filled in the same order as listed in the view. In your view, start_date is the first, and it filled first. If end date always should be filled fist just move it to the top in view class

cfme/tests/candu/test_gap_collection.py
digitronik: As per our discussion on IRC, fixture parameter will run before provider. So, better to go with same. Its faster (for one provider run all `vm-hourly`, `vm-daily`, and `host-hourly`, `host-daily`) and nice id.

digitronik:
@rlbabyuk: I added changes

PR #7307, opened 2018-05-30
[1LP][RFR] Added test_automate_git_domain_removed_from_disk

quarckster:
{{pytest: -v -k test_automate_git_domain_removed_from_disk}}

### DIFF
--- a/cfme/automate/import_export.py
+++ b/cfme/automate/import_export.py
@@ -83,7 +83,7 @@ class AutomateGitRepository(Navigatable):
         git_select = self.create_view(GitImportSelectorView)
         assert git_select.is_displayed
         git_select.flash.assert_no_error()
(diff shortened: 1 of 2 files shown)


cfme/automate/import_export.py
```diff
@@ -83,7 +84,7 @@ def import_domain_from(self, branch=None, tag=None):
         git_select = self.create_view(GitImportSelectorView)
         assert git_select.is_displayed
         git_select.flash.assert_no_error()
-        assert git_select.fill(self.fill_values_branch_select(branch, tag))
```
quarckster: Doesn't make sense if want to import from origin/master.

psav: Is this **REQUIRED**?

quarckster: For this test is not. I found this issue when I tried to import a domain from master branch.

PR #7311, opened 2018-05-30
[1LP][RFR]adding azure provider tests

lkhomenk:
Purpose or Intent
=================
__Extending__ provider tests

{{pytest: --long-running --use-provider complete -k test_azure cfme/tests/cloud/test_providers.py}}

### DIFF
--- a/cfme/tests/cloud/test_providers.py
+++ b/cfme/tests/cloud/test_providers.py
@@ -11,7 +11,7 @@ from cfme import test_requirements
 from cfme.base.credential import Credential
 from cfme.cloud.instance import Instance
 from cfme.cloud.provider import CloudProvider
-from cfme.cloud.provider.azure import AzureProvider
+from cfme.cloud.provider.azure import AzureProvider, AzureEndpoint
 from cfme.cloud.provider.ec2 import EC2Provider
 from cfme.cloud.provider.gce import GCEProvider
 from cfme.cloud.provider.openstack import OpenStackProvider, RHOSEndpoint
@@ -20,6 +20,7 @@ from cfme.common.provider_views import (
 from cfme.rest.gen_data import _creating_skeleton as creating_skeleton
 from cfme.rest.gen_data import arbitration_profiles as _arbitration_profiles
 from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils import conf
 from cfme.utils.rest import (
     assert_response,
     delete_resources_from_collection,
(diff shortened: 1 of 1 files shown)


cfme/tests/cloud/test_providers.py
```diff
@@ -353,6 +354,68 @@ def test_api_port_max_character_validation_cloud(appliance):
...
+def test_azure_subscription_required(request, provider):
+    """
+    Tests that provider can't be added w/o subscription
+
+    Metadata:
+        test_flag: crud
+    """
+    provider.subscription_id = ''
+    request.addfinalizer(provider.delete_if_exists)
+    flash = 'Credential validation was not successful: ' \
```
rlbabyuk: Required: Use brackets here instead of \
('Credential validation was not successful: '
 'Incorrect credentials - check your Azure Subscription ID')

cfme/tests/cloud/test_providers.py
```diff
@@ -353,6 +354,68 @@ def test_api_port_max_character_validation_cloud(appliance):
...
+        crud_provider.validate_stats()
+        prov_inventory.append((crud_provider.name,
+                               crud_provider.num_vm(),
+                               crud_provider.num_template()))
+
+    for index, prov_a in enumerate(prov_inventory[:-1]):
+        for prov_b in prov_inventory[index + 1:]:
+            assert prov_a[1] != prov_b[1], "Same num_vms for {} and {}".format(prov_a[0],
+                                                                               prov_b[0])
+            assert prov_a[2] != prov_b[2], "Same num_templates for {} and {}".format(prov_a[0],
```
rlbabyuk: Required: use soft_assert here

cfme/tests/cloud/test_providers.py
```diff
@@ -353,6 +354,68 @@ def test_api_port_max_character_validation_cloud(appliance):
...
+    Verifies that different azure providers have different resources access
+
+    Steps:
+    1. Add all Azure providers
+    2. Compare their VMs/Templates
+
+    Metadata:
+        test_flag: crud
+    """
+    providers = conf.cfme_data.get("management_systems", {})
```
psav: **REQUIRED:** I would prefer to see the filter mechanism here rather than a new mechanism, we have a flag here that would not be respected, and global filters that would be ignored too. I see the filter mechanism as the only way to do it.

cfme/tests/cloud/test_providers.py
```diff
@@ -353,6 +354,68 @@ def test_api_port_max_character_validation_cloud(appliance):
...
+    azure_providers = []
+    for provider in providers.items():
+        if provider[1].get('type') == 'azure':
+            azure_providers.append(provider)
+    prov_inventory = []
+    for provider in azure_providers:
```
psav: **REQUIRED:** There is no protection here against there being only 1 azure provider.

cfme/tests/cloud/test_providers.py
lkhomenk: your advice simplified everything totally

PR #7313, opened 2018-05-31
[1LP][RFR]fixing azure event tests

lkhomenk:
Purpose or Intent
=================

__Fixing__  automation
{{pytest: --use-provider azure cfme/tests/cloud/test_cloud_events.py}}


lkhomenk:
needs MR 520

lkhomenk:
@izapolsk locally it works - can't pass PRT. As I recall - it's a know issue - we depend on the provider which runs appliance, right?

PR #7318, opened 2018-06-01
[1LP][RFR] SystemImage and ISO datastore changes

jan-zmeskal:
So, this PR does many things, let's look at them:

- Providing __new model__ of ISO SystemImage that can be found in ISO datastore. I mean this:
image
I did this because I want to be able to change `Type` of such ISO image. It might seem like an overkill. I also thought so. But then I realize I cannot really use existing models/views/navigation steps for this purpose without too much hacking. The approach that was used originally (`iso_datastore.set_iso_image_type`) simply was not working, because it was using wrong view (`ISODatastoreAll`). I believe this proposed approach is more in line with the overall philosophy of the framework and is extensible if the need arises. And it will (at least for me) in the near future.

- `test_iso_datastore_crud` now runs not only if provider definition contains `iso_datastore` key, it must also be set to `True`. This makes much more sense. Up to this point, it was executed whet a provider had `iso_datastore: False`, which I am sure was not the intention.

- `setup_provider` was used as a fixture for `test_iso_provision_from_template`. I moved it to be a fixture of `datastore_init` fixture. At that point, we already have to have provider, otherwise this fixture fails.

I hope there's no big issue with this. The reality is that `test_iso_provision_from_template` is not being executed at all at the moment. I want to get it to working state at least for RHV-CFME integration project and these are some of the prerequisities I need.

{{pytest: cfme/tests/infrastructure/test_iso_datastore.py::test_iso_datastore_crud -vv}}

### DIFF
--- a/cfme/infrastructure/pxe.py
+++ b/cfme/infrastructure/pxe.py
@@ -678,21 +678,33 @@ class PXESystemImageTypeEditView(PXESystemImageTypeForm):
     cancel = Button('Cancel')
 
 
-class SystemImage(Updateable, Navigatable):
+@attr.s
+class SystemImage(Updateable, BaseEntity):
+    """Model of an ISO System Image in CFME.
 
-    def __init__(self, name=None, image_type=None, datastore=None, appliance=None):
-        Navigatable.__init__(self, appliance=appliance)
-        self.name = name
-        self.image_type = image_type
-        self.datastore = datastore
+    Args:
+        name: The name of the System Image. It's the same as ISO filename in ISO domain
+        image_type: SystemImageType object
+        datastore: ISODatastore object
+    """
+    name = attr.ib(default=None)
+    image_type = attr.ib(default=None)
+    datastore = attr.ib(default=None)
 
     def set_image_type(self):
+        """Changes the Type field in Basic Information table for System Image in UI."""
         view = navigate_to(self, 'Edit')
-        changed = view.image_type.fill_with(self.image_type)
+        changed = view.image_type.fill_with(self.image_type.name)
         if changed:
             view.save.click()
         else:
             view.cancel.click()
+
+
+@attr.s
+class SystemImageCollection(BaseCollection):
+
+    ENTITY = SystemImage
 
 
 class PXESystemImageDeatilsView(PXEMainView):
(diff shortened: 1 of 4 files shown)


cfme/infrastructure/pxe.py
```diff
@@ -678,6 +678,68 @@ class PXESystemImageTypeEditView(PXESystemImageTypeForm):
     cancel = Button('Cancel')
 
 
+class SystemImage(Updateable, Navigatable):
```
quarckster: **REQUIRED**
Please make it as a collection.

jan-zmeskal: @quarckster That's actually one thing I'm not so sure about. I don't think that SystemImage really fits into definition of entity, since you cannot really create or delete one.

quarckster: It doesn't matter, you can instantiate an entity from a collection. Also we should'n use `Navigatable` directly.

jan-zmeskal: @quarckster Okay, I'll look into that once I have some time. I suppose it will take some time since I never created one.

dajoRH:
Lint report for commit 93a7e004c5343e8ce13dd7ea46998b6424ba7985:

`cfme/tests/infrastructure/test_iso_provisioning.py`:
- :red_circle: Line 9:1: **F401** *'cfme.infrastructure.pxe.SystemImage' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

jan-zmeskal:
@quarckster I have transformed `SystemImage` into collection/entity. Hope I did not omit anything important.

dajoRH:
Lint report for commit dd0476c6a141da026dc9e030c5ac1c02505b0925:

`cfme/infrastructure/pxe.py`:
- :red_circle: Line 717:11: **N801** *class names should use CapWords convention*

**No commit flaws detected**.

*CFME QE Bot*

jan-zmeskal:
PRT fails on 5.10 because it [cannot log into appliance](http://10.16.45.124/7QKbFhOk/artifacts/cfme/tests/infrastructure/test_iso_datastore.py/test_iso_datastore_crud[rhevm-4.1]/filedump-exception_screenshot.png).

PR #7319, opened 2018-06-01
[RFR][1LP] Map Tags to Collections

rlbabyuk:
Purpose or Intent
=================
Convert map tagging to collections
{{pytest: -vvv -k test_paginator}}


rlbabyuk:
Failures for 5.8 not caused by this PR changes

PR #7323, opened 2018-06-03
[WIPTEST] Fix smart management tests

shalomnaim1:
{{ pytest: cfme/tests/containers/test_smart_management.py --use-provider ocp-37-hawk }}

Fix smart management broken tests

### DIFF
--- a/cfme/common/__init__.py
+++ b/cfme/common/__init__.py
@@ -348,6 +348,7 @@ class Taggable(TaggableCommonBase):
             details (bool): set False if tag should be added for list selection,
                             default is details page
         """
+
         step = 'EditTagsFromDetails' if details else 'EditTags'
         view = navigate_to(self, step)
         added_tag, updated = self._assign_tag_action(view, tag)
--- a/cfme/common/__init__.py
+++ b/cfme/common/__init__.py
@@ -274,7 +274,7 @@ class TaggableCommonBase(object):
         """
         if not tag:
             tag = self._set_random_tag(view)
-        category_name = tag.category.display_name
+        category_name = tag.category
         tag_name = tag.display_name
         # Handle nested view.form and where the view contains form widgets
         try:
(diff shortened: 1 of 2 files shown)


dajoRH:
*CFME QE Bot*

cfme/common/provider_views.py
```diff
@@ -115,6 +116,7 @@ class ProviderDetailsSummaryView(View):
         represents Details page when it is switched to Summary aka Tables view
         """
         summary = ParametrizedSummaryTable()
+        smart_management = SummaryTable(title="Smart Management")
```
jawatts: You do not need this, the Smart Management table is defined by the ParametrizedSummaryTable.

Example:
```python
view = navigate_to(provider, 'Details')
fields = view.entities.summary('Smart Management').fields
```

cfme/common/__init__.py
```diff
@@ -274,7 +274,7 @@ def _assign_tag_action(self, view, tag):
         """
         if not tag:
             tag = self._set_random_tag(view)
-        category_name = tag.category.display_name
+        category_name = tag.category
```
jawatts: This is not correct. category_name should be a string, not an object. 

https://github.com/ManageIQ/integration_tests/blob/master/cfme/configure/configuration/region_settings.py#L75

cfme/common/provider_views.py
shalomnaim1: @jawatts Thanks, I will try this

cfme/common/__init__.py
shalomnaim1: @jawatts For some reason ```tag.category``` does not contain the attr ```display_name```
when I investigated this I found that ```tag. category``` is the category name.

jawatts: @shalomnaim1 That is because you need to instantiate the category object. In the smart management test, you were passing it as a string.

shalomnaim1: @jawatts So what it the right way to use this object?
could you please show me an example?

jawatts: @shalomnaim1 Here is an example: https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/cloud_infra_common/test_vm_instance_analysis.py#L768

dajoRH:
*CFME QE Bot*

PR #7325, opened 2018-06-04
[WIPTEST] Added test_embed_tower_exec_play_against_amazon

quarckster:
{{pytest: -v -k test_embed_tower_exec_play_against_amazon --long-running}}


quarckster:
5.8 failed because embedded ansible role cannot be enabled. Should be fixed in the next build.

cfme/tests/ansible/test_embedded_ansible_services.py
```diff
@@ -441,3 +457,27 @@ def test_ansible_group_id_in_payload(service_catalog, service_request, service):
...
+    def _revert():
+        with update(ansible_catalog_item):
+            ansible_catalog_item.provisioning = {
+                "playbook": "dump_all_variables.yml",
+                "cloud_type": "<Choose>"
+            }
+
+    request.addfinalizer(_revert)
+    service_request = service_catalog.order()
+    service_request.wait_for_request(method="ui", num_sec=200 * 60, delay=120)
```
bsquizz: Just a nitpick here -- does the delay need to be 120sec? Could we shave some time off the test execution with a shorter delay?

cfme/tests/ansible/test_embedded_ansible_services.py
```diff
@@ -441,3 +457,27 @@ def test_ansible_group_id_in_payload(service_catalog, service_request, service):
...
+        ansible_catalog_item.provisioning = {
+            "playbook": "list_ec2_instances.yml",
+            "cloud_type": "Amazon",
+            "cloud_credential": ansible_amazon_credential.name
+        }
+
+    def _revert():
+        with update(ansible_catalog_item):
+            ansible_catalog_item.provisioning = {
+                "playbook": "dump_all_variables.yml",
```
bsquizz: I assume we're hard-coding this playbook name because it is usually the first in the drop-down? Should we revert with something a little 'smarter' that does not use a hard-coded name? I.e. revert to <whatever playbook no. 1 is>

quarckster: I don't think so, tests in this module use exactly that playbook and only few of them use other playbooks.

quarckster:
@bsquizz please mark you review comments either **REQUIRED** or **OPTIONAL**

cfme/tests/ansible/test_embedded_ansible_services.py
```diff
@@ -441,3 +457,27 @@ def test_ansible_group_id_in_payload(service_catalog, service_request, service):
...
+                "cloud_type": "<Choose>"
+            }
+
+    request.addfinalizer(_revert)
```
mshriver: Consider decorating cleanup functions instead of a separate call - for the future :)

cfme/tests/ansible/test_embedded_ansible_services.py
```diff
@@ -441,3 +457,27 @@ def test_ansible_group_id_in_payload(service_catalog, service_request, service):
...
+    service_request = service_catalog.order()
+    service_request.wait_for_request(method="ui", num_sec=200 * 60, delay=20)
```
mshriver: I'm thinking this is a typo for a 20 minute timeout? 200 minutes is far too long.

PR #7336, opened 2018-06-06
[1LP][RFR] Remove arg in template_upload_rhevm

mshriver:
extra arg was put into upload_template call, removed, unecessary

### DIFF
--- a/scripts/template_upload_rhevm.py
+++ b/scripts/template_upload_rhevm.py
@@ -146,6 +146,7 @@ def template_from_ova(mgmt, username, password, rhevip, edomain, ovaname, ssh_cl
         temp_template_name: temporary template name (this template will be deleted)
     """
     try:
+        change_edomain_state(mgmt, 'active', edomain, provider)
         if mgmt.get_template_from_storage_domain(temp_template_name, edomain):
             logger.info("RHEVM:%r Warning: found another template with this name.", provider)
             logger.info("RHEVM:%r Skipping this step. Attempting to continue...", provider)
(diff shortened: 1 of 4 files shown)


scripts/template_upload_rhevm.py
```diff
@@ -198,8 +200,14 @@ def import_template(mgmt, edomain, sdomain, cluster, temp_template_name, provide
             logger.info("RHEVM:%r The template failed to import on data domain", provider)
             sys.exit(127)
         logger.info("RHEVM:%r successfully imported template on data domain", provider)
+        logger.info('RHEVM:%r updating nic on template')
```
izapolsk: REQUIRED: it seems you missed argument

scripts/template_upload_rhevm.py
```diff
@@ -177,9 +178,10 @@ def template_from_ova(mgmt, username, password, rhevip, edomain, ovaname, ssh_cl
         logger.info("RHEVM:%r successfully created template from ova file:", provider)
     except Exception:
         logger.exception("RHEVM:%r template_from_ova failed:", provider)
+        raise
 
 
-def import_template(mgmt, edomain, sdomain, cluster, temp_template_name, provider):
+def import_template(mgmt, cfme_data, edomain, sdomain, cluster, temp_template_name, provider):
```
izapolsk: honestly, I don't like such way to pass data. it makes everyone to follow the way how data is stored in our yamls.

izapolsk: I'd rather update this method to accept properly prepared dict or config object

izapolsk: ah, nevermind, I didn't notice that I switched from wrapanapi to our internal scripts. ignore above statements.

PR #7337, opened 2018-06-07
[1LP][RFR] Fixing Container Images perform_smartstate_analysis with correct task name

brahmanim:
Following cfme/configure/tasks.py update that has been done,

I had to fix cfme/containers/image.py perform_smartstate_analysis with correct task name,
That includes "Container Image Analysis: " prefix on task name.

Also, an update regarding  perform_smartstate_analysis_multiple_images,
update the following functions regarding TasksCollection:
is_finished, is_successfully_finished and wait_for_finished.

{{pytest: cfme/tests/containers/test_containers_smartstate_analysis.py -v --use-provider ocp-37-hawk }}

### DIFF
--- a/cfme/configure/tasks.py
+++ b/cfme/configure/tasks.py
@@ -3,6 +3,7 @@
 """ Module dealing with Configure/Tasks section.
 """
 import attr
+import re
 
 from navmazing import NavigateToAttribute
 from widgetastic.exceptions import RowNotFound
@@ -226,8 +227,8 @@ class TasksCollection(BaseCollection):
     def is_finished(self, *tasks):
         view = navigate_to(self, self.tab)
         tab_view = getattr(view.tabs, self.tab.lower())
-        for row in tab_view.table.rows:
-            if row.name in tasks and row.state.lower() != "finished":
+        for row in tab_view.table.rows():
+            if row.task_name.text in tasks and row.state.text.lower() != "finished":
                 return False
         return True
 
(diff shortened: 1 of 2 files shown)


cfme/containers/image.py
```diff
@@ -203,24 +203,26 @@ def perform_smartstate_analysis_multiple_images(
         # task_name change from str to regular expression
         # the str compile on tasks module
         image_enities_names = []
+        delay = 5
+        silent_failure = True
```
lkhomenk: QUESTION: why are there hardcoded?

brahmanim: No special reason, Just to understand your question, what is your suggestion?
Are you suggest to use this variable values directly on the function call?

lkhomenk: @brahmanim  delay is already defaulted to 5 - https://github.com/ManageIQ/integration_tests/pull/7337/files/6bc76133010fd1e4abcf5f2273c39d001e4639c8#diff-1fe1d4bbe572be5ebf426e918f68f31fR261
and `silent_failure` better to pass as attribute directly into `is_successfully_finished`. I don't see a need in additional local variables

brahmanim: OK, just wanted to be sure before I make changes.

brahmanim: Since i Will move back to WIP, Are you have any forther comments?

lkhomenk: nothing else ;) PR is fine. and even more - green PRT is here!

brahmanim:
@lkhomenk Local Variables, delay & silent_failure, removed.

cfme/containers/image.py
```diff
@@ -209,18 +209,18 @@ def perform_smartstate_analysis_multiple_images(
         images_view.toolbar.configuration.item_select(
             'Perform SmartState Analysis', handle_alert=True)
         for image_entity in image_entities:
-            image_enities_names.append(image_entity.name)
+            image_enities_names.append("Container Image Analysis: '{}'".format(image_entity.name))
```
izapolsk: REQUIRED: typo

cfme/configure/tasks.py
```diff
@@ -226,17 +227,24 @@ def delete_all(self):
...
+            if row.task_name.text in tasks and row.state.text.lower() != "finished":
                 return False
         return True
 
     def is_successfully_finished(self, silent_failure=False, *tasks):
         view = navigate_to(self, self.tab)
         tab_view = getattr(view.tabs, self.tab.lower())
         rows = []
+        # expected_status support also regular expression pattern
+        expected_status = re.compile('finished', re.IGNORECASE)
```
izapolsk: OPTIONAL: I'd rather use "in" or "contains" but this also is ok.

izapolsk: yeah, there is no good way to avoid using regexp

PR #7338, opened 2018-06-07
[RFR] [NOTTEST] Improve locator of LineChart

digitronik:
Purpose or Intent
=================
- Improve locator so it will applicable for Planning Line chart and other as well
- Added locator so user can directly set locator.

### DIFF
--- a/widgetastic_manageiq/__init__.py
+++ b/widgetastic_manageiq/__init__.py
@@ -4064,7 +4064,7 @@ class LineChart(Widget, ClickableMixin):
        chart_name = LineChart(id='miq_chart_parent_candu_0')
     """
 
-    ROOT = ParametrizedLocator(".//div[@id='candu_charts_div']//div[@id='{@id}']")
+    ROOT = ParametrizedLocator("{@locator}")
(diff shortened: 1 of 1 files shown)


widgetastic_manageiq/__init__.py
```diff
@@ -4073,10 +4073,15 @@ class LineChart(Widget, ClickableMixin):
     tooltip = Table(locator='.//div[contains(@class,"c3-tooltip-container")]/table')
     LEGENDS = ".//*[contains(@class, 'c3-legend-item c3-legend-item-')]"
 
-    def __init__(self, parent, id, logger=None):
+    def __init__(self, parent, id=None, locator=None, logger=None):
         """Create the widget"""
         Widget.__init__(self, parent, logger=logger)
-        self.id = id
+        if id:
+            self.locator = ".//*[@id={}]".format(quote(id))
```
izapolsk: OPTIONAL: @digitronik, changes look good but I don't like using such fully abstract xpath.

digitronik: @izapolsk: Accept, but I am not getting supporting elements to improve more static xpath.

PR #7340, opened 2018-06-07
[1LP][WIPTEST] Test: azure_provider_discovery_error

digitronik:
Purpose or Intent
=================
- Automating test  RHCF3-11608 
- This test similar to the provider discovery, only with false data also discovery will start but provider will not apper. 
 
{{pytest: cfme/tests/cloud/test_providers.py::test_discovery_error_azure_cloud}}

### DIFF
--- a/cfme/tests/cloud/test_providers.py
+++ b/cfme/tests/cloud/test_providers.py
@@ -9,7 +9,6 @@ from widgetastic.exceptions import MoveTargetOutOfBoundsException
 
 from cfme import test_requirements
 from cfme.base.credential import Credential
-from cfme.cloud.instance import Instance
 from cfme.cloud.provider import CloudProvider
 from cfme.cloud.provider.azure import AzureProvider
 from cfme.cloud.provider.ec2 import EC2Provider
@@ -97,8 +96,8 @@ def test_discovery_password_mismatch_validation_cloud(appliance):
 
 
 @pytest.mark.tier(3)
-@pytest.mark.uncollectif(lambda: store.current_appliance.version >= '5.9',
-                         reason='no more support for cloud provider discovery')
+@pytest.mark.uncollectif(lambda appliance: appliance.version >= '5.9',
+                        reason='no more support for cloud provider discovery')
 def test_discovery_error_azure_cloud(appliance):
     """ Test Azure discovery with feck data
 
--- a/cfme/tests/cloud/test_providers.py
+++ b/cfme/tests/cloud/test_providers.py
@@ -19,7 +19,6 @@ from cfme.common.provider_views import (
 from cfme.rest.gen_data import _creating_skeleton as creating_skeleton
 from cfme.rest.gen_data import arbitration_profiles as _arbitration_profiles
 from cfme.utils.appliance.implementations.ui import navigate_to
-from cfme.utils.log import logger
 from cfme.utils.providers import list_providers, ProviderFilter
 from cfme.utils.rest import (
     assert_response,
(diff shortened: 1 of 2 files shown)


dajoRH:
*CFME QE Bot*

cfme/tests/cloud/test_providers.py
```diff
@@ -94,6 +96,43 @@ def test_discovery_password_mismatch_validation_cloud(appliance):
     view.flash.assert_message('Password/Verify Password do not match')
 
 
+@pytest.mark.tier(3)
+@pytest.mark.uncollectif(lambda: store.current_appliance.version >= '5.9',
+                         reason='no more support for cloud provider discovery')
```
quarckster: **REQUIRED**
```python
@pytest.mark.uncollectif(lambda appliance: appliance.version >= '5.9',
                         reason='no more support for cloud provider discovery')
```

cfme/tests/cloud/test_providers.py
```diff
@@ -94,6 +96,43 @@ def test_discovery_password_mismatch_validation_cloud(appliance):
...
+    initial_count = len(view.entities.entity_names)
+
+    collection.discover(cred, AzureProvider)
+    view = appliance.browser.create_view(CloudProvidersView)
+    view.flash.assert_success_message('Cloud Providers: Discovery successfully initiated')
+
+    try:
+        collection.wait_for_new_provider()
+    except TimedOutError:
+        logger.warning('Timed out waiting for new provider, continuing')
```
quarckster: **QUESTION**
What's the point in this exception?

digitronik: It's not necessary but we are handling exception so better to log it.

quarckster: Ok, so why do you wait for a new provider if it's not necessary?

quarckster:
```python
with pytest.raises(TimedOutError):
    collection.wait_for_new_provider()
```
looks much better

cfme/tests/cloud/test_providers.py
```diff
@@ -94,6 +94,43 @@ def test_discovery_password_mismatch_validation_cloud(appliance):
     view.flash.assert_message('Password/Verify Password do not match')
 
 
+@pytest.mark.tier(3)
+@pytest.mark.uncollectif(lambda appliance: appliance.version >= '5.9',
+                        reason='no more support for cloud provider discovery')
+def test_discovery_error_azure_cloud(appliance):
+    """ Test Azure discovery with feck data
```
mshriver: 'fake', typo :)

cfme/tests/cloud/test_providers.py
```diff
@@ -94,6 +94,43 @@ def test_discovery_password_mismatch_validation_cloud(appliance):
...
+
+    prerequisites:
+        * appliance supporting discovery
+
+    Steps:
+        * Navigate Cloud provider discovery and select Azure
```

cfme/tests/cloud/test_providers.py
mshriver: Though this might be my new favorite fake curse word, as it sounds just like another four letter F word.

cfme/tests/cloud/test_providers.py
```diff
@@ -94,6 +94,43 @@ def test_discovery_password_mismatch_validation_cloud(appliance):
...
+    view = navigate_to(collection, 'All')
+    initial_count = len(view.entities.entity_names)
+
+    collection.discover(cred, AzureProvider)
+    view = appliance.browser.create_view(CloudProvidersView)
+    view.flash.assert_success_message('Cloud Providers: Discovery successfully initiated')
+
+    # While waiting for new provider, TimeOutError will come (Negative Test)
+    with pytest.raises(TimedOutError):
+        collection.wait_for_new_provider()
```
mshriver: I think for this to be effective for use in tests like this, the method you're using needs to be updated to take a timeout kwarg, and pass it to the wait_for call. Currently it waits 1000s, or ~16 minutes, which is far too long for a test like this to just wait for an exception to occur that we're fully expecting.  Please add the required kwarg, and pass a much shorter timeout (maybe only 60s?) for this test. Currently it takes ~22m to run.

https://github.com/ManageIQ/integration_tests/blob/e4a5fa4e37f467f4e86784477ae596b948b58ab3/cfme/cloud/provider/__init__.py#L189

Secondly, is there ever another flash message for when discovery fails? Any further validation you could do that the discovery process has actually completed and failed and not just timed out?

digitronik: @mshriver I added a timeout. Now, it's taking 507 sec. Further validation not possible. I have to go for log check; no flash message for discovery fails.

PR #7295, opened 2018-05-25
[1LP][RFR] Adding new widget to capture migrations not started/completed list in v2v

kedark3:
Purpose or Intent
=================

__Adding widget__ for v2v migration not started/completed list

**UPSTREAM ONLY**

{{pytest: cfme/tests/v2v/test_v2v.py -v }}

PR #7343, opened 2018-06-07
[1LP][RFR] Added new widget Infra Mappings List

kedark3:
Purpose or Intent
=================

__Adding New Widget__ for Infrastructure Mapping List. It is partially COMPLETE, needs few enhancements, which have not been delivered by ManageIQ Devel Team yet, as they provide those, I will extend this PR, if they come before this PR is merged, I will update PR with those changes, else submit a different PR later. 

**UPDATE**: Aforementioned enhancements  still far out in devel hence I'd like to request this PR to merge. Enhancements will be submitted in different PR later.

**UPDATE 2**:Received changes from Devel, updated PR to reflect everything. It should be complete now.

{{pytest: cfme/tests/v2v/test_v2v.py -v }}

### DIFF
--- a/widgetastic_manageiq/__init__.py
+++ b/widgetastic_manageiq/__init__.py
@@ -4437,3 +4437,157 @@ class MigrationPlansList(Widget):
             raise TypeError("Fill can only take one value from list to click on.")
         self.select_plan(value)
         return True
+
+
+class InfraMappingList(Widget):
+    """Represents the list of Infrastructure Mappings."""
+
+    ROOT = ParametrizedLocator('.//div[contains(@class,{@list_class|quote})]')
+    ITEM_LOCATOR = './div[contains(@class,"list-group-item")]'
+    # ITEM_TEXT_LOCATOR helps locate name of the Migration plan. ITEM_LOCATOR.text does not suffice.
+    # Also, ITEM_LOCATOR does not yield element which can be clicked to navigate to details page.
+    ITEM_TEXT_LOCATOR = './/div[contains(@class,"list-group-item-heading")]'
+    ITEM_STATUS_LOCATOR = './/div[contains(@class,"list-group-item-text")]'
+    ITEM_CLUSTERS_LOCATOR = './/div[./span[contains(@class,"pficon-cluster")]]'
+    ITEM_NETWORKS_LOCATOR = './/div[./span[contains(@class,"pficon-network")]]'
+    ITEM_DATASTORES_LOCATOR = './/div[./span[contains(@class,"fa-database")]]'
+    # Locates Associated Plans Button
+    ITEM_ASSOCIATED_PLANS_BUTTON_LOCATOR = './/div[./span[contains(@class,"pficon-catalog")]]'
+    # Locates Associated Plans list
+    ITEM_ASSOCIATED_PLANS_LIST_LOCATOR = ('.//div[contains(@class,'
+    '"infra-mapping-associated-plans-row")]')
+    SOURCE_ITEMS_LIST_LOCATOR = './/div[contains(@class,"infra-mapping-item-source")]/div'
+    TARGET_ITEMS_LIST_LOCATOR = './/div[contains(@class,"infra-mapping-item-target")]'
+    ITEM_EXPAND_LOCATOR = './span[contains(@class,"fa-angle-down")]'
+    ITEM_COLLAPSE_DETAILS_BUTTON_LOCATOR = './/span[contains(@class, "pficon-close")]'
+
+    def __init__(self, parent, list_class, logger=None):
+        Widget.__init__(self, parent, logger=logger)
+        self.list_class = list_class
+
+    @property
+    def all_items(self):
+        return [self.browser.text(item) for item in self.browser.elements(self.ITEM_TEXT_LOCATOR)]
+
+    def _get_map_element(self, map_name):
+        for item in self.browser.elements(self.ITEM_LOCATOR):
+            el = self.browser.element(".//*[@class='list-group-item-heading']", parent=item)
+            if self.browser.text(el) == map_name:
+                return item
+        raise ItemNotFound("Infra Mapping: {} not found".format(map_name))
+
+    def expand_map_clusters(self, map_name):
+        """Click on clusters to expand to show details."""
+        el = self._get_map_element(map_name)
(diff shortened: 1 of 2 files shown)


widgetastic_manageiq/__init__.py
```diff
@@ -4356,3 +4356,153 @@ def fill(self, value):
             return False
         self._all_item_elements[value or self.selected_item].click()
         return True
+
+class MigrationPlansList(Widget):
```
kedark3: Ignore this Widget, as this code is already being reviewed in the #7295 , which when merged, I will rebase this PR off-of master and make sure of correct commit history.

dajoRH:
Lint report for commit fa329fd25b4a02090283fa321ac72e07b89ebe68:

- :bangbang: Line 4360:1: **E302** *expected 2 blank lines, found 1*

**No commit flaws detected**.

*CFME QE Bot*

widgetastic_manageiq/__init__.py
```diff
@@ -4356,3 +4356,154 @@ def fill(self, value):
             return False
         self._all_item_elements[value or self.selected_item].click()
         return True
+
+
+class MigrationPlansList(Widget):
```
Please remove it from the PR, because you have it in #7295

mshriver:
WIPTEST as dependent PR isn't merged, and description still says its in progress.

kedark3:
:Received changes from Devel, updated PR to reflect everything. It should be complete now.

widgetastic_manageiq/__init__.py
```diff
@@ -4437,3 +4437,150 @@ def fill(self, value):
...
+
+    def get_map_status(self, map_name):
+        el = self._get_map_element(map_name)
+        return self.browser.text(self.browser.element(self.ITEM_STATUS_LOCATOR,
+             parent=el))
+
+    def _get_sources(self, map_name):
+        el = self._get_map_element(map_name)
+        for item in self.browser.elements(self.SOURCE_ITEMS_LIST_LOCATOR, parent=el):
+            return item.text
```
```python
self.browser.text(item)
```

widgetastic_manageiq/__init__.py
```diff
@@ -4437,3 +4437,150 @@ def fill(self, value):
...
+
+    def _get_targets(self, map_name):
+        el = self._get_map_element(map_name)
+        for item in self.browser.elements(self.TARGET_ITEMS_LIST_LOCATOR, parent=el):
+            return item.text
```
```python
return self.browser.text(item)
```

widgetastic_manageiq/__init__.py
```diff
@@ -4437,3 +4437,150 @@ def fill(self, value):
             raise TypeError("Fill can only take one value from list to click on.")
         self.select_plan(value)
         return True
+
+
+class InfraMappingList(Widget):
+    """Represents the list of Infrastructure Mappings."""
+
+    ROOT = ParametrizedLocator('.//div[contains(@class,{@list_class|quote})]')
+    ITEM_LOCATOR = ParametrizedLocator('./div[contains(@class,"list-group-item")]')
```
mshriver: Why is this `ParametrizedLocator`? Doesn't look like it needs to be.

widgetastic_manageiq/__init__.py
```diff
@@ -4437,3 +4437,150 @@ def fill(self, value):
...
+        for item in self.browser.elements(self.SOURCE_ITEMS_LIST_LOCATOR, parent=el):
+            return self.browser.text(item)
+
+    def _get_targets(self, map_name):
+        el = self._get_map_element(map_name)
+        for item in self.browser.elements(self.TARGET_ITEMS_LIST_LOCATOR, parent=el):
+            return self.browser.text(item)
+
+    def get_map_source_clusters(self, map_name):
+        if not self.is_clusters_expanded(map_name):
```
mshriver: I think you should bake all of these checks into the `expand_map_<thing>` function calls, so anything trying to expand the element doesn't have to check first.

Also, these need docblocks.

widgetastic_manageiq/__init__.py
```diff
@@ -4437,3 +4437,150 @@ def fill(self, value):
...
+        return self._get_targets(map_name)
+
+    def get_associated_plans_count(self, map_name):
+        try:
+            el = self._get_map_element(map_name)
+            return self.browser.text(self.browser.element(self.ITEM_ASSOCIATED_PLANS_BUTTON_LOCATOR,
+                 parent=el))
+        except NoSuchElementException:
+            # Plan with no associated plans
+            return None
```
mshriver: Since the function is for `get_<>_count`, I think this should return 0 when there's no button so its a predictable return type.

Also, what's the need/use for this method, when there is `get_associated_plans`, and any caller could just look at the length of that return.

widgetastic_manageiq/__init__.py
```diff
@@ -4437,3 +4437,150 @@ def fill(self, value):
...
+
+    def get_associated_plans(self, map_name):
+        if not self.is_associated_plans_expanded(map_name):
+            self.expand_map_associated_plans(map_name)
+        el = self._get_map_element(map_name)
+        return self.browser.text(self.browser.element(self.ITEM_ASSOCIATED_PLANS_LIST_LOCATOR,
```
mshriver: Feels a bit strange that this isn't returning a list of names, since its plural? If there's some reason why it doesn't or can't, please describe the behavior and what you expect this element to locate in the docblock.

widgetastic_manageiq/__init__.py
kedark3: good catch. ;)

widgetastic_manageiq/__init__.py
kedark3: This method will help me read the count which is displayed on the top bar in infra mappings list and then with get_associated_plans I can look at list length and validate that count is correctly displayed on the screen. Attaching a screenshot of how it looks on the page. 
image

widgetastic_manageiq/__init__.py
kedark3: We cannot split it to return a list as there is no way to differentiate, its all single block of text under a div. I have asked them if they can divide those in different divs/li tags, if they decide to do so, we can revisit this and update the widget. For now, it just returns a string.

widgetastic_manageiq/__init__.py
mshriver: Picture worth 1000 words :)

PR #7352, opened 2018-06-11
[1LP][RFR] Fix CMQE Persistent Volume Fixture

jawatts:
Remove wrapanapi call to create a persistent volume if one no longer exists, the create method was removed in https://github.com/ManageIQ/wrapanapi/pull/255/files. 

{{ pytest: cfme/tests/containers/test_relationships.py --use-provider cmqe }}


mshriver:
NEEDS TO INCLUDE wrapanapi 2.9.10 requirement before merge!

PR #7355, opened 2018-06-12
[1LP][RFR] workaround for some bug in ansible pod

izapolsk:
Ansible pod starts very slow first time. So, some tests fail. This is temporary workaround.


cfme/utils/appliance/__init__.py
```diff
@@ -1718,6 +1719,11 @@ def wait_for_embedded_ansible(self, timeout=900):
         Args:
             timeout: Number of seconds to wait until timeout (default ``900``)
         """
+        if self.is_pod and BZ(1576744, forced_streams=['5.9']).blocks:
```
lkhomenk: REQUIRED: could you please add `TODO` here? so it could be easier to find in the future

PR #7365, opened 2018-06-15
[WIP] Making small_vm fixture function scoped

jan-zmeskal:
TBD


cfme/tests/infrastructure/test_vm_reconfigure.py
```diff
@@ -43,7 +43,7 @@ def reconfigure_vm(vm, config):
                 .format(vars(config.hw), config.disks))
 
 
-@pytest.fixture(scope='module')
+@pytest.fixture(scope='function')
```
rlbabyuk: Required: remove (scope='function') as  @pytest.fixture as by default function scoped

quarckster: @jan-zmeskal It would be nice if you add a description in this fixture that explains why it should a function scoped.

jan-zmeskal: @quarckster Done

PR #7366, opened 2018-06-15
Fix miq table sort_order, update cmqe sort test

mshriver:


cfme/tests/containers/test_tables_sort.py
```diff
@@ -31,20 +31,24 @@
...
-    current_view.toolbar.view_selector.select('List View')
+    view.toolbar.view_selector.select('List View')
 
-    for col, header_text in enumerate(current_view.entities.elements.headers):
+    for col, header_text in enumerate(view.entities.elements.headers):
 
         if not header_text:
             continue
-        current_view.entities.paginator.set_items_per_page(1000)
+        view.entities.paginator.set_items_per_page(1000)
```
jawatts: Question:
Do we need to set the items per page each time to 1000 or can we just set it to 1000 on the initial navigation

jawatts:
Since you are updating this test, can we move this out the containers and into the the webui tests?

cfme/tests/containers/test_tables_sort.py
mshriver: good call, I'll move this up

mshriver:
@jawatts I'd rather leave moving it out to you, if you can.

mshriver:
There is a single failure, against container providers, in 5.8.4.4 in PRT.  The other cases are passing, and it seems to be mixing up columns at some point in the test. The result shows table sorted by hostname, but the `header_text` variable was the next column, ip_address. @jawatts is going to do some step-through to see why that resource type and version specifically is failing.

widgetastic_manageiq/__init__.py
```diff
@@ -446,7 +450,7 @@ def sort_order(self):
         # for filtering the None out
         try:
             return filter(
-                None, re.search(r'(sorting_|fa-sort-)(asc|desc)', klass).groups())[0]
+                None, re.search(r'(sorting_|fa-sort-)(asc|desc)', klass).groups())[1]
```
mshriver: Should have just left the PR at this one line change .... :)

mshriver:
So, the failures in PRT don't have anything to do with the code I'm changing here.

We've observed a behavior where the anchor (link text) for the column is being clicked, as shown in screenshots with the text highlighted after having been clicked. For some reason, either the page doesn't sort, or the failure and screenshot happen so fast that the sorting hadn't finished yet, and the UI wasn't actually ready.

I would like to proceed with review of this PR, and will be re-running PRT to try and identify a pattern to the failure and get some passing runs on the latest commit.

@jawatts @psav you may be interested in this, as it appears to be browser interaction causing a test failure.

psav:
Oh great! More browser interaction failure. Can we see the problem locally, ie.....if we get the screenshot and then pause at the end of the test, do we see that in actuality it did sort and we didn't wait long enough? I'm assuming there is no table cache that we are needing to clear?

PR #7374, opened 2018-06-18
[RFR] Force sprout use miq version instead of its own regexps

izapolsk:
it demands some changes in miq version. which are in progress


mshriver:
It was my intention to bring this in with the template_upload work I'm doing, and have these changes in my branch for that, but 5.10z stream has forced our hand. Sorry you're having to shoehorn this in @izapolsk.

mshriver:
We'll want to include the miq_version package bump in here too, though it doesn't look dependent on it.

PR #7384, opened 2018-06-20
[1LP][RFR] Ext-auth group switching test

mshriver:
Utilizing users with an evmgroup and a custom ext-auth group, this test configures the necessary authentication mode, adds a group, logs in, and tests that groups can be switched for ext-auth users.

{{ pytest: cfme/tests/integration/test_cfme_auth.py -k test_user_group_switching --long-running }}

### DIFF
--- a/cfme/tests/integration/test_cfme_auth.py
+++ b/cfme/tests/integration/test_cfme_auth.py
@@ -289,19 +289,28 @@ def test_user_group_switching(appliance, auth_user_data, auth_mode, user_type, a
                               soft_assert, request):
     """Test switching groups on a single user, between retreived group and built-in group"""
     UserGroups = namedtuple("UserGroups", "user, retrieved, groups")
-    user_group_tuples = [
-        UserGroups(
-            appliance.collections.users.simple_user(
-                user.username.replace(' ', '-') if user_type == 'upn' else user.username,
-                credentials[user.password]['password'],
-                fullname=user.fullname),
-            retrieve_group(appliance, auth_mode, user.username, group, auth_provider),
-            user.groups
-        )
-        for user in auth_user_data
-        for group in user.groups
-        # pick non-evm group when there are multiple groups for the user
-        if 'evmgroup' not in group.lower() and len(user.groups) > 1]
+    user_group_tuples = []
+    for user in auth_user_data:
+        if len(user.groups) > 1:  # skip users with only one group
+            for group in user.groups:
+                # pick non-evm group when there are multiple groups for the user
+                if 'evmgroup' not in group.lower():
+                    name_arg = (user.username.replace(' ', '-')
+                                if user_type == 'upn'
+                                else user.username)
+                    user_group_tuples.append(UserGroups(
+                        # make a simple_user instance, no CFME user create
+                        appliance.collections.users.simple_user(name_arg,
+                                                                credentials[user.password].password,
+                                                                fullname=user.fullname),
+                        # create group in CFME via retrieve_group which looks it up on auth_provider
+                        retrieve_group(appliance, auth_mode, user.username, group, auth_provider),
+                        # The users complete list of groups
+                        user.groups))
+                else:
+                    logger.info('Skipping user "%s" group "%s", its an evm group', user, group)
+        else:
+            logger.info('Skipping user for group switching test, only one group: %s', user)
 
     logger.info('USER_GROUP_TUPLES: %r', user_group_tuples)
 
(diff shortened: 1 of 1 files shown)


cfme/tests/integration/test_cfme_auth.py
```diff
@@ -275,3 +277,68 @@ def test_login_local_group(appliance, local_users, local_group, soft_assert):
...
+                user.username.replace(' ', '-') if user_type == 'upn' else user.username,
+                credentials[user.password]['password'],
+                fullname=user.fullname),
+            retrieve_group(appliance, auth_mode, user.username, group, auth_provider),
+            user.groups
+        )
+        for user in auth_user_data
+        for group in user.groups
+        # pick non-evm group when there are multiple groups for the user
+        if 'evmgroup' not in group.lower() and len(user.groups) > 1]
```
quarckster: **OPTIONAL**
It's not the best using of the list comprehension. I would make regular nested `for` loops.

cfme/tests/integration/test_cfme_auth.py
```diff
@@ -275,3 +277,68 @@ def test_login_local_group(appliance, local_users, local_group, soft_assert):
...
                         .format(local_group.description, display_groups))
+
+
+@pytest.mark.tier(1)
+@pytest.mark.ignore_stream('5.8')
+@pytest.mark.uncollectif(lambda auth_mode: auth_mode == 'amazon',
+                         'Amazon auth_data needed for group switch testing')
+def test_user_group_switching(appliance, auth_user_data, auth_mode, user_type, auth_provider,
+                              soft_assert, request):
+    """Test switching groups on a single user, between retreived group and built-in group"""
```
jawatts: Typo: retrieved

cfme/tests/integration/test_cfme_auth.py
jawatts: I would agree here, it is quite the list comprehension. if you do not want to change it, maybe add some comments so the next tester does not spend a lot of time on this.

cfme/tests/integration/test_cfme_auth.py
```diff
@@ -275,3 +277,68 @@ def test_login_local_group(appliance, local_users, local_group, soft_assert):
...
+    logger.info('USER_GROUP_TUPLES: %r', user_group_tuples)
+
+    for test_user in user_group_tuples:
+        with test_user.user:
+            view = navigate_to(appliance.server, 'LoggedIn', wait_for_view=True)
+            display_name = view.current_fullname
+            display_groups = [name for name in view.group_names]
+            active_group = view.current_groupname
+            # Check there are multiple groups displayed
+            soft_assert(len(display_groups) > 1, 'Only a single group is displayed for the user')
```
bsquizz: *OPTIONAL*
Should this be a hard assert, because it seems like your test can't really continue properly if there's only 1 group?

cfme/tests/integration/test_cfme_auth.py
mshriver: You guys make very good points, I originally avoided it because of the level of nesting required. Updating it now, but be warned it nests like 5 times :)

cfme/tests/integration/test_cfme_auth.py
```diff
@@ -275,3 +277,68 @@ def test_login_local_group(appliance, local_users, local_group, soft_assert):
...
+                        'user full name "{}" did not match UI display name "{}"'
+                        .format(test_user.user, display_name))
+            # Not checking current group, determined by group priority
+            # check retreived group is there
+            soft_assert(test_user.retrieved.description in display_groups,
+                        'user group "{}" not displayed in UI groups list "{}"'
+                        .format(test_user.retrieved.description, display_groups))
+
+            # change to the other group
+            view.change_group(display_other_group)
```
bsquizz: *OPTIONAL*
The other option is, if you keep the above as a soft_assert, maybe only run this portion of the test if display_other_group is not 'None' ? I am just trying to avoid the test failing with some hard-to-decipher error.

cfme/tests/integration/test_cfme_auth.py
jawatts: I think this should be a hard assert, if there is only one group, there is no use to continue

mshriver: So, responding to both of you, no, this should still be a soft assert.

Reason being, the test isn't parametrized on users, only on auth_provider and mode (basically). So I generate the list of users relevant to that provider/mode from the yaml, and then iterate over those users. I don't want any test failures for a single user (within reason, I'm not diapering here) to block execution of testing on the rest of the users.

Combined with @bsquizz comment below, the test will stop for that user if for some reason there's only one group displayed when we expect 2+

mshriver:
Test failure for external auth mode because the appliance didn't have an FQDN, not in scope here.

PR #7386, opened 2018-06-20
[1LP][RFR] Blocking test_child_tenant_quota_enforce_via_lifecycle_infra with GH7385

jan-zmeskal:
{{pytest: cfme/tests/infrastructure/test_child_tenant.py::test_child_tenant_quota_enforce_via_lifecycle_infra --long-running --use-provider rhv41}}

PRT was actually successful: http://10.16.4.32/trackerbot/pr/run/25164#0JoWcFYF

### DIFF
--- a/cfme/tests/infrastructure/test_child_tenant.py
+++ b/cfme/tests/infrastructure/test_child_tenant.py
@@ -8,6 +8,7 @@ from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.markers.env_markers.provider import ONE_PER_TYPE
 from cfme.provisioning import do_vm_provisioning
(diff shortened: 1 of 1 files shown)


anurag03:
@jan-zmeskal As the test case is not causing issue for 5.8 appliance we can have a skip condition which can exclude skip for 5.8 appliance and just skip the test case for 5.9 appliance.

jan-zmeskal:
@anurag03 That's a very good remark. The thing is that when I try define unblocking based on appliance version like this:
`unblock=lambda provider, appliance: not provider.one_of(RHEVMProvider) or appliance.version < '5.9'`
I get this error:
image
I searched for `appliance.version` usage in cfme/tests and it seems it's only used with `uncollectif` method and the `@pytest.mark.meta` decorator does not have access to it.

jan-zmeskal:
@anurag03 I figured it out.

PR #7327, opened 2018-06-05
[1LP][RFR] Initial setup for v2v

Yadnyawalkya:
Currently executing via,
```pytest test_v2v.py --use-provider complete --provider-limit 2 --use-template-cache```

Tasks done:
	1. Added mutli-provider usage
	2. Removed pytest_generate_tests from tests
        3. Moved common fixtures to v2v.py

Steps:
        1.Add rhvm and nvc providers and return provider objects
        2.Validate host credentials
        3.Assign tags to conversation host
        ~~4.Import automate domain via github repo~~

{{pytest: cfme/tests/v2v/test_v2v.py --use-template-cache --use-provider rhv42 --use-provider vsphere6-nested --provider-limit 2 -vvvv}}

PR #7388, opened 2018-06-20
[1LP][RFR] V2V migration tests fixtures

kedark3:
Purpose or Intent
=================

__Adding fixtures__ for migration tests. These fixtures will enable us to perform migration tests. Currently only fixtures are added. Initial  provider setup is taken from https://github.com/ManageIQ/integration_tests/pull/7327 and actual tests will be added in the separate PR, tests in this PR are just for demonstration purpose.

~~**NOTE** results of execution of these test will depend on https://gitlab.cee.redhat.com/cfme-qe/cfme-qe-yamls/merge_requests/545~~

{{pytest: cfme/tests/v2v/test_v2v_migrations.py  --use-provider rhv42 --use-provider vsphere6-nested --provider-limit 2 -vvvv}}

### DIFF
--- a/cfme/tests/v2v/test_v2v_migrations.py
+++ b/cfme/tests/v2v/test_v2v_migrations.py
@@ -1,0 +1,227 @@
+"""Test to validate basic navigations.Later to be replaced with End-to-End functional testing."""
+import fauxfactory
+import pytest
+
+from widgetastic.utils import partial_match
+
+from cfme.fixtures.provider import setup_or_skip
+from cfme.infrastructure.provider.rhevm import RHEVMProvider
+from cfme.infrastructure.provider.virtualcenter import VMwareProvider
+from cfme.utils import testgen
+from cfme.utils.appliance.implementations.ui import navigate_to
+
+pytestmark = [pytest.mark.ignore_stream('5.8')]
+
+
+# TODO: Following function is due for refactor as soon as PR#7408 is merged.
+def pytest_generate_tests(metafunc):
+    """This is parametrizing over the provider types and creating permutations of provider pairs,
+       adding ids and argvalues."""
+    argnames1, argvalues1, idlist1 = testgen.providers_by_class(metafunc, [VMwareProvider])
+    argnames2, argvalues2, idlist2 = testgen.providers_by_class(metafunc, [RHEVMProvider])
+
+    new_idlist = []
+    new_argvalues = []
+    new_argnames = ['nvc_prov', 'rhvm_prov']
+
+    for index1, argvalue_tuple1 in enumerate(argvalues1):
+        for index2, argvalue_tuple2 in enumerate(argvalues2):
+            new_idlist.append('{}-{}'.format(idlist1[index1], idlist2[index2]))
+            new_argvalues.append((argvalue_tuple1[0], argvalue_tuple2[0]))
+    testgen.parametrize(metafunc, new_argnames, new_argvalues, ids=new_idlist, scope="module")
+
+
+@pytest.fixture(scope='module')
+def create_provider(request, nvc_prov, rhvm_prov):
+    """ Fixture to setup providers """
+    setup_or_skip(request, nvc_prov)
+    setup_or_skip(request, rhvm_prov)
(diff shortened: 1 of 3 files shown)


dajoRH:
Lint report for commit db232f2470dbd92a9fd6331ff13ed13133cad041:

`cfme/tests/v2v/test_v2v_migrations.py`:
- :red_circle: Line 9:1: **F401** *'cfme.utils.appliance.implementations.ui.navigate_to' imported but unused*
- :red_circle: Line 11:1: **F401** *'cfme.utils.log.logger' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

kedark3:
@mshriver kindly review if possible :smile_cat:

cfme/tests/v2v/test_v2v_migrations.py
```diff
@@ -0,0 +1,172 @@
...
+    """ Fixture to setup providers """
+    if not nvc_prov.exists:
+        nvc_prov.create_rest()
+    if not rhvm_prov.exists:
+        rhvm_prov.create_rest()
+    # yield
+    # if nvc_prov.exists:
+    #     nvc_prov.delete_rest()
+    # if rhvm_prov.exists:
+    #     rhvm_prov.delete_rest()
```
quarckster: **REQUIRED**
Please remove commented out lines.

cfme/tests/v2v/test_v2v_migrations.py
```diff
@@ -0,0 +1,172 @@
...
+                                            form_data_single_datastore):
+    infrastructure_mapping_collection = appliance.collections.infrastructure_mapping
+    mapping = infrastructure_mapping_collection.create(form_data_single_datastore)
+    view = navigate_to(infrastructure_mapping_collection, 'All')
+    assert mapping.name in view.infra_mapping_list.read()
+
+
+@pytest.mark.parametrize('form_data_single_network', [['VM Network', 'ovirtmgmt'],
+                            ['DPortGroup', 'ovirtmgmt']], indirect=True)
+def test_single_network_single_vm_mapping_crud(appliance, create_provider,
```
quarckster: **REQUIRED**
CRUD is create, read, update, delete. In this test you do only create.

mshriver: This was actually at my request, given what is currently supported in the framework. I wanted to see some valid tests of the fixtures/generation via the available create method. These CF entities don't block from provider removal, or make for any additional load on the provider. I believe its okay to not clean them up at this time, and the tests will be updated as the framework support is added for Update/Delete.

mshriver: Though I was expecting a TODO + comments here explaining that intention and need for test expansion... ;)

kedark3: @mshriver thanks for jumping in on this, I will add TODO. @quarckster thanks for review, updating my PR.

mshriver:
PRT check your `{{ pytest: }}` line, it collected rhv41 because if you check setup.txt in the PRT artifacts, its calling py.test with `--use-provider default`

kedark3:
This PR needs to be rebased and merged AFTER #7416

mshriver:
We are seeing some inconsistent behavior in tests related to the modal being displayed, and triggering a browser close.

The check that's causing the browser close is within CFMENavigateStep.pre_badness_check, but at the moment I'm not able to determine why this is triggered _after_ navigation has already completed, and has returned a view (that was waited for).  The logs show completion of the navigate step. The next call is a view.form.fill, which is ultimately using widgetastic.core.widget.View.fill. The Widgetastic browser plugin, and the MIQ extension of it, have `ENSURE_PAGE_SAFE` methods that don't check for the generic modals, and should not be catching the modal.

There is some risk in accepting this, as there are some modifications to the view and navigation, but I do not believe they're directly related to the inconsistent behavior that we're seeing.  Sometimes local tests pass completely with no issue.

We can continue to debug and address these test failures in a followup PR, so that we can start getting some automation against V2V.

PR #7395, opened 2018-06-22
[1LP][RFR] Change required to support dual browser setup in container

psav:
* Forces chrome to run in --no-sandbox mode with Remote webdriver
* Forces Marionette to False for Firefox, which is required for older
  FF versions

This is really just to allow us to support the fact that 5.10 MUST at
the moment run under Chrome, but that we want to allow original tests
of 5.9 and 5.8 to run under Firefox still.

NOTE: This is temporary and fully expected to be removed once FF 60.1
      is out.

### DIFF
--- a/cfme/utils/browser.py
+++ b/cfme/utils/browser.py
@@ -7,7 +7,6 @@ from collections import namedtuple
 from shutil import rmtree
 from string import Template
 from tempfile import mkdtemp
-from six.moves.urllib_error import URLError
 
 import os
 import requests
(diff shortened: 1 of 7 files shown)


cfme/utils/browser.py
```diff
@@ -246,6 +246,19 @@ def from_conf(cls, browser_conf):
             atexit.register(wharf.checkin)
             return cls(WharfFactory(webdriver_class, browser_kwargs, wharf))
         else:
+            if webdriver_name == "Remote":
+                if browser_conf[
+                        'webdriver_options'][
+                            'desired_capabilities']['browserName'].lower() == 'chrome':
+                    browser_kwargs['desired_capabilities']['chromeOptions'] = {}
```
quarckster: I have nice chrome option here in my env.local.yaml:
```yaml
chromeOptions:
    args:
        - --disable-infobar
```

PR #7398, opened 2018-06-22
[RFR] Only delete marionette on chrome if its there

mshriver:
keyerror trying to delete it when it doesn't exist


jawatts:
:+1:

PR #7347, opened 2018-06-08
[WIPTEST] Move the cfme test framework to wrapanapi 3.0

bsquizz:
Get everything in sync with the class-based wrapanapi changes introduced in https://github.com/ManageIQ/wrapanapi/pull/240

* Try to keep most of the "on_provider" stuff in wrapanapi
* Modify anywhere `wrapanapi` or `provider.mgmt` type calls are used in framework & tests

{{ pytest: --use-provider complete --long-running cfme/tests/cloud_infra_common/*.py }}

PR #7400, opened 2018-06-25
[1LP][RFR] Update Container test project dashboard according to wrapanapi changes

brahmanim:
Following changes in wrapanapi that affect our container tests, 
I needed to update test_project_dashboard function with new wrapanapi methods
And attributes name.

This test should run on appliance version 5.9 and above.

{{pytest: cfme/tests/containers/test_projects_dashboard.py -v --use-provider ocp-37-prom }}

### DIFF
--- a/cfme/tests/containers/test_projects_dashboard.py
+++ b/cfme/tests/containers/test_projects_dashboard.py
@@ -27,27 +27,37 @@ def get_api_object_counts(appliance, project_name, provider):
     """ Fetches amount of Containers/Services/Images from the API per selected project name"""
     assert isinstance(provider, ContainersProvider)
     return {
-        Container: len(provider.mgmt.list_container(project_name=project_name)),
-        Service: len(provider.mgmt.list_service(project_name=project_name)),
-        Image: len(get_container_images_amt(provider, project_name))
+        Container: len(flatten_list(provider.mgmt.list_container(namespace=project_name),
+                                    flattened_list=[])),
+        Service: len(provider.mgmt.list_service(namespace=project_name)),
+        # TODO: remove sorted set when wrapanapi updates will return unique image IDs
+        # number of unique images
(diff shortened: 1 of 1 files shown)


mshriver:
This depends on a wrapanapi change that's merged an in 3.0.5, but this doesn't include a requirement update.

The jump to wrapanapi 3.0 is non-trivial.

@bsquizz is working on wrapanapi 3.0 integration in PR #7347 , and this should be reviewed after that has been merged and we're up-to-date on wrapanapi.

brahmanim:
@mshriver I manage to do all the changes on this test, with no dependency on new wrapanapi 3.X.

In addition, there is a small change in wrapanapi [list_image_id](https://github.com/ManageIQ/wrapanapi/blob/eeed8cbd91c113944227c2fe1070c3dfec89d326/wrapanapi/systems/container/rhopenshift.py#L257), that I implement on current PR on [get_api_object_counts](https://github.com/brahmanim/integration_tests/blob/36df5171958e699ca7aec274d9bbbdfd0c72b559/cfme/tests/containers/test_projects_dashboard.py#L35), with TODO remark.
So there is no dependency on new wrapanapi.

Can i move back the PR to RFR for review?

bsquizz:
Took a look at this to confirm ... there shouldn't be any impact from wrapanapi 3.x here. The openshift provider hasn't changed in wrapanapi 3.0 so container tests shouldn't see much change.

mshriver:
@brahmanim If that's the case you should update your PR description, it still states that it is dependent on wrapanapi #280 (which is what I was working from when I marked it WIP)

brahmanim:
@mshriver @bsquizz You are right, There was a dependency on the beginning of this PR,
Then I manage to avoid it by implementation on the test.
I remove this dependency from PR description and move back to RFR.

PR #7404, opened 2018-06-25
[1LP][RFR] Add virt_type Support for Downstream OpenShiftProvider

jawatts:
Updated view_value_mapping method in OpenshiftProvider class to enable virt_type for downstream builds, "Virtualization" drop down was added in downstream 510.

Updated .get to default to "Disabled" by default if the value is not found. 

Fixed variable naming. 

{{ pytest: cfme/tests/containers/test_node.py cfme/tests/containers/test_static_custom_attributes.py  --use-provider cmqe }}

### DIFF
--- a/cfme/common/provider_views.py
+++ b/cfme/common/provider_views.py
@@ -531,7 +531,7 @@ class ContainerProviderAddViewUpdated(ContainerProviderAddView, ContainerProvide
         for widget_name in self.COND_WIDGETS:
             widget = getattr(self, widget_name)
             if widget.is_displayed:
-                widget.fill(values.get(widget_name))
+                widget.fill(values.get(widget_name) or 'Disabled')
 
 
 class PhysicalProviderAddView(ProviderAddView):
--- a/cfme/containers/provider/openshift.py
+++ b/cfme/containers/provider/openshift.py
@@ -145,14 +145,14 @@ class OpenshiftProvider(ContainersProvider):
                 'image_tag': self.image_tag,
                 'cve_loc': self.cve_loc
             }
+            if self.appliance.version >= '5.10':
+                mapping['virt_type'] = self.virt_type
         else:
             mapping['metrics_type'] = None
             mapping['alerts_type'] = None
+            mapping['virt_type'] = None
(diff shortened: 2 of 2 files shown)


jawatts:
@quarckster PRT failures due to test_static_custom_attributes test. Fixes for the test will be handled in a different PR.

cfme/common/provider_views.py
```diff
@@ -542,7 +542,7 @@ def before_fill(self, values):
         for widget_name in self.COND_WIDGETS:
             widget = getattr(self, widget_name)
             if widget.is_displayed:
-                widget.fill(values.get(widget_name))
+                widget.fill(values.get(widget_name, 'Disabled'))
```
psav: **REQUIRED:** This will fill ```['prov_type', 'metrics_type', 'alerts_type', 'virt_type']``` with 'Disabled' now instead of, I'm assuming 'None'. Is this the right thing to do for ALL the widgets?

cfme/containers/provider/openshift.py
```diff
@@ -145,15 +145,15 @@ def view_value_mapping(self):
                 'image_tag': self.image_tag,
                 'cve_loc': self.cve_loc
             }
+        elif self.appliance.version >= '5.10':
```
psav: **REQUIRED:** Forgive me, but this won't fire will it? If the appliance is '5.10' then it will hit the first conditional, as it's also over '5.9' and won't actually do what you want. Also is it the case that if it's over '5.10' you don't want any of the other fields populated? So if it worked as I believe you intended, you woulnd't ever see ```mapping['alerts_type']``` being set.

cfme/common/provider_views.py
```diff
@@ -531,7 +531,7 @@ def before_fill(self, values):
         for widget_name in self.COND_WIDGETS:
             widget = getattr(self, widget_name)
             if widget.is_displayed:
-                widget.fill(values.get(widget_name))
+                widget.fill(values.get(widget_name) or 'Disabled')
```
jawatts: Set to "Disabled" if value of widget is None

PR #7409, opened 2018-06-26
[1LP][RFR] Update server name setting, auth test BZ, auth test loop fix

mshriver:
Simplify the workaround for server name setting, don't call `server.name` again in an `except` block that was entered because of a call to `server.name` in the first place.

Add BZ blocker for 510z downstream to auth tests

Update auth test list comprehension to handle a user with no groups.


mshriver:
Thanks for the review @rlbabyuk, I'm actually going to close this though, as I've decided to change parametrization in the auth test, and @psav requested the server name change _not_ go in as we're leaning on development for fixes.

PR #7410, opened 2018-06-27
[WIP] Adding navigation step for InfraVm to Provision.

rhrazdil:
Currently our Provision test cases fail because InfraVm doesn't have Provision navigation target.
There may be a better way to fix this, because I'm not really sure why this stopped working. So I'm open to suggestions. 

{{pytest: cfme/tests/infrastructure/test_pxe_provisioning.py::test_pxe_provision_from_template --use-provider complete --long-running -vv}}

### DIFF
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -1412,6 +1412,7 @@ class TemplatesAll(CFMENavigateStep):
 
 
 @navigator.register(InfraVmCollection, 'Provision')
+@navigator.register(InfraVm, 'Provision')
 class ProvisionVM(CFMENavigateStep):
     VIEW = ProvisionView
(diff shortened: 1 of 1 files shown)


quarckster:
I believe the fix should be in [cfme/provisioning.py](https://github.com/ManageIQ/integration_tests/blob/master/cfme/provisioning.py#L22).
```diff
- view = navigate_to(vm, 'Provision')
+ view = navigate_to(vm.parent, 'Provision')
```

rhrazdil:
Hi @quarckster , the approach you suggest is discussed  by @jan-zmeskal in #7378, but the results show that this causes some other test cases to break down.
I'll investigate further after PTO.

rhrazdil:
Closing as I see that the fix is indeed included in PR #7347.

PR #7415, opened 2018-06-27
[1LP][RFR]Update cloud/infra tagging tests

rlbabyuk:
Purpose or Intent
=================
Remove duplication for cloud anf infra providers, after selectors update

### DIFF
--- a/cfme/tests/cloud_infra_common/test_tag_objects.py
+++ b/cfme/tests/cloud_infra_common/test_tag_objects.py
@@ -3,14 +3,15 @@
 import pytest
 
 from cfme.cloud.keypairs import KeyPair
-from cfme.cloud.provider import CloudProvider
+from cfme.cloud.provider.ec2 import EC2Provider
 from cfme.infrastructure.provider import InfraProvider
-from cfme.fixtures.provider import setup_one_or_skip
-from cfme.utils.providers import ProviderFilter
+from cfme.markers.env_markers.provider import ONE_PER_TYPE
 from cfme.utils.appliance.implementations.ui import navigate_to
 
 pytestmark = [
-    pytest.mark.tier(2)
+    pytest.mark.tier(2),
+    pytest.mark.usefixtures('setup_provider'),
+    pytest.mark.provider([InfraProvider, EC2Provider], scope='module', selector=ONE_PER_TYPE)
 ]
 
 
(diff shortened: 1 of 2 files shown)


cfme/tests/cloud_infra_common/test_tag_objects.py
```diff
@@ -5,58 +5,48 @@
...
-from cfme.fixtures.provider import setup_one_or_skip
-from cfme.utils.providers import ProviderFilter
+from cfme.markers.env_markers.provider import ONE_PER_CATEGORY
 from cfme.utils.appliance.implementations.ui import navigate_to
 
 pytestmark = [
-    pytest.mark.tier(2)
+    pytest.mark.tier(2),
+    pytest.mark.usefixtures('setup_provider'),
+    pytest.mark.provider([InfraProvider, CloudProvider], scope='module', selector=ONE_PER_CATEGORY)
```
mshriver: provider marker should be function scoped

cfme/tests/cloud_infra_common/test_tag_objects.py
```diff
@@ -5,58 +5,48 @@
...
-    return setup_one_or_skip(request, filters=[prov_filter])
-
-
 def get_collection_entity(appliance, collection_name, destination, provider):
     if collection_name in ['infra_provider', 'cloud_provider']:
         return provider
     else:
         collection = getattr(appliance.collections, collection_name)
-        view = navigate_to(provider, destination) if destination else navigate_to(collection, 'All')
+        collection.filters = {'provider': provider}
```
mshriver: I think since this test collection and parametrization has changed enough that its worth mixing it up a bit.

I think it would be a lot more clear to define the destination string within the test_items tuples instead of inferring them here.

rlbabyuk: done

cfme/tests/cloud_infra_common/test_tag_objects.py
```diff
@@ -117,25 +106,26 @@ def _tagging_check(test_item, tag_place):
     return _tagging_check
 
 
+@pytest.mark.uncollectif(lambda provider: provider.one_of(InfraProvider))
```
psav: **REQUIRED:** I believe you can now do overrides. 
```python
@pytest.mark.provider([CloudProvider], override=True)
```

PR #7416, opened 2018-06-27
[1LP][RFR]Adding appliance method to enable-disable migration UI, and test it

kedark3:
Purpose or Intent
=================

__Extending IPAppliance__ to add new methods that can enable/disable migration UI.

and
__Adding test__ to enable and disable migration UI and validate if it shows/hides 'Migration' option on navigation bar.

{pytest: cfme/tests/v2v/test_v2v.py -k 'test_migration_ui_enable_disable' -vvvv}

### DIFF
--- a/cfme/tests/v2v/test_v2v.py
+++ b/cfme/tests/v2v/test_v2v.py
@@ -3,18 +3,23 @@ import pytest
 
 from cfme.utils.appliance.implementations.ui import navigate_to
 
-pytestmark = [pytest.mark.ignore_stream('5.8', '5.9')]
+pytestmark = [pytest.mark.ignore_stream('5.8')]
 
 
-@pytest.mark.tier(0)
-def test_compute_migration_modal(appliance):
-    """Test takes an appliance and tries to navigate to migration page and open mapping wizard.
+@pytest.fixture(scope="module")
+def disable_migration_ui(appliance):
+    appliance.disable_migration_ui()
 
-    This is a dummy test just to check navigation.
-    # TODO : Replace this test with actual test.
-    """
-    infra_mapping_collection = appliance.collections.v2v_mappings
-    view = navigate_to(infra_mapping_collection, 'All')
-    assert view.is_displayed
-    view = navigate_to(infra_mapping_collection, 'Add')
-    assert view.is_displayed
+
+def test_migration_ui_enable_disable(appliance, disable_migration_ui):
+    appliance.enable_migration_ui()
+    view = navigate_to(appliance.server, 'Dashboard', wait_for_view=True)
+    nav_tree = view.navigation.nav_item_tree()
+    assert 'Migration' in nav_tree['Compute'], ('Trying to find Migration in'
+           ' nav_tree, should be present :{}'.format(nav_tree))
+    appliance.disable_migration_ui()
+    view.browser.refresh()
+    view = navigate_to(appliance.server, 'Dashboard', wait_for_view=True)
+    nav_tree = view.navigation.nav_item_tree()
+    assert 'Migration' not in nav_tree['Compute'], ('Trying to find Migration in'
+           ' nav_tree, should be absent :{}'.format(nav_tree))
--- a/cfme/utils/appliance/__init__.py
+++ b/cfme/utils/appliance/__init__.py
@@ -2443,6 +2443,19 @@ class IPAppliance(object):
             raise ValueError('No support for coll_type: "{}" collection name lookup'
                              .format(coll_type))
 
+    def _switch_migration_ui(self, enable):
+        self.update_advanced_settings({'product': {'transformation': enable}})
+        self.appliance.server.logout()
+        self.restart_evm_service()
+        self.wait_for_evm_service()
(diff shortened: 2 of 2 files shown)


cfme/utils/appliance/__init__.py
```diff
@@ -2443,6 +2443,18 @@ def provider_based_collection(self, provider, coll_type='vms'):
...
+        self.update_advanced_settings({'product': {'transformation': True}})
+        self.restart_evm_service()
+        self.wait_for_evm_service()
+        self.wait_for_web_ui()
+
+    def disable_migration_ui(self):
+        self.update_advanced_settings({'product': {'transformation': False}})
+        self.restart_evm_service()
+        self.wait_for_evm_service()
+        self.wait_for_web_ui()
```
quarckster: **REQUIRED**
I guess this is only for 5.9.
```python
def _switch_migration_ui(self,  enable):
    self.update_advanced_settings({'product': {'transformation': enable}})
    self.restart_evm_service()
    self.wait_for_evm_service()
    self.wait_for_web_ui()

def enable_migration_ui(self)
    self._switch_migration_ui(self,  True)

def disable_migration_ui(self)
   self. _switch_migration_ui(self,  False)
```

kedark3: made the changes.

kedark3:
**Update: trackerbot/pr/run/24914**: it has passed for 59z, 510z will not pass because there is some issue with automation not being able to login to cfme appliance. I think its because of old browser. I reques to merge this pr based on 59z results.

cfme/utils/appliance/__init__.py
```diff
@@ -2443,6 +2443,18 @@ def provider_based_collection(self, provider, coll_type='vms'):
             raise ValueError('No support for coll_type: "{}" collection name lookup'
                              .format(coll_type))
 
+    def _switch_migration_ui(self, enable):
+        self.update_advanced_settings({'product': {'transformation': enable}})
```
psav: **REQUIRED:** I think the better action here is to explicitly log out, before restarting the evm service.

cfme/tests/v2v/test_v2v.py
```diff
@@ -2,19 +2,20 @@
...
-    """
-    infra_mapping_collection = appliance.collections.v2v_mappings
-    view = navigate_to(infra_mapping_collection, 'All')
-    assert view.is_displayed
-    view = navigate_to(infra_mapping_collection, 'Add')
-    assert view.is_displayed
+def test_migration_ui_enable_disable(appliance):
+    appliance.enable_migration_ui()
+    view = navigate_to(appliance.server, 'LoggedIn', wait_for_view=True)
+    nav_tree = view.navigation.nav_item_tree()
```
psav: **REQUIRED:** Here, we should not be navigating to 'LoggedIn' but 'Dashboard'. It is a proper _further_ destination. Same as below.

kedark3:
@psav @mshriver updated this PR with requested changes. Kindly review as convenient.

cfme/tests/v2v/test_v2v.py
```diff
@@ -2,19 +2,20 @@
...
-    infra_mapping_collection = appliance.collections.v2v_mappings
-    view = navigate_to(infra_mapping_collection, 'All')
-    assert view.is_displayed
-    view = navigate_to(infra_mapping_collection, 'Add')
-    assert view.is_displayed
+def test_migration_ui_enable_disable(appliance):
+    appliance.enable_migration_ui()
+    view = navigate_to(appliance.server, 'Dashboard', wait_for_view=True)
+    nav_tree = view.navigation.nav_item_tree()
+    logger.debug('Trying to find Migration in nav_tree, should be present : %s', nav_tree)
```
mshriver: Consider instead passing this string as the failure message for the assert below it, so when it fails the AssertionError contains relevant text.

cfme/tests/v2v/test_v2v.py
```diff
@@ -2,19 +2,20 @@
...
-    This is a dummy test just to check navigation.
-    # TODO : Replace this test with actual test.
-    """
-    infra_mapping_collection = appliance.collections.v2v_mappings
-    view = navigate_to(infra_mapping_collection, 'All')
-    assert view.is_displayed
-    view = navigate_to(infra_mapping_collection, 'Add')
-    assert view.is_displayed
+def test_migration_ui_enable_disable(appliance):
+    appliance.enable_migration_ui()
```
mshriver: So, this test is a little bit ambiguous in that we don't really know the state of the migration feature at the start, and this `enable_migration_ui` call may not actually be doing anything.

Test should either enable..disable..enable, or should have a fixture that disables during setup.

kedark3: @mshriver initially appliance has the state `Disabled`. So enable-disable works.

mshriver: It does now, but will it always? Will all versions keep this configuration?

kedark3: Good question, not sure of this, I will add fixture to disable it before test.

PR #7418, opened 2018-07-02
[1LP][RFR] Automate test for help menu

valaparthvi:
Add a new test case to check if editing items in the help menu actually updates the help menu.
{{ pytest: cfme/tests/configure/test_help_menu.py }}

Note: This test runs and passes successfully on my local environment, but fails here, since my gpg key hasn't been merged into cfme-qe-yamls yet.

### DIFF
--- a/cfme/tests/configure/test_help_menu.py
+++ b/cfme/tests/configure/test_help_menu.py
@@ -1,8 +1,12 @@
 import fauxfactory
+
 from cfme.utils.appliance.implementations.ui import navigate_to
 
 
 def test_update_help_menu(appliance):
+    """Test case to check if editing items in help menu actually
+    updates the help menu.
+    """
     region = appliance.collections.regions.instantiate()
     updates = {'documentation_title': fauxfactory.gen_alpha()}
     view = region.set_help_menu_configuration(updates)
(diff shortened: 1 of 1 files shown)


cfme/tests/configure/test_help_menu.py
```diff
@@ -0,0 +1,11 @@
+import fauxfactory
+from cfme.utils.appliance.implementations.ui import navigate_to
```
quarckster: **REQUIRED**
Please follow our style guides http://cfme-tests.readthedocs.io/en/master/guides/dev_guide.html#code-style.

> In addition to being broken up into the three sections of standard library, third-party, and the local application, imports should be sorted alphabetically. ‘import’ lines within those sections still come before ‘from ... import’ lines:

Just put a blank line between third-party and the local imports.

cfme/tests/configure/test_help_menu.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def test_update_help_menu(appliance):
```
quarckster: **REQUIRED**
Please add a description for this test. http://cfme-tests.readthedocs.io/en/master/guides/documenting.html#documenting-tests

valaparthvi:
My bad! I didn't take a look at the guideline. But, I've updated the changes now.

PR #7424, opened 2018-07-04
[1LP][RFR] Fixed provider and hosts relationships

quarckster:
Purpose
=================

`provider.hosts` should return a hosts collection filtered out by the provider.

{{pytest: -v -k "test_schedule_timer or test_vm_migrate or test_host_drift_analysis" --long-running}}

### DIFF
--- a/cfme/infrastructure/host.py
+++ b/cfme/infrastructure/host.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 """A model of an Infrastructure Host in CFME."""
 import attr
+import six
+
 from manageiq_client.api import APIException
 from navmazing import NavigateToSibling, NavigateToAttribute
 from selenium.common.exceptions import NoSuchElementException
@@ -248,8 +250,14 @@ class Host(BaseEntity, Updateable, Pretty, PolicyProfileAssignable, Taggable):
         # TODO: Move to Sentaku
         try:
             host = self.appliance.rest_api.collections.hosts.get(name=self.name)
-            host.action.edit(credentials={"userid": credentials['default'].principal,
-                                          "password": credentials['default'].secret})
+            if isinstance(credentials['default'], six.string_types):
+                creds = self.Credential.from_config(credentials["default"])
+            elif isinstance(credentials['default'], self.Credential):
+                creds = credentials['default']
+            else:
+                raise TypeError("credentials must be a dict or a Credential instance")
+            host.action.edit(credentials={"userid": creds.principal,
+                                          "password": creds.secret})
         except APIException:
             return False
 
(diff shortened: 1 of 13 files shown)


dajoRH:
Lint report for commit 9747217cc4233d71511f7caa1230194a2f213753:

`cfme/tests/cloud_infra_common/test_vm_instance_analysis.py`:
- :bangbang: Line 151:52: **E999** *SyntaxError: invalid syntax*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

quarckster:
Failed tests are not related to the PR changes.

cfme/infrastructure/host.py
```diff
@@ -485,11 +493,14 @@ def create(self, name, provider=None, credentials=None, hostname=None, ip_addres
         view.flash.assert_success_message(flash_message)
         return host
 
-    def all(self, provider):
+    def all(self):
```
psav: **REQUIRED:** So we are dropping the provider requirement and simply using the parent as the ```provider```, but the problem is that if we access the hosts collection from the appliance, then the parent will be an appliance object which will not work. I also think that given the simplicity of the function, we could perhaps also look to obtaining it via REST instead of the UI, which would be much quicker. We can also then think about dropping hosts from the appliance list of collections, and simple instantiating it as a provider based collection as you have done here. Obviously it is applicable to obtain a host without a provider, the problem is _as I remember_ that our Host object doesn't allow you to instantiate it without giving a provider.

quarckster: @psav 
* Testing this PR I was able successfully to get:
```python
hosts = appliance.collections.hosts.all()
```
as well as
```python
hosts = provider.hosts.all()
```
because in the first case `self.filters.get('parent') == None`
* `provider` is not required argument, so a host can be successfully instantiated without it
* if we drop the hosts collections the only way to get all aplliance's hosts will be iterating through all providers:
```python
hosts = [host for host in provider.hosts.all() for provider in appliance.collections.infra_providers.all()]
```
* Agree with REST, I will rework that.

psav: Ahhhhhh, for some reason i thought host required provider!!
Yeh lets get REST in, should be simple.

quarckster: @psav it seems there is no way to get host of a provider via REST:
```python
In [58]: appliance.rest_api.collections.providers[0].subcollections
Out[58]:
[u'cloud_networks',
 u'cloud_subnets',
 u'cloud_tenants',
 u'custom_attributes',
 u'flavors',
 u'load_balancers',
 u'policies',
 u'policy_profiles',
 u'security_groups',
 u'tags',
 u'vms']
```

PR #7426, opened 2018-07-04
[1LP][RFR] nasty old workaround adding zero inside build number is removed

izapolsk:
build number 5.9.3.4 has become 5.9.3.04 because of that workaround. I doubt this is correct behavior.
So, this PR intends to remove that workaround

### DIFF
--- a/scripts/template_upload_all.py
+++ b/scripts/template_upload_all.py
@@ -81,8 +81,6 @@ def template_name(image_link, image_ts, checksum_link, version=None):
             return "cfme-nightly-{}-{}".format(version, result[0][:-2])
         elif version:
             # proper build
-            if len(version) == 4:
-                version = version[:-1] + '0' + version[-1:]
             return "cfme-{}-{}{}{}".format(version, image_ts, image_dt.hour, image_dt.minute)
         else:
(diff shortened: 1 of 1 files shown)


scripts/template_upload_all.py
```diff
@@ -81,8 +81,6 @@ def template_name(image_link, image_ts, checksum_link, version=None):
             return "cfme-nightly-{}-{}".format(version, result[0][:-2])
         elif version:
             # proper build
-            if len(version) == 4:
-                version = version[:-1] + '0' + version[-1:]
```
mshriver: This was intentional, and AFAIK its driven by polarion templates for our builds being in the format: `x.y.z.a`, where `a` is always zero padded.

This is mirrored in other places in the scripts having to do with miq_version and polarion upload. If we want to remove this zero padding we'll have to make a more complete change.

Again, this all falls into the bucket of changes I'm trying to make across our tooling for modular template upload.

izapolsk: I'll think over this once more and maybe try to make versions keep points.

izapolsk:
it was agreed to retain build number as is with dots in order to avoid all those problems with parsing.
@mshriver or me will do that soon.

PR #7437, opened 2018-07-06
[1LP][RFR]Extra navigation in test_sdn_crud.py

ganeshhubale:
* This test case needs to go to details page of provider. 
  which can be done by ```view = navigate_to(network_provider, 'Details')``` - It navigates to network > providers > Details page 
* ```view = navigate_to(provider, 'Details')``` - It navigates to compute > provider  > Details Page. This   **extra navigation** is not required for this test case. Because it is not used in test execution.

{{ pytest: cfme/tests/networks/test_sdn_crud.py }}

### DIFF
--- a/cfme/tests/networks/test_sdn_crud.py
+++ b/cfme/tests/networks/test_sdn_crud.py
@@ -22,7 +22,6 @@ def test_sdn_crud(provider, appliance):
     Metadata:
         test_flag: sdn
     """
-    view = navigate_to(provider, 'Details')
     collection = appliance.collections.network_providers.filter({'provider': provider})
     network_provider = collection.all()[0]
 


mshriver:
@ganeshhubale Nothing wrong with removing this extra navigation, but a note about navigation in general, for future reference.  Apologies if I'm repeating information you know.

The `CFMENavigateStep` implementation that is executed when `navigate_to` is used will always check `am_i_here` or `VIEW.is_displayed` before navigating, and throughout the nested navigation steps.

What this means, in practice, is that during a test, re-navigation to a page, or potentially duplicate navigation won't actually slow the tests down, and won't actually force navigation.  It will generally take the shortest path available to get to the destination.  If we're already on the providers 'All' page, and our destination is 'Details', we won't back out to the dashboard and start over again - we'll just go straight from 'All' to the specific entity via the `step` method.

PR #7438, opened 2018-07-06
[1LP][RFR] Automate test for deleting a report from saved reports list 

valaparthvi:
This automation tests deletion of a saved report from the saved reports list. It creates a custom report, saves it, navigates to the All Saved Reports, iterates through all the saved reports, and checks the report that was recently created and deletes it.

Note: Currently it takes some time to iterate through the list and read the data. This could be improved, but I'm not aware of another way right now. (Fixed)

{{ pytest: cfme/tests/intelligence/reports/test_crud.py -k test_reports_delete_saved_report }}


valaparthvi:
Using `row.name` didn't give the desired output, it returned a `TableColumn` object. But it certainly helped in decreasing the time. Using `row.name.read()` takes lesser time and gives desired output. Thanks @rlbabyuk !

cfme/tests/intelligence/reports/test_crud.py
```diff
@@ -240,3 +241,21 @@ def test_import_report(appliance):
...
+@test_requirements.report
+def test_reports_delete_saved_report(custom_report_values, appliance):
+    """This test case deletes reports from the Saved Reports list.
+    """
+    report = appliance.collections.reports.create(**custom_report_values)
+    report.queue(wait_for_finish=True)
+    view = navigate_to(appliance.collections.saved_reports, 'All')
+    # iterates through every row and checks if the 'Name' column matches the given value
+    for row in view.table.rows():
+        if row.name.read() == report.title:
```
rlbabyuk: also, we can use row.name.text instead of read()

cfme/tests/intelligence/reports/test_crud.py
```diff
@@ -240,3 +241,21 @@ def test_import_report(appliance):
...
+            row[0].check()
+    view.configuration.item_select(
+        item='Delete selected Saved Reports', handle_alert=True)
+    report.delete()
```
rlbabyuk: Required: Add a finalizer for this deletion
request.addfinalizer(report.delete)
and move it right aftre report create

cfme/tests/intelligence/reports/test_crud.py
```diff
@@ -240,3 +241,21 @@ def test_import_report(appliance):
...
+@pytest.mark.tier(3)
+@test_requirements.report
+def test_reports_delete_saved_report(custom_report_values, appliance):
+    """This test case deletes reports from the Saved Reports list.
+    """
+    report = appliance.collections.reports.create(**custom_report_values)
+    report.queue(wait_for_finish=True)
+    view = navigate_to(appliance.collections.saved_reports, 'All')
+    # iterates through every row and checks if the 'Name' column matches the given value
+    for row in view.table.rows():
```
rlbabyuk: Required: As SavedReport is a collection you can instantiate an entity and use simple saved_report.delete() as we have such method for it, and don't iterate all list
saved_report = appliance.collections.saved_reports.instantiate(params)

valaparthvi: Using `saved_report.delete()` would go the the `Details` page of that saved report and delete it from there. But, I wrote this test case to see if we can select reports from the Saved Reports list and delete them.

valaparthvi:
Hey @rlbabyuk , I've made the changes you requested. But my tests are failing on 5.9 and 5.10 for some reason. I'm getting `Incorrect username or password error` there. I think it's got nothing to do with my changes, so you can review it.

cfme/tests/intelligence/reports/test_crud.py
```diff
@@ -240,3 +241,21 @@ def test_import_report(appliance):
...
+    report = appliance.collections.reports.create(**custom_report_values)
+    request.addfinalizer(report.delete)
+    report.queue(wait_for_finish=True)
+    view = navigate_to(appliance.collections.saved_reports, 'All')
+    # iterates through every row and checks if the 'Name' column matches the given value
+    for row in view.table.rows():
+        if row.name.text == report.title:
+            row[0].check()
+    view.configuration.item_select(
+        item='Delete selected Saved Reports', handle_alert=True)
```
psav: **REQUIRED:** This is a great looking test, but there are a few things we should address.
1) The most important, is that though you invoke the button, you never check to see if the saved report is actually gone. We should at least refresh the list and ensure it is no longer there.
2) We should investigate making the saved report have a method to "delete". It is possible people would like to reuse this behaviour.
The first of these points is REQUIRED. The second is OPTIONAL, but should have an issue raised against it.

valaparthvi:
Hey @psav, I've added the changes you addressed. There are 2 things though.

1. I have made one minor change. Creating a custom report takes time, so instead of that, I have instantiated an entity of `Reports` and queued it, which takes lesser time. It is something @rlbabyuk suggested some time back, but I did not understand it at the time.

2. I am not sure I completely understand the second point about `delete` method. If I use saved_report's `delete` method, it will navigate to it's `Details` page and delete the saved report from there. But with this test, I am checking if we can delete reports from `All Saved Reports` list. And besides, we are already using `delete` method in other test cases(example: test_custom_report_crud).

PR #7442, opened 2018-07-09
[1LP][RFR] Add Bugzilla default URL for Travis

mshriver:
There is corresponding RH CFME QE yaml change.

Thanks @izapolsk for raising the issue of failing travis runs

### DIFF
--- a/cfme/utils/bz.py
+++ b/cfme/utils/bz.py
@@ -4,7 +4,7 @@ from bugzilla import Bugzilla as _Bugzilla
 from collections import Sequence
 
 from cached_property import cached_property
-from cfme.utils.conf import cfme_data, credentials
+from cfme.utils.conf import cfme_data, credentials, env
 from cfme.utils.log import logger
 from cfme.utils.version import (
     LATEST, Version, current_version, appliance_build_datetime, appliance_is_downstream)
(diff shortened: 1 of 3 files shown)


cfme/utils/bz.py
```diff
@@ -78,16 +78,18 @@ def default_product(self):
...
         if url is None:
-            raise Exception("No Bugzilla URL specified!")
-        cr_root = cfme_data.get("bugzilla", {}).get("credentials")
-        username = credentials.get(cr_root, {}).get("username")
-        password = credentials.get(cr_root, {}).get("password")
-        return cls(
-            url=url, user=username, password=password, cookiefile=None,
-            tokenfile=None, product=product)
+            url = 'https://bugzilla.redhat.com/xmlrpc.cgi'
+            logger.warning("No Bugzilla URL specified in conf, using default: %s", url)
```
mshriver: This looks way cleaner just using an `or` or providing as default to the `.get()`, but this way we can have indication that we ignored or didn't find BZ URL.

cfme/utils/bz.py
```diff
@@ -78,16 +78,18 @@ def default_product(self):
 
     @classmethod
     def from_config(cls):
-        url = cfme_data.get("bugzilla", {}).get("url")
-        product = cfme_data.get("bugzilla", {}).get("product")
+        bz_conf = env.get('bugzilla', {})  # default empty so we can call .get() later
```
izapolsk: I also thought about this. good idea !

PR #7447, opened 2018-07-10
[1LP][RFR]Fix dict key for test_db_backup_schedule

rlbabyuk:
Purpose or Intent
=================
{{pytest:: -k test_db_backup_schedule}}


psav:
For future ref this one should have been tagged for rc-regression fix :) It's so small

PR #7449, opened 2018-07-10
[1LP][RFR] Fixed creation of Nuage Provider

psav:
Nuage needed to correctly pass the class for instantiation


sasoc:
With this changes error is gone and Nuage provider works fine. Thank you :)

PR #7450, opened 2018-07-10
[1LP][RFR] Add Nuage to supportability template

sasoc:
Nuage is now in supportability templete so test get collected.


sasoc:
cc @miha-plesko

miha-plesko:
LGTM, @psav wdyt?

PR #7452, opened 2018-07-10
[WIPTEST] Set provider key to provider hostname

sasoc:
If you send `provider_key` as dictionary to `get_mgmt` function you get an error:
```
> if provider_key not in PROVIDER_MGMT_CACHE:
E TypeError: unhashable type: 'AttrDict'
```
So now `provider_key` is set as `provider_key['type']`

This PR is also rebased to https://github.com/ManageIQ/integration_tests/pull/7450

### DIFF
--- a/cfme/utils/providers.py
+++ b/cfme/utils/providers.py
@@ -387,6 +387,9 @@ def get_mgmt(provider_key, providers=None, credentials=None):
         provider_kwargs['provider_key'] = provider_key
     provider_kwargs['logger'] = logger
 
+    if isinstance(provider_key, Mapping):
+        provider_key = provider_key['hostname']
+
     if provider_key not in PROVIDER_MGMT_CACHE:
         mgmt_instance = get_class_from_type(provider_data['type']).mgmt_class(**provider_kwargs)
         PROVIDER_MGMT_CACHE[provider_key] = mgmt_instance
--- a/conf/supportability.yaml.template
+++ b/conf/supportability.yaml.template
@@ -53,6 +53,8 @@
         - 3.9
     physical:
         - lenovo
+    networks:
+        - nuage
 'master':
   providers:
     infra:
@@ -82,3 +84,5 @@
         - 3.9
     physical:
         - lenovo
+    networks:
+        - nuage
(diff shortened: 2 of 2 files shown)


sasoc:
cc @miha-plesko

miha-plesko:
It looks correct, but I'm not sure how it could work for other providers. @psav is Nuage the only one passing provider hash instead provider key here? Should we try to avoid this somehow or can we leave it like this?

cfme/utils/providers.py
```diff
@@ -358,6 +358,7 @@ def get_mgmt(provider_key, providers=None, credentials=None):
     # TODO rename the parameter; might break things
     if isinstance(provider_key, Mapping):
         provider_data = provider_key
+        provider_key = provider_key['type']
```
mshriver: I don't think that this is an assumption that we can make in all cases when a dictionary is passed instead of a bare key.

There are checks throughout the rest of the method that check provider_key type before acting on it, and I think we should consider modifying the check on L389/390 that's raising the exception to make the check based on the `provider_key` type.  See below for example.

dajoRH:
*CFME QE Bot*

sasoc:
@mshriver Thank you for your review. I have changed `type` to `hostname` as you suggested and moved it lower. Since `get_mgmt` returns cache, it should be stored in this case also.  What do you think, is it better like this?

cfme/utils/providers.py
```diff
@@ -387,6 +387,9 @@ def get_mgmt(provider_key, providers=None, credentials=None):
         provider_kwargs['provider_key'] = provider_key
     provider_kwargs['logger'] = logger
 
+    if isinstance(provider_key, Mapping):
```
mshriver: I'm still not confident in this change, and I think we need to look at approaching this from how the function is called and how he kwargs are used.

This exact check of the type of provider_key is already done at L359, setting provider_key to the 'name' attribute of the provider_data dictionary.  I'm not sure why that isn't functional in your case.

If you're trying to create a mgmt object for something that's not in cfme_data.yaml, then the custom dictionary should be passed as the second argument, `providers`.

miha-plesko:
Is this change still needed? Nuage CRUD test perfectly works for me on latest master without such modifications.

miha-plesko:
@sasoc can we close this one or is actually needed?

PR #7471, opened 2018-07-12
[1LP][RFR] ocp servers cleanup script

izapolsk:
Requires: https://github.com/ManageIQ/wrapanapi/pull/302

### DIFF
--- a/scripts/ocp_cleanup.py
+++ b/scripts/ocp_cleanup.py
@@ -1,0 +1,102 @@
+import argparse
+import re
+import sys
+
+from cfme.containers.provider.openshift import OpenshiftProvider
+from cfme.utils.providers import list_providers, ProviderFilter
+from cfme.utils.log import logger, add_stdout_handler
+
+add_stdout_handler(logger)
+
+global_scc_names = ['anyuid', 'privileged']
+
+default_namespace_regex = ['^.*?-project-',
+                           '^jenkins-.*',
+                           '^.*?s-appl-',
+                           '^test',
+                           '^long-test',
+                           '^external-test']
+
+
+def parse_cmd_line():
+    parser = argparse.ArgumentParser(argument_default=None)
+    parser.add_argument('-s', '--cleanup-scc', action='store_true',
+                        help="removes openshift sa from system scc which no longer belong to any "
+                             "appliance project")
+    parser.add_argument('text_to_match', nargs='*', default=default_namespace_regex,
+                        help='Regex in the name of scc->sa to be affected, '
+                             'can be used multiple times')
+    return parser.parse_args()
+
+
+def match(matchers, vm_name):
+    """Check whether the given vm name matches any of the regex match objects
+
+    Args:
+        matchers (list): A list of regex objects with match() method
+        vm_name (string): The name of the VM to compare
+    Returns:
+        bool: whether or not the vm_name is matched by ANY of the regex in the list
+    """
+    for matcher in matchers:
+        if matcher.match(vm_name):
+            return True
+    else:
+        return False
+
+
+def delete_stale_sa(provider, text_to_match):
+    """ Checks global Security Context Constrains for stale Service Account records created during
+        appliance deployment and removes such records if appliance doesn't exist any longer
+
+    Args:
+        provider: provider object
+        text_to_match: (list) regexps which sa should match to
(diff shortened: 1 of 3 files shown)


dajoRH:
Lint report for commit aa7f0e99c2abbb359fc28362957b50847e9a260c:

`scripts/ocp_cleanup.py`:
- :red_circle: Line 5:1: **F401** *'cfme.utils.providers.get_mgmt' imported but unused*
- :bangbang: Line 33:1: **E305** *expected 2 blank lines after class or function definition, found 1*

**No commit flaws detected**.

*CFME QE Bot*

scripts/ocp_cleanup.py
```diff
@@ -0,0 +1,95 @@
...
+                           '^jenkins-*',
+                           '^*?s-appl-',
+                           '^test',
+                           '^long-test',
+                           '^external-test']
+
+
+def parse_cmd_line():
+    parser = argparse.ArgumentParser(argument_default=None)
+    parser.add_argument('-s', '--cleanup-scc', action='store_true',
```
quarckster: **QUESTION**
This argument is not used here, isn't it?

izapolsk: good catch. there were more cleanup functions but I removed them along with args. 
I'll add this argument back to script

mshriver:
ManageIQ/wrapanapi#302 merged and released with 3.0.13, please include frozen update.

scripts/ocp_cleanup.py
```diff
@@ -0,0 +1,95 @@
...
+
+if __name__ == "__main__":
+    args = parse_cmd_line()
+    errors = 0
+    pf = ProviderFilter(classes=[OpenshiftProvider], required_fields=[('use_for_sprout', True)])
+    providers = list_providers(filters=[pf], use_global_filters=False)
+    for prov in providers:
+        # ping provider
+        try:
+            prov.list_project()
```
mshriver: This is so much better then a bunch of the current script implementations to actually do some pinging and pollute the logs unnecessarily. Nice!

scripts/ocp_cleanup.py
```diff
@@ -0,0 +1,95 @@
+import argparse
+import sys
+
+from cfme.containers.provider.openshift import OpenshiftProvider
+from cfme.utils.providers import list_providers, ProviderFilter
+from cfme.utils.log import create_sublogger, add_stdout_handler
+
+logger = create_sublogger('ocp_cleanup')
+add_stdout_handler(logger)
```
mshriver: I think there's an issue with calling add_stdout_handler here, since create_sublogger returns a NamedLoggerAdapter:

```
  File "scripts/ocp_cleanup.py", line 9, in <module>
    add_stdout_handler(logger)
  File "/home/mshriver/repos/integration_tests/cfme/utils/log.py", line 531, in add_stdout_handler
    for handle in logger.handlers:
AttributeError: 'NamedLoggerAdapter' object has no attribute 'handlers'
```

Looks like you'll need to pass logger.logger, using the `logging.LoggerAdapter` attribute that stores the original logger object, to add_stdout_handler.

izapolsk: I didn't expect that. fixed.

mshriver: I wouldn't have either, its a bit odd of behavior, but useful since the logger's an easily accessible attribute of the wrapper class.

requirements/frozen_docs.py2.txt
```diff
@@ -41,8 +41,10 @@ funcsigs==1.0.2
 functools32==3.2.3.post2
 future==0.16.0
 futures==3.0.5
+gevent==1.3.5
```
mshriver: gevent is included in the mocked modules for the docs build, so it doesn't actually need to be installed.

requirements/frozen_docs.py2.txt
```diff
@@ -172,6 +174,7 @@ Werkzeug==0.14.1
 widgetastic.core==0.21.3
 widgetastic.patternfly==0.0.34
 widgetsnbextension==3.2.1
+wrapanapi==3.0.14
```
mshriver: I had just removed wrapanapi from docs, because its all mocked. this was causing the doc env build to fail.

PR #7473, opened 2018-07-13
[1LP][RFR] support for azure ec2 and rhv graph

digitronik:
Purpose or Intent
=================

{{pytest: cfme/tests/candu/test_vm_graph.py -k test_graph_screen -v --use-provider complete}}

### DIFF
--- a/cfme/tests/candu/test_vm_graph.py
+++ b/cfme/tests/candu/test_vm_graph.py
@@ -1,6 +1,7 @@
 import pytest
 
 from cfme import test_requirements
+from cfme.cloud.provider import CloudProvider
 from cfme.cloud.provider.azure import AzureProvider
 from cfme.cloud.provider.ec2 import EC2Provider
 from cfme.infrastructure.provider.rhevm import RHEVMProvider
@@ -15,7 +16,8 @@ pytestmark = [
     test_requirements.c_and_u,
     pytest.mark.usefixtures('setup_provider'),
     pytest.mark.provider([VMwareProvider, RHEVMProvider, EC2Provider, AzureProvider],
-                         required_fields=[(['cap_and_util', 'capandu_vm'], 'cu-24x7')])
+                        scope='module',
+                        required_fields=[(['cap_and_util', 'capandu_vm'], 'cu-24x7')])
 ]
 
 VM_GRAPHS = ['vm_cpu', 'vm_cpu_state', 'vm_memory', 'vm_disk', 'vm_network']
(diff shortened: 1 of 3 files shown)


digitronik:
Two test will fail for network graphs but is ok. 
We need to add network creation script on cu-24x7 for the cloud provider. Its required infra issue not related with code or logic.

Local Results:
http://pastebin.test.redhat.com/622361

cfme/tests/candu/test_vm_graph.py
```diff
@@ -15,7 +18,8 @@
     test_requirements.c_and_u,
     pytest.mark.usefixtures('setup_provider'),
     pytest.mark.provider([VMwareProvider, RHEVMProvider, EC2Provider, AzureProvider],
-                         required_fields=[(['cap_and_util', 'capandu_vm'], 'cu-24x7')])
+                        scope='module',
```
rlbabyuk: Required: provider should be function scope, remove scope='module'

cfme/tests/candu/test_vm_graph.py
```diff
@@ -81,11 +85,39 @@ def refresh():
...
+        tb_data : Data from Utilization table
+        gp_data : Data from Utilization graph
+        legends : Legends in graph; which will help for comparison
+        tolerance : Its error which we have to allow while comparison
+    """
+    for row in tb_data:
+        for key, data in gp_data.items():
+            if any([re.match(key, item) for item in row['Date/Time'].split()]):
+                for leg in legends:
+                    tb = row[leg].replace(',', '').replace('%', '').split()
```
rlbabyuk: Required: use variables full names

cfme/tests/candu/test_vm_graph.py
```diff
@@ -81,11 +85,39 @@ def refresh():
...
+                    if tb:
+                        tb = round(float(tb[0]), 1)
+                        gp = round(
```
rlbabyuk: Required: full name please

cfme/tests/candu/test_vm_graph.py
```diff
@@ -81,11 +85,39 @@ def refresh():
...
+                            float(data[leg].replace(',', '').replace('%', '').split()[0]), 1)
+                        assert abs(tb - gp) <= tolerance
```
rlbabyuk: Required: add a message what we assert

cfme/tests/candu/test_vm_graph.py
```diff
@@ -81,11 +85,39 @@ def refresh():
...
+                    else:
+                        logger.warning("No {leg} data captured for DateTime: {dt}".format(
+                            leg=leg, dt=row['Date/Time']))
+
+
+@pytest.mark.uncollectif(lambda provider, interval, graph_type:
```
rlbabyuk: Required: You can override provider here and specify what you need
ex. @pytest.mark.provider([VMwareProvider], override=True)

digitronik: @rlbabyuk: It's not possible with provider overriding. It can be used only when the uncollect decision depends only on the `provider`. But, In this case, we are uncollecting things with `provider`, `Interval` and `graph type`.

cfme/tests/candu/test_vm_graph.py
```diff
@@ -81,11 +84,40 @@ def refresh():
...
+                    table_item = row[leg].replace(',', '').replace('%', '').split()
+                    if table_item:
+                        table_item = round(float(table_item[0]), 1)
+                        graph_item = round(
+                            float(data[leg].replace(',', '').replace('%', '').split()[0]), 1)
+                        cmp_data = abs(table_item - graph_item) <= tolerance
+                        assert cmp_data, "compare graph and table readings with tolerance"
+                    else:
+                        logger.warning("No {leg} data captured for DateTime: {dt}".format(
+                            leg=leg, dt=row['Date/Time']))
```
psav: **REQUIRED:** This code has already been merged once. As it is quite complex, is there a reason why we cannot move this to a common location please?

digitronik: @psav: As per your suggestion. I moved to `compare_data` to `__init__.py`. which will share to all graph tests files. In same PR, I am including `test_host_graph.py` changes.

digitronik:
@psav done with PRT. Azure network fails as expected due to lack of network activity..... Please, review it one time.

PR #7477, opened 2018-07-15
[1LP][RFR] Add test for validating provider column in all projects table

shalomnaim1:
{{ pytest : cfme/tests/containers/test_projects_dashboard.py::test_project_has_provider --use-provider ocp-36-hawk }}

### DIFF
--- a/cfme/tests/containers/test_projects_dashboard.py
+++ b/cfme/tests/containers/test_projects_dashboard.py
@@ -103,3 +103,26 @@ def test_projects_dashboard_icons(provider, appliance, soft_assert, container_pr
                 containers_cls.__name__, api_values[containers_cls], statusbox_value
             )
         )
+
+
+def test_project_has_provider(appliance, soft_assert, provider):
+    """
+    Test provider name existence in all projects table.
+    Steps:
+      * navigate to all project page
+      * get through all the project to ensure that the provider column isn't
+        empty on each on each of the projects
+    """
+    projects_collection = appliance.collections.container_projects
+
+    all_project_view = navigate_to(projects_collection, "All")
+    all_tables_rows = all_project_view.entities.get_all()
+
+    assert all_tables_rows, "No table row was found"
+
+    for row in all_tables_rows:
+        curr_project_name = row.data["name"]
+        curr_project_provider = row.data["provider"]
+
+        soft_assert(curr_project_provider,
+                    "No Provider found for project {name}".format(name=curr_project_name))


dajoRH:
Lint report for commit e01be1c0a55f85ee247a31e84054e79357e61330:

`cfme/tests/containers/test_projects_dashboard.py`:
- :bangbang: Line 107:1: **E302** *expected 2 blank lines, found 1*
- :bangbang: Line 117:101: **E501** *line too long (113 > 100 characters)*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/containers/test_projects_dashboard.py
```diff
@@ -103,3 +103,17 @@ def test_projects_dashboard_icons(provider, appliance, soft_assert, container_pr
                 containers_cls.__name__, api_values[containers_cls], statusbox_value
             )
         )
+
+
+def test_project_has_provider(appliance, soft_assert):
```
rlbabyuk: Required: Add doc string, with info what we are testing here

shalomnaim1: @rlbabyuk doc string added

shalomnaim1:
@rlbabyuk Please approve

cfme/tests/containers/test_projects_dashboard.py
```diff
@@ -103,3 +103,24 @@ def test_projects_dashboard_icons(provider, appliance, soft_assert, container_pr
...
+    Steps:
+      * navigate to all project page
+      * get through all the project to ensure that the provider column isn't
+        empty on each on each of the projects
+    """
+    projects_collection = appliance.collections.container_projects
+
+    all_project_view = navigate_to(projects_collection, "All")
+    all_tables_rows = all_project_view.entities.get_all()
+
```
izapolsk: REQUIRED: I'd add assert statement that all_tables_rows isn't empty. Otherwise, this test will test nothing if there isn't projects.

cfme/tests/containers/test_projects_dashboard.py
```diff
@@ -103,3 +103,24 @@ def test_projects_dashboard_icons(provider, appliance, soft_assert, container_pr
                 containers_cls.__name__, api_values[containers_cls], statusbox_value
             )
         )
+
+
+def test_project_has_provider(appliance, soft_assert):
```
izapolsk: REQUIRED: I think you need to add provider fixture here. If it is absent and you try to run this test only, it is possible that provider won't be added.

cfme/tests/containers/test_projects_dashboard.py
shalomnaim1: @izapolsk Well this is a good point, I am working to this...

cfme/tests/containers/test_projects_dashboard.py
shalomnaim1: @izapolsk At the beginning of this module there is a call for ```setup_provider```, I am not sure but I think that provider is added when any test is triggered on this suite, right?

https://github.com/shalomnaim1/integration_tests/blob/1871af20e6849deb2edc5491795ec4b06ba3c02f/cfme/tests/containers/test_projects_dashboard.py#L12

izapolsk: setup_provider fixture works on top of provider fixture. You provider fixture has scope="function", what means that provider has to be created and removed for every test where provider fixture is present.

This is rather bug that your provider is present in test where provider fixture isn't present.

So, I think you should either add provider fixture to test or change scope to module for mark.provider.
Second is preferable because it won't make pytest add and remove provider for every test in file.

shalomnaim1: @izapolsk I think I got it, I added ```setup_provider```

PR #7486, opened 2018-07-17
[RFR] make sync_template_trackerbot script ignore templates with empty name

izapolsk:
it turned out that some providers have templates w/o name. 
When such templates are processed by sync_template_tracker script, those are 1. added to trackerbot; 2. all templates are removed  from trackerbot.
this PR adds some protection from such templates.


mshriver:
Hey, you're removing my feature! :P

PR #7503, opened 2018-07-19
[1LP][RFR] Fixed creation of custom buttons

quarckster:
{{pytest: -v cfme/tests/automate/test_buttons.py::test_button_crud}}


jkrocil:
Tested locally; works. This fixes 10 tests in automate in 5.10.

PR #7510, opened 2018-07-20
[RFR][1LP] Fixed automate schema crud

quarckster:
{{pytest: -v cfme/tests/automate/test_class.py::test_schema_crud}}

### DIFF
--- a/cfme/automate/explorer/klass.py
+++ b/cfme/automate/explorer/klass.py
@@ -6,7 +6,7 @@ from cached_property import cached_property
 from copy import copy
 
 from navmazing import NavigateToAttribute, NavigateToSibling
-from widgetastic.utils import ParametrizedLocator, ParametrizedString
+from widgetastic.utils import ParametrizedLocator, ParametrizedString, Version, VersionPick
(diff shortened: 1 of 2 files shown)


jawatts:
:+1:

cfme/automate/explorer/klass.py
```diff
@@ -371,8 +371,11 @@ def row_id(self):
             max_time = Input(name=ParametrizedString('fields_max_time_{@row_id}'))
 
             def delete(self):
-                self.browser.click(
-                    './/img[@alt="Click to delete this field from schema"]', parent=self)
+                if self.browser.product_version < '5.10':
```
jawatts: Question:
Why not use VersionPick here?

quarckster: VersionPick is a descriptor and it's supposed to be used for class attributes. For view instances `self.browser.product_version` should be used.

psav: Yeh VersionPick could be used here, though it is generally thought to be more efficient to do a simple comparison here.

PR #7520, opened 2018-07-23
[1LP][RFR] Fix in the case a required field is a string

psav:
* In the case of nested keys, sometimes different providers have set
things up differently and have a string where another provider has a
set of nested keys.
* This makes that allowable when provider filtering.

### DIFF
--- a/cfme/tests/infrastructure/test_tenant_quota.py
+++ b/cfme/tests/infrastructure/test_tenant_quota.py
@@ -254,7 +254,7 @@ def test_setting_child_quota_more_than_parent(tenants_setup, parent_quota, child
 
 @pytest.mark.long_running
 @pytest.mark.provider([VMwareProvider, RHEVMProvider], override=True, scope="module",
-                      required_fields=['templates', 'small_template'], selector=ONE_PER_TYPE)
+                      required_fields=[['templates', 'small_template']], selector=ONE_PER_TYPE)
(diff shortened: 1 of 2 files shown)


mshriver:
Marking rc-regression because this impacts test collection and we have plenty of current uses of lists in `required_fields`

mshriver:
collection with `long_running` + `complete` before change:
```
Uncollection Stats:
 manual: 19
 uncollectif: 821
 5977 tests left after all uncollection
```

Same collection, after:
```
Uncollection Stats:
 manual: 19
 uncollectif: 821
 5975 tests left after all uncollections
```

Collection difference:
```
<Module 'cfme/tests/infrastructure/test_tenant_quota.py'>
  <Function 'test_vm_migration_after_assigning_tenant_quota[virtualcenter-max_cores]'>
  <Function 'test_vm_migration_after_assigning_tenant_quota[rhevm-max_cores]'>
```

This is due to the required_fields not using a nested list for that test function. I tried to make the required change in your branch - but it appears you created this branch from an older commit, and I am not sure of a way to cleanly bring the change into your branch without rewriting history.

All that needs to be done is wrap the current required_fields list here: 
https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/infrastructure/test_tenant_quota.py#L257

Then collection should be the same.

PR #7522, opened 2018-07-23
[RFR][1LP] Cleanup vms handle error

mshriver:
Running locally with wrapanapi fixes.

### DIFF
--- a/requirements/frozen_docs.py3.txt
+++ b/requirements/frozen_docs.py3.txt
@@ -35,8 +35,10 @@ fauxfactory==3.0.2
 flake8==3.5.0
 funcsigs==1.0.2
 future==0.16.0
+gevent==1.3.5
 gitdb2==2.0.3
 GitPython==2.1.10
+greenlet==0.4.14
 html5lib==1.0.1
 idna==2.6
 importscan==0.1
--- a/scripts/cleanup_old_vms.py
+++ b/scripts/cleanup_old_vms.py
@@ -10,6 +10,7 @@ from multiprocessing import Manager, Pool
 
 import pytz
 from tabulate import tabulate
+from wrapanapi.exceptions import VMInstanceNotFound
 
 from cfme.utils.log import logger, add_stdout_handler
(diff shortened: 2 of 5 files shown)


requirements/frozen_docs.py3.txt
```diff
@@ -35,8 +35,10 @@ fauxfactory==3.0.2
 flake8==3.5.0
 funcsigs==1.0.2
 future==0.16.0
+gevent==1.3.5
```
mshriver: These are mocked in the docs build, but freeze_all is going to keep adding them, so FINE lets let it try to do a docs build with them. :)

scripts/cleanup_old_vms.py
```diff
@@ -188,7 +193,7 @@ def scan_vm(provider_key, vm, delta, match_queue, scan_failure_queue):
         logger.info('%r: VM %r did not match age requirement', provider_key, vm.name)
 
 
-def delete_vm(provider_key, vm, age, result_queue):
```
psav: **REQUIRED:** These seem to be going against the new model of always passing in the objects, I know there must be a reason for this, can you explain it a little?

mshriver: Yep - the objects being passed into the function are done through the pool_manager, and have to be pickled. The wrapanapi VM objects aren't pickle-able.  I know I was hitting it on trying to get the namedtuple results, and changed them going into the function as well, but I'll double check that behavior.

mshriver: Part of the reasoning is that this script is running independent of any tests, and the VM could have been deleted by tests or users between scanning the provider, scanning the vm, and trying to delete.  So each time we call `get_vm` on it to fetch it, though this is secondary and we could have vm.exists traps.

psav:
@mshriver and I discussed some of the patterns that we are not particularly happy with. We have devised a potential forward plan for this, but we are happy with what we have, on the proviso that we add an underscore to the bad pattern functions.

PR #7523, opened 2018-07-24
[1LP][NOTEST][RFR] This PR will add wait between validate and save operation.

sbulage:


cfme/tests/cloud_infra_common/test_vm_instance_analysis.py
```diff
@@ -167,7 +168,9 @@ def set_agent_creds(appliance, request, provider):
     ssa_agent_view = view.browser.create_view(EC2EndpointForm)
     ssa_agent_view.smart_state_docker.fill({'username': username, 'password': password})
     ssa_agent_view.default.validate.click()
+    time.sleep(35)
```
sbulage: After validation, we have to wait for approx.30-35 sec.  for credentials to be validated from docker.io. Tried all possible conditions which can be use in `wait_for` but not showing required output hence `time.sleep` is added.

rlbabyuk: @sbulage please add bz, or explanation that we discussed

sbulage: @rlbabyuk yeah sure, I will add it after discussion with devel.

cfme/tests/cloud_infra_common/test_vm_instance_analysis.py
```diff
@@ -167,7 +168,10 @@ def set_agent_creds(appliance, request, provider):
     ssa_agent_view = view.browser.create_view(EC2EndpointForm)
     ssa_agent_view.smart_state_docker.fill({'username': username, 'password': password})
     ssa_agent_view.default.validate.click()
+    if BZ(1608023, forced_streams=['5.9']).blocks:
```
rlbabyuk: Question: How about 5.10? is it working ok

cfme/tests/cloud_infra_common/test_vm_instance_analysis.py
```diff
@@ -167,7 +168,10 @@ def set_agent_creds(appliance, request, provider):
...
+        time.sleep(35)
     view.save.click()
+    view.browser.refresh()
```
rlbabyuk: Question: why we need this view.browser.refresh() here

sbulage: @rlbabyuk After saving creds, it won't reflect immediately and navigate to other page.

psav:
We REALLY need a comment in here stating how bad this is :) As well as the BZ :)

cfme/tests/cloud_infra_common/test_vm_instance_analysis.py
```diff
@@ -171,7 +172,11 @@ def set_agent_creds(appliance, request, provider):
     ssa_agent_view = view.browser.create_view(EC2EndpointForm)
     ssa_agent_view.smart_state_docker.fill({'username': username, 'password': password})
     ssa_agent_view.default.validate.click()
+    # TODO: Remove time.sleep asap BZ fix.
+    if BZ(1608023, forced_streams=['5.9', '5.10']).blocks:
+        time.sleep(35)
```
mshriver: gross.

sbulage: :(

mshriver: I've had to do it in the auth tests too, at least its wrapped in a BZ.

PR #7524, opened 2018-07-24
[1LP][RFR] Modify is_displayed method in every provider's Edit view

valaparthvi:
Modified `is_displayed` methods inside the following edit views:
1. `InfraProviderEditView`
2. `PhysicalProviderEditView`
3. `ContainerProviderEditView`

This change was made to fix the `is_displayed` methods of the above mentioned views, which returned `False` even if the page was displayed.

### DIFF
--- a/cfme/common/provider_views.py
+++ b/cfme/common/provider_views.py
@@ -564,62 +564,20 @@ class ProviderEditView(ProviderAddView):
 
     @property
     def is_displayed(self):
-        return self.logged_in_as_current_user
+        provider_obj = self.context['object']
+        expected_title = ("Edit {type} Providers '{name}'".format(
+            type=provider_obj.string_name, name=provider_obj.name))
+
+        return (self.logged_in_as_current_user and
+                self.title.text == expected_title and
+                self.navigation.currently_selected == [
+                    'Compute',
+                    ('Clouds' if provider_obj.string_name == "Cloud" else provider_obj.string_name),
+                    'Providers'
+                ])
 
 
-class InfraProviderEditView(ProviderEditView):
-    """
-     represents Infra Provider Edit View
-    """
-    @property
-    def is_displayed(self):
-        expected_title = ("Edit Infrastructure Providers '{name}'".format(
-            name=self.context['object'].name))
-        return (super(InfraProviderEditView, self).is_displayed and
-                self.navigation.currently_selected == ['Compute', 'Infrastructure', 'Providers'] and
-                self.title.text == expected_title)
-
-
-class PhysicalProviderEditView(ProviderEditView):
-    """
-     represents Provider Edit View
-    """
-    @property
-    def is_displayed(self):
-        expected_title = ("Edit Physical Infrastructure Providers '{name}'"
-                          .format(name=self.context['object'].name))
-        return (super(PhysicalProviderEditView, self).is_displayed and
-                self.navigation.currently_selected ==
-                ['Compute', 'Physical Infrastructure', 'Providers'] and
-                self.title.text == expected_title)
-
-
-class CloudProviderEditView(ProviderEditView):
-    """
-     represents Cloud Provider Edit View
-    """
-    @property
-    def is_displayed(self):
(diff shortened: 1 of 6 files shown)


cfme/common/provider_views.py
```diff
@@ -572,9 +573,11 @@ class InfraProviderEditView(ProviderEditView):
     """
     @property
     def is_displayed(self):
+        expected_title = ("Edit Infrastructure Providers '{name}'".format(
```
rlbabyuk: Question: This works for 5.10 and 5.9 versions?

valaparthvi: Yes, it works for both 5.10 and 5.9 versions.

psav: Sooo, I'd like you to investigate this
**REQUIRED:** Inside each of these classes there is a ```string_name``` which is going to be either ```Cloud``` or ```Infrastructure``` allowing you to do ```self.context['object'].string_name``` as a replacement for the static strings above. This means you could almost push this definition up into the common file. Please experiment and see if it works.

valaparthvi:
@psav I need your opinion here. I was thinking to change the `is_displayed` method of `ProviderEditView` to this:
```
    @property
    def is_displayed(self):
        expected_title = ("Edit {type} Providers '{name}'".format(
            type=self.context['object'].string_name,
            name=self.context['object'].name))

        return (self.logged_in_as_current_user and self.title.text == expected_title and
                self.navigation.currently_selected == [
                    'Compute', '{}'.format(self.context['object'].string_name), 'Providers'])
```

With this change, we can remove `CloudProviderEditView`, `ContainerProviderEditView`, `InfraProviderEditView`, `PhysicalProviderEditView` and make other required changes in the corresponding navigation views. 

What do you think about it?

dajoRH:
*CFME QE Bot*

dajoRH:
Lint report for commit 888c0cebd7e9644ecbe4e2f0194744bb2fea69c8:

`cfme/common/provider.py`:
- :bangbang: Line 738:101: **E501** *line too long (105 > 100 characters)*

**No commit flaws detected**.

*CFME QE Bot*

valaparthvi:
@psav So, I've made the following changes:
1. Modify `ProviderEditView`'s `is_displayed` method - use `self.view['object'].string_name` and `self.view['object'].name` instead of explicit `string_name` and `name`. I had to add a special case for cloud provider, since the there's an extra `s` at the end of it's `string_name`. :roll_eyes: 

2. Remove `CloudProviderEditView`, `InfraProviderEditView`, `PhysicalProviderEditView` and `ContainerProviderEditView`.

3. Modify import statements(I have set the modules in alphabetical order, so you might see some unrelated changes) and other places that used the different EditViews.

dajoRH:
*CFME QE Bot*

izapolsk:
as far as I can recall we decided that forms like Edit form, should always have is_displayed = False.
So, I'm not sure we need this change at all. I'll look at it closer a bit later.

valaparthvi:
@izapolsk I started working on this PR because `is_displayed` was returning `False`(check the PR description). I don't understand why it should intentionally return `False`.

izapolsk:
@valaparthvi, there were several reasons. one of them is that, it should be possible to navigate to this form second time. If you are already on this page, navmazing won't allow you to navigate second time because is_displayed == True.

valaparthvi:
@izapolsk  Can you point me towards a case where it is required to navigate a second time?

izapolsk:
@valaparthvi, yep, different tests when Edit form is filled in with wrong values. there were some issues from external teams when their tests failed because couldn't navigate to edit form again.

valaparthvi:
@izapolsk I am at a lost here. As far as I remember, `is_displayed` method of `InfraProviderEditView`, `CloudProviderEditView` returned `False`. Do you suggest I make changes such that `is_displayed` method of these views return `False`?

izapolsk:
@valaparthvi, nevermind, it seems we changed our mind but I forgot about that. Those views had to return False. However, those have some comparison rules in is_displayed for some reason. So, it's probably ok. code looks good to me. so, I'm merging it.

PR #7530, opened 2018-07-25
[RFR] Fix ValueError in host yaml lookup

mshriver:
the `host_data,` type unpacking raises a ValueError when there are multiple items in the list.


mshriver:
PRT 25432 collection failure because of an assertion in pytest_generate_tests.  _/me prepares to throw a trout at someone_

PR #7534, opened 2018-07-26
[1LP][RFR] provision tests/views fixes

rlbabyuk:
Purpose or Intent
=================
1. Fixed template select in provision table
2. Added BZ blocker for 5.10

{{pytest: cfme/tests/cloud_infra_common/test_provisioning.py -k test_provision_approval}}

### DIFF
--- a/cfme/common/vm_views.py
+++ b/cfme/common/vm_views.py
@@ -374,6 +374,11 @@ class ProvisionView(BaseLoggedInPage):
     image_table = Table('//div[@id="pre_prov_div"]//table')
 
     @View.nested
+    class sidebar(View):  # noqa
+        extend_button = Text(locator='//div[contains(@class, "resize-right")]')
+        decrease_button = Text(locator='//div[contains(@class, "resize-left")]')
+
+    @View.nested
     class form(BasicProvisionFormView):  # noqa
         """First page of provision form is image selection
         Second page of form is tabbed with nested views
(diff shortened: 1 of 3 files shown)


cfme/utils/appliance/implementations/ui.py
```diff
@@ -172,6 +172,9 @@ def _check():
             result = self.browser.execute_script(self.ENSURE_PAGE_SAFE, silent=True)
             # TODO: Logging
             return bool(result)
+        time.sleep(0.2)
+        # added this sleep as loading, and other elements that we are checking
+        # in ENSURE_PAGE_SAFE script appears on page after script is run
```
psav: **REQUIRED:** I don't see why this is needed, we have silent failures on here and wait for 2 seconds overall? Any failure will jsut cause a retry in 0.2 seconds.

rlbabyuk: removed

cfme/common/vm_views.py
```diff
@@ -373,6 +373,11 @@ class ProvisionView(BaseLoggedInPage):
     breadcrumb = BreadCrumb()
     image_table = Table('//div[@id="pre_prov_div"]//table')
 
+    @View.nested
+    class sidebar(View):  # noqa
+        extend_button = Text(locator='//div[contains(@class, "resize-right")]')
```
psav: **REQUIRED:** Please use either:
```extend / reduce```
or
```increase / decrease```

File: cfme/common/vm_views.py [hunk 3]
```diff
@@ -382,6 +387,14 @@ class form(BasicProvisionFormView):  # noqa
         submit_button = Button('Submit')  # Submit for 2nd page, tabular form
         cancel_button = Button('Cancel')
 
+        def _select_template(self, template_name, provider_name):
+            try:
+                row = self.parent.image_table.row(name=template_name, provider=provider_name)
+            except IndexError:
+                raise TemplateNotFound('Cannot find template "{}" for provider "{}"'
+                                       .format(template_name, provider_name))
+            row.click()
```
psav: **REQUIRED:** How can we click if we hit an exception finding it?

cfme/common/vm_views.py
rlbabyuk: done

cfme/common/vm_views.py
rlbabyuk: done

PR #7536, opened 2018-07-26
[WIP] Applying workaround for chrome

jan-zmeskal:


cfme/utils/appliance/implementations/ui.py
```diff
@@ -218,7 +218,9 @@ def after_click(self, element, locator):
         # page_dirty is set to None because otherwise if it was true, all next ensure_page_safe
         # calls would check alert presence which is enormously slow in selenium.
         if not isinstance(locator, (Input)):
-            if self.browser.browser_type.lower() == 'firefox' and self.browser.browser_version > 45:
+            browser_type = self.browser.browser_type.lower()
+            if browser_type == 'chrome' \
```
mshriver: Use parenthesis instead of line-continuation please.

mshriver:
PRT will be of limited relevance because it is only running FF

PR #7544, opened 2018-07-27
[1LP][RFR] Added relationship between Openstack Infra provider and Openstack Node

quarckster:
This PR adds ability to get all openstack nodes from an openstack provider:
```python
nodes = provider.nodes.all()
```
Requires #7549

### DIFF
--- a/cfme/infrastructure/provider/openstack_infra.py
+++ b/cfme/infrastructure/provider/openstack_infra.py
@@ -8,6 +8,7 @@ from wrapanapi.systems import OpenstackInfraSystem
 
 from cfme.common.provider import EventsEndpoint, SSHEndpoint, DefaultEndpoint, DefaultEndpointForm
 from cfme.common.provider_views import BeforeFillMixin, ProviderNodesView
+from cfme.infrastructure.openstack_node import OpenstackNodeCollection
 from cfme.infrastructure.provider import InfraProvider
 from cfme.utils.appliance.implementations.ui import navigate_to, CFMENavigateStep, navigator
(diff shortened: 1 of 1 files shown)


cfme/infrastructure/openstack_node.py
```diff
@@ -41,3 +41,15 @@ class OpenstackNodeCollection(HostsCollection):
...
+    def all(self):
+        """returning all openstack nodes objects"""
+        from cfme.infrastructure.provider.openstack_infra import OpenstackInfraProvider
+        parent = self.filters.get('parent')
+        if parent:
+            parent.rest_api_entity.reload(attributes=['nodes'])
+            nodes = parent.rest_api_entity.nodes
+        else:
+            return super(OpenstackNodeCollection, self).all()
+        provider = parent if isinstance(parent, OpenstackInfraProvider) else None
```
psav: **REQUIRED:** Am I right in assuming the main reason for this separate function is that we need to protect against other providers and we need to specify thta we are talking about an OpenStack type provider. In which case. Can we not assign a the OpenstackNodeCollection object to know that it requires OpenstackInfraProvider as the provider as an attribute, so that we can have a common function? Or do you think the specificity is better here? I just worry about the functionality of the Hosts function here is going to change and that we will not know about its usage here?

quarckster: I was thinking openstack provider has different attribute to access its nodes, I checked it in appliance with RHOS UC and I found it's the same. So this code is not needed at all.

PR #7546, opened 2018-07-27
[RFR] Fixed destination error

psav:
* The NavigationDestinationNotFound is internal to navmazing and though
  it could be raised here, it would probably give an errorneous error :)
  It would say something like, "can't find destination [Page] in object,
  available destinations [Page]" which would lead to confusion.
* If we really can't navigate, let the error bubble.

This is +212 in latest jenkins run

### DIFF
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -8,15 +8,6 @@ from copy import copy
 import attr
 import fauxfactory
 import re
-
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit b3f8170cd251a1cfe4765795755410a548c24126:

- :bangbang: Line 17:1: **E122** *continuation line missing indentation or outdented*

**No commit flaws detected**.

*CFME QE Bot*

PR #7547, opened 2018-07-27
[1LP][RFR] Host collections are now accessible from the provider

psav:
Host collections are now accessible from the provider

### DIFF
--- a/cfme/tests/infrastructure/test_advanced_search_host.py
+++ b/cfme/tests/infrastructure/test_advanced_search_host.py
@@ -4,9 +4,8 @@ from itertools import dropwhile
 
 import fauxfactory
 import pytest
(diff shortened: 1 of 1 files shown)


cfme/tests/infrastructure/test_advanced_search_host.py
```diff
@@ -20,7 +19,7 @@ def host_collection(appliance):
 def hosts(infra_provider, host_collection):
     view = navigate_to(host_collection, 'All')
     view.entities.search.remove_search_filters()
-    return host_collection.all(infra_provider)
+    return infra_provider.collections.hosts.all()
```
quarckster: **OPTIONAL**
```python
return infra_provider.hosts.all()
```

PR #7549, opened 2018-07-27
[1LP][RFR] HostsCollection.all() uses REST API

quarckster:
Intent
==============
`all()` method in HostsCollection uses REST API. Thanks @mkoura for the help.

{{pytest: -v -k "test_schedule_timer or test_vm_migrate or test_host_drift_analysis" --long-running}}

### DIFF
--- a/cfme/common/provider.py
+++ b/cfme/common/provider.py
@@ -136,11 +136,15 @@ class BaseProvider(Taggable, Updateable, Navigatable, BaseEntity):
         return self.type_name
 
     @property
+    def rest_api_entity(self):
+        return self.appliance.rest_api.collections.providers.get(name=self.name)
+
+    @property
     def id(self):
         """"
         Return the ID associated with the specified provider name
         """
-        return self.appliance.rest_api.collections.providers.find_by(name=self.name)[0].id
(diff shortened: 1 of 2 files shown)


quarckster:
Travis failed with the error: "AttributeError: 'DummyAppliance' object has no attribute 'rest_api'"

cfme/common/provider.py
```diff
@@ -97,6 +97,7 @@ class BaseProvider(Taggable, Updateable, Navigatable, BaseEntity):
     def __attrs_post_init__(self):
         # attr.ib(convert=prepare_endpoints) doesn't work correctly. this is workaround
         self.endpoints = prepare_endpoints(self.endpoints)
+        self.rest_api_entity = self.appliance.rest_api.collections.providers.get(name=self.name)
```
psav: **REQUIRED:** This breaks ```get_crud``` if the provider hasn't been added to the appliance. We should do some checking here, remember we can instantiate without it being on the appliance. I think perhaps it would be better to have a sync method. This will perhaps look for it and sync certain values over time, but perhaps to start with it just syncs the id?

PR #7552, opened 2018-07-27
[1LP][RFR] fix ui button name to add Physical Infrastructure Provider

matheuscmelo:
this PR fixes the name of the button the test should click to add a new physical provider

### DIFF
--- a/cfme/physical/provider/__init__.py
+++ b/cfme/physical/provider/__init__.py
@@ -157,5 +157,5 @@ class Add(CFMENavigateStep):
 
     def step(self):
         self.prerequisite_view.toolbar.configuration.item_select(
-            'Add a New Physical Provider'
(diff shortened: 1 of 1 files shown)


cfme/physical/provider/__init__.py
```diff
@@ -157,5 +157,5 @@ class Add(CFMENavigateStep):
 
     def step(self):
         self.prerequisite_view.toolbar.configuration.item_select(
-            'Add a New Physical Provider'
+            'Add a New Physical Infrastructure Provider'
```
quarckster: **QUESTION**
Does it work with both 5.9 and 5.10?

matheuscmelo: yes, it does.
5.9:
screenshot_20180730_132935
5.10:
screenshot_20180730_133204

PR #7554, opened 2018-07-29
[1LP][RFR] Update ContainerProviderEditView is_displayed to use ProviderEditView

brahmanim:
Fixing ContainerProviderEditView is_displayed method, to use ProviderEditView 
is_displayed instead of ProviderAddView one.

{{pytest: cfme/tests/containers/test_provider_configuration_menu.py -v --use-provider ocp-39-hawk}}

### DIFF
--- a/cfme/common/provider_views.py
+++ b/cfme/common/provider_views.py
@@ -608,7 +608,7 @@ class ContainerProviderEditView(ProviderEditView):
     """
     @property
     def is_displayed(self):
-        return (super(ProviderEditView, self).is_displayed and
+        return (super(ContainerProviderEditView, self).is_displayed and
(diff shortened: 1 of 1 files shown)


cfme/common/provider_views.py
```diff
@@ -608,7 +608,7 @@ class ContainerProviderEditView(ProviderEditView):
     """
     @property
     def is_displayed(self):
-        return (super(ProviderEditView, self).is_displayed and
+        return (ProviderEditView.is_displayed and
```
digitronik: Required:
I think better to go with the same format for all. You can use `ContainerProviderEditView` with `super` method only.

brahmanim: Thank you for the comment, I have update ContainerProviderEditView accordingly.

PR #7558, opened 2018-07-30
[1LP][RFR] Added 2 new volume backup tests

jadhaj:
Purpose or Intent
=================
Added 2 tests:
1. test_create_volume_backup
2. test_create_volume_incremental_backup

{{ pytest: cfme/tests/openstack/cloud/test_volume_backup.py --use-provider rhos11 }}

### DIFF
--- a/cfme/tests/openstack/cloud/test_volume_backup.py
+++ b/cfme/tests/openstack/cloud/test_volume_backup.py
@@ -1,0 +1,74 @@
+"""Tests for Openstack cloud volume Backups"""
+
+import fauxfactory
+import pytest
+
+from cfme.cloud.provider.openstack import OpenStackProvider
+from cfme.utils.log import logger
+
+pytestmark = [
+    pytest.mark.usefixtures("setup_provider"),
+    pytest.mark.provider([OpenStackProvider])
+]
+
+
+VOLUME_SIZE = 1
+
+
+@pytest.fixture(scope='function')
+def volume_backup(appliance, provider):
+    volume_collection = appliance.collections.volumes
+    storage_manager = '{} Cinder Manager'.format(provider.name)
+    backup_collection = appliance.collections.volume_backups.filter({'provider': provider})
+
+    # create new volume
+    volume = volume_collection.create(name=fauxfactory.gen_alpha(),
+                                      storage_manager=storage_manager,
+                                      tenant=provider.data['provisioning']['cloud_tenant'],
+                                      size=VOLUME_SIZE,
+                                      provider=provider)
+
+    # create new backup for crated volume
+    if volume.status == 'available':
+        backup_name = fauxfactory.gen_alpha()
+        volume.create_backup(backup_name)
+        volume_backup = backup_collection.instantiate(backup_name, provider)
+        yield volume_backup
+    else:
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit 5bdfb81f3cc2ed630e9ae79c561c719ec00c0799:

`cfme/tests/openstack/cloud/test_volume_backup.py`:
- :bangbang: Line 17:1: **E302** *expected 2 blank lines, found 1*

**No commit flaws detected**.

*CFME QE Bot*

digitronik:
Question:
why you are not improving `cfme/tests/storage/test_volume_backup.py` rather than committing new files? It's better to add in `storage` so, it will help to add `AWS` provider support in `parametrize` way.

jadhaj:
@digitronik 
We have already our cfme/tests/openstack/cloud/test_volumes.py while there is also  cfme/tests/storage/test_volume.py. I think it is mostly because theses tests forms our automation regression which do not include all possible tests either automated or not.

cfme/tests/openstack/cloud/test_volume_backup.py
```diff
@@ -0,0 +1,74 @@
...
+
+import fauxfactory
+import pytest
+
+from cfme.cloud.provider.openstack import OpenStackProvider
+from cfme.utils.log import logger
+
+pytestmark = [
+    pytest.mark.usefixtures("setup_provider_modscope"),
+    pytest.mark.provider([OpenStackProvider], scope='module')
```
sshveta: Required: provider should be function scope, remove scope='module'

sshveta:
Please check PRT results .

cfme/tests/openstack/cloud/test_volume_backup.py
jadhaj: Will remove it and resubmit. just can you provide an explanation on the difference because it  worked like that.

Regarding PRT results, according to @jawatts, looks like it is a problem in PRT system

jadhaj: BTW< i got below errors when remove scope:
peMismatch: You tried to access the 'function' scoped fixture 'provider' with a 'module' scoped request object, involved factories
cfme/fixtures/provider.py:281:  def setup_provider_modscope(request, provider)
../cfme_venv/lib/python2.7/site-packages/_pytest/fixtures.py:242:  def get_direct_param_fixture_func(request)
______________________ ERROR at setup of test_create_volume_incremental_backup[openstack-11] ______________________
ScopeMismatch: You tried to access the 'function' scoped fixture 'provider' with a 'module' scoped request object, invo

sshveta: I think you will have to use 
pytest.mark.usefixtures("setup_provider"),
pytest.mark.provider([OpenStackProvider]) 
both of them have to be function scoped I think .

Provider markers have to function scoped now because it takes care of adding/removing providers when necessary, and controlling number of providers per appliance .

jawatts: @sshveta is correct here, provider should always be marked function scoped. As for PRT, 5.9 should be working but 510 is still having issues.

PR #7563, opened 2018-07-30
[RFR] Added table object to ContainerObjectAllBaseView

tkhamis:
{{ pytest : cfme/tests/containers/test_tables_fields.py   --use-provider ocp-39-hawk }}

Added table object to ContainerObjectAllBaseView

### DIFF
--- a/cfme/containers/provider/__init__.py
+++ b/cfme/containers/provider/__init__.py
@@ -584,10 +584,6 @@ class ContainerObjectAllBaseView(ProvidersView):
         # We use 'in' for this condition since when we use search it'll include (Names with "...")
         return self.SUMMARY_TEXT in self.summary.text
 
-    @property
-    def table(self):
-        return self.entities.elements
-
 
 class ContainerObjectDetailsEntities(View):
     properties = SummaryTable(title="Properties")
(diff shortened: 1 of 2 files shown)


cfme/containers/provider/__init__.py
```diff
@@ -584,6 +584,10 @@ def is_displayed(self):
         # We use 'in' for this condition since when we use search it'll include (Names with "...")
         return self.SUMMARY_TEXT in self.summary.text
 
+    @property
+    def table(self):
+        return self.entities.elements
```
jawatts: entities is Conditional widget here and this property is dependent on a potential state of the condition. 
https://github.com/ManageIQ/integration_tests/blob/master/widgetastic_manageiq/__init__.py#L3248

We should not define a property in this case.

In this case, the test is what needs to be fixed. 

Example for getting the column rows on the provider all view:
```
In[29]: view.entities.elements.headers
Out[29]: (None, None, u'Name', u'Hostname', u'IP Address', u'Port', u'Type')
```

jawatts:
:+1: one comment, please rebase your commits.

tkhamis:
done

tkhamis:
@jawatts can I change it to (1LP)RFR?

jawatts:
@tkhamis only first level reviewers can change it to 1LP(I am not one), please move it back to RFR.

PR #7568, opened 2018-07-31
[1LP][RFR] Fix clicking checkboxes on HostDriftHistory view

rhrazdil:
Host drift analysis fails on:
E       assert False
E        +  where False = <cfme.common.host_views.HostDriftAnalysis object at 0x7f20545a7ad0>.is_displayed

The problem begins on the following line:
https://github.com/ManageIQ/integration_tests/blob/5f4466953e7b0007b7810fbec30018e65115be56/cfme/tests/infrastructure/test_host_drift_analysis.py#L115

This line is supposed to check first two rows representing latest 2 analysis. What actually happens is that no line is selected. The reason is that indexes 0 and 1 both point to the first row, so the first row is selected and deselected afterwards.

Now, at this point I changed the indexes from 0,1 to 1,2 so that first and second row is selected correctly.

Better way to deal with this may be fixing the selection logic instead.

{{ pytest: cfme/tests/infrastructure/test_host_drift_analysis.py cfme/tests/cloud_infra_common/test_vm_instance_analysis.py::test_drift_analysis }}

### DIFF
--- a/cfme/tests/infrastructure/test_host_drift_analysis.py
+++ b/cfme/tests/infrastructure/test_host_drift_analysis.py
@@ -109,7 +109,6 @@ def test_host_drift_analysis(appliance, request, a_host, soft_assert, set_host_c
         message="Waiting for Drift History count to increase",
         fail_func=appliance.server.browser.refresh
     )
-
     # check drift difference
     soft_assert(a_host.equal_drift_results(
         '{} (1)'.format(added_tag.category.display_name), 'My Company Tags', 0, 1),
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -883,7 +883,7 @@ class VM(BaseVM):
         Args:
             drift_section (str): Title text of the row to compare
             section (str): Accordion section where the change happened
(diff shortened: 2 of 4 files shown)


rhrazdil:
PRT fails on connection problems.

cfme/tests/infrastructure/test_host_drift_analysis.py
```diff
@@ -109,7 +109,6 @@ def test_host_drift_analysis(appliance, request, a_host, soft_assert, set_host_c
         message="Waiting for Drift History count to increase",
         fail_func=appliance.server.browser.refresh
     )
-
     # check drift difference
     soft_assert(a_host.equal_drift_results(
         '{} (1)'.format(added_tag.category.display_name), 'My Company Tags', 0, 1),
```
rlbabyuk: Question: Does it works for host?

rhrazdil: Ahh, no. I forgot to change it for host... I will add it in new PR.

PR #7571, opened 2018-07-31
[1LP][RFR] Update openstack/cloud/test_instances

mshriver:
Fix power control


mshriver:
Travis failure unrelated and fixed by #7572

PR #7573, opened 2018-07-31
[WIPTEST] Fix for issue when after_click unintentionally closes alert during provider removal in FF

izapolsk:
FF workaround in after_click closes alert as a result some code which handles alerts stopped working in 5.10 and FF > 45
REQUIRES: https://github.com/RedHatQE/widgetastic.core/pull/109/files


cfme/common/provider.py
```diff
@@ -543,7 +543,9 @@ def delete(self, cancel=True):
         if not cancel:
             msg = ('Delete initiated for 1 {} Provider from '
                    'the {} Database'.format(self.string_name, self.appliance.product_name))
-            view.flash.assert_success_message(msg)
+            main_view = self.create_view(navigator.get_class(self, 'All').VIEW)
```
mshriver: :tada:

cfme/utils/appliance/implementations/ui.py
```diff
@@ -214,13 +214,14 @@ def before_click(self, element, locator):
         # this is necessary in order to handle unexpected alerts like "Abandon Changes"
         self.browser.page_dirty = self.page_has_changes
 
-    def after_click(self, element, locator):
+    def after_click(self, element, locator, ignore_ajax=None):
```
mshriver: We must be careful about this, because the class method we're overriding does not take additional kwargs.

izapolsk: I updated class method in wt.core as well

mshriver: I see your REQUIRES line now! :bowing_man:

izapolsk:
with this fix I'm going down the rabbit hole. it causes new issues.
I'm going to close it and try to refactor after_click workaround

izapolsk:
I removed after click patch and fixed two root causes in WT.

requirements/frozen.py2.txt
```diff
@@ -305,8 +305,10 @@ wcwidth==0.1.7
 webencodings==0.5.1
 websocket-client==0.40.0
 Werkzeug==0.14.1
-widgetastic.core==0.21.6
-widgetastic.patternfly==0.0.37
+# widgetastic.core==0.21.6
+git+git://github.com/izapolsk/widgetastic.core@ff60_fix
+#widgetastic.patternfly==0.0.37
+git+git://github.com/izapolsk/widgetastic.patternfly@ff60_dropdown_fix
```
mshriver: Merged/released as 0.0.38

izapolsk:
all these changes and some other patches have been added to https://github.com/ManageIQ/integration_tests/pull/7771/files.
So, this PR isn't necessary any longer.

PR #7579, opened 2018-08-01
[1LP][RFR] Fixed TypeError while collecting hosts

ganeshhubale:
It fixes 
1. 'type error' 
2. 'HostCollection object is not iterable error'

{{ pytest: cfme/tests/infrastructure/test_tenant_quota.py::test_vm_migration_after_assigning_tenant_quota --use-provider complete --long-running }}

### DIFF
--- a/cfme/tests/infrastructure/test_tenant_quota.py
+++ b/cfme/tests/infrastructure/test_tenant_quota.py
@@ -107,7 +107,7 @@ def small_vm(provider, small_template_modscope):
 @pytest.fixture()
 def check_hosts(small_vm, provider):
     """Fixture to return host"""
-    if provider.hosts != 1:
+    if len(provider.hosts.all()) != 1:
         view = navigate_to(small_vm, 'Details')
         vm_host = view.entities.summary('Relationships').get_text_of('Host')
         hosts = [vds.name for vds in provider.hosts.all() if vds.name not in vm_host]
--- a/cfme/tests/infrastructure/test_tenant_quota.py
+++ b/cfme/tests/infrastructure/test_tenant_quota.py
@@ -253,7 +253,7 @@ def test_setting_child_quota_more_than_parent(tenants_setup, parent_quota, child
 
 
 @pytest.mark.long_running
-@pytest.mark.provider([VMwareProvider, RHEVMProvider], override=True, scope="module",
+@pytest.mark.provider([VMwareProvider], override=True, scope="module",
                       required_fields=[['templates', 'small_template']], selector=ONE_PER_TYPE)
 @pytest.mark.parametrize(
     ['set_roottenant_quota', 'custom_prov_data'],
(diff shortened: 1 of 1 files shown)


cfme/tests/infrastructure/test_tenant_quota.py
```diff
@@ -107,10 +107,10 @@ def small_vm(provider, small_template_modscope):
 @pytest.fixture()
 def check_hosts(small_vm, provider):
     """Fixture to return host"""
-    if len(provider.hosts) != 1:
+    if provider.hosts != 1:
```
rlbabyuk: Required: this will be always true as provider.hosts is hosts collection object
you need to use provider.hosts.all() this will return you a list all available hosts and len(provider.hosts.all()) will do the trick

cfme/tests/infrastructure/test_tenant_quota.py
```diff
@@ -253,7 +253,7 @@ def test_setting_child_quota_more_than_parent(tenants_setup, parent_quota, child
 
 
 @pytest.mark.long_running
-@pytest.mark.provider([VMwareProvider, RHEVMProvider], override=True, scope="module",
```
rlbabyuk: Question: Why did you remove RHEVMProvider provider?

cfme/tests/infrastructure/test_tenant_quota.py
```diff
@@ -253,7 +253,7 @@ def test_setting_child_quota_more_than_parent(tenants_setup, parent_quota, child
...
+@pytest.mark.provider([VMwareProvider], override=True, scope="module",
```
rlbabyuk: Request: we don't use scope="module" for providers, please remove

cfme/tests/infrastructure/test_tenant_quota.py
ganeshhubale: RHEVMProvider provider  is not providing multiple hosts. Because of this, vm migration gets failed.

cfme/tests/infrastructure/test_tenant_quota.py
ganeshhubale: Thanks

cfme/tests/infrastructure/test_tenant_quota.py
ganeshhubale: @rlbabyuk 
To fix this issue. If I removed **scope='module'** and changed it to **scope='function'** then I need some changes in following files too:
```
1. cfme/tests/infrastructure/test_tenant_quota.py:95:  def small_vm(provider, small_template_modscope) 
2. cfme/fixtures/provider.py:357:  def small_template_modscope(provider) 
```
In above files: need to change **scope='function'**. 
If it is not affecting other tests then we can go ahead with these changes.
please let me know if you have any suggestions for me.

rlbabyuk: @ganeshhubale I think in this case we can leave provider scope as module, not to change other fixtures

PR #7587, opened 2018-08-02
[1LP][RFR] Add password requirements test.

jaryn:
Purpose or Intent
=================
__Adding tests__ for password requirements to cover case RHCF3-14524

Shriver: Editing pytest line, azure isn't default provider, needs to be collected against.
{{ pytest: -v cfme/tests/cloud/test_providers.py::test_azure_instance_password_requirements --use-provider azure }}

### DIFF
--- a/cfme/tests/cloud/test_providers.py
+++ b/cfme/tests/cloud/test_providers.py
@@ -519,8 +519,10 @@ def test_azure_instance_password_requirements(appliance, request,
                "abcdefgh_1A"):
         view.form.customize.fill({"root_password": pw})
         view.form.submit_button.click()
-        wait_for(lambda: message in (m.text for m in view.flash.messages),
+        wait_for(lambda: message in (m.read() for m in view.flash.messages),
                  fail_condition=False, num_sec=10, delay=.1)
+        for m in view.flash.messages:
+            m.dismiss()
 
 
 @pytest.mark.tier(3)
--- a/cfme/tests/cloud/test_providers.py
+++ b/cfme/tests/cloud/test_providers.py
@@ -488,7 +488,8 @@ def test_select_key_pair_none_while_provisioning(appliance, request, has_no_clou
 
 @pytest.mark.tier(3)
 @pytest.mark.provider([AzureProvider], override=True)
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit 762e479af9b8d7f541e847ccfdaa33638f3f74de:

- :bangbang: Line 518:54: **E202** *whitespace before '}'*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/cloud/test_providers.py
```diff
@@ -484,6 +485,42 @@ def test_select_key_pair_none_while_provisioning(appliance, request, has_no_clou
...
+        "admin_username": "some_value",
+    })
+
+    for pw in ("abcdefghijkl_",
+               "ABCDEFGHIJKL_",
+               "ABCDEFGHIJKLa",
+               "abcdefgh_1A"):
+        view.form.customize.fill({"root_password": pw})
+        view.form.submit_button.click()
+        wait_for(lambda: message in (m.text for m in view.flash.messages),
```
rlbabyuk: Question: why we need to use wait_for here, as we can use soft_assert(view.flash.assert_messeges(message)) here to check all options?

jaryn: We need to wait for the messages to appear.

mshriver:
PRT 26229 false positive, nothing collected. Re-running with modified pytest line to include azure provider.

cfme/tests/cloud/test_providers.py
mshriver: @jaryn Are you intending to make the above changes to the test? Thoughts? Marking WIPTEST until I hear back :)

jaryn: I will change the code to use `m.read()` instead of accessing `m.text` if you think it is better. Although the result is equivalent, so I don't really see a reason.

`m.dismiss()` is not needed. The messages are dismissed upon IIRC `submit_button.click()` method call, but it feels better to dismiss explicitly.

I cannot simply use soft_assert as I need to wait for the message to show up.

jaryn:
@mshriver Please take a look on how am I dealing with the FlashMessages now.

PR #7588, opened 2018-08-02
[1LP][RFR] VM TagWise graph for host

digitronik:
Purpose or Intent
=================
{{pytest: cfme/tests/candu/test_host_graph.py::test_tagwise -v --use-provider complete}}

### DIFF
--- a/cfme/configure/configuration/region_settings.py
+++ b/cfme/configure/configuration/region_settings.py
@@ -12,7 +12,6 @@ from cfme.modeling.base import BaseCollection, BaseEntity
 from cfme.utils import conf
 from cfme.utils.appliance import Navigatable, NavigatableMixin
 from cfme.utils.appliance.implementations.ui import navigator, CFMENavigateStep, navigate_to
-from cfme.utils.blockers import BZ
 from cfme.utils.log import logger
 from cfme.utils.pretty import Pretty
 from cfme.utils.update import Updateable
@@ -220,7 +219,8 @@ class Category(Pretty, BaseEntity, Updateable):
         else:
             view.save_button.click()
 
-        view = self.create_view(navigator.get_class(self.parent, 'All').VIEW)
+        view = view.browser.create_view(CompanyCategoriesAllView)
+        view.wait_displayed(50)
         assert view.is_displayed
         view.flash.assert_no_error()
 
--- a/cfme/configure/configuration/region_settings.py
+++ b/cfme/configure/configuration/region_settings.py
@@ -135,8 +135,8 @@ class CompanyCategoriesAllView(RegionView):
     @property
     def is_displayed(self):
         return (
-            self.company_categories.is_active() and
-            self.table.is_displayed
+            self.tags.is_active() and
+            self.company_categories.is_active()
         )
 
 
@@ -206,7 +206,7 @@ class Category(Pretty, BaseEntity, Updateable):
                 cancel:
         """
         view = navigate_to(self, 'Edit')
-        view.fill({
+        change = view.fill({
             'name': updates.get('name'),
             'display_name': updates.get('display_name'),
             'long_description': updates.get('description'),
@@ -214,13 +214,13 @@ class Category(Pretty, BaseEntity, Updateable):
             'single_value': updates.get('single_value'),
             'capture_candu': updates.get('capture_candu'),
         })
-        if cancel:
+
+        if cancel or not change:
             view.cancel_button.click()
         else:
             view.save_button.click()
 
-        view = view.browser.create_view(CompanyCategoriesAllView)
-        view.wait_displayed(50)
+        view = self.create_view(navigator.get_class(self.parent, 'All').VIEW)
         assert view.is_displayed
         view.flash.assert_no_error()
 
--- a/cfme/fixtures/candu.py
+++ b/cfme/fixtures/candu.py
@@ -41,19 +41,20 @@ def collect_data(appliance, provider, interval='hourly', back='7.days'):
 
 
 @pytest.fixture(scope="module")
-def enable_candu_cat(appliance):
+def enable_candu_category(appliance):
     """Enable capture C&U Data for tag category location"""
     collection = appliance.collections.categories
-    location_cat = collection.instantiate(name="location", display_name="Location")
-    if not location_cat.capture_candu:
-        location_cat.update(updates={"capture_candu": True})
-    return location_cat
+    location_category = collection.instantiate(name="location", display_name="Location")
+    if not location_category.capture_candu:
+        location_category.update(updates={"capture_candu": True})
+    return location_category
 
 
 @pytest.fixture(scope="function")
-def candu_tag_vm(provider, enable_candu_cat):
+def candu_tag_vm(provider, enable_candu_category):
+    """Add location tag to VM"""
     collection = provider.appliance.provider_based_collection(provider)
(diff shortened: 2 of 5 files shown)


digitronik:
For this PR- https://github.com/ManageIQ/integration_tests/pull/7618 need to merge.
and
Need to fix BZ https://bugzilla.redhat.com/show_bug.cgi?id=1601915

digitronik:
@mshriver: can you please review it?

cfme/fixtures/candu.py
```diff
@@ -38,3 +38,26 @@ def collect_data(appliance, provider, interval='hourly', back='7.days'):
         vm.perf_capture({}, {}.ago.utc, Time.now.utc)\""
         .format(provider.id, repr(vm_name), repr(interval), back))
     return ret.success
+
+
+@pytest.fixture(scope="module")
+def enable_candu_category(appliance):
+    """Enable capture C&U Data for tag category location"""
```
nachandr: Could you please describe how to navigate to that page?

nachandr:
@digitronik , Just making sure I'm not missing something.Do we have automated tests for grouping C&U graphs for clusters by tags?

cfme/fixtures/candu.py
digitronik: @nachandr `location_category.update` will take care of it. `update` method will call to `category` Detail navigation.

cfme/tests/candu/test_host_graph.py
```diff
@@ -157,3 +160,81 @@ def refresh():
...
+def test_tagwise(provider, interval, graph_type, gp_by, candu_tag_vm, enable_candu):
+    """Test Host graphs group by VM tag for hourly and Daily
+
+    prerequisites:
+        * C&U enabled appliance
+        * C&U data collection enabled for Tag category
+        * VM should be taged with proper tag category
+
+    Steps:
+        * If possible capture historical data for host and vm
```
nachandr: Isn't this step a prerequisite?

digitronik: We can not take as `prerequisite`. prerequisite means necessary. It's optional to reduce runtime we are collecting historical data. Now, a runtime for those 45 test nearly 2.30 hr. If we remove it. we can not go for daily and hourly also it will take 30 min * provider+ other.

digitronik: Thats why I used `If possible`.

cfme/fixtures/candu.py
nachandr: REQUIRED   
Replace this comment Enable capture C&U Data for tag category location 

with this

Enable capture C&U Data for tag category location by navigating to the Configuration -> Region page.Click 'Tags' tab , select required company category under 'My Company Categories' and enable 'Capture C & U Data' for the category.

cfme/tests/candu/test_host_graph.py
nachandr: Just remove the 'If possible' from the sentence since historical data for host and vm needs to be captured.

digitronik:
@nachandr no, Its first initiative for tagged graph test. As per my knowledge, `host` allows `vm` tag.
The cluster will allow both `vm` and `host` tag. With reference of this PR anyone can automate for the cluster. :+1:

cfme/tests/candu/__init__.py
```diff
@@ -25,3 +25,61 @@ def compare_data(table_data, graph_data, legends, tolerance=1):
...
+    Note: Mainly, when we check graph for some tag the unit in table reading missing. The unit
+        conversion totally depends on manual observation.
+
+    BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1367560
+    """
+    for row in table_data:
+        for key, data in graph_data.items():
+            if any([re.match(key, item) for item in row["Date/Time"].split()]):
+                for leg in legends:
+                    table_item = re.split(" |%", row[leg].replace(",", ""))
```
mshriver: This needs in-line commenting for intent and processing steps

cfme/tests/candu/__init__.py
```diff
@@ -25,3 +25,61 @@ def compare_data(table_data, graph_data, legends, tolerance=1):
...
+                    tb_value, tb_unit = table_item if len(table_item) > 1 else (table_item[0], None)
+                    if table_item:
+                        graph_item = re.split(" |%", data[leg].replace(",", ""))
+                        gp_value, gp_unit = (
+                            graph_item if len(graph_item) > 1 else (graph_item[0], None)
+                        )
+
+                        if not tb_unit:
```
mshriver: I suggest you take a look at utilizing bitmath for this instead:

https://pypi.org/project/bitmath/1.0.2-3/

digitronik: @mshriver: I had experience with this lib; Its overall about `bits` conversions. I am not getting other usages of it in our `integration test` except this `workaround`. But, As its supporting `SI` and `NSI` both unit types. It's possible to play with `frequency` conversion also a little bit `str` parsing required.

I think it will be helpful for `storage` and other `utilization` scope; I am trying to play with `SI` unit of it with string parsing. If success, I will include this lib in requirements else we will move with a workaround.

digitronik: I tried with it but fails. I found this `bitmath` lib only supporting `bit` related operation not supporting other units like `MHz`. It will not provide a complete solution for this problem.
```
In [52]: x = bitmath.parse_string('1 GiB')
In [53]: x.MiB
Out[53]: MiB(1024.0)
In [54]: x = bitmath.parse_string('1 GB')
In [55]: x.MB
Out[55]: MB(1000.0)
In [56]: x = bitmath.parse_string('1 GHz')
---------------------------------------------------------------------------
ValueError 
```
Also, main issue will raise when there is no unit for table data. In this case we have again go with observation and conditions. 
Finally, I am on the decision that better go with the normal workaround.
@mshriver  what you think?

cfme/configure/configuration/region_settings.py
```diff
@@ -136,8 +135,8 @@ class CompanyCategoriesAllView(RegionView):
     @property
     def is_displayed(self):
         return (
-            self.company_categories.is_active() and
-            self.table.is_displayed
```
mshriver: Why removing this table check?

cfme/tests/candu/__init__.py
```diff
@@ -25,3 +25,72 @@ def compare_data(table_data, graph_data, legends, tolerance=1):
...
+    Note: Mainly, when we check graph for some tag the unit in table reading missing. The unit
+        conversion totally depends on manual observation.
+
+    BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1367560
+    """
+    for row in table_data:
+        for key, data in graph_data.items():
+            if any([re.match(key, item) for item in row["Date/Time"].split()]):
+                for leg in legends:
+                    # For table item data separating unit, removing `,` and removing "%" Unit
```
mshriver: :+1:  Thanks for adding comments, great!

cfme/configure/configuration/region_settings.py
digitronik: @mshriver It was an option change means removal of a `table is_displayed` accepted, but when I was checking `CompanyCategoriesAllView` I found the table is not a unique thing which we always want while writing `is_displayed` method so, I just replace unique thing means checking `company_categories` tag is active or not with table. I am assuring you it will effect on any test.

PR #7594, opened 2018-08-03
[1LP][RFR] Fixed test_class_crud

quarckster:
{{pytest: -v -k "test_class_crud or test_instance_crud or test_namespace_crud"}}


cfme/automate/explorer/namespace.py
```diff
@@ -26,6 +27,7 @@ def is_displayed(self):
             self.in_explorer and
             self.title.text == 'Automate Namespace "{}"'.format(self.context['object'].name) and
             self.datastore.is_opened and
+            BZ(1611969, forced_streams=['5.10']).blocks or
```
psav: **REQUIRED:** Is it prudent to put parentheses around the OR block to make the reader a little more careful as you have a chain of ANDs and an OR?

PR #7605, opened 2018-08-06
[RFR] remove bad json deserialization statements in sprout and etc

izapolsk:
trackerbot's issue with wrong parsing JSONField's data has been fixed. Trackerbot should provide custom data in correct format. So, old bad decode statements aren't necessary now and can be removed


mshriver:
Trackerbot code updated with correct serialization.

izapolsk:
tested clone template. works well.

PR #7610, opened 2018-08-07
[1LP][RFR] This PR will fix the scope of test and updated with Analysis profile create method.

sbulage:
PRT run:
{{ pytest: cfme/tests/infrastructure/test_ssa_vddk.py -vvvv --long-running --use-provider vsphere6-nested }}


mshriver:
@sbulage keep an eye on either #7614 or #7615  too, you might see those failures as well.

sbulage:
PRT is failing on specific tests which is due to VDDK and Vsphere version backward support.

PR #7611, opened 2018-08-08
[1LP][RFR]  Add new test for provider test resume workers

tkhamis:
{{ pytest : cfme/tests/containers/test_pause_resume_workers.py --use-provider ocp-39-hawk }}
Add new test for provider test resume workers.
This is an implementation for Polarion test cases '10790',  '10791'

### DIFF
--- a/cfme/tests/containers/test_pause_resume_workers.py
+++ b/cfme/tests/containers/test_pause_resume_workers.py
@@ -20,7 +20,6 @@ def check_ems_state_in_diagnostics(appliance, provider):
             return True
     except Exception:
         return False
-    return False
 
 
 def test_pause_and_resume_provider_workers(appliance, provider):
--- a/cfme/tests/containers/test_pause_resume_workers.py
+++ b/cfme/tests/containers/test_pause_resume_workers.py
@@ -36,7 +36,6 @@ def check_ems_state_in_diagnostics(appliance, provider):
             return True
         else:
             return False
-
     except Exception:
         return False
     return False
@@ -56,23 +55,15 @@ def test_pause_and_resume_provider_workers(appliance, provider):
         7. navigate to : User -> Configuration -> Diagnostics ->  Workers
         8. Validate the ems_ workers are started
     """
-
     view = navigate_to(provider, "Details")
-
     pause_option, resume_option = get_pause_resume_buttons(view)
-
     # pause the provider
     view.toolbar.configuration.item_select(pause_option, handle_alert=True)
-
     ems_worker_state = check_ems_state_in_diagnostics(appliance, provider)
-
     assert not ems_worker_state, "Diagnostics shows that workers are running after pause provider"
 
     view = navigate_to(provider, "Details")
-
     # resume the provider
     view.toolbar.configuration.item_select(resume_option, handle_alert=True)
-
     ems_worker_state = check_ems_state_in_diagnostics(appliance, provider)
-
     assert ems_worker_state, "Diagnostics shows that workers are not running after resume provider"
--- a/cfme/tests/containers/test_pause_resume_workers.py
+++ b/cfme/tests/containers/test_pause_resume_workers.py
@@ -1,0 +1,84 @@
+import pytest
+from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.containers.provider import ContainersProvider
+
+pytestmark = [
+    pytest.mark.usefixtures('setup_provider'),
+    pytest.mark.tier(1),
+    pytest.mark.provider([ContainersProvider], scope='function')
+]
+
+
+def get_pause_resume_buttons(view):
+    # ensure resume option is exist
+    resume_option = filter(lambda item: "resume" in item.lower(), view.toolbar.configuration.items)
+    assert resume_option, "Not resume provider option is available in the configuration menu"
+    resume_option = resume_option.pop()
+
+    # ensure pause option is exsit
+    pause_option = filter(lambda item: "pause" in item.lower(), view.toolbar.configuration.items)
+    assert pause_option, "Not pause provider option is available in the configuration menu"
+    pause_option = pause_option.pop()
+
+    return pause_option, resume_option
+
+
+@pytest.mark.polarion('10790', '10791')
+def test_pause_and_resume_provider_workers(appliance, provider):
+    """
+    Basic workers testing for pause and resume for a container provider
+    Tests steps:
+        1. Navigate to provider page
+        2. Pause the provider
+        3. navigate to : User -> Configuration -> Diagnostics ->  Workers
+        4. Validate the ems_ workers are not found
+        5. Navigate to provider page
+        6. Resume the provider
+        7. navigate to : User -> Configuration -> Diagnostics ->  Workers
+        8. Validate the ems_ workers are started
+    """
+
+    view = navigate_to(provider, "Details")
+
+    pause_option, resume_option = get_pause_resume_buttons(view)
+
+    # pause the provider
+    view.toolbar.configuration.item_select(pause_option, handle_alert=True)
+
+    ems_worker_state = check_ems_state_in_diagnostics(appliance, provider)
+
+    assert not ems_worker_state, "Diagnostics shows that workers are running after pause provider"
+
+    view = navigate_to(provider, "Details")
+
+    # resume the provider
+    view.toolbar.configuration.item_select(resume_option, handle_alert=True)
+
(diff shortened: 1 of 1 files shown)


cfme/tests/containers/test_pause_resume_workers.py
```diff
@@ -0,0 +1,84 @@
...
+
+        uri_queue_name = getattr(row, "URI / Queue Name")
+
+        assert name, "'Name' column not found in workers table"
+
+        assert uri_queue_name, "'URI / Queue Name' column not found in workers table"
+
+        if "Event Monitor for Provider" in name.text\
+                and provider.name in name.text \
+                and "ems_" in uri_queue_name.text:
```
sshveta: Required : Use brackets here instead of \ 

        if ("Event Monitor for Provider" in name.text
              and provider.name in name.text 
              and "ems_" in uri_queue_name.text) :

cfme/tests/containers/test_pause_resume_workers.py
```diff
@@ -0,0 +1,76 @@
...
+    ems_worker_is_running = False
+    for row in workers_view.workers_table.rows():
+        name = getattr(row, "Name")
+        uri_queue_name = getattr(row, "URI / Queue Name")
+        assert name, "'Name' column not found in workers table"
+        assert uri_queue_name, "'URI / Queue Name' column not found in workers table"
+        if ("Event Monitor for Provider" in name.text and
+                provider.name in name.text and
+                "ems_" in uri_queue_name.text):
+            ems_worker_is_running = True
```
mshriver: There's no need for this variable name and setting it before the loop for concern of referencing it before definition.

Just return the result of the three-statement boolean that you're using to set false.

cfme/tests/containers/test_pause_resume_workers.py
```diff
@@ -0,0 +1,76 @@
...
+
+    return pause_option, resume_option
+
+
+def check_ems_state_in_diagnostics(appliance, provider):
+    workers_view = navigate_to(appliance.collections.diagnostic_workers, 'AllDiagnosticWorkers')
+    workers_view.browser.refresh()
+    ems_worker_is_running = False
+    for row in workers_view.workers_table.rows():
+        name = getattr(row, "Name")
```
mshriver: `getattr()` raises exception, thus the assert below is moot since the exception will occur here, not set 'None'. A default must be passed to `getattr`.

cfme/tests/containers/test_pause_resume_workers.py
```diff
@@ -0,0 +1,84 @@
...
+    assert not ems_worker_state, "Diagnostics shows that workers are running after pause provider"
+
+    view = navigate_to(provider, "Details")
+
+    # resume the provider
+    view.toolbar.configuration.item_select(resume_option, handle_alert=True)
+
+    ems_worker_state = check_ems_state_in_diagnostics(appliance, provider)
+
+    assert ems_worker_state, "Diagnostics shows that workers are not running after resume provider"
```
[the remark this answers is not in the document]
mshriver: Its fine to break up the test a bit with some empty lines, but separating each with an empty line doesn't help readability

cfme/tests/containers/test_pause_resume_workers.py
```diff
@@ -0,0 +1,76 @@
...
+pytestmark = [
+    pytest.mark.usefixtures('setup_provider'),
+    pytest.mark.tier(1),
+    pytest.mark.provider([ContainersProvider], scope='function')
+]
+
+
+def get_pause_resume_buttons(view):
+    # ensure resume option is exists
+    resume_option = filter(lambda item: "resume" in item.lower(), view.toolbar.configuration.items)
```
mshriver: I don't like this implementation, for a few reasons - one being lambdas should only be used when necessary, and this isn't such a case. This can easily be done without using a lambda by not using the filter method.

Second, its possible there will be multiple configuration menu items with either 'resume' or 'pause' in them, its easy to imagine MIQ features which change the items in the menu.

The specific text that's expected for a given version of cloudforms should be used within the test in order to complete the action.

If this is commonly used, and we don't want to have multiple tests defining a `cfme.utils.version.VersionPicker` object for these items, then it should be defined along with the provider class as an attribute. Or there should be a convenience method added to make the selection (like `provider.pause_worker()` and `provider.resume_worker()`

This ultimately just creates a bit of an odd pattern, and won't catch the instance where multiple items match.

cfme/tests/containers/test_pause_resume_workers.py
```diff
@@ -0,0 +1,69 @@
+from cfme.containers.provider import ContainersProvider
+from cfme.utils.appliance.implementations.ui import navigate_to
+import pytest
```
mshriver: Imports need to be sorted, pytest is a pip installable package which should be above the local imports, with an empty line between them.  More details in the docs.

cfme/tests/containers/test_pause_resume_workers.py
```diff
@@ -0,0 +1,69 @@
...
+        for r in rows:
+            return True
+        else:
+            return False
+    except Exception:
+        return False
+    return False
+
+
+@pytest.mark.polarion('10790', '10791')
```
mshriver: We're removing these in #7634, so you can drop this line

cfme/tests/containers/test_pause_resume_workers.py
```diff
@@ -0,0 +1,69 @@
...
+    return pause_option, resume_option
+
+
+def check_ems_state_in_diagnostics(appliance, provider):
+    workers_view = navigate_to(appliance.collections.diagnostic_workers, 'AllDiagnosticWorkers')
+    workers_view.browser.refresh()
+    try:
+        rows = workers_view.workers_table.rows(
+            name='Event Monitor for Provider: {}'.format(provider.name))
+        for r in rows:
```
mshriver: This is a bit of an odd pattern - from the looks of it rows yields or returns `None` when no rows match. Thoughts on a pattern like:

```
try:
    rows = table.rows(name='filter')
except Exception:
    return False
else:
    return rows is not None
```

tkhamis: rows is always not None , since table.rows(name='filter') returns a generator and not a list.
I have to check that the generator does not return any elements.
I tried your suggestion while writing the code, and it didn't work, it will always return true.

sshveta: How about - 
        try:
            for r in workers_view.workers_table.rows(
                 name='Event Monitor for Provider: {}'.format(provider.name))
                 return True
        except Exception:
                return False

mshriver: Thanks for confirming @tkhamis, you're very right, its always a generator. I think we should take advantage of that fact, and using @sshveta example get a little more explicit with the exceptions we're handling here.

Little snippet from ipython:
```
# v is view of instances table (doesn't matter for example)
In [19]: v.entities.elements.rows().next().read()
Out[19]: 
{0: u'',
 1: u'',
 u'Allocated Size': u'170 GB',
 u'Architecture': u'64 bit',
 u'Availability Zone': <redacted>,
 u'Compliant': u'',
 u'Flavor': u'm1.small',
 u'IP Addresses': <redacted>,
 u'Last Analysis Time': u'',
 u'Load Balancer Status': u'',
 u'Name': u'cu-24x7',
 u'Provider': <redacted>,
 u'Region': u'Region 0'}

# Example of explicit pattern, using next() and StopIteration exception
In [20]: def check(name_match):
    ...:     try:
    ...:         v.entities.elements.rows(name=name_match).next()
    ...:         return True
    ...:     except StopIteration:
    ...:         # no matching rows
    ...:         return False
    ...:     

In [21]: check('cu-24x7')
Out[21]: True

In [22]: check('junk-name-fake')
Out[22]: False
```

cfme/tests/containers/test_pause_resume_workers.py
```diff
@@ -0,0 +1,49 @@
...
+def check_ems_state_in_diagnostics(appliance, provider):
+    workers_view = navigate_to(appliance.collections.diagnostic_workers, 'AllDiagnosticWorkers')
+    workers_view.browser.refresh()
+    try:
+        if workers_view.workers_table.rows(
+                name='Event Monitor for Provider: {}'.format(provider.name)).next():
+            return True
+    except Exception:
+        return False
+    return False
```
quarckster: **REQUIRED**
This `return` is never reached.

PR #7613, opened 2018-08-08
[1LP][RFR] Add Hypervisor hostname tests

jadhaj:
Purpose or Intent
=================
Added new tests: 
test_hypervisor_hostname
test_hypervisor_hostname_views

{{ pytest: cfme/tests/openstack/infrastructure/test_hosts.py::test_hypervisor_hostname cfme/tests/openstack/infrastructure/test_hosts.py::test_hypervisor_hostname_views }}

### DIFF
--- a/cfme/tests/openstack/infrastructure/test_hosts.py
+++ b/cfme/tests/openstack/infrastructure/test_hosts.py
@@ -9,6 +9,8 @@ pytestmark = [
     pytest.mark.provider([OpenstackInfraProvider], scope='module'),
 ]
 
+VIEWS = ["List View", "Tile View"]
+
 
 @pytest.fixture(scope="module")
 def host_collection(appliance):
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit 8f95be53b27a4f8118a1818106c90bfe9af02e66:

`cfme/tests/openstack/infrastructure/test_hosts.py`:
- :heavy_exclamation_mark: Line 142:1: **W391** *blank line at end of file*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/openstack/infrastructure/test_hosts.py
```diff
@@ -112,3 +112,30 @@ def test_host_zones_assigned(host_collection, provider):
         view = navigate_to(host, 'Details')
         result = view.entities.summary('Relationships').get_text_of('Availability Zone')
         assert result, "Availability zone doesn't specified"
+
+
+def test_hypervisor_hostname(host_collection, appliance, soft_assert):
+    hvisors = {
+        hv.ipaddress: hv.hypervisor_hostname for hv in appliance.rest_api.collections.hosts.all}
+
```
mshriver: I think it would be more useful to test the hypervisor hostnames in CFME against those on the provider itself. Can these be pulled through wrapanapi instead of from CFME rest API?  All we're comparing here is the IP/hostname pairs in the rest API vs the IP/hostname pairs in the UI.

jadhaj: I am not sure i am familiar with wrapanapi. You mean maybe using provider.mgmt.api.hosts.list() or  provider.mgmt.iapi.node.list()

mshriver: wrapanapi is the library that wraps all provider API's to create a consistent API directly to the provider under test for use within the framework.

`provider.mgmt` is a wrapanapi `System` object, and will have methods to retrieve different types of objects from the provider.

In the case of openstack, there is the `OpenstackSystem` type, and the `OpenstackInfraSystem` type. Assuming you're working with openstack infra here, there is a `provider.mgmt.list_hosts()` method available:
https://github.com/ManageIQ/wrapanapi/blob/master/wrapanapi/systems/openstack_infra.py#L127

This will return a list of node names, there's also a `list_nodes` method which will return more useful Node objects.

jadhaj: i think there also another issue here. Let's assume above methods are working, they don't return another unique value like ip address besides node name.
this unique value is needed to build the dictionary which is used for comparison afterwards.

Using zip to iterate through both list will not help also because there is no guarantee of lists order

your advise

mshriver: I see what you mean, it returns Node namedTuple objects that have name and UUID, if the UUID can't be matched to the UI, we should look at updating wrapanapi to return a more rich object with more attributes from the `i_node` API objects.

jadhaj: After you have fixed authentication issue, i have changed tests based on that. instead of building a hash, i assert against a list of nodes returned by wrapanapi mgmt list_nodes method

please let me know if that is OK

PR #7614, opened 2018-08-08
[WIPTEST] fixing typo in test_vm_instance_analysis

izapolsk:

PR #7592, opened 2018-08-02
This PR will fix host creds issue.

sbulage:
PRT run:
{{ pytest: cfme/tests/cloud_infra_common/test_vm_instance_analysis.py -k test_ssa_vm -vvvv --long-running --use-provider vsphere6-nested }}

PR #7615, opened 2018-08-08
[RFR] AnalysisProfile creation in SSA

mshriver:
#7614 is hitting this too, but I think more than just the init call needs to be changed.

# Note to reviewers
PRT results aren't clean - many tests are skipped because of host credentials setting.

At least one testcase gets to the point where the fixture structure has executed, and this block of code to create the analysis profile was executed without error.


sbulage:
@mshriver I already did modification in PR 7592 can you please take look in it ..?

mshriver:
@sbulage thanks, didn't come up in my search.

So, I'll close this, but I think you need to reconsider your modification in #7592 because it now creates, checks if it exists, and deletes, which is not correct.

sbulage:
@mshriver thanks :) I will update it accordingly

PR #7616, opened 2018-08-08
Add secgroup port

jaryn:
Purpose or Intent
=================

__Fixing__  test_sdn_nsg_firewall_rules. It is currently not doing network provider refresh which is incorrect. It should be refreshing to get new port rules.
__Updating tests__ to new wrapanapi changes.


dajoRH:
Lint report for commit 6d399700642810b42d6f08cc0c9013f7cc6256f6:

`cfme/networks/security_group.py`:
- :red_circle: Line 0:0: **EPY3** *Usage of .iteritems*

`cfme/tests/networks/test_sdn_inventory_collection.py`:
- :red_circle: Line 134:5: **F821** *undefined name 'wait_for'*
- :bangbang: Line 140:13: **E126** *continuation line over-indented for hanging indent*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
Lint report for commit 7d672c7f2076c317930608b8a75bef4381c9943c:

`cfme/networks/security_group.py`:
- :red_circle: Line 0:0: **EPY3** *Usage of .iteritems*

**No commit flaws detected**.

*CFME QE Bot*

cfme/networks/provider/__init__.py
```diff
@@ -301,6 +301,9 @@ def all(self):
 
         return network_providers
 
+    def instantiate(self, prov_class, *args, **kwargs):
```
digitronik: Required:
Remove this; we already have `instantiate` method
https://github.com/ManageIQ/integration_tests/blob/master/cfme/networks/provider/__init__.py#L323

cfme/networks/security_group.py
```diff
@@ -54,7 +54,16 @@ def all(self):
         else:
             view = navigate_to(self, 'All')
         list_networks_obj = view.entities.get_all(surf_pages=True)
-        return [self.instantiate(name=s.name) for s in list_networks_obj]
+
+        collection = (self.instantiate(name=s.name) for s in list_networks_obj)
```
digitronik: Optional:
Instate of using `collection` I would use other names like `security_groups`.

cfme/networks/security_group.py
```diff
@@ -54,7 +54,16 @@ def all(self):
...
+        return list(a for a in collection if apply_filters(a, self.filters))
```
digitronik: Question:
If you are planning to return `list`; why you are not using directly `list comprehension`?

cfme/networks/security_group.py
```diff
@@ -54,7 +54,16 @@ def all(self):
...
+
+
+def apply_filters(item, filters):
```
digitronik: Required:
Please move it to `collection class` and restrict the users from direct use by changing the method name to  `_apply_filters()`

cfme/networks/security_group.py
```diff
@@ -54,7 +54,16 @@ def all(self):
...
+    for attr_name, val in filters.items():
+        if getattr(item, attr_name) != val:
```
digitronik: Question:
I found `SecurityGroup` main attribute is `name`. You are also trying to filter it with a name. 
Why are you trying to implement this separate method?
Is security group allow to create duplicate names?

jaryn: 1) I do not know what is meant by main attribute.

2) `apply_filters` is written so it is general -- can be used with any object, therefore it doesn't belong to any specific class. Maybe it could be in some base class though. Best would be to have it in some file and import it when needed.

dajoRH:
Lint report for commit 52f355836058025dc9c488373534223882cafe12:

`cfme/networks/provider/__init__.py`:
- :bangbang: Line 305:5: **E303** *too many blank lines (2)*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/networks/test_sdn_inventory_collection.py
```diff
@@ -130,14 +131,16 @@ def test_sdn_nsg_firewall_rules(provider, appliance, secgroup_with_rule):
...
     network_provider = collection.all()[0]
+    network_provider.refresh_provider_relationships()
+    wait_for(network_provider.is_refreshed, func_kwargs=dict(refresh_delta=10), timeout=600)
     view = navigate_to(network_provider, 'Details')
     parent_name = view.entities.relationships.get_text_of("Parent Cloud Provider")
     assert parent_name == provider.name
 
-    # Navigate to secgroup.
-    collection = appliance.collections.network_security_groups.filter({'name': secgroup_with_rule})
+    collection = appliance.collections.network_security_groups.filter(
```
digitronik: Required:
same `secgroup_collection`

cfme/tests/networks/test_sdn_inventory_collection.py
```diff
@@ -130,14 +131,16 @@ def test_sdn_nsg_firewall_rules(provider, appliance, secgroup_with_rule):
...
+        {'name': secgroup_with_rule})
     secgroup = collection.all()[0]
     view = navigate_to(secgroup, 'Details')
 
-    assert 'Port Range' == view.entities.firewall_rules[0][3].text
-    assert '22' == view.entities.firewall_rules[1][3].text
+    assert 'Port' == view.entities.firewall_rules[1][3].text
+    assert '22' == view.entities.firewall_rules[2][3].text
```
digitronik: Required:
assert int(view.entities.firewall_rules[2][3].text) == 22

cfme/networks/security_group.py
```diff
@@ -54,7 +54,16 @@ def all(self):
         else:
             view = navigate_to(self, 'All')
         list_networks_obj = view.entities.get_all(surf_pages=True)
-        return [self.instantiate(name=s.name) for s in list_networks_obj]
+
+        collection = (self.instantiate(name=s.name) for s in list_networks_obj)
+        return [a for a in collection if apply_filters(a, self.filters)]
```
digitronik: Required:
I am suspect It will not return security group objects when I will call without a filter.

cfme/tests/networks/test_sdn_inventory_collection.py
jaryn: Your suggestion does exactly the same as my version, but it is a bit more complex .

dajoRH:
Lint report for commit df1e76d0dd9276a63692e7d5b27df45c57a9ad4d:

`cfme/networks/provider/__init__.py`:
- :bangbang: Line 305:5: **E303** *too many blank lines (2)*

**No commit flaws detected**.

*CFME QE Bot*

cfme/networks/security_group.py
jaryn: You are not right:
```
In [4]: def apply_filters(item, filters):
   ...:     for attr_name, val in filters.items():
   ...:         if getattr(item, attr_name) != val:
   ...:             return False
   ...:     return True
   ...: 

In [5]: class A(object):
   ...:     def __init__(self, foo):
   ...:         self.foo = foo
   ...:         

In [7]: collection = [A(str(i)) for i in range(10)]

In [13]: [a.foo for a in collection if apply_filters(a, {})]
Out[13]: ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9']

In [14]: [a.foo for a in collection if apply_filters(a, {'foo': '3'})]
Out[14]: ['3']
```

cfme/tests/networks/test_sdn_inventory_collection.py
digitronik: I think; Numerical thing should not compare as a string.

jaryn: @digitronik `'22'` is string. `'22' == '22'` is perfectly ok. Doing `int('22') == 22` is just adding additional step

digitronik: :P, @jaryn its simple to understand and the suggestion is for good practice. `firewall rule port` is number like port `8080`. When we assert such things we should have to convert `string` to mathematical quantity before assert. The thumb rule is that always do operations in the same domain what it represents. 
`"22"` represents a specific ASCII symbol (also known as a "character")
`22` represent a mathematical number.

In this case its ok but if we read `"22"` l.e. `string`. It's most likely the mathematical quantity, but it could have different meanings depending on what other characters are nearby. like "jaryn22" :). You are free to choose what you like...

cfme/tests/networks/test_sdn_inventory_collection.py
```diff
@@ -125,16 +129,18 @@ def test_sdn_nsg_firewall_rules(provider, appliance, secgroup_with_rule):
...
-    collection = appliance.collections.network_security_groups.filter({'name': secgroup_with_rule})
-    secgroup = collection.all()[0]
+    secgrp_collection = appliance.collections.network_security_groups.filter(
+        {'name': secgroup_with_rule})
+    secgroup = secgrp_collection.all()[0]
     view = navigate_to(secgroup, 'Details')
 
-    assert 'Port' == view.entities.firewall_rules[0][3].text
-    assert '22' == view.entities.firewall_rules[1][3].text
+    assert 'Port' == view.entities.firewall_rules[1][3].text
```
mshriver: Why the change here to which row is indexed? There needs to be an in-line comment explaining these 'magic numbers' at the minimum, so we're not guessing at what row _should_ have been there.

jaryn:
@mshriver @digitronik The apply_filters method was renamed to match_items and is being added in #8018.

cfme/tests/networks/test_sdn_inventory_collection.py
```diff
@@ -125,16 +129,25 @@ def test_sdn_nsg_firewall_rules(provider, appliance, secgroup_with_rule):
...
-    assert 'Port' == view.entities.firewall_rules[0][3].text
-    assert '22' == view.entities.firewall_rules[1][3].text
+    if appliance.version < '5.10':
+        # The table has one header row. The first non-header row has column
+        # names.
+        assert 'Port' == view.entities.firewall_rules[1][3].text
+        assert '22' == view.entities.firewall_rules[2][3].text
+    else:
+        # The table has two header rows. We cannot access the second one with
+        # widgetastic. So let's hope the column of index 3 is the Port Range
```
mshriver: Let's get an issue written against wt.core for handling two header tables

PR #7617, opened 2018-08-08
[1LP][RFR] Fixed test_provisioning_dialog_crud

quarckster:
{{pytest: cfme/tests/automate/test_provisioning_dialogs.py::test_provisioning_dialog_crud -v}}

### DIFF
--- a/cfme/tests/automate/test_provisioning_dialogs.py
+++ b/cfme/tests/automate/test_provisioning_dialogs.py
@@ -27,7 +27,7 @@ def test_provisioning_dialog_crud(appliance):
     assert dialog.exists
 
     with update(dialog):
-        dialog.diag_type = collection.HOST_PROVISION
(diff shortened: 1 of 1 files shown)


cfme/tests/automate/test_provisioning_dialogs.py
```diff
@@ -27,7 +27,7 @@ def test_provisioning_dialog_crud(appliance):
     assert dialog.exists
 
     with update(dialog):
-        dialog.diag_type = collection.HOST_PROVISION
+        dialog.diag_type = collection.VM_MIGRATE
```
sshveta: Question : Can you pls comment why is this change needed ? 
Host_provision is still an option for provisioning dialogs.

quarckster: It seems it's not anymore in CFME 5.10

PR #7602, opened 2018-08-06
[1LP][RFR] Cover provision cases with vlan profiles <No Profile> and <Use template nics>

rhrazdil:
Extending test_provision_dialog to include test cases that provision VM with vnic profiles 'No Profile' and 'Use template nics'. 

For RFE page refer to https://bugzilla.redhat.com/show_bug.cgi?id=1449157, comment #4.

{{pytest: cfme/tests/infrastructure/test_provisioning_dialog.py::test_provisioning_vnic_profiles --long-running}}

PR #7619, opened 2018-08-08
[1LP][RFR] Expand provision test to check more vlan options for RHV provider

istein1:
Verified:
Test pass, but sometimes there is a problem with the cleanup- VM removal.
This is handled by rhrazdil PR ManageIQ/wrapanapi#329

A RHV only addition:
Expanding the provision rest testing, to include RHV vlan added options:
 <Empty> & Profile name (network name). 

This should cover customer Bug 1574351 - prov.set_vlan() method didn't set the vnicprofiles identifier.

VM provision Rest api, network options, for RHV provider:
1) "vlan": "3a8dce01-dd59-4a46-8a6f-823acccda79f" (specific vnic profile ID) 
    - Not covered yet. 
2) "vlan": <Empty> (no vNIC profile) 
    - Covered in this PR, by 'test_provision_vlan'
3) "vlan": <Template> (use template's default vNIC profile) 
   - Already covered in existing code, by 'test_provision' 
4) "vlan": "profile_name(network_name)" (Specific vnic profile&network, for example ovirtmgmt (ovirtmgmt)) 
   - Covered in this PR, by 'test_provision_vlan'

### DIFF
--- a/cfme/tests/infrastructure/test_provisioning_rest.py
+++ b/cfme/tests/infrastructure/test_provisioning_rest.py
@@ -111,6 +111,60 @@ def test_provision(request, appliance, provider, provision_data):
     assert found_vms, 'VM `{}` not found'.format(vm_name)
 
 
+@pytest.mark.rhv2
+@pytest.mark.uncollectif(lambda provider: not provider.one_of(RHEVMProvider),
+                         reason='These profile names are RHV specific.')
+@pytest.mark.parametrize('vnic_profile', ['empty_vnic_profile', 'specific_vnic_profile'])
+def test_provision_vlan(request, appliance, provision_data, vnic_profile, provider):
+    """Tests provision via REST API for vlan Empty/Specific vNic profile.
+    Prerequisities:
+        * Have a provider set up with templates suitable for provisioning.
+    Steps:
+        * POST /api/provision_requests (method ``create``) the JSON with provisioning data. The
+            request is returned.
+        * Apart from the usual provisioning settings, set vlan
+          to values <Empty>/profile_name (network_name)
+        * Query the request by its id until the state turns to ``finished`` or ``provisioned``.
+        * Check the VM's vNic profile is as expected.
+    Metadata:
+        test_flag: rest, provision
+    """
+    vm_name = provision_data['vm_fields']['vm_name']
+    profile_name = provider.data["provisioning"]["vlan"]
+    if vnic_profile == 'empty_vnic_profile':
+        vnic_profile = '<Empty>'
+    # For 'specific_vnic_profile'
+    else:
+        vnic_profile = '{0} ({0})'.format(profile_name)
+    provision_data['vm_fields']['vlan'] = vnic_profile
+    request.addfinalizer(lambda: clean_vm(appliance, provider, vm_name))
+    appliance.rest_api.collections.provision_requests.action.create(**provision_data)
+    assert_response(appliance)
+    provision_request = appliance.collections.requests.instantiate(description=vm_name,
+                                                                   partial_check=True)
+    provision_request.wait_for_request()
+    msg = "Provisioning failed with the message {}".format(
+        provision_request.rest.message)
+    assert provision_request.is_succeeded(), msg
+    found_vms = appliance.rest_api.collections.vms.find_by(name=vm_name)
+    assert found_vms, 'VM `{}` not found'.format(vm_name)
+    # Check the VM vNIC
+    vm = appliance.collections.infra_vms.instantiate(vm_name, provider)
+    nics = vm.mgmt.get_nics()
+    assert nics, 'The VM should have a NIC attached.'
+    # Check the vNIC network profile
+    profile = nics[0].vnic_profile
+    if vnic_profile == '<Empty>':
+        assert not profile, 'vNIC profile is not Empty as expected.'
+    # For 'specific_vnic_profile'
+    else:
+        vnic_srv = provider.mgmt.api.system_service().vnic_profiles_service()
+        profile_via_provider = vnic_srv.profile_service(profile.id).get()
+        assert profile_via_provider.name == profile_name, \
+            "The vNIC profile name is {}, but should be {}".\
+            format(profile_via_provider.name, profile_name)
+
+
 @pytest.mark.rhv3
 @pytest.mark.meta(server_roles="+notifier")
 def test_provision_emails(request, provision_data, provider, appliance, smtp_test):


cfme/tests/infrastructure/test_provisioning_rest.py
```diff
@@ -111,6 +111,43 @@ def test_provision(request, appliance, provider, provision_data):
...
+        * Query the request by its id until the state turns to ``finished`` or ``provisioned``.
+    Metadata:
+        test_flag: rest, provision
+    """
+    vm_name = provision_data['vm_fields']['vm_name']
+    if vnic_profile == 'no_vnic_profile':
+        vnic_profile = '<Empty>'
+    elif vnic_profile == 'profile_name(network_name)':
+        vlan_data = provider.data["provisioning"]["vlan"]
+        vnic_profile = vlan_data + ' (' + vlan_data + ')'
```
rhrazdil: Can vnic_profile contain something different that 'no_vnic_profile' or 'profile_name(network_name)'? If yes, what happens then?

rhrazdil: I'm thinking this might look a little bit more readable: vnic_profile =  '{0} ({0})'.format(vlan_data)

cfme/tests/infrastructure/test_provisioning_rest.py
```diff
@@ -111,6 +111,43 @@ def test_provision(request, appliance, provider, provision_data):
     assert found_vms, 'VM `{}` not found'.format(vm_name)
 
 
+@pytest.mark.rhv2
+@pytest.mark.uncollectif(lambda provider: not provider.one_of(RHEVMProvider),
+                         reason='These profile names are RHV specific.')
+@pytest.mark.parametrize('vnic_profile', ['no_vnic_profile', 'profile_name(network_name)'])
```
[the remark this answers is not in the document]
istein1: Fixed.

cfme/tests/infrastructure/test_provisioning_rest.py
istein1: Fixed

istein1: Nope. It is a test for these 2 parameters, if more parameters will be added later on, the if statement will be adjusted.

istein1:
Verified the added tests pass, and that the created VMs are cleaned up at the end.

cfme/tests/infrastructure/test_provisioning_rest.py
```diff
@@ -111,6 +111,43 @@ def test_provision(request, appliance, provider, provision_data):
...
+    appliance.rest_api.collections.provision_requests.action.create(**provision_data)
+    assert_response(appliance)
+    provision_request = appliance.collections.requests.instantiate(description=vm_name,
+                                                                   partial_check=True)
+    provision_request.wait_for_request()
+    msg = "Provisioning failed with the message {}".format(
+        provision_request.rest.message)
+    assert provision_request.is_succeeded(), msg
+    found_vms = appliance.rest_api.collections.vms.find_by(name=vm_name)
+    assert found_vms, 'VM `{}` not found'.format(vm_name)
```
jan-zmeskal: Here we are checking if the VM has been provisioned. While that's definitely an important check, I believe check of the new VM's actual NICs is in place here, since this test case is focused on this area. You could do something like this:
`vm = appliance.collections.infra_vms.instantiate(vm_name, provider)`
and then you have this at your disposal:
`vm.mgmt.get_nics()`
By these means you can check that the VM actually (or does not have) the expcted NIC configuration.

I am saying this because I was running this specific test case with `vnic_profile == 'no_vnic_profile'` and the provisioned VM had a NIC like [this](https://up1-vrutkovs.int.open.paas.redhat.com/#rFIK31aKn14oSAw-7Xo9_Q) even though that the used template has not had any NICs. This might be a product bug. But regardless of whether this is a bug or not, I believe it would be good to have such a check here.

rhrazdil: @jan-zmeskal I think the problem here is in naming. The `vnic_profile == 'no_vnic_profile'` you mention actually refers to Empty profile (on line 133). Although this profile is called Empty (which I believe means no QoS limitation or qota), it is a profile. Hence there is a NIC. If you create a VM from a template and do not select any profile (not even the Empty one), the created VM won't have a NIC. This is a desired behaviour AFAIK.

So I would suggest renaming the `no_vnic_profile` to `empty_vnic_profile`. Additionally, `no_vnic_profile` may be tested by something  like: `vnic_profile = None` or `vnic_profile = '<Use template nics>'` (where both options are effectively the same, so I would prefer the second one for better readability).

jan-zmeskal: @rhrazdil Thank you very much for clarification. So it means that the behaviour I observed is not a product bug, which is good. However I still think my main point is valid - IMO it would be reasonable to check presence/absence of the desired NIC(s) via wrapanapi in the final part of the test.

rhrazdil: @jan-zmeskal Yes, I agree. By the way, I played around a little with provisioning with different vnic profile configurations and I discovered that if you select in CFME provisioning dialog option 'No profile', the provisioned VM does have a NIC with '<Empty>' profile. I'll file a bug for this naming discrepancy because I think I'm not the only one who may find it confusing.

jan-zmeskal: @rhrazdil You definitely are not - I got lost in this as well.

istein1: * Regarding the "<Empty>" option:
This is from https://bugzilla.redhat.com/show_bug.cgi?id=1449157 doc text:

3. Provision a virtual machine with one vNic where the vNic has no profile.

"vm_fields" : {
    ...
    "vlan" : "<Empty>"
  }

The history was that in the UI we had option "<Empty>" added, and the corresponding phrase in REST api was also "Empty".
There were complains on the misleading naming in the UI,  so the UI changed from <Empty> into <No Profile>:
https://github.com/ManageIQ/manageiq-providers-ovirt/pull/151

However in REST api, the "<Empty>" still remained.

* Regarding the test name: I shall rename the test name from "no_vnic_profile" into "empty_vnic_profile"

* Regarding adding the test a check of the actual vnic profile set - I'll need to work on this further - so moving the PR back to [WIP].

dajoRH:
Lint report for commit 9338c3a914016aef38c0bfa4d636e2b09944ad06:

`cfme/tests/infrastructure/test_provisioning_rest.py`:
- :bangbang: Line 165:1: **E303** *too many blank lines (3)*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/infrastructure/test_provisioning_rest.py
```diff
@@ -111,6 +111,56 @@ def test_provision(request, appliance, provider, provision_data):
...
+    nics = vm.mgmt.get_nics()
+    assert nics, 'The VM should have a NIC attached.'
+    # Check the vNIC network profile
+    profile = nics[0].vnic_profile
+    if vnic_profile == '<Empty>':
+        assert not profile, 'The vNIC profile should be empty.'
+    else:
+        vlan_data = provider.data["provisioning"]["vlan"]
+        vlan_data_formatted = '{0}/{0}'.format(vlan_data)
+        assert profile != vlan_data_formatted, \
```
jan-zmeskal: @istein1 I don't understand this comparison here. Should not we compare that profile **is** what we expect it to be?

jan-zmeskal: @istein1 Yes, I understand that, but why is the comparison operator `!=`? Shouldn't it be `==`? Especially given the message that gets printed when `assert` fails:
`"The vNIC profile should be {}".format(vlan_data_formatted)`

When we say "should be", we should probably check for equality (`==`), right?

jan-zmeskal: Hi @istein1, I played around for a big and discover two interesting things. Check this ipython output (I obtained it by placing breakpoint after line 155:
```python
In [74]: type(profile)
Out[74]: ovirtsdk4.types.VnicProfile

In [75]: profile.name

In [76]: profile.id
Out[76]: '5a0a3b4e-7a12-440d-9f7a-b367ec9d8b53'

In [77]: profile_via_provider = provider.mgmt.api.system_service().vnic_profiles_service().profile_service(profile.id).get()

In [78]: type(profile_via_provider)
Out[78]: ovirtsdk4.types.VnicProfile

In [79]: profile_via_provider.name
Out[79]: 'ovirtmgmt'

In [80]: assert profile_via_provider.name == provider.data['provisioning']['vlan']

In [81]:
```
Commentary: In step 76, we can see that `profile` that you obtained returns its ID. That's good, however in step 75 `profile` does not return its name. I think this might be an oVirt API bug, see further. In step 77, I obtain the very same profile but this time via `provider.mgmt`. In step 78, you can see that it is the very same type - `ovirtsdk4.types.VnicProfile`. However, this time it returns its name (step 79). Such obtained value can be used for comparison as I illustrated in step 80. It works but it's not very pretty, therefore I recommend researching whether this is indeed oVirtAPI bug and if yes, reporting it.

dajoRH:
Lint report for commit d803b4fded9dc59b72449bfdddf6a084fefb6089:

`cfme/tests/infrastructure/test_provisioning_rest.py`:
- :bangbang: Line 161:9: **E122** *continuation line missing indentation or outdented*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/infrastructure/test_provisioning_rest.py
```diff
@@ -111,6 +111,58 @@ def test_provision(request, appliance, provider, provision_data):
...
+    # Check the VM vNIC
+    vm = appliance.collections.infra_vms.instantiate(vm_name, provider)
+    nics = vm.mgmt.get_nics()
+    assert nics, 'The VM should have a NIC attached.'
+    # Check the vNIC network profile
+    profile = nics[0].vnic_profile
+    if vnic_profile == '<Empty>':
+        assert not profile, 'The vNIC profile should be empty.'
+    else:
+        profile_via_provider = provider.mgmt.api.system_service().\
```
izapolsk: REQUIRED: please use () instead of back slash or reformat code.

izapolsk:
by the way, has that wrapanapi PR been released and added to integration tests requirements ?

rhrazdil:
@izapolsk If you mean the PR for cleaning VMs that was released and added here https://github.com/ManageIQ/integration_tests/pull/7602

cfme/tests/infrastructure/test_provisioning_rest.py
```diff
@@ -111,6 +111,58 @@ def test_provision(request, appliance, provider, provision_data):
...
+    found_vms = appliance.rest_api.collections.vms.find_by(name=vm_name)
+    assert found_vms, 'VM `{}` not found'.format(vm_name)
+    # Check the VM vNIC
+    vm = appliance.collections.infra_vms.instantiate(vm_name, provider)
+    nics = vm.mgmt.get_nics()
+    assert nics, 'The VM should have a NIC attached.'
+    # Check the vNIC network profile
+    profile = nics[0].vnic_profile
+    if vnic_profile == '<Empty>':
+        assert not profile, 'The vNIC profile should be empty.'
```
jan-zmeskal: Here I would also print out what **is** the actual value of `profile` if `assert not profile` fails. The same goes for assertion on line 163. We should not only see the **expected** value, but also the obtained one.

cfme/tests/infrastructure/test_provisioning_rest.py
istein1: Fixed. Thanks.

cfme/tests/infrastructure/test_provisioning_rest.py
istein1: Thanks Jan. 
For the vnic profile Empty assert - if the vnic profile is not empty, it will complicate the code, if I'll fetch here the profile name. 
Also I am not sure that in this specific case it is important to know what is the profile name. What's important is that the profile is not empty, as expected.

For the specific vnic profile assert, I've added a print of the actual vnic profile name, in case it is not the expected one. Please check the updated code.

jan-zmeskal: Hi Ilanit, in python, it is okay to print a variable even if you expect it to be `None`. What will happen is that `None` will literally get converted to string "None". But it's okay, not a big deal. I propose to move forward with the PR.

PR #7620, opened 2018-08-08
[RFR] generic shepherd fix for openshift

izapolsk:
It turned out that last changes made generic shepherd choose wrong template for app name generation. therefore, wrong appliance name was generated for ocp appliances. 
This commit should fix such issue.

This change needs some local testing before merge.


mshriver:
want to bump wrapanapi version with this PR?

izapolsk:
yep, I also bear this in mind. one moment.

PR #7631, opened 2018-08-10
[RFR] fixed version check in sprout

izapolsk:
It turned out that 5.10 version comparison as a string doesn't work right. So, ansible url fix was applied to 5.9 as well.


izapolsk:
passes local testing on 5.9 and 5.10

PR #7637, opened 2018-08-11
[1LP][RFR] Chargeback CRUD operations: Parameterizing tests

nachandr:
Parameterizing CRUD tests on compute/storage + fixed/variable to minimize code duplication

All tests passed locally.

### DIFF
--- a/cfme/tests/intelligence/test_chargeback.py
+++ b/cfme/tests/intelligence/test_chargeback.py
@@ -7,7 +7,6 @@ import pytest
 import cfme.intelligence.chargeback.rates as cb
 from cfme import test_requirements
 from cfme.rest.gen_data import rates as _rates
-from cfme.utils.blockers import BZ
 from cfme.utils.rest import (
     assert_response,
     delete_resources_from_collection,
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit 92d92f95c60dd1d88ea6ec299dc9237ef788d659:

`cfme/tests/intelligence/test_chargeback.py`:
- :red_circle: Line 42:9: **F841** *local variable 'rate' is assigned to but never used*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/intelligence/test_chargeback.py
```diff
@@ -119,67 +116,68 @@ def _cleanup():
...
-        storage_rate.fields = {
-            'Fixed Storage Cost 1': with_random_per_time(fixed_rate='500'),
-            'Allocated Disk Storage': with_random_per_time(fixed_rate='100'),
-        }
-    view = storage_rate.create_view(
-        navigator.get_class(storage_rate, 'Details').VIEW)
+    cb_rate.create()
+    assert cb_rate.exists
+    with update(cb_rate):
+        cb_rate.description = cb_rate.description + "-edited"
```
quarckster: **REQURIED**
Please use `format`

cfme/tests/intelligence/test_chargeback.py
```diff
@@ -24,42 +24,41 @@
 ]
 
 
-# TODO refactor to parametrize and collapse into CRUD tests
```
mshriver: Nice to see these go away sometimes!

cfme/tests/intelligence/test_chargeback.py
```diff
@@ -68,42 +67,49 @@ def chargeback_storage_rate(rate_type='fixed'):
...
-        navigator.get_class(chargeback_compute_rate, 'Add').VIEW)
+    view = cb_rate.create_view(
+        navigator.get_class(cb_rate, 'Add').VIEW)
     view.flash.assert_message('Description has already been taken', t='error')
     # cancel form, check all redirect
     view.cancel_button.click()
-    view = chargeback_compute_rate.create_view(
-        navigator.get_class(chargeback_compute_rate, 'All').VIEW)
+    view = cb_rate.create_view(
+        navigator.get_class(cb_rate, 'All').VIEW)
```
mshriver: Ah, these still don't have collection classes.

cfme/tests/intelligence/test_chargeback.py
```diff
@@ -68,42 +67,49 @@ def chargeback_storage_rate(rate_type='fixed'):
...
-@pytest.mark.parametrize('storage_chargeback',
-                         ['fixed', 'variable'],
-                         ids=['fixed_storage_rate', 'variable_storage_rate'])
-@pytest.mark.meta(blockers=[
-    BZ(1532368, forced_streams='5.9',
-       unblock=lambda storage_chargeback: 'variable' not in storage_chargeback)])
-def test_add_new_storage_chargeback(storage_chargeback, request):
-    cb_rate = chargeback_storage_rate(storage_chargeback)
+@pytest.mark.parametrize('chargeback_rate',
+                         ['compute-fixed-add', 'compute-variable-add', 'storage-fixed-add',
```
mshriver: Did you consider parametrizing these components separately instead of as one long list?  The ids are so close to the actual parameters, it seems they could be generated in a list comprehension with a `format()` call.

This is just a bit difficult to read and parse out, since compute/storage and fixed/variable are mixed up, and they're organized by the action, which is the last field.

something like
```
@pytest.mark.parametrize('rate_resource', ['compute', 'storage'])
@pytest.mark.parametrize('rate_type', ['fixed', 'variable'])
@pytest.mark.parametrize('rate_action', ['add', 'delete', 'edit'])
```

Then call the test `test_chargeback_rates` instead of trying to put 'rate' in the parameter ID.

mshriver: If you split parametrization like this, you can use the `chargeback_[storage|compute]_rate` method as a fixture that references rate_resource, and then do yield/cleanup there instead of within the test.  Just something to consider.

mshriver: Why split add/edit/delete into separate parametrization, instead of the test method doing create/edit/delete on both types  (and both resources) as a single CRUD test? This would remove one layer of parametrization - effectively this still makes for as many tests as there were before, just combined into a single method with if blocks.

nachandr: @mshriver , I've turned chargeback_rate into a fixture now and the cleanup is happening in the fixture.

I have add/edit/delete as separate tests since those operations are available as separate options on the UI.

dajoRH:
*CFME QE Bot*

cfme/tests/intelligence/test_chargeback.py
nachandr: Collection class needs to be added to this file as well :https://github.com/ManageIQ/integration_tests/blob/master/cfme/intelligence/chargeback/rates.py#L136

I could submit a follow up PR to add collection class to rates.py and test_chargeback.py .

PR #7642, opened 2018-08-14
[1LP][RFR] Fixing clod tests for supporting the CFME 5.10 automation

dkholodo:
Were fixed selectors for correct working automation tests not only with CFME 5.8 and CFME 5.9 but with CFME 5.10 too.
Was updated returning object of parent cloud provider regarding version of the CFME

{{ pytest: cfme/tests/openstack/cloud/test_networks.py -k 'test_create_network or test_create_subnet or test_create_router_with_gateway or test_clear_router_gateway or test_add_gateway_to_router' }}

### DIFF
--- a/cfme/networks/cloud_network.py
+++ b/cfme/networks/cloud_network.py
@@ -42,7 +42,10 @@ class CloudNetwork(Taggable, BaseEntity):
     def parent_provider(self):
         """ Return object of parent cloud provider """
         view = navigate_to(self, 'Details')
-        provider_name = view.entities.relationships.get_text_of('Parent ems cloud')
+        if self.appliance.version >= '5.10':
+            provider_name = view.entities.relationships.get_text_of('Parent Cloud Provider')
+        else:
+            provider_name = view.entities.relationships.get_text_of('Parent ems cloud')
         return providers.get_crud_by_name(provider_name)
 
     @property
--- a/cfme/networks/views.py
+++ b/cfme/networks/views.py
@@ -427,7 +427,7 @@ class NetworkRouterAddView(BaseLoggedInPage):
     """ Represents Add NetworkRouters page """
     network_manager = Select(id='ems_id')
     router_name = TextInput(name='name')
-    ext_gateway = BootstrapSwitch(id='network_router_external_gateway')
+    ext_gateway = BootstrapSwitch(name='external_gateway')
     network_name = Select(name='cloud_network_id')
     subnet_name = Select(name='cloud_subnet_id')
     cloud_tenant = Select(name='cloud_tenant_id')
@@ -441,7 +441,7 @@ class NetworkRouterAddView(BaseLoggedInPage):
 class NetworkRouterEditView(BaseLoggedInPage):
     """ Represents Edit NetworkRouters page """
     router_name = TextInput(name='name')
-    ext_gateway = BootstrapSwitch(id='network_router_external_gateway')
+    ext_gateway = BootstrapSwitch(name='external_gateway')
     network_name = Select(name='cloud_network_id')
     subnet_name = Select(name='cloud_subnet_id')
     save = Button('Save')
@@ -565,7 +565,7 @@ class SubnetDetailsToolBar(View):
 class SubnetAddView(BaseLoggedInPage):
     """ Represents Add view of subnet """
     title = Text('//div[@id="main-content"]//h1')
-    network_manager = Select(id='ems_id')
+    network_manager = Select(name='ems_id')
     cloud_tenant = Select(name='cloud_tenant_id')
     network = Select(name='network_id')
     subnet_name = TextInput(name='name')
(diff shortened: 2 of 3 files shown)


cfme/networks/views.py
```diff
@@ -420,7 +420,7 @@ class NetworkRouterAddView(BaseLoggedInPage):
     """ Represents Add NetworkRouters page """
     network_manager = Select(id='ems_id')
     router_name = TextInput(name='name')
-    ext_gateway = BootstrapSwitch(id='network_router_external_gateway')
+    ext_gateway = BootstrapSwitch(name='external_gateway')
```
sshveta: Question : Is this both for 5.9 and 5.10 ?

cfme/networks/views.py
```diff
@@ -555,7 +555,7 @@ class SubnetDetailsToolBar(View):
 class SubnetAddView(BaseLoggedInPage):
     """ Represents Add view of subnet """
     title = Text('//div[@id="main-content"]//h1')
-    network_manager = Select(id='ems_id')
+    network_manager = Select(name='ems_id')
```
sshveta: Question : Is this both for 5.9 and 5.10 ?

cfme/networks/views.py
```diff
@@ -434,7 +434,7 @@ def is_displayed(self):
 class NetworkRouterEditView(BaseLoggedInPage):
     """ Represents Edit NetworkRouters page """
     router_name = TextInput(name='name')
-    ext_gateway = BootstrapSwitch(id='network_router_external_gateway')
+    ext_gateway = BootstrapSwitch(name='external_gateway')
```
sshveta: Same question

cfme/networks/views.py
dkholodo: Yes the same for 5.8, 5.9 and 5.10

cfme/networks/views.py
dkholodo: Yes the same for 5.8, 5.9 and 5.10

cfme/networks/views.py
dkholodo: Yes the same for 5.8, 5.9 and 5.10

sshveta:
Okay please comment on 5.9 failure .

   def test_create_network(network, provider):
        """Creates private cloud network and verifies it's relationships"""
        assert network.exists
        assert network.parent_provider.name == provider.name
>       assert network.cloud_tenant == provider.data['tenant']
E       KeyError: 'tenant'

dajoRH:
Lint report for commit d3b74a9948e2718ccea099e301694f0610bc3228:

* 26693b78544f8d2830343ec03226f1d9c2590234
* d3b74a9948e2718ccea099e301694f0610bc3228

*CFME QE Bot*

cfme/networks/cloud_network.py
```diff
@@ -42,7 +42,10 @@ def exists(self):
     def parent_provider(self):
         """ Return object of parent cloud provider """
         view = navigate_to(self, 'Details')
-        provider_name = view.entities.relationships.get_text_of('Parent ems cloud')
+        if self.appliance.version >= '5.10':
+            provider_name = view.entities.relationships.get_text_of('Parent Cloud Provider')
+        else:
+            provider_name = view.entities.relationships.get_text_of('Parent ems cloud')
```
jawatts: **Question**: Why not use the VersionPick object here? 
```
parent_cloud_text = VersionPick({Version.lowest(): Parent ems cloud',
                                '5.10': 'Parent Cloud Provider'}
                                ).pick(self.appliance.version)
provider_name = view.entities.relationships.get_text_of(parent_cloud_text)```

File: cfme/tests/openstack/cloud/test_networks.py [hunk 5]
```diff
@@ -52,7 +52,7 @@ def create_subnet(appliance, provider, network):
 def create_router(appliance, provider, ext_gw, ext_network=None, ext_subnet=None):
     collection = appliance.collections.network_routers
     router = collection.create(name=fauxfactory.gen_alpha(),
-                               tenant=provider.data['tenant'],
+                               tenant=provider.data['provisioning']['cloud_tenant'],
```
jawatts: For good practice, I would use a .get() here instead of accessing the dict directly. That way the test won't throw a KeyError if that key does not exist. What do you think?

cfme/tests/openstack/cloud/test_networks.py
```diff
@@ -114,7 +114,7 @@ def test_create_network(network, provider):
     """Creates private cloud network and verifies it's relationships"""
     assert network.exists
     assert network.parent_provider.name == provider.name
-    assert network.cloud_tenant == provider.data['tenant']
+    assert network.cloud_tenant == provider.data['provisioning']['cloud_tenant']
```
jawatts: I would do the .get() here as well. and the other assertions below

cfme/networks/cloud_network.py
```diff
@@ -42,7 +43,10 @@ def exists(self):
     def parent_provider(self):
         """ Return object of parent cloud provider """
         view = navigate_to(self, 'Details')
-        provider_name = view.entities.relationships.get_text_of('Parent ems cloud')
+        parent_cloud_text = VersionPick({Version.lowest(): 'Parent ems cloud',
```
izapolsk: OPTIONAL: I don't think you need to resolve VersionPick here since widgetastic is able to resolve VersionPicks automatically.

izapolsk:
I updated your PRT command and restarted PRT.

mshriver:
PRT tests for 5.9 are failing from a timeout when creating subnets.

PR #7646, opened 2018-08-14
[1LP][RFR] Fixing analysis profile creation for SSA tests

jan-zmeskal:
Purpose of this PR is twofold:
1. Fix instantiating/creating of analysis profile. It used to throw this:
> TypeError: 'AnalysisProfileCollection' object is not callable

2. Renaming used analysis profile to anything different than "default". SSA used to work only when "default" analysis profile was used. However, since this [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1553808) has been fixed, user should be able to provide any name for their analysis profile. 
Verification run: https://rhv-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/CFME-DEV/job/cfme-5.9-rhv-4.2-integration-flow-dev/268/

{{pytest: cfme/tests/cloud_infra_common/test_vm_instance_analysis.py --long-running --use-provider rhv41}}

### DIFF
--- a/cfme/tests/cloud_infra_common/test_vm_instance_analysis.py
+++ b/cfme/tests/cloud_infra_common/test_vm_instance_analysis.py
@@ -338,17 +338,19 @@ def ssa_analysis_profile(appliance):
     for file in ssa_expect_files:
         collected_files.append({"Name": file, "Collect Contents?": True})
 
-    analysis_profile_name = 'default'
+    analysis_profile_name = 'custom'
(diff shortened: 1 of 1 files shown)


mshriver:
Well done, just saw this in the automation results and I'm glad I searched first!

jan-zmeskal:
@mshriver Well, I did not see that merge coming. :-D The thing is that I still haven't verified properly than the change of analysis profile name does not cause any trouble. But I guess it's okay. Because if some tests fail now because of it, that will be an actual product bug.

PR #7652, opened 2018-08-15
[1LP][RFR] Fix KeyError in test_rest_services for dialog_service_name

mshriver:
its version based, little while ago 'dialog_service_name' was added, but that's not available in 5.9z, so version pick and include the older regex search for the name in the expected message syntax.

few lines later the service request is pulled from the rest api collections again, and the name verified.

### DIFF
--- a/cfme/rest/gen_data.py
+++ b/cfme/rest/gen_data.py
@@ -166,14 +166,7 @@ def services(request, appliance, provider, service_dialog=None, service_catalog=
     assert 'error' not in service_request.message.lower(), ('Provisioning failed: `{}`'
                                                             .format(service_request.message))
 
-    service_name = VersionPicker({
-        Version.lowest(): lambda: re.search(r'\[({}[0-9-]*)\] '
-                                            .format(template_subcollection.name),
-                                            service_request.message)
-                                    .group(1),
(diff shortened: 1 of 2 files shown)


mshriver:
Reqesting re-review, created common method within the test function, left single use in gen_data as-is.

cfme/rest/gen_data.py
```diff
@@ -160,10 +163,17 @@ def _order_finished():
...
     wait_for(_order_finished, num_sec=2000, delay=10)
-    assert 'error' not in service_request.message.lower(), \
-        'Provisioning failed with the message `{}`'.format(service_request.message)
-
-    service_name = str(service_request.options['dialog']['dialog_service_name'])
+    assert 'error' not in service_request.message.lower(), ('Provisioning failed: `{}`'
+                                                            .format(service_request.message))
+
+    service_name = VersionPicker({
+        Version.lowest(): lambda: re.search(r'\[({}[0-9-]*)\] '
```
bsquizz: Sorry to harp on the "centralized helper method" thing, since 2 other reviewers have too, but this is basically the same as `def _get_dialog_service_name` in test_rest_services right? So should that helper live in utils somewhere?

mshriver:
98% passing in PRT, exceptions targeted are resolved.

PR #7657, opened 2018-08-16
[1LP][RFR] Custom Button: Group crud parametrization for All objects

digitronik:
Purpose or Intent
=================
- Added ColorPicker in create method
- Create attributs imporve and changes in fill method
- sorted inputs 
- Added parametrize test for group crud for all objects 5.9 and 5.10

{{pytest: cfme/tests/automate/test_buttons.py::test_button_group_crud -v}}

### DIFF
--- a/cfme/automate/buttons.py
+++ b/cfme/automate/buttons.py
@@ -2,20 +2,30 @@
 import attr
 import re
 
-from navmazing import NavigateToSibling, NavigateToAttribute
-
+from navmazing import NavigateToAttribute, NavigateToSibling
 from widgetastic.utils import ParametrizedString
+from widgetastic.widget import (
+    Checkbox,
+    ColourInput,
+    ConditionalSwitchableView,
+    ParametrizedView,
+    Text,
+    View,
+)
 from widgetastic.xpath import quote
-from widgetastic.widget import Checkbox, ConditionalSwitchableView, ParametrizedView, Text, View
-from widgetastic_manageiq import (SummaryFormItem, FonticonPicker, PotentiallyInvisibleTab,
-                                  RadioGroup)
 from widgetastic_patternfly import BootstrapSelect, Button, CandidateNotFound, Input
 
 from cfme.modeling.base import BaseCollection, BaseEntity
-from cfme.utils.appliance.implementations.ui import navigator, navigate_to, CFMENavigateStep
+from cfme.utils.appliance.implementations.ui import CFMENavigateStep, navigate_to, navigator
 from cfme.utils.blockers import BZ
 from cfme.utils.update import Updateable
 from cfme.utils.wait import wait_for
+from widgetastic_manageiq import (
+    FonticonPicker,
+    PotentiallyInvisibleTab,
+    RadioGroup,
+    SummaryFormItem,
+)
 
 from . import AutomateCustomizationView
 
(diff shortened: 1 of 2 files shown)


digitronik:
Local results for 5.10:

http://pastebin.test.redhat.com/632979

for 5.9 run PR https://github.com/ManageIQ/integration_tests/pull/7638 need to merge.

cfme/automate/buttons.py
```diff
@@ -539,41 +558,45 @@ class ButtonGroupCollection(BaseCollection):
     TEMPLATE = "VM Template and Image"
     VM_INSTANCE = "VM and Instance"
 
-    def instantiate(self, text, hover, type, image=None):
+    def instantiate(self, text, hover, type, image=None, display=None, icon_color=None):
         if image:
             pass
         elif self.appliance.version < '5.9':
```
sshveta: This can be removed now . We are not checking for versions less that 5.9 anymore

cfme/automate/buttons.py
```diff
@@ -539,41 +558,45 @@ class ButtonGroupCollection(BaseCollection):
...
+        return self.ENTITY.from_collection(self, text, hover, type, image, display, icon_color)
 
-    def create(self, text, hover, type, image=None):
+    def create(self, text, hover, type, image=None, display=None, icon_color=None):
         self.type = type
+
+        # Icon selection is Mandatory
         if image:
             pass
         elif self.appliance.version < '5.9':
```
sshveta: Same here

cfme/automate/buttons.py
```diff
@@ -539,41 +558,45 @@ class ButtonGroupCollection(BaseCollection):
...
-        })
+            'image': image,
+            'display': display,
+            'icon_color': icon_color}
+        )
         view.add_button.click()
         view = self.create_view(ButtonGroupObjectTypeView)
 
         view.flash.assert_no_error()
         if self.appliance.version < '5.9':
```
sshveta: Same as above . Remove version check for less that 5.9

cfme/automate/buttons.py
digitronik: @sshveta still we have 5.8 branches. I will add card to remove it.

sshveta: :+1:

cfme/automate/buttons.py
```diff
@@ -2,21 +2,32 @@
...
 from cfme.utils.update import Updateable
 from cfme.utils.version import Version
 from cfme.utils.wait import wait_for
+from widgetastic_manageiq import (
+    FakeWidget,
+    FonticonPicker,
+    PotentiallyInvisibleTab,
+    RadioGroup,
+    SummaryFormItem,
+)
```
quarckster: **REQUIRED**
This import should be in the third-party section.

digitronik: @quarckster  `wt_manageiq` we treat as internal right not like `wt.core` and `wt.pfly`

quarckster: @mshriver is something changed? I was thinking that eventually widgetastic_manageiq will go to a separate repo.

mshriver: It can/should eventually be separated - but until then its a local import and should be treated as such, imported in the 3rd group.

dajoRH:
*CFME QE Bot*

PR #7658, opened 2018-08-16
[1LP][RFR] Remove deadlock from ssh client.

jaryn:
We were reading only from the stdout, therefore if there was something
available on stderr, it was available forever, causing deadlock.


izapolsk:
@jaryn, please sign your commit. PRT won't run w/o that.
I'll merge your PR as soon as PRT is done

Issue #7603, opened 2018-08-06
[Bug] If BZ is reported for version 5.9 and fixed in 5.10, it is not skipped


jan-zmeskal:
Firstly, I noticed this behaviour because of this [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1545240). `test_query_provider_attributes` [checks](https://github.com/ManageIQ/integration_tests/blob/ef2fc2dd243b35cf96007c81d3642e02711a4f11/cfme/tests/cloud_infra_common/test_rest_providers.py#L66) if this BZ blocks. At some point during processing of this BZ, [Bugzilla.resolve_blocker](https://github.com/ManageIQ/integration_tests/blob/ef2fc2dd243b35cf96007c81d3642e02711a4f11/cfme/utils/bz.py#L145) is called. The bug is added to the `filtered` set. But when execution gets to this [point](https://github.com/ManageIQ/integration_tests/blob/ef2fc2dd243b35cf96007c81d3642e02711a4f11/cfme/utils/bz.py#L196), `check_fixed_in` returns `False`. Therefore `bug` is not returned and instead `None` is returned. As a consequence, `False` is also returned during resolving property [blocks](https://github.com/ManageIQ/integration_tests/blob/ef2fc2dd243b35cf96007c81d3642e02711a4f11/cfme/utils/blockers.py#L187) of `BZ`.  As far as I understand, this is not expected behaviour. I believe this bug should resolve as blocking if it is in one of [Bugzilla.open_states](https://github.com/ManageIQ/integration_tests/blob/ef2fc2dd243b35cf96007c81d3642e02711a4f11/cfme/utils/bz.py#L107). It's nice that it's fixed in 5.10, but I still need to test 5.9 and skip this bug if necessary.

Additional info: I know that at time of reporting this issue, the BZ is acutally in ON_QA state. But that should still should make it blocking as I have ON_QA state included in env.yaml in `bugzilla[skip]`.

Also, [here](http://pastebin.test.redhat.com/627464) you can see how flow of `resolve_blockers` actually looks like when I was following it manually.

Sorry for the text wall, just wanted to spare you some debugging. :-)

jan-zmeskal:
Creating #7604 as a workaround.

mshriver:
I believe this has been resolved with @jaryn contribution here:

https://github.com/ManageIQ/integration_tests/pull/8208/files#diff-89e51f6d56a899c026153169db2576a5

[...]

PR #7661, opened 2018-08-16
[WIPTEST] Change BZ resolve_blockers boolean

mshriver:
resolve_blockers was returning a bug if check_fixed_in was returning true, which is backwards.

The method should return a bug if that bug is a 'blocker' and is in the right series and is _not_ fixed in the same series.

FIXES 7603

This is very odd to me, as this isn't something that has changed recently. Have we been incorrectly resolving blockers this whole time?

### DIFF
--- a/cfme/scripting/ipyshell.py
+++ b/cfme/scripting/ipyshell.py
@@ -7,7 +7,9 @@ IMPORTS = [
     'from cfme.utils import conf',
     'from cfme.fixtures.pytest_store import store',
     'from cfme.utils.appliance.implementations.ui import navigate_to',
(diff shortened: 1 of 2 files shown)


mshriver:
@jan-zmeskal Haven't tested beyond your example PR in #7603  but it looks like this has been wrong for a while if this is actually it.

We're calling a method to check if the series for the fixed_in field matches our version, and we want to return a bug if it has _not_ been fixed in series. I believe this is causing the inversion that you raised.

jan-zmeskal:
@mshriver Thank you so much for looking into this!

PR #7662, opened 2018-08-17
[1LP][RFR] additional sprout client/server fixes in order to fix the issue caused by reboot during rename vm

izapolsk:
1. sprout plugin returned appliances using pool.fulfilled attribute what is incorrect. it should use pool.finished instead.
2. rename_appliance task was made synchronous in order to get rid of cases when provided appliance is rebooted afterwards

### DIFF
--- a/sprout/appliances/tasks.py
+++ b/sprout/appliances/tasks.py
@@ -697,10 +697,10 @@ def apply_lease_times_after_pool_fulfilled(self, appliance_pool_id, time_minutes
         pool.logger.info("Applying lease time and renaming appliances")
         for appliance in pool.appliances:
             apply_lease_times.delay(appliance.id, time_minutes)
-        rename_appliances_for_pool.delay(pool.id)
         with transaction.atomic():
             pool.finished = True
             pool.save(update_fields=['finished'])
+            pool.logger.info("Pool {} setup is finished".format(appliance_pool_id))
     else:
         # Look whether we can swap any provisioning appliance with some in shepherd
         pool.logger.info("Pool isn't fulfilled yet")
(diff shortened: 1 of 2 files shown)


sprout/appliances/tasks.py
```diff
@@ -1584,17 +1589,31 @@ def delete_template_from_provider(self, template_id):
 
 
 @singleton_task()
-def appliance_rename(self, appliance_id, new_name):
-    self.logger.info("Start renaming process for {} to {}".format(appliance_id, new_name))
+def appliance_rename(self, appliance_id):
     try:
         appliance = Appliance.objects.get(id=appliance_id)
+        new_name = '{}_'.format(appliance.appliance_pool.owner.username)
```
mshriver: Consider moving this new code out of the try block since there's no exception handling for anything within these lines, just the appliance object get.

sprout/appliances/tasks.py
```diff
@@ -1584,17 +1589,31 @@ def delete_template_from_provider(self, template_id):
...
+            appliance.template.date.strftime("%y%m%d"),
+            fauxfactory.gen_alphanumeric(length=4))
+        self.logger.info("Start renaming process for {} to {}".format(appliance_id, new_name))
     except ObjectDoesNotExist:
+        self.logger.warning("No such appliance {} in sprout".format(appliance_id))
         return None
     if not appliance.provider.is_working:
         raise RuntimeError('Provider is not working.')
-    if not appliance.provider.allow_renaming:
+    if (not appliance.provider.allow_renaming or appliance.is_openshift or
```
mshriver: Looks like this could be checked immediately after pulling the appliance object, before making additional calls/queries to generate the new name

izapolsk:
sprout behaves correctly with this patch. merging it.

PR #7666, opened 2018-08-19
[1LP][RFR] Add flavor methods and tests

jadhaj:
Purpose or Intent
=================
Add test test_create_instance_with_zero_disk_flavor  including related mothods

{{ pytest: cfme/tests/openstack/cloud/test_flavors.py }}

### DIFF
--- a/cfme/cloud/flavor.py
+++ b/cfme/cloud/flavor.py
@@ -2,14 +2,17 @@
 """
 import attr
 
-from navmazing import NavigateToAttribute
-from widgetastic_patternfly import Dropdown, Button, View, BreadCrumb
+from navmazing import NavigateToAttribute, NavigateToSibling
+from widgetastic_patternfly import (Accordion, BootstrapSwitch, BreadCrumb, Button,
+                                    Dropdown, TextInput, View)
+
+from widgetastic.widget import Text, Select
 
 from cfme.base.ui import BaseLoggedInPage
 from cfme.common import Taggable
 from cfme.exceptions import FlavorNotFound, ItemNotFound
 from cfme.modeling.base import BaseEntity, BaseCollection
-from cfme.utils.appliance.implementations.ui import CFMENavigateStep, navigator
+from cfme.utils.appliance.implementations.ui import CFMENavigateStep, navigator, navigate_to
 from widgetastic_manageiq import (
     BaseEntitiesView, ItemsToolBarViewSelector, SummaryTable, Text, Table, Accordion, ManageIQTree,
     PaginationPane, Search)
@@ -27,6 +30,7 @@ class FlavorView(BaseLoggedInPage):
 class FlavorToolBar(View):
     policy = Dropdown('Policy')
     download = Dropdown('Download')
+    configuration = Dropdown('Configuration')
     view_selector = View.nested(ItemsToolBarViewSelector)
 
 
@@ -38,6 +42,7 @@ class FlavorEntities(BaseEntitiesView):
 class FlavorDetailsToolBar(View):
     policy = Dropdown('Policy')
     download = Button(title='Download summary in PDF format')
+    configuration = Dropdown('Configuration')
 
 
 class FlavorDetailsAccordion(View):
@@ -98,6 +103,38 @@ class FlavorDetailsView(FlavorView):
     entities = FlavorDetailsEntities()
 
 
+class FlavorAddEntities(View):
+    breadcrumb = BreadCrumb()
+    title = Text('//div[@id="main-content"]//h1')
+
+
+class FlavorAddForm(View):
+    """ Represents Add Flavor page """
+    provider = Select(name='ems_id')
+    flavor_name = TextInput(name='name')
+    ram_size = TextInput(name='ram')
+    vcpus = TextInput(name='vcpus')
+    disk_size = TextInput(name='disk')
+    swap_size = TextInput(name='swap')
+    rxtx_factor = TextInput(name='rxtx_factor')
+    public = BootstrapSwitch(name='is_public')
+    add = Button('Add')
+    cancel = Button('Cancel')
+
+
+class FlavorAddView(FlavorView):
+    @property
+    def is_displayed(self):
+        expected_title = "Add a new Flavor"
+        return (
+            self.in_availability_zones and
+            self.entities.title.text == expected_title and
+            self.entities.breadcrumb.active_location == expected_title)
+
+    entities = View.nested(FlavorAddEntities)
+    form = View.nested(FlavorAddForm)
+
+
 @attr.s
 class Flavor(BaseEntity, Taggable):
     """
@@ -107,11 +144,61 @@ class Flavor(BaseEntity, Taggable):
 
     name = attr.ib()
     provider = attr.ib()
+    ram = attr.ib(default=None)
+    vcpus = attr.ib(default=None)
+    disk = attr.ib(default=None)
+    swap = attr.ib(default=None)
+    rxtx = attr.ib(default=None)
+    is_public = attr.ib(default=True)
+
+    def delete(self, cancel=False):
+        """Delete current falvor """
+        view = navigate_to(self, 'Details')
+        view.toolbar.configuration.item_select('Remove Flavor', handle_alert=not cancel)
+        view.flash.assert_no_error()
+
+    def refresh(self):
+        """Refresh provider relationships and browser"""
+        self.provider.refresh_provider_relationships()
+        self.browser.refresh()
+
+    @property
+    def exists(self):
+        try:
+            navigate_to(self, 'Details')
+            return True
+        except FlavorNotFound:
+            return False
 
 
 @attr.s
 class FlavorCollection(BaseCollection):
     ENTITY = Flavor
+
+    def create(self, name, provider, ram, vcpus, disk, swap, rxtx, is_public=True, cancel=False):
+        """Create new falvor"""
+        view = navigate_to(self, 'Add')
+        form_params = {'provider': provider.name,
+                       'flavor_name': name,
+                       'ram_size': ram,
+                       'vcpus': vcpus,
+                       'disk_size': disk,
+                       'swap_size': swap,
+                       'rxtx_factor': rxtx,
+                       'public': is_public}
+
+        view.form.fill(form_params)
+
+        if cancel:
+            view.form.cancel.click()
+
+        else:
+            view.form.add.click()
+
+        view = self.create_view(FlavorAllView)
+        view.flash.assert_no_error()
+        flavor = self.instantiate(name, provider, ram, vcpus, disk, swap, rxtx, is_public)
+        return flavor
 
 
 @navigator.register(FlavorCollection, 'All')
@@ -136,3 +223,12 @@ class FlavorDetails(CFMENavigateStep):
             raise FlavorNotFound('Could not locate flavor "{}" on provider {}'
                                  .format(self.obj.name, self.obj.provider.name))
         row.click()
+
+
+@navigator.register(FlavorCollection, 'Add')
+class FlavorAdd(CFMENavigateStep):
+    prerequisite = NavigateToSibling('All')
+    VIEW = FlavorAddView
+
+    def step(self):
+        self.prerequisite_view.toolbar.configuration.item_select('Add a new Flavor')
--- a/cfme/tests/openstack/cloud/test_flavors.py
+++ b/cfme/tests/openstack/cloud/test_flavors.py
@@ -1,0 +1,125 @@
+"""Tests for Openstack cloud Flavors"""
+
+import fauxfactory
+import pytest
+
+from cfme.cloud.instance.openstack import OpenStackInstance
+from cfme.cloud.provider.openstack import OpenStackProvider
+from cfme.utils.appliance.implementations.ui import navigator, navigate_to
+from cfme.utils.wait import wait_for
+from widgetastic.utils import partial_match
+
+pytestmark = [
+    pytest.mark.usefixtures("setup_provider"),
+    pytest.mark.provider([OpenStackProvider], scope='function')
+]
+
+
+DISK_SIZE = 1
+RAM = 64
+VCPUS = 1
+SWAP_SIZE = 1
+RXTX = 1
+ZERO_DISK_SIZE = 0
+
+
+@pytest.fixture(scope='function')
+def zero_disk_flavor(provider):
+    api_flv = provider.mgmt.api.flavors.create(fauxfactory.gen_alpha(), RAM, VCPUS, ZERO_DISK_SIZE)
+    wait_for(provider.is_refreshed, func_kwargs=dict(refresh_delta=10), timeout=600)
+    zero_disk_flavor = provider.appliance.collections.cloud_flavors.instantiate(
+        api_flv.name, provider)
+
+    yield zero_disk_flavor
+
+    if zero_disk_flavor.exists:
+        zero_disk_flavor.delete()
+
+
+@pytest.fixture(scope='function')
+def new_instance(provider, zero_disk_flavor):
+    flavor_name = zero_disk_flavor.name
+    prov_data = provider.data['provisioning']
+    prov_form_data = {
+        'request': {'email': fauxfactory.gen_email(),
+                    'first_name': fauxfactory.gen_alpha(),
+                    'last_name': fauxfactory.gen_alpha()},
+        'catalog': {'num_vms': '1',
+                    'vm_name': fauxfactory.gen_alpha()},
+        'environment': {'cloud_network': prov_data['cloud_network'],
+                        'cloud_tenant': prov_data['cloud_tenant']},
+        'properties': {'instance_type': partial_match(flavor_name)},
+    }
+
+    instance_name = prov_form_data['catalog']['vm_name']
+
+    try:
+        instance = provider.appliance.collections.cloud_instances.create(
+            instance_name,
+            provider,
+            prov_form_data, find_in_cfme=True
+        )
+
+    except KeyError:
+        # some yaml value wasn't found
+        pytest.skip('Unable to find an image map in provider "{}" provisioning data: {}'
+                    .format(provider, prov_data))
+
+    yield instance
+
+    instance.cleanup_on_provider()
+
+
+@pytest.mark.ignore_stream('5.9')
+def test_create_instance_with_zero_disk_flavor(new_instance, soft_assert):
+    view = navigate_to(new_instance, 'Details')
+    prov_data = new_instance.provider.data['provisioning']
+    power_state = view.entities.summary('Power Management').get_text_of('Power State')
+    assert power_state == OpenStackInstance.STATE_ON
+
+    vm_tmplt = view.entities.summary('Relationships').get_text_of('VM Template')
+    soft_assert(vm_tmplt == prov_data['image']['name'])
+
+    flavors = [f.name for f in new_instance.provider.mgmt.api.flavors.list()]
+    flavor = view.entities.summary('Relationships').get_text_of('Flavor')
+    soft_assert(flavor in flavors)
+
+    # Assert other relationships in a loop
+    props = [('Availability Zone', 'availability_zone'),
+             ('Cloud Tenants', 'cloud_tenant'),
+             ('Virtual Private Cloud', 'cloud_network')]
+
+    for p in props:
+        v = view.entities.summary('Relationships').get_text_of(p[0])
+        soft_assert(v == prov_data[p[1]])
+
+
+def test_flavor_crud(appliance, provider, request):
+    collection = appliance.collections.cloud_flavors
+    flavor = collection.create(name=fauxfactory.gen_alpha(),
+                               provider=provider,
+                               ram=RAM,
+                               vcpus=VCPUS,
+                               disk=DISK_SIZE,
+                               swap=SWAP_SIZE,
+                               rxtx=RXTX)
+
+    @request.addfinalizer
+    def cleanup():
+        if flavor.exists:
+            flavor.delete()
+
+    view = appliance.browser.create_view(navigator.get_class(collection, 'All').VIEW)
+    view.flash.assert_success_message(
+        'Add of Flavor "{}" was successfully initialized.'.format(flavor.name))
+
+    wait_for(lambda: flavor.exists, delay=5, timeout=600, fail_func=flavor.refresh,
+             message='Wait for flavor to appear')
+
+    flavor.delete()
+    view = appliance.browser.create_view(navigator.get_class(collection, 'All').VIEW)
+    view.flash.assert_success_message(
+        'Delete of Flavor "{}" was successfully initiated.'.format(flavor.name))
+
+    wait_for(lambda: not flavor.exists, delay=5, timeout=600, fail_func=flavor.refresh,
+             message='Wait for flavor to appear')
--- a/cfme/cloud/instance/openstack.py
+++ b/cfme/cloud/instance/openstack.py
@@ -4,6 +4,7 @@ from navmazing import NavigateToSibling
 from widgetastic.widget import View, NoSuchElementException
 from widgetastic_patternfly import Button, BootstrapSelect
 from widgetastic_manageiq import CheckboxSelect, Select, Input
+from widgetastic.utils import partial_match
 
 from cfme.exceptions import OptionNotAvailable, DestinationNotFound
 from cfme.common.vm_views import RightSizeView
@@ -65,6 +66,12 @@ class OpenStackInstance(Instance):
         view = navigate_to(self, 'DetachVolume')
         view.form.volume.fill(name)
         view.form.submit_button.click()
+
+    def reconfigure(self, flavor):
+        view = navigate_to(self, 'Reconfigure')
+        view.form.flavor.fill(partial_match(flavor))
+        view.form.submit_button.click()
+        view.flash.assert_no_error()
 
     @property
     def volume_count(self):
@@ -167,7 +174,7 @@ class MigrateView(CloudInstanceView):
 class ReconfigureView(CloudInstanceView):
     @View.nested
     class form(View):  # noqa
-        flavor = BootstrapSelect('flavor')
+        flavor = Select(name='flavor_id')
         submit_button = Button('Submit')
         cancel_button = Button('Cancel')
 


dajoRH:
Lint report for commit 7c37f2c640feb14ecbd9bbfcd06c6420feb6bf16:

`cfme/cloud/flavor.py`:
- :red_circle: Line 6:1: **F401** *'widgetastic_patternfly.BootstrapSelect' imported but unused*

`cfme/tests/openstack/cloud/test_flavors.py`:
- :red_circle: Line 7:1: **F401** *'cfme.utils.appliance.implementations.ui.navigate_to' imported but unused*
- :red_circle: Line 8:1: **F401** *'cfme.utils.blockers.BZ' imported but unused*
- :red_circle: Line 9:1: **F401** *'cfme.utils.log.logger' imported but unused*
- :red_circle: Line 10:1: **F401** *'cfme.utils.update.update' imported but unused*
- :bangbang: Line 22:7: **E225** *missing whitespace around operator*

Hmmm


**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
Lint report for commit d2fa66cfaf408fd0e2976af6d338aea3d9cedea9:

`cfme/tests/openstack/cloud/test_flavors.py`:
- :bangbang: Line 43:101: **E501** *line too long (103 > 100 characters)*

**No commit flaws detected**.

*CFME QE Bot*

cfme/cloud/flavor.py
```diff
@@ -108,11 +133,60 @@ class Flavor(BaseEntity, Taggable):
...
+                 fail_func=self.browser.refresh)
+
+    @property
+    def exists(self):
+        try:
+            navigate_to(self, 'Details')
+        except FlavorNotFound:
+            return False
+        else:
+            return True
```
sshveta: This can be done as : 
@property
    def exists(self):
        try:
            navigate_to(self, 'Details')
            return True
        except FlavorNotFound:
            return False

cfme/tests/openstack/cloud/test_flavors.py
```diff
@@ -0,0 +1,67 @@
...
+import fauxfactory
+import pytest
+
+from cfme.cloud.provider.openstack import OpenStackProvider
+from cfme.utils.wait import wait_for
+
+pytestmark = [
+    pytest.mark.usefixtures("setup_provider_modscope"),
+    pytest.mark.provider([OpenStackProvider], scope='module')
+]
```
sshveta: Required : 
pytestmark = [
    pytest.mark.usefixtures("setup_provider"),
    pytest.mark.provider([OpenStackProvider], scope='function')
]

dajoRH:
Lint report for commit 2cb78aebf2b6e3f3a46b7731817298ddc2cf131f:

`cfme/cloud/flavor.py`:
- :red_circle: Line 9:1: **F811** *redefinition of unused 'View' from line 6*

**No commit flaws detected**.

*CFME QE Bot*

cfme/cloud/flavor.py
```diff
@@ -98,6 +104,23 @@ def is_displayed(self):
...
+    ram_size = TextInput(name='ram')
+    vcpus = TextInput(name='vcpus')
+    disk_size = TextInput(name='disk')
+    swap_size = TextInput(name='swap')
+    rxtx_factor = TextInput(name='rxtx_factor')
+    public = BootstrapSwitch(name='is_public')
+    add = Button('Add')
+
+    @property
+    def is_displayed(self):
```
sshveta: Is_displayed should check if the view is displayed and return True or False based on that .
Example : 
https://github.com/ManageIQ/integration_tests/blob/master/cfme/services/catalogs/catalog_items/ansible_catalog_items.py#L200

sshveta:
One minor comment on is_displayed method

dajoRH:
Lint report for commit 30e9eabe558bf4fc0b7346c2b035bd569186fec6:

`cfme/cloud/flavor.py`:
- :bangbang: Line 111:1: **E302** *expected 2 blank lines, found 1*
- :bangbang: Line 127:17: **E126** *continuation line over-indented for hanging indent*

**No commit flaws detected**.

*CFME QE Bot*

sshveta:
waiting for PRT

cfme/cloud/flavor.py
```diff
@@ -108,11 +142,59 @@ class Flavor(BaseEntity, Taggable):
...
+                       'rxtx_factor': rxtx,
+                       'public': is_public}
+
+        view.fill(form_params)
+        view.add.click()
+        view.flash.assert_success_message(
+            'Add of Flavor "{}" was successfully initialized.'.format(name))
+        flavor = self.instantiate(name, provider)
+        wait_for(provider.is_refreshed, func_kwargs=dict(refresh_delta=10), timeout=600)
+        wait_for(lambda: flavor.exists, timeout=100, fail_func=flavor.browser.refresh,
```
jawatts: perhaps ```flavor.name in view.entities.all_entity_names``` ?

cfme/cloud/flavor.py
```diff
@@ -108,11 +142,59 @@ class Flavor(BaseEntity, Taggable):
...
+                       'swap_size': swap,
+                       'rxtx_factor': rxtx,
+                       'public': is_public}
+
+        view.fill(form_params)
+        view.add.click()
+        view.flash.assert_success_message(
+            'Add of Flavor "{}" was successfully initialized.'.format(name))
+        flavor = self.instantiate(name, provider)
+        wait_for(provider.is_refreshed, func_kwargs=dict(refresh_delta=10), timeout=600)
```
jawatts: Question: Why to two ```wait_for``` here?

cfme/cloud/flavor.py
```diff
@@ -108,11 +142,59 @@ class Flavor(BaseEntity, Taggable):
...
+                       'ram_size': ram,
+                       'vcpus': vcpus,
+                       'disk_size': disk,
+                       'swap_size': swap,
+                       'rxtx_factor': rxtx,
+                       'public': is_public}
+
+        view.fill(form_params)
+        view.add.click()
+        view.flash.assert_success_message(
```
jawatts: When you add a new flavor, the page is redirected to the parent view. I believe you would have to create a new view here correct?

cfme/cloud/flavor.py
```diff
@@ -108,11 +142,59 @@ class Flavor(BaseEntity, Taggable):
...
+                       'flavor_name': name,
+                       'ram_size': ram,
+                       'vcpus': vcpus,
+                       'disk_size': disk,
+                       'swap_size': swap,
+                       'rxtx_factor': rxtx,
+                       'public': is_public}
+
+        view.fill(form_params)
+        view.add.click()
```
jawatts: Question: What if you want to cancel adding a new flavor. Perhaps pass a new param ```cancel``` to the ```create``` method and implement this use case.

jawatts:
Looks like there are some failures in PRT. Sprout failed to provision appliances for 59z but test failures in 510z.

cfme/cloud/flavor.py
jadhaj: I intended to get flavor object to be returned. Does instantiate creates a new one ?

cfme/cloud/flavor.py
jadhaj: Sorry but didn't get your point, can you clarify please

cfme/cloud/flavor.py
jadhaj: Wanted to be sure that provider was refreshed before checking existence

cfme/cloud/flavor.py
jadhaj: Will fix that

cfme/cloud/flavor.py
jadhaj: Also i saw examples where instantiate is used:
https://github.com/ManageIQ/integration_tests/blob/master/cfme/storage/volume.py#L333

cfme/cloud/flavor.py
jadhaj: Add refresh method and removed redundant wait

cfme/cloud/flavor.py
```diff
@@ -108,11 +143,66 @@ class Flavor(BaseEntity, Taggable):
...
+        view.toolbar.configuration.item_select('Remove Flavor', handle_alert=True)
+        view.flash.assert_success_message(
+            'Delete of Flavor "{}" was successfully initiated.'.format(self.name))
+
+        if wait:
+            self.wait_for_disappear()
+
+    def wait_for_disappear(self):
+        """Wait for flavor to disappear"""
+        wait_for(self.provider.is_refreshed, func_kwargs=dict(refresh_delta=10), timeout=300)
```
jawatts: Question: Why are you waiting for the provider to refresh here? Would just waiting for the Flavor object to not present in the table not be enough?

cfme/cloud/flavor.py
```diff
@@ -108,11 +143,66 @@ class Flavor(BaseEntity, Taggable):
...
+        wait_for(lambda: not self.exists,
+                 timeout=300,
+                 message='Wait for flavor to disappear',
+                 delay=20,
+                 fail_func=self.browser.refresh)
```
jawatts: Thought... why not call self.refresh here, which calls the method ```refresh()``` on line 165 below, then we could possible remove the wait for on line 158. What do you think? If you do agree, I think moving the wait into the delete method would be the next step.

cfme/cloud/flavor.py
```diff
@@ -108,11 +143,66 @@ class Flavor(BaseEntity, Taggable):
...
+            form_params = {'provider': provider.name,
+                           'flavor_name': name,
+                           'ram_size': ram,
+                           'vcpus': vcpus,
+                           'disk_size': disk,
+                           'swap_size': swap,
+                           'rxtx_factor': rxtx,
+                           'public': is_public}
+
+            view.fill(form_params)
```
jawatts: What do you think about this:
```
if cancel:
    view.cancel.click()
    flash_message = "Add of Flavor cancelled by user."
else:
    view.add.click()
    flash_message = 'Add of Flavor "{}" was successfully initialized.'.format(name)
```
Then below you can assert on ```flash_message```

cfme/cloud/flavor.py
```diff
@@ -108,11 +143,66 @@ class Flavor(BaseEntity, Taggable):
...
+            view.add.click()
+            view.flash.assert_success_message(
```
jawatts: When you add a new flavor, it redirects you to the parent page. You will have to create the view again. 

```view = self.create_view(FlavorAllView)```

File: cfme/cloud/flavor.py [hunk 12]
```diff
@@ -108,11 +143,66 @@ class Flavor(BaseEntity, Taggable):
     name = attr.ib()
     provider = attr.ib()
 
+    def delete(self, wait=True):
```
jawatts: I suggest passing ```cancel=False``` as a param here, which will give the capability of how the alert popup is handled.

cfme/cloud/flavor.py
```diff
@@ -108,11 +143,66 @@ class Flavor(BaseEntity, Taggable):
...
+        """Delete current falvor """
+        view = navigate_to(self, 'Details')
+        view.toolbar.configuration.item_select('Remove Flavor', handle_alert=True)
```
jawatts: Then here ```view.toolbar.configuration.item_select('Remove Flavor', handle_alert=not cancel)```.

cfme/cloud/flavor.py
jadhaj: Yes, right. because of this redirection the assert message is working properly.
After this point I am not using the view anymore, all i need is to return flavor object to user

How can creating the  View help me doing that ?

dajoRH:
Lint report for commit b5ccab9a74a652a2d11e1a5c7ce0817754970b05:

`cfme/cloud/flavor.py`:
- :bangbang: Line 146:45: **E202** *whitespace before ')'*

**No commit flaws detected**.

*CFME QE Bot*

jadhaj:
@jawatts I think i have rectified all your comments except the related to view = self.create_view(FlavorAllView) which i responded above

jadhaj:
@jawatts Regarding PRT failure, I am not sure why is that
What i am sure about is that all tests passed on my env with 5.10

sshveta:
@jadhaj : There are PRT failures for 5.9 and 5.10 . 

           def test_create_zero_disk_flavor(zero_disk_flavor):
>       assert zero_disk_flavor.exists
E       AssertionError: assert False
E        +  where False = Flavor(name=u'GgqCThyuRP', provider=OpenStackProvider(endpoints={'default': <c...e, keystone_v3_domain_id='default', tenant_mapping=False, infra_provider=None)).exists

cfme/cloud/flavor.py
```diff
@@ -108,11 +143,65 @@ class Flavor(BaseEntity, Taggable):
...
+        view.fill(form_params)
+
+        if cancel:
+            view.cancel.click()
+            flash_message = "Add of Flavor cancelled by user."
+            view.flash.assert_message(flash_message)
+        else:
+            view.add.click()
+            flash_message = 'Add of Flavor "{}" was successfully initialized.'.format(name)
+            view.flash.assert_success_message(flash_message)
```
jawatts: You almost got it! But starting at line 199, it should not be in the else right? In the if/else block, it is controlling which action button you are clicking and setting the appropriate ```flash_message``` variable for both the cancel and add cases. Both cases will direct you to the parent view, in this case the FlavorAllView. 

The reason we want to create a new view here is because the previous view was set to the FlavorAddView. After a click, either cancel or add, the view is now on the parent view (FlavorAllView). 

Here is an example I just ran through:
```In[3]: collection = appliance.collections.cloud_flavors
In[4]: view = navigate_to(collection, 'Add')
In[5]: view
Out[5]: <cfme.cloud.flavor.FlavorAddView at 0x7f3fa9760e50>
In[6]: view.cancel.click()
In[7]: view
Out[7]: <cfme.cloud.flavor.FlavorAddView at 0x7f3fa9760e50>
```
So when you are checking the flash message, you want to check it on the parent view. 

Also, another thing I just realized. When you cancel adding a Flavor, it is not a success message. You will want to use ```view.flash.assert_message()```. You can use that for both add and cancel cases

dajoRH:
Lint report for commit 07b5e961f9c851eed34f387fc84bc151cdd834c1:

`cfme/cloud/flavor.py`:
- :bangbang: Line 211:1: **E303** *too many blank lines (3)*

**No commit flaws detected**.

*CFME QE Bot*

cfme/cloud/flavor.py
```diff
@@ -108,11 +143,69 @@ class Flavor(BaseEntity, Taggable):
...
+            flash_message = 'Add of Flavor "{}" was successfully initialized.'.format(name)
+
+        view = self.create_view(FlavorAllView)
+        view.flash.assert_message(flash_message)
+        flavor = self.instantiate(name, provider)
+
+        if wait:
+            wait_for(lambda: flavor.exists, delay=50, timeout=600, fail_func=flavor.refresh,
+                 message='Wait for flavor to appear')
+            return flavor
```
jawatts: So if ```wait=False``` this method will never return. ```return flavor``` will have to be moved out of this if block.

cfme/cloud/flavor.py
```diff
@@ -108,11 +143,69 @@ class Flavor(BaseEntity, Taggable):
...
+            return True
+        except FlavorNotFound:
+            return False
+
 
 @attr.s
 class FlavorCollection(BaseCollection):
     ENTITY = Flavor
 
+    def create(self, name, provider, ram, vcpus, disk, swap, rxtx, is_public=True, cancel=False,
```
jawatts: A doc block is needed here as well.

jadhaj: Done

cfme/cloud/flavor.py
```diff
@@ -108,11 +143,71 @@ class Flavor(BaseEntity, Taggable):
...
+
+        if cancel:
+            view.cancel.click()
+            flash_message = "Add of Flavor cancelled by user."
+        else:
+            view.add.click()
+            flash_message = 'Add of Flavor "{}" was successfully initialized.'.format(name)
+
+        view = self.create_view(FlavorAllView)
+        view.flash.assert_message(flash_message)
```
mshriver: This specific message shouldn't be asserted within the CRUD method, for reasons stated above in `delete`. Just assert no error, and add these message assertions to the CRUD test for the entity.

cfme/cloud/flavor.py
```diff
@@ -108,11 +143,71 @@ class Flavor(BaseEntity, Taggable):
...
 
 @attr.s
 class FlavorCollection(BaseCollection):
     ENTITY = Flavor
 
+    def create(self, name, provider, ram, vcpus, disk, swap, rxtx, is_public=True, cancel=False,
+            wait=True):
+        """Create new falvor"""
+        view = navigate_to(self, 'Add')
+        form_params = {'provider': provider.name,
```
mshriver: If all of these parameters are what define a flavor entity that can be created in MIQ, then the entity model should include them as attributes. Otherwise there's no way to model a flavor that we create, and assert any of its attributes against what is created on the provider itself. (verifying MIQ behavior.

cfme/cloud/flavor.py
```diff
@@ -108,11 +143,71 @@ class Flavor(BaseEntity, Taggable):
...
+        else:
+            view.add.click()
+            flash_message = 'Add of Flavor "{}" was successfully initialized.'.format(name)
+
+        view = self.create_view(FlavorAllView)
+        view.flash.assert_message(flash_message)
+        flavor = self.instantiate(name, provider)
+
+        if wait:
+            wait_for(lambda: flavor.exists, delay=50, timeout=600, fail_func=flavor.refresh,
```
mshriver: Though this page performance is poor, I think this delay needs to be lowered as much as possible. Is there a reason for it being as high as 50s? Can it be called every 5 seconds instead? We don't want to wait 50s if we could have waited 15

cfme/tests/openstack/cloud/test_flavors.py
```diff
@@ -0,0 +1,67 @@
...
+    zero_disk_flavor = provider.appliance.collections.cloud_flavors.instantiate(
+        api_flv.name, provider)
+
+    yield zero_disk_flavor
+
+    if zero_disk_flavor.exists:
+        zero_disk_flavor.delete(wait=False)
+
+
+def test_create_flavor(flavor):
```
mshriver: These should all be combined into test_flavor_crud(), such that it can be parametrized on various flavor spec combinations.

Here is where UI flash messages should be asserted against after creating, deleting, updating a flavor.

cfme/tests/openstack/cloud/test_flavors.py
```diff
@@ -0,0 +1,67 @@
...
+    assert flavor.exists
+
+
+def test_delete_flavor(flavor):
+    flavor.delete()
+    assert not flavor.exists
+
+
+def test_create_zero_disk_flavor(zero_disk_flavor):
```
mshriver: I don't think this is a valid creation test, as the fixture is creating the flavor on the provider directly via its API, which has nothing to do with MIQ.

This is really a test of provider refresh with a zero_disk_flavor, and a test of deleting through UI.

Again this should all be one test, not a separate create/delete test pair.

cfme/tests/openstack/cloud/test_flavors.py
```diff
@@ -0,0 +1,67 @@
...
+RAM = 64
+VCPUS = 1
+SWAP_SIZE = 1
+RXTX = 1
+ZERO_DISK_SIZE = 0
+
+
+@pytest.fixture(scope='function')
+def flavor(appliance, provider):
+    collection = appliance.collections.cloud_flavors
```
mshriver: please put this on one line since the args are split anyway, `flavor = appliance.collections.cloud_flavors.create(` will fit on 100 chars.

dajoRH:
Lint report for commit 9b19fa7981a52583aeb9307568e99ad02d8116b6:

`cfme/cloud/instance/openstack.py`:
- :red_circle: Line 12:1: **F401** *'cfme.utils.wait.wait_for' imported but unused*
- :red_circle: Line 73:9: **F841** *local variable 'current_flavor' is assigned to but never used*

**No commit flaws detected**.

*CFME QE Bot*

cfme/cloud/flavor.py
```diff
@@ -98,6 +103,38 @@ def is_displayed(self):
...
+    @property
+    def is_displayed(self):
+        expected_title = "Add a new Flavor"
+        return (
+            self.in_availability_zones and
+            self.entities.title.text == expected_title and
+            self.entities.breadcrumb.active_location == expected_title)
+
+    entities = View.nested(FlavorAddEntities)
+    form = View.nested(FlavorAddForm)
```
mshriver: :tada:

cfme/cloud/flavor.py
```diff
@@ -108,11 +145,75 @@ class Flavor(BaseEntity, Taggable):
...
     provider = attr.ib()
 
+    def delete(self, cancel=False):
+        """Delete current falvor """
+        view = navigate_to(self, 'Details')
+        view.toolbar.configuration.item_select('Remove Flavor', handle_alert=not cancel)
+        view.flash.assert_no_error()
+
+    @property
+    def vcpu(self):
```
mshriver: Bit of a miscommunication here I think.

So when creating a flavor in the collection, you pass these arguments: `ram`, `vcpus`, `disk`, `swap`, `rxtx`, `is_public`.

These are attributes of the flavor entity that you are modeling, and creating in the UI.

We don't want our model of the flavor entity to return the contents of the UI in its properties.
We want to use that model to CRUD the entity in MIQ, and then validate that the UI fields for the entity match the attributes of the model we tried to CRUD.

`ram`, `vcpus`, `disk`, `swap`, `rxtx`, `is_public` should all be class attributes, defined using `attr.ib()`, with defaults where necessary (is_public).

Then we can verify that `test_flavor.ram == view.entities.properties.get_text_of('Memory').split()[0]`

cfme/tests/openstack/cloud/test_flavors.py
```diff
@@ -0,0 +1,142 @@
...
+        )
+
+    except KeyError:
+        # some yaml value wasn't found
+        pytest.skip('Unable to find an image map in provider "{}" provisioning data: {}'
+                    .format(provider, prov_data))
+
+    yield instance
+    try:
+        instance.cleanup_on_provider()
```
mshriver: No need for try/except here, cleanup_on_provider checks if it exists, and if it fails to delete we want to see the teardown failure.

cfme/tests/openstack/cloud/test_flavors.py
```diff
@@ -0,0 +1,142 @@
...
+
+    flavors = [f.name for f in new_instance.provider.mgmt.api.flavors.list()]
+    flavor = view.entities.summary('Relationships').get_text_of('Flavor')
+    soft_assert(flavor in flavors)
+
+    # Assert other relationships in a loop
+    props = [('Availability Zone', 'availability_zone'),
+             ('Cloud Tenants', 'cloud_tenant')]
+
+    if current_version() >= '5.7':
```
mshriver: We support CFME 59+ at the moment, so this doesn't need to be version dependent and can always be included.

cfme/tests/openstack/cloud/test_flavors.py
```diff
@@ -0,0 +1,142 @@
...
+
+    vm_tmplt = view.entities.summary('Relationships').get_text_of('VM Template')
+    soft_assert(vm_tmplt == prov_data['image']['name'])
+
+    flavors = [f.name for f in new_instance.provider.mgmt.api.flavors.list()]
+    flavor = view.entities.summary('Relationships').get_text_of('Flavor')
+    soft_assert(flavor in flavors)
+
+    # Assert other relationships in a loop
+    props = [('Availability Zone', 'availability_zone'),
```
mshriver: **Optional**

Widgetastic has a util function you can import to do this for you - just define the UI string (p = 'Availability Zone') and then use `widgetastic.utils.attributize_string(p)` to get `availability_zone`.

This means you don't have to define tuples and have `get_text_of(p[0])` and `prov_data[p[1]]`, you can just define strings and have `get_text_of(p)` and `prov_data.get(attributize_string(p))`

mshriver: https://github.com/RedHatQE/widgetastic.core/blob/master/src/widgetastic/utils.py#L493

mshriver: This wouldn't work for Virtual Private Cloud though, so, ignore me :)

cfme/tests/openstack/cloud/test_flavors.py
```diff
@@ -0,0 +1,142 @@
...
+DISK_SIZE = 1
+RAM = 64
+VCPUS = 1
+SWAP_SIZE = 1
+RXTX = 1
+ZERO_DISK_SIZE = 0
+
+
+@pytest.fixture(scope='function')
+def flavor(appliance, provider):
```
mshriver: Where is this fixture used? Is it for future test cases?

cfme/tests/openstack/cloud/test_flavors.py
```diff
@@ -0,0 +1,142 @@
...
+                               rxtx=RXTX)
+
+    view = appliance.browser.create_view(navigator.get_class(collection, 'All').VIEW)
+    view.flash.assert_success_message(
+        'Add of Flavor "{}" was successfully initialized.'.format(flavor.name))
+
+    wait_for(lambda: flavor.exists, delay=5, timeout=600, fail_func=flavor.refresh,
+             message='Wait for flavor to appear')
+
+    assert flavor.exists
```
mshriver: Note wait_for returns a tuple of the result of the function from the last call, and the time waited. You can capture that tuple and use it instead of calling exists again:

```
flavor_exists, _ = wait_for(<etc>)
assert flavor_exists
```

But its kind of a moot assert, because if it is False, wait_for would have timed out.

cfme/tests/openstack/cloud/test_flavors.py
jadhaj: It is not used, so maybe better off be removed

cfme/cloud/flavor.py
jadhaj: Sorry, what is wrong here ?

mshriver: nothing! :) Like the structure, thanks!

dajoRH:
Lint report for commit a0392e713077648e300faceef128044767f42707:

`cfme/cloud/flavor.py`:
- :bangbang: Line 200:49: **E231** *missing whitespace after ','*

`cfme/tests/openstack/cloud/test_flavors.py`:
- :heavy_exclamation_mark: Line 120:1: **W391** *blank line at end of file*

**No commit flaws detected**.

*CFME QE Bot*

cfme/cloud/flavor.py
```diff
@@ -107,12 +144,62 @@ class Flavor(BaseEntity, Taggable):
 
     name = attr.ib()
     provider = attr.ib()
+    ram = attr.ib()
```
mshriver: These are now required attributes, and the test is hitting a TypeError because its instantiating with only name/provider.

These attributes should be optional so we can instantiate with just name and provider.

```
provider = OpenStackProvider(endpoints={'default': <cfme.infrastructure.provider.openstac...ne, keystone_v3_domain_id='default', tenant_mapping=False, infra_provider=None)

    def zero_disk_flavor(provider):
        api_flv = provider.mgmt.api.flavors.create(fauxfactory.gen_alpha(), RAM, VCPUS, ZERO_DISK_SIZE)
        wait_for(provider.is_refreshed, func_kwargs=dict(refresh_delta=10), timeout=600)
        zero_disk_flavor = provider.appliance.collections.cloud_flavors.instantiate(
>           api_flv.name, provider)

cfme/tests/openstack/cloud/test_flavors.py:31: 
cfme/modeling/base.py:119: in instantiate
    return self.ENTITY.from_collection(self, *args, **kwargs)

cls = <class 'cfme.cloud.flavor.Flavor'>
collection = FlavorCollection(filters={})
k = ('KdDEpdUHXb', OpenStackProvider(endpoints={'default': <cfme.infrastructure.provider.openstac...ne, keystone_v3_domain_id='default', tenant_mapping=False, infra_provider=None))
kw = {}

    @classmethod
    def from_collection(cls, collection, *k, **kw):
>       return cls(collection, *k, **kw)
E       TypeError: __init__() takes at least 9 arguments (4 given)

cfme/modeling/base.py:152: TypeError
```

jadhaj: How could they be made optional ?
might be with default=None

jawatts: Just like this ```attr.ib(default=None)```

cfme/tests/openstack/cloud/test_flavors.py
```diff
@@ -0,0 +1,119 @@
...
+             ('Virtual Private Cloud', 'cloud_network')]
+
+    for p in props:
+        v = view.entities.summary('Relationships').get_text_of(p[0])
+        soft_assert(v == prov_data[p[1]])
+
+
+def test_flavor_crud(appliance, provider):
+    collection = appliance.collections.cloud_flavors
+    flavor = collection.create(name=fauxfactory.gen_alpha(),
```
mshriver: I think it would be good to build some cleanup into this test by using `request.addfinalizer` method call or decorator.

After calling create, we should register a delete method call, so that any subsequent failures of flash message asserts, exists wait_for calls don't prevent the flavor from getting deleted as cleanup for the test.

cfme/tests/openstack/cloud/test_flavors.py
```diff
@@ -0,0 +1,125 @@
...
+
+    flavor.delete()
+    view = appliance.browser.create_view(navigator.get_class(collection, 'All').VIEW)
+    view.flash.assert_success_message(
+        'Delete of Flavor "{}" was successfully initiated.'.format(flavor.name))
+
+    wait_for(lambda: not flavor.exists, delay=5, timeout=600, fail_func=flavor.refresh,
+             message='Wait for flavor to appear')
+
+    @request.addfinalizer
```
mshriver: yep - just the kind of thing I was looking to see for cleanup!

Now, its placement in the test function matters too - Please move this up to right after the call to `collection.create`, before the `view` is even set in order to check flash messages.

Defining the method there won't execute it right away - `request` will just register the method and save it for the teardown phase of test execution.  Then it'll get run after the test itself finishes (pretty much where you've defined it here).  With the method defined at the bottom, if say the call to `appliance.browser.create_view` results in an exception, the cleanup method won't be registered yet, and won't run - leaving a new 'test' flavor behind on the provider.

mshriver:
PRT failed because the image needs updated for sprout checkouts.

PR #7667, opened 2018-08-19
[1LP][RFR] Add test for checking for typos in metrics endpoint type

shalomnaim1:
{{ pytest : cfme/tests/containers/test_basic_metrics.py::test_typo_in_metrics_endpoint_type --use-provider ocp-39-hawk }}

### DIFF
--- a/cfme/tests/containers/test_basic_metrics.py
+++ b/cfme/tests/containers/test_basic_metrics.py
@@ -120,8 +120,8 @@ def test_flash_msg_not_contains_html_tags(provider):
 
 
 def test_typo_in_metrics_endpoint_type(provider):
-        view = navigate_to(provider, "Details")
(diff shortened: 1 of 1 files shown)


shalomnaim1:
@quarckster Done

cfme/tests/containers/test_basic_metrics.py
```diff
@@ -117,3 +117,15 @@ def test_flash_msg_not_contains_html_tags(provider):
         is_translated_to_html = True
 
     assert is_translated_to_html, "Flash massage contains HTML tags"
+
+
+def test_typo_in_metrics_endpoint_type(provider):
+    """
+    This test based on bz1436698
```
jawatts: **Question**: Is this the correct bug?  It looks like this bug has to do with a report header being wrong

PR #7671, opened 2018-08-20
[WIP] Add ad-hoc test support for node object

shalomnaim1:
Add parametrization for ad hoc metrics to support testing for both provider and node objects

{{ pytest : cfme/tests/containers/test_ad_hoc_metrics.py --use-provider ocp-39-hawk }}

### DIFF
--- a/cfme/containers/node.py
+++ b/cfme/containers/node.py
@@ -11,7 +11,7 @@ from cfme.common import Taggable, TagPageView, PolicyProfileAssignable
 from cfme.common.vm_console import ConsoleMixin
 from cfme.containers.provider import (Labelable,
     ContainerObjectAllBaseView, LoggingableView, ContainerObjectDetailsBaseView,
-    GetRandomInstancesMixin)
+    GetRandomInstancesMixin, AdHocMetricsView)
 from cfme.modeling.base import BaseCollection, BaseEntity
 from cfme.utils.appliance.implementations.ui import (CFMENavigateStep, navigator,
                                                      navigate_to)
@@ -190,5 +190,13 @@ class Timelines(CFMENavigateStep):
         """Navigate to the Timelines page"""
         self.prerequisite_view.toolbar.monitoring.item_select('Timelines')
 
-# TODO Need Ad hoc Metrics
+
+@navigator.register(Node, 'AdHoc')
+class AdHocMain(CFMENavigateStep):
+    VIEW = AdHocMetricsView
+    prerequisite = NavigateToSibling('Details')
+
+    def step(self):
+        self.prerequisite_view.toolbar.monitoring.item_select('Ad hoc Metrics')
+
 # TODO Need External Logging
--- a/cfme/tests/containers/test_ad_hoc_metrics.py
+++ b/cfme/tests/containers/test_ad_hoc_metrics.py
@@ -2,7 +2,8 @@ import pytest
 import requests
 
 from cfme.utils.log import logger
-from cfme.containers.provider import ContainersProvider
+from cfme.containers.provider import ContainersProvider, ContainersProviderCollection
+from cfme.containers.node import NodeCollection
 from cfme.markers.env_markers.provider import providers
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.providers import ProviderFilter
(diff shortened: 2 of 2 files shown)


cfme/tests/containers/test_ad_hoc_metrics.py
```diff
@@ -2,7 +2,8 @@
 import requests
 
 from cfme.utils.log import logger
```
jawatts: Please put these in alphabetical order.

cfme/tests/containers/test_ad_hoc_metrics.py
```diff
@@ -16,6 +17,13 @@
                          scope='function')
 ]
 
+TEST_ITEMS = [ContainersProviderCollection, NodeCollection]
```
jawatts: I am thinking to change this to '''TEST_ITEMS = ['containers_providers', 'container_nodes']```

cfme/tests/containers/test_ad_hoc_metrics.py
```diff
@@ -16,6 +17,13 @@
...
+
+
+def get_test_object(appliance, provider, test_item):
+    return (
+        provider if test_item is ContainersProviderCollection else test_item(appliance).all().pop())
```
jawatts: And instead here:
```return(provider if test_item == 'containers_providers' else getattr(appliance.collections, test_item).all().pop())```

What do you think? We can get rid of those imports.

File: cfme/tests/containers/test_ad_hoc_metrics.py (hunk 2 above)
Comment (shalomnaim1) [reply cfme/tests/containers/test_ad_hoc_metrics.py]: I need the collection objects, is ```container_nodes``` and ```containers_providers``` is a collection object?

File: cfme/tests/containers/test_ad_hoc_metrics.py (hunk 3 above)
Comment (shalomnaim1) [reply cfme/tests/containers/test_ad_hoc_metrics.py]: @jawatts Personally, I prefer the long line, but I will change this if you want...

Comment (mshriver) [reply cfme/tests/containers/test_ad_hoc_metrics.py]: @shalomnaim1 @jawatts the important distinction here is the instantiation of the collection class. Collection classes should not be instantiated directly through `__new__`, but with a call to `appliance.collections` - in this case through `getattr()` since the collection name (the attribute of `appliance.collections`) is in a variable.

File: cfme/tests/containers/test_ad_hoc_metrics.py (hunk 2 above)
Comment (mshriver) [reply cfme/tests/containers/test_ad_hoc_metrics.py]: Its not a collection object, but combined with @jawatts comment below, its the name of the collection that you want to get from the appliance.

Comment (shalomnaim1) [reply cfme/tests/containers/test_ad_hoc_metrics.py]: @mshriver amm interesting, you suggest to have the name of the required collection, and use ```getattr(app.collections, <collection name>)``` to actually get it?

Comment (shalomnaim1) [reply cfme/tests/containers/test_ad_hoc_metrics.py]: This is how the function ```get_test_object``` function looks like:

```python
TEST_ITEMS = ["containers_providers", "container_nodes"]


def get_test_object(appliance, provider, test_item):
    test_item = getattr(appliance.collections, test_item)
    return (
        provider if test_item is ContainersProviderCollection else test_item(appliance).all().pop())
```

mshriver: @shalomnaim1  close - but  the call to instantiate test_item and pass it appliance `test_item(appliance)` doesn't look right at all, as appliance.collections.test_item will return you a collection instance, not a collection class.

cfme/containers/node.py
```diff
@@ -190,5 +190,13 @@ def step(self):
         """Navigate to the Timelines page"""
         self.prerequisite_view.toolbar.monitoring.item_select('Timelines')
 
-# TODO Need Ad hoc Metrics
+
+@navigator.register(Node, 'AdHoc')
```
jawatts: **Question**: Thoughts on making this ```AdHocMetrics```? I think being more specific here would be better. I know it would also have to change for https://github.com/ManageIQ/integration_tests/blob/master/cfme/containers/provider/__init__.py#L472

cfme/containers/node.py
```diff
@@ -190,5 +190,13 @@ def step(self):
...
+class AdHocMain(CFMENavigateStep):
```
jawatts: **Question**: What do you think about calling this ```class AdHocMetrics``` ? Same for here: https://github.com/ManageIQ/integration_tests/blob/master/cfme/containers/provider/__init__.py#L473

cfme/containers/node.py
```diff
@@ -11,7 +11,7 @@
 from cfme.common.vm_console import ConsoleMixin
 from cfme.containers.provider import (Labelable,
     ContainerObjectAllBaseView, LoggingableView, ContainerObjectDetailsBaseView,
-    GetRandomInstancesMixin)
+    GetRandomInstancesMixin, AdHocMetricsView)
```
jawatts: **Required**: Sort the imported names alphabetically

dajoRH:
*CFME QE Bot*

PR #7672, opened 2018-08-20
[1LP][WIPTEST] Skipping AMQP event with BZ

digitronik:
Purpose or Intent
=================
- Avoiding to fill `AMQP` even credentials in create method for OpenStack.
- We are facing AMQP credential validation BZ[0]

[0] https://bugzilla.redhat.com/show_bug.cgi?id=1618700

### DIFF
--- a/cfme/cloud/provider/openstack.py
+++ b/cfme/cloud/provider/openstack.py
@@ -7,6 +7,7 @@ from cfme.common.provider import EventsEndpoint
 from cfme.exceptions import ItemNotFound
 from cfme.infrastructure.provider.openstack_infra import RHOSEndpoint, OpenStackInfraEndpointForm
 from cfme.services.catalogs.catalog_items import OpenStackCatalogItem
+from cfme.utils.blockers import BZ
 from . import CloudProvider
 
 
@@ -85,9 +86,10 @@ class OpenStackProvider(CloudProvider):
         endpoints = {}
         endpoints[RHOSEndpoint.name] = RHOSEndpoint(**prov_config['endpoints'][RHOSEndpoint.name])
 
-        endp_name = EventsEndpoint.name
-        if prov_config['endpoints'].get(endp_name):
-            endpoints[endp_name] = EventsEndpoint(**prov_config['endpoints'][endp_name])
+        if not BZ(1618700, forced_streams=["5.9", "5.10", "upstream"]).blocks:
+            endp_name = EventsEndpoint.name
+            if prov_config["endpoints"].get(endp_name):
+                endpoints[endp_name] = EventsEndpoint(**prov_config["endpoints"][endp_name])
 
         from cfme.utils.providers import get_crud
         infra_prov_key = prov_config.get('infra_provider_key')
--- a/cfme/cloud/provider/openstack.py
+++ b/cfme/cloud/provider/openstack.py
@@ -8,6 +8,7 @@ from cfme.exceptions import ItemNotFound
 from cfme.infrastructure.provider.openstack_infra import RHOSEndpoint, OpenStackInfraEndpointForm
 from cfme.services.catalogs.catalog_items import OpenStackCatalogItem
 from cfme.utils.blockers import BZ
+from cfme.utils.log import logger
 from . import CloudProvider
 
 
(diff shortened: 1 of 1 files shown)


digitronik:
travis failure not due code `INTERNALERROR`

sshveta:
@psav : please review

cfme/cloud/provider/openstack.py
```diff
@@ -85,9 +86,10 @@ def from_config(cls, prov_config, prov_key):
         endpoints = {}
         endpoints[RHOSEndpoint.name] = RHOSEndpoint(**prov_config['endpoints'][RHOSEndpoint.name])
 
-        endp_name = EventsEndpoint.name
-        if prov_config['endpoints'].get(endp_name):
-            endpoints[endp_name] = EventsEndpoint(**prov_config['endpoints'][endp_name])
+        if not BZ(1618700, forced_streams=["5.9", "5.10", "upstream"]).blocks:
```
mshriver: If we're forcing this on all three streams, it won't unblock when the BZ is fixed. Additionally this just silently does not configure events on the provider. Tests relying on this configuration will still collect, and will just 'randomly' fail at various points as the provider configuration is incomplete.

This BZ needs to be applied to the relevant tests that require events configuration and collect against RHOS providers, and skipping events setup on RHOS needs to be logged as an error for visibility.

sshveta: Yea I too discussed that with Nikhil and he said the provider gets added by this change without AMQP and hence the tests pass .

digitronik: @sshveta @mshriver: I completely disagree with `provider configuration is incomplete`

1. As per my understanding, AMQP even is completely optional, except if we want events on `AMQP`.
2. I am not able to find a single test which depends on specific AMQP event handling. Please, point out me; if any. 
2. AMQP plays an important role but not with CFME but at OpenStack side means.  OpenStack-ceilometer-notification service is running on the OpenStack controller nodes and consumes messages from the AMPQ queues to build meter and event data/samples.
3. AMQP validation broken not just now. Its was failing from many other builds and developers not that much interested in a fix. Please, have a quick look at [comment-9](https://bugzilla.redhat.com/show_bug.cgi?id=1505159).
4. Its affecting core method of integration test framework l.e. `provider create` with UI. While adding a provider to `cfme` `AMQP` is completely optional so, my mind raised some questions:
why are we adding it by default? 
Is there any problem with default `ceilometer` event?
5. When I concern this issue with OpenStack QE team members they told me that they are also testing OSP without AMQP validation. 


Mike, I don't understand, If the problem affects all streams then how can we block without mentioning all streams?

mshriver: Understood that AMQP is an optional endpoint when configuring openstack - the provider configuration is incomplete in the sense that the yaml for the provider includes an AMQP endpoint, and after configuration, that endpoint isn't present.

Noted that there are currently no tests relying on AMQP, but should someone go to write one the provider won't ever get the required endpoint configured.

At a minimum, this needs to be logged as a `warning` that we're skipping part of the configuration due to a BZ.

I think it would be good to note this BZ in the yaml, and comment out the AMQP endpoint.

mshriver:
Travis failure isn't your fault, I think this is a valid use of BZ blocker.  #7705 should resolve that.

cfme/cloud/provider/openstack.py
```diff
@@ -82,12 +84,17 @@ def deployment_helper(self, deploy_args):
...
-        endpoints = {}
-        endpoints[RHOSEndpoint.name] = RHOSEndpoint(**prov_config['endpoints'][RHOSEndpoint.name])
+        endpoints = {
+            RHOSEndpoint.name: RHOSEndpoint(**prov_config['endpoints'][RHOSEndpoint.name])
+        }
 
-        endp_name = EventsEndpoint.name
-        if prov_config['endpoints'].get(endp_name):
-            endpoints[endp_name] = EventsEndpoint(**prov_config['endpoints'][endp_name])
+        event_endpoint_config = prov_config['endpoints'].get(EventsEndpoint.name, {})
```
mshriver: @digitronik changed this up a bit to explicitly check for AMQP (yes I realize that's really the only event endpoint type :) )

Let me know what you think, I've tested it with AMQP and a bogus event stream type in yamls, it logged correctly and included the bogus endpoint in the provider object (won't configure, but just testing if a new type is added).

digitronik: @mshriver I like `logger` with `warning` :1st_place_medal:

digitronik:
@mshriver what you think, can we move forward this?

mshriver:
PRT failures unrelated.

PR #7673, opened 2018-08-20
[WIPTEST] Fix: region name for ec2

digitronik:
Purpose or Intent
=================
- Fix for region name selection problem for ec2
- For Version 5.10 "Northern" replace with "N." like US West (N. California)

{{pytest: cfme/tests/cloud/test_providers.py::test_cloud_provider_add_with_bad_credentials --use-provider ec2west -v}}


cfme/cloud/provider/ec2.py
```diff
@@ -80,11 +80,16 @@ def view_value_mapping(self):
     def from_config(cls, prov_config, prov_key, appliance=None):
         """Returns the EC" object from configuration"""
         endpoint = EC2Endpoint(**prov_config['endpoints']['default'])
+        region_name = prov_config["region_name"]
+        # Note: for Version 5.10 "Northern" replace with "N." like US West (N. California)
+        if cls.appliance.version >= "5.10" and "Northern" in region_name:
```
mshriver: Don't have to include the check for `'Northern' in region_name`, because `replace` doesn't raise any exception on no match.

cfme/cloud/provider/ec2.py
```diff
@@ -80,11 +80,16 @@ def view_value_mapping(self):
     def from_config(cls, prov_config, prov_key, appliance=None):
         """Returns the EC" object from configuration"""
         endpoint = EC2Endpoint(**prov_config['endpoints']['default'])
+        region_name = prov_config["region_name"]
```
mshriver: This seems like something that should go into `view_value_mapping`, not `from_config`, if its only impacted in 5.10 where the region is displayed in the UI.  If the change is related to how EC2 represents its region names, then we should update YAML - but this appears to be directly tied to the CFME version, not EC2 interface.

Where are we seeing the impact from this change?

digitronik: @mshriver In tests where we select `region name` like `crud` test else you can take a simple `create` method with UI.

cfme/cloud/provider/ec2.py
digitronik: we won't need an exception in conditional check.  if no match means good to go with the same name. right?

mshriver: You don't need to check that 'Northern' is in the string, before calling `string.replace('Northern', 'other')`, the call is a no-op since its not there.

The only check needed is that the appliance version is 5.10 or above, if so try to replace 'Northern' with 'N.'.

cfme/cloud/provider/ec2.py
mshriver: What are all the contexts in MIQ where this region name is now truncated to a single character?

If everywhere, then I think its appropriate to make this version differentiation here in `from_config`.  If its only during provider creation that this is truncated within the form, and elsewhere in the UI its still displayed as 'Northern California' still - then this code should go into `view_value_mapping`, and the provider object's attribute should stay as the full name.

digitronik: @mshriver: I had talk with @mmojzis  on this issue. He pointed out [PR[1]](https://github.com/ManageIQ/manageiq-providers-amazon/pull/436/files) in which devel getting correct names via `aws-sdk` for ruby.
That means, region_name of UI should be changed but I am not 100% sure on this.

I think we should go with `from_config` for now and add people to become a whistleblower if it breaks some tests.

[1] https://github.com/ManageIQ/manageiq-providers-amazon/pull/436/files

mshriver: Tested locally a bit, expanded the PRT test collection, waiting to see its results before merging.

PR #7675, opened 2018-08-20
[1LP][RFR]collection of menus

reubenvarghese1:
Purpose or Intent
=================
* Collection of Menus


cfme/intelligence/reports/menus.py
```diff
@@ -171,7 +172,15 @@ def manage_folder(self, group, folder=None):
...
-@navigator.register(ReportMenu)
+@attr.s
+class ReportMenusCollection(BaseCollection):
+    """Collection object for the
+    :py:class:'cfme.intelligence.reports.ReportMenu'."""
+
+    ENTITY = ReportMenu
+
+
+@navigator.register(ReportMenusCollection)
```
mshriver: I think this navigation class should be registered against the entity.

Note that the step method references `self.obj.group`, which will raise an AttributeError against the collection - the collection doesn't have a `group` attribute, the entity does.

entry_points.txt
```diff
@@ -67,6 +67,7 @@ infra_providers = cfme.infrastructure.provider:InfraProviderCollection
 infra_templates = cfme.infrastructure.virtual_machines:InfraTemplateCollection
 infra_vms = cfme.infrastructure.virtual_machines:InfraVmCollection
 map_tags = cfme.configure.configuration.region_settings:MapTagsCollection
+menus = cfme.intelligence.reports.menus:ReportMenusCollection
```
mshriver: More descriptive collection name please, like `report_menus` or `intel_report_menus`

cfme/intelligence/reports/menus.py
```diff
@@ -43,14 +45,13 @@ def is_displayed(self):
...
+@attr.s
+class ReportMenu(BaseEntity):
     """
         This is a fake class mainly needed for navmazing navigation.
 
     """
-    def __init__(self, appliance=None):
-        Navigatable.__init__(self, appliance)
-        self.group = None
+    group = attr.ib(default=None)
```
mshriver: I don't think this should be attr defined, as its not an option that can be passed to the class's constructor, but is set through `go_to_group` method. It should just be a class attribute.

reubenvarghese1:
@mshriver , changes made, kindly do check it out :)

cfme/tests/intelligence/reports/test_menus.py
```diff
@@ -28,8 +27,8 @@ def shuffle(l):
 
 
 @pytest.fixture(scope="function")
-def report_menus(group):
-    report_menus = menus.ReportMenu()
+def report_menus(group, appliance):
+    report_menus = appliance.collections.intel_report_menus.ReportMenu()
```
mshriver: Here we need to use the instantiate method, so that the class constructor (from attr) will get all of the attributes it needs - like appliance.

BaseEntity or BaseCollection derived classes should pretty much never be instantiated directly.

cfme/intelligence/reports/menus.py
```diff
@@ -171,6 +172,14 @@ def manage_folder(self, group, folder=None):
             view.save_button.click()
 
 
+@attr.s
+class ReportMenusCollection(BaseCollection):
+    """Collection object for the
+    :py:class:'cfme.intelligence.reports.ReportMenu'."""
```
mshriver: Please checkout napoleon/sphinx guidelines for docblocks, this should be on one line, with the second line empty. This fits in 100 chars, and should just be on a single line.

reubenvarghese1:
@mshriver Sorry for the trouble, changes made!
Btw, why is that after instantiating, we no longer are required to call the ReportMenu() method?

mshriver:
No worries, its a learning process!

Nope, the `.instantiate()` call on the collection object will return an entity object (instance).

To be clear (and a pedant, sorry) `ReportMenu` is a class, and `ReportMenu()` is a call to `ReportMenu.__new__` method, calling `ReportMenu()` a method is a bit ambiguous  or misleading ;)

reubenvarghese1:
Oh, Now it all makes sense!(haha,these tiny details can really get me confused at times )

reubenvarghese1:
@mshriver Do the new changes look good?

reubenvarghese1:
@izapolsk One of the tests, I am unable to access the logs. It would great if you could let me know the issue :)

izapolsk:
PRT looks good.

reubenvarghese1:
Great, thanks!

reubenvarghese1:
@mshriver Do these changes look good?

PR #7681, opened 2018-08-21
[1LP][RFR] Fix snapshot memory checkbox for 5.10

apagac:
This is a fix for snapshot memory checkbox in appliance version 5.10 and above.

The original checkbox from 5.9 has been replaced by a something that looks like a BootstrapSwitch, but that is only a decoy. The Wigetastic BootstrapSwitch cannot be used here, the layout of divs is different and there is no input and no id. So I created new class SnapshotMemorySwitch inheriting from Widget with some basic operations defined, also with read and fill operations, so it works like any other widget.

I've put this into widgetastic_manageiq, since I suspect this kind of switch is used only in snapshots. There also exist other examples of kind-of-a-BootstrapSwitch already in different parts of the framework.

This should fix all of the snapshot failures currently appearing in jenkins for 5.10.

{{ pytest: -v --long-running cfme/tests/infrastructure/test_snapshot.py -k test_memory_checkbox --use-provider vsphere65-nested }}

### DIFF
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -49,7 +49,7 @@ from cfme.utils.conf import cfme_data
 from cfme.utils.log import logger
 from cfme.utils.pretty import Pretty
 from cfme.utils.providers import get_crud_by_name
-from cfme.utils.version import UPSTREAM, VersionPicker
+from cfme.utils.version import UPSTREAM, Version, VersionPicker
 from cfme.utils.wait import wait_for
 from widgetastic_manageiq import (
     Accordion,
@@ -60,7 +60,8 @@ from widgetastic_manageiq import (
     Table,
     TimelinesView,
     CompareToolBarActionsView,
-    Search
+    Search,
+    SnapshotMemorySwitch
 )
 from widgetastic_manageiq.vm_reconfigure import DisksTable
 
@@ -386,7 +387,9 @@ class InfraVmSnapshotAddView(InfraVmView):
     title = Text('#explorer_title_text')
     name = TextInput('name')
     description = TextInput('description')
-    snapshot_vm_memory = Checkbox('snap_memory')
+    snapshot_vm_memory = VersionPicker({
+        Version.lowest(): Checkbox('snap_memory'),
+        '5.10': SnapshotMemorySwitch()})
     create = Button('Create')
     cancel = Button('Cancel')
 
--- a/widgetastic_manageiq/__init__.py
+++ b/widgetastic_manageiq/__init__.py
@@ -5129,7 +5129,7 @@ class MonitorStatusCard(ParametrizedView):
         return self.click_title()
 
 
-class SnapshotMemorySwitch(Widget):
+class SnapshotMemorySwitch(Widget, ClickableMixin):
     """ This replaces the old-style checkbox when working with snapshot memory.
 
     At first glance on the page it looks like a BootstrapSwitch but it really isn't.
@@ -5140,28 +5140,26 @@ class SnapshotMemorySwitch(Widget):
     ROOT = '//div[contains(@class, "bootstrap-switch-snap_memory")]'
     ON_LOCATOR = './/div[contains(@class, "bootstrap-switch-on")]'
 
-    def switch(self):
-        self.browser.click(self.__locator__)
-
     @property
     def is_set(self):
         return bool(self.browser.elements(self.ON_LOCATOR))
 
     def switch_on(self):
         if not self.is_set:
-            self.switch()
-        return True
+            self.click()
+            return True
+        else:
+            return False
 
     def switch_off(self):
         if self.is_set:
-            self.switch()
-        return True
+            self.click()
+            return True
+        else:
+            return False
 
     def fill(self, value):
-        if value:
-            self.switch_on()
-        else:
-            self.switch_off()
+        return (self.switch_on if value else self.switch_off)()
 
     def read(self):
         return self.is_set
--- a/widgetastic_manageiq/__init__.py
+++ b/widgetastic_manageiq/__init__.py
@@ -5127,3 +5127,41 @@ class MonitorStatusCard(ParametrizedView):
 
     def click(self):
         return self.click_title()
+
+
+class SnapshotMemorySwitch(Widget):
+    """ This replaces the old-style checkbox when working with snapshot memory.
+
+    At first glance on the page it looks like a BootstrapSwitch but it really isn't.
+    Basic switch controls are implemented here along with fill and read methods,
+    so this should be usable as any other widget.
+    """
+
+    ROOT = '//div[contains(@class, "bootstrap-switch-snap_memory")]'
+    ON_LOCATOR = './/div[contains(@class, "bootstrap-switch-on")]'
+
+    def switch(self):
+        self.browser.click(self.__locator__)
+
+    @property
+    def is_set(self):
+        return bool(self.browser.elements(self.ON_LOCATOR))
+
+    def switch_on(self):
+        if not self.is_set:
+            self.switch()
+        return True
+
+    def switch_off(self):
+        if self.is_set:
+            self.switch()
+        return True
+
+    def fill(self, value):
+        if value:
+            self.switch_on()
+        else:
+            self.switch_off()
+
+    def read(self):
+        return self.is_set


apagac:
Currently fails on 5.10 during test vm provision. Have no idea why, can't reproduce locally.

widgetastic_manageiq/__init__.py
```diff
@@ -33,6 +33,57 @@
...
+    def is_set(self):
+        if self.browser.selenium.find_elements_by_xpath(self.ON_LOCATOR):
+            return True
+        return False
+
+    def switch_on(self):
+        try:
+            if not self.is_set:
+                self.switch()
+        except Exception as e:
```
jan-zmeskal: Does this `try` block actually do anything?

dajoRH:
*CFME QE Bot*

widgetastic_manageiq/__init__.py
apagac: Sure. If the switch is set (meaning it's ON), it leaves it as is. If the switch is not set (meaning it's OFF), it sets the switch to ON.
So after running the switch_on() function, the switch will always be ON.

dajoRH:
*CFME QE Bot*

mshriver:
@apagac please rebase, RHV team has identified this failure as a blocker for their QE.

lina-nikiforova:
@apagac can you please fix the Travis error? As @mshriver mentioned it's a blocker for RHV testing and we need this PR to be merged

dajoRH:
*CFME QE Bot*

dajoRH:
Lint report for commit 365d250ad132bb6c23b65352f3aa947ef0729de2:

- :bangbang: Line 750:101: **E501** *line too long (104 > 100 characters)*

- :bangbang: Line 5139:101: **E501** *line too long (144 > 100 characters)*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
Lint report for commit 2fefe44dadcc6c947abf2ccbfe0272b903ab351c:

- :bangbang: Line 750:101: **E501** *line too long (104 > 100 characters)*

- :bangbang: Line 5139:101: **E501** *line too long (144 > 100 characters)*

**No commit flaws detected**.

*CFME QE Bot*

cfme/infrastructure/virtual_machines.py
```diff
@@ -8,17 +8,27 @@
 import attr
 import fauxfactory
 import re
-from navmazing import NavigateToSibling, NavigationDestinationNotFound, NavigateToAttribute
+from navmazing import (
```
mshriver: This look like you're applying black formatting to the file while you're making changes. Our pre-commit is currently setup to exclude the cfme/ folder from black formatting. We'll apply it all at once in a PR that doesn't try to change functionality, just formatting.

Trying to apply this formatting here makes it much more difficult to identify and review your actual functional changes to the module, and I think it would be better to use the pre-commit config that comes with integration_tests, and not apply it individually like this.

apagac: Thanks for the explanation Mike. I was just playing around trying to figure things out.
Will change this PR as suggested.

mshriver: No worries - I like how you're being pro-active!  We'll be able to get this reviewed and merged quicker with focused changes though :)

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -42,7 +42,7 @@ def provision_vm(provider, template):
                                                               template.name)
 
     if not provider.mgmt.does_vm_exist(vm_name):
-        vm.create_on_provider(find_in_cfme=True, allow_skip="default")
+        vm.create_on_provider(find_in_cfme=False, allow_skip="default")
```
jawatts: **Question** Why did you decide to set this to false? Would it not be good to verify that the VM exists before continuing?

widgetastic_manageiq/__init__.py
```diff
@@ -61,6 +61,58 @@
...
+
+    def __locator__(self):
+        return self.LOCATOR
+
+    def switch(self):
+        self.browser.click(self.LOCATOR)
+
+    @property
+    def is_set(self):
+        if self.browser.selenium.find_elements_by_xpath(self.ON_LOCATOR):
```
mshriver: We don't want to make any direct calls to selenium unless absolutely necessary, and I don't think it is here. Why not use `self.browser.elements(self.ON_LOCATOR)`?

widgetastic_manageiq/__init__.py
```diff
@@ -61,6 +61,58 @@
 from cfme.exceptions import ItemNotFound
 
 
+class SnapshotMemorySwitch(Widget):
```
mshriver: Please move the widget to the end of the module.

widgetastic_manageiq/__init__.py
```diff
@@ -61,6 +61,58 @@
...
+    """ This replaces the old-style checkbox when working with snapshot memory.
+
+    At first glance on the page it looks like a BootstrapSwitch but it really isn't.
+    Basic switch controls are implemented here along with fill and read methods,
+    so this should be usable as any other widget.
+    """
+
+    LOCATOR = u'//div[contains(@class, "bootstrap-switch-snap_memory")]'
```
mshriver: You should use `ROOT` here, and remove your `__locator__` property. Widgetastic will create it automatically when `ROOT` is used.

widgetastic_manageiq/__init__.py
```diff
@@ -61,6 +61,58 @@
...
+    ON_LOCATOR = u'//div[contains(@class, "bootstrap-switch-on")]'
```
mshriver: This should almost certainly be a relative xpath expression: `'.//div[]'`

widgetastic_manageiq/__init__.py
```diff
@@ -61,6 +61,58 @@
...
+
+    def __init__(self, parent, logger=None):
```
mshriver: Unnecessary constructor since there are no statements made. If this constructor were to be included, it should make the call to the parent class `__init__` through `super()`, not directly through the class name.

widgetastic_manageiq/__init__.py
mshriver: Additionally, since the element isn't actually used, and you aren't catching exceptions when the element isn't present, This can just be `return self.browser.elements(self.ON_LOCATOR)`

widgetastic_manageiq/__init__.py
mshriver: @apagac I don't think that is quite what @jan-zmeskal  was getting at with his comment.

You have a code block of:
```
try:
    <some_statement>
except:
    raise
```

This is equivalent to:
```
<some_statement>
```

There is nothing done with the exception is caught here, and this try/except block is extraneous.  Is there an exception you were encountering during testing or expecting from the check of is_set or call to self.switch?  

If so those exceptions should be specifically listed here, and then there should at the very least be logging, if not modification of the exception or changes in the functions behavior due to the exception being encountered.

apagac: There indeed was an exception I was hitting back in the day I wrote this. I'll check for it and fix.

cfme/tests/infrastructure/test_snapshot.py
apagac: Originally, this was a workaround for navigation errors. Not relevant anymore.

apagac:
Updated according to comments. Tested locally and by PRT.

Removed unnecessary workarounds that were in place due to errors being encountered in the time of writing this code.

widgetastic_manageiq/__init__.py
```diff
@@ -5127,3 +5127,41 @@ def click_body_action(self):
...
+
+    At first glance on the page it looks like a BootstrapSwitch but it really isn't.
+    Basic switch controls are implemented here along with fill and read methods,
+    so this should be usable as any other widget.
+    """
+
+    ROOT = '//div[contains(@class, "bootstrap-switch-snap_memory")]'
+    ON_LOCATOR = './/div[contains(@class, "bootstrap-switch-on")]'
+
+    def switch(self):
```
mshriver: I don't think there's a real need for `switch`, just `switch_on` and `switch_off`.  `switch` itself, in that context, becomes a bit ambiguous. 

Please remove this method, and include `ClickableMixin` in your class declaration.  The `switch_on` and `switch_off` methods can then call `self.click()`, which I think is a more explicit action than `switch` alone.

PR #7682, opened 2018-08-21
[1LP][RFR] Update httpd workaround for 59z

mshriver:
Apply the httpd workaround on all streams

### DIFF
--- a/cfme/utils/appliance/__init__.py
+++ b/cfme/utils/appliance/__init__.py
@@ -468,7 +468,7 @@ class IPAppliance(object):
     def fix_httpd_issue(self, log_callback=None):
         # This is workaround for issue when httpd cannot start after restart because
         # it didn't properly clean up semaphores
-        if self.version >= '5.10' and not self.is_pod:
+        if not self.is_pod:
             with self.ssh_client as ssh:
                 filename = 'httpd.service'
                 src = os.path.join('/usr/lib/systemd/system', filename)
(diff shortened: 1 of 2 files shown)


cfme/utils/appliance/__init__.py
```diff
@@ -482,8 +483,8 @@ def fix_httpd_issue(self, log_callback=None):
         """
                 full_cmd = """echo '{cmd}' >> {dst}""".format(cmd=exec_pre, dst=dst)
                 assert ssh.run_command(full_cmd).success
-                assert ssh.run_command('systemctl daemon-reload').success
-                assert ssh.run_command('systemctl restart httpd').success
+                self.daemon_reload.restart()
```
mshriver: The SystemdService method verifies the RC on restart and raises an exception on failure.

izapolsk: why restart ? it will cause error.  I'd add reload method to SystemdService class instead

mshriver: OH sorry I totally mis-read this.

cfme/utils/appliance/__init__.py
```diff
@@ -482,9 +482,9 @@ def fix_httpd_issue(self, log_callback=None):
         """
                 full_cmd = """echo '{cmd}' >> {dst}""".format(cmd=exec_pre, dst=dst)
                 assert ssh.run_command(full_cmd).success
-                assert ssh.run_command('systemctl daemon-reload').success
-                assert ssh.run_command('systemctl restart httpd').success
-                self.wait_for_web_ui(timeout=1800, log_callback=log_callback)
+                self.httpd.daemon_reload()
+                self.httpd.restart()
+                wait_for(lambda: self.httpd.running, timeout=1800)
```
mshriver: This is run during configuration, where webui likely won't be up.

PR #7683, opened 2018-08-21
[1LP][RFR] Check that Nuage events are emitted and received

miha-plesko:
With this commit we add a test to verify that events are being recieved by MIQ.

PR #7686, opened 2018-08-21
[RFR] Don't throw away events too fast

miha-plesko:
With this commit we prevent RestEventListener from updating last
event ID too eagerly. Here's what happens otherwise:

```
1 for exp_event in self._events_to_listen:
2  matched_events = self.get_next_portion() # <-- fetch events since "last ID"
3  for event_entity in matched_events:
4    if MATCHES:
5      update "last ID" globally # <-- this is wrong!!!
```

First FOR loops over expected events (line 1). For each of expected events
it fetches events from API, but only with event ID greater than "last ID"
(line 2). Then it checks if event matches current expectation (line 4) and
it updates the "last ID" for everyone (line 5). BUUUM!

When topmost FOR loop iterates over 2nd, 3rd, 4th,... event expectation
it only fatches events since just updated "last ID". And if it happens that
the event we're expecting occured before that "last ID", we will never
even attempt to match it.

As a result, we observe random behavior. Somethimes test fails, sometimes it
works. It depends on whether we're lucky to be receiving events in same order
with same speed as we're consuming them.

/cc @mshriver @quarckster @ohiliazo

### DIFF
--- a/cfme/utils/events.py
+++ b/cfme/utils/events.py
@@ -183,7 +183,8 @@ class RestEventListener(Thread):
                 exp_event = {'event': evt,
                              'callback': callback,
                              'matched_events': [],
-                             'first_event': first_event}
+                             'first_event': first_event,
+                             'last_id': None}
                 self._events_to_listen.append(exp_event)
                 logger.info("event {} is added to listening queue.".format(evt))
             else:
@@ -220,7 +221,7 @@ class RestEventListener(Thread):
                 if exp_event['first_event'] and len(exp_event['matched_events']):
                     continue
 
-                matched_events = self.get_next_portion(exp_event['event'])
+                matched_events = self.get_next_portion(exp_event['event'], exp_event['last_id'])
 
                 if not matched_events:
                     continue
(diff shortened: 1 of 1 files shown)


miha-plesko:
Hey guys, took me entire day to figure this one out. Kindly ask for opinion. Wondering if no one ever used `register_event` fixture with more than one event? And then we come and use it with 12 event expectations. Ain't I the lucky one, hahaha.

mshriver:
Believe I understand the context of the randomized behavior/failure that you're seeing, but its not clear to me how this is safe to remove this completely from processing.

This looks like an intended use of the `RestEventListener.set_last_record` method, which will set it based on the event streams.

Once the outermost for loop has completed, won't we want the _last_processed_id attribute updated at that point?

dajoRH:
Lint report for commit 26a4b531f254d7b8fb68110cd90a7132f5385b49:

`cfme/tests/networks/nuage/test_events_are_triggered.py`:
- :red_circle: Line 3:1: **F401** *'time' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

miha-plesko:
Thanks for pointing out @mshriver, you are right. What we really need here is that "last ID" is set per expectation that we're checking, not just globally for all events. Different expectations captures different events (because they apply different API filters) so it makes no sense to then mix IDs.

Just pushed such approach. So what we do now is we memorize global "last ID" which points to the time when event catcher fixture was started, and we never modify it later. Expectations then start by using this value. As soon as they obtain some events and check them, they set local "last ID" for themselves so that in next tick they don't check same events again.

Looking forward to your opinion!

izapolsk:
@miha-plesko, you're right this is bug introduced when we were replacing db based event listener with rest one.  I don't think we have to keep track of last_event_id for each expected event. I'd rather obtain last event id and limited queries by that id.

miha-plesko:
@izapolsk thanks for commenting. Imagine there are three events in DB:

- event 1 (id=1) (evt_type=a)
- event 2 (id=2) (evt_type=b)
- event 3 (id=3) (evt_type=c)

I'm thinking what if first expected event is "c". So it fetches all three events via API and sets "last ID" to 3. In this case when expected event "a" will want to check events, it'll only fetch events with ID > 3. So it'll never get any event really ---> and the test will fail saying there was no event of type "a".

Right?

mshriver:
@miha-plesko please restrict PRT to relevant tests - the latest update with  being able to specify the ID when registering events to listen to, and the process loop checking against it looks good.

miha-plesko:
@mshriver I'm not sure what you mean by "restrict PRT to relevant tests"? So the test that I'm trying to run, and does not run without this patch is here https://github.com/ManageIQ/integration_tests/pull/7683.

With respect to comment from @izapolsk (see above) I've implemented alternative solution. So now we have two PRs open to fix this very same issue:

a) https://github.com/ManageIQ/integration_tests/pull/7686 (stores "last ID" per expectation)
b) https://github.com/ManageIQ/integration_tests/pull/7706 (stores a global "last ID" in a way that most restrictive expectations still get all the events while most permisive expectations might get events more than once).

Both work for me, I don't really mind any of those. Can we please pick one and merge it?

mshriver:
Closing in favor of 7706

PR #7697, opened 2018-08-22
[1LP][RFR] Use version.lowest instead of 5.9

mshriver:
@quarckster I'm thinking in the future when we remove deprecated CFME versions from the pickers, we replace with Version.lowest() instead of a specific version - When an older appliance version is run against this, it ends up getting None from the `.pick()` call and fails in odd ways during collection.

In this case if you're running against 5.8, you get a 'wrong' version of the role_dict, but that's more expected as you're running against unsupported version.

Thoughts? Not something I feel too strongly about, but hit it while template_tester was running against an older appliance version, from the master branch.


quarckster:
@mshriver I'm totally ok with that.

PR #7700, opened 2018-08-23
[1LP][RFR][NOTEST] Add BZ wrapper for test host credentials

rhrazdil:
Wrapping host creds validation tests because of [this BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1619626).

The problem is that when SSH keys change on hosts, CFME fails to accept the new one. Once the bug is fixed, the next step would be to update the tests so that if an alert informing about new ssh keys shows up, 'OK' button is clicked and new keys accepted.

### DIFF
--- a/cfme/tests/infrastructure/test_individual_host_creds.py
+++ b/cfme/tests/infrastructure/test_individual_host_creds.py
@@ -47,7 +47,9 @@ def get_host_data_by_name(provider_key, host_name):
                      unblock=lambda creds: creds == 'default'),
                   BZ(1584280, forced_streams=['5.9'],
                      unblock=lambda provider, creds: not provider.one_of(RHEVMProvider) or
-                     creds != 'web_services')])
+                     creds != 'web_services'),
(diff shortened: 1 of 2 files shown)


cfme/tests/infrastructure/test_individual_host_creds.py
```diff
@@ -105,7 +107,9 @@ def _refresh():
                   BZ(1584246, forced_streams=['5.8', '5.9'],
                      unblock=lambda provider: not provider.one_of(VMwareProvider)),
                   BZ(1584261, forced_streams=['5.8'],
-                     unblock=lambda creds: creds == 'default')])
+                     unblock=lambda creds: creds == 'default'),
```
quarckster: **REQUIRED**
Can you please remove obsolete blockers here?

rhrazdil: @quarckster I'm don't see any obsolete blockers, all of them are NEW except the new one which is already ASSIGNED.

PR #7706, opened 2018-08-24
[RFR] Don't throw away events too fast (alternative solution)

miha-plesko:
With this commit we prevent RestEventListener from updating last
event ID too eagerly. Here's what happens otherwise:

```
1 for exp_event in self._events_to_listen:
2  matched_events = self.get_next_portion() # <-- fetch events since "last ID"
3  for event_entity in matched_events:
4    if MATCHES:
5      update "last ID" globally # <-- this is wrong!!!
```

First FOR loops over expected events (line 1). For each of expected events
it fetches events from API, but only with event ID greater than "last ID"
(line 2). Then it checks if event matches current expectation (line 4) and
it updates the "last ID" for everyone (line 5). BUUUM!

When topmost FOR loop iterates over 2nd, 3rd, 4th,... event expectation
it only fatches events since just updated "last ID". And if it happens that
the event we're expecting occured before that "last ID", we will never
even attempt to match it.

As a result, we observe random behavior. Somethimes test fails, sometimes it
works. It depends on whether we're lucky to be receiving events in same order
with same speed as we're consuming them.

Alternative solution to https://github.com/ManageIQ/integration_tests/pull/7686, but here we respect @izapolsk 's request to avoid local "current event ID"s.

/cc @izapolsk @mshriver

### DIFF
--- a/cfme/utils/events.py
+++ b/cfme/utils/events.py
@@ -214,6 +214,10 @@ class RestEventListener(Thread):
         """
         while not self._stop_event.is_set():
             sleep(1)
+
+            # Memorize how far in time each expectation came.
+            last_processed_ids = []
+
             for exp_event in self._events_to_listen:
 
                 # Skip if event has occurred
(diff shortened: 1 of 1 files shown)


miha-plesko:
@izapolsk as discussed on gitter, I've implemented a solution where we avoid storing local "last ID"s per expectation. Instead, we now memorize "how far did the most conservative expectation" come and store this as a global "last ID". Tested with https://github.com/ManageIQ/integration_tests/pull/7683 and it works as expected.

Looking forward to your opinion!

miha-plesko:
Correct me if I'm wrong but I think @izapolsk votes for this solution. @mshriver you okay with it?

miha-plesko:
Closing in favor of https://github.com/ManageIQ/integration_tests/pull/7722, thanks @izapolsk !

PR #7707, opened 2018-08-24
[RFR][NOTEST] Removed case-sensitivity of webdriver option.

sbulage:


cfme/utils/browser.py
```diff
@@ -258,7 +258,7 @@ def from_conf(cls, browser_conf):
                 browser_kwargs['desired_capabilities']['marionette'] = False
             return cls(WharfFactory(webdriver_class, browser_kwargs, wharf))
         else:
-            if webdriver_name == "Remote":
```
quarckster: **REQUIRED**
Please describe how did you face this issue?

sbulage: While explaining environment to Summer interns, couldn't open the browser in Selenium server. Later found that I have made typo at `webdriver_name` in the environment so I wanted to incorporate these changes in our own framework.

sbulage: I will update it accordingly.

izapolsk: I'm doubtful of necessity of this feature. Possible values of that field are Firefox, Chrome and etc. Are you going to make them case insensitive as well ?
In addition, my memory reminds me that we pass this value to webdriver itself and I'm not entirely sure it supports lower case value.

sbulage: @izapolsk I have tried with `Firefox` and `Chrome` in place of `Remote`. Able to open the browser in Selenium server as well as outside of it.
I am only making `Remote` `webdriver_name` to lower case.

PR #7714, opened 2018-08-28
[1LP][RFR] Adding test_provision_cloud_init_payload for RHEVM

jan-zmeskal:
I believe the test case (and its docstring) are pretty self-explanatory. What is maybe not so clear are changes in cfme/common/vm_views.py. They are two:
- I added RadioGroup for Address Mode
- I rearranged the button definitions so they are in the same order as in UI. This can be significant because some of the "later" fields are dependent on Customize Type.

Also, what might not be very clear is that why I am not specifying Customization Template. The reason is that there is already one pre-defined (`oVirt cloud-init`) that perfectly suits this purpose.

Also, why am I using pure `paramiko` as opposed to `cfme.utils.ssh`? The reason is that I need to SSH to the provisioned VM using IPv6 and I just could not get it to work with `SSHClient` from `cfme.utils.ssh`.

Jenkins verification: https://rhv-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/cfme-5.9-rhv-4.2-integration-test-dev/187

New verification (after changes to `net_check`): https://rhv-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/cfme-5.9-rhv-4.2-integration-test-dev/202/

New verification after comments from @digitronik and @mshriver: https://rhv-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/cfme-5.10-rhv-4.2-integration-test-dev/27/console

### DIFF
--- a/cfme/tests/cloud_infra_common/test_cloud_init_provisioning.py
+++ b/cfme/tests/cloud_infra_common/test_cloud_init_provisioning.py
@@ -8,6 +8,7 @@ from cfme.cloud.provider.azure import AzureProvider
 from cfme.cloud.provider.gce import GCEProvider
 from cfme.cloud.provider.openstack import OpenStackProvider
 from cfme.infrastructure.provider import InfraProvider
+from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.infrastructure.provider.scvmm import SCVMMProvider
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.infrastructure.pxe import get_template_from_config
@@ -27,6 +28,22 @@ pytestmark = [
     pytest.mark.meta(server_roles="+automate"),
     pytest.mark.provider(gen_func=providers, filters=[pf1, pf2], scope="module")
 ]
+
+
+def find_global_ipv6(vm):
+    """
+    Find global IPv6 on a VM if present.
+
+    Args:
+        vm: InfraVm object
+
+    Returns: IPv6 as a string if found, False otherwise
+    """
+    all_ips = vm.mgmt.all_ips
+    for ip in all_ips:
+        if ':' in ip and not ip.startswith('fe80'):
+            return ip
+    return False
 
 
 @pytest.fixture(scope="module")
@@ -98,3 +115,78 @@ def test_provision_cloud_init(appliance, request, setup_provider, provider, prov
     with ssh.SSHClient(hostname=connect_ip, username=provisioning['ci-username'],
                        password=provisioning['ci-pass']) as ssh_client:
         wait_for(ssh_client.uptime, num_sec=200, handle_exception=True)
+
+
+@pytest.mark.rhv3
+@pytest.mark.uncollectif(lambda provider, appliance:
+    not provider.one_of(RHEVMProvider) or appliance.version < '5.10.0.12',
+    reason='Required customization template is only available in CFME 5.10.0.12 or greater.')
+def test_provision_cloud_init_payload(appliance, request, setup_provider, provider, provisioning,
+                                      vm_name):
+    """
+    Tests that options specified in VM provisioning dialog in UI are properly passed as a cloud-init
+    payload to the newly provisioned VM.
+
+    Metadata:
+        test_flag: cloud_init, provision
+    """
+    image = provisioning.get('ci-image', None)
+    if not image:
+        pytest.skip('No ci-image found in provider specification.')
+    note = ('Testing provisioning from image {image} to vm {vm} on provider {provider}'.format(
+        image=image, vm=vm_name, provider=provider.key))
+    logger.info(note)
+
+    ci_payload = {
+        'root_password': 'mysecret',
+        'address_mode': 'Static',
+        'hostname': 'cimachine',
+        'ip_address': '169.254.0.1',
+        'subnet_mask': '29',
+        'gateway': '169.254.0.2',
+        'dns_servers': '169.254.0.3',
+        'dns_suffixes': 'virt.lab.example.com',
+        'custom_template': {'name': 'oVirt cloud-init'}
+    }
+
+    inst_args = {
+        'request': {'notes': note},
+        'customize': {'customize_type': 'Specification'},
+        'template_name': image
+    }
+
+    inst_args['customize'].update(ci_payload)
+    logger.info('Instance args: {}'.format(inst_args))
+
+    # Provision VM
+    collection = appliance.provider_based_collection(provider)
+    instance = collection.create(vm_name, provider, form_values=inst_args)
+    request.addfinalizer(instance.cleanup_on_provider)
+    provision_request = provider.appliance.collections.requests.instantiate(vm_name,
+                                                                            partial_check=True)
+    provision_request.wait_for_request()
+
+    connect_ip = wait_for(find_global_ipv6, func_args=[instance], num_sec=600, delay=20).out
+    logger.info('Connect IP: {}'.format(connect_ip))
+
+    # Connect to the newly provisioned VM
+    with ssh.SSHClient(hostname=connect_ip,
+                       username='root',
+                       password=ci_payload['root_password']) as ssh_client:
+        # Check that correct hostname has been set
+        hostname_cmd = ssh_client.run_command('hostname')
+        assert hostname_cmd.success
+        assert hostname_cmd.output.strip() == ci_payload['hostname']
+
+        # Obtain network configuration script for eth0 and store it in a list
+        network_cfg_cmd = ssh_client.run_command('cat /etc/sysconfig/network-scripts/ifcfg-eth0')
+        assert network_cfg_cmd.success
+        config_list = network_cfg_cmd.output.split('\n')
+
+        # Compare contents of network script with cloud-init payload
+        assert 'BOOTPROTO=none' in config_list, 'Address mode was not set to static'
+        assert 'IPADDR={}'.format(ci_payload['ip_address']) in config_list
+        assert 'PREFIX={}'.format(ci_payload['subnet_mask']) in config_list
+        assert 'GATEWAY={}'.format(ci_payload['gateway']) in config_list
+        assert 'DNS1={}'.format(ci_payload['dns_servers']) in config_list
+        assert 'DOMAIN={}'.format(ci_payload['dns_suffixes']) in config_list
--- a/cfme/utils/net.py
+++ b/cfme/utils/net.py
@@ -66,8 +66,9 @@ def net_check(port, addr=None, force=False):
     if port not in _ports[addr] or force:
         # First try DNS resolution
         try:
-            addr = socket.gethostbyname(addr)
-
+            addr_info = socket.getaddrinfo(addr, port)[0]
+            sockaddr = addr_info[4]
+            addr = sockaddr[0]
             # Then try to connect to the port
             try:
                 socket.create_connection((addr, port), timeout=10)
--- a/cfme/common/vm_views.py
+++ b/cfme/common/vm_views.py
@@ -337,19 +337,21 @@ class BasicProvisionFormView(View):
     class customize(Tab):  # noqa
         TAB_NAME = 'Customize'
         # Common
-        dns_servers = Input(name='customize__dns_servers')
-        dns_suffixes = Input(name='customize__dns_suffixes')
         customize_type = BootstrapSelect('customize__sysprep_enabled')
-        specification_name = Table('//div[@id="prov_vc_div"]/table')
-        admin_username = Input(name='customize__root_username')
         root_password = Input(name='customize__root_password')
-        linux_host_name = Input(name='customize__linux_host_name')
-        linux_domain_name = Input(name='customize__linux_domain_name')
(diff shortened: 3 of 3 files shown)


cfme/tests/cloud_infra_common/test_cloud_init_provisioning.py
```diff
@@ -28,6 +30,17 @@
...
 
+def find_global_ipv6(vm):
+    """
+    Find global IPv6 on a VM if present and return it as a string. Otherwise, return False.
+    """
+    all_ips = vm.mgmt.all_ips
+    for ip in all_ips:
+        if ':' in ip and not ip.startswith('fe80'):
+            return ip
+    return False
```
rhrazdil: I would recommend to return None to be consistent with wrapanapi.

cfme/tests/cloud_infra_common/test_cloud_init_provisioning.py
```diff
@@ -96,3 +109,79 @@ def test_provision_cloud_init(appliance, request, setup_provider, provider, prov
     with ssh.SSHClient(hostname=connect_ip, username=provisioning['ci-username'],
                        password=provisioning['ci-pass']) as ssh_client:
         wait_for(ssh_client.uptime, num_sec=200, handle_exception=True)
+
+
+@pytest.mark.rhv3
+@pytest.mark.provider([RHEVMProvider], selector=LATEST, scope='function', override=True)
+@pytest.mark.uncollectif(lambda appliance: appliance.version < '5.10.0.12')
```
rhrazdil: Please, provide reason parameter for the uncollect.

dajoRH:
Lint report for commit afc2532232aa0b24a9da032dff98c7b74beb8f1f:

`cfme/tests/cloud_infra_common/test_cloud_init_provisioning.py`:
- :red_circle: Line 16:1: **F401** *'cfme.markers.env_markers.provider.LATEST' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
Lint report for commit 9b1a25093e4cb430a2df47d400041a83cc07966d:

`cfme/tests/cloud_infra_common/test_cloud_init_provisioning.py`:
- :red_circle: Line 16:1: **F401** *'cfme.markers.env_markers.provider.LATEST' imported but unused*

* 9b1a25093e4cb430a2df47d400041a83cc07966d

*CFME QE Bot*

cfme/tests/cloud_infra_common/test_cloud_init_provisioning.py
```diff
@@ -28,6 +30,17 @@
 ]
 
 
+def find_global_ipv6(vm):
+    """
+    Find global IPv6 on a VM if present and return it as a string. Otherwise, return None.
+    """
+    all_ips = vm.mgmt.all_ips
+    for ip in all_ips:
+        if ':' in ip and not ip.startswith('fe80'):
```
istein1: Is there a way to replace the hardcoded 'fe80' here with a definition?

cfme/tests/cloud_infra_common/test_cloud_init_provisioning.py
```diff
@@ -96,3 +109,81 @@ def test_provision_cloud_init(appliance, request, setup_provider, provider, prov
...
+        pytest.skip('No ci-image found in provider specification.')
+    note = ('Testing provisioning from image {} to vm {} on provider {}'.format(
+        image, vm_name, provider.key))
+    logger.info(note)
+
+    ci_payload = {
+        'root_password': 'mysecret',
+        'address_mode': 'Static',
+        'hostname': 'cimachine',
+        'ip_address': '10.1.1.10',
```
mshriver: These need to be non-routable or defined within and pulled from yamls.

Please see the other tests here for how payload is defined in the yamls and resolved for tests.

cfme/tests/cloud_infra_common/test_cloud_init_provisioning.py
```diff
@@ -96,3 +109,81 @@ def test_provision_cloud_init(appliance, request, setup_provider, provider, prov
...
+    request.addfinalizer(instance.cleanup_on_provider)
+    provision_request = provider.appliance.collections.requests.instantiate(vm_name,
+                                                                            partial_check=True)
+    provision_request.wait_for_request()
+
+    connect_ip = wait_for(find_global_ipv6, func_args=[instance], num_sec=600, delay=20).out
+    logger.info('Connect IP: {}'.format(connect_ip))
+
+    # Connect to the newly provisioned VM
+    ssh_client = paramiko.SSHClient()
```
mshriver: Please use `cfme.utils.ssh.SSHClient`.  It inherits Paramiko, and supports using the client in a context manager to close and control transport connection between exec calls.

Use that context manager below to wrap the use of the client to call `run_command`, and remove the call to close being registered with `request.addfinalizer`

jan-zmeskal: Hi @mshriver, this is something I've mentioned in the description of the PR - for some reason, I was not able to get `cfme.utils.ssh.SSHClient` to work with IPv6.

cfme/tests/cloud_infra_common/test_cloud_init_provisioning.py
jan-zmeskal: What about this?
ip_address: 169.254.0.1
subnet_mask: 29
gateway: 169.254.0.2
dns_servers: 169.254.0.3

This should not be routable as specified in this IANA doc: https://www.iana.org/assignments/iana-ipv4-special-registry/iana-ipv4-special-registry.xhtml

That request I understand. But why should we need to have this specified in yamls, given that this is just a dummny configuration? No matter in which environment you run it, it should work the same.

cfme/tests/cloud_infra_common/test_cloud_init_provisioning.py
mshriver: Gotcha, missed that in your description, my bad. Let's look into why its not working with IPv6, as it should just provide some convenience methods for Paramiko's library.

jan-zmeskal: @mshriver I think I know where the problem with wrapped paramiko might be. Working on a fix.

cfme/utils/net.py
```diff
@@ -67,15 +67,15 @@ def net_check(port, addr=None, force=False):
         # First try DNS resolution
         try:
             addr = socket.gethostbyname(addr)
-
-            # Then try to connect to the port
+        except:
+            _ports[addr][port] = False
+        finally:
+            # Try to connect to the port
```
jan-zmeskal: Hi @mshriver, using `cfme.utils.ssh` now works, but I had to change order of checks here. The reason is that even though `socket.gethostbyname` won't work with IPv6, `socket.create_connection` can still be used. I believe that this change should make `net_check` still work as originally intented. But I would be dishonest if I said that I truly understand everything that's going on in the fucntion.

cfme/utils/net.py
```diff
@@ -66,8 +66,9 @@ def net_check(port, addr=None, force=False):
     if port not in _ports[addr] or force:
         # First try DNS resolution
         try:
-            addr = socket.gethostbyname(addr)
-
+            addr_info = socket.getaddrinfo(addr, port)[0]
+            sockaddr = addr_info[4]
+            addr = sockaddr[0]
```
jan-zmeskal: This should work for `addr` being hostname, IPv4 and IPv6.
I followed advice by @rhrazdil.

cfme/tests/cloud_infra_common/test_cloud_init_provisioning.py
```diff
@@ -29,6 +30,17 @@
 ]
 
 
+def find_global_ipv6(vm):
+    """
+    Find global IPv6 on a VM if present and return it as a string. Otherwise, return None.
```
digitronik: Required:
This should be an `inline` comment.

cfme/tests/cloud_infra_common/test_cloud_init_provisioning.py
```diff
@@ -98,3 +110,81 @@ def test_provision_cloud_init(appliance, request, setup_provider, provider, prov
...
+
+    # Connect to the newly provisioned VM
+    with ssh.SSHClient(hostname=connect_ip,
+                       username='root',
+                       password=ci_payload['root_password']) as ssh_client:
+
+        # Check that correct hostname has been set
+        _, stdout, stderr = ssh_client.exec_command('hostname')
+        error, output = stderr.read(), stdout.read()
+        assert error == '', 'There was an error during getting hostname: {}'.format(error)
```
digitronik: Question:
Instate of using `exec_command` why are you not going with `run_command`?. It will help you to assert `success` and you will get direct `output`. Check small example..

```In [25]: x = ssh.SSHClient(hostname=appliance.hostname, username='root', password='smartvm')

In [26]: x.connect()

In [27]: x.connected
Out[27]: True

In [28]: host_name = x.run_command('hostname')

In [29]: host_name.success
Out[29]: True

In [30]: host_name.output
Out[30]: u'dhcp-8-196-130.cfme2.lab.eng.rdu2.redhat.com\n' ```

File: cfme/tests/cloud_infra_common/test_cloud_init_provisioning.py [hunk 10]
```diff
@@ -98,3 +110,81 @@ def test_provision_cloud_init(appliance, request, setup_provider, provider, prov
...
+        assert output.strip() == ci_payload['hostname']
+
+        # Obtain network configuration script for eth0 and store it in a list
+        _, stdout, stderr = ssh_client.exec_command('cat /etc/sysconfig/network-scripts/ifcfg-eth0')
```
digitronik: Required:
same as above...

cfme/tests/cloud_infra_common/test_cloud_init_provisioning.py
mshriver: @digitronik  why do you say so, not sure that I agree?

If anything I would request that the `Return` values are properly specified in the `Return:` header of the docblock, as opposed to including them in the first line of the summary.

cfme/tests/cloud_infra_common/test_cloud_init_provisioning.py
jan-zmeskal: @digitronik I tried it after @rhrazdil suggested it. Problem with that is that if this function returns `None` if case of failure, `wait_for` considers it finished and does not perform another round.

cfme/tests/cloud_infra_common/test_cloud_init_provisioning.py
```diff
@@ -98,3 +110,81 @@ def test_provision_cloud_init(appliance, request, setup_provider, provider, prov
...
+    Tests that options specified in VM provisioning dialog in UI are properly passed as a cloud-init
+    payload to the newly provisioned VM.
+
+    Metadata:
+        test_flag: cloud_init, provision
+    """
+    image = provisioning.get('ci-image', None)
+    if not image:
+        pytest.skip('No ci-image found in provider specification.')
+    note = ('Testing provisioning from image {} to vm {} on provider {}'.format(
```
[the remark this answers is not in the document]
jan-zmeskal: I think this is still quite easy to parse but yeah, I can do it with keys.

cfme/tests/cloud_infra_common/test_cloud_init_provisioning.py
```diff
@@ -98,3 +110,81 @@ def test_provision_cloud_init(appliance, request, setup_provider, provider, prov
...
+    }
+
+    inst_args = {
+        'request': {'notes': note},
+        'customize': {'customize_type': 'Specification'},
+        'template_name': image
+    }
+
+    inst_args['customize'].update(ci_payload)
+    logger.info('Instance args: {}'.format(inst_args))
```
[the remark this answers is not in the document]
jan-zmeskal: To be  consistent with `test_provision_cloud_init`: https://github.com/ManageIQ/integration_tests/blob/ad4fe37ef9874098a51568524e77157d635b87bf/cfme/tests/cloud_infra_common/test_cloud_init_provisioning.py#L83

cfme/tests/cloud_infra_common/test_cloud_init_provisioning.py
jan-zmeskal: Nice hint, thank you!

cfme/tests/cloud_infra_common/test_cloud_init_provisioning.py
digitronik: Ahhh I miss return in doc string. Just checked format. I agreed with your statement Return block needs.

cfme/tests/cloud_infra_common/test_cloud_init_provisioning.py
```diff
@@ -29,6 +30,21 @@
 ]
 
 
+def find_global_ipv6(vm):
+    """
+    Find global IPv6 on a VM if present.
+
+    Args:
+        vm: InfraVm object
```
digitronik: Required:
single blank line after one block...

PR #7716, opened 2018-08-28
[1LP][RFR] fixed AttributeError in test_paginator.py tests

lina-is-here:
{{ pytest: -v cfme/tests/configure/test_paginator.py }}

Fixed AttributeError but test doesn't work for 5.10 due to BZ. So also added blocker for test_paginator

### DIFF
--- a/cfme/tests/configure/test_paginator.py
+++ b/cfme/tests/configure/test_paginator.py
@@ -89,8 +89,8 @@ def schedule(appliance):
     schedule.delete()
 
 
-@pytest.mark.meta(blockers=[
-                  BZ(1623091, forced_streams=['5.10'])])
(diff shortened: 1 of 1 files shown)


cfme/tests/configure/test_paginator.py
```diff
@@ -88,6 +89,8 @@ def schedule(appliance):
     schedule.delete()
 
 
+@pytest.mark.meta(blockers=[
```
izapolsk: Q: as far as I understand this BZ blocks only one or two test cases not whole test, right ?
Please update statement if so.

lina-nikiforova: Thanks for the review, you're right, BZ affects not the whole test. Only those test cases for VMDB Summary view.
Updated the blocker statement.

PR #7729, opened 2018-08-29
Fix list provider vms

mshriver:
Jenkins already running from this branch.


mshriver:
Since this has been tested in CI and only impacts utility script, I'm going to merge after travis passes.

PR #7733, opened 2018-08-30
[1LP][RFR] New Set of v2v migration tests

kedark3:
Purpose or Intent
=================

__Adding tests__ for feature v2v  : 

- Migration using SSH Transformation method
- Migration preserving Retirement, Power State, Tags and ownership
- Migration cancellation Test
- Migration Scheduling Test
- Migrations with multiple conversion-hosts 
**UPDATE:** Following two tests moved out and will be added in different PR.
- Migration cancellation Test
- Migration Scheduling Test

{{pytest: cfme/tests/v2v/test_v2v_migrations.py --use-provider rhv42 --use-provider vsphere6-nested --provider-limit 2 -vvvv --long-running }}


dajoRH:
*CFME QE Bot*

jawatts:
@kedark3 Can you check out the PRT failures for this PR

cfme/tests/v2v/test_v2v_migrations.py
```diff
@@ -441,3 +443,391 @@ def test_conversion_host_tags(appliance, v2v_providers):
...
+        infrastructure_mapping_collection.delete(mapping)
+
+    migration_plan_collection = appliance.collections.v2v_plans
+
+    migration_plan = migration_plan_collection.create(
+        name="plan_{}".format(fauxfactory.gen_alphanumeric()), description="desc_{}"
+        .format(fauxfactory.gen_alphanumeric()), infra_map=mapping.name,
+        vm_list=form_data_multiple_vm_obj_single_datastore[1], start_migration=True)
+    # as migration is started, try to track progress using migration plan request details page
+    view = appliance.browser.create_view(navigator.get_class(migration_plan_collection, 'All').VIEW)
```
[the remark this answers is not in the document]
kedark3: @sshveta wouldn't it be better to keep using current method, in case we change the class name for actual view, the navigation destination will know which class to point to, removing need to update this test. I think we should keep it as is.

kedark3:
Hey Y'all, PRT seems to have worked well. Many cases passed, some skipped maybe due to provider issues(mainly providers cannot create VMs on timely manner so it times out) and failures are due to Migration of VM took longer than an hour, current timeout is set to an hour and depending on provider load most migrations should finish within that timeframe.

kedark3:
Something is wrong with infra mapping wizard, moving back to WIPTEST

cfme/fixtures/v2v.py
```diff
@@ -44,23 +50,29 @@ def extract_tag(tag):
 
     if len(tags_set) < 2 or not tags_set.issubset(valid_tags):
         host_obj.remove_tags(tags=tags)
-        return True
-    return False
+        return False
```
mshriver: Its not clear why this boolean is switched, because its not actually clear what this return is supposed to represent :)

Add a bit more to the docblock description as to what this boolean is supposed to represent, so that the intent is clear to the caller.

cfme/fixtures/v2v.py
```diff
@@ -44,23 +50,29 @@ def extract_tag(tag):
...
 
 
 @pytest.fixture(scope='function')
-def conversion_tags(appliance, host_creds):
+def conversion_tags(request, appliance, host_creds):
     """Assigning tags to conversation host"""
     tag1 = appliance.collections.categories.instantiate(
         display_name='V2V - Transformation Host *').collections.tags.instantiate(
         display_name='t')
+    if hasattr(request, 'param') and request.param == 'SSH':
```
mshriver: Looks like `transformation_method` could just equal the param here? Reasonable to force callers to parametrize on valid transform methods?

cfme/fixtures/v2v.py
```diff
@@ -44,23 +50,29 @@ def extract_tag(tag):
...
+    else:
+        transformation_method = 'VDDK'
     tag2 = appliance.collections.categories.instantiate(
         display_name='V2V - Transformation Method').collections.tags.instantiate(
-        display_name='VDDK')
-    if _tag_cleanup(host_creds, tag1, tag2):
-        host_creds.add_tags(tags=(tag1, tag2))
+        display_name=transformation_method)
+    for host in host_creds:
+        if not _tag_cleanup(host, tag1, tag2):
```
mshriver: This is a bit ambiguous to me, as the intent of _tag_cleanup's boolean return isn't clear. Please add some comments here for the intent - you're trying to add tags when the call to cleanup does not return true - which means cleanup did ... what?

cfme/tests/v2v/test_v2v_migrations.py
```diff
@@ -133,24 +99,12 @@ def _cleanup():
...
-          request_details_list.is_in_progress(vm))
-        return not(request_details_list.is_in_progress(vm) and (clock_reading1 < clock_reading2))
-
-    wait_for(func=_get_plan_status, message="migration plan is in progress, be patient please",
+    vms = request_details_list.read()
+
+    wait_for(func=view.get_plan_status, message="migration plan is in progress, be patient please",
      delay=5, num_sec=3600)
-    assert (request_details_list.is_successful(vm) and not request_details_list.is_errored(vm))
+    assert (request_details_list.is_successful(vms[0]) and
```
mshriver: I think there's a bit of an assumption here that could cause some exceptions - and really its a hidden test assertion.

Add a check that there is a vm in `vms` list, before indexing it, when it may be empty.

cfme/tests/v2v/test_v2v_migrations.py
```diff
@@ -441,3 +299,283 @@ def test_conversion_host_tags(appliance, v2v_providers):
...
+    view.migration_plans_not_started_list.schedule_migration(migration_plan.name, after_mins=1)
+    view.migr_dropdown.item_select('In Progress Plans')
+    wait_for(func=view.progress_card.is_plan_started, func_args=[migration_plan.name],
+        message="migration plan is starting, be patient please", delay=5, num_sec=150,
+        handle_exception=True)
+
+    # wait until plan is in progress
+    wait_for(func=view.get_plan_status, func_args=[migration_plan.name],
+        message="migration plan is in progress, be patient please",
+        delay=5, num_sec=3600)
```
mshriver: heck of a wait time, let's cut these down to 1800 please

cfme/tests/v2v/test_v2v_migrations.py
```diff
@@ -441,3 +299,283 @@ def test_conversion_host_tags(appliance, v2v_providers):
...
+        migration_plan_in_progress_tracker = []
+        for vm in vms:
+            clock_reading1 = request_details_list.get_clock(vm)
+            time.sleep(1)  # wait 1 sec to see if clock is ticking
+            logger.info("For vm %s, current message is %s", vm,
+                request_details_list.get_message_text(vm))
+            current_progress_text = request_details_list.get_progress_description(vm)
+            current_data_transferred, total_data_to_transfer = map(float, re.findall("\d+\.\d+",
+                current_progress_text))
+            if current_data_transferred > (total_data_to_transfer / 2):
```
mshriver: Can we just wait until its a 3rd done? or a 5th? 10th? Thoughts on what's reasonable here, I'm thinking 10% might really be sufficient for a process cancel test, at least starting. If its an area we want to automate for more coverage, we can start to push the limits of canceling immediately, and at 99%, etc.

cfme/tests/v2v/test_v2v_migrations.py
```diff
@@ -441,3 +299,283 @@ def test_conversion_host_tags(appliance, v2v_providers):
...
+    view.migr_dropdown.item_select("Completed Plans")
+    view.wait_displayed()
+    logger.info("For plan %s, migration status after completion: %s, total time elapsed: %s",
+        migration_plan.name, view.migration_plans_completed_list.get_vm_count_in_plan(
+            migration_plan.name), view.migration_plans_completed_list.get_clock(
+            migration_plan.name))
+    assert view.migration_plans_completed_list.is_plan_succeeded(migration_plan.name)
+
+
+@pytest.mark.parametrize('form_data_vm_obj_single_datastore', [['nfs', 'nfs', rhel7_minimal]],
```
jawatts: Question: why is ```nfs`` here twice?

cfme/tests/v2v/test_v2v_migrations.py
```diff
@@ -441,3 +299,283 @@ def test_conversion_host_tags(appliance, v2v_providers):
...
+    indirect=True)
+@pytest.mark.parametrize('power_state', ['RUNNING', 'STOPPED'])
+def test_single_vm_migration_power_state_tags_retirement(request, appliance, v2v_providers,
+                                    host_creds, conversion_tags,
+                                    form_data_vm_obj_single_datastore,
+                                    power_state):
+    # Test VM migration power state and tags are preserved
+    vm = form_data_vm_obj_single_datastore[1][0]
```
jawatts: As @mshriver stated above, some clarity on the indexing here and below would be helpful.

cfme/tests/v2v/test_v2v_migrations.py
```diff
@@ -441,3 +299,283 @@ def test_conversion_host_tags(appliance, v2v_providers):
...
+    migrated_vm = rhv_prov.mgmt.get_vm((form_data_vm_obj_single_datastore[1][0]).name)
+    assert power_state in migrated_vm.state
+    # check tags
+    vm_obj = appliance.collections.infra_vms.instantiate(migrated_vm.name, rhv_prov)
+    owner_tag = None
+    for t in vm_obj.get_tags():
+        if tag.display_name in t.display_name:
+            owner_tag = t
+    assert owner_tag is not None and tag.display_name in owner_tag.display_name
+    assert 'Never' not in vm_obj.retirement_date
```
jawatts: Maybe a comment here on what ```Never``` means? Is it if a provider hasn't been migrated it would show Never?

cfme/tests/v2v/test_v2v_migrations.py
```diff
@@ -441,3 +299,283 @@ def test_conversion_host_tags(appliance, v2v_providers):
...
+
+
+@pytest.mark.parametrize('host_creds, form_data_multiple_vm_obj_single_datastore', [['multi-host',
+    ['nfs', 'nfs', [rhel7_minimal, ubuntu16_template, rhel69_template, win7_template]]]],
```
jawatts: same question as above, two nfs?

cfme/v2v/migrations.py
```diff
@@ -495,6 +532,33 @@ def sort_by(self, option):
             raise ItemNotFound("Migration plan is in Not Started State,"
                 " hence sort_by dropdown not visible")
 
+    def get_plan_status(self, vms_count=5):
+        """Reuturn True or False, depending on migration plan status.
```
mshriver: typo on 'Reuturn', and please format this docblock more - first line should describe the action/use for the caller.

These comments should be in `Returns:` header

mshriver: I think returning `True` on both any plan running, and no plans running, is an ambiguous implementation for the caller.

Why not return `None` when there are no plans in progress?

I think the function may be a bit mis-named, as its not actually returning a status, which I would expect from `get_plan_status`. Does True mean 'in progress'? finished? finished with success?

Consider either actually returning a status/`None` or renaming to `plan_in_progress` or something to that effect.

jawatts: ^ agree

widgetastic_manageiq/__init__.py
```diff
@@ -4931,6 +4996,15 @@ def is_errored(self, vm_name):
             # This means element not present, so no error
             return False
 
+    def is_cancelled(self, vm_name):
+        """See if cross icon indicating failure present."""
```
jawatts: Following @mshriver comment above, lets clean up these doc blocks

cfme/fixtures/v2v.py
kedark3: I did not want to force all to send this parameter as by default VDDK is easier method(in terms of env setup) to set up and all tests can use it as default.

cfme/tests/v2v/test_v2v_migrations.py
```diff
@@ -441,3 +299,283 @@ def test_conversion_host_tags(appliance, v2v_providers):
...
+
+
+@pytest.mark.parametrize('conversion_tags, form_data_vm_obj_single_datastore', [['SSH',
+    ['nfs', 'nfs', rhel7_minimal]]], indirect=True)
+def test_single_vm_migration_with_ssh(request, appliance, v2v_providers, host_creds,
+                                    conversion_tags,
+                                    form_data_vm_obj_single_datastore):
+    infrastructure_mapping_collection = appliance.collections.v2v_mappings
+
+    mapping = infrastructure_mapping_collection.create(form_data_vm_obj_single_datastore[0])
```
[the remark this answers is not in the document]
kedark3: form_data is [0] and vm_obj[1] in all the fixtures that have similar names. Would you like me to comment it in each test?

cfme/tests/v2v/test_v2v_migrations.py
kedark3: I will reduce time out for all those tests that are migrating small VM(only 1 VM). For some tests, there are more than 1 VMs being migrated and those are big in size. 30 mins may fall short.

cfme/tests/v2v/test_v2v_migrations.py
kedark3: This is something I can improve on in a different PR. Definitely I'd like to try cancellation at different phases. But initially just checking functionality itself is good enough. I'll add a `#TODO` ? Will submit a PR in early next week .

cfme/tests/v2v/test_v2v_migrations.py
kedark3: source and target both needs to be `nfs` storage.So, I have combinations like:
`nfs` to `nfs`
`nfs` to `iscsi` and so on.

cfme/tests/v2v/test_v2v_migrations.py
kedark3: I'll add a comment on this. `Never` means retirement date not set. If `Never` is not there, that means retirement is set, which is what I want to assert.

cfme/tests/v2v/test_v2v_migrations.py
kedark3: Same as above.

cfme/fixtures/v2v.py
```diff
@@ -18,21 +18,29 @@ def v2v_providers(request, second_provider, provider):
...
-    yield host
-    host.remove_credentials_rest()
+    hosts = provider.hosts.all()
+    try:
+        hosts = hosts if getattr(request, 'param', '') == 'multi-host' else hosts[0:1]
+        for host in hosts:
+            host_data, = [data for data in provider.data['hosts'] if data['name'] == host.name]
+            host.update_credentials_rest(credentials=host_data['credentials'])
+    except Exception:
+        # if above throws ValueError or TypeError or other exception, just skip the test
```
mshriver: :+1: I think its fine to diaper here.

kedark3:
For Run 26692, 
1) test_single_vm_migration_power_state_tags_retirement failed as the feature itself is buggy, and yet to be fixed in ManageIQ, this accounts for 2 failures.
2) test_migration_different_os failed 1 time as it could not migrate one of the 2 VMs, and it could be issue with provider or VM itself. Test did work as expected.
3) Remaining 2 failures are due to Alert notification being present on screen. This is something I need to review, for Schedule_migration and Cancel_migration tests. 

I will remove those two tests from this PR and fix it and put up a new PR together for those 2 tests. what say @mshriver ?

kedark3:
@mshriver thanks for merge :)

PR #7735, opened 2018-08-31
[WIP] Refactor physical server stat calls

matejart:
With this patch we suggest a better way of calling `server_stats` and `server_inventory` in the integration tests.


dajoRH:
*CFME QE Bot*

PR #7739, opened 2018-08-31
[1LP][RFR] Fixed vm provisioning error in quota test

ganeshhubale:
{{ pytest: cfme/tests/cloud/test_quota_tagging.py::test_quota_tagging_cloud_via_lifecycle --use-provider complete }}

### DIFF
--- a/cfme/tests/cloud/test_quota_tagging.py
+++ b/cfme/tests/cloud/test_quota_tagging.py
@@ -149,7 +149,7 @@ def test_quota_tagging_cloud_via_lifecycle(request, appliance, provider, prov_da
     recursive_update(prov_data, {
         'request': {'email': 'test_{}@example.com'.format(fauxfactory.gen_alphanumeric())}})
     prov_data.update({'template_name': template_name})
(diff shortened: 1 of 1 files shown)


cfme/common/vm.py
```diff
@@ -531,6 +531,7 @@ def create(self, vm_name, provider, form_values=None, cancel=False, check_existi
                              handle_exception=True, num_sec=600)
                 elif override:
                     logger.info('Overriding exception to check failure condition.')
+                    return vm
```
digitronik: Question: For which provider you are getting `Exception` error? I am suspect about returning `vm` object in `override`.

ganeshhubale: I am getting Exception error for **openstack** provider.

digitronik: Required: 
Then no need to return `vm` object `override` will handle an exception. As per your test no need of `vm object`. debug your failures.

ganeshhubale:
@digitronik 
PR is updated with required changes please take a look at it.
Thank you.

PR #7740, opened 2018-08-31
[1LP][RFR] Adding some fixes for UI problems in 5.8

jan-zmeskal:
There are some minor differences between messages in UI in 5.8. This should take care of that. Also, at some point in SSUI, there is a notification that obscures upper right menu. I added **super-lazy workaround** using `time.sleep(15)`. I know it is ugly, but hey, it's almost deprecated branch and there's IMO no value in fixing it properly.

Jenkins verification: https://rhv-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/cfme-5.8-rhv-4.2-integration-flow-dev/95/

### DIFF
--- a/cfme/services/myservice/ssui.py
+++ b/cfme/services/myservice/ssui.py
@@ -165,7 +165,7 @@ def set_ownership(self, owner, group):
                'select_group': group})
     view.save_button.click()
     view = self.create_view(DetailsMyServiceView)
-    assert view.is_displayed
+    view.wait_displayed('15s')  # WA for StaleElementException in 5.8 testing
     # TODO - remove sleep when BZ 1518954 is fixed
     time.sleep(10)
     if self.appliance.version >= "5.8":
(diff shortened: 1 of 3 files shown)


cfme/services/myservice/ssui.py
```diff
@@ -165,6 +165,7 @@ def set_ownership(self, owner, group):
                'select_group': group})
     view.save_button.click()
     view = self.create_view(DetailsMyServiceView)
+    time.sleep(15)  # WA for StaleElementException in 5.8 testing
```
digitronik: Required:
@jan-zmeskal: I think you can add some wait to display this view like `view.wait_displayed("15s")`. I know first, you have to check `is_displayed` method for this view.

cfme/services/myservice/ssui.py
```diff
@@ -184,6 +185,7 @@ def edit_tags(self, tag, value):
     view.add_tag.click()
     view.save.click()
     view = self.create_view(DetailsMyServiceView)
+    time.sleep(15)
```
digitronik: similar wait for a view display.

cfme/services/service_catalogs/ui.py
```diff
@@ -88,7 +88,7 @@ def order(self):
     if self.ansible_dialog_values:
         view.fill(self.ansible_dialog_values)
     msg = "Order Request was Submitted"
-    msg_type = "info"
+    msg_type = "success"
```
digitronik: :+1:

PR #7741, opened 2018-08-31
[1LP][RFR] Fixing getting VM state for 5.8

jan-zmeskal:
The name of file that is used to get VM state is different in 5.8. I am not using `VersionPicker` because this fix is specific for 5.8-breakpoint-patched branch.

Jenkins verification: https://rhv-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/cfme-5.8-rhv-4.2-integration-flow-dev/92/

### DIFF
--- a/requirements/frozen.py2.txt
+++ b/requirements/frozen.py2.txt
@@ -247,7 +247,7 @@ python-string-utils==0.6.0
 python-swiftclient==3.5.0
 pytz==2018.4
 pyvcloud==19.1.2
-pyvmomi==6.7.0
+pyvmomi==6.7.0.2018.9
 pywinrm==0.3.0
 PyYAML==3.12
 pyzmq==17.0.0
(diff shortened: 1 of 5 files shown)


requirements/frozen.py2.txt
```diff
@@ -308,7 +308,7 @@ Werkzeug==0.14.1
 widgetastic.core==0.21.6
 widgetastic.patternfly==0.0.37
 widgetsnbextension==3.2.1
-wrapanapi==3.0.23
+wrapanapi==3.0.24
```
mshriver: Please make this update across all the frozen files, or remove.

jan-zmeskal: Requirements updated by running `python freeze_all.py --upgrade-only wrapanapi`.

mshriver: (discussed offline a bit) The freeze_all brought in new requirements from local env that we don't want.

they're now limited to what packages we want - but we need these updates in both py2 and py3 files.

PR #7751, opened 2018-09-03
[1LP][RFR][NOTEST] fix for deployment role entity

digitronik:
Purpose or Intent
=================
- No need of quad icon normal `BaseEntitiesView` will work on All page views.
- It  will fix all navigation failures facing by @dkholodo 
- Not adding the test as we don't have `under cloud` provider.
- Tested it locally. working awesome. thanks to @dkholodo for an appliance. 

Local  Results:
http://pastebin.test.redhat.com/640753

### DIFF
--- a/cfme/infrastructure/deployment_roles.py
+++ b/cfme/infrastructure/deployment_roles.py
@@ -4,19 +4,23 @@ from functools import partial
 import attr
 from navmazing import NavigateToAttribute, NavigateToSibling
 from widgetastic.exceptions import NoSuchElementException
-from widgetastic.widget import ParametrizedView, View, Text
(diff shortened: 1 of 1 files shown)


cfme/infrastructure/deployment_roles.py
```diff
@@ -1,22 +1,26 @@
 """ A model of an Infrastructure Deployment roles in CFME"""
+import attr
 from functools import partial
 
-import attr
```
quarckster: Actually it's a correct place for this import.

dkholodo:
@digitronik Thanks for this PR. Now it's working.

PR #7752, opened 2018-09-04
[1LP][RFR] Fix test_datastores_summary test on 5.8

valaparthvi:
The test case fails when the number of VMs is greater than 999. Take for example the number of VMs is 1248. The data received by reading rows is `1,248`. Converting `1,248` to integer fails, since `int()` fails to convert a string with ',' in between. This change removes the ',', which allows successful string conversion.

{{ pytest: cfme/tests/intelligence/reports/test_canned_corresponds.py -k test_datastores_summary }}

### DIFF
--- a/cfme/tests/intelligence/reports/test_canned_corresponds.py
+++ b/cfme/tests/intelligence/reports/test_canned_corresponds.py
@@ -222,7 +222,7 @@ def test_datastores_summary(soft_assert, appliance, request):
         row['Free Space'] = extract_num(row['Free Space'])
         row['Total Space'] = extract_num(row['Total Space'])
         row['Number of Hosts'] = int(row['Number of Hosts'])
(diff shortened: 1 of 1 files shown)


valaparthvi:
@digitronik  I checked on 5.9 and again on 5.8, but for some reason, I didn't get any data where the number of VMs was greater than 1000, but we can keep this fix if we ever encounter such a thing again.

izapolsk:
and one more thing, please add : after pytest. dockerbot goes crazy when it gets malformed prt command strings

valaparthvi:
@izapolsk Got it! I saw some error related to dockerbot on hubber, didn't realize it had something to do with the malformed command string, and since this change was specifically for 5.8, I didn't pay much attention to the error. But, I'll keep your comments in mind next time, thanks :+1:

PR #7756, opened 2018-09-04
[1LP][RFR] Change sprout client default auth behavior

mshriver:
Remove default credentials lookup for sprout client, using only env vars or a passed creds key

This prevents groups with a credentials file that contains a 'sprout' key from defaulting to that user for anyone using the credentials file.
This was causing regular users to make sprout requests as a CI user instead of themselves.

Testing:

All changes tested locally, CI configuration changes ready, for passing `--sprout-user-key` to `miq-runtest`

### DIFF
--- a/cfme/fixtures/cli.py
+++ b/cfme/fixtures/cli.py
@@ -6,9 +6,8 @@ from six import iteritems
 
 import cfme.utils.auth as authutil
 from cfme.test_framework.sprout.client import SproutClient
-from cfme.utils.conf import cfme_data, credentials, auth_data
+from cfme.utils.conf import credentials, auth_data
 from cfme.utils.log import logger
-from cfme.utils.version import get_stream
 from cfme.utils.wait import wait_for
 
 TimedCommand = namedtuple('TimedCommand', ['command', 'timeout'])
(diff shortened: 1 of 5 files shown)


cfme/fixtures/cli.py
```diff
@@ -18,10 +18,10 @@
 
 
 @contextmanager
-def fqdn_appliance(appliance, preconfigured, count):
+def fqdn_appliance(appliance, preconfigured, count, config):
     version = appliance.version.vstring
     stream = get_stream(appliance.version)
-    sprout_client = SproutClient.from_config()
+    sprout_client = SproutClient.from_config(sprout_user_key=config.option.sprout_user_key or None)
```
mshriver: This part is untested.

cfme/fixtures/cli.py
```diff
@@ -18,25 +17,24 @@
...
+def fqdn_appliance(appliance, preconfigured, count, config):
+    sprout_client = SproutClient.from_config(sprout_user_key=config.option.sprout_user_key or None)
     apps, request_id = sprout_client.provision_appliances(
-        provider_type='rhevm', count=count, version=version, preconfigured=preconfigured,
-        stream=stream)
+        provider_type='rhevm',
+        count=count,
+        version=appliance.version.vstring,
+        preconfigured=preconfigured,
+        stream=appliance.version.stream(),
```
jawatts: Super small, but you don't need that comma here

cfme/test_framework/sprout/plugin.py
```diff
@@ -74,8 +80,12 @@ def mangle_in_sprout_appliances(config):
     """
     provision_request = SproutProvisioningRequest.from_config(config)
 
-    mgr = config._sprout_mgr = SproutManager()
-    requested_appliances = mgr.request_appliances(provision_request)
+    mgr = config._sprout_mgr = SproutManager(config.option.sprout_user_key)
+    try:
+        requested_appliances = mgr.request_appliances(provision_request)
+    except AuthException:
+        log.exception('Sprout client not authenticated, please provide env vars or sprout_user_key')
```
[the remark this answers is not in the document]
mshriver: No, wanted to actually have the exception logged along with the plain-text description. Why do you suggest using `error()` here?

cfme/fixtures/cli.py
mshriver: true, not needed, but when we enable `Black` it'll add it in these situations anyway:

`Black will add trailing commas to expressions that are split by comma where each element is on its own line. This includes function signatures.`

cfme/test_framework/sprout/plugin.py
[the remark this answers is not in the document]
izapolsk: I can't recall such log level but maybe I'm wrong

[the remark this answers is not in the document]
izapolsk: I've just checked. nevermind. my bad

cfme/fixtures/cli.py
RonnyPfannschmidt: its absolutely required for better and saner diffs on later changes

PR #7758, opened 2018-09-05
[1LP][RFR] fix for quadicon

lina-nikiforova:
{{ pytest: -v cfme/tests/cloud_infra_common/ --use-provider complete }}

This commit undo some changes in #7502 as 'quadicon' property is accessible again.
The removed lines didn't work if page is opened in List view and it is opened in this view by default for Images, so the test test_tag_cloud_objects was broken.


mshriver:
PRT is showing some failures, but none related to quadicon. Logs show successful access of quadicon properties/states.

PR #7761, opened 2018-09-05
[1LP][WIPTEST] New upgrade tests added and old tests moved from test_db_migrate

lcouzens:
Added new upgrade tests for multiple scenarios and moved old upgrade tests from test_db_migrate to test_appliance_upgrade.
{{pytest: -v -k "test_upgrade"}}
REQUIRES PRS: 7760 and 7759

### DIFF
--- a/cfme/fixtures/cli.py
+++ b/cfme/fixtures/cli.py
@@ -251,7 +251,6 @@ def ha_appliances_with_providers(ha_multiple_preupdate_appliances, app_creds):
     command_set = ("ap", "", "5", "2", app0_ip, "", pwd, "", "2", "0", "y", app0_ip, "", "", "",
                    TimedCommand(pwd, 360), "")
     apps2.appliance_console.run_commands(command_set)
-    apps2.wait_for_evm_service()
     apps2.wait_for_web_ui()
     # Configure primary replication node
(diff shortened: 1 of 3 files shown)


dajoRH:
*CFME QE Bot*

lcouzens:
5.10 'Failures' are actually passes due to latest build just changing during the test run, test run is 5.10.0.21 but tests upgrade to 5.10.0.22, Remaining issues are sprout and network related issues.

mshriver:
I'd like to figure out why this isn't getting single sprout appliances, are we missing templates? very low pass rates in PRT

lcouzens:
I have looked at this before but can't figure out what the problem is with sprout. There are ssh errors which I would guess are either ipv6 issues popping up again or just network drops/hangs. As for sprout not provisioning it makes no sense, the templates we require are there and availability should be fine, the only thing I am doing different is requesting from provider_type rhevm, But all the provisions from these test are handle by the same context manage [0] so it makes no sense that some provision and other don't. 
[0] https://github.com/ManageIQ/integration_tests/blob/4eb5ae05e525f8c226ead452787264fe0afc2ff1/cfme/fixtures/cli.py#L148

PR #7774, opened 2018-09-06
[1LP][RFR]Catalog name appended by My_Company in 5.10

sshveta:
{{pytest: cfme/tests/services/test_myservice.py -v --use-provider vsphere55}}

In 5.10 catalog name is displayed with My Company/catalog_name.
It affects some tests but tests cannot be run since 5.10 tests are failing because of issue 
https://bugzilla.redhat.com/show_bug.cgi?id=1626232.

Blocking tests with this BZ in another upcoming PR

### DIFF
--- a/cfme/services/catalogs/catalog_items/__init__.py
+++ b/cfme/services/catalogs/catalog_items/__init__.py
@@ -14,6 +14,7 @@ from cfme.services.catalogs import ServicesCatalogView
 from cfme.utils.appliance.implementations.ui import navigator, CFMENavigateStep, navigate_to
 from cfme.utils.pretty import Pretty
 from cfme.utils.update import Updateable
+from cfme.utils.version import LOWEST, VersionPicker
 from cfme.utils.wait import wait_for
 
 
@@ -299,6 +300,15 @@ class BaseCatalogItem(BaseEntity, Updateable, Pretty, Taggable):
         except CandidateNotFound:
             return False
 
+    @property
+    def catalog_name(self):
+        # In 5.10 catalog name is appended with 'My Company'
+        cat_name = VersionPicker({
+            LOWEST: getattr(self.catalog, 'name', None),
+            '5.10': 'My Company/{}'.format(getattr(self.catalog, 'name', None))
+        }).pick(self.appliance.version)
+        return cat_name
+
 
 @attr.s
 class CloudInfraCatalogItem(BaseCatalogItem):
@@ -320,7 +330,7 @@ class CloudInfraCatalogItem(BaseCatalogItem):
                 'name': self.name,
                 'description': self.description,
                 'display': self.display_in,
-                'select_catalog': getattr(self.catalog, 'name', None),
+                'select_catalog': self.catalog_name,
                 'select_dialog': self.dialog
             },
             'request_info': {'provisioning': self.prov_data}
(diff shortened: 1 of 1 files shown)


cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -315,12 +315,15 @@ class CloudInfraCatalogItem(BaseCatalogItem):
 
     @property
     def fill_dict(self):
+        # In 5.10 catalog name is appended with 'My Company'
+        catalog_name = getattr(self.catalog, 'name', None)
+        cat_name = catalog_name if self.appliance.version < '5.9' else 'My Company/' + catalog_name
```
Please go with `string` formatting style like
`"My Company/{}".format(catalog_name)`

As per inline comment form, 5.10 its comes with `My Company` right? so condition should `<5.10` what you think?

cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -341,11 +344,14 @@ class NonCloudInfraCatalogItem(BaseCatalogItem):
 
     @cached_property
     def _fill_dict(self):
+        # In 5.10 catalog name is appended with 'My Company'
+        catalog_name = getattr(self.catalog, 'name', None)
+        cat_name = catalog_name if self.appliance.version < '5.9' else 'My Company/' + catalog_name
```
same as above

cfme/services/catalogs/catalog_items/__init__.py
sshveta: yup that was a typo .

dajoRH:
Lint report for commit d85d85377a2d8716699a9f3f1235c89ab88bbd53:

`cfme/services/catalogs/catalog_items/__init__.py`:
- :bangbang: Line 320:101: **E501** *line too long (108 > 100 characters)*
- :bangbang: Line 349:101: **E501** *line too long (108 > 100 characters)*

**No commit flaws detected**.

*CFME QE Bot*

digitronik:
updated `--use-provider vsphere55`

digitronik:
Talk with @sshveta test are passing locally;

cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -315,12 +315,16 @@ class CloudInfraCatalogItem(BaseCatalogItem):
 
     @property
     def fill_dict(self):
+        # In 5.10 catalog name is appended with 'My Company'
```
mshriver: I think this would be more readable as a `VersionPicker`, and a bit more versatile as its a `fill_dict` property, that would be used to fill a view, where the version pick will be automatically resolved.

Thoughts?

```
from cfme.utils.version import VersionPicker, LOWEST

cat_name = VersionPicker(
    LOWEST: getattr(self.catalog, 'name', None),
    '5.10': 'My Company/{}'.format(getattr(self.catalog, 'name', None))
)
```

mshriver: This looks like the same implementations and dict returns in both classes, and they both inherit `BaseCatalogItem`. Please double check me on this, and move this method to `BaseCatalogItem` if that's the case.

digitronik: @mshriver I dont think we have to go with `VersionPick`. specifally where we are providing data to view. Yes, `VersionPick` is  best option in `Views` but other than `View` no need to import `VersionPicker` as its dynamic data generation. As per my understanding for `locators` assignment except `ParametrizeView` we con't deciside vesions so `VersionPick` is right but other palces where we can use `appliance.version`.

PR #7778, opened 2018-09-07
[WIPTEST] restoring parsing quad removed by PR #7758

izapolsk:
quadicon was removed replaced wit quad in 5.10. So, this code should definitely be restored.

{{ pytest: -v cfme/tests/cloud_infra_common/ --use-provider complete }}


mshriver:
Prime example of why we need 510z testing working in PRT :)

PR #7789, opened 2018-09-10
[RFR] Fix fixtures for sprout appliances, pass pytestconfig

mshriver:
I broke this in #7756 with only passing pytestconfig in the first fixture using temp_appliance method.


mshriver:
@jaryn thanks for pointing this out!

PR #7793, opened 2018-09-11
[1LP][RFR] Testing cloud quota for 'user' and 'group' by provisioning instance via lifecycle

ganeshhubale:
- Testing cloud quota of user and group with openstack provider
- Provisioning instance with more than assigned quota

{{ pytest: cfme/tests/cloud/test_quota_tagging.py::test_cloud_quota_by_lifecycle --use-provider complete }}


ganeshhubale:
Thanks @sshveta @mshriver

PR #7795, opened 2018-09-11
[RFR] Fix getting sprout auth.

jaryn:
If bool(user_key) resolved to false, None was used as user/password. The
environment variable was never effective.


jkrocil:
@mshriver This was blocking env vars from being used when `sprout_user_key` wasnt specified.

jkrocil:
Tested locally; both env and yamls. Merging.

PR #7797, opened 2018-09-11
[1LP][RFR] Allow plugins to return value

sasoc:
Artifactor gets triggered by `fire_art_hook` to handle events after they happen, Artifactor then tells plugin that certain event has occurred and the plugin then reacts accordingly, while `fire_art_hook` does not expect any response. So in other words we tell plugin to do something and then we expect nothing to be returned. But in some cases plugin does return some value, which we expect immediately. With this commit `fire_art_hook` and `fire_art_test_hook` now return value that they get from plugins.

For example we can look at Merkyl plugin that checks log content, which was logged after test started. Merkyl has `merkyl_inspector` fixture that has `get_log` method. This method uses `fire_art_test_hook` to tell plugin to get log content, plugin does return that content but then it gets lost in `fire_art_hook`.

### DIFF
--- a/cfme/fixtures/artifactor_plugin.py
+++ b/cfme/fixtures/artifactor_plugin.py
@@ -150,12 +150,12 @@ def fire_art_hook(config, hook, **hook_args):
     if client is None:
         assert UNDER_TEST, 'missing artifactor is only valid for inprocess tests'
     else:
-        client.fire_hook(hook, **hook_args)
+        return client.fire_hook(hook, **hook_args)
 
 
 def fire_art_test_hook(node, hook, **hook_args):
     name, location = get_test_idents(node)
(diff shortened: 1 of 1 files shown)


sasoc:
cc @miha-plesko

miha-plesko:
@psav so this little change is needed in order to make merkyl work again. Currently, an error is raised because merkyl fixture expects a value to be returned by `fire_art_test_hook('get_log_merkyl')`, see here: https://github.com/ManageIQ/integration_tests/blob/master/cfme/fixtures/merkyl.py#L28

miha-plesko:
@izapolsk thanks for approve, can we get it in, please, because we have some dependent PRs that won't work without this fix.

PR #7800, opened 2018-09-11
[1LP][RFR][NOTEST] Removing test_custom_vm_report from RHV regression suite

jan-zmeskal:
So, why remove this from RHV-CFME regression suite?

1) This does not really test any RHV specific functionality. The main point of this test case is that custom report can be successfully created and therefore I don't see much reason to keep it in specific RHV-CFME integration suite.

2) This test case works quite problematically in you environment because:
2.1) The fixture `report_vms` [requires](https://github.com/ManageIQ/integration_tests/blob/23c846c40c23eef0d1f5b1c1619f026abb2665c7/cfme/tests/intelligence/reports/test_report_corresponds.py#L32) that there are **at least two VMs present on a provider**. Recently, I created some PRs that make sure that our tests clean after themselves. And since we have much cleaner env now, this test case started to fail. There just aren't two random VMs on RHV when it is running in our suite.
2.2) Even if there were two random VMs, the test case itself skips all the VMs that start with "test_" as you can see [here](https://github.com/ManageIQ/integration_tests/blob/23c846c40c23eef0d1f5b1c1619f026abb2665c7/cfme/tests/intelligence/reports/test_report_corresponds.py#L45). If our env, all the VMs that we use start with "test_", because all of them are (temporarily) created by test cases. We don't provide any VMs as part of our test env.

Now, I still think that the basic problem here is suboptimal design of the test case and fixture itself. I especially see the problem in the fact that `report_vms` depends on two VMs being present, but does not ensure this will be the case. Nor does it invoke any other fixture that would ensure this. That I think is a design problem. 
However, it probably works well for CFME QE and as I explained in point 1, I don't believe we need this test case. Therefore I suggest just to remove it from our suite.
@istein1, @rhrazdil please provide your opinion. If you're okay with that, I'll proceed and move the PR to [RFR] state.

### DIFF
--- a/cfme/tests/intelligence/reports/test_report_corresponds.py
+++ b/cfme/tests/intelligence/reports/test_report_corresponds.py
@@ -34,7 +34,6 @@ def report_vms(appliance, infra_provider):
 
 
 @pytest.mark.tier(3)
-@pytest.mark.rhv3
 @pytest.mark.meta(blockers=[BZ(1531600, forced_streams=['5.9'])])
 @test_requirements.report
 def test_custom_vm_report(soft_assert, report_vms):


rhrazdil:
@jan-zmeskal Nice elaboration, I agree.

istein1:
As the test adds no value for RHV testing, I agree it can be removed.

PR #7801, opened 2018-09-11
[1LP][RFR] Use specific keyserver.

jaryn:
The fedoraproject keyserver seems to always answer even repetetive
queries contrary to whatever is in the servers pool which is used by default.


mshriver:
@RonnyPfannschmidt Can you rebuild the PRT image for us please, to incorporate this change?

PR #7802, opened 2018-09-11
[1LP][RFR] Add logrotate tests.

jaryn:
__Adding tests__ for logrotate on appliance.

### DIFF
--- a/cfme/tests/cli/test_appliance_log_rotate.py
+++ b/cfme/tests/cli/test_appliance_log_rotate.py
@@ -1,0 +1,45 @@
+import datetime
+
+import dateutil
+
+
+LOGS = ["/var/www/miq/vmdb/log/evm.log",
+        "/var/opt/rh/rh-postgresql95/lib/pgsql/data/pg_log/postgresql.log"]
+
+
+def advance_appliance_date_by_day(appliance):
+    """Advance date on the appliance by one day."""
+    txt_date = appliance.ssh_client.run_command('date --rfc-3339=ns').output
+    appliance_date = dateutil.parser.parse(txt_date)
+    td = datetime.timedelta(days=1)
+    advanced_txt_date = (appliance_date + td).strftime("%Y-%m-%d %H:%M:%S%z")
+    appliance.ssh_client.run_command("date -s '{}'".format(advanced_txt_date))
+
+
+def test_appliance_log_rotate(temp_appliance_preconfig_funcscope):
+    """ Checks whether the log is logrotated daily.
+
+    Polarion:
+        assignee: jhenner
+        casecomponent: Appliance
+        initialEstimate: 1/12h
+        startsin: 5.6
+    """
+    appliance = temp_appliance_preconfig_funcscope
+    assert appliance.ssh_client.run_command("/etc/cron.daily/logrotate").success
+
+    initial_log_files = {}
+    for log_path in LOGS:
+        initial_log_files[log_path] = appliance.ssh_client.run_command(
+            "ls -1 {}*".format(log_path)).output.split('\n')
+        appliance.ssh_client.run_command(
+            "echo 'Ensure line in logs' >> {}".format(log_path))
+
+    # Perform the logrotate.
+    advance_appliance_date_by_day(appliance)
+    assert appliance.ssh_client.run_command("/etc/cron.daily/logrotate").success
+
+    for log_path in LOGS:
+        adv_time_log_files = appliance.ssh_client.run_command("ls -1 {}*".format(
+            log_path)).output.split('\n')
+        assert set(initial_log_files[log_path]) < set(adv_time_log_files)
--- a/cfme/tests/test_appliance.py
+++ b/cfme/tests/test_appliance.py
@@ -310,61 +310,6 @@ def test_appliance_executing_script():
 
 
 @pytest.mark.manual
-@pytest.mark.tier(2)
-def test_appliance_log_rotate_postgres():
-    """
-    [root@host-192-168-55-133 ~]# cd /var/opt/rh/rh-
-    postgresql95/lib/pgsql/data/pg_log
-    [root@host-192-168-55-133 pg_log]# ll
-    total 204
-    -rw-------. 1 postgres postgres 108218 Nov 10 14:39 postgresql.log
-    -rw-------. 1 postgres postgres  10776 Nov  3 03:34
-    postgresql.log-20161103.gz
-    -rw-------. 1 postgres postgres   8095 Nov  4 03:26
-    postgresql.log-20161104.gz
-    -rw-------. 1 postgres postgres   7544 Nov  5 03:20
-    postgresql.log-20161105.gz
-    -rw-------. 1 postgres postgres   8385 Nov  6 03:22
-    postgresql.log-20161106.gz
-    -rw-------. 1 postgres postgres   7637 Nov  7 03:26
-    postgresql.log-20161107.gz
-    -rw-------. 1 postgres postgres   8621 Nov  8 03:09
-    postgresql.log-20161108.gz
-    -rw-------. 1 postgres postgres  13083 Nov  9 03:28
-    postgresql.log-20161109.gz
-    -rw-------. 1 postgres postgres  14880 Nov 10 03:05
-    postgresql.log-20161110.gz
-    # are the log files being compressed and archived each day ( like
-    appliance_console.log-20161109.gz )
-
-    Polarion:
-        assignee: jhenner
-        casecomponent: Appliance
-        initialEstimate: 1/12h
-        startsin: 5.6
-    """
-    pass
-
-
-@pytest.mark.manual
-@pytest.mark.tier(2)
-def test_appliance_log_rotate_evm():
-    """
-    [root@host-192-168-55-133 pg_log]# vmdb
-    [root@host-192-168-55-133 vmdb]# cd log
-    # are the log files being compressed and archived each day ( like
-    appliance_console.log-20161109.gz )
-
-    Polarion:
-        assignee: jhenner
-        casecomponent: Appliance
-        initialEstimate: 1/12h
-        startsin: 5.6
-    """
-    pass
-
-
-@pytest.mark.manual
 @pytest.mark.tier(1)
 def test_appliance_log_crond():
     """


dajoRH:
Lint report for commit ede2196d5c63941e23377e4395c42195b3540945:

`cfme/tests/cli/test_appliance_log_rotate.py`:
- :red_circle: Line 1:1: **F401** *'pytest' imported but unused*
- :bangbang: Line 20:13: **E126** *continuation line over-indented for hanging indent*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/cli/test_appliance_log_rotate.py
```diff
@@ -0,0 +1,37 @@
+import datetime
+import dateutil
+
+
+def advance_appliance_date_by_day(appliance):
+    txt_date = appliance.ssh_client.run_command('date --rfc-3339=ns').output
+    appliance_date = dateutil.parser.parse(txt_date)
+    td = datetime.timedelta(days=1)
+    advanced_txt_date = (appliance_date + td).strftime("%Y-%m-%d %H:%M:%S%z")
+    appliance.ssh_client.run_command("date -s '{}'".format(advanced_txt_date))
```
jan-zmeskal: Should not we undo what we've caused there? I did not run these tests locally. However it seems to me that all the tests that will run _after_ these two will leave a wrong timestamp in the appliance logs.

digitronik: @jaryn Mine Intension is reducing the complexity as you are trying to increase timedelta by `1 day` in python. Its simple to fire single command in state of it using a shell. I would like to suggest you try this If fails let me know.

cfme/tests/cli/test_appliance_log_rotate.py
```diff
@@ -0,0 +1,38 @@
+import datetime
+
+import dateutil
+
+
+def advance_appliance_date_by_day(appliance):
+    txt_date = appliance.ssh_client.run_command('date --rfc-3339=ns').output
```
jawatts: Could you add a doc block here

cfme/tests/cli/test_appliance_log_rotate.py
```diff
@@ -0,0 +1,38 @@
...
+        log_path)).output.split('\n')
+    assert set(initial_log_files) < set(adv_time_log_files)
+
+
+def test_appliance_vmdb_log_rotate(temp_appliance_preconfig_funcscope):
+    log = "/var/www/miq/vmdb/log/evm.log"
+    log_rotate_check(temp_appliance_preconfig_funcscope, log)
+
+
+def test_appliance_postgres_log_rotate(temp_appliance_preconfig_funcscope):
```
jawatts: So these two methods look to be the same thing. Have you thought about creating a global variable ```CFME_LOGS```, for example, which would be a dict of the different logs to rotate and paramertize the test of ```CFME_LOGS``` This will reduce the

cfme/tests/cli/test_appliance_log_rotate.py
```diff
@@ -0,0 +1,38 @@
...
+
+import dateutil
+
+
+def advance_appliance_date_by_day(appliance):
+    txt_date = appliance.ssh_client.run_command('date --rfc-3339=ns').output
+    appliance_date = dateutil.parser.parse(txt_date)
+    td = datetime.timedelta(days=1)
+    advanced_txt_date = (appliance_date + td).strftime("%Y-%m-%d %H:%M:%S%z")
+    appliance.ssh_client.run_command("date -s '{}'".format(advanced_txt_date))
```
jawatts: I am debating this myself, but would it be good to reset the time back to what it was? @digitronik thoughts?

cfme/tests/cli/test_appliance_log_rotate.py
jan-zmeskal: If we're only going to have these two tests, then it's maybe not necessary. However I definitely agree that what @jawatts proposes would be a better design. If you plan to add more tests in the future, I also propose to go that way.

cfme/tests/cli/test_appliance_log_rotate.py
```diff
@@ -0,0 +1,38 @@
...
+    configured_appliance.ssh_client.run_command("/etc/cron.daily/logrotate")
+
+    initial_log_files = configured_appliance.ssh_client.run_command("ls -1 {}*".format(
+        log_path)).output.split('\n')
+    configured_appliance.ssh_client.run_command(
+        "echo 'Ensure line in logs' >> {}".format(log_path))
+    advance_appliance_date_by_day(configured_appliance)
+
+    # Perform the logrotate.
+    configured_appliance.ssh_client.run_command("/etc/cron.daily/logrotate")
```
jan-zmeskal: I would also consider asserting return code of this operation. It's just one line of code and might save some time during analysis when this test case fails. It's optional, though - I don't see it as required.

cfme/tests/cli/test_appliance_log_rotate.py
jaryn: Well the name of this function is saying it all so I didn't think it was necessary to document it.

jaryn: Well the name of this function is saying it all so I didn't think it was necessary to document it.

cfme/tests/cli/test_appliance_log_rotate.py
jaryn: The time is set on temp appliance, so I don't think it is necessary to reset the time back.

cfme/tests/cli/test_appliance_log_rotate.py
jaryn: No. I don't plan to add any more tests.

jaryn: No. I don't plan to add any more tests.

cfme/tests/cli/test_appliance_log_rotate.py
jaryn: The time is set on temp appliance, so I don't think it is necessary to reset the time back.

cfme/tests/cli/test_appliance_log_rotate.py
jaryn: I think the problem will be obvious from the traceback better than from any return code.

cfme/tests/cli/test_appliance_log_rotate.py
jaryn: No. I don't plan to add any more tests.

cfme/tests/cli/test_appliance_log_rotate.py
jaryn: OK, although I think the name says it all.

jaryn: OK, although I think the name says it all.

jaryn: OK, although I think the name says it all.

jaryn: OK, although I think the name says it all.

cfme/tests/cli/test_appliance_log_rotate.py
jawatts: @jaryn I would still paramertize this, I do not think we should make this assumption that no more tests will be added and it reduce code duplications

cfme/tests/cli/test_appliance_log_rotate.py
```diff
@@ -0,0 +1,39 @@
+import datetime
+
+import dateutil
+
+
+def advance_appliance_date_by_day(appliance):
+    '''Advance date on the appliance by one day.'''
```
jawatts: Please use double quotes

jaryn: Done

cfme/tests/cli/test_appliance_log_rotate.py
jawatts: @jaryn did you try @digitronik request here? Would this work for you?

cfme/tests/cli/test_appliance_log_rotate.py
jawatts: @jaryn please parameterize this.

jaryn:
@jawatts The results are visible even without the pytest statement. I think it is like so that in case the file starts with `test_` all `test_` methods in the file are executed as PR tests

cfme/tests/cli/test_appliance_log_rotate.py
jawatts: @jaryn you are correct here. :+1:

cfme/tests/cli/test_appliance_log_rotate.py
jawatts: @jaryn Please paramertize this test. This is a required change.

cfme/tests/cli/test_appliance_log_rotate.py
jawatts: @jaryn Since this is moving back to WIPTEST because of the parameterization comment, I would suggest trying @digitronik suggestion. It is more readable and simpler.

jawatts: @jan-zmeskal To answer your initial question here, I don't think it is necessary to reset the time here. This test is being run with a temp appliance.

cfme/tests/cli/test_appliance_log_rotate.py
```diff
@@ -0,0 +1,37 @@
...
+def advance_appliance_date_by_day(appliance):
+    """Advance date on the appliance by one day."""
+    txt_date = appliance.ssh_client.run_command('date --rfc-3339=ns').output
+    appliance_date = dateutil.parser.parse(txt_date)
+    td = datetime.timedelta(days=1)
+    advanced_txt_date = (appliance_date + td).strftime("%Y-%m-%d %H:%M:%S%z")
+    appliance.ssh_client.run_command("date -s '{}'".format(advanced_txt_date))
+
+
+@pytest.mark.parametrize('log_path', LOGS, ids=IDS)
```
jawatts: :+1:

cfme/tests/cli/test_appliance_log_rotate.py
```diff
@@ -0,0 +1,37 @@
...
+def test_appliance_log_rotate(temp_appliance_preconfig_funcscope, log_path):
+    ''' Checks whether the log is logrotated daily. '''
```
jawatts: **Required** This line should be double quoted

cfme/tests/cli/test_appliance_log_rotate.py
```diff
@@ -0,0 +1,37 @@
...
+    appliance = temp_appliance_preconfig_funcscope
+    appliance.ssh_client.run_command("/etc/cron.daily/logrotate")
```
jawatts: @jaryn What do you think about checking the return code here  and in line 33 below, if the ssh command fails the rest of the test fails.

cfme/tests/cli/test_appliance_log_rotate.py
```diff
@@ -0,0 +1,44 @@
...
+    """ Checks whether the log is logrotated daily.
+
+    Polarion:
+        assignee: jhenner
+        casecomponent: appl
+        initialEstimate: 1/12h
+        startsin: 5.6
+    """
+    appliance = temp_appliance_preconfig_funcscope
+    assert appliance.ssh_client.run_command("/etc/cron.daily/logrotate").success
```
izapolsk: what is it for ?

cfme/tests/cli/test_appliance_log_rotate.py
```diff
@@ -0,0 +1,44 @@
+import datetime
+
+import dateutil
+import pytest
+
+LOGS = ["/var/www/miq/vmdb/log/evm.log",
```
izapolsk: Please consider comparing both files for one test run. Your test requests new appliance for every log file,  it doesn't look efficient to me

dajoRH:
*CFME QE Bot*

dajoRH:
Lint report for commit a77ecb1a8b01b66c7d22c69a3ddc7c00b9ba4369:

`cfme/tests/cli/test_appliance_log_rotate.py`:

**No commit flaws detected**.

*CFME QE Bot*

PR #7776, opened 2018-09-07
[1LP][RFR] Fixing getting retirement state

jan-zmeskal:
Main point of this PR is the fact that field 'Retirement state' in VM details view has been changed to 'Retirement State' in 5.10. However, once I fixed that I also encountered a product [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1626971), that is influencing all tests that invoke [set_retirement_date](https://github.com/ManageIQ/integration_tests/blob/401acc6d20e8860710799ae95bd57eb6eeb5ada9/cfme/common/vm.py#L764) method. Therefore providing a BZ wrapper.

Jenkins: https://rhv-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/CFME-DEV/job/cfme-5.9-rhv-4.2-integration-flow-dev/332/console

PR #7804, opened 2018-09-11
[WIPTEST] Adding BZ1627758 and fixing test_retirement_now

jan-zmeskal:
1) Why am I modifying `verify_retirement_state`?
Because in CFME 5.10, you are redirected to a view with Requests when you do Lifecycle -> Retire VM.
I did not use `VersionPicker` since call to `navigate_to(retire_vm, 'Details')` does not cause any problem in 5.9 - we're already there.

2) Why am I wrapping so many tests with [BZ1627758](https://bugzilla.redhat.com/show_bug.cgi?id=1627758)?
Recently there was a change in datepicker. Its timezone is now determined by local time zone of your browser. However, the success flash message is always displayed in UTC. Therefore we cannot really verify now that the flash message displayed after setting retirement date is correct. 

Example: I am running this test case in my time zone (CEST) and set retirement date to 12th of September 23:00. The flash message will say that it was set to 12th of September 21:00 **UTC**. That's because CEST is UTC+2. If the test case with the very same configuration is run let's say in India Standard Time (UTC+5:30), the flash message will say that retirement date was set to 13th of September 4:30. 
In the current state, we have no way how to check the flash message deterministically (at least no that I know of). 

Honestly, I think this is more of a bug than a feature. However, the dev seem to [disagree](https://bugzilla.redhat.com/show_bug.cgi?id=1626971#c10). Therefore I created this [RFE](https://bugzilla.redhat.com/show_bug.cgi?id=1627758) to make time zone of datepicker configurable. If we have this, we can set it to UTC and reliably check the flash message.

{{pytest: cfme/tests/cloud_infra_common/test_retirement.py::test_retirement_now --long-running -vv}}

### DIFF
--- a/cfme/tests/automate/test_common_methods.py
+++ b/cfme/tests/automate/test_common_methods.py
@@ -9,6 +9,7 @@ from cfme import test_requirements
 from cfme.infrastructure.virtual_machines import InfraVmSummaryView
 from cfme.infrastructure.provider import InfraProvider
 from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.blockers import BZ
 from cfme.utils.generators import random_vm_name
 from cfme.utils.log import logger
 from cfme.utils.wait import wait_for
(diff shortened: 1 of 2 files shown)


jan-zmeskal:
Hi @mshriver, thank you for merging this so promptly. The thing is though that I wanted to add a comment there that the fix of `verify_retirement_state` won't really work until #7776 is merged.

PR #7808, opened 2018-09-11
[1LP][RFR] Fix test generation in test_register_appliance

mshriver:
This was not recognizing reg_method argname, changing the parametrize call and scope resolves this, while maintaining the same parametrization result.

{{ pytest: --long-running cfme/tests/configure/test_register_appliance.py -k test_rh_registration }}


mshriver:
PRT failures are timeout errors and things not related to the fixture exception this was causing.

PR #7812, opened 2018-09-12
[RFR] Don't crash fixture if Merkyl plugin returns bad response

miha-plesko:
With this commit we prevent crash of entire `merkyl_inspector` fixture in case Merkyl plugin timeouts. Thing is that Merkyl service on appliance sometimes timeouts before replying to our requests. And currently we crash hard when this occurs. But we shouldn't - because Merkyl service usually returns valid response upon next probe.

So with this commit we simulate empty response as opposed to crashing. The real fix, obviously, would be to fix merkyl service itself, but that's out of our current scope.

### DIFF
--- a/cfme/fixtures/merkyl.py
+++ b/cfme/fixtures/merkyl.py
@@ -28,7 +28,7 @@ class MerkylInspector(object):
         res = fire_art_test_hook(
             self.node, 'get_log_merkyl', ip=self.ip,
             filename=log_name, grab_result=True)
-        return res['merkyl_content']
+        return res.get('merkyl_content', '')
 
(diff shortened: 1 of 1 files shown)


miha-plesko:
@izapolsk so here we have another tiny framework fix that caused us quite some headaches :) I suspect Merkyl timeouts happen if evm.log file gets too large.

@psav do you have any similar experience with this? Would you find it useful if I provide a PR on https://github.com/psav/merkyl to support limiting number of returned log lines?

Something like:
<pre>
http://10.10.43.130:8192/get/evm.log<b>?tail=1000</b>
</pre>
would then return only last 1k lines of evm.log.

PR #7818, opened 2018-09-12
[1LP][RFR]Adding new Test: Migration Plans Sorting type/Order

kedark3:
Purpose or Intent
=================

__Adding tests__ for Migration Plans Sorting type/Order and Archive plans

{{pytest: cfme/tests/v2v/test_v2v_migrations_ui.py --use-provider rhv42 --use-provider vsphere67-nested --provider-limit 2 -vvvv --long-running -k 'test_v2v_ui_migration_plan_sorting' }}

### DIFF
--- a/cfme/tests/v2v/test_v2v_migrations_ui.py
+++ b/cfme/tests/v2v/test_v2v_migrations_ui.py
@@ -4,11 +4,13 @@ import pytest
 
 from widgetastic.exceptions import NoSuchElementException
 
+from cfme.fixtures.provider import rhel7_minimal
 from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.markers.env_markers.provider import ONE_PER_VERSION
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.tests.services.test_service_rbac import new_user, new_group, new_role
+from cfme.utils.wait import wait_for
 from cfme.v2v.migrations import MigrationPlanRequestDetailsView
 
 pytestmark = [
@@ -245,6 +247,73 @@ def test_v2v_ui_set2(appliance, v2v_providers, form_data_single_datastore, soft_
     soft_assert(set(view.migration_request_details_list.read()) == set(vm_selected))
     view = navigate_to(infrastructure_mapping_collection, 'All')
     view.migration_plans_not_started_list.migrate_plan(plan_name)
+
+    # Test Archive completed migration  plan
+    wait_for(func=view.plan_in_progress, func_args=[plan_name], delay=5, num_sec=300,
+        handle_exception=True)
+    view.migr_dropdown.item_select("Completed Plans")
+    view.wait_displayed()
+    view.migration_plans_completed_list.archive_plan(plan_name)
+    view.migr_dropdown.item_select("Archived Plans")
+    view.wait_displayed()
+    soft_assert(plan_name in view.migration_plans_archived_list.read())
+
+
+@pytest.mark.parametrize('form_data_multiple_vm_obj_single_datastore', [['nfs', 'nfs',
+    [rhel7_minimal, rhel7_minimal]]], indirect=True)
+def test_v2v_ui_migration_plan_sorting(appliance, v2v_providers,
+        form_data_multiple_vm_obj_single_datastore, soft_assert):
+    infrastructure_mapping_collection = appliance.collections.v2v_mappings
+    migration_plan_collection = appliance.collections.v2v_plans
+
+    mapping = infrastructure_mapping_collection.create(
+        form_data_multiple_vm_obj_single_datastore[0])
+
+    # creating two plans, with 1 VM in each, hence vm_list[0]/[1]
+    plan1 = migration_plan_collection.create(
+        name="plan_{}".format(fauxfactory.gen_alphanumeric()), description="desc_{}"
+        .format(fauxfactory.gen_alphanumeric()), infra_map=mapping.name,
+        vm_list=[form_data_multiple_vm_obj_single_datastore.vm_list[0]], start_migration=False)
+    plan2 = migration_plan_collection.create(
+        name="plan_{}".format(fauxfactory.gen_alphanumeric()), description="desc_{}"
+        .format(fauxfactory.gen_alphanumeric()), infra_map=mapping.name,
+        vm_list=[form_data_multiple_vm_obj_single_datastore.vm_list[1]], start_migration=False)
+
+    view = navigate_to(migration_plan_collection, 'All')
+
+    plans_list_before_sort = view.migration_plans_not_started_list.read()
+    view.sort_direction.click()
+    plans_list_after_sort = view.migration_plans_not_started_list.read()
+    soft_assert(plans_list_before_sort != plans_list_after_sort)
+
+    view.migration_plans_not_started_list.schedule_migration(plan1.name, after_mins=5)
+    view.wait_displayed()
+    view.migration_plans_not_started_list.schedule_migration(plan2.name, after_mins=10)
+    view.wait_displayed()
+    view.sort_type_dropdown.item_select("Scheduled Time")
+    plans_list_before_sort = view.migration_plans_not_started_list.read()
+    view.sort_direction.click()
+    plans_list_after_sort = view.migration_plans_not_started_list.read()
+    soft_assert(plans_list_before_sort != plans_list_after_sort)
+    view.migration_plans_not_started_list.schedule_migration(plan1.name)
+    view.wait_displayed()
+    view.migration_plans_not_started_list.schedule_migration(plan2.name)
+
+    view.migration_plans_not_started_list.migrate_plan(plan1.name)
+    view.migr_dropdown.item_select("Not Started Plans")
+    view.wait_displayed()
+    view.migration_plans_not_started_list.migrate_plan(plan2.name)
+    view.migr_dropdown.item_select("In Progress Plans")
+    wait_for(func=view.plan_in_progress, func_args=[plan1.name], delay=5, num_sec=300,
+        handle_exception=True)
+    wait_for(func=view.plan_in_progress, func_args=[plan2.name], delay=5, num_sec=300,
+        handle_exception=True)
+    view.migr_dropdown.item_select("Completed Plans")
+    view.wait_displayed()
+    plans_list_before_sort = view.migration_plans_completed_list.read()
+    view.sort_direction.click()
+    plans_list_after_sort = view.migration_plans_completed_list.read()
+    soft_assert(plans_list_before_sort != plans_list_after_sort)
 
 
 def test_migration_rbac(appliance, new_credential, v2v_providers):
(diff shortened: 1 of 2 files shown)


digitronik:
@kedark3 please maintain PR flow
https://github.com/ManageIQ/integration_tests/wiki/PR-Process

dajoRH:
*CFME QE Bot*

kedark3:
Tests failing in PRT due to Browser being old. Works on Chrome on my local system.

digitronik:
@kedark3 please provide local results in such case

kedark3:
@digitronik @jawatts @mshriver please review. 
Note: /pr/run/26848 did pass, it failed in TEARDOWN while trying to delete the VM, seems like a vCenter issue. Nothing to do with testm test did pass.

widgetastic_manageiq/__init__.py
```diff
@@ -4274,12 +4274,37 @@ def schedule_migration(self, plan_name, cancel=False, after_mins=1):
         try:
             el = self._get_plan_element(plan_name)
             schedule_button = self.browser.element(self.ITEM_SCHEDULE_BUTTON_LOCATOR, parent=el)
```
digitronik: Optional:
I would use a `widget` inside a `widget` for such things.

widgetastic_manageiq/__init__.py
```diff
@@ -4274,12 +4274,37 @@ def schedule_migration(self, plan_name, cancel=False, after_mins=1):
...
-                for i in range(after_mins):
-                    self.root_browser.click(self.ITEM_MODAL_MINUTE_INCREMENT_LOCATOR)
+                try:
+                    for i in range(after_mins):
+                        self.root_browser.click(self.ITEM_MODAL_MINUTE_INCREMENT_LOCATOR)
+                except NoSuchElementException:  # Throws this in FF browser
+                    self.logger.info(
+                        "Schedule using dateTimePicker failed, using datetime and timedelta."
+                    )
+                    from datetime import datetime, timedelta
```
digitronik: Required:
Remove this import and put in a header as we already imported `date` from `datetime`  lib.

widgetastic_manageiq/__init__.py
```diff
@@ -4274,12 +4274,37 @@ def schedule_migration(self, plan_name, cancel=False, after_mins=1):
...
+
+                    current_time = datetime.strptime(
+                        schedule_time_input.get_attribute("value"), "%m/%d/%Y %I:%M %p"
+                    )
+                    schedule_time_input.clear()
+                    schedule_time_input.send_keys(
+                        datetime.strftime(
+                            current_time + timedelta(seconds=60 * after_mins), "%m/%d/%Y %I:%M %p"
```
digitronik: Required:
No need to convert `minutes` into `seconds`. Its core python powerfull implementation you can directly go with `minutes` tag; lib will take care of it. 
```
In [7]: timedelta(minutes=1)
Out[7]: datetime.timedelta(0, 60)

In [8]: timedelta(seconds=1 * 60)
Out[8]: datetime.timedelta(0, 60)
```

cfme/tests/v2v/test_v2v_migrations_ui.py
```diff
@@ -241,3 +243,69 @@ def test_v2v_ui_set2(appliance, v2v_providers, form_data_single_datastore, soft_
     soft_assert(set(view.migration_request_details_list.read()) == set(vm_selected))
     view = navigate_to(infrastructure_mapping_collection, 'All')
     view.migration_plans_not_started_list.migrate_plan(plan_name)
+
+    # Test Archive completed migration  plan
+    wait_for(func=view.plan_in_progress, func_args=[plan_name], delay=5, num_sec=300,
+        handle_exception=True)
+    view.migr_dropdown.item_select("Completed Plans")
+    view.wait_displayed()
```
digitronik: I am suspect about such use of `wait_displayed` every time. I think `wt` internally take care of this.

cfme/tests/v2v/test_v2v_migrations_ui.py
```diff
@@ -241,3 +243,69 @@ def test_v2v_ui_set2(appliance, v2v_providers, form_data_single_datastore, soft_
...
+    wait_for(func=view.plan_in_progress, func_args=[plan1.name], delay=5, num_sec=300,
+        handle_exception=True)
+    wait_for(func=view.plan_in_progress, func_args=[plan2.name], delay=5, num_sec=300,
+        handle_exception=True)
+    view.migr_dropdown.item_select("Completed Plans")
+    view.wait_displayed()
+    plans_list_before_sort = view.migration_plans_completed_list.read()
+    view.sort_direction.click()
+    plans_list_after_sort = view.migration_plans_completed_list.read()
+    soft_assert(not (plans_list_before_sort == plans_list_after_sort))
```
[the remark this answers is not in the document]
kedark3: Nikhil, I think it is appropriate to use `soft_assert` here. At each of the assertion, I want to know what the result was and I also don't want my test to fail immediately if any of these assertions do not hold up, hence not using `assert`. Please help me understand why do you think its not necessary.

cfme/tests/v2v/test_v2v_migrations_ui.py
```diff
@@ -241,3 +243,69 @@ def test_v2v_ui_set2(appliance, v2v_providers, form_data_single_datastore, soft_
...
+
+    view.migration_plans_not_started_list.schedule_migration(plan1.name, after_mins=5)
+    view.wait_displayed()
+    view.migration_plans_not_started_list.schedule_migration(plan2.name, after_mins=10)
+    view.wait_displayed()
+    view.sort_type_dropdown.item_select("Scheduled Time")
+    plans_list_before_sort = view.migration_plans_not_started_list.read()
+    view.sort_direction.click()
+    plans_list_after_sort = view.migration_plans_not_started_list.read()
+    soft_assert(not (plans_list_before_sort == plans_list_after_sort))
```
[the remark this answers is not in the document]
kedark3: Do you mean instead of `not` I should use `!=` ? If so, I really don't have an explanation on why. I just picked whatever came to my mind :) Do you think one is better than other?

widgetastic_manageiq/__init__.py
kedark3: `Button` ? I think I tried that and for some reason did not work that I can't recall.

widgetastic_manageiq/__init__.py
kedark3: I think I did try that earlier and since it didn't work I switched to `seconds`.  I'll try again and replace.

widgetastic_manageiq/__init__.py
kedark3: Ok, I can do that. I was thinking that we will not import this unless we need to use it. hence put it in the code there instead of top. Do you still want me to put it on top?

cfme/tests/v2v/test_v2v_migrations_ui.py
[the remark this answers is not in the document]
digitronik: != is better option in such case

cfme/tests/v2v/test_v2v_migrations_ui.py
kedark3: So these `view.wait_displayed()` are required here. When I navigate to this `view` initially, `wt` will do the wait by default as you said. Then what I do is, I am selecting different dropdowns in that view, and each dropdown item opens a different section, which shows `spinner` for brief duration, so using `view.wait_displayed()` I am really avoiding any operations before the spinner is gone. If I remove the wait_displayed() the test would pretty much be unreliable and fail inconsistently. You can try this on your local system and see what happens.

widgetastic_manageiq/__init__.py
digitronik: Yes, It should at import only. Datetime already imported so better to import in same.

kedark3:
Thanks for Reviews @digitronik , please elaborate more on :+1: 

"This widget is not directly part of wt.core or wt.pfy it's optional suggestion but while implementing widget we have to use a model of widget inside widget

I found ClickableMixin and extract text from that web element. You can do it easily by assigning that locator/id to base widgets like Text and others. It will reduce LOC and readability."

Do you want me to implement a Widget for DateTimeInput box?

cfme/tests/v2v/test_v2v_migrations_ui.py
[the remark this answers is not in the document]
kedark3: Can you help me understand the rationale. I did search a little bit on the internet to try to find what is recommended or preferred  and why. I did not find a good answer. Also, in our codebase, there are instances of `not` being used with `==` just like this, so not sure why they are allowed, but this one is not. Trying to expand my knowledge of the topic. Thanks.

widgetastic_manageiq/__init__.py
kedark3: Can you point me to an example please?

digitronik: @kedark3 you can refer wt.core or wt.pfy you will get lots of examples.

cfme/tests/v2v/test_v2v_migrations_ui.py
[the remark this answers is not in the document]
digitronik: `!=`, `==` etc are comparator operators. Those are pointing to an equality test according to their `__eq__` or `__cmp__` methods.  This method return type is `bool`.

Now, In your our case, `A != B` will return us exact output what we expect. but we are doing   `A == B` it returning `bool` out and then with negation operator (`not`) we are trying to get what we expected.

```
In [1]: A = 1
In [2]: B = 2
In [3]: x = A == B
In [4]: x
Out[4]: False
In [5]: not(x)
Out[5]: True

In [6]: # If we have a facility and direct operator why not to use it?
In [7]: A!=B
Out[7]: True
```
As per my basic understanding
`!=` is `c_operator` and `not` is `py_operator` as `cpython` use `c` as base lang.

If you like to go in deep then let me know. I would like to point you `not` implementation from `CPython`.

kedark3:
To be merged after #7866

dajoRH:
*CFME QE Bot*

cfme/tests/v2v/test_v2v_migrations_ui.py
```diff
@@ -243,6 +245,72 @@ def test_v2v_ui_set2(appliance, v2v_providers, form_data_single_datastore, soft_
...
+    infrastructure_mapping_collection = appliance.collections.v2v_mappings
+    migration_plan_collection = appliance.collections.v2v_plans
+
+    mapping = infrastructure_mapping_collection.create(
+        form_data_multiple_vm_obj_single_datastore[0])
+
+    plan1 = migration_plan_collection.create(
+        name="plan_{}".format(fauxfactory.gen_alphanumeric()), description="desc_{}"
+        .format(fauxfactory.gen_alphanumeric()), infra_map=mapping.name,
+        vm_list=[form_data_multiple_vm_obj_single_datastore[1][0]], start_migration=False)
```
mshriver: isn't `form_data_multuiple_vm_ob_single_datastore` returning a namedtuple now? Need a comment at the very least for what we're expecting these indices to be.

kedark3: fixed.

PR #7833, opened 2018-09-17
[1LP][RFR] Support reset log for merkyl plugin

sasoc:
With this commit we implement method that resets log for merkyl plugin. Now we can clear content of a log and continue tailing it.

This is needed when a log is tailed for a duration of the whole session and there are multiple tests being ran. For example two test can search for the same log entry. If first one succeeds and we don't clear the log content after, second one will just find the log content of the first test and succeed.

### DIFF
--- a/artifactor/plugins/merkyl.py
+++ b/artifactor/plugins/merkyl.py
@@ -39,6 +39,7 @@ class Merkyl(ArtifactorBasePlugin):
         self.register_plugin_hook("teardown_merkyl", self.finish_session)
         self.register_plugin_hook("get_log_merkyl", self.get_log)
         self.register_plugin_hook("add_log_merkyl", self.add_log)
+        self.register_plugin_hook("reset_log_merkyl", self.reset_log)
 
     def configure(self):
         self.files = self.data.get("log_files", [])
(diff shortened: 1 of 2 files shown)


sasoc:
cc @miha-plesko

artifactor/plugins/merkyl.py
```diff
@@ -71,6 +72,15 @@ def get_log(self, test_name, test_location, filename):
         content = doc.content
         return {"merkyl_content": content}, None
 
+    @ArtifactorBasePlugin.check_configured
+    def reset_log(self, test_name, test_location, filename):
+        test_ident = "{}/{}".format(test_location, test_name)
```
digitronik: Required:
Please use keys for string formatting

artifactor/plugins/merkyl.py
```diff
@@ -71,6 +72,15 @@ def get_log(self, test_name, test_location, filename):
...
+        ip = self.tests[test_ident].ip
+
+        base, tail = os.path.split(filename)
+        url = "http://{}:{}/reset/{}".format(ip, self.port, tail)
```
digitronik: Required:
as about formatting

artifactor/plugins/merkyl.py
```diff
@@ -71,6 +72,15 @@ def get_log(self, test_name, test_location, filename):
         content = doc.content
         return {"merkyl_content": content}, None
 
+    @ArtifactorBasePlugin.check_configured
+    def reset_log(self, test_name, test_location, filename):
+        test_ident = "{}/{}".format(test_location, test_name)
+        ip = self.tests[test_ident].ip
+
+        base, tail = os.path.split(filename)
```
digitronik: Required:
`base` not going to use in code please go with ` _, tail = os.path.split(filename)`

sasoc:
@digitronik Thank you for your review, I've made the changes you requested.

miha-plesko:
@psav this one is pretty straight-forward, it just slightly extends the Merkyl plugin/fixture, yet needed for our test.

PR #7836, opened 2018-09-17
[WIP] Add test_router_crud

jaryn:
__Adding tests__ for router CRUD.

juwatts adding pytest:
{{ pytest: cfme/tests/networks/test_sdn_crud.py --use-provider complete }}

### DIFF
--- a/cfme/networks/network_router.py
+++ b/cfme/networks/network_router.py
@@ -42,7 +42,9 @@ class NetworkRouter(Taggable, BaseEntity):
         """Deletes current cloud router"""
         view = navigate_to(self, 'Details')
         view.toolbar.configuration.item_select('Delete this Router', handle_alert=True)
-        view.flash.assert_success_message('Delete initiated for 1 Network Router.')
+        wait_for(lambda: view.flash.messages, fail_condition=[], num_sec=10, delay=.1)
+        message = 'Delete initiated for 1 Network Router.'
+        view.flash.assert_success_message(message)
 
     def edit(self, name=None, change_external_gw=None, ext_network=None, ext_network_subnet=None):
         """Edit this router
--- a/cfme/networks/network_router.py
+++ b/cfme/networks/network_router.py
@@ -185,6 +185,10 @@ class Details(CFMENavigateStep):
     def step(self):
         self.prerequisite_view.entities.get_entity(name=self.obj.name, surf_pages=True).click()
 
+    def resetter(self):
+        """Reset the view"""
+        self.view.browser.refresh()
+
 
 @navigator.register(NetworkRouterCollection, 'Add')
 class AddRouter(CFMENavigateStep):
--- a/cfme/networks/views.py
+++ b/cfme/networks/views.py
@@ -445,7 +445,7 @@ class NetworkRouterAddView(BaseLoggedInPage):
 
     @property
     def is_displayed(self):
-        return (super(BaseLoggedInPage, self).is_displayed and
+        return (self.logged_in_as_current_user and
                 self.navigation.currently_selected == ['Networks', 'Network Routers'] and
                 self.title.text == 'Add New Network Router')
 
--- a/cfme/networks/views.py
+++ b/cfme/networks/views.py
@@ -434,6 +434,7 @@ class NetworkRouterDetailsView(BaseLoggedInPage):
 
 class NetworkRouterAddView(BaseLoggedInPage):
     """ Represents Add NetworkRouters page """
+    title = Text('//div[@id="main-content"]//h1')
     network_manager = Select(id='ems_id')
     router_name = TextInput(name='name')
     ext_gateway = BootstrapSwitch(name='external_gateway')
@@ -442,7 +443,11 @@ class NetworkRouterAddView(BaseLoggedInPage):
     cloud_tenant = Select(name='cloud_tenant_id')
     add = Button('Add')
 
-    is_displayed = displayed_not_implemented
+    @property
+    def is_displayed(self):
+        return (super(BaseLoggedInPage, self).is_displayed and
+                self.navigation.currently_selected == ['Networks', 'Network Routers'] and
+                self.title.text == 'Add New Network Router')
 
 
 class NetworkRouterEditView(BaseLoggedInPage):
@@ -546,7 +551,7 @@ class SecurityGroupDetailsView(BaseLoggedInPage):
     @property
     def is_displayed(self):
         return (super(BaseLoggedInPage, self).is_displayed and
-                self.navigation.currently_selected == ['Networks', 'Security Groups'] and
+                self.navigation.currently_selected == ['Networks', 'Providers'] and
                 self.title.text == '{name} (Summary)'.format(name=self.context['object'].name))
 
 
--- a/cfme/tests/networks/test_sdn_crud.py
+++ b/cfme/tests/networks/test_sdn_crud.py
@@ -1,9 +1,11 @@
 import pytest
+import fauxfactory
 
 from cfme.cloud.provider.azure import AzureProvider
 from cfme.cloud.provider.ec2 import EC2Provider
 from cfme.cloud.provider.gce import GCEProvider
 from cfme.cloud.provider.openstack import OpenStackProvider
+from cfme.utils.wait import wait_for
 from cfme.utils.appliance.implementations.ui import navigate_to
 
 pytestmark = [
@@ -41,3 +43,25 @@ def test_sdn_crud(provider, appliance):
     provider.wait_for_delete()
 
     assert not network_provider.exists
+
+
+@pytest.mark.ignore_stream('5.9')
+@pytest.mark.provider([OpenStackProvider], scope='function', override=True)
+def test_router_crud(provider, appliance):
+    router_name = 'test_router_crud-{}'.format(fauxfactory.gen_alphanumeric().lower())
+    routers = appliance.collections.network_routers
+    router = routers.create(name=router_name,
+                            provider=provider,
+                            tenant=provider.data.networks.cloud_tenant,
+                            network_manager=provider.data.networks.network_manager)
+    assert router.exists
+
+    router.delete()
+    provider.refresh_provider_relationships()
+
+    def router_doesnt_exists():
+        # Navigate somewhere else then to Router Details to enforce page refresh.
+        collection = appliance.collections.network_providers.filter({'provider': provider})
+        return router.name not in (r.name for r in collection.all())
+
+    wait_for(router_doesnt_exists, fail_condition=False, timeout=240, delay=1)
--- a/cfme/tests/networks/test_sdn_crud.py
+++ b/cfme/tests/networks/test_sdn_crud.py
@@ -59,9 +59,4 @@ def test_router_crud(provider, appliance):
     router.delete()
     provider.refresh_provider_relationships()
 
-    def router_doesnt_exists():
-        # Navigate somewhere else then to Router Details to enforce page refresh.
-        collection = appliance.collections.network_providers.filter({'provider': provider})
-        return router.name not in (r.name for r in collection.all())
-
-    wait_for(router_doesnt_exists, fail_condition=False, timeout=240, delay=1)
+    wait_for(lambda: not router.exists, fail_condition=False, timeout=240, delay=1)


cfme/networks/network_router.py
```diff
@@ -42,7 +42,10 @@ def delete(self):
         """Deletes current cloud router"""
         view = navigate_to(self, 'Details')
         view.toolbar.configuration.item_select('Delete this Router', handle_alert=True)
-        view.flash.assert_success_message('Delete initiated for 1 Network Router.')
+        message = 'Delete initiated for 1 Network Router.'
+        wait_for(lambda: message in (m.read() for m in view.flash.messages),
```
jawatts: Suggestion, why not ```wait_for(lambda: view.flash.messages, fail_func=[], num_sec=10, delay=.1)``` and then assert ```view.flash.assert_success_message('Delete initiated for 1 Network Router.')```

cfme/tests/networks/test_sdn_crud.py
```diff
@@ -41,3 +43,21 @@ def test_sdn_crud(provider, appliance):
     provider.wait_for_delete()
 
     assert not network_provider.exists
+
+
+@pytest.mark.uncollectif(lambda appliance: appliance.version < '5.10',
```
jawatts: **Required** Please use ```@pytest.mark.ignore_stream('5.9')```

cfme/tests/networks/test_sdn_crud.py
```diff
@@ -41,3 +43,21 @@ def test_sdn_crud(provider, appliance):
...
+                         reason='Delete button was added in CFME 5.10.')
+@pytest.mark.provider([OpenStackProvider], scope='function', override=True)
+def test_router_crud(provider, appliance):
+    router_name = 'test_router_crud-{}'.format(fauxfactory.gen_alphanumeric().lower())
+    routers = appliance.collections.network_routers
+    router = routers.create(router_name, provider, provider.data.networks.cloud_tenant,
```
jawatts: **Optional** Please pass the keys along with the values in this create call and break on separate lines for each key/value.

cfme/tests/networks/test_sdn_crud.py
```diff
@@ -41,3 +43,21 @@ def test_sdn_crud(provider, appliance):
...
+                            provider.data.networks.network_manager)
+    router.delete()
```
jawatts: **Required** I believe it would be a good idea to check that the router was successfully created and being tracked by CFME

cfme/tests/networks/test_sdn_crud.py
```diff
@@ -41,3 +43,21 @@ def test_sdn_crud(provider, appliance):
...
+    provider.refresh_provider_relationships()
+
+    def router_doesnt_exists():
```
jawatts: **Question** Why not just use the NetworkRouter exists method?

cfme/tests/networks/test_sdn_crud.py
```diff
@@ -41,3 +43,21 @@ def test_sdn_crud(provider, appliance):
...
+        router = (r for r in appliance.collections.network_routers.all() if r.name == router_name)
+        return not any(router)
+
+    wait_for(router_doesnt_exists, fail_condition=False, timeout=120, delay=1)
```
jawatts: **Optional** I think it would be good to have an assert here. 

example:
```router_exists, _ = wait_for(lambda: router.exists, fail_condition=False, timeout=120, delay=1)
   assert router_exists
```

or

``` assert not router.exists```

**Comment by dajoRH** [comment]
Lint report for commit 7cfd3153a0ddb9a82c0eee7a62d9e0e6d46a3f26:

`cfme/tests/networks/test_sdn_crud.py`:
- :red_circle: Line 53:5: **F841** *local variable 'router' is assigned to but never used*

**No commit flaws detected**.

*CFME QE Bot*

File: cfme/tests/networks/test_sdn_crud.py [hunk 7]
```diff
@@ -41,3 +43,26 @@ def test_sdn_crud(provider, appliance):
...
+                            tenant=provider.data.networks.cloud_tenant,
+                            network_manager=provider.data.networks.network_manager)
+    assert router.exists
+
+    router.delete()
+    provider.refresh_provider_relationships()
+
+    def router_doesnt_exists():
+        # Navigate somewhere else then to Router Details to enforce page refresh.
+        collection = appliance.collections.network_providers
```
jaryn: @mshriver @quarckster Hi guys. The router.exists goes to router Details page to check whether the router exists, but once it is there, it doesn't try to refresh. Do you guys have some suggestion how can I do the refresh in some nicer way then navigating here and there?

cfme/networks/network_router.py
mshriver: so, @jawatts suggestion only waits for there to be a flash message, not a matching flash message. Then once there is a flash message, it calls `assert_success_message` - which does two things. First it asserts there are no error messages, and second it asserts that the flash message given (Delete initiated...) is a success type message, and exists.  assert_message or assert_success_message handle getting the text from each `FlashMessage` widget generated by `FlashMessages`.

cfme/tests/networks/test_sdn_crud.py
mshriver: You can implement a `resetter()` method for the 'details' navigation destination that calls browser or UI toolbar refresh widgets.  This will be called after `prerequisite` and `step`.

mshriver: Note `resetter` is called even if navigation isn't needed (its on details page already, skips all navigation, but will call resetter.

dajoRH:
*CFME QE Bot*

cfme/tests/networks/test_sdn_crud.py
jaryn: There are problems with the `not router.exists()` approach. It is simpler to just list all the routers and check whether the router is among them then to use negative of `router.exists`

dajoRH:
*CFME QE Bot*

cfme/tests/networks/test_sdn_crud.py
jawatts: @jaryn question above still remains. Why not use the existing ```exist``` method?

jawatts:
One questions and restarting PRT with updated pytest statement

cfme/tests/networks/test_sdn_crud.py
```diff
@@ -41,3 +43,25 @@ def test_sdn_crud(provider, appliance):
...
+
+    router.delete()
+    provider.refresh_provider_relationships()
+
+    def router_doesnt_exists():
+        # Navigate somewhere else then to Router Details to enforce page refresh.
+        collection = appliance.collections.network_providers.filter({'provider': provider})
+        return router.name not in (r.name for r in collection.all())
+
+    wait_for(router_doesnt_exists, fail_condition=False, timeout=240, delay=1)
```
jawatts: If using the previous exists method, this could be updated to 
```wait_for(lamba: not router.exists, fail_condition=False, timeout=240, delay=1)```

File: cfme/tests/networks/test_sdn_crud.py [hunk 9]
```diff
@@ -41,3 +43,25 @@ def test_sdn_crud(provider, appliance):
...
+                            provider=provider,
+                            tenant=provider.data.networks.cloud_tenant,
+                            network_manager=provider.data.networks.network_manager)
+    assert router.exists
+
+    router.delete()
+    provider.refresh_provider_relationships()
+
+    def router_doesnt_exists():
+        # Navigate somewhere else then to Router Details to enforce page refresh.
```
jawatts: @jaryn Why not use the exists ```exists``` method: https://github.com/ManageIQ/integration_tests/blob/master/cfme/networks/network_router.py#L74

cfme/tests/networks/test_sdn_crud.py
jaryn: The router.exists is not refreshing the page once it already is on the destination, so it will time-out.

cfme/tests/networks/test_sdn_crud.py
jaryn: Because it is not refreshing the page.

cfme/networks/views.py
```diff
@@ -442,7 +443,11 @@ class NetworkRouterAddView(BaseLoggedInPage):
     cloud_tenant = Select(name='cloud_tenant_id')
     add = Button('Add')
 
-    is_displayed = displayed_not_implemented
+    @property
+    def is_displayed(self):
+        return (super(BaseLoggedInPage, self).is_displayed and
```
mshriver: Please use `self.logged_in_as_current_user` directly, since that's the only real check that we want from `BaseLoggedInPage` anyway.

cfme/tests/networks/test_sdn_crud.py
mshriver: condensing conversation to L67

cfme/tests/networks/test_sdn_crud.py
mshriver: Condensing conversation to L67

cfme/tests/networks/test_sdn_crud.py
mshriver: I think the correct way to solve that behavior is by implementing a resetter for the details nav destination for the router. That way navigation to the page will refresh by default, even if it short-cuts `step` because `am_i_here` was true.  This is ultimately a workaround, in the test, that would have to be duplicated throughout any tests trying to effectively check if a router exists.

jaryn: I implemented the resetter, but now the wait_for(router.exists...) fails this way:
```
================================================ FAILURES ================================================
_____________________________________ test_router_crud[openstack-11] _____________________________________

provider = OpenStackProvider(endpoints={'default': <cfme.infrastructure.provider.openstac...ne, keystone_v3_domain_id='default', tenant_mapping=False, infra_provider=None)
appliance = IPAppliance.from_json('{"container": null, "ui_port": 443, "ssh_port": 22, "op...dev": false, "version": "5.10.0.22", "db_host": null, "browser_steal": false}')

    @pytest.mark.ignore_stream('5.9')
    @pytest.mark.provider([OpenStackProvider], scope='function', override=True)
    def test_router_crud(provider, appliance):
        router_name = 'test_router_crud-{}'.format(fauxfactory.gen_alphanumeric().lower())
        routers = appliance.collections.network_routers
        router = routers.create(name=router_name,
                                provider=provider,
                                tenant=provider.data.networks.cloud_tenant,
                                network_manager=provider.data.networks.network_manager)
        assert router.exists
    
        router.delete()
    
>       wait_for(lambda: not router.exists, fail_condition=False, timeout=240, delay=1)

cfme/tests/networks/test_sdn_crud.py:62: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../work-venv/lib/python2.7/site-packages/wait_for/__init__.py:145: in wait_for
    out = func(*func_args, **func_kwargs)
cfme/tests/networks/test_sdn_crud.py:62: in <lambda>
    wait_for(lambda: not router.exists, fail_condition=False, timeout=240, delay=1)
cfme/networks/network_router.py:78: in exists
    navigate_to(self, 'Details')
../work-venv/lib/python2.7/site-packages/navmazing/__init__.py:113: in navigate
    return nav(cls_or_obj, self).go(0, *args, **kwargs)
cfme/utils/appliance/implementations/ui.py:563: in go
    message="Waiting for view [{}] to display".format(view.__class__.__name__)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

func = <function <lambda> at 0x7f663d7d3aa0>, func_args = [], func_kwargs = {}
logger = <cfme.utils.log.TraceLogger object at 0x7f665a47f950>
kwargs = {'message': 'Waiting for view [NetworkRouterDetailsView] to display', 'num_sec': 10}
st_time = 1541501381.144015, total_time = 8, num_sec = 10, expo = False
f_code = <code object <lambda> at 0x7f6649650c30, file "/home/jhenner/work/integration_tests/cfme/utils/appliance/implementations/ui.py", line 562>
line_no = 562
filename = '/home/jhenner/work/integration_tests/cfme/utils/appliance/implementations/ui.py'

    def wait_for(func, func_args=[], func_kwargs={}, logger=None, **kwargs):
        """Waits for a certain amount of time for an action to complete
        Designed to wait for a certain length of time,
        either linearly in 1 second steps, or exponentially, up to a maximum.
        Returns the output from the function once it completes successfully,
        along with the time taken to complete the command.
        Note: If using the expo keyword, the returned elapsed time will be inaccurate
            as wait_for does not know the exact time that the function returned
            correctly, only that it returned correctly at last check.
        Args:
            func: A function to be run
            func_args: A list of function arguments to be passed to func
            func_kwargs: A dict of function keyword arguments to be passed to func
            num_sec: An int describing the number of seconds to wait before timing out.
            timeout: Either an int describing the number of seconds to wait before timing out. Or a
                :py:class:`timedelta` object. Or a string formatted like ``1h 10m 5s``. This then sets
                the ``num_sec`` variable.
            expo: A boolean flag toggling exponential delay growth.
            message: A string containing a description of func's operation. If None,
                defaults to the function's name.
            fail_condition: An object describing the failure condition that should be tested
                against the output of func. If func() == fail_condition, wait_for continues
                to wait. Can be a callable which takes the result and returns boolean whether to fail.
                You can also specify it as a  set, that way it checks whether it is present in the
                iterable.
            handle_exception: A boolean controlling the handling of excepetions during func()
                invocation. If set to True, in cases where func() results in an exception,
                clobber the exception and treat it as a fail_condition.
            delay: An integer describing the number of seconds to delay before trying func()
                again.
            fail_func: A function to be run after every unsuccessful attempt to run func()
            quiet: Do not write time report to the log (default False)
            very_quiet: Do not log unless there was an error (default False). Implies quiet.
            silent_failure: Even if the entire attempt times out, don't throw a exception.
        Returns:
            A tuple containing the output from func() and a float detailing the total wait time.
        Raises:
            TimedOutError: If num_sec is exceeded after an unsuccessful func() invocation.
        """
        logger = logger or default_hidden_logger
        st_time = time.time()
        total_time = 0
    
        num_sec = _get_timeout_secs(kwargs)
        expo = kwargs.get('expo', False)
    
        f_code, line_no, filename, message = _get_context(func, kwargs.get('message', None))
    
        fail_condition = kwargs.get('fail_condition', False)
        fail_condition_check = _get_failcondition_check(fail_condition)
        handle_exception = kwargs.get('handle_exception', False)
        delay = kwargs.get('delay', 1)
        fail_func = kwargs.get('fail_func', None)
        very_quiet = kwargs.get("very_quiet", False)
        quiet = kwargs.get("quiet", False) or very_quiet
        silent_fail = kwargs.get("silent_failure", False)
    
        t_delta = 0
        tries = 0
        if not very_quiet:
            logger.debug('Started {} at {}'.format(message, st_time))
        while t_delta <= num_sec:
            try:
                tries += 1
                out = func(*func_args, **func_kwargs)
            except Exception as e:
                logger.info("wait_for hit an exception: {}: {}".format(type(e).__name__, e))
                if handle_exception:
                    out = fail_condition
                    logger.info("Call failed with following exception, but continuing "
                                "as handle_exception is set to True")
                else:
                    logger.info(
                        "Wait for {} took {} tries and {} seconds "
                        "before failure from an exception.".format(
                            message, tries, time.time() - st_time))
                    raise
            if out is fail_condition or fail_condition_check(out):
                time.sleep(delay)
                total_time += delay
                if expo:
                    delay *= 2
                if fail_func:
                    fail_func()
            else:
                duration = time.time() - st_time
                if not quiet:
                    logger.debug('Took {:0.2f} to do {}'.format(duration, message))
                if not very_quiet:
                    logger.debug(
                        'Finished {} at {}, {} tries'.format(message, st_time + t_delta, tries))
                return WaitForResult(out, duration)
            t_delta = time.time() - st_time
        if not very_quiet:
            logger.debug('Finished at {}'.format(st_time + t_delta))
        if not silent_fail:
            logger.error("Couldn't complete {} at {}:{} in time, took {:0.2f}, {} tries".format(message,
                filename, line_no, t_delta, tries))
            logger.error('The last result of the call was: {}'.format(out))
>           raise TimedOutError("Could not do {} at {}:{} in time".format(message, filename, line_no))
E           TimedOutError: Could not do Waiting for view [NetworkRouterDetailsView] to display at /home/jhenner/work/integration_tests/cfme/utils/appliance/implementations/ui.py:562 in time

../work-venv/lib/python2.7/site-packages/wait_for/__init__.py:180: TimedOutError
```

jaryn: @jawatts @mshriver do you have any suggestion on how to proceed?

dajoRH:
*CFME QE Bot*

PR #7842, opened 2018-09-18
[1LP][RFR] Adding ability to add VMware Host creds, as this is new requirement for v2v tests

kedark3:
Purpose or Intent
=================

__Extending__ cfme/fixtures/v2v.py  ability to add VMware Host creds, as this is new requirement for v2v tests, currently for 5.10, probably will end up in 5.9 builds as well. No harm in adding creds in current 5.9 tests as well. 

{{ pytest:  cfme/tests/v2v/test_v2v_migrations.py --use-provider rhv42 --use-provider vsphere6-nested --provider-limit 2 -vvvv --long-running -k 'test_dual_datastore_dual_vm_migration' }}

### DIFF
--- a/cfme/fixtures/v2v.py
+++ b/cfme/fixtures/v2v.py
@@ -1,10 +1,14 @@
 import fauxfactory
+import itertools
 import pytest
 from riggerlib import recursive_update
 from widgetastic.utils import partial_match
 
 from cfme.fixtures.provider import setup_or_skip
+from cfme.infrastructure.provider.rhevm import RHEVMProvider
+from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.utils.generators import random_vm_name
+from cfme.utils.log import logger
 
 
 @pytest.fixture(scope='function')
@@ -19,19 +23,41 @@ def v2v_providers(request, second_provider, provider):
 
 @pytest.fixture(scope='function')
 def host_creds(request, v2v_providers):
-    """Add credentials to conversation host"""
-    provider = v2v_providers[1]
-    hosts = provider.hosts.all()
+    """Add credentials to VMware and RHV hosts."""
+    if len(v2v_providers) > 2:
+        pytest.skip("There are more than two providers in v2v_providers fixture,"
+                    "which is invalid, skipping.")
     try:
-        hosts = hosts if getattr(request, 'param', '') == 'multi-host' else hosts[0:1]
-        for host in hosts:
-            host_data, = [data for data in provider.data['hosts'] if data['name'] == host.name]
+        vmware_provider, rhv_provider = None, None
+        for provider in v2v_providers:
+            if provider.one_of(VMwareProvider):
+                vmware_provider = provider
+            elif provider.one_of(RHEVMProvider):
+                rhv_provider = provider
+            else:
+                pytest.skip("Provider {} is not a valid provider for v2v tests".
+                    format(provider.name))
+        if not vmware_provider or not rhv_provider:
+            pytest.skip("Something went wrong with VMware or RHV provider, skipping test.")
+
+        vmware_hosts = vmware_provider.hosts.all()
+        for host in vmware_hosts:
+            host_data, = [data for data in vmware_provider.data['hosts']
+                          if data['name'] == host.name]
+            host.update_credentials_rest(credentials=host_data['credentials'])
+
+        rhv_hosts = rhv_provider.hosts.all()
+        rhv_hosts = rhv_hosts if getattr(request, 'param', '') == 'multi-host' else rhv_hosts[0:1]
+        for host in rhv_hosts:
+            host_data, = [data for data in rhv_provider.data['hosts'] if data['name'] == host.name]
             host.update_credentials_rest(credentials=host_data['credentials'])
     except Exception:
         # if above throws ValueError or TypeError or other exception, just skip the test
+        logger.exception("Exception when trying to add the host credentials.")
         pytest.skip("No data for hosts in providers, failed to retrieve hosts and add creds.")
-    yield hosts
-    for host in hosts:
+    # only yield RHV hosts as they will be required to tag with conversion_tags fixture
+    yield rhv_hosts
+    for host in itertools.chain(rhv_hosts, vmware_hosts):
         host.remove_credentials_rest()
 
 


dajoRH:
Lint report for commit a70cd086d003a6c757286a4f9b4f84ef7c5bfb72:

`cfme/fixtures/v2v.py`:
- :bangbang: Line 48:1: **E303** *too many blank lines (3)*

**No commit flaws detected**.

*CFME QE Bot*

kedark3:
@mshriver please review.

cfme/fixtures/v2v.py
```diff
@@ -20,15 +21,23 @@ def v2v_providers(request, second_provider, provider):
 @pytest.fixture(scope='function')
 def host_creds(request, v2v_providers):
     """Add credentials to conversation host"""
```
jawatts: I think the doc block here needs to be updated to better explain what is happening in this fixture.

cfme/fixtures/v2v.py
```diff
@@ -20,15 +21,23 @@ def v2v_providers(request, second_provider, provider):
...
-    provider = v2v_providers[1]
-    hosts = provider.hosts.all()
     try:
+        vmware_provider = v2v_providers[0]
```
jawatts: The assumption that the vmware provider will be the 0th element and the rhv provider will be the second is a bit dangerous. Doesn't this depend solely on the order in which they are passed with the --use-provider option?

cfme/fixtures/v2v.py
```diff
@@ -20,15 +21,23 @@ def v2v_providers(request, second_provider, provider):
...
+        hosts = vmware_provider.hosts.all()
+        for host in hosts:
+            host_data, = [data for data in vmware_provider.data['hosts']
+             if data['name'] == host.name]
```
jawatts: Indentation looks wrong here.

cfme/fixtures/v2v.py
```diff
@@ -20,15 +21,23 @@ def v2v_providers(request, second_provider, provider):
...
+            host.update_credentials_rest(credentials=host_data['credentials'])
+
+        rhv_provider = v2v_providers[1]
+        hosts = rhv_provider.hosts.all()
```
jawatts: It seems to me that you are overwriting any changes you made for the vmware host creds here?

cfme/fixtures/v2v.py
```diff
@@ -20,15 +21,23 @@ def v2v_providers(request, second_provider, provider):
...
         hosts = hosts if getattr(request, 'param', '') == 'multi-host' else hosts[0:1]
         for host in hosts:
-            host_data, = [data for data in provider.data['hosts'] if data['name'] == host.name]
+            host_data, = [data for data in rhv_provider.data['hosts'] if data['name'] == host.name]
             host.update_credentials_rest(credentials=host_data['credentials'])
     except Exception:
         # if above throws ValueError or TypeError or other exception, just skip the test
+        logger.exception("Exception when trying to add the host credentials.")
         pytest.skip("No data for hosts in providers, failed to retrieve hosts and add creds.")
     yield hosts
```
jawatts: **Question** Why are you only yielding the rhv hosts and not the vmware?

cfme/fixtures/v2v.py
```diff
@@ -20,15 +21,23 @@ def v2v_providers(request, second_provider, provider):
...
     for host in hosts:
```
jawatts: Should both the rhv and vmware hosts have the credentials removed?

cfme/fixtures/v2v.py
kedark3: Becase only RHV hosts are needed in further operation, which is host tagging.

cfme/fixtures/v2v.py
kedark3: Good point, we should remove creds for both.

cfme/fixtures/v2v.py
kedark3: I have not noticed any difference in provider order based on `--use-provider` because,  according to the code below, `second_provider` is always VMware and `v2v_providers` fixture returns the providers as 
`second_provider, provider` (VMware, RHV)

```
pytestmark = [
    pytest.mark.ignore_stream('5.8'),
    pytest.mark.provider(
        classes=[RHEVMProvider],
        selector=ONE_PER_VERSION,
        required_flags=['v2v']
    ),
    pytest.mark.provider(
        classes=[VMwareProvider],
        selector=ONE_PER_VERSION,
        fixture_name='second_provider',
        required_flags=['v2v']
    )
]
```

kedark3: Yep, that could be a trouble, I'll add checks there. :+1:

cfme/fixtures/v2v.py
```diff
@@ -19,19 +23,38 @@ def v2v_providers(request, second_provider, provider):
...
-        hosts = hosts if getattr(request, 'param', '') == 'multi-host' else hosts[0:1]
-        for host in hosts:
-            host_data, = [data for data in provider.data['hosts'] if data['name'] == host.name]
+        vmware_provider, rhv_provider = None, None
+        for provider in v2v_providers:
+            if provider.one_of(VMwareProvider):
+                vmware_provider = provider
+            elif provider.one_of(RHEVMProvider):
+                rhv_provider = provider
+        if not vmware_provider and not rhv_provider:
```
jawatts: What do you think about turning this into an `else` ? 

```
if provider.one_of(VMwareProvider):
    vmware_provider = provider
elif provider.one_of(RHEVMProvider):
￼     rhv_provider = provider
else:
    pytest.skip("Provider {} is not a valid provider for v2v testings.format(provider.name)")
```
Here is what I am thinking: You will have only 2 providers in the v2v_providers list (because of your check above). If one of them does not land in vmware or rhevm provider type, `vmware_provider` or `rhv_provider` is going to be `None`. If this happens on the first iteration of the for loop, then we do not need to bother continuing the second.

cfme/fixtures/v2v.py
```diff
@@ -19,19 +23,38 @@ def v2v_providers(request, second_provider, provider):
...
+            pytest.skip("Something Wrong with vmware_provider or rhv_provider, skipping.")
+
+        vmware_hosts = vmware_provider.hosts.all()
+        for host in vmware_hosts:
+            host_data, = [data for data in vmware_provider.data['hosts']
+                        if data['name'] == host.name]
```
jawatts: this indentation looks off, it should align with the 'd' in 'data'

cfme/fixtures/v2v.py
```diff
@@ -19,19 +23,38 @@ def v2v_providers(request, second_provider, provider):
...
             host.update_credentials_rest(credentials=host_data['credentials'])
     except Exception:
         # if above throws ValueError or TypeError or other exception, just skip the test
+        logger.exception("Exception when trying to add the host credentials.")
         pytest.skip("No data for hosts in providers, failed to retrieve hosts and add creds.")
-    yield hosts
-    for host in hosts:
+    # only yield RHV hosts as they will be required to tag with conversion_tags fixture
+    yield rhv_hosts
+    for host in itertools.chain(rhv_hosts, vmware_hosts):
```
jawatts: I had to look up the documentation for itertools.chain(), very cool.

cfme/fixtures/v2v.py
kedark3: Yes that may also work. I'll update the PR right away.

cfme/fixtures/v2v.py
kedark3: which data? there are many data in there :P LOL. and my Flake8Lint does not complain

jawatts: the first data :) pycharm complains it is "under-indented` when I checked out your branch

cfme/fixtures/v2v.py
```diff
@@ -19,19 +23,41 @@ def v2v_providers(request, second_provider, provider):
...
+    """Add credentials to VMware and RHV hosts."""
+    if len(v2v_providers) > 2:
+        pytest.skip("There are more than two providers in v2v_providers fixture,"
+                    "which is invalid, skipping.")
     try:
-        hosts = hosts if getattr(request, 'param', '') == 'multi-host' else hosts[0:1]
-        for host in hosts:
-            host_data, = [data for data in provider.data['hosts'] if data['name'] == host.name]
+        vmware_provider, rhv_provider = None, None
+        for provider in v2v_providers:
```
jawatts: @mshriver I was thinking about this last night, would this be better in the `v2v_providers` fixture? It can then return a keyed dict and in here we can pull assign based on the keys?

mshriver: Gotcha - discussed a bit with @kedark3, going to allow this for now.

Ultimately what I'd like to see for these tests/fixtures is use of explicit provider fixture names, and have them as parameters for any V2V fixtures that do setup on those providers. For example, `'rhv_provider'` and `'vmw_provider'` fixture names that are parameters for this fixture. Any test using this fixture should mark providers using those fixture names.

PR #7845, opened 2018-09-19
[RFR] Fix template ownership test

apagac:
This is a fix/redo of an existing ownership test.

The original test had two flaws:
1. If an exception occurred, we never got to the template deletion
2. If the test went fine, the template was deleted only from CFME database - not from provider.

This caused templates with the name "test-ownrs*" stack up on vmware providers and that is bad.
Also, the 'set_ownership' method now takes an user instance as an argument, not a string.

On 5.10 this will fail because: https://bugzilla.redhat.com/show_bug.cgi?id=1622952


apagac:
Added
@pytest.mark.meta(blockers=[BZ(1622952, forced_streams=['5.10'])])

PR #7849, opened 2018-09-19
[1LP][RFR]Generic service 5.10 fix

sshveta:
1) Marked tests with BZ.

2) Catalog name is appended with 'My company' in catalog bundles.

3) Candidatenotfound exception is raised when catalog node not found.


jawatts:
PRT for 510z has 3 failures due to failed logins. 

@sshveta is everything working locally?

sshveta:
> PRT for 510z has 3 failures due to failed logins.
> 
> @sshveta is everything working locally?

@jawatts Yes . Tests passing locally .

PR #7851, opened 2018-09-20
[1LP][RFR] 5.8 - add flash message for snapshot

apagac:
Quick add flash message for 5.8. Fixes failures in snapshot testing for 5.8.


apagac:
So two things:
1. This is a fix for 5.8, made against the '5.8-breakpoint-patched' branch. I don't see a reason for concerning myself with 5.10 on this particular branch
2. I can't very well use LOWEST, because there are differences between 5.9 and 5.8 strings.

'5.9': "Revert to a Snapshot initiated for 1 VM and Instance from the CFME Database"
'5.8': "Revert To Snapshot initiated for 1 VM and Instance from the CFME Database"

PR #7852, opened 2018-09-20
[RFR] fixed check for ansible logs

lina-nikiforova:
{{ pytest: -v cfme/tests/configure/test_log_depot_operation.py::test_collect_multiple_servers }}

1. fixed issue with provisioning temporary appliances in sprout (AuthException: Authentication failed! Method request_appliances needs authentication!)

2. fixed check for ansible logs introduced in #7805 (I don't want to enable ansible and check its logs in every test, e.g. collecting logs from multiple servers)

### DIFF
--- a/cfme/test_framework/sprout/client.py
+++ b/cfme/test_framework/sprout/client.py
@@ -85,9 +85,7 @@ class SproutClient(object):
     def from_config(cls, **kwargs):
         host = env.get("sprout", {}).get("hostname", "localhost")
         port = env.get("sprout", {}).get("port", 8000)
-        user_key = kwargs.pop('sprout_user_key', None)
-        if user_key is None:
-            user_key = 'sprout'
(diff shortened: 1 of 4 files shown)


cfme/test_framework/sprout/client.py
```diff
@@ -85,7 +85,9 @@ def __getattr__(self, attr):
     def from_config(cls, **kwargs):
         host = env.get("sprout", {}).get("hostname", "localhost")
         port = env.get("sprout", {}).get("port", 8000)
-        user_key = kwargs.pop('sprout_user_key') if 'sprout_user_key' in kwargs else None
+        user_key = kwargs.pop('sprout_user_key', None)
+        if user_key is None:
+            user_key = 'sprout'
```
mshriver: Will ask that we do _not_ do this, as it restores the default sprout user behavior that I removed just a while ago.  We want to force the user to explicitly pass a sprout user cred key, or have their env vars setup with their sprout creds

PR #7853, opened 2018-09-21
[1LP][RFR] Convert `cfme/tests/networks` to package

sasoc:
Folder `cfme/tests/networks` is now a package. We changed this because we will need to import some of its modules.

For example we have a very useful fixture called [with_nuage_sandbox](https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/networks/nuage/test_events_are_triggered.py#L17-L61), which creates enterprise and its entities. It also returns object with their values, which we can then compare to ui and database values. This fixture will be needed for many tests and we want to be able to import it.


sasoc:
cc @miha-plesko

sasoc:
@miha-plesko description has been updated, I hope its ok now.

mshriver:
@miha-plesko @sasoc No fixtures should have to be imported into test modules for use.

Fixtures should be registered here:
https://github.com/ManageIQ/integration_tests/blob/master/cfme/test_framework/pytest_plugin.py#L37

They will then be available in the global fixture list, and can be added as parameters to any test module's fixtures or tests.

PR #7861, opened 2018-09-24
[1LP][RFR] Add test for migration with rbac

Yadnyawalkya:
{{pytest: cfme/tests/v2v/test_v2v_migrations_ui.py -k test_migration_rbac --use-provider rhv42 --use-provider vsphere67-nested --provider-limit 2 -vvvv}}

### DIFF
--- a/cfme/tests/v2v/test_v2v_migrations_ui.py
+++ b/cfme/tests/v2v/test_v2v_migrations_ui.py
@@ -257,8 +257,8 @@ def test_migration_rbac(appliance, new_credential, conversion_tags, host_creds, 
         view = navigate_to(appliance.server, 'Dashboard', wait_for_view=True)
         nav_tree = view.navigation.nav_item_tree()
         # Checks migration option is disabled in navigation
-        assert 'Migration' not in nav_tree['Compute'], ('Trying to find Migration in nav_tree, '
-                                                        'should be absent :{}'.format(nav_tree))
+        assert 'Migration' not in nav_tree['Compute'], ('Migration found in nav tree, '
+                                                        'rbac should not allow this')
 
     product_features = [(['Everything'], False), (['Everything'], True)]
     role.update({'product_features': product_features})
@@ -266,5 +266,5 @@ def test_migration_rbac(appliance, new_credential, conversion_tags, host_creds, 
         view = navigate_to(appliance.server, 'Dashboard', wait_for_view=True)
         nav_tree = view.navigation.nav_item_tree()
         # Checks migration option is enabled in navigation
-        assert 'Migration' in nav_tree['Compute'], ('Trying to find Migration in nav_tree, '
-                                                    'should be absent :{}'.format(nav_tree))
+        assert 'Migration' in nav_tree['Compute'], ('Migration not found in nav tree, '
+                                                    'rbac should allow this')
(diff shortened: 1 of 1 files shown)


Yadnyawalkya:
I imported tests since I had to reuse many existing functions, not sure it is good, @digitronik need thoughts. 5.9 PRT passed!

cfme/tests/v2v/test_v2v_migrations_ui.py
```diff
@@ -241,3 +242,29 @@ def test_v2v_ui_set2(appliance, v2v_providers, form_data_single_datastore, soft_
...
+    group = new_group(appliance=appliance, role=role.name)
+    user = new_user(appliance=appliance, group=group, credential=new_credential)
+
+    product_features = [(['Everything', 'Compute', 'Migration'], False)]
+    role.update({'product_features': product_features})
+    with user:
+        view = navigate_to(appliance.server, 'Dashboard', wait_for_view=True)
+        nav_tree = view.navigation.nav_item_tree()
+        # Checks migration option is disabled in navigation
+        assert 'Migration' not in nav_tree['Compute'], ('Trying to find Migration in nav_tree, '
```
digitronik: Question  and one Suggestion:
So here `Migration` will `disable` or `hidden`? I am if disable then it `key` will still available. 

`nav_tree["Compute"]` will return you `dict`  so better to go with `nav_tree["Compute"].has_key("Migration")`

cfme/tests/v2v/test_v2v_migrations_ui.py
```diff
@@ -241,3 +242,29 @@ def test_v2v_ui_set2(appliance, v2v_providers, form_data_single_datastore, soft_
...
+                                                        'should be absent :{}'.format(nav_tree))
+
+    product_features = [(['Everything'], False), (['Everything'], True)]
+    role.update({'product_features': product_features})
+    with user:
+        view = navigate_to(appliance.server, 'Dashboard', wait_for_view=True)
+        nav_tree = view.navigation.nav_item_tree()
+        # Checks migration option is enabled in navigation
+        assert 'Migration' in nav_tree['Compute'], ('Trying to find Migration in nav_tree, '
```
digitronik: Same as above

cfme/tests/v2v/test_v2v_migrations_ui.py
Yadnyawalkya: btw @digitronik, when I run flake8 ignoring `E128`, it says `W601 .has_key() is deprecated, use 'in'`, so I guess If I push this changes lint-bot will catch me!

cfme/tests/v2v/test_v2v_migrations_ui.py
mshriver: This is still the case, the message on assert failure does not make sense here.

In both cases (above assertion, and the one on L260) I think the message can be simplified.
```
assert 'Migration' not in nav_tree['Compute'], ('Migration found in nav tree, rbac should not allow')
# ...
assert 'Migration' in nav_tree['Compute'], ('Migration not found in nav tree, rbac should allow')

Yadnyawalkya: updated messages @mshriver  :+1:

PR #7863, opened 2018-09-24
[WIP] Validate Nuage subnet inventory

sasoc:
With this commit we create NuageSubnet and NuageSubnetCollection, which we then use to validate details page of Nuage Subnet.

We use `with_nuage_sandbox` to create Nuage enterprise and its entities, including subnet. Sandbox then gives us values to compare with later.

NuageSubnetCollection has `find_by_ems_ref` function that finds requested NuageSubnet in database by ems_ref and returns it.

ValidateStatsMixin has `validate_stats` that uses ui methods from NuageSubnet to get details page values. These are then compared to those acquired from sandbox.

We also use different steps to navigate to details page. We go through provider using steps like this:
```Nertwork -> Providers -> Provider (by name) -> Cloud Subnets -> Subnet (by name)```
By using this steps we make sure that we are on the details page of Nuage Cloud Subnet.

Depends on: https://github.com/ManageIQ/integration_tests/pull/7858

### DIFF
--- a/cfme/networks/provider/nuage.py
+++ b/cfme/networks/provider/nuage.py
@@ -7,6 +7,7 @@ from wrapanapi.systems import NuageSystem
 from cfme.common.provider import DefaultEndpoint, DefaultEndpointForm, EventsEndpoint
 from cfme.common.provider_views import BeforeFillMixin
 from cfme.networks.security_group import SecurityGroupCollection
+from cfme.networks.subnet import Subnet, SubnetCollection
 from cfme.utils import version
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.varmeth import variable
@@ -42,6 +43,30 @@ class NuageEndpointForm(View):
         password = Input('amqp_password')
 
         validate = Button('Validate')
+
+
+class ValidateStatsMixin(object):
+    def validate_stats(self, expected_stats):
+        """ Validates that the details page matches the subnet's inventory.
+
+        Args:
+            expected_stats: dictionary of values to be compered to UI values,
+            where keys have the same names as NuageSubnet ui functions
+
+        This method is given expected stats as an argument and those are then matched
+        against the UI. An AssertionError exception will be raised if the stats retrieved
+        from the UI do not match those from expected stats.
+
+        IMPORTANT: Please make sure NuageSubnet implements counterpart ui method for each of
+        the keys in expected_stats dictionary.
+        """
+
+        # Make sure we are on the subnet details page and refresh it
+        navigate_to(self, "DetailsThroughProvider")
+        self.browser.selenium.refresh()
+
+        cfme_stats = {stat: getattr(self, stat)(method='ui') for stat in expected_stats}
+        assert cfme_stats == expected_stats
 
 
 @attr.s(hash=False)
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by sasoc** [comment]
cc @miha-plesko

File: cfme/networks/provider/nuage.py [hunk 1]
```diff
@@ -155,3 +156,99 @@ def num_floating_ip(self):
...
+        subnets_table = self.appliance.db.client['cloud_subnets']
+        subnet = None
+        try:
+            subnet_query = (
+                self.appliance.db.client.session
+                    .query(subnets_table.name, subnets_table.ems_ref)
+                    .filter(subnets_table.ems_ref == ems_ref)).all()[0]
+            subnet = self.instantiate(name=subnet_query.name, ems_ref=subnet_query.ems_ref)
+        finally:
+            return subnet
```
miha-plesko: Nah, there are multiple things that bother me: first, we suppress all the errors even potential syntax errors with try-finally. Second, we call `.all()[0]` on query instead `.first()`. Third, we the docstring is not capitalized. Can I suggest something like:

```python
def find_by_ems_ref(self, ems_ref):
        """Fetch NuageSubnet instance with provided ems_ref"""
        subnets_table = self.appliance.db.client['cloud_subnets']
        subnet = self.appliance.db.client.session.query(subnets_table.name, subnets_table.ems_ref).filter(subnets_table.ems_ref == ems_ref).first()

        return self.instantiate(name=subnet.name, ems_ref=subnet.ems_ref) if subnet else None
```

Btw, the ORM that is used here is sqlalchemy, which comes with as powerful ORM as the one shipped with Django. Here's the doc on [.first()](http://docs.sqlalchemy.org/en/rel_0_9/orm/query.html#sqlalchemy.orm.query.Query.first) method. It returns `None` if no records are found so no need for try-catch anymore.

cfme/tests/networks/nuage/test_nuage_inventory.py
```diff
@@ -0,0 +1,45 @@
...
+        'network_protocol_value': sandbox['subnet'].ip_type.lower(),
+        'network_manager_value': provider.name,
+        'cloud_tenant_value': sandbox['enterprise'].name,
+        'network_router_value': sandbox['domain'].name,
+        'network_ports_num': len(sandbox['subnet'].vports),
+        'security_groups_num': sum(len(port.policy_groups) for port in sandbox['subnet'].vports),
+    }
+    provider.refresh_provider_relationships()
+    subnet, _ = wait_for(
+        lambda: appliance.collections.nuage_network_subnets.find_by_ems_ref(sandbox['subnet'].id)
```
miha-plesko: Yeah so as discussed offline I'd prefer if each iteration of the VMBD probing would log a line like

```
Looking for subnet with ems_ref %s in the VMBD...
```

That's because when user runs the test, she can have no idea what the test is currently doing unless we log it.

sasoc:
@miha-plesko thank you for the review, changes have been pushed.

cfme/networks/provider/nuage.py
```diff
@@ -44,6 +46,27 @@ class events(Tab, BeforeFillMixin):
...
 
+class ValidateStatsMixin(object):
+    def validate_stats(self, host_stats, ui=False):
+        """ Validates that the detail page matches the subnet's inventory.
+
+        Args:
+            host_stats: dictionary of values to be compered to UI values,
+            where keys have the same names as NuageSubnet ui functions
+
+        This method gets host stats and matches them against the UI. An exception will
```
miha-plesko: _matches them against the UI_ - that part is not always true they way it's implemented now. In fact, it's not even true in default case because we set default to
```
ui=False
```
right?

Since we currently only support UI approach and we don't have desire to be checking DB really (trying to simulate what user sees as much as possible) I'd suggest to just drop `ui=False` parameter of our `validate_stats` function and always go for UI method. We can easily support both methods later should need arise.

cfme/networks/provider/nuage.py
```diff
@@ -44,6 +46,27 @@ class events(Tab, BeforeFillMixin):
         validate = Button('Validate')
 
 
+class ValidateStatsMixin(object):
+    def validate_stats(self, host_stats, ui=False):
+        """ Validates that the detail page matches the subnet's inventory.
```
miha-plesko: It's "deatil**s** page", not "detail page". Please fix everywhere

cfme/networks/provider/nuage.py
miha-plesko: Also: _This method gets host stats and matches them against the UI_ sounds much like the method itself somehow retrieves host stats. Please reword, we want it to be clear that we're passing dictionary of expected values. No need to even name them as "host_stats" but maybe rather "expected_stats".

Sorry for nitpicking, but it pisses me off when I look at other providers how confusing they are. They say "get_stats" and mean PhysicalServer status like "on" or "off". Then they say "validate_stats" and they mean server details. So confusing 😄 

Also, please add a short clarification like _Please make sure NuageSubnet implements counterpart method for each of the keys in `expected_stats` dictionary._

cfme/networks/provider/nuage.py
```diff
@@ -44,6 +46,27 @@ class events(Tab, BeforeFillMixin):
...
+
+        Args:
+            host_stats: dictionary of values to be compered to UI values,
+            where keys have the same names as NuageSubnet ui functions
+
+        This method gets host stats and matches them against the UI. An exception will
+        be raised if the stats retrieved from the UI do not match those from host stats.
```
miha-plesko: _An exception will be raised_ - please be specific. If I wanted to catch the exception I'd need to know excpetion class.

cfme/tests/networks/nuage/test_nuage_inventory.py
miha-plesko: I see you put logging into `find_by_ems_ref` method directly, but we actually only want to log those strings here because we're looping and waiting. The method itself shouldn't log IMAO - if we call it from elsewhere, logging wouldn't be desired.

dajoRH:
Lint report for commit b0d49799351d547cb621b7f17b04cc65aec73de7:

`cfme/networks/provider/nuage.py`:
- :red_circle: Line 13:1: **F401** *'cfme.utils.log.logger' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/networks/nuage/test_nuage_inventory.py
```diff
@@ -0,0 +1,47 @@
...
+        'gateway_value': sandbox['subnet'].gateway,
+        'network_protocol_value': sandbox['subnet'].ip_type.lower(),
+        'network_manager_value': provider.name,
+        'cloud_tenant_value': sandbox['enterprise'].name,
+        'network_router_value': sandbox['domain'].name,
+        'network_ports_num': len(sandbox['subnet'].vports),
+        'security_groups_num': sum(len(port.policy_groups) for port in sandbox['subnet'].vports),
+    }
+    provider.refresh_provider_relationships()
+    logger.info('Looking for subnet with ems_ref {} in the VMBD...'.format(sandbox['subnet'].id))
```
miha-plesko: @sasoc does this really log upon every iteration? I think it'll only log once and then user will see nothing in log anymore. So I was thinking more like to include the logging in the lambda below. If lambda gets too ugly, then just go with regular function.

sasoc:
@miha-plesko I've changed it so it logs in every iteration. Please take a look if its ok now.

sasoc:
@miha-plesko Changes that you requested have been pushed.

sasoc:
@miha-plesko I've changed some thing as we discussed.
1. I've implemented `object_in_vmdb_with_timeout` so now other entities can use it also.
2. I've changed steps when navigating through ui, so now it goes like this:
``` Nertwork -> Providers -> Provider (by name) -> Cloud Subnets -> Subnet (by name)```

Can you please take a look at this changes.

mshriver:

PR #7864, opened 2018-09-24
[1LP][RFR][NOTEST] Need to separate out Managers  from provider

digitronik:
Purpose or Intent
=================
- We  are getting error like this
```
    def get_crud_by_name(provider_name):
        """ Creates a Provider object given a management_system name in cfme_data.
    
        Usage:
            get_crud_by_name('My RHEV 3.6 Provider')
    
        Returns: A Provider object that has methods that operate on CFME
        """
        for provider_key, provider_data in providers_data.items():
            if provider_data.get("name") == provider_name:
                return get_crud(provider_key)
>       raise NameError("Could not find provider {}".format(provider_name))
E       NameError: Could not find provider Embedded Ansible Automation Manager
```
- we have seperate out all manager those are `child provider` not `main provider`

Note: Test not adding as will not reproduce with fresh appliance... need a proper run or cloud provider with infra having cluster :D

### DIFF
--- a/cfme/infrastructure/cluster.py
+++ b/cfme/infrastructure/cluster.py
@@ -334,7 +334,7 @@ class ClusterCollection(BaseCollection):
             providers_db = {
                 prov.id: get_crud_by_name(prov.name)
                 for prov in providers
-                if "Manager" not in prov.name
+                if not getattr(prov, "parent_ems_id", False)
             }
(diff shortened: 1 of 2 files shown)


digitronik:
I like @mkoura  suggestion of attributes offline. trying those.

PR #7865, opened 2018-09-24
[1LP][RFR] Add summary view selection to cloud details nav resetter

mshriver:
Infra details view had this default, cloud should as well.

resetter will select summary view by default on provider pages. use_resetter=false, or explicit dashboard view selection where needed.

### DIFF
--- a/cfme/infrastructure/provider/__init__.py
+++ b/cfme/infrastructure/provider/__init__.py
@@ -331,9 +331,9 @@ class Details(CFMENavigateStep):
         self.prerequisite_view.entities.get_entity(name=self.obj.name, surf_pages=True).click()
 
     def resetter(self):
-        """Reset view and selection"""
+        """Reset view to Summary View if available"""
         view_selector = self.view.toolbar.view_selector
-        if view_selector.selected != 'Summary View':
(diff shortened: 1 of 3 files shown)


cfme/infrastructure/host.py
```diff
@@ -644,6 +644,15 @@ def step(self):
             self.prerequisite_view.navigation.select("Compute", "Infrastructure", "Nodes")
 
 
+@navigator.register(HostsCollection)
+class AllForProvider(CFMENavigateStep):
+    VIEW = HostsView
+    prerequisite = NavigateToSibling('All')
+
+    def step(self, *args, **kwargs):
+        pass
```
mshriver: Currently working on this

cfme/infrastructure/provider/__init__.py
```diff
@@ -333,7 +333,7 @@ def step(self):
     def resetter(self):
         """Reset view and selection"""
```
jawatts: Should this be updated to describe which view it is resetting to?

mshriver: sure

PR #7866, opened 2018-09-24
[1LP][RFR]V2v tests mac addr

kedark3:
Purpose or Intent
=================

__Extending__ v2v_tests to check MAC address of the VMs pre and post migration.

{{pytest: cfme/tests/v2v/test_v2v_migrations.py --use-provider rhv42 --use-provider vsphere65-nested --provider-limit 2 -vvvv --long-running  -k 'test_single_datastore_single_vm_migration' }}

### DIFF
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -181,6 +181,7 @@ class BaseVM(BaseEntity, Pretty, Updateable, PolicyProfileAssignable, Taggable, 
     @property
     def mac_address(self):
         """Fetches MAC Address of VM"""
+        # TODO: We should update this with wrapanapi method when it becomes available.
         view = navigate_to(self, "Details", use_resetter=False)
         return view.entities.summary('Properties').get_text_of('MAC Address')
 
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -179,6 +179,12 @@ class BaseVM(BaseEntity, Pretty, Updateable, PolicyProfileAssignable, Taggable, 
         return self.mgmt.ip
 
     @property
+    def mac_address(self):
+        """Fetches MAC Address of VM"""
+        view = navigate_to(self, "Details", use_resetter=False)
+        return view.entities.summary('Properties').get_text_of('MAC Address')
+
+    @property
     def is_retired(self):
         """Check retirement status of vm"""
         view = navigate_to(self, "Details", use_resetter=False)
--- a/cfme/fixtures/v2v.py
+++ b/cfme/fixtures/v2v.py
@@ -81,8 +81,8 @@ def _tag_cleanup(host_obj, tag1, tag2):
     # are not subset of valid tags, we still remove them.
     if len(tags_set) < 2 or not tags_set.issubset(valid_tags):
         host_obj.remove_tags(tags=tags)
-        return False
-    return True
+        return True
+    return False
 
 
 @pytest.fixture(scope='function')
--- a/cfme/fixtures/v2v.py
+++ b/cfme/fixtures/v2v.py
@@ -81,8 +81,8 @@ def _tag_cleanup(host_obj, tag1, tag2):
     # are not subset of valid tags, we still remove them.
     if len(tags_set) < 2 or not tags_set.issubset(valid_tags):
         host_obj.remove_tags(tags=tags)
-        return True
-    return False
+        return False
+    return True
 
 
 @pytest.fixture(scope='function')
--- a/cfme/fixtures/v2v.py
+++ b/cfme/fixtures/v2v.py
@@ -1,6 +1,7 @@
 import fauxfactory
 import itertools
 import pytest
+from collections import namedtuple
 from riggerlib import recursive_update
 from widgetastic.utils import partial_match
 
@@ -11,14 +12,32 @@ from cfme.utils.generators import random_vm_name
 from cfme.utils.log import logger
 
 
+FormDataVmObj = namedtuple(
+    'FormDataVmObj', ['form_data', 'vm_list']
+)
+
+
 @pytest.fixture(scope='function')
 def v2v_providers(request, second_provider, provider):
     """ Fixture to setup providers """
-    setup_or_skip(request, second_provider)
-    setup_or_skip(request, provider)
-    yield second_provider, provider
-    second_provider.delete_if_exists(cancel=False)
-    provider.delete_if_exists(cancel=False)
+    V2vProviders = namedtuple(
+        'V2vProviders', ['vmware_provider', 'rhv_provider']
+    )
+    vmware_provider, rhv_provider = None, None
+    for provider in [second_provider, provider]:
+        if provider.one_of(VMwareProvider):
+            vmware_provider = provider
+        elif provider.one_of(RHEVMProvider):
+            rhv_provider = provider
+        else:
+            pytest.skip("Provider {} is not a valid provider for v2v tests".
+                format(provider.name))
+    v2v_providers = V2vProviders(vmware_provider=vmware_provider, rhv_provider=rhv_provider)
+    setup_or_skip(request, v2v_providers.vmware_provider)
+    setup_or_skip(request, v2v_providers.rhv_provider)
+    yield v2v_providers
+    v2v_providers.vmware_provider.delete_if_exists(cancel=False)
+    v2v_providers.rhv_provider.delete_if_exists(cancel=False)
 
 
 @pytest.fixture(scope='function')
@@ -28,17 +47,8 @@ def host_creds(request, v2v_providers):
         pytest.skip("There are more than two providers in v2v_providers fixture,"
                     "which is invalid, skipping.")
     try:
-        vmware_provider, rhv_provider = None, None
-        for provider in v2v_providers:
-            if provider.one_of(VMwareProvider):
-                vmware_provider = provider
-            elif provider.one_of(RHEVMProvider):
-                rhv_provider = provider
-            else:
-                pytest.skip("Provider {} is not a valid provider for v2v tests".
-                    format(provider.name))
-        if not vmware_provider or not rhv_provider:
-            pytest.skip("Something went wrong with VMware or RHV provider, skipping test.")
+        vmware_provider = v2v_providers.vmware_provider
+        rhv_provider = v2v_providers.rhv_provider
 
         vmware_hosts = vmware_provider.hosts.all()
         for host in vmware_hosts:
@@ -168,10 +178,10 @@ def form_data_multiple_vm_obj_single_datastore(request, appliance, second_provid
             }
         }
     })
-    vm_obj = []
+    vm_list = []
     for template_name in request.param[2]:
-        vm_obj.append(get_vm(request, appliance, second_provider, template_name))
-    return form_data, vm_obj
+        vm_list.append(get_vm(request, appliance, second_provider, template_name))
+    return FormDataVmObj(form_data=form_data, vm_list=vm_list)
 
 
 @pytest.fixture(scope='function')
@@ -243,7 +253,7 @@ def form_data_dual_vm_obj_dual_datastore(request, appliance, second_provider, pr
     # creating 2 VMs on two different datastores and returning its object list
     vm_obj1 = get_vm(request, appliance, second_provider, request.param[0][2], request.param[0][0])
     vm_obj2 = get_vm(request, appliance, second_provider, request.param[1][2], request.param[1][0])
-    return form_data, [vm_obj1, vm_obj2]
+    return FormDataVmObj(form_data=form_data, vm_list=[vm_obj1, vm_obj2])
 
 
 @pytest.fixture(scope='function')
@@ -270,7 +280,7 @@ def form_data_vm_obj_dual_nics(request, appliance, second_provider, provider):
         }
     })
     vm_obj = get_vm(request, appliance, second_provider, request.param[2])
-    return form_data, [vm_obj]
+    return FormDataVmObj(form_data=form_data, vm_list=[vm_obj])
 
 
 @pytest.fixture(scope='function')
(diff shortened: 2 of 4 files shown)


kedark3:
In run  26850, 4 out of 5 tests passed. 1 Failed due to migration failed, not due to any of the code in this PR.

cfme/common/vm.py
```diff
@@ -178,6 +178,12 @@ def ip_address(self):
         """Fetches IP Address of VM"""
         return self.mgmt.ip
 
+    @property
+    def mac_address(self):
+        """Fetches MAC Address of VM"""
+        view = navigate_to(self, "Details", use_resetter=False)
+        return view.entities.summary('Properties').get_text_of('MAC Address')
```
jawatts: **Question** Why could we not use wrapanapi to pull the mac address? Or do you want to get it from CFME exclusively?

cfme/tests/v2v/test_v2v_migrations.py
```diff
@@ -43,6 +43,14 @@ def test_single_datastore_single_vm_migration(request, appliance, v2v_providers,
     def _cleanup():
         infrastructure_mapping_collection.delete(mapping)
 
+    src_vm_obj = form_data_vm_obj_single_datastore[1][0]
```
jawatts: I would like to get this test as a whole away from using the indexing.

cfme/tests/v2v/test_v2v_migrations.py
```diff
@@ -110,6 +125,12 @@ def _cleanup():
      delay=5, num_sec=2700)
     assert (request_details_list.is_successful(vms[0]) and
         not request_details_list.is_errored(vms[0]))
+    # validate MAC address matches between source and target VMs
+    src_vm = form_data_vm_obj_single_network[1][0]
+    rhv_prov = v2v_providers[1]
+    collection = rhv_prov.appliance.provider_based_collection(rhv_prov)
+    migrated_vm = collection.instantiate(src_vm.name, rhv_prov)
+    assert src_vm.mac_address == migrated_vm.mac_address
```
jawatts: Seems like this code is duplicated in a couple places, thoughts on breaking it out into a single method?

cfme/tests/v2v/test_v2v_migrations.py
```diff
@@ -154,6 +175,13 @@ def _cleanup():
         soft_assert(request_details_list.is_successful(vm) and
          not request_details_list.is_errored(vm))
 
+    rhv_prov = v2v_providers[1]
+    collection = rhv_prov.appliance.provider_based_collection(rhv_prov)
+    # validate MAC address matches between source and target VMs
+    for src_vm in form_data_dual_vm_obj_dual_datastore[1]:
+        migrated_vm = collection.instantiate(src_vm.name, rhv_prov)
+        soft_assert(src_vm.mac_address == migrated_vm.mac_address)
```
jawatts: Same comment as above. Also see soft_assert being used here, is this because you want to go through all items?

kedark3: yes, we want to go through all the items. Don't want to fail if one failed. hence soft_assert.

cfme/common/vm.py
kedark3: Will resolve this as TODO as per the discussion we had offline.

cfme/tests/v2v/test_v2v_migrations.py
kedark3: Fixed it, check it out @jawatts :)

kedark3:
@jawatts @mshriver kindly review, need to merge this as I need to rebase all of my other PRs after this is merged and then only can proceed merging other PRs.

kedark3:
In /pr/run/26958 -- Because of re-factoing of fixtures, I ran tests against entire v2v module, 2 failures are due to known bz and 1 failure is due to migration could not complete before timeout. Nothing to do with code, re-factoring went successful. Skips are due to timeout during VM  provisioning.

cfme/fixtures/v2v.py
```diff
@@ -14,9 +15,12 @@
 @pytest.fixture(scope='function')
 def v2v_providers(request, second_provider, provider):
     """ Fixture to setup providers """
+    v2v_providers = namedtuple(
+        'v2v_providers', ['vmware_provider', 'rhv_provider']
+    )
     setup_or_skip(request, second_provider)
     setup_or_skip(request, provider)
-    yield second_provider, provider
+    yield v2v_providers(vmware_provider=second_provider, rhv_provider=provider)
```
jawatts: **Required** I think we are still making the assumption that the rhv provider will always be provider and vmware will always be the second provider, I think we should move away from this assumption and specifically check.

cfme/fixtures/v2v.py
```diff
@@ -152,6 +152,9 @@ def _form_data(second_provider, provider):
 @pytest.fixture(scope='function')
 def form_data_multiple_vm_obj_single_datastore(request, appliance, second_provider, provider):
     # this fixture will take list of N VM templates via request and call get_vm for each
+    form_data_multiple_vm_obj_single_datastore = namedtuple(
```
jawatts: **Required** This tuple should be defined at the module level and instantiated for each method.

cfme/fixtures/v2v.py
kedark3: will check that here.

cfme/fixtures/v2v.py
kedark3: I thought of that, but wanted to let it come from reviewers as I didn't know it will be accepted.

cfme/fixtures/v2v.py
```diff
@@ -11,12 +12,31 @@
 from cfme.utils.log import logger
 
 
+form_data_vm_obj = namedtuple(
```
jawatts: Same camel case comment here

kedark3: Done

cfme/fixtures/v2v.py
```diff
@@ -11,12 +12,31 @@
...
+            rhv_provider = provider
+        else:
+            pytest.skip("Provider {} is not a valid provider for v2v tests".
+                format(provider.name))
+    if not vmware_provider or not rhv_provider:
+        pytest.skip("Something went wrong with VMware or RHV provider, skipping test.")
     setup_or_skip(request, second_provider)
     setup_or_skip(request, provider)
-    yield second_provider, provider
+    yield v2v_providers(vmware_provider=vmware_provider, rhv_provider=rhv_provider)
```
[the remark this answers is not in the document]
kedark3: Done

cfme/fixtures/v2v.py
```diff
@@ -11,12 +12,31 @@
...
+            vmware_provider = provider
+        elif provider.one_of(RHEVMProvider):
+            rhv_provider = provider
+        else:
+            pytest.skip("Provider {} is not a valid provider for v2v tests".
+                format(provider.name))
+    if not vmware_provider or not rhv_provider:
+        pytest.skip("Something went wrong with VMware or RHV provider, skipping test.")
     setup_or_skip(request, second_provider)
     setup_or_skip(request, provider)
```
jawatts: We should be using the values of the namedtuple here instead of `provider` and `second_provider`

cfme/fixtures/v2v.py
```diff
@@ -11,12 +12,31 @@
...
-    yield second_provider, provider
+    yield V2vProviders(vmware_provider=vmware_provider, rhv_provider=rhv_provider)
```
jawatts: the namedtuple should be instantiated before the yield

cfme/fixtures/v2v.py
```diff
@@ -28,17 +48,13 @@ def host_creds(request, v2v_providers):
...
-            if provider.one_of(VMwareProvider):
-                vmware_provider = provider
-            elif provider.one_of(RHEVMProvider):
-                rhv_provider = provider
-            else:
-                pytest.skip("Provider {} is not a valid provider for v2v tests".
-                    format(provider.name))
-        if not vmware_provider or not rhv_provider:
-            pytest.skip("Something went wrong with VMware or RHV provider, skipping test.")
+        if (v2v_providers.vmware_provider.one_of(VMwareProvider)and
```
jawatts: I think this is redudunt now that it is checked in the `v2v_providers` fixture

cfme/tests/v2v/test_v2v_migrations.py
```diff
@@ -36,19 +43,25 @@ def test_single_datastore_single_vm_migration(request, appliance, v2v_providers,
...
     infrastructure_mapping_collection = appliance.collections.v2v_mappings
-    # form_data_vm_obj_single_datastore has form_data at [0]
-    mapping = infrastructure_mapping_collection.create(form_data_vm_obj_single_datastore[0])
+    mapping = infrastructure_mapping_collection.create(form_data_vm_obj_single_datastore.form_data)
 
     @request.addfinalizer
     def _cleanup():
         infrastructure_mapping_collection.delete(mapping)
 
+    src_vm_obj = form_data_vm_obj_single_datastore.vm_obj[0]
```
jawatts: what is the 0th index here? Can you put a comment that vm_obj is a list and you are getting the 0th element because.

cfme/tests/v2v/test_v2v_migrations.py
```diff
@@ -293,7 +326,7 @@ def test_conversion_host_tags(appliance, v2v_providers):
             display_name='V2V - Transformation Method')
             .collections.tags.instantiate(display_name='VDDK'))
 
-    host = v2v_providers[1].hosts.all()[0]
+    host = v2v_providers.rhv_provider.hosts.all()[0]
```
jawatts: what is the 0th index here? Same comment as about, lets put this clarification

kedark3: added comment

cfme/tests/v2v/test_v2v_migrations.py
kedark3: Added

kedark3:
Couple of tests failed as the VMs did not get the same IP address pre and post migration.

cfme/tests/v2v/test_v2v_migrations.py
```diff
@@ -67,7 +80,12 @@ def _cleanup():
         migration_plan.name, view.migration_plans_completed_list.get_vm_count_in_plan(
             migration_plan.name), view.migration_plans_completed_list.get_clock(
             migration_plan.name))
+    # validate MAC address matches between source and target VMs
     assert view.migration_plans_completed_list.is_plan_succeeded(migration_plan.name)
+    src_vm = form_data_vm_obj_single_datastore.vm_obj[0]
+    migrated_vm = get_migrated_vm_obj(src_vm, v2v_providers.rhv_provider)
+    assert src_vm.mac_address == migrated_vm.mac_address
+    assert src_vm_ip_addr == migrated_vm.ip_address
```
mshriver: I'm not sure this is necessarily something that we can validate against, as its highly dependent on DHCP server and client configuration. The target provider may be served by a different DHCP server even, in a different subnet, than the source machine. I think it should only validate that the MAC addresses match, as that's all the migration guarantees.

cfme/tests/v2v/test_v2v_migrations.py
```diff
@@ -36,19 +43,25 @@ def test_single_datastore_single_vm_migration(request, appliance, v2v_providers,
                                             conversion_tags,
                                             form_data_vm_obj_single_datastore):
     infrastructure_mapping_collection = appliance.collections.v2v_mappings
-    # form_data_vm_obj_single_datastore has form_data at [0]
-    mapping = infrastructure_mapping_collection.create(form_data_vm_obj_single_datastore[0])
+    mapping = infrastructure_mapping_collection.create(form_data_vm_obj_single_datastore.form_data)
```
mshriver: :tada:

cfme/tests/v2v/test_v2v_migrations.py
```diff
@@ -36,19 +43,25 @@ def test_single_datastore_single_vm_migration(request, appliance, v2v_providers,
...
 
     @request.addfinalizer
     def _cleanup():
         infrastructure_mapping_collection.delete(mapping)
+    # vm_obj is a list, with only 1 VM object, hence [0]
+    src_vm_obj = form_data_vm_obj_single_datastore.vm_obj[0]
```
mshriver: Consider calling `.pop()` instead of indexing

cfme/tests/v2v/test_v2v_migrations.py
```diff
@@ -67,7 +80,12 @@ def _cleanup():
         migration_plan.name, view.migration_plans_completed_list.get_vm_count_in_plan(
             migration_plan.name), view.migration_plans_completed_list.get_clock(
             migration_plan.name))
+    # validate MAC address matches between source and target VMs
     assert view.migration_plans_completed_list.is_plan_succeeded(migration_plan.name)
+    src_vm = form_data_vm_obj_single_datastore.vm_obj[0]
```
mshriver: You already have this named as `src_vm_obj`, why re-name it and index the fixture list again?

cfme/fixtures/v2v.py
```diff
@@ -11,14 +12,32 @@
 from cfme.utils.log import logger
 
 
+FormDataVmObj = namedtuple(
+    'FormDataVmObj', ['form_data', 'vm_obj']
```
mshriver: Consider calling the second attribute `vm_list` since its always a list.  That'll make it more clear/readable having `v2v_fixture_name.vm_list.pop()` instead of `v2v_fixture_name.vm_obj[0]`

cfme/tests/v2v/test_v2v_migrations.py
```diff
@@ -278,6 +305,12 @@ def _cleanup():
         soft_assert(request_details_list.is_successful(vm) and
          not request_details_list.is_errored(vm))
 
+    src_vms_list = form_data_multiple_vm_obj_single_datastore[1]
```
mshriver: still indexing the fixture tuple here

cfme/tests/v2v/test_v2v_migrations.py
kedark3: Should we maybe make it soft_assert?

kedark3: @jawatts it is a test case to see 1) if the MAC and 2) IP Addresses remain same, but IP address is best attempt, not a guarantee.

PR #7882, opened 2018-09-26
[WIP] Update release util, py3 compat

mshriver:
Using Popen was making stdout a bytestring, which you can't split with a regular string
Since we're only looking at stdout here, just use getoutput() instead of Popen()
This gives us a string, which we can then split with a string

## Note to Reviewers

PRT won't run this, but you can locally very easily with `miq release`.

Example args: 
`miq release --brief 17.36.0`
`miq release --full --line-limit 200 --old-tag 17.35.0 17.36.0`

### DIFF
--- a/cfme/utils/release.py
+++ b/cfme/utils/release.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python2
 from __future__ import print_function
-
 from collections import defaultdict
+from functools import reduce
 import datetime
 import re
 import requests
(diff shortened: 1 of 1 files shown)


cfme/utils/release.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env python2
 from __future__ import print_function
-
 from collections import defaultdict
+from functools import reduce
```
digitronik: :+1:

digitronik:
@mshriver I think it will break for `py2`. Please have a look one time.

PR #7883, opened 2018-09-27
[RFR] Use log validator for targeted refresh

sasoc:
Targeted refesh fixture searches `evm.log` to check if refresh  was started for specific entity. Until this commit merkyl plugin was used to tail `evm.log`, now LogValidator is used to tail it. This change was made because LogValidator is prefered over merkyl as ssh access is all you need for it. While merkyl needs daemon to run on appliance and its failure can affect targeted refresh test.

Upon migrating to the new approach, we had to enhance LogValidator a bit to support our use case. It's that until now it was always used to search logs exactly once (and fail test if not found), while we needed to periodically retry the thing and only fail test if not fulfilled by the end of timeout.

Depends on https://github.com/ManageIQ/integration_tests/pull/7858

### DIFF
--- a/cfme/fixtures/networks.py
+++ b/cfme/fixtures/networks.py
@@ -1,15 +1,15 @@
 import pytest
-import re
 from argparse import Namespace
 from contextlib import contextmanager
 
 from cfme.exceptions import NeedleNotFoundInLog
 from cfme.utils.log import logger
+from cfme.utils.log_validator import LogValidator
 from cfme.utils.wait import wait_for, TimedOutError
 
 
 @pytest.fixture
-def targeted_refresh(merkyl_setup, merkyl_inspector):
+def targeted_refresh():
     """
     This fixture tests whether targeted refresh was triggered for given targets.
     It basically tails evm.log to see whether a line like this has occured:
@@ -28,19 +28,16 @@ def targeted_refresh(merkyl_setup, merkyl_inspector):
                     targeted_refresh.register_target('ref-123456', 'Subnet named TEST')
                     targeted_refresh.register_target('ref-aaaabb', 'Router named TEST')
     """
-    evm_log = '/var/www/miq/vmdb/log/evm.log'
     needle_template = r'^.*Collection of targets with id.*:ems_ref=>"{}".*$'
-    merkyl_inspector.add_log(evm_log)
-    merkyl_inspector.reset_log(evm_log)
-    targets = set()  # { (ems_ref, comment), (ems_ref, comment), ... }
+    targets = {}  # { needle: comment }
+    log_validator = LogValidator('/var/www/miq/vmdb/log/evm.log')
+    log_validator.fix_before_start()
 
     def check_log():
-        logger.info('Looking for %s needles in evm.log: %s', len(targets), [t[1] for t in targets])
-        content = merkyl_inspector.get_log(evm_log)
-        for target in set(targets):
-            if target[0].search(content):
-                logger.info('Found needle %s', target[1])
-                targets.remove(target)
+        logger.info('Looking for %s needles in evm.log: %s', len(targets), list(targets.values()))
+        log_validator.update_matched_patterns(list(targets))
+        for revealed_pattern in log_validator.validate_logs(auto_fail=False):
+            logger.info('Found needle %s', targets.pop(revealed_pattern))
         return len(targets) == 0
 
     @contextmanager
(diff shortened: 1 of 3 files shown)


sasoc:
cc @miha-plesko

cfme/utils/log_validator.py
```diff
@@ -46,16 +46,23 @@ def __init__(self, remote_filename, **kwargs):
...
         for line in self._remote_file_tail:
             if self._check_skip_logs(line):
                 continue
             self._check_fail_logs(line)
             self._check_match_logs(line)
-        self._verify_match_logs()
+        if _auto_fail:
+            self._verify_match_logs()
+        else:
+            return list(self.matches.keys())
```
miha-plesko: We can always return, no need to put it in else. Also, can go with `return list(self.matches)` here as well.

cfme/fixtures/networks.py
```diff
@@ -29,18 +29,20 @@ def test_something(targeted_refresh)
                     targeted_refresh.register_target('ref-aaaabb', 'Router named TEST')
     """
     evm_log = '/var/www/miq/vmdb/log/evm.log'
```
miha-plesko: I think we no longer need this in separate variable as we only use it once. Please consider passing it directly.

cfme/fixtures/networks.py
```diff
@@ -29,18 +29,20 @@ def test_something(targeted_refresh)
...
-    needle_template = r'^.*Collection of targets with id.*:ems_ref=>"{}".*$'
```
miha-plesko: Is there no match if you leave those? I think we should leave them, unless there is a reason why not.

cfme/fixtures/networks.py
```diff
@@ -29,18 +29,20 @@ def test_something(targeted_refresh)
...
-    merkyl_inspector.add_log(evm_log)
-    merkyl_inspector.reset_log(evm_log)
-    targets = set()  # { (ems_ref, comment), (ems_ref, comment), ... }
+    needle_template = r'.*Collection of targets with id.*:ems_ref=>"{}".*'
+    targets = {}
```
miha-plesko: So for a set we had a comment to document its structure. Please consider doing something similar for dict.

cfme/fixtures/networks.py
```diff
@@ -29,18 +29,20 @@ def test_something(targeted_refresh)
...
-        content = merkyl_inspector.get_log(evm_log)
-        for target in set(targets):
-            if target[0].search(content):
-                logger.info('Found needle %s', target[1])
-                targets.remove(target)
+        logger.info('Looking for %s needles in evm.log: %s' % (
+            len(targets), list(targets.values())))
+        log_validator.update_matched_patterns(list(targets.keys()))
+        found_targets = log_validator.validate_logs(_auto_fail=False)
+        if len(found_targets) > 0:
```
miha-plesko: No need for this check. If an empty list will be returned, then block won't be executed anyway.

cfme/fixtures/networks.py
```diff
@@ -29,18 +29,20 @@ def test_something(targeted_refresh)
...
+            for target in found_targets:
+                logger.info('Found needle %s' % targets[target])
+                targets.pop(target, None)
```
miha-plesko: Why would you default to anything if we never even capture the result? So what `.pop(k)` does is it removes key from dictionary and returns it's counterpart value. Knowing this, we could compress L41-45 into a two-liner:

```python
for revealed_pattern in log_validator.validate_logs(auto_fail=False):
  logger.info('Found needle %s', targets.pop(revealed_pattern))
```

cfme/fixtures/networks.py
```diff
@@ -29,18 +29,20 @@ def test_something(targeted_refresh)
...
-            if target[0].search(content):
-                logger.info('Found needle %s', target[1])
-                targets.remove(target)
+        logger.info('Looking for %s needles in evm.log: %s' % (
+            len(targets), list(targets.values())))
+        log_validator.update_matched_patterns(list(targets.keys()))
+        found_targets = log_validator.validate_logs(_auto_fail=False)
+        if len(found_targets) > 0:
+            for target in found_targets:
+                logger.info('Found needle %s' % targets[target])
```
miha-plesko: We're not allowed to use `%` operator for string interpolation anymore.

cfme/utils/log_validator.py
```diff
@@ -46,16 +46,23 @@ def __init__(self, remote_filename, **kwargs):
         self._remote_file_tail = SSHTail(remote_filename, **kwargs)
         self.matches = {}
 
+    def update_matched_patterns(self, new_patterns):
```
miha-plesko: A docstring would be nice to have here. We need to elaborate our usecase and reason why we implemented this. So please add a comment telling this function is used when periodically probing the log for a list of patterns where we want to shrink the list when targets are found.

sasoc:
@miha-plesko Thank you for the review. I've made changes that you requested.

digitronik:
@sasoc https://github.com/ManageIQ/integration_tests/pull/7858 duplication of work

sasoc:
@digitronik The PR this one was depending on was merged and I've rebased this one to master.

mshriver:

PR #7886, opened 2018-09-27
[1LP][RFR] Add leading whitespace password test

apagac:
This is a new test. Adds simple test for checking if an user is able to be created & login with password containing leading space character.

I will refactor the whole file later, now just adding tests to increase coverage.

{{ pytest: -v cfme/tests/test_login.py -k test_leading_whitespace_password }}

### DIFF
--- a/cfme/tests/test_login.py
+++ b/cfme/tests/test_login.py
@@ -94,3 +94,27 @@ def test_update_password(context, request, appliance):
         appliance.server.login(user)
 
     user.delete()
+
+
+@pytest.mark.parametrize('context', [ViaUI])
+def test_leading_whitespace_password(context, request, appliance):
+    """ Test password with leading whitespace """
+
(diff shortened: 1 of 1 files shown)


cfme/tests/test_login.py
```diff
@@ -89,3 +89,27 @@ def test_update_password(context, request, appliance):
...
+        groups=user_group
+    )
+
+    # Try to login with new user
+    logged_in_page = appliance.server.login(user)
+    assert logged_in_page.is_displayed
+    logged_in_page.logout()
+
+    # Delete the user
+    user.delete()
```
izapolsk: REQUIRED: if assert or something else  happen before this line. User won't be removed. Please handle this by either creating appropriate yield fixture or adding request.addfinalizer.

apagac:
Will be implemented as a part of another test.

PR #7890, opened 2018-09-27
[1LP][WIPTEST] new test_provider_log_level

lina-nikiforova:
Shriver: expanding pytest coverage to include all modified cases
{{ pytest: -v cfme/tests/configure/test_logs.py --use-provider complete }}

### DIFF
--- a/cfme/tests/configure/test_logs.py
+++ b/cfme/tests/configure/test_logs.py
@@ -50,16 +50,16 @@ def test_provider_log_rotate(appliance, provider, log_exists):
     Metadata:
         test_flag: log
     """
-    if log_exists:
-        appliance.ssh_client.run_command("logrotate -f /etc/logrotate.d/miq_logs.conf")
-        logs_count = appliance.ssh_client.run_command(
-            "ls -l /var/www/miq/vmdb/log/{}.log*|wc -l".format(
-                provider.log_name
-            )
-        ).output
-        assert logs_count > 1, "{}.log wasn't rotated by default miq_logs.conf".format(
+    assert log_exists, "Log file {}.log doesn't exist".format(provider.log_name)
+    appliance.ssh_client.run_command("logrotate -f /etc/logrotate.d/miq_logs.conf")
+    logs_count = appliance.ssh_client.run_command(
+        "ls -l /var/www/miq/vmdb/log/{}.log*|wc -l".format(
             provider.log_name
         )
+    ).output
+    assert logs_count > 1, "{}.log wasn't rotated by default miq_logs.conf".format(
+        provider.log_name
+    )
 
 
 def test_provider_log_updated(appliance, provider, log_exists):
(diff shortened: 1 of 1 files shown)


cfme/tests/configure/test_logs.py
```diff
@@ -88,3 +89,34 @@ def test_provider_log_updated(appliance, provider, log_exists):
...
+        # set log level to debug
+        appliance.server.update_advanced_settings(
+            {'log': {'level_{}'.format(provider.log_name): 'debug'}})
+        wait_for(provider.is_refreshed, func_kwargs=dict(refresh_delta=10), timeout=600)
+        debug_in_logs = appliance.ssh_client.run_command(
+            "cat /var/www/miq/vmdb/log/{}.log | grep DEBUG".format(provider.log_name)).output
+        # set log level back
+        appliance.server.update_advanced_settings(
+            {'log': {'level_{}'.format(provider.log_name): log_level}})
+        assert 'DEBUG' in debug_in_logs
```
digitronik: Required:
@lina-nikiforova I would like to go like this 
```
debug_in_logs = appliance.ssh_client.run_command(
            "cat /var/www/miq/vmdb/log/{}.log | grep DEBUG".format(provider.log_name))
assert debug_in_logs.success
```

lina-nikiforova: 1. @digitronik log_exists can't be `None`, it's easier `True` or `False`
https://docs.python.org/2/library/functions.html#bool
But in case of `False` the test is indeed passed. Thanks.
I'm adding assert for it in this and other tests but keep plain "test_log_exists" as well

2. Thanks, I'm changing it

cfme/tests/configure/test_logs.py
```diff
@@ -49,16 +50,16 @@ def test_provider_log_rotate(appliance, provider, log_exists):
...
-        ).output
-        assert logs_count > 1, "{}.log wasn't rotated by default miq_logs.conf".format(
+    assert log_exists, "Log file {}.log doesn't exist".format(provider.log_name)
+    appliance.ssh_client.run_command("logrotate -f /etc/logrotate.d/miq_logs.conf")
+    logs_count = appliance.ssh_client.run_command(
+        "ls -l /var/www/miq/vmdb/log/{}.log*|wc -l".format(
             provider.log_name
         )
+    ).output
+    assert logs_count > 1, "{}.log wasn't rotated by default miq_logs.conf".format(
```
mshriver: I believe you'll need to explicitly cast this to an `int()`, as the output from SSH call will be a string. Trying to compare string and int with `>` results in TypeError

cfme/tests/configure/test_logs.py
```diff
@@ -74,17 +75,48 @@ def test_provider_log_updated(appliance, provider, log_exists):
...
+    # set log level to debug
+    appliance.server.update_advanced_settings(
+        {'log': {'level_{}'.format(provider.log_name): 'debug'}})
+    wait_for(provider.is_refreshed, func_kwargs=dict(refresh_delta=10), timeout=600)
+    debug_in_logs = appliance.ssh_client.run_command(
+        "cat /var/www/miq/vmdb/log/{}.log | grep DEBUG".format(provider.log_name))
+    # set log level back
+    appliance.server.update_advanced_settings(
+        {'log': {'level_{}'.format(provider.log_name): log_level}})
+    assert debug_in_logs.success
```
mshriver: Seeing this failing in PRT against two providers in one version and several in the other - please comment on what's going on with this check? Does BZ 1633656 cover more providers than thought?

lina-nikiforova: @mshriver I updated the blocker as there's one more BZ, but for the remaining providers I have all the tests passed locally, however they're failing in PRT, can you please take a look why this can happen?

PR #7893, opened 2018-09-27
[WIP] Adding initial tests for physical overview page

matheuscmelo:
This PR adds initial tests for the Physical Overview page.


dajoRH:
Lint report for commit 6a4ad8e37800bfc92d5806fdafb8118b4814c231:

`cfme/physical/overview.py`:
- :red_circle: Line 6:1: **F401** *'cfme.utils.wait.wait_for' imported but unused*
- :bangbang: Line 23:101: **E501** *line too long (103 > 100 characters)*

`cfme/tests/physical_infrastructure/ui/test_physical_overview.py`:
- :heavy_exclamation_mark: Line 7:35: **W291** *trailing whitespace*
- :bangbang: Line 15:1: **E302** *expected 2 blank lines, found 1*
- :bangbang: Line 20:1: **E302** *expected 2 blank lines, found 1*

Hmmm


**No commit flaws detected**.

*CFME QE Bot*

cfme/physical/overview.py
```diff
@@ -0,0 +1,32 @@
+from navmazing import NavigateToAttribute
+from widgetastic_manageiq import StatusBox
```
digitronik: Required:
Please sort imports as per libs
1. python core
2. third party
3. local

cfme/physical/overview.py
```diff
@@ -0,0 +1,32 @@
...
+from cfme.base.login import BaseLoggedInPage
+from cfme.utils.appliance.implementations.ui import CFMENavigateStep, navigator
+from cfme.utils.appliance import Navigatable
+
+
+class PhysicalOverview(Navigatable):
```
digitronik: Required:
Need collection. we are not currently using this. 
Please refer this current design model for `Integration test`
https://cfme-tests.readthedocs.io/en/master/guides/collections.html

cfme/physical/overview.py
```diff
@@ -0,0 +1,32 @@
...
+    storages = StatusBox('Storages')
+    switches = StatusBox('Switches')
+
+    @property
+    def is_displayed(self):
+        return self.navigation.currently_selected == ["Compute",
+            "Physical Infrastructure", "Overview"]
+
+
+@navigator.register(PhysicalOverview)
```
digitronik: Required:
Short name for navigation destination like `All`

cfme/tests/physical_infrastructure/ui/test_physical_overview.py
```diff
@@ -0,0 +1,25 @@
...
+    pytest.mark.provider([LenovoProvider])]
+
+
+def test_physical_overview_page():
+    view = navigate_to(PhysicalOverview, 'All')
+    assert view.is_displayed
+
+
+def test_physical_overview_servers_number(appliance, provider):
+    view = navigate_to(PhysicalOverview, 'All')
```
digitronik: Required:
Please register it in `entty_pioints` with a collection and then use this

cfme/tests/physical_infrastructure/ui/test_physical_overview.py
```diff
@@ -0,0 +1,25 @@
...
+    servers = appliance.collections.physical_servers.all(provider)
```
digitronik: :+1:

cfme/physical/overview.py
matheuscmelo: Hi @digitronik, does it really need a collection as it's a view, not an entity?

digitronik: my bad

cfme/physical/overview.py
matheuscmelo: - no core imports
- navmazing and widgetstatic
- local imports
aren't they correct?

digitronik: `widgetastic_manageiq` is part of the integration test. Ya, we are planning to move it but for now teat it as a local import.

cfme/physical/overview.py
digitronik: As per offline conversation. Its simple overview page no need of `BaseEntity` like container `overview` page.

mshriver: This is an unnecessary object that the navigation destination has been registered against, and should not exist.

The navigation destination that brings the browser to the `PhysicalOverviewView` should be registered against the physical provider collection.  That's the best contextual collection for this page in the UI, as its an overview of all providers, and their collected stats.

I think there's no need for a new module here, and this could just be in `cfme/physical/provider/__init__.py - but if you'd like to separate this into its own module, I'm not opposed to that structure.  Just import `PhysicalProviderCollection` and use it for navigation class registration.

cfme/physical/overview.py
```diff
@@ -0,0 +1,33 @@
...
+from cfme.utils.appliance.implementations.ui import CFMENavigateStep, navigator
+from cfme.utils.appliance import Navigatable
+
+
+class PhysicalOverview(Navigatable):
+    pass
+
+
+class PhysicalOverviewView(BaseLoggedInPage):
+    providers = StatusBox('Providers')
```
mshriver: There is a ParametrizedStatusBox available now, if you'd like to use it instead of creating separate widgets - your call.

cfme/tests/physical_infrastructure/ui/test_physical_overview.py
mshriver: Along with the comment above, this navigation destination should be called 'Overview', and should be registered against the provider collection. Here, you can get `appliance.collections.physical_providers` to navigate with.

cfme/tests/physical_infrastructure/ui/test_physical_overview.py
```diff
@@ -0,0 +1,25 @@
...
```
mshriver: I would switch the order of your statements here, since its possible that an `all()` method call could use the UI and navigate.

Though it may currently use DB, or REST to get the physical server list, the method may change, or be implemented with Sentaku - making this test unstable.

Think about the comparison made here as well - the `all()` method that you called will be getting the number of servers from CFME's database or rest interface. You're then comparing the UI number against that - which ends up being a weak comparison. A much stronger comparison, and test, would be to query the number of servers from the provider's REST interface directly via a wrapanapi call to `provider.mgmt.list_servers()`

cfme/tests/physical_infrastructure/ui/test_physical_overview.py
```diff
@@ -0,0 +1,25 @@
...
+    assert view.servers.value == len(servers)
+
+
+def test_physical_overview_switches_number(appliance):
+    view = navigate_to(PhysicalOverview, 'All')
```
mshriver: Same here as above - fetch the all() count first and then navigate and use the view from navigation second.

Or better yet - query wrapanapi first, and view second :)

matheuscmelo: Hey @mshriver, thanks for your review. As far as I know wrapanapi fetches information from only 1 provider, and Overview page informs counts servers/switches/racks... from all physical providers connected to MIQ. What would be a better way to handle this issue?

matheuscmelo: Also, fetch switches from wrapanapi depends on [wrapanapi/294](https://github.com/ManageIQ/wrapanapi/pull/294)

dajoRH:
Lint report for commit 500e94f815ad43319f7fe5382e28e0ec7710a24b:

`cfme/physical/provider/__init__.py`:
- :bangbang: Line 166:1: **E302** *expected 2 blank lines, found 1*

**No commit flaws detected**.

*CFME QE Bot*

PR #7894, opened 2018-09-27
[1LP][RFR] Add Physical Storage entity

matheuscmelo:
This PR adds the Physical Storage entity to the framework.

### DIFF
--- a/cfme/physical/physical_storage.py
+++ b/cfme/physical/physical_storage.py
@@ -1,0 +1,207 @@
+# -*- coding: utf-8 -*-
+"""A model of an Infrastructure PhysicalStorage in CFME."""
+import attr
+from navmazing import NavigateToAttribute
+from cached_property import cached_property
+from lxml.html import document_fromstring
+from widgetastic_patternfly import Dropdown, Accordion, BreadCrumb
+from widgetastic.widget import Text, View
+from widgetastic_manageiq import (
+    BaseEntitiesView,
+    JSBaseEntity,
+    ItemsToolBarViewSelector,
+    SummaryTable,
+    ManageIQTree
+)
+from cfme.common import PolicyProfileAssignable, Taggable
+from cfme.exceptions import ItemNotFound
+from cfme.modeling.base import BaseEntity, BaseCollection
+from cfme.utils.appliance.implementations.ui import CFMENavigateStep, navigate_to, navigator
+from cfme.utils.pretty import Pretty
+from cfme.utils.providers import get_crud_by_name
+from cfme.utils.update import Updateable
+from cfme.utils.wait import wait_for
+from cfme.base.login import BaseLoggedInPage
+
+
+class ComputePhysicalInfrastructureStoragesView(BaseLoggedInPage):
+    """Common parts for Storage views."""
+    title = Text('.//div[@id="center_div" or @id="main-content"]//h1')
+
+    @property
+    def in_compute_physical_infrastructure_storages(self):
+        return (self.logged_in_as_current_user and
+                self.navigation.currently_selected == ["Compute", "Physical Infrastructure",
+                                                       "Storages"])
+
+
+class PhysicalStorageEntity(JSBaseEntity):
+    @property
+    def data(self):
+        data_dict = super(PhysicalStorageEntity, self).data
+        if 'quadicon' in data_dict and data_dict['quadicon']:
+            quad_data = document_fromstring(data_dict['quadicon'])
+            data_dict['no_port'] = int(quad_data.xpath(self.QUADRANT.format(pos="a"))[0].text)
+            data_dict['state'] = quad_data.xpath(self.QUADRANT.format(pos="b"))[0].get('style')
+            data_dict['vendor'] = quad_data.xpath(self.QUADRANT.format(pos="c"))[0].get('alt')
+            data_dict['health_state'] = quad_data.xpath(self.QUADRANT.format(pos="d"))[0].get('alt')
+        return data_dict
+
+
+class PhysicalStorageDetailsToolbar(View):
+    """Represents physical toolbar and its controls."""
+    configuration = Dropdown(text="Configuration")
+
+
+class PhysicalStorageDetailsEntities(View):
+    """Represents Details page."""
+    properties = SummaryTable(title="Properties")
+    management_networks = SummaryTable(title="Management Networks")
+    relationships = SummaryTable(title="Relationships")
+    power_management = SummaryTable(title="Power Management")
+    firmwares = SummaryTable(title="Firmwares")
+    ports = SummaryTable(title="Ports")
+
+
+class PhysicalStorageDetailsView(ComputePhysicalInfrastructureStoragesView):
+    """Main PhysicalStorage details page."""
+    breadcrumb = BreadCrumb()
+    toolbar = View.nested(PhysicalStorageDetailsToolbar)
+    entities = View.nested(PhysicalStorageDetailsEntities)
+
+    @property
+    def is_displayed(self):
+        title = "{name} (Summary)".format(name=self.context["object"].name)
+        return (self.in_compute_physical_infrastructure_storages and
+                self.breadcrumb.active_location == title)
+
+
+class PhysicalStoragesToolbar(View):
+    """Represents hosts toolbar and its controls."""
+    configuration = Dropdown(text="Configuration")
+    view_selector = View.nested(ItemsToolBarViewSelector)
+
+
+class PhysicalStorageSideBar(View):
+    """Represents left side bar. It usually contains navigation, filters, etc."""
+
+    @View.nested
+    class filters(Accordion): # noqa
+        tree = ManageIQTree()
+
+
+class PhysicalStorageEntitiesView(BaseEntitiesView):
+    """Represents the physical storage view"""
+    @property
+    def entity_class(self):
+        return PhysicalStorageEntity
+
+
+class PhysicalStoragesView(ComputePhysicalInfrastructureStoragesView):
+    toolbar = View.nested(PhysicalStoragesToolbar)
+    sidebar = View.nested(PhysicalStorageSideBar)
+    including_entities = View.include(PhysicalStorageEntitiesView, use_parent=True)
+
+    @property
+    def is_displayed(self):
+        return (self.in_compute_physical_infrastructure_storages and
+                self.title.text == "Physical Storages")
+
+
+@attr.s
+class PhysicalStorage(BaseEntity, Updateable, Pretty, PolicyProfileAssignable, Taggable):
+    """Model of an Physical Storage in cfme.
+
+    Args:
+        name: Name of the physical storage.
+
+    Usage:
+        mystorage = PhysicalStorage(name='sample_Storage')
+        mystorage.create()
+
+    """
+    pretty_attrs = ['name']
+
+    name = attr.ib()
+    provider = attr.ib()
+
+    INVENTORY_TO_MATCH = ['Product Name', 'Serial Number', 'Description',
+                          'Health State', 'Drive Bays', 'Enclosure Count']
+
+    def get_value_from_summary(self, value):
+        """Returns the value of a summary table
+
+        Usage:
+            mystorage = PhysicalStorage(name='sample_Storage')
+            mystorage.get_value_from_summary(value)
+
+        Args:
+            value (str): The Storage property you want to retrieve from Details page.
+        """
+        view = navigate_to(self, "Details")
+        return view.entities.properties.get_text_of(value)
+
+    def exists(self):
+        """Checks if the physical_Storage exists in the UI.
+
+        Returns: :py:class:`bool`
+        """
+        view = navigate_to(self.parent, "All")
+        try:
+            view.entities.get_entity(name=self.name, surf_pages=True)
+            return True
+        except ItemNotFound:
+            return False
+
+    @cached_property
+    def db_id(self):
+        return self.appliance.physical_storage_id(self.name)
+
+
+@attr.s
+class PhysicalStorageCollection(BaseCollection):
+    """Collection object for the :py:class:`cfme.infrastructure.host.PhysicalStorage`."""
+
+    ENTITY = PhysicalStorage
+
+    def select_entity_rows(self, *physical_storages):
+        """ Select all physical storage objects """
+        view = navigate_to(self, 'All')
+
+        for physical_storage in physical_storages:
+            view.entities.get_entity(name=physical_storage.name, surf_pages=True).check()
+
+    def all(self):
+        """returning all physical_storages objects"""
+        physical_storage_table = self.appliance.db.client['physical_storages']
+        ems_table = self.appliance.db.client['ext_management_systems']
+        physical_storage_query = (
+            self.appliance.db.client.session
+                .query(physical_storage_table.name, ems_table.name)
+                .join(ems_table, physical_storage_table.id == ems_table.id))
+        provider = self.filters.get('provider')
+        if provider:
+            physical_storage_query = physical_storage_query.filter(ems_table.name == provider.name)
+        physical_storages = []
+        for name, ems_name in physical_storage_query.all():
+            physical_storages.append(self.instantiate(name=name,
+                                    provider=provider or get_crud_by_name(ems_name)))
+        return physical_storages
+
+
+@navigator.register(PhysicalStorageCollection, 'All')
+class All(CFMENavigateStep):
+    VIEW = PhysicalStoragesView
+    prerequisite = NavigateToAttribute("appliance.server", "LoggedIn")
+
+    def step(self):
+        self.prerequisite_view.navigation.select("Compute", "Physical Infrastructure", "Storages")
+
+
+@navigator.register(PhysicalStorage, 'Detail')
+class Details(CFMENavigateStep):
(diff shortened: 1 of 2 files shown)


cfme/common/physical_storage_views.py
```diff
@@ -0,0 +1,135 @@
+# -*- coding: utf-8 -*-
```
matheuscmelo: Not sure if this should be here or inside `physical` folder

dajoRH:
Lint report for commit 6110ab69930a017d9dd4a8a864744f4d04e6ff32:

`cfme/physical/physical_storage.py`:
- :red_circle: Line 4:1: **F401** *'pdb' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

cfme/common/physical_storage_views.py
```diff
@@ -0,0 +1,125 @@
...
+from lxml.html import document_fromstring
+from widgetastic_manageiq import (
+    BaseEntitiesView,
+    JSBaseEntity,
+    BreadCrumb,
```
digitronik: Required:
Please import this from `wt.pfy`

cfme/common/physical_storage_views.py
```diff
@@ -0,0 +1,125 @@
...
+    management_networks = SummaryTable(title="Management Networks")
+    relationships = SummaryTable(title="Relationships")
+    power_management = SummaryTable(title="Power Management")
+    firmwares = SummaryTable(title="Firmwares")
+    ports = SummaryTable(title="Ports")
+
+
+class PhysicalStorageDetailsView(ComputePhysicalInfrastructureStoragesView):
+    """Main PhysicalStorage details page."""
+    breadcrumb = BreadCrumb(locator='.//ol[@class="breadcrumb"]')
```
digitronik: Required:
Already this locator set for you in `wt.pfy` as `ROOT` please refer this
https://github.com/RedHatQE/widgetastic.patternfly/blob/master/src/widgetastic_patternfly/__init__.py#L1821

No need to pass locator

cfme/common/physical_storage_views.py
```diff
@@ -0,0 +1,125 @@
...
+    @property
+    def is_displayed(self):
+        title = "{name} (Summary)".format(name=self.context["object"].name)
+        return (self.in_compute_physical_infrastructure_storages and
+                self.breadcrumb.active_location == title)
+
+
+class PhysicalStorageProvisionView(BaseLoggedInPage):
+    """Represents the Provision Physical Storage page."""
+    breadcrumb = BreadCrumb(locator='.//ol[@class="breadcrumb"]')
```
digitronik: Required:
As above

cfme/common/physical_storage_views.py
```diff
@@ -0,0 +1,125 @@
...
+        return self.breadcrumb.active_location == title
+
+
+class PhysicalStoragesToolbar(View):
+    """Represents hosts toolbar and its controls."""
+    configuration = Dropdown(text="Configuration")
+    view_selector = View.nested(ItemsToolBarViewSelector)
+
+    @ParametrizedView.nested
+    class custom_button(ParametrizedView):  # noqa
```
digitronik: Question:
You need this static custom button method? I think better to go with dynamic as we need to check custom button dynamic as per group and button hover.

cfme/common/physical_storage_views.py
```diff
@@ -0,0 +1,125 @@
...
+class PhysicalStorageSideBar(View):
+    """Represents left side bar. It usually contains navigation, filters, etc."""
+
+    @View.nested
+    class filters(Accordion): # noqa
+        tree = ManageIQTree()
+
+
+class PhysicalStorageEntitiesView(BaseEntitiesView):
+    """Represents the view with different items like hosts."""
```
digitronik: Required:
hosts?

cfme/physical/physical_storage.py
```diff
@@ -0,0 +1,231 @@
...
+
+@attr.s
+class PhysicalStorage(BaseEntity, Updateable, Pretty, PolicyProfileAssignable, Taggable):
+    """Model of an Physical Storage in cfme.
+
+    Args:
+        name: Name of the physical Storage.
+
+    Usage:
+
```
digitronik: Required:
remove this blank line

cfme/physical/physical_storage.py
```diff
@@ -0,0 +1,231 @@
...
+        mystorage = PhysicalStorage(name='sample_Storage')
+        mystorage.create()
+
+    """
+    pretty_attrs = ['name']
+
+    name = attr.ib()
+    provider = attr.ib(default=None)
```
digitronik: Question:
Why default value `None`?

cfme/physical/physical_storage.py
```diff
@@ -0,0 +1,231 @@
...
+
+    INVENTORY_TO_MATCH = ['product_name', 'serial_number', 'description',
+                          'health_state', 'drive_bays', 'enclosure_count']
+
+    def load_details(self, refresh=False):
```
digitronik: Required:
No need to add this method. It should cover in test only not in model

cfme/physical/physical_storage.py
```diff
@@ -0,0 +1,231 @@
...
+
+        Args:
+            refresh (bool): Whether to perform the page refresh, defaults to False
+        """
+        view = navigate_to(self, "Details")
+        if refresh:
+            view.browser.refresh()
+            view.flush_widget_cache()
+
+    def _execute_button(self, button, option, handle_alert=False):
```
digitronik: Required:
Its related to a custom button. I will suggest you, please go with dynamic. As the custom button is created dynamically not a static part of a model page.

cfme/physical/physical_storage.py
```diff
@@ -0,0 +1,231 @@
...
+        view = navigate_to(self, "Details")
+        view.toolbar.custom_button(button).item_select(option, handle_alert=handle_alert)
+        return view
+
+    def _execute_action_button(self, button, option, handle_alert=True, **kwargs):
```
digitronik: Required:
similar to above

cfme/physical/physical_storage.py
```diff
@@ -0,0 +1,231 @@
...
+            self._wait_for_state_change(desired_state, target, provider, view)
+        elif handle_alert:
+            wait_for(
+                lambda: view.flash.is_displayed,
+                message="Wait for the handle alert to appear...",
+                num_sec=5,
+                delay=2
+            )
+
+    def product_name(self):
```
digitronik: Required:
It should be property

cfme/physical/physical_storage.py
```diff
@@ -0,0 +1,231 @@
...
+        view = navigate_to(self, "Details")
+        return view.entities.properties.get_text_of("Product Name")
+
+    def serial_number(self):
```
digitronik: Required;
same property

cfme/physical/physical_storage.py
```diff
@@ -0,0 +1,231 @@
...
+        view = navigate_to(self, "Details")
+        return view.entities.properties.get_text_of("Serial Number")
+
+    def health_state(self):
```
digitronik: Required:
same

cfme/physical/physical_storage.py
```diff
@@ -0,0 +1,231 @@
...
+        view = navigate_to(self, "Details")
+        return view.entities.properties.get_text_of("Health State")
+
+    def enclosure_count(self):
```
digitronik: Optional:
If those are not going to use repeatedly in your test then no need to add in a model page. You can easily use in a test whenever you want. The model page should be standard and effective methods or property.

cfme/physical/physical_storage.py
```diff
@@ -0,0 +1,231 @@
...
+        view = navigate_to(self, "Details")
+        return view.entities.properties.get_text_of("Enclosure Count")
+
+    def drive_bays(self):
```
digitronik: same

cfme/physical/physical_storage.py
```diff
@@ -0,0 +1,231 @@
...
+        view = navigate_to(self, "Details")
+        return view.entities.properties.get_text_of("Drive Bays")
+
+    def uuid(self):
```
digitronik: same

cfme/physical/physical_storage.py
```diff
@@ -0,0 +1,231 @@
...
+        view = navigate_to(self, "Details")
+        return view.entities.properties.get_text_of("UUID")
+
+    def description(self):
```
digitronik: same

cfme/physical/physical_storage.py
```diff
@@ -0,0 +1,231 @@
...
+            desired_state (str): 'on' or 'off'
+            target (str): The name of the method that most be used to compare with the desired_state
+            provider (object): 'LenovoProvider'
+            view (object): The view that most be refreshed to verify if the value was changed
+            timeout (int): Specify amount of time (in seconds) to wait until TimedOutError is raised
+            delay (int): Specify amount of time (in seconds) to repeat each time.
+        """
+
+        def _is_state_changed():
+            self.refresh(provider, handle_alert=True)
```
digitronik: Question;
Where is the refresh method in `BaseEntity`?
you can use `provider.refresh_provider_relationships`
Also,

Required:
You can accept provider as mandatory so you can use `self.provider` here.

cfme/physical/physical_storage.py
```diff
@@ -0,0 +1,231 @@
...
+
+    @property
+    def exists(self):
+        """Checks if the physical_Storage exists in the UI.
+
+        Returns: :py:class:`bool`
+        """
+        view = navigate_to(self.parent, "All")
+        try:
+            view.entities.get_entity(name=self.name, surf_pages=True)
```
digitronik: :+1:

cfme/physical/physical_storage.py
```diff
@@ -0,0 +1,231 @@
...
+        except ItemNotFound:
+            return False
+        else:
+            return True
+
+    @cached_property
+    def db_id(self):
+        return self.appliance.physical_storage_id(self.name)
+
+    def wait_to_appear(self):
```
digitronik: Required:
`wait_to_appear` and `wait_for_delete` both are depends on `exists` properly better to handle those in test only.

cfme/physical/physical_storage.py
```diff
@@ -0,0 +1,231 @@
...
+
+@attr.s
+class PhysicalStorageCollection(BaseCollection):
+    """Collection object for the :py:class:`cfme.infrastructure.host.PhysicalStorage`."""
+
+    ENTITY = PhysicalStorage
+
+    def select_entity_rows(self, *physical_storages):
+        """ Select all physical storage objects """
+        physical_storages = list(physical_storages)
```
digitronik: Required:
no need to make immutable

cfme/physical/physical_storage.py
```diff
@@ -0,0 +1,231 @@
...
+        checked_physical_storages = list()
+        view = navigate_to(self, 'All')
+
+        for physical_storage in physical_storages:
+            view.entities.get_entity(name=physical_storage.name, surf_pages=True).check()
+            checked_physical_storages.append(physical_storage)
```
digitronik: Required:
No need to append those as you are returning View.

Question:
why you are returning `view`?

cfme/physical/physical_storage.py
```diff
@@ -0,0 +1,231 @@
...
+                .query(physical_storage_table.name, ems_table.name)
+                .join(ems_table, physical_storage_table.id == ems_table.id))
+        provider = self.filters.get('provider')
+        if provider:
+            physical_storage_query = physical_storage_query.filter(ems_table.name == provider.name)
+        physical_storages = []
+        for name, ems_name in physical_storage_query.all():
+            physical_storages.append(self.instantiate(name=name,
+                                    provider=provider or get_crud_by_name(ems_name)))
+        return physical_storages
```
digitronik: :+1:

cfme/physical/physical_storage.py
```diff
@@ -0,0 +1,231 @@
...
+
+    def find_by(self, ph_name, provider=None):
```
digitronik: Question 
Why you need this method? You have all method doing the same thing for you. You can filter out in a test as well.

cfme/physical/physical_storage.py
```diff
@@ -0,0 +1,231 @@
...
+            provider = self.filters.get('provider')
+            physical_storage_query = physical_storage_query.filter(ems_table.name == provider.name)
+
+        for name, ems_name in physical_storage_query.all():
+            if ph_name == name:
+                return self.instantiate(name=name, provider=provider or get_crud_by_name(ems_name))
+
+
+@navigator.register(PhysicalStorageCollection)
+class All(CFMENavigateStep):
```
digitronik: Required:
please use short name for navigation destinations

cfme/physical/physical_storage.py
```diff
@@ -0,0 +1,231 @@
...
+    VIEW = PhysicalStoragesView
+    prerequisite = NavigateToAttribute("appliance.server", "LoggedIn")
+
+    def step(self):
+        self.prerequisite_view.navigation.select("Compute", "Physical Infrastructure", "Storages")
+
+
+@navigator.register(PhysicalStorage)
```
digitronik: Required:
Same navigation destination short name.

entry_points.txt
```diff
@@ -83,6 +83,7 @@ openstack_nodes = cfme.infrastructure.openstack_node:OpenstackNodeCollection
 orchestration_templates = cfme.services.catalogs.orchestration_template:OrchestrationTemplatesCollection
 physical_providers = cfme.physical.provider:PhysicalProviderCollection
 physical_servers = cfme.physical.physical_server:PhysicalServerCollection
+physical_storages = cfme.physical.physical_storage:PhysicalStorageCollection
```
digitronik: :+1:

cfme/physical/physical_storage.py
sshveta: Required :   
            @property
            def exists(self):
            """Checks if the physical_Storage exists in the UI.
            Returns: :py:class:`bool`
            """
            view = navigate_to(self.parent, "All")
            try:
                view.entities.get_entity(name=self.name, surf_pages=True)
                return True
            except ItemNotFound:
                return False

dajoRH:
Lint report for commit 11b6e5a2e1878ace89bd8784cc5c5bb6f51c716c:

`cfme/physical/physical_storage.py`:
- :red_circle: Line 8:1: **F401** *'widgetastic.utils.Parameter' imported but unused*
- :red_circle: Line 9:1: **F401** *'widgetastic.widget.ParametrizedView' imported but unused*
- :red_circle: Line 21:1: **F401** *'cfme.utils.log.logger' imported but unused*
- :bangbang: Line 28:1: **E302** *expected 2 blank lines, found 1*
- :bangbang: Line 79:1: **E302** *expected 2 blank lines, found 1*

Hmmm


**No commit flaws detected**.

*CFME QE Bot*

matheuscmelo:
Hi @digitronik and @sshveta, thanks for your review. Can you check?

cfme/physical/physical_storage.py
mshriver: Navigate to Details page, the step to find the entity is already built into its navigation `step()` method.

This ends up trying to reflect the `prerequisite` + `step` that Navmazing covers within the exists method directly, instead of taking advantage of the framework already in place.

See all other exists method implementations, they navigate to 'Details' - if they don't they should ;)

cfme/physical/physical_storage.py
mshriver: **Required**

I don't like this pattern of trying to reflect the UI summary table values as entity properties.  Please remove from this PR scope, or include and let's discuss some better options to reduce all this boilerplate.

This could be done with a `_getattr_` implementation that catches attributes that are detail properties and returns them from the UI, and otherwise resolves `super().__getattr__()`

cfme/physical/physical_storage.py
```diff
@@ -0,0 +1,227 @@
...
+    """
+    pretty_attrs = ['name']
+
+    name = attr.ib()
+    provider = attr.ib()
+
+    INVENTORY_TO_MATCH = ['Product Name', 'Serial Number', 'Description',
+                          'Health State', 'Drive Bays', 'Enclosure Count']
+
+    def get_value_from_summary(self, value):
```
matheuscmelo: @mshriver I added this as a separate method for code clarity and so we'll not have to change behavior if we add others functionalities which retrieves attributes from others sources. Don't know if this name is good enough though.

cfme/physical/physical_storage.py
```diff
@@ -0,0 +1,227 @@
...
+            mystorage = PhysicalStorage(name='sample_Storage')
+            mystorage.get_value_from_summary(value)
+
+        Args:
+            value (str): The Storage property you want to retrieve from Details page.
+        """
+        view = navigate_to(self, "Details")
+        return view.entities.properties.get_text_of(value)
+
+    def _wait_for_state_change(self, desired_state, target, provider, view, timeout=300, delay=10):
```
digitronik: Required:
as you used `_` at start indicating it's for internal use. I am not found any use of this method. 

I would not make this as part of the model but used in the test. If still, you like to make part of the model then please avoid the use of arguments like `provider` and `view`. Those you can take from the object only like `self.provider`.

PR #7895, opened 2018-09-27
[1LP][RFR]Extending test to include additional scenario

kedark3:
Purpose or Intent
=================

__Extending__ Mappings test to include additional scenario for host_refresh after mapping is added and check if network is still available.

{{pytest: cfme/tests/v2v/test_v2v_migrations_ui.py --use-provider rhv42 --use-provider vsphere6-nested --provider-limit 2 -vvvv --long-running -k 'test_infra_mapping_ui_assertions' }}

### DIFF
--- a/cfme/tests/v2v/test_v2v_migrations_ui.py
+++ b/cfme/tests/v2v/test_v2v_migrations_ui.py
@@ -31,7 +31,7 @@ pytestmark = [
 
 @pytest.mark.parametrize('form_data_single_datastore', [['nfs', 'nfs']], indirect=True)
 def test_infra_mapping_ui_assertions(appliance, v2v_providers, form_data_single_datastore,
-                                    soft_assert):
+                                    host_creds, conversion_tags, soft_assert):
     # TODO: This test case does not support update
     # as update is not a supported feature for mapping.
     infrastructure_mapping_collection = appliance.collections.v2v_mappings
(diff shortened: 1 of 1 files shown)


kedark3:
This test is supposed to Fail at this time due to a bug https://bugzilla.redhat.com/show_bug.cgi?id=1619431

cfme/tests/v2v/test_v2v_migrations_ui.py
```diff
@@ -50,6 +50,15 @@ def test_infra_mapping_ui_assertions(appliance, v2v_providers, form_data_single_
      mapping_list.get_map_source_networks(mapping.name)[0])
     soft_assert(mapping.form_data['network'].values()[0]['mappings'][0]['target'][0].format() in
      mapping_list.get_map_target_networks(mapping.name)[0])
+
+    rhv_prov = v2v_providers[1]
```
jawatts: **Required** You should be using the v2v namedtuple here.

kedark3: updated this and elsewhere. :)

cfme/tests/v2v/test_v2v_migrations_ui.py
```diff
@@ -50,6 +50,15 @@ def test_infra_mapping_ui_assertions(appliance, v2v_providers, form_data_single_
      mapping_list.get_map_source_networks(mapping.name)[0])
     soft_assert(mapping.form_data['network'].values()[0]['mappings'][0]['target'][0].format() in
      mapping_list.get_map_target_networks(mapping.name)[0])
+
+    rhv_prov = v2v_providers.rhv_provider
+    host = rhv_prov.hosts.all()[0]
+    host.refresh(cancel=True)
+    view = navigate_to(infrastructure_mapping_collection, 'All', wait_for_view=True)
+    mapping_list = view.infra_mapping_list
+    soft_assert(mapping_list.get_map_source_networks(mapping.name) != [])
```
digitronik: Optional:
@kedark3: when we compare with the empty list. use of `not` is more pythonic than it like...

```
In [8]: y = []

In [9]: x
Out[9]: [1, 2, 3]

In [10]: not x
Out[10]: False

In [11]: not y
Out[11]: True

In [12]: not x
Out[12]: False
```

kedark3:
Failed PRT because #8011 needs to be merged before.

cfme/tests/v2v/test_v2v_migrations_ui.py
kedark3: `not mapping_list.get_map_source_networks(mapping.name)` yields `false` as the list is not empty, and assertion fails, as assertion wants `true` so instead of saying `not mapping_list.get_map_source_networks(mapping.name) == False` I chose to use original `!= []`

PR #7898, opened 2018-09-28
[WIP] Validate Nuage network router inventory

sasoc:
With this commit we create Nuage network router and Nuage network router collection, which is then used to validate its inventory.

We create sandbox with Nuage enterprise and its entities, including network router. We use this sandbox to gather all the stats we expect to be shown on the details page. Then we find and navigate to created network router where we compare its UI stats with expected ones.

Depends on: https://github.com/ManageIQ/integration_tests/pull/7858

### DIFF
--- a/cfme/networks/provider/nuage.py
+++ b/cfme/networks/provider/nuage.py
@@ -7,6 +7,7 @@ from wrapanapi.systems import NuageSystem
 from cfme.common.provider import DefaultEndpoint, DefaultEndpointForm, EventsEndpoint
 from cfme.common.provider_views import BeforeFillMixin
 from cfme.networks.security_group import SecurityGroupCollection
+from cfme.networks.subnet import Subnet, SubnetCollection
 from cfme.utils import version
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.varmeth import variable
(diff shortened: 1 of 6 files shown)


sasoc:
cc @miha-plesko

cfme/networks/provider/nuage.py
```diff
@@ -44,6 +45,30 @@ class events(Tab, BeforeFillMixin):
         validate = Button('Validate')
 
 
+class ValidateStatsMixin(object):
```
miha-plesko: I think we introduced this mixin in one of the other PRs, no need to copy-paste it with this commit

cfme/tests/networks/nuage/test_nuage_inventory.py
```diff
@@ -51,3 +51,39 @@ def subnet_from_vmdb():
...
+            sandbox['domain'].id)
+        )
+        return appliance.collections.nuage_network_routers.find_by_ems_ref(sandbox['domain'].id)
+
+    router, _ = wait_for(
+        router_from_vmdb,
+        num_sec=60,
+        delay=5,
+        fail_condition=None
+    )
```
miha-plesko: I see we're now copy-pasting L77-88 from subnet, only this time its for tenant. Think we could introduce auxilary function, say

```python
def object_in_vmdb_with_timeout(table, ems_ref):
   ...
```

So we could reuse it for every entity then?

```python
object_in_vmdb_with_timeout('cloud_tenants', 123)
```

sasoc:
@miha-plesko I've implemented `object_in_vmdb_with_timeout` as you suggested and also changed ui pathing.

mshriver:

PR #7900, opened 2018-09-28
[RFR] Fix ovirt template upload failure

nachandr:
The ovirt templates were not being uploaded to the Sprout providers.

I have made changes to cfme/utils/template/rhevm.py so that the regular expression matches both ovirt and rhevm.

Ovirt template files  are named like this 
manageiq-ovirt-master-201809272000-96d0b65c0f.qc2
 
RHV template files are named like this:
cfme-rhevm-5.10.0.17-1.x86_64.qcow2

I have also tested the PR locally .

### DIFF
--- a/cfme/utils/template/base.py
+++ b/cfme/utils/template/base.py
@@ -22,8 +22,6 @@ from cfme.utils.log import logger
 from cfme.utils.providers import get_mgmt
 from cfme.utils.ssh import SSHClient
 
-import pytest
-
 NUM_OF_TRIES = 3
 lock = Lock()
 
@@ -127,7 +125,7 @@ class ProviderTemplateUpload(object):
             raise TemplateUploadException("Cannot get image URL.")
         else:
             image_name = self.image_pattern.findall(string_from_url)
-            if len(image_name) == 1:
+            if len(image_name):
                 # return 0th element of image_name, likely multiple formats for same provider type
                 return '/'.join([self.image_url, image_name[0]])  # TODO support multiple formats
 
--- a/cfme/utils/template/base.py
+++ b/cfme/utils/template/base.py
@@ -125,9 +125,18 @@ class ProviderTemplateUpload(object):
             raise TemplateUploadException("Cannot get image URL.")
         else:
             image_name = self.image_pattern.findall(string_from_url)
-            if len(image_name):
+            if len(image_name) == 1:
                 # return 0th element of image_name, likely multiple formats for same provider type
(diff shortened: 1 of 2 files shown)


cfme/utils/template/base.py
```diff
@@ -125,9 +125,18 @@ def raw_image_url(self):
             raise TemplateUploadException("Cannot get image URL.")
         else:
             image_name = self.image_pattern.findall(string_from_url)
-            if len(image_name):
+            if len(image_name) == 1:
                 # return 0th element of image_name, likely multiple formats for same provider type
                 return '/'.join([self.image_url, image_name[0]])  # TODO support multiple formats
+            elif len(image_name) > 1:
+                if 'manageiq' in self.image_url:  # Upstream ovirt build
+                    index = (i for i, string in enumerate(image_name) if 'ovirt' in string
```
digitronik: Required:
No need to find an index for a particular image. I would like to go with
`image_name = [img for img in image if 'ovirt' in img  and 'qc2' in img][0]`

mshriver: So I think any checks like this that specific to a provider type, should be handled in the respective provider's upload class.

This is addressing the TODO on L130, handling multiple formats.

I'm considering turning raw_image_url into a method that takes a parameter of a callback for determining which format to choose.

This can also be handled in the template_upload script itself, and iterating over all raw_image_urls returned by the method.

dajoRH:
Lint report for commit 8ead0a9833ae96afaff24cdd0dc17301f93e4d06:

`cfme/utils/template/base.py`:
- :red_circle: Line 25:1: **F401** *'pytest' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

cfme/utils/template/base.py
nachandr: @digitronik , I have updated the regex in cfme/utils/template/rhevm.py to match *rhevm*.qcow2 or *ovirt*.qc2 files,
so the changes in cfme/utils/template/base.py are not needed anymore for RHV/Ovirt.

However, we still have to add support for multiple image formats for same provider type.I will be doing that in a follow up PR.

cfme/utils/template/base.py
```diff
@@ -125,9 +125,18 @@ def raw_image_url(self):
...
+                        and 'qc2' in string).next()
+                    return '/'.join([self.image_url, image_name[index]])
+                elif 'cfme' in self.image_url:     # Downstream RHV build
```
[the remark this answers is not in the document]
nachandr: This is not relevant any more with my latest commit.

PR #7903, opened 2018-10-01
[1LP][RFR] Quota Test For Cloud: Testing tenant quota via UI and SSUI 

ganeshhubale:
- This PR tests Tenant Quota in UI and SSUI by selecting default entry point(field_entry_point) while creating catalog item.
- Added new fixture:
    - **set_default()**
        - This modification applied on all tests which are creating **catalog item** from this file. 
        - This fixture used to modify **field_entry_point**
        - Some of the tests from this file do not require **field_entry_point** except new test -  ```test_service_cloud_tenant_quota_with_default_entry_point```
- **provider_type** while creating catalog item
    - It is used to differentiate ``` field_entry_point``` for provider types because field_entry_point is different for **cloud** providers and **infra** providers.

{{ pytest: cfme/tests/cloud/test_tenant_quota.py -k 'test_service_cloud_tenant_quota_with_default_entry_point or test_tenant_quota_enforce_via_service_cloud' -qsvvv }}

### DIFF
--- a/cfme/services/catalogs/catalog_items/__init__.py
+++ b/cfme/services/catalogs/catalog_items/__init__.py
@@ -20,19 +20,15 @@ from cfme.utils.wait import wait_for
 
 class EntryPoint(Input):
     def fill(self, value):
-        current_value = self.value
-        if value == current_value:
-            return False
-        # Clear and type everything
-        self.browser.click(self)
-        self.browser.clear(self)
-        self.browser.send_keys(value, self)
-        self.parent_view.modal.cancel.click()
-        # After changing default path values of 'retirement_entry_point', 'reconfigure_entry_point'
-        # or 'provisioning_entry_point'; 'Select Entry Point Instance'(pop up) occures which also
-        # helps to select these paths using tree structure. Here we have already filled these values
-        # through text input. So to ignore this pop up clicking on cancel button is required.
-        return True
+        if super(EntryPoint, self).fill(value):
+            self.parent_view.modal.cancel.click()
+            # After changing default path values of 'retirement_entry_point',
+            # 'reconfigure_entry_point' or 'provisioning_entry_point';
+            # 'Select Entry Point Instance'(pop up) occures which also helps to select these paths
+            # using tree structure. Here we have already filled these values through text input.
+            # So to ignore this pop up clicking on cancel button is required.
+            return True
+        return False
 
 
 # Views
--- a/cfme/services/catalogs/catalog_items/__init__.py
+++ b/cfme/services/catalogs/catalog_items/__init__.py
@@ -488,13 +488,13 @@ class CatalogItemsCollection(BaseCollection):
     def instantiate(self, catalog_item_class, *args, **kwargs):
         return catalog_item_class.from_collection(self, *args, **kwargs)
 
-    def create(self, catalog_item_class, provider_type=None, field_entry_point=False, *args,
+    def create(self, catalog_item_class, provider, field_entry_point=False, *args,
                **kwargs):
         """Creates a catalog item in the UI.
 
         Args:
             catalog_item_class: type of a catalog item
-            provider_type: Used to distinguish default entry point path
+            provider: Used to distinguish default entry point path
             field_entry_point: Used to add default entry point
             *args: see the respectful catalog item class
             **kwargs: see the respectful catalog item class
@@ -503,19 +503,20 @@ class CatalogItemsCollection(BaseCollection):
             An instance of catalog_item_class
         """
         cat_item = self.instantiate(catalog_item_class, *args, **kwargs)
-        view = navigate_to(cat_item, 'Add')
+        view = navigate_to(cat_item, "Add")
         if field_entry_point:
-            if provider_type == 'cloud':
-                view.basic_info.field_entry_point.fill("")
-                view.basic_info.modal.tree.click_path('Datastore', 'ManageIQ (Locked)', 'Cloud',
-                                                      'VM', 'Provisioning', 'StateMachines',
-                                                      'ProvisionRequestApproval', 'Default')
-            elif provider_type == 'infra':
-                view.basic_info.field_entry_point.fill("")
-                view.basic_info.modal.tree.click_path('Datastore', 'ManageIQ (Locked)',
-                                                      'Infrastructure', 'VM', 'Provisioning',
-                                                      'StateMachines', 'ProvisionRequestApproval',
-                                                      'Default')
+            prov_type = "Cloud" if provider.category == "cloud" else "Infrastructure"
+            view.basic_info.field_entry_point.fill("")
+            view.basic_info.modal.tree.click_path(
+                "Datastore",
+                "ManageIQ (Locked)",
+                prov_type,
+                "VM",
+                "Provisioning",
+                "StateMachines",
+                "ProvisionRequestApproval",
+                "Default",
+            )
             view.basic_info.modal.include_domain.fill(True)
             view.basic_info.modal.apply.click()
         view.fill(cat_item.fill_dict)
--- a/cfme/services/catalogs/catalog_items/__init__.py
+++ b/cfme/services/catalogs/catalog_items/__init__.py
@@ -488,20 +488,25 @@ class CatalogItemsCollection(BaseCollection):
     def instantiate(self, catalog_item_class, *args, **kwargs):
         return catalog_item_class.from_collection(self, *args, **kwargs)
 
-    def create(self, catalog_item_class, provider, field_entry_point=False, *args,
-               **kwargs):
+    def create(self, catalog_item_class, *args, **kwargs):
         """Creates a catalog item in the UI.
 
         Args:
             catalog_item_class: type of a catalog item
-            provider: Used to distinguish default entry point path
-            field_entry_point: Used to add default entry point
             *args: see the respectful catalog item class
             **kwargs: see the respectful catalog item class
 
         Returns:
             An instance of catalog_item_class
         """
+        if args:
+            """args should not contain any value while instantiating cat_item.
+               Hence released 'provider' and 'field_entry_point' elements from args tuple."""
+            provider, field_entry_point = args[0], args[1]
+            release_extra_args = list(args)
+            release_extra_args.pop(1)
+            release_extra_args.pop(0)
+            args = tuple(release_extra_args)
         cat_item = self.instantiate(catalog_item_class, *args, **kwargs)
         view = navigate_to(cat_item, "Add")
         if field_entry_point:
--- a/cfme/services/catalogs/catalog_items/__init__.py
+++ b/cfme/services/catalogs/catalog_items/__init__.py
@@ -488,11 +488,14 @@ class CatalogItemsCollection(BaseCollection):
     def instantiate(self, catalog_item_class, *args, **kwargs):
         return catalog_item_class.from_collection(self, *args, **kwargs)
 
-    def create(self, catalog_item_class, *args, **kwargs):
+    def create(self, catalog_item_class, provider_type=None, field_entry_point=False, *args,
+               **kwargs):
         """Creates a catalog item in the UI.
 
         Args:
             catalog_item_class: type of a catalog item
+            provider_type: Used to distinguish default entry point path
+            field_entry_point: Used to add default entry point
             *args: see the respectful catalog item class
             **kwargs: see the respectful catalog item class
 
@@ -501,6 +504,20 @@ class CatalogItemsCollection(BaseCollection):
         """
         cat_item = self.instantiate(catalog_item_class, *args, **kwargs)
         view = navigate_to(cat_item, 'Add')
+        if field_entry_point:
+            if provider_type == 'cloud':
+                view.basic_info.field_entry_point.fill("")
+                view.basic_info.modal.tree.click_path('Datastore', 'ManageIQ (Locked)', 'Cloud',
+                                                      'VM', 'Provisioning', 'StateMachines',
+                                                      'ProvisionRequestApproval', 'Default')
+            elif provider_type == 'infra':
+                view.basic_info.field_entry_point.fill("")
+                view.basic_info.modal.tree.click_path('Datastore', 'ManageIQ (Locked)',
+                                                      'Infrastructure', 'VM', 'Provisioning',
+                                                      'StateMachines', 'ProvisionRequestApproval',
+                                                      'Default')
+            view.basic_info.modal.include_domain.fill(True)
+            view.basic_info.modal.apply.click()
         view.fill(cat_item.fill_dict)
         view.add.click()
         view = self.create_view(AllCatalogItemView)
--- a/cfme/services/catalogs/catalog_items/__init__.py
+++ b/cfme/services/catalogs/catalog_items/__init__.py
@@ -352,6 +352,12 @@ class CloudInfraCatalogItem(BaseCatalogItem):
     domain = attr.ib(default='ManageIQ (Locked)')
     provider = attr.ib(default=None)
     item_type = None
+    provisioning_entry_point = attr.ib(default=(
+        "/Service/Provisioning/StateMachines/ServiceProvision_Template/CatalogItemInitialization"))
+    retirement_entry_point = attr.ib(
+        default="/Service/Retirement/StateMachines/ServiceRetirement/Default"
+    )
+    reconfigure_entry_point = attr.ib(default=None)
 
     @property
     def fill_dict(self):
@@ -361,7 +367,10 @@ class CloudInfraCatalogItem(BaseCatalogItem):
                 'description': self.description,
                 'display': self.display_in,
                 'select_catalog': self.catalog_name,
-                'select_dialog': self.dialog
+                'select_dialog': self.dialog,
+                'provisioning_entry_point': self.provisioning_entry_point,
+                'retirement_entry_point': self.retirement_entry_point,
+                'reconfigure_entry_point': self.reconfigure_entry_point
             },
             'request_info': {'provisioning': self.prov_data}
         }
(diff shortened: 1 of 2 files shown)


dajoRH:
*CFME QE Bot*

cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -472,11 +472,14 @@ def ANSIBLE_PLAYBOOK(self):  # noqa
     def instantiate(self, catalog_item_class, *args, **kwargs):
         return catalog_item_class.from_collection(self, *args, **kwargs)
 
-    def create(self, catalog_item_class, *args, **kwargs):
+    def create(self, catalog_item_class, provider_type=None, field_entry_point=False, *args,
```
digitronik: Are you able to access `provider` `attribute` in a method?
if yes then 
Required: you can use internally `provider.category`. don't pass it form test to `create` method

cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -485,6 +488,20 @@ def create(self, catalog_item_class, *args, **kwargs):
         """
         cat_item = self.instantiate(catalog_item_class, *args, **kwargs)
         view = navigate_to(cat_item, 'Add')
+        if field_entry_point:
+            if provider_type == 'cloud':
+                view.basic_info.field_entry_point.fill("")
```
`field entry_point` is common for both `infra` and `cloud` so you can put this about conditions.

cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -485,6 +488,20 @@ def create(self, catalog_item_class, *args, **kwargs):
...
+                view.basic_info.modal.tree.click_path('Datastore', 'ManageIQ (Locked)', 'Cloud',
+                                                      'VM', 'Provisioning', 'StateMachines',
+                                                      'ProvisionRequestApproval', 'Default')
+            elif provider_type == 'infra':
+                view.basic_info.field_entry_point.fill("")
+                view.basic_info.modal.tree.click_path('Datastore', 'ManageIQ (Locked)',
+                                                      'Infrastructure', 'VM', 'Provisioning',
```
In this tree path `Cloud/Infrastructure` changes. So, better to manipulate that only... what you say?

cfme/tests/cloud/test_tenant_quota.py
```diff
@@ -143,3 +156,48 @@ def test_tenant_quota_enforce_via_service_cloud(request, appliance, provider, se
...
+    6. Order the catalog item via UI and SSUI individually
+    """
+
+    with appliance.context.use(context):
+        service_catalogs = ServiceCatalogs(appliance, catalog_item.catalog, catalog_item.name)
+        if context is ViaSSUI:
+            service_catalogs.add_to_shopping_cart()
+        service_catalogs.order()
+    # nav to requests page to check quota validation
+    request_description = 'Provisioning Service [{0}] from [{0}]'.format(catalog_item.name)
```
digitronik: I would like to use `key` than `Index`

cfme/services/catalogs/catalog_items/__init__.py
ganeshhubale: @digitronik 
I am not able access **provider** attribute directly in ```create``` method but  I can pass provider attribute to create method from test like instead of passing ```provider=provider.category```; we can pass this - ```provider=provider```
Please let me know if you have any other solution around it.
Thank you.

cfme/services/catalogs/catalog_items/__init__.py
ganeshhubale: Yes, this is very useful. I am updating PR with this suggestion.

sshveta:
@ganeshhubale : Can you please explain the test case here.
How does selecting default entry point value impact quota ?

sshveta:
Talked to Ganesh and changing default entry point does not impact quota at all .
So we are making unnecessary changes to catalog file which is being used in lot of services tests.
I suggest to close this PR and test quota with other required conditions.

ganeshhubale:
@sshveta @mshriver 
We are already checking whether quota is working with other conditions or not. 
This test case is needed because quota has to be checked if its working with default entry point.

cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -472,19 +472,37 @@ def ANSIBLE_PLAYBOOK(self):  # noqa
     def instantiate(self, catalog_item_class, *args, **kwargs):
         return catalog_item_class.from_collection(self, *args, **kwargs)
 
-    def create(self, catalog_item_class, *args, **kwargs):
+    def create(self, catalog_item_class, provider, field_entry_point=False, *args,
```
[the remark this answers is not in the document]
ganeshhubale: @sshveta 
I used args for passing parameters(provider and set_default) to create method.
Please review the  changes and let me know if if you have any suggestions around it.

sshveta:
One small change

cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -483,8 +483,31 @@ def create(self, catalog_item_class, *args, **kwargs):
         Returns:
             An instance of catalog_item_class
         """
+        if args:
+            """args should not contain any value while instantiating cat_item.
+               Hence released 'provider' and 'field_entry_point' elements from args tuple."""
+            provider, field_entry_point = args[0], args[1]
```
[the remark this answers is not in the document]
ganeshhubale: @sshveta 
Thanks for this suggestion with explanation. It is better than using args for this test case.
Please take a look at modifications - using kwargs to pass parameters.
Let me know if this PR needs any changes.

cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -483,8 +483,33 @@ def create(self, catalog_item_class, *args, **kwargs):
         Returns:
             An instance of catalog_item_class
         """
+        if kwargs:
+            provider = kwargs.get("provider", None)
```
mshriver: Remove default get param of `None` in these calls.

cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -483,8 +483,33 @@ def create(self, catalog_item_class, *args, **kwargs):
         Returns:
             An instance of catalog_item_class
         """
+        if kwargs:
```
mshriver: this check is unnecessary, kwargs will be defined, and you're using `get` to reference expected keys.

cfme/tests/cloud/test_tenant_quota.py
```diff
@@ -16,6 +16,15 @@
 ]
 
 
+@pytest.fixture
+def set_default(request):
+    value = request.param
+    if value:
```
mshriver: You can just `return bool(value)` here.

cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -483,8 +483,33 @@ def create(self, catalog_item_class, *args, **kwargs):
...
+                """
+                kwargs.pop("provider")
+                kwargs.pop("field_entry_point")
         cat_item = self.instantiate(catalog_item_class, *args, **kwargs)
-        view = navigate_to(cat_item, 'Add')
+        view = navigate_to(cat_item, "Add")
+        if field_entry_point:
+            prov_type = "Cloud" if provider.category == "cloud" else "Infrastructure"
+            view.basic_info.field_entry_point.fill("")
+            view.basic_info.modal.tree.click_path(
```
mshriver: This is all hardcoded fill values that are specific to a test or a set of assumptions.  They can be kept as a default value (given the updates I describe below)

Overall, this does not seem like the correct way to approach setting this value for a catalog item within its CRUD method.

field_entry_point should be an attribute of the catalog item, passed to its instantiate method as a kwarg.  With it passed directly to create, there's no modeling of the value after the catalog item has been created - we don't have a reference for what the value should be in the UI, we only have what's actually displayed.

With field_entry_point existing as a CloudInfraCatalogItem attribute, it can be included in the `fill_dict` property that's used in this `create` method to fill the form.  field_entry_point is already a widget defined on this form.

If the widget does not fill correctly, as you appear to be doing some custom fill actions here (filling with emptry sting, then using the modal tree) we can evaluate the structure of the fill dict and the catalog item attributes so that they are set and then fill correctly.  We may need to use before_fill for that view.

cfme/tests/cloud/test_tenant_quota.py
```diff
@@ -16,6 +16,15 @@
 ]
 
 
+@pytest.fixture
+def set_default(request):
```
mshriver: Can you explain why this has to be an indirect fixture? Its simply a boolean value that you can directly parametrize on

cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -483,8 +483,33 @@ def create(self, catalog_item_class, *args, **kwargs):
...
+                """kwargs should not contain value of 'provider' and 'field_entry_point'
+                    while instantiating cat_item. Hence released 'provider' and 'field_entry_point'
+                    elements from kwargs.
+                """
+                kwargs.pop("provider")
+                kwargs.pop("field_entry_point")
         cat_item = self.instantiate(catalog_item_class, *args, **kwargs)
-        view = navigate_to(cat_item, 'Add')
+        view = navigate_to(cat_item, "Add")
+        if field_entry_point:
```
mshriver: This is checking field_entry_point, and then making a reference to `provider.category` - where `provider` could be `None`.  It should be checking both, and per my other comments may not even be necessary here when field_entry_point is filled via `fill_dict`.

Why not also make provider a CloudInfraCatalogItem attribute, that's optional to maintain current implementations?

cfme/tests/cloud/test_tenant_quota.py
ganeshhubale: Hi @mshriver 
- Now this fixture does not return only ```boolean``` value. So I have modified this fixture and it returns three values:  ```provisioning_entry_point```, ```reconfigure_entry_point```, ```retirement_entry_point```.
- Out of these three values ```provisioning_entry_point``` is required for this test. Hence I put  other two values to be ```None```.
- And the tests which does not require these values, this fixture returns ```None``` for those test.

cfme/services/catalogs/catalog_items/__init__.py
ganeshhubale: @mshriver 
- I have removed this check and updated the way of filling ```entry_point``` while creating catalog items.
- I have added three  checks for [three values](https://github.com/ManageIQ/integration_tests/pull/7903#discussion_r239798559) as I passed these values using ```kwargs```.
- About ```provider``` object: Now we don't need this in create method.

cfme/services/catalogs/catalog_items/__init__.py
ganeshhubale: @mshriver 
- As you said, I tried to remove all hard-coded fill values that are specific to  this test.
- Instead of using ```before_fill``` method or including ```field_entry_point``` to ```fill_dict```. I have added [three checks for three path values](https://github.com/ManageIQ/integration_tests/pull/7903/files#diff-3a85f436b2ce6046c949bfc47f9b8b73R488) in create method.
- Let me know, if this is a better way to go with or we must have to either add ```field_entry_point``` to ```fill_dict``` or use ```before_fill``` method.


Thank you for this explanation.

cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -483,8 +484,26 @@ def create(self, catalog_item_class, *args, **kwargs):
...
+        retirement_entry_point = kwargs.get("retirement_entry_point", None)
+        kwargs.pop("provisioning_entry_point")
+        kwargs.pop("reconfigure_entry_point")
+        kwargs.pop("retirement_entry_point")
         cat_item = self.instantiate(catalog_item_class, *args, **kwargs)
-        view = navigate_to(cat_item, 'Add')
+        view = navigate_to(cat_item, "Add")
+        if provisioning_entry_point:
+            view.basic_info.field_entry_point.fill(provisioning_entry_point)
+            view.basic_info.modal.cancel.click()
```
mshriver: What is the purpose of this cancel click?

This still bakes view behavior into the `Entity.create` method, instead of defining that behavior in the view's fill methods.

Separating the entry point fields is good - but they still should be attributes of the catalog item entity class, and included in the kwargs used to instantiate a catalog item.

The changes you've implemented are an improvement, but in my view don't go far enough to correctly model the behavior of the MIQ entity and the MIQ views.

This type of behavior should be part of the `widget.after_fill` method of the widgets used for these entry point fields.

```python
## double check me on the kwargs to after_fill here, writing as example not as runnable code

class FieldEntryPoint(Input):
    def after_fill(self, was_changed):
        if was_changed:
            self.parent_view.modal.cancel.click()
        else:
            pass
```

The above widget would be used for `BasicInfoForm` widgets for the entry points.

Then you can just have a single call to `view.fill(cat_item.fill_dict)` that will also fill the entry point fields and click modal cancel buttons where appropriate.

If there is a need to fill entry points before the rest of the form, this can be achieved with `BasicInfoForm.before_fill` implementation.

digitronik: @mshriver:  I am suspect about this  implementation; please clear my doubts. 

 `Input` is  `widget` and `after_fill` is `View` class method. `Widget` class is  used to create `Input` widget. `Widget` class is parent for `View` class as per inheritance. So as per me this will not work.

`fill` method call will point `Input` widget `fill` method; not `View` class `fill`.  so as per me `aftert_fill` method will not to call. This is magic method for  `Views` not for `Widget`.
I will vote for simple `fill` method overwrite  in this case. 

```
class FieldEntryPoint(Input):
    def fill(self, value):
        current_value = self.value
        if value == current_value:
            return False
        self.browser.click(self)
        self.browser.clear(self)
        self.browser.send_keys(value, self)
        self.parent_view.modal.cancel.click()
        return True
```

cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -18,6 +18,23 @@
...
+class EntryPoint(Input):
+    def fill(self, value):
+        current_value = self.value
+        if value == current_value:
+            return False
+        # Clear and type everything
+        self.browser.click(self)
+        self.browser.clear(self)
+        self.browser.send_keys(value, self)
+        self.parent_view.modal.cancel.click()
```
mshriver: Let's not re-implement `widgetastic.widget.TextInput.Fill` here, let's use `super().fill()` here to get through filling of the field, and then add on the `parent_view` access and widget interaction if `TextInput.Fill` returns True.

cfme/services/catalogs/catalog_items/__init__.py
mshriver: @digitronik you're absolutely right on widget vs view here, good call. Made comment on the new code, I think we can reduce duplication, but this is the correct approach, and the in-line comments in the current commit are good.

cfme/services/catalogs/catalog_items/__init__.py
ganeshhubale: @mshriver 
Can you please explain it for me with example. I am not getting what you are expecting.
Thank you :)

mshriver: So all of the actions in this fill method are the exact same as `TextInput.fill()` method, which is a parent class of this specific Input derived class.

By using `super().fill()`, we'll resolve the `fill()` method from our parent class, and it will execute the fill steps up through calling `self.browser.send_keys()`. Then you can check if `super().fill()` returns True (field was filled in), you can additionally call `self.parent_view.modal.cancel.click()`.

That way any changes to how `TextInput.fill()` works don't have to be copied here as well.

ganeshhubale: Thank you for explanation :)

ganeshhubale: Thanks @mshriver and @digitronik 
Finally, I am able to modify ```fill``` method as shown below:
```

class EntryPoint(Input):
    def fill(self, value):
        if super(EntryPoint, self).fill(value):
            self.parent_view.modal.cancel.click()
            # After changing default path values of 'retirement_entry_point',
            # 'reconfigure_entry_point' or 'provisioning_entry_point';
            # 'Select Entry Point Instance'(pop up) occures which also helps to select these paths
            # using tree structure. Here we have already filled these values through text input.
            # So to ignore this pop up clicking on cancel button is required.
            return True
        return False
```

cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -499,8 +513,20 @@ def create(self, catalog_item_class, *args, **kwargs):
         Returns:
             An instance of catalog_item_class
         """
+        provisioning_entry_point = kwargs.pop("provisioning_entry_point", None)
```
mshriver: Why are these fields only passed to the collection create method, and not included as catalog item attributes? These are fields that are part of a catalog item definition, and `cat_item.fill_dict` property could just include them so that they're filled by the default fill mechanisms.

ganeshhubale: If I add these three fields to the ```fill_dict``` method. then my observations are like:
1. When we execute ```cat_item.fill_dict``` then we can see these fileds are filled with our values. But it is not able to execute ```self.parent_view.modal.cancel.click()```. So that **pop-up** window not get cancelled after filling these fields.

cfme/tests/cloud/test_tenant_quota.py
```diff
@@ -12,8 +12,27 @@
...
+       But other tests does not require these values since those tests takes default values hence
+       providing default value. So in this file, this fixture - 'set_default'
+       must be used in all tests of quota which are related to services where catalog item needs to
+       be created with specific values for these entries.
+    """
+
+    provisioning_entry_point = (
+        "/Service/Provisioning/StateMachines/ServiceProvision_Template/CatalogItemInitialization")
+    if request.param:
+        provisioning_entry_point = ("/ManageIQ (Locked)/" + provider.string_name
```
mshriver: Use `''.format()` for this composition, not string concatenation.

cfme/tests/cloud/test_tenant_quota.py
```diff
@@ -155,3 +178,62 @@ def test_tenant_quota_enforce_via_service_cloud(request, appliance, provider, se
     def delete():
         provision_request.remove_request()
         catalog_item.delete()
+
+
+# Args of parametrize is the list of navigation parameters to order catalog item.
+@pytest.mark.parametrize('context', [ViaSSUI, ViaUI])
```
mshriver: Formatting a bit off here, keep the comment block together and the parametrize lines together, instead of this splitting the comment block please.

PR #7904, opened 2018-10-01
[1LP][RFR] Automate test: Create schedule for base report

valaparthvi:
This PR adds a test case, which creates a schedule for a base report and asserts the schedule was created without any error.

{{pytest: cfme/tests/intelligence/reports/test_crud.py::test_reports_crud_schedule_for_base_report_once --use-template-cache -sqvvv}}

### DIFF
--- a/cfme/intelligence/reports/reports.py
+++ b/cfme/intelligence/reports/reports.py
@@ -11,6 +11,7 @@ from widgetastic_manageiq import (PaginationPane, Table, ReportToolBarViewSelect
 from widgetastic_manageiq.expression_editor import ExpressionEditor
 from widgetastic_patternfly import Button, Input, BootstrapSelect, Tab, CandidateNotFound
 
+from cfme.intelligence.reports.schedules import NewScheduleView
 from cfme.modeling.base import BaseCollection, BaseEntity
 from cfme.utils import ParamClassName
 from cfme.utils.appliance.implementations.ui import navigator, CFMENavigateStep, navigate_to
@@ -312,6 +313,42 @@ class Report(BaseEntity, Updateable):
     def saved_reports(self):
         return self.collections.saved_reports
 
+    def create_schedule(self, name=None, description=None, active=None,
+            timer=None, emails=None, email_options=None):
+
+        view = navigate_to(self, "ScheduleReport")
+        view.fill({
+            "name": name,
+            "description": description,
+            "active": active,
+            "run": timer.get("run"),
+            "time_zone": timer.get("time_zone"),
+            "starting_date": timer.get("starting_date"),
+            "hour": timer.get("hour"),
+            "minute": timer.get("minute"),
+            "emails_send": bool(emails),
+            "emails": emails,
+            "send_if_empty": email_options.get("send_if_empty"),
+            "send_txt": email_options.get("send_txt"),
+            "send_csv": email_options.get("send_csv"),
+            "send_pdf": email_options.get("send_pdf")
+        })
+
+        view.add_button.click()
+        view.flash.assert_no_error()
+
+        schedule = self.appliance.collections.schedules.instantiate(
+            name=name or self.menu_name,
+            description=description,
+            filter=(self.company_name, self.subtype, self.menu_name),
+            active=active,
+            emails=emails,
+            email_options=email_options,
+        )
+
+        assert schedule.exists
+        return schedule
+
     def queue(self, wait_for_finish=False):
         view = navigate_to(self, "Details")
         view.report_info.queue_button.click()
@@ -555,6 +592,20 @@ class ReportDetails(CFMENavigateStep):
         )
         self.view.report_info.select()
 
+
+@navigator.register(Report, "ScheduleReport")
+class ReportSchedule(CFMENavigateStep):
+    VIEW = NewScheduleView
+    prerequisite = NavigateToAttribute("appliance.server", "CloudIntelReports")
+
+    def step(self):
+        self.prerequisite_view.reports.tree.click_path(
+            "All Reports",
+            self.obj.type or self.obj.company_name,
+            self.obj.subtype or "Custom",
+            self.obj.menu_name
+        )
+        self.prerequisite_view.configuration.item_select("Add a new Schedule")
 
 @navigator.register(SavedReport, "Details")
 class SavedReportDetails(CFMENavigateStep):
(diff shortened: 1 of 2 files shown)


cfme/tests/intelligence/reports/test_crud.py
```diff
@@ -264,3 +264,23 @@ def test_reports_delete_saved_report(appliance, request):
...
+    report = appliance.collections.reports.instantiate(
+        type="Configuration Management",
+        subtype="Virtual Machines",
+        menu_name="Hardware Information for VMs",
+    )
+    view = navigate_to(report, "Details")
+    view.configuration.item_select("Add a new Schedule")
+    add_view = report.create_view(NewScheduleView)
+    data = {"hour": "12", "minute": "10"}
+    add_view.fill(data)
```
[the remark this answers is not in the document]
valaparthvi: When I go to the Details page of a report and click on 'Add new Schedule', it opens up the form with some details pre-filled, I want to preserve those details and only make a few changes, before adding the schedule. Using the `create` method would direct to 'Add new Schedule' page and load an empty form.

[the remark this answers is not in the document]
valaparthvi: And I am only creating a schedule here, not deleting it, so I don't think it's CRUD operation. Although, I could change the test name to something shorter.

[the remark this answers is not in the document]
digitronik: better you can import this `NewScheduleView` in `report.py`. create a method in `Report` class.

valaparthvi:
@digitronik 
I have made the changes you requested.
1) Add a `ReportSchedule` view, which can be used to add a schedule from report's `Details` page.
2) Add a `create_schedule` method, which creates a schedule for the base report.
3) Modify test in accordance with the new changes.

cfme/intelligence/reports/reports.py
```diff
@@ -312,6 +313,42 @@ def delete(self, cancel=False):
     def saved_reports(self):
         return self.collections.saved_reports
 
+    def create_schedule(self, name=None, description=None, active=None,
```
digitronik: Required:
Why all argument default value `None`?

cfme/intelligence/reports/reports.py
```diff
@@ -312,6 +313,42 @@ def delete(self, cancel=False):
...
+            "send_txt": email_options.get("send_txt"),
+            "send_csv": email_options.get("send_csv"),
+            "send_pdf": email_options.get("send_pdf")
+        })
+
+        view.add_button.click()
+        view.flash.assert_no_error()
+
+        schedule = self.appliance.collections.schedules.instantiate(
+            name=name or self.menu_name,
```
digitronik: Question:
can you please explain this?

cfme/tests/intelligence/reports/test_crud.py
```diff
@@ -264,3 +263,26 @@ def test_reports_delete_saved_report(appliance, request):
     view.configuration.item_select(
         item='Delete selected Saved Reports', handle_alert=True)
     assert not report.exists
+
+
+def test_reports_create_schedule_for_base_report_once(appliance, request):
```
digitronik: Required:
This should be a crud operation.
Please create and delete this scheduled report properly and assert for both
remove finalizer

cfme/intelligence/reports/reports.py
```diff
@@ -556,6 +593,20 @@ def step(self):
...
 
+@navigator.register(Report, "ScheduleReport")
+class ReportSchedule(CFMENavigateStep):
+    VIEW = NewScheduleView
+    prerequisite = NavigateToAttribute("appliance.server", "CloudIntelReports")
+
+    def step(self):
+        self.prerequisite_view.reports.tree.click_path(
+            "All Reports",
+            self.obj.type or self.obj.company_name,
```
digitronik: I am suspect about these or. are those required?

cfme/intelligence/reports/reports.py
valaparthvi: That is in the case when no argument value is passed in the data.

cfme/intelligence/reports/reports.py
valaparthvi: This instantiates the newly created schedule and returns it.

cfme/intelligence/reports/reports.py
valaparthvi: It is required to navigate to the details page of the report and then select `Add a new Schedule` from the configuration menu.

valaparthvi:
@digitronik I have added a new `ReportScheduleView`, to fix the failing `is_displayed` method. I had done a few things incorrectly, but have now fixed it.

cfme/intelligence/reports/reports.py
```diff
@@ -225,6 +226,24 @@ def is_displayed(self):
         )
 
 
+class ReportScheduleView(SchedulesFormCommon):
```
mshriver: Hang on I'm pretty sure all of this already exists in the schedules.py module:
https://github.com/ManageIQ/integration_tests/blob/master/cfme/intelligence/reports/schedules.py

cfme/tests/intelligence/reports/test_crud.py
```diff
@@ -264,3 +263,26 @@ def test_reports_delete_saved_report(appliance, request):
     view.configuration.item_select(
         item='Delete selected Saved Reports', handle_alert=True)
     assert not report.exists
+
+
+def test_reports_crud_schedule_for_base_report_once(appliance, request):
```
mshriver: I think this case is covered by test_schedule_queue?  

https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/intelligence/reports/test_reports_schedules.py

This uses several schedules from the cfme/data/ directory.

valaparthvi:
Hey @mshriver, we have views and objects to create a completely new schedule, but as far as I know, there is no way to create a schedule for an existing report. Take this for example:
1) Navigate to a report's `Details` page.
2) Click on `Configuration` and select `Add a new schedule`.
3) You'll see a form with some pre-filled data such as `Filters`, `Name` and `Description`.

Take a look at [this](https://up1-vrutkovs.int.open.paas.redhat.com/#oyZ7jTaXqjU8xs6ig-Uv_g).
I don't think we have any view or object that can directly create a schedule for a report like this.

And the test case is similar to `test_schedule_queue`, but `test_reports_crud_schedule_for_base_report_once` uses the new view and method.

But, there is one drawback I hadn't noticed before. The method `create_schedule` will fail if no data is passed since the `Add` button on New Schedule page does not activate until and unless some change is made in the form.

Let me know what you think.

PR #7906, opened 2018-10-01
[1LP][RFR] Automate Test: Add schedule for custom report

valaparthvi:
This PR adds a test case, which creates a custom report and checks if a schedule can be created for it.

{{pytest: cfme/tests/intelligence/reports/test_crud.py::test_crud_custom_report_schedule --use-template-cache -sqvvv }}

### DIFF
--- a/cfme/tests/intelligence/reports/test_crud.py
+++ b/cfme/tests/intelligence/reports/test_crud.py
@@ -11,7 +11,6 @@ from cfme.utils.path import data_path
 from cfme.utils.rest import assert_response
 from cfme.utils.update import update
 from cfme.utils.wait import wait_for_decorator
-from widgetastic.exceptions import NoSuchElementException, MoveTargetOutOfBoundsException
 
 
 report_crud_dir = data_path.join("reports_crud")
@@ -302,12 +301,7 @@ def test_create_custom_report_schedule(
         "Custom",
         custom_report.menu_name,
     )
-    try:
-        custom_report_schedule = appliance.collections.schedules.create(**schedule_data)
-    except (MoveTargetOutOfBoundsException, NoSuchElementException):
-        # Temporary workaround - running the `create` method first time raises exceptions,
-        # but running it again works fine.
-        custom_report_schedule = appliance.collections.schedules.create(**schedule_data)
+    custom_report_schedule = appliance.collections.schedules.create(**schedule_data)
     assert custom_report_schedule.exists
 
     request.addfinalizer(custom_report.delete)
--- a/cfme/tests/intelligence/reports/test_crud.py
+++ b/cfme/tests/intelligence/reports/test_crud.py
@@ -49,6 +49,13 @@ def schedule_data(request):
         return yaml.load(rep_yaml)
 
 
+@pytest.fixture(scope="function")
+def get_custom_report(appliance, custom_report_values):
+    custom_report = appliance.collections.reports.create(**custom_report_values)
+    yield custom_report
+    custom_report.delete()
+
+
 @pytest.mark.rhel_testing
 @pytest.mark.sauce
 @pytest.mark.tier(3)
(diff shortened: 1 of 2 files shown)


cfme/tests/intelligence/reports/test_crud.py
```diff
@@ -264,3 +266,21 @@ def test_reports_delete_saved_report(appliance, request):
...
+):
+    custom_report = appliance.collections.reports.create(**custom_report_values)
+    schedule_data["filter"] = (
+        "My Company (All Groups)",
+        "Custom",
+        custom_report.menu_name,
+    )
+    try:
+        custom_report_schedule = appliance.collections.schedules.create(**schedule_data)
+    except (MoveTargetOutOfBoundsException, NoSuchElementException):
```
digitronik: Que:
`MoveTargetOutOfBoundsException` still comming?

cfme/tests/intelligence/reports/test_crud.py
```diff
@@ -264,3 +266,21 @@ def test_reports_delete_saved_report(appliance, request):
...
+        custom_report_schedule = appliance.collections.schedules.create(**schedule_data)
```
sshveta: Question : 
Your try and except both are doing the same thing??
custom_report_schedule = appliance.collections.schedules.create(**schedule_data)

cfme/tests/intelligence/reports/test_crud.py
```diff
@@ -264,3 +266,21 @@ def test_reports_delete_saved_report(appliance, request):
     view.configuration.item_select(
         item='Delete selected Saved Reports', handle_alert=True)
     assert not report.exists
+
+
+def test_reports_create_schedule_for_custom_report(
```
digitronik: Required:
Better to called it `test_custom_schedule_report_crud` what you say? and please delete what you created.

cfme/tests/intelligence/reports/test_crud.py
valaparthvi: Yes. The command fails the first time. Locally it gives me a `NoSuchElementException` and on PRT it gives `MoveTargetOutOfBoundsException`. I don't understand why, but the command works fine when I run it again.

cfme/tests/intelligence/reports/test_crud.py
valaparthvi: I am getting the error when the test runs on PRT, but locally it gives a `NoSuchElementException`.

cfme/tests/intelligence/reports/test_crud.py
valaparthvi: Okay, we can name it that. I delete report at the end of the test, which automatically deletes the schedule.

cfme/tests/intelligence/reports/test_crud.py
sshveta: Okay please write a comment about it then .
It's only through code that this command fails for the first time right ?
If in UI also this happens then it might be a BZ , something to investigate .

cfme/tests/intelligence/reports/test_crud.py
```diff
@@ -264,3 +266,26 @@ def test_reports_delete_saved_report(appliance, request):
...
+    custom_report = appliance.collections.reports.create(**custom_report_values)
+    schedule_data["filter"] = (
+        "My Company (All Groups)",
+        "Custom",
+        custom_report.menu_name,
+    )
+    try:
+        custom_report_schedule = appliance.collections.schedules.create(**schedule_data)
+    except (MoveTargetOutOfBoundsException, NoSuchElementException):
+        # Temporary workaround - running the `create` method first time raises exceptions,
```
jawatts: @valaparthvi I don't think we should be needing to try this twice, it should work the first time. Please spend some more time debugging what is happening. This may be a BZ.

Moving the PR back to WIPTEST

dajoRH:
Lint report for commit 2b56948a4385fb5093855d57fb7c7cd16c151651:

`cfme/intelligence/reports/schedules.py`:
- :bangbang: Line 192:9: **E265** *block comment should start with '# '*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/intelligence/reports/test_crud.py
```diff
@@ -264,3 +265,21 @@ def test_reports_delete_saved_report(appliance, request):
...
+    custom_report = appliance.collections.reports.create(**custom_report_values)
+    schedule_data["filter"] = (
+        "My Company (All Groups)",
+        "Custom",
+        custom_report.menu_name,
+    )
+    custom_report_schedule = appliance.collections.schedules.create(**schedule_data)
+    assert custom_report_schedule.exists
+    custom_report_schedule.delete(cancel=False)
+    request.addfinalizer(custom_report.delete)
```
digitronik: Required:
Move this `Finalizer` after `L276` it will increase readability. I think, creating `custom_report` and deletion is a part of a `setup` for this test. If possible, I would suggest you to create separate `fixture` which will create `custom_report` for you and delete at the time of `tear down`.

valaparthvi: @digitronik I've added a separate fixture `get_custom_report` for the test.

PR #7909, opened 2018-10-01
[RFR][NOTEST] Adding 5.10 forced_stream to test_operations_vm_on

jan-zmeskal:
Problem described in https://bugzilla.redhat.com/show_bug.cgi?id=1571254 occurs in 5.10 as well.


cfme/tests/intelligence/reports/test_canned_corresponds.py
```diff
@@ -120,8 +120,7 @@ def test_cluster_relationships(appliance, soft_assert):
 
 
 @pytest.mark.rhv2
-@pytest.mark.meta(blockers=[BZ(1504010, forced_streams=['5.7', '5.8', 'upstream']),
-                            BZ(1571254, forced_streams=['5.8', '5.9'])])
+@pytest.mark.meta(blockers=[BZ(1571254, forced_streams=['5.8', '5.9', '5.10'])])
```
sshveta: Required : We can remove 5.8 .
@pytest.mark.meta(blockers=[BZ(1571254, forced_streams=['5.9', '5.10'])])

jan-zmeskal: @sshveta Actually, I am not a fan of that. I remember the last time we thought there wouldn't be any further 5.8 builds...

jan-zmeskal: @sshveta I think it's worth keeping 5.8 for now in case we need it.

sshveta: @jan-zmeskal : There is a separate branch 5.8-breakpoint-patched for 5.8 only.

jan-zmeskal: @sshveta True that

PR #7913, opened 2018-10-02
[WIPTEST]  Custom Button: automate test for cloud objects

digitronik:
Purpose or Intent
=================

{{pytest:  cfme/tests/automate/custom_button/test_cloud_objects.py::test_custom_button_automate -sqvv}}


digitronik:
https://github.com/ManageIQ/integration_tests/pull/8004 need to merge for `clear filter`

cfme/tests/automate/custom_button/test_cloud_objects.py
```diff
@@ -236,3 +238,94 @@ def test_custom_button_dialog(appliance, dialog, request, setup_objs, button_gro
...
+
+                paginator.check_all()
+            else:
+                entity_count = 1
+
+            # Clear the automation log
+            clean = appliance.ssh_client.run_command(
+                'echo -n "" > ' "/var/www/miq/vmdb/log/automation.log"
+            )
+            assert clean.success
```
sshveta: We can assert for the command directly 

assert appliance.ssh_client.run_command( 'echo -n "" > ' "/var/www/miq/vmdb/log/automation.log")

cfme/tests/automate/custom_button/test_cloud_objects.py
```diff
@@ -236,3 +238,93 @@ def test_custom_button_dialog(appliance, dialog, request, setup_objs, button_gro
...
+
+
+@pytest.mark.meta(
+    blockers=[BZ(1628224, forced_streams=["5.10"], unblock=lambda submit: submit != "Submit all")]
+)
+@pytest.mark.uncollectif(
+    lambda appliance, button_group: not bool([obj for obj in OBJ_TYPE_59 if obj in button_group])
+    and appliance.version < "5.10"
+)
+@pytest.mark.parametrize("submit", SUBMIT, ids=["_".join(item.split()) for item in SUBMIT])
```
mshriver: Please use `item.replace(' ', '_')` here instead of a split+join

cfme/tests/automate/custom_button/test_cloud_objects.py
```diff
@@ -236,3 +238,93 @@ def test_custom_button_dialog(appliance, dialog, request, setup_objs, button_gro
...
+            # Submit all: single request for all entity execution
+            # One by one: separate requests for all entity execution
+            expected_count = 1 if submit == "Submit all" else entity_count
+            try:
+                wait_for(
+                    log_request_check,
+                    [appliance, expected_count],
+                    timeout=600,
+                    message="Check for expected request count",
+                    delay=20,
```
mshriver: Please lower the delay for this check, why wait so long between?

digitronik: @mshriver we need more `delay` as its doing `ssh` and `grep` for `request count` in `automation.log`. support, `submit` is  `one-by-one` and total 20 objects available then it will fire `20` requests in `automation.log` one after another. As per your comment, I reduced to 10s but continues `ssh` and `grep automation.log` not good so I  was added that delay.

PR #7889, opened 2018-09-27
[1LP][RFR] Custom Button:  Import & Export RFE automate

digitronik:
Purpose or Intent
=================
Adding test for RFE on 5.10 for import and export fuctionality of custom button
https://bugzilla.redhat.com/show_bug.cgi?id=1164306

{{pytest: cfme/tests/automate/custom_button/test_import_export.py -v}}

PR #7914, opened 2018-10-02
[1LP][RFR] Custom Button: Container object initial display test

digitronik:
Purpose or Intent
=================
- Introducing container custom button objects  display test
1. CONTAINER_IMAGE
2. CONTAINER_NODE
3. CONTAINER_POD
4. CONTAINER_PROJECT
5. CONTAINER_TEMPLATE
6. CONTAINER_VOLUME

{{pytest: cfme/tests/automate/custom_button/test_container_objects.py -v --use-provider ocp-39-hawk}}


cfme/tests/automate/custom_button/test_container_objects.py
```diff
@@ -0,0 +1,106 @@
...
+
+DISPLAY_NAV = {
+    "Single entity": ["Details"],
+    "List": ["All"],
+    "Single and list": ["All", "Details"],
+}
+
+
+@pytest.fixture(
+    params=CONTAINER_OBJECTS, ids=[obj.capitalize() for obj in CONTAINER_OBJECTS], scope="module"
```
jaryn: It seems like capitalize is not needed any more.

cfme/tests/automate/custom_button/test_container_objects.py
```diff
@@ -0,0 +1,106 @@
...
+    elif obj_type == "CONTAINER_VOLUME":
+        obj = appliance.collections.container_volumes.all()[0]
+    else:
+        logger.error("No object collected for custom button object type '{}'".format(obj_type))
+    return obj
+
+
+@pytest.mark.ignore_stream('5.9')
+@pytest.mark.parametrize(
+    "display", DISPLAY_NAV.keys(), ids=["_".join(item.split()) for item in DISPLAY_NAV.keys()]
```
jaryn: I would say that use of `replace()` is more straightforward:

```
In [1]: " sf df".replace(" ", "_")
Out[1]: '_sf_df'
```

cfme/tests/automate/custom_button/test_container_objects.py
digitronik: @jaryn we won't take ids in full capital letters. So its required

cfme/tests/automate/custom_button/test_container_objects.py
digitronik: @jaryn I think its more pythonic than replacing blank space with `_`

cfme/tests/automate/custom_button/test_container_objects.py
```diff
@@ -0,0 +1,106 @@
...
+from cfme.utils.log import logger
+
+
+pytestmark = [
+    pytest.mark.tier(2),
+    pytest.mark.usefixtures("setup_provider"),
+    pytest.mark.provider([ContainersProvider], selector=ONE_PER_TYPE),
+]
+
+CONTAINER_OBJECTS = [
```
jawatts: **Required** This should a list of the collection classes, ie:
```
CONTAINER_COLLECTIONS = [
"container_images",
"container_nodes",
"container_pods",
"container_projects",
"container_templates",
"container_volumes",
]``` 

If we do this, we can use it to call the collection classes directly.

File: cfme/tests/automate/custom_button/test_container_objects.py [hunk 4]
```diff
@@ -0,0 +1,106 @@
...
+    yield button_gp, request.param
+    button_gp.delete_if_exists()
+
+
+@pytest.fixture()
+def setup_obj(appliance, button_group):
+    """ Setup object for specific custom button object type."""
+    obj_type = button_group[1]
+
+    if obj_type == "CONTAINER_IMAGE":
```
jawatts: If we parameterize on the collection classes, we then can replace this large if/elif block with ```obj = getattr(appliance.collections, data_set.collection_obj).all())[0]```

Also there are instances where, specifically for openshift pods, where a pod be spun up to do a scan and then be destroyed (This happens when SSA is ran). This will cause you tests to fail because it will not be able to navigate to the destination. For the container tests, I also added an additional steps to check if the object exists.

digitronik: :P Our  views matches at some points. Yup, I was also thinking same at the  beginning of custom button automation but,

When I was planning automation structure for a custom button with parametrization, we had 28 custom button objects out  of that 12 from 5.9 and other included in 5.10. I thought about this idea at the initial level but there are some problems like,
1. Two ways to for parametrization
    a. go with `collection attributes` name
    b. go with actual `object TYPE` name

If we select to go with an attribute:
- I have to change Object type initialization in `ButtonGroup` collection class which will affect other test means I need to change a LOT.
- I have to add workaround with like `TYPE` name `CONTAINER_IMAGE` and collection name is `container_images`. As we have to provider type name while creating `button` and `button group`.
- Some objects are there like `VM_INSTANCE` where the collection initialization like 
`appliance.provider_based_collection(provider)`
- Sometimes I won't need to initialize a collection like for `provider` custom button object, `provider` itself acts as object.
- Some objects are like where I won't have the support of `all` method 
```
appliance.collections.tenants.get_root_tenant()
appliance.collections.users.instantiate(name="Administrator")
```
-  Some object will be there in which we need to create an object like `Cloud Volumes` or `Cloud Tenent`. 

I considered all of those possible conditions and I decided to go with `if` `else` block to take action for initializing an object either with `all` method, or normal initialization of the hardcoded object or create an object.

I know big `if else` block is not good; but it's not bad as well where we are trying to maintain similarity between all files means cloud, infra, generic, service, container object types. If someone understand one file, he can understand all of them related to custom button automation. Let me know your views.  

Yes, we can go `getattr` for this file with a small workaround in string formatting or change entries in `ButtonGroup` class.

digitronik: @jawatts our views matches at some points. :P Yup, I was also thinking same at the  beginning of custom button automation but,

When I was planning automation structure for a custom button with parametrization, we had 28 custom button objects out  of that 12 from 5.9 and other included in 5.10. I thought about this idea at the initial level but there are some problems like,
1. Two ways to for parametrization
    a. go with `collection attributes` name
    b. go with actual `object TYPE` name

If we select to go with an attribute:
- I have to change Object type initialization in `ButtonGroup` collection class which will affect other test means I need to change a LOT.
- I have to add workaround with like `TYPE` name `CONTAINER_IMAGE` and collection name is `container_images`. As we have to provider type name while creating `button` and `button group`.
- Some objects are there like `VM_INSTANCE` where the collection initialization like 
`appliance.provider_based_collection(provider)`
- Sometimes I won't need to initialize a collection like for `provider` custom button object, `provider` itself acts as object.
- Some objects are like where I won't have the support of `all` method 
```
appliance.collections.tenants.get_root_tenant()
appliance.collections.users.instantiate(name="Administrator")
```
-  Some object will be there in which we need to create an object like `Cloud Volumes` or `Cloud Tenent`. 

I considered all of those possible conditions and I decided to go with `if` `else` block to take action for initializing an object either with `all` method, or normal initialization of the hardcoded object or create an object.

I know big `if else` block is not good; but it's not bad as well where we are trying to maintain similarity between all files means cloud, infra, generic, service, container object types. If someone understand one file, he can understand all of them related to custom button automation. Let me know your views.  

Yes, we can go `getattr` for this file with a small workaround in string formatting or change entries in `ButtonGroup` class.

jawatts: @digitronik Alright lets keep it for now but I think we should drop the if/elif block and do some string manipulation then. Change it to lower, make it plural, then use `get_attr`. I would still recommend making the `exists` call as well.

digitronik: @jawatts Agreed, If no other custom button object added by dev. No object manipulation needed then definitely I will shift it to `get_attr`. I know its not possible with other but for the container, I will add.

digitronik: @jawatts can please check it. I think now exist call not needed. Exception will handle it.

digitronik:
Need #7889 to pass PRT

cfme/tests/automate/custom_button/test_container_objects.py
jawatts: @digitronik You will still need it. It is a raise condition I have been fighting. The `all()` method will still return an object, but by the time you check it in the UI, it is gone.

jawatts: @digitronik I thought we were going to use `get_attr` as well on the manipulated strings but keep `CONTAINER_OBJECTS` the same

jawatts:
5.10 is failing because of https://bugzilla.redhat.com/show_bug.cgi?id=1640304

Found this bug last week.

cfme/tests/automate/custom_button/test_container_objects.py
```diff
@@ -0,0 +1,107 @@
...
+@pytest.fixture()
+def setup_obj(appliance, provider, button_group):
+    """ Setup object for specific custom button object type."""
+    obj_type = button_group[1]
+
+    try:
+        if obj_type == "PROVIDER":
+            obj = provider
+        else:
+            obj = getattr(appliance.collections, obj_type.lower()).all()[0]
```
jawatts: :+1:

cfme/tests/automate/custom_button/test_container_objects.py
```diff
@@ -0,0 +1,107 @@
...
+    except IndexError:
+        pytest.skip("Object not found for {obj} type".format(obj=obj_type))
+
+    if not obj.exists:
```
jawatts: :+1:

cfme/tests/automate/custom_button/test_container_objects.py
mshriver: six in one, half dozen in the other.

PR #7915, opened 2018-10-02
[WIP] BZ1635178 workaround

jan-zmeskal:
TBD


cfme/base/__init__.py
```diff
@@ -130,7 +131,8 @@ def get_master(self):
             name = server.name
         except AttributeError:
             logger.error('The EVM has no name, setting it to EVM')
-            if self.appliance.version == LATEST or self.appliance.is_pod:
+            if self.appliance.version == LATEST or self.appliance.is_pod \
```
izapolsk: please use () instead of \

PR #7916, opened 2018-10-03
[1LP][RFR] new test_250_vmware_hosts_loading

lina-nikiforova:
{{ pytest: -v cfme/tests/infrastructure/test_host.py::test_250_vmware_hosts_loading }}

### DIFF
--- a/cfme/tests/infrastructure/test_host.py
+++ b/cfme/tests/infrastructure/test_host.py
@@ -4,7 +4,7 @@ import random
 import socket
 
 from cfme.base.credential import Credential
-from cfme.common.host_views import HostsEditView, HostsView
+from cfme.common.host_views import HostsEditView
 from cfme.common.provider_views import ProviderNodesView
 from cfme.infrastructure.provider import InfraProvider
 from cfme.infrastructure.provider.rhevm import RHEVMProvider
@@ -20,6 +20,8 @@ pytestmark = [
     pytest.mark.tier(3),
     pytest.mark.provider([InfraProvider], required_fields=['hosts'], scope='module'),
 ]
+
+VIEWS = ('Grid View', 'Tile View', 'List View')
 
 
 @pytest.fixture(scope='module')
@@ -182,7 +184,8 @@ def test_tag_host_after_provider_delete(provider, appliance, setup_provider, req
     request.addfinalizer(lambda: host.remove_tag(tag))
 
 
-def test_250_vmware_hosts_loading(appliance, create_250_hosts):
+@pytest.mark.parametrize('view_type', VIEWS)
+def test_250_vmware_hosts_loading(appliance, create_250_hosts, view_type):
     """
     Test to automate BZ1580569
     """
@@ -194,8 +197,8 @@ def test_250_vmware_hosts_loading(appliance, create_250_hosts):
     )
     assert rails_console.success
 
-    # Check it loads in the UI fast
-    host_collection = appliance.collections.hosts
-    view = host_collection.appliance.browser.create_view(HostsView)
-    load_view = wait_for(lambda: view.entities.paginator.set_items_per_page(1000))
-    assert load_view.duration < 60
+    # Check it loads fast in the UI
+    view = navigate_to(appliance.collections.hosts, 'All')
+    view.entities.paginator.set_items_per_page(1000)
+    view.toolbar.view_selector.select(view_type)
+    wait_for(view.entities.get_first_entity, timeout=60, message='Wait for the view')
--- a/cfme/tests/infrastructure/test_host.py
+++ b/cfme/tests/infrastructure/test_host.py
@@ -4,7 +4,7 @@ import random
 import socket
 
 from cfme.base.credential import Credential
-from cfme.common.host_views import HostsEditView
+from cfme.common.host_views import HostsEditView, HostsView
 from cfme.common.provider_views import ProviderNodesView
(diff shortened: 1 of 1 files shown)


lina-nikiforova:
Since PRT is failing on 510z, posting local results ran on CFME 5.10.0.17:

```
Appliance's streams: [5.10, downstream]
========================================================================================================================= test session starts ==========================================================================================================================
platform linux2 -- Python 2.7.5, pytest-3.4.1, py-1.5.3, pluggy-0.6.0 -- /home/anikifor/cfme/bin/python
cachedir: .pytest_cache
rootdir: /home/anikifor/projects/integration_tests, inifile:
plugins: wait-for-1.0.9, pycharm-0.5.0, manageiq-integration-tests-17.36.0rc1.dev22+g704d904.d20181002
collected 1 item                                                                                                                                                                                                                                                       
Uncollection Stats:
 manual: 0
 uncollectif: 0
 1 tests left after all uncollections

cfme/tests/infrastructure/test_host.py::test_250_vmware_hosts_loading 
PASSED

====================================================================================================================== 1 passed in 182.66 seconds ======================================================================================================================

```

cfme/tests/infrastructure/test_host.py
```diff
@@ -46,6 +46,24 @@ def host_ips(provider):
...
+    assert script_downloaded.success, script_downloaded.output
+
+    create_250_hosts = appliance.ssh_client.run_command(
+        "cd /var/www/miq/vmdb && bin/rails r bz_1580569_db_replication_script.rb")
+    assert create_250_hosts.success
+    yield
+
+    appliance.ssh_client.run_command("rm -f /var/www/miq/vmdb/bz_1580569_db_replication_script.rb")
+    appliance.ssh_client.run_rails_console("[Host].each(&:delete_all)")
+    appliance.delete_all_providers()
```
jawatts: **Question** Why delete all the providers here? Is it because you removed all the hosts being tracked by CFME, even the ones outside of the 250?

cfme/tests/infrastructure/test_host.py
```diff
@@ -162,3 +180,22 @@ def test_tag_host_after_provider_delete(provider, appliance, setup_provider, req
...
+    # due to a poorly optimized database query
+    rails_console = appliance.ssh_client.run_rails_console(
+        "MiqReport.load_from_view_options(Host, User.where(:userid => 'admin').first)"
+        ".paged_view_search", timeout=60
+    )
+    assert rails_console.success
+
+    # Check it loads in the UI fast
+    host_collection = appliance.collections.hosts
+    view = host_collection.appliance.browser.create_view(HostsView)
```
jawatts: **Optional** I think you can combine 198 and 199 since `host_collection` is not used anywhere else

cfme/tests/infrastructure/test_host.py
```diff
@@ -162,3 +180,22 @@ def test_tag_host_after_provider_delete(provider, appliance, setup_provider, req
     provider.wait_for_delete()
     tag = host.add_tag()
     request.addfinalizer(lambda: host.remove_tag(tag))
+
+
+def test_250_vmware_hosts_loading(appliance, create_250_hosts):
```
jawatts: **Question** Does this test only apply to CFME 5.10? Or has it been backported to 5.9z stream?

cfme/tests/infrastructure/test_host.py
mshriver: Alternatively, why use create_view() here and not a call to `navigate_to()`? I don't see where the navigation would occur, explicitly, anywhere otherwise (in the fixture)

cfme/tests/infrastructure/test_host.py
```diff
@@ -162,3 +180,22 @@ def test_tag_host_after_provider_delete(provider, appliance, setup_provider, req
...
+    rails_console = appliance.ssh_client.run_rails_console(
+        "MiqReport.load_from_view_options(Host, User.where(:userid => 'admin').first)"
+        ".paged_view_search", timeout=60
+    )
+    assert rails_console.success
+
+    # Check it loads in the UI fast
+    host_collection = appliance.collections.hosts
+    view = host_collection.appliance.browser.create_view(HostsView)
+    load_view = wait_for(lambda: view.entities.paginator.set_items_per_page(1000))
```
mshriver: I'm curious about the behavior that you expect here, `set_items_per_page` is implemented with `ReportDataControllerMixin`, which uses javascript to make this page change.  I imagine that JS command to change the per-page count is returning before the page is actually done rendering the entities.

I think you should consider making a query against `entities.get_first_entity` on the page in order to check that the entities are actually rendered, or just a wait that `entities.table.is_displayed` in a list view.

Consider parametrizing the test against tile, quad, and list views to cover all of those cases of capturing how long it took for the entities to be visible.

cfme/tests/infrastructure/test_host.py
lina-nikiforova: Yes. I want to do clean up at the end, the script for the test generates 6 providers like ems_01 and so on, it's easy to do by this method

cfme/tests/infrastructure/test_host.py
lina-nikiforova: thanks, Mike!

cfme/tests/infrastructure/test_host.py
lina-nikiforova: This test applies to both streams

PR #7917, opened 2018-10-03
[1LP][RFR] Modify report download method for 5.10

valaparthvi:
This PR modifies `download` method for Report. If the appliance version is greater than 5.10, it looks for `Print or export as PDF` option while downloading a report.

{{pytest: cfme/tests/intelligence/test_download_report.py --use-template-cache -sqvvv}}

### DIFF
--- a/cfme/tests/intelligence/test_download_report.py
+++ b/cfme/tests/intelligence/test_download_report.py
@@ -1,51 +1,21 @@
 # -*- coding: utf-8 -*-
 import pytest
-import os
-import shutil
-from cfme.utils.wait import wait_for
-from cfme.utils import browser
-
-
-def clean_temp_directory():
-    """ Clean the temporary directory.
-
-    """
-    for root, dirs, files in os.walk(browser.firefox_profile_tmpdir):
-        for f in files:
-            os.unlink(os.path.join(root, f))
-        for d in dirs:
-            shutil.rmtree(os.path.join(root, d))
-
-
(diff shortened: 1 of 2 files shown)


digitronik:
@pvala please add test. Is there no test for download? Also add local results for 5.10

valaparthvi:
@digitronik Changed the old `test_download_report_firefox` to `test_download_report` which will work for all browser, removed all the other configurations specific to old firefox test and the test now just downloads the report.

cfme/tests/intelligence/test_download_report.py
```diff
@@ -1,51 +1,21 @@
...
-@pytest.mark.parametrize("filetype", ["txt", "csv"])
-def test_download_report_firefox(needs_firefox, infra_provider, report, filetype):
-    """ Download the report as a file and check whether it was downloaded. """
-    extension = "." + filetype
-    clean_temp_directory()
+@pytest.mark.parametrize("filetype", ["txt", "csv", "pdf"])
+def test_download_report(infra_provider, report, filetype):
+    """
+    Download the report as a file .
+    """
```
digitronik: Inline please

digitronik:
@valaparthvi please debug 5.9 failures at downloading `file type`  and tear down deletion.

valaparthvi:
@digitronik I've tested this locally and it works fine there, but I don't understand why the PRT test is failing. What do you suggest I do?

PR #7919, opened 2018-10-03
[1LP][RFR] Move widgetastic exception import

mshriver:
Imported from the wrong module, just moving it.


cfme/configure/configuration/diagnostics_settings.py
```diff
@@ -3,7 +3,8 @@
 from navmazing import NavigateToAttribute, NavigateToSibling
 from widgetastic_patternfly import Input, BootstrapSelect, Button, Dropdown
 from widgetastic_manageiq import SummaryFormItem, Table
-from widgetastic.widget import View, RowNotFound, Text
+from widgetastic.exceptions import RowNotFound
```
izapolsk: :+1:

PR #7923, opened 2018-10-04
[1LP][WIP] Refactor login tests

apagac:
Added 'new_user_create' and 'new_user_instantiate' functions to basic login testing.

These are going to be helpful for new automation that will be added to this file.

{{ pytest: -v cfme/tests/test_login.py -k 'test_update_password or test_bad_password' }}

### DIFF
--- a/cfme/tests/test_login.py
+++ b/cfme/tests/test_login.py
@@ -12,6 +12,35 @@ from cfme.utils.version import LOWEST, UPSTREAM, VersionPicker
 pytestmark = pytest.mark.usefixtures('browser')
 
 
+def new_cred(username, password):
+    username = username or conf.credentials['default']['username']
+    password = password or conf.credentials['default']['password']
+    cred = Credential(principal=username, secret=password)
+    return cred
+
+
+def new_user_instantiate(appliance, username=None, password=None, name=None):
+    cred = new_cred(username=username, password=password)
+    name = name or 'Administrator'
+    user = appliance.collections.users.instantiate(credential=cred, name=name)
+    return user
+
+
+def new_user_create(appliance, request, username=None, password=None, group=None, name=None):
+    cred = new_cred(username=username, password=password)
+    user_group = appliance.collections.groups.instantiate(
+        description=group or "EvmGroup-vm_user"
+    )
+    name = name or username
+    user = appliance.collections.users.create(
+        name=name,
+        credential=cred,
+        groups=user_group
+    )
+    request.addfinalizer(user.delete)
+    return user
+
+
 @pytest.mark.rhel_testing
 @test_requirements.drift
 @pytest.mark.tier(1)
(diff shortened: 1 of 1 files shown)


cfme/tests/test_login.py
```diff
@@ -12,6 +12,35 @@
...
+def new_cred(username, password):
+    username = username or conf.credentials['default']['username']
+    password = password or conf.credentials['default']['password']
+    cred = Credential(principal=username, secret=password)
+    return cred
+
+
+def new_user_instantiate(appliance, username=None, password=None, name=None):
+    cred = new_cred(username=username, password=password)
+    name = name or 'Administrator'
```
mshriver: Please use fauxfactory to get a randomized name so that subsequent test runs don't fail just because the appliance wasn't completely cleaned up.

cfme/tests/test_login.py
```diff
@@ -12,6 +12,35 @@
...
+    user = appliance.collections.users.instantiate(credential=cred, name=name)
+    return user
+
+
+def new_user_create(appliance, request, username=None, password=None, group=None, name=None):
+    cred = new_cred(username=username, password=password)
+    user_group = appliance.collections.groups.instantiate(
+        description=group or "EvmGroup-vm_user"
+    )
+    name = name or username
```
mshriver: both of these could be None, you handle the default username when creating the `Credential` instance though, so you can use it here to default the name:
` name = name or cred.principal`

cfme/tests/test_login.py
```diff
@@ -71,13 +100,13 @@ def test_update_password(context, request, appliance):
 
     # First, create a temporary new user
     username = 'user_temp_{}'.format(fauxfactory.gen_alphanumeric(4).lower())
-    new_creds = Credential(principal=username, secret='redhat')
-    user_group = appliance.collections.groups.instantiate(description="EvmGroup-vm_user")
-    user = appliance.collections.users.create(
-        name=username,
-        credential=new_creds,
-        groups=user_group
+    user = new_user_create(
```
mshriver: This is changing test behavior by not passing secret, which is explicitly set in the old test. Reasoning for the change?

cfme/tests/test_login.py
```diff
@@ -12,6 +12,35 @@
...
+
+
+def new_user_instantiate(appliance, username=None, password=None, name=None):
+    cred = new_cred(username=username, password=password)
+    name = name or 'Administrator'
+    user = appliance.collections.users.instantiate(credential=cred, name=name)
+    return user
+
+
+def new_user_create(appliance, request, username=None, password=None, group=None, name=None):
```
psav: This is acting like a fixture, by adding the finalizer, personally I think this should be a fixture this actually then makes the test focus solely on updating the password and not on creating the user.

dajoRH:
*CFME QE Bot*

PR #7924, opened 2018-10-04
[WIP] Changing dockerbot default browser to chrome

mshriver:
This is for MIQ Gaprindashvilli+ testing. Older FF versions hit a host of automation issues, which are being debugged. For now switch dockerbot (and thus PRT) to use chrome by default.


mshriver:
I am withdrawing this PR, opting to modify the docker config on PRT instead of defaulting dockerbot itself to chrome.

PR #7926, opened 2018-10-04
[1LP][RFR] Fix CMQE Custom Attribute Test Case

jawatts:
Fix custom cttribute test case to match the implantation of the GUI. 

{{ pytest: cfme/tests/containers/test_static_custom_attributes.py --use-provider ocp-36-hawk }}


digitronik:
LGTM :+1:  still problem with PRT for 5.10

PR #7930, opened 2018-10-04
[1LP][RFR]ProviderClusters.is_displayed had a typo and always returned False

izapolsk:
{{pytest: cfme/tests/infrastructure/test_timelines.py --use-provider=vsphere6-nested}}


izapolsk:
Tests failed but issue with ProviderClusters disappeared. So, I think this change can be merged.

PR #7943, opened 2018-10-05
[1LP][RFR] Update Taggable.add_tag nav error handling

mshriver:
Raise common exception when nav to Details fails for a Taggable entity

Right now we're hitting a `TimedOutError` on Taggable entities that are returning `False` for `is_displayed` on their `Details` view, I think its more fit to raise an ItemNotFound error here consistently. Happens to handle case where Taggable class doesn't have details nav defined either.

### DIFF
--- a/cfme/common/__init__.py
+++ b/cfme/common/__init__.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 import six
 import random
-from navmazing import NavigateToSibling
+from navmazing import NavigateToSibling, NavigationDestinationNotFound
 from widgetastic.exceptions import NoSuchElementException, RowNotFound
 from widgetastic_patternfly import (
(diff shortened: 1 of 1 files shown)


cfme/common/__init__.py
```diff
@@ -397,8 +398,16 @@ def get_tags(self, tenant="My Company Tags"):
...
         Returns: :py:class:`list` List of Tag objects
+
+        Raises:
+            ItemNotFound: when nav destination DNE, or times out on wait (is_displayed was false)
         """
-        view = navigate_to(self, 'Details')
+        try:
+            view = navigate_to(self, 'Details')
+        except (NavigationDestinationNotFound, DestinationNotFound):
+            raise ItemNotFound('Details page does not exit for: {}'.format(self))
```
kedark3: typo : exit <> exists

PR #7945, opened 2018-10-06
[RFR] Remove wait_for_evm_service and fix call to evmserverd.running

mshriver:


cfme/utils/appliance/__init__.py
```diff
@@ -1478,7 +1478,7 @@ def restart_evm_rude(self, log_callback=None):
             log_callback('Waiting for evm service to stop')
             try:
                 wait_for(
-                    self.evmserverd.running, num_sec=120, fail_condition=True, delay=5,
+                    lambda: self.evmserverd.running, num_sec=120, fail_condition=True, delay=5,
```
mshriver: This is the bug fix, rest of PR is `removing wait_for_evm_service`

PR #7954, opened 2018-10-08
[RFR] Fix: is_displayed for template details

digitronik:
Purpose or Intent
=================
{{pytest: cfme/tests/automate/custom_button/test_infra_objects.py::test_custom_button_display -v}}

5.10  two test failing due to host bug 1636120

### DIFF
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -265,10 +265,13 @@ class InfraVmDetailsView(InfraVmView):
 
     @property
     def is_displayed(self):
-        expected_titles = [
-            'VM and Instance "{}"'.format(self.context["object"].name),
-            'VM Template and Image "{}"'.format(self.context["object"].name),
-        ]
+        if isinstance(self.context['object'], InfraVm):
+            expected_title = 'VM and Instance "{}"'.format(self.context["object"].name)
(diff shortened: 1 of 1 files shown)


cfme/infrastructure/virtual_machines.py
```diff
@@ -265,7 +265,10 @@ class TemplatesToolbar(InfraGenericDetailsToolbar):
 
     @property
     def is_displayed(self):
-        expected_name = self.context['object'].name
+        expected_titles = [
+            'VM and Instance "{}"'.format(self.context["object"].name),
```
mshriver: Discussed offline a bit - this is a crossover of VM/Template that we need to split up a bit to keep the navigation and view detection accurate and reliable.

We either need two views (separate class for template details view), or is_displayed need to detect what type of contextual object is there, and set the title respective to the class.

That will drive whether or not the navigation destination needs to be changed. If we have a separate view for Template, then we'll need either two navigation destination classes or a `VIEW` property that detects the object type and returns the correct view matching the class.

digitronik: @mshriver I selected the first option.

mshriver:
PRT isn't complete, but its showing passing VM/Template tests for test_custom_button_display parametrization.  Some failures from HostDetailsView, not in scope for PR.

PR #7955, opened 2018-10-08
[WIP] Let PXESystemImageEditView inherit is_displayed from parent class

jan-zmeskal:
Ignore for now

### DIFF
--- a/cfme/infrastructure/pxe.py
+++ b/cfme/infrastructure/pxe.py
@@ -716,10 +716,6 @@ class PXESystemImageDeatilsView(PXEMainView):
 
 class PXESystemImageEditView(PXEMainView):
 
-    @property
-    def is_displayed(self):
-        return False
-
     image_type = BootstrapSelect(id='image_typ')
     save = Button('Save')
     reset = Button('Reset')


mshriver:
I think I know what you're trying to fix, but I don't think this is the right way to go about it in this case - the `PXEMainView.is_displayed` only looks at the vertical navigation selection.

This is a result of wait_for_view being the default behavior of navigate_to. Since this is an edit page, and its a fine practice to default to never matching the view - you'll want to update any calls to navigate to this view to pass `wait_for_view=0`.

We may be implementing a pattern to have such `is_displayed` methods raise `NotImplemented`, so if you'd like to wait a bit that may be a better fix than updating the nav calls.

jan-zmeskal:
Hi @mshriver, you're right. Already yesterday I realized that `wait_for_view=0` is much better solution. I didn't have time to change this PR so I at least let it hang there as a reminder for me. I don't understand the last part of your message though. Are you preparing a change that will fix this more globally. Or should I still create a PR for `wait_for_view=0`?

mshriver:
@jan-zmeskal the last comment I made was about an idea I had for how to handle these views where we explicitly do not want to implement is_displayed, but don't want to force callers to account for the fact that it returns False.

@izapolsk ran with the idea and opened up #7958

jan-zmeskal:
@mshriver I see. If I understand it correctly, after izapolsk's PR is merged, we'll have to `raise NotImplementedError` in the views that now return `False` in `is_displayed`, right?

mshriver:
That's correct, but we're still debating the merits of that pattern. Please raise any concerns or ideas you have about it in #7958.

In the meantime, passing `wait_for_view=0` or implementing the is_displayed method are both viable fixes for nav.

jan-zmeskal:
@mshriver I am in no rush as far as fixing ISO provision test case goes, since the feature itself does not work right now. I just want to ask: Could we create some GH issue where we could track progress on this?

dajoRH:
*CFME QE Bot*

PR #7956, opened 2018-10-08
[1LP][RFR] Fix is_displayed for create and edit dialog

niyazRedhat:
{{ pytest: cfme/tests/automate/test_service_dialog.py::test_reorder_elements -v}}
{{ pytest: cfme/tests/automate/test_service_dialog.py::test_dropdownlist_dialog_element -v}}

### DIFF
--- a/cfme/automate/dialogs/__init__.py
+++ b/cfme/automate/dialogs/__init__.py
@@ -28,7 +28,8 @@ class AutomateCustomizationView(BaseLoggedInPage):
 
 
 class DialogForm(AutomateCustomizationView):
-    title = Text('#explorer_title_text')
+    title = Text('//div[@id= "main-content"]//h1')
+    sub_title = Text('//div[@id= "main-content"]//h2')
 
     label = Input(id='name')
     description = Input(id="description")
@@ -42,8 +43,8 @@ class AddDialogView(DialogForm):
     @property
     def is_displayed(self):
         return (
-            self.in_customization and self.service_dialogs.is_opened and
-            self.create_tab.is_displayed
+            self.in_customization and self.title.text == "Automate Customization" and
+            self.sub_title.text == "General" and self.create_tab.is_displayed
         )
 
 
@@ -54,8 +55,8 @@ class EditDialogView(DialogForm):
     @property
     def is_displayed(self):
         return (
-            self.in_customization and self.service_dialogs.is_opened and
-            self.label.text == self.label
+            self.in_customization and self.title.text == "Automate Customization" and
+            self.sub_title.text == "General" and self.label.read() == self.context['object'].label
         )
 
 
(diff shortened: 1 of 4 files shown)


cfme/automate/dialogs/__init__.py
```diff
@@ -54,8 +55,8 @@ class EditDialogView(DialogForm):
     @property
     def is_displayed(self):
         return (
-            self.in_customization and self.service_dialogs.is_opened and
-            self.label.text == self.label
+            self.in_customization and self.title.text == "Automate Customization" and
+            self.sub_title.text == "General" and self.label.text == self.label
```
mshriver: I don't think this is a good check - `self.text.label` won't equal `self.label` - `self.label` is the widget object, if the label (name) cannot be modified, you can validate `self.text.label` against `self.context['object'].label`.

Alternatively, if this is a weak check (edit page), return False here and update any navigation to this view to pass `wait_for_view=0`

cfme/automate/dialogs/__init__.py
```diff
@@ -42,7 +42,7 @@ class AddDialogView(DialogForm):
     @property
     def is_displayed(self):
         return (
-            self.in_customization and self.service_dialogs.is_opened and
```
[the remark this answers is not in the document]
mshriver: Updates here look good!

mshriver:

cfme/automate/dialogs/__init__.py
```diff
@@ -54,8 +55,8 @@ class EditDialogView(DialogForm):
...
-            self.label.text == self.label
+            self.in_customization and self.title.text == "Automate Customization" and
+            self.sub_title.text == "General" and bool(self.label.read())
```
mshriver: What's the purpose of the `bool(self.label.read())` check? just that the label field is populated?

I think you need to check that `self.label.read() == self.context['object'].label` - that the content of the label field matches the Dialog's `label` attribute.  If this can be changed during edit, then this may be a weak check that we're on the correct page. It may be better to just `return False` for this method, or raise NotImplemented per the change coming in #7958

mshriver: Looking at the UI a bit, I don't think we should try to match this edit view via is_displayed, this will not be robust. The label can be modified, navigation to this edit page should always be performed via its prerequisite tree.

niyazRedhat: @mshriver rather than removing all `is_displayed` checks better to go with check for label (`TextInput` box).
1. If we go for add page `TextInput` box read will be blank.
2. If we go for edit then `TextInput` box read will be `old name`.

I think those check is sufficient to check for edit view.

mshriver: @niyazRedhat it looks like you ended up changing this to matching the object label?

I think there's an assumption in the check you were suggesting above - that the page has just been navigated to.

It's possible the browser is on an Add page, that has been partially filled in. If we tried to navigate to Edit, and were only checking that the label field is filled, it would short-circuit navigation and think it was on the right page.

PR #7958, opened 2018-10-08
[1LP][RFR] replace view.is_displayed->False with NotImplementedError

izapolsk:
{{pytest: --long-running --use-provider=complete cfme/tests/infrastructure/test_timelines.py cfme/tests/cloud/test_cloud_timelines.py}}

### DIFF
--- a/cfme/utils/appliance/implementations/ssui.py
+++ b/cfme/utils/appliance/implementations/ssui.py
@@ -154,6 +154,10 @@ class SSUINavigateStep(NavigateStep):
         waited = False
         try:
             here = self.am_i_here()
+        except NotImplementedError:
+            nav_args['wait_for_view'] = 0
+            self.log_message(
+                "is_displayed not implemented for {} view".format(self.VIEW or ""), level="warn")
         except Exception as e:
             self.log_message(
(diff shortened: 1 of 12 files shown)


cfme/utils/appliance/implementations/ssui.py
```diff
@@ -105,6 +105,8 @@ def am_i_here(self):
             return self.view.is_displayed
         except (AttributeError, NoSuchElementException):
             return False
+        except NotImplementedError:
+            return True
```
mshriver: I think this check should actually go into the call to `self.am_i_here()` at L156/158 within `go`.

Then we can set `nav_args[wait_for_view] = 0` and stop it from waiting for the view - instead of defaulting a NotImplemented `is_displayed` to return True.

mshriver:

mshriver:
this is part of resolving #7960

dajoRH:
*CFME QE Bot*

matejart:
I just noticed a small typo in the message of the `NotImplementedError`:

> This view doesn't **implement** unique markers to check in is_displayed

mshriver:
@matejart I pushed a commit, likely while you were reading the PR, with a change to the wording there.

PR #7962, opened 2018-10-09
[1LP][RFR][NOTEST] removed blocker for test_paginator

lina-nikiforova:
removed blocker as BZ1623091 is verified


lina-nikiforova:
@sshveta yes, it will

PR #7963, opened 2018-10-09
[1LP][RFR] Create custom role in quota tests instead of super_admin

rhrazdil:
__Added__ custom role instead of using super_administrator role. 
Motivation behind the change is comment 6 in the following BZ https://bugzilla.redhat.com/show_bug.cgi?id=1636958#c6 .

TL;DR
Quotas should not be applied to Super_administrator role because of possible (and quite real) unpredictable behaviour.

juwatts:
{{ pytest: cfme/tests/infrastructure/test_project_quota.py --use-provider complete --long-running }}

### DIFF
--- a/cfme/tests/infrastructure/test_project_quota.py
+++ b/cfme/tests/infrastructure/test_project_quota.py
@@ -60,10 +60,20 @@ def new_project(appliance):
 
 
 @pytest.fixture(scope='module')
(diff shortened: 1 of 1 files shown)


digitronik:
@rhrazdil tests are deselected on PRT. Might be long-running tag needs. Please check it.

Issue #7960, opened 2018-10-09
Resolve navigation failures due to wait_for_view


matejart:
In [this commit](https://github.com/ManageIQ/integration_tests/commit/27c7ba9f8ee8797812c16774f3111357eda5472d#diff-8f65f83f3de1a967375f00c9f99ed6ccR514) the type of `nav_args["wait_for_view"]` changed from boolean to a numeric value. However, its usage [in condition at line 554 of cfme/utils/appliance/implementations/ui.py](https://github.com/ManageIQ/integration_tests/commit/27c7ba9f8ee8797812c16774f3111357eda5472d#diff-8f65f83f3de1a967375f00c9f99ed6ccR554) remains as a boolean check.

If someone set `wait_for_view` to `0`, they might mean "I want to check, but not wait for the result", while the logic would interpret that as `False` and skip the check altogether.

mshriver:
I'm going to work on Issue2 above, `Message: Could not find an element Locator(by='xpath', locator=u'//div[@id="settings_co_categories"]/table')`

digitronik:
I would like to work on Issue-3. Which related to `LineChart` tooltip. Let me first ask @nachandr.

kedark3:
Issue I see when navigating host:

> raise ItemNotFound('Timed out navigating to details for: {}'.format(self))
> E           ItemNotFound: Timed out navigating to details for: Host(name=u'env-rhv42-03.cfme2.lab.eng.rdu2.redhat.com', provider=RHEVMProvider(endpoints={'default': <cfme.infrastructure.provider.rhevm.RHEVMEndpoint object at 0x7f76f8198c10>}, name='rhv42', key='rhv42', zone='default', start_ip='10.8.197.37', end_ip='10.8.197.37', provider_data=None), hostname=None, ip_address=None, custom_ident=None, host_platform=None, ipmi_address=None, mac_address=None, credentials={}, ipmi_credentials=None, interface_type='lan')
> 

It is happening on the function call `host_obj.get_tags()`

mshriver:
There is a bug in wt.core, introduced with row/colspan support, that is impacting any of our views where a table widget is nested under a Tab - this applies to any tests that use the provisioning form, or tags.

https://github.com/RedHatQE/widgetastic.core/pull/121

izapolsk:
I suppose this issue has already been fixed.

digitronik:
@izapolsk agreed with you we are not facing this issue in results. I vote for `close`

PR #7964, opened 2018-10-09
[1LP][RFR] fix group navigation displayed

digitronik:
Purpose or Intent
=================
{{pytest: cfme/tests/automate/custom_button/test_buttons.py -v}}

### DIFF
--- a/cfme/automate/buttons.py
+++ b/cfme/automate/buttons.py
@@ -133,8 +133,7 @@ class EditButtonView(ButtonFormCommon):
     def is_displayed(self):
         return (
             self.in_customization and
-            # TODO: vvv BUG
-            self.title.text.startswith('Adding a new Button') and
+            self.title.text.startswith('Editing Button') and
             self.buttons.is_dimmed and
             self.buttons.is_opened and
             self.buttons.tree.currently_selected == [
@@ -311,7 +310,7 @@ class ButtonCollection(BaseCollection):
             image = 'Button Image 1'
         else:
             image = 'fa-user'
-        view = navigate_to(self, 'Add', wait_for_view=0)
+        view = navigate_to(self, 'Add')
         view.options.fill({'type': type})
         view.fill({
             'options': {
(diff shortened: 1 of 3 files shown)


mshriver:

cfme/automate/buttons.py
```diff
@@ -437,9 +436,14 @@ class NewButtonGroupView(ButtonGroupFormCommon):
 
     @property
     def is_displayed(self):
+        if self.browser.appliance.version < "5.10":
+            expected_title = "Adding a new Buttons Group"
+        else:
+            expected_title = "Adding a new Button Group"
```
jan-zmeskal: It seems like a place where `VersionPicker` should be used.

cfme/automate/buttons.py
```diff
@@ -453,13 +457,18 @@ class EditButtonGroupView(ButtonGroupFormCommon):
 
     @property
     def is_displayed(self):
+        if self.browser.appliance.version < "5.10":
+            expected_title = "Editing Buttons Group"
+        else:
+            expected_title = "Editing Button Group"
```
jan-zmeskal: The same as above

cfme/automate/buttons.py
digitronik: @jan-zmeskal personally I avoid using `version pick` in a view definition; where I can easily access version attribute of an appliance.

cfme/automate/buttons.py
```diff
@@ -573,6 +582,7 @@ class ButtonGroupCollection(BaseCollection):
     VM_INSTANCE = "VM and Instance"
 
     def instantiate(self, text, hover, type, image=None, display=None, icon_color=None):
+        self.type = type
```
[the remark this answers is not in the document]
digitronik: ahhhh; This is a bad implementation but yes, to provide a type to collection class; we have to assign it like this.
you can check `create` a method https://github.com/ManageIQ/integration_tests/blob/master/cfme/automate/buttons.py#L585

I am planning to change the structure of `button.py` but for now let's go with it.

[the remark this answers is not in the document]
digitronik: http://pastebin.test.redhat.com/655607

a collection should know object type as per model implementation.
1. when I created object > collection know type is `Provider` so everything is ok
2. but If I instantiate any other object with different type > collection does not know about new object type. to avoid this I am trying to provide new object type.

I know this is bad for now. it's the only solution; I found to avoid this situation.

PR #7973, opened 2018-10-10
[1LP][RFR] fixed more is_displayed methods which I missed before

izapolsk:
There were some other views returning False which didn't match pattern. 
Those should also raise NotImplementedError until those are correctly defined.


izapolsk:
@RonnyPfannschmidt, I agree but mixin isn't obvious. 
I'll give it a thought and fix it soon.

PR #7975, opened 2018-10-10
[1LP][RFR] Add test for scl packages.

jaryn:
Purpose or Intent
=================
__Adding tests__ for non-existence rh-ruby packages in 5.10:
https://polarion.engineering.redhat.com/polarion/#/project/RHCF3/workitem?id=RHCF3-42352
{ pytest: -v cfme/tests/test_appliance.py::test_appliance_console_packages }

### DIFF
--- a/cfme/tests/test_appliance.py
+++ b/cfme/tests/test_appliance.py
@@ -180,3 +180,9 @@ def test_keys_included(appliance, soft_assert):
         soft_assert(appliance.ssh_client.run_command(
             "test -e /var/www/miq/vmdb/certs/{}".format(k)).success,
             "{} was not included in the build".format(k))
+
+
+@pytest.mark.ignore_stream("5.9")
+def test_appliance_console_packages(appliance):
+    """Test that we have no scl packages installed."""
+    assert appliance.ssh_client.run_command('scl --list | grep -v rh-ruby').success


cfme/tests/test_appliance.py
```diff
@@ -180,3 +180,8 @@ def test_keys_included(appliance, soft_assert):
         soft_assert(appliance.ssh_client.run_command(
             "test -e /var/www/miq/vmdb/certs/{}".format(k)).success,
             "{} was not included in the build".format(k))
+
+
+@pytest.mark.ignore_stream("5.9")
+def test_appliance_console_packages(appliance):
+    assert appliance.ssh_client.run_command('scl --list | grep -v rh-ruby').success
```
jawatts: **Question** Are you expecting certain packages to be there? This is what I get when I run this command:

```[root@dhcp-8-198-63 ~]# scl --list | grep -v rh-ruby
rh-nodejs8
rh-postgresql95
[root@dhcp-8-198-63 ~]# 
```

I think we should be checking the output from this.

jawatts:
Also, please add pytest statement

cfme/tests/test_appliance.py
jaryn: I have just added a docstring for the test that explains that intention is to test for packages not being present. grep -v is inverse match. grep will exit with zero only if there are no packages matching. There is no need to check stdout.

jawatts: AH! I missed the `-v`, thanks for clarification!

PR #7976, opened 2018-10-10
[RFR] Fix message in unique name tests

apagac:
This is a quick fix for a expected error message in two tests:
- test_unique_project_name_on_parent_level
- test_unique_tenant_name_on_parent_level

I suppose I could have used versionpick. My chosen approach is quicker and easier though.

{{ pytest: -v cfme/tests/configure/test_access_control.py -k test_unique_tenant_name_on_parent_level }}


cfme/tests/configure/test_access_control.py
```diff
@@ -1059,8 +1059,10 @@ def tenant_unique_tenant_project_name_on_parent_level(request, appliance, object
         parent=object_type.get_root_tenant())
     if appliance.version < '5.9':
```
sshveta: This check for 5.8 can be removed now

PR #7977, opened 2018-10-11
[1LP][RFR] Test for new feature of VM: Rename  the VM

ganeshhubale:
{{ pytest: cfme/tests/infrastructure/test_vm_ownership.py::test_rename_vm --use-provider vsphere67-nested  -v }}

### DIFF
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -442,28 +442,29 @@ class BaseVM(BaseEntity, Pretty, Updateable, PolicyProfileAssignable, Taggable, 
         view = self.create_view(navigator.get_class(self, 'Details').VIEW)
         view.flash.assert_success_message(msg)
 
-    def rename_vm(self, new_vm_name, click_cancel=False, click_reset=False):
+    def rename_vm(self, new_vm_name, cancel=False, reset=False):
         """Rename the VM
 
         Args:
             new_vm_name: object for renaming vm
-            click_cancel (bool): Whether to cancel form submission
-            click_reset (bool): Whether to reset form after filling
+            cancel (bool): Whether to cancel form submission
+            reset (bool): Whether to reset form after filling
         """
         view = navigate_to(self, 'Rename')
-        view.form.fill({'vm_name': new_vm_name})
-        if click_reset:
-            view.form.reset_button.click()
-            view.flash.assert_no_error()
-            view.form.cancel_button.click()
-
-        elif click_cancel:
-            view.form.cancel_button.click()
-            view.flash.assert_no_error()
-
-        else:
-            # save the form
-            view.form.save_button.click()
+        changed = view.vm_name.fill(new_vm_name)
+        if changed:
+            if reset:
+                view.reset_button.click()
+                view.flash.assert_no_error()
+                view.cancel_button.click()
+            else:
+                # save the form
+                view.save_button.click()
+                view.flash.assert_no_error()
+                return self.appliance.collections.infra_vms.instantiate(new_vm_name, self.provider,
+                                                                        self.template_name)
+        if cancel:
+            view.cancel_button.click()
             view.flash.assert_no_error()
 
 
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -442,37 +442,29 @@ class BaseVM(BaseEntity, Pretty, Updateable, PolicyProfileAssignable, Taggable, 
         view = self.create_view(navigator.get_class(self, 'Details').VIEW)
         view.flash.assert_success_message(msg)
 
-    def rename_vm(self, new_vm_name, old_vm_name, click_cancel=False, click_reset=False):
+    def rename_vm(self, new_vm_name, click_cancel=False, click_reset=False):
         """Rename the VM
 
         Args:
             new_vm_name: object for renaming vm
-            old_vm_name: object for assertion
             click_cancel (bool): Whether to cancel form submission
             click_reset (bool): Whether to reset form after filling
         """
-        view = navigate_to(self, 'RenameVM', wait_for_view=0)
+        view = navigate_to(self, 'Rename')
         view.form.fill({'vm_name': new_vm_name})
         if click_reset:
             view.form.reset_button.click()
-            view.flash.assert_message('All changes have been reset', 'warning')
-            # Cancel after reset
-            assert view.form.is_displayed
+            view.flash.assert_no_error()
             view.form.cancel_button.click()
 
         elif click_cancel:
             view.form.cancel_button.click()
-            view.flash.assert_message('Rename of VM "{}" was cancelled by the user'
-                                      .format(old_vm_name))
+            view.flash.assert_no_error()
 
         else:
             # save the form
             view.form.save_button.click()
-            view = self.create_view(navigator.get_class(self, 'Details').VIEW)
-            view.wait_displayed("60s")
-            view.flash.assert_success_message('Rename of Virtual Machine "{}" has been initiated'
-                                              .format(old_vm_name))
-            return True
+            view.flash.assert_no_error()
 
 
 @attr.s
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -442,6 +442,38 @@ class BaseVM(BaseEntity, Pretty, Updateable, PolicyProfileAssignable, Taggable, 
         view = self.create_view(navigator.get_class(self, 'Details').VIEW)
         view.flash.assert_success_message(msg)
 
+    def rename_vm(self, new_vm_name, old_vm_name, click_cancel=False, click_reset=False):
+        """Rename the VM
+
+        Args:
+            new_vm_name: object for renaming vm
+            old_vm_name: object for assertion
+            click_cancel (bool): Whether to cancel form submission
+            click_reset (bool): Whether to reset form after filling
+        """
+        view = navigate_to(self, 'RenameVM', wait_for_view=0)
+        view.form.fill({'vm_name': new_vm_name})
+        if click_reset:
+            view.form.reset_button.click()
+            view.flash.assert_message('All changes have been reset', 'warning')
+            # Cancel after reset
+            assert view.form.is_displayed
+            view.form.cancel_button.click()
+
+        elif click_cancel:
+            view.form.cancel_button.click()
+            view.flash.assert_message('Rename of VM "{}" was cancelled by the user'
+                                      .format(old_vm_name))
+
+        else:
+            # save the form
+            view.form.save_button.click()
+            view = self.create_view(navigator.get_class(self, 'Details').VIEW)
+            view.wait_displayed("60s")
+            view.flash.assert_success_message('Rename of Virtual Machine "{}" has been initiated'
+                                              .format(old_vm_name))
+            return True
+
 
 @attr.s
 class BaseVMCollection(BaseCollection):
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -442,7 +442,7 @@ class BaseVM(BaseEntity, Pretty, Updateable, PolicyProfileAssignable, Taggable, 
         view = self.create_view(navigator.get_class(self, 'Details').VIEW)
         view.flash.assert_success_message(msg)
 
-    def rename_vm(self, new_vm_name, cancel=False, reset=False):
+    def rename(self, new_vm_name, cancel=False, reset=False):
         """Rename the VM
 
         Args:
@@ -461,8 +461,8 @@ class BaseVM(BaseEntity, Pretty, Updateable, PolicyProfileAssignable, Taggable, 
                 # save the form
                 view.save_button.click()
                 view.flash.assert_no_error()
-                return self.appliance.collections.infra_vms.instantiate(new_vm_name, self.provider,
-                                                                        self.template_name)
+                self.name = new_vm_name
+                return self
         if cancel:
             view.cancel_button.click()
             view.flash.assert_no_error()
--- a/cfme/common/vm_views.py
+++ b/cfme/common/vm_views.py
@@ -483,6 +483,23 @@ class SetOwnershipView(BaseLoggedInPage):
     is_displayed = displayed_not_implemented
 
 
+class RenameVmView(BaseLoggedInPage):
+    """Rename VM page for VMs"""
+
+    title = Text('#explorer_title_text')
+
+    @View.nested
+    class form(View):  # noqa
+        vm_name = Input("name")
+        save_button = Button('Save')
+        reset_button = Button('Reset')
+        cancel_button = Button('Cancel')
+
+    @property
+    def is_displayed(self):
+        return self.title.read() == 'Renaming VM "{}"'.format(self.context['object'].name)
+
+
 class ManagementEngineView(BaseLoggedInPage):
     """
     Edit management engine relationship page
--- a/cfme/common/vm_views.py
+++ b/cfme/common/vm_views.py
@@ -487,13 +487,10 @@ class RenameVmView(BaseLoggedInPage):
     """Rename VM page for VMs"""
 
     title = Text('#explorer_title_text')
-
-    @View.nested
-    class form(View):  # noqa
-        vm_name = Input("name")
-        save_button = Button('Save')
-        reset_button = Button('Reset')
-        cancel_button = Button('Cancel')
+    vm_name = Input("name")
+    save_button = Button('Save')
+    reset_button = Button('Reset')
+    cancel_button = Button('Cancel')
 
     @property
     def is_displayed(self):
--- a/cfme/common/vm_views.py
+++ b/cfme/common/vm_views.py
@@ -497,7 +497,7 @@ class RenameVmView(BaseLoggedInPage):
 
     @property
     def is_displayed(self):
-        return self.title.read() == 'Renaming VM "{}"'.format(self.context['object'].name)
+        return self.title.text == 'Renaming VM "{}"'.format(self.context['object'].name)
 
 
 class ManagementEngineView(BaseLoggedInPage):
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -34,7 +34,8 @@ from cfme.common.vm_views import (
     VMDetailsEntities,
     VMToolbar,
     VMEntities,
-    SetOwnershipView
+    SetOwnershipView,
+    RenameVmView
 )
 from cfme.exceptions import (
     DestinationNotFound,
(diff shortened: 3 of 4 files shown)


dajoRH:
*CFME QE Bot*

cfme/common/vm.py
```diff
@@ -438,6 +438,38 @@ def unset_ownership(self):
         view = self.create_view(navigator.get_class(self, 'Details').VIEW)
         view.flash.assert_success_message(msg)
 
+    def rename_vm(self, new_vm_name, old_vm_name, click_cancel=False, click_reset=False):
```
digitronik: Required:
No need of `old_vm_name` argument

cfme/common/vm.py
```diff
@@ -438,6 +438,38 @@ def unset_ownership(self):
...
+            view.form.cancel_button.click()
+            view.flash.assert_message('Rename of VM "{}" was cancelled by the user'
+                                      .format(old_vm_name))
+
+        else:
+            # save the form
+            view.form.save_button.click()
+            view = self.create_view(navigator.get_class(self, 'Details').VIEW)
+            view.wait_displayed("60s")
+            view.flash.assert_success_message('Rename of Virtual Machine "{}" has been initiated'
```
digitronik: Required:
Please avoid asserting such messages in the `model`. you can assert this message in test

cfme/infrastructure/virtual_machines.py
```diff
@@ -1479,6 +1480,15 @@ def step(self, *args, **kwargs):
         self.prerequisite_view.toolbar.configuration.item_select('Set Ownership')
 
 
+@navigator.register(InfraVm, 'RenameVM')
```
digitronik: Optional:
I would use only `Rename`

cfme/common/vm_views.py
```diff
@@ -483,6 +483,23 @@ class form(View):  # noqa
...
+    @View.nested
+    class form(View):  # noqa
+        vm_name = Input("name")
+        save_button = Button('Save')
+        reset_button = Button('Reset')
+        cancel_button = Button('Cancel')
+
+    @property
+    def is_displayed(self):
+        return self.title.read() == 'Renaming VM "{}"'.format(self.context['object'].name)
```
digitronik: Required:
`self.title.text`

cfme/common/vm_views.py
```diff
@@ -483,6 +483,23 @@ class form(View):  # noqa
     is_displayed = displayed_not_implemented
 
 
+class RenameVmView(BaseLoggedInPage):
+    """Rename VM page for VMs"""
+
+    title = Text('#explorer_title_text')
+
+    @View.nested
+    class form(View):  # noqa
```
digitronik: Optional:
No need of view nesting for such small view

cfme/common/vm.py
```diff
@@ -438,6 +438,38 @@ def unset_ownership(self):
...
+                                              .format(old_vm_name))
+            return True
```
digitronik: Question:
This return for?
In which condition it will rerun `False`?

cfme/common/vm.py
```diff
@@ -438,6 +438,38 @@ def unset_ownership(self):
...
+            click_cancel (bool): Whether to cancel form submission
+            click_reset (bool): Whether to reset form after filling
+        """
+        view = navigate_to(self, 'RenameVM', wait_for_view=0)
+        view.form.fill({'vm_name': new_vm_name})
+        if click_reset:
+            view.form.reset_button.click()
+            view.flash.assert_message('All changes have been reset', 'warning')
+            # Cancel after reset
+            assert view.form.is_displayed
```
digitronik: Required:
Not needed. Navigation will take of this.

cfme/common/vm.py
```diff
@@ -438,6 +438,38 @@ def unset_ownership(self):
...
+    def rename_vm(self, new_vm_name, old_vm_name, click_cancel=False, click_reset=False):
+        """Rename the VM
+
+        Args:
+            new_vm_name: object for renaming vm
+            old_vm_name: object for assertion
+            click_cancel (bool): Whether to cancel form submission
+            click_reset (bool): Whether to reset form after filling
+        """
+        view = navigate_to(self, 'RenameVM', wait_for_view=0)
```
digitronik: Required:
we use `wait_for_view=0` only in such a condition when we are not in a position to check `is_displayed`. As per my understanding, you have `is_displayed` method.

cfme/tests/infrastructure/test_vm_ownership.py
```diff
@@ -85,3 +91,33 @@ def test_set_vm_owner(self, appliance, vm, from_detail):
...
+    vm = provider.appliance.collections.infra_vms.instantiate(random_vm_name(context='rename'),
+                                                              provider,
+                                                              small_template_modscope.name)
+    vm.create_on_provider(find_in_cfme=True, allow_skip="default")
+    vm.refresh_relationships()
+    yield vm
+    vm.cleanup_on_provider()
+
+
+@pytest.mark.uncollectif(lambda appliance: appliance.version < '5.10')
```
digitronik: Required:
Use:
`@pytest.mark.ignore_stream("5.9")`

cfme/tests/infrastructure/test_vm_ownership.py
```diff
@@ -1,11 +1,17 @@
+import fauxfactory
 import pytest
 
 from cfme import test_requirements
 from cfme.infrastructure.provider import InfraProvider
+from cfme.infrastructure.provider.virtualcenter import VMwareProvider
+from cfme.markers.env_markers.provider import ONE_PER_TYPE
```
digitronik: Required: you can import this in L8.

cfme/tests/infrastructure/test_vm_ownership.py
```diff
@@ -85,3 +91,35 @@ def test_set_vm_owner(self, appliance, vm, from_detail):
...
+                                                              provider,
+                                                              small_template_modscope.name)
+    vm.create_on_provider(find_in_cfme=True, allow_skip="default")
+    vm.refresh_relationships()
+    yield vm
+    vm.cleanup_on_provider()
+
+
+@pytest.mark.ignore_stream('5.9')
+@pytest.mark.provider([VMwareProvider], override=True, scope="module", selector=ONE_PER_TYPE)
```
digitronik: Required:
Please, Provider scope should always `function` not `module`

cfme/tests/infrastructure/test_vm_ownership.py
```diff
@@ -85,3 +91,35 @@ def test_set_vm_owner(self, appliance, vm, from_detail):
...
+    3. Navigate to details page of VM
+    4. Click on Configuration > Rename this VM > Enter new name
+    5. Click on submit
+    6. Check whether VM is renamed or not
+    """
+    view = navigate_to(small_vm, 'Details')
+    small_vm.rename_vm(new_vm_name="test-{}".format(fauxfactory.gen_alphanumeric()))
+    view.wait_displayed("60s")
+    msg = 'Rename of Virtual Machine "{vm_name}" has been initiated'.format(vm_name=small_vm.name)
+    assert view.flash.read()[0] == msg
```
digitronik: Question:
Why you are asserting in this way?

ganeshhubale: @digitronik 
Yes, I changed the way of asserting and used **view.flash.assert_success_message()**.

cfme/common/vm.py
```diff
@@ -442,6 +442,30 @@ def unset_ownership(self):
         view = self.create_view(navigator.get_class(self, 'Details').VIEW)
         view.flash.assert_success_message(msg)
 
+    def rename_vm(self, new_vm_name, click_cancel=False, click_reset=False):
```
digitronik: @ganeshhubale please change argument names like `cancel` and `reset`

cfme/common/vm.py
```diff
@@ -442,6 +442,30 @@ def unset_ownership(self):
...
+        """Rename the VM
+
+        Args:
+            new_vm_name: object for renaming vm
+            click_cancel (bool): Whether to cancel form submission
+            click_reset (bool): Whether to reset form after filling
+        """
+        view = navigate_to(self, 'Rename')
+        view.form.fill({'vm_name': new_vm_name})
+        if click_reset:
```
digitronik: Required:
`Reset` button will activate only when the name is changed. If we send the same name it will not activate, right? better to capture changes... like
```
changed = view.form.fill({'vm_name': new_vm_name})
if changed:
     do operations
```

cfme/common/vm.py
```diff
@@ -442,6 +442,30 @@ def unset_ownership(self):
...
+            view.flash.assert_no_error()
+            view.form.cancel_button.click()
+
+        elif click_cancel:
+            view.form.cancel_button.click()
+            view.flash.assert_no_error()
+
+        else:
+            # save the form
+            view.form.save_button.click()
```
digitronik: Required:
After rename VM. we need to return `VM object` with the `new name` (instantiate)

ganeshhubale:
@digitronik 
Thanks for your suggestions with explanation. It helped me a lot while improvising this test.
Now this test is working as expected with suggested changes.

digitronik:
@ganeshhubale LGTM :+1: one small question
Can you please check created `VM` properly deleted at `teardown` or not? I am suspect about it.

ganeshhubale:
@digitronik 
- Yes, VM was not properly deleted at teardown because of change in vm object i.e. ```changed_vm_name```.
- Now, I have added ```request.addfinalizer(changed_vm_name.cleanup_on_provider)``` which stops and deletes the VM at teardown.
- I found this result in cfme.log file.
```
2018-11-22 11:10:56,584 [I] [wrapanapi.systems.virtualcenter.VMWareVirtualMachine] VM <name=test-G9UkBDNdNb> in actionable state. current state: VmState.RUNNING, ensuring state: VmState.STOPPED) (/home/ghubale/cfme_workspace/cfme_venv/lib/python2.7/site-packages/wrapanapi/entities/vm.py:221)
2018-11-22 11:10:58,480 [I] [wrapanapi.systems.virtualcenter.VMWareVirtualMachine]  Stopping vSphere VM test-G9UkBDNdNb (/home/ghubale/cfme_workspace/cfme_venv/lib/python2.7/site-packages/wrapanapi/systems/virtualcenter.py:508)
2018-11-22 11:11:17,162 [I] [wrapanapi.systems.virtualcenter.VMWareVirtualMachine]  Deleting vSphere VM/template test-G9UkBDNdNb (/home/ghubale/cfme_workspace/cfme_venv/lib/python2.7/site-packages/wrapanapi/systems/virtualcenter.py:166)

```

cfme/tests/infrastructure/test_vm_ownership.py
```diff
@@ -85,3 +90,38 @@ def test_set_vm_owner(self, appliance, vm, from_detail):
...
+
+@pytest.fixture(scope='module')
+def small_vm(provider, small_template_modscope):
+    vm = provider.appliance.collections.infra_vms.instantiate(random_vm_name(context='rename'),
+                                                              provider,
+                                                              small_template_modscope.name)
+    vm.create_on_provider(find_in_cfme=True, allow_skip="default")
+    vm.refresh_relationships()
+    yield vm
+    vm.cleanup_on_provider()
```
digitronik: Required:
```
if vm.exists:
    vm.cleanup_on_provider()
```

cfme/tests/infrastructure/test_vm_ownership.py
```diff
@@ -85,3 +90,38 @@ def test_set_vm_owner(self, appliance, vm, from_detail):
...
+    6. Check whether VM is renamed or not
+    """
+    view = navigate_to(small_vm, 'Details')
+    changed_vm_name = small_vm.rename_vm(new_vm_name="test-{}".
+                                         format(fauxfactory.gen_alphanumeric()))
+    view.flash.wait_displayed("20s")
+    msg = 'Rename of Virtual Machine "{vm_name}" has been initiated'.format(vm_name=small_vm.name)
+    view.flash.assert_success_message(msg)
+    assert changed_vm_name.exists
+    request.addfinalizer(changed_vm_name.cleanup_on_provider)
```
digitronik: I would `remove` in place of adding a finalizer else add it after object creation step.

ganeshhubale: yes, renamed vm object to it's original state so that it could be removed properly.

cfme/tests/infrastructure/test_vm_ownership.py
```diff
@@ -85,3 +90,39 @@ def test_set_vm_owner(self, appliance, vm, from_detail):
...
+    vm.create_on_provider(find_in_cfme=True, allow_skip="default")
+    vm.refresh_relationships()
+    yield vm
+    if vm.exists:
+        vm.cleanup_on_provider()
+
+
+@pytest.mark.ignore_stream('5.9')
+@pytest.mark.provider([VMwareProvider], override=True, scope="module", selector=ONE_PER_TYPE)
+def test_rename_vm(request, appliance, setup_provider, small_vm):
```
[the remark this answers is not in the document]
ganeshhubale: @izapolsk 
I removed ```request, appliance and setup_provider``` parameters from this test as these are not required to execute this test.

cfme/tests/infrastructure/test_vm_ownership.py
```diff
@@ -85,3 +90,39 @@ def test_set_vm_owner(self, appliance, vm, from_detail):
...
+    Steps:
+    1. Add VMware provider
+    2. Provision VM
+    3. Navigate to details page of VM
+    4. Click on Configuration > Rename this VM > Enter new name
+    5. Click on submit
+    6. Check whether VM is renamed or not
+    """
+    view = navigate_to(small_vm, 'Details')
+    changed_vm_name = small_vm.rename_vm(new_vm_name="test-{}".
```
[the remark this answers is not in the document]
ganeshhubale: This is the VM object. Not VM name. So changed it's name to **changed_vm**.

cfme/common/vm.py
```diff
@@ -442,6 +442,31 @@ def unset_ownership(self):
         view = self.create_view(navigator.get_class(self, 'Details').VIEW)
         view.flash.assert_success_message(msg)
 
+    def rename_vm(self, new_vm_name, cancel=False, reset=False):
```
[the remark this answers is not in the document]
mshriver: @ganeshhubale this method needs to be moved so that it is properly scoped.

If its only supported for vsphere vms, I agree with @izapolsk and @digitronik that it should exist in `InfraVm`, or even as a method in `VMwareProvider` that takes a vm object and a new name, returning a newly instantiated object with the new name.

cfme/common/vm.py
```diff
@@ -442,6 +442,31 @@ def unset_ownership(self):
...
+        if changed:
+            if reset:
+                view.reset_button.click()
+                view.flash.assert_no_error()
+                view.cancel_button.click()
+            else:
+                # save the form
+                view.save_button.click()
+                view.flash.assert_no_error()
+                return self.appliance.collections.infra_vms.instantiate(new_vm_name, self.provider,
```
mshriver: Why instantiate a new VM instead of just setting `self.name = new_vm_name`?

cfme/tests/infrastructure/test_vm_ownership.py
mshriver: Has not been updated.

cfme/tests/infrastructure/test_vm_ownership.py
```diff
@@ -85,3 +90,38 @@ def test_set_vm_owner(self, appliance, vm, from_detail):
...
+    5. Click on submit
+    6. Check whether VM is renamed or not
+    """
+    view = navigate_to(small_vm, 'Details')
+    changed_vm = small_vm.rename_vm(new_vm_name="test-{}".format(fauxfactory.gen_alphanumeric()))
+    view.flash.wait_displayed("20s")
+    msg = 'Rename of Virtual Machine "{vm_name}" has been initiated'.format(vm_name=small_vm.name)
+    view.flash.assert_success_message(msg)
+    assert changed_vm.exists
+    changed_vm.rename_vm(small_vm.name)
```
mshriver: If you actually modify the `name` attribute of the InfraVm object, instead of instantiating a new one, then the cleanup will work just fine without having to do this rename back to the original.

cfme/tests/infrastructure/test_vm_ownership.py
```diff
@@ -85,3 +90,38 @@ def test_set_vm_owner(self, appliance, vm, from_detail):
...
+    4. Click on Configuration > Rename this VM > Enter new name
+    5. Click on submit
+    6. Check whether VM is renamed or not
+    """
+    view = navigate_to(small_vm, 'Details')
+    changed_vm = small_vm.rename_vm(new_vm_name="test-{}".format(fauxfactory.gen_alphanumeric()))
+    view.flash.wait_displayed("20s")
+    msg = 'Rename of Virtual Machine "{vm_name}" has been initiated'.format(vm_name=small_vm.name)
+    view.flash.assert_success_message(msg)
+    assert changed_vm.exists
```
mshriver: Is there a need to wait for this exists call, since the rename operation could still be in progress?

cfme/common/vm.py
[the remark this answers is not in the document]
mshriver: Also, this can just be called `rename()`, its already a method on the InfraVm class (should be ;) )

cfme/tests/infrastructure/test_vm_ownership.py
ganeshhubale: Yes. we need to check whether ```old_vm``` has changed it's name and it is exist with ```new_name``` or not.

cfme/common/vm.py
ganeshhubale: Yes, I have updated it with your suggestion. As we are only changing name of VM. So we don't need to reinitialize whole VM object again.
Thanks for this suggestion.

cfme/tests/infrastructure/test_vm_ownership.py
ganeshhubale: I did not change because it was clashing with old implementations. Now I have modified and used ```small_template``` fixture which is of  **function** scope to get vm template and changed scope of ```small_vm`` fixture to **function** scope.
Now I can able to run this test with **function** scope.
Please let me know, if you any suggestions around it.
Thank you @mshriver @digitronik

cfme/tests/infrastructure/test_vm_ownership.py
ganeshhubale: Thank you @mshriver 
This modification saved more time which was taken to perform renaming operation on VM again.

cfme/tests/infrastructure/test_vm_ownership.py
mshriver: I added commit for this

PR #7980, opened 2018-10-11
[WIPTEST] Network view fix

digitronik:
# Purpose or Intent
=================
- added some small fixes in network views
- tested most of with miq shell
http://pastebin.test.redhat.com/656198

- not going to cover all tests but adding one test  for @dkholodo as he only pointed me this

{{pytest: cfme/tests/networks -v}}

### DIFF
--- a/cfme/networks/views.py
+++ b/cfme/networks/views.py
@@ -419,7 +419,7 @@ class NetworkRouterDetailsView(BaseLoggedInPage):
     @property
     def is_displayed(self):
         return (super(BaseLoggedInPage, self).is_displayed and
-                self.navigation.currently_selected == ['Networks', 'Providers'] and
+                self.navigation.currently_selected == ['Networks', 'Network Routers'] and
                 self.title.text == '{name} (Summary)'.format(name=self.context['object'].name))
 
 
@@ -543,7 +543,7 @@ class SecurityGroupDetailsView(BaseLoggedInPage):
     @property
     def is_displayed(self):
         return (super(BaseLoggedInPage, self).is_displayed and
-                self.navigation.currently_selected == ['Networks', 'Providers'] and
+                self.navigation.currently_selected == ['Networks', 'Security Groups'] and
                 self.title.text == '{name} (Summary)'.format(name=self.context['object'].name))
 
 
@@ -665,7 +665,7 @@ class OneProviderSubnetView(BaseLoggedInPage):
         title = '{name} (All Cloud Subnets)'.format(name=self.context['object'].name)
         return (super(BaseLoggedInPage, self).is_displayed and
                 self.navigation.currently_selected == ['Networks', 'Providers'] and
-                self.title.text == title)
+                self.entities.title.text == title)
 
 
 class OneProviderBalancerView(BaseLoggedInPage):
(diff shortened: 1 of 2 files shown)


digitronik:
part of issue: https://github.com/ManageIQ/integration_tests/issues/7960

izapolsk:
@digitronik, please add more tests. because your changes may potentially break other network tests.

digitronik:
@izapolsk let's run all network test against this PR so I will get idea if any test failure due to wait_for

cfme/networks/views.py
```diff
@@ -783,6 +784,7 @@ def is_displayed(self):
 
 class FloatingIpDetailsView(BaseLoggedInPage):
     """ Represents detail view of floating ip """
+    title = Text('//div[@id="main-content"]//h1')
```
mshriver: I think the title should be within entities, so these view structure are consistent.

cfme/networks/views.py
```diff
@@ -798,3 +800,16 @@ class entities(View):  # noqa
...
+
+class OneProviderFloatingIpView(BaseLoggedInPage):
+    """ Represents Floating Ip all for specific Network provider """
+    toolbar = View.nested(OneProviderComponentsToolbar)
+    including_entities = View.include(FloatingIpEntities, use_parent=True)
+
+    @property
+    def is_displayed(self):
+        title = '{name} (All Floating IPs)'.format(name=self.context['object'].name)
+        return (super(BaseLoggedInPage, self).is_displayed and
```
mshriver: Consider calling `self.logged_in_as_current_user` directly instead of the call to `super().is_displayed`

mshriver:
@digitronik Thanks for providing some local testing results via the pastebin.

The pattern I would like to see when testing these views is:
1. Navigate and store the view
2. assert the view is displayed
3. in the browser session, navigate to any other page (good to try the other network pages)
4. assert view.is_displayed is False

In order to make sure the view doesn't falsely return True on other network pages.

digitronik:
@mshriver I added suggestions. As per the last run, the failures were not related to changes.

cfme/networks/views.py
```diff
@@ -735,7 +736,7 @@ def is_displayed(self):
         title = '{name} (All Security Groups)'.format(name=self.context['object'].name)
         return (super(BaseLoggedInPage, self).is_displayed and
                 self.navigation.currently_selected == ['Networks', 'Providers'] and
-                self.title.text == title)
+                self.entities.title.text == title)
```
mshriver: I don't think `title` widget exists on the entities view in this case. `SecurityGroupEntities` doesn't have it, and `BaseEntitiesView` doesn't have it.

The title widget definition on L729 above needs to be moved into the `SecurityGroupEntities` view class.

digitronik: No need of it. It was my bad. entities have a `title`

mshriver: alright, so the title widget that was added won't be needed.

digitronik: 👍 I removed Mike. It was my bad I was unknown about `BaseEntityConditionalView` even not checked with MIQ shell.

mshriver:
PRT failures don't look related.

PR #7982, opened 2018-10-11
[WIPTEST] 5.10 migration fixes

Yadnyawalkya:
Fixed added:
1. wait_for for table load
2. fixed diff modal titles for 5.9 and 5.10
3. fixed validation messages due to 5.10 UI changes
4. Removed useless test_v2v.py

{{pytest: cfme/tests/v2v/test_v2v_migrations_ui.py -k test_v2v_ui_set1 -k test_v2v_ui_set1 --use-provider rhv42 --use-provider vsphere67-nested --provider-limit 2 -vvvv}}

{{pytest: cfme/tests/v2v/test_csv_import.py --use-provider rhv42 --use-provider vsphere67-nested --provider-limit 2 -vvvv }}


cfme/tests/v2v/test_csv_import.py
```diff
@@ -60,10 +61,14 @@ def import_and_check(appliance, infra_map, error_text, filetype='csv', content=F
     except UnexpectedAlertPresentException:
         pass
     if table_hover:
-        wait_for(lambda: plan_view.vms.table.is_displayed,
-                 timeout=60, message='Wait for VMs view', delay=2)
+        wait_for(lambda: plan_view.vms.is_displayed,
+                 timeout=60, message='Wait for VMs view', delay=5)
         if table_hover is 'duplicate':
-            plan_view.vms.table[2][1].widget.click()
+            if appliance.version >= '5.10.0.19':
```
digitronik: Optional:
No need to go for a specific version. mostly we test with the latest.

jawatts: I do not think we need this version check. Since we are on 5.10.0.21 now, is testing going to occur on builds before this?

cfme/tests/v2v/test_csv_import.py
```diff
@@ -60,10 +61,14 @@ def import_and_check(appliance, infra_map, error_text, filetype='csv', content=F
...
+                # Version check due to change in order of valid vms
+                plan_view.vms.table[0][1].widget.click()
+            else:
+                plan_view.vms.table[2][1].widget.click()
```
digitronik: Please inline comment for this table column `widget`.

cfme/tests/v2v/test_csv_import.py
Yadnyawalkya: @jawatts @digitronik, this version check added with respect to 5.9 basically.
Agree with views https://github.com/ManageIQ/integration_tests/pull/7982#discussion_r228158800, changing it to stream-wise, `5.10` instead of `5.10.0.19`

Yadnyawalkya:
@digitronik @jawatts 5.9 passed, 5.10 failing due to UI changes, kk's https://github.com/ManageIQ/integration_tests/pull/8021 should fix that.

PR #7983, opened 2018-10-11
[1LP][RFR] new test for advanced setting for remote servers

lina-nikiforova:
{{ pytest: -v cfme/tests/configure/test_remote_server_tabs.py }}

new test to address new feature (in 5.10) https://bugzilla.redhat.com/show_bug.cgi?id=1536524

I had to modify is_displayed and navigation for Server as it depends whether the server is remote or not

### DIFF
--- a/cfme/tests/configure/test_remote_server_tabs.py
+++ b/cfme/tests/configure/test_remote_server_tabs.py
@@ -1,0 +1,43 @@
+# -*- coding: utf-8 -*-
+
+import pytest
+
+from cfme.utils.appliance.implementations.ui import navigate_to
+
+
+@pytest.fixture(scope="module")
+def configured_external_appliance(temp_appliance_preconfig, app_creds_modscope,
+                                  temp_appliance_unconfig):
+    hostname = temp_appliance_preconfig.hostname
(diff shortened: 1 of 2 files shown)


cfme/tests/configure/test_remote_server_tabs.py
```diff
@@ -0,0 +1,39 @@
...
+    temp_appliance_unconfig.appliance_console_cli.configure_appliance_external_join(hostname,
+        app_creds_modscope['username'], app_creds_modscope['password'], 'vmdb_production',
+        hostname, app_creds_modscope['sshlogin'], app_creds_modscope['sshpass'])
+    temp_appliance_unconfig.evmserverd.start()
+    temp_appliance_unconfig.evmserverd.wait_for_running()
+    temp_appliance_unconfig.wait_for_web_ui()
+    return temp_appliance_unconfig
+
+
+@pytest.mark.uncollectif(lambda appliance: appliance.version < '5.10')
```
digitronik: Required:
If this test not applicable to `5.10` then please use `ignore_stream`. We use `uncollectif`, when we like to `uncollect` for specific `attribute`, `object` or some `condition`. 

you can say some condition except only `stream`. I would like to use `only` in my statement because we will get some condition where we checked both `stream` and other `condition`.

PR #7988, opened 2018-10-11
[RFR]Myservice view fix

sshveta:
1) View is_displayed fix - Part of resolving #7960
2) corrected Text displayed for set ownership for both 5.9 and 5.10

{{pytest: cfme/tests/services/test_myservice.py --use-provider vsphere55 --long-running}}

### DIFF
--- a/cfme/services/myservice/ui.py
+++ b/cfme/services/myservice/ui.py
@@ -6,6 +6,7 @@ from widgetastic_patternfly import Input, BootstrapSelect, Dropdown, Button, Can
 from cfme.base.login import BaseLoggedInPage
 from cfme.common import TagPageView
 from cfme.common.vm_views import VMDetailsEntities
+from cfme.exceptions import displayed_not_implemented
 from cfme.services.myservice import MyService
 from cfme.services.requests import RequestsView
 from cfme.utils.appliance import MiqImplementationContext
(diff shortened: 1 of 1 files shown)


cfme/services/myservice/ui.py
```diff
@@ -148,7 +148,7 @@ def is_displayed(self):
         return (
             self.in_myservices and
             self.myservice.is_opened and
-            self.title.text == 'Set Ownership of Service "{}"'.format(self.context['object'].name))
+            self.title.text == 'Set Ownership for Service')
```
mshriver: This does not appear to uniquely identify the page for the service that ownership is being set for.

I think the correct thing to do in this case, if there's no other unique elements on the page, is to `raise NotImplementedError`.  Checkout the existing implementations to use a consistent string in this instance.

cfme/services/myservice/ui.py
```diff
@@ -191,7 +191,7 @@ class ServiceVMDetailsView(VMDetailsEntities):
     @property
     def is_displayed(self):
         return (
-            self.in_myservices and self.myservice.is_opened and
+            self.myservice.is_opened and
```
mshriver: Anything you can validate against to replace the call to `self.in_myservices`?

sshveta: ServiceVMDetailsView(VMDetailsEntities) 
ServiceVMDetailsView is inherting VMDetailsEntities which does not have in_myservice.

So can't implement self.in_myservices

PR #7992, opened 2018-10-12
[WIPTEST] Fix is_displayed method for InstanceDetailsView

nachandr:
Fix is_displayed method for InstanceDetailsView
Part of resolving #7960

{{pytest: cfme/tests/candu/test_vm_graph.py}}

### DIFF
--- a/cfme/cloud/instance/__init__.py
+++ b/cfme/cloud/instance/__init__.py
@@ -128,7 +128,7 @@ class InstanceDetailsView(CloudInstanceView):
             relationship_provider_name = relationships.get_text_of('Cloud Provider')
         except (NameError, NoSuchElementException):
             logger.warning('No "Cloud Provider" Relationship, assume instance view not displayed')
-            return False
+            raise NotImplementedError("This view has no unique markers for is_displayed check")
         return (
             self.in_cloud_instance and
             self.entities.title.text == 'Instance "{}"'.format(expected_name) and


cfme/cloud/instance/__init__.py
```diff
@@ -128,7 +128,7 @@ def is_displayed(self):
             relationship_provider_name = relationships.get_text_of('Cloud Provider')
         except (NameError, NoSuchElementException):
             logger.warning('No "Cloud Provider" Relationship, assume instance view not displayed')
-            return False
+            raise NotImplementedError("This view has no unique markers for is_displayed check")
```
digitronik: @nachandr can you please change it to `is_displayed = displayed_not_implemented`... please refer https://github.com/ManageIQ/integration_tests/pull/7989

digitronik: Please refer https://github.com/ManageIQ/integration_tests/pull/7989  for an exception `displayed_not_implemented`

digitronik:
Linking to issue https://github.com/ManageIQ/integration_tests/issues/7960

cfme/cloud/instance/__init__.py
mshriver: @digitronik I don't think that property will be useful here, as this is a special case where our ability to identify the screen is dynamic.

In this case we'll need to `raise NotImplementedError` directly.

digitronik: :+1: I  thought better to return `displayed_not_implemented`

PR #7995, opened 2018-10-15
[1LP][RFR] Limit tests to one provider per type

nachandr:
{{pytest: cfme/tests/candu/test_vm_graph.py --use-provider complete --collect-only }}

### DIFF
--- a/cfme/tests/candu/test_vm_graph.py
+++ b/cfme/tests/candu/test_vm_graph.py
@@ -7,6 +7,7 @@ from cfme.cloud.provider.azure import AzureProvider
 from cfme.cloud.provider.ec2 import EC2Provider
 from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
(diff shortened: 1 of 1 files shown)


cfme/tests/candu/test_vm_graph.py
```diff
@@ -17,7 +18,7 @@
     pytest.mark.tier(3),
     test_requirements.c_and_u,
     pytest.mark.usefixtures('setup_provider'),
-    pytest.mark.provider([VMwareProvider, RHEVMProvider, EC2Provider, AzureProvider],
+    pytest.mark.provider([VMwareProvider, RHEVMProvider, EC2Provider, AzureProvider], selector=ONE,
```
mshriver: I think you want ONE_PER_TYPE here, as this only collects and ONE, resulting in collection only against vmware as it matches the required_fields.

nachandr: @mshriver , That's correct.

My intent is to limit the tests to one per type, so I should be using ONE_PER_TYPE here.

I've updated the PR.

PR #7999, opened 2018-10-17
[WIPTEST] Use ensure_page_safe on provisioning form

### LINKED ISSUE #7965: [Bug] VM provision fails on NoSuchElementException
When you try to provision VM via Lifecycle Dropdown, you encounter this error:

> NoSuchElementException: Message: Could not find an element u'.//*[(self::input or self::textarea) and @name="service__vm_name"]'

See for example [here](http://infra-trackerbot2.cfme2.lab.eng.rdu2.redhat.com/ostriz/dini/?stream_report=downstream-510z&source=jenkins#cfme/tests/cloud_infra_common/test_provisioning.py/test_provision_approval[virtualcenter-6.7-edit]).
It happens just when UI is switching from "Request" tab to "Catalog" tab. The most weird thing is however, that such web element is indeed present on the page. See below:

image

My guess is that `widgetastic` tries to fill this field before the whole page is actually loaded. I was trying to fix it by overloading general `fill` method in `Tab`. I added few seconds sleep there (just for experimental purposes), but it did not help. Therefore creating this issue.

Additional info:
- I think I first encountered it on CFME 5.10.0.18
- Also, the very same goes for other tabs beyond "Catalog". If you for example don't fill anything in Catalog, you encounter the same issue on "Environment" tab and so on.

mshriver:
Within before_fill, the static sleep was sometimes failing with an activity spinner that took longer than 3 seconds to load the 2nd part of the provisioning form

@jan-zmeskal I'm interested in seeing how this impacts the various provisioning / provisioning dialog tests that are failing to find their elements.

We may need to include some allowance for delay between tab selection and sub-widget access as well.

{{ pytest: cfme/tests/infrastructure/test_provisioning_dialog.py cfme/tests/cloud_infra_common/test_provisioning.py --use-provider default --long-running -vvv }}

FIXES #7965

### DIFF
--- a/cfme/common/vm_views.py
+++ b/cfme/common/vm_views.py
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
 import os
-from time import sleep
 
 from lxml.html import document_fromstring
+from wait_for import wait_for
 from widgetastic.widget import (
     View, TableRow, Text, TextInput, ParametrizedView, Image, ConditionalSwitchableView)
 from widgetastic_patternfly import (
@@ -355,9 +355,7 @@ class ProvisionView(BaseLoggedInPage):
                 self.parent.sidebar.decrease_button.click()
                 self._select_template(template_name, provider_name)
             self.continue_button.click()
-            # TODO timing, wait_displayed is timing out and can't get it to stop in is_displayed
-            sleep(3)
-            self.flush_widget_cache()
+            wait_for(self.browser.plugin.ensure_page_safe, delay=.1, num_sec=10)
 
     is_displayed = displayed_not_implemented
 


cfme/common/vm_views.py
```diff
@@ -355,9 +355,7 @@ def before_fill(self, values):
                 self.parent.sidebar.decrease_button.click()
                 self._select_template(template_name, provider_name)
             self.continue_button.click()
-            # TODO timing, wait_displayed is timing out and can't get it to stop in is_displayed
-            sleep(3)
-            self.flush_widget_cache()
+            self.plugin.ensure_page_safe()
```
jan-zmeskal:
```suggestion
            self.browser.plugin.ensure_page_safe()
```

jan-zmeskal:
@mshriver I tried running it with my proposed change, but I am still getting
`NoSuchElementException: Message: Could not find an element Locator(by='xpath', locator=u'//div[@id="prov_vm_div"]/table')`

cfme/common/vm_views.py
mshriver: Thanks @jan-zmeskal, realized after updating that this is some new GH thing for suggested change - would have clicked 'apply suggestion' otherwise.

Debugging, looks like this may actually be related to the rowspan implementation. Its trying to check table elements before `child_widget_accessed` is executed, so the widget isn't displayed.

mshriver: https://github.com/RedHatQE/widgetastic.core/pull/121

PR #8004, opened 2018-10-17
[1LP][RFR] Test custom button dialog

digitronik:
Purpose or Intent
=================
{{pytest: -k "test_custom_button_dialog" -v}}

### DIFF
--- a/cfme/tests/automate/custom_button/__init__.py
+++ b/cfme/tests/automate/custom_button/__init__.py
@@ -1,5 +1,8 @@
 # Common stuff for custom button testing
 
+from widgetastic.widget import View, Text, TextInput
+from widgetastic_patternfly import Button
+
 
 def check_log_requests_count(appliance, parse_str=None):
     """ Method for checking number of requests count in automation log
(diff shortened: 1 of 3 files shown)


cfme/tests/automate/custom_button/test_infra_objects.py
```diff
@@ -193,3 +196,63 @@ def test_custom_button_automate(appliance, request, submit, setup_obj, button_gr
...
+    dialog_view = view.browser.create_view(TextInputDialogView)
+    dialog_view.wait_displayed()
+    assert dialog_view.service_name.fill("Custom Button Execute")
+    dialog_view.submit.click()
+
+    if not (
+        BZ(bug_id=1640592, forced_streams=["5.9", "5.10"]).blocks and obj_type == "TEMPLATE_IMAGE"
+    ):
+        view.wait_displayed("60s")
+    view.flash.assert_message("Order Request was Submitted")
```
sshveta: Question : Can you also check if the order request completed successfully for both cloud and infra objects ?

digitronik: @sshveta agreed with you... but custom button simple dialog submission did not create any request. So for better testing, I am adding `request` checks from `automation log`

cfme/tests/automate/custom_button/__init__.py
```diff
@@ -27,3 +30,17 @@ def log_request_check(appliance, expected_count):
...
+    """ This is view comes on different custom button objects for dialog execution"""
+
+    title = Text("#explorer_title_text")
+    service_name = TextInput(id="service_name")
+    submit = Button("Submit")
+    cancel = Button("Cancel")
+
+    @property
+    def is_displayed(self):
+        # This is only for wait for view
```
mshriver: This is fine like it is for now, but it sounds like you need to make a decision on either raising `NotImplementedError` here via `cfme.exceptions.displayed_not_implemented` property, or add some more specific checks against the title as well.

PR #8005, opened 2018-10-17
[RFR] [NOTEST] Added unzipping of manageiq templates for ec2 and gce

mmojzis:
Added unzipping of manageiq images, which are for some reason zipped for scvmm, gce, hyperv and azure.
It also changes image_name to unzipped image and then removes zipped archived from system.
It works for EC2. Can't test on GCE.

### DIFF
--- a/cfme/utils/template/base.py
+++ b/cfme/utils/template/base.py
@@ -1,4 +1,5 @@
 import os
+import re
 import tarfile
 
 
@@ -240,46 +241,45 @@ class ProviderTemplateUpload(object):
 
     @log_wrap("download image locally")
     def download_image(self):
+        suffix = re.compile(
+            r'^.*?[.](?P<ext>tar\.gz|tar\.bz2|\w+)$').match(self.image_name).group('ext')
         # Check if file exists already:
         if path.isfile(self.local_file_path):
             logger.info('Local image found, skipping download: %s', self.local_file_path)
-            return True
-
-        # Download file to cli-tool-client
-        try:
-            request.urlretrieve(self.raw_image_url, self.local_file_path)
-        except URLError:
-            logger.exception('Failed download of image using urllib')
-            return False
+            if suffix not in ['zip', 'tar.gz']:
+                return True
         else:
-            return True
+            # Download file to cli-tool-client
+            try:
+                request.urlretrieve(self.raw_image_url, self.local_file_path)
+            except URLError:
+                logger.exception('Failed download of image using urllib')
+                return False
 
-    @log_wrap('unzip image')
-    def unzip_image(self, archive_type='zip'):
-        """
-        Unzips image and then changes image name to extracted one.
+        # Unzips image  when suffix is zip or tar.gz and then changes image name to extracted one.
+        # For EC2 and SCVMM images is zip used and for GCE is tar.gz used.
 
-        For EC2 and SCVMM images is zip used and for GCE is tar.gz used.
-        """
-        zip_path = self.image_name
+        archive_path = self.image_name
         try:
-            if archive_type == 'zip':
-                archive = ZipFile(zip_path)
+            if suffix == 'zip':
+                archive = ZipFile(archive_path)
                 zipinfo = archive.infolist()
                 self._unzipped_file = zipinfo[0].filename
-            elif archive_type == 'tar.gz':
-                archive = tarfile.open(zip_path, "r:gz")
+            elif suffix == 'tar.gz':
+                archive = tarfile.open(archive_path, "r:gz")
                 self._unzipped_file = archive.firstmember.name
             else:
-                logger.exception("Archive type {} is not implemented yet.".format(archive_type))
-                return False
+                return True
+            if path.isfile(self.image_name):
+                os.remove(self.image_name)
+            logger.info('Image archived - unzipping as : %s', self._unzipped_file)
             archive.extractall()
             archive.close()
             # remove the archive
-            os.remove(zip_path)
+            os.remove(archive_path)
             return True
         except Exception:
-            logger.exception("{} archive unzip failed.".format(archive_type))
+            logger.exception("{} archive unzip failed.".format(suffix))
             return False
 
     @log_wrap('add template to glance')
--- a/cfme/utils/template/base.py
+++ b/cfme/utils/template/base.py
@@ -1,13 +1,18 @@
+import os
+import tarfile
+
+
 from contextlib import closing
 from os import path
 from threading import Lock
-
+from zipfile import ZipFile
 from cached_property import cached_property
 from fauxfactory import gen_alphanumeric
 from glanceclient import Client
 from six.moves.urllib.request import urlopen
 from six.moves.urllib.error import URLError
 from six.moves.urllib import request
+
 
 from cfme.cloud.provider.openstack import OpenStackProvider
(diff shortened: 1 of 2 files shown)


cfme/utils/template/base.py
```diff
@@ -254,6 +262,34 @@ def download_image(self):
...
+    @log_wrap('unzip image')
+    def unzip_image(self, archive_type='zip'):
+        """
+        Unzips image and then changes image name to extracted one.
+
+        For EC2 and SCVMM images is zip used and for GCE is tar.gz used.
+        """
+        zip_path = self.image_name
+        try:
+            if archive_type == 'zip':
```
mshriver: I would think you can detect the archive type from the raw_image_url, just by looking at the file extension?  Then this doesn't have to take any arguments.

cfme/utils/template/base.py
```diff
@@ -254,6 +262,34 @@ def download_image(self):
         else:
             return True
 
+    @log_wrap('unzip image')
+    def unzip_image(self, archive_type='zip'):
```
mshriver: Consider making this part of download_image, called automatically when the downloaded file is `.zip` or `.tar.gz`. Then the `run` methods don't have to separately call this to unpack the archive.

cfme/utils/template/ec2.py
```diff
@@ -71,6 +71,10 @@ def run(self):
         if not self.download_image():
             return False
 
+        if 'miq' in self.template_name:
```
mshriver: if `unzip_image()` detects the file type and only unpacks when appropriate, then there is no need to check for template name substrings to explicitly call this. Combined with it being called by download_image and there's no need for this additional call here.

cfme/utils/template/gce.py
```diff
@@ -66,10 +66,21 @@ def create_template(self):
...
 
     def run(self):
-        self.download_image()
-        self.create_bucket()
-        self.upload_image()
-        self.create_template()
+        if not self.download_image():
+            return False
+
+        if 'miq' in self.template_name:
```
mshriver: Again I'd much rather unzip_image just skip when the file isn't a zip or tag.gz archive, and not have to explicitly check upstream/downstream via some substring. If the miq_version.TemplateName changes format and starts using `manageiq` instead of `miq`, then this breaks.

cfme/utils/template/gce.py
```diff
@@ -66,10 +66,21 @@ def create_template(self):
         return True
 
     def run(self):
-        self.download_image()
-        self.create_bucket()
-        self.upload_image()
-        self.create_template()
+        if not self.download_image():
+            return False
```
mshriver: Good call on wrapping these to return when they fail.

cfme/utils/template/base.py
```diff
@@ -131,10 +137,12 @@ def raw_image_url(self):
 
     @property
     def image_name(self):
-        """ Returns filename of an image.
+        """ Returns filename of an downloaded or unzipped image.
 
         Example output: cfme-type-version-x86_64-vhd
         """
+        if self._unzipped_file:
```
mshriver: can just be `return self._unzipped_file or self.raw_image_url.split('/')[-1]`

cfme/utils/template/base.py
mmojzis: Yes, it makes sense. I had problems with tar.gz so I didn't do it this way, but reworked.

cfme/utils/template/base.py
mmojzis: Done

cfme/utils/template/ec2.py
mmojzis: Done

cfme/utils/template/base.py
mmojzis: Done

cfme/utils/template/base.py
```diff
@@ -234,19 +241,46 @@ def deploy_template(self):
 
     @log_wrap("download image locally")
     def download_image(self):
+        suffix = re.compile(
+            r'^.*?[.](?P<ext>tar\.gz|tar\.bz2|\w+)$').match(self.image_name).group('ext')
         # Check if file exists already:
         if path.isfile(self.local_file_path):
             logger.info('Local image found, skipping download: %s', self.local_file_path)
-            return True
+            if suffix not in ['zip', 'tar.gz']:
```
mshriver: `suffix` uses `self.image_name`, whereas this is operating off of `self.local_file_path` - I think that you need to compile the suffix regex from `self.local_file_path` to be specific here that you're checking the _downloaded_ file that was found has been unpacked

cfme/utils/template/base.py
```diff
@@ -234,19 +241,46 @@ def deploy_template(self):
...
-        else:
+            if suffix == 'zip':
+                archive = ZipFile(archive_path)
+                zipinfo = archive.infolist()
+                self._unzipped_file = zipinfo[0].filename
+            elif suffix == 'tar.gz':
+                archive = tarfile.open(archive_path, "r:gz")
+                self._unzipped_file = archive.firstmember.name
+            else:
+                return True
```
mshriver: I don't necessarily like this pattern of burying the return when suffix isn't of the archive types.

Again I think suffix should be checking against the downloaded file here, not the image_name that's parsed from URL.

May I suggest a bit different structure, that returns more immediately when there's no need to further process the file, instead of doing so within a logic block that is also removing files and re-naming object attributes.

Something along the lines of:

``` python
ARCHIVE_TYPES = ['zip', 'tar.gz']  # should be class constant
if suffix_from_downloaded_file not in ARCHIVE_TYPES:
    return True
else:
    logger.info('Unpacking archived image: %s', self.image_name)
    try:
        ## diapered try block for creating archive
    except Exception:
         ## diaper message
         return False
    try:
         ## big diapered try block for removing the downloaded archive, regardless of the creation failing (re-download next time anyway)
    except Exception:
        ## diaper message
        return False
    try:
         # diapered try block for extracting, closing archive, returning True
     except Exception:
          # diaper message
          return False

File: cfme/utils/template/base.py [hunk 9]
```diff
@@ -234,19 +241,46 @@ def deploy_template(self):
...
+            if path.isfile(self.image_name):
+                os.remove(self.image_name)
+            logger.info('Image archived - unzipping as : %s', self._unzipped_file)
+            archive.extractall()
+            archive.close()
+            # remove the archive
+            os.remove(archive_path)
```
mshriver: `archive_path` is equal to `self.image_name`, how is this call different than the one at L274?

mmojzis: It isn't equal as image_name returns `return self._unzipped_file or self.raw_image_url.split("/")[-1]` and _unzipped_file is already set. I need to delete original image_name, not image_name after setting _unzipped_file.

mmojzis: Also deleting image_name on L274 deletes file that might be already extracted in previous run so it doesn't throw an error that file already exists. so deleting image_name is not equal deleting archive_name.

cfme/utils/template/base.py
mmojzis: Actually local_file_path uses image_name anyway. I don't need whole path I think?
local_file_path = project_path.join(self.image_name).strpath
I will rather should everywhere use image_name than local_file_path.

PR #8007, opened 2018-10-18
[1LP][RFR] WaitTab and wt.core update for Table

mshriver:
testing table on tab fix

{{ pytest: cfme/tests/cloud_infra_common/test_provisioning.py cfme/tests/candu/test_host_graph.py --use-provider default --long-running -vvv }}

part of fixing #7960

### DIFF
--- a/cfme/configure/tasks.py
+++ b/cfme/configure/tasks.py
@@ -81,7 +81,7 @@ class Task(BaseEntity):
         name: Name of the task.
         user: User who initiated the task.
         server: Cfme server.
-        tab: WaitTab where current task located.
+        tab: Tab where current task located.
 
     """
     OK = 'ok'
--- a/cfme/configure/tasks.py
+++ b/cfme/configure/tasks.py
@@ -8,8 +8,8 @@ import re
 from navmazing import NavigateToAttribute
 from widgetastic.exceptions import RowNotFound
 from widgetastic.widget import View
-from widgetastic_manageiq import BootstrapSelect, Button, CheckboxSelect, Table
-from widgetastic_patternfly import Dropdown, Tab
+from widgetastic_manageiq import BootstrapSelect, Button, CheckboxSelect, Table, WaitTab
+from widgetastic_patternfly import Dropdown
 
 from cfme.base.login import BaseLoggedInPage
 from cfme.modeling.base import BaseCollection, BaseEntity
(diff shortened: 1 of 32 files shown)


mshriver:
@digitronik note Jenkins is running this branch as its default for 595 at the moment. PRT will take quite a while to get through even the subset of tests I have included, but its ran about 20 so far (as of writing). There are test failures, but they aren't necessarily related to this change. There will be followup PRs to deal with test specific failures, or other View related failures. Thanks for review!

cfme/configure/tasks.py
```diff
@@ -81,7 +81,7 @@ class Task(BaseEntity):
         name: Name of the task.
         user: User who initiated the task.
         server: Cfme server.
-        tab: Tab where current task located.
+        tab: WaitTab where current task located.
```
izapolsk: unintentional rename ?

widgetastic_manageiq/__init__.py
```diff
@@ -3660,7 +3660,31 @@ def fill(self, value):
         return True
 
 
-class PotentiallyInvisibleTab(Tab):
+class WaitTab(Tab):
```
izapolsk: please add a docstring and some todo. we need to fix ensure page safe later

widgetastic_manageiq/__init__.py
```diff
@@ -3660,7 +3660,31 @@ def fill(self, value):
...
+        _to_try = 2
+        _tries = 0
+        while _tries < _to_try:
+            _tries += 1
+            self.select()
+            try:
+                wait_for(self.is_active, delay=.1, num_sec=2)
+            except TimedOutError:
+                self.logger.exception(
+                    "Tab was not active in 3s{}".format(", retrying" if _tries < _to_try else "")
```
jawatts: should this be 2s?

mshriver: yeah

widgetastic_manageiq/__init__.py
```diff
@@ -3660,7 +3660,31 @@ def fill(self, value):
...
+    def child_widget_accessed(self, widget):
+        _to_try = 2
+        _tries = 0
+        while _tries < _to_try:
+            _tries += 1
+            self.select()
+            try:
+                wait_for(self.is_active, delay=.1, num_sec=2)
+            except TimedOutError:
+                self.logger.exception(
```
izapolsk: warn ?

widgetastic_manageiq/__init__.py
```diff
@@ -3660,7 +3660,31 @@ def fill(self, value):
         return True
 
 
-class PotentiallyInvisibleTab(Tab):
+class WaitTab(Tab):
+    def child_widget_accessed(self, widget):
+        _to_try = 2
+        _tries = 0
+        while _tries < _to_try:
```
izapolsk: 2 attempts ? you mentioned 3 when you talked about it.

widgetastic_manageiq/__init__.py
```diff
@@ -3660,7 +3660,31 @@ def fill(self, value):
...
+            try:
+                wait_for(self.is_active, delay=.1, num_sec=2)
+            except TimedOutError:
+                self.logger.exception(
+                    "Tab was not active in 3s{}".format(", retrying" if _tries < _to_try else "")
+                )
+                if _tries >= _to_try:  # raise the TimedOutError if we've tried enough times
+                    raise
+            else:
+                break  # no retry
```
izapolsk: you could put break right after wait

widgetastic_manageiq/__init__.py
```diff
@@ -3660,7 +3660,31 @@ def fill(self, value):
...
+        try:
+            widget.wait_displayed(timeout=2)
+        except TimedOutError:
+            self.logger.exception("Tab widget not displayed, continuing...")
+            pass
```
izapolsk: not necessary

widgetastic_manageiq/__init__.py
mshriver: Just two tries, don't want it taking up more time than that for workaround

widgetastic_manageiq/__init__.py
mshriver: could, exception log gives us the exception as well as the passed message.

widgetastic_manageiq/__init__.py
mshriver: wanted to keep 1 statement in the try when writing, but yep going to collapse this now. Thanks!

PR #8008, opened 2018-10-18
[RFR] Change nav targets for report prerequisites

mshriver:
View timing out in 59z

SavedReportDetailsView timing out in PRT, debugging further. This fixed 4/6 tests so far though.

{{ pytest: cfme/tests/intelligence/reports/test_generate_report.py --use-provider default --long-running }}

Part of resolving #7960

### DIFF
--- a/cfme/intelligence/reports/reports.py
+++ b/cfme/intelligence/reports/reports.py
@@ -18,7 +18,7 @@ from cfme.utils.blockers import BZ
 from cfme.utils.pretty import Pretty
 from cfme.utils.timeutil import parsetime
 from cfme.utils.update import Updateable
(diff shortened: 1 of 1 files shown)


cfme/intelligence/reports/reports.py
```diff
@@ -176,20 +178,22 @@ def all(cls, browser):
 
     @property
     def is_displayed(self):
+        expected_tree = [
```
jawatts: Question, isn't this under the "Saved Reports" navigation tree

mshriver: In this case, no, its the `Saved Reports` tab on a `Report` details page.

PR #8009, opened 2018-10-18
[RFR] Update SystemSchedule views for zone reference

mshriver:
{{ pytest: cfme/tests/configure/test_db_backup_schedule.py --long-running --use-provider default }}

Related to fixing #7960


jawatts:
PRT is skipping for both tests.

mshriver:
@jawatts known, its part of the PR content - the VM for log depot doesn't have IP, so I skipped in fixture. Not a failure I'm trying to address in this PR, I'd rather have the test author address that.

PR #8011, opened 2018-10-18
[1LP][RFR]Fixing navigation on host collection via Compute-Infrastructure-Providers

kedark3:
Purpose or Intent
=================

__Fixing__ host collection navigation to use correct view when navigating via Compute->Infrastructure->Providers.

{{pytest: cfme/tests/v2v/test_v2v_migrations.py --use-provider rhv42 --use-provider vsphere65-nested --provider-limit 2 -vvvv --long-running  -k 'test_single_datastore_single_vm_migration'}}

### DIFF
--- a/cfme/common/host_views.py
+++ b/cfme/common/host_views.py
@@ -276,5 +276,6 @@ class ProviderAllHostsView(HostsView):
     def is_displayed(self):
         return (
             self.navigation.currently_selected == ["Compute", "Infrastructure", "Providers"] and
-            self.title.text == "{} (All Managed Hosts)".format(self.context["object"].name)
+            self.title.text == "{} (All Managed Hosts)".format(
+                self.context["object"].filters['parent'].name)
         )
--- a/cfme/infrastructure/host.py
+++ b/cfme/infrastructure/host.py
@@ -19,7 +19,8 @@ from cfme.common.host_views import (
     HostDriftHistory,
     HostEditView,
     HostsView,
-    HostTimelinesView
+    HostTimelinesView,
+    ProviderAllHostsView
 )
 from cfme.exceptions import ItemNotFound
 from cfme.infrastructure.datastore import HostAllDatastoresView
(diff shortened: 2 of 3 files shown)


cfme/common/host_views.py
```diff
@@ -276,5 +276,6 @@ class ProviderAllHostsView(HostsView):
     def is_displayed(self):
         return (
             self.navigation.currently_selected == ["Compute", "Infrastructure", "Providers"] and
-            self.title.text == "{} (All Managed Hosts)".format(self.context["object"].name)
+            self.title.text == "{} (All Managed Hosts)".format(self.context["object"]
```
kedark3: This was required as HostCollection object did not have `name` attribute in itself

mshriver: consider breaking this after `.format(`

```
thing == '{} (title part)'.format(
    self.some_attributes.name
)

**Comment by kedark3** [comment]
@mshriver if this does not timeout/itemnotfound on host page in PRT then we are good to merge it.

**Comment by kedark3** [comment]
Linking to #7960

File: cfme/infrastructure/host.py [hunk 2]
```diff
@@ -634,7 +635,14 @@ def create_from_config(self, host_key):
 
 @navigator.register(HostsCollection)
 class All(CFMENavigateStep):
-    VIEW = HostsView
+
+    @property
+    def VIEW(self):
```
digitronik: please use  `# noqa` its not as per definition standard.

kedark3: will do. But, what is that? sorry I don't know.

cfme/v2v/migrations.py
```diff
@@ -414,7 +414,8 @@ class vms(View):  # noqa
 
         @property
         def is_displayed(self):
-            return self.filter_by_dropdown.is_displayed
+            return (self.filter_by_dropdown.is_displayed and
+                 (len(self.browser.elements(".//div[contains(@class,'spinner')]")) == 0))
```
mshriver: Where are you seeing this behavior?

ensure_page_safe should be handling this before navigation is attempted.

kedark3: it happens in modal.

PR #8014, opened 2018-10-19
[WIP] Expand Nuage sandbox

sasoc:
With this commit we expand Nuage sandbox with shared network resources (cloud network) and floating ip.  We create a small cloud network and use it to create floating ip, which is then associated with vm vport. We need both of those entities for future testing (e.g. inventory validation)

Sandbox fixture deploys Nuage enterprise (cloud tenant) and its entities. We can use it to get values of those entities.

### DIFF
--- a/cfme/fixtures/nuage.py
+++ b/cfme/fixtures/nuage.py
@@ -1,4 +1,5 @@
 import pytest
+import random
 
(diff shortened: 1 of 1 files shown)


sasoc:
cc @miha-plesko

mshriver:

PR #8015, opened 2018-10-19
[1LP][RFR] Download images before uploading to glance

mshriver:


cfme/utils/template/base.py
```diff
@@ -279,7 +279,10 @@ def glance_upload(self):
             # add location for image on standalone glance
             client.images.add_location(glance_image.id, self.raw_image_url, {})
         else:
-            client.images.upload(glance_image.id, self.raw_image_url)
+            if self.download_image():
```
digitronik: :+1:

PR #8017, opened 2018-10-19
[RFR] Updated CMQE SSL Test to Handle No Metrics

jawatts:
{{ pytest: cfme/tests/containers/test_ssl_providers.py::test_add_mertics_provider_ssl --use-provider ocp-311 }}


cfme/tests/containers/test_ssl_providers.py
```diff
@@ -117,9 +117,12 @@ def test_add_mertics_provider_ssl(provider, appliance, test_item,
             Hawkular Endpoint = SSL trusting custom CA/SSL without validation/SSL
         * Assert that provider was added successfully
         """
+    if not provider.endpoints.get('metrics', False):
```
mshriver: `hasattr()` would read a lot better I think

mshriver: My bad, its not AttrDict, its just a dictionary.

PR #8018, opened 2018-10-19
Add match_items() and first()

jaryn:
Purpose or Intent
=================
Add some helper methods for filtering and picking first item.

### DIFF
--- a/cfme/modeling/base.py
+++ b/cfme/modeling/base.py
@@ -190,3 +190,18 @@ def parent_of_type(obj, klass):
     for x in _walk_to_obj_root(obj):
         if isinstance(x, klass):
             return x
+
+
+def match_items(items, filter):
+    ''' Return only items matching filter dict. '''
+    for item in items:
+        for attr_name, val in filter.items():
+            if getattr(item, attr_name) != val:
+                break
+        else:
+            yield item
+
+
+def first(items):
+    ''' Return the first item in the iterable. '''
+    return next(iter(items))


cfme/modeling/base.py
```diff
@@ -190,3 +190,18 @@ def parent_of_type(obj, klass):
     for x in _walk_to_obj_root(obj):
         if isinstance(x, klass):
             return x
+
+
+def match_items(items, filter):
+    ''' Return only items matching filter dict. '''
```
mshriver: double quotes on these docblocks please

mshriver: Consider making this a BaseCollection classmethod instead of just this standalone method, for clarity during use and ability to call `self.match_items(self.all(), self.filters)`

cfme/modeling/base.py
```diff
@@ -190,3 +190,18 @@ def parent_of_type(obj, klass):
...
+    for item in items:
+        for attr_name, val in filter.items():
+            if getattr(item, attr_name) != val:
+                break
+        else:
+            yield item
+
+
+def first(items):
+    ''' Return the first item in the iterable. '''
```
mshriver: I think this should definitely be a BaseCollection method, if its included at all. I'm hesitant to include it because its a single line wrapper that doesn't handle StopIteration to return None, or anything 'special' that would make it valuable beyond the caller just using next() or .pop() themselves.

jaryn: This method is intended to be used when single item is expected to appear in some list _or generator_. So added value is that it handles both same way.

We can discuss adding handling StopIteration and raising some other exception instead, if you think it something we should do.

I've just realized I can be checking we got single and only value. So perhaps I should do that and rename this to `singletone` or `only` or something like that.

What do you think @mshriver

PR #8021, opened 2018-10-22
[WIP] Fixing navigation for cfme 510z and fixing a test

kedark3:
Purpose or Intent
=================

__Fixing__ Navigation as cfme 510z has different menu options as well as fixing `vm_obj` call by replacing it with `vm_list`

{{pytest: cfme/tests/v2v/test_v2v_migrations.py --use-provider rhv42 --use-provider vsphere67-nested --provider-limit 2 -vvvv --long-running -k 'test_single_network_single_vm_migration' }}

### DIFF
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -183,7 +183,11 @@ class BaseVM(BaseEntity, Pretty, Updateable, PolicyProfileAssignable, Taggable, 
         """Fetches MAC Address of VM"""
         # TODO: We should update this with wrapanapi method when it becomes available.
         view = navigate_to(self, "Details", use_resetter=False)
-        return view.entities.summary('Properties').get_text_of('MAC Address')
+        mac_addr_field = VersionPicker({
+            LOWEST: 'MAC Address',
+                    '5.10': 'MAC Addresses'
+        })
+        return view.entities.summary('Properties').get_text_of(mac_addr_field)
 
     @property
     def is_retired(self):
--- a/cfme/tests/v2v/test_v2v_migrations.py
+++ b/cfme/tests/v2v/test_v2v_migrations.py
@@ -68,8 +68,7 @@ def test_single_datastore_single_vm_migration(request, appliance, v2v_providers,
     wait_for(func=view.plan_in_progress, func_args=[migration_plan.name],
         message="migration plan is in progress, be patient please",
         delay=5, num_sec=1800)
-
-    view.migr_dropdown.item_select("Completed Plans")
+    view.switch_to("Completed Plans")
     view.wait_displayed()
     logger.info("For plan %s, migration status after completion: %s, total time elapsed: %s",
         migration_plan.name, view.migration_plans_completed_list.get_vm_count_in_plan(
@@ -207,7 +206,7 @@ def test_dual_nics_migration(request, appliance, v2v_providers, host_creds, conv
         message="migration plan is in progress, be patient please",
         delay=5, num_sec=2700)
 
-    view.migr_dropdown.item_select("Completed Plans")
+    view.switch_to("Completed Plans")
     view.wait_displayed()
     logger.info("For plan %s, migration status after completion: %s, total time elapsed: %s",
         migration_plan.name, view.migration_plans_completed_list.get_vm_count_in_plan(
@@ -248,7 +247,7 @@ def test_dual_disk_vm_migration(request, appliance, v2v_providers, host_creds, c
         message="migration plan is in progress, be patient please",
         delay=5, num_sec=3600)
 
-    view.migr_dropdown.item_select("Completed Plans")
+    view.switch_to("Completed Plans")
     view.wait_displayed()
     logger.info("For plan %s, migration status after completion: %s, total time elapsed: %s",
         migration_plan.name, view.migration_plans_completed_list.get_vm_count_in_plan(
@@ -370,7 +369,7 @@ def test_single_vm_migration_with_ssh(request, appliance, v2v_providers, host_cr
         message="migration plan is in progress, be patient please",
         delay=5, num_sec=1800)
 
-    view.migr_dropdown.item_select("Completed Plans")
+    view.switch_to("Completed Plans")
     view.wait_displayed()
     logger.info("For plan %s, migration status after completion: %s, total time elapsed: %s",
         migration_plan.name,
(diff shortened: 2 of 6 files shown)


kedark3:
I need to update View and Test Cases around these. Will merge after that @mshriver . Target is  Wed Morning. Wed is new 510z build.

cfme/v2v/migrations.py
```diff
@@ -682,13 +682,28 @@ def create(self, name, infra_map, vm_list, description=None, csv_import=False,
...
 class All(CFMENavigateStep):
     VIEW = MigrationDashboardView
     prerequisite = NavigateToAttribute('appliance.server', 'LoggedIn')
 
     def step(self):
-        self.prerequisite_view.navigation.select('Compute', 'Migration')
+        if self.obj.appliance.version < '5.10':
+            self.prerequisite_view.navigation.select('Compute', 'Migration')
+        else:
+            self.prerequisite_view.navigation.select('Compute', 'Migration', 'Overview')
```
mshriver: Consider just checking the type of `self.obj` in order to select `'Overview'` or `'Infrastructure Mappings'` for navigation. Then this can be a single nav class registered against both `InfrastructureMappingCollection` and `MigrationPlanCollection`

kedark3: I need to create 2 different views for Navigation to `All` for `InfraMapping` and `MigrationPlan` so wouldn't it be easier just to split them in two classes?

cfme/tests/v2v/test_v2v_migrations.py
```diff
@@ -517,18 +517,12 @@ def test_migration_special_char_name(request, appliance, v2v_providers, host_cre
     def _cleanup():
         infrastructure_mapping_collection.delete(mapping)
 
-    src_vm_obj = form_data_vm_obj_single_datastore.vm_obj[0]
```
kedark3: No longer required.

cfme/v2v/migrations.py
```diff
@@ -325,6 +331,8 @@ class MigrationDashboardView(BaseLoggedInPage):
 
     @property
     def is_displayed(self):
+        # TODO Uncomment next line, remove line after that
```
kedark3: This will be required to be done after next 5.10z build, as I built this code against upstream and there is a bug on vertical navigation in upstream. Bug should be resolved in next 5.10 build.

cfme/v2v/migrations.py
```diff
@@ -363,17 +371,51 @@ def plan_in_progress(self, plan_name):
             return False
 
 
+class MigrationDashboardView59z(MigrationDashboardView):
```
kedark3: Dashboard for 59z has `infra_mapping_list` while 510z moves it to a separate page.

cfme/v2v/migrations.py
```diff
@@ -363,17 +371,51 @@ def plan_in_progress(self, plan_name):
...
+    infra_mapping_list = InfraMappingList("infra-mappings-list-view")
+
+    @property
+    def is_displayed(self):
+        return (self.navigation.currently_selected == ['Compute', 'Migration'] and
+            (len(self.browser.elements(".//div[contains(@class,'spinner')]")) == 0) and
+            (len(self.browser.elements('.//div[contains(@class,"card-pf")]')) > 0))
+
+
+class InfrastructureMappingView(BaseLoggedInPage):
```
kedark3: This is an entire separate page, with many elements similar to what we had in `MigrationPlanRequestDetailsView` , so re-using some of those Paginator things by renaming those from `MigrationPlanRequestDetailsPaginator` to `v2vPaginator`, etc.

cfme/v2v/migrations.py
```diff
@@ -363,17 +371,51 @@ def plan_in_progress(self, plan_name):
...
+            ['Compute', 'Migration'] and
+            # ['Compute', 'Migration', 'Infrastructure Mappings'] and
+            len(self.browser.elements(".//div[contains(@class,'spinner')]")) == 0 and
+            (self.create_infrastructure_mapping.is_displayed or
+                self.infra_mapping_list.is_displayed))
+
+
 class AddInfrastructureMappingView(View):
     form = InfraMappingWizard()
+    form_title_text = VersionPicker({
```
kedark3: This was failing to `.pick()` right version/item while it was inside `is_displayed` so moved it out, now it works as expected.

cfme/v2v/migrations.py
```diff
@@ -363,17 +372,52 @@ def plan_in_progress(self, plan_name):
...
+    # USed to select sort by options like 'Name', 'Number of Associated Plans'
+    sort_by_dropdown = SelectorDropdown('id', 'sortTypeMenu')
+
+    @property
+    def is_displayed(self):
+        # TODO: Remove 1st condition, once /manageiq-v2v/issues/726 fix is backported to 510z
+        return ((self.navigation.currently_selected ==
+            ['Compute', 'Migration'] or self.navigation.currently_selected ==
+            ['Compute', 'Migration', 'Infrastructure Mappings']) and
+            len(self.browser.elements(".//div[contains(@class,'spinner')]")) == 0 and
```
mshriver: consider expressing this as `not self.browser.elements()`, which is returning an empty list on no locator match.

cfme/v2v/migrations.py
```diff
@@ -606,7 +649,9 @@ def create(self, form_data):
         return infra_map
 
     def delete(self, mapping):
-        view = navigate_to(self, 'All')
+        view = navigate_to(self, 'All', wait_for_view=True)
```
mshriver: `wait_for_view` kwarg is an integer number of seconds to wait. Default is 10s

cfme/common/vm.py
```diff
@@ -183,7 +183,7 @@ def mac_address(self):
         """Fetches MAC Address of VM"""
         # TODO: We should update this with wrapanapi method when it becomes available.
         view = navigate_to(self, "Details", use_resetter=False)
-        return view.entities.summary('Properties').get_text_of('MAC Address')
+        return view.entities.summary('Properties').get_text_of('MAC Addresses')
```
mshriver: Needs to be version picked, this is still 'Address' in 5.9.5

kedark3:
This PR was opened during GitHub outage. Seems to have issues updating eventhough I commit changes. Re-opening it.

PR #8022, opened 2018-10-22
[WIP] Validate nuage network ports inventory

sasoc:
With this commit we create Nuage network port and Nuage network port collection, which we then use to validate details page of Nuage network port.

We create sandbox with Nuage enterprise and its entities, including network port. We use this sandbox to gather all the stats we expect to be shown on the details page. Then we find and navigate to created network port where we compare its UI stats with expected ones.

We also use different steps to navigate to details page. We go through provider using steps like this:
```
Nertwork -> Providers -> Provider (by name) -> Network Ports -> Network Port (by name)
```
By using this steps we make sure that we are on the details page of Nuage network port.

Depends on: https://github.com/ManageIQ/integration_tests/pull/8014

### DIFF
--- a/cfme/networks/network_port.py
+++ b/cfme/networks/network_port.py
@@ -101,3 +101,12 @@ class Details(CFMENavigateStep):
 
     def step(self):
         self.prerequisite_view.entities.get_entity(name=self.obj.name, surf_pages=True).click()
+
+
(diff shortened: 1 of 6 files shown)


sasoc:
cc @miha-plesko

mshriver:

PR #8026, opened 2018-10-23
[WIPTEST] Fix ProviderAllHostsView for test_infra_provider_relationships

rhrazdil:
Fix ProviderAllHostsView for test_infra_provider_relationships

{{ pytest: cfme/tests/cloud_infra_common/test_relationships.py::test_infra_provider_relationships }}

### DIFF
--- a/cfme/common/host_views.py
+++ b/cfme/common/host_views.py
@@ -274,8 +274,12 @@ class ProviderAllHostsView(HostsView):
 
     @property
     def is_displayed(self):
-        return (
-            self.navigation.currently_selected == ["Compute", "Infrastructure", "Providers"] and
-            self.title.text == "{} (All Managed Hosts)".format(
-                self.context["object"].filters['parent'].name)
-        )
+        if self.navigation.currently_selected != ["Compute", "Infrastructure", "Providers"]:
+            return False
+
+        title = "{} (All Managed Hosts)"
+        if hasattr(self.context["object"], "name"):
+            return self.title.text == title.format(self.context["object"].name)
+        elif hasattr(self.context["object"], "filters"):
+            return self.title.text == title.format(self.context["object"].filters["parent"].name)
+        return False


cfme/common/host_views.py
```diff
@@ -276,6 +276,5 @@ class ProviderAllHostsView(HostsView):
     def is_displayed(self):
         return (
             self.navigation.currently_selected == ["Compute", "Infrastructure", "Providers"] and
-            self.title.text == "{} (All Managed Hosts)".format(
-                self.context["object"].filters['parent'].name)
+            self.title.text == "{} (All Managed Hosts)".format(self.context["object"].name)
```
mshriver: I don't think this fix will work - but you're absolutely on the right path.

So this view is used in two cases (at least) - `HostsCollection` 'All' nav destination in `cfme/infrastructure/host.py` and for `InfraProvider` 'Hosts' destination in `cfme/infrastructure/provider/__init__.py`

In the HostsCollection destination, the view is only selected when a parent or provider filter is applied to the collection.

Here, we can't assume that the contextual object is the provider - just like we can't assume its the HostsCollection either.  The current code assumes there's a parent filter, your change assumes that the context object is a provider.

This should check for the presence of a parent/provider filter, just like the VIEW property in HostsCollection 'All' nav destination.  The other thing to pay attention to is how the view may be used in a test.

mshriver:
Part of resoliving #7960

cfme/common/host_views.py
rhrazdil: @mshriver Thank you for the explanation. I see it's a bit more complicated than I originally thought.

cfme/common/host_views.py
```diff
@@ -274,8 +274,12 @@ class ProviderAllHostsView(HostsView):
...
-        )
+        if self.navigation.currently_selected != ["Compute", "Infrastructure", "Providers"]:
+            return False
+
+        title = "{} (All Managed Hosts)"
+        if hasattr(self.context["object"], "name"):
+            return self.title.text == title.format(self.context["object"].name)
+        elif hasattr(self.context["object"], "filters"):
+            return self.title.text == title.format(self.context["object"].filters["parent"].name)
+        return False
```
mshriver: This is an improved pattern, but still making an assumption that 'parent' is the filter that's set that will result in a KeyError when the collection is filtered otherwise.  I would also prefer to retain the pattern of returning a list of boolean-joined statements instead of having several return statements.  Python will short circuit and return as soon as `and` joined statements result in `False` - so having the `navigation.currently_selected` first means that if it fails, we won't run any other checks.

The filter selection needs to account for `parent` and `provider` filters, can I suggest something like this:

```
# include some logging messages too when making assumptions about what you're looking at
from cfme.modeling.base import BaseEntity, BaseCollection
expected_title = "{} (All Managed Hosts)"
obj = self.context['object']
is_entity = getattr(obj, 'name', False) and isinstance(obj, BaseEntity)
is_filtered = isinstance(obj, BaseCollection) and obj.filters
if is_entity:
    # object has name attribute and is BaseEntity derived, assuming its a provider
    matched_title = self.title.text == expected_title.format(obj.name)
elif is_filtered:
    filter = obj.filters.get('parent') or obj.filters.get('provider')
    if filter:
        matched_title = self.title.text == expected_title.format(filter.name)
    else:
        matched_title = False  # filtered, but not with either filter we expect
else:
    matched_title = False  # not an entity with a name, or a filtered collection

return (
    self.navigation.currently_selected == ['Compute', 'Infrastructure', 'Providers'] and
    matched_title
)

mshriver:
@rhrazdil has moved onto another team, and I'm going to close this PR and take over fixing this issue.

PR #8029, opened 2018-10-23
[RFR]Orchestration view is_displayed fix

sshveta:
1) Removed assertions to be moved to test in following PR
2) Fixed some views
3) Changed order of filling fields as provider dropdown appears once template is selected.

{{ pytest: --use-provider rhos11 cfme/tests/services/test_provision_stack.py::test_provision_stack }}


lina-nikiforova:
@sshveta can you add some tests to this PR?

mshriver:
The failure in PRT actually looks like a potential BZ, not a problem with the test. the flash message is looking for a field to be filled that is disabled.

PR #8032, opened 2018-10-24
[1LP][RFR] Add cloud provider dashboard tests

jadhaj:
Purpose or Intent
=================
Add test_dashboard_card_tenants

{{ pytest: cfme/tests/openstack/cloud/test_provider.py::test_dashboard_card_tenants cfme/tests/openstack/cloud/test_provider.py::test_dashboard_card_security_groups }}

### DIFF
--- a/cfme/tests/openstack/cloud/test_provider.py
+++ b/cfme/tests/openstack/cloud/test_provider.py
@@ -32,3 +32,22 @@ def test_dashboard_card_availability_zones(provider):
     view.toolbar.view_selector.select('Dashboard View')
     dashboard_card = view.entities.cards("Availability Zones")
     assert dashboard_card.value == len(provider.mgmt.api.availability_zones.list())
+
+
+@pytest.mark.ignore_stream('5.9')
+def test_dashboard_card_tenants(provider):
+    collection = provider.appliance.collections.cloud_tenants
+    view = navigate_to(provider, 'Details')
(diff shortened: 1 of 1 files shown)


cfme/tests/openstack/cloud/test_provider.py
```diff
@@ -32,3 +32,22 @@ def test_dashboard_card_availability_zones(provider):
...
+
+
+@pytest.mark.ignore_stream('5.9')
+def test_dashboard_card_tenants(provider):
+    collection = provider.appliance.collections.cloud_tenants
+    view = navigate_to(provider, 'Details')
+    view.toolbar.view_selector.select('Dashboard View')
+    card_tenants = view.entities.cards("Cloud Tenants").value
+    view = navigate_to(collection, 'All')
+    assert card_tenants == view.entities.paginator.items_amount
```
jaryn: Shouldn't we prefer the amounts in CFME to be checked against provider API instead of against CFME? In other words I don't want to believe CFME itself as source of the truth. I would rather believed the API.

jadhaj: @jaryn  Right and  this is what i did in previous 6 tests of dashbaord  cards. sometimes there is no correlation between UI and API, for instance in Tenants. if admin is not member of a tenant , it will no be reflected in UI . I already encountered such a deployment in PRT where number of tenants returned by API are much more than those shown in UI and again it is related to admin membership on tenants

Issue #8033, opened 2018-10-24
[Bug] view.form.fill_with in do_vm_provisioning returns False


jan-zmeskal:
It is called [here](https://github.com/ManageIQ/integration_tests/blob/master/cfme/provisioning.py#L23). Even though a lot of stuff is filled there during creating provision request, it still returns `False`. The same happen when you call just `fill` with the same provisioning data.

Here's my [conversation](https://up1-vrutkovs.int.open.paas.redhat.com/#B5e4i3cBdRyfhWdLnWI3ig) with @izapolsk and @mshriver.

jan-zmeskal:
It seems like the issue has been resolved already by some PR not referenced here

PR #8039, opened 2018-10-25
[1LP][RFR][NOTEST] Updating ISO provisioning test

jan-zmeskal:
This PR does a couple of things:

1) We started using RHEL7.5 image in ISO provisioning testing. That's why I am adding _wget_ installation to the Kickstart script. We need to make a callback to CFME in the end, but _wget_ is not present on clean RHEL7.5 box.

2) Removing BZ1613326 since it's been verified.

3) Adding #8033 as a blocker.

4) BZ1642924 now causes that e-mail about approval of provision request is not sent. Therefore I suggest we don't check it while the bug is open. It's probably not a good idea to block a big chunk of tests just because of that.

5) Adding `addr_mode` as a new value in `provisioning_data`. You can choose either "Static" or "DHCP". If you don't provide anything in cfme_data, then this is just ignored.

6) Setting VM to be powered off after provision. This is in accord with the documentation.

7) Setting timeout for 1800 seconds. My experience is that ISO provision of RHEL7.5 with our Kickstart takes about 23 minutes, so 1500 seconds is really cutting it close. 

Note about testing: Unfortunately, I cannot really verify this PR as it stands now. That's because #8033 is blocking automation. However, I tested it on our Jenkins with a workaround (adding `return True` to the end of `before_fill` of `ProvisionView.form`). Here's the result: https://rhv-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/cfme-5.10-rhv-4.2-integration-test-dev/43/

Shriver: wanted to see PRT results including 8033 behavior, feel free to remove my commit :)
{{ pytest: --use-provider complete --long-running cfme/tests/infrastructure/test_iso_provisioning.py }}

### DIFF
--- a/cfme/provisioning.py
+++ b/cfme/provisioning.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.blockers import BZ
 from cfme.utils.log import logger
 from cfme.utils.wait import wait_for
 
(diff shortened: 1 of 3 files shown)


sshveta:
> @sshveta Hi, I am very sorry, but that test case is not part of RHV-CFME regression suite. These days I have very little resources available for work on RHV-CFME tests. I cannot really invest my time in anything else.

Ok No worries I have someone from service team implement that .

mshriver:
ISO provisioning tests failing because rhv41 provider is missing expected ISO image, not related to changes here.

PR #8042, opened 2018-10-26
[1LP][RFR] added check for advanced tab for region and zones and fixed navigation

lina-nikiforova:
{{ pytest: -v cfme/tests/configure/test_paginator.py::test_paginator }}

1. Added check for the Advanced tab on Region and Zone level (https://bugzilla.redhat.com/show_bug.cgi?id=1082155)
2. Fixed some navigation issues

### DIFF
--- a/cfme/base/ui.py
+++ b/cfme/base/ui.py
@@ -8,7 +8,7 @@ from selenium.webdriver.common.keys import Keys
 from widgetastic.utils import Version, VersionPick
 from widgetastic.widget import View, Table, Text, Image, FileInput
 from widgetastic_manageiq import (ManageIQTree, Checkbox, AttributeValueForm, TimelinesView,
-                                  ParametrizedSummaryTable, SummaryFormItem, WaitTab)
+                                  SummaryFormItem, WaitTab)
 from widgetastic_patternfly import (Accordion, Input, Button, Dropdown, DatePicker,
                                     FlashMessages, BootstrapSelect, CheckableBootstrapTreeview)
 
@@ -620,12 +620,11 @@ class ServerDatabaseView(ConfigurationView):
         return (
             self.in_configuration and
             self.summary.is_displayed and
-            self.title.text == 'VMDB Summary'
+            self.accordions.database.is_opened
         )
 
 
 class DatabaseSummaryView(ServerDatabaseView):
-    summary = ParametrizedSummaryTable()
 
     @property
     def is_displayed(self):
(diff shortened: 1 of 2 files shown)


cfme/base/ui.py
```diff
@@ -1077,6 +1079,20 @@ def step(self):
         self.prerequisite_view.help_menu.select()
 
 
+@navigator.register(Region)
```
digitronik: Optional:
But, I like to go with registration name like L1357 you added. rather than taking `class` name.

cfme/base/ui.py
```diff
@@ -1077,6 +1079,20 @@ def step(self):
...
+class Advanced(CFMENavigateStep):
+    VIEW = RegionView
+    prerequisite = NavigateToSibling('Details')
+
+    def am_i_here(self):
+        return (
+            self.view.is_displayed and self.view.advanced.is_displayed and
```
digitronik: :+1: as a common view for all tabs and check for `advance` tab

cfme/base/ui.py
```diff
@@ -620,19 +620,17 @@ def is_displayed(self):
...
 
 
 class DatabaseSummaryView(ServerDatabaseView):
-    summary = ParametrizedSummaryTable()
 
     @property
     def is_displayed(self):
         return (
-            self.summary.is_displayed and
-            self.summary.is_active() and
```
digitronik: Question:
why you removed `summary` active check?

cfme/base/ui.py
```diff
@@ -620,19 +620,17 @@ def is_displayed(self):
         return (
             self.in_configuration and
             self.summary.is_displayed and
-            self.title.text == 'VMDB Summary'
```
digitronik: Question:
we are not getting any common `title` for all `tabs`?

lina-nikiforova: Yes. For the `ServerDatabaseView` we need to be sure we are in the Database section, for that I added `self.accordions.database.is_opened` below.
The title is different for each tab in this section and it's checked in the respective view for the tab, L627-665.

cfme/base/ui.py
lina-nikiforova: I'd keep it like that, otherwise I need to re-write navigation registration for other tabs in `Region`. Now they are in the same pattern.

lina-nikiforova: oh no, I meant I need to re-write it to be written in the same way, not because it will be broken :)

PR #8044, opened 2018-10-29
[WIPTEST] fix is_displayed of AttachVolumeView

jadhaj:
Purpose or Intent
=================
{{ pytest: cfme/tests/openstack/cloud/test_instances.py::test_instance_attach_volume }}


mshriver:
PRT failures aren't related to the volume view, this simple change is the correct way to implement views that we don't confirm view on.

PR #8053, opened 2018-10-31
[1LP][RFR] Add support for volume type field

jadhaj:
Purpose or Intent
=================
Add support for volume type field

{{ pytest: cfme/tests/openstack/cloud/test_volumes.py::test_create_volume }}

### DIFF
--- a/cfme/storage/volume.py
+++ b/cfme/storage/volume.py
@@ -102,9 +102,10 @@ class VolumeAddEntities(View):
 
 class VolumeAddForm(View):
     storage_manager = BootstrapSelect(name='storage_manager_id')
-    volume_name = TextInput(name='name')
-    size = TextInput(name='size')
     tenant = BootstrapSelect(name='cloud_tenant_id')
+    volume_name = TextInput(name='name')
+    volume_type = BootstrapSelect(name='volume_type')
+    size = TextInput(name='size')
     add = Button('Add')
     cancel = Button('Cancel')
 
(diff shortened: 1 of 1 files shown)


cfme/storage/volume.py
```diff
@@ -105,6 +105,7 @@ class VolumeAddForm(View):
     volume_name = TextInput(name='name')
     size = TextInput(name='size')
     tenant = BootstrapSelect(name='cloud_tenant_id')
+    type = BootstrapSelect(name='volume_type')
```
jaryn: I would recommend using `volume_type` for this field to not shadow the builtin `type`

jadhaj: Done

cfme/storage/volume.py
```diff
@@ -105,6 +105,7 @@ class VolumeAddForm(View):
     volume_name = TextInput(name='name')
     size = TextInput(name='size')
     tenant = BootstrapSelect(name='cloud_tenant_id')
+    volume_type = BootstrapSelect(name='volume_type')
```
digitronik: Optional:
Please order view elements if possible like...
```
storage_manager = BootstrapSelect(name='storage_manager_id')
tenant = BootstrapSelect(name='cloud_tenant_id')
volume_name = TextInput(name='name')
volume_type = BootstrapSelect(name='volume_type')
size = TextInput(name='size')
```
It will help in collective fill.

jadhaj: Done

digitronik: awesome thanks :+1:

PR #8056, opened 2018-10-31
[1LP][RFR] Add pytest.mark.tier(3) for all tests in test_embedded_ansible_automate

john-dupuy:
Add tier metadata for test_alert_run_ansible_playbook so that it matches tier on Polarion.


digitronik:
@john-dupuy can you please check others tests on the same file? Also if possible set global tire.

PR #8057, opened 2018-10-31
[RFR] Fix sprout orphans

mshriver:
Exception when this is run by celery-beat against cm-pod1/2 since their `list_vms` mgmt function returns a list of names instead of objects.

Continuing to debug orphan behavior, just small change to consider on its own.

```
2018-10-31 17:13:37,547 [ERROR] 'str' object has no attribute 'name'
  File "/opt/sprout/integration_tests/sprout/appliances/tasks.py", line 125, in wrapped_task
    return task(self, *args, **kwargs)
  File "/opt/sprout/integration_tests/sprout/appliances/tasks.py", line 1985, in synchronize_untracked_vms_in_provider
    if Appliance.objects.filter(name=vm.name, template__provider=provider).count() != 0:
AttributeError: 'str' object has no attribute 'name'
```

### DIFF
--- a/sprout/appliances/tasks.py
+++ b/sprout/appliances/tasks.py
@@ -1993,7 +1993,10 @@ def synchronize_untracked_vms_in_provider(self, provider_id):
         # This provider does not have VMs
         return
     for vm in sorted(provider_api.list_vms()):
-        if Appliance.objects.filter(name=vm.name, template__provider=provider).count() != 0:
+        if (
+            Appliance.objects.filter(name=getattr(vm, 'name', vm),
+                                     template__provider=provider).count() != 0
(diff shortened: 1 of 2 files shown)


sprout/appliances/tasks.py
```diff
@@ -1993,7 +1993,10 @@ def synchronize_untracked_vms_in_provider(self, provider_id):
         # This provider does not have VMs
         return
     for vm in sorted(provider_api.list_vms()):
-        if Appliance.objects.filter(name=vm.name, template__provider=provider).count() != 0:
+        if (
```
izapolsk: :+1:

PR #8066, opened 2018-11-02
[1LP][RFR] Fixed Assertion error in system_image_duplicate_name for 5.10z

ganeshhubale:
{{ pytest: cfme/tests/infrastructure/test_system_image_type.py -k 'test_system_image_duplicate_name_error_validation' -v }}

Fixed Error:
```
>               provision_type=SystemImageType.VM_OR_INSTANCE)
E           AssertionError: Pattern 'Name has already been taken' not found in 'assert_no_error: error: Name is not unique within region 0'

cfme/tests/infrastructure/test_system_image_type.py:37: AssertionError
AssertionError
Pattern 'Name has already been taken' not found in 'assert_no_error: error: Name is not unique within region 0'
```

### DIFF
--- a/cfme/tests/infrastructure/test_system_image_type.py
+++ b/cfme/tests/infrastructure/test_system_image_type.py
@@ -31,7 +31,12 @@ def test_system_image_duplicate_name_error_validation(appliance):
     sys_image_type = collection.create(
         name=name,
         provision_type=SystemImageType.VM_OR_INSTANCE)
-    with pytest.raises(Exception, match='Name has already been taken'):
+    error_message = (
+        "Name has already been taken"
+        if appliance.version < "5.10"
+        else "Name is not unique within region 0"
+    )
+    with pytest.raises(Exception, match=error_message):
         collection.create(
             name=name,
             provision_type=SystemImageType.VM_OR_INSTANCE)


cfme/tests/infrastructure/test_system_image_type.py
```diff
@@ -31,7 +31,12 @@ def test_system_image_duplicate_name_error_validation(appliance):
     sys_image_type = collection.create(
         name=name,
         provision_type=SystemImageType.VM_OR_INSTANCE)
-    with pytest.raises(Exception, match='Name has already been taken'):
+    error_message = (
```
jawatts: **Required** Have you thought about using `VersionPicker()` here. https://github.com/ManageIQ/integration_tests/blob/master/cfme/utils/version.py#L69

```
error-message = VersionPicker({
            LOWEST: "Name has already been taken",
            '5.10': '"Name is not unique within region 0"
        }).pick(self.appliance.version)
```

ganeshhubale: @jawatts 
I would like to know how will it effect tests performance with use of **VersionPicker** over **appliance.version**
Because yesterday I discussed with @digitronik as he said, It would be good to use **VersionPicker** in views but not in test.

jawatts: @ganeshhubale I was wrong, `VersionPicker`should be used when you do not have the appliance object. So you define the versionpicker object, and time comes for it to get picked,  its supposed to call pick on itself with the context's browser.version or appliance.version. 

Great question though!

PR #8067, opened 2018-11-02
[1LP][RFR]  Pre and post migration tests

Yadnyawalkya:
{{pytest: cfme/tests/v2v/test_v2v_ansible.py --use-provider vsphere65-nested --use-provider rhv42 --provider-limit 2 -vvvv}}

### DIFF
--- a/cfme/fixtures/ansible_fixtures.py
+++ b/cfme/fixtures/ansible_fixtures.py
@@ -22,7 +22,7 @@ def ansible_repository(appliance, wait_for_ansible):
     repositories = appliance.collections.ansible_repositories
     repository = repositories.create(
         name=fauxfactory.gen_alpha(),
-        url=cfme_data.ansible_links.repositories.embedded_ansible,
+        url=cfme_data.ansible_links.playbook_repositories.embedded_ansible,
         description=fauxfactory.gen_alpha())
     view = navigate_to(repository, "Details")
     if appliance.version < "5.9":
--- a/cfme/fixtures/ansible_fixtures.py
+++ b/cfme/fixtures/ansible_fixtures.py
@@ -5,6 +5,7 @@ from cfme.services.catalogs.catalog import Catalog
 from cfme.services.service_catalogs import ServiceCatalogs
 from cfme.services.myservice import MyService
 from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.conf import cfme_data
 from cfme.utils.wait import wait_for
 
 
@@ -21,7 +22,7 @@ def ansible_repository(appliance, wait_for_ansible):
     repositories = appliance.collections.ansible_repositories
     repository = repositories.create(
         name=fauxfactory.gen_alpha(),
-        url="https://github.com/quarckster/ansible_playbooks",
+        url=cfme_data.ansible_links.repositories.embedded_ansible,
         description=fauxfactory.gen_alpha())
     view = navigate_to(repository, "Details")
     if appliance.version < "5.9":
--- a/cfme/fixtures/ansible_fixtures.py
+++ b/cfme/fixtures/ansible_fixtures.py
@@ -20,10 +20,13 @@ def wait_for_ansible(appliance):
 @pytest.fixture(scope="module")
 def ansible_repository(appliance, wait_for_ansible):
     repositories = appliance.collections.ansible_repositories
-    repository = repositories.create(
-        name=fauxfactory.gen_alpha(),
-        url=cfme_data.ansible_links.playbook_repositories.embedded_ansible,
-        description=fauxfactory.gen_alpha())
+    try:
+        repository = repositories.create(
+            name=fauxfactory.gen_alpha(),
+            url=cfme_data.ansible_links.playbook_repositories.embedded_ansible,
+            description=fauxfactory.gen_alpha())
+    except KeyError:
+        pytest.skip("Skipping since no such key found in yaml")
     view = navigate_to(repository, "Details")
     if appliance.version < "5.9":
         refresh = view.browser.refresh
--- a/cfme/fixtures/cli.py
+++ b/cfme/fixtures/cli.py
@@ -344,7 +344,7 @@ def ansible_repository(appliance):
     repositories = appliance.collections.ansible_repositories
     repository = repositories.create(
         name=fauxfactory.gen_alpha(),
-        url=cfme_data.ansible_links.repositories.embedded_ansible,
+        url=cfme_data.ansible_links.playbook_repositories.embedded_ansible,
         description=fauxfactory.gen_alpha()
     )
     view = navigate_to(repository, "Details")
--- a/cfme/fixtures/cli.py
+++ b/cfme/fixtures/cli.py
@@ -342,11 +342,14 @@ def appliance_with_providers(appliance_preupdate):
 @pytest.fixture(scope="module")
 def ansible_repository(appliance):
     repositories = appliance.collections.ansible_repositories
-    repository = repositories.create(
-        name=fauxfactory.gen_alpha(),
-        url=cfme_data.ansible_links.playbook_repositories.embedded_ansible,
-        description=fauxfactory.gen_alpha()
-    )
+    try:
+        repository = repositories.create(
+            name=fauxfactory.gen_alpha(),
+            url=cfme_data.ansible_links.playbook_repositories.embedded_ansible,
+            description=fauxfactory.gen_alpha()
+        )
+    except KeyError:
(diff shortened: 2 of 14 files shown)


Yadnyawalkya:
@digitronik  can you start reviewing this. 5.9 passed, meanwhile I will test this out on 5.10.0.24 manually.
Failure with 5.10.0.25 on PRT is expected since devel changed migration UI location and till not I am not aware of where it is, searching :laughing:  

I will update 5.10 code and get back to you in while.

cfme/tests/v2v/test_v2v_ansible.py
```diff
@@ -0,0 +1,195 @@
...
+)
+def test_migration_playbooks(request, appliance, v2v_providers, host_creds, conversion_tags,
+                             wait_for_ansible, ansible_repository, credentials_collection,
+                             form_data_vm_obj_single_datastore):
+    """Test for migrating vms with pre and post playbooks"""
+    CREDENTIALS = (
+        "Machine",
+        {
+            "username": credentials[
+                v2v_providers.vmware_provider.data.templates.get("rhel7_minimal").creds
```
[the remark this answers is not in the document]
Yadnyawalkya: Good idea! correcting!

cfme/tests/v2v/test_v2v_ansible.py
```diff
@@ -0,0 +1,195 @@
+import fauxfactory
+import pytest
+
+from cfme import test_requirements
+from cfme.fixtures.provider import rhel7_minimal
```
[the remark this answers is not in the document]
Yadnyawalkya: so no, I tried that thing and surprisingly it is not working like that, we need to import this.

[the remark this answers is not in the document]
Yadnyawalkya: > I don't think you need to import this... its `global` registered fixture.

so no, I tried that thing and surprisingly it is not working like that, so we need to import this.

Yadnyawalkya:
@digitronik added 1LP changes, 5.9 passed on PRT, 5.10.0.24 also passing locally after #8194 fix, it should not pass 5.10.0.25 since migration options in not in UI. https://gitter.im/ManageIQ/v2v?at=5bf6d72f97a8982b9a498582

cfme/tests/v2v/test_v2v_ansible.py
```diff
@@ -0,0 +1,181 @@
...
+
+@pytest.fixture(scope="module")
+def wait_for_ansible(appliance):
+    appliance.wait_for_embedded_ansible()
+
+
+@pytest.fixture(scope="module")
+def credentials_collection(appliance):
+    """Fixture to add machine credentials"""
+    return appliance.collections.ansible_credentials
```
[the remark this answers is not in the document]
Yadnyawalkya: There are two more tests we are going to add here like `test_ansible_log_download` so I thought may be I can add fixture and use that in other future test functions. but agree @mshriver, I can remove this and add it directly in function instead of using it as a fixture.

cfme/tests/v2v/test_v2v_ansible.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+@pytest.fixture(scope="module")
+def ansible_repository(appliance):
```
mshriver: Since this depends on `wait_for_ansible`, that fixture should be a parameter here.  They're both module so it should be simple.  That way the dependency is built in, and the test won't fail if the fixture order changes in the test definition.

Yadnyawalkya:
PR657 need to be merge to get all ansible repo into cfme_data.yaml, want to wait till that happens.

Yadnyawalkya:
PR657 (cfme-qe-yamls) need to be merge to get all ansible repo into cfme_data.yaml, like to wait till that happens.

cfme/fixtures/ansible_fixtures.py
```diff
@@ -21,7 +22,7 @@ def ansible_repository(appliance, wait_for_ansible):
     repositories = appliance.collections.ansible_repositories
     repository = repositories.create(
         name=fauxfactory.gen_alpha(),
-        url="https://github.com/quarckster/ansible_playbooks",
+        url=cfme_data.ansible_links.playbook_repositories.embedded_ansible,
```
mshriver: Now that this is coming from a yaml lookup, it needs to be wrapped in a try block to skip the test when this is missing, instead of raising KeyError/AttributeError.

cfme/fixtures/cli.py
```diff
@@ -344,7 +343,9 @@ def appliance_with_providers(appliance_preupdate):
 def ansible_repository(appliance):
     repositories = appliance.collections.ansible_repositories
     repository = repositories.create(
-        fauxfactory.gen_alpha(), REPOSITORIES[0], description=fauxfactory.gen_alpha()
+        name=fauxfactory.gen_alpha(),
+        url=cfme_data.ansible_links.playbook_repositories.embedded_ansible,
```
mshriver: This needs to be wrapped to capture any exceptions on yaml value lookup.

cfme/tests/ansible/test_embedded_ansible_actions.py
```diff
@@ -37,7 +37,7 @@ def ansible_repository(appliance, wait_for_ansible):
     repositories = appliance.collections.ansible_repositories
     repository = repositories.create(
         name=fauxfactory.gen_alpha(),
-        url="https://github.com/quarckster/ansible_playbooks",
+        url=cfme_data.ansible_links.playbook_repositories.embedded_ansible,
```
mshriver: This needs to be wrapped to capture any exceptions on yaml value lookup.

cfme/tests/ansible/test_embedded_ansible_automate.py
```diff
@@ -42,7 +42,7 @@ def ansible_repository(appliance, wait_for_ansible):
     repositories = appliance.collections.ansible_repositories
     repository = repositories.create(
         name=fauxfactory.gen_alpha(),
-        url="https://github.com/quarckster/ansible_playbooks",
+        url=cfme_data.ansible_links.playbook_repositories.embedded_ansible,
```
mshriver: This needs to be wrapped to capture any exceptions on yaml value lookup.

cfme/tests/ansible/test_embedded_ansible_basic.py
```diff
@@ -114,8 +115,8 @@
 ]
 
 REPOSITORIES = [
-    "https://github.com/quarckster/ansible_playbooks",
-    "https://github.com/patchkez/ansible_playbooks"
+    cfme_data.ansible_links.playbook_repositories.embedded_ansible,
```
mshriver: This needs to be wrapped to capture any exceptions on yaml value lookup.

cfme/tests/ansible/test_embedded_ansible_services.py
```diff
@@ -45,7 +46,7 @@ def ansible_repository(appliance, wait_for_ansible):
     repositories = appliance.collections.ansible_repositories
     repository = repositories.create(
         name=fauxfactory.gen_alpha(),
-        url="https://github.com/quarckster/ansible_playbooks",
+        url=cfme_data.ansible_links.playbook_repositories.embedded_ansible,
```
mshriver: This needs to be .... you get the idea :D

cfme/v2v/migrations.py
```diff
@@ -373,10 +372,18 @@ def plan_in_progress(self, plan_name):
...
-                    "time elapsed for migration: %s",
-                    plan_name, is_plan_visible,
-                    plan_time_elapsed)
+                try:
+                    plan_time_elapsed = self.progress_card.get_clock(plan_name)
+                    new_msg = "time elapsed for migration: {time}".format(time=plan_time_elapsed)
+                except NoSuchElementException:
+                    new_msg = "playbook is executing.."
+                    pass
+                msg = "For plan {plan_name}, is plan in progress: {visibility}, {message}".format(
```
mshriver: **Optional**

No need to create new var here, its only used once in the log message.

cfme/v2v/migrations.py
```diff
@@ -712,14 +726,16 @@ class MigrationPlanCollection(BaseCollection):
     ENTITY = MigrationPlan
 
     def create(self, name, infra_map, vm_list, description=None, csv_import=False,
-               start_migration=False):
+               pre_playbook=False, post_playbook=False, start_migration=False):
```
mshriver: I think `None` is a more appropriate default here.

Yadnyawalkya: uhmm..let me think..so yes we are just checking in condition so `None` also will be good in this condition

PR #8068, opened 2018-11-02
[1LP][NOTEST][WIPTEST] Cleanup snapshots

mmojzis:
Added delete_snapshots method to ec2cleanup script.
Depends on this wrapanapi PR: https://github.com/ManageIQ/wrapanapi/pull/342/files
Tested.

### DIFF
--- a/requirements/frozen_docs.py2.txt
+++ b/requirements/frozen_docs.py2.txt
@@ -172,6 +172,7 @@ Werkzeug==0.14.1
 widgetastic.core==0.30.2
 widgetastic.patternfly==0.0.39
 widgetsnbextension==3.4.2
+wrapanapi==3.0.31
 wrapt==1.10.11
 xmltodict==0.11.0
 yaycl==0.3.0
--- a/scripts/ec2cleanup.py
+++ b/scripts/ec2cleanup.py
@@ -23,6 +23,8 @@ def parse_cmd_line():
                         help='List of ENIs, which should be excluded. ENI ID is allowed.')
     parser.add_argument('--exclude_stacks', nargs='+',
                        help='List of Stacks, which should be excluded')
+    parser.add_argument('--exclude_snapshots', nargs='+',
+                        help='List of snapshots, which should be excluded. Snapshot ID is allowed')
     parser.add_argument('--stack-template',
                         help='Stack name template to be removed', default="test", type=str)
     parser.add_argument("--output", dest="output", help="target file name, default "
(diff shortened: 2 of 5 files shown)


mshriver:
Please include an update to wrapanapi 3.0.31 to pull in the required methods.

scripts/ec2cleanup.py
```diff
@@ -164,10 +166,34 @@ def delete_stacks(provider_mgmt, excluded_stacks, stack_template, output):
...
+    snapshot_list = []
+    provider_name = provider_mgmt.kwargs['name']
+    try:
+        for snapshot in provider_mgmt.list_own_snapshots():
+            snapshot_id = snapshot.get("SnapshotId")
+            if excluded_snapshots and snapshot_id in excluded_snapshots:
+                logger.info(" Excluding Snapshot with id: %r", snapshot_id)
+                continue
+            else:
+                if provider_mgmt.delete_snapshot(snapshot_id=snapshot_id):
```
mshriver: Are there any snapshot ID's we'll want to exclude by default? this appears to delete all snapshots owned by 'self' on ec2

mmojzis: I don't think there are any snapshots that we will want to exclude. I think we should rather make AMIs from snapshots that we will want to use in the future, but I added this exclude option because it's used in all other methods so I wanted to make it unified.

scripts/ec2cleanup.py
```diff
@@ -189,6 +215,10 @@ def ec2cleanup(exclude_volumes, exclude_eips, exclude_elbs, exclude_enis, exclud
                       excluded_stacks=exclude_stacks,
                       stack_template=stack_template,
                       output=output)
+        logger.info("Deleting snapshots...")
+        delete_snapshots(provider_mgmt=provider_mgmt,
```
izapolsk: Q: shouldn't your methods pass something to sys.exit ? 
what's the reason for having sys.exit if nothing is passed there ?

izapolsk:
please don't forget about wrapanapi requirement.

requirements/frozen_docs.py2.txt
```diff
@@ -172,6 +172,7 @@ Werkzeug==0.14.1
 widgetastic.core==0.30.2
 widgetastic.patternfly==0.0.39
 widgetsnbextension==3.4.2
+wrapanapi==3.0.31
```
mshriver: have to update py3 frozen files as well

mmojzis: How can I do that? By running freeze_all.py in python3? So I have to download all the requirements for python3?

mshriver: Correct, create a py3 virtualenv ( can use quickstart for it if you want) and activate it, then run freeze_all.

mmojzis: Done

PR #8072, opened 2018-11-02
[RFR] Fix arg name in requirements/freeze.py

mshriver:
L64 passes this as args.keep, opting to set that name here via dest instead of changing L64 to use keep_venv

### DIFF
--- a/requirements/freeze.py
+++ b/requirements/freeze.py
@@ -18,7 +18,7 @@ os.environ.pop("PYTHONPATH", "")
 
 parser = argparse.ArgumentParser(description=__doc__.strip())
 parser.add_argument("--venv", default=None)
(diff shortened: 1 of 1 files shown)


requirements/freeze.py
```diff
@@ -18,7 +18,7 @@
 
 parser = argparse.ArgumentParser(description=__doc__.strip())
 parser.add_argument("--venv", default=None)
-parser.add_argument("--keep-venv", action="store_true")
+parser.add_argument("--keep-venv", action="store_true", dest='keep')
```
mshriver: L64 passes this as `args.keep`, opting to set that name here via `dest` instead of changing L64 to use `keep_venv`.

PR #8080, opened 2018-11-05
[1LP][RFR] Test for Quota with Catalog_bundle: Verifying quota limit by ordering catalog bundle over the assigned quota 

ganeshhubale:
- This test case verifies the quota assigned by automation method for user and group
   is working correctly for the infra providers by ordering catalog bundle.
- Added new fixture
    - **catalog bundle()**
        - **catalog_item_list** : this is used to add more than one catalog items to list and pass them to catalog bundle.
        - Here in this test, only one catalog item is created and it is passed to add in catalog bundle.
        - catalog_item_list is taken because list of catalog items are accepted by catalog bundle's create method

{{ pytest: cfme/tests/infrastructure/test_quota_tagging.py -k 'test_quota_catalog_bundle_infra' -v }}

### DIFF
--- a/cfme/services/catalogs/catalog_items/catalog_bundles.py
+++ b/cfme/services/catalogs/catalog_items/catalog_bundles.py
@@ -103,7 +103,7 @@ class CatalogBundlesCollection(BaseCollection):
                 "StateMachines", "ServiceProvision_Template", "CatalogItemInitialization")
             view.apply_button.click()
         for cat_item in catalog_items:
-            view.resources.fill({'select_resource': cat_item.name})
+            view.resources.fill({'select_resource': cat_item})
         view.add_button.click()
         view.flash.assert_success_message('Catalog Bundle "{}" was added'.format(name))
         view = self.create_view(AllCatalogItemView)
--- a/cfme/services/catalogs/catalog_items/catalog_bundles.py
+++ b/cfme/services/catalogs/catalog_items/catalog_bundles.py
@@ -103,7 +103,7 @@ class CatalogBundlesCollection(BaseCollection):
                 "StateMachines", "ServiceProvision_Template", "CatalogItemInitialization")
             view.apply_button.click()
         for cat_item in catalog_items:
-            view.resources.fill({'select_resource': cat_item})
+            view.resources.fill({'select_resource': cat_item.name})
         view.add_button.click()
         view.flash.assert_success_message('Catalog Bundle "{}" was added'.format(name))
         view = self.create_view(AllCatalogItemView)
--- a/cfme/tests/infrastructure/test_quota_tagging.py
+++ b/cfme/tests/infrastructure/test_quota_tagging.py
@@ -87,6 +87,23 @@ def catalog_item(appliance, provider, dialog, catalog, prov_data):
         catalog_item.delete()
 
 
+@pytest.fixture
+def catalog_bundle(appliance, dialog, catalog, catalog_item):
+    catalog_item_list = []
+    catalog_item_list.append(catalog_item)
+    collection = appliance.collections.catalog_bundles
+    catalog_bundle = collection.create(name='test_{}'.format(fauxfactory.gen_alphanumeric()),
+                                       catalog_items=catalog_item_list,
+                                       description='test catalog bundle',
+                                       display_in=True,
+                                       catalog=catalog,
+                                       dialog=dialog)
+
+    yield catalog_bundle
+    if catalog_bundle.exists:
+        catalog_bundle.delete()
+
+
 @pytest.fixture(scope='module')
 def max_quota_test_instance(appliance, test_domain):
     miq = appliance.collections.domains.instantiate('ManageIQ')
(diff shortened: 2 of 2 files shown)


cfme/services/catalogs/catalog_items/catalog_bundles.py
```diff
@@ -103,7 +103,7 @@ def create(self, name, catalog_items=None, catalog=None, description=None, displ
                 "StateMachines", "ServiceProvision_Template", "CatalogItemInitialization")
             view.apply_button.click()
         for cat_item in catalog_items:
-            view.resources.fill({'select_resource': cat_item})
+            view.resources.fill({'select_resource': cat_item.name})
```
sshveta: Required :
This will break the services bundle tests .
You have to pass the name from your tests 
Ex : https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/services/test_generic_service_catalogs.py#L82

cfme/tests/infrastructure/test_quota_tagging.py
```diff
@@ -87,6 +87,23 @@ def catalog_item(appliance, provider, dialog, catalog, prov_data):
         catalog_item.delete()
 
 
+@pytest.fixture
+def catalog_bundle(appliance, dialog, catalog, catalog_item):
+    catalog_item_list = []
+    catalog_item_list.append(catalog_item)
+    collection = appliance.collections.catalog_bundles
+    catalog_bundle = collection.create(name='test_{}'.format(fauxfactory.gen_alphanumeric()),
+                                       catalog_items=catalog_item_list,
```
sshveta: catalog_items=[catalog_item.name]
No  need for  catalog_item_list = [] if there is only one catalog item

cfme/tests/infrastructure/test_quota_tagging.py
```diff
@@ -317,3 +334,44 @@ def test_quota_infra(request, appliance, provider, setup_provider, admin_email,
...
     request.addfinalizer(provision_request.remove_request)
     assert provision_request.row.reason.text == "Quota Exceeded"
+
+
+@pytest.mark.parametrize('context', [ViaSSUI, ViaUI])
+@pytest.mark.parametrize(
+    ['custom_prov_data'],
+    [
+        [{'hardware': {'memory': '4096'}}],
+        [{}],
```
sshveta: Question : 
What is the purpose of this empty array ?

cfme/tests/infrastructure/test_quota_tagging.py
```diff
@@ -317,3 +334,44 @@ def test_quota_infra(request, appliance, provider, setup_provider, admin_email,
...
+    """
+    prov_data.update(custom_prov_data)
+    with appliance.context.use(context):
+        service_catalogs = ServiceCatalogs(appliance, catalog_bundle.catalog, catalog_bundle.name)
+        if context is ViaSSUI:
+            service_catalogs.add_to_shopping_cart()
+        service_catalogs.order()
+    # nav to requests page to check quota validation
+    request_description = 'Provisioning Service [{0}] from [{0}]'.format(catalog_bundle.name)
+    provision_request = appliance.collections.requests.instantiate(request_description)
```
sshveta: Please check if this request part will work from SSUI.
There is no navigation to request page in SUI .

sshveta:
Please run this test from SUI context as well.

cfme/tests/infrastructure/test_quota_tagging.py
ganeshhubale: ```[{}],``` - This parameterization is for selecting storage while provisioning VM/instance. But we don't need to select storage because by default storage is already more than assigned storage quota which is 2GB maximum.

cfme/tests/infrastructure/test_quota_tagging.py
ganeshhubale: This test case orders the service_catalog from SSUI and checks in UI that ordered request is exceeded quota or not.

ganeshhubale:
@sshveta  
I have updated this PR with requested changes.
Please take a look at it and let me know if you have any suggestions.
Thank you.

cfme/tests/infrastructure/test_quota_tagging.py
sshveta: :+1:

cfme/tests/infrastructure/test_quota_tagging.py
mshriver: This logic needs to be in an in-line comment, there are a few assumptions in that statement that could breakdown as the application goes through changes.  Someone else trying to maintain this test won't be aware of the assumptions made.  Please either explicitly parametrize on storage of sufficient size, or include in-line comment that this parameter item is for storage, and the assumptions about default sizes and quota settings.

ganeshhubale: @mshriver 
Thanks for your suggestion. It is really useful for someone new. I have added in-line comment for empty array as below.  Let me know if you have any suggestions around this comment.

> [{}],  # This parameterization is for selecting storage while provisioning VM.
        # But it is not possible to parameterize storage size.
        # Becuase it is defined in disk formats(Thin, Thick, Default and other types).
        # Also it varies with providers.
        # But we don't need to select storage because by default storage is already more than
        #  assigned storage quota which is 2GB maximum.

PR #8081, opened 2018-11-05
[1LP][RFR] Refarctor Nuage fixture

aljazkosir:
Refactor Nuage fixture so we have an option to reuse it with different test configuration.

/cc @miha-plesko

### DIFF
--- a/cfme/fixtures/nuage.py
+++ b/cfme/fixtures/nuage.py
@@ -5,10 +5,8 @@ from wrapanapi.utils.random import random_name
 from cfme.utils.log import logger
 
 
-@pytest.fixture
-def with_nuage_sandbox(networks_provider):
-    nuage = networks_provider.mgmt
-    sandbox = box = {}
+def create_basic_sandbox(nuage):
+    box = {}
 
     # Create empty enterprise aka 'sandbox'.
     enterprise = box['enterprise'] = nuage.create_enterprise()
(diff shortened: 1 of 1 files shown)


cfme/fixtures/nuage.py
```diff
@@ -45,9 +44,18 @@ def with_nuage_sandbox(networks_provider):
     box['l2_group'] = box['l2_domain'].create_child(
         nuage.vspk.NUPolicyGroup(name=random_name()))[0]
 
+    return box
+
+
+@pytest.fixture
+def with_nuage_sandbox(networks_provider):
+    nuage = networks_provider.mgmt
+    sandbox = create_nuage_sandbox(networks_provider)
```
miha-plesko: Please pass `nuage` handle directly to the `create_nuage_sandbox` function instead passing `networks_provider` fixture. We want the auxilary function be as pure as possible.

cfme/fixtures/nuage.py
```diff
@@ -5,10 +5,9 @@
 from cfme.utils.log import logger
 
 
-@pytest.fixture
-def with_nuage_sandbox(networks_provider):
+def create_nuage_sandbox(networks_provider):
```
miha-plesko: May I suggest to name it `create_basic_sandbox` instead? In the future we plan to come up with an extended sandbox as well (with Nuage templates and all) where we'll want to first create this basic setup and then extend it.

miha-plesko: Also, please modify the method to accept `nuage` as parameter, as suggested by the comment below.

aljazkosir: I modified both function name and input parameter as suggested.

PR #8083, opened 2018-11-05
[WIP] This PR will fix set_hostname test and updates datetime test.

sbulage:
PRT Run:
{{ pytest: cfme/tests/cli/test_appliance_cli.py -k  test_appliance_console_cli_set_hostname -qsvvvv }}

### DIFF
--- a/cfme/tests/cli/test_appliance_cli.py
+++ b/cfme/tests/cli/test_appliance_cli.py
@@ -30,13 +30,12 @@ requires_59 = pytest.mark.uncollectif(
 
 
 @requires_59
-def test_appliance_console_cli_datetime(temp_appliance_preconfig_funcscope):
+def test_appliance_console_cli_datetime(appliance):
     """Grab fresh appliance and set time and date through appliance_console_cli and check result"""
-    app = temp_appliance_preconfig_funcscope
-    app.ssh_client.run_command("appliance_console_cli --datetime 2020-10-20T09:59:00")
+    appliance.ssh_client.run_command("appliance_console_cli --datetime 2020-10-20T09:59:00")
 
     def date_changed():
-        return app.ssh_client.run_command("date +%F-%T | grep 2020-10-20-10:00").success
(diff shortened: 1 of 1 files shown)


cfme/tests/cli/test_appliance_cli.py
```diff
@@ -30,13 +30,12 @@
 
 
 @requires_59
-def test_appliance_console_cli_datetime(temp_appliance_preconfig_funcscope):
+def test_appliance_console_cli_datetime(appliance):
```
digitronik: :+1:

izapolsk:
@sbulage, please explain your changes. why appliance instead of that temporary func scope appliance ?

sbulage:
@izapolsk temporary func scope appliance is waiting from Sprout to get the appliance, which sometimes to tend to failure.

cfme/tests/cli/test_appliance_cli.py
```diff
@@ -30,13 +30,12 @@
...
     """Grab fresh appliance and set time and date through appliance_console_cli and check result"""
-    app = temp_appliance_preconfig_funcscope
-    app.ssh_client.run_command("appliance_console_cli --datetime 2020-10-20T09:59:00")
+    appliance.ssh_client.run_command("appliance_console_cli --datetime 2020-10-20T09:59:00")
```
mshriver: This test isn't resetting the datetime configuration, which is why it was using temp_appliance_preconfig_funcscope.  If that fixture is failing to pull a sprout appliance, we need to resolve that. I think using appliance here will have negative impacts on tests that follow.

mshriver: Not sure where hostname assertion comes into play here in this test function?  I see in the test below hostname is used changed and asserted against - but again isn't reset after the test is done, because the test was running on a one-off appliance.  We need to resolve the core failures here when using one-off appliances, or need to reliably reset the changes made in these tests.

sbulage:
set_hostname is working properly Closing PR.

PR #8084, opened 2018-11-05
[1LP][RFR] Update Alert docstring and add ability to create alert with only a 'description'

john-dupuy:
Purpose or Intent
=================
__Extending__  the `alert_collection.create()` method to work when only passed a description. The `create()` method now defaults to have `timeline_event=True` and `driving_event='Hourly Timer'`. The latter is selected from the dropdown menu on the `AlertCollection` "Add" page. 

__Updating__ docstring for `Alert` to reflect the proper way to create new alerts; via `alert_collection.create()` method

{{ pytest: --long-running cfme/tests/control/test_basic.py::test_alert_crud }}

### DIFF
--- a/cfme/control/explorer/alerts.py
+++ b/cfme/control/explorer/alerts.py
@@ -301,15 +301,16 @@ class AlertCollection(BaseCollection):
             Note: Since alerts in CFME require a description, driving_event, and one
             of snmp_trap, emails, timeline_event, or mgmt_event, we set defaults to
             timeline_event=True, and driving_event=<first_item_from_dropdown>
+            We select the first item from the dropdown for efficiency.
 
             This allows creation of an alert only by a description. e.g.
-            >>> alert = appliance.collection.alerts.create('my_alert_description')
+            >>> alert = appliance.collections.alerts.create('my_alert_description')
             >>> alert.delete()
         """
         view = navigate_to(self,"Add")
         if driving_event is None:
-            view.driving_event.select_by_visible_text(partial_match('H'))
-            driving_event = view.driving_event.read()
+            driving_event = view.driving_event.all_options[1].text
+        # instantiate the alert
         alert = self.instantiate(description, severity=severity, active=active, based_on=based_on,
             evaluate=evaluate, driving_event=driving_event,
             notification_frequency=notification_frequency, snmp_trap=snmp_trap, emails=emails,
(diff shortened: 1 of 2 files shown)


cfme/control/explorer/alerts.py
```diff
@@ -291,12 +295,28 @@ class AlertCollection(BaseCollection):
...
+            of snmp_trap, emails, timeline_event, or mgmt_event, we set defaults to
+            timeline_event=True, and driving_event=<item_from_dropdown>
+
+            This allows creation of an alert only by a description. e.g.
+            >>> alert = appliance.collections.alerts.create('my_alert_description')
+            >>> alert.delete()
+        """
+        view = navigate_to(self,"Add")
+        if driving_event is None:
+            view.driving_event.select_by_visible_text(partial_match('H'))
```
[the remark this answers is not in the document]
john-dupuy: @digitronik you mean replacing `H` with `Hourly`? The idea here is that we will be able to match at least one of those driving events if the text changes in the dropdown menu. Right now, this is selecting `Hourly Timer` as the default driving event, but if you think it is more clear to put in `Hourly`, that's fine by me.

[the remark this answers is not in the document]
digitronik: Yes, matching `Hourly` is better than only `H` character.

cfme/control/explorer/alerts.py
```diff
@@ -291,12 +295,28 @@ class AlertCollection(BaseCollection):
...
+            of snmp_trap, emails, timeline_event, or mgmt_event, we set defaults to
+            timeline_event=True, and driving_event=<item_from_dropdown>
+
+            This allows creation of an alert only by a description. e.g.
+            >>> alert = appliance.collections.alerts.create('my_alert_description')
+            >>> alert.delete()
+        """
+        view = navigate_to(self,"Add")
+        if driving_event is None:
+            view.driving_event.select_by_visible_text(partial_match('Hourly'))
```
mshriver: In-line comment as to why hourly is selected here

mshriver: Consider using `all_options` instead of actually selecting and then reading to get the full text.

john-dupuy: See recent update -- I select the first item from the `all_options` now

PR #8089, opened 2018-11-06
[1LP][RFR] Fix myservice ssui crud

niyazRedhat:
{{pytest: cfme/tests/ssui/test_ssui_myservice.py::test_myservice_crud --use-provider vsphere67-nested -svvv }}

Due to this [1] bug I am replacing cost center 001 to 002 
[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1646925

### DIFF
--- a/cfme/tests/ssui/test_ssui_myservice.py
+++ b/cfme/tests/ssui/test_ssui_myservice.py
@@ -33,7 +33,6 @@ pytestmark = [
 
 @pytest.mark.rhel_testing
 @pytest.mark.rhv1
-@pytest.mark.meta(blockers=[BZ(1646925, forced_streams=['5.10'])])
 @pytest.mark.parametrize('context', [ViaSSUI])
 def test_myservice_crud(appliance, setup_provider, context, order_service):
     """Test Myservice crud in SSUI.
@@ -46,8 +45,7 @@ def test_myservice_crud(appliance, setup_provider, context, order_service):
         my_service = MyService(appliance, catalog_item.name)
         my_service.set_ownership("Administrator", "EvmGroup-approver")
         my_service.update({'description': '{}_edited'.format(catalog_item.name)})
-        if appliance.version > "5.8":
-            my_service.edit_tags("Cost Center", "Cost Center 001")
+        my_service.edit_tags("Cost Center", "Cost Center 002")
         my_service.delete()
 
 


cfme/tests/ssui/test_ssui_myservice.py
```diff
@@ -48,7 +48,7 @@ def test_myservice_crud(appliance, setup_provider, context, order_service):
         my_service.set_ownership("Administrator", "EvmGroup-approver")
         my_service.update({'description': '{}_edited'.format(catalog_item.name)})
         if appliance.version > "5.8":
-            my_service.edit_tags("Cost Center", "Cost Center 001")
+            my_service.edit_tags("Cost Center", "Cost Center 002")
```
digitronik: If possible remove above 5.8 condition

niyazRedhat:
As this test expect either 001 or 002, i changed that to 002 so that this test shouldn't blocked by bugzilla.

cfme/tests/ssui/test_ssui_myservice.py
mshriver: @niyazRedhat bringing discussion about BZ impact here so we can have thread and not just PR comments.

Given: 1) don't want to block test on BZ, 2) does not really matter which cost center we pick for this test

I say we:
1) take @digitronik suggestion and remove the 5.8 block that this is in
2) add in-line comment about BZ impacting tag visibility
3) use wt.core.util.partial_match here, passing `partial_match('Cost Center')` to just pick the first visible cost center tag.  `edit_tags` method just passes this value to a `fill()` call, so this should work fine.

niyazRedhat: @mshriver I tried to add partial_match but it is not working on ssui edit tag form.
Bug will be fixed in a month. Now can we go for  ```Cost Center 002``` ?

Traceback
```ipdb> view.form.tag_name.fill(partial_match("Cost Center"))
*** ValueError: Cannot locate option with visible text: u'Cost Center'. Available options: u'Select a value to assign', u'Cost Center 002'
```

mshriver: I would like to address the failure of partial_match in this case. Do you have a python notebook you can share, or even just a set of miq shell commands to reproduce? The snippet you included there clearly includes 'Cost Center', so it should have matched.

PR #8091, opened 2018-11-07
[1LP][RFR]Catalog view is_displayed fix

sshveta:
1) Added wait for a view .
2) fixed is_displayed for a view 
3) Fixed test catalog permission test

### DIFF
--- a/cfme/services/catalogs/catalog.py
+++ b/cfme/services/catalogs/catalog.py
@@ -99,8 +99,7 @@ class Catalog(BaseEntity, Updateable, Pretty, Taggable):
         else:
             view.cancel_button.click()
         view = self.create_view(DetailsCatalogView, override=updates)
-        view.wait_displayed()
-        assert view.is_displayed
+        assert view.wait_displayed()
         view.flash.assert_no_error()
         if changed:
             view.flash.assert_message(
--- a/cfme/services/catalogs/catalog.py
+++ b/cfme/services/catalogs/catalog.py
@@ -8,9 +8,10 @@ from navmazing import NavigateToAttribute, NavigateToSibling
 
 from cfme.common import Taggable
 from cfme.modeling.base import BaseCollection, BaseEntity
-from cfme.utils.update import Updateable
-from cfme.utils.pretty import Pretty
 from cfme.utils.appliance.implementations.ui import navigator, CFMENavigateStep, navigate_to
+from cfme.utils.pretty import Pretty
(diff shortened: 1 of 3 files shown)


sshveta:
Failures are not related to changes made.

mshriver:
Part of resolving #7960

cfme/services/catalogs/catalog.py
```diff
@@ -98,6 +99,7 @@ def update(self, updates):
         else:
             view.cancel_button.click()
         view = self.create_view(DetailsCatalogView, override=updates)
+        wait_for(lambda: view.is_displayed, delay=0.5, num_sec=2)
```
mshriver: If the 0.2s default delay in wt.core.View.wait_displayed will work here, please use `view.wait_displayed` instead of wrapping it. Default timeout is 10s for wait_displayed.

mshriver: I've opened a wt.core enhancement to allow you to pass that delay to `wait_displayed`, and to assert against its return (so you can remove L101/103 below)

sshveta: OK will do mshriver

cfme/services/catalogs/catalog.py
```diff
@@ -98,6 +99,7 @@ def update(self, updates):
         else:
             view.cancel_button.click()
         view = self.create_view(DetailsCatalogView, override=updates)
+        view.wait_displayed()
```
jawatts: Lines 102 and 103 can be combined into a single line:
``` assert view.wait_displayed()```

`wait_displayed()` returns the boolean of `wait_for`

jawatts:
@sshveta could you look at why PRT is failing on 5.10

cfme/services/catalogs/catalog.py
jawatts: @sshveta I jumped the gun here, looks like that change hasn't been released yet :) @mshriver will be doing it after lunch

sshveta: OO I made the changes :)

sshveta: @jawatts : Failures are in test_access_control.py  since all the tests ran . Not related to changes made by me.

mshriver: wt.core 0.30.3 included with #8055, I've restarted PRT here as now wait_displayed should be returning boolean.

PR #8092, opened 2018-11-07
[1LP][RFR] Fix alert_timeline_cpu by making change for TimelinesChart class

john-dupuy:
Purpose or Intent
=================

__Fixing__ test_alert_timeline_cpu because currently it is failing in the `_prepare_event()` method of `TimelinesChart` class. The fix is to remove whitespace from `br.tail` and to add a `try` `except` clause around the `get_groups()` method.

{{ pytest: --long-running cfme/tests/control/test_alerts.py::test_alert_timeline_cpu }}

### DIFF
--- a/widgetastic_manageiq/__init__.py
+++ b/widgetastic_manageiq/__init__.py
@@ -2573,14 +2573,19 @@ class TimelinesChart(View):
         node = document_fromstring(evt)
         # lxml doesn't replace <br> with \n in this case. so this has to be done by us
         for br in node.xpath("*//br"):
-            br.tail = "\n" + br.tail if br.tail else "\n"
+            br.tail = "\n" + br.tail.rstrip() if br.tail and br.tail.rstrip() else "\n"
 
         # parsing event and preparing its attributes
         event = self.TimelinesEvent()
         for line in node.text_content().split("\n"):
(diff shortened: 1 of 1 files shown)


widgetastic_manageiq/__init__.py
```diff
@@ -2573,14 +2573,19 @@ def _prepare_event(self, evt, category):
         node = document_fromstring(evt)
         # lxml doesn't replace <br> with \n in this case. so this has to be done by us
         for br in node.xpath("*//br"):
-            br.tail = "\n" + br.tail if br.tail else "\n"
+            br.tail = "\n" + br.tail.rstrip() if br.tail and br.tail.rstrip() else "\n"
```
digitronik: @john-dupuy I don't think we need `br.tail and br.tail.rstrip()` only `br.tail.rstrip()` check is sufficient. what do you think?

john-dupuy: @digitronik Are you referring to the `if br.tail and br.tail.rstrip()` line? The problem is that `br.tail` is either a string or `None` and if we just have `if br.tail.rstrip()` we are trying to call a method that doesn't exist for a `None` object.

PR #8104, opened 2018-11-08
[1LP][RFR] fixed ProviderTemplatesView and removed BZ blocker

lina-nikiforova:
{{ pytest: -v cfme/tests/infrastructure/test_infra_tag_filters_combination.py::test_tagvis_tag_datacenter_combination }}

This PR fixes ProviderTemplatesView that affected 2 tests:
```
test_tagvis_tag_datacenter_combination[infra_templates-infra-not_visible]
test_tagvis_tag_datacenter_combination[infra_templates-infra-visible]
```

Also removed the BZ blocker.

### DIFF
--- a/cfme/common/provider_views.py
+++ b/cfme/common/provider_views.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 from lxml.html import document_fromstring
+from widgetastic.utils import Version
 
 from widgetastic_patternfly import BreadCrumb, Dropdown, BootstrapSelect
 from widgetastic.exceptions import NoSuchElementException
@@ -7,6 +8,7 @@ from widgetastic.widget import View, Text, ConditionalSwitchableView
 
 from cfme.base.login import BaseLoggedInPage
 from cfme.common.host_views import HostEntitiesView
+from cfme.utils.version import VersionPicker
 from widgetastic_manageiq import (
     ParametrizedSummaryTable,
     Button,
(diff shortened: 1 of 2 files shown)


cfme/common/provider_views.py
```diff
@@ -249,7 +249,10 @@ class ProviderTemplatesView(ProviderVmsTemplatesView):
 
     @property
     def is_displayed(self):
-        title = '{name} (All Miq Templates)'.format(name=self.context['object'].name)
+        if self.extra.appliance.version < '5.10':
```
mshriver: First use I've seen of the extra attribute!

mshriver: This looks like a good place to use `VersionPick`, since it is within a view method, and should pick itself when needed.

cfme/common/provider_views.py
```diff
@@ -249,15 +251,15 @@ class ProviderTemplatesView(ProviderVmsTemplatesView):
...
     def is_displayed(self):
-        if self.browser.product_version < '5.10':
-            title = '{name} (All VM Templates)'.format(name=self.context['object'].name)
-        else:
-            title = '{name} (All VM Templates and Images)'.format(name=self.context['object'].name)
-
+        title = VersionPicker({
+            Version.lowest(): '{name} (All VM Templates)'.format(name=self.context['object'].name),
+            '5.10': '{name} (All VM Templates and Images)'.format(name=self.context['object'].name),
+        }).pick(self.extra.appliance.version)
```
jawatts: **Question** Does this still work without calling `.pick` ? I am thinking it may not be needed here

lina-nikiforova: @jawatts I tried and it didn't. After that I added `.pick`

jawatts: @lina-nikiforova Thanks, I was curious!

PR #8107, opened 2018-11-08
[1LP][RFR]Service Link test fix

sshveta:
{{pytest: cfme/tests/services/test_myservice.py::test_service_link -v --long-running}}

ServiceVMDetailsView inherits VMDetailsEntities
which does not have navigation and other things so removed them and fixed object .

### DIFF
--- a/cfme/services/myservice/ui.py
+++ b/cfme/services/myservice/ui.py
@@ -186,12 +186,7 @@ class ReconfigureServiceView(SetOwnershipForm):
 class ServiceVMDetailsView(VMDetailsEntities):
     @property
     def is_displayed(self):
-        return (
-            self.logged_in_as_current_user and
-            self.navigation.currently_selected == ['Services', 'MyServices'] and
-            self.myservice.is_opened and
-            self.title.text == 'VM and Instance "{}"'.format(self.context['object'].name)
-        )
+        return self.title.text == 'VM and Instance "{}"'.format(self.context['object'].vm_name)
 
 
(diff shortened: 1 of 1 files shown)


cfme/services/myservice/ui.py
```diff
@@ -186,12 +186,7 @@ def is_displayed(self):
 class ServiceVMDetailsView(VMDetailsEntities):
     @property
     def is_displayed(self):
-        return (
-            self.logged_in_as_current_user and
-            self.navigation.currently_selected == ['Services', 'MyServices'] and
-            self.myservice.is_opened and
-            self.title.text == 'VM and Instance "{}"'.format(self.context['object'].name)
-        )
+        return self.title.text == 'VM and Instance "{}"'.format(self.context['object'].vm_name)
```
mshriver: Are you 100% sure this won't match other screens?  I'm thinking it would be better to add the necessary widgets here to the view since they aren't provided by `VMDetailsEntities`. Then you can still validate you're on MyServices and the title is matching.

sshveta: @mshriver : The VM details page anyways do not have the navigation selected which was being checked before as it leaves myservice page and navigates to VM details page .
SO even after adding necessary things to the view the only additional check that I will be able to do is 
self.logged_in_as_current_user. 

Do you still want me to make changes ?

mshriver: No, that's fine @sshveta, trusting your judgement :)

PR #8112, opened 2018-11-09
[1LP][RFR] Quota test for Infra: Testing tenant quota via UI and SSUI 

ganeshhubale:
- To test Tenant Quota in UI and SSUI by selecting field entry points.
- Quota has to be checked if it is working with field entry points also.

{{ pytest: cfme/tests/infrastructure/test_tenant_quota.py -k 'test_service_infra_tenant_quota_with_default_entry_point' --use-provider=rhv42 -v }}

Note:
This test is depend on PR #7903

### DIFF
--- a/cfme/tests/infrastructure/test_tenant_quota.py
+++ b/cfme/tests/infrastructure/test_tenant_quota.py
@@ -18,9 +18,29 @@ pytestmark = [
     test_requirements.quota,
     pytest.mark.meta(server_roles="+automate"),
     test_requirements.vm_migrate,
-    pytest.mark.usefixtures('setup_provider', 'uses_infra_providers'),
+    pytest.mark.usefixtures('setup_provider'),
     pytest.mark.provider([VMwareProvider, RHEVMProvider], scope="module", selector=ONE_PER_TYPE)
 ]
+
+
+@pytest.fixture
+def set_default(provider, request):
+    """This fixture is used to return paths for provisioning_entry_point, reconfigure_entry_point
+       and retirement_entry_point. The value of 'provisioning_entry_point' is required while
+       creating new catalog item in 'test_service_infra_tenant_quota_with_default_entry_point' test.
+       But other tests does not require these values since those tests takes default values hence
+       providing default value. So in this file, this fixture - 'set_default'
+       must be used in all tests of quota which are related to services where catalog item needs to
+       be created with specific values for these entries.
+    """
+    with_prov = (
+        "/ManageIQ (Locked)/{}/VM/Provisioning/StateMachines/ProvisionRequestApproval/Default"
+        .format(provider.string_name)
+    )
+    default = (
+        "/Service/Provisioning/StateMachines/ServiceProvision_Template/CatalogItemInitialization"
+    )
+    return with_prov if request.param else default
 
 
 @pytest.fixture
@@ -80,17 +100,21 @@ def set_roottenant_quota(request, roottenant, appliance):
     roottenant.set_quota(**{'{}_cb'.format(field): False})
 
 
-@pytest.fixture
-def catalog_item(appliance, provider, dialog, catalog, prov_data):
+@pytest.fixture(scope='function')
+def catalog_item(appliance, provider, dialog, catalog, prov_data, set_default):
     collection = appliance.collections.catalog_items
-    yield collection.create(
+    catalog_item = collection.create(
         provider.catalog_item_type,
         name='test_{}'.format(fauxfactory.gen_alphanumeric()),
         description='test catalog',
         display_in=True,
         catalog=catalog,
         dialog=dialog,
-        prov_data=prov_data)
+        prov_data=prov_data,
+        provider=provider,
+        provisioning_entry_point=set_default)
+    yield catalog_item
+    catalog_item.delete()
 
 
 @pytest.fixture(scope='module')
@@ -166,18 +190,19 @@ def test_tenant_quota_enforce_via_lifecycle_infra(appliance, provider, set_roott
 # indirect is the list where we define which fixtures are to be passed values indirectly.
 @pytest.mark.parametrize('context', [ViaSSUI, ViaUI])
 @pytest.mark.parametrize(
-    ['set_roottenant_quota', 'custom_prov_data', 'extra_msg'],
+    ['set_roottenant_quota', 'custom_prov_data', 'extra_msg', 'set_default'],
     [
-        [('cpu', '2'), {'hardware': {'num_sockets': '8'}}, ''],
-        [('storage', '0.01'), {}, ''],
-        [('memory', '2'), {'hardware': {'memory': '4096'}}, ''],
-        [('vm', '1'), {'catalog': {'num_vms': '4'}}, '###'],
+        [('cpu', '2'), {'hardware': {'num_sockets': '8'}}, '', False],
+        [('storage', '0.01'), {}, '', False],
+        [('memory', '2'), {'hardware': {'memory': '4096'}}, '', False],
+        [('vm', '1'), {'catalog': {'num_vms': '4'}}, '###', False],
     ],
-    indirect=['set_roottenant_quota', 'custom_prov_data'],
+    indirect=['set_roottenant_quota', 'custom_prov_data', 'set_default'],
     ids=['max_cpu', 'max_storage', 'max_memory', 'max_vms']
 )
 def test_tenant_quota_enforce_via_service_infra(request, appliance, context, set_roottenant_quota,
-                                                extra_msg, custom_prov_data, catalog_item):
+                                                extra_msg, set_default, custom_prov_data,
+                                                catalog_item):
     """Tests quota enforcement via service infra
 
     Polarion:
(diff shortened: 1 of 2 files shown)


dajoRH:
*CFME QE Bot*

dajoRH:
*CFME QE Bot*

dajoRH:
Lint report for commit 8394f752ca3ca3d9ed9ab2205b5f34e43be334a9:

`cfme/tests/infrastructure/test_tenant_quota.py`:
- :bangbang: Line 203:71: **E231** *missing whitespace after ','*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/infrastructure/test_tenant_quota.py
```diff
@@ -23,6 +23,26 @@
...
+       be created with specific values for these entries.
+    """
+    with_prov = (
+        "/ManageIQ (Locked)/{}/VM/Provisioning/StateMachines/ProvisionRequestApproval/Default"
+        .format(provider.string_name)
+    )
+    default = (
+        "/Service/Provisioning/StateMachines/ServiceProvision_Template/CatalogItemInitialization"
+    )
+    return with_prov if request.param else default
```
valaparthvi: Why not do something like this:
```
if request.param:
    return_value = 'With Provider"
else:
    return_value = "Default"
return return_value
```

cfme/tests/infrastructure/test_tenant_quota.py
```diff
@@ -316,3 +339,63 @@ def test_vm_migration_after_assigning_tenant_quota(appliance, small_vm, set_root
...
+    # nav to requests page to check quota validation
+    request_description = 'Provisioning Service [{name}] from [{name}]'.format(
+        name=catalog_item.name)
+    provision_request = appliance.collections.requests.instantiate(request_description)
+    provision_request.wait_for_request(method='ui')
+
+    @request.addfinalizer
+    def delete():
+        provision_request.remove_request()
+        catalog_item.delete()
```
valaparthvi: Why not delete this inside the fixture that creates this catalog item?

cfme/tests/infrastructure/test_tenant_quota.py
ganeshhubale: Thanks @valaparthvi 
But It will not be more readable as you will get to know that, you are only returning ```return_value```.
Also it's not that difficult to read if we followed existing approach. So no readability issue. I would like to go with existing one.

cfme/tests/infrastructure/test_tenant_quota.py
ganeshhubale: Good catch!! @valaparthvi

cfme/tests/infrastructure/test_tenant_quota.py
valaparthvi: You can always add a comment to indicate what type of value is being returned.

ganeshhubale:
@valaparthvi agree! 
- Removed ```uses_infra_provider```.
>     1. This test file uses both `setup_provider` and `uses_infra_provider`, why not just use one of them?

cfme/tests/infrastructure/test_tenant_quota.py
```diff
@@ -80,17 +100,21 @@ def set_roottenant_quota(request, roottenant, appliance):
...
         description='test catalog',
         display_in=True,
         catalog=catalog,
         dialog=dialog,
-        prov_data=prov_data)
+        prov_data=prov_data,
+        provider=provider,
+        provisioning_entry_point=set_default)
+    yield catalog_item
+    catalog_item.delete()
```
digitronik: Optional:
I think you need to check exist or not before delete it.

cfme/tests/infrastructure/test_tenant_quota.py
```diff
@@ -316,3 +337,59 @@ def test_vm_migration_after_assigning_tenant_quota(appliance, small_vm, set_root
...
+# indirect is the list where we define which fixtures are to be passed values indirectly.
+@pytest.mark.parametrize('context', [ViaSSUI, ViaUI])
+@pytest.mark.parametrize(
+    ['set_roottenant_quota', 'custom_prov_data', 'extra_msg', 'set_default'],
+    [
+        [('cpu', '2'), {'hardware': {'num_sockets': '8'}}, '', True],
+        [('storage', '0.01'), {}, '', True],
+        [('memory', '2'), {'hardware': {'memory': '4096'}}, '', True],
+        [('vm', '1'), {'catalog': {'num_vms': '4'}}, '###', True],
+    ],
```
digitronik: Optional:
I think @ganeshhubale we can put this as global variable as `dict` and then parametrize.

ganeshhubale: There are other test cases in other files too which needs same improvement. 
I will make new PR for it.
Thanks for suggesting

digitronik:
Waiting for PRT...

digitronik:
@ganeshhubale please check 5.9 run

ganeshhubale:
@digitronik 
Please take a look at it.
Now it's working as expected after re-basing this branch.

cfme/tests/infrastructure/test_tenant_quota.py
```diff
@@ -166,18 +191,19 @@ def test_tenant_quota_enforce_via_lifecycle_infra(appliance, provider, set_roott
...
+        [('memory', '2'), {'hardware': {'memory': '4096'}}, '', False],
+        [('vm', '1'), {'catalog': {'num_vms': '4'}}, '###', False],
     ],
-    indirect=['set_roottenant_quota', 'custom_prov_data'],
+    indirect=['set_roottenant_quota', 'custom_prov_data', 'set_default'],
     ids=['max_cpu', 'max_storage', 'max_memory', 'max_vms']
 )
 def test_tenant_quota_enforce_via_service_infra(request, appliance, context, set_roottenant_quota,
-                                                extra_msg, custom_prov_data, catalog_item):
+                                                extra_msg, set_default, custom_prov_data,
```
mshriver: Why is `set_default` included here? Its not used by the test, and its not actually doing any setup.

cfme/tests/infrastructure/test_tenant_quota.py
```diff
@@ -18,11 +18,31 @@
...
     test_requirements.vm_migrate,
-    pytest.mark.usefixtures('setup_provider', 'uses_infra_providers'),
+    pytest.mark.usefixtures('setup_provider'),
     pytest.mark.provider([VMwareProvider, RHEVMProvider], scope="module", selector=ONE_PER_TYPE)
 ]
 
 
+@pytest.fixture
+def set_default(provider, request):
+    """This fixture is used to return paths for provisioning_entry_point, reconfigure_entry_point
```
mshriver: Thanks for the detailed description, but it doesn't make sense to me that this is a fixture. Its not doing any setup, its simply returning a string, or a formatted string.

This could just be a statement within the `catalog_item` fixture, with `set_default` being changed to a more descriptive indirect parameter, like `default_entry_point`. 

Is there value to testing both of these settings for any test using a catalog item? Consider actually parametrizing this fixture directly.

cfme/tests/infrastructure/test_tenant_quota.py
```diff
@@ -316,3 +338,59 @@ def test_vm_migration_after_assigning_tenant_quota(appliance, small_vm, set_root
...
+        testSteps:
+            1. Add infrastructure provider
+            2. Set quota for root tenant - 'My Company'
+            3. Navigate to services > catalogs
+            4. Create catalog item with selecting following field entry points:
+                a.provisioning_entry_point = /ManageIQ (Locked)/Infrastructure/VM/Provisioning
+                /StateMachines/ProvisionRequestApproval/Default
+                b.retirement_entry_point = /Service/Retirement/StateMachines/ServiceRetirement
+                /Default
+            5. Add other information required in catalog for provisioning VM
```
mshriver: All of these steps are actually setup, as they're executed within fixtures. The actual `testSteps` are just ordering the catalog.

cfme/tests/infrastructure/test_tenant_quota.py
ganeshhubale: @mshriver 
- ```set_default``` fixture is used by this test case for differentiating this test case from the test cases which are using ``` entry points``` other than default.
- Here this test case needs the ```catalog_item``` fixture and ```set_default``` is called by the ```catalog_item``` fixture.
- As I said this test case uses default entry points. So the fixture ```set_default``` returns default value of ```provisioning_entry_point```.
- ```set_default``` is parametrized for such a test cases and it differentiates the tests with default entry points from tests with custom entry points.

cfme/tests/infrastructure/test_tenant_quota.py
ganeshhubale: @mshriver 
- I am agree with the statement that ```Its not doing any setup; its simply returning a string, or a formatted string.```
- But this needs to be different fixture because as I mentioned in the doc string of ```set_default``` fixture; In future some test cases may require all the three custom entry points. **Also I have called ```provider``` fixture from this fixture.**
- Parametrising this fixture directly may not be good idea if we want to pass all the three entry points to ```catalog item``` fixture at a time.
- Right now, there are  only two test cases for this scenario(custom entry points) [one for cloud](https://github.com/ManageIQ/integration_tests/pull/7903/files) and [other one for infra providers](https://github.com/ManageIQ/integration_tests/pull/8112/files)
- I would like to discuss about ```set_default``` fixture more in future for such a scenarios.

cfme/tests/infrastructure/test_tenant_quota.py
mshriver: Obviously from my comments I'm not too much of a fan of this structure, but I'm not going to block your PR on it. As you said above, lets continue to discuss how we can better structure this test parametrization overall.

ganeshhubale: Sure :)
Thanks @mshriver

PR #8114, opened 2018-11-09
[RFR][NOTEST] Added cleanup for SQS queues

mmojzis:
Deleting SQS queues older than 14 days.
This wrapanapi PR required for this PR: https://github.com/ManageIQ/wrapanapi/commit/291b029fdcc5860218201d694f3408a3a76289b9
Also after merge please keep rerunning ec2cleanup for at least one day as there are so many queues and only 1k are deleted by one run. @mshriver

### DIFF
--- a/scripts/ec2cleanup.py
+++ b/scripts/ec2cleanup.py
@@ -1,5 +1,6 @@
 import argparse
 import sys
+import time
 from datetime import datetime, timedelta
 from tabulate import tabulate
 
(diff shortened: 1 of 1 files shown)


scripts/ec2cleanup.py
```diff
@@ -189,8 +192,34 @@ def delete_snapshots(provider_mgmt, excluded_snapshots, output):
...
+    current_timestamp = time.time()
+    try:
+        queues = provider_mgmt.list_queues_with_creation_timestamps()
+        for queue in queues:
+            if excluded_queues and queue in excluded_queues:
+                logger.info(" Excluding Queue with url: %r", queue)
+                continue
+            else:
+                # 1209600 is 14 days in seconds
+                if float(current_timestamp) - float(queues[queue]) > 1209600:
```
digitronik: Required:
`current_timestamp` will always `float` value. please remove conversion.

scripts/ec2cleanup.py
```diff
@@ -189,8 +192,34 @@ def delete_snapshots(provider_mgmt, excluded_snapshots, output):
...
+                    provider_mgmt.delete_sqs_queue(queue)
+                    queue_list.append([provider_name, queue])
+        logger.info("  Deleted Queues: %r", queue_list)
+        with open(output, 'a+') as report:
+            if queue_list:
```
digitronik: Optional:
I would open a file for writing only if `queue_list` available.

scripts/ec2cleanup.py
```diff
@@ -215,6 +244,10 @@ def ec2cleanup(exclude_volumes, exclude_eips, exclude_elbs, exclude_enis, exclud
                       excluded_stacks=exclude_stacks,
                       stack_template=stack_template,
                       output=output)
+        logger.info("Deleting old queues...")
```
digitronik: :+1:

scripts/ec2cleanup.py
mmojzis: You are right. That makes sense.

scripts/ec2cleanup.py
mmojzis: Fixed.

PR #8115, opened 2018-11-09
[1LP][WIPTEST] Add test for validation of Nuage Cloud Tenant inventory

aljazkosir:
This test creates sandbox with Nuage cloud tenant and its entities. We then navigate to Cloud tenant (which we created in previous step) through provider, where we compare UI stats with sandbox stats.
We also extended fixture for nuage sandbox and made it 'module' scoped which will help us with the future inventory tests.

\cc @miha-plesko

### DIFF
--- a/cfme/cloud/tenant.py
+++ b/cfme/cloud/tenant.py
@@ -10,6 +10,7 @@ from cfme.base.ui import BaseLoggedInPage
 from cfme.common import Taggable
 from cfme.exceptions import TenantNotFound, DestinationNotFound
 from cfme.modeling.base import BaseCollection, BaseEntity
+from cfme.networks import ValidateStatsMixin
 from cfme.utils.appliance.implementations.ui import CFMENavigateStep, navigator, navigate_to
 from cfme.utils.log import logger
 from cfme.utils.wait import wait_for, TimedOutError
@@ -101,12 +102,41 @@ class ProviderTenantAllView(TenantAllView):
 
     @property
     def is_displayed(self):
-        return (
-            self.logged_in_as_current_user and
-            self.navigation.currently_selected == ['Compute', 'Clouds', 'Providers'] and
-            self.entities.title.text == '{} (All Cloud Tenants)'.format(
-                self.context['object'].name)
-        )
+        def get_provider_navigation(provider):
+            # Return navigation path based on a type of a provider
+            # TODO: Store navigation target as class attribute in Provider classes
+            if getattr(provider, 'name', False) and provider.category == 'cloud':
+                return ['Compute', 'Clouds', 'Providers']
+            elif getattr(provider, 'name', False) and provider.category == 'networks':
+                return ['Networks', 'Providers']
+            return []
+
+        expected_title = '{} (All Cloud Tenants)'
+        obj = self.context['object']
+        is_entity = getattr(obj, 'name', False) and isinstance(obj, BaseEntity)
+        is_filtered = isinstance(obj, BaseCollection) and obj.filters
+        provider = obj.filters.get('parent') or obj.filters.get('provider') if is_filtered else None
+
+        if is_entity:
+            logger.debug('Tenant view context object is assumed to be provider: %r', obj)
+            matched_title = self.entities.title.text == expected_title.format(obj.name)
+            matched_navigation = self.navigation.currently_selected == get_provider_navigation(obj)
+        elif provider and hasattr(provider, 'name'):
+            # filtered collection, use provider object's name and provider object's navigation
+            logger.debug(
+                'Tenant view context object has provider related to view with name attribute: %r',
+                obj.filters
+            )
+            matched_title = self.entities.title.text == expected_title.format(provider.name)
+            matched_navigation = (
+                self.navigation.currently_selected == get_provider_navigation(provider)
+            )
+        else:
+            # not an entity with a name, or a filtered collection
+            matched_title = False
+            matched_navigation = False
+
+        return self.logged_in_as_current_user and matched_navigation and matched_title
 
 
 class TenantDetailsView(TenantView):
@@ -173,7 +203,7 @@ class TenantEditView(TenantView):
 
 
 @attr.s
-class Tenant(BaseEntity, Taggable):
+class Tenant(BaseEntity, Taggable, ValidateStatsMixin):
     """Tenant Class
 
     """
@@ -335,7 +365,17 @@ class TenantAll(CFMENavigateStep):
 @navigator.register(Tenant, 'Details')
 class TenantDetails(CFMENavigateStep):
     VIEW = TenantDetailsView
-    prerequisite = NavigateToAttribute('parent', 'All')
+
+    def prerequisite(self, *args, **kwargs):
+        """
+        Navigate through provider or collection filter if it exists else navigate through collection
+        object """
+        # Here we assume that every tenant has a parent collection
+        filter = self.obj.provider or self.obj.parent.filters.get('parent')
+        if filter:
+            return navigate_to(filter, 'CloudTenants')
+        else:
+            return navigate_to(self.obj.parent, 'All')
 
     def step(self, *args, **kwargs):
         """Navigate to the details page"""
--- a/cfme/fixtures/nuage.py
+++ b/cfme/fixtures/nuage.py
@@ -1,8 +1,8 @@
 import pytest
 
 from wrapanapi.utils.random import random_name
-
 from cfme.utils.log import logger
+from cfme.utils.wait import wait_for
 
 
 def create_basic_sandbox(nuage):
@@ -10,7 +10,7 @@ def create_basic_sandbox(nuage):
 
     # Create empty enterprise aka 'sandbox'.
     enterprise = box['enterprise'] = nuage.create_enterprise()
-    logger.info('Created sandbox enterprise {} ({})'.format(enterprise.name, enterprise.id))
+    logger.info('Created sandbox enterprise %s (%s)', enterprise.name, enterprise.id)
 
     # Fill the sandbox with some entities.
     # Method `create_child` returns a tuple (object, connection) and we only need object.
@@ -57,4 +57,28 @@ def with_nuage_sandbox(networks_provider):
     # Destroy the sandbox.
     enterprise = sandbox['enterprise']
     nuage.delete_enterprise(enterprise)
-    logger.info('Destroyed sandbox enterprise {} ({})'.format(enterprise.name, enterprise.id))
+    logger.info('Destroyed sandbox enterprise %s (%s)', enterprise.name, enterprise.id)
+
+
+@pytest.fixture(scope='module')
+def with_nuage_sandbox_modscope(appliance, setup_provider_modscope, provider):
+    nuage = provider.mgmt
+    sandbox = create_basic_sandbox(nuage)
+    enterprise = sandbox['enterprise']
+    logger.info('Performing a full refresh, so sandbox %s appears in the database', enterprise.name)
+    provider.refresh_provider_relationships()
+    wait_for(provider.is_refreshed, func_kwargs=dict(refresh_delta=5), timeout=600, delay=10)
+
+    # Check if tenant exists in database, if not fail test immediately
+    tenants_table = appliance.db.client['cloud_tenants']
+    tenant = (appliance.db.client.session
+              .query(tenants_table.name, tenants_table.ems_ref)
+              .filter(tenants_table.ems_ref == enterprise.id).first())
+    assert tenant is not None, 'Nuage sandbox tenant inventory missing: {}'.format(enterprise.name)
+
+    # Let integration test do whatever it needs to do.
+    yield sandbox
+
+    # Destroy the sandbox.
+    nuage.delete_enterprise(enterprise)
+    logger.info('Destroyed sandbox enterprise %s (%s)', enterprise.name, enterprise.id)
--- a/cfme/networks/__init__.py
+++ b/cfme/networks/__init__.py
@@ -1,0 +1,29 @@
+from cfme.utils.appliance.implementations.ui import navigate_to
+
+
+class ValidateStatsMixin(object):
+    # TODO: Move this class to a higher level, where could be useful for thing beyond this
+    # network module, maybe BaseEntity
+
+    def ui_details_value(self, path, property_name):
+        view = navigate_to(self, "Details", use_resetter=False)
+        return getattr(view.entities, path).get_text_of(property_name)
+
+    def validate_stats(self, expected_stats):
+        """ Validates that the details page matches the expected stats.
+
+        Args:
+            expected_stats: dictionary of values to be compered to UI values
+
+        This method is given expected stats as an argument and those are then matched
+        against the UI. An AssertionError exception will be raised if the stats retrieved
+        from the UI do not match those from expected stats.
+
+        IMPORTANT: Please make sure inherited classes implements `ui_details_value` which fetches
+        the value of a stat
+        """
+
+        cfme_stats = {
+            (path, stat): self.ui_details_value(path, stat) for path, stat in expected_stats
+        }
+        assert cfme_stats == expected_stats
--- a/cfme/networks/provider/__init__.py
+++ b/cfme/networks/provider/__init__.py
@@ -2,6 +2,7 @@ import attr
 from cached_property import cached_property
 from navmazing import NavigateToSibling, NavigateToAttribute
 
+from cfme.cloud.tenant import ProviderTenantAllView
 from cfme.common import Taggable
 from cfme.common.provider import BaseProvider, prepare_endpoints
 from cfme.exceptions import DestinationNotFound
@@ -364,6 +365,20 @@ class Edit(CFMENavigateStep):
     def step(self):
         self.prerequisite_view.entities.get_entity(name=self.obj.name, surf_pages=True).check()
         self.prerequisite_view.toolbar.configuration.item_select('Edit Selected Network Provider')
+
+
+@navigator.register(NetworkProvider, 'CloudTenants')
+class OpenCloudTenants(CFMENavigateStep):
+    prerequisite = NavigateToSibling('Details')
+    VIEW = ProviderTenantAllView
+
+    def step(self):
+        item = 'Cloud Tenants'
+        item_amt = int(self.prerequisite_view.entities.relationships.get_text_of(item))
+        if item_amt > 0:
+            self.prerequisite_view.entities.relationships.click_at(item)
+        else:
+            raise DestinationNotFound("This provider doesn't have {item}".format(item=item))
 
 
 @navigator.register(NetworkProvider, 'CloudSubnets')
--- a/cfme/networks/provider/nuage.py
+++ b/cfme/networks/provider/nuage.py
@@ -1,7 +1,10 @@
 import attr
 import copy
+
 from widgetastic.widget import View
 from widgetastic_patternfly import BootstrapSelect, Input, Button
+
+from cfme.cloud.tenant import TenantCollection
 from wrapanapi.systems import NuageSystem
 
 from cfme.common.provider import DefaultEndpoint, DefaultEndpointForm, EventsEndpoint
@@ -68,6 +71,7 @@ class NuageProvider(NetworkProvider):
 
     _collections = {
         'security_groups': SecurityGroupCollection,
+        'cloud_tenants': TenantCollection
     }
 
     key = attr.ib(default=None)
--- a/cfme/tests/networks/nuage/test_inventory.py
+++ b/cfme/tests/networks/nuage/test_inventory.py
@@ -1,0 +1,20 @@
+import pytest
+
+from cfme.networks.provider.nuage import NuageProvider
+
+pytestmark = [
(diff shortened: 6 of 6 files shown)


cfme/tests/networks/nuage/test_inventory.py
```diff
@@ -0,0 +1,28 @@
...
+    groups = sum([len(d.policy_groups) for d in sandbox['enterprise'].domains])
+    l2_vports = sum([len(d.vports.get()) for d in sandbox['enterprise'].l2_domains])
+    vports = sum([len(d.vports.get()) for d in sandbox['enterprise'].domains])
+
+    expected_stats = {
+        'num_cloud_subnets': len(subnets) + len(sandbox['enterprise'].l2_domains.get()),
+        'num_network_routers': len(sandbox['enterprise'].domains),
+        'num_security_groups': l2_groups + groups,
+        'num_network_ports': l2_vports + vports
+    }
```
miha-plesko: So sorry to mislead you earlier @aljazkosir but we know all those values upfront - because we know exactly what we created in the sandbox. Can you please just hard-code them here?

```
expected_stats = {
	'num_cloud_subnets': 1,
	'num_network_routers': 1,
	'num_security_groups': 2,
	'num_network_ports': 4
}
```

The idea is that we grab only those variables from sandbox that we don't know upfront (e.g. tenant name - because tenant name is randomized upon creation).

aljazkosir:
@miha-plesko thanks for the review, I have applied the changes you suggested. Let me know if there is anything else I could improve.

cfme/tests/networks/nuage/test_inventory.py
```diff
@@ -0,0 +1,21 @@
...
+    tenant_name = sandbox['enterprise'].name
+    tenant = provider.cloud_tenants.get_by_name(tenant_name)
+
+    expected_stats = {
+        'num_cloud_subnets': 2,
+        'num_network_routers': 1,
+        'num_security_groups': 2,
+        'num_network_ports': 4
+    }
+    tenant.validate_stats(expected_stats)
```
miha-plesko: Please just pass the expected hash inline, no need to introduce `expected_stats` variable at all.

cfme/fixtures/nuage.py
```diff
@@ -58,3 +58,26 @@ def with_nuage_sandbox(networks_provider):
...
+    sandbox = create_basic_sandbox(nuage)
+    logger.info('Performing a full refresh, this might take a while.')
+    wait_for(provider.is_refreshed, func_kwargs=dict(refresh_delta=10), timeout=600, delay=10)
+
+    # Check if tenant exists in database, if not fail test immediately
+    tenants_table = appliance.db.client['cloud_tenants']
+    tenant = (appliance.db.client.session
+              .query(tenants_table.name, tenants_table.ems_ref)
+              .filter(tenants_table.ems_ref == sandbox['enterprise'].id).first())
+    assert tenant is not None
```
mshriver: Please include some contextual description here since we're going to hit a setup phase AssertionError here. Something like:

```python
assert tenant is not None, 'Nuage sandbox tenant inventory missing: {}'.format(sandbox)
```

cfme/fixtures/nuage.py
```diff
@@ -58,3 +58,26 @@ def with_nuage_sandbox(networks_provider):
     enterprise = sandbox['enterprise']
     nuage.delete_enterprise(enterprise)
     logger.info('Destroyed sandbox enterprise {} ({})'.format(enterprise.name, enterprise.id))
+
+
+@pytest.fixture(scope='module')
+def with_nuage_sandbox_modscope(appliance, setup_provider_modscope, provider):
+    nuage = provider.mgmt
+    sandbox = create_basic_sandbox(nuage)
+    logger.info('Performing a full refresh, this might take a while.')
```
mshriver: Please include more context here in the log message - your message on L83 is perfect, including some information about the sandbox.

cfme/networks/views.py
```diff
@@ -550,6 +550,26 @@ def is_displayed(self):
                 self.title.text == '{name} (Summary)'.format(name=self.context['object'].name))
 
 
+class CloudTenantEntities(BaseEntitiesView):
```
mshriver: I think that these views are duplicating many of the views already defined in `cfme/cloud/tenant.py`, and the only unique view here is the one for a network provider's tenant collection.  Please re-use views where possible.

cfme/networks/provider/nuage.py
```diff
@@ -77,6 +81,10 @@ class NuageProvider(NetworkProvider):
     def __attrs_post_init__(self):
         self.parent = self.appliance.collections.network_providers
 
+    @property
+    def cloud_tenants(self):
+        return self.collections.cloud_tenants
```
mshriver: I think that this qualifies as a 'convenience method', and should not be included. The caller should reference entity collections directly through BaseEntity.collections, this property provides no additional functionality.

cfme/networks/__init__.py
```diff
@@ -0,0 +1,20 @@
+class ValidateStatsMixin(object):
+    def validate_stats(self, expected_stats):
```
mshriver: This is nice, but should likely be abstracted to a bit higher level where its useful for things beyond this network module. We have validate_stats built-in to the base provider, maybe BaseEntity could include this mixin? A TODO would be nice for now.

cfme/networks/views.py
```diff
@@ -653,6 +673,37 @@ class OneProviderComponentsToolbar(View):
     view_selector = View.nested(ItemsToolBarViewSelector)
 
 
+class OneProviderCloudTenantsView(BaseLoggedInPage):
+    """ Represents whole All Tenant page for provider"""
```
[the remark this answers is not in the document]
aljazkosir: Thanks, I wasn't aware of this file. Will take a look and remove my duplicates.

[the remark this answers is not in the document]
mshriver: Don't hesitate to reach out if you run into an trouble, or its not clear to how to design the view to be useful on multiple navigation destinations, or how to attach navigation destinations to different entities/collections.

aljazkosir:
Thanks for a review @mshriver, I pushed the changes you suggested. I was not able to test the new view for other providers except for NuageProvider. Please, if you can run a test that uses this view and different provider.

\cc @miha-plesko

cfme/cloud/tenant.py
```diff
@@ -101,12 +101,42 @@ class ProviderTenantAllView(TenantAllView):
...
-            self.navigation.currently_selected == ['Compute', 'Clouds', 'Providers'] and
-            self.entities.title.text == '{} (All Cloud Tenants)'.format(
-                self.context['object'].name)
-        )
+        def get_provider_navigation(provider):
+            # Return navigation path based on a type of a provider
+            from cfme.networks.provider import NetworkProvider
+            from cfme.cloud.provider import CloudProvider
+            if getattr(provider, 'name', False) and isinstance(provider, CloudProvider):
+                return ['Compute', 'Infrastructure', 'Providers']
```
miha-plesko: Cloud providers are under Compute > Cloud > Providers

cfme/cloud/tenant.py
```diff
@@ -101,12 +101,42 @@ class ProviderTenantAllView(TenantAllView):
...
+            elif getattr(provider, 'name', False) and isinstance(provider, NetworkProvider):
+                return ['Networks', 'Providers']
+            return []
+
+        expected_title = '{} (All Cloud Tenants)'
+        obj = self.context['object']
+        is_entity = getattr(obj, 'name', False) and isinstance(obj, BaseEntity)
```
miha-plesko: I'd prefer if we do it the same way as in L117:

```python
is_entity = isinstance(obj, BaseEntity) and obj.name
```

cfme/networks/provider/__init__.py
```diff
@@ -24,8 +25,7 @@
     OneProviderNetworkRouterView,
     OneProviderSecurityGroupView,
     OneProviderSubnetView,
-    NetworkProviderEditView
-)
+    NetworkProviderEditView)
```
miha-plesko: Please exclude this unrelated change from commit.

cfme/cloud/tenant.py
aljazkosir: The problem here is that `name` attribute does not exists in the definition of `BaseEntity` and this could result in an error in some cases. While `filter` is in the definition of `BaseCollection`

cfme/networks/cloud_tenant.py
```diff
@@ -0,0 +1,72 @@
...
+from cfme.utils import version
+from cfme.utils.appliance.implementations.ui import navigator, CFMENavigateStep, navigate_to
+
+
+@attr.s
+class CloudTenant(Taggable, BaseEntity, ValidateStatsMixin):
+    """Class representing cloud tenants"""
+    in_version = ('5.10', version.LATEST)
+    category = "networks"
+    string_name = 'CloudTenant'
```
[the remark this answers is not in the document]
aljazkosir: I was following the pattern of other similar classes, for example: [cloud_network.py](https://github.com/ManageIQ/integration_tests/blob/master/cfme/networks/cloud_network.py)

cfme/networks/cloud_tenant.py
```diff
@@ -0,0 +1,72 @@
...
+from cfme.modeling.base import BaseCollection, BaseEntity, parent_of_type
+from cfme.networks import ValidateStatsMixin
+from cfme.utils import version
+from cfme.utils.appliance.implementations.ui import navigator, CFMENavigateStep, navigate_to
+
+
+@attr.s
+class CloudTenant(Taggable, BaseEntity, ValidateStatsMixin):
+    """Class representing cloud tenants"""
+    in_version = ('5.10', version.LATEST)
```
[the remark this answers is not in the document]
aljazkosir: As far as I know, this line is for the tests, so we can skip the test based on the appliance version

cfme/fixtures/nuage.py
```diff
@@ -58,3 +58,26 @@ def with_nuage_sandbox(networks_provider):
...
+              .query(tenants_table.name, tenants_table.ems_ref)
+              .filter(tenants_table.ems_ref == enterprise.id).first())
+    assert tenant is not None, 'Nuage sandbox tenant inventory missing: {}'.format(enterprise.name)
+
+    # Let integration test do whatever it needs to do.
+    yield sandbox
+
+    # Destroy the sandbox.
+    nuage.delete_enterprise(enterprise)
+    logger.info('Destroyed sandbox enterprise {} ({})'.format(enterprise.name, enterprise.id))
```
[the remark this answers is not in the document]
aljazkosir: I would rather create new PR for this and fix all string formatting in this file, because those lines already existed before I started implementing new test. I didn't want to change things that weren't related.

cfme/networks/cloud_tenant.py
```diff
@@ -0,0 +1,72 @@
...
+    category = "networks"
+    string_name = 'CloudTenant'
+    quad_name = None
+    db_types = ["CloudTenant"]
+
+    name = attr.ib()
+    provider_obj = attr.ib(default=None)
```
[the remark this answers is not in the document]
aljazkosir: As far as I know, Cloud Tenant can exist without a provider, hence the `default=None`

cfme/cloud/tenant.py
```diff
@@ -101,12 +101,42 @@ class ProviderTenantAllView(TenantAllView):
...
-            self.entities.title.text == '{} (All Cloud Tenants)'.format(
-                self.context['object'].name)
-        )
+        def get_provider_navigation(provider):
+            # Return navigation path based on a type of a provider
+            from cfme.networks.provider import NetworkProvider
+            from cfme.cloud.provider import CloudProvider
+            if getattr(provider, 'name', False) and isinstance(provider, CloudProvider):
+                return ['Compute', 'Clouds', 'Providers']
+            elif getattr(provider, 'name', False) and isinstance(provider, NetworkProvider):
```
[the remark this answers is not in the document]
aljazkosir: Yes, this is correct... every `Provider` has a `name` in its definition, I applied this suggestion.

cfme/networks/cloud_tenant.py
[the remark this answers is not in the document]
mshriver: I'm actually not aware of where this is consumed. I see it in vcloud provider, container provider, and network classes, but no fixture or plugin handling of this during collection.

cfme/networks/cloud_tenant.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+@attr.s
+class CloudTenantCollection(BaseCollection):
+    """Collection object for CloudTenant object"""
+    ENTITY = CloudTenant
+
+    def all(self):
+        view = navigate_to(self.filters.get('parent'), 'CloudTenants')
+        cloud_tenants = view.entities.get_all(surf_pages=True)
```
mshriver: You're only using the names here, just use `view.entities.all_entity_names` property.

cfme/tests/networks/nuage/test_inventory.py
```diff
@@ -0,0 +1,20 @@
...
+]
+
+
+def test_tenant_details(setup_provider_modscope, provider, with_nuage_sandbox_modscope):
+    sandbox = with_nuage_sandbox_modscope
+    tenant_name = sandbox['enterprise'].name
+    tenant = provider.collections.cloud_tenants.instantiate(name=tenant_name)
+
+    tenant.validate_stats({
+        'num_cloud_subnets': 2,
```
mshriver: Where are these values coming from? They're too 'magic' of numbers, and need to be pulled from somewhere consistent, if that's the sandbox fixture.

cfme/networks/cloud_tenant.py
```diff
@@ -0,0 +1,72 @@
...
+from cfme.cloud.tenant import TenantDetailsView
+from cfme.common import Taggable
+from cfme.modeling.base import BaseCollection, BaseEntity, parent_of_type
+from cfme.networks import ValidateStatsMixin
+from cfme.utils import version
+from cfme.utils.appliance.implementations.ui import navigator, CFMENavigateStep, navigate_to
+
+
+@attr.s
+class CloudTenant(Taggable, BaseEntity, ValidateStatsMixin):
```
mshriver: This still appears to be a duplication of the Tenant/TenantCollection classes that already exist in `cfme/cloud/tenant.py`, or at least poorly named for its intent.

The tenants that this is apparently tied to, when you're on the page what vertical navigation is it under? Are they the same as the cloud provider tenants?

If so, then the convenience methods implemented here could be implemented elsewhere, or a mechanism with less boilerplate could be used.  There could be a single navigation destination for a Tenant details page, that could take into account any filters applied to its parent collection.

cfme/networks/cloud_tenant.py
```diff
@@ -0,0 +1,72 @@
...
+    name = attr.ib()
+    provider_obj = attr.ib(default=None)
+
+    @property
+    def provider(self):
+        from cfme.networks.provider import NetworkProvider
+        return parent_of_type(self, NetworkProvider)
+
+    @property
+    def num_cloud_subnets(self):
```
mshriver: There are many more of these fields on the page, I don't think its a good pattern to try to implement these boilerplate methods to get the relationship values from the tables.

cfme/networks/cloud_tenant.py
```diff
@@ -0,0 +1,72 @@
...
+@navigator.register(CloudTenant, 'Details')
+class Details(CFMENavigateStep):
+    VIEW = TenantDetailsView
+
+    def prerequisite(self, *args, **kwargs):
+        """Navigate through provider if it exists else navigate through parent object"""
+        if self.obj.provider:
+            return navigate_to(self.obj.provider, 'CloudTenants')
+        else:
+            return navigate_to(self.obj.parent, 'CloudTenants')
```
mshriver: The parent could well be a CloudTenantCollection, that will not have a CloudTenants destination registered. This destination is only registered against NetworkProvider.

Again this comes back to duplication of Tenant / CloudTenant when they appear to be the same entity in MIQ.

The navigation here needs to be structured so that navigation to a Tenant uses a prerequisite that accounts for the appropriate layers of the model. The parent of the `Tenant` is a `TenantCollection`, but that `TenantCollection` could have a parent of: `Appliance`, `CloudProvider`, `NetworkProvider`.  This parent could be due to how the class was created (`appliance.collections.cloud_tenants` vs `provider.collections.tenants`) or due to a filter applied to the collection after it was created.

cfme/cloud/tenant.py
```diff
@@ -101,12 +101,42 @@ class ProviderTenantAllView(TenantAllView):
...
-            self.logged_in_as_current_user and
-            self.navigation.currently_selected == ['Compute', 'Clouds', 'Providers'] and
-            self.entities.title.text == '{} (All Cloud Tenants)'.format(
-                self.context['object'].name)
-        )
+        def get_provider_navigation(provider):
+            # Return navigation path based on a type of a provider
+            from cfme.networks.provider import NetworkProvider
+            from cfme.cloud.provider import CloudProvider
+            if isinstance(provider, CloudProvider):
```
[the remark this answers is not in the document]
aljazkosir: I agree, but I think it's better to create new PR for this enhancement. I will compare by `category` for now and add a TODO message. Later I will create new PR for this.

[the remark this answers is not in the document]
miha-plesko: @aljazkosir I see all cloud providers already contain `provider.string_name = 'Cloud'` and Nuage already contains `provider.string_name = 'Network'` so I don't see why we shouldn't just switch to it right away? It's for free, event less imports :)

cfme/networks/cloud_tenant.py
miha-plesko: I think it's a common pattern to do it like this @mshriver (we've seen it with other providers), but what you say makes sense. @aljazkosir maybe we can implement following method instead:

```python
def ui_details_value(property_name)
  return view.entities.relationships.get_text_of(property_name)
```

and then update ValidateStatsMixin to just make use of it. Should be quite trivial and indeed, as @mshriver suggests, quite less boilerplate.

cfme/tests/networks/nuage/test_inventory.py
miha-plesko: We know what sandbox we create with the fixture so those numbers have a right to be hard-coded. But I feel you @mshriver since you don't know the sandbox, it's fuzzy, thanks for pointing it out.

May I suggest we fix this in a followup PR because we'll need to refactor return value of the fixture a bit and adjust other Nuage tests as well?

cfme/cloud/tenant.py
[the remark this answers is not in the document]
aljazkosir: Yes, comparing by string name will be implemented. I meant to create new PR for the enhancement that will store navigation target as class attribute as @mshriver mentioned:

> Even better - CloudProvider, InfraProvider, NetworkProvider (etc) could be enhanced to store a navigation target as a class attribute.

cfme/networks/cloud_tenant.py
aljazkosir: I pushed new changes, navigation now first checks if we have a provider or filter on collection and navigate through that. If we don't have provider or filter we navigate through default TenantCollection object `('Compute', 'Clouds', 'Tenants')`

I would kindly ask you to review our code again @mshriver :)

cfme/cloud/tenant.py
mshriver: @aljazkosir yep, that's exactly why I structured these two checks in this way in some of the other all methods :+1:

cfme/cloud/tenant.py
```diff
@@ -335,7 +365,17 @@ def resetter(self):
...
-    prerequisite = NavigateToAttribute('parent', 'All')
+
+    def prerequisite(self, *args, **kwargs):
+        """
+        Navigate through provider or collection filter if it exists else navigate through collection
+        object """
+        # Here we assume that every tenant has a parent collection
+        filter = self.obj.provider or self.obj.parent.filters.get('parent')
+        if filter:
+            return navigate_to(filter, 'CloudTenants')
```
mshriver: This navigation destination has only been registered for a `NetworkProvider` parent, but is not registered for a `CloudProvider` parent.  If I have a tenant object with a parent `TenantCollection` that is filtered, or that has a `provider` attribute set, and that provider is a `CloudProvider`, this call with raise `NavigationDestinationNotFound`.

A navigation class must be registered against `CloudProvider`, using the `ProviderTenantAllView`.

mshriver:
In local testing the updates to ProviderTenantAllView.is_displayed are working for cloud providers when I manually create the view with a TenantCollection context (no navigation destination for CloudProvider) that is filtered on a cloud provider.  Some additional tweaks are necessary for this to be a more complete implementation of the view and the paths for getting there.

cfme/cloud/tenant.py
miha-plesko: If Cloud provider would pay so much attention to our Networks provider as we do it for it, then the view would have worked for us out of the box, haha :) I don't mind supporting it for Cloud provider as well, but perhaps we could do it in a followup PR?

mshriver: I can dig that - hooking into existing framework comes with more responsibilities than building the foundations ;)

Ultimately my goal here is to make sure that all the providers live in harmony together, utilizing the same base model/view/navigation code wherever possible.

PR #8121, opened 2018-11-09
[1LP][RFR] Update test_vm_instance_analysis, unique ssa agent name

mshriver:
Believe this was the cause of many duplicate 'test_smartstate' instances in our ec2 region.  The duplicate name instances don't get cleaned up by normal means, and should have unique id's.


mshriver:
@sbulage updated

PR #8126, opened 2018-11-12
[1LP][RFR][NOTEST] Blocking test_provision_vm_to_virtual_network with GH8128

jan-zmeskal:
Blocker


jan-zmeskal:
Hi @jawatts, I pushed to this branch even though it is 1LP because I found another serious blocker for the test case: https://bugzilla.redhat.com/show_bug.cgi?id=1649320

PR #8132, opened 2018-11-12
[1LP][RFR] Update xunit fixture, include group-id from xunit.conf

mshriver:
remove containers from blacklist


mshriver:
Thanks for quick review @mkoura

PR #8133, opened 2018-11-13
[1LP][RFR] Fixing Condition CRUD test 

### LINKED ISSUE #8103: [Bug] Condition creation fails with StaleElementReferenceException
Unable to create a condition via `appliance.collections.conditions.create()` method

cf. `cfme/control/explorer/conditions.py`

e.g. [here](https://github.com/ManageIQ/integration_tests/blob/0751cabee21c8ff1b25cc1b87aee21fa981b5ca7/cfme/control/explorer/conditions.py#L237)
``` 
condition = condition_collection.create(
    condition_class,
    expression="fill_field({} : Address )".format(
        condition_class.FIELD_VALUE),
    scope="fill_field({} : Name, INCLUDES, {})".format(
        condition_class.FIELD_VALUE, fauxfactory.gen_alpha())
)
```

This will raise `StaleElementReferenceException: Message: Couldn't handle it` and appears to impact CFME 5.9 as well as 5.10. Once at this page, it is not possible to navigate via the Accordion, as it is grayed out. Seems to be impacting several test cases, but is most relevant for tests in:
`cfme/tests/control/test_basic.py` [here](https://github.com/ManageIQ/integration_tests/blob/0751cabee21c8ff1b25cc1b87aee21fa981b5ca7/cfme/tests/control/test_basic.py#L530)

john-dupuy:
Purpose or Intent
=================

__Fixing__ the condition CRUD test. This requires: 
__1)__ Updating the `condition_collection.create()` method to use `view.wait_displayed()` and update `ExpressionEditor` to use `show_loc` to select the proper buttons. 
__2)__ Putting in a workaround for [BZ 1649057](https://bugzilla.redhat.com/show_bug.cgi?id=1649057) in `BootstrapSelect` for the expression_editor
__3)__ Fixing `EditConditionView` `is_displayed()` method
__4)__ Adding a view for all the conditions in a condition class, since the appliance redirects the user to this page when a condition is deleted. 

FIXES #8103

{{ pytest: --long-running cfme/tests/control/test_basic.py::test_condition_crud }}

### DIFF
--- a/cfme/control/explorer/conditions.py
+++ b/cfme/control/explorer/conditions.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 import attr
-from wait_for import wait_for
 
 from navmazing import NavigateToAttribute, NavigateToSibling
 from widgetastic.widget import Text, TextInput, Widget
@@ -77,8 +76,8 @@ class ConditionFormCommon(ControlExplorerView):
 
     title = Text("#explorer_title_text")
     description = Input(name="description")
-    scope = ExpressionEditor("//img[@alt='Edit this Scope']")
-    expression = ExpressionEditor("//img[@alt='Edit this Expression']")
+    scope = ExpressionEditor("//button[normalize-space(.)='Define Scope']")
+    expression = ExpressionEditor("//button[normalize-space(.)='Define Expression']")
     notes = TextInput(name="notes")
 
     cancel_button = Button("Cancel")
@@ -86,7 +85,6 @@ class ConditionFormCommon(ControlExplorerView):
 
 class NewConditionView(ConditionFormCommon):
     add_button = Button("Add")
-    scope_button = Button("Define Scope")
 
     @property
     def is_displayed(self):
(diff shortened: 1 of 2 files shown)


mshriver:
@john-dupuy you can use GH 'smart text' or 'smart labels' or whatever it calls them to tie this PR and the issue together.

In your PR description, on a line by itself, put `FIXES #8103`.  This will tie the two together, and when this PR is merged, the issue will automatically be closed.

cfme/control/explorer/conditions.py
```diff
@@ -86,6 +86,7 @@ class ConditionFormCommon(ControlExplorerView):
 
 class NewConditionView(ConditionFormCommon):
     add_button = Button("Add")
+    scope_button = Button("Define Scope")
```
mshriver: Let's re-arrange this a bit so that the scope and expression editors click their required buttons before the widget is filled.

mshriver: `ExpressionEditor.show_loc` may actually be functional for this use

cfme/control/explorer/conditions.py
```diff
@@ -77,8 +76,8 @@ class ConditionFormCommon(ControlExplorerView):
 
     title = Text("#explorer_title_text")
     description = Input(name="description")
-    scope = ExpressionEditor("//img[@alt='Edit this Scope']")
-    expression = ExpressionEditor("//img[@alt='Edit this Expression']")
+    scope = ExpressionEditor(".//button[normalize-space(.)='Define Scope']")
```
mshriver: If you don't use relative xpath here, you shouldn't have to modify `ExpressionEditor.enable_editor`

PR #8139, opened 2018-11-14
[1LP][RFR] fixed test_collect_multiple_servers tests

lina-nikiforova:
{{ pytest: -v cfme/tests/configure/test_log_depot_operation.py::test_collect_multiple_servers }}

### DIFF
--- a/cfme/configure/configuration/diagnostics_settings.py
+++ b/cfme/configure/configuration/diagnostics_settings.py
@@ -275,6 +275,12 @@ class CollectLogsBase(Pretty, NavigatableMixin, Updateable):
             view = navigate_to(self, 'DiagnosticsCollectLogs')
         last_collection = self.last_collection
         # Initiate the collection
+        wait_for(lambda: view.toolbar.collect.item_enabled(selection),
+                 delay=20,
+                 timeout=7*60,
+                 handle_exception=True,
+                 fail_condition=False,
+                 fail_func=self.browser.refresh)
         view.toolbar.collect.item_select(selection, handle_alert=None)
         if self.browser.alert_present:
             self.browser.handle_alert(prompt=self.ALERT_PROMPT)
@@ -328,7 +334,6 @@ class CollectLogsBase(Pretty, NavigatableMixin, Updateable):
 
 
 # ======================= Server Collect Logs ============================
-
 class ServerCollectLogsToolbar(View):
     edit = Button(title="Edit the Log Depot settings for the selected Server")
     collect = Dropdown('Collect Logs')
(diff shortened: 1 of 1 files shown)


cfme/configure/configuration/diagnostics_settings.py
```diff
@@ -358,6 +373,30 @@ def in_server_collect_logs(self):
         )
 
 
+class ServerCollectLogsSlaveView(ServerCollectLogsView):
+    @property
+    def is_displayed(self):
+        slave_server = self.context['object'].appliance.server.slave_servers[0]
```
jawatts: **Question** What is the 0th element here?

cfme/configure/configuration/diagnostics_settings.py
```diff
@@ -343,11 +349,20 @@ class ServerCollectLogsView(ServerDiagnosticsView):
 
     @property
     def is_displayed(self):
+        tree = self.accordions.diagnostics.tree.currently_selected
+        selected_server = tree[-1].split(': ')[-1]
+        try:
+            name, sid, current = selected_server.split()
```
jawatts: **Required** I don't think this is a good assumption to make, what is name or sid ended up raising a ValueError. Perhaps look at the length before the split.

cfme/configure/configuration/diagnostics_settings.py
lina-nikiforova: I'm adding inline comment there, it's the first slave server (out of one)

cfme/configure/configuration/diagnostics_settings.py
```diff
@@ -358,6 +376,32 @@ def in_server_collect_logs(self):
...
+        )
+
+
+class ServerCollectLogsMasterView(ServerCollectLogsView):
+    @property
+    def is_displayed(self):
+        master_server = self.context['object'].appliance.server
+        return (
+            self.in_server_collect_logs and
+            self.title.text == 'Diagnostics Server "{name} [{sid}]" (current)'.format(
```
izapolsk: Q: I have a distinct feeling of deja vu :).  Could this code be wrapped up into function or variable and used in all these views ?

cfme/configure/configuration/diagnostics_settings.py
```diff
@@ -428,7 +472,7 @@ def is_displayed(self):
             self.collectlogs.is_displayed and
             self.collectlogs.is_active and
             self.title.text == 'Diagnostics Zone "{}" (current)'.format(
-                self.context['object'].description))
+                self.context['object'].appliance.default_zone.description))
```
izapolsk: Q: could you explain why default zone ?

lina-nikiforova: `self.context['object'].description` gives an error, so I used `appliance.default_zone` as log collection tests use only one zone, which is default.

cfme/configure/configuration/diagnostics_settings.py
```diff
@@ -328,6 +334,16 @@ def collect_current(self):
 
 
 # ======================= Server Collect Logs ============================
+# common part of is_displayed check for ServerCollectLogsSlaveView and ServerCollectLogsMasterView
+def server_collect_logs_displayed(view, server, slave=True):
```
izapolsk: okay, this is closer to what I thought about but I'd implement it a little bit different.

```python
class ServerCollectLogsView(ServerDiagnosticsView):
  title_template = 'Diagnostics Server "{name} {sid}"{current}'

  def form_expected_title(self, one, two, three):
    return self.title_template.format()

  @property
  def is_displayed(self):
    return in_server_collect_logs and self.title.text == self.form_expected_title()


class ServerCollectLogsSlaveView(ServerCollectLogsView):
  @property
    def is_displayed(self):
        return in_server_collect_logs and self.title.text == self.form_expected_title()  

class ServerCollectLogsMasterView(ServerCollectLogsView):
    @property
    def is_displayed(self):
        return in_server_collect_logs and self.title.text == self.form_expected_title()  
```

lina-nikiforova: @izapolsk  yep, yours is much better :+1: 
I'm gonna implement this

izapolsk: yeah, it is a pity that we managed to wrap up only one statement. I thought it could be made more neat :(

PR #8148, opened 2018-11-15
[1LP][RFR] whitelist test_v2v in xunit_tools

mshriver:
This is effectively temporary, as #7690 will replace this use.


cfme/fixtures/xunit_tools.py
```diff
@@ -15,6 +15,8 @@
 whitelist = [
     r'cfme/tests/infrastructure/test_quota_tagging.py::test_.*\[.*rhe?v',
     r'test_tenant_quota.py',
+    r'test_v2v.*',
+    r'test_migration.*'
```
kedark3: `this would be test_*migration*`

mshriver: whitelisting whole directory actually.

mkoura: How about just `r'cfme/tests/v2v` ?

PR #8154, opened 2018-11-16
[1LP][RFR] fixed finalizer cleanup for test_tagvis_infra_object

lina-nikiforova:
{{ pytest: -v cfme/tests/cloud_infra_common/test_tag_objects.py::test_tagvis_infra_object }}

1. fixed finalizer cleanup for test_tagvis_infra_object
2. fixed checking visibility for hosts

### DIFF
--- a/cfme/fixtures/tag.py
+++ b/cfme/fixtures/tag.py
@@ -2,6 +2,7 @@ import fauxfactory
 import pytest
 
 from cfme.base.credential import Credential
+from cfme.infrastructure.host import Host
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.blockers import BZ
 from cfme.utils.log import logger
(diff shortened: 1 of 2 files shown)


cfme/fixtures/tag.py
```diff
@@ -119,6 +120,10 @@ def _check_item_visibility(vis_object, vis_expect):
                 vis_object.remove_tag(tag=tag)
         with user_restricted:
             try:
+                if isinstance(vis_object, Host):
+                    # need to remove the link to the provider from the host,
+                    # so the navigation goes Compute -> Infrastructure -> Hosts, not Providers
+                    vis_object.parent.filters.update({'provider': None})
```
digitronik: :+1: 

@mshriver we are getting this diffraction due to use of `collection.filter` (provider key) object in navigation. I think we have to take look at it.

Issue #8152, opened 2018-11-16
[Bug] Waiting for Task to be successfully finished throws an error


jan-zmeskal:
[Task.is_successfully_finished](https://github.com/ManageIQ/integration_tests/blob/3e277887f1769b7f2cb558d0ef39b66b096f0501/cfme/configure/tasks.py#L155) throws this [error](http://pastebin.test.redhat.com/670018). It's a blocker for tests using SSA, becase they call [BaseVM.smartstate_scan](https://github.com/ManageIQ/integration_tests/blob/3e277887f1769b7f2cb558d0ef39b66b096f0501/cfme/common/vm.py#L341).

PR #8156, opened 2018-11-16
[1LP][RFR] Making activation of SmartProxy Affinity idempotent

jan-zmeskal:
I cannot really provide valid verification, since SSA tests are blocker by #8152


jawatts:
Looks good, tested this change locally.

PR #8165, opened 2018-11-19
[WIP][NOTEST] sprout providers clean up

izapolsk:


sprout/sprout/settings.py
```diff
@@ -158,6 +158,19 @@
     }
 }
 
+CLEANUP_RULES = [
+    # appliance name , max appliance lifetime
+    {'name': r'^test', 'lifetime': dict(hours=1)},
```
mshriver: I'm thinking these don't need to be in this cleanup, since they aren't related to any machines that sprout deploys. The intent of these cleanup strings in our current VM cleanup is for test resources, which I don't think sprout should be trying to cleanup.

We shouldn't be creating test resources on sprout providers - and if we do (for whatever reason) I think I would like to keep cleanup of VM's named this way to the cleanup_old_vms script, which can be easily run against sprout providers.

Thoughts?

sprout/sprout/settings.py
```diff
@@ -158,6 +158,19 @@
...
+    {'name': r'^(long-test|i-)', 'lifetime': dict(hours=6)},
+    {'name': r'^(external-test|s_tpl|s_appl)', 'lifetime': dict(hours=12)},
+    {'name': r'^jenkins-template', 'lifetime': dict(hours=4)},
```
mshriver: I'm thinking we make all of these  `r'^jenkins(-|_)'` - so we match dash or underscore and don't have to adjust in future

mshriver: above comment for jenkins-template, jenkins, sprout, dockerbot

sprout/appliances/tasks.py
```diff
@@ -152,6 +152,121 @@ def kill_unused_appliances(self):
...
+@singleton_task()
+def delete_nonexistent_appliances(self):
+    """Goes through orphaned appliances' objects and deletes them from the database."""
+    expiration_time = (timezone.now() - timedelta(**settings.ORPHANED_APPLIANCE_GRACE_TIME))
+    for appliance in Appliance.objects.filter(ready=True).all():
+        if appliance.name in redis.renaming_appliances:
+            continue
+        if appliance.power_state == Appliance.Power.ORPHANED:
+            if appliance.power_state_changed > expiration_time:
+                # Ignore it for now
```
mshriver: Thinking we should log this skip case, either info/debug level.

sprout/appliances/tasks.py
```diff
@@ -152,6 +152,121 @@ def kill_unused_appliances(self):
...
+                continue
+            self.logger.info(
+                "I will delete orphaned appliance {}/{}".format(appliance.id, appliance.name))
```
mshriver: Consider including provider info in log message so we don't have to look it up.

sprout/appliances/tasks.py
```diff
@@ -152,6 +152,121 @@ def kill_unused_appliances(self):
...
+                    # Someone managed to delete the appliance pool before
+                    appliance.appliance_pool = None
+                    appliance.save(update_fields=['appliance_pool'])
+                    appliance.delete()
+                else:
+                    raise  # No diaper pattern here!
+    # If something happened to the appliance provisioning process, just delete it to remove
+    # the garbage. It will be respinned again by shepherd.
+    # Grace time is specified in BROKEN_APPLIANCE_GRACE_TIME
+    expiration_time = (timezone.now() - timedelta(**settings.BROKEN_APPLIANCE_GRACE_TIME))
```
mshriver: Almost seems like a 2nd singleton_task, one for orphaned appliances and one for broken appliances?

sprout/appliances/tasks.py
```diff
@@ -152,6 +152,121 @@ def kill_unused_appliances(self):
...
+    for appliance in Appliance.objects.filter(ready=False, marked_for_deletion=False).all():
+        if appliance.status_changed < expiration_time:
+            self.logger.info("Killing broken appliance {}/{}".format(appliance.id, appliance.name))
```
mshriver: again consider including provider name/key as well in this log message.

sprout/sprout/settings.py
izapolsk: I guess you're right. Off the top of my head, appliances which are used in tests aren't renamed to test-*

sprout/appliances/tasks.py
izapolsk: yep, there even one more task. 
We have there:
1. kill unused appliances;
2. kill nonexistent appliances; 
3. kill lost appliances.

I don't want to combine everything into one task because celery prefers to have many small tasks instead of one big.

sprout/appliances/tasks.py
```diff
@@ -152,6 +152,130 @@ def kill_unused_appliances(self):
...
+            expiration_time = timezone.now() - timedelta(**rule['lifetime'])
+            for name in vm_names:
+                if not re.match(rule['name'], name):
+                    continue
+
+                if provider.provider_type != 'openshift':
+                    try:
+                        vm_creation_time = provider.api.get_vm(name).creation_time
+                    except Exception as e:
+                        self.logger.exception("Couldn't get vm {vm} timestamp (prov: {p}) "
```
mshriver: If this exception occurs, vm_creation_time won't be set and L240 will raise exception yes?

izapolsk: it will just put exception warn in the log file and proceed to next vm

mshriver: Nvm, the continue breaks us out.

PR #8172, opened 2018-11-20
[1LP][RFR] Custom button: Service and Generic object test

digitronik:
Purpose or Intent
=================
- Service and Generic object custom button tests base
- display test for service and generic object
- Fix sometime `ssui` login problem with `wait`

{{pytest: cfme/tests/automate/custom_button/test_service_objects.py -vvv}}

### DIFF
--- a/cfme/base/ssui.py
+++ b/cfme/base/ssui.py
@@ -125,7 +125,6 @@ LOGIN_METHODS = ['click_on_login', 'press_enter_after_password']
 @MiqImplementationContext.external_for(Server.login, ViaSSUI)
 def login(self, user=None, method=LOGIN_METHODS[-1]):
     if not user:
-        from cfme.configure.access_control import User
         username = conf.credentials['default']['username']
         password = conf.credentials['default']['password']
         cred = Credential(principal=username, secret=password)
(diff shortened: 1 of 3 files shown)


cfme/base/ssui.py
```diff
@@ -125,7 +125,6 @@ def address(self):
 @MiqImplementationContext.external_for(Server.login, ViaSSUI)
 def login(self, user=None, method=LOGIN_METHODS[-1]):
     if not user:
-        from cfme.configure.access_control import User
```
digitronik: Note: Imported but not used

cfme/services/myservice/ui.py
```diff
@@ -339,6 +340,9 @@ class MyServiceAll(CFMENavigateStep):
     def step(self, *args, **kwargs):
         self.prerequisite_view.navigation.select('Services', 'My Services')
 
+    def resetter(self, *args, **kwargs):
+        self.view.myservice.tree.click_path('Active Services')
```
digitronik: Note: reset view so it will maintain `All` view every time...

cfme/base/ssui.py
```diff
@@ -145,7 +144,7 @@ def login(self, user=None, method=LOGIN_METHODS[-1]):
     time.sleep(3)
     login_view.flash.assert_no_error()
     self.browser.plugin.ensure_page_safe()
-    assert logged_in_view.is_displayed
+    logged_in_view.wait_displayed('10s')
```
sshveta: you can assert for this too .
assert logged_in_view.wait_displayed('10s')

digitronik: @sshveta I think, no need to assert... It will raise `Timeout` error. If fails to land within `10s` after a call.

PR #8173, opened 2018-11-20
[1LP][RFR] fix test_vmware_alarm_selection_does_not_fail by restructuring test

john-dupuy:
Purpose or Intent
=================

__Fixing__ `test_vmware_alarm_selection_does_not_fail` by restructuring the test to assert that an alert with `evaluate=("VMware Alarm", {})` is successfully created. Previously, this test would expect the alert creation to fail. It is written against BZ 1231889. 

{{ pytest: --long-running cfme/tests/control/test_bugs.py::test_vmware_alarm_selection_does_not_fail }}

### DIFF
--- a/cfme/tests/control/test_bugs.py
+++ b/cfme/tests/control/test_bugs.py
@@ -282,35 +282,25 @@ def test_control_identical_descriptions(request, create_function, collections, a
 
 
 @pytest.mark.meta(blockers=[1231889], automates=[1231889])
-def test_vmware_alarm_selection_does_not_fail(alert_collection):
+def test_vmware_alarm_selection_does_not_fail(request, alert_collection):
     """Test the bug that causes CFME UI to explode when VMware Alarm type is selected.
+        We assert that the alert using this type is simply created. Then we destroy
+        the alert.
 
     Metadata:
         test_flag: alerts
     """
-    alert_obj = alert_collection.instantiate(
-        "Trigger by CPU {}".format(fauxfactory.gen_alpha(length=4)),
-        active=True,
-        based_on="VM and Instance",
-        evaluate=("VMware Alarm", {}),
-        notification_frequency="5 Minutes"
-    )
     try:
-        alert_collection.create(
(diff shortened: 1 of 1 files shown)


izapolsk:
@john-dupuy, this test is created to test some old bug - `automates=[1231889]`. it expects no error during alert creation. Are you sure your changes are valid in such conditions ?

john-dupuy:
@izapolsk, this test is weird. Apparently a bug occurred when VMWare Alarm was selected in the Alert creation -- the page would "explode". This test is simply ensuring that the bug does not occur and in order for the test to pass -- the alert must fail to be created. My changes to `alert_collection.create()` function were making the alert successfully and so the test was failing. c.f. 

https://github.com/ManageIQ/integration_tests/blob/3f5a0053c07c9c45f99cda830c43808552117937/cfme/tests/control/test_bugs.py#L298-L313

izapolsk:
@john-dupuy, I doubt it should fail. I smell a rat. 
If I understand this test right, it should check that appliance doesn't get broken after some option is chosen in some dialog. So, it should just fail if appliance gets broken. Otherwise, it should finish alert creation with such option and that's it.

So, I'd ask you to fully revise this test and rewrite it if it is necessary.

john-dupuy:
@izapolsk more specifically, it is testing that whether the appropriate flash message appears:
```
     view.flash.assert_message("At least one of E-mail, SNMP Trap, Timeline Event, or" 
         " Management Event must be configured")
```

This flash message will appear only when the alert creation fails. I agree that since the bug is simply that CFME explodes when "VMware" alarm is selected, the test should be simplified. Perhaps the test was written in this way so that no additional cleanup was necessary, since the alert was never created. Regardless, I will work on a restructuring of this test.

john-dupuy:
@jawatts Yes, I noticed this last night and submitted https://bugzilla.redhat.com/show_bug.cgi?id=1654896 against this issue. I will add a blocker.

PR #8174, opened 2018-11-20
[RFR]v2v edit mapping test

sshveta:
{{ pytest: cfme/tests/v2v/test_v2v_migrations_ui.py --use-provider rhv42 --use-provider vsphere67-nested --provider-limit 2 -vvvv --long-running}}

### DIFF
--- a/cfme/tests/v2v/test_v2v_migrations.py
+++ b/cfme/tests/v2v/test_v2v_migrations.py
@@ -562,16 +562,3 @@ def test_migration_special_char_name(request, appliance, v2v_providers, host_cre
     src_vm = form_data_vm_obj_single_datastore.vm_list[0]
     migrated_vm = get_migrated_vm_obj(src_vm, v2v_providers.rhv_provider)
     assert src_vm.mac_address == migrated_vm.mac_address
-
-
-@pytest.mark.parametrize('form_data_vm_obj_single_datastore', [['nfs', 'nfs', rhel7_minimal],],
-                        indirect=True)
-def test_edit_mapping_description(request, appliance, host_creds,
-                                conversion_tags, form_data_vm_obj_single_datastore):
-    infrastructure_mapping_collection = appliance.collections.v2v_mappings
-    mapping = infrastructure_mapping_collection.create(form_data_vm_obj_single_datastore.form_data)
-    edited_form_data = {
-        'general': {
-            'description': "my edited description"}
-    }
-    mapping.update(edited_form_data)
--- a/cfme/tests/v2v/test_v2v_migrations.py
+++ b/cfme/tests/v2v/test_v2v_migrations.py
@@ -562,3 +562,16 @@ def test_migration_special_char_name(request, appliance, v2v_providers, host_cre
     src_vm = form_data_vm_obj_single_datastore.vm_list[0]
     migrated_vm = get_migrated_vm_obj(src_vm, v2v_providers.rhv_provider)
     assert src_vm.mac_address == migrated_vm.mac_address
+
+
+@pytest.mark.parametrize('form_data_vm_obj_single_datastore', [['nfs', 'nfs', rhel7_minimal],],
+                        indirect=True)
+def test_edit_mapping_description(request, appliance, host_creds,
+                                conversion_tags, form_data_vm_obj_single_datastore):
+    infrastructure_mapping_collection = appliance.collections.v2v_mappings
+    mapping = infrastructure_mapping_collection.create(form_data_vm_obj_single_datastore.form_data)
+    edited_form_data = {
+        'general': {
+            'description': "my edited description"}
+    }
+    mapping.update(edited_form_data)
--- a/cfme/tests/v2v/test_v2v_migrations_ui.py
+++ b/cfme/tests/v2v/test_v2v_migrations_ui.py
@@ -372,16 +372,14 @@ def test_migration_rbac(appliance, new_credential, v2v_providers):
                                                     'rbac should allow this')
 
 
-@pytest.mark.uncollectif(lambda appliance: appliance.version < '5.10')
+@pytest.mark.ignore_stream("5.9")
 @pytest.mark.parametrize('form_data_single_datastore', [['nfs', 'nfs']], indirect=True)
 def test_edit_mapping_description(appliance, v2v_providers, form_data_single_datastore,
                                     host_creds, conversion_tags, soft_assert):
     infrastructure_mapping_collection = appliance.collections.v2v_mappings
     mapping = infrastructure_mapping_collection.create(form_data_single_datastore)
-    edited_name = 'edited_name_{}'.format(fauxfactory.gen_alphanumeric())
     edited_form_data = {
         'general': {
-            'name': edited_name,
             'description': "my edited description"},
         'cluster': {},
         'datastore': {},
@@ -392,4 +390,4 @@ def test_edit_mapping_description(appliance, v2v_providers, form_data_single_dat
     view = navigate_to(infrastructure_mapping_collection, 'All')
     infrastructure_mapping_collection.find_mapping(mapping)
     mapping_list = view.infra_mapping_list
-    soft_assert(str(mapping_list.get_map_description(mapping.name)) == mapping.description)
+    soft_assert(str(mapping_list.get_map_description(mapping.name)) == "my edited description")
--- a/cfme/tests/v2v/test_v2v_migrations_ui.py
+++ b/cfme/tests/v2v/test_v2v_migrations_ui.py
@@ -390,7 +390,6 @@ def test_edit_mapping_description(appliance, v2v_providers, form_data_single_dat
 
     view = navigate_to(infrastructure_mapping_collection, 'All')
     if appliance.version >= '5.10':
-        mapping_name = infrastructure_mapping_collection.find_mapping(mapping)
-        soft_assert(mapping_name == edited_name)
-        soft_assert(mapping.description in view.infra_mapping_list.read())
-        soft_assert(mapping.description == "my edited description")
+        infrastructure_mapping_collection.find_mapping(mapping)
+        mapping_list = view.infra_mapping_list
+        soft_assert(str(mapping_list.get_map_description(mapping.name)) == mapping.description)
--- a/cfme/tests/v2v/test_v2v_migrations_ui.py
+++ b/cfme/tests/v2v/test_v2v_migrations_ui.py
@@ -5,7 +5,7 @@ import pytest
 from widgetastic.exceptions import NoSuchElementException
 
 from cfme.exceptions import ItemNotFound
-from cfme.fixtures.provider import small_template, rhel7_minimal
+from cfme.fixtures.provider import small_template
 from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.markers.env_markers.provider import ONE_PER_VERSION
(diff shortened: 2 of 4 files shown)


cfme/tests/v2v/test_v2v_migrations.py
```diff
@@ -562,3 +562,16 @@ def _cleanup():
     src_vm = form_data_vm_obj_single_datastore.vm_list[0]
     migrated_vm = get_migrated_vm_obj(src_vm, v2v_providers.rhv_provider)
     assert src_vm.mac_address == migrated_vm.mac_address
+
+
+@pytest.mark.parametrize('form_data_vm_obj_single_datastore', [['nfs', 'nfs', rhel7_minimal],],
+                        indirect=True)
+def test_edit_mapping_description(request, appliance, host_creds,
```
kedark3: Looks good. Maybe we can add few more tests around different edits and validate this by checking the Infra Mapping details in the `InfraMappingsList`

cfme/tests/v2v/test_v2v_migrations.py
```diff
@@ -562,3 +562,16 @@ def _cleanup():
...
+                                conversion_tags, form_data_vm_obj_single_datastore):
+    infrastructure_mapping_collection = appliance.collections.v2v_mappings
+    mapping = infrastructure_mapping_collection.create(form_data_vm_obj_single_datastore.form_data)
+    edited_form_data = {
+        'general': {
+            'description': "my edited description"}
+    }
+    mapping.update(edited_form_data)
```
kedark3: we need to assert this information being updated.

cfme/v2v/migrations.py
```diff
@@ -125,10 +125,12 @@ def fill(self, values):
                        ...
                     }
         """
-        for mapping in values['mappings']:
-            self.source_clusters.fill(mapping['sources'])
-            self.target_clusters.fill(mapping['target'])
-            self.add_mapping.click()
+        if values:
```
kedark3: :+1:

cfme/v2v/migrations.py
```diff
@@ -664,6 +679,11 @@ class InfrastructureMapping(BaseEntity):
     description = attr.ib(default=None)
     form_data = attr.ib(default=None)
 
+    def update(self, form_data):
```
kedark3: **Question**: Why do you think this should be in Entity while Create and Delete were in Collection? We may need to consider all the operations and their placement.

cfme/tests/v2v/test_v2v_migrations.py
sshveta: Kedar will add that in next stage . For now I wanted to check if the framework to edit mapping works , so starting with just editing description

cfme/v2v/migrations.py
sshveta: This should be in entity as update is for a single mapping . generally we add delete too in Baseentity and Methods like create in collection .

jawatts:
@sshveta Before I take a look, can you take a look at travis and PRT? Both are failing. Thank you!

cfme/tests/v2v/test_v2v_migrations_ui.py
```diff
@@ -370,3 +370,27 @@ def test_migration_rbac(appliance, new_credential, v2v_providers):
         # Checks migration option is enabled in navigation
         assert 'Migration' in nav_tree['Compute'], ('Migration not found in nav tree, '
                                                     'rbac should allow this')
+
+
+@pytest.mark.parametrize('form_data_vm_obj_single_datastore', [['nfs', 'nfs', rhel7_minimal],],
```
kedark3: **REQUIRED**  :You can use `form_data_single_datastore` and avoid using `rhel7_minimal`.

cfme/v2v/migrations.py
```diff
@@ -125,14 +126,14 @@ def fill(self, values):
...
-            self.target_clusters.fill(mapping['target'])
-            self.add_mapping.click()
+        if values:
+            for mapping in values['mappings']:
+                self.source_clusters.fill(mapping['sources'])
+                self.target_clusters.fill(mapping['target'])
+                self.add_mapping.click()
+
         was_change = not self.mappings_tree.is_empty
-        if was_change:
```
kedark3: **Question**: Why should we remove this check?

cfme/v2v/migrations.py
```diff
@@ -166,16 +167,16 @@ def fill(self, values):
...
+                if self.cluster_selector.is_displayed:
+                    self.cluster_selector.fill(cluster)
+                for mapping in values[cluster]['mappings']:
+                    self.source_datastores.fill(mapping['sources'])
+                    self.target_datastores.fill(mapping['target'])
+                    self.add_mapping.click()
         was_change = not self.mappings_tree.is_empty
-        if was_change:
-            self.logger.info("Fill operation was successful, click Next.")
+        if self.next_btn.wait_displayed():
```
kedark3: **Question**: Can we check this for each page of the form? And also, wait_displayed() would be wrong,  maybe we should check `is_enabled `? Because when we don't have any mappings, it is usually greyed out.

cfme/v2v/migrations.py
sshveta: @kedar : edit form does not work with 'was_change' . only some values are changed while editing and hence the form does not change , but we still need to click on next button

cfme/v2v/migrations.py
sshveta: @kedark3 :without wait displayed , buttons are not clicked , it needs some time to display

cfme/tests/v2v/test_v2v_migrations.py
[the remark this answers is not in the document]
sshveta: Done

cfme/tests/v2v/test_v2v_migrations_ui.py
sshveta: Done

cfme/tests/v2v/test_v2v_migrations_ui.py
```diff
@@ -370,3 +370,26 @@ def test_migration_rbac(appliance, new_credential, v2v_providers):
...
+            'name': edited_name,
+            'description': "my edited description"},
+        'cluster': {},
+        'datastore': {},
+        'network': {}
+    }
+    mapping.update(edited_form_data)
+
+    view = navigate_to(infrastructure_mapping_collection, 'All')
+    if appliance.version >= '5.10':
```
kedark3: Why is this only for `>5.10` ??

cfme/v2v/migrations.py
kedark3: Did you try to run any other tests with these changes? We don't want this to break other things. Maybe try to run other UI tests and see if they still work.

cfme/tests/v2v/test_v2v_migrations_ui.py
jawatts: In additions to @kedark3, there are no checks for < 5.10

cfme/v2v/migrations.py
jawatts:
```In[10]: a = {'mappings':[]}
In[11]: if a:
   ...:     print "This is a test"
   ...: 
This is a test
```
I don't think this test really gets you very much. We should explicitly check that all the right keys are there.

cfme/v2v/migrations.py
```diff
@@ -166,16 +167,17 @@ def fill(self, values):
...
                     }
         """
-        for cluster in values:
-            if self.cluster_selector.is_displayed:
-                self.cluster_selector.fill(cluster)
-            for mapping in values[cluster]['mappings']:
-                self.source_datastores.fill(mapping['sources'])
-                self.target_datastores.fill(mapping['target'])
-                self.add_mapping.click()
+        if values:
```
jawatts: same comment as above

cfme/v2v/migrations.py
```diff
@@ -166,16 +167,17 @@ def fill(self, values):
...
+                if self.cluster_selector.is_displayed:
+                    self.cluster_selector.fill(cluster)
+                for mapping in values[cluster]['mappings']:
+                    self.source_datastores.fill(mapping['sources'])
+                    self.target_datastores.fill(mapping['target'])
+                    self.add_mapping.click()
         was_change = not self.mappings_tree.is_empty
-        if was_change:
-            self.logger.info("Fill operation was successful, click Next.")
+        self.cancel_btn.wait_displayed()
```
jawatts: Same comment @kedark3 made above about removing `if was_change`

cfme/tests/v2v/test_v2v_migrations_ui.py
sshveta: Not applicable for version 5.9 . Uncollected.

cfme/v2v/migrations.py
sshveta: will see the result of this in PRT and then run other tests too

cfme/v2v/migrations.py
```diff
@@ -664,6 +678,17 @@ class InfrastructureMapping(BaseEntity):
     description = attr.ib(default=None)
     form_data = attr.ib(default=None)
 
+    def update(self, form_data):
```
jawatts: It be good to have some description on what is expected for form_data

cfme/v2v/migrations.py
```diff
@@ -664,6 +678,17 @@ class InfrastructureMapping(BaseEntity):
...
+        view = navigate_to(self, 'Edit', wait_for_view=20)
+        assert view.wait_displayed()
+        view.form.fill(form_data)
+        if form_data['general']['name']:
```
jawatts: This will throw an exception is `general` or `name` do not exist

cfme/tests/v2v/test_v2v_migrations_ui.py
```diff
@@ -370,3 +370,26 @@ def test_migration_rbac(appliance, new_credential, v2v_providers):
         # Checks migration option is enabled in navigation
         assert 'Migration' in nav_tree['Compute'], ('Migration not found in nav tree, '
                                                     'rbac should allow this')
+
+
+@pytest.mark.uncollectif(lambda appliance: appliance.version < '5.10')
```
jawatts: Please use ```@pytest.mark.ignore_stream("5.9")``` here

cfme/v2v/migrations.py
sshveta: @jawatts : These Views are filled only if the form_data has respective data in it .
For ex: If 
form_data = {
        'general': {
            'description': "my edited description"},
        'cluster': {},
        'datastore': {},
        'network': {}
    }

So if cluster is present only then that view is filled . and if values check if it has mappings or empty , so working in this case

sshveta:
PRT passed for this test . Now running other tests.

kedark3:
:+1:  LGTM, PRT failures are not relevant to PR in Run # 28650

cfme/v2v/migrations.py
mshriver: I think you two are in a way talking across each other here, both in a way missing the point a bit :)

1. Justin's correct, checking values is empty before iterating without actually doing something with that information (like logging that the clusters fill was empty) doesn't get us much.

2. Shveta's correct too, its not a problem that its empty in the fill context, and we just ultimately want to click next after skipping this.

All of this is due to the fact that we have a fill dictionary (that really should be a list anyway) that doesn't match our view/widget structure and names, so we have to try and mangle the data to fill, instead of letting `View.fill` manage it for us.

In this case, I don't think you should add a check of `if values`, because `values['mappings']` could still raise an exception anyway (the check is incomplete).

Simply using `for mapping in (values.get('mappings', []) if isinstance(values, dict) else []):` protects from the KeyError, values not being a dict, and short-circuits the loop if the `values` dict is empty or contains an unexpected key.

I have plenty of comments about the fill dicts not matching the view structure, but I think that might be out of scope for the changes proposed here.  Doing so would remove our need to have these corner cases during fill in the first place.

```
In [3]: values = {}                                                                                                                        

In [4]: for mapping in (values.get('mappings', []) if isinstance(values, dict) else []): 
   ...:     print('true') 
   ...:                                                                                                                                    

In [5]: values = {'thing'}                                                                                                                 

In [6]: for mapping in (values.get('mappings', []) if isinstance(values, dict) else []): 
   ...:     print('true') 
   ...:                                                                                                                                    

In [7]: values = []                                                                                                                        

In [8]: for mapping in (values.get('mappings', []) if isinstance(values, dict) else []): 
   ...:     print('true') 
   ...:                                                                                                                                    

In [9]: values = 'stick'                                                                                                                   

In [10]: for mapping in (values.get('mappings', []) if isinstance(values, dict) else []): 
    ...:     print('true') 
    ...:                                                                                                                                   

In [11]: values = {'mappings':[]}                                                                                                          

In [12]: for mapping in (values.get('mappings', []) if isinstance(values, dict) else []): 
    ...:     print('true') 
    ...:      
    ...:                                                                                                                                   

In [13]: values = {'mappings':[1]}                                                                                                         

In [14]: for mapping in (values.get('mappings', []) if isinstance(values, dict) else []): 
    ...:     print('true') 
    ...:      
    ...:                                                                                                                                   
true

In [15]: values = {'mappings':[1, 2]}                                                                                                      

In [16]: for mapping in (values.get('mappings', []) if isinstance(values, dict) else []): 
    ...:     print('true') 
    ...:      
    ...:                                                                                                                                   
true
true

```

sshveta:
Got into some rebase issues . opening new PR https://github.com/ManageIQ/integration_tests/pull/8242
Closing this one

PR #8183, opened 2018-11-22
[1LP][RFR] Custom Button: Cloud-Visibility and Enablement

digitronik:
Purpose or Intent
=================
- Visibility and enablement for cloud custom button objects

{{pytest: cfme/tests/automate/custom_button/test_cloud_objects.py::test_custom_button_expression -vvv}}

### DIFF
--- a/cfme/tests/automate/custom_button/test_cloud_objects.py
+++ b/cfme/tests/automate/custom_button/test_cloud_objects.py
@@ -303,3 +303,65 @@ def test_custom_button_automate(appliance, request, submit, setup_objs, button_g
                 assert False, "Expected {} requests not found in automation log".format(
                     str(expected_count)
                 )
+
+
+@pytest.mark.uncollectif(
+    lambda appliance, button_group: not bool([obj for obj in OBJ_TYPE_59 if obj in button_group])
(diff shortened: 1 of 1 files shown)


digitronik:
For `5.10` test will skip on the current build. It's due to provider refresh bug.

jawatts:
@digitronik is this bug fixed in the latest 5.10 build? Or is it worth adding the blocker to the test case?

cfme/tests/automate/custom_button/test_cloud_objects.py
```diff
@@ -303,3 +303,65 @@ def test_custom_button_automate(appliance, request, submit, setup_objs, button_g
...
+        system="Request",
+        request="InspectMe",
+        **exp
+    )
+    request.addfinalizer(button.delete_if_exists)
+
+    tag_cat = appliance.collections.categories.instantiate(
+        name="department", display_name="Department"
+    )
+    tag = tag_cat.collections.tags.instantiate(name="engineering", display_name="Engineering")
```
mshriver: The test up to here seems to be duplicated across many of these tests, and would likely be better as a fixture with some indirect parametrization.  (TODO)

PR #8191, opened 2018-11-22
[1LP][RFR] added all method for keypairs, tenant and flavor collections

lina-nikiforova:
{{ pytest: -v cfme/tests/cloud_infra_common/test_tag_objects.py::test_tag_cloud_objects }}

fixed 
```
AttributeError: 'KeyPairCollection' object has no attribute 'all'
AttributeError: 'TenantCollection' object has no attribute 'all'
AttributeError: 'FlavorCollection' object has no attribute 'all'
```
that I introduced by fixing `test_tag_infra_objects tests` in #8144

### DIFF
--- a/cfme/cloud/flavor.py
+++ b/cfme/cloud/flavor.py
@@ -13,6 +13,7 @@ from cfme.common import Taggable
 from cfme.exceptions import FlavorNotFound, ItemNotFound
 from cfme.modeling.base import BaseEntity, BaseCollection
 from cfme.utils.appliance.implementations.ui import CFMENavigateStep, navigator, navigate_to
+from cfme.utils.providers import get_crud_by_name
 from widgetastic_manageiq import (
     BaseEntitiesView, ItemsToolBarViewSelector, SummaryTable, Text, Table, Accordion, ManageIQTree,
     PaginationPane, Search)
@@ -209,6 +210,22 @@ class FlavorCollection(BaseCollection):
         view.flash.assert_no_error()
         flavor = self.instantiate(name, provider, ram, vcpus, disk, swap, rxtx, is_public)
         return flavor
+
+    def all(self):
+        provider = self.filters.get('provider')  # None if no filter, need for entity instantiation
+        view = navigate_to(self, 'All')
+        result = []
+        for _ in view.entities.paginator.pages():
+            flavors = view.entities.get_all()
+            for flavor in flavors:
+                if provider is not None:
+                    if flavor.data['cloud_provider'] == provider.name:
+                        entity = self.instantiate(flavor.data['name'], provider)
+                else:
+                    entity = self.instantiate(flavor.data['name'],
+                                              get_crud_by_name(flavor.data['cloud_provider']))
+                result.append(entity)
+        return result
 
 
 @navigator.register(FlavorCollection, 'All')
--- a/cfme/cloud/keypairs.py
+++ b/cfme/cloud/keypairs.py
@@ -120,7 +120,7 @@ class KeyPair(BaseEntity, Taggable):
     _param_name = "KeyPair"
 
     name = attr.ib()
-    provider = attr.ib()
+    provider = attr.ib(default=None)
     public_key = attr.ib(default="")
 
     def delete(self, cancel=False, wait=False):
@@ -207,6 +207,10 @@ class KeyPairCollection(BaseCollection):
         assert view.is_displayed
         view.flash.assert_success_message(flash_message)
         return self.instantiate(name, provider, public_key=public_key)
+
+    def all(self):
+        view = navigate_to(self, 'All')
+        return [self.instantiate(name) for name in view.entities.all_entity_names]
 
 
 @navigator.register(KeyPairCollection, 'All')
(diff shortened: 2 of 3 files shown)


lina-nikiforova:
@digitronik thanks, good point! I have rewritten these methods and it works without filtering.

cfme/cloud/flavor.py
```diff
@@ -210,6 +211,16 @@ def create(self, name, provider, ram, vcpus, disk, swap, rxtx, is_public=True, c
         flavor = self.instantiate(name, provider, ram, vcpus, disk, swap, rxtx, is_public)
         return flavor
 
+    def all(self):
+        provider = self.filters.get('provider')  # None if no filter, need for entity instantiation
+        view = navigate_to(self, 'All')
+        result = []
+        for e in view.entities.get_all():
+            entity = self.instantiate(e.data['name'], provider or
+                                      get_crud_by_name(e.data['cloud_provider']))
```
digitronik: Required:
If you pass `OSP` with a `filter` it will return `ec2` entities with a provider as `OSP`. what do you think?

cfme/cloud/tenant.py
```diff
@@ -316,6 +317,16 @@ def delete(self, *tenants):
         # TODO: Assert deletion flash message for selected tenants
         # it is not shown in current UI, so not asserting
 
+    def all(self):
+        provider = self.filters.get('provider')  # None if no filter, need for entity instantiation
+        view = navigate_to(self, 'All')
+        result = []
+        for e in view.entities.get_all():
+            entity = self.instantiate(e.data['name'], provider or
```
digitronik: Required:
same as above...

cfme/cloud/flavor.py
```diff
@@ -210,6 +211,16 @@ def create(self, name, provider, ram, vcpus, disk, swap, rxtx, is_public=True, c
         flavor = self.instantiate(name, provider, ram, vcpus, disk, swap, rxtx, is_public)
         return flavor
 
+    def all(self):
+        provider = self.filters.get('provider')  # None if no filter, need for entity instantiation
+        view = navigate_to(self, 'All')
+        result = []
+        for e in view.entities.get_all():
```
digitronik: Required:
It will not cover all entities. I think we should go with `surf_pages=True`. It will cover all `entities` even if more pages.  same suggestion of other as well.

cfme/cloud/tenant.py
```diff
@@ -316,6 +317,23 @@ def delete(self, *tenants):
...
+        provider = self.filters.get('provider')  # None if no filter, need for entity instantiation
+        view = navigate_to(self, 'All')
+        result = []
+        for _ in view.entities.paginator.pages():
+            tenants = view.entities.get_all()
+            for tenant in tenants:
+                if provider is not None:
+                    if tenant.data['cloud_provider'] == provider.name:
+                        entity = self.instantiate(tenant.data['name'],
+                                                  get_crud_by_name(tenant.data['cloud_provider']))
```
digitronik: Required:
@lina-nikiforova you have a `provider object` in your hand, so no need to create it with a name.

cfme/cloud/flavor.py
```diff
@@ -210,6 +211,23 @@ def create(self, name, provider, ram, vcpus, disk, swap, rxtx, is_public=True, c
...
+        provider = self.filters.get('provider')  # None if no filter, need for entity instantiation
+        view = navigate_to(self, 'All')
+        result = []
+        for _ in view.entities.paginator.pages():
+            flavors = view.entities.get_all()
+            for flavor in flavors:
+                if provider is not None:
+                    if flavor.data['cloud_provider'] == provider.name:
+                        entity = self.instantiate(flavor.data['name'],
+                                              get_crud_by_name(flavor.data['cloud_provider']))
```
digitronik: Required:
no need to create a provider object using `get_crud_by_name`. use it  directly

cfme/cloud/keypairs.py
```diff
@@ -208,6 +208,16 @@ def create(self, name, provider, public_key=None, cancel=False):
         view.flash.assert_success_message(flash_message)
         return self.instantiate(name, provider, public_key=public_key)
 
+    def all(self):
+        view = navigate_to(self, 'All')
+        result = []
+        for _ in view.entities.paginator.pages():
+            for name in view.entities.entity_names:
+                # provider is needed to instantiate but keypairs are not tied to the provider
```
mshriver: With what you're pointing out here, I think you should just update the `attr.ib` definition with a default on the Keypair entity:
https://github.com/ManageIQ/integration_tests/blob/master/cfme/cloud/keypairs.py#L123

cfme/cloud/keypairs.py
```diff
@@ -208,6 +208,16 @@ def create(self, name, provider, public_key=None, cancel=False):
         view.flash.assert_success_message(flash_message)
         return self.instantiate(name, provider, public_key=public_key)
 
+    def all(self):
+        view = navigate_to(self, 'All')
+        result = []
+        for _ in view.entities.paginator.pages():
```
mshriver: Understood why you used paginator on the other items, but here you're only using the names from the table. I think you can safely use `view.entityes.all_entity_names` here to not have to surf pages.  Then this can even be squashed into a list comprehension if you want.

PR #8195, opened 2018-11-25
[1lP][RFR] Add test_volume_attach_detach_instance

jadhaj:
Purpose or Intent
=================
Add test: test_volume_attach_detach_instance

{{ pytest: cfme/tests/openstack/cloud/test_volumes.py::test_volume_attach_detach_instance }}

### DIFF
--- a/cfme/storage/volume.py
+++ b/cfme/storage/volume.py
@@ -3,7 +3,7 @@ import attr
 
 from navmazing import NavigateToSibling, NavigateToAttribute
 from widgetastic.widget import TextInput
-from widgetastic_patternfly import BreadCrumb, Button, Dropdown
+from widgetastic_patternfly import BreadCrumb, Button, Dropdown, Input
 from widgetastic_manageiq import Search
 
 from widgetastic.widget import View, Text, NoSuchElementException
@@ -165,6 +165,50 @@ class VolumeSnapshotView(VolumeView):
     cancel = Button('Cancel')
 
 
+class VolumeAttachInstanceEntities(View):
+    breadcrumb = BreadCrumb()
+    title = Text('//div[@id="main-content"]//h1')
+
+
+class AttachInstanceView(VolumeView):
+    @property
+    def is_displayed(self):
+        expected_title = 'Attach Cloud Volume "{name}"'.format(name=self.context['object'].name)
+        return (
+                self.in_volume and
+                self.entities.title.text == expected_title and
+                self.entities.breadcrumb.active_location == expected_title)
+
+    instance = BootstrapSelect('vm_id')
+    mountpoint = Input(name='device_path')
+    attach = Button('Attach')
+    cancel = Button('Cancel')
+    reset = Button('Reset')
+
+    entities = View.nested(VolumeAttachInstanceEntities)
+
+
+class VolumeDetachInstanceEntities(View):
+    breadcrumb = BreadCrumb()
+    title = Text('//div[@id="main-content"]//h1')
+
+
+class DetachInstanceView(VolumeView):
+    @property
+    def is_displayed(self):
+        expected_title = 'Detach Cloud Volume "{name}"'.format(name=self.context['object'].name)
+        return (
+                self.in_volume and
+                self.entities.title.text == expected_title and
+                self.entities.breadcrumb.active_location == expected_title)
+
+    instance = BootstrapSelect(name='vm_id')
+    detach = Button('Detach')
+    cancel = Button('Cancel')
+
+    entities = View.nested(VolumeDetachInstanceEntities)
+
+
 @attr.s
 class Volume(BaseEntity, Updateable, Taggable):
     name = attr.ib()
@@ -243,6 +287,40 @@ class Volume(BaseEntity, Updateable, Taggable):
                 view.save.click()
                 return snapshot_collection.instantiate(name, self.provider)
 
+    def attach_instance(self, name, mountpoint=None, cancel=False, reset=False):
+        view = navigate_to(self, 'AttachInstance')
+
+        # Reset and Attach buttons are only active when view is changed
+        changed = view.fill({'instance': name, 'mountpoint': mountpoint})
+
+        if cancel or not changed:
+            if not changed:
+                logger.info("attach_instance: the form was unchanged")
+            self.click.cancel()
+
+        elif changed:
+            if reset:
+                view.reset.click()
+            else:
+                view.attach.click()
+                view = self.create_view(VolumeDetailsView)
+                view.flash.assert_no_error()
+
+    def detach_instance(self, name, cancel=False):
+        view = navigate_to(self, 'DetachInstance')
+
+        # Detach button is only active when view is changed
+        changed = view.instance.fill(name)
+
+        if cancel or not changed:
+            if not changed:
+                logger.info("detach_instance: the form was unchanged")
+            self.click.cancel()
+        elif changed:
+            view.detach.click()
+            view = self.create_view(VolumeDetailsView)
+            view.flash.assert_no_error()
+
     @property
     def exists(self):
         try:
@@ -303,6 +381,16 @@ class Volume(BaseEntity, Updateable, Taggable):
         """
         view = navigate_to(self, 'Details')
         return int(view.entities.relationships.get_text_of('Cloud Volume Snapshots'))
+
+    @property
+    def instance_count(self):
+        """ number of instances attached to volume.
+
+        Returns:
+            :py:class:`int` instance count.
+        """
+        view = navigate_to(self, 'Details')
+        return int(view.entities.relationships.get_text_of('Instances'))
 
 
 @attr.s
@@ -463,3 +551,23 @@ class VolumeDetailEditTag(CFMENavigateStep):
 
     def step(self, *args, **kwargs):
         self.prerequisite_view.toolbar.policy.item_select('Edit Tags')
+
+
+@navigator.register(Volume, 'AttachInstance')
+class AttachInstance(CFMENavigateStep):
+    VIEW = AttachInstanceView
+    prerequisite = NavigateToSibling('Details')
+
+    def step(self, *args, **kwargs):
+        self.prerequisite_view.toolbar.configuration.item_select('Attach this Cloud Volume to an '
+                                                                 'Instance')
+
+
+@navigator.register(Volume, 'DetachInstance')
+class DetachInstance(CFMENavigateStep):
+    VIEW = DetachInstanceView
+    prerequisite = NavigateToSibling('Details')
+
+    def step(self, *args, **kwargs):
+        self.prerequisite_view.toolbar.configuration.item_select('Detach this Cloud Volume from an '
+                                                                 'Instance')
--- a/cfme/tests/openstack/cloud/test_volumes.py
+++ b/cfme/tests/openstack/cloud/test_volumes.py
@@ -4,7 +4,7 @@ import fauxfactory
 import pytest
 
 from cfme.cloud.provider.openstack import OpenStackProvider
-from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.appliance.implementations.ui import navigate_to, navigator
 from cfme.utils.log import logger
 from cfme.utils.update import update
 from cfme.utils.wait import wait_for, wait_for_decorator
@@ -65,6 +65,16 @@ def volume_with_type(appliance, provider):
         provider.mgmt.capi.volume_types.delete(vol_type)
 
 
+@pytest.fixture(scope='function')
+def new_instance(provider):
+    instance_name = fauxfactory.gen_alpha()
+    collection = provider.appliance.provider_based_collection(provider)
+    instance = collection.create_rest(instance_name, provider)
+    yield instance
+
+    instance.cleanup_on_provider()
+
+
 @pytest.mark.regression
 def test_create_volume(volume, provider):
     assert volume.exists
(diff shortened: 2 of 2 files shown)


cfme/storage/volume.py
```diff
@@ -165,6 +165,23 @@ class VolumeSnapshotView(VolumeView):
     cancel = Button('Cancel')
 
 
+class AttachInstanceView(VolumeView):
+    is_displayed = displayed_not_implemented
```
digitronik: Question:
can you please explain to me why we should not write `is_displayed` property? I think we are getting `title` and `instance` dropdown to check

cfme/storage/volume.py
```diff
@@ -165,6 +165,23 @@ class VolumeSnapshotView(VolumeView):
...
+
+    instance = BootstrapSelect('vm_id')
+    mountpoint = Input(name='device_path')
+    attach = Button('Attach')
+    cancel = Button('Cancel')
+
+
+class DetachInstanceView(VolumeView):
+    is_displayed = displayed_not_implemented
```
digitronik: same question as above....

cfme/storage/volume.py
```diff
@@ -463,3 +506,23 @@ class VolumeDetailEditTag(CFMENavigateStep):
 
     def step(self, *args, **kwargs):
         self.prerequisite_view.toolbar.policy.item_select('Edit Tags')
+
+
+@navigator.register(Volume, 'AttachInstance')
```
digitronik: :+1:

cfme/storage/volume.py
```diff
@@ -243,6 +260,22 @@ def create_snapshot(self, name, cancel=False, reset=False):
                 view.save.click()
                 return snapshot_collection.instantiate(name, self.provider)
 
+
+    def attach_instance(self, name, mountpoint=None):
+        view = navigate_to(self, 'AttachInstance')
+        view.fill({'instance': name})
+        view.fill({'mountpoint': mountpoint})
```
digitronik: Required:
Jad please use `.fill` for view directly no need to use individual widget `fill` method...
like
```
view.fill({'instance': name, 'mountpoint': mountpoint})
```

cfme/storage/volume.py
```diff
@@ -243,6 +260,22 @@ def create_snapshot(self, name, cancel=False, reset=False):
...
+        view.attach.click()
+        view.flash.assert_no_error()
+
+
+    def detach_instance(self, name):
+        view = navigate_to(self, 'DetachInstance')
+        view.fill({'instance': name})
```
digitronik: Here you can use like
`view.instance.fill(name)`

cfme/storage/volume.py
```diff
@@ -243,6 +260,22 @@ def create_snapshot(self, name, cancel=False, reset=False):
                 view.save.click()
                 return snapshot_collection.instantiate(name, self.provider)
 
+
+    def attach_instance(self, name, mountpoint=None):
```
digitronik: Required:
As per your view; User can choice here `cancel` as well. Let's take that in mind and write this method; what you say?
Take `cancle=False` default

cfme/storage/volume.py
```diff
@@ -243,6 +260,22 @@ def create_snapshot(self, name, cancel=False, reset=False):
...
+        view = navigate_to(self, 'AttachInstance')
+        view.fill({'instance': name})
+        view.fill({'mountpoint': mountpoint})
+        view.attach.click()
+        view.flash.assert_no_error()
+
+
+    def detach_instance(self, name):
```
digitronik: same `cancel`. It will help us with writing `Non-Functional tests`

cfme/storage/volume.py
```diff
@@ -304,6 +337,16 @@ def snapshots_count(self):
         view = navigate_to(self, 'Details')
         return int(view.entities.relationships.get_text_of('Cloud Volume Snapshots'))
 
+    @property
+    def instance_count(self):
```
digitronik: :+1:

cfme/tests/openstack/cloud/test_volumes.py
```diff
@@ -110,3 +120,29 @@ def test_edit_volume_with_type(volume_with_type, appliance):
...
     assert not volume_with_type.exists
+
+
+@pytest.mark.regression
+def test_volume_attach_detach_instance(volume, new_instance, appliance):
+    initial_instance_count = volume.instance_count
+    volume.attach_instance(new_instance.name)
+    view = appliance.browser.create_view(navigator.get_class(volume, 'Details').VIEW)
+    view.flash.assert_success_message(
+        'Attaching Cloud Volume "{}" to {} finished'.format(volume.name, new_instance.name))
```
digitronik: Required:
@jadhaj Please use `keys` while formatting... string

cfme/tests/openstack/cloud/test_volumes.py
```diff
@@ -110,3 +120,29 @@ def test_edit_volume_with_type(volume_with_type, appliance):
...
+    @wait_for_decorator(delay=10, timeout=300,
+                        message="Waiting for volume to be attached to instance")
+    def volume_attached_to_instance():
+        new_instance.refresh_relationships()
+        return volume.instance_count > initial_instance_count
+
+    volume.detach_instance(new_instance.name)
+    view = appliance.browser.create_view(navigator.get_class(volume, 'Details').VIEW)
+    view.flash.assert_success_message(
+        'Detaching Cloud Volume "{}" from {} finished'.format(volume.name, new_instance.name))
```
digitronik: `keys`

cfme/storage/volume.py
jadhaj: Actually you are right. I just thoght is is like some methods where is_displayed is not implemented. working on that !

jadhaj:
@digitronik local results:

2018-11-28 06:51:32,741 [I] [cfme] ------- test_volume_attach_detach_instance[openstack-11] result: passed -------- (cfme/fixtures/log.py:70)

cfme/storage/volume.py
```diff
@@ -165,6 +165,49 @@ class VolumeSnapshotView(VolumeView):
...
+    def is_displayed(self):
+        expected_title = 'Attach Cloud Volume "{name}"'.format(name=self.context['object'].name)
+        return (
+                self.in_volume and
+                self.entities.title.text == expected_title and
+                self.entities.breadcrumb.active_location == expected_title)
+
+    instance = BootstrapSelect(name='vm_id')
+    detach = Button('Detach')
+    cancel = Button('Cancel')
```
jawatts: Is there a `Reset` in the detach view as well? If so, it should be added as a `Button()` here

cfme/storage/volume.py
```diff
@@ -165,6 +165,49 @@ class VolumeSnapshotView(VolumeView):
...
+        expected_title = 'Attach Cloud Volume "{name}"'.format(name=self.context['object'].name)
+        return (
+                self.in_volume and
+                self.entities.title.text == expected_title and
+                self.entities.breadcrumb.active_location == expected_title)
+
+    instance = BootstrapSelect('vm_id')
+    mountpoint = Input(name='device_path')
+    attach = Button('Attach')
+    cancel = Button('Cancel')
```
jawatts: On 5.10.0.27, there is a `Reset` button. This should be included as well

cfme/storage/volume.py
```diff
@@ -243,6 +286,28 @@ def create_snapshot(self, name, cancel=False, reset=False):
...
+        view.flash.assert_no_error()
+
+    def detach_instance(self, name, cancel=False):
+        view = navigate_to(self, 'DetachInstance')
+        view.instance.fill(name)
+
+        if cancel:
+            view.cancel.click()
+        else:
+            view.detach.click()
```
jawatts: same comment about `Reset`

cfme/tests/openstack/cloud/test_volumes.py
```diff
@@ -110,3 +120,31 @@ def test_edit_volume_with_type(volume_with_type, appliance):
 def test_delete_volume_with_type(volume_with_type):
     volume_with_type.delete()
     assert not volume_with_type.exists
+
+
+@pytest.mark.regression
+def test_volume_attach_detach_instance(volume, new_instance, appliance):
```
jawatts: I would like to see `Reset` and `Cancel` tested as well, may be out of scope for this PR. Add a `TODO` perhaps?

cfme/tests/openstack/cloud/test_volumes.py
```diff
@@ -110,3 +120,31 @@ def test_edit_volume_with_type(volume_with_type, appliance):
...
+    view = appliance.browser.create_view(navigator.get_class(volume, 'Details').VIEW)
+    view.flash.assert_success_message(
+        'Attaching Cloud Volume "{name}" to {instance} finished'.format(
+            name=volume.name, instance=new_instance.name))
+
+    @wait_for_decorator(delay=10, timeout=300,
+                        message="Waiting for volume to be attached to instance")
+    def volume_attached_to_instance():
+        new_instance.refresh_relationships()
+        return volume.instance_count > initial_instance_count
```
jawatts: @jadhaj Have you thought about ```return volume.instance_count == initial_instance_count + 1``` here? This test should be the only one adding this specific volume to an instance. This would make it a more specific check.

cfme/storage/volume.py
jadhaj: No Reset button at detach View

cfme/tests/openstack/cloud/test_volumes.py
jadhaj: I think also it is out of scope for this PR. Will add a TODO

cfme/tests/openstack/cloud/test_volumes.py
jadhaj: Actually i thought about  return volume.instance_count ==1 and  return volume.instance_count ==0 without using initial_instance_count var

cfme/storage/volume.py
```diff
@@ -243,6 +287,31 @@ def create_snapshot(self, name, cancel=False, reset=False):
...
                 return snapshot_collection.instantiate(name, self.provider)
 
+    def attach_instance(self, name, mountpoint=None, cancel=False, reset=False):
+        view = navigate_to(self, 'AttachInstance')
+        changed = view.fill({'instance': name, 'mountpoint': mountpoint})
+
+        if changed:
+            if reset:
+                view.reset.click()
+            elif cancel:
```
jawatts: **Required** The cancel button is always active and not dependent on the view changing. 

Also, I would add a comment saying that reset and attach are only active when the view is changed

cfme/storage/volume.py
```diff
@@ -243,6 +287,31 @@ def create_snapshot(self, name, cancel=False, reset=False):
...
+        view.flash.assert_no_error()
+
+    def detach_instance(self, name, cancel=False):
+        view = navigate_to(self, 'DetachInstance')
+        view.instance.fill(name)
+
+        if cancel:
+            view.cancel.click()
+        else:
+            view.detach.click()
```
jawatts: Similiar comment to above, is detach only active if the view was changed? If so, this needs to be captured. 

```if changed:
        view.detach.click()```

File: cfme/tests/openstack/cloud/test_volumes.py [hunk 18]
```diff
@@ -110,3 +120,32 @@ def test_edit_volume_with_type(volume_with_type, appliance):
 def test_delete_volume_with_type(volume_with_type):
     volume_with_type.delete()
     assert not volume_with_type.exists
+
+
+@pytest.mark.regression
+def test_volume_attach_detach_instance(volume, new_instance, appliance):
+    # TODO: Test Reset and Cancel
```
jawatts: :+1:

cfme/tests/openstack/cloud/test_volumes.py
jawatts: @jadhaj I think it be better to be specific here, so I going to mark this as a **Required** change.

cfme/storage/volume.py
```diff
@@ -243,6 +287,36 @@ def create_snapshot(self, name, cancel=False, reset=False):
                 view.save.click()
                 return snapshot_collection.instantiate(name, self.provider)
 
+    def attach_instance(self, name, mountpoint=None, cancel=False, reset=False):
+        view = navigate_to(self, 'AttachInstance')
+
+        # Reset and Attach buttons are only active when view is changed
+        changed = view.fill({'instance': name, 'mountpoint': mountpoint})
+
+        if cancel:
```
jawatts: I think this is close, but we are missing a condition. What is the form was not changed? What do you think about updating the cancel statement to:

```
if cancel or not was_changed:
    if not was_changed:
        logger.info("attach_instance: the form was unchanged")
    self.click.cancel()
```

cfme/storage/volume.py
```diff
@@ -243,6 +287,36 @@ def create_snapshot(self, name, cancel=False, reset=False):
...
+        view.flash.assert_no_error()
+
+    def detach_instance(self, name, cancel=False):
+        view = navigate_to(self, 'DetachInstance')
+
+        # Detach button is only active when view is changed
+        changed = view.instance.fill(name)
+
+        if cancel:
+            view.cancel.click()
```
jawatts: same comment above

cfme/storage/volume.py
```diff
@@ -243,6 +287,36 @@ def create_snapshot(self, name, cancel=False, reset=False):
...
+
+        elif changed:
+            if reset:
+                view.reset.click()
+            else:
+                view.attach.click()
+
+        view.flash.assert_no_error()
```
jawatts: @jadhaj Was there a reason this did not create a view? I believe this would be beneficial, then you would not have to create the view everytime in your tests.

cfme/storage/volume.py
jadhaj: Right, I agree

cfme/storage/volume.py
jadhaj: @jawatts Sorry but i am not sure i understand what is needed here

jadhaj: @jawatts I guess you mean to create a View and return it at end of method ?
The reason is that  I just did not see examples on the code for such behavior 
Usually we return instances not Views

jawatts: @jadhaj you are correct! It would be something like 

```view = self.create_view(VolumeDetailsView)```

That way when you are assert no errors it will be agaisnt the` VolumeDetailsView`

File: cfme/storage/volume.py [hunk 22]
```diff
@@ -243,6 +287,42 @@ def create_snapshot(self, name, cancel=False, reset=False):
...
+
+        if cancel or not changed:
+            if not changed:
+                logger.info("detach_instance: the form was unchanged")
+            self.click.cancel()
+        elif changed:
+            view.detach.click()
+
+        view = self.create_view(VolumeDetailsView)
+        view.flash.assert_no_error()
```
jawatts: same comment as above about moving under the `elif changed` block

PR #8199, opened 2018-11-26
[1LP][WIP] Fix test retire service ssui

niyazRedhat:
{{pytest: cfme/tests/ssui/test_ssui_myservice.py::test_retire_service_ssui --use-provider vsphere67-nested -svvv }}

### DIFF
--- a/cfme/services/myservice/ssui.py
+++ b/cfme/services/myservice/ssui.py
@@ -11,6 +11,7 @@ from cfme.utils.appliance.implementations.ssui import (
     navigator, SSUINavigateStep, navigate_to, ViaSSUI
 )
 from cfme.utils.wait import wait_for
+from cfme.utils.version import VersionPicker, Version
 from widgetastic_manageiq import (SSUIlist, SSUIDropdown, Notification, SSUIAppendToBodyDropdown)
 from . import MyService
 
@@ -230,14 +231,9 @@ def launch_vm_console(self, catalog_item):
 @MiqImplementationContext.external_for(MyService.retire, ViaSSUI)
 def retire(self):
     view = navigate_to(self, 'Retire')
-    view.retire.click()
-    view = self.create_view(MyServicesView)
-    assert wait_for(
-        lambda: view.is_displayed, delay=3, num_sec=300,
-        message="waiting for view to be displayed"
-    )
-    # TODO - remove sleep when BZ 1518954 is fixed
-    time.sleep(10)
+    if self.appliance.version < "5.10":
+        view.retire.click()
+    assert view.wait_displayed("10s")
     assert view.notification.assert_message("{} was retired.".format(self.name))
 
 
@@ -329,12 +325,13 @@ class MyServiceEditTags(SSUINavigateStep):
 
 @navigator.register(MyService, 'Retire')
 class MyServiceRetire(SSUINavigateStep):
-    VIEW = RetireServiceView
+    # TODO - remove check when BZ 1653288 is fixed
+    VIEW = VersionPicker({
+        Version.lowest(): RetireServiceView,
+        '5.10': MyServicesView
+    })
 
     prerequisite = NavigateToSibling('Details')
 
     def step(self):
-        if self.appliance.version >= "5.8":
-            self.prerequisite_view.lifecycle.item_select('Retire')
-        else:
-            self.prerequisite_view.lifecycle.item_select('Retire Now')
+        self.prerequisite_view.lifecycle.item_select('Retire')


cfme/services/myservice/ssui.py
```diff
@@ -329,12 +330,15 @@ def step(self):
 
 @navigator.register(MyService, 'Retire')
 class MyServiceRetire(SSUINavigateStep):
-    VIEW = RetireServiceView
+    # TODO - remove check when BZ 1653288 is fixed
+    @property
+    def VIEW(self): #noqa
```
digitronik: :+1:

cfme/services/myservice/ssui.py
```diff
@@ -230,14 +230,15 @@ def launch_vm_console(self, catalog_item):
...
-    view = self.create_view(MyServicesView)
-    assert wait_for(
-        lambda: view.is_displayed, delay=3, num_sec=300,
-        message="waiting for view to be displayed"
-    )
-    # TODO - remove sleep when BZ 1518954 is fixed
-    time.sleep(10)
+    if self.appliance.version <= "5.9":
+        view.retire.click()
+        view = self.create_view(MyServicesView)
```
digitronik: Required:
I think you won't need it; as your navigation `(L232)` returning you same view.

cfme/services/myservice/ssui.py
```diff
@@ -230,14 +230,15 @@ def launch_vm_console(self, catalog_item):
...
     view = navigate_to(self, 'Retire')
-    view.retire.click()
-    view = self.create_view(MyServicesView)
-    assert wait_for(
-        lambda: view.is_displayed, delay=3, num_sec=300,
-        message="waiting for view to be displayed"
-    )
-    # TODO - remove sleep when BZ 1518954 is fixed
-    time.sleep(10)
+    if self.appliance.version <= "5.9":
```
sshveta: How about for 5.10 ? Services cannot be retired in 5.10 ?

cfme/services/myservice/ssui.py
```diff
@@ -230,14 +230,15 @@ def launch_vm_console(self, catalog_item):
...
+        view.retire.click()
+        view = self.create_view(MyServicesView)
+        assert wait_for(
```
sshveta: You can do 
assert view.wait_displayed() 
instead now  and pass the delay .

cfme/services/myservice/ssui.py
niyazRedhat: Services are getting retired in 5.10 but confirmation pop-up window is not appearing .
so I raised the bug https://bugzilla.redhat.com/show_bug.cgi?id=1653288

sshveta: Then we should mark the test for 5.10 with this BZ

niyazRedhat: We can still retire this vm for now, but we want this stretched use case to be considered for safer side (we don't want customer to click retire by mistake so we have raised issue). However, we have added #to-do there and we will add pop-up thing as well as this thing get fixed in UI, so final thought, it should be not blocked.

sshveta: Okay I thought its being ignored for 5.10 as of now . 
If it's being tested for 5.9 and 5.10 both with a workaround then that is good :+1:

cfme/services/myservice/ssui.py
```diff
@@ -329,12 +324,15 @@ def step(self):
 
 @navigator.register(MyService, 'Retire')
 class MyServiceRetire(SSUINavigateStep):
-    VIEW = RetireServiceView
+    # TODO - remove check when BZ 1653288 is fixed
+    @property
+    def VIEW(self): #noqa
```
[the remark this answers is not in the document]
niyazRedhat: @izapolsk in 5.10 confirmation pop-up window for retiring the service is not appearing so service is retiring without confirmation.
In 5.9 confirmation pop-up window appear after clicking on retire.
I added check. 
``` view.retire.click()``` is part of confirmation popup and for 5.10 its not needed as i said confirmation pop-up window is missing.

```    if self.appliance.version < "5.10":
              view.retire.click()
````

File: cfme/services/myservice/ssui.py [hunk 6]
```diff
@@ -230,14 +230,9 @@ def launch_vm_console(self, catalog_item):
...
     view = navigate_to(self, 'Retire')
-    view.retire.click()
-    view = self.create_view(MyServicesView)
-    assert wait_for(
-        lambda: view.is_displayed, delay=3, num_sec=300,
-        message="waiting for view to be displayed"
-    )
-    # TODO - remove sleep when BZ 1518954 is fixed
-    time.sleep(10)
+    if self.appliance.version < "5.10":
```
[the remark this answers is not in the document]
niyazRedhat: @izapolsk in 5.10 confirmation pop-up window for retiring the service is not appearing so service is retiring without confirmation.
In 5.9 confirmation pop-up window appear after clicking on retire.
I added check.
```view.retire.click()``` is part of confirmation popup and for 5.10 its not needed as i said confirmation pop-up window is missing.

File: cfme/services/myservice/ssui.py (hunk 5 above)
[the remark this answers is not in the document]
Comment (niyazRedhat) [reply cfme/services/myservice/ssui.py]:
```
self = VersionPick({Version('lowest'): <class 'cfme.services.myservice.ssui.RetireServiceView'>, '5.10': <class 'cfme.services.myservice.ssui.MyServicesView'>})
o = <cfme.services.myservice.ssui.MyServiceRetire object at 0x7fdd3d933310>, type = <class 'cfme.services.myservice.ssui.MyServiceRetire'>

    def __get__(self, o, type=None):
        if o is None:
            # On a class, therefore not resolving
            return self
    
>       result = self.pick(o.browser.product_version)
E       AttributeError: 'MyServiceRetire' object has no attribute 'browser'

../../../../cfme_venv/lib/python2.7/site-packages/widgetastic/utils.py:312: AttributeError
```

I think it should be VersionPicker not VersionPick. since here in mycase we don't have `browser` to navigation class `MyServiceRetire`, @izapolsk  please confirm.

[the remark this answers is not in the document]
Comment (niyazRedhat) [reply cfme/services/myservice/ssui.py]: @izapolsk please review I have added VersionPicker and its Locally passing

File: cfme/services/myservice/ssui.py (hunk 6 above)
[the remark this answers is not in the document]
Comment (mshriver) [reply cfme/services/myservice/ssui.py]: **Required**

@niyazRedhat you've had to describe this to both @sshveta and @izapolsk in this PR - this information needs to be in-line here as a comment block, listing the BZ's involved.  Otherwise other contributors aren't aware of what's happening here, or if/when this version check should be removed.

File: cfme/services/myservice/ssui.py [hunk 7]
```diff
@@ -329,12 +325,13 @@ def step(self):
 
 @navigator.register(MyService, 'Retire')
 class MyServiceRetire(SSUINavigateStep):
-    VIEW = RetireServiceView
+    # TODO - remove check when BZ 1653288 is fixed
```
mshriver: This should be done dynamically based on `BZ(1653288).blocks`, otherwise tests will pass now and fail when the BZ is fixed because they'll have the wrong view again.

Ultimately I don't think this should really be changed in the framework to account for the BZ.  The framework should reflect, as much as possible, the application's intended design.  I believe we should avoid coding workarounds into our modeling, especially statically for BZ's.  

I think these changes are confusing a few different things:

1. Is the page displayed on navigation via selecting `Lifecycle->Retire` really version dependent, or is it BZ dependent?
2a. If the view is version dependent, then this should stay
2b. If the view is BZ dependent, this should not be changed, it should be the view that _should_ be displayed
3. Handling of the wrong view displayed due to BZ should be in `retire` method, not navigation
4. Tests could be skipped instead based on BZ, instead of writing workaround ( 3 ) into the retire method

niyazRedhat:
The BZ 1653288 is going to fixed in next build. closing this for now.

PR #8201, opened 2018-11-26
[RFR][NOTEST] Blocking test_service_start with GH8200

jan-zmeskal:
Blocker


cfme/tests/ssui/test_ssui_myservice.py
```diff
@@ -70,6 +70,7 @@ def _finalize():
 
 
 @pytest.mark.rhv3
+@pytest.mark.meta(blockers=[GH('ManageIQ/integration_tests:8200')])
```
izapolsk: @jan-zmeskal, please explain why this is bloker. 
As far as I understand you can always move necessary registration statement to one of always imported files.
Besides that you can always add an import statement here as a workaround.

PR #8204, opened 2018-11-27
[1LP][RFR] visibility and enablement test for container

digitronik:
Purpose or Intent
=================
{{pytest: cfme/tests/automate/custom_button/test_container_objects.py::test_custom_button_expression -vv --use-provider ocp-39-hawk}}

### DIFF
--- a/cfme/tests/automate/custom_button/test_container_objects.py
+++ b/cfme/tests/automate/custom_button/test_container_objects.py
@@ -172,3 +172,68 @@ def test_custom_button_dialog(appliance, dialog, request, setup_obj, button_grou
         )
     except TimedOutError:
         assert False, "Expected 1 requests not found in automation log"
+
+
+@pytest.mark.uncollectif(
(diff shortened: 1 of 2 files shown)


cfme/tests/automate/custom_button/test_container_objects.py
```diff
@@ -172,3 +172,68 @@ def test_custom_button_dialog(appliance, dialog, request, setup_obj, button_grou
...
+    )
+    request.addfinalizer(button.delete_if_exists)
+
+    tag_cat = appliance.collections.categories.instantiate(
+        name="department", display_name="Department"
+    )
+    tag = tag_cat.collections.tags.instantiate(name="engineering", display_name="Engineering")
+
+    view = navigate_to(setup_obj, "Details")
+    custom_button_group = Dropdown(view, group.hover)
```
mshriver: I think a better pattern for using these would be to create a parametrized view that we can pass the group.hover to, so that the widget is better tied to the view.

PR #8169, opened 2018-11-20
[1LP][RFR] Test network router details

aljazkosir:
This test creates sandbox with Nuage cloud tenant and its entities. We then navigate to Cloud tenant's network routers through provider, where we compare UI stats with sandbox stats.

This PR depends on: https://github.com/ManageIQ/integration_tests/pull/8150

Please review 3rd commit as the first and the second one are related to other PRs.

\cc @miha-plesko

PR #8205, opened 2018-11-27
[1LP][RFR] Test network port details

aljazkosir:
This test creates sandbox with Nuage cloud tenant and its entities. We then navigate to Network port through Nuage network provider, where we compare UI stats of ports with sandbox stats.
We test two different navigations, one navigates through L3 subnet, other one through L2 subnet.

This PR depends on: https://github.com/ManageIQ/integration_tests/pull/8169

Please review 4th commit as the first, second and third one are related to other PRs.

\cc @miha-plesko

### DIFF
--- a/cfme/networks/__init__.py
+++ b/cfme/networks/__init__.py
@@ -1,5 +1,6 @@
 from cfme.modeling.base import BaseCollection
-from cfme.utils.appliance.implementations.ui import navigate_to, navigator
+from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.appliance.implementations.ui import navigator
 
 
 class ValidateStatsMixin(object):
--- a/cfme/networks/__init__.py
+++ b/cfme/networks/__init__.py
@@ -1,4 +1,5 @@
-from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.modeling.base import BaseCollection
+from cfme.utils.appliance.implementations.ui import navigate_to, navigator
 
 
 class ValidateStatsMixin(object):
@@ -6,7 +7,11 @@ class ValidateStatsMixin(object):
     # network module, maybe BaseEntity
 
     def ui_details_value(self, path, property_name):
-        view = navigate_to(self, "Details", use_resetter=False)
+        is_filtered = isinstance(self.parent, BaseCollection) and self.parent.filters
+        if is_filtered and 'DetailsFromParent' in navigator.list_destinations(self):
+            view = navigate_to(self, 'DetailsFromParent', use_resetter=False)
+        else:
+            view = navigate_to(self, 'Details', use_resetter=False)
         return getattr(view.entities, path).get_text_of(property_name)
 
     def validate_stats(self, expected_stats):
--- a/cfme/networks/network_port.py
+++ b/cfme/networks/network_port.py
@@ -3,11 +3,14 @@ from navmazing import NavigateToAttribute
 
 from cfme.common import Taggable
 from cfme.exceptions import ItemNotFound
+from cfme.exceptions import DestinationNotFound
 from cfme.modeling.base import BaseCollection
 from cfme.modeling.base import BaseEntity
 from cfme.modeling.base import parent_of_type
 from cfme.networks.views import NetworkPortDetailsView
 from cfme.networks.views import NetworkPortView
+from cfme.networks import ValidateStatsMixin
+
 from cfme.utils import version
 from cfme.utils.appliance.implementations.ui import CFMENavigateStep
 from cfme.utils.appliance.implementations.ui import navigate_to
@@ -15,7 +18,7 @@ from cfme.utils.appliance.implementations.ui import navigator
 
 
 @attr.s
-class NetworkPort(Taggable, BaseEntity):
+class NetworkPort(Taggable, BaseEntity, ValidateStatsMixin):
     """Class representing network ports in sdn"""
     in_version = ('5.8', version.LATEST)
     category = "networks"
@@ -101,8 +104,19 @@ class All(CFMENavigateStep):
 
 @navigator.register(NetworkPort, 'Details')
 class Details(CFMENavigateStep):
-    prerequisite = NavigateToAttribute('parent', 'All')
     VIEW = NetworkPortDetailsView
+    prerequisite = NavigateToAttribute('parent', 'All')
 
     def step(self, *args, **kwargs):
         self.prerequisite_view.entities.get_entity(name=self.obj.name, surf_pages=True).click()
+
+
+@navigator.register(NetworkPort, 'DetailsFromParent')
+class DetailsFromParent(Details):
+    def prerequisite(self, *args, **kwargs):
+        is_filtered = isinstance(self.obj.parent, BaseCollection) and self.obj.parent.filters
+        filtered = (self.obj.parent.filters.get('parent') if is_filtered else None)
+        if is_filtered:
+            return navigate_to(filtered, 'NetworkPorts')
+        else:
+            raise DestinationNotFound("Parent filter missing on object {obj}".format(obj=self.obj))
--- a/cfme/networks/network_port.py
+++ b/cfme/networks/network_port.py
@@ -2,15 +2,14 @@ import attr
 from navmazing import NavigateToAttribute
 
 from cfme.common import Taggable
-from cfme.exceptions import ItemNotFound
 from cfme.exceptions import DestinationNotFound
+from cfme.exceptions import ItemNotFound
 from cfme.modeling.base import BaseCollection
 from cfme.modeling.base import BaseEntity
 from cfme.modeling.base import parent_of_type
+from cfme.networks import ValidateStatsMixin
 from cfme.networks.views import NetworkPortDetailsView
 from cfme.networks.views import NetworkPortView
-from cfme.networks import ValidateStatsMixin
-
 from cfme.utils import version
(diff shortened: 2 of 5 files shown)


aljazkosir:
@mshriver Sorry for the delay. This PR is now rebased with master. Once https://github.com/ManageIQ/integration_tests/pull/8169 gets merged we can continue with this.

aljazkosir:
@mshriver All dependent commits got merged. This one is now ready to be reviewed :)

cfme/tests/networks/nuage/test_inventory.py
```diff
@@ -82,3 +82,101 @@ def test_network_router_details(setup_provider_modscope, provider, with_nuage_sa
...
+    tenant = provider.collections.cloud_tenants.instantiate(name=tenant_name, provider=provider)
+    subnet = tenant.collections.subnets.instantiate(name=subnet_name)
+    network_port = subnet.collections.network_ports.instantiate(name=vport_name)
+
+    network_port.validate_stats({
+        ('properties', 'Name'): vport_name,
+        ('properties', 'Type'): 'ManageIQ/Providers/Nuage/Network Manager/Network Port/Container',
+        ('relationships', 'Cloud tenant'): tenant_name,
+        ('relationships', 'Cloud subnets'): '1',
+    })
```
miha-plesko: @aljazkosir I'm thinking we could extract common into a function as @digitronik suggested:

```python
def test_network_port_l2_container(...)
  assert_network_port(...)

def test_network_port_l2_vm(...)
  assert_network_port(...)

def assert_network_port(subnet_name, vport_name, expected_stats):
  """Helper function"""
  ...
```

Think that would make sense?

cfme/networks/network_port.py
```diff
@@ -96,8 +97,16 @@ def resetter(self):
...
 @navigator.register(NetworkPort, 'Details')
 class Details(CFMENavigateStep):
-    prerequisite = NavigateToAttribute('parent', 'All')
     VIEW = NetworkPortDetailsView
 
+    def prerequisite(self, *args, **kwargs):
+        """Navigate through filter parent if it exists else navigate through parent object"""
+        is_filtered = isinstance(self.obj.parent, BaseCollection) and self.obj.parent.filters
+        filter = (self.obj.parent.filters.get('parent') if is_filtered else None)
+        if is_filtered:
```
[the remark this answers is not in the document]
aljazkosir: In case parent is being filtered we would navigate through that filter.

cfme/networks/views.py
```diff
@@ -340,6 +340,26 @@ def is_displayed(self):
                 self.entities.title.text == 'Network Ports')
 
 
+class NetworkEntityWithNetworkPortView(BaseLoggedInPage):
```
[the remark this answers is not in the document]
aljazkosir: I named like that so any other Network entity with ports can use it.

aljazkosir:
I included some changes you suggested, as for the parametrization I think it's easier to read the way it is. Because L2 ports have different path than L3 ports, we would need two helper functions or some sort of if clause, but this will only make it harder to read.

\cc @digitronik

cfme/tests/networks/nuage/test_inventory.py
digitronik: @aljazkosir 

LGTM; please can you try some parametrize like
```
@pytest.mark.parametrize("port", ["vm_vport", "cont_vport"], ids=["vm", "container"])
@pytest.mark.parametrize("domain", ["default", "l2_domain"])
def test_network_port(port, domain):
    # some manuculation related to port and domain as per test
    print(port, domain)  # print conbination
```
```
test.py::test_network_port[default-vm] ('vm_vport', 'default')
PASSED
test.py::test_network_port[default-container] ('cont_vport', 'default')
PASSED
test.py::test_network_port[l2_domain-vm] ('vm_vport', 'l2_domain')
PASSED
test.py::test_network_port[l2_domain-container] ('cont_vport', 'l2_domain')
PASSED
```
I thing good to  parametrize. Please let me know if you face problem.

dajoRH:
*CFME QE Bot*

cfme/tests/networks/nuage/test_inventory.py
aljazkosir: This won't work since "default" and "l2_domain" ports follow different paths, "l2_domain" doesn't go through router while L3 does. I think it is clearer the way it is.

\cc @miha-plesko @mshriver

miha-plesko: @digitronik I had a discussion with @aljazkosir and we kind of agreed that in tests it's better to be a more explicit rather than overdo DRY. I understand we're trying to reuse as much as possible, but in this case I'd say the way @aljazkosir did it is just the best...

dajoRH:
Lint report for commit 056e7f0d0ae1a6b2296fcb7a911bf689a2963f63:

`cfme/networks/subnet.py`:
- :heavy_exclamation_mark: Line 231:69: **W292** *no newline at end of file*

**No commit flaws detected**.

*CFME QE Bot*

cfme/networks/views.py
```diff
@@ -374,6 +374,23 @@ class my_filters(Accordion):  # noqa
...
+    def is_displayed(self):
+        obj = self.context['object']
+        from cfme.networks.subnet import Subnet
+        matched_title = ('{name} (All Network Ports)'.format(
+            name=self.context['object'].name) == self.entities.title.text
+        )
+        if isinstance(obj, Subnet):
+            matched_navigation = self.navigation.currently_selected == ['Networks', 'Subnets']
+        else:
+            matched_navigation = False
```
digitronik: As we are restricting by checking navigation tree pointing to `Subnets`. I think better to use `SubnetNetworkPortView` and remove this check of `Subnet` class. This view will not going to applicable to other `NetworkEntitys`.

aljazkosir: Sure, we meant to create a view which be applicable for other network entities. But sure, since is currently not used anywhere else.

cfme/networks/network_port.py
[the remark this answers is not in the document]
mshriver: This is a pattern which exists in other network classes, but I'm beginning to think that its not a good one to continue repeating.  

@miha-plesko @izapolsk @jawatts @digitronik @psav Please weigh in if you can.

I don't think that this is a level at which we should apply filtering to navigation.

We are codifying the steps necessary in the UI to get to the details page of a NetworkPort, which we reasonably expect a 'default' navigation path of navigating to the collection's page ('All'), then select the entity.

What is represented here is an additional navigation path through the UI - which is determined by the filtering on the entity object's collection. I think these are better represented as separate navigation destinations for the entity - navigation through a parent object.

My concern is that navigation to this destination is not very explicit when used in code. I create a filtered (or not) collection, instantiate or create an entity in that collection, and then navigate to its details page.  Whether or not I filtered ends up changing how I have navigated to the page. 

If we were to have separate destinations, then my navigation would be explicit - with the same process of collection + entity + navigation, I can now navigate via two different paths, regardless of the filter I have applied. 

I'm thinking we should represent these navigation paths with destinations like:
1. `Details` - default navigation from collection, regardless of filtering
2. `DetailsFromParent` - navigation that looks for a parent filter, navigates from parent destination

Some concerns I have with the change are the code duplication in the step method, the `VIEW` class used, but that is minimal for most navigation (single line statements)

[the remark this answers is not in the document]
miha-plesko: @mshriver your suggestion sounds reasonable. @aljazkosir do you see any issue here?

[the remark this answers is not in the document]
aljazkosir: Fine with me, I can include changes for this (NetworkPort) entity in this PR and create a new PR for other entities that uses the same pattern. Regarding the code duplication, we can always just inherit the the class that has the same `step` and `VIEW` and only change `prerequisite` attribute? In this case `Details` class will have both `step` and `VIEW` defined and `DetailsFromParent` can just inherit `Details` class. What do you think @mshriver ?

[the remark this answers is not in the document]
aljazkosir: I pushed changes that include this.

cfme/networks/__init__.py
```diff
@@ -1,12 +1,17 @@
...
 
 
 class ValidateStatsMixin(object):
     # TODO: Move this class to a higher level, where could be useful for thing beyond this
     # network module, maybe BaseEntity
 
     def ui_details_value(self, path, property_name):
-        view = navigate_to(self, "Details", use_resetter=False)
+        is_filtered = isinstance(self.parent, BaseCollection) and self.parent.filters
+        if is_filtered and 'DetailsFromParent' in navigator.list_destinations(self):
```
mshriver: :+1:

dajoRH:
*CFME QE Bot*

cfme/networks/views.py
```diff
@@ -374,6 +374,18 @@ class my_filters(Accordion):  # noqa
         tree = ManageIQTree()
 
 
+class SubnetNetworkPortView(NetworkPortView):
+    """ Represents a Subnet network with port view"""
+    title = Text('//div[@id="main-content"]//h1')
+
+    @property
+    def is_displayed(self):
+        return (super(BaseLoggedInPage, self).is_displayed and
```
mshriver: I think a better call here is to not use `super` and directly call `self.logged_in_as_current_user`, as this will account for the context (UI, SSUI) and the active appliance user.

mshriver: Pushing a commit with this change, and a rebase due to the import order sorting PR #8487 being merged.

PR #8208, opened 2018-11-27
[1LP][RFR] Fix dashboard tests

jaryn:
Purpose or Intent
=================

__Fixing__ test_menuwidget_crud.

There were some typo-like mishaps.

{{ py.test: cfme/tests/intelligence/reports/test_crud.py }}

### DIFF
--- a/cfme/intelligence/reports/dashboards.py
+++ b/cfme/intelligence/reports/dashboards.py
@@ -12,6 +12,7 @@ from cfme.utils.appliance import Navigatable
 from cfme.utils.appliance.implementations.ui import navigator, CFMENavigateStep, navigate_to
 from cfme.utils.pretty import Pretty
 from cfme.utils.update import Updateable
+from cfme.utils.wait import wait_for
 from . import CloudIntelReportsView
 
 
@@ -55,7 +56,7 @@ class NewDashboardView(DashboardFormCommon):
             self.dashboards.tree.currently_selected == [
                 "All Dashboards",
                 "All Groups",
-                self.context["object"].group
+                self.context["object"]._group
             ]
         )
 
@@ -68,7 +69,7 @@ class EditDashboardView(DashboardFormCommon):
     def is_displayed(self):
         return (
             self.in_intel_reports and
-            self.title.text == "Editing Dashboard {}".format(self.context["object"].name) and
+            self.title.text == 'Editing Dashboard "{}"'.format(self.context["object"].name) and
             self.dashboards.is_opened and
             self.dashboards.tree.currently_selected == [
                 "All Dashboards",
@@ -172,7 +173,7 @@ class Dashboard(BaseEntity, Updateable, Pretty):
         else:
             view.cancel_button.click()
         view = self.create_view(DashboardDetailsView, override=updates)
-        assert view.is_displayed
+        wait_for(lambda: view.is_displayed, fail_condition=False, timeout=10)
         view.flash.assert_no_error()
         if self.appliance.version < "5.9":
             success_msg = 'Dashboard "{}" was saved'.format(self.name)
--- a/cfme/intelligence/reports/dashboards.py
+++ b/cfme/intelligence/reports/dashboards.py
@@ -12,7 +12,6 @@ from cfme.utils.appliance import Navigatable
 from cfme.utils.appliance.implementations.ui import navigator, CFMENavigateStep, navigate_to
 from cfme.utils.pretty import Pretty
 from cfme.utils.update import Updateable
-from cfme.utils.wait import wait_for
 from . import CloudIntelReportsView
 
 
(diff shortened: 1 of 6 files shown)


cfme/intelligence/reports/widgets/__init__.py
```diff
@@ -123,7 +123,7 @@ def create(self, widget_class, *args, **kwargs):
         view = navigate_to(dashboard_widget, "Add")
         view.fill_with(dashboard_widget.fill_dict, on_change=view.add_button.click)
         view = dashboard_widget.create_view(AllDashboardWidgetsView)
-        assert view.is_displayed
+        wait_for(lambda: view.is_displayed, fail_condition=False, timeout=10)
```
valaparthvi: You might want to use `view.wait_displayed()` instead of `wait_for`.

cfme/intelligence/reports/dashboards.py
```diff
@@ -172,7 +173,7 @@ def update(self, updates):
         else:
             view.cancel_button.click()
         view = self.create_view(DashboardDetailsView, override=updates)
-        assert view.is_displayed
+        wait_for(lambda: view.is_displayed, fail_condition=False, timeout=10)
```
valaparthvi: `view.wait_displayed()` instead of `wait_for`

valaparthvi:
Additionally, you might want to use `use_resetter=False` on [L#166](https://github.com/ManageIQ/integration_tests/pull/8208/files#diff-1135fda3b54df4d13b1f88347ec62ef0L166) and [L#61](https://github.com/ManageIQ/integration_tests/pull/8208/files#diff-29e866572668278cd1b520033a72c2a5L61), since recent changes reset the Reports page every time it's navigated to.

cfme/tests/intelligence/reports/test_crud.py
```diff
@@ -118,6 +118,7 @@ def test_reports_disable_enable_schedule(schedule_data, appliance):
 
 @pytest.mark.sauce
 @pytest.mark.tier(3)
+@pytest.mark.meta(blockers=[BZ(1653796, forced_streams=["5.9", "5.10"])])
```
valaparthvi: This BZ has already been verified, so you might want to remove this meta blocker.

cfme/intelligence/reports/widgets/__init__.py
jaryn: No. The assert was failing because some condition in is_displayed was not met, but when waited-for, it passes after some time.

jaryn:
> Additionally, you might want to use `use_resetter=False` on [L#166](https://github.com/ManageIQ/integration_tests/pull/8208/files#diff-1135fda3b54df4d13b1f88347ec62ef0L166) and [L#61](https://github.com/ManageIQ/integration_tests/pull/8208/files#diff-29e866572668278cd1b520033a72c2a5L61), since recent changes reset the Reports page every time it's navigated to, and avoiding resetter will be more effective.

I don't understand. Where should I use it?

cfme/tests/intelligence/reports/test_crud.py
jaryn: It is verified in 5.10, but not for the 5.9. It is still broken in 5.9

cfme/intelligence/reports/widgets/__init__.py
jaryn: You are probably right. I though you meant `view.is_displayed.

jaryn:
> > @jaryn, on L#166, we're navigating to the EditView, so you can use `use_resetter` like this:
> > `view = navigate_to(self, "Edit", use_resetter=False)`.
> > Same for L#61.
> 
> Looks like you've forgotten to add `use_resetter` on [L#61](https://github.com/ManageIQ/integration_tests/pull/8208/files#diff-29e866572668278cd1b520033a72c2a5L61).

I added it to L#61, but I don't know why is it better. Everything works the same.

valaparthvi:
Some of the `widget_crud` tests were failing before I used `use_resetter`, but they passed once I set it to `False`.

And since you're testing all the tests of `test_crud.py`, with the current changes, all the tests(except `test_dashboard_crud`) are passing.

dajoRH:
*CFME QE Bot*

cfme/tests/intelligence/test_dashboard.py
```diff
@@ -66,6 +66,7 @@ def test_widgets_operation(dashboards, widgets, soft_assert, infra_provider):
 
 @pytest.mark.rhel_testing
 @pytest.mark.parametrize("number_dashboards", range(1, 4))
+@pytest.mark.meta(blockers=[BZ(1666712, forced_streams=['5.10'])])
```
mshriver: I don't think forced_streams is necessary here - your fixes to the blocker resolution removes this need! :)

jaryn: Done

PR #8209, opened 2018-11-28
[RFR] fixed indent for test_tagvis_cloud_object and removed if for appliance.version < 5.8

lina-nikiforova:
{{ pytest: -v  cfme/tests/cloud_infra_common/test_tag_objects.py::test_tagvis_cloud_object --use-provider=complete }}

1. fixed indent for `test_tagvis_cloud_object` so it will be executed 
2. removed the line with `appliance.version < '5.9'` as 5.8 in the separate branch and is not supported in master

PRT will not pass until #8191 is merged


mshriver:
PRT was not complete, but showed passing results covering the changes here.

Issue #8200, opened 2018-11-26
Power dropdown no longer available for service in SSUI


jan-zmeskal:
Hello, I bumped into this while running [test_service_start](https://github.com/ManageIQ/integration_tests/blob/1160bd8e03db0c88e152317bad55faec3ee961e5/cfme/tests/ssui/test_ssui_myservice.py#L74). The test case successfully orders VM. Problem happens when it switches to SSUI, navigates to service's details and tries to stop the VM. It happens [here](https://github.com/ManageIQ/integration_tests/blob/1160bd8e03db0c88e152317bad55faec3ee961e5/cfme/tests/ssui/test_ssui_myservice.py#L87). 

According to definition of [DetailsMyServiceView](https://github.com/ManageIQ/integration_tests/blob/1160bd8e03db0c88e152317bad55faec3ee961e5/cfme/services/myservice/ssui.py#L40), power_operations is denined like this:
`power_operations = SSUIDropdown('Power Operations')`
However, the framework throws this error when trying to click it:

> E NoSuchElementException: Message: Could not find an element Locator(by='xpath', locator='.//span[contains(@class, "dropdown") and .//button[contains(@class, "dropdown-toggle")]/span[normalize-space(.)="Power Operations"]]')

Also, if you take a look at the current state of service detail view in SSUI, there is no dropdown with text "Power": 
![image](https://user-images.githubusercontent.com/22600243/49017852-2dbde280-f18a-11e8-9ca6-9327bedf1d7f.png)

Instead, there is this little kebab menu right of the Access dropdown. If you exanpd it, it actually contains power operations. DetailsMyServiceView needs to be updated to reflect this. I was able to achieve clicking the kebab me by replacing [this line](https://github.com/ManageIQ/integration_tests/blob/1160bd8e03db0c88e152317bad55faec3ee961e5/cfme/services/myservice/ssui.py#L247) with this:
`view.browser.click('.//div[contains(@class,"dropdown-kebab-pf")]/button')`
but this is definitely not the correct way. The main problem here is that the kebab element misses any unique identifier like text or id. Therefore I am putting it here, since you probable know better how to tackle these issues.

mshriver:
@jan-zmeskal thanks for the heads up and documenting this thoroughly.  @sshveta and @digitronik can you check this out for us?

digitronik:
@mshriver I assigned myself.
Added fix for it in `ssui` model page.

jan-zmeskal:
[...]

PR #7600, opened 2018-08-06
[1LP][RFR] Removed unneeded VersionPicks

quarckster:
Intent
=================
- `widgetastic.utils.VersionPick` replaced by `cfme.utils.VersionPicker`
- removed unneeded version picks

{{pytest: -v -k "test_automate_method_inputs_crud or test_host_good_creds or test_infra_provider_crud" --long-running}}

PR #8213, opened 2018-11-29
[1LP][RFR] Fix power operation

digitronik:
Purpose or Intent
=================
{{pytest: cfme/tests/ssui/test_ssui_myservice.py::test_service_start -v  --use-provider vsphere65-nested --long-running}}

### DIFF
--- a/cfme/services/myservice/ssui.py
+++ b/cfme/services/myservice/ssui.py
@@ -11,7 +11,7 @@ from cfme.utils.appliance import MiqImplementationContext
 from cfme.utils.appliance.implementations.ssui import (
     navigator, SSUINavigateStep, navigate_to, ViaSSUI
 )
-from cfme.utils.version import Version, VersionPick
+from cfme.utils.version import VersionPicker, LATEST
 from cfme.utils.wait import wait_for
 from widgetastic_manageiq import (SSUIlist, SSUIDropdown, Notification, SSUIAppendToBodyDropdown)
 from . import MyService
@@ -49,16 +49,16 @@ class DetailsMyServiceView(MyServicesView):
 
     notification = Notification()
     policy = SSUIDropdown('Policy')
-    power_operations = VersionPick(
-        {Version.lowest(): SSUIDropdown("Power Operations"), "5.10": Kebab()}
+    power_operations = VersionPicker(
+        {LATEST: SSUIDropdown("Power Operations"), "5.10": Kebab()}
     )
     access_dropdown = SSUIAppendToBodyDropdown('Access')
     remove_service = Button("Remove Service")
     configuration = SSUIDropdown('Configuration')
     lifecycle = SSUIDropdown('Lifecycle')
     console_button = Button(tooltip="HTML5 console", classes=['open-console-button'])
-    resource_power_status = PowerIcon("//i[@class='pficon fa fa-power-off' or "
-                                      "@class='pficon pficon-ok']")
+    resource_power_status = PowerIcon(".//span/i[contains(@class, 'pficon') and contains("
+                                      "@uib-tooltip,'Power State')]")
 
 
 class ServiceEditForm(MyServicesView):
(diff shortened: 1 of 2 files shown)


digitronik:
Related to https://github.com/ManageIQ/integration_tests/issues/8200

digitronik:
aahhh  I think,  we need to fix `power_status`  `locator` to run this test against changes.

digitronik:
@niyazRedhat can you please fetch this PR and check whether your test working or not?

cfme/services/myservice/ssui.py
```diff
@@ -6,10 +6,12 @@
 from widgetastic_patternfly import Input, Button
 
 from cfme.base.ssui import SSUIBaseLoggedInPage
+from cfme.dashboard import Kebab
 from cfme.utils.appliance import MiqImplementationContext
 from cfme.utils.appliance.implementations.ssui import (
     navigator, SSUINavigateStep, navigate_to, ViaSSUI
 )
+from cfme.utils.version import Version, VersionPick
```
sshveta: Please Use : 
from cfme.utils.version import VersionPicker, LOWEST

cfme/services/myservice/ssui.py
```diff
@@ -47,14 +49,16 @@ def is_displayed(self):
 
     notification = Notification()
     policy = SSUIDropdown('Policy')
-    power_operations = SSUIDropdown('Power Operations')
+    power_operations = VersionPick(
+        {Version.lowest(): SSUIDropdown("Power Operations"), "5.10": Kebab()}
```
sshveta: Use VersionPicker instead

cfme/services/myservice/ssui.py
```diff
@@ -244,14 +248,11 @@ def retire(self):
...
-        lambda: view.is_displayed, delay=3, num_sec=300,
-        message="waiting for view to be displayed"
-    )
-    # TODO - remove sleep when BZ 1518954 is fixed
-    time.sleep(10)
+    if self.appliance.version < "5.10":
+        view.power_operations.item_select(power)
+    else:
+        view.power_operations.select(power)
+    view.wait_displayed('60s')
```
sshveta: :+1:

cfme/services/myservice/ssui.py
digitronik: @sshveta can you please explain why?

sshveta: @digitronik : Actually there was a PR by Dmitry sometime back in which he wrapped 
cfme.utils.version.pick to  a method of VersionPicker and removed versionpick from most of the places . I have been using versionpicker since then .
https://github.com/ManageIQ/integration_tests/pull/7543 and 
https://github.com/ManageIQ/integration_tests/pull/7600

digitronik: But its completely `optional`  as its wrapped. In my condition; I am not using an extended feature of `ViersionPicker` so its valid as per `WT` guideline and current implementation; 

As per my observation and understanding with current `collection` implementation no need of `VersionPicker`.

Cases where  we need to take  action as per `version`:
1. In `views class`:
where we can not access `appliance` object at the initial stage (`l.e.` `locator`, `Widget` selection) then go with `VersionPick`.
2. In `views methods` or `properties`:
where possible to access `appliance` object directly free to go with a condition.
3. In `Navigation`:
If we need to return `view` as per `appliance version` then we can go with `VIEW property` means
```
@property
def VIEW(self):
    return view1 if self.appliance.version < 'x.y' else view2
```

I think it's better to make one guideline; where we should accept `VersionPick`, `VersionPicker` and `appliance.version` condition. adding `tech-debt` tag.

digitronik: @psav @quarckster @mshriver @izapolsk @RonnyPfannschmidt Please have a look... It will help to avoid confusion...

quarckster: @digitronik The intent of creating `VersionPicker` to unify different version picker implementations. I would insist to use `VersionPicker`.

izapolsk: I'd add that VersionPicker is prefered for widgetastic stuff because it can be correctly resolved by wt methods and etc

PR #8218, opened 2018-11-30
[1LP][RFR] Add fixture to enable v2v migration UI in downstream 510z

mshriver:
Migration was disabled, this fixture will add it back to permissions for downstream 5.10 if its missing.

Session scoped so anything using v2v_providers will get it enabled for the entire session duration.

Local testing successful with workaround already applied and without.

{{ pytest: --long-running --provider-limit=2 --use-provider rhv42 --use-provider vsphere67-nested cfme/tests/v2v/test_v2v_migrations_ui.py }}

### DIFF
--- a/cfme/fixtures/v2v.py
+++ b/cfme/fixtures/v2v.py
@@ -18,8 +18,24 @@ FormDataVmObj = namedtuple(
 )
 
 
+@pytest.fixture(scope='session')
+def v2v_downstream_enable(appliance):
+    """downstream 5.10 public beta has migration option disabled
+        Edit config and restart evm to enable it for testing
+        Not needed after GA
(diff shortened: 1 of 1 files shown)


mshriver:
@Yadnyawalkya I moved the evm restart into the if statement  - and this is session scoped, so it should only run at the start of the test session, not at every function.  Even if it does, after its applied the check for the value will make it skip the next time it runs.  I pushed the commit changing the evmserver restart indent right before you made your comment though, I think this is a non-issue.

jawatts:
Lets make sure this works first :)

mshriver:
Thanks for quick response and testing guys!! waiting for travis mostly.

PR #8221, opened 2018-12-03
[1LP][RFR] CB: service automate

digitronik:
Purpose or Intent
=================
- Added To-Do  for paginator on ssui views
- Added new test for sevice  automate

{{pytest:  cfme/tests/automate/custom_button/test_service_objects.py::test_custom_button_automate -vv}}

### DIFF
--- a/cfme/tests/automate/custom_button/test_service_objects.py
+++ b/cfme/tests/automate/custom_button/test_service_objects.py
@@ -4,10 +4,12 @@ import fauxfactory
 from widgetastic_patternfly import Dropdown
 
 from cfme.services.myservice import MyService
+from cfme.tests.automate.custom_button import log_request_check
 from cfme.utils.appliance import ViaREST, ViaUI, ViaSSUI
 from cfme.utils.appliance.implementations.ui import navigate_to as ui_nav
 from cfme.utils.appliance.implementations.ssui import navigate_to as ssui_nav
 from cfme.utils.blockers import BZ
+from cfme.utils.wait import TimedOutError, wait_for
 
 
 pytestmark = [pytest.mark.tier(2)]
@@ -19,6 +21,8 @@ DISPLAY_NAV = {
     "List": ["All"],
     "Single and list": ["All", "Details"],
 }
+
+SUBMIT = ["Submit all", "One by one"]
 
 
 @pytest.fixture(scope="module")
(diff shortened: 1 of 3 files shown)


cfme/tests/automate/custom_button/test_service_objects.py
```diff
@@ -153,3 +157,99 @@ def test_custom_button_display(request, appliance, context, display, objects, bu
...
+            else:
+                entity_count = 1
+
+            # Clear the automation log
+            assert appliance.ssh_client.run_command(
+                'echo -n "" > /var/www/miq/vmdb/log/automation.log'
+            )
+
+            custom_button_group.item_select(button.text)
+            if context != ViaSSUI:
```
sshveta: What happens in SSUI case . Is  there a way to assert in SSUI ?

digitronik: `log request` check is our main check. SSUI not showing msg. I don't think its important as well.

jawatts: I think this is worth putting in as a comment ^

digitronik: @jawatts no need of this... Its expected behavior of SSUI.

cfme/tests/automate/custom_button/test_service_objects.py
```diff
@@ -153,3 +157,99 @@ def test_custom_button_display(request, appliance, context, display, objects, bu
...
+        * Check for button group and button
+        * Select/execute button from group dropdown for selected entities
+        * Check for the proper flash message related to button execution
+        * Check automation log requests. Submitted as per selected submit option or not.
+        * Submit all: single request for all entities execution
+        * One by one: separate requests for all entities execution
+
+    Bugzillas:
+        * 1650066
+    """
```
mshriver: Please add polarion docblock with at least assignee and initialestimate fields.

cfme/tests/automate/custom_button/test_service_objects.py
mshriver: I think it does need to be in a comment in-line, as its only a _current_ expected behavior, and obviously not every contributor will be aware of this expectation. :)

**Required**

mshriver: I also think this would be better as a positive assertion of the ViaUI context, than negative assertion against ViaSSUI.

digitronik: :D added

PR #8225, opened 2018-12-03
[1LP][RFR] Fixed assertion error while creating  new group

ganeshhubale:
- This PR fixes the Assertion Error while creating new group with blank description. 

{{ pytest: cfme/tests/configure/test_access_control.py -k 'test_group_description_required_error_validation' -v }}

Fixed Error:
```
>           group_collection.create(description=None, role='EvmRole-approver')
E           AssertionError: Pattern 'Description can't be blank' not found in 'Could not do function <lambda>() at /var/ci/workspace/downstream-510z-rc/cfme_venv/lib/python2.7/site-packages/widgetastic/widget/base.py:493 in time'

cfme/tests/configure/test_access_control.py:471: AssertionError
AssertionError
Pattern 'Description can't be blank' not found in 'Could not do function <lambda>() at /var/ci/workspace/downstream-510z-rc/cfme_venv/lib/python2.7/site-packages/widgetastic/widget/base.py:493 in time'
```

### DIFF
--- a/cfme/configure/access_control/__init__.py
+++ b/cfme/configure/access_control/__init__.py
@@ -836,11 +836,7 @@ class GroupCollection(BaseCollection):
                 not having appropriate permissions OR delete is not allowed
                 for currently selected user
         """
-        if self.appliance.version < "5.8":
-            flash_blocked_msg = ("Description has already been taken")
-        else:
-            flash_blocked_msg = "Description is not unique within region {}".format(
-                self.appliance.server.zone.region.number)
+        flash_blocked_msg = "Description can't be blank"
 
         view = navigate_to(self, 'Add')
 
--- a/cfme/configure/access_control/__init__.py
+++ b/cfme/configure/access_control/__init__.py
@@ -836,7 +836,13 @@ class GroupCollection(BaseCollection):
                 not having appropriate permissions OR delete is not allowed
                 for currently selected user
         """
-        flash_blocked_msg = "Description can't be blank"
+        flash_blocked_msg = (
+            "Description is not unique within region {}".format(
+                self.appliance.server.zone.region.number
+            )
+            if description
+            else "Description can't be blank"
+        )
 
         view = navigate_to(self, 'Add')
 
--- a/cfme/configure/access_control/__init__.py
+++ b/cfme/configure/access_control/__init__.py
@@ -836,13 +836,13 @@ class GroupCollection(BaseCollection):
                 not having appropriate permissions OR delete is not allowed
                 for currently selected user
         """
-        flash_blocked_msg = (
-            "Description is not unique within region {}".format(
-                self.appliance.server.zone.region.number
+        if description:
+            flash_blocked_msg = (
+                "Description is not unique within region {}".format(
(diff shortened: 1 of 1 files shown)


cfme/configure/access_control/__init__.py
```diff
@@ -837,10 +837,9 @@ def create(self, description=None, role=None, tenant="My Company", ldap_credenti
                 for currently selected user
         """
         if self.appliance.version < "5.8":
```
digitronik: Required:
@ganeshhubale remove  this condition fro msg  as we are not supporting  version less than `5.9` against `master`

cfme/tests/infrastructure/test_vm_power_control.py
```diff
@@ -397,7 +397,12 @@ def test_no_power_controls_on_archived_vm(testing_vm, archived_vm, soft_assert):
         test_flag: power_control
     """
     view = navigate_to(testing_vm, 'AnyProviderDetails', use_resetter=False)
-    soft_assert(not view.toolbar.power.is_displayed, "Power displayed in archived VM's details!")
+    if appliance.version < 5.10:
```
digitronik: Optional:
better to have  single `assert`

cfme/configure/access_control/__init__.py
```diff
@@ -836,11 +836,7 @@ def create(self, description=None, role=None, tenant="My Company", ldap_credenti
                 not having appropriate permissions OR delete is not allowed
                 for currently selected user
         """
-        if self.appliance.version < "5.8":
-            flash_blocked_msg = ("Description has already been taken")
-        else:
-            flash_blocked_msg = "Description is not unique within region {}".format(
```
izapolsk: REQUIRED: This message look different from your message. 
Besides that, you update code inside framework. So, this code can potentially be used by other tests.
So, to recap I have a suspicion that your change will fix one test but break a bunch of other tests. 
Please make sure that other tests won't be broken and include other tests using this method into PRT.
Btw, please look once more at code where this flash message is used.

ganeshhubale: @izapolsk  
 Yes, I agree with you. This method is used in various tests. 
- My observations:
    - If we do not provide description while creating group then it is showing flash message as ```Description can't be blank```.
    - If we provide description which is similar to other existing group then it shows flash message as ```"Description is not unique within region {}".format(self.appliance.server.zone.region.number```   
- I modified this PR in a way where it will not break other test cases.

Please, let me know if you have any other suggestions.
Thank you.

cfme/configure/access_control/__init__.py
```diff
@@ -836,11 +836,13 @@ def create(self, description=None, role=None, tenant="My Company", ldap_credenti
...
-                self.appliance.server.zone.region.number)
+        flash_blocked_msg = (
+            "Description is not unique within region {}".format(
+                self.appliance.server.zone.region.number
+            )
+            if description
```
digitronik: It's valid but we should not go more than three line for inline `if else` block. It's difficult to read...

PR #8227, opened 2018-12-03
[WIP] Add a test for events in Redfish physical infrastructure provider

matejart:
Purpose or Intent
=================

__Adding tests__ for Redfish phsyical infrastructure provider to cover event collection functionality implemented in Hammer.

Events generated depend on the OEM who created the Redfish server, so the test does not assume that any particular Redfish action would generate any specific event. So we rely on the server to generate a stream of events on its own. When testing against the [Redfish mock-up server](https://github.com/xlab-si/redfish-recordings), this assumption is covered.

Related Bugzilla bug:
  * [#1574828](https://bugzilla.redhat.com/show_bug.cgi?id=1574828)


matejart:
`register_event` works 👍  thank you!

cfme/tests/physical_infrastructure/test_redfish_physical_infra_events.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def test_get_redfish_events_any(setup_provider_funcscope, register_event):
+    """
+    Test that the provider accounts for any Redfish-related event.
+
+    The test assumes that events are generated regularly without needing to
+    trigger any action.
+    """
+    register_event(source=SOURCE)
```
digitronik: As per my understanding, `register_event` yield `event_listener`. better to check that... what do you think?

jawatts: @digitronik you are correct here. But if you look at the fixture, there is a `pytest_runtest_call` method defined.

PR #8230, opened 2018-12-03
[WIPTEST]Mark test cases with blockers and optimize the test cases

valaparthvi:
1. Add one more asserting statement to all the test cases, which checks the value of `default items per page` in the paginator on a particular page.
2. Add one more value(5) to the `value` fixture.

{{pytest: cfme/tests/configure -k "page_per_item" --use-template-cache -sqvvvv}}


mshriver:
PRT looking great

PR #8234, opened 2018-12-04
[1LP][RFR] unifying flash message xpaths and fixing issue in hammer

izapolsk:
Many views have their own xpaths for FlashMessages. We need to unify that. 
Besides that there are flash messages stored in their own flash_text_div block in hammer. This also has to be sorted out
REQUIRES: https://github.com/RedHatQE/widgetastic.patternfly/pull/76

{{ pytest: cfme/tests/infrastructure/test_providers.py cfme/tests/automate/test_domain.py --long-running }}

### DIFF
--- a/cfme/base/ui.py
+++ b/cfme/base/ui.py
@@ -1620,7 +1620,7 @@ class AutomateSimulation(CFMENavigateStep):
 class AutomateImportExportBaseView(BaseLoggedInPage):
     # TODO This is currently overiding the base flash and should be renamed and efforts made
     # to update assocaited tests
-    flash = FlashMessages('.//div[@id="flash_msg_div"]')
+    flash = FlashMessages('div.import-flash-message')
     title = Text('.//div[@id="main-content"]//h1')
 
     @property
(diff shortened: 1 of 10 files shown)


cfme/base/ui.py
```diff
@@ -1620,7 +1620,7 @@ def step(self):
 class AutomateImportExportBaseView(BaseLoggedInPage):
     # TODO This is currently overiding the base flash and should be renamed and efforts made
     # to update assocaited tests
-    flash = FlashMessages('div.import-flash-message')
```
izapolsk: need to double check this. this change can potentially break some tests

mshriver:
Included infra/test_providers, which does some flash message asserting, looking great. Might want to include some automate tests using the view that you're concerned with.

izapolsk:
@mshriver, please review. PRT errors aren't related. I'll fix them in another PR.

mshriver:
The updates to wt.patternfly have gotten much more in depth, so scope has changed a bit here. Need to make sure we're covering a good variety of flash message use in the tests selected for PRT.

izapolsk:
@mshriver, there is one failure but it isn't related to FlashMessages changes. In addition, it appeared after code where previous error occurred .
I assume this is just flaky test and this error should go after restarting PRT.

PR #8236, opened 2018-12-04
[RFR] Testing Redfish physical server detail page's buttons

matejart:
__Adding tests__ for feature Redfish physical infrastructure provider to test that pressing power, LED identification and lifecycle menu buttons of a pysical server details works. This functionality is implemented in Hammer.

Depends on:
  * [#8234](https://github.com/ManageIQ/integration_tests/pull/8234) for the fix of flash message assertion

Related Bugzilla bugs:
  * https://bugzilla.redhat.com/show_bug.cgi?id=1574816
  * https://bugzilla.redhat.com/show_bug.cgi?id=1574817
  * https://bugzilla.redhat.com/show_bug.cgi?id=1574818
  * https://bugzilla.redhat.com/show_bug.cgi?id=1574820
  * https://bugzilla.redhat.com/show_bug.cgi?id=1574821

/cc. @miha-plesko

### DIFF
--- a/cfme/tests/physical_infrastructure/test_redfish_physical_server_details.py
+++ b/cfme/tests/physical_infrastructure/test_redfish_physical_server_details.py
@@ -2,14 +2,27 @@
 import pytest
 
 from cfme.physical.provider.redfish import RedfishProvider
+from cfme.utils.appliance.implementations.ui import navigate_to
 
 pytestmark = [pytest.mark.provider([RedfishProvider], scope="function")]
 
 
 @pytest.fixture(scope="function")
 def physical_server(appliance, provider, setup_provider_funcscope):
-    # Get and return the first physical server
+    """Get and return the first physical server."""
     yield appliance.collections.redfish_physical_servers.all(provider)[0]
+
+
+def assert_message(physical_server, expected_message):
+    """
+    Assert that the physical server details view displays the requested message.
+
+    Args:
+      physical_server: check the details view of this physical server.
+      expected_message: the message we expect to be displayed in the flash alert
+    """
+    view = navigate_to(physical_server, "Details")
+    view.flash.assert_message(expected_message)
 
 
 def test_redfish_physical_server_details_stats(physical_server):
@@ -20,3 +33,29 @@ def test_redfish_physical_server_details_stats(physical_server):
         initialEstimate: None
     """
     physical_server.validate_stats(ui=True)
+
+
+def test_redfish_power_buttons(physical_server, provider):
+    """
+    Test that pressing of the power buttons for physical server succeeds.
+
+    The test assumes that the buttons can be pressed in any order and that we
+    will get a flash that tells us of success regardless of the state that the
+    physical server is in. Here we only test that the request in the gui
+    succeeds.
+    """
+    power_actions = [
+        ("power_off", lambda: physical_server.power_off()),
+        ("power_on", lambda: physical_server.power_on()),
+        ("power_off_now", lambda: physical_server.power_off_immediately()),
+        ("restart", lambda: physical_server.restart()),
+        ("restart_now", lambda: physical_server.restart_immediately()),
+        ("turn_on_loc_led", lambda: physical_server.turn_on_led()),
+        ("turn_off_loc_led", lambda: physical_server.turn_off_led()),
+        ("blink_loc_led", lambda: physical_server.turn_blink_led()),
+    ]
+
+    for action_name, action in power_actions:
+        action()
+        assert_message(physical_server, "Requested {} of selected item.".format(
+            action_name))


cfme/tests/physical_infrastructure/test_redfish_physical_server_buttons.py
```diff
@@ -0,0 +1,76 @@
...
+      provider: the RedfishProvider instance
+      expected_message: the message we expect to be displayed in the flash alert
+    """
+    view = provider.create_view(PhysicalServerDetailsView, physical_server)
+    view.flash.assert_message(expected_message)
+
+
+# Power Button
+def test_redfish_server_power_off(physical_server, provider):
+    physical_server.power_off()
```
miha-plesko: What if server is powered off already? I mean, you pick first physical server from VMDB so you cannot know, right? I assume the "Power Off" button will be grayed-out in this case which will result in test failure. Similar goes for other tests.

cfme/tests/physical_infrastructure/test_redfish_physical_server_buttons.py
```diff
@@ -0,0 +1,76 @@
...
+
+def test_redfish_server_restart_immediately(physical_server, provider):
+    physical_server.restart_immediately()
+    assert_message(physical_server, provider, "Requested restart_now of selected item.")
+
+
+# Identify Button
+def test_redfish_server_turn_on_led(physical_server, provider):
+    physical_server.turn_on_led()
+    assert_message(physical_server, provider, "Requested turn_on_loc_led of selected item.")
```
miha-plesko: Interesting why `turn_on_loc_led` is not localized into a human-readable name. A bug somewhere in CloudForms I guess :)

cfme/tests/physical_infrastructure/test_redfish_physical_server_buttons.py
```diff
@@ -0,0 +1,76 @@
...
+
+def test_redfish_server_turn_blink_led(physical_server, provider):
+    physical_server.turn_blink_led()
+    assert_message(physical_server, provider, "Requested blink_loc_led of selected item.")
+
+
+# Lifecycle Button
+def test_redfish_server_lifecycle_provision(physical_server):
+    view = navigate_to(physical_server, "Provision")
+    assert view.is_displayed
```
miha-plesko: Did we forget to press the button here or is it hidden in `navigate_to`?

cfme/tests/physical_infrastructure/test_redfish_physical_server_buttons.py
```diff
@@ -0,0 +1,76 @@
+# -*- coding: utf-8 -*-
+import pytest
```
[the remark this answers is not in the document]
matejart: 👍

cfme/tests/physical_infrastructure/test_redfish_physical_server_buttons.py
```diff
@@ -0,0 +1,76 @@
...
+from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.common.physical_server_views import (
```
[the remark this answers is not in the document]
matejart: 👍

cfme/tests/physical_infrastructure/test_redfish_physical_server_buttons.py
```diff
@@ -0,0 +1,76 @@
...
+    PhysicalServerDetailsView,
+)
+from cfme.physical.provider.redfish import RedfishProvider
+
+pytestmark = [pytest.mark.provider([RedfishProvider], scope="function")]
+
+
+@pytest.fixture(scope="function")
```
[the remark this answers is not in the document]
matejart: Thank you, good to know! If it's ok, I'll keep it for now for clarity reasons.

cfme/tests/physical_infrastructure/test_redfish_physical_server_buttons.py
```diff
@@ -0,0 +1,76 @@
...
+def physical_server(appliance, provider, setup_provider_funcscope):
+    """Get and return the first physical server."""
+    yield appliance.collections.redfish_physical_servers.all(provider)[0]
```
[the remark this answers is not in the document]
matejart: In this case we access one instance - whichever instance - so the return value is not a collection.

cfme/tests/physical_infrastructure/test_redfish_physical_server_buttons.py
```diff
@@ -0,0 +1,76 @@
...
+      physical_server: check the details view of this physical server.
+      provider: the RedfishProvider instance
+      expected_message: the message we expect to be displayed in the flash alert
+    """
+    view = provider.create_view(PhysicalServerDetailsView, physical_server)
+    view.flash.assert_message(expected_message)
+
+
+# Power Button
+def test_redfish_server_power_off(physical_server, provider):
```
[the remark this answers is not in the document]
matejart: I prefer to keep the part of the context in the test name regardless of the file the test resides in, because it helps when using the `-k` switch of the `pytest`. If anything, the tests should be renamed to `test_redfish_physical_server_power_off`.

cfme/tests/physical_infrastructure/test_redfish_physical_server_buttons.py
matejart: All the buttons are always available, so no option is greyed-out as a function of the server's known state.

cfme/tests/physical_infrastructure/test_redfish_physical_server_buttons.py
matejart: None of them are. I am guessing that the flash just displays the raw enum value.

cfme/tests/physical_infrastructure/test_redfish_physical_server_buttons.py
matejart: This does not test the button pressing functionality (which apparently is not an action but just a link to the "Provision" page, so we just do it directly.

miha-plesko: IMAO we needn't this test because I assume you'd soon be implementing a more complete test that will actually perform provisioning; and it'll click on Lifecycle > Provision Physical Server button anyway. That being said, if you want to keep it here, please add a docstring that we're just testing whether button redirects us to the desired page.

cfme/tests/physical_infrastructure/test_redfish_physical_server_buttons.py
```diff
@@ -0,0 +1,76 @@
...
+def assert_message(physical_server, provider, expected_message):
+    """
+    Assert that the physical server details view displays the requested message.
+
+    Args:
+      physical_server: check the details view of this physical server.
+      provider: the RedfishProvider instance
+      expected_message: the message we expect to be displayed in the flash alert
+    """
+    view = provider.create_view(PhysicalServerDetailsView, physical_server)
```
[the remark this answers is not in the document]
matejart: 👍

cfme/tests/physical_infrastructure/test_redfish_physical_server_buttons.py
matejart: 👍

matejart:
@mshriver this PR is now ready for review.
Thank you!

cfme/tests/physical_infrastructure/test_redfish_physical_server_details.py
```diff
@@ -2,16 +2,29 @@
...
+from cfme.utils.appliance.implementations.ui import navigate_to
 
 pytestmark = [pytest.mark.provider([RedfishProvider], scope="function")]
 
 
 @pytest.fixture(scope="function")
 def physical_server(appliance, provider, setup_provider_funcscope):
-    # Get and return the first physical server
+    """Get and return the first physical server."""
     yield appliance.collections.redfish_physical_servers.all(provider)[0]
```
mshriver: Consider wrapping this so that the test can cleanly skip when there is no physical server available on the provider.

Otherwise the test ends in an ERROR state with IndexError.

cfme/tests/physical_infrastructure/test_redfish_physical_server_details.py
```diff
@@ -20,3 +33,45 @@ def test_redfish_physical_server_details_stats(physical_server):
         initialEstimate: None
     """
     physical_server.validate_stats(ui=True)
+
+
+# Power Button
+def test_redfish_server_power_off(physical_server, provider):
```
mshriver: This appears to be setup to assume that this specific set of tests is executing in sequence?  Not an acceptable pattern if so, as there are too many failure modes introduced. We cannot guarantee that these tests will execute on the same appliance, by the same pytest slave, much less in sequence.   What if I just decided to run the power_on test in isolation - the vm is already powered on, did I really test anything?

If it is assumed that the machine is in a powered on state when this test runs, that needs to be built into this test function, or into a fixture used by this test.

Same goes for the following tests that may be assuming a specific power state at the start of the test.

Few things, will enumerate:
1. Docblocks for these test methods, explaining any assumptions about the server state, or whether the state matters
2. Consider parametrizing some here to reduce the number of methods, maybe a single `test_redfish_server_power_control` test that goes through the power operations in turn.
3. Some assertion that the server actually went into the requested state? Currently this is just testing that I can make that button click in MIQ, but not that it has any effect on the server.  Look at using the wrapanapi classes/methods for this power state assertion.

mshriver: Now that I said all of this, I see this thread:
https://github.com/ManageIQ/integration_tests/pull/8236#discussion_r242804860

So I would then boil down my issues to:
1. The assumptions made here (all power operations are available always in UI) need to be in the test docblock
2. There is no need to have these separated into individual test functions, this needs to be parametrized.  Consider a namedTuple of button + flash message, and parametrize over a list of tuples for a single `test_redfish_power_buttons` test function.

There are also assumptions being made in calling `server.power_off` or `power_on` functions, in that they are using a UI context to click some button.  If the test is just that these buttons are available, and that clicking them produces flash messages, than I think the test should directly navigate to server details and use the widget methods to assert that the options are available, then click them.

Ultimately I think the test should be filled out to actually assert a power state, call a power operation, and then actually assert the power state change in addition to the flash message check.  We have power control test modules in the cloud/infra or cloud_infra_common folder, I suggest you check out how they're structured.

matejart: The only assumption we have is that any button can be pressed at any time regardless of the server's state. The test just makes sure that we can queue a request. So the tests can be run in any order. I will write that  in the docblocks to make it clearer.

I kept the single button press in each test method because I've taken the example from [Lenovo](https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/physical_infrastructure/ui/test_physical_server_details_buttons.py#L31) and it seemed simple and understandable enough. Of course in the interest of reducing the number of test methods (and thus the test execution time), I can put them in a single test as you suggested. Within this test I can also do the widget tests as you suggested.

Regarding the complexity of the depth of these tests, our intent is really to test that the GUI is properly wired with the provider in that clicking a button results in a reported success on the GUI. Admittedly we currently lack the assertion that clicking a certain button actually confers the corresponding command to the back-end. But we do not want to test the physical servers' state transitions here, because this part is very Redfish-vendor specific, so results might vary depending on the server the provider instance is connected to.

matejart: Done.

matejart:
I believe I addressed your comments now.

cfme/tests/physical_infrastructure/test_redfish_physical_server_details.py
```diff
@@ -20,3 +33,29 @@ def test_redfish_physical_server_details_stats(physical_server):
         initialEstimate: None
     """
     physical_server.validate_stats(ui=True)
+
+
+def test_redfish_power_buttons(physical_server, provider):
+    """
+    Test that pressing of the power buttons for physical server succeeds.
+
+    The test assumes that the buttons can be pressed in any order and that we
```
mshriver: :+1:

PR #8237, opened 2018-12-04
[RFR] added check for models and dialogs in log collection

lina-nikiforova:
{{ pytest: -v cfme/tests/configure/test_log_depot_operation.py::test_collect_log_depot }}

1. added check for models and dialogs (new feature introduced in 5.10)
2. During testing I saw some failures for ansible logs, so it's different if we're looking at the file named "Archive..." or "Current...".


digitronik:
@lina-nikiforova you need to rebase...

PR #8238, opened 2018-12-04
[1LP][RFR] Update appliance name

mshriver:
The appliance name should be a property, not an attribute, as it must be fetched from the appliance.

I did some searching through the framework, but did not find any uses of the appliance rename, or of its name attribute directly, where the property is not preferred.

### DIFF
--- a/cfme/base/__init__.py
+++ b/cfme/base/__init__.py
@@ -7,17 +7,14 @@ import sentaku
 from cfme.modeling.base import BaseCollection, BaseEntity
 from cfme.utils import ParamClassName
 from cfme.utils.appliance.implementations.ui import navigate_to
-from cfme.utils.blockers import BZ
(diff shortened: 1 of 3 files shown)


cfme/base/__init__.py
```diff
@@ -126,21 +126,11 @@ def get_master(self):
         else:
             server = server_collection.find_by(is_master=True)[0]
 
-        # TODO This entire hack needs to be removed once we are able to edit the server name again
         try:
-            name = server.name
+            name = getattr(server, 'name')
         except AttributeError:
-            logger.error('The EVM has no name, setting it to EVM')
-            if (self.appliance.version == LATEST or
```
mshriver: @izapolsk I've tested locally to confirm that the upstream appliances have a name set

PR #8239, opened 2018-12-05
[1LP][RFR] Fixed KeyError and AssertionError in inline automate method

ganeshhubale:
- This test was skipping because of ```Key Error``` while getting ```small_template``` of vSphere provider.
- Also fixed ```Assertion Error``` which is occurred while asserting flash message of data validation in ```MethodAddView```

{{ pytest: cfme/tests/automate/test_vmware_methods.py -k 'test_vmware_vimapi_hotadd_disk' --long-running --use-provider complete -v }}

### DIFF
--- a/cfme/automate/explorer/method.py
+++ b/cfme/automate/explorer/method.py
@@ -472,8 +472,8 @@ class MethodCollection(BaseCollection):
             validate = False
         if validate and not BZ(1499881, forced_streams=['5.9']).blocks:
             add_page.validate_button.click()
-            add_page.flash.assert_no_error()
             add_page.wait_displayed()
+            add_page.flash.assert_no_error()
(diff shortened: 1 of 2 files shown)


cfme/automate/explorer/method.py
```diff
@@ -472,6 +473,7 @@ def create(
         if validate and not BZ(1499881, forced_streams=['5.9']).blocks:
             add_page.validate_button.click()
             add_page.flash.assert_no_error()
+            add_page.wait_displayed()
```
mshriver: This should be above the flash message call to `assert_no_error` as well, just up one line please :)

ganeshhubale: sure @mshriver

PR #8242, opened 2018-12-05
[1LP][RFR]v2v Edit mapping test case

sshveta:
{{ pytest: cfme/tests/v2v/test_v2v_migrations_ui.py --use-provider rhv42 --use-provider vsphere67-nested --provider-limit 2 -vvvv --long-running -k test_edit_mapping_description}}

Reference PR 8174

### DIFF
--- a/cfme/v2v/migrations.py
+++ b/cfme/v2v/migrations.py
@@ -678,7 +678,6 @@ class InfrastructureMapping(BaseEntity, Updateable):
 
     def update(self, updates):
         view = navigate_to(self, 'Edit', wait_for_view=20)
-        assert view.wait_displayed()
         view.form.fill(updates)
 
 
@@ -839,8 +838,10 @@ class EditInfrastructureMapping(CFMENavigateStep):
     prerequisite = NavigateToAttribute('parent', 'All')
 
     def step(self):
-        self.prerequisite_view.infra_mapping_list.edit_mapping(
-            self.obj.name)
+        try:
+            self.prerequisite_view.infra_mapping_list.edit_mapping(self.obj.name)
+        except NoSuchElementException:
+            raise ItemNotFound("Mapping {} not found".format(self.obj.name))
 
 
 @navigator.register(MigrationPlanCollection, 'All')
--- a/cfme/v2v/migrations.py
+++ b/cfme/v2v/migrations.py
@@ -20,6 +20,7 @@ from cfme.base.login import BaseLoggedInPage
 from cfme.exceptions import ItemNotFound
 from cfme.modeling.base import BaseCollection, BaseEntity
 from cfme.utils.appliance.implementations.ui import navigator, CFMENavigateStep, navigate_to
+from cfme.utils.update import Updateable
 from cfme.utils.version import Version, VersionPicker
 from cfme.utils.wait import wait_for
 
@@ -72,6 +73,7 @@ class InfraMappingFormControlButtons(View):
     # common footer buttons for first 3 pages
     back_btn = Button('Back')
     next_btn = Button('Next')
+    save = Button('Save')
     cancel_btn = Button('Cancel')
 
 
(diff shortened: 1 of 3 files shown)


cfme/v2v/migrations.py
```diff
@@ -166,16 +166,16 @@ def fill(self, values):
...
             if self.cluster_selector.is_displayed:
                 self.cluster_selector.fill(cluster)
             for mapping in values[cluster]['mappings']:
                 self.source_datastores.fill(mapping['sources'])
                 self.target_datastores.fill(mapping['target'])
                 self.add_mapping.click()
         was_change = not self.mappings_tree.is_empty
-        if was_change:
-            self.logger.info("Fill operation was successful, click Next.")
+        self.cancel_btn.wait_displayed()
```
jawatts: **Question** Why are we waiting for the cancel button to be displayed here? A comment would be nice.

cfme/v2v/migrations.py
```diff
@@ -166,16 +166,16 @@ def fill(self, values):
...
+        if self.next_btn.is_displayed:
```
jawatts: What happens if the next button is not changed here?

cfme/v2v/migrations.py
sshveta: Done

cfme/v2v/migrations.py
sshveta: Removed this line . next button is always displayed

cfme/v2v/migrations.py
```diff
@@ -211,17 +210,21 @@ def fill(self, values):
...
             if self.cluster_selector.is_displayed:
                 self.cluster_selector.fill(cluster)
             for mapping in values[cluster]['mappings']:
                 self.source_networks.fill(mapping['sources'])
                 self.target_networks.fill(mapping['target'])
                 self.add_mapping.click()
         was_change = not self.mappings_tree.is_empty
-        if was_change:
-            self.logger.info("Fill operation was successful, click Next.")
+        # Cancel button is the common button on all pages so
```
mshriver: TODO:

I believe these actions would be more appropriate in after_fill, since its not part of actually filling these form values, and before_fill/after_fill would allow for each modal page to wait for elements before starting to fill, and to click next/save where appropriate.

sshveta: @mshriver :allow me to do this in next PR please where I am adding one more test case . 
Will make this change too .

cfme/v2v/migrations.py
```diff
@@ -815,6 +832,17 @@ def step(self):
...
 
 
+@navigator.register(InfrastructureMapping, 'Edit')
+class EditInfrastructureMapping(CFMENavigateStep):
+    VIEW = EditInfrastructureMappingView
+
+    prerequisite = NavigateToAttribute('parent', 'All')
+
+    def step(self):
+        self.prerequisite_view.infra_mapping_list.edit_mapping(
```
mshriver: I think there needs to be a check here on the return of `edit_mapping`, since if its `False` we have failed to navigate and shouldn't leave it to timeout on wait_displayed.

I think you should raise `ItemNotFound` on a `False` return from edit_mapping.

cfme/v2v/migrations.py
```diff
@@ -658,12 +670,17 @@ def plan_in_progress(self, vms_count=5):
...
-class InfrastructureMapping(BaseEntity):
+class InfrastructureMapping(BaseEntity, Updateable):
     """Class representing v2v infrastructure mappings"""
     name = attr.ib()
     description = attr.ib(default=None)
     form_data = attr.ib(default=None)
 
+    def update(self, updates):
+        view = navigate_to(self, 'Edit', wait_for_view=20)
+        assert view.wait_displayed()
```
mshriver: `wait_for_view` explicitly waited for 20s above, why additionally `wait_displayed` here on the same view? Is there some issue with the view getting displayed in a reasonable time? Do we need to trigger `is_displayed` on more than just the title appearing?

cfme/v2v/migrations.py
```diff
@@ -815,6 +831,19 @@ def step(self):
...
+@navigator.register(InfrastructureMapping, 'Edit')
+class EditInfrastructureMapping(CFMENavigateStep):
+    VIEW = EditInfrastructureMappingView
+
+    prerequisite = NavigateToAttribute('parent', 'All')
+
+    def step(self):
+        try:
+            self.prerequisite_view.infra_mapping_list.edit_mapping(self.obj.name)
+        except NoSuchElementException:
```
mshriver: Not quite - edit_mapping is already catching `NoSuchElementException` and returning a boolean - here we need to check against the boolean return and raise.

PR #8244, opened 2018-12-06
[1LP][RFR] Fix ssui dashboard test

niyazRedhat:
{{ pytest: cfme/tests/ssui/test_ssui_dashboard.py::test_current_service --long-running -v }}

### DIFF
--- a/cfme/services/dashboard/ssui.py
+++ b/cfme/services/dashboard/ssui.py
@@ -70,8 +70,7 @@ def results(self):
     view = self.create_view(MyServicesView)
     result = view.results.text
     int_result = re.search(r'\d+', result).group()
-    if self.appliance.version >= "5.8":
-        assert int_result == self.num_of_rows()
+    assert int_result == self.num_of_rows()
     return int_result
 
 


cfme/services/dashboard/ssui.py
```diff
@@ -70,8 +70,9 @@ def results(self):
     view = self.create_view(MyServicesView)
     result = view.results.text
     int_result = re.search(r'\d+', result).group()
-    if self.appliance.version >= "5.8":
-        assert int_result == self.num_of_rows()
+    # if self.appliance.version >= "5.8":
+    #     assert int_result == self.num_of_rows()
+    # TODO Uncomment above check when ManageIQ/integration_tests/issues/8243 is solved
```
jawatts: **Required** Have you considered using ```.blocks``` here instead? 
```In[7]: GH('ManageIQ/integration_tests:8243').blocks
Out[7]: True
```
https://github.com/ManageIQ/integration_tests/blob/master/cfme/utils/blockers.py#L124

That way, we won't have to manually remove this and open up another PR to uncomment these lines. In addition, I think the if statement checking that the appliance is greater and equal to 5.8 can be removed since we are no longer testing 5.8 or lower.

PR #8248, opened 2018-12-06
[1LP][RFR] added is_displayed for UtilizationView, fixed is_displayed for ServiceCatalogsDefaultView and added TimedOutError to exception for test_pull_splitter_persistence

lina-nikiforova:
{{ pytest: -v cfme/tests/webui/test_splitter.py::test_pull_splitter_persistence }}

1. missing `is_displayed` for `UtilizationView` led to broken navigation
2. `self.in_service_catalogs` is more specific for the `ServiceCatalogsDefaultView`
3. Added `TimedOutError` as some `is_displayed` checks rely on accordion, that can be hidden during this test

### DIFF
--- a/cfme/optimize/utilization.py
+++ b/cfme/optimize/utilization.py
@@ -22,6 +22,15 @@ class UtilizationView(BaseLoggedInPage):
             "Utilization",
         ]
 
+    @property
+    def is_displayed(self):
+        # region = CFME Region: Region 0 [0]'
+        region = self.extra.appliance.region()
+        return (
+            self.in_utilization and
+            self.title.text == 'Region "{}" Utilization Trend Summary'.format(region)
+        )
+
 
 @attr.s
 class Utilization(BaseEntity):
--- a/cfme/tests/webui/test_splitter.py
+++ b/cfme/tests/webui/test_splitter.py
@@ -1,6 +1,7 @@
 from xml.sax.saxutils import quoteattr, unescape
 
 import pytest
+from wait_for import TimedOutError
 
 # from cfme.cloud.instance import Instance
 # from cfme.infrastructure.config_management import ConfigManager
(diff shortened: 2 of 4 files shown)


cfme/tests/webui/test_splitter.py
```diff
@@ -72,9 +73,10 @@ def test_pull_splitter_persistence(request, appliance, model_object, destination
     navigate_to(appliance.server, 'Dashboard')
     try:
         navigate_to(model_object, destination)
-    except (TypeError, CannotScrollException):
+    except (TypeError, CannotScrollException, TimedOutError):
```
izapolsk: are you sure it won't hide some other TimedOutError ?

cfme/optimize/utilization.py
```diff
@@ -22,6 +22,16 @@ def in_utilization(self):
             "Utilization",
         ]
 
+    @property
+    def is_displayed(self):
+        # region = CFME Region: Region 0 [0]'
+        region = self.extra.appliance.server_region_string()
+        region_in_title = region.split(': ')[1].split(' [')[0]
```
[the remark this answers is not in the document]
lina-nikiforova: not sure it's needed for many things, but still looks better now :)
added `region` to the IPAppliance, to the same place where I took `server_region_string`

cfme/tests/webui/test_splitter.py
lina-nikiforova: Currently it works fine.
Actually, I got this error `TimedOutError('Could not do Waiting for view [ConfigurationView] to display at .../cfme/utils/appliance/implementations/ui.py:562 in time',)`. @izapolsk Is there any better way to pass this exception? it happens because `is_displayed` is based on accordions and in this test it's fine

izapolsk: hm, okay. let's merge it as it then.

PR #8250, opened 2018-12-06
[1LP][RFR] ipa external auth test in pod appliance

izapolsk:
{{pytest: cfme/tests/pod/test_appliance_crud.py -k test_pod_appliance_basic_ipa_auth --use-provider cm-pod2 --long-running --use-template-cache }}
REQUIRES: https://github.com/ManageIQ/wrapanapi/pull/349, https://github.com/ManageIQ/integration_tests/pull/8394


cfme/fixtures/authentication.py
```diff
@@ -13,6 +13,27 @@ def amazon_auth_provider():
         pytest.skip('amazon auth provider not found in auth_data.auth_providers, skipping test.')
 
 
+@pytest.fixture(scope='function')
+def ipa_auth_provider():
+    try:
+        return authutil.get_auth_crud('freeipa01')
```
mshriver: I don't think we should hardcode this, you should be able to parametrize `prov_key` based on what types of auth providers in the yaml, and use `auth_provider` fixture with the prov_key that matches the yaml.

freeipa01 is specific to our yaml config.

izapolsk: I just used the same approach from previous fixture in this file.

cfme/tests/pod/test_appliance_crud.py
```diff
@@ -121,6 +125,20 @@ def template_folder(template):
     return os.path.join(upload_folder, template['name'])
 
 
+@pytest.fixture(scope='function')
+def setup_ipa_auth_provider(temp_pod_appliance, ipa_auth_provider):
```
jawatts: **Question** Looks you have two fixtures doing the same thing?

izapolsk: hehe, you're right. I apparently moved it to global fixtures file and forgot to remove it here

cfme/tests/pod/test_appliance_crud.py
```diff
@@ -447,3 +485,129 @@ def test_pod_appliance_db_backup_restore(temp_pod_appliance, provider, setup_pro
...
+     - enable external httpd authentication in appliance
+     - deploy latest configmap generator
+     - generate new configuration
+     - deploy new httpd configmap configuration
+     - restart httpd pod
+     - to login to appliance using external credentials
+
+    Polarion:
+        assignee: izapolsk
+        initialEstimate: 1/2h
```
jawatts: **Required** Should have a casecomponent here as well

jawatts:
Looks like travis failed and prt failed

dajoRH:
*CFME QE Bot*

cfme/tests/pod/test_appliance_crud.py
izapolsk: nope, I was wrong both are necessary. since first fixture for regular appliance whereas second one overrides first one for custom appliance.

cfme/tests/pod/test_appliance_crud.py
```diff
@@ -219,6 +258,7 @@ def test_crud_pod_appliance(temp_pod_appliance, provider, setup_provider):
 
     Polarion:
         assignee: izapolsk
+        casecomponent: PodAppliance
```
jawatts: I think you want Containers here, but we should confirm with Sudhir. If he wants a different name, I can add it to Polarion and we would need to update https://github.com/ManageIQ/integration_tests/blob/master/conf/polarion_tools.yaml#L94

jawatts:
Moving to WIPTEST until we get an answer on what we want to casecomponent to be

PR #8252, opened 2018-12-07
[1LP][RFR] Added rest migrations tests

Yadnyawalkya:
Purpose or Intent
=================

__Adding tests__ for feature X to cover cases Y and Z. They were implemented in product version 5.10.0.27, 5.9.6.5.

{{pytest: cfme/tests/v2v/test_rest_migrations.py --use-provider vsphere65-nested --use-provider rhv42 --provider-limit 2 -vvvv}}

### DIFF
--- a/cfme/tests/v2v/test_rest_migrations.py
+++ b/cfme/tests/v2v/test_rest_migrations.py
@@ -1,0 +1,120 @@
+"""Test to check v2v migration rest API"""
+import fauxfactory
+import pytest
+
+from cfme.infrastructure.provider.rhevm import RHEVMProvider
+from cfme.infrastructure.provider.virtualcenter import VMwareProvider
+from cfme.markers.env_markers.provider import ONE_PER_VERSION
+
+
+pytestmark = [
+    pytest.mark.provider(
+        classes=[RHEVMProvider],
+        selector=ONE_PER_VERSION,
+        required_flags=["v2v"]
+    ),
+    pytest.mark.provider(
+        classes=[VMwareProvider],
+        selector=ONE_PER_VERSION,
+        fixture_name="second_provider",
+        required_flags=["v2v"]
+    )
+]
+
+
+@pytest.fixture(scope="function")
+def get_clusters(appliance, v2v_providers):
+    clusters = {}
+    source_cluster = v2v_providers.vmware_provider.data.get("clusters")[0]
+    target_cluster = v2v_providers.rhv_provider.data.get("clusters")[0]
+    cluster = appliance.rest_api.collections.clusters.all
+    for cl in cluster:
+        if cl.name == source_cluster:
+            clusters["source"] = "/api/clusters/{}".format(cl.id)
+        elif cl.name == target_cluster:
+            clusters["destination"] = "/api/clusters/{}".format(cl.id)
+    return clusters
+
+
+@pytest.fixture(scope="function")
+def get_datastores(appliance, v2v_providers):
+    datastores = {}
+    datastore = appliance.rest_api.collections.data_stores.all
+    source_ds = [
+        i.name for i in v2v_providers.vmware_provider.data.datastores if i.type == "nfs"][0]
+    target_ds = [
+        i.name for i in v2v_providers.rhv_provider.data.datastores if i.type == "nfs"][0]
+    for ds in datastore:
+        if ds.name == source_ds:
+            datastores["source"] = "/api/data_stores/{}".format(ds.id)
+        elif ds.name == target_ds:
+            datastores["destination"] = "/api/data_stores/{}".format(ds.id)
+    return datastores
+
+
+@pytest.fixture(scope="function")
+def get_networks(appliance, v2v_providers):
+    networks = {}
+    source_network = v2v_providers.vmware_provider.data.get("vlans", [None])[0]
+    target_network = v2v_providers.rhv_provider.data.get("vlans", [None])[0]
+    network = appliance.rest_api.collections.lans.all
+    for n in network:
+        if n.name == source_network:
+            networks["source"] = "/api/lans/{}".format(n.id)
+        elif n.name == target_network:
+            networks["destination"] = "/api/lans/{}".format(n.id)
+    return networks
+
+
+def test_rest_mapping_create(request, appliance, get_clusters, get_datastores, get_networks):
+    """Tests infrastructure mapping create"""
+    collection = appliance.rest_api.collections.transformation_mappings.action.create(
+        name=fauxfactory.gen_alphanumeric(),
+        description=fauxfactory.gen_alphanumeric(),
+        state="draft",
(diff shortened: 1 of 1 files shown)


cfme/tests/v2v/test_rest_migrations.py
```diff
@@ -0,0 +1,120 @@
...
+                get_networks
+            ]
+        }
+        for _ in range(2)
+    ]
+    mapping = collection.action.create(*data)
+
+    @request.addfinalizer
+    def _cleanup():
+        for m in mapping:
```
digitronik: Question:
`L107` creating more than one items... as you are iterating for deletion?

cfme/tests/v2v/test_rest_migrations.py
```diff
@@ -0,0 +1,120 @@
...
+
+@pytest.fixture(scope="function")
+def get_clusters(appliance, v2v_providers):
+    clusters = {}
+    source_cluster = v2v_providers.vmware_provider.data.get("clusters")[0]
+    target_cluster = v2v_providers.rhv_provider.data.get("clusters")[0]
+    cluster = appliance.rest_api.collections.clusters.all
+    for cl in cluster:
+        if cl.name == source_cluster:
+            clusters["source"] = "/api/clusters/{}".format(cl.id)
```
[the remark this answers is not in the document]
Yadnyawalkya: I also thought of adding `object.href` at time you raised this comment. Earlier it was accepting only url path, HTTP request kind of following,
```
{
 "name": "map1",
 "description": "clusters to clusters",
 "state": "draft",
 "transformation_mapping_items": [
     { "source": "/api/clusters/1", "destination": "/api/clusters/3" },
     { "source": "/api/data_stores/17", "destination": "/api/data_stores/6" },
     { "source": "/api/lans/1", "destination": "/api/lans/8" }     
   ]
}
```
Then I reached out to devel and as per other API implementation standard now we can also specify full url there and hence `object.href` will work. Thanks :)

cfme/tests/v2v/test_rest_migrations.py
Yadnyawalkya: Yup! Here I am checking collective delete.

cfme/tests/v2v/test_rest_migrations.py
```diff
@@ -0,0 +1,126 @@
...
+        fixture_name="second_provider",
+        required_flags=["v2v"]
+    )
+]
+
+
+@pytest.fixture(scope="function")
+def get_clusters(appliance, v2v_providers):
+    clusters = {}
+    source_cluster = v2v_providers.vmware_provider.data.get("clusters")[0]
```
mshriver: Please wrap these in a `try/except` block to catch the IndexError and call `pytest.skip` instead of an exception raising to the result.

cfme/tests/v2v/test_rest_migrations.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def test_rest_mapping_bulk_delete_from_collection(
+        request, appliance, get_clusters, get_datastores, get_networks):
+    """Tests infrastructure mapping bulk delete from collection.
+
+    Bulk delete operation deletes all specified resources that exist. When the
+    resource doesn't exist at the time of deletion, the corresponding result
+    has "success" set to false.
+    """
```
mshriver: Polarion docblock, who owns this, and what's the intialestimate (at least)?

cfme/tests/v2v/test_rest_migrations.py
Yadnyawalkya: sure, updating other indexing too. :+1:

cfme/tests/v2v/test_rest_migrations.py
```diff
@@ -0,0 +1,126 @@
...
+    clusters = {}
+    source_cluster = v2v_providers.vmware_provider.data.get("clusters")[0]
+    target_cluster = v2v_providers.rhv_provider.data.get("clusters")[0]
+    cluster_db = {
+        cluster.name: cluster for cluster in appliance.rest_api.collections.clusters.all
+    }
+
+    if source_cluster in cluster_db.keys():
+        clusters["source"] = cluster_db[source_cluster].href
+    if target_cluster in cluster_db.keys():
```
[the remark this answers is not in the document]
Yadnyawalkya: Did not added `else` since wanted to check both `source` and `target` things; also did not add `elif` since if first condition passed, it will not check second condition which is not we wanted.

Updated try/expect/skip there, let me know if you feel this is too much (which I believe) so will wrap two try/expect into one.

cfme/tests/v2v/test_rest_migrations.py
Yadnyawalkya: Updated! :100:

cfme/tests/v2v/test_rest_migrations.py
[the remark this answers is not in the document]
Yadnyawalkya: @mshriver did not added `else` since wanted to check both `source` and `target` things; also did not add `elif` since if first condition passed, it will not check second condition which is not we wanted.

Updated try/expect/skip there, let me know if you feel this is too much (which I believe) so will wrap two try/expect into one.

cfme/tests/v2v/test_rest_migrations.py
```diff
@@ -0,0 +1,183 @@
...
+        classes=[VMwareProvider],
+        selector=ONE_PER_TYPE,
+        fixture_name="second_provider",
+        required_flags=["v2v"]
+    )
+]
+
+
+@pytest.fixture(scope="function")
+def get_clusters(appliance, v2v_providers):
```
mshriver: These fixtures are repeated in many of the tests, with slightly different implementations. Some tests may need slightly different sources, but I think that the v2v team needs to evaluate creating a few common cluster/network/datastore fixtures in `cfme/fixtures/v2v.py` for use in all the tests.

cfme/tests/v2v/test_rest_migrations.py
```diff
@@ -0,0 +1,183 @@
...
+
+    try:
+        if target_network in network_db.keys():
+            networks["destination"] = network_db[target_network].href
+    except KeyError:
+        pytest.skip("Network:{target_network} not found in {network_list}".format(
+            target_network=target_network, network_list=network_db.keys()))
+    return networks
+
+
```
mshriver: Tier and requirement markers?

cfme/tests/v2v/test_rest_migrations.py
Yadnyawalkya: @mshriver  good point here, if we need to implement this then we need to rethink + redesign it from scratch since source and target entity is first point of implementation. So this can be technically possible but .. may be I will put this in WIP and work on this and try to minimize structure and lets see hows it look.

cc @sshveta  @kedark3  I am on it, thoughts?

PR #8254, opened 2018-12-10
Add test for validation of subnbet creation on Network Router

aljazkosir:
We navigate to Network router and click `Edit` button inside toolbar. Then we click on `Create L3 Cloud Subnet` which opens up a popup, we fill the popup with predefined values and check for the right response message.

Please review only 5th commit as other are from dependant PRs.

\cc @miha-plesko


cfme/tests/networks/nuage/test_inventory.py
```diff
@@ -180,3 +181,18 @@ def test_network_port_l2_container(setup_provider_modscope, provider, with_nuage
         ('relationships', 'Cloud tenant'): tenant_name,
         ('relationships', 'Cloud subnets'): '1',
     })
+
+
+def test_router_add_subnet(setup_provider_modscope, provider, with_nuage_sandbox_modscope):
```
miha-plesko: Please move this test into a new nuage/operations.py file. Also, please use the regular `with_nuage_sandbox` fixture here, not the modscope one because we're adding subnet to it and it's better to discard it afterwards.

cfme/tests/networks/nuage/test_inventory.py
```diff
@@ -180,3 +181,18 @@ def test_network_port_l2_container(setup_provider_modscope, provider, with_nuage
...
+
+    We navigate to router through Provider > Tenant > Network Router
+    """
+    sandbox = with_nuage_sandbox_modscope
+    tenant_name = sandbox['enterprise'].name
+    router_name = sandbox['domain'].name
+    tenant = provider.collections.cloud_tenants.instantiate(name=tenant_name, provider=provider)
+    router = tenant.collections.routers.instantiate(name=router_name)
+
+    router.add_subnet(random_name(), '100.100.100.0', '255.255.255.0', '100.100.100.1')
```
miha-plesko: Can we add a `wait_for` to wait and see if the subnet actually got created?

mshriver:
@aljazkosir this needs rebased before it can be reviewed. Please @-me when you've updated

aljazkosir:
@mshriver Sorry for the delay. This PR is now rebased with master. But it's dependant on two other PRS.
First one: https://github.com/ManageIQ/integration_tests/pull/8169
Second one: https://github.com/ManageIQ/integration_tests/pull/8205

dajoRH:
Lint report for commit c75353df07be6815b720f46b0c0b740b0aa2865b:

`cfme/networks/subnet.py`:
- :heavy_exclamation_mark: Line 231:69: **W292** *no newline at end of file*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

aljazkosir:
Thanks for merging dependent commits! This one should now be ready for a review :)

\cc @mshriver

cfme/tests/networks/nuage/test_operations.py
```diff
@@ -0,0 +1,39 @@
...
+    Ensure that subnet is added on network router
+
+    We navigate to router through Provider > Tenant > Network Router
+    """
+    sandbox = with_nuage_sandbox
+    tenant_name = sandbox['enterprise'].name
+    router_name = sandbox['domain'].name
+    tenant = provider.collections.cloud_tenants.instantiate(name=tenant_name, provider=provider)
+    router = tenant.collections.routers.instantiate(name=router_name)
+    subnet_name = random_name()
```
digitronik: Optional:
I would you `fauxfactory` to create random name.

cfme/tests/networks/nuage/test_operations.py
```diff
@@ -0,0 +1,39 @@
...
+    router.add_subnet(subnet_name, '100.100.100.0', '255.255.255.0', '100.100.100.1')
```
digitronik: Q: are you sure about such hard-coded value?

cfme/networks/network_router.py
```diff
@@ -55,6 +56,16 @@ def add_interface(self, subnet_name):
...
 
+    def add_subnet(self, subnet_name, address, netmask, gateway):
+        view = navigate_to(self, 'AddSubnet')
+        view.fill({'subnet_name': subnet_name,
+                   'address': address,
+                   'netmask': netmask,
+                   'gateway': gateway})
+        view.add.click()
+        success_msg = 'Creating subnet {subnet}'.format(subnet=subnet_name)
+        view.flash.assert_success_message(success_msg)
```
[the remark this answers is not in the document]
aljazkosir: I was following the pattern other similar methods have, for example if you look at the L46 (method `add_interface`) also has the similar assert function.

cfme/networks/views.py
```diff
@@ -518,6 +519,17 @@ class NetworkRouterAddInterfaceView(BaseLoggedInPage):
...
 
+class NetworkRouterAddCloudSubnetView(BaseLoggedInPage):
+    """ Represents Add Interface to Network Router page """
+    subnet_name = TextInput(id='name')
+    address = TextInput(id='address')
+    netmask = TextInput(id='netmask')
+    gateway = TextInput(id='gateway')
+    add = Button('Add')
+
+    is_displayed = displayed_not_implemented
```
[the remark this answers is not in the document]
aljazkosir: Same here, I was following pattern of the other edit/add popup views. I checked the code and I couldn't find any popups with the `is_displayed` method implemented, maybe because it's a popup and doesn't work like the other pages...

cfme/tests/networks/nuage/test_operations.py
```diff
@@ -0,0 +1,39 @@
+import pytest
+
+from cfme.utils.log import logger
+from cfme.utils.wait import wait_for
+from wrapanapi.utils.random import random_name
```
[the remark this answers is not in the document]
aljazkosir: I moved it to the third party imports.

cfme/tests/networks/nuage/test_operations.py
aljazkosir: Function `random_name()` is imported from the other file, where it is calling `fauxfactory` to generate a random name :).

cfme/tests/networks/nuage/test_operations.py
aljazkosir: It's just some random IPs, so the popup is filled with this data and we can check if the subnet was really created.

cfme/networks/network_router.py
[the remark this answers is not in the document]
digitronik: I agreed. its old implementation. but for flexibility we should just check for msg type means `error` or not.

cfme/networks/views.py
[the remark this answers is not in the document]
digitronik: @aljazkosir here also I'm agreed with you... but what you think we should implement more robust  implementation or not? I would like to add `is_displayed` with some unique checks. so it will help us to improve checks. 

https://github.com/ManageIQ/integration_tests/pull/8254/files#diff-78b559a596dffca17c9934f224cafa47R244

To help navmazing we should write more unique view checks.

cfme/tests/networks/nuage/test_operations.py
digitronik: other than getway... I'm able to get random data.but thanks for answer.

cfme/networks/network_router.py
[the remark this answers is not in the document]
aljazkosir: Sure, fine with me :)

cfme/networks/views.py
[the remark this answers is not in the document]
aljazkosir: Yes, I think it would be better to have some unique checks. I added is_displayed function, which checks if module title is "Create L3 Cloud Subnet'".

miha-plesko:
@mshriver I think you gave thumbsup on this one during the call last time. Did we perhaps just forget to merge it, or is there something to improve?

cfme/networks/network_router.py
```diff
@@ -55,6 +56,15 @@ def add_interface(self, subnet_name):
...
         view.flash.assert_success_message(success_msg)
 
+    def add_subnet(self, subnet_name, address, netmask, gateway):
+        view = navigate_to(self, 'AddSubnet')
+        view.fill({'subnet_name': subnet_name,
+                   'address': address,
+                   'netmask': netmask,
+                   'gateway': gateway})
+        view.add.click()
+        view.flash.assert_no_error()
```
mshriver: For the record this would be a good pattern to push to the method above, `add_interface`, and the CRUD methods for this class.   Make specific flash message assertions in the test using this method, just make `assert_no_error()` calls in the method itself.

cfme/networks/views.py
```diff
@@ -518,6 +519,20 @@ class NetworkRouterAddInterfaceView(BaseLoggedInPage):
...
+    modal_title = Text('//div[@class="modal-header"]//h4')
+    subnet_name = TextInput(id='name')
+    address = TextInput(id='address')
+    netmask = TextInput(id='netmask')
+    gateway = TextInput(id='gateway')
+    add = Button('Add')
+
+    @property
+    def is_displayed(self):
+        return self.modal_title.text == 'Create L3 Cloud Subnet'
```
mshriver: There is no check here that this is also the add page for NetworkRouter CloudSubnet, and not just the basic CloudSubnet view.  Note that `class SubnetAddView` in this same module does not implement is_displayed - and I don't think this view should either.  Navigation to add CloudSubnet to a specific networkRouter needs to validate (or always navigate) that its adding the subnet to the correct NetworkRouter - and we aren't on an `Add` page from navigating through a different NetworkRouter.

cfme/networks/views.py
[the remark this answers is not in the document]
mshriver: The reason, AFAIK, that these views are specifically not using is_displayed is because of the interconnected nature of the collections and entities. Just checking the title here is not sufficient, as this is a cloud subnet tied to a specific Network Router entity.

Sometimes adding is_displayed makes the framework _less_ robust, as the check is insufficient and produces false positives - making for strange test failures or behavior, or even worse a false positive test result.

cfme/networks/views.py
mshriver: Just saw the old comments on this, and @digitronik suggestion to implement a method. I commented there as well, but figured I would complete the loop on this newer commit.

Apologies for the back and forth @aljazkosir, but I think you were right in the first place to not implement is_displayed.  It means this view always navigates, but that is better than incorrectly identifying a view and adding the subnet to the wrong NetworkRouter entity.

cfme/networks/network_router.py
mshriver: To be clear - does not need to be changed in this PR, can be a separate enhancement PR.

mshriver:
@aljazkosir We have changed to using `reorder-python-imports` for pre-commit, and your imports are slightly out of order, which is why travis is failing:
https://travis-ci.org/ManageIQ/integration_tests/jobs/498082942#L533

You can either make that change yourself and commit, or you can setup pre-commit in your local environment. Installing the pre-commit hooks locally means when you try to `git commit`, it will run all of our hooks from `.pre-commit-config.yaml`, just like travis does. This will fail, and automatically apply the necessary changes to your working tree, allowing you to review them before committing again.

https://pre-commit.com/#usage

cfme/tests/networks/nuage/test_operations.py
```diff
@@ -0,0 +1,40 @@
+import pytest
+
+from wrapanapi.utils.random import random_name
```
mshriver: I think it would be better to use fauxfactory directly for this purpose. 

https://pypi.org/project/fauxfactory/

cfme/tests/networks/nuage/test_operations.py
mshriver: Please use a non-routable network (10.0.0.0/24, 172.16.0.0/16, 192.168.0.0/16) unless the UI blocks these from being added.

cfme/tests/networks/nuage/test_operations.py
mshriver: Noted @aljazkosir but wrapanpi could change its random_name method, as its a util for use with wrapanpi, not necessarily for direct use in test. It could even just change the default to something incompatible with the subnet name field. Using fauxfactory directly allows you to control why type (alphanumeric, numeric, alpha) and length directly.

cfme/tests/networks/nuage/test_operations.py
```diff
@@ -0,0 +1,40 @@
...
+    sandbox = with_nuage_sandbox
+    tenant_name = sandbox['enterprise'].name
+    router_name = sandbox['domain'].name
+    tenant = provider.collections.cloud_tenants.instantiate(name=tenant_name, provider=provider)
+    router = tenant.collections.routers.instantiate(name=router_name)
+    subnet_name = random_name()
+    router.add_subnet(subnet_name, '100.100.100.0', '255.255.255.0', '100.100.100.1')
+    subnet = get_object_from_db_with_timeout(provider.appliance, 'cloud_subnets', subnet_name)
+
+    assert subnet is not None
```
mshriver: I think this test should be doing a few additional things to make it more complete in its validation.

`router.add_subnet` should (in my opinion) be instantiating and returning a Subnet entity instance.  You should be able to validate its existence by calling its `exists` method (which should navigate to its details page, through the Router entity).

The test should also be validating that the record pulled from the DB has a connection to the network router entity, not that its a generic subnet.

aljazkosir: I added the database check which validates that a subnet has a proper network_router. I also added subnet instance to the return statement of `router.add_subnet`. But do we really need to call `exists` method, shouldn't database check be enough?

cfme/networks/network_router.py
aljazkosir: I think I would rather create a new PR for this, since it's an enhancement related to other entities. :)

cfme/tests/networks/nuage/test_operations.py
mshriver: I'll leave that up to you, but validation that the entity is in the DB is not the same as validation that the entity is displayed correctly in the UI, and can be navigated to through the NetworkRouter.

cfme/tests/networks/nuage/test_operations.py
mshriver: This change hasn't been made yet, but the PR looks merge-ready otherwise - @aljazkosir Please confirm whether or not you can change this subnet IP range.

mshriver:
Collection looks good locally using RH's cfme_data.yaml.

cfme/tests/networks/nuage/test_operations.py
aljazkosir: Ops, I though I pushed this, it appears I only commited it though..

cfme/tests/networks/nuage/test_operations.py
aljazkosir: Hmm, you are right..., i think adding one extra assert won't hurt anyone :)

PR #8256, opened 2018-12-10
[RFR]v2v edit mapping fields

sshveta:
{{ pytest: cfme/tests/v2v/test_v2v_migrations_ui.py --use-provider rhv42 --use-provider vsphere67-nested --provider-limit 2 -vvvv --long-running -k test_edit_mapping_fields}}

1) Extended edit mapping test case to edit fields like datastore and network too .
2) refactored some code .

### DIFF
--- a/cfme/fixtures/v2v.py
+++ b/cfme/fixtures/v2v.py
@@ -165,7 +165,7 @@ def get_vm(request, appliance, second_provider, template, datastore='nfs'):
     request.addfinalizer(lambda: vm_obj.cleanup_on_provider())
     return vm_obj
 
-
+@pytest.fixture(scope='function')
 def _form_data(second_provider, provider):
     form_data = {
         'general': {
@@ -251,6 +251,28 @@ def form_data_single_network(request, second_provider, provider):
         }
     })
     return form_data
+
+
+@pytest.fixture(scope='function')
+def edited_form_data(second_provider, provider):
+    edited_form_data = {
+        'general': {
+            'description': "my edited description"},
+        'cluster': {},
+        'datastore': {
+            'Cluster ({})'.format(provider.data.get('clusters')[0]): {
+                'mappings': [_form_data_mapping('datastores', second_provider, provider,
+                                                'iscsi', 'iscsi')]
+            }
+        },
+        'network': {
+            'Cluster ({})'.format(provider.data.get('clusters')[0]): {
+                'mappings': [_form_data_mapping('vlans', second_provider, provider,
+                                                'DPortGroup', 'Storage - VLAN 33')]
+            }
+        }
+    }
+    return edited_form_data
 
 
 @pytest.fixture(scope='function')
--- a/cfme/fixtures/v2v.py
+++ b/cfme/fixtures/v2v.py
@@ -165,7 +165,7 @@ def get_vm(request, appliance, second_provider, template, datastore='nfs'):
     request.addfinalizer(lambda: vm_obj.cleanup_on_provider())
     return vm_obj
 
-@pytest.fixture(scope='function')
+
 def _form_data(second_provider, provider):
     form_data = {
         'general': {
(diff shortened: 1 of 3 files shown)


sshveta:
Test test_edit_mapping_fields passed in PRT .
Other failures are not related to the changes made

cfme/v2v/migrations.py
```diff
@@ -132,12 +138,11 @@ def fill(self, values):
             self.target_clusters.fill(mapping['target'])
             self.add_mapping.click()
         was_change = not self.mappings_tree.is_empty
-        self.next_btn.click()
+        self.after_fill()
```
jawatts: I don't think this is quite right here. `after_fill` is called by `View.fill()`

https://github.com/RedHatQE/widgetastic.core/blob/master/src/widgetastic/widget/base.py#L987

cfme/v2v/migrations.py
```diff
@@ -174,13 +179,11 @@ def fill(self, values):
                 self.target_datastores.fill(mapping['target'])
                 self.add_mapping.click()
         was_change = not self.mappings_tree.is_empty
-        self.next_btn.wait_displayed()
-        self.next_btn.click()
+        self.after_fill()
```
jawatts: same comment as above

cfme/v2v/migrations.py
sshveta: @jawatts : I also thought so .. but something is different in the fill here. 
The after_fill was not being called automatically .
May be  @mshriver will be able to tell why it is not called .

cfme/fixtures/v2v.py
```diff
@@ -165,7 +165,7 @@ def get_vm(request, appliance, second_provider, template, datastore='nfs'):
     request.addfinalizer(lambda: vm_obj.cleanup_on_provider())
     return vm_obj
 
-
+@pytest.fixture(scope='function')
```
kedark3: Hey Shveta, can you help me understand why are we making this as a fixture? Ytale had made it more of like a helper function when he designed it.

cfme/tests/v2v/test_v2v_migrations_ui.py
```diff
@@ -427,21 +427,21 @@ def test_migration_rbac(appliance, new_credential, v2v_providers):
 
 
 @pytest.mark.ignore_stream("5.9")
-@pytest.mark.parametrize('form_data_single_datastore', [['nfs', 'nfs']], indirect=True)
-def test_edit_mapping_description(appliance, v2v_providers, form_data_single_datastore,
-                                    host_creds, conversion_tags, soft_assert):
+def test_edit_mapping_fields(appliance, v2v_providers, _form_data,
```
kedark3: I think here you are using `_form_data` that I referred in comment above. 
**Suggestion**  Can we rather write a new fixture that encapsulates the `_form_data` call instead? That way we leave `_form_data` as it was intended to work/be used.

cfme/fixtures/v2v.py
sshveta: Hey Kedar , yes I needed a simple form_data to fill and edit and didn't want to make the test case complicated with parametrization just for editing hence used form_data

mshriver: That doesn't make it clear to me why it was marked as fixture? If you just needed the form data, you could have called it as a function within your test call phase?

cfme/tests/v2v/test_v2v_migrations_ui.py
mshriver: **Required**

`_form_data` itself was not designed to be a fixture, there are other fixtures in `cfme/fixtures/v2v` that use it as a method to generate a base dictionary.

cfme/v2v/migrations.py
mshriver: This method is overriding View.fill(), and yes needs to call after_fill() just like View.fill() does.  The call needs to pass was_change, and the overriding implementation should define after_fill to take the value, even if its not used, for consistency.

PR #8258, opened 2018-12-11
[RFR] 5.10: Fixed main navigation in UI

Yadnyawalkya:
Purpose or Intent
=================
 
5.10.0.28 navigations - 

screenshot from 2018-12-11 14-40-57

Tested locally, small UI fix affecting V2V automation job.


Yadnyawalkya:
@izapolsk can you take look at this, we wanted it to be in asap.
This is causing failures on v2v Jenkin jobs.

Yadnyawalkya:
@izapolsk can you take look at this, we wanted it to be in asap.
This is causing failures on v2v Jenkin jobs. Tried to contact you from RC, seems like it is problematic today.

kedark3:
:+1:

PR #8259, opened 2018-12-11
[1LP][RFR] Add test from flavors management RFE

jadhaj:
Purpose or Intent
=================
Added test: test_create_instance_with_private_flavor

{{ pytest: cfme/tests/openstack/cloud/test_flavors.py::test_create_instance_with_private_flavor }}

### DIFF
--- a/cfme/tests/openstack/cloud/test_flavors.py
+++ b/cfme/tests/openstack/cloud/test_flavors.py
@@ -6,6 +6,7 @@ import pytest
 from cfme.cloud.instance.openstack import OpenStackInstance
 from cfme.cloud.provider.openstack import OpenStackProvider
 from cfme.utils.appliance.implementations.ui import navigator, navigate_to
+from cfme.utils.blockers import BZ
 from cfme.utils.wait import wait_for
 from widgetastic.utils import partial_match
 
@@ -38,6 +39,7 @@ def zero_disk_flavor(provider):
 
 @pytest.fixture(scope='function')
 def private_flavor(appliance, provider):
+    prov_data = provider.data['provisioning']
     collection = appliance.collections.cloud_flavors
     private_flavor = collection.create(name=fauxfactory.gen_alpha(),
                                provider=provider,
@@ -46,7 +48,8 @@ def private_flavor(appliance, provider):
                                disk=DISK_SIZE,
                                swap=SWAP_SIZE,
                                rxtx=RXTX,
-                               is_public=False)
+                               is_public=False,
+                               tenant=prov_data['cloud_tenant'])
 
     view = appliance.browser.create_view(navigator.get_class(collection, 'All').VIEW)
     view.flash.assert_success_message(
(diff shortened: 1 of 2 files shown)


cfme/tests/openstack/cloud/test_flavors.py
```diff
@@ -199,3 +236,45 @@ def test_flavor_details(appliance, soft_assert, private_flavor):
...
+    prov_data = provider.data['provisioning']
+    power_state = view.entities.summary('Power Management').get_text_of('Power State')
+    assert power_state == OpenStackInstance.STATE_ON
+
+    vm_tmplt = view.entities.summary('Relationships').get_text_of('VM Template')
+    soft_assert(vm_tmplt == prov_data['image']['name'])
+
+    flavors = [f.name for f in provider.mgmt.api.flavors.list()]
+    flavor = view.entities.summary('Relationships').get_text_of('Flavor')
+    soft_assert(flavor in flavors)
```
digitronik: Required:
I don't think here you need to use `soft_assert` from the provider side. and `L252` you can use as flavor has `all` method. More stick to `cfme` than a provider. what do you think?
`flavor in appliance.collections.cloud_flavors.all()`

cfme/tests/openstack/cloud/test_flavors.py
```diff
@@ -199,3 +236,45 @@ def test_flavor_details(appliance, soft_assert, private_flavor):
...
+
+    # Assert other relationships in a loop
+    props = [('Availability Zone', 'availability_zone'),
```
digitronik: Optional:
I would treat this as `database` and create `dict` for it.

cfme/tests/openstack/cloud/test_flavors.py
```diff
@@ -199,3 +236,45 @@ def test_flavor_details(appliance, soft_assert, private_flavor):
...
+             ('Virtual Private Cloud', 'cloud_network')]
+
+    for p in props:
+        v = view.entities.summary('Relationships').get_text_of(p[0])
+        soft_assert(v == prov_data[p[1]])
+
+
+    collection = provider.appliance.collections.cloud_flavors
+    flavor_obj = collection.instantiate(flavor, provider)
+    soft_assert(flavor_obj.instance_count == 1)
```
digitronik: Required:
Please avoid using `soft_assert`. `Instance` count should be `1`; so we should have to `assert` it.

cfme/tests/openstack/cloud/test_flavors.py
```diff
@@ -199,3 +236,45 @@ def test_flavor_details(appliance, soft_assert, private_flavor):
...
+
+    flavor_obj.delete()
+    view = provider.appliance.browser.create_view(navigator.get_class(collection, 'All').VIEW)
```
digitronik: Required:
In such case please just use
`view = naviagate_to(collection, "All")`
If you are on the same page. then it will automatically avoid redundant navigation; Actually @jadhaj that's beauty of `navimazing`.

cfme/tests/openstack/cloud/test_flavors.py
jadhaj: The logic was not to compare cfme to itself but if it is OK to use lavor in appliance.collections.cloud_flavors.all(), I am OK with that

PR #8265, opened 2018-12-13
[1LP][RFR] Update "test_network_provider_add_with_bad_credentials" test

aljazkosir:
Update "test_network_provider_add_with_bad_credentials" test with new
warning about failed login.

\cc @miha-plesko, @mshriver

### DIFF
--- a/cfme/tests/networks/test_network_providers.py
+++ b/cfme/tests/networks/test_network_providers.py
@@ -51,7 +51,10 @@ def test_network_provider_add_with_bad_credentials(provider):
     default_credentials = provider.default_endpoint.credentials
 
     # default settings
-    flash = 'Login failed due to a bad username or password.'
+    if provider.appliance.version < '5.10':
+        flash = 'Login failed due to a bad username or password.'
+    else:
+        flash = 'Login failed due to a bad username, password or unsupported API version.'
     default_credentials.principal = "bad"
     default_credentials.secret = 'notyourday'
 


cfme/tests/networks/test_network_providers.py
```diff
@@ -51,7 +51,7 @@ def test_network_provider_add_with_bad_credentials(provider):
     default_credentials = provider.default_endpoint.credentials
 
     # default settings
-    flash = 'Login failed due to a bad username or password.'
+    flash = 'Login failed due to a bad username, password or unsupported API version.'
```
jawatts: **Question** Is this change in both 5.9 and 5.10? If it isn't, it will have to be picked based on the version of the appliance.

aljazkosir: Yes, you are right. 5.9 and earlier have old value. I pushed new changes.

cfme/tests/networks/test_network_providers.py
```diff
@@ -51,7 +51,9 @@ def test_network_provider_add_with_bad_credentials(provider):
...
+    if provider.appliance < 5.10:
+        flash = 'Login failed due to a bad username or password.'
```
jawatts: Close! Here you are getting the IPappliance object. 
```In[31]: provider.appliance
Out[31]: IPAppliance.from_json('{"container": null, "ui_port": 443, "ssh_port": 22, "openshift_creds": {}, "ui_protocol": "https", "hostname": "10.8.198.243", "db_port": 5432, "project": null, "is_dev": false, "version": "5.10.0.29", "db_host": null, "browser_steal": false}')
```
You will want to go one step further and call the version property:
```In[30]: provider.appliance.version
Out[30]: Version('5.10.0.29')
```

aljazkosir: oh, yes, of course ... Should be good now :)

izapolsk: and could you put both statements in one if/else block to make it more neat

aljazkosir: done :)

PR #8266, opened 2018-12-13
[1LP][RFR] CB: test check quotes

digitronik:
Purpose or Intent
=================
automating BZ-1646905

{{pytest: cfme/tests/automate/custom_button/test_buttons.py::test_custom_button_quotes -v}}

### DIFF
--- a/cfme/tests/automate/custom_button/test_buttons.py
+++ b/cfme/tests/automate/custom_button/test_buttons.py
@@ -5,6 +5,9 @@ import pytest
 from widgetastic_patternfly import Dropdown
 
 from cfme import test_requirements
+from cfme.cloud.provider.openstack import OpenStackProvider
+from cfme.infrastructure.provider.virtualcenter import VMwareProvider
+from cfme.markers.env_markers.provider import ONE_PER_TYPE
 from cfme.tests.automate.custom_button import OBJ_TYPE, OBJ_TYPE_59, TextInputDialogView
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.blockers import BZ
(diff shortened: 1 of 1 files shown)


cfme/tests/automate/custom_button/test_buttons.py
```diff
@@ -257,3 +262,63 @@ def test_open_url_availability(appliance):
         view.add_button.click()
         view.flash.assert_message("URL can be opened only by buttons for a single entity")
     view.cancel_button.click()
+
+
+@pytest.mark.provider([VMwareProvider], override=True, scope="function", selector=ONE_PER_TYPE)
+@pytest.mark.uncollectif(
```
jawatts: note for reviewers: no bug for 5.9 has been cloned yet. 

@digitronik I think using `@pytest.mark.ignore_stream("5.9")' would be better here

digitronik: > note for reviewers: no bug for 5.9 has been cloned yet.
> 
> @digitronik I think using `@pytest.mark.ignore_stream("5.9")' would be better here

@jawatts `ignore_stream` option good if test is `not applicable for specific version`. Its will not going to provide specific reason. Means if feature comes in `5.10`  and I don't want to run that test against `5.9` then this option  is `better` but not good to use it when `uncollected`  is `temporary`. 

> I'd vote for `pytest.mark.meta(blockers=[BZ(1646905, forced_streams=['5.9'])])`

@mshriver your `vote` always valuable :)... I was selected same option initially but after that I understand one behavior of `BZ` options. I logged this BZ against `5.9` but its fixed for `5.10`. so `BZ` will not `uncollect` it.

After understanding both scenarios. I selected `uncollefctif` as  its providing me proper `reason` attribute.  which  will remind me every time about this BZ.

cfme/tests/automate/custom_button/test_buttons.py
```diff
@@ -2,8 +2,13 @@
 import fauxfactory
 import pytest
 
+from widgetastic_patternfly import Dropdown
+
 from cfme import test_requirements
-from cfme.tests.automate.custom_button import OBJ_TYPE, OBJ_TYPE_59
+from cfme.cloud.provider.openstack import OpenStackProvider
```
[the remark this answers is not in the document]
digitronik: It was my  bad.

PR #8267, opened 2018-12-13
[1LP][RFR] CB: Added test for open url

digitronik:
- Added test for open url functionality

{{pytest: cfme/tests/automate/custom_button/test_infra_objects.py::test_open_url -vvv}}

### DIFF
--- a/cfme/tests/automate/custom_button/test_infra_objects.py
+++ b/cfme/tests/automate/custom_button/test_infra_objects.py
@@ -391,16 +391,13 @@ def test_open_url(request, setup_obj, button_group, method):
         initialEstimate: 1/2
         caseimportance: high
         testSteps:
-            prerequisites:
-                * Appliance with Infra provider
-
-            Steps:
-                * Create ruby method for url functionality
-                * Create custom button group with the Object type
-                * Create a custom button with open_url option and respective method
-                * Navigate to object Detail page
-                * Execute custom button
-                * Check new tab open or not with respective url
+            1. Appliance with Infra provider
+            2. Create ruby method for url functionality
+            3. Create custom button group with the Object type
+            4. Create a custom button with open_url option and respective method
+            5. Navigate to object Detail page
+            6. Execute custom button
+            7. Check new tab open or not with respective url
     """
 
     group, obj_type = button_group
@@ -431,6 +428,11 @@ def test_open_url(request, setup_obj, button_group, method):
     open_url_window = set(view.browser.selenium.window_handles) ^ {main_window}
 
     view.browser.selenium.switch_to_window(open_url_window.pop())
+
+    @request.addfinalizer
+    def _reset_window():
+        if view.browser.selenium.current_window_handle != main_window:
+            view.browser.selenium.close()
+            view.browser.selenium.switch_to_window(main_window)
+
     assert "example.com" in view.browser.url
-    view.browser.selenium.close()
-    view.browser.selenium.switch_to_window(main_window)
--- a/cfme/tests/automate/custom_button/test_infra_objects.py
+++ b/cfme/tests/automate/custom_button/test_infra_objects.py
@@ -64,7 +64,7 @@ def method(cls):
         name="inst_{}".format(fauxfactory.gen_alphanumeric(4)),
         fields={"meth1": {"value": meth.name}},
     )
-    yield meth
+    yield instance
     meth.delete_if_exists()
     instance.delete_if_exists()
 
--- a/cfme/tests/automate/custom_button/test_infra_objects.py
+++ b/cfme/tests/automate/custom_button/test_infra_objects.py
@@ -52,11 +52,11 @@ def method(cls):
         name="meth_{}".format(fauxfactory.gen_alphanumeric(4)),
         script=dedent(
             """
-        # add google url to open
-        vm = $evm.root['vm']
-        $evm.log(:info, "Opening google url")
-        vm.remote_console_url = "https://google.com"
-        """
+            # add google url to open
+            vm = $evm.root['vm']
+            $evm.log(:info, "Opening url")
+            vm.remote_console_url = "http://example.com"
+            """
         ),
     )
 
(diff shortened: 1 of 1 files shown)


cfme/tests/automate/custom_button/test_infra_objects.py
```diff
@@ -336,3 +377,56 @@ def test_custom_button_expression(appliance, request, setup_obj, button_group, e
...
+        * Create custom button group with the Object type
+        * Create a custom button with open_url option and respective method
+        * Navigate to object Detail page
+        * Execute custom button
+        * Check new tab open or not with respective url
+
+    Polarion:
+        assignee: ndhandre
+        initialEstimate: 1/2
+
```
jawatts: missing caseimportance

cfme/tests/automate/custom_button/test_infra_objects.py
```diff
@@ -28,6 +30,45 @@
...
+@pytest.fixture(scope="module")
+def method(cls):
+    meth = cls.methods.create(
+        name="meth_{}".format(fauxfactory.gen_alphanumeric(4)),
+        script=dedent(
+            """
+        # add google url to open
+        vm = $evm.root['vm']
+        $evm.log(:info, "Opening google url")
+        vm.remote_console_url = "https://google.com"
```
mshriver: I don't see a need to actually resolve a domain here, especially one that ends up loading everything google will. The URL doesn't need to resolve or render a page for the test to validate the browser URL, so let's use an IP/domain from the IANA special-use list for the test:

https://www.iana.org/assignments/special-use-domain-names/special-use-domain-names.xhtml

cfme/tests/automate/custom_button/test_infra_objects.py
```diff
@@ -339,3 +380,57 @@ def test_custom_button_expression(appliance, request, setup_obj, button_group, e
...
+        system="Request",
+        request=method.name,
+    )
+    request.addfinalizer(button.delete_if_exists)
+
+    view = navigate_to(setup_obj, "Details")
+    custom_button_group = Dropdown(view, group.hover)
+    assert custom_button_group.has_item(button.text)
+
+    # To-Do: Move windows handling functionality to browser
```
mshriver: We have the ConsoleMixin class to provide some browser interaction for selecting windows, which would be good to move into browser itself. 

https://github.com/ManageIQ/integration_tests/blob/master/cfme/common/vm_console.py

digitronik:
@mshriver failure not related to this newly added test... It was due https://github.com/ManageIQ/integration_tests/commit/1428426e350d79142d016bc29b91a6b15e2ba6df#diff-798139e1f33422c95de0f06621f10e01R332 . I am suspect those are version dependent change means `chrome browser versions`. I don't want to go  for this changes in same PR.

cfme/tests/automate/custom_button/test_infra_objects.py
```diff
@@ -28,6 +30,45 @@
...
+@pytest.fixture(scope="module")
+def method(cls):
+    meth = cls.methods.create(
+        name="meth_{}".format(fauxfactory.gen_alphanumeric(4)),
+        script=dedent(
+            """
+            # add google url to open
+            vm = $evm.root['vm']
+            $evm.log(:info, "Opening url")
+            vm.remote_console_url = "http://example.com"
```
mshriver: :+1:

cfme/tests/automate/custom_button/test_infra_objects.py
```diff
@@ -339,3 +380,57 @@ def test_custom_button_expression(appliance, request, setup_obj, button_group, e
...
+    wait_for(
+        lambda: len(view.browser.selenium.window_handles) > initial_count,
+        timeout=120,
+        message="Check for window open",
+    )
+    open_url_window = set(view.browser.selenium.window_handles) ^ {main_window}
+
+    view.browser.selenium.switch_to_window(open_url_window.pop())
+    assert "example.com" in view.browser.url
+    view.browser.selenium.close()
```
mshriver: These need to be in a finalizer, in case the assertion fails. Can't leave the browser on the wrong tab.

cfme/tests/automate/custom_button/test_infra_objects.py
```diff
@@ -339,3 +380,62 @@ def test_custom_button_expression(appliance, request, setup_obj, button_group, e
...
+@pytest.mark.uncollectif(lambda button_group: "VM_INSTANCE" not in button_group)
+def test_open_url(request, setup_obj, button_group, method):
+    """ Test Open url functionality of custom button.
+
+    Polarion:
+        assignee: ndhandre
+        initialEstimate: 1/2
+        caseimportance: high
+        testSteps:
+            prerequisites:
```
mshriver: If testSteps is going to be used, it needs to be an enumerated list of steps.

```
Polarion:
    testSteps:
        1. create ruby method
        2. create custom button
        3. etc

File: cfme/tests/automate/custom_button/test_infra_objects.py [hunk 7]
```diff
@@ -339,3 +380,59 @@ def test_custom_button_expression(appliance, request, setup_obj, button_group, e
...
+    initial_count = len(view.browser.selenium.window_handles)
+    main_window = view.browser.selenium.current_window_handle
+    custom_button_group.item_select(button.text)
+
+    wait_for(
+        lambda: len(view.browser.selenium.window_handles) > initial_count,
+        timeout=120,
+        message="Check for window open",
+    )
+    open_url_window = set(view.browser.selenium.window_handles) ^ {main_window}
```
izapolsk: REQUIRED: it works but this is wrong operation.  you use symmetric difference instead of (difference) subtract.

digitronik:
Added  separate PR for failures. As, it was affecting  nearly `46` test in orstriz report. one liner but need to test against all  test in class file. 
https://github.com/ManageIQ/integration_tests/pull/8316

PR #8271, opened 2018-12-14
[RFR] Adds config for polarion docstrings flake8 plugin

mkoura:
Adds config file so the flake8 plugin installed in polarion_docstrings can check the Polarion metadata values.

### DIFF
--- a/conf/polarion_tools.yaml
+++ b/conf/polarion_tools.yaml
@@ -1,5 +1,7 @@
 default_fields:
+    arch: ""
     assignee: ""
+    automation_script: ""
     caseautomation: automated
     casecomponent: "-"
     caseimportance: high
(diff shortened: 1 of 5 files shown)


conf/polarion_tools.yaml
```diff
@@ -0,0 +1,111 @@
...
+    - tags
+    - teardown
+    - testtier
+    - testtype
+    - upstream
+
+docstrings:
+    required_fields:
+        - assignee
+        - initialEstimate
```
mshriver: I believe we want caseimportance to be required as well.

mshriver: Ignore me, default is `high`, we'll leave that for now, can easily change in the future.

PR #8276, opened 2018-12-14
[1LP][RFR] Templates checksum

mmojzis:
Added checking of templates whether they were downloaded correctly.
Fails only when checksum of downloaded image is not equal to checksum from SHA256SUM file.
When SHA256SUM checksum parsing fails then downloaded file is not checked and compared.
Tested.

### DIFF
--- a/cfme/utils/template/base.py
+++ b/cfme/utils/template/base.py
@@ -1,3 +1,4 @@
+import hashlib
 import os
 import re
 from contextlib import closing
@@ -294,6 +295,35 @@ class ProviderTemplateUpload(object):
             except URLError:
                 logger.exception('Failed download of image using urllib')
                 return False
+
+        # Download checksum from url
+        checksum = None
+        try:
+            response = request.urlopen('{}/SHA256SUM'.format(self.image_url))
+            checksums = response.read()
+            for line in checksums.split("\n"):
+                if self.image_name in line:
+                    checksum = line.strip().split()[0]
+        except URLError:
+            logger.exception('Failed download of checksum using urllib')
+        if not checksum:
+            logger.exception('Failed to get checksum of image from url')
+        else:
+            # Get checksum of downloaded file
+            sha256 = hashlib.sha256()
+            image_sha256 = None
+            try:
+                with open(self.image_name, 'rb') as f:
+                    for block in iter(lambda: f.read(65536), b''):
+                        sha256.update(block)
+                image_sha256 = sha256.hexdigest()
+            except Exception:
+                logger.exception('Failed to get checksum of image')
+            if image_sha256 and not checksum == image_sha256:
+                logger.exception('Local image checksum does not match checksum from url')
+                return False
+            else:
+                logger.info('Local image checksum matches checksum from url')
 
         # Unzips image  when suffix is zip or tar.gz and then changes image name to extracted one.
         # For EC2 and SCVMM images is zip used and for GCE is tar.gz used.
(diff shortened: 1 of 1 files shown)


cfme/utils/template/base.py
```diff
@@ -256,6 +257,35 @@ def download_image(self):
                 logger.exception('Failed download of image using urllib')
                 return False
 
+        # Download checksum from url
+        checksum = None
```
izapolsk: REQUIRED: please make it another method. this method is already big but your changes make it huge.

mmojzis: @izapolsk Okay, so this new method should be called from download_image() or called separately from here https://github.com/ManageIQ/integration_tests/blob/master/cfme/utils/template/ec2.py#L69 ?

izapolsk: @mmojzis, it depends on whether you want it to be run for all providers or ec2 only.
I'd rather you called it from run() since it would look more straightforward and clear.

mmojzis: @izapolsk I would like to run it for template of any type.

izapolsk: Although please take note that some templates like openshift ones may not have checksum files

mmojzis: Yes, I can see it: http://file.cloudforms.lab.eng.rdu2.redhat.com/builds/cfme/5.10/stable/SHA256SUM
But actually template upload doesn't stop when checksum wasn't found:
https://github.com/ManageIQ/integration_tests/pull/8276/files#diff-c9ddd76cc9e9ceda813ce94622b76309R270

izapolsk: but I'd rather change log level to warning in such case.

mmojzis: Makes sense

mmojzis: @mshriver What do you think about this? Where to put that method?

mshriver: The image is not downloaded for all provider upload types, so I'm not sure what our checksum validation case would be for when we didn't download the image (and it was directly uploaded).

Assuming this is only validating checksum of locally downloaded images, I think it can be called from `download_image` instead of from `run`.

mmojzis: Created separate method that is called in download_image method. Also changed logger to warn when there is no SHA sum for downloaded image.
This can be tested when devs start building EC2 images.

mmojzis: Forgot it can be tested with 5.10. So tested with ec2 and 5.10. Working well.

cfme/utils/template/base.py
```diff
@@ -277,6 +278,38 @@ def deploy_template(self):
...
+        try:
+            response = request.urlopen('{}/SHA256SUM'.format(self.image_url))
+            checksums = response.read().decode('utf-8')
+            for line in checksums.split("\n"):
+                if self.image_name in line:
+                    checksum = line.strip().split()[0]
+        except URLError:
+            logger.warn('Failed download of checksum using urllib')
+        if not checksum:
+            logger.warn('Failed to get checksum of image from url')
```
mshriver: I think this should `return False` at this point as well, instead of leaving it to return None at the bottom of the function.  At this point the 'step' has failed.

mmojzis: Okay, when it returns False. Does it interrupt the job? Because this is not such a big deal to interrupt template upload.

mshriver: IIRC both False or None will trigger the job 'failing'. I think it should fail if the checksum doesn't match, forcing us to investigate why (we can always re-run)

mshriver: Actually, ignore me. Because this is called by download_image, it won't stop the job. download_image returning False/None will stop the upload, because that step is wrapped.

PR #8278, opened 2018-12-17
[1LP][RFR]Fixed permissions service catalog item

niyazRedhat:
{{pytest: cfme/tests/services/test_catalog_item.py::test_permissions_catalog_item_add --long-running -vvv}}

### DIFF
--- a/cfme/tests/services/test_catalog_item.py
+++ b/cfme/tests/services/test_catalog_item.py
@@ -5,7 +5,7 @@ from selenium.common.exceptions import NoSuchElementException
 
 import cfme.tests.configure.test_access_control as tac
 from cfme import test_requirements
-from cfme.services.catalogs.catalog_items import AllCatalogItemView
+from cfme.services.catalogs.catalog_items import AllCatalogItemView, AddCatalogItemView
 from cfme.base.login import BaseLoggedInPage
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.blockers import BZ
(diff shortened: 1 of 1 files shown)


mshriver:
Is this an RC regression?

cfme/tests/services/test_catalog_item.py
```diff
@@ -147,35 +147,64 @@ def test_edit_tags(catalog_item):
     catalog_item.remove_tag(tag)
 
 
-@pytest.mark.skip('Catalog items are converted to collections. Refactoring is required')
-@pytest.mark.meta(blockers=[BZ(1531512, forced_streams=["5.8", "5.9", "upstream"])])
-def test_catalog_item_duplicate_name(catalog_item):
+@pytest.mark.meta(blockers=[BZ(1531512, forced_streams=["5.9", "5.10"])])
```
digitronik: Question:
Is this issue fixed on `upstream`?

cfme/tests/services/test_catalog_item.py
```diff
@@ -147,35 +147,64 @@ def test_edit_tags(catalog_item):
...
+        appliance.collections.catalog_items.GENERIC,
+        name=cat_item_name,
+        description="my catalog item",
+        display_in=True,
+        catalog=catalog,
+        dialog=dialog
+    )
+    view = cat_item.create_view(AllCatalogItemView)
+    assert view.is_displayed
+    view.flash.assert_success_message('Service Catalog Item "{}" was added'.format(
```
digitronik: Required: Its looks like less than `100` can you please add inline

cfme/tests/services/test_catalog_item.py
niyazRedhat: No, but we don't test our automation on upstream.

mshriver: We certainly do, and all framework and automation components must be written to run against MIQ.

cfme/tests/services/test_catalog_item.py
```diff
@@ -147,35 +147,63 @@ def test_edit_tags(catalog_item):
...
+            name='test_item_{}'.format(fauxfactory.gen_alphanumeric()),
+            description="my catalog item",
+            display_in=True,
+            catalog=catalog,
+            dialog=dialog
+        )
+        request.addfinalizer(lambda: cat_item.delete())
+    test_product_features = [['Everything', 'Services', 'Catalogs Explorer', 'Catalog Items']]
+    test_actions = {'Add Catalog Item': _create_catalog}
+    tac.single_task_permission_test(appliance, test_product_features, test_actions)
```
izapolsk: REQUIRED: procedures shouldn't use test in name.  those can be collected by py.test and treated as a test.
Please rename this. It can be done in another PR. In addition, this isn't good practice to create cross references between tests. Can you move stuff used in several test suites to some common location.

Ah, it seems you try to call another test which in turn calls one more test. Such implementation is frowned upon.
Please rework this stuff.

izapolsk: We can merge your current changes if you want to rework this stuff in another PR. Please let me know about your plans.

niyazRedhat: @izapolsk I will add those changes in another PR.

PR #8283, opened 2018-12-17
[1LP][RFR] Ensure Nuage AE Class contains auth attributes in AE Schema

aljazkosir:
With this commit we are testing nuage automation, more specifically we are testing if variables exist in Nuage class, which is navigated through automate explorer tree (ManageIQ/System/Event/EmsEvent/Nuage).

\cc @miha-plesko 

There was an issue because clicking happens at the middle of an element, and at the middle of an element there was a chevron for expanding tree. This chevron only expanded the tree not actually clicking on it. We made use of the buttons showed on the image to make the explorer larger so chevron is not exactly at the middle of the element.
automate_explorer

### DIFF
--- a/cfme/automate/explorer/__init__.py
+++ b/cfme/automate/explorer/__init__.py
@@ -8,7 +8,7 @@ from cfme.base import Server
 from cfme.base.login import BaseLoggedInPage
 from cfme.base.ui import automate_menu_name
 from cfme.utils.appliance.implementations.ui import navigator, CFMENavigateStep
-from widgetastic_manageiq import Accordion, ManageIQTree
+from widgetastic_manageiq import Accordion, ManageIQTree, Splitter
 
 
 class AutomateExplorerView(BaseLoggedInPage):
@@ -26,6 +26,7 @@ class AutomateExplorerView(BaseLoggedInPage):
     @View.nested
     class datastore(Accordion):  # noqa
         tree = ManageIQTree()
+        splitter = Splitter()
 
     configuration = Dropdown('Configuration')
 
--- a/cfme/tests/networks/nuage/test_automation.py
+++ b/cfme/tests/networks/nuage/test_automation.py
@@ -1,0 +1,14 @@
+def test_auth_attributes_of_nuage_ae_class(appliance):
+    """
+        Ensure Nuage AE Class contains auth attributes in AE Schema
+    """
+    expected_fields = ['nuage_password', 'nuage_username', 'nuage_enterprise', 'nuage_url',
+                       'nuage_api_version']
+    manageiq_domain = appliance.collections.domains.instantiate(name='ManageIQ')
+    system_namespace = manageiq_domain.namespaces.instantiate(name='System')
+    event_namespace = system_namespace.namespaces.instantiate(name='Event')
+    ems_event_namespace = event_namespace.namespaces.instantiate(name='EmsEvent')
+    nuage_class = ems_event_namespace.classes.instantiate(name='Nuage')
+    schema_fields = nuage_class.schema.schema_field_names
+
+    assert all(expected in schema_fields for expected in expected_fields)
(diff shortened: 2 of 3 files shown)


miha-plesko:
@aljazkosir please mark this as `[RFR]` or maybe even `[1LP][RFR]` because it works perfectly in my local env as well and the code is as readable as it can get

cfme/tests/networks/nuage/test_automation.py
```diff
@@ -0,0 +1,11 @@
+def test_authentication(appliance):
```
mshriver: This needs a docblock describing the test, and likely a more descriptive method name. Its not clear to me how this is testing authentication? Looks like a test of the default automation schema fields?

Does a nuage provider need to be configured in order to run this test?  Is it relevant against MIQ and downstream CFME versions?

cfme/automate/explorer/__init__.py
```diff
@@ -28,6 +28,8 @@ class datastore(Accordion):  # noqa
         tree = ManageIQTree()
 
     configuration = Dropdown('Configuration')
+    increase_button = Text(locator='//div[contains(@class, "resize-right")]')
```
mshriver: TODO (optional)

I would like to see configuration widget moved to a `toolbar` nested view, and increase_button/decrease_button moved to a `sidebar` nested view.

Only other use of this locator is in ProvisionView, where its part of the sidebar. This would provide a more consistent view structure.  Does not need to be changed here, as it would require additional changes to anything using `AutomateExplorerView.configuration`

mshriver: **Required**

On second look, this is the `widgetastic_manageiq.Splitter` widget.  This widget should be used here, with calls to `Splitter.pull_right` during navigation. Consider using `Splitter.reset` after navigation, in CFMENavigateStep.resetter() method implementation (or just in CFMENavigateStep.step() ).

If the locator for `Splitter` isn't working here (it just uses the fa-angle-left|right classes) then we should adjust the widget accordingly.

aljazkosir: Hmm, nice, I wasn't aware of this `Splitter`. I pushed new changes.

cfme/tests/networks/nuage/test_automation.py
```diff
@@ -0,0 +1,14 @@
+def test_auth_attributes_of_nuage_ae_class(appliance):
```
mshriver: Great! Thanks @aljazkosir

miha-plesko:
@mshriver thanks for amazing response time and merge. Let me describe the big picture behind this test. 

Nuage provider integration for CloudForms is currently focused on capturing events and responding to them with Embedded Ansible driven Automation - [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1574890); we even have a [blogpost](https://tech.xlab.si/posts/nuage-ansible-in-automate) on this topic, with [video demo](https://www.youtube.com/watch?v=YeXtnjTY67s). There was an interesting issue to overcome upon integration:

#### How to pass Nuage credentials to the playbook?
Playbook needs to have access to the Nuage credentials in order to consume API, but since in CFME 5.10.x.x passing credentials via REST API is not supported yet, we implemented a workaround: user is given an option to specify Nuage credentials via AE Instance attributes.

auth_attribs

The test in this PR is needed to ensure that Nuage AE Class has those 5 authentication attributes present in the built-in AE Domain, please see screenshot. You are right, this specific test does nothing but checks Schema attributes. Zero configuration is needed for test to succeed because we're just testing the built-in AE Schema, you don't even need Nuage provider yet.

However, @aljazkosir is currently implementing a more complex followup test where entire process will be carried out to _actually_ test that those Attributes are accessible by playbooks. That test will require Embedded Ansible role activated and Nuage provider added. The agenda that @aljazkosir is currently implementing is:

- add Embedded Ansible repo with sample playbook
- create empty domain and copy AE Instance from the built-in domain
- create AE Method of type 'playbook' for it with some sample playbook
- trigger event on Nuage
- verify that playbook was triggered and has access to the authentication variables and access to the event payload - by checking the logs

At this point I must thank you guys for making integration tests framework so amazing: as far as we've analyzed, 90% of the fixtures are already implemented and we'll just reuse/copy them 👏 which will make this rather complex test scenario quite easy to implement. 

@mshriver sorry for rather long reply, please let me know if it answers your questions.

PR #8284, opened 2018-12-17
[1LP][RFR] Update validate_chargeback_report cost comparisons

mshriver:
Seeing some cost comparison failures, DRYing the cost comparison functions in chargeback tests.

Ultimately this isn't resolving the failing comparisons between the chargeback reports and estimations. @nachandr is looking into these failures, which are occurring in PRT.

{{ pytest: cfme/tests/intelligence/reports/test_validate_chargeback_report.py --use-provider complete --long-running }}


nachandr:
The change through this PR to DRY the cost calculation and the assertion looks good to me.

PR #8289, opened 2018-12-18
[1LP][RFR][NOTEST] update cfme.utils for basic Notebooks

digitronik:
Purpose or Intent
=================

just change `utils` import to `cfme.utils`. 
Those are basic notebooks for cfme...


digitronik:
@izapolsk yes, I tested basics and added smaller changes in imports only.

PR #8293, opened 2018-12-19
[1LP][RFR] removed generated manual test cases and respectively updated existing test cases

izapolsk:


cfme/tests/pod/test_appliance_crud.py
```diff
@@ -208,7 +208,7 @@ def test_crud_pod_appliance(temp_pod_appliance, provider, setup_provider):
         test_flag: podtesting
 
     Polarion:
-        assignee: None
+        assignee: izapolsk
```
jawatts: missing `caseimportance` on some of these. Also need to fill out initialEstimate

PR #8303, opened 2018-12-20
[1LP][RFR] Moving manual multi-tenancy test

mnadeem92:
Purpose or Intent
=================

### DIFF
--- a/cfme/tests/configure/test_access_control.py
+++ b/cfme/tests/configure/test_access_control.py
@@ -1398,3 +1398,409 @@ def test_superadmin_tenant_admin_crud(appliance, group_collection):
         user.name = "{}_edited".format(user.name)
     user.delete()
     assert not user.exists
+
+
+@pytest.mark.manual
+@pytest.mark.ignore_stream("upstream")
+def test_tenant_unique_catalog():
+    """
+    Catalog name is unique per tenant. Every tenant can have catalog with
+    name "catalog" defined.
+
+    Polarion:
+        assignee: mnadeem
+        casecomponent: config
+        caseposneg: negative
+        initialEstimate: 1/2h
+        startsin: 5.5
+    """
+    pass
+
+
+@pytest.mark.manual
+@pytest.mark.ignore_stream("upstream")
+@pytest.mark.tier(2)
+def test_tenant_visibility_service_template_catalogs_all_parents():
+    """
+    Members of child tenants can see service templates which are visible
+    in parent tenants.
+
+    Polarion:
+        assignee: mnadeem
+        casecomponent: config
+        caseimportance: medium
+        initialEstimate: 1/2h
+        startsin: 5.5
+    """
+    pass
+
+
+@pytest.mark.manual
+@pytest.mark.ignore_stream("upstream")
+@pytest.mark.tier(2)
+def test_tenant_visibility_services_all_childs():
+    """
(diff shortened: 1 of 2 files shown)


cfme/tests/configure/test_multi_tenancy_manual.py
```diff
@@ -0,0 +1,438 @@
+# -*- coding: utf-8 -*-
+# pylint: skip-file
+"""Manual tests"""
+
+import pytest
+
+from cfme import test_requirements
+
+pytestmark = [pytest.mark.ignore_stream('5.9', '5.10', 'upstream')]
```
mshriver: Need to collect against at least one stream

cfme/tests/configure/test_multi_tenancy_manual.py
```diff
@@ -0,0 +1,438 @@
+# -*- coding: utf-8 -*-
+# pylint: skip-file
```
mshriver: Why is this included here? We want to lint all python

cfme/tests/configure/test_multi_tenancy_manual.py
```diff
@@ -0,0 +1,438 @@
...
+import pytest
+
+from cfme import test_requirements
+
+pytestmark = [pytest.mark.ignore_stream('5.9', '5.10', 'upstream')]
+
+
+@pytest.mark.manual
+@test_requirements.cfme_tenancy
+def test_tenant_unique_catalog():
```
mshriver: Lots of opportunity for automation here :)

cfme/tests/configure/test_multi_tenancy_manual.py
mnadeem92: All test applicable to both 5.9 and 5.10, so removing both.

cfme/tests/configure/test_multi_tenancy_manual.py
mnadeem92: I tried to keep the same syntax as set in test_manual.py, 
Thank you for the suggestion, I have removed it now.

cfme/tests/configure/test_multi_tenancy_manual.py
mnadeem92: Yes, automating these test cases is on my TODO list.

cfme/tests/configure/test_multi_tenancy_manual.py
```diff
@@ -0,0 +1,434 @@
+import pytest
+
+from cfme import test_requirements
+
+pytestmark = [pytest.mark.ignore_stream("upstream")]
+
+
+@pytest.mark.manual
```
digitronik: Required:
This file is for `manual tests` so better to have this `global marker`

mnadeem92: Done, along with manual, i have marked @test_requirements.cfme_tenancy as well, Cause the file is dedicated for manual tests related to multi-tenancy only.

izapolsk:
@mnadeem92, If I understand your changes right, you are going to move your tests to another file and automate them one by one.  am I right ?
It will entail renaming file and marking all not automated tests as manual ones.
Please remove manual from your file name then.
Please mark every test case as manual.

izapolsk:
let's wait for PRT and Travis results.

PR #8304, opened 2018-12-20
[1LP][RFR] ipv6 handing in sprout

izapolsk:
REQUIRES: https://github.com/ManageIQ/wrapanapi/pull/350

with updating wrapanapi to 3.0, rhev ip started returning ipv6 until ipv4 was available. This ip isn't accessaible and isn't parsed well in appliance atm.
So, we need to wait for ipv4 and return appliance only then.

### DIFF
--- a/cfme/utils/net.py
+++ b/cfme/utils/net.py
@@ -152,3 +152,39 @@ def is_pingable(ip_addr):
     except Exception as e:
         logger.exception(e)
         return False
+
+
+def is_ipv4(ip_addr):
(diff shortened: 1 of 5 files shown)


cfme/utils/net.py
```diff
@@ -152,3 +152,39 @@ def is_pingable(ip_addr):
     except Exception as e:
         logger.exception(e)
         return False
+
+
+def is_ipv4(ip_addr):
```
mshriver: :tada:

PR #8305, opened 2018-12-20
[1LP][RFR] Refactor cleanup_old_vms

mshriver:
Rework the multiprocess structure in cleanup_old_vms to layer the process/thread use across providers, and to avoid two things:

1. failing to delete vms because of duplicate names
2. skipping deletion of vms with unknown creation dates

I've also removed the prompt, instead opting for a 'dry-run' kind of approach, leaving this as the default via keeping the --force arg.

Unfortunately `Pool.starmap`, is only python 3.

### DIFF
--- a/scripts/cleanup_old_vms.py
+++ b/scripts/cleanup_old_vms.py
@@ -1,12 +1,15 @@
-#!/usr/bin/env python2
+#! /usr/bin/env python
+# uses Pool.starmap, which is py3 only
 import argparse
 import datetime
-from datetime import timedelta
 import re
 import sys
 from collections import namedtuple
-from operator import attrgetter
+from datetime import timedelta
 from multiprocessing import Manager, Pool
+from multiprocessing.pool import ThreadPool
+from operator import attrgetter
+from textwrap import dedent
 
 import pytz
 from tabulate import tabulate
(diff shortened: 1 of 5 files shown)


scripts/cleanup_old_vms.py
```diff
@@ -134,21 +97,54 @@ def scan_provider(provider_key, matchers, match_queue, scan_failure_queue):
...
-        match_queue.put(VmProvider(provider_key, vm.name))
 
-    non_text_matching = set(vm_list) - set(text_matched_vms)
-    logger.info(
-        '%r: NOT matching text filters: %r', provider_key, [vm.name for vm in non_text_matching])
-    logger.info(
-        '%r: MATCHED text filters: %r', provider_key, [vm.name for vm in text_matched_vms])
+    logger.info('%r: NOT matching text filters: %r',
+                provider_key,
+                set([v.name for v in vm_list]) - set([v.name for v in text_matched_vms]))
```
izapolsk: {for in} ?

scripts/cleanup_old_vms.py
```diff
@@ -160,63 +156,57 @@ def scan_vm(provider_key, vm_name, delta, match_queue, scan_failure_queue):
...
         except Exception:  # noqa
-            failure = True
-            logger.exception('%r: Exception getting status for %r', provider_key, vm_name)
+            logger.exception('%r: Exception getting status for %r', provider_key, vm.name)
             status = NULL
-    finally:
-        if failure:
-            scan_failure_queue.put(VmReport(provider_key, vm_name, FAIL, status, NULL))
-            return
+            pass
```
izapolsk: looks like forgotten development bits

mshriver: nah I just always leave `pass` in, but nobody seems to like when I do :)

scripts/cleanup_old_vms.py
mshriver: not sure what you're getting at here - I'm making two lists of VM names, subtracting the matched names from the full name list in order to display what did not match the given text filter.  This is so on output we can see if there's some problem with the text matching.

I use set() to make sure I end up with unique names, since some providers can have duplicate vm names.

PR #8319, opened 2018-12-26
[RFR] using wait_widget_fill_strategy

izapolsk:
There was an issue in some tests when widget was not filled in because it turned up after fill operation.
widgetastic.core got a new fill strategy for views which waited for each new widget to show up.

{{pytest: --long-running --use-template-cache --use-provider=azure -k test_provision_stack cfme/tests/services/test_provision_stack.py}}

REQUIRES: 
https://github.com/RedHatQE/widgetastic.core/pull/133


requirements/frozen.py2.txt
```diff
@@ -317,7 +317,8 @@ wcwidth==0.1.7
 webencodings==0.5.1
 websocket-client==0.40.0
 Werkzeug==0.14.1
-widgetastic.core==0.30.3
+#widgetastic.core==0.30.3
+git+git://github.com/izapolsk/widgetastic.core@wait_fill_widgets#widgetastic.core
```
mshriver: Needs a real version before RFR/mergable ;)

izapolsk:
PRT looks fine now. Failed tests aren't related to my changes.

cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -140,6 +137,17 @@ def is_displayed(self):
         )
 
 
+class AddOrchestrationCatalogItemView(AddCatalogItemView):
+    fill_strategy = WaitFillViewStrategy()
+
+    @property
+    def widget_names(self):
+        return ['title', 'select_catalog', 'select_dialog',
```
mshriver: Please put inline comment here that you're setting widget_names in order to control the fill order of this view, but the widgets are the same as AddCatalogItemView

PR #8323, opened 2018-12-27
[RFR] Test Embedded Ansible integration for Nuage

aljazkosir:
Test that Nuage events trigger Embedded Ansible automation and that playbook has access to authentication attributes and event data.

Specifically, we copy AE Instance 'ManageIQ/System/Event/EmsEvent/Nuage/nuage_enterprise_create' from default domain into our own domain and customize it's 'meth5' attribute to invoke Embedded Ansible playbook which prints authentication attributes and event data into evm.log. This test then triggers a 'nuage_enterprise_create' event and waits for appropriate line to appear in evm.log.

Prerequisites:
Following content needs to be present in cfme_data.yaml in order to fetch correct Ansible repository:

```yaml
# cfme_data.yaml
ansible_links:
  playbook_repositories:
    embedded_ansible: https://github.com/xlab-si/integration-tests-nuage-automation.git
```

Video explaining what the test is doing: 
https://drive.google.com/file/d/1weHHLjQ_0t1OrXECaDDxoGjzjYwwJ18R/view

\cc @miha-plesko

### DIFF
--- a/cfme/automate/explorer/method.py
+++ b/cfme/automate/explorer/method.py
@@ -243,6 +243,7 @@ class MethodAddView(AutomateExplorerView):
     machine_credential = PlaybookBootstrapSelect('provisioning_machine_credential_id')
     hosts = Input('provisioning_inventory')
     max_ttl = Input('provisioning_execution_ttl')
+    logging_output = PlaybookBootstrapSelect('provisioning_log_output')
     escalate_privilege = BootstrapSwitch('provisioning_become_enabled')
     verbosity = PlaybookBootstrapSelect('provisioning_verbosity')
     playbook_input_parameters = PlaybookInputParameters()
@@ -280,6 +281,7 @@ class MethodEditView(AutomateExplorerView):
     machine_credential = PlaybookBootstrapSelect('provisioning_machine_credential_id')
     hosts = Input('provisioning_inventory')
     max_ttl = Input('provisioning_execution_ttl')
+    logging_output = PlaybookBootstrapSelect('provisioning_log_output')
     escalate_privilege = BootstrapSwitch('provisioning_become_enabled')
     verbosity = PlaybookBootstrapSelect('provisioning_verbosity')
     playbook_input_parameters = PlaybookInputParameters()
@@ -312,7 +314,7 @@ class Method(BaseEntity, Copiable):
 
     def __init__(self, collection, name=None, display_name=None, location='inline', script=None,
                  data=None, repository=None, playbook=None, machine_credential=None, hosts=None,
-                 max_ttl=None, escalate_privilege=None, verbosity=None,
+                 max_ttl=None, logging_output=None, escalate_privilege=None, verbosity=None,
                  playbook_input_parameters=None, cancel=False, validate=True, inputs=None):
         super(Method, self).__init__(collection)
 
@@ -327,6 +329,7 @@ class Method(BaseEntity, Copiable):
         self.machine_credential = machine_credential
         self.hosts = hosts
         self.max_ttl = max_ttl
+        self.logging_output = logging_output
         self.escalate_privilege = escalate_privilege
         self.verbosity = verbosity
         self.playbook_input_parameters = playbook_input_parameters
@@ -433,7 +436,7 @@ class MethodCollection(BaseCollection):
     def create(
             self, name=None, display_name=None, location='inline', script=None, data=None,
             cancel=False, validate=True, repository=None, playbook=None, machine_credential=None,
-            hosts=None, max_ttl=None, escalate_privilege=None, verbosity=None,
+            hosts=None, max_ttl=None, logging_output=None, escalate_privilege=None, verbosity=None,
             playbook_input_parameters=None, inputs=None):
         add_page = navigate_to(self, 'Add')
         add_page.fill({'location': location})
@@ -458,6 +461,7 @@ class MethodCollection(BaseCollection):
                 'machine_credential': machine_credential,
                 'hosts': hosts,
                 'max_ttl': max_ttl,
+                'logging_output': logging_output,
                 'escalate_privilege': escalate_privilege,
                 'verbosity': verbosity,
                 'playbook_input_parameters': playbook_input_parameters
@@ -488,6 +492,7 @@ class MethodCollection(BaseCollection):
                 machine_credential=machine_credential,
                 hosts=hosts,
                 max_ttl=max_ttl,
+                logging_output=logging_output,
                 escalate_privilege=escalate_privilege,
                 verbosity=verbosity,
                 playbook_input_parameters=playbook_input_parameters,
--- a/cfme/fixtures/ansible_fixtures.py
+++ b/cfme/fixtures/ansible_fixtures.py
@@ -6,6 +6,7 @@ from cfme.services.myservice import MyService
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.blockers import BZ
 from cfme.utils.conf import cfme_data
+from cfme.utils.log import logger
 from cfme.utils.wait import wait_for
 
 
@@ -29,26 +30,27 @@ def ansible_repository(request, appliance, wait_for_ansible):
         ...
     """
     repositories = appliance.collections.ansible_repositories
-    playbooks = cfme_data.ansible_links.playbook_repositories
     try:
-        playbook = getattr(request, 'param', 'embedded_ansible')
+        playbooks_yaml = cfme_data.ansible_links.playbook_repositories
+        playbook_name = getattr(request, 'param', 'embedded_ansible')
         repository = repositories.create(
             name=fauxfactory.gen_alpha(),
-            url=getattr(playbooks, playbook),
-            description=fauxfactory.gen_alpha())
-    except KeyError:
-        pytest.skip("Skipping since no such key found in yaml")
+            url=getattr(playbooks_yaml, playbook_name),
+            description=fauxfactory.gen_alpha()
+        )
+    except (KeyError, AttributeError):
+        message = "Missing ansible_links content in cfme_data, cannot setup repository"
+        logger.exception(message)  # log the exception for debug of the missing content
+        pytest.skip(message)
     view = navigate_to(repository, "Details")
-    refresh = view.toolbar.refresh.click
     wait_for(
         lambda: view.entities.summary("Properties").get_text_of("Status") == "successful",
         timeout=60,
-        fail_func=refresh
+        fail_func=view.toolbar.refresh.click
     )
     yield repository
 
-    if repository.exists:
-        repository.delete()
+    repository.delete_if_exists()
 
 
 @pytest.fixture(scope="module")
--- a/cfme/fixtures/ansible_fixtures.py
+++ b/cfme/fixtures/ansible_fixtures.py
@@ -18,12 +18,23 @@ def wait_for_ansible(appliance):
 
 
 @pytest.fixture(scope="module")
-def ansible_repository(appliance, wait_for_ansible):
+def ansible_repository(request, appliance, wait_for_ansible):
+    """
+    By default cfme_data.ansible_links.playbook_repositories.embedded_ansible is set for the url,
+    but you can specify it explicitly with @pytest.mark.parametrize decorator on your test function.
+
+    Example:
+    @pytest.mark.parametrize('ansible_repository', ['nuage'], indirect=True)
+    def test_function(ansible_repository):
+        ...
+    """
     repositories = appliance.collections.ansible_repositories
+    playbooks = cfme_data.ansible_links.playbook_repositories
     try:
+        playbook = getattr(request, 'param', 'embedded_ansible')
         repository = repositories.create(
             name=fauxfactory.gen_alpha(),
-            url=cfme_data.ansible_links.playbook_repositories.embedded_ansible,
+            url=getattr(playbooks, playbook),
             description=fauxfactory.gen_alpha())
     except KeyError:
         pytest.skip("Skipping since no such key found in yaml")
--- a/cfme/fixtures/embedded_ansible.py
+++ b/cfme/fixtures/embedded_ansible.py
@@ -1,0 +1,101 @@
+import pytest
+from wrapanapi.utils.random import random_name
+
+
+def instantiate_namespace(domain_name, namespace_levels):
+    namespace = domain_name
+    for namespace_name in namespace_levels:
+        namespace = namespace.namespaces.instantiate(namespace_name)
+    return namespace
+
+
+def copy_ae_instance(appliance, instance_fqdn, domain_name):
+    """
+    Copy AE Instance into a domain. We split 'instance_fqdn' which is a full string path to existing
+    AE instance. Then based on the values we get from split, proper classes are instantiated.
+    AE Instance is copied to provided domain name, then instantiated and returned under new domain.
+    Example:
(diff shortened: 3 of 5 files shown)


cfme/fixtures/embedded_ansible.py
```diff
@@ -0,0 +1,81 @@
...
+    destination_domain = appliance.collections.domains.instantiate(destination_domain_name)
+    destination_namespace = create_namespaces(destination_domain, source)
+    return destination_namespace.classes.instantiate(source_class_name)\
+        .instances.instantiate(source_instance_name)
+
+
+@pytest.fixture
+def domain(request, appliance):
+    """
+    Created new domain with random name, if we want we can also provide a custom name with
```
miha-plesko: Be careful, first line in docstring is special so you shouldn't break it:

```python

"""
Create a new AE Domain.

By default, a random name is generated for the Domain, but you can specify it explicitly with @pytest.mark.parametrize
decorator on your test function.

Example:
@pytest.mark.parametrize('domain', ['TestDomain'], indirect=True)
def test_my_automation(domain):
  ...
"""
```

cfme/fixtures/embedded_ansible.py
```diff
@@ -0,0 +1,81 @@
...
+    @pytest.mark.parametrize test decorator.
+
+    Example: @pytest.mark.parametrize('domain', ['TestDomain'], indirect=True)
+    """
+    dc = appliance.collections.domains
```
miha-plesko: What does `dc` stand for? I think we should just do

```python
domain = appliance.collections.domains.create(...)
```

When possible, I'd suggest we use full names instead short-hand names (`domain` vs `d`)

cfme/fixtures/embedded_ansible.py
```diff
@@ -0,0 +1,81 @@
...
+    d = dc.create(
+        name=getattr(request, 'param', None) or random_name(),
```
miha-plesko:
```python
name=getattr(request, 'param', random_name()),
```

cfme/fixtures/embedded_ansible.py
```diff
@@ -0,0 +1,81 @@
...
+        enabled=True)
+    yield d
+    d.delete()
+
+
+@pytest.fixture
+def copy_ae_instance_to_domain_name(request, appliance):
```
miha-plesko: Since we don't make use of this fixture ATM I'd just remove it. It's simple enough to implement it later if needed.

cfme/fixtures/embedded_ansible.py
```diff
@@ -0,0 +1,81 @@
...
+    [('ManageIQ/System/Event/EmsEvent/Nuage/nuage_enterprise_create', 'TestDomain')], indirect=True)
+
+    :return: copied AE instance with provided existing domain
+    """
+    src, dest = request.param
+    return copy_ae_instance(appliance, src, dest)
+
+
+@pytest.fixture
+def copy_ae_instance_to_domain(request, appliance, domain):
```
miha-plesko: Please rename to `copy_ae_instance_to_new_domain`

cfme/fixtures/embedded_ansible.py
```diff
@@ -0,0 +1,81 @@
...
+        .instances.instantiate(source_instance_name).copy_to(destination_domain_name)
+
+    destination_domain = appliance.collections.domains.instantiate(destination_domain_name)
+    destination_namespace = create_namespaces(destination_domain, source)
+    return destination_namespace.classes.instantiate(source_class_name)\
+        .instances.instantiate(source_instance_name)
+
+
+@pytest.fixture
+def domain(request, appliance):
```
miha-plesko: Please rename to `custom_ae_domain`

cfme/fixtures/embedded_ansible.py
```diff
@@ -0,0 +1,81 @@
...
+    This is going to create new domain and copy AE instance to it. We can also choose name of the
+    new domain by using one more decorator, see domain fixture docstring
+
+    Example:
+
+    @pytest.mark.parametrize('copy_ae_instance_to_domain',
+    ['ManageIQ/System/Event/EmsEvent/Nuage/nuage_enterprise_create'], indirect=True)
+
+    :return: copied AE instance with new domain
+    """
```
miha-plesko: """
    Copy AE Instance to a newly created AE Domain.

    This fixture requires at least Instance fqdn parameter, while newly created domain name
    can also be customized:

    @pytest.mark.parametrize(
      'copy_ae_instance_to_new_domain',
      ['ManageIQ/System/Event/EmsEvent/Nuage/nuage_enterprise_create'],
      indirect=True
    )
    @pytest.mark.parametrize(
      'custom_ae_domain',
      ['CustomDomainName'],
      indirect=True
    )
    def test_my_automation(copy_ae_instance_to_domain):
      ...

    :return: newly created AE Instance
    """

cfme/fixtures/embedded_ansible.py
```diff
@@ -0,0 +1,81 @@
...
+    src = request.param
+    return copy_ae_instance(appliance, src, domain.name)
```
miha-plesko: Nah, just make it one-liner:

```python
return copy_ae_instance(appliance, request.param, custom_ae_domain.name)
```

cfme/tests/networks/nuage/test_automation.py
```diff
@@ -12,3 +22,52 @@ def test_auth_attributes_of_nuage_ae_class(appliance):
...
+    nuage_ae_class = copy_ae_instance_to_domain.klass
+    ae_method = nuage_ae_class.methods.create(name="printout", location='playbook',
+                                              repository=ansible_repository.name,
+                                              playbook='printout.yaml',
+                                              machine_credential="CFME Default Credential",
+                                              logging_output='Always')
+    username = random_name()
+    with update(copy_ae_instance_to_domain):
+        copy_ae_instance_to_domain.fields = {
+            'nuage_username': {'value': username},
```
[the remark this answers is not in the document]
aljazkosir: There is actually a problem with password field, because password field has a different `id` value (`id='inst_password_value'`) inside HTML tag then the rest of the fields (`id='inst_value')`. File `cfme/automate/explorer/instance.py` line 117 and 124.

\cc @miha-plesko

aljazkosir:
I included most of the changes, issue with password mentioned above still exists and parametrizing cfme_data with custom url.

\cc @miha-plesko

dajoRH:
Lint report for commit 895c323fcd161959a127c83075d9ee9129611590:

`cfme/fixtures/embedded_ansible.py`:
- :bangbang: Line 42:101: **E501** *line too long (122 > 100 characters)*

`cfme/tests/networks/nuage/test_automation.py`:
- :heavy_exclamation_mark: Line 91:73: **W292** *no newline at end of file*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

aljazkosir:
I rebased and added parametrization to ansible playbook url.

\cc @miha-plesko

cfme/tests/networks/nuage/test_automation.py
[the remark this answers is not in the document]
miha-plesko: Yaaay. Okay, I see you managed to resolve this by simply not including the password, which works with me 👍

aljazkosir:
@mshriver As requested I edited the pull request with the video explaining what the test is doing.

miha-plesko:
Thanks @aljazkosir I think this PR is now ready to go

cfme/automate/explorer/method.py
```diff
@@ -433,7 +434,7 @@ def tree_path(self):
     def create(
             self, name=None, display_name=None, location='inline', script=None, data=None,
             cancel=False, validate=True, repository=None, playbook=None, machine_credential=None,
-            hosts=None, max_ttl=None, escalate_privilege=None, verbosity=None,
+            hosts=None, max_ttl=None, logging_output=None, escalate_privilege=None, verbosity=None,
```
mshriver: This attribute should be on the `Method` entity, and passed to the instantiate call as well.

cfme/fixtures/ansible_fixtures.py
```diff
@@ -18,12 +18,25 @@ def wait_for_ansible(appliance):
 
 
 @pytest.fixture(scope="module")
-def ansible_repository(appliance, wait_for_ansible):
+def ansible_repository(request, appliance, wait_for_ansible):
+    """
+    By default cfme_data.ansible_links.playbook_repositories.embedded_ansible is set for the url,
```
mshriver: :+1: Great enhancement

cfme/fixtures/embedded_ansible.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def create_namespaces(d, namespace_names):
+    namespace = None
+    for namespace_name in namespace_names:
+        if namespace:
+            namespace = namespace.namespaces.instantiate(namespace_name)
+        else:
+            namespace = d.namespaces.instantiate(namespace_name)
+    return namespace
```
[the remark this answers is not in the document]
mshriver: Yes, I think we need some docblock context here in what the function is actually doing here.

Its odd to me, at a read, why only the last namespace instantiated is returned, when multiple are being instantiated.

cfme/fixtures/embedded_ansible.py
```diff
@@ -0,0 +1,81 @@
...
+    :param domain_name: name of exisitng AE Domain where we want to copy to e.g. 'TestDomain'
+    :return: the new AE Instance
+    """
+    source = instance_fqdn.split('/')
+    source_domain_name = source.pop(0)
+    source_instance_name = source.pop(-1)
+    source_class_name = source.pop(-1)
+    source_domain = appliance.collections.domains.instantiate(source_domain_name)
+    namespace = open_namespace(source_domain, source)
+    namespace.classes.instantiate(source_class_name)\
```
mshriver: Please do not use `\` line continuation, wrap with parenthesis or change syntax of calls.

Also this call chain is not clear - please include a comment in-line in plain english of what resource you're trying to use and copy

cfme/fixtures/embedded_ansible.py
```diff
@@ -0,0 +1,81 @@
...
+def open_namespace(domain_name, namespace_levels):
+    namespace = domain_name
+    for namespace_name in namespace_levels:
+        namespace = namespace.namespaces.instantiate(namespace_name)
+    return namespace
+
+
+def copy_ae_instance(appliance, instance_fqdn, domain_name):
+    """
+    Copy AE Instance into a domain.
```
mshriver: This seems like a complex method to have within the fixtures module, instead of as a method on Instance. 

I see that the `copy_to` method that you're using on the AE Instance, defined here, has a `TODO` for 'Namespace!', which is actually not clear to me what needs to be done there.

https://github.com/ManageIQ/integration_tests/blob/c94723bf2bb7aebd66a1e5e8130fb76f2ff935c2/cfme/automate/explorer/common.py#L30

Can this `Copiable.copy_to` method be expanded to cover your use case here?

Otherwise, can you provide more description of exactly what is going on in this method - there are calls to pop components off of the instance fqdn that aren't clear, and then chained calls to instantiate entities within collections that are not commented and not clear.

cfme/automate/explorer/method.py
mshriver: @aljazkosir This is required - the Entity for this collection doesn't have this attribute. It should be stored with the entity object, otherwise its invisible to the framework outside of this create method, where its passed to the form fill.

cfme/fixtures/embedded_ansible.py
aljazkosir: `Copiable.copy_to` requires instance of an object, but the way that test decoration works, we can't provide AE instance to a fixture. We instead provide a full string path to AE instance, proper classes are then generated inside this function.

I split the call chains into multiple variables so the call is more clear and also provided some explanation and an example.

cfme/fixtures/embedded_ansible.py
```diff
@@ -0,0 +1,99 @@
...
+    'ManageIQ' as source_domain
+    'nuage_enterprise_create' as source AE instance
+    'Nuage' as source AE class
+    the remaining strings left after 'pop' calls are the AE namespaces which are instantiated
+    through 'instantiate_namespace' function
+    AE instance is then copied to a new domain. We also instantiate all the destination classes so
+    we get proper parent-child structure.
+
+
+    :param appliance: appliance
```
mshriver: Thanks so much for adding this docblock content, this is great and lays out your intent very well.

Please use sphynx-napoleon style for these docblocks.
https://cfme-tests.readthedocs.io/en/master/guides/documenting.html?highlight=napoleon

This section should look more like:
```
Args:
    appliance: Appliance object
    instance_fqdn (str): full path <yada>
```

aljazkosir: Updated :)

cfme/tests/networks/nuage/test_automation.py
```diff
@@ -12,3 +22,74 @@ def test_auth_attributes_of_nuage_ae_class(appliance):
...
     assert all(expected in schema_fields for expected in expected_fields)
+
+
+@pytest.mark.parametrize(
+    'copy_ae_instance_to_new_domain',
+    ['ManageIQ/System/Event/EmsEvent/Nuage/nuage_enterprise_create'],
+    indirect=True)
+@pytest.mark.parametrize(
+    'ansible_repository',
+    [cfme_data.ansible_links.playbook_repositories.nuage],
```
mshriver: This breaks collection of tests when this entry isn't present in the cfme_data.yaml that is available.  Don't want to have a direct cfme_data reference here where it can cause an exception.

Consider parametrizing the playbook name (nuage), and have the fixture wrap the lookup in try/except to skip the test when that playbook repo name isn't available.

```
================ ERRORS ================
________ ERROR at setup of cfme/tests/networks/nuage/test_automation.py ___________
cfme/tests/networks/nuage/test_automation.py:33: in <module>
    [cfme_data.ansible_links.playbook_repositories.nuage],
.py2env/lib/python2.7/site-packages/lya/lya.py:68: in __getattr__
    if not (k.startswith('__') or k.startswith('_OrderedDict__')): return self[k]
E   KeyError: 'nuage'
```

aljazkosir: I changed how we pass the playbook into the fixture.

aljazkosir:
@mshriver as requested I?m attaching cfme log for this test:
[cfme.log](https://github.com/ManageIQ/integration_tests/files/3028430/cfme.log)

dajoRH:
Lint report for commit 7e0bab75fca9325e19ad928fda826b17cc8f3952:

`cfme/tests/networks/nuage/test_automation.py`:
- :red_circle: Line 5:1: **F401** *'cfme.utils.conf.cfme_data' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

cfme/fixtures/ansible_fixtures.py
```diff
@@ -18,12 +18,23 @@ def wait_for_ansible(appliance):
...
+    but you can specify it explicitly with @pytest.mark.parametrize decorator on your test function.
+
+    Example:
+    @pytest.mark.parametrize('ansible_repository', ['nuage'], indirect=True)
+    def test_function(ansible_repository):
+        ...
+    """
     repositories = appliance.collections.ansible_repositories
+    playbooks = cfme_data.ansible_links.playbook_repositories
```
mshriver: I'm going to push a commit tweaking how this is referenced.

Its important that this attribute lookup chain is within the try block below, because a cfme_data file without `ansible_links` or `playbook_repositories` maps will cause an exception during the fixture execution.

PR #8324, opened 2018-12-27
[1LP][RFR] fix ansible tests  against pod appliance

izapolsk:
{{pytest: --long-running --use-template-cache --use-provider=cm-pod2 cfme/tests/ansible}}

### DIFF
--- a/cfme/automate/explorer/method.py
+++ b/cfme/automate/explorer/method.py
@@ -301,8 +301,7 @@ class MethodEditView(AutomateExplorerView):
         return (
             self.in_explorer and
             self.datastore.is_opened and
-            self.title.text == 'Editing Automate Method "{}"'.format(
-                self.context['object'].name) and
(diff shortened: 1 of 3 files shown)


cfme/automate/explorer/method.py
```diff
@@ -301,8 +301,7 @@ def is_displayed(self):
         return (
             self.in_explorer and
             self.datastore.is_opened and
-            self.title.text == 'Editing Automate Method "{}"'.format(
-                self.context['object'].name) and
+            'Automate Method [{}'.format(self.context['object'].name) in self.title.text and
```
mshriver: This appears to weaken the is_displayed check, but it appears this view is used in one navigation destination. Although there is no explicit check for an 'edit(ing)' string/element on the page, the Add page title is markedly different. Inclusion of the method name in the title should be sufficient.

PR #8328, opened 2018-12-28
[1LP][RFR] Apply multi-conversion host setup to every rhv host

Yadnyawalkya:
{{pytest: cfme/tests/v2v/test_v2v_migrations.py --use-provider vsphere67-nested --use-provider rhv42 --provider-limit 2 -vvvv -k 'test_single_datastore_single_vm_migration[virtualcenter-6.7-rhevm-4.2-form_data_vm_obj_single_datastore0]'}}

Purpose or Intent
=================

1. __Fixing__ X because it is currently doing Y which is incorrect. It should be doing Z.
2. So changes in UI now gives conversion host warning if host is already been used. This happened multiple times and thus caused many timeouts in V2V. With changes we are setting all RHV hosts as a conversion host so it will not wait for particular host and proceeds without mentioned warning,

### DIFF
--- a/cfme/fixtures/v2v.py
+++ b/cfme/fixtures/v2v.py
@@ -73,7 +73,6 @@ def host_creds(request, v2v_providers):
             host.update_credentials_rest(credentials=host_data['credentials'])
 
         rhv_hosts = rhv_provider.hosts.all()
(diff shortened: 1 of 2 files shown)


Yadnyawalkya:
@digitronik 5.10 PRT result is just a fluke, test should also work for 5.10 like it does on 5.9. Tested locally, it is working on 5.10. btw it has nothing to deal with migration itself and changes are just gonna affects initial setup process.

PR #8329, opened 2018-12-31
Fix: One more check for customization view related to button

digitronik:
Purpose or Intent
=================
- Fix : 
```
TimedOutError
Could not do Waiting for view [ButtonGroupObjectTypeView]
```
- Explanation:
Button group add navigation prerequisite is
`prerequisite = NavigateToAttribute('appliance.server', 'AutomateCustomization')`

`AutomateCustomization` view is not proper... even automation stuck in `Add new button page`. this view `is_displayed  satisfy`. so better to check button is dimmed or not. 

{{pytest: cfme/tests/automate/custom_button/test_infra_objects.py::test_custom_button_automate}}

### DIFF
--- a/cfme/automate/__init__.py
+++ b/cfme/automate/__init__.py
@@ -21,7 +21,7 @@ class AutomateCustomizationView(BaseLoggedInPage):
 
     @property
     def is_displayed(self):
-        return self.in_customization
+        return self.in_customization and not self.buttons.is_dimmed
 
     @View.nested
     class provisioning_dialogs(Accordion):  # noqa


izapolsk:
@digitronik, I tried to implement this via resetter but it brought some complexity. Then I replaced it with am_i_here for AutomationCustomization where it  checked that there were no dimmed accordions. And finally I noticed that all accordions are dimmed in such case and it is enough to just check state of any of them.
So, your fix should be enough in such case.

cfme/automate/__init__.py
```diff
@@ -21,7 +21,7 @@ def in_customization(self):
 
     @property
     def is_displayed(self):
-        return self.in_customization
+        return self.in_customization and not self.buttons.is_dimmed
```
izapolsk: please add a comment explaining this change

PR #8330, opened 2018-12-31
[1LP][RFR] Adding a new test to test required dialog element in ssui order page

niyazRedhat:
{{ pytest: cfme/tests/ssui/test_ssui_service_dialog.py::test_tag_dialog_catalog_item_ssui -v }}

### DIFF
--- a/cfme/tests/ssui/test_ssui_service_dialog.py
+++ b/cfme/tests/ssui/test_ssui_service_dialog.py
@@ -1,0 +1,76 @@
+# -*- coding: utf-8 -*-
+import fauxfactory
+import pytest
+
+from cfme import test_requirements
+from cfme.services.service_catalogs import ServiceCatalogs
+from cfme.utils.appliance import ViaSSUI
+from cfme.utils.appliance.implementations.ssui import navigate_to
+
+
+@pytest.fixture(scope="function")
+def tagcontrol_dialog(appliance):
+    service_dialog = appliance.collections.service_dialogs
+    dialog = "dialog_{}".format(fauxfactory.gen_alphanumeric())
+    element_data = {
+        'element_information': {
+            'ele_label': "Service Level",
+            'ele_name': "service_level",
+            'ele_desc': "service_level_desc",
+            'choose_type': "Tag Control",
+        },
+        'options': {
+            'field_category': "Service Level",
+            'field_required': True
+        }
+    }
+    sd = service_dialog.create(label=dialog, description="my dialog")
+    tab = sd.tabs.create(tab_label='tab_{}'.format(fauxfactory.gen_alphanumeric()),
+                         tab_desc="my tab desc")
+    box = tab.boxes.create(box_label='box_{}'.format(fauxfactory.gen_alphanumeric()),
+                           box_desc="my box desc")
+    box.elements.create(element_data=[element_data])
+    yield sd
+    sd.delete_if_exists()
+
+
+@pytest.fixture(scope="function")
+def catalog(appliance):
+    catalog = "cat_{}".format(fauxfactory.gen_alphanumeric())
+    cat = appliance.collections.catalogs.create(name=catalog, description="my catalog")
+    yield cat
+    if cat.exists:
+        cat.delete()
+
+
+@pytest.fixture(scope="function")
+def catalog_item(appliance, tagcontrol_dialog, catalog):
+    catalog_item = appliance.collections.catalog_items.create(
+        appliance.collections.catalog_items.GENERIC,
+        name=fauxfactory.gen_alphanumeric(),
+        description="my catalog", display_in=True, catalog=catalog,
+        dialog=tagcontrol_dialog)
+    yield catalog_item
+    if catalog_item.exists:
+        catalog_item.delete()
+
+
+@test_requirements.ssui
+def test_tag_dialog_catalog_item_ssui(appliance, catalog_item):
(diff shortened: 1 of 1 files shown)


cfme/tests/ssui/test_ssui_service_dialog.py
```diff
@@ -0,0 +1,70 @@
...
+            'field_required': True
+        }
+    }
+    sd = service_dialog.create(label=dialog, description="my dialog")
+    tab = sd.tabs.create(tab_label='tab_{}'.format(fauxfactory.gen_alphanumeric()),
+        tab_desc="my tab desc")
+    box = tab.boxes.create(box_label='box_{}'.format(fauxfactory.gen_alphanumeric()),
+        box_desc="my box desc")
+    box.elements.create(element_data=[element_data])
+    yield sd
```
digitronik: Question:
Please add proper  `teardown`. whatever we created for setup always we destroy at end of test.

cfme/tests/ssui/test_ssui_service_dialog.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+@pytest.fixture(scope="function")
+def catalog(appliance):
+    catalog = "cat_" + fauxfactory.gen_alphanumeric()
```
digitronik: Required:
User `.format` for string formatting.

cfme/tests/ssui/test_ssui_service_dialog.py
```diff
@@ -0,0 +1,70 @@
...
+    cat = appliance.collections.catalogs.create(name=catalog, description="my catalog")
+    yield cat
+
+
+@pytest.fixture(scope="function")
+def catalog_item(appliance, tagcontrol_dialog, catalog):
+    item_name = fauxfactory.gen_alphanumeric()
```
digitronik: Optional:
If you are not using `name` multiple time then better to use directly rather than variable declaration.

cfme/tests/ssui/test_ssui_service_dialog.py
```diff
@@ -0,0 +1,70 @@
...
+    catalog_item = appliance.collections.catalog_items.create(
+        appliance.collections.catalog_items.GENERIC,
+        name=item_name,
+        description="my catalog", display_in=True, catalog=catalog,
+        dialog=tagcontrol_dialog)
+    return catalog_item
+
+
+@pytest.mark.parametrize('context', [ViaSSUI])
+def test_tagdialog_catalog_item_ssui(appliance, context, catalog_item):
```
digitronik: Your context already pointing its ssui test. I would give better name. I would add `_` in between `tag` and `dialog`

digitronik: Question:
If its part of `ssui`. As per my understanding its specific for `ssui` then why you are parametrizing? Second though are you going to add `ViaUI` as well?

niyazRedhat: @digitronik yes this is ssui test for that we have to add context for ssui login.In this test I will test only ViaSSUI.

niyazRedhat: Made changes

cfme/tests/ssui/test_ssui_service_dialog.py
```diff
@@ -0,0 +1,71 @@
...
+        }
+    }
+    sd = service_dialog.create(label=dialog, description="my dialog")
+    tab = sd.tabs.create(tab_label='tab_{}'.format(fauxfactory.gen_alphanumeric()),
+        tab_desc="my tab desc")
+    box = tab.boxes.create(box_label='box_{}'.format(fauxfactory.gen_alphanumeric()),
+        box_desc="my box desc")
+    box.elements.create(element_data=[element_data])
+    yield sd
+    request.addfinalizer(sd.delete_if_exists)
```
digitronik: Required:
when we use `yield` then we wont need to  add `Finalizer`. At `teardown`; part after `yield` going to call.

niyazRedhat: yes right

cfme/tests/ssui/test_ssui_service_dialog.py
```diff
@@ -0,0 +1,74 @@
...
+    catalog_item = appliance.collections.catalog_items.create(
+        appliance.collections.catalog_items.GENERIC,
+        name=fauxfactory.gen_alphanumeric(),
+        description="my catalog", display_in=True, catalog=catalog,
+        dialog=tagcontrol_dialog)
+    yield catalog_item
+    if catalog_item.exists:
+        catalog_item.delete()
+
+
```
mshriver: Please mark with a requirement

cfme/tests/ssui/test_ssui_service_dialog.py
mshriver: This was marked resolved, but I'm not seeing any cleanup within the fixture here or within the test function? Am I missing something or was it missed?

mshriver: Wait it looks like I missed a line? or my diff updated? anyway, ignore me.

cfme/tests/ssui/test_ssui_service_dialog.py
mshriver: Adding this so I can merge as this looks good otherwise.

PR #8331, opened 2019-01-03
[WIPTEST] added navigation for cloud tenant

lina-nikiforova:
{{ pytest: -v cfme/tests/cloud_infra_common/test_tag_objects.py::test_tag_cloud_objects }}

I added navigation for cloud tenants to fix the tests
```
test_tag_cloud_objects[cloud_tenants-cloud-details]
test_tag_cloud_objects[cloud_tenants-cloud-list]
```


cfme/cloud/provider/__init__.py
```diff
@@ -19,7 +19,7 @@
 from cfme.utils.log import logger
 from cfme.utils.pretty import Pretty
 from cfme.utils.wait import wait_for
-from widgetastic_manageiq import TimelinesView, ItemsToolBarViewSelector
+from widgetastic_manageiq import TimelinesView, ItemsToolBarViewSelector, Table
```
digitronik: Please, do sorting...

cfme/cloud/provider/__init__.py
```diff
@@ -75,6 +75,23 @@ def is_displayed(self):
     including_entities = View.include(VMEntities, use_parent=True)
 
 
+class CloudProviderTenantsView(BaseLoggedInPage):
+    """
+    The collection page for cloud tenants
+    """
```
digitronik: Required:
Inline comment please. I think `docstring` not matching with intention means class name.

cfme/cloud/provider/__init__.py
```diff
@@ -75,6 +75,23 @@ def is_displayed(self):
...
+    @property
+    def is_displayed(self):
+        return (
+            self.breadcrumb.locations[0] == 'Cloud Providers' and
```
digitronik: Optional:
I would go with `self.breadcrumb.active_location`. It will be unique check. If we go with `locations[0]` there may be a chance that other view can satisfy this condition means `Datastore` or `Other`...

cfme/cloud/provider/__init__.py
digitronik: I think you might have look over this view. I am suspect this view already available.
https://github.com/ManageIQ/integration_tests/blob/674b36716e2cd47632ff5ca0ac0b13d9467da2fe/cfme/cloud/tenant.py#L105

lina-nikiforova: @digitronik right...

PR #8333, opened 2019-01-04
[1LP][RFR] Move all yaml.load calls to yaml.safe_load

mshriver:
4.1 release was reverted, 4.2 is not yet available, 3.13 flagged as high severity because of code execution through load calls

Should not have any impact on the modified calls.

### DIFF
--- a/cfme/tests/cloud/test_cloud_events.py
+++ b/cfme/tests/cloud/test_cloud_events.py
@@ -34,7 +34,7 @@ def test_manage_nsg_group(appliance, provider, register_event):
 
     def add_cmp(_, y):
         # In 5.9 version `y` is a dict, not a yaml stream.
-        data = yaml.load(y) if appliance.version < '5.9' else y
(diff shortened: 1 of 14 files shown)


cfme/tests/cloud/test_cloud_events.py
```diff
@@ -90,7 +90,7 @@ def test_vm_capture(appliance, request, provider, register_event):
 
     def cmp_function(_, y):
         # In 5.9 version `y` is a dict, not a yaml stream.
-        data = yaml.load(y) if appliance.version < '5.9' else y
+        data = yaml.safe_load(y) if appliance.version < '5.9' else y
```
jawatts: **Optional** don't think we need the < 5.9 here, but probably out of scope for this PR.

mshriver: Yeah, just gonna leave this one in for now.

PR #8344, opened 2019-01-09
[1LP][RFR] updated BZ for templates power control and assertion for template grid

lina-nikiforova:
1. Removed `if` for `5.9` in first assertion as in `5.9.7.0` Power dropdown is not displayed (not just disabled)
2. The test fails at second assertion, that's due to BZ1634713, so I updated the blocker.

I didn't add the PRT as the test is skipped.

### DIFF
--- a/cfme/tests/infrastructure/test_vm_power_control.py
+++ b/cfme/tests/infrastructure/test_vm_power_control.py
@@ -367,9 +367,10 @@ class TestVmDetailsPowerControlPerProvider(object):
 
 
 @pytest.mark.rhv3
-@pytest.mark.meta(blockers=[BZ(1496383, forced_streams=['5.9', '5.10', 'upstream'])])
+@pytest.mark.meta(blockers=[BZ(1634713, forced_streams=['5.9', '5.10', 'upstream'])])
 def test_no_template_power_control(provider, soft_assert):
     """ Ensures that no power button is displayed for templates.
+    See BZ1496383 and BZ1634713
 
     Prerequisities:
         * An infra provider that has some templates.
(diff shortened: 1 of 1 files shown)


cfme/tests/infrastructure/test_vm_power_control.py
```diff
@@ -391,10 +391,7 @@ def test_no_template_power_control(provider, soft_assert):
     """
     view = navigate_to(provider, 'ProviderTemplates')
     view.toolbar.view_selector.select('Grid View')
-    if provider.appliance.version < '5.10':
-        soft_assert(not view.toolbar.power.is_enabled, "Power enabled in template grid view!")
-    else:
-        soft_assert(not view.toolbar.power.is_displayed, "Power displayed in template grid view!")
+    soft_assert(not view.toolbar.power.is_displayed, "Power displayed in template grid view!")
```
ganeshhubale: @lina-nikiforova 
According to me, as we are asserting for once in this test case. So it would be better approach if we use ordinary ```assert``` instead of ```soft_assert```

lina-nikiforova: @ganeshhubale  there're 2 more soft asserts in this test, e.g. https://github.com/ManageIQ/integration_tests/pull/8344/files#diff-cbdf78293430cdea45837cb22c5774a2R406

ganeshhubale: My bad! Thanks for explaination. :)

cfme/tests/infrastructure/test_vm_power_control.py
```diff
@@ -367,7 +367,7 @@ def test_start_from_suspend(
 
 
 @pytest.mark.rhv3
-@pytest.mark.meta(blockers=[BZ(1496383, forced_streams=['5.9', '5.10', 'upstream'])])
```
digitronik: @lina-nikiforova I went  though BZ. which is interesting; It will be great, if you add it in `doc-string` for reference.

PR #8348, opened 2019-01-09
[1LP]RFR] Add new tests Service Dialog Elements with regex validation in catalog item

niyazRedhat:
{{ pytest: cfme/tests/services/test_dialog_regex_validation_in_catalog.py::test_dialog_element_regex_validation -v }}

### DIFF
--- a/cfme/tests/services/test_dialog_regex_validation_in_catalog.py
+++ b/cfme/tests/services/test_dialog_regex_validation_in_catalog.py
@@ -1,0 +1,68 @@
+# -*- coding: utf-8 -*-
+import fauxfactory
+import pytest
+
+from wait_for import wait_for
+
+from cfme import test_requirements
+from cfme.services.service_catalogs import ServiceCatalogs
+from cfme.utils.appliance.implementations.ui import navigate_to
+
+
+@pytest.fixture(scope="function")
+def dialog_cat_item(appliance, catalog):
+    service_dialog = appliance.collections.service_dialogs
+    dialog = "dialog_{}".format(fauxfactory.gen_alphanumeric())
+    ele_name = fauxfactory.gen_alphanumeric()
+    element_data = {
+        'element_information': {
+            'ele_label': "ele_{}".format(fauxfactory.gen_alphanumeric()),
+            'ele_name': ele_name,
+            'ele_desc': fauxfactory.gen_alphanumeric(),
+            'choose_type': "Text Box"
+        },
+        'options': {
+            'validation_switch': True,
+            'validation': "^([a-z0-9]+_*)*[a-z0-9]+$"
+            }
+        }
+    if appliance.version < '5.10':
+        element_data["options"].pop("validation_switch", None)
+    sd = service_dialog.create(label=dialog, description="my dialog")
+    tab = sd.tabs.create(tab_label="tab_{}".format(fauxfactory.gen_alphanumeric()),
+                         tab_desc="my tab desc")
+    box = tab.boxes.create(box_label="box_{}".format(fauxfactory.gen_alphanumeric()),
+                           box_desc="my box desc")
+    box.elements.create(element_data=[element_data])
+
+    catalog_item = appliance.collections.catalog_items.create(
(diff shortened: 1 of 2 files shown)


cfme/tests/services/test_dialog_regex_validation_in_catalog.py
```diff
@@ -0,0 +1,71 @@
...
+    box = tab.boxes.create(box_label="box_{}".format(fauxfactory.gen_alphanumeric()),
+                           box_desc="my box desc")
+    box.elements.create(element_data=[element_data])
+    request.addfinalizer(sd.delete_if_exists)
+
+    item_name = fauxfactory.gen_alphanumeric()
+    catalog_item = appliance.collections.catalog_items.create(
+        appliance.collections.catalog_items.GENERIC,
+        name=item_name,
+        description="my catalog",
```
ganeshhubale: @niyazRedhat 
According to me, instead of creating extra variable ```item_name``` just use ```name=fauxfactory.gen_alphanumeric()```. Although this variable is used once.

cfme/tests/services/test_dialog_regex_validation_in_catalog.py
```diff
@@ -0,0 +1,70 @@
...
+        element_data["options"].pop("validation_switch", None)
+    sd = service_dialog.create(label=dialog, description="my dialog")
+    tab = sd.tabs.create(tab_label="tab_{}".format(fauxfactory.gen_alphanumeric()),
+                         tab_desc="my tab desc")
+    box = tab.boxes.create(box_label="box_{}".format(fauxfactory.gen_alphanumeric()),
+                           box_desc="my box desc")
+    box.elements.create(element_data=[element_data])
+    request.addfinalizer(sd.delete_if_exists)
+
+    catalog_item = appliance.collections.catalog_items.create(
```
sshveta: Required :
We should create a fixture for generic catalog item as it's used in multiple test cases .

cfme/tests/services/test_dialog_regex_validation_in_catalog.py
```diff
@@ -0,0 +1,70 @@
...
+    Polarion:
+        assignee: nansari
+        caseimportance: high
+        initialEstimate: 1/16h
+    """
+    catalog_item, ele_name = dialog_cat_item
+    service_catalogs = ServiceCatalogs(appliance, catalog_item.catalog, catalog_item.name)
+    view = navigate_to(service_catalogs, 'Order')
+    view.fields(ele_name).fill("!@#%&")
+    wait_for(lambda: view.submit_button.disabled, timeout=7)
```
sshveta: why are we waiting here ? 
We can just assert if submit button is disabled or not .

cfme/tests/services/test_dialog_regex_validation_in_catalog.py
```diff
@@ -0,0 +1,70 @@
...
+    view.fields(ele_name).fill("test_123")
+    wait_for(lambda: not view.submit_button.disabled, timeout=7)
```
sshveta: Same as above .
Please assert.

cfme/tests/services/test_dialog_regex_validation_in_catalog.py
ganeshhubale: @sshveta I worked with @niyazRedhat on this issue. 
- Here what happens is ```view.fields(ele_name).fill("!@#%&")```(filling value in text) and checking ```view.submit_button.disabled```(whether button is disabled or not) works simultaneously. Thus test case fails to check button is disabled or not.
- So here we need to wait for sometime to see button is disabled.

sshveta: :+1:

cfme/tests/services/test_dialog_regex_validation_in_catalog.py
niyazRedhat: Here I am using custom dialog in generic catalog item.This is not common scenario.
This is  good idea for global fixture for generic catalog item. will add fixture in future.

PR #8351, opened 2019-01-10
[RFR] Add fix to delete vm templates created during rest tests

valaparthvi:
1. Delete the templates that were created during test runs.
Test case responsible: [test_rest_templates.py](https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/infrastructure/test_rest_templates.py#L30)

Teardown code for `template` fixture did not work well, causing failed template deletion. Also, using API to delete template only deletes it from the appliance inventory and not from the actual provider. Using wrapanapi instead, to delete the template from provider.

{{ pytest: cfme/tests/infrastructure/test_rest_templates.py::test_set_ownership --use-template-cache -sqvvv }}

### DIFF
--- a/cfme/tests/infrastructure/test_rest_templates.py
+++ b/cfme/tests/infrastructure/test_rest_templates.py
@@ -1,12 +1,15 @@
 # -*- coding: utf-8 -*-
 import pytest
 
+from wrapanapi.exceptions import NotFoundError, MultipleItemsError
+
 from cfme import test_requirements
 from cfme.infrastructure.provider import InfraProvider
 from cfme.markers.env_markers.provider import ONE
 from cfme.rest.gen_data import mark_vm_as_template
 from cfme.rest.gen_data import vm as _vm
 from cfme.utils.blockers import BZ
+from cfme.utils.log import logger
 from cfme.utils.rest import (
     assert_response,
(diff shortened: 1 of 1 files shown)


cfme/rest/gen_data.py
```diff
@@ -487,6 +487,15 @@ def mark_vm_as_template(appliance, provider, vm_name):
     return appliance.rest_api.collections.templates.get(name=vm_name)
 
 
+def delete_templates_from_provider(provider, *template_names):
+    """Helper function to delete templates from provider with the help of wrapanapi.
+    """
+    for template in provider.mgmt.list_templates():
+        if template.name in template_names:
+            template.delete()
+    return True
```
john-dupuy: Why is `True` returned here?

mshriver: This is a costly way of doing this operation, since providers could have thousands of templates.  The `wrapanapi` `get_template` method should be used for this.

Second, I would argue against the inclusion of this function at all, since this is something that can be expressed concisely in-line in list comprehension.

valaparthvi: @mshriver I didn't realize we had a `get_template` method, guess I must've missed it somehow! And if this the case, we definitely do not need an external function for deleting templates.

PR #8354, opened 2019-01-11
[RFR] removing keep_alive = False workaround

izapolsk:
it seems keepalive works right. So, this workaround should be removed if extended testing proves no errors

REQUIRES: 
https://github.com/RedHatQE/cfme_dockerfiles/pull/8
https://github.com/RedHatQE/cfme_dockerfiles/pull/9


mshriver:
Dockerfiles have been merged, wharf servers being updated to account for this. 

Local testers will have to update their containers in order to get the keepalive proxy.

PR #8356, opened 2019-01-11
[1LP][RFR] new test_host_refresh_relationships

lina-nikiforova:
{{ pytest: -v cfme/tests/cloud_infra_common/test_relationships.py::test_host_refresh_relationships }}

I created this test in response to https://bugzilla.redhat.com/show_bug.cgi?id=1658240

### DIFF
--- a/cfme/tests/cloud_infra_common/test_relationships.py
+++ b/cfme/tests/cloud_infra_common/test_relationships.py
@@ -307,6 +307,28 @@ def test_provider_refresh_relationship(provider, setup_provider):
     wait_for_relationship_refresh(provider)
 
 
+@pytest.mark.rhv2
+@pytest.mark.provider([InfraProvider])
+def test_host_refresh_relationships(provider, setup_provider):
+    """ Test that host refresh doesn't fail
+
+    https://bugzilla.redhat.com/show_bug.cgi?id=1658240
+
+    Polarion:
+        assignee: ghubale
+        casecomponent: infra
+        caseimportance: high
+        initialEstimate: 1/8h
+        testSteps:
+            1. Go to a host summary page in cfme
+            2. From configuration -> select "Refresh Relationships and Power State"
+            3. No error, host inventory properly refreshes
+    """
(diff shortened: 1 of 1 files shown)


cfme/tests/cloud_infra_common/test_relationships.py
```diff
@@ -307,6 +307,28 @@ def test_provider_refresh_relationship(provider, setup_provider):
...
+        casecomponent: infra
+        caseimportance: high
+        initialEstimate: 1/8h
+        testSteps:
+            1. Go to a host summary page in cfme
+            2. From configuration -> select "Refresh Relationships and Power State"
+            3. No error, host inventory properly refreshes
+    """
+    host = provider.hosts.all()[0]
+    host.refresh(cancel=True)
```
john-dupuy: Why does `cancel=True` here? I can see that in `cfme.infrastructure.host` the `refresh` method takes `cancel` as the `handle_alert` argument that is passed to `item_select`. Why not call it `handle_alert` if that is the case?

digitronik: @john-dupuy she is handling `alert` by passing argument `True` as me model method. I would like to understand meaning of your query. are you pointing to `cancel` keyword?

cfme/tests/cloud_infra_common/test_relationships.py
```diff
@@ -307,6 +307,28 @@ def test_provider_refresh_relationship(provider, setup_provider):
     wait_for_relationship_refresh(provider)
 
 
+@pytest.mark.rhv2
```
digitronik: This tag is for `rhv` team as per my understating... but awesome :+1:

cfme/tests/cloud_infra_common/test_relationships.py
john-dupuy: @digitronik yes, I am just confused as to why that keyword is called `cancel`? since it is not cancelling the refresh

PR #8358, opened 2019-01-11
[1LP][RFR] Fixed is_displayed of MethodEditView

ganeshhubale:
{{ pytest: cfme/tests/automate -k 'test_method_crud or test_domain_lock_unlock or test_domain_cannot_delete_builtin or test_domain_cannot_edit_builtin or test_priority or test_automate_method_inputs_crud' -v }}

Fixed Error:
```
>           raise TimedOutError("Could not do {} at {}:{} in time".format(message, filename, line_no))
E           TimedOutError: Could not do Waiting for view [MethodEditView] to display at /var/ci/workspace/downstream-59z-tests-master/integration_tests/cfme/utils/appliance/implementations/ui.py:562 in time

../cfme_venv/lib/python2.7/site-packages/wait_for/__init__.py:180: TimedOutError
TimedOutError
Could not do Waiting for view [MethodEditView] to display at /var/ci/workspace/downstream-59z-tests-master/integration_tests/cfme/utils/appliance/implementations/ui.py:562 in time
```

This PR fixes:

 - Fixed is_displyed of MethodEditView
 - Removed "< 5.7" dependencies
 - Assigned tests to 'ghubale'
 - Updated xpath of add_field for 5.10

### DIFF
--- a/cfme/automate/explorer/__init__.py
+++ b/cfme/automate/explorer/__init__.py
@@ -41,6 +41,10 @@ class AutomateExplorer(CFMENavigateStep):
 
 
 def check_tree_path(actual, desired, partial=False):
+    # keyword argument - 'partial'  is introduced here because of unexpected behaviour of automate
+    # tree accordions. In this case, actual and desired trees will not be same which creates problem
+    # For more details please refer:
+    # https://github.com/ManageIQ/integration_tests/pull/8358#discussion_r247770688
     if len(actual) != len(desired) and not partial:
         return False
     # We don't care about icons because we also match titles, which give the type away
--- a/cfme/automate/explorer/__init__.py
+++ b/cfme/automate/explorer/__init__.py
@@ -40,8 +40,8 @@ class AutomateExplorer(CFMENavigateStep):
         self.view.navigation.select(*automate_menu_name(self.obj.appliance) + ['Explorer'])
 
 
-def check_tree_path(actual, desired):
-    if len(actual) != len(desired):
+def check_tree_path(actual, desired, partial=False):
+    if len(actual) != len(desired) and not partial:
         return False
     # We don't care about icons because we also match titles, which give the type away
     for actual_item, desired_item in zip(actual, without_icons(desired)):
--- a/cfme/tests/automate/test_method.py
+++ b/cfme/tests/automate/test_method.py
@@ -40,7 +40,7 @@ def klass(request, namespace):
 
 @pytest.mark.sauce
 @pytest.mark.tier(2)
-def test_method_crud(klass):
+def test_method_crud(appliance, klass):
     """
     Polarion:
         assignee: ghubale
@@ -61,6 +61,9 @@ def test_method_crud(klass):
         method.name = fauxfactory.gen_alphanumeric(8)
         method.script = "bar"
     assert method.exists
+    navigate_to(appliance.server, 'Dashboard')
+    # Navigation to 'Dashboard' is required to navigate to details page of method.
+    # So that 'Edit this Method' option from configuration will be selected from details page.
     with update(method):
         method.name = origname
     assert method.exists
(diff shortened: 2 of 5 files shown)


cfme/tests/automate/test_method.py
```diff
@@ -60,6 +60,9 @@ def test_method_crud(klass):
         method.name = fauxfactory.gen_alphanumeric(8)
         method.script = "bar"
     assert method.exists
+    navigate_to(appliance.server, 'Dashboard')
```
izapolsk: doesn't this mean that is_displayed is incorrect for this page ?

mshriver: @ganeshhubale thanks for the thorough descriptions of your findings, that helps a lot!

I think that this is ultimately just a workaround for the behavior you are seeing, you haven't fixed anything in the framework, just addressed it in this one test. If any other test edits a method twice, it will run into the same behavior.

I think the real problem here lies with both the implementation of `Method.update` (snippet below) and the implementation of `MethodEditView.is_displayed`
```python
if changed:
            view.save_button.click()
        else:
            view.cancel_button.click()
        view = self.create_view(MethodDetailsView, override=updates)
        view.wait_displayed()
```

Here we are assuming the view that MIQ lands on after the save/cancel button is clicked. This assumption is wrong and needs to be corrected.

Second assumption we're making is in `Method.is_displayed`, where we assume how the page was navigated to by checking the accordion tree.  MIQ does not behave that way, and will not always have the method selected in the accordion.  It could be argued that this is actually an MIQ bug, and we could write a BZ that would likely be treated as an RFE.

Thoughts:
1. The flash message assertions in `Method.update` should not be there, they should be in the `Method` crud test itself. (tangentially related)
2. `Method.update` should only use the view to call `view.flash_messages.assert_no_error`. It should not actually assert the view is displayed, or it should allow for one or the other to be displayed. The flash message assertions could be completed by creating a view for `BaseLoggedInPage`
3. There are other Entities and Views in the framework that handle this by having multiple navigation methods - 'EditFromDetails' kind of destinations, that specify the navigation path - via collection, via dashboard, via details page, etc. This is basically the source of both problems - assumptions on both ends of how the page was navigated to.

So I think we would require two fixes here, and have one optional enhancement:
Fix1: Update is_displayed to not rely on the accordion, use other page elements, title and partial accordion may be enough
Fix2: Update `Method.update` to just use `BaseLoggedInPage` for flash message assertion, don't assert the view itself is displayed. Alternatively check for any of the expected views being displayed
Optional1: Move flash message string checks to `test_method_crud`, only `assert_no_error` in `update` itself.

dajoRH:
*CFME QE Bot*

cfme/tests/automate/test_method.py
```diff
@@ -55,6 +56,8 @@ def test_method_crud(klass):
         location='inline',
         script='$evm.log(:info, ":P")',
     )
+    view = appliance.browser.create_view(ClassDetailsView)
```
digitronik: As per my understanding only to create view you inherited `appliance`. As per me any entity object capable to access `create_view`. Please try this may be it will work `method.create_view(ClassDetailsView)`

cfme/automate/explorer/__init__.py
```diff
@@ -40,8 +40,8 @@ def step(self):
         self.view.navigation.select(*automate_menu_name(self.obj.appliance) + ['Explorer'])
 
 
-def check_tree_path(actual, desired):
-    if len(actual) != len(desired):
+def check_tree_path(actual, desired, partial=False):
```
digitronik: Please, add one line comment about `partial` intention.

cfme/tests/automate/test_method.py
ganeshhubale: Thank you :)

PR #8361, opened 2019-01-14
[RFR] added shortcut for create_view() and wait_displayed()

izapolsk:
There are many places where we use 
```
view = create_view()
view.wait_displayed()
```
or 
```
view = create_view()
assert view.is_displayed
```
This change is going to replace all such calls with
```
view = create_view(wait='10s')
```
what will call wait_displayed after view creation

### DIFF
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -925,8 +925,7 @@ class VM(BaseVM):
             view.form.cancel.click()
             msg = 'Set/remove retirement date was cancelled by the user'
         if self.DETAILS_VIEW_CLASS is not None:
-            view = self.create_view(self.DETAILS_VIEW_CLASS)
-            assert view.is_displayed
+            view = self.create_view(self.DETAILS_VIEW_CLASS, wait='5s')
         view.flash.assert_success_message(msg)
 
     def equal_drift_results(self, drift_section, section, *indexes):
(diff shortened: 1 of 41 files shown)


cfme/common/vm.py
```diff
@@ -925,7 +925,7 @@ def set_retirement_date(self, when=None, offset=None, warn=None):
             view.form.cancel.click()
             msg = 'Set/remove retirement date was cancelled by the user'
         if self.DETAILS_VIEW_CLASS is not None:
-            view = self.create_view(self.DETAILS_VIEW_CLASS)
+            view = self.create_and_wait_view(self.DETAILS_VIEW_CLASS)
             assert view.is_displayed
```
digitronik: we can remove this assertion if you want.

cfme/utils/appliance/__init__.py
```diff
@@ -3187,6 +3187,11 @@ def create_view(self, view_class, o=None, override=None):
         return self.appliance.browser.create_view(
             view_class, additional_context={'object': new_obj})
 
+    def create_and_wait_view(self, view_class, o=None, override=None, timeout='10s'):
```
mshriver: Thought:

Did you consider implementation of `create_view(wait=None)`, where `wait` kwarg can be set to a timeout value. Existing `create_view` just modified to check for `wait` to be not-`None` and wait_displayed the given number of seconds.

cfme/services/catalogs/orchestration_template.py
```diff
@@ -178,7 +178,6 @@ def copy_template(self, template_name, content, draft=None, description=None):
                    })
         view.add_button.click()
         view.wait_displayed()
-        assert view.is_displayed
```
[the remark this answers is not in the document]
izapolsk: this is removed because wait_displayed already does this check. so, it jsut doesn't bring anything.

cfme/utils/appliance/__init__.py
izapolsk: this was just a light bulb moment. I hadn't considered embedding it in existing call because it would have required some analysis.

Do you think it's worth replacing it with parameter ?

PR #8363, opened 2019-01-14
[1LP][RFR] Manual Custom button

digitronik:
Purpose or Intent
=================
-  Moved custom button manuals to respective file with `manual` tag
- Deleted already automated manual tests
- Improved polarion tags

### DIFF
--- a/cfme/tests/automate/custom_button/test_buttons.py
+++ b/cfme/tests/automate/custom_button/test_buttons.py
@@ -327,3 +327,78 @@ def test_custom_button_quotes(appliance, provider, setup_provider, dialog, reque
 
     dialog_view.submit.click()
     view.flash.assert_message("Order Request was Submitted")
+
+
+@pytest.mark.manual
+def test_custom_button_simulation():
+    """ Test whether custom button works with simulation option
+
+    Bugzillas:
+        * 1535215
+
+    Polarion:
+        assignee: ndhandre
+        initialEstimate: 1/8h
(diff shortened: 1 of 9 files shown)


cfme/tests/automate/custom_button/test_buttons.py
```diff
@@ -328,3 +328,78 @@ def test_custom_button_quotes(appliance, provider, setup_provider, dialog, reque
 
     dialog_view.submit.click()
     view.flash.assert_message("Order Request was Submitted")
+
+
+@pytest.mark.manual
+def test_custom_button_simulation():
+    """ Test whether custom button works with simulation option
+
+    Bugzillas:
```
john-dupuy: @digitronik In some places you say `Bugzillas` and in others you say `Bugzilla` or `Reference BZs`, do we have a standard for referencing a BZ in the docstring? If not, it may be good to have one.

digitronik: We don't have standard or guideline. I used this in my test for my reference. Its helped me to find out bz which I faced related to that test. I don't think we need standard for things which are not a tag. We implement standard when we need filter out from doc strings. But its good to use standard related to BZ. I will push this idea thanks

john-dupuy: @digitronik LGTM

dajoRH:
*CFME QE Bot*

PR #8369, opened 2019-01-16
[1LP][RFR] New tagging updates

jawatts:
Added the ability to add tags from the dashboard view. Default functionality has remained the same with this change. 

{{ pytest: cfme/tests/cloud_infra_common/test_tag_objects.py --use-provider complete }}

### DIFF
--- a/cfme/common/__init__.py
+++ b/cfme/common/__init__.py
@@ -321,19 +321,32 @@ class TaggableCommonBase(object):
 class Taggable(TaggableCommonBase):
     """
     This class can be inherited by any class that honors tagging.
-    Class should have following
+
+    Classes inheriting this class should have following defined:
 
     * 'Details' navigation
     * 'Details' view should have entities.smart_management SummaryTable widget
-    * 'EditTags' navigation
-    * 'EditTags' view should have nested 'form' view with 'tags' table widget
-    * Suggest using class cfme.common.TagPageView as view for 'EditTags' nav
+    * 'EditTagsFromDetails' navigation
+    * 'EditTagsFromDetails' At a minimum, views should include EditTagsFromDetails class. View
+    should have nested 'form' view with 'tags' table widget
+    'In addition, the prerequisite should be NavigateToSibling('Details')
+
+    Optional classes that can be defined:
+
+    'EditTags':
+    * 'All' navigation
+    * Same as EditTagsFromDetails but the prerequisite should be NavigateToSibling('All')
+    'EditTagsFromDashboard':
+    * 'Dashboard' navigation
+    * Same as EditTagsFromDetails but the prerequisite should be NavigateToSibling('Dashboard')
+
+    Suggest using class cfme.common.TagPageView as view for all 'EditTags' nav
 
     This class provides functionality to assign and unassigned tags for page models with
     standardized widgetastic views
     """
 
-    def add_tag(self, tag=None, cancel=False, reset=False, details=True):
+    def add_tag(self, tag=None, cancel=False, reset=False, details=True, dashboard=False):
         """ Add tag to tested item
 
         Args:
@@ -342,8 +355,14 @@ class Taggable(TaggableCommonBase):
             reset: set True to reset already set up tag
             details (bool): set False if tag should be added for list selection,
                             default is details page
+            dashboard (bool): Set to True if tag should be added via the Dashboard view
         """
-        step = 'EditTagsFromDetails' if details else 'EditTags'
+        if details and not dashboard:
+            step = 'EditTagsFromDetails'
+        elif dashboard:
+            step = 'EditTagsFromDashboard'
+        else:
+            step = 'EditTags'
         view = navigate_to(self, step)
         added_tag, updated = self._assign_tag_action(view, tag)
         # In case if field is not updated cancel the edition
--- a/cfme/containers/project.py
+++ b/cfme/containers/project.py
@@ -23,7 +23,7 @@ class ProjectDetailsView(ContainerObjectDetailsBaseView):
     SUMMARY_TEXT = 'Container Projects'
 
 
-class ProjectDashboardView(ContainerObjectDetailsBaseView):
+class ProjectDashboardView(ProjectDetailsView):
 
     @property
     def is_displayed(self):
@@ -111,11 +111,9 @@ class Details(CFMENavigateStep):
         self.prerequisite_view.entities.get_entity(name=self.obj.name,
                                                    surf_pages=not search_visible,
                                                    use_search=search_visible).click()
-        self.view.toolbar.view_selector.select("Summary View")
 
     def resetter(self):
-        if (self.appliance.version.is_in_series('5.9') and
-                self.view.toolbar.view_selector.is_displayed):
+        if self.view.toolbar.view_selector.is_displayed:
             self.view.toolbar.view_selector.select("Summary View")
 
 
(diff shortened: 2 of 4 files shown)


dajoRH:
Lint report for commit 55dfb2506a81252d0f54113161afced8b7fdfbbb:

`cfme/tests/containers/test_tags.py`:
- :bangbang: Line 18:26: **E265** *block comment should start with '# '*

**No commit flaws detected**.

*CFME QE Bot*

cfme/common/__init__.py
```diff
@@ -321,19 +321,28 @@ def _tags_action(self, view, cancel, reset):
...
-    * 'EditTags' navigation
-    * 'EditTags' view should have nested 'form' view with 'tags' table widget
+    * 'EditTagsFromDetails' navigation
+    * 'EditTagsFromDetails' At a minimum, views should include EditTagsFromDetails class. View
+    should have nested 'form' view with 'tags' table widget
+    'In addition, the prerequisite should be NavigateToSibling('Details')
+
+    Optional classes that can be defined:
+
+    * 'Dashbaord' navigation
```
lina-nikiforova: Just a small typo in `Dashboard` :)

jawatts: Good catch!

cfme/common/__init__.py
```diff
@@ -321,19 +321,28 @@ def _tags_action(self, view, cancel, reset):
...
-    Class should have following
+
+    Classes inheriting this class should have following defined:
 
     * 'Details' navigation
     * 'Details' view should have entities.smart_management SummaryTable widget
-    * 'EditTags' navigation
-    * 'EditTags' view should have nested 'form' view with 'tags' table widget
+    * 'EditTagsFromDetails' navigation
+    * 'EditTagsFromDetails' At a minimum, views should include EditTagsFromDetails class. View
```
lina-nikiforova: Q: Can you explain to me this part? I checked a few classes that inherit `Taggable` and I don't see this navigation defined.

jawatts: Absolutely. Here is an example:
https://github.com/ManageIQ/integration_tests/blob/master/cfme/common/__init__.py#L498

I have observed a couple places where we have the class name wrong. The class has a preq of navigating to details but the class name is `EditTags`. Here is an example of that:
https://github.com/ManageIQ/integration_tests/blob/master/cfme/containers/replicator.py#L107

`EditTags` should have the preq of navigating to `All`. I should add that into to the doc block.

lina-nikiforova: Thanks for the explanation!

cfme/tests/containers/test_tags.py
```diff
@@ -44,37 +50,67 @@ def get_collection_entity(appliance, collection_name, provider):
...
+        object_tags.category.display_name == tag.category.display_name and
+        object_tags.display_name == tag.display_name for object_tags in tags), (
+        "{}: {} not in ({})".format(tag.category.display_name, tag.display_name, str(tags)))
+
+    obj_under_test.remove_tag(tag=tag, details=details)
+
+    post_remove_tags = obj_under_test.get_tags()
+    if post_remove_tags:
+        for post_tags in post_remove_tags:
+            soft_assert(
```
digitronik: Q:
why `soft_assert`? As per tag behavior if this condition not satisfy then test should fails. Please, correct me about this behavior.

cfme/tests/containers/test_tags.py
```diff
@@ -44,37 +50,67 @@ def get_collection_entity(appliance, collection_name, provider):
...
+    verify_tags(obj_under_test=obj_under_test, tag=tag, details=tag_place, dashboard=False,
+                soft_assert=soft_assert)
 
-    tags = obj_under_test.get_tags()
 
-    assert any(
-        object_tags.category.display_name == tag.category.display_name and
-        object_tags.display_name == tag.display_name for object_tags in tags), (
-        "{}: {} not in ({})".format(tag.category.display_name, tag.display_name, str(tags)))
+@pytest.mark.parametrize('test_param', bz_1665284_test_items, ids=[bti for bti in
```
digitronik: Required:
No need to add `ids`; If you are not manipulating ids with different tags. You are using same names.

cfme/tests/containers/test_tags.py
```diff
@@ -44,37 +50,67 @@ def get_collection_entity(appliance, collection_name, provider):
...
-def test_tag_container_objects(soft_assert, test_param, appliance, provider, request, tag,
+def test_tag_container_objects(soft_assert, test_param, appliance, provider, tag,
                                tag_place):
     """ Test for container items tagging action from list and details pages
 
     Polarion:
         assignee: juwatts
-        caseimportance: medium
         casecomponent: Containers
+        caseimportance: medium
```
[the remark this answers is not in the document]
jawatts: this was from rebase.

cfme/tests/containers/test_tags.py
```diff
@@ -44,10 +48,32 @@ def get_collection_entity(appliance, collection_name, provider):
...
+        for post_tags in post_remove_tags:
+            assert(
+                post_tags.category.display_name != tag.category.display_name and
+                post_tags.display_name != tag.display_name)
+
+
+@pytest.mark.parametrize('test_param', container_test_items)
 @pytest.mark.parametrize('tag_place', [True, False], ids=['details', 'list'])
-def test_tag_container_objects(soft_assert, test_param, appliance, provider, request, tag,
+def test_tag_container_objects(test_param, appliance, provider, tag,
```
digitronik: Its <100 char.

dajoRH:
*CFME QE Bot*

cfme/containers/project.py
```diff
@@ -111,11 +111,9 @@ def step(self):
         self.prerequisite_view.entities.get_entity(name=self.obj.name,
                                                    surf_pages=not search_visible,
                                                    use_search=search_visible).click()
-        self.view.toolbar.view_selector.select("Summary View")
 
     def resetter(self):
-        if (self.appliance.version.is_in_series('5.9') and
```
jawatts: I don't think we need to check versions anymore. 5.9 and 5.10 both has "Summary" view.

cfme/containers/project.py
```diff
@@ -135,10 +133,15 @@ def resetter(self):
...
+class EditTagsFromDetails(CFMENavigateStep):
     VIEW = TagPageView
     prerequisite = NavigateToSibling('Details')
 
     def step(self):
         self.prerequisite_view.toolbar.policy.item_select('Edit Tags')
+
+
+@navigator.register(Project, 'EditTagsFromDashboard')
+class EditTagsFromDashboard(EditTagsFromDetails):
```
izapolsk: REQUIRED: I suppose this is typo. please check this out again

cfme/containers/provider/__init__.py
```diff
@@ -149,6 +149,15 @@ def summary_text(self):
         return self.SUMMARY_TEXT
 
 
+class ContainersProviderDashboardView(ContainerProviderDetailsView):
+
+    @property
+    def is_displayed(self):
+        return(
```
izapolsk: REQUIRED: typo ?

cfme/containers/project.py
izapolsk: @psav, could you remind me if this is ok ?  ^^
Afaik, we had some issues with step inheritance before.

jawatts: I don't see a typo here. What are you seeing?

cfme/containers/provider/__init__.py
jawatts: where do you see the typo?

cfme/common/__init__.py
```diff
@@ -321,19 +321,32 @@ def _tags_action(self, view, cancel, reset):
...
+    should have nested 'form' view with 'tags' table widget
+    'In addition, the prerequisite should be NavigateToSibling('Details')
+
+    Optional classes that can be defined:
+
+    'EditTags':
+    * 'All' navigation
+    * Same as EditTagsFromDetails but the prerequisite should be NavigateToSibling('All')
+    'EditTagsFromDasboard':
```
jawatts: Typo here, fixing it

cfme/containers/provider/__init__.py
izapolsk: there should be at least some space between return and (

cfme/containers/project.py
izapolsk: you inherit one step from another one. I thought this is typo. I got later that you did this intentionally.
let's wait for @psav reply. I'd like to have a confirmation.

psav: Hmmmm - I'm not sure it's a supported config, but perhaps it could be, @izapolsk can you remember the issues we faced?

mshriver: There's not much code that gets repeated for these navigation destinations, and I'm a bit hesitant to connect them in this way, because of the potential changes to navigation that would then impact multiple pages.

jawatts: I will change it

PR #8374, opened 2019-01-17
[RFR] Only use REST API for GET/POST of Advanced Settings

jawatts:
REST API is available in 5.9, use this for the GET/POST of advanced settings instead of rails.

{{ pytest:  cfme/tests/platform/test_advanced_config.py --use-provider complete }}

### DIFF
--- a/cfme/tests/containers/test_blacklisted_container_events.py
+++ b/cfme/tests/containers/test_blacklisted_container_events.py
@@ -2,7 +2,6 @@ import fauxfactory
 import pytest
 import time
 
-from copy import deepcopy
 from wrapanapi.systems.container.rhopenshift import ApiException
 
 from cfme.containers.provider import ContainersProvider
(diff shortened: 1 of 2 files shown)


dajoRH:
Lint report for commit ebe18c750c9a94024585303dc47ef4cd439e392e:

- :bangbang: Line 2089:5: **E303** *too many blank lines (2)*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/containers/test_blacklisted_container_events.py
```diff
@@ -69,6 +68,18 @@
 }
 
 
+def restore_advanced_settings(appliance):
+    """Restores the Advanced Settings config"""
+    if appliance.version < '5.10':
```
john-dupuy: Why not use `VersionPicker` here?

cfme/tests/containers/test_blacklisted_container_events.py
```diff
@@ -69,6 +68,18 @@
 }
 
 
+def restore_advanced_settings(appliance):
```
john-dupuy: Will `update_advanced_settings` be modified in the future to behave in this way? Or is this a temporary workaround?

cfme/tests/containers/test_blacklisted_container_events.py
jawatts: I could but VersionPicker is generally used when you do not have the appliance object.

cfme/tests/containers/test_blacklisted_container_events.py
jawatts: In what way are you referring to here?

john-dupuy: Nevermind, I see that here you are only changing a single setting that depends on the version of the appliance.

PR #8377, opened 2019-01-21
[1LP][RFR] Auth: add logout after evm restart

apagac:
This fixes numerous 'CandidateNotFound' errors for both 5.9 and 5.10.

The error:

> CandidateNotFound
> path: ('Zones', 'Zone: Default Zone (current)', 'Server:  [1] (current)'), message: Could not find the item CFME Region: Region 0 [0]/Zones/Zone: Default Zone (current)/Server:  [1] (current) in Boostrap tree settings_tree, cause: Was not found in Zone: Default Zone (current)

Cause:
During auth testing, evmserverd is restarted; after that, we wait for an UI to appear. What in fact does happen in automation is that the UI never disappears. The browser sits there and seems like the server never restarted. This causes appliance to behave unexpectedly when the automation tries to navigate or perform actions in UI.
The appliance does not logout the user after evmserverd restart. BZ in progress.

Proposed fix:
Added server.logout after the evmserverd restart. This forces automation to log in again and everything works as expected from there.

{{ pytest: -v --long-running cfme/tests/integration/test_cfme_auth.py -k test_login_evm_group --use-provider vsphere67-nested }}

### DIFF
--- a/cfme/fixtures/authentication.py
+++ b/cfme/fixtures/authentication.py
@@ -74,3 +74,6 @@ def configure_auth(appliance, auth_mode, auth_provider, user_type, request, fix_
     appliance.server.authentication.auth_settings = original_config
     appliance.evmserverd.restart()
     appliance.wait_for_web_ui()
+    # After evmserverd restart, we need to logout from the appliance in the UI.
+    # Otherwise the UI would be in a bad state and produce errors while testing.
+    appliance.server.logout()


apagac:
Looking at the pass rate on trackerbot, it seems much better and no 'CandidateNotFound' error is present.

PRs to fix remaining issues incoming.

cfme/fixtures/authentication.py
```diff
@@ -74,3 +74,4 @@ def configure_auth(appliance, auth_mode, auth_provider, user_type, request, fix_
     appliance.server.authentication.auth_settings = original_config
     appliance.evmserverd.restart()
     appliance.wait_for_web_ui()
+    appliance.server.logout()
```
jawatts: Another options here is if you call 'appliance.server.browser.refresh`, you will be redirected to the logout page as well.

**Required** Please add a comment here explaining why you are logging out so it is clear for the next engineer that looks at this fixture.

jawatts:
Note to reviewers: as @apagac stated above this PR fixes the `CandidateNotFound`exception only. In the 5.10 and 5.9 runs, that exception is no longer present.

cfme/fixtures/authentication.py
apagac: Comment added.

apagac:
Moving to [1LP][RFR] without waiting for PRT results as there was no change in the code, just added comment.

PR #8378, opened 2019-01-21
[1LP][RFR] CB events infra object and Mixing

digitronik:
Purpose or Intent
=================
{{pytest: cfme/tests/automate/custom_button/test_infra_objects.py::test_custom_button_events -v}}

### DIFF
--- a/cfme/common/__init__.py
+++ b/cfme/common/__init__.py
@@ -658,3 +658,61 @@ class UtilizationMixin(object):
     # @cached_property
     # def utilization(self):
     #     return Utilization(self)
+
+
+class CustomButtonEventsView(View):
+    """Class represents common custom button events page in CFME UI"""
+
+    breadcrumb = BreadCrumb()
+    title = Text('//div[@id="main-content"]//h1')
+    table = Table('//*[@class="miq-data-table"]/table')
+
+    @property
+    def is_displayed(self):
+        expected_title = "{} (All Custom Button Events)".format(self.context["object"].name)
+        return (
+            self.title.text == expected_title and self.breadcrumb.active_location == expected_title
+        )
+
+
+class CustomButtonEventsMixin(object):
+    @property
+    def cb_event_count(self):
+        view = navigate_to(self, "Details")
+        ent = view.entities
+        table = ent.summary("Relationships") if hasattr(ent, "summary") else ent.relationships
+        # Some objects have filed text `Custom button events` in relationship table
+        cb_text = (
+            "Custom Button Events"
+            if "Custom Button Events" in table.fields
+            else "Custom button events"
+        )
+        return int(table.get_text_of(cb_text))
+
+    def get_button_events(self):
+        try:
+            view = navigate_to(self, "ButtonEvents")
+            return view.table.read()
+        except DestinationNotFound:
+            return []
+
+
+@navigator.register(CustomButtonEventsMixin, "ButtonEvents")
+class CustomButtonEvents(CFMENavigateStep):
+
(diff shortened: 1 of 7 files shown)


cfme/common/__init__.py
```diff
@@ -658,3 +658,61 @@ class UtilizationMixin(object):
...
+        )
+
+
+class CustomButtonEventsMixin(object):
+    @property
+    def cb_event_count(self):
+        view = navigate_to(self, "Details")
+        ent = view.entities
+        table = ent.summary("Relationships") if hasattr(ent, "summary") else ent.relationships
+        # Some objects have filed text `Custom button events` in relationship table
```
mshriver: Please submit a BZ against this inconsistency in the UI, with UX whiteboard.

cfme/common/__init__.py
```diff
@@ -658,3 +658,61 @@ class UtilizationMixin(object):
...
+    prerequisite = NavigateToSibling("Details")
+
+    def step(self):
+        ent = self.prerequisite_view.entities
+        table = ent.summary("Relationships") if hasattr(ent, "summary") else ent.relationships
+        # Some objects have filed text `Custom button events` in relationship table
+        cb_text = (
+            "Custom Button Events"
+            if "Custom Button Events" in table.fields
+            else "Custom button events"
```
mshriver: I don't like the duplication between this and the convenience methods provided by the mixin class above.  Let's find a better way to capture this poor UX workaround, and the variance in the table name/access.

Potentially going as far as updating all of the classes where you expect to add the mixin, changing their entities definitions to use ParametrizedSummaryTable consistently.  Then you can eliminate the ternary statement above.  This would likely require changes in the tests related to the framework views that are updated.

cfme/common/__init__.py
```diff
@@ -658,3 +658,61 @@ class UtilizationMixin(object):
...
+        )
+        return int(table.get_text_of(cb_text))
+
+    def get_button_events(self):
```
mshriver: Couple of thoughts here:

1. Is a '*_count' method really necessary when this returns a list that the caller could just use len() on?

2.  Assuming we do need a '*_count' method, let's make this mixin consistent, and have it provide two properties, `button_event_count` and `button_events`.  Given the duplicated string/attribute selection in navigation and the count method, I'd lean towards keeping the Mixin skinny, and just linking Collections/Entities to a `ButtonEvents` navigation.

3. Why would `DestinationNotFound` occur when the nav class is registered a few lines below?

4. Ultimately these two convenience methods don't really buy that much convenience in the tests. The calls to navigate to 'Details' are already being made in the test, and then the value you could pull from the view you already have. Burying navigation in these convenience methods tends to end up with unexpected view changes during tests.  Convenience in one test is a poor assumption in another.

5. This class _could_ just be empty, and a placeholder to connect things to the navigation destination.

cfme/tests/automate/custom_button/test_infra_objects.py
```diff
@@ -455,3 +455,57 @@ def _reset_window():
...
+            unblock=lambda button_group, btn_dialog: not ("HOSTS" in button_group and btn_dialog),
+        )
+    ]
+)
+@pytest.mark.ignore_stream("5.9")
+@pytest.mark.parametrize("btn_dialog", [False, True], ids=["simple", "dialog"])
+def test_custom_button_events(request, dialog, setup_obj, button_group, btn_dialog):
+    """Test custom button events
+
+    Polarion:
```
mshriver: :tada:

PR #8379, opened 2019-01-21
[1LP][RFR] _select template for all pages

lina-nikiforova:
{{ pytest: -v cfme/tests/cloud_infra_common/test_provisioning.py::test_provision_with_tag }}

`test_provision_with_tag[ec2]` was failing because `def _select_template` was looking for the template only on the first page and the template was not there.
There are tons of templates on ec2 provider, so some iteration is needed.

### DIFF
--- a/cfme/common/vm_views.py
+++ b/cfme/common/vm_views.py
@@ -317,6 +317,7 @@ class ProvisionView(BaseLoggedInPage):
     title = Text('#explorer_title_text')
     breadcrumb = BreadCrumb()
     image_table = Table('//div[@id="pre_prov_div"]//table')
+    paginator = PaginationPane()
 
     @View.nested
     class sidebar(View):  # noqa
@@ -334,8 +335,10 @@ class ProvisionView(BaseLoggedInPage):
 
         def _select_template(self, template_name, provider_name):
             try:
-                row = self.parent.image_table.row(name=template_name, provider=provider_name)
-                row.click()
+                self.parent.paginator.find_row_on_pages(self.parent.image_table,
+                                                        name=template_name,
+                                                        provider=provider_name).click()
+            # image was not found, therefore raise an exception
             except IndexError:
                 raise TemplateNotFound('Cannot find template "{}" for provider "{}"'
                                        .format(template_name, provider_name))


lina-nikiforova:
PRT failures are for `openstack-11`  that are not relevant to this change

cfme/common/vm_views.py
```diff
@@ -317,6 +317,7 @@ class ProvisionView(BaseLoggedInPage):
     title = Text('#explorer_title_text')
     breadcrumb = BreadCrumb()
     image_table = Table('//div[@id="pre_prov_div"]//table')
```
digitronik: @lina-nikiforova can you please confirm that are you able to cover this table with `BaseEntitiesView`. Just simple though comes in my mind. If yes then it will solve your problem.

cfme/common/vm_views.py
```diff
@@ -333,12 +334,15 @@ class form(BasicProvisionFormView):  # noqa
...
-                raise TemplateNotFound('Cannot find template "{}" for provider "{}"'
-                                       .format(template_name, provider_name))
+            for _ in self.parent.paginator.pages():
+                try:
+                    self.parent.image_table.row(name=template_name, provider=provider_name).click()
+                    return
+                except IndexError:
+                    continue
+            # image was not found, therefore raise an exception
+            raise TemplateNotFound('Cannot find template "{}" for provider "{}"'
```
digitronik: I would use
```
for ....
else:
    rasie ...
```
what you think?

digitronik:
For now moving to WIPTEST.

cfme/common/vm_views.py
jawatts: Have you considered putting this ins a `for/else`?

cfme/common/vm_views.py
```diff
@@ -333,10 +334,14 @@ class form(BasicProvisionFormView):  # noqa
         cancel_button = Button('Cancel')
 
         def _select_template(self, template_name, provider_name):
-            try:
-                row = self.parent.image_table.row(name=template_name, provider=provider_name)
-                row.click()
-            except IndexError:
+            for _ in self.parent.paginator.pages():
+                try:
+                    self.parent.image_table.row(name=template_name, provider=provider_name).click()
```
digitronik: I think it was my bad; 
can you please try:
```
view.paginator.find_row_on_pages(view.image_table, name=template_name)
```

Let me know its works or not.

lina-nikiforova: @digitronik awesome, thanks! It works beautifully!

mshriver: Yay you used my function!

PR #8392, opened 2019-01-23
[1LP][RFR] Disk modes cannot be specified when adding disk to VM in RHV provider

lina-nikiforova:
{{ pytest: -v cfme/tests/infrastructure/test_vm_reconfigure.py::test_vm_reconfig_add_remove_disk_cold  --long-running }}

As written here https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/infrastructure/test_vm_reconfigure.py#L111

disk modes are not supported for RHV

### DIFF
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -895,6 +895,7 @@ class InfraVm(VM):
                      See VMConfiguration.get_changes_to_fill to see expected format of the data
             cancel: `False` if we want to submit the changes, `True` otherwise
         """
+        from cfme.infrastructure.provider.rhevm import RHEVMProvider
         if not new_configuration and not changes:
             raise TypeError(
                 "You must provide either new configuration or changes to apply.")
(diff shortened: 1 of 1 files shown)


cfme/infrastructure/virtual_machines.py
```diff
@@ -895,6 +895,7 @@ def reconfigure(self, new_configuration=None, changes=None, cancel=False):
                      See VMConfiguration.get_changes_to_fill to see expected format of the data
             cancel: `False` if we want to submit the changes, `True` otherwise
         """
+        from cfme.infrastructure.provider.rhevm import RHEVMProvider
```
john-dupuy: Why not put this with the imports at the top of the file?

cfme/infrastructure/virtual_machines.py
```diff
@@ -934,19 +935,24 @@ def reconfigure(self, new_configuration=None, changes=None, cancel=False):
 
                 row = vm_recfg.disks_table.click_add_disk()
                 row.type.fill(disk.type)
-                row.mode.fill(mode)
+                if not self.provider.one_of(RHEVMProvider):
```
john-dupuy: :+1:

cfme/infrastructure/virtual_machines.py
izapolsk: infinite imports loop ?

PR #8403, opened 2019-01-25
[1LP][RFR] Update physical infrastructure test collection

mshriver:
physical infra tests were collecting when no provider was available, although they needed a provider to be present for the test and fixtures to work.

Updated fixtures to cleanly skip instead of raise IndexError when a provider doesn't have the necessary physical entities present.

Fixes one FixtureLookupError created by a method that wasn't decorated as a fixture.

### DIFF
--- a/cfme/tests/physical_infrastructure/test_physical_rack.py
+++ b/cfme/tests/physical_infrastructure/test_physical_rack.py
@@ -4,7 +4,8 @@ import pytest
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.physical.provider.lenovo import LenovoProvider
 
-pytestmark = [pytest.mark.tier(3), pytest.mark.provider([LenovoProvider], scope="function")]
+pytestmark = [pytest.mark.tier(3),
+              pytest.mark.provider([LenovoProvider], scope="function")]
 
 
 @pytest.fixture(scope="module")
@@ -14,7 +15,7 @@ def physical_rack(appliance, provider, setup_provider):
         physical_racks = appliance.collections.physical_racks.filter({"provider": provider}).all()
         return physical_racks[0]
     except IndexError:
-        pytest.skip("No rack resource found")
+        pytest.skip("No rack resource on provider")
 
 
 def test_physical_rack_details_dropdowns(physical_rack):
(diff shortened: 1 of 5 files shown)


mshriver:
@matejart @matheuscmelo Thoughts on these changes to collection?

Should be no functional impacts to the tests themselves - just changes to how they collect, and that they skip if they are collected with a provider available, but its missing a resource.

cfme/tests/physical_infrastructure/test_physical_rack.py
```diff
@@ -27,19 +28,14 @@ def test_physical_rack_details_dropdowns(physical_rack):
...
     """Navigate to the physical racks page and verify that the dropdown menus are present
 
     Polarion:
         assignee: rhcf3_machine
         initialEstimate: 1/4h
     """
-    physical_racks_view = navigate_to(physical_rack_collection, 'All')
+    physical_racks_view = navigate_to(appliance.collections.physical_racks, 'All')
 
     configuration_items = physical_racks_view.toolbar.configuration.items
```
mshriver: This menu item is disabled without a provider configured. This is why I added physical_rack to its parameters list.

matejart:
In terms of the Redfish physical infrastructure provider tests, the tests are meant to work against a mock-up Redfish server, where we know that if the provider is set and configured, the resources tested should be available.
That said, we only have a `physical_server` fixture in the `test_redfish_physical_server_details.py` that actually expects one physical server to be present. The other tests will run without doing anything (but still succeeding).
So if the aim is to use the integration tests against some actual provider that may or may not have the expected resources, then skipping sounds reasonable.

cfme/tests/physical_infrastructure/ui/test_physical_switch_list.py
```diff
@@ -18,7 +18,7 @@ def test_physical_switches_view_displayed(appliance):
     assert physical_switches_view.is_displayed
 
 
-def test_physical_switches_view_dropdowns(appliance):
+def test_physical_switches_view_dropdowns(appliance, physical_switch):
```
john-dupuy: Is the reason you need `physical_switch` the same as above -- i.e. the menu item is disabled without the provider configured?

mshriver:
@matejart Sounds good - yeah my aim is to change the behavior of the test collection and execution when there is no provider available for collection - should not change collection when there _is_ a provider. It will change failure behavior (to a skip) if the provider is there. We always intend to skip during setup, instead of fail, when a resource is not available.  These modules are marked for LenovoProvider btw, not RedfishProvider or generic PhysicalInfrastructureProvider.

cfme/tests/physical_infrastructure/ui/test_physical_switch_list.py
mshriver: Yep - Including physical_switch here adds the dependency on `provider` + `setup_provider`, which will prevent the test from getting collected if there is no provider in the yamls.  The dropdowns are disabled when there aren't entities present.

If there is a provider in the yamls, but that provider doesn't actually have a switch available, then the test will skip in setup via the fixture instead of raising an exception.

PR #8406, opened 2019-01-28
[1LP][RFR] Fixing catalog exists property

### LINKED ISSUE #8405: Catalog exists property throws "CandidateNotFound" Exception
When we call <catalog_obj>.exists , it throws  "CandidateNotFound" Exception, where as in our test cases  we expect a Boolean value "False"
>> https://github.com/ManageIQ/integration_tests/blob/master/cfme/services/catalogs/catalog.py#L119 

Steps to Reproduce: <catalog_obj>.exists
Actual Result:  Trace-back of   "CandidateNotFound" Exception
Expected Result: False

mnadeem92:
Catalog exists property throws "CandidateNotFound" Exception
This patch except the exception and return False Output
Closes #8405 

{{pytest: cfme/tests/services/test_service_rbac.py::test_service_rbac_catalog -v}}

### DIFF
--- a/cfme/services/catalogs/catalog.py
+++ b/cfme/services/catalogs/catalog.py
@@ -3,7 +3,7 @@ import attr
 from widgetastic.utils import Parameter
 from widgetastic.widget import Text
 from widgetastic_manageiq import MultiBoxSelect
-from widgetastic_patternfly import Button, Input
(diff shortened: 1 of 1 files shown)


izapolsk:
@mnadeem92, please add tests it fixes to PRT.

PR #8407, opened 2019-01-28
[WIPTEST] Update catalogbundle View/Collection.create

mshriver:
View changes have broken the create method's fill of the form. This is impacting ~35 tests, tried to pick one of the shorter ones.

@ganeshhubale please take a look as this is related to the view changes for the entry points in that view.

Not sure that this will completely fix the `create` method, but it should at least resolve the `AttributeError` triggered because of the invalid widget name.

{{ pytest: --long-running -vv --use-provider complete cfme/tests/services/test_catalog_item.py -k test_tagvis_catalog_bundle }}

### DIFF
--- a/cfme/services/catalogs/catalog_items/catalog_bundles.py
+++ b/cfme/services/catalogs/catalog_items/catalog_bundles.py
@@ -18,7 +18,7 @@ class CatalogBundleFormView(ServicesCatalogView):
     @View.nested
     class basic_info(WaitTab):  # noqa
         TAB_NAME = 'Basic Info'
-        included_form = View.include(BasicInfoForm)
+        included_form = View.include(BasicInfoForm, use_parent=True)
 
     @View.nested
     class resources(WaitTab):  # noqa
@@ -73,7 +73,7 @@ class CatalogBundle(NonCloudInfraCatalogItem):
         view = self.create_view(DetailsCatalogItemView, override=updates, wait='10s')
         view.flash.assert_no_error()
 
-
+@attr.s
 class CatalogBundlesCollection(BaseCollection):
     ENTITY = CatalogBundle
 
@@ -95,9 +95,9 @@ class CatalogBundlesCollection(BaseCollection):
             'select_catalog': cat_name,
             'select_dialog': dialog
         })
-        if view.basic_info.field_entry_point.value == "":
-            view.basic_info.fill({'field_entry_point': ''})
-            view.basic_info.tree.click_path(
+        if view.basic_info.provisioning_entry_point.value == "":
+            view.basic_info.fill({'provisioning_entry_point': ''})
+            view.basic_info.modal.tree.click_path(
                 "Datastore", domain, "Service", "Provisioning",
                 "StateMachines", "ServiceProvision_Template", "CatalogItemInitialization")
             view.apply_button.click()


cfme/services/catalogs/catalog_items/catalog_bundles.py
```diff
@@ -95,9 +95,9 @@ def create(self, name, catalog_items=None, catalog=None, description=None, displ
             'select_catalog': cat_name,
             'select_dialog': dialog
         })
-        if view.basic_info.field_entry_point.value == "":
-            view.basic_info.fill({'field_entry_point': ''})
-            view.basic_info.tree.click_path(
+        if view.basic_info.provisioning_entry_point.value == "":
```
ganeshhubale: @mshriver 
- I am agree with you. This issue occurred because of changes for the entry points in ```BasicInfoForm``` view. Also as you said we have fixed ```AttributeError``` for wrong widget names but need to improve ```create``` method.

- According to my observations,
    - While creating ```catalog bundle```: **Provisioning Entry Point** and **Retirement Entry Point** has default values. So this check ```if view.basic_info.provisioning_entry_point.value == "":``` will always return **false**. Hence we don't need this block of code if there is no such condition where we will get  **Provisioning Entry Point** is **None**.
    - Screenshot of catalog bundle basic info view:
cat_bundle_entry

ganeshhubale:
@mshriver 
- As we have already, improved the ```fill``` method of entry points widgets. So in this fix, we can add these entry points in ```  view.basic_info.fill```.
- I improved ```fill``` method as below and it worked as expected locally.
- This will help in future if we want to pass different values for these entry points.
```
default_prov = (
        "/Service/Provisioning/StateMachines/ServiceProvision_Template/CatalogBundleInitialization")
    default_retire = (
        "/Service/Retirement/StateMachines/ServiceRetirement/Default")

    def create(self, name, catalog_items=None, catalog=None, description=None, display_in=None,
               dialog=None, provisioning_entry_point=default_prov, reconfigure_entry_point=None,
               retirement_entry_point=default_retire, domain="ManageIQ (Locked)"):
        # TODO Move this logic into the view, the main obstacle is filling 'catalog_items'
        view = navigate_to(self, 'Add')

        # In 5.10 catalog name is appended with 'My Company'
        cat_name = VersionPicker({
            LOWEST: getattr(catalog, 'name', None),
            '5.10': 'My Company/{}'.format(getattr(catalog, 'name', None))
        }).pick(self.appliance.version)
        view.basic_info.fill({
            'name': name,
            'description': description,
            'display': display_in,
            'select_catalog': cat_name,
            'select_dialog': dialog,
            'provisioning_entry_point': provisioning_entry_point,
            'reconfigure_entry_point': reconfigure_entry_point,
            'retirement_entry_point': retirement_entry_point,
        })

        for cat_item in catalog_items:
            view.resources.fill({'select_resource': cat_item})
        view.add_button.click()
        view.flash.assert_success_message('Catalog Bundle "{}" was added'.format(name))
        view = self.create_view(AllCatalogItemView, wait='10s')
        view.flash.assert_no_error()
        return self.instantiate(name, catalog_items=catalog_items, catalog=catalog,
                                description=description, display_in=display_in, dialog=dialog,
                                domain=domain)

```
Thank you.
Ganesh

mshriver:
@ganeshhubale Thanks! Please open a PR with those changes, and we'll check it out. I'm going to merge the small changes I have here, but look forward to you building on them!

PR #8411, opened 2019-01-29
[1LP][RFR] Group enhancement

digitronik:
Purpose or Intent
=================
- Added `MutiBoxOrderedSelect` selection widget in `wt.miq`
- Enhance Button group create method with selection
- Added new test for  checking ordered changes

{{pytest: cfme/tests/automate/custom_button/test_buttons.py::test_custom_button_order_sort -v}}

### DIFF
--- a/cfme/automate/buttons.py
+++ b/cfme/automate/buttons.py
@@ -26,8 +26,8 @@ from cfme.utils.appliance.implementations.ui import CFMENavigateStep
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.appliance.implementations.ui import navigator
 from cfme.utils.update import Updateable
-from widgetastic_manageiq import MultiBoxOrderedSelect
 from widgetastic_manageiq import FonticonPicker
+from widgetastic_manageiq import MultiBoxOrderedSelect
 from widgetastic_manageiq import PotentiallyInvisibleTab
 from widgetastic_manageiq import RadioGroup
 from widgetastic_manageiq import SummaryFormItem
@@ -713,12 +713,13 @@ class ButtonGroup(BaseEntity, Updateable):
         image: Icon of Group.
         display: Group name display on Button.
         icon_color: Icon Color of Group.
+        assign_buttons: List of assigned buttons to Group.
     """
 
     text = attr.ib()
     hover = attr.ib()
     type = attr.ib()
-    image = attr.ib()
+    image = attr.ib(default="fa-user")
     display = attr.ib(default=None)
     icon_color = attr.ib(default=None)
     assign_buttons = attr.ib(default=None)
@@ -803,21 +804,6 @@ class ButtonGroupCollection(BaseCollection):
     TENANT = "Tenant"
     TEMPLATE_IMAGE = "VM Template and Image"
     VM_INSTANCE = "VM and Instance"
-
-    def instantiate(
-        self,
-        text,
-        hover,
-        type,
-        image="fa-user",
-        display=None,
-        icon_color=None,
-        assign_buttons=None,
-    ):
-        self.type = type
-        return self.ENTITY.from_collection(
-            self, text, hover, type, image, display, icon_color, assign_buttons
-        )
 
     def create(
         self,
--- a/cfme/automate/buttons.py
+++ b/cfme/automate/buttons.py
@@ -26,6 +26,7 @@ from cfme.utils.appliance.implementations.ui import CFMENavigateStep
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.appliance.implementations.ui import navigator
 from cfme.utils.update import Updateable
+from widgetastic_manageiq import ButtonMultiBoxSelect
 from widgetastic_manageiq import FonticonPicker
 from widgetastic_manageiq import PotentiallyInvisibleTab
 from widgetastic_manageiq import RadioGroup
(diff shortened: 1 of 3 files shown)


widgetastic_manageiq/__init__.py
```diff
@@ -5141,3 +5141,63 @@ def click_body_action(self):
 
     def click(self):
         return self.click_title()
+
+
+class MultiBoxOrderedSelect(MultiBoxSelect):
```
sshveta: Question : There is already a widget called MultiBoxSelect . 
Can you modify that  to your requirements ?

digitronik: Yes, I am modifying `MultiBoxSelect`... Its not fallowing `sequence` of values... So its different than that widget. It will take care of `Order` of values... pointing index of values.

sshveta: :+1:

widgetastic_manageiq/__init__.py
```diff
@@ -5141,3 +5141,63 @@ def click_body_action(self):
...
+            available_items=available_items,
+            chosen_items=chosen_items,
+            move_into=move_into,
+            move_from=move_from,
+            logger=logger,
+        )
+        self.move_up = move_up
+        self.move_down = move_down
+
+    def fill(self, values):
```
jawatts: **Question** Could you use `DefaultFillViewStrategy` here? Define that in the `MultiBoxSelect` class?

mshriver: I think this is required, this (and `MultiBoxSelect`) need to use fill_strategy to define this, as it doesn't handle before_fill/after_fill calls and breaks some expectations of filling widgets.

dajoRH:
*CFME QE Bot*

widgetastic_manageiq/__init__.py
digitronik: > **Question** Could you use `DefaultFillViewStrategy` here? Define that in the `MultiBoxSelect` class?
@jawatts as per me. when we not specify any `fill_strategy` then `DefaultFillViewStrategy` will select by default. For reference please check `View` from `wt.core`.

digitronik: > I think this is required, this (and `MultiBoxSelect`) need to use fill_strategy to define this, as it doesn't handle before_fill/after_fill calls and breaks some expectations of filling widgets.

@mshriver I am confused about this. I tried to understand but fails as `before_fill` and `after_fill` are hooks; which I am not using. He we are checking order with values not just values. Can you please help me to understand this? I would like to improve this.

digitronik:
for now moving to rfr please free to move wip.

cfme/automate/buttons.py
```diff
@@ -795,19 +802,33 @@ class ButtonGroupCollection(BaseCollection):
     TEMPLATE_IMAGE = "VM Template and Image"
     VM_INSTANCE = "VM and Instance"
 
-    def instantiate(self, text, hover, type, image=None, display=None, icon_color=None):
+    def instantiate(
+        self,
+        text,
+        hover,
+        type,
+        image="fa-user",
```
mshriver: I don't think this is the correct way to define this collection ENTITY's attributes.

1. This entire function definition should be removed, as `BaseCollection.instantiate` covers it.

2. `self.type = type` on L815, we're setting the collection instance `type` attribute on instantiation of a ENTITY. We could instantiate multiple entities on this same collection instance, with different `type` passed each time. I'm not sure at the moment of the purpose of the `ButtonGroupCollection.type` attribute, but this appears to be an inconsistent implementation

3. The default for image that you're trying to define here ('fa-user') contradicts the ENTITY.image `attr.ib()` definition, which does not provide a default. If this entity attribute does not need to be passed during instantiation, then this default should be provided in the ENTITY attibute specification. (L719 above)

widgetastic_manageiq/__init__.py
```diff
@@ -5205,3 +5205,63 @@ def fill(self, value):
...
+    move_into_button = Button(title=Parameter("@move_into"))
+    move_from_button = Button(title=Parameter("@move_from"))
+    move_up_button = Button(title=Parameter("@move_up"))
+    move_down_button = Button(title=Parameter("@move_down"))
+
+    def __init__(
+        self,
+        parent,
+        available_items,
+        chosen_items,
```
mshriver: We are losing the default ID strings that are defined in `MultiBoxSelect` and used for the Parametrized locators. since there are no defaults here, and these values are passed into the parent constructor.

cfme/automate/buttons.py
digitronik: @mshriver I discuss this  thing before.... nearly  same question. https://github.com/ManageIQ/integration_tests/pull/7964#discussion_r223990458

related to `images`; here already overwrite instantiate method. attr.ib default will not effective then.

If you want I will refactor overall Button model page.... we can even remove object mapping also dependency etc. but, This  implementation working awesome and not even creating problem in automating new test.

mshriver: Do you disagree with 1 or 3 here, or just 2?  Sounds like just 2 needs some discussion, but 1 and 3 are required.

mshriver: For #2, it sounds like we need, in the collection, to store _all_ of the types of buttons that have been instantiated through the collection, not just the _last_ type that was instantiated.

dajoRH:
*CFME QE Bot*

cfme/automate/buttons.py
digitronik: Added changes

PR #8416, opened 2019-01-29
[1LP][RFR] Fix Ansible Custom Attribute Failing Test for Containers

jawatts:
{{ pytest: cfme/tests/containers/test_manageiq_ansible_custom_attributes.py --use-provider ocp-37-hawk }}

### DIFF
--- a/cfme/tests/containers/test_manageiq_ansible_custom_attributes.py
+++ b/cfme/tests/containers/test_manageiq_ansible_custom_attributes.py
@@ -36,8 +36,9 @@ def ansible_custom_attributes():
     remove_tmp_files()
 
 
-def verify_custom_attributes(provider, custom_attributes_to_verify):
+def verify_custom_attributes(appliance, provider, custom_attributes_to_verify):
     view = navigate_to(provider, 'Details')
+    appliance.server.browser.refresh()
     assert view.entities.summary('Custom Attributes').is_displayed
     for custom_attribute in custom_attributes_to_verify:
         assert (
@@ -45,7 +46,7 @@ def verify_custom_attributes(provider, custom_attributes_to_verify):
                 str(custom_attribute['value'])))
 
 
-def test_manageiq_ansible_add_custom_attributes(ansible_custom_attributes, provider):
+def test_manageiq_ansible_add_custom_attributes(appliance, ansible_custom_attributes, provider):
     """This test checks adding a Custom Attribute using Ansible script via Manage IQ module
         Steps:
         1. 'add_custom_attributes.yml script runs against the appliance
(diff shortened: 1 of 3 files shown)


dajoRH:
Lint report for commit 2c1177425a2066994783f3460ceee4bc1b4fd951:

`cfme/tests/containers/test_manageiq_ansible_custom_attributes.py`:
- :bangbang: Line 70:58: **E251** *unexpected spaces around keyword / parameter equals*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/containers/test_manageiq_ansible_custom_attributes.py
```diff
@@ -36,16 +36,17 @@ def ansible_custom_attributes():
     remove_tmp_files()
 
 
-def verify_custom_attributes(provider, custom_attributes_to_verify):
+def verify_custom_attributes(appliance, provider, custom_attributes_to_verify):
     view = navigate_to(provider, 'Details')
+    appliance.server.browser.refresh()
```
digitronik: Q:
after navigating to `Detail` page you need browser refresh for checking `custom Attributes`. Is there any BZ?

cfme/tests/containers/test_manageiq_ansible_custom_attributes.py
```diff
@@ -154,11 +163,12 @@ def test_manageiq_ansible_add_custom_attributes_bad_user(ansible_custom_attribut
     run_result = run_ansible('add_custom_attributes_bad_user')
     assert 'Authentication failed' in run_result
     view = refresh_and_navigate(provider, 'Details')
+    appliance.server.browser.refresh()
```
digitronik: same as above

cfme/tests/containers/test_manageiq_ansible_custom_attributes.py
jawatts: So this isn't so much a CloudForms problem more of a nuance with the test. This test adds and removes custom attributes via an ansible playbook. After the first navigation, the view will stay on the details page since it is already there,which causes the test to fail because new custom attributes have been added or removed via ansible. I thought about adding a `ressetter` to the Details class but thought this made more sense.

mshriver: I would actually lean toward using resetter here - any reason not to? Was it still sometimes refreshing before the information was available?

mshriver: Merging for test fixes, this can be considered and easily added to the provider nav.

PR #8421, opened 2019-01-31
[1LP][RFR] Allow currently_selected to work with trees with multiple roots

john-dupuy:
Purpose or Intent
=================

__Extending__ `ManageIQTree` such that it inherits `BootstrapTreeview` rather than being equivalent to it, this will allow us to define MIQ specific functions for the `BootstrapTreeview`s that are present in MIQ. 

I have modified the `currently_selected` property to be able to deal with trees that have more than one root. As a smoke test, I have modified the Reports Timeline view to use this method. 

{{ pytest: --long-running --use-provider vsphere6-nested cfme/tests/intelligence/test_reports_with_timelines.py }}

### DIFF
--- a/widgetastic_manageiq/__init__.py
+++ b/widgetastic_manageiq/__init__.py
@@ -68,7 +68,35 @@ class DynamicTableAddError(Exception):
 
 
 # TODO: replace below calls with direct calls later
-ManageIQTree = BootstrapTreeview
+class ManageIQTree(BootstrapTreeview):
+    """ Refactor of BootstrapTreeview to add currently_selected for times
+    which have more than 1 root. Also a root_items property."""
+
+    @property
+    def root_items(self):
+        return self.browser.elements(self.ROOT_ITEMS, parent=self)
+
+    @property
+    def currently_selected(self):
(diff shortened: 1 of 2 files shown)


john-dupuy:
Sprout provisioning error for the 5.10 check.

widgetastic_manageiq/__init__.py
```diff
@@ -68,7 +68,35 @@ class DynamicTableAddError(Exception):
 
 
 # TODO: replace below calls with direct calls later
-ManageIQTree = BootstrapTreeview
+class ManageIQTree(BootstrapTreeview):
+    """ Refactor of BootstrapTreeview to add currently_selected for times
```
jawatts: The wording here is a bit confusing

widgetastic_manageiq/__init__.py
```diff
@@ -68,7 +68,35 @@ class DynamicTableAddError(Exception):
 
 
 # TODO: replace below calls with direct calls later
```
jawatts: can this TODO be removed?

PR #8422, opened 2019-02-02
[1LP][RFR] modify fill_rsa_endpoint method and add tenant_mapping options in provider create_rest method

valaparthvi:

Changes that have been integrated from @mkoura's PR #6476:
1. Add `tenant_mapping_enabled` attribute while using create_rest method.
2. Add `_fill_amqp_endpoint_dicts`.

Other Changes:
1. Modify `_fill_rsa_endpoint_dicts` to take `rsa_keypair` endpoints if they exist, even if they're not accessible directly.

{{ pytest: cfme/tests -k "test_cloud_provider_crud or test_create_provider" --use-provider={osp13-ims,rhos13} --use-template-cache -sqvvvv }}

### DIFF
--- a/cfme/cloud/provider/openstack.py
+++ b/cfme/cloud/provider/openstack.py
@@ -3,7 +3,7 @@ from wrapanapi.systems import OpenstackSystem
 
 from . import CloudProvider
 from cfme.cloud.instance.openstack import OpenStackInstance
-from cfme.common.provider import EventsEndpoint
+from cfme.common.provider import EventsEndpoint, SSHEndpoint
 from cfme.exceptions import ItemNotFound
 from cfme.infrastructure.provider.openstack_infra import OpenStackInfraEndpointForm
 from cfme.infrastructure.provider.openstack_infra import RHOSEndpoint
@@ -95,6 +95,10 @@ class OpenStackProvider(CloudProvider):
                 logger.warning('Skipping AMQP event config due to BZ 1618700')
             else:
                 endpoints[EventsEndpoint.name] = EventsEndpoint(**event_endpoint_config)
+
+        rsa_endpoint_config = prov_config['endpoints'].get(SSHEndpoint.name, {})
+        if rsa_endpoint_config:
+            endpoints[SSHEndpoint.name] = EventsEndpoint(**rsa_endpoint_config)
 
         from cfme.utils.providers import get_crud
         infra_prov_key = prov_config.get('infra_provider_key')
--- a/cfme/cloud/provider/openstack.py
+++ b/cfme/cloud/provider/openstack.py
@@ -3,7 +3,8 @@ from wrapanapi.systems import OpenstackSystem
 
 from . import CloudProvider
 from cfme.cloud.instance.openstack import OpenStackInstance
-from cfme.common.provider import EventsEndpoint, SSHEndpoint
+from cfme.common.provider import EventsEndpoint
+from cfme.common.provider import SSHEndpoint
 from cfme.exceptions import ItemNotFound
 from cfme.infrastructure.provider.openstack_infra import OpenStackInfraEndpointForm
 from cfme.infrastructure.provider.openstack_infra import RHOSEndpoint
@@ -98,7 +99,7 @@ class OpenStackProvider(CloudProvider):
 
         rsa_endpoint_config = prov_config['endpoints'].get(SSHEndpoint.name, {})
         if rsa_endpoint_config:
-            endpoints[SSHEndpoint.name] = EventsEndpoint(**rsa_endpoint_config)
+            endpoints[SSHEndpoint.name] = SSHEndpoint(**rsa_endpoint_config)
 
         from cfme.utils.providers import get_crud
         infra_prov_key = prov_config.get('infra_provider_key')
--- a/cfme/common/provider.py
+++ b/cfme/common/provider.py
@@ -377,25 +377,17 @@ class BaseProvider(Taggable, Updateable, Navigatable, BaseEntity, CustomButtonEv
         Helper method for ``self.create_rest``
         """
         if "rsa_keypair" not in self.endpoints:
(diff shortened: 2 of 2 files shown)


cfme/common/provider.py
```diff
@@ -361,20 +365,61 @@ def _fill_rsa_endpoint_dicts(self, provider_attributes, connection_configs):
         Helper method for ``self.create_rest``
         """
         if "rsa_keypair" not in self.endpoints:
-            return
+            if "rsa_keypair" not in self.data.endpoints:
```
Yadnyawalkya: :+1: :100:

cfme/common/provider.py
```diff
@@ -361,20 +365,61 @@ def _fill_rsa_endpoint_dicts(self, provider_attributes, connection_configs):
...
+                "auth_type": "ssh_keypair",
+            }
+        )
+
+    def _fill_amqp_endpoint_dicts(self, provider_attributes, connection_configs):
+        """Fills dicts with AMQP events endpoint data.
+
+        Helper method for ``self.create_rest``
+        """
+        if "events" not in self.endpoints:
```
mshriver: I think this should be more specific, since there could be multiple types of event endpoints. Ran into this before with container endpoints.

cfme/common/provider.py
mshriver: Why is this additional check against data done here? This should just be processing the endpoints that were set when the provider object was instantiated via the BaseProvider attribute.

This is not the right way to approach handling filling this endpoint value.

`BaseProvider.endpoints` is a constructor attribute, and is set when the provider object is instantiated. This is generally via `cfme.utils.providers.get_crud` method, which will use the individual provider class `from_config` method. In the case of openstack, that's here:

https://github.com/ManageIQ/integration_tests/blob/master/cfme/cloud/provider/openstack.py#L86

This is obviously only creating a single entry in the endpoints dictionary, using the RHOSEndpoint class that's actually defined in the openstack_infra module at the next link. Note OpenstackInfraProvider has its own `from_config` method, which handles all three types of endpoints for openstack. I assume this change is being made to support the rsa endpoint for the IMS provider, which is _not_ OpenstackInfra.

https://github.com/ManageIQ/integration_tests/blob/master/cfme/infrastructure/provider/openstack_infra.py#L17

This needs to be addressed by updating the `OpenstackProvider.from_config()` method to actually process any endpoints other than `default` for openstack.  There (likely) won't be any changes to the method here, as its already processing the endpoint information in exactly the same way, you have just added references to the yaml data directly instead of the provider instance.

mshriver:
This does need to have PRT testing performed, after the yaml for the provider which you expect to setup with an RSA endpoint is available.  It needs to, at a minimum, pass a CRUD test for that provider that includes a credential check.

cfme/common/provider.py
Yadnyawalkya: @valaparthvi  I mistakenly deleted my comment which resulted in deletion of @mshriver's following comment (really sorry for that); Mike said, 

"Why is this additional check against data done here? This should just be processing the endpoints that were set when the provider object was instantiated via the BaseProvider attribute.

This is not the right way to approach handling filling this endpoint value.

BaseProvider.endpoints is a constructor attribute, and is set when the provider object is instantiated. This is generally via cfme.utils.providers.get_crud method, which will use the individual provider class from_config method. In the case of openstack, that's here:

https://github.com/ManageIQ/integration_tests/blob/master/cfme/cloud/provider/openstack.py#L86

This is obviously only creating a single entry in the endpoints dictionary, using the RHOSEndpoint class that's actually defined in the openstack_infra module at the next link. Note OpenstackInfraProvider has its own from_config method, which handles all three types of endpoints for openstack. I assume this change is being made to support the rsa endpoint for the IMS provider, which is not OpenstackInfra.

https://github.com/ManageIQ/integration_tests/blob/master/cfme/infrastructure/provider/openstack_infra.py#L17

This needs to be addressed by updating the OpenstackProvider.from_config() method to actually process any endpoints other than default for openstack. There (likely) won't be any changes to the method here, as its already processing the endpoint information in exactly the same way, you have just added references to the yaml data directly instead of the provider instance."

cfme/cloud/provider/openstack.py
```diff
@@ -96,6 +96,10 @@ def from_config(cls, prov_config, prov_key):
             else:
                 endpoints[EventsEndpoint.name] = EventsEndpoint(**event_endpoint_config)
 
+        rsa_endpoint_config = prov_config['endpoints'].get(SSHEndpoint.name, {})
+        if rsa_endpoint_config:
+            endpoints[SSHEndpoint.name] = EventsEndpoint(**rsa_endpoint_config)
```
mshriver: This should be SSHEndpoint instance no? (typo?)

cfme/common/provider.py
valaparthvi: Even if there are multiple types of event endpoints, it does a further check to make sure that the event is `amqp`. Shouldn't that be fine?

mshriver:
I think that at a minimum, `test_cloud_provider_crud` should be used for PRT, as we have CFME yaml file with an rsa endpoint defined. 

rhos11 is unusable, looks like it was the only cloud provider with ampq endpoint defined. The ampq endpoint helper needs tested in PRT before merge. Can you update osp14-ims to include the events endpoint, or work with @mnadeem92 to get one defined in the yamls for the new rhos13 deployment?

Then the CRUD test can at least be used to validate all of the endpoints working.

https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/cloud/test_providers.py#L172

cfme/cloud/provider/openstack.py
valaparthvi: Huge typo! :roll_eyes:

valaparthvi:
@mshriver Do you think additional checks must be added to the test to make sure endpoints other than the default are passed and validated?

PR #8425, opened 2019-02-04
[WIPTEST] added uncollection for test_provider_log_level

lina-nikiforova:
{{ pytest: -v cfme/tests/configure/test_logs.py::test_provider_log_level --use-provider=complete }}

Here I removed blocker for BZ1640718 as this is verified for `5.10`. However, this "bug" will not be fixed for `5.9` so I added `uncollectif` for this case.

Other changes happened due to black formatting.


dajoRH:
*CFME QE Bot*

PR #8426, opened 2019-02-05
[RFR] Fixed AttributeError for AddCatalogItemView

ganeshhubale:
{{ pytest: cfme/tests/automate/test_git_import.py::test_automate_git_domain_displayed_in_service -v }}

Fixed Error:
```
>       view.field_entry_point.fill("")
E       AttributeError: 'AddCatalogItemView' object has no attribute 'field_entry_point'

cfme/tests/automate/test_git_import.py:66: AttributeError
AttributeError
'AddCatalogItemView' object has no attribute 'field_entry_point'
```

### DIFF
--- a/cfme/services/catalogs/catalog_items/__init__.py
+++ b/cfme/services/catalogs/catalog_items/__init__.py
@@ -31,6 +31,14 @@ class EntryPoint(Input):
             return True
         return False
 
+    def click(self):
+        # After clicking on widget - 'retirement_entry_point', 'reconfigure_entry_point' or
+        # 'provisioning_entry_point'; 'Select Entry Point Instance'(pop up) occures which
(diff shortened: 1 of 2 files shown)


cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -31,6 +31,14 @@ def fill(self, value):
             return True
         return False
 
+    def click(self):
```
This is not way to go for this. I would inherit `ClickableMixin` form `wt.core`. It will provide you `click()` method.

so, try like this.
```
class EntryPoint(Input, ClickableMixin):
```

Let me, its work or not.

ganeshhubale: Yes, It is providing ```click()``` method.
Thanks :)

PR #8431, opened 2019-02-05
[WIPTEST] Remove old BZ, update test_restart_workers

mshriver:
Removing closed BZ, updating docblock


jawatts:
Note: False positive for 5.10, test is skipping. Working for 5.9

PR #8434, opened 2019-02-06
[1LP][RFR] CB: Simulation

digitronik:
Purpose or Intent
=================

{{pytest: cfme/tests/automate/custom_button/test_buttons.py::test_custom_button_simulation -v}}

### DIFF
--- a/cfme/automate/buttons.py
+++ b/cfme/automate/buttons.py
@@ -16,6 +16,7 @@ from widgetastic.widget import (
 from widgetastic.xpath import quote
 from widgetastic_patternfly import BootstrapSelect, Button, CandidateNotFound, Input
 
+from cfme.base.ui import AutomateSimulationView
 from cfme.modeling.base import BaseCollection, BaseEntity
 from cfme.utils.appliance.implementations.ui import CFMENavigateStep, navigate_to, navigator
 from cfme.utils.update import Updateable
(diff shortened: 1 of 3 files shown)


digitronik:
Added log check as well...

cfme/tests/automate/custom_button/test_buttons.py
```diff
@@ -351,7 +366,44 @@ def test_custom_button_simulation():
...
+    button = gp.buttons.create(
+        text="Btn_{}".format(fauxfactory.gen_alphanumeric(2)),
+        hover="Hover_{}".format(fauxfactory.gen_alphanumeric(2)),
+        system="Request",
+        request="InspectMe",
+    )
+    request.addfinalizer(button.delete_if_exists)
+
+    # Clear the automation log
+    assert appliance.ssh_client.run_command('echo -n "" > /var/www/miq/vmdb/log/automation.log')
```
Yadnyawalkya: :+1: :100:

cfme/tests/automate/custom_button/test_buttons.py
```diff
@@ -351,7 +366,44 @@ def test_custom_button_simulation():
     Bugzilla:
         1535215
     """
-    pass
+    if button_tag == "Evm":
```
Yadnyawalkya: @digitronik, you are only checking for EVM and provider object type, right? LGTM.

digitronik: Yes, I don't think we need to check all object... `Simulation` mainly cover by `Automate` area. In case of `custom button` `evm` and `build` these are major type. I would like to take one one test from those.

cfme/automate/buttons.py
```diff
@@ -16,6 +16,7 @@
 from widgetastic.xpath import quote
 from widgetastic_patternfly import BootstrapSelect, Button, CandidateNotFound, Input
 
+from cfme.base.ui import AutomateSimulationView
```
sshveta: Question : Any idea why is AutomateSimulationView in base.ui file ? There is simulation.py in automate folder . 
Not in this PR but I think it should be moved to respective automate classes

cfme/tests/automate/custom_button/test_buttons.py
mshriver: This is ultimately a bit of a hack - there are tools for dealing with the logs, and grepping for certain content. It automatically handles only checking the log during the test function execution. Okay for now as `log_request_check` already exists, but it should be replaced.

cfme/automate/buttons.py
digitronik: @sshveta: I agreed with your point. But I try to avoid changes in others FAs without permission. As per my understanding it is old stuff. I will take a look on this thanks for pointing this.

PR #8437, opened 2019-02-06
[1LP][RFR][NOTEST] Remove BZ meta and unassign a test case

valaparthvi:
1. Unassign a test case: `test_provisioning`
2. Remove BZ meta - not required anymore.

### DIFF
--- a/cfme/intelligence/reports/reports.py
+++ b/cfme/intelligence/reports/reports.py
@@ -324,10 +324,7 @@ class Report(BaseEntity, Updateable):
 
     @company_name.default
     def company_name_default(self):
-        if self.appliance.version < "5.9":
-            return "My Company (All EVM Groups)"
-        else:
-            return "My Company (All Groups)"
+        return "My Company (All Groups)"
 
     def update(self, updates):
         view = navigate_to(self, "Edit")
@@ -336,7 +333,7 @@ class Report(BaseEntity, Updateable):
             view.save_button.click()
         else:
             view.cancel_button.click()
-        view = self.create_view(ReportDetailsView, override=updates, wait="60s")
+        view = self.create_view(ReportDetailsView, override=updates, wait="5s")
         view.flash.assert_no_error()
         if changed:
             view.flash.assert_message(
--- a/cfme/intelligence/reports/reports.py
+++ b/cfme/intelligence/reports/reports.py
@@ -336,7 +336,7 @@ class Report(BaseEntity, Updateable):
             view.save_button.click()
         else:
             view.cancel_button.click()
-        view = self.create_view(ReportDetailsView, override=updates, wait="5s")
+        view = self.create_view(ReportDetailsView, override=updates, wait="60s")
         view.flash.assert_no_error()
         if changed:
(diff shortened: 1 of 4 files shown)


cfme/intelligence/reports/reports.py
```diff
@@ -330,7 +330,7 @@ def update(self, updates):
             view.save_button.click()
         else:
             view.cancel_button.click()
-        view = self.create_view(ReportDetailsView, override=updates, wait="5s")
+        view = self.create_view(ReportDetailsView, override=updates, wait="60s")
```
john-dupuy: @valaparthvi Why does increasing the `wait` time allow your test to pass?

cfme/tests/perf/workloads/test_provisioning.py
```diff
@@ -77,9 +77,7 @@ def test_provisioning(appliance, request, scenario):
     Memory Monitor creates graphs and summary at the end of each scenario.
 
     Polarion:
-        assignee: pvala
-        casecomponent: Rest
-        caseimportance: high
+        assignee: rhcf3_machine
```
john-dupuy: Why are you un-assigning this test case?

cfme/intelligence/reports/reports.py
valaparthvi: I was getting a `NoSuchElementException` for `Report Info` tab which appears after the report is edited. Giving it some time, helped in avoiding the error.

cfme/tests/perf/workloads/test_provisioning.py
valaparthvi: It is a performance-related test case, which I had mistakenly assigned to myself, so assigning it back rhcf3_machine.

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -256,17 +256,7 @@ def verify():
...
-        BZ(
-            1662638,
-            forced_streams=["5.9", "5.10", "upstream"],
-            unblock=lambda provider: provider.one_of(VMwareProvider)
-            and provider.version == "6.5",
-        )
-    ]
-)
-@pytest.mark.parametrize('auto', [True, False], ids=["Auto", "Manual"])
+@pytest.mark.parametrize('auto', [False], ids=["Manual"])
```
lina-nikiforova: Why did you remove "Auto" and "True" parametrization?

valaparthvi: Oh shoot! Committed that by mistake. Had changed it for local testing.

cfme/intelligence/reports/reports.py
jawatts: @valaparthvi What if you did a browser refresh after creating the view instead waiting 60s, would that solve the issue? Seems like a long wait.

valaparthvi: Okay, I'll see if browser refresh helps, or if there's some other way.

digitronik:
@valaparthvi in such case best to run tests. It helps to check status.

PR #8443, opened 2019-02-07
[1LP][RFR] This test for checking invalid name of namespaces

ganeshhubale:
{{ pytest: cfme/tests/automate/ -k 'test_wrong_namespace_name or test_wrong_domain_name' -v }}

### DIFF
--- a/cfme/tests/automate/test_domain.py
+++ b/cfme/tests/automate/test_domain.py
@@ -173,8 +173,10 @@ def test_domain_name_wrong(appliance):
         initialEstimate: 1/60h
         tags: automate
     """
-    with pytest.raises(Exception, match='Name may contain only'):
-        appliance.collections.domains.create(name='with space')
+    view = navigate_to(appliance.collections.domains, 'Add')
+    view.name.fill('Dummy Domain')
+    view.add_button.click()
+    view.flash.assert_message('Name may contain only alphanumeric and _ . - $ characters')
 
 
 @pytest.mark.tier(2)
--- a/cfme/tests/automate/test_domain.py
+++ b/cfme/tests/automate/test_domain.py
@@ -163,7 +163,7 @@ def test_domain_cannot_edit_builtin(appliance):
 
 
 @pytest.mark.tier(2)
-def test_domain_name_wrong(appliance):
+def test_domain_name_wrong(request, appliance):
     """
     Polarion:
         assignee: ghubale
@@ -173,10 +173,15 @@ def test_domain_name_wrong(appliance):
         initialEstimate: 1/60h
         tags: automate
     """
-    view = navigate_to(appliance.collections.domains, 'Add')
-    view.name.fill('Dummy Domain')
+    wrong_domain = 'Dummy Domain'
+    domain = appliance.collections.domains
+    view = navigate_to(domain, 'Add')
+    view.name.fill(wrong_domain)
     view.add_button.click()
     view.flash.assert_message('Name may contain only alphanumeric and _ . - $ characters')
+    wrong_domain = domain.instantiate(name=wrong_domain)
+    request.addfinalizer(wrong_domain.delete_if_exists)
+    assert not wrong_domain.exists
 
 
 @pytest.mark.tier(2)
--- a/cfme/tests/automate/test_domain.py
+++ b/cfme/tests/automate/test_domain.py
@@ -3,6 +3,7 @@ import fauxfactory
 import pytest
 
 from cfme import test_requirements
+from cfme.automate.explorer.domain import DomainAddView
 from cfme.automate.import_export import AutomateGitRepository
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.update import update
@@ -178,9 +179,9 @@ def test_wrong_domain_name(request, appliance):
     """
     wrong_domain = 'Dummy Domain'
     domain = appliance.collections.domains
-    view = navigate_to(domain, 'Add')
-    view.name.fill(wrong_domain)
-    view.add_button.click()
+    with pytest.raises(AssertionError):
+        domain.create(name=wrong_domain)
+    view = domain.create_view(DomainAddView)
     view.flash.assert_message('Name may contain only alphanumeric and _ . - $ characters')
     wrong_domain = domain.instantiate(name=wrong_domain)
     request.addfinalizer(wrong_domain.delete_if_exists)
--- a/cfme/tests/automate/test_domain.py
+++ b/cfme/tests/automate/test_domain.py
@@ -163,8 +163,11 @@ def test_domain_cannot_edit_builtin(appliance):
 
 
 @pytest.mark.tier(2)
-def test_domain_name_wrong(request, appliance):
-    """
+def test_wrong_domain_name(request, appliance):
+    """To test whether domain is creating with wrong name or not.
+       wrong_domain: 'Dummy Domain' (This is invalid name of Domain because there is space
+       in the name)
+
     Polarion:
         assignee: ghubale
         casecomponent: Automate
--- a/cfme/tests/automate/test_namespace.py
+++ b/cfme/tests/automate/test_namespace.py
@@ -85,3 +85,25 @@ def test_duplicate_namespace_disallowed(parent_namespace):
         parent_namespace.namespaces.create(
             name=ns.name,
             description=ns.description)
+
+
+@pytest.mark.tier(2)
+def test_namespace_name_wrong(appliance, domain):
+    """
+    Polarion:
+        assignee: ghubale
+        casecomponent: Automate
+        caseimportance: medium
+        caseposneg: negative
+        initialEstimate: 1/60h
+        tags: automate
+        testSteps:
+            1. Navigate to Automation> Automate> Explorer
+            2. Try to create namespace with name `Dummy Domain` (I put space which is invalid)
+            3. Should give proper flash message
+
+    Bugzilla:
+        1650071
+    """
+    with pytest.raises(Exception, match='Name may contain only'):
+        domain.namespaces.create(name='with space')
--- a/cfme/tests/automate/test_namespace.py
+++ b/cfme/tests/automate/test_namespace.py
@@ -3,6 +3,7 @@ import fauxfactory
 import pytest
 
 from cfme import test_requirements
+from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.update import update
 
 pytestmark = [test_requirements.automate]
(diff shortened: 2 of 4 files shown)


dajoRH:
Lint report for commit d7ccbcb26dc22afafaa7a9296aac3e2a1cbf5ca7:

`cfme/tests/automate/test_domain.py`:
- :red_circle: Line 305:7: **N801** *class names should use CapWords convention*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
Lint report for commit 29bda56f9547c34a1400820c3ca9d372ae779059:

`cfme/tests/automate/test_domain.py`:
- :red_circle: Line 305:7: **N801** *class names should use CapWords convention*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/automate/test_namespace.py
```diff
@@ -94,3 +94,25 @@ def test_duplicate_namespace_disallowed(request, parent_namespace):
...
+        assignee: ghubale
+        casecomponent: Automate
+        caseimportance: medium
+        caseposneg: negative
+        initialEstimate: 1/60h
+        tags: automate
+        testSteps:
+            1. Navigate to Automation> Automate> Explorer
+            2. Try to create namespace with name `Dummy Domain` (I put space which is invalid)
+            3. Should give proper flash message
```
jawatts: **Required** Please separate this between `testSteps` and `expectedResults` 

So:
testSteps:
            1. Navigate to Automation> Automate> Explorer
            2. Try to create namespace with name `Dummy Domain` (I put space which is invalid)
expectedResults:
            1.
            2. Should give proper flash message

cfme/tests/automate/test_namespace.py
```diff
@@ -94,3 +94,25 @@ def test_duplicate_namespace_disallowed(request, parent_namespace):
...
+
+    Bugzilla:
+        1650071
+    """
+    with pytest.raises(Exception, match='Name may contain only'):
+        domain.namespaces.create(name='with space')
```
jawatts: **Required** I think it be better to not use the `DomainCollection` class `create()` method in this case. It would be a better test of the GUI to navigate to the create view, fill it (with the bad name), and perform a `view.flash.assert_message()` on the view. Then verify the domain was not created. What do you think?

ganeshhubale: @jawatts  Thank for suggestion.
I am agree with you.
I have modified this test case and  ```test_domain_name_wrong``` in this way.
Please take a look and let me know if it needs more improvements.

cfme/tests/automate/test_domain.py
```diff
@@ -182,8 +182,10 @@ def test_domain_name_wrong(appliance):
         initialEstimate: 1/60h
         tags: automate
     """
-    with pytest.raises(Exception, match='Name may contain only'):
-        appliance.collections.domains.create(name='with space')
+    view = navigate_to(appliance.collections.domains, 'Add')
+    view.name.fill('Dummy Domain')
+    view.add_button.click()
+    view.flash.assert_message('Name may contain only alphanumeric and _ . - $ characters')
```
jawatts: **Required** I think it would be good to extend this even more and verify that the `Dummy Domain` was not added even though a flash error appeared by checking the all view.

cfme/tests/automate/test_namespace.py
```diff
@@ -94,3 +95,29 @@ def test_duplicate_namespace_disallowed(request, parent_namespace):
...
+            1.
+            2. Should give proper flash message
+
+    Bugzilla:
+        1650071
+    """
+    view = navigate_to(domain.namespaces, 'Add')
+    view.name.fill('Dummy Namespace')
+    view.add_button.click()
+    view.flash.assert_message('Name may contain only alphanumeric and _ . - $ characters')
```
jawatts: Same comment as above.

cfme/tests/automate/test_domain.py
ganeshhubale: @jawatts 
According to me, 
There are two ways to check if ```dummy domain``` is exists or not.
1. By instantiating domain object with name: ```dummy domain```
```
wrong_domain = domain.instantiate(name=wrong_domain)
assert not wrong_domain.exists
```
2. Checking in list of domains on all page
```
view = navigate_to(domain, 'All')
all_domains = view.domains.read()
for i in range(len(all_domains)):
     assert wrong_domain != all_domains[i]['Name']
```
 I have updated PR with first way. 
Please let me know, if it needs more changes.

cfme/tests/automate/test_domain.py
```diff
@@ -182,8 +182,15 @@ def test_domain_name_wrong(appliance):
         initialEstimate: 1/60h
         tags: automate
     """
-    with pytest.raises(Exception, match='Name may contain only'):
-        appliance.collections.domains.create(name='with space')
+    wrong_domain = 'Dummy Domain'
```
digitronik: @ganeshhubale please add comment or steps in `doc-string`. as you took "Dummy Domain" other contributes  don't know that its not going to allow space.

cfme/tests/automate/test_domain.py
```diff
@@ -182,8 +182,15 @@ def test_domain_name_wrong(appliance):
...
+    domain = appliance.collections.domains
+    view = navigate_to(domain, 'Add')
```
digitronik: Optional:
@ganeshhubale I think better to go with create domain and use `pytest.raise(AssiersionError)`. As `flash.assert_no_error()` should raise this error. This comment totally optional.

cfme/tests/automate/test_domain.py
```diff
@@ -172,7 +172,7 @@ def test_domain_cannot_edit_builtin(appliance):
 
 
 @pytest.mark.tier(2)
-def test_domain_name_wrong(appliance):
+def test_domain_name_wrong(request, appliance):
```
sshveta: Test name  should be test_wrong_domain_name()

cfme/tests/automate/test_domain.py
sshveta: I agree with this ^ Ganesh . 
Please create domain using collection 

     domain = appliance.collections.domains.create(name=wrong_domain, description=fauxfactory.gen_alpha(),
                                                                                            enabled=enabled)
and then check for flash mesage
and 
assert not  domain.exists

cfme/tests/automate/test_namespace.py
```diff
@@ -94,3 +95,32 @@ def test_duplicate_namespace_disallowed(request, parent_namespace):
...
+        casecomponent: Automate
+        caseimportance: medium
+        caseposneg: negative
+        initialEstimate: 1/60h
+        tags: automate
+        testSteps:
+            1. Navigate to Automation> Automate> Explorer
+            2. Try to create namespace with name `Dummy Domain` (I put space which is invalid)
+        expectedResults:
+            1.
```
sshveta: Remove Empty step 1 .

cfme/tests/automate/test_namespace.py
sshveta: Same as above , please create domain using collection then just check for flash message( by creating view) and  assert namespace not exist

cfme/tests/automate/test_domain.py
ganeshhubale: @digitronik @sshveta 
- I am agree with both of you.
- Please see this discussion, as per [this comment](https://github.com/ManageIQ/integration_tests/pull/8443#discussion_r259083403) by @jawatts ; I too think that it's good to go with GUI testing while creating domain/namespace.
- Also sometimes what happens is that, while creating ```class``` with invalid name it gives flash message like this: **Error during 'add': Validation failed: MiqAeClass: Name may contain only alphanumeric and _ . - characters**
- As I see this is error, It might be bug. So I have added comment for it in [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1650071#c14)
- Previously it was as you both said.
- I would be happy to update with suggestions from all of you @sshveta @digitronik @jawatts 

Please let me know, which one is best way for this test case.

Thank you,
Ganesh

cfme/tests/automate/test_namespace.py
ganeshhubale: @sshveta 
As I see [suggestions](https://github.com/ManageIQ/integration_tests/pull/8443#discussion_r259077545) from @jawatts @mshriver .
This is the way of writing ```expectedResults```.

dajoRH:
*CFME QE Bot*

cfme/tests/automate/test_domain.py
jawatts: @ganeshhubale @digitronik @sshveta So I agree with @digitronik @sshveta if we were not checking for a specific error message. It seems like we have a BZ around the error message, so I would argue we should explicitly be checking the flash message in this case and not just an `AssertionError`

cfme/tests/automate/test_domain.py
jawatts: Option 1 here is good by me, thanks for adding that in

cfme/tests/automate/test_namespace.py
```diff
@@ -85,3 +86,35 @@ def test_duplicate_namespace_disallowed(parent_namespace):
...
+
+    Bugzilla:
+        1650071
+    """
+    wrong_namespace = 'Dummy Namespace'
+    view = navigate_to(domain.namespaces, 'Add')
+    view.name.fill(wrong_namespace)
+    view.add_button.click()
+    view.flash.assert_message('Name may contain only alphanumeric and _ . - $ characters')
+    wrong_namespace = domain.namespaces.instantiate(name=wrong_namespace)
```
jawatts: I like what you did in the previous test with the finalizer, we should put that here as well. If that namespace does get created, we want to clean it up.

cfme/tests/automate/test_domain.py
digitronik: My view on this:
@ganeshhubale @jawatts @sshveta really its debatable matter. As per me it should contributor privilege so I added `Optional` comment. I don't want to manipulate his/her way of codding as both way are correct as per me.

@sshveta your comment is also valid:
1. Its negative test case.... we are expecting some error mainly `AssertionError` due to `assert_no_error()`. I remember my patch with `Mike` guidance that this method should allow only error not `warning` or other.
2. As `pytest.raise` capture error it will come out of method and ready for other operation. So @sshveta suggestion means asserting exact flash will work as `flash` is part of `BaseEntityView` we can catch from any view.

jawatts: @digitronik agree with you here. 

@ganeshhubale I think the takeaway here for you, code it how you feel is correct (both ways here are correct). But we should check the explicit flash message and make sure the object is not created.

ganeshhubale: Thanks for suggestions @digitronik @sshveta @digitronik 
I will check ```AssertionError``` using ```pytest.raise``` then will check ```flash``` message and ```domain.exist``` or not.

PR #8446, opened 2019-02-08
[RFR] fix rhevm template scripts

izapolsk:
template test and upload scripts for rhevm are broken. Fixing that issue

### DIFF
--- a/requirements/frozen.py3.txt
+++ b/requirements/frozen.py3.txt
@@ -204,7 +204,7 @@ oslo.i18n==3.22.1
 oslo.serialization==2.28.1
 oslo.utils==3.37.1
 ovirt-engine-sdk-python==4.3.0
-packaging==18.0
+packaging==19.0
 pandocfilters==1.4.2
 paramiko==2.4.2
 parsedatetime==2.4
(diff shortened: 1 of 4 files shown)


requirements/frozen.py3.txt
```diff
@@ -218,6 +216,7 @@ pickleshare==0.7.5
 pika==0.12.0
 Pillow==5.3.0
 pluggy==0.6.0
+polarion-docstrings==0.15
```
mshriver: We can pickup `0.16`

requirements/frozen_docs.py2.txt
```diff
@@ -122,6 +123,7 @@ pyOpenSSL==18.0.0
 pyparsing==2.2.2
 pytesseract==0.2.5
 pytest==3.4.1
+pytest-polarion-collect==0.10
```
mshriver: We can pickup 0.11

requirements/frozen.py3.txt
izapolsk: freeze_all script picked up this version

requirements/frozen_docs.py2.txt
izapolsk: same here ^^

PR #8452, opened 2019-02-11
[1LP][RFR] Fix No providers available skip message.

jaryn:
The propety of the provider class is not "dereferenced" when
accessed from class, so we get something like this in the skip message:

```
In [2]: VMwareProvider.type
Out[2]: <property at 0x7fc9090b0470>
```
which is not too helpful. It is better to use the type_name

{ py.test: cfme/tests/cli/test_appliance_console_db_restore.py::test_appliance_console_restore_pg_basebackup_replicated }

### DIFF
--- a/cfme/fixtures/cli.py
+++ b/cfme/fixtures/cli.py
@@ -346,7 +346,7 @@ def provider_app_crud(provider_class, appliance):
         prov.appliance = appliance
         return prov
     except IndexError:
-        pytest.skip("No {} providers available (required)".format(provider_class.type))
+        pytest.skip("No {} providers available (required)".format(provider_class.type_name))
 
 
 def provision_vm(request, provider):


cfme/fixtures/cli.py
```diff
@@ -346,7 +346,7 @@ def provider_app_crud(provider_class, appliance):
         prov.appliance = appliance
         return prov
     except IndexError:
-        pytest.skip("No {} providers available (required)".format(provider_class.type))
+        pytest.skip("No {} providers available (required)".format(provider_class.type_name))
```
john-dupuy: Nice catch! LGTM :+1:

digitronik:
@jaryn nice catch :+1: we have `4 fixtures` which using this `method`. Please run at least one test which is using this anyone of this fixture. It will be great.

jaryn:
@digitronik As the code I touched is executed only when list_providers_by_class(provider_class) is [] there is no use of running this method with any fixture if I don't somehow make it []. I don't know how.

PR #8455, opened 2019-02-11
[1LP][RFR][NOTEST] Reassign generic object test and remove manual test case

john-dupuy:
Assigning `test_generic_object_definition_crud` to myself and removing a manual test case that is automated. 

{{ pytest: --long-running cfme/tests/generic_objects/test_definitions.py }}

### DIFF
--- a/cfme/tests/generic_objects/test_definitions.py
+++ b/cfme/tests/generic_objects/test_definitions.py
@@ -14,12 +14,15 @@ pytestmark = [test_requirements.generic_objects]
 
 
 @pytest.mark.sauce
+@pytest.mark.tier(0)
 @pytest.mark.parametrize('context', [ViaREST, ViaUI])
 def test_generic_object_definition_crud(appliance, context, soft_assert):
     """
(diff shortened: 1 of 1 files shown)


cfme/tests/generic_objects/test_definitions.py
```diff
@@ -14,11 +14,14 @@
...
+@pytest.mark.tier(0)
 @pytest.mark.parametrize('context', [ViaREST, ViaUI])
 def test_generic_object_definition_crud(appliance, context, soft_assert):
     """
     Polarion:
-        assignee: dmisharo
+        assignee: jdupuy
+        casecomponent: GenericObjects
+        caseimportance: high
         initialEstimate: 1/30h
```
digitronik: Are you sure about it (2 min)?

john-dupuy: It is quick since it is just creating a generic object via Rest and the UI. Maybe ~5 min is closer, I can change it.

PR #8459, opened 2019-02-12
[WIPTEST] Fix pytest_generate_tests for test_vm_instance_analysis

jawatts:
{{ pytest: cfme/tests/cloud_infra_common/test_vm_instance_analysis.py --use-provider complete --collect-only --long-running }}

### DIFF
--- a/cfme/tests/cloud_infra_common/test_vm_instance_analysis.py
+++ b/cfme/tests/cloud_infra_common/test_vm_instance_analysis.py
@@ -98,6 +98,7 @@ ssa_expect_files = [
 
 
 def pytest_generate_tests(metafunc):
+
     argnames, argvalues, idlist = testgen.providers_by_class(
         metafunc, [CloudProvider, InfraProvider], required_fields=['vm_analysis_new'])
     argnames.append('analysis_type')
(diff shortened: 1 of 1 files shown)


cfme/tests/cloud_infra_common/test_vm_instance_analysis.py
```diff
@@ -106,11 +107,18 @@ def pytest_generate_tests(metafunc):
...
-            new_argvalues.append([args["provider"], vm_analysis_key])
+        if 'vms' in vma_data:
+            vms = vma_data.vms
+            for vm_analysis_key in vms:
+                # Set VM name here
+                new_idlist.append('{}-{}'.format(idlist[index], vm_analysis_key))
+                new_argvalues.append([args["provider"], vm_analysis_key])
+        else:
+            # TODO FIXME
+            logger.error("Provider {p} does not have the correct keys in the yaml"
```
mshriver: When using logger, use `%s` in the string and pass the value to be formatted as a 2nd argument.

sshveta: @mshriver : can you please tell why to use %s in logger

mshriver: This is a build-in facility in the python logger that we're taking advantage of, instead of making our own calls to format().

https://docs.python.org/2/library/logging.html#logging.Logger.debug

mshriver:
Verified this does not impact test collection for the module.

cfme/tests/cloud_infra_common/test_vm_instance_analysis.py
sshveta: :+1:

PR #8460, opened 2019-02-12
[1LP][RFR] Fix for test_custom_button_expression by allowing tags to be read

john-dupuy:
{{ pytest: --long-running cfme/tests/services/test_myservice.py -vvv }}

Fixing `test_custom_button_expression` which was facing this error:

```
>           tag_table = entities.summary('Smart Management')
E           AttributeError: 'GridView' object has no attribute 'summary'

cfme/common/__init__.py:433: AttributeError
AttributeError
'GridView' object has no attribute 'summary'
```

### DIFF
--- a/cfme/tests/automate/custom_button/test_service_objects.py
+++ b/cfme/tests/automate/custom_button/test_service_objects.py
@@ -408,7 +408,7 @@ def test_custom_button_expression(appliance, context, objects, button_group, vis
             obj.remove_tag(tag)
 
     with appliance.context.use(context):
-        view = navigate_to(obj, dest_name)
+        view = navigate_to(obj, dest_name, wait_for_view=15)
         custom_button_group = (
             CustomButtonSSUIDropdwon(view, group.text)
             if context is ViaSSUI
--- a/cfme/services/myservice/ui.py
+++ b/cfme/services/myservice/ui.py
@@ -104,22 +104,20 @@ class MyServiceDetailsToolbar(MyServiceToolbar):
 
 class MyServiceDetailsEntities(View):
     """Represents Details page."""
+    # TODO: use a ParametrizedSummaryTable eventually to do this
     summary = ParametrizedView.nested(ParametrizedSummaryTable)
+    properties = SummaryTable(title='Properties')
+    lifecycle = SummaryTable(title='Lifecycle')
+    relationships = SummaryTable(title='Relationships')
+    vm = SummaryTable(title='Totals for Service VMs ')
+    smart_management = SummaryTable(title='Smart Management')
+    generic_objects = SummaryTable(title='Generic Objects')
 
 
 class MyServiceDetailView(MyServicesView):
     title = Text('#explorer_title_text')
     toolbar = View.nested(MyServiceDetailsToolbar)
-    # entities = View.nested(BaseNonInteractiveEntitiesView)  FIXME need new entities widget here
-
-    @View.nested
-    class details(View):  # noqa
-        properties = SummaryTable(title='Properties')
-        lifecycle = SummaryTable(title='Lifecycle')
-        relationships = SummaryTable(title='Relationships')
-        vm = SummaryTable(title='Totals for Service VMs ')
-        smart_mgmt = SummaryTable(title='Smart Management')
-        generic_objects = SummaryTable(title='Generic Objects')
+    entities = View.nested(MyServiceDetailsEntities)
 
     @View.nested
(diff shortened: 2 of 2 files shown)


cfme/common/__init__.py
```diff
@@ -429,8 +429,11 @@ def get_tags(self, tenant="My Company Tags"):
         entities = view.entities
         if hasattr(entities, 'smart_management'):
             tag_table = entities.smart_management
+        elif hasattr(view.details, 'smart_mgmt'):
```
mshriver: I think it would be better to pick one name for this widget and stick with it. Moving all of the views to use ParametrizedSummaryTable is a bigger change, but I think that would be the optimal solution.

We're adding band-aids by doing this, and enabling poor consistency. While this change _works_, I would much much rather see work put in to make the views (and tests) consistent.

mshriver: If we're going to accept this as a temporary 'fix' for the test failure, I want to see a FIXME here at a minimum. But I would rather just address the non-consistent views _now_.

john-dupuy: I definitely agree with this, I will start working on ensuring consistency across our views and tests for tag_tables. :+1:

mshriver: Moving back to WIP since the PR will change significantly. Thanks @john-dupuy!

john-dupuy:
@digitronik can you please check to see if my changes will impact tests elsewhere? I looked in the services tests and don't think that any of them will be impacted, but you know this area better than I do.

cfme/tests/automate/custom_button/test_service_objects.py
```diff
@@ -406,6 +406,7 @@ def test_custom_button_expression(appliance, context, objects, button_group, vis
 
     with appliance.context.use(context):
         view = navigate_to(obj, dest_name)
+        view.wait_displayed()
```
digitronik: Please remove this no need of it. Navigation will take care of this.

jawatts: Unless it will take longer than 10 seconds for the view. I don't think it hurts to have this here.

john-dupuy: While testing locally I noticed that sometimes the test would fail because the page was not fully displayed, this seemed to help that issue.

john-dupuy:
@digitronik/reviewers check test failure reports for the PRT runs, these appear to be unrelated to the changes that I've made in this PR.

cfme/tests/automate/custom_button/test_service_objects.py
digitronik: @jawatts yup `10s` is my point.

@john-dupuy let me help you to understand our `navigation`. @izapolsk and @mshriver already provided us facility to deal with such navigation problems.
https://github.com/ManageIQ/integration_tests/blob/master/cfme/utils/appliance/implementations/ui.py#L518
so default `wait_for_view` is `10s`. `view.wait_displayed` default value is `10s` so no need of extra `wait_for_view`. Yup its not hurting.

In your case if you are facing `lag` in view display while navigation. As per me you can try:
```
view = navigate_to(obj, dest_name, wait_for_view=10+)
```

dajoRH:
*CFME QE Bot*

cfme/tests/automate/custom_button/test_service_objects.py
john-dupuy: @digitronik thanks for the response. I knew about the `wait` arg in `create_view` but was unaware about the `wait_for_view` arg in `navigate_to`. I will update this PR to use this!

sshveta:
@john-dupuy : can you also run test_myservice.py once with these changes .
Just want to make sure this doesn't break anything there.

cfme/tests/automate/custom_button/test_service_objects.py
```diff
@@ -405,7 +405,7 @@ def test_custom_button_expression(appliance, context, objects, button_group, vis
             obj.remove_tag(tag)
 
     with appliance.context.use(context):
-        view = navigate_to(obj, dest_name)
+        view = navigate_to(obj, dest_name, wait_for_view=10)
```
digitronik: `10` is default value

john-dupuy:
@sshveta running tests against that file now

sshveta:
> @sshveta running tests against that file now

Results looks good , One failure in myservice is not related to this change.

PR #8462, opened 2019-02-12
[1LP][RFR] Add *args and **kwargs to All step() and resetter() methods

jawatts:
Add *args and **kwargs to All step() and resetter() methods that were missing them. 

{{ pytest: --long-running cfme/tests/configure/test_visual_cloud.py cfme/tests/configure/test_default_views_infra.py --use-provider complete }}

### DIFF
--- a/cfme/ansible/credentials.py
+++ b/cfme/ansible/credentials.py
@@ -407,7 +407,7 @@ class AnsibleCredentials(CFMENavigateStep):
     VIEW = CredentialsListView
     prerequisite = NavigateToSibling("LoggedIn")
 
-    def step(self):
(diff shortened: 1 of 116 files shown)


digitronik:
Could explain why you need this `*arg` and `**kwargs` on larger scale? Its user privilege for method. As per current implementation its not required. 
Is there any changes we are planning for `CFMENavigateStep`?

mshriver:
@digitronik This is for consistency across these Navmazing method implementations. It allows for passing of args/kwargs through to the methods without errors when a navigation kwarg is added. We're looking at some enhancements for CFMENavigateStep (and/or Navmazing itself) that may pass some additional arguments down.

PR #8463, opened 2019-02-12
[1LP][RFR] Update Appliance.fix_ntp_clock

mshriver:
EVM was overwriting any changes we were making to chrony.conf.

Instead of trying to mangle chrony.conf, configure NTP later in the process, and use EVM UI to do it.

EVM comes with default public pool NTP, an unconfigured CFME image comes with RHEL NTP pools, and we want to use the ones specified in cfme_data for testing.


dajoRH:
Lint report for commit a0a569a2a25d845a237775ded3871d960312faea:

- :bangbang: Line 1114:14: **E131** *continuation line unaligned for hanging indent*

**No commit flaws detected**.

*CFME QE Bot*

sprout/appliances/tasks.py
```diff
@@ -1203,9 +1203,6 @@ def appliance_power_on(self, appliance_id):
                 # VM is running now.
                 sync_appliance_hw.delay(appliance.id)
                 sync_provider_hw.delay(appliance.template.provider.id)
-                # fixes time synchronization
```
mshriver: I don't think that this needs to or should happen during power on - an unconfigured appliance will have the default NTP servers for the image produced by CFME (rhel NTP pool).  A configured appliance will have either the EVM default (public ntp pools) or the NTP servers set during configuration.

PR #8466, opened 2019-02-13
[1LP][RFR] a few useful chrome browser options

izapolsk:
Ref: https://github.com/RedHatQE/cfme_dockerfiles/pull/11


izapolsk:
@john-dupuy, actually I fixed that error in docker image. I guess this option should slightly speed up that call.

PR #8468, opened 2019-02-13
[1LP][RFR] Add test cases for BZ

john-dupuy:
Adding in a automated test case for a BZ.

https://bugzilla.redhat.com/show_bug.cgi?id=1656873

{{ pytest: --long-running cfme/tests/configure/test_workers.py::test_set_memory_threshold_in_ui }}

### DIFF
--- a/cfme/tests/configure/test_workers.py
+++ b/cfme/tests/configure/test_workers.py
@@ -1,34 +1,24 @@
 # -*- coding: utf-8 -*-
 import pytest
 
+from collections import namedtuple
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.blockers import BZ
 from cfme.utils.wait import wait_for
 
-DROPDOWNS = [
-    'generic_worker_threshold',
-    'cu_data_collector_worker_threshold',
-    'event_monitor_worker_threshold',
-    'connection_broker_worker_threshold',
-    'reporting_worker_threshold',
-    'web_service_worker_threshold',
-    'priority_worker_threshold',
-    'cu_data_processor_worker_threshold',
-    'refresh_worker_threshold',
-    'vm_analysis_collectors_worker_threshold'
-]
+Dropdown = namedtuple('Dropdown', 'dropdown id')
 
-IDS = [
-    'generic',
-    'cu_data_coll',
-    'event_monitor',
-    'conn_broker',
-    'reporting',
-    'web_service',
-    'priority',
-    'cu_data_proc',
-    'refresh',
-    'vm_analysis'
+DROPDOWNS = [
+    Dropdown('generic_worker_threshold', 'generic'),
+    Dropdown('cu_data_collector_worker_threshold', 'cu_data_coll'),
+    Dropdown('event_monitor_worker_threshold', 'event_monitor'),
+    Dropdown('connection_broker_worker_threshold', 'conn_broker'),
+    Dropdown('reporting_worker_threshold', 'reporting'),
+    Dropdown('web_service_worker_threshold', 'web_service'),
+    Dropdown('priority_worker_threshold', 'priority'),
+    Dropdown('cu_data_processor_worker_threshold', 'cu_data_proc'),
+    Dropdown('refresh_worker_threshold', 'refresh'),
+    Dropdown('vm_analysis_collectors_worker_threshold', 'vm_analysis')
 ]
 
 
@@ -56,9 +46,9 @@ def test_restart_workers(appliance):
              message="Wait for all original workers are back online")
 
 
-@pytest.mark.tier(0)
-@pytest.mark.parametrize("dropdown", DROPDOWNS, ids=IDS)
-@pytest.mark.meta(blockers=[BZ(1656873)])
+@pytest.mark.tier(2)
+@pytest.mark.parametrize("dropdown", [x.dropdown for x in DROPDOWNS], ids=[x.id for x in DROPDOWNS])
+@pytest.mark.meta(blockers=[BZ(1656873, forced_streams="5.9")])
 def test_set_memory_threshold_in_ui(appliance, dropdown):
     """
     Bugzillas:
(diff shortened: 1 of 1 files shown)


cfme/tests/control/test_alerts.py
```diff
@@ -470,3 +470,37 @@ def test_alert_hardware_reconfigured(request, configure_fleecing, alert_collecti
...
+            7. Assign the Alert Profile to the Server
+            8. In advanced config, change:
+                  events:
+                  :disk_usage_gt_percent: 80
+                to:
+                  events:
+                  :disk_usage_gt_percent: 1
+            9. dd a file in /var/www/miq/vmdb/log large enough to trigger 1% disk usage
+        expectedResults:
+            9. the alert should fire, and the event of type
```
jawatts: **Required** You will still need steps 1-8 here, just leave those blank.

cfme/tests/webui/test_general_ui.py
```diff
@@ -431,3 +431,29 @@ def test_configure_icons_roles_by_server():
...
+        testSteps:
+            1. Navigate to Configuration
+            2. Select the workers tab
+            3. Set memory threshold of "Generic Workers" to a value above 1 GB
+            4. Hit save
+            5. Set the memory threshold of "Event Monitor" to a value above 1 GB
+            6. Hit save
+        expectedResults:
+            4. the change should be reflected in the dropdown
+            6. same as 4.
```
jawatts: Same as above

cfme/tests/control/test_alerts.py
john-dupuy: Thanks @jawatts, I am automating this test, so I will remove this particular test_case from this PR anyway.

mshriver:
Two things:

1. I think this would be a better place for the test:
https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/configure/test_workers.py

2. I think we should go straight to automation here, as the view and navigation already exists for the workers count/threshold page:

https://github.com/ManageIQ/integration_tests/blob/master/cfme/base/ui.py#L479

john-dupuy:
@mshriver I agree with both these points. Regarding 1, I have moved the test to `test_workers`. Regarding 2. I have created a simple test case that checks this for each of the workers.

cfme/tests/configure/test_workers.py
```diff
@@ -27,3 +54,42 @@ def test_restart_workers(appliance):
     wait_for(lambda: len(pids) == len(worker.get_all_worker_pids()),
              fail_func=worker.parent.reload_workers_page, num_sec=1800, delay=10,
              message="Wait for all original workers are back online")
+
+
+@pytest.mark.tier(0)
+@pytest.mark.parametrize("dropdown", DROPDOWNS, ids=IDS)
```
mshriver: Consider using a namedTuple for the pairing, to keep them in order should we get new worker types. Explicitly parametrize here against `[t.dropdown for t in DROPDOWN_TUPLES]` and pass `ids=[t.id for t in DROPDOWN_TUPLES]`

cfme/tests/configure/test_workers.py
```diff
@@ -27,3 +54,42 @@ def test_restart_workers(appliance):
...
+    """
+    view = navigate_to(appliance.server, 'Workers')
+    mem_threshold = getattr(view.workers, dropdown)
+    before = mem_threshold.selected_option
+    mem_threshold.select_by_visible_text("1.1 GB")
+    view.workers.save.click()
+    view.wait_displayed()
+    after = mem_threshold.selected_option
+
+    assert before != after
```
mshriver: Not sure about this check here - defaults might get updated to 1.1GB per worker and then this fails.

Thought: check that `1.1` isn't what is already selected, select 1.2 instead if so

This way the assertion that after matches the value we selected is a complete assertion that it _changed_.

cfme/tests/configure/test_workers.py
john-dupuy: I always wondered if there was an easy way to do something like this in python. I will update this to use a namedtuple. :+1:

cfme/tests/configure/test_workers.py
john-dupuy: While unlikely that this would be the default, I agree that it is good to check and make sure that this is not the case before executing the test.

cfme/tests/configure/test_workers.py
```diff
@@ -27,3 +44,45 @@ def test_restart_workers(appliance):
     wait_for(lambda: len(pids) == len(worker.get_all_worker_pids()),
              fail_func=worker.parent.reload_workers_page, num_sec=1800, delay=10,
              message="Wait for all original workers are back online")
+
+
+@pytest.mark.tier(0)
```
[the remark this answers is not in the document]
john-dupuy: @izapolsk, I initially thought it was tier(0) based on the fact that the key functionality of this page is broken. A page that lets you make changes should be reflective of the changes that you make. Though, since the changes are actually made, just not communicated to you, I think that perhaps a tier(2) would be more appropriate.

cfme/tests/configure/test_workers.py
```diff
@@ -27,3 +44,45 @@ def test_restart_workers(appliance):
...
+        caseimportance: medium
+        initialEstimate: 1/6h
+        testSteps:
+            1. Navigate to Configuration
+            2. Select the workers tab
+            3. Set memory threshold of any dropdown to a value above 1 GB
+            4. Hit save
+        expectedResults:
+            1.
+            2.
```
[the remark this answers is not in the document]
john-dupuy: @izapolsk, @jawatts mentioned to me that such empty rows were necessary for the polarion metadata, this was there is an expectedResult for each test step, which is useful for manual cases. Though this test is now automated, they are not necessary but help explain the test.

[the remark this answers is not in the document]
izapolsk: okay, ty. I think  - 6. blabla is also descriptive enough

PR #8479, opened 2019-02-15
[1LP][RFR]Fix is_displayed for service retirement date

niyazRedhat:
Purpose or Intent
=================

{{pytest: cfme/tests/services/test_myservice.py -k test_retire_service_on_date --long-running --use-provider vsphere6-nested -v}}

### DIFF
--- a/cfme/services/myservice/ui.py
+++ b/cfme/services/myservice/ui.py
@@ -189,8 +189,9 @@ class ServiceRetirementView(ServiceRetirementForm):
         return (
             self.in_myservices and
             self.myservice.is_opened and
-            self.myservice.tree.currently_selected == self.context['object'].name and
-            self.title.text == 'Set/Remove retirement date for Service')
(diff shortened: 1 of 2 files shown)


cfme/services/myservice/ui.py
```diff
@@ -174,7 +174,6 @@ def is_displayed(self):
         return (
             self.in_myservices and
             self.myservice.is_opened and
-            self.myservice.tree.currently_selected == self.context['object'].name and
```
john-dupuy: @niyazRedhat What is the error that you are getting here? `currently_selected` should work for this Accordion, check the Accordion name on L54 of this file. I think that we should keep this check so that the page is associated with the service name.

digitronik: @niyazRedhat why are you removing this?

niyazRedhat: @john-dupuy @digitronik  I made changes please have a look.

PR #8480, opened 2019-02-15
[1LP][RFR] Infra mapping

sshveta:
# PR 8399 

{{ pytest: cfme/tests/v2v/test_infrastructure_mapping.py --use-provider vsphere67-ims --use-provider rhv42 --provider-limit 2 --long-running }}

### DIFF
--- a/cfme/fixtures/v2v_fixtures.py
+++ b/cfme/fixtures/v2v_fixtures.py
@@ -127,9 +127,10 @@ def create_tags(appliance, transformation_method):
         transformation_method: VDDK/SSH
 
     """
+    # t is for True in V2V - Transformation Host * tag
     tag1 = appliance.collections.categories.instantiate(
         display_name="V2V - Transformation Host *"
-    ).collections.tags.instantiate(display_name="transformation_host")
+    ).collections.tags.instantiate(display_name="t")
     tag2 = appliance.collections.categories.instantiate(
         display_name="V2V - Transformation Method"
     ).collections.tags.instantiate(display_name=transformation_method)
@@ -149,23 +150,24 @@ def set_conversion_instance_for_rhev(appliance, transformation_method, rhev_host
         transformation_method : vddk or ssh as per test requirement
         rhev_hosts: hosts in rhev to configure for conversion
     """
-    # Delete all prior conversion hosts otherwise it creates duplicate entries
-    delete_hosts = appliance.ssh_client.run_rails_command("'ConversionHost.delete_all'")
-    if not delete_hosts.success:
-        pytest.skip("Failed to delete all conversion hosts:".format(delete_hosts.out))
 
     for host in rhev_hosts:
         # set conversion host via rails console
         if appliance.version >= "5.10":
+            # Delete all prior conversion hosts otherwise it creates duplicate entries
+            delete_hosts = appliance.ssh_client.run_rails_command("'ConversionHost.delete_all'")
+            if not delete_hosts.success:
+                pytest.skip("Failed to delete all conversion hosts:".format(delete_hosts.output))
+
             set_conv_host = appliance.ssh_client.run_rails_command(
                 "'r = Host.find_by(name:{host});\
             c_host = ConversionHost.create(name:{host},resource:r);\
             c_host.{method}_transport_supported = true;\
             c_host.save'".format(host=json.dumps(host.name),
-                                     method=transformation_method.lower())
+                                 method=transformation_method.lower())
             )
             if not set_conv_host.success:
-                pytest.skip("Failed to set conversion hosts:".format(set_conv_host.out))
+                pytest.skip("Failed to set conversion hosts:".format(set_conv_host.output))
         else:
             tag1, tag2 = create_tags(appliance, transformation_method)
             # if _tag_cleanup() returns True, means all tags were removed
--- a/cfme/fixtures/v2v_fixtures.py
+++ b/cfme/fixtures/v2v_fixtures.py
@@ -1,0 +1,545 @@
+import json
+from collections import namedtuple
+
+import fauxfactory
+import pytest
+from riggerlib import recursive_update
+from widgetastic.utils import partial_match
+
+from cfme.cloud.provider.openstack import OpenStackProvider
+from cfme.fixtures.provider import setup_or_skip
+from cfme.infrastructure.provider.rhevm import RHEVMProvider
+from cfme.infrastructure.provider.virtualcenter import VMwareProvider
+from cfme.utils.generators import random_vm_name
+from cfme.utils.log import logger
+from cfme.v2v.infrastructure_mapping import InfrastructureMapping as InfraMapping
+
+
+FormDataVmObj = namedtuple("FormDataVmObj", ["infra_mapping_data", "vm_list"])
+V2vProviders = namedtuple("V2vProviders", ["vmware_provider", "rhv_provider", "osp_provider"])
+
+
+@pytest.fixture(scope="module")
+def v2v_provider_setup(request, appliance, source_provider, provider):
+    """ Fixture to setup providers """
+    vmware_provider, rhv_provider, osp_provider = None, None, None
+    for v2v_provider in [source_provider, provider]:
+        if v2v_provider.one_of(VMwareProvider):
+            vmware_provider = v2v_provider
+            setup_or_skip(request, vmware_provider)
+        elif v2v_provider.one_of(RHEVMProvider):
+            rhv_provider = v2v_provider
+            setup_or_skip(request, rhv_provider)
+        elif v2v_provider.one_of(OpenStackProvider):
+            osp_provider = v2v_provider
+            setup_or_skip(request, osp_provider)
+        else:
+            pytest.skip("Provider {} is not a valid provider for v2v tests".format(provider.name))
+    v2v_providers = V2vProviders(vmware_provider=vmware_provider,
+                                 rhv_provider=rhv_provider,
+                                 osp_provider=osp_provider)
+
+    # Transformation method can be vddk or ssh
+    if hasattr(request, "param") and request.param == "SSH":
+        transformation_method = "SSH"
+    else:
+        transformation_method = "VDDK"
+
+    # set host credentials for Vmware and RHEV hosts
+    host_credentials(appliance, transformation_method, v2v_providers)
+
+    yield v2v_providers
+    for v2v_provider in v2v_providers:
+        if v2v_provider is not None:
+            v2v_provider.delete_if_exists(cancel=False)
+
+
+def host_credentials(appliance, transformation_method, v2v_providers):
+    """ Sets up host credentials for vmware and rhv providers
+        for RHEV migration.
+        For migration with OSP only vmware(source) provider
+        host credentials need to be added.
+        These credentials are automatically removed once the
+        provider is deleted in clean up.
+
+    Args:
+        appliance
+        transformation_method : vddk or ssh to be used in configuring conversion host
+        v2v_providers: vmware (and rhev in case of RHV migration ) , osp not needed.
+    """
+    provider_list = [v2v_providers.vmware_provider]
+    rhv_hosts = None
+
+    if v2v_providers.rhv_provider is not None:
+        rhv_hosts = v2v_providers.rhv_provider.hosts.all()
+        provider_list.append(v2v_providers.rhv_provider)
+
+    try:
+        for v2v_provider in provider_list:
+            hosts = v2v_provider.hosts.all()
+            for host in hosts:
+                host_data = [data for data in v2v_provider.data['hosts']
+                             if data['name'] == host.name]
+                if not host_data:
+                    pytest.skip("No host data")
+                host.update_credentials_rest(credentials=host_data[0]['credentials'])
+    except Exception:
+        logger.exception("Exception when trying to add the host credentials.")
+        pytest.skip("No data for hosts in providers, failed to retrieve hosts and add creds.")
+    # Configure conversion host for RHEV migration
+    if rhv_hosts is not None :
+        set_conversion_instance_for_rhev(appliance, transformation_method, rhv_hosts)
+    if v2v_providers.osp_provider is not None:
+        set_conversion_instance_for_osp(appliance, transformation_method,
+                                        v2v_providers.osp_provider)
+
+
+def _tag_cleanup(host_obj, tag1, tag2):
+    """
+        Clean Up Tags
+
+        Returns: Boolean True if all Tags were removed/cleaned
+        or False means all required Tags are present on host.
+    """
+
+    def extract_tag(tag):
+        # Following strip will remove extra asterisk from tag assignment
+        return tag.category.display_name.strip(" *"), tag.display_name
+
+    valid_tags = {extract_tag(tag1), extract_tag(tag2)}
+    tags = host_obj.get_tags()
+    tags_set = set(map(extract_tag, tags))
+    # we always neeed 2 tags for migration, if total is less than 2
+    # don't bother checking what tag was it, just remove it and
+    # then add all required tags via add_tags() call. or if tags on host
+    # are not subset of valid tags, we still remove them.
+    if len(tags_set) < 2 or not tags_set.issubset(valid_tags):
+        host_obj.remove_tags(tags=tags)
+        return True
+    return False
+
+
+def create_tags(appliance, transformation_method):
+    """
+    Create tags V2V - Transformation Host * and V2V - Transformation Method
+    Args:
+        appliance:
+        transformation_method: VDDK/SSH
+
+    """
+    tag1 = appliance.collections.categories.instantiate(
+        display_name="V2V - Transformation Host *"
+    ).collections.tags.instantiate(display_name="transformation_host")
+    tag2 = appliance.collections.categories.instantiate(
+        display_name="V2V - Transformation Method"
+    ).collections.tags.instantiate(display_name=transformation_method)
+    return tag1, tag2
+
+
+def set_conversion_instance_for_rhev(appliance, transformation_method, rhev_hosts):
+    """Assigning tags to conversion host.
+       In 5.9 all rhev hosts and tagged with tags V2V - Transformation Host=t
+       and V2V - Transformation Method=vddk/ssh.These tags are automatically removed
+       once the provider is deleted.
+
+       In 5.10 rails console commands are run to configure all the rhev hosts.
+
+    Args:
+        appliance:
+        transformation_method : vddk or ssh as per test requirement
+        rhev_hosts: hosts in rhev to configure for conversion
+    """
+    # Delete all prior conversion hosts otherwise it creates duplicate entries
+    delete_hosts = appliance.ssh_client.run_rails_command("'ConversionHost.delete_all'")
+    if not delete_hosts.success:
+        pytest.skip("Failed to delete all conversion hosts:".format(delete_hosts.out))
+
+    for host in rhev_hosts:
+        # set conversion host via rails console
+        if appliance.version >= "5.10":
+            set_conv_host = appliance.ssh_client.run_rails_command(
+                "'r = Host.find_by(name:{host});\
+            c_host = ConversionHost.create(name:{host},resource:r);\
+            c_host.{method}_transport_supported = true;\
+            c_host.save'".format(host=json.dumps(host.name),
+                                     method=transformation_method.lower())
+            )
+            if not set_conv_host.success:
+                pytest.skip("Failed to set conversion hosts:".format(set_conv_host.out))
+        else:
+            tag1, tag2 = create_tags(appliance, transformation_method)
+            # if _tag_cleanup() returns True, means all tags were removed
+            if _tag_cleanup(host, tag1, tag2):
+                # so we call add_tags to add only required tags
+                host.add_tags(tags=(tag1, tag2))
+
+
+def set_conversion_instance_for_osp(appliance, transformation_method, osp_provider):
+    """
+    Rails console command
+    ====================
+    res = Vm.find_by(name: 'my_osp_instance')
+    conversion_host = ConversionHost.create(name: res.name, resource: res)
+    conversion_host.vddk_transport_supported = true
+    conversion_host.save
+
+    Args:
+        appliance
+        transformation_method: vddk or ssh
+        osp_provider: OSP
+    """
+
+    # Delete all prior conversion hosts otherwise it creates duplicate entries
+    appliance.ssh_client.run_rails_command("'ConversionHost.delete_all'")
+
+    try:
+        conversion_instances = osp_provider.data.get('conversion_instances', [])
+    except KeyError:
+        pytest.skip("No conversion instance on provider.")
+
+    for instance in conversion_instances:
+        set_conv_host = appliance.ssh_client.run_rails_command(
+            "'r = Vm.find_by(name:{vm});\
+        c_host = ConversionHost.create(name:r.name, resource: r);\
+        c_host.{method}_transport_supported = true;\
+        c_host.save'".format(
+                vm=json.dumps(instance),
+                method=transformation_method.lower(),
+            )
+        )
+        if not set_conv_host.success:
+            pytest.skip("Failed to set conversion hosts:".format(set_conv_host.out))
+
+
+def get_vm(request, appliance, source_provider, template, datastore=None):
+    """ Helper method that takes template , source provider and datastore
+        and creates VM on source provider to migrate .
+
+    Args:
+        request
+        appliance:
+        source_provider: Provider on which vm is created
+        template: Template used for creating VM
+        datastore: datastore in which VM is created. If no datastore
+                   is provided then by default VM is created on nfs datastore
+
+        returns: Vm object
+    """
+    if datastore is None:
+        datastore = "nfs"
+    source_datastores_list = source_provider.data.get("datastores", [])
+    source_datastore = [d.name for d in source_datastores_list if d.type == datastore][0]
+    collection = source_provider.appliance.provider_based_collection(source_provider)
+    vm_name = random_vm_name("v2v-auto")
+    vm_obj = collection.instantiate(
+        vm_name, source_provider, template_name=template(source_provider)["name"]
+    )
+    power_on_vm = True
+    if template.__name__ == "win10_template":
+        # Need to leave this off, otherwise migration fails
+        # because when migration process tries to power off the VM if it is powered off
+        # and for win10, it hibernates and that state of filesystem is unsupported
+        power_on_vm = False
+    vm_obj.create_on_provider(
+        timeout=2400,
+        find_in_cfme=True,
+        allow_skip="default",
+        datastore=source_datastore,
+        power_on=power_on_vm,
+    )
+    request.addfinalizer(lambda: vm_obj.cleanup_on_provider())
+    return vm_obj
+
+
+def get_source_data(provider, component, default_value):
+    try:
+        data = (provider.data.get(component, [])[0])
+    except IndexError:
+        data = default_value
+    return data
+
+
+def infra_mapping_default_data(source_provider, provider):
+    """
+    Default data for infrastructure mapping form.
+    It is used in other methods to recursive update the data according
+    to parameters in tests.
+
+    Args:
+        source_provider: Vmware provider
+        provider: Target rhev/OSP provider
+    """
+    infra_mapping_data = {
+        "name": "infra_map_{}".format(fauxfactory.gen_alphanumeric()),
+        "description": "Single Datastore migration of VM from {ds_type1} to {ds_type2}".format(
+            ds_type1="nfs", ds_type2="nfs"
+        ),
+        "clusters": [component_generator("clusters", source_provider, provider)],
+        "datastores": [component_generator(
+            "datastores", source_provider, provider,
+            get_source_data(source_provider, "datastores", "nfs").type,
+            get_source_data(provider, "datastores", "nfs").type)],
+        "networks": [
+            component_generator("vlans", source_provider, provider,
+                                get_source_data(source_provider, "vlans", "VM Network"),
+                                get_source_data(provider, "vlans", "ovirtmgmt"))
+        ],
+    }
+    return infra_mapping_data
+
+
+@pytest.fixture(scope="function")
+def mapping_data_multiple_vm_obj_single_datastore(request, appliance, source_provider, provider):
+    # this fixture will take list of N VM templates via request and call get_vm for each
+    cluster = provider.data.get("clusters", [False])[0]
+    if not cluster:
+        pytest.skip("No data for cluster available on provider.")
+
+    infra_mapping_data = infra_mapping_default_data(source_provider, provider)
+    recursive_update(
+        infra_mapping_data,
+        {
+            "description": "Single Datastore migration of VM from {ds_type1} to {ds_type2},".format(
+                ds_type1=request.param[0], ds_type2=request.param[1]
+            ),
+            "networks": [
+                component_generator("vlans", source_provider, provider, "VM Network", "ovirtmgmt")
+            ],
+        },
+    )
+    vm_list = []
+    for template_name in request.param[2]:
+        vm_list.append(get_vm(request, appliance, source_provider, template_name))
+    return FormDataVmObj(infra_mapping_data=infra_mapping_data, vm_list=vm_list)
+
+
+@pytest.fixture(scope="function")
+def mapping_data_single_datastore(request, source_provider, provider):
+    infra_mapping_data = infra_mapping_default_data(source_provider, provider)
+    recursive_update(
+        infra_mapping_data,
+        {
+            "description": "Single Datastore migration of VM from {ds_type1} to {ds_type2},".format(
+                ds_type1=request.param[0], ds_type2=request.param[1]
+            ),
+            "datastores": [
+                component_generator(
+                    "datastores", source_provider, provider, request.param[0], request.param[1]
+                )
+            ],
+        },
+    )
+    return infra_mapping_data
+
+
+@pytest.fixture(scope="function")
+def mapping_data_single_network(request, source_provider, provider):
+    infra_mapping_data = infra_mapping_default_data(source_provider, provider)
+    recursive_update(
+        infra_mapping_data,
+        {
+            "description": "Single Network migration of VM from {vlan1} to {vlan2},".format(
+                vlan1=request.param[0], vlan2=request.param[1]
+            ),
+            "networks": [
+                component_generator(
+                    "vlans", source_provider, provider, request.param[0], request.param[1]
+                )
+            ],
+        },
+    )
+    return infra_mapping_data
+
+
+@pytest.fixture(scope="function")
+def edited_mapping_data(request, source_provider, provider):
+    infra_mapping_data = infra_mapping_default_data(source_provider, provider)
+    edited_form_data = {
+        "description": "my edited description",
+        "clusters": {},
+        "datastores": {},
+        "networks": [
+            component_generator(
+                "vlans", source_provider, provider, request.param[1][0], request.param[1][1]
+            )
+        ],
+    }
+    return infra_mapping_data, edited_form_data
+
+
+@pytest.fixture(scope="function")
+def mapping_data_dual_vm_obj_dual_datastore(request, appliance, source_provider, provider):
+    vmware_nw = source_provider.data.get("vlans", [None])[0]
+    rhvm_nw = provider.data.get("vlans", [None])[0]
+    cluster = provider.data.get("clusters", [False])[0]
+    if not vmware_nw or not rhvm_nw or not cluster:
+        pytest.skip("No data for source or target network in providers.")
+
+    infra_mapping_data = infra_mapping_default_data(source_provider, provider)
+    recursive_update(
+        infra_mapping_data,
+        {
+            "description": "Dual DS migration of VM from {dss1} to {dst1},& from {dss2} to {dst2}".
+            format(dss1=request.param[0][0],
+                   dst1=request.param[0][1],
+                   dss2=request.param[1][0],
+                   dst2=request.param[1][1]),
+            "datastores": [
+                component_generator(
+                    "datastores",
+                    source_provider,
+                    provider,
+                    request.param[0][0],
+                    request.param[0][1],
+                ),
+                component_generator(
+                    "datastores",
+                    source_provider,
+                    provider,
+                    request.param[1][0],
+                    request.param[1][1],
+                ),
+            ],
+            "networks": [
+                component_generator(
+                    "vlans",
+                    source_provider,
+                    provider,
+                    source_provider.data.get("vlans")[0],
+                    provider.data.get("vlans")[0],
+                )
+            ],
+        },
+    )
+    # creating 2 VMs on two different datastores and returning its object list
+    vm_obj1 = get_vm(request, appliance, source_provider, request.param[0][2], request.param[0][0])
+    vm_obj2 = get_vm(request, appliance, source_provider, request.param[1][2], request.param[1][0])
+    return FormDataVmObj(infra_mapping_data=infra_mapping_data, vm_list=[vm_obj1, vm_obj2])
+
+
+@pytest.fixture(scope="function")
+def mapping_data_vm_obj_dual_nics(request, appliance, source_provider, provider):
+    vmware_nw = source_provider.data.get("vlans", [None])[0]
+    rhvm_nw = provider.data.get("vlans", [None])[0]
+    cluster = provider.data.get("clusters", [False])[0]
+    if not vmware_nw or not rhvm_nw or not cluster:
+        pytest.skip("No data for source or target network in providers.")
+
+    infra_mapping_data = infra_mapping_default_data(source_provider, provider)
+    recursive_update(
+        infra_mapping_data,
+        {
+            "description": "Dual DS migration of VM from {dss1} to {dst1},& from {dss2} to {dst2}".
+            format(dss1=request.param[0][0],
+                   dst1=request.param[0][1],
+                   dss2=request.param[1][0],
+                   dst2=request.param[1][1]),
+            "networks": [
+                component_generator(
+                    "vlans", source_provider, provider, request.param[0][0], request.param[0][1]
+                ),
+                component_generator(
+                    "vlans", source_provider, provider, request.param[1][0], request.param[1][1]
+                ),
+            ],
+        },
+    )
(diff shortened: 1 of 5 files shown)


dajoRH:
Lint report for commit e8841a6a0034e849b1eff75d06ce0f292f1de65a:

`cfme/fixtures/v2v.py`:
- :red_circle: Line 416:13: **F841** *local variable 'type' is assigned to but never used*

`cfme/tests/v2v/test_infrastructure_mapping.py`:
- :bangbang: Line 44:101: **E501** *line too long (106 > 100 characters)*

`cfme/v2v/infrastructure_mapping.py`:
- :red_circle: Line 12:1: **F401** *'widgetastic.utils.Parameter' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
Lint report for commit 3ca62596d9366de101c5983ecb42b946d38e17c1:

`cfme/fixtures/v2v.py`:
- :red_circle: Line 13:1: **N813** *camelcase imported as lowercase*

`cfme/v2v/infrastructure_mapping.py`:
- :red_circle: Line 96:11: **N801** *class names should use CapWords convention*
- :red_circle: Line 100:11: **N801** *class names should use CapWords convention*
- :red_circle: Line 104:11: **N801** *class names should use CapWords convention*

**No commit flaws detected**.

*CFME QE Bot*

cfme/fixtures/v2v.py
```diff
@@ -10,11 +10,12 @@
 from cfme.fixtures.provider import setup_or_skip
 from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
+from cfme.v2v.infrastructure_mapping import InfrastructureMapping as InfraMapping
 from cfme.utils.generators import random_vm_name
 from cfme.utils.log import logger
```
digitronik: Required:
@sshveta need to refactor import with new style.

cfme/fixtures/v2v.py
```diff
@@ -163,32 +165,28 @@ def get_vm(request, appliance, source_provider, template, datastore="nfs"):
...
-        },
+def infra_mapping_default_data(source_provider, provider):
+    """
+    Default data for infrastructure mapping form.
+    It is used in other methods to recursive update the data accoriding
+    to parameters in tests.
+
+    :param source_provider: Vmware provider
+    :param provider: Target rhev/OSP provider
+    :return:
```
digitronik: I like this doc string style helped lot in doc building... but we are using different...
https://gist.github.com/mfalesni/11694e0d563c80c3917ce6012400b93f
Its old please correct me if missing something.

cfme/fixtures/v2v.py
```diff
@@ -310,56 +278,44 @@ def form_data_dual_vm_obj_dual_datastore(request, appliance, source_provider, pr
...
-                            "vlans",
-                            source_provider,
-                            provider,
-                            source_provider.data.get("vlans")[0],
-                            provider.data.get("vlans")[0],
-                        )
-                    ]
-                }
-            },
+            "description": "Dual DS migration of VM from {} to {},& from {} to {}".format(
```
digitronik: Required:
Please us keying for string formatting.

cfme/fixtures/v2v.py
```diff
@@ -370,119 +326,101 @@ def form_data_vm_obj_dual_nics(request, appliance, source_provider, provider):
...
-                            "vlans",
-                            source_provider,
-                            provider,
-                            request.param[1][0],
-                            request.param[1][1],
-                        ),
-                    ]
-                }
-            },
+            "description": "Dual DS migration of VM from {} to {},& from {} to {}".format(
```
digitronik: same as above

cfme/tests/v2v/test_infrastructure_mapping.py
```diff
@@ -0,0 +1,48 @@
+"""Test to validate End-to-End migrations- functional testing."""
+import pytest
+
+from cfme.fixtures.provider import rhel7_minimal
+from cfme.infrastructure.provider.rhevm import RHEVMProvider
+from cfme.infrastructure.provider.virtualcenter import VMwareProvider
+from cfme.markers.env_markers.provider import ONE_PER_VERSION, ONE_PER_TYPE
+
```
digitronik: Required:
need one per line import now.

cfme/v2v/infrastructure_mapping.py
```diff
@@ -0,0 +1,298 @@
...
+    cancel_btn = Button("Cancel")
+    save = Button("Save")
+    create = Button("Create")
+    add_mapping = Button("Add Mapping")
+    remove_mapping = Button("Remove Selected")
+    remove_all_mappings = Button("Remove All")
+    mappings_tree = InfraMappingTreeView(tree_class="treeview")
+
+
+class ComponentView(View):  # noqa
```
digitronik: Q: why `#noqa` ?

cfme/v2v/infrastructure_mapping.py
```diff
@@ -0,0 +1,298 @@
...
+    buttons = View.include(InfraMappingCommonButtons, use_parent=True)
+    comp_name = None
```
digitronik: Q: I don't understand meaning of `comp_name` as None

cfme/v2v/infrastructure_mapping.py
```diff
@@ -0,0 +1,298 @@
...
+    class DatastoreComponent(InfraMappingComponent):  # noqa
+        comp_name = "datastore"
+
+    @attr.s
+    class NetworkComponent(InfraMappingComponent):  # noqa
+        comp_name = "network"
+
+    name = attr.ib()
+    description = attr.ib()
+    clusters = attr.ib(validator=attr.validators.instance_of(list))
```
digitronik: :100:

cfme/v2v/infrastructure_mapping.py
sshveta: Oh I changed the name but forgot to remove #noqa

cfme/fixtures/v2v.py
sshveta: yes that was default generated doc string

dajoRH:
Lint report for commit 6a30b7e12cc6b0fbba64ef233b88e894597b5fcb:

`cfme/fixtures/v2v.py`:
- :bangbang: Line 288:13: **E124** *closing bracket does not match visual indentation*
- :bangbang: Line 337:13: **E124** *closing bracket does not match visual indentation*
- :bangbang: Line 405:101: **E501** *line too long (102 > 100 characters)*

**No commit flaws detected**.

*CFME QE Bot*

digitronik:
@sshveta as you removed from `_form_data` import affected in tests `cfme.fixtures.v2v import _form_data`

sshveta:
> @sshveta as you removed from `_form_data` import affected in tests `cfme.fixtures.v2v import _form_data`

@digitronik : Will do that in a following PR with total refactoring of tests , as this PR is going to break jenkins tests , we might run v2v jenkins on an old branch .

cfme/fixtures/v2v.py
```diff
@@ -370,119 +328,106 @@ def form_data_vm_obj_dual_nics(request, appliance, source_provider, provider):
...
     vm_obj = get_vm(request, appliance, source_provider, request.param[2])
-    return FormDataVmObj(form_data=form_data, vm_list=[vm_obj])
+    return FormDataVmObj(infra_mapping_data=infra_mapping_data, vm_list=[vm_obj])
 
 
-def _form_data_mapping(selector, source_provider, provider, source_list=None, target_list=None):
+def component_generator(selector, source_provider, provider, source_list=None, target_list=None):
+    """
+    Component generator method to generate a dict of source and target
+    components(cluster/datastore/networks).
```
mshriver: Please expand this description a bit as the function is doing a few things:

Its looking up component names in provider yaml, and skipping the test is relevant information is not provided.

cfme/fixtures/v2v.py
```diff
@@ -370,119 +328,106 @@ def form_data_vm_obj_dual_nics(request, appliance, source_provider, provider):
...
+
+    Args:
+        selector: can be cluster/datastore/network
+        source_provider: vmware provider to migrate from
+        provider:  rhev or osp provider or target provider to migrate to
+        source_list: source cluster/datastore/network on vmware provider to migrate from
+        target_list: target cluster/ds/network to migrate to
+
+    """
```
mshriver: As this method will be imported/used by multiple fixtures or test modules, consider a check here against selector before continuing, raising ValueError if its not an expected selector.

cfme/fixtures/v2v.py
```diff
@@ -370,119 +328,106 @@ def form_data_vm_obj_dual_nics(request, appliance, source_provider, provider):
...
 
 
-def _form_data_mapping(selector, source_provider, provider, source_list=None, target_list=None):
+def component_generator(selector, source_provider, provider, source_list=None, target_list=None):
+    """
+    Component generator method to generate a dict of source and target
+    components(cluster/datastore/networks).
+
+    Args:
+        selector: can be cluster/datastore/network
```
mshriver: These are plural below, not singular.

cfme/fixtures/v2v.py
```diff
@@ -370,119 +328,106 @@ def form_data_vm_obj_dual_nics(request, appliance, source_provider, provider):
...
+
+    """
     source_data = source_provider.data.get(selector, [])
     target_data = provider.data.get(selector, [])
+
     if selector is "clusters":
-        source = source_data if source_data else None
-        target = target_data if target_data else None
-        skip_test = not source or not target
+        sources = source_data if source_data else None
```
mshriver: These statements can be expressed as `sources = source_data or None`

cfme/fixtures/v2v.py
```diff
@@ -370,119 +328,106 @@ def form_data_vm_obj_dual_nics(request, appliance, source_provider, provider):
...
+        targets = target_data if target_data else None
+        component = InfraMapping.ClusterComponent(
+            [partial_match(sources[0])], [partial_match(targets[0])]
```
mshriver: This will cause a TypeError if `sources` was set to `None` above.

Logic needs to be reworked a bit here to make this flow a bit better.

There are a few options:
1. Check `source_data` and `target_data` at ~L410 and skip immediately if they're empty
2. Set component to `None` if `sources` or `targets` is empty, modify `skip_test` setting on L430 to check if component is `None`

cfme/fixtures/v2v.py
mshriver: You're also passing `'vlans'` as a selector instead of `'networks'`, but the logic here doesn't explicitly check for either. Which is correct?

cfme/fixtures/v2v.py
```diff
@@ -370,119 +328,106 @@ def form_data_vm_obj_dual_nics(request, appliance, source_provider, provider):
...
+    Component generator method to generate a dict of source and target
+    components(cluster/datastore/networks).
+
+    Args:
+        selector: can be cluster/datastore/network
+        source_provider: vmware provider to migrate from
+        provider:  rhev or osp provider or target provider to migrate to
+        source_list: source cluster/datastore/network on vmware provider to migrate from
+        target_list: target cluster/ds/network to migrate to
+
```
mshriver: Returns?

cfme/fixtures/v2v.py
```diff
@@ -370,119 +328,106 @@ def form_data_vm_obj_dual_nics(request, appliance, source_provider, provider):
...
+def component_generator(selector, source_provider, provider, source_list=None, target_list=None):
+    """
+    Component generator method to generate a dict of source and target
+    components(cluster/datastore/networks).
+
+    Args:
+        selector: can be cluster/datastore/network
+        source_provider: vmware provider to migrate from
+        provider:  rhev or osp provider or target provider to migrate to
+        source_list: source cluster/datastore/network on vmware provider to migrate from
```
mshriver: These descriptions don't sound accurate given how they're used below - for processing datastores, its comparing the datastore type property to `source_list`, and for networks its just checking if they're equal. These appear to be type strings, not a list of any kind?

cfme/fixtures/v2v.py
```diff
@@ -163,32 +165,28 @@ def get_vm(request, appliance, source_provider, template, datastore="nfs"):
...
+    """
+    infra_mapping_data = {
+        "name": "infra_map_{}".format(fauxfactory.gen_alphanumeric()),
+        "description": "Single Datastore migration of VM from {ds_type1} to {ds_type2}".format(
+            ds_type1="nfs", ds_type2="nfs"
+        ),
+        "clusters": [component_generator("clusters", source_provider, provider)],
+        "datastores": [component_generator("datastores", source_provider, provider, "nfs", "nfs")],
+        "networks": [
+            component_generator("vlans", source_provider, provider, "VM Network", "ovirtmgmt")
```
mshriver: It looks like you're picking specific values here to pass, and hard coding them into the test. `VM Network`  and `ovirtmgmt` are network/vlan names on the providers themselves, and should be pulled from yaml, not hardcoded.

cfme/fixtures/v2v.py
```diff
@@ -163,32 +165,28 @@ def get_vm(request, appliance, source_provider, template, datastore="nfs"):
...
+    to parameters in tests.
+
+    Args:
+        source_provider: Vmware provider
+        provider: Target rhev/OSP provider
+    """
+    infra_mapping_data = {
+        "name": "infra_map_{}".format(fauxfactory.gen_alphanumeric()),
+        "description": "Single Datastore migration of VM from {ds_type1} to {ds_type2}".format(
+            ds_type1="nfs", ds_type2="nfs"
```
mshriver: `nfs` is just the default parametrization value for datastore, and I don't think should be hardcoded in this default. Why not include `request` here and pick from `request.param` instead, so the test controls which datastore types are parametrized instead of the default being hardcoded.

cfme/fixtures/v2v.py
```diff
@@ -198,108 +196,78 @@ def form_data_multiple_vm_obj_single_datastore(request, appliance, source_provid
...
-            }
-        },
+        "description": "my edited description",
+        "clusters": {},
+        "datastores": [
+            component_generator("datastores", source_provider, provider, "iscsi", "iscsi")
+        ],
+        "networks": [
+            component_generator(
+                "vlans", source_provider, provider, "DPortGroup", "Storage - VLAN 33"
```
mshriver: Again, hardcoded values that should be pulled from provider data.

cfme/v2v/infrastructure_mapping.py
```diff
@@ -0,0 +1,304 @@
...
+from widgetastic_manageiq import InfraMappingTreeView
+from widgetastic_manageiq import MultiSelectList
+from widgetastic_manageiq import V2VPaginatorPane
+
+
+class InfrastructureMappingView(BaseLoggedInPage):
+    title = Text("#explorer_title_text")
+
+    @property
+    def in_explorer(self):
```
mshriver: Consider calling this `in_mapping_explorer` to be more explicit when additional views are using this for is_displayed

cfme/v2v/infrastructure_mapping.py
```diff
@@ -0,0 +1,304 @@
...
+        fill_strategy = WaitFillViewStrategy("15s")
+
+        def after_fill(self, was_change):
+            if not self.parent.add_mapping.disabled:
+                self.parent.add_mapping.click()
+
+    def fill(self, values):
+        was_change = True
+        for mapping in values:
+            self.MappingFillView(self.comp_name).fill(mapping)
```
mshriver: For completeness, please define `ComponentView.comp_name` as 'None' above, so the class has the attribute set. The nested definitions include it, but the base class does not.

cfme/fixtures/v2v.py
sshveta: For Networks the parameter in yaml is vlan . So when it is not cluster or datastore , it is creating network component .

cfme/fixtures/v2v.py
sshveta: @mshriver : This is default data for a mapping. It is not used in tests . The mapping used in tests is derived from this default data by doing recursive updates from request.params.

cfme/fixtures/v2v.py
sshveta: This is just to create a data structure for default data . 
Not used in tests directly .
Only used after doing recursive updates from request.params

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,449 @@
...
+from cfme.v2v.infrastructure_mapping import InfrastructureMapping as InfraMapping
+
+
+FormDataVmObj = namedtuple("FormDataVmObj", ["infra_mapping_data", "vm_list"])
+
+
+@pytest.fixture(scope="module")
+def v2v_provider_setup(request, source_provider, provider):
+    """ Fixture to setup providers """
+    V2vProviders = namedtuple("V2vProviders", ["vmware_provider", "rhv_provider", "osp_provider"])
```
mshriver: Consider defining this up with FormDataVmObj, instead of locally here.

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,449 @@
...
+    vmware_provider, rhv_provider = None, None
+    for provider in [source_provider, provider]:
```
mshriver: I think its a bad practice to overload `provider` here, you need a separate variable name for the loop iterations

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,449 @@
...
+            rhv_provider = provider
+            setup_or_skip(request, rhv_provider)
+        elif provider.one_of(OpenStackProvider):
+            osp_provider = provider
+            setup_or_skip(request, osp_provider)
+        else:
+            pytest.skip("Provider {} is not a valid provider for v2v tests".format(provider.name))
+    v2v_providers = V2vProviders(vmware_provider=vmware_provider,
+                                 rhv_provider=rhv_provider,
+                                 osp_provider=osp_provider)
```
mshriver: `osp_provider` will cause `UnboundLocalError` if neither provider was Openstack, because it wasn't set to None on L26 with the others.

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,449 @@
...
+    yield v2v_providers
+    v2v_providers.vmware_provider.delete_if_exists(cancel=False)
+    v2v_providers.rhv_provider.delete_if_exists(cancel=False)
+    v2v_providers.osp_provider.delete_if_exists(cancel=False)
+
+
+@pytest.fixture(scope="module")
+def add_host_creds(request, v2v_providers):
+    """Add credentials to VMware and RHV hosts."""
+    if len(v2v_providers) > 2:
```
mshriver: This will always be the case, as the fixture above populates all 3, though one will be set to `None`

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,449 @@
...
+    # then add all required tags via add_tags() call. or if tags on host
+    # are not subset of valid tags, we still remove them.
+    if len(tags_set) < 2 or not tags_set.issubset(valid_tags):
+        host_obj.remove_tags(tags=tags)
+        return True
+    return False
+
+
+@pytest.fixture(scope="module")
+def assign_conversion_tags(request, appliance, host_creds):
```
mshriver: Is this supposed to be `add_host_creds`, the fixture above?

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,449 @@
...
+            "which is invalid, skipping."
+        )
+    try:
+        vmware_provider = v2v_providers.vmware_provider
+        rhv_provider = v2v_providers.rhv_provider
+        vmware_hosts = vmware_provider.hosts.all()
+        try:
+            for host in vmware_hosts:
+                host_data, = [
+                    data for data in vmware_provider.data["hosts"] if data["name"] == host.name
```
mshriver: So, some things to consider that I think make this less than ideal:
1. If there are multiple matching hosts that are in the yaml and returned by `provider.hosts.all()`, this hits ValueError - is that really invalid? Do we want to just pick the first host in that case? Isn't there potential that we'll want to update credentials on _all_ the hosts available to the provider, since any could be used in the test?

2. The call to `host.update_credentials_rest()` could be make directly in the list comprehension, since we're not using that host data otherwise. 

3. The code duplication here could be eliminated by creating `vmware_hosts` and `rhv_hosts` outside of the try block, and then iterating over each. The same list comprehension is being used against both host lists.

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,449 @@
...
+        transformation_method = "VDDK"
+    tag2 = appliance.collections.categories.instantiate(
+        display_name="V2V - Transformation Method"
+    ).collections.tags.instantiate(display_name=transformation_method)
+    appliance.ssh_client.run_rails_command("'ConversionHost.delete_all'")
+    for host in host_creds:
+        # if _tag_cleanup() returns True, means all tags were removed
+        if _tag_cleanup(host, tag1, tag2):
+            # so we call add_tags to add only required tags
+            host.add_tags(tags=(tag1, tag2))
```
mshriver: Is this adding the categories too? Or do you expect the categories above to already exist?

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,449 @@
...
+            # set conversion host via rails console
+            if appliance.version >= "5.10":
```
mshriver: Is there an else for this clause? Do we not need conversion host set for CFME 5.9?

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,449 @@
...
+                if isinstance(host.provider, RHEVMProvider):
```
mshriver: Please use `host.provider.one_of(RHEVMProvider)` here

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,449 @@
...
+                        type=json.dumps(resource_type),
+                        method=transformation_method.lower(),
+                    )
+                )
+    yield
+    for host in host_creds:
+        host.remove_tags(tags=(tag1, tag2))
+
+
+def get_vm(request, appliance, source_provider, template, datastore="nfs"):
```
mshriver: This looks like a fixture?

`datastore` kwarg is inaccurate, its a type of datastore.

Need docblocks explaining use, assumptions, etc.

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,449 @@
...
+    # are not subset of valid tags, we still remove them.
+    if len(tags_set) < 2 or not tags_set.issubset(valid_tags):
+        host_obj.remove_tags(tags=tags)
+        return True
+    return False
+
+
+@pytest.fixture(scope="module")
+def assign_conversion_tags(request, appliance, host_creds):
+    """Assigning tags to conversation host"""
```
mshriver: Please expand this docblock, its not descriptive enough.

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,449 @@
...
+        power_on=power_on_vm,
+    )
+    request.addfinalizer(lambda: vm_obj.cleanup_on_provider())
+    return vm_obj
+
+
+def infra_mapping_default_data(source_provider, provider):
+    """
+    Default data for infrastructure mapping form.
+    It is used in other methods to recursive update the data accoriding
```
mshriver: typo, `according`

cfme/fixtures/v2v.py
mshriver: This is not true, as `mapping_data_multiple_vm_obj_single_datastore` does not modify clusters or datastores, and `mapping_data_single_datastore` does not modify clusters or networks.

If we're going to have some default values here, we need to justify where they are coming from. Are they default network names for VMware and RHEVM? If so we at least need to state them here - but again, why not pull them from the provider provisioning data? You have both provider objects available.

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,449 @@
...
+    Gets the provider data based on selector from cfme_data.yaml and creates
+    InfraMapping.component(source_list, target_list) object
+
+    Test is skipped if no source or target data is found
+
+    Args:
+        selector: can be clusters/datastores/networks
+        source_provider: vmware provider to migrate from
+        provider:  rhev or osp provider or target provider to migrate to
+        source_type: string source datastores/networks on vmware provider to migrate from.
```
mshriver: :+1:

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,449 @@
...
+    """
+    Component generator method to generate a dict of source and target
+    components(clusters/datastores/networks).
+    Gets the provider data based on selector from cfme_data.yaml and creates
+    InfraMapping.component(source_list, target_list) object
+
+    Test is skipped if no source or target data is found
+
+    Args:
+        selector: can be clusters/datastores/networks
```
mshriver: `clusters/datastores/vlans`

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,449 @@
...
+        source_type: string source datastores/networks on vmware provider to migrate from.
+                     Ex: if source_type is "iscsi". Provider data is checked for datastore with type
+                     iscsi and that datastores name is used.
+        target_type: string target datastores/networks to migrate to
+
+        returns : InfraMapping.component(source_list, target_list) object
+
+    """
+    if selector not in "clusters" or "datastores" or "vlans":
+        raise ValueError("Please specify cluster, datatsore or network(vlans) selector!")
```
mshriver: typo, `datastore`

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,449 @@
...
+        provider:  rhev or osp provider or target provider to migrate to
+        source_type: string source datastores/networks on vmware provider to migrate from.
+                     Ex: if source_type is "iscsi". Provider data is checked for datastore with type
+                     iscsi and that datastores name is used.
+        target_type: string target datastores/networks to migrate to
+
+        returns : InfraMapping.component(source_list, target_list) object
+
+    """
+    if selector not in "clusters" or "datastores" or "vlans":
```
mshriver: this should be a list, `if selector not in ['clusters', 'datastores', 'vlans']`

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,449 @@
...
+        raise ValueError("Please specify cluster, datatsore or network(vlans) selector!")
+
+    source_data = source_provider.data.get(selector, [])
+    target_data = provider.data.get(selector, [])
+    component = None
+    if selector is "clusters":
```
mshriver: I still think the logic is a bit broken here

If `source_data` was empty, `sources[0]` will raise TypeError because `sources` is now `None` and has no `__get__` method.

You only want to look at the selector if `source_data` and `target_data` are not empty. You can call `pytest.skip` immediately, or move the selector checks into an `if` block, so that when they're empty it skips straight to the existing check+skip.

cfme/tests/v2v/test_infrastructure_mapping.py
```diff
@@ -0,0 +1,49 @@
...
+        classes=[RHEVMProvider], selector=ONE_PER_VERSION, required_flags=["v2v"], scope="module"
+    ),
+    pytest.mark.provider(
+        classes=[VMwareProvider],
+        selector=ONE_PER_TYPE,
+        fixture_name="source_provider",
+        required_flags=["v2v"],
+        scope="module",
+    ),
+    pytest.mark.usefixtures("v2v_provider_setup", "add_host_creds", "assign_conversion_tags"),
```
mshriver: I think it would be good to structure the fixtures such that `add_host_creds` and `assign_conversion_tags` are included just by the fixture tree, and are not required to be specified directly in the test module or function.

cfme/fixtures/v2v_fixtures.py
sshveta: This is not required , removing

cfme/v2v/infrastructure_mapping.py
```diff
@@ -0,0 +1,305 @@
...
+class AddInfrastructureMappingView(InfrastructureMappingForm):
+    form_title_text = VersionPicker(
+        {
+            Version.lowest(): "Infrastructure Mapping Wizard",
+            "5.10": "Create Infrastructure Mapping",
+        }
+    )
+
+    @property
+    def is_displayed(self):
```
mshriver: Looks like this could be moved to `InfrastructureMappingForm.is_displayed`, since the child classes are both setting `form_title_text`

cfme/v2v/infrastructure_mapping.py
```diff
@@ -0,0 +1,305 @@
...
+        }
+
+    def update(self, updates):
+        """
+        Args:
+            updates: An entity instance, that we can call form_values on
+        Returns:
+        """
+        view = navigate_to(self, "Edit", wait_for_view=20)
+        # TODO recursive update of self.form_values and updates.form_values
```
mshriver: You can use riggerlib for this, should be pretty straightforward with the fill_dict methods implemented already.

cfme/tests/v2v/test_infrastructure_mapping.py
sshveta: OSP does not required add_host_creds and assign_conversion_tags fixture , hence cannot combine these with v2v_provider fixture . They need to be separate so that only v2v_provider_Setup can be called in case of OSP

mshriver: Gotcha. So, if they are required by v2v_provider_setup, and they check the `provider` type - just yielding immediately when its OSP - then all V2V tests that use provider setup would get those actions by default. Thoughts?

cfme/fixtures/v2v_fixtures.py
mshriver: Cool!

cfme/fixtures/v2v_fixtures.py
sshveta: Yes categories are present. V2V - Transformation Host and V2V - Transformation Method

cfme/fixtures/v2v_fixtures.py
sshveta: For CFME 5.9 only tags are assigned to RHV hosts.That is enough for set up

dajoRH:
Lint report for commit d47a6fc170cefb4466fe3deec3e45cd06dddd03e:

`cfme/fixtures/v2v_fixtures.py`:
- :bangbang: Line 62:101: **E501** *line too long (102 > 100 characters)*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
Lint report for commit eba2e689ae7f53f382036129c9fd6927ca3e85ef:

`cfme/fixtures/v2v_fixtures.py`:
- :bangbang: Line 62:101: **E501** *line too long (102 > 100 characters)*

`cfme/v2v/infrastructure_mapping.py`:
- :red_circle: Line 2:1: **F401** *'riggerlib.recursive_update' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,468 @@
...
+@pytest.fixture(scope="module")
+def host_credentials(request, v2v_provider_setup):
+    """ Sets up host credentials for vmware and rhv providers.
+    Args:
+        request:
+        v2v_provider_setup:
+    """
+    v2v_providers = v2v_provider_setup
+    provider_list = [v2v_providers.vmware_provider, v2v_providers.rhv_provider]
+    rhv_hosts = v2v_providers.rhv_provider.hosts.all()
```
jawatts: what if v2v_providers.rhv_provider is `None` in this case, this would throw an `AttributeError`

dajoRH:
Lint report for commit 4a51e0d516a1cb9d04ccf89dd62043b6e2f59d65:

`cfme/fixtures/v2v_fixtures.py`:
- :red_circle: Line 1:1: **F401** *'itertools' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

sshveta:
Not sure why travis is failing . It shows green in the details

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,522 @@
...
+        Rails console command
+        ====================
+        res = Vm.find_by(name: 'my_osp_instance')
+        conversion_host = ConversionHost.create(name: res.name, resource: res)
+        conversion_host.vddk_transport_supported = true
+        conversion_host.save
+    """
+    # Delete all prior conversion hosts otherwise it creates duplicate entries
+    appliance.ssh_client.run_rails_command("'ConversionHost.delete_all'")
+    conversion_instance = osp_provider.data.get('conversion_instance1', None)
```
Yadnyawalkya: @sshveta  we also need to consider usecase where migration already in progress on conversion host and since that host is busy, UI will show you warning like 'Conversion is in progress'. To avoid that in [RHV we usually specify more than one conversion hosts](https://github.com/ManageIQ/integration_tests/pull/8480/files#diff-ed6030ba2f5ff1aa837d6e8cbfaf433cR146), we should use the same for OSP as well.

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,522 @@
...
+
+    resource_type = "VmOrTemplate"
+    appliance.ssh_client.run_rails_command(
+        "'r = Vm.find_by(name:{vm});\
+    c_host = ConversionHost.create(name:{vm},resource_id:r.id,resource_type:{type});\
```
Yadnyawalkya: @sshveta  you are using deprecated rails commands here, no need to specify `resource_id` and `resource_type` now.
We should add 

```
res = Vm.find_by(name: 'my_osp_instance')
conversion_host = ConversionHost.create(name: res.name, resource: res)
conversion_host.vddk_transport_supported = true # appropriate transport method
conversion_host.save
```

Please update it accordingly to RHV's conversion hosts as well, only change is, 
```
res = Host.find_by(name: 'my_rhv_host')
```

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,522 @@
...
+
+       In 5.10 rails console commands are run to configure all the rhev hosts.
+
+    Args:
+        appliance:
+        transformation_method : vddk or ssh as per test requirement
+        rhev_hosts: number of hosts in rhev to configure for conversion
+    """
+
+    tag1 = appliance.collections.categories.instantiate(
```
Yadnyawalkya: @sshveta  we should only run tag thing for 5.9 since we don't need them on 5.10. Please specify it in else part of 5.10 or something like that.  This can save our 10s :dagger:

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,531 @@
...
+        host credentials need to be added.
+        These credentials are automatically removed once the
+        provider is deleted in clean up.
+
+    Args:
+        appliance
+        transformation_method : vddk or ssh to be used in configuring conversion host
+        v2v_providers: vmware (and rhev in case of RHV migration ) , osp not needed.
+    """
+    provider_list = [v2v_providers.vmware_provider]
```
jawatts: This make the assumption that `v2v_providers` will always have a vmware provider. Is this safe to assume? What if vmware is `None` in `v2v_provider_setup()`

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,531 @@
...
+    """
+    Create tags V2V - Transformation Host * and V2V - Transformation Method
+    Args:
+        appliance:
+        transformation_method: VDDK/SSH
+
+    """
+    tag1 = appliance.collections.categories.instantiate(
+        display_name="V2V - Transformation Host *"
+    ).collections.tags.instantiate(display_name="t")
```
jawatts: `t` does not seem very descriptive here.

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,531 @@
...
+       In 5.9 all rhev hosts and tagged with tags V2V - Transformation Host=t
+       and V2V - Transformation Method=vddk/ssh.These tags are automatically removed
+       once the provider is deleted.
+
+       In 5.10 rails console commands are run to configure all the rhev hosts.
+
+    Args:
+        appliance:
+        transformation_method : vddk or ssh as per test requirement
+        rhev_hosts: number of hosts in rhev to configure for conversion
```
jawatts: `rhev_hosts` is a list of hosts right? Not a `int` or `number`

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,531 @@
...
+    """
+    # Delete all prior conversion hosts otherwise it creates duplicate entries
+    appliance.ssh_client.run_rails_command("'ConversionHost.delete_all'")
```
jawatts: Should we make sure that this command passes before continuing? If this fails, we will have duplicate entries

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,531 @@
...
+        if appliance.version >= "5.10":
+            appliance.ssh_client.run_rails_command(
+                "'r = Host.find_by(name:{host});\
+            c_host = ConversionHost.create(name:{host},resource:r);\
+            c_host.{method}_transport_supported = true;\
+            c_host.save'".format(
+                    host=json.dumps(host.name),
+                    method=transformation_method.lower(),
+                )
+            )
```
jawatts: I think it would be good to check the RC of this. Is there anyway to also verify that the conversion was indeed set correctly?

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,531 @@
...
+
+
+def set_conversion_instance_for_osp(appliance, transformation_method, osp_provider):
+    """
+    Args:
+        appliance
+        transformation_method: vddk or ssh
+        osp_provider: OSP
+
+        Rails console command
```
jawatts: This shouldn't be under the `Args`.

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,531 @@
...
+        ====================
+        res = Vm.find_by(name: 'my_osp_instance')
+        conversion_host = ConversionHost.create(name: res.name, resource: res)
+        conversion_host.vddk_transport_supported = true
+        conversion_host.save
+    """
+    # Delete all prior conversion hosts otherwise it creates duplicate entries
+    appliance.ssh_client.run_rails_command("'ConversionHost.delete_all'")
```
jawatts: same comment about checking output

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,531 @@
...
+        provider: Target rhev/OSP provider
+    """
+    infra_mapping_data = {
+        "name": "infra_map_{}".format(fauxfactory.gen_alphanumeric()),
+        "description": "Single Datastore migration of VM from {ds_type1} to {ds_type2}".format(
+            ds_type1="nfs", ds_type2="nfs"
+        ),
+        "clusters": [component_generator("clusters", source_provider, provider)],
+        "datastores": [component_generator("datastores", source_provider, provider,
+                                           (source_provider.data.get("datastores", [])[0]).type,
```
jawatts: This will throw an `IndexError` if the result of the `get` is an empty list

cfme/fixtures/v2v_fixtures.py
sshveta: @jawatts : Yes Vmware provider is always there as source provider .
Only target provider can be different

dajoRH:
Lint report for commit f8f5b5e41c7380c14b0576ff57d92587367eff8a:

`cfme/fixtures/v2v_fixtures.py`:
- :bangbang: Line 161:29: **E126** *continuation line over-indented for hanging indent*
- :bangbang: Line 279:101: **E501** *line too long (102 > 100 characters)*

**No commit flaws detected**.

*CFME QE Bot*

cfme/v2v/infrastructure_mapping.py
```diff
@@ -0,0 +1,305 @@
...
+    def fill(self, values):
+        was_change = True
+        for mapping in values:
+            self.MappingFillView(self.comp_name).fill(mapping)
+        else:
+            was_change = False
+        self.after_fill(was_change)
+        return was_change
+
+    def after_fill(self, was_change):
```
jawatts: **Question** Do you want to click next regardless if it was changed or not?

dajoRH:
Lint report for commit 9cbf86b636a09a71253123e2c9cdfb10e6fa071b:

`cfme/fixtures/v2v_fixtures.py`:
- :bangbang: Line 165:38: **E127** *continuation line over-indented for visual indent*

**No commit flaws detected**.

*CFME QE Bot*

cfme/v2v/infrastructure_mapping.py
sshveta: Yes , because there are multiple forms one after other and any page can be changed

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,545 @@
...
+    ).collections.tags.instantiate(display_name="transformation_host")
+    tag2 = appliance.collections.categories.instantiate(
+        display_name="V2V - Transformation Method"
+    ).collections.tags.instantiate(display_name=transformation_method)
+    return tag1, tag2
+
+
+def set_conversion_instance_for_rhev(appliance, transformation_method, rhev_hosts):
+    """Assigning tags to conversion host.
+       In 5.9 all rhev hosts and tagged with tags V2V - Transformation Host=t
```
jawatts: `t` needs to be updated here since it was changed in `create_tags()`

cfme/fixtures/v2v_fixtures.py
```diff
@@ -0,0 +1,545 @@
...
+    conversion_host.save
+
+    Args:
+        appliance
+        transformation_method: vddk or ssh
+        osp_provider: OSP
+    """
+
+    # Delete all prior conversion hosts otherwise it creates duplicate entries
+    appliance.ssh_client.run_rails_command("'ConversionHost.delete_all'")
```
jawatts: I don't see the RC check here

cfme/fixtures/v2v_fixtures.py
sshveta: So I added a comment t is for True and one of the values listed in V2V - Transformation Host. 
Cannot be changed.

PR #8483, opened 2019-02-16
[1LP][RFR] fixing broken provider object comparison

izapolsk:
{{ pytest: --long-running --use-template-cache -k 'smoke or test_ipappliance_managed_providers ' cfme }}

### DIFF
--- a/cfme/cloud/provider/__init__.py
+++ b/cfme/cloud/provider/__init__.py
@@ -76,7 +76,7 @@ class CloudProviderImagesView(BaseLoggedInPage):
     including_entities = View.include(VMEntities, use_parent=True)
 
 
-@attr.s(hash=False)
+@attr.s(cmp=False)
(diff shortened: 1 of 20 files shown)


cfme/cloud/provider/__init__.py
```diff
@@ -76,7 +76,7 @@ def is_displayed(self):
     including_entities = View.include(VMEntities, use_parent=True)
 
 
-@attr.s(hash=False)
```
mshriver: Can you describe your intent here?

If `cmp=False` and `hash` not set, hash will default to the base class `__hash__` method. Currently we are explicitly defining the providers as _not hashable_, this will change them to use `BaseProvider.__hash__`.

^ Is that the intent? If so, cool.

izapolsk: @mshriver, I also had such opinion before looking into this issue. atts is nasty beast and doesn't work as it is expected. 
it doesn't pick up `__hash__`.  I messed around with it for several hours until I managed to make it pick up 
__hash__ and __eq__.

PR #8486, opened 2019-02-18
[1LP][RFR] This will fix control-action tests for Ansible.

sbulage:
PRT run:
{{ pytest: cfme/tests/ansible/test_embedded_ansible_basic.py -qsvvvv --long-running }}

### DIFF
--- a/cfme/control/explorer/actions.py
+++ b/cfme/control/explorer/actions.py
@@ -205,8 +205,7 @@ class Action(BaseEntity, Updateable, Pretty):
             assert view.is_displayed
             view.flash.assert_no_error()
         else:
-            view = self.create_view(ActionsAllView)
-            view.wait_displayed('15s')
+            view = self.create_view(ActionsAllView, wait="15s")
             view.flash.assert_success_message(
                 'Action "{}": Delete successful'.format(self.description))
 
--- a/cfme/control/explorer/actions.py
+++ b/cfme/control/explorer/actions.py
@@ -4,6 +4,7 @@ import attr
 from cached_property import cached_property
 from navmazing import NavigateToAttribute
 from navmazing import NavigateToSibling
+from widgetastic.utils import WaitFillViewStrategy
 from widgetastic.widget import Checkbox
 from widgetastic.widget import Text
 from widgetastic.widget import View
@@ -85,6 +86,7 @@ class NewActionView(ActionFormCommon):
 
 
 class EditActionView(ActionFormCommon):
+    fill_strategy = WaitFillViewStrategy()
     title = Text("#explorer_title_text")
 
     save_button = Button("Save")
--- a/cfme/control/explorer/actions.py
+++ b/cfme/control/explorer/actions.py
@@ -4,7 +4,6 @@ import attr
 from cached_property import cached_property
 from navmazing import NavigateToAttribute
 from navmazing import NavigateToSibling
-from widgetastic.utils import WaitFillViewStrategy
 from widgetastic.widget import Checkbox
 from widgetastic.widget import Text
 from widgetastic.widget import View
@@ -86,7 +85,6 @@ class NewActionView(ActionFormCommon):
 
 
 class EditActionView(ActionFormCommon):
-    fill_strategy = WaitFillViewStrategy()
     title = Text("#explorer_title_text")
 
     save_button = Button("Save")
@@ -208,7 +206,7 @@ class Action(BaseEntity, Updateable, Pretty):
             view.flash.assert_no_error()
         else:
             view = self.create_view(ActionsAllView)
-            assert view.is_displayed
+            view.wait_displayed('15s')
             view.flash.assert_success_message(
                 'Action "{}": Delete successful'.format(self.description))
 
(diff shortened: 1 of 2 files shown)


dajoRH:
Lint report for commit 54ddceb339146db2e278ce5c5cdea1160adba550:

`cfme/tests/ansible/test_embedded_ansible_basic.py`:
- :red_circle: Line 253:13: **F821** *undefined name 'appliance'*

**No commit flaws detected**.

*CFME QE Bot*

sbulage:
DS-59z is failing due to text box visibility after selecting Specific Machine. Tried with WaitFillStrategy, REST API etc. but still not working but in every scenario it is passing on DS-510z.

cfme/tests/ansible/test_embedded_ansible_basic.py
```diff
@@ -296,17 +265,19 @@ def test_control_crud_ansible_playbook_action(request, catalog_item, action_coll
             "inventory": {
                 "specific_hosts": True,
                 "hosts": ipaddr
-            }
+            },
         }
     view = navigate_to(action, "Edit")
     assert view.description.value == new_descr
     assert view.run_ansible_playbook.inventory.hosts.value == ipaddr
+    view.browser.refresh()
```
jawatts: **Required** We try to avoid using a browser refresh unless absolutely necessary, it increases the length of our test runs significantly. I would suggest renavigating here with the `Force=True` arg.

cfme/tests/ansible/test_embedded_ansible_basic.py
```diff
@@ -317,20 +288,28 @@ def test_control_add_ansible_playbook_action_invalid_address(request, catalog_it
...
     )
-    request.addfinalizer(action.delete_if_exists)
+
+    @request.addfinalizer
+    def _finalizer():
+        if action.exists:
+            action_id = appliance.rest_api.collections.actions.get(
+                description=action.description
+            ).id
+            appliance.rest_api.collections.actions.action.delete(id=action_id)
```
jawatts: **Required** This finalizer seems to be used twice here, perhaps break it out into a separate method.

cfme/tests/ansible/test_embedded_ansible_basic.py
```diff
@@ -317,20 +288,28 @@ def test_control_add_ansible_playbook_action_invalid_address(request, catalog_it
...
+
     assert action.exists
     view = navigate_to(action, "Edit")
     assert view.run_ansible_playbook.inventory.hosts.value == "invalid_address_!@#$%^&*"
+    view.browser.refresh()
```
jawatts: Same comment about navigating here, please use `Force` arg

cfme/control/explorer/actions.py
```diff
@@ -206,7 +206,7 @@ def delete(self, cancel=False):
             view.flash.assert_no_error()
         else:
             view = self.create_view(ActionsAllView)
-            assert view.is_displayed
+            view.wait_displayed('15s')
```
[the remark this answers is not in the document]
sbulage: @jawatts Yes, tried still taking time to reload and eventually test(s) failing.

cfme/tests/ansible/test_embedded_ansible_basic.py
sbulage: @jawatts Both processes are equal time consuming even re-navigating back to same page will take more time.

cfme/tests/ansible/test_embedded_ansible_basic.py
sbulage: See above answer on re-navigation.

cfme/tests/ansible/test_embedded_ansible_basic.py
digitronik: @sbulage as per my observation... wrong to use `browser refresh` as well `force=True` will totally wrong approach.
I think you are adding work around to land on `Details` page after value assertion right?
If I am right you can use simple approach like
```
view.cancel_button.click()
``` 
Please try and let me know.

cfme/control/explorer/actions.py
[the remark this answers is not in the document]
digitronik: It should work
```
view = self.create_view(ActionsAllView, wait="15s")
```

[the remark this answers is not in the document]
sbulage: I will give try once and let you know.

cfme/tests/ansible/test_embedded_ansible_basic.py
sbulage: Trying... let you know.

cfme/tests/ansible/test_embedded_ansible_basic.py
sbulage: @jawatts Updating it.

izapolsk:
@sbulage,  looks good so far but major part of changes is just cosmetic changes. What exactly have you fixed ?
In addition, since you modify fixtures used by many tests, I update your PRT to see overall state of tests.
This should prove us that nothing is broken by these changes.

sbulage:
@izapolsk 2 tests were failing, I have fixed those and updated with Assignee. Also updated with finalizer.

PR #8490, opened 2019-02-19
[1LP][RFR] Adding global fixtures and formatting tests according to it

ganeshhubale:
- Added global fixtures for automate
    - ```domain``` : This fixture used to create automate domain - Datastore/Domain
    - ```namespace```: This fixture used to create automate namespace - Datastore/Domain/Namespace
    - ```class```: This fixture used to create automate class - Datastore/Domain/Namespace/Class
- Removed deprecated code from tests
- Removed less than ```5.9``` dependency
- Used global fixtures in automate tests

Note: Running the test cases which uses newly added global fixtures
 
{{ pytest: cfme/tests/automate/ -k "test_duplicate_domain_disallowed or test_domain_edit_enabled or test_class_crud or test_schema_crud or test_schema_duplicate_field_disallowed or test_duplicate_class_disallowed or test_same_class_name_different_namespace or test_class_display_name_unset_from_ui or test_duplicate_instance_disallowed or test_instance_display_name_unset_from_ui or test_namespace_crud or test_namespace_delete_from_table or test_duplicate_namespace_disallowed or test_domain_present or test_vmware_vimapi_hotadd_disk" -vvv --long-running }}

### DIFF
--- a/cfme/fixtures/automate.py
+++ b/cfme/fixtures/automate.py
@@ -1,0 +1,32 @@
+import fauxfactory
+import pytest
+
+
+@pytest.fixture(scope='module')
+def domain(appliance):
+    """This fixture used to create automate domain - Datastore/Domain"""
+    domain = appliance.collections.domains.create(
+        name=fauxfactory.gen_alpha(),
+        description=fauxfactory.gen_alpha(),
+        enabled=True)
+    yield domain
+    domain.delete_if_exists()
+
+
+@pytest.fixture(scope="module")
+def namespace(domain):
+    """This fixture used to create automate namespace - Datastore/Domain/Namespace"""
+    yield domain.namespaces.create(
+        name=fauxfactory.gen_alpha(),
+        description=fauxfactory.gen_alpha()
+    )
+
+
+@pytest.fixture(scope="module")
+def klass(namespace):
+    """This fixture used to create automate class - Datastore/Domain/Namespace/Class"""
+    yield namespace.classes.create(
+        name=fauxfactory.gen_alpha(),
+        display_name=fauxfactory.gen_alpha(),
+        description=fauxfactory.gen_alpha()
+    )
--- a/cfme/fixtures/automate.py
+++ b/cfme/fixtures/automate.py
@@ -16,17 +16,21 @@ def domain(appliance):
 @pytest.fixture(scope="module")
 def namespace(domain):
     """This fixture used to create automate namespace - Datastore/Domain/Namespace"""
-    yield domain.namespaces.create(
+    namespace = domain.namespaces.create(
         name=fauxfactory.gen_alpha(),
         description=fauxfactory.gen_alpha()
     )
+    yield namespace
+    namespace.delete_if_exists()
 
 
 @pytest.fixture(scope="module")
 def klass(namespace):
     """This fixture used to create automate class - Datastore/Domain/Namespace/Class"""
-    yield namespace.classes.create(
+    klass = namespace.classes.create(
         name=fauxfactory.gen_alpha(),
         display_name=fauxfactory.gen_alpha(),
         description=fauxfactory.gen_alpha()
     )
+    yield klass
+    klass.delete_if_exists()
--- a/cfme/tests/automate/test_class.py
+++ b/cfme/tests/automate/test_class.py
@@ -3,28 +3,15 @@ import fauxfactory
 import pytest
 
 from cfme import test_requirements
-from cfme.automate.explorer.domain import DomainCollection
-from cfme.utils.blockers import BZ
 from cfme.utils.update import update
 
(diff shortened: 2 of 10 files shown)


cfme/fixtures/automate.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+@pytest.fixture(scope="module")
+def namespace(domain):
+    """This fixture used to create automate namespace - Datastore/Domain/Namespace"""
+    yield domain.namespaces.create(
+        name=fauxfactory.gen_alpha(),
+        description=fauxfactory.gen_alpha()
+    )
+
```
lina-nikiforova: Q: Why don't you have some cleanup here afterwards?

cfme/fixtures/automate.py
```diff
@@ -0,0 +1,32 @@
...
+
+@pytest.fixture(scope="module")
+def klass(namespace):
+    """This fixture used to create automate class - Datastore/Domain/Namespace/Class"""
+    yield namespace.classes.create(
+        name=fauxfactory.gen_alpha(),
+        display_name=fauxfactory.gen_alpha(),
+        description=fauxfactory.gen_alpha()
+    )
```
lina-nikiforova: Q: Same as above

cfme/tests/automate/test_class.py
```diff
@@ -49,6 +36,7 @@ def test_class_crud(namespace):
         initialEstimate: 1/30h
         tags: automate
     """
+    namespace = get_namespace
```
lina-nikiforova: I think it's better to refactor names than reassign like this

cfme/tests/automate/test_domain.py
```diff
@@ -196,12 +187,6 @@ def test_domain_lock_unlock(request, appliance):
         caseimportance: medium
         tags: automate
     """
-    domain = appliance.collections.domains.create(
-        name=fauxfactory.gen_alpha(),
-        description=fauxfactory.gen_alpha(),
-        enabled=True)
-
-    request.addfinalizer(domain.delete)
     ns1 = domain.namespaces.create(name='ns1')
```
lina-nikiforova: Q: Why in previous 2 cases you had `assert domain.exists` but in this case not?

cfme/fixtures/automate.py
ganeshhubale: @lina-nikiforova 
- I am agree with you.
- But ```namepaces```, ```classes```, ````methods````, ```instances``` etc. are not require cleanup because these things are completely dependent ```domain``` creation. 
- Without creating domain we can not proceed for these things.
- Still I have added cleanup for these fixtures.

cfme/tests/automate/test_class.py
```diff
@@ -3,28 +3,15 @@
...
-        enabled=True)
-    yield d
-    d.delete()
-
-
 @pytest.fixture(
     scope="function",
     params=["plain", "nested_existing"])
-def namespace(request, domain):
+def get_namespace(request, domain):
```
digitronik: As you refactoring this @ganeshhubale I would like in state of `return` we can go with `yield`. It will help you to clean what you created. what you say?

cfme/tests/automate/test_domain.py
```diff
@@ -187,7 +178,7 @@ def test_domain_name_wrong(appliance):
 
 
 @pytest.mark.tier(2)
-def test_domain_lock_unlock(request, appliance):
+def test_domain_lock_unlock(domain, appliance):
```
digitronik: Optional: but fiy @ganeshhubale when we have have some object entity and for small use-case `appliance` you can get with `entity(domain).appliance`. Not required for this implementation.

cfme/tests/automate/test_vmware_methods.py
```diff
@@ -83,11 +68,10 @@ def testing_vm(setup_provider, provider):
         vm.cleanup_on_provider()
 
 
-@pytest.mark.meta(blockers=[1211627, BZ(1311221, forced_streams=['5.5'])])
 def test_vmware_vimapi_hotadd_disk(
-        appliance, request, testing_group, provider, testing_vm, domain, cls):
-    """ Tests hot adding a disk to vmware vm.
-    This test exercises the ``VMware_HotAdd_Disk`` method, located in ``/Integration/VMware/VimApi``
+        appliance, request, testing_group, testing_vm, domain, cls):
+    """Tests hot adding a disk to vmware vm. This test exercises the ``VMware_HotAdd_Disk`` method,
```
digitronik: I think single `` will be great.

cfme/tests/automate/test_class.py
ganeshhubale: Yes, I am agree with you.
Used ```yield``` in get_namespace.

PR #8492, opened 2019-02-19
[1LP][RFR] Add new manual test case for Node Utlization graphs

jawatts:
Adding new manual test case of the node utilization graphs on the Container provider dashboard.


cfme/tests/containers/test_overview_data_integrity.py
```diff
@@ -82,3 +82,35 @@ def test_containers_overview_data_integrity(appliance, soft_assert):
                 cls.__name__, api_values[cls], statusbox_value
             )
         )
+
+
+@pytest.mark.manual
+@pytest.mark.tier(1)
```
digitronik: For this file we already have global `tire(1)`

jawatts: Good catch. I will fix that

PR #8495, opened 2019-02-21
[1LP][RFR] New test for product features with custom roles

ganeshhubale:
Purpose or Intent
=================
- This test case is based on new RFR for 5.10
- This RFE includes managing ```Dialogs``` and ```Quota``` via RBAC
- This test case is added because of this BZ: [1655012](https://bugzilla.redhat.com/show_bug.cgi?id=1655012)
- Also with this new test case, I have updated poarion tags of tests assigned to me in same file.

{{ pytest: cfme/tests/configure/test_analysis_profiles.py::test_custom_role_modify_for_dynamic_product_feature -vv }}


cfme/tests/configure/test_analysis_profiles.py
```diff
@@ -333,3 +336,70 @@ def test_analysis_profile_description_validation(analysis_profile_collection):
...
+    tenant = appliance.collections.tenants.create(
+        name=TENANT_NAME,
+        description="tenant_des{}".format(fauxfactory.gen_alphanumeric()),
+        parent=appliance.collections.tenants.get_root_tenant(),
+    )
+    request.addfinalizer(tenant.delete)
+    role = appliance.collections.roles.instantiate(name='EvmRole-tenant_quota_administrator')
+    copied_role = role.copy()
+    request.addfinalizer(copied_role.delete)
+    copied_role.update({'product_features': product_features})
```
sshveta: Required : assert for result here , whether the role got updated or not .

cfme/tests/configure/test_analysis_profiles.py
```diff
@@ -24,6 +25,8 @@
     {'Name': files_list[0]['Name'],
      'Collect Contents?': not files_list[0]['Collect Contents?']}]
 
+TENANT_NAME = "tenant_{}".format(fauxfactory.gen_alphanumeric())
+
 
 def events_check(updates=False):
     form_bug = BZ(1485953, forced_streams=['5.7', '5.8', 'upstream'])
```
sshveta: This line can be removed now? If not completely then 5.7 and 5.8

ganeshhubale: @sshveta 
This bug is closed as ```WONTFIX```. Hence removing only ```5.7 and 5.8``` dependency.

cfme/tests/configure/test_analysis_profiles.py
ganeshhubale: Added assert statements for checking whether nodes are updated or not in that role.

cfme/tests/configure/test_access_control.py
```diff
@@ -1336,17 +1339,17 @@ def test_tenant_quota_input_validate(appliance):
...
         assignee: ghubale
         casecomponent: Infra
         initialEstimate: 1/20h
+        tags: quota
+        testSteps:
+            1. Login as an 'Administrator' user
+            2. Navigate to configuration > access control > tenants
+            3. Select default tenant('My Company') from tenants table
+            4. Delete using 'configuration > Delete selected items'
+            5. Check whether default tenant is deleted or not
```
jawatts: **Required** Please break this up between `testSteps` and `expectedResults`

cfme/tests/configure/test_analysis_profiles.py
```diff
@@ -24,9 +25,11 @@
     {'Name': files_list[0]['Name'],
      'Collect Contents?': not files_list[0]['Collect Contents?']}]
 
+TENANT_NAME = "tenant_{}".format(fauxfactory.gen_alphanumeric())
+
 
 def events_check(updates=False):
-    form_bug = BZ(1485953, forced_streams=['5.7', '5.8', 'upstream'])
+    form_bug = BZ(1485953, forced_streams=['upstream'])
```
jawatts: Does this not apply to 5.9 or 5.10? Just upstream that is does not work?

cfme/tests/configure/test_analysis_profiles.py
```diff
@@ -333,3 +336,77 @@ def test_analysis_profile_description_validation(analysis_profile_collection):
...
+            3. Update newly created custom role by doing uncheck in to options provided under
+               automation > automate > customization > Dialogs > modify > edit/add/copy/delete
+               > uncheck for any tenant
+            4. Or Update newly created custom role by doing uncheck in to options provided under
+               Settings > Configuration > Access Control > Tenants > Modify > Manage Quotas
+               > uncheck for any tenant
+            5. You will see save button is not enabled but if you changed 'Name' or
+               'Access Restriction for Services, VMs, and Templates' then save button is getting
+               enabled.
+            6. It updates changes only when we checked or unchecked for all of the tenants under
```
jawatts: same comment as above

cfme/tests/configure/test_analysis_profiles.py
```diff
@@ -333,3 +336,77 @@ def test_analysis_profile_description_validation(analysis_profile_collection):
...
     view.flash.assert_message("Description can't be blank")
     view.cancel.click()
+
+
+@test_requirements.rbac
+@pytest.mark.tier(1)
+@pytest.mark.ignore_stream('5.9')
+@pytest.mark.parametrize(
+    'product_features',
+    [  # Navigation for product features trees
```
digitronik: Optional:
I think we should put this as global veritable not like this.... and then parametrize. It looks like mess in test pedestrianization. its my personal view.

jawatts: Totally agree here Nikhil, was going to make the same comment.

cfme/tests/configure/test_access_control.py
digitronik: As per me it should contributors privilege else we need to add large change in integration test. For every step we need expectedResults.

jawatts: I talked to Mike about this, for automated test we do not need to separate it so strictly. So this can be **Optional**

cfme/tests/configure/test_analysis_profiles.py
jawatts: I want to say make this **Required** since we both feel the same here

ganeshhubale: @jawatts @digitronik 
I have created global list and then parametrised it.
Thank for suggestion.

cfme/tests/configure/test_analysis_profiles.py
ganeshhubale: Added ```expectedResults:``` tag.

cfme/tests/configure/test_analysis_profiles.py
ganeshhubale: @jawatts @digitronik 
Please help me with this comment.

- I am not sure if this is exist on ```5.10 or 5.9```.
- I went through BZ but I am not getting clear idea behind this BZ.
- As it was reported for ```5.7, 5.8```, also according to [this comment](https://bugzilla.redhat.com/show_bug.cgi?id=1485953#c9); this BZ does not exist on 5.10 or 5.9.
- For more confirmation, I would like to ask @mshriver about it; as it was reported by Mike.
- For now I am updating with removing this ``` form_bug```.

Thank you

digitronik: I think you should remove `upstream` as well.

jawatts: @digitronik agree. Looks like @ganeshhubale did remove it in the lastest.

cfme/tests/configure/test_analysis_profiles.py
```diff
@@ -24,13 +24,107 @@
...
+                "Copy",
+                "Copy ({})".format(TENANT_NAME),
+            ],
+            False,
+        )
+    ],
+]
+
 
 def events_check(updates=False):
```
mshriver: This function appears to be pretty useless without the BZ check. Whatever is calling it can select which list to use based on the arg passed here.

Please just remove the function, and update any callers.

cfme/tests/configure/test_analysis_profiles.py
```diff
@@ -24,13 +24,107 @@
     {'Name': files_list[0]['Name'],
      'Collect Contents?': not files_list[0]['Collect Contents?']}]
 
+TENANT_NAME = "tenant_{}".format(fauxfactory.gen_alphanumeric())
+
+PRODUCT_FEATURES = [
```
mshriver: This makes for pretty unreadable parametrization of the tests. The first two items have 6 strings in common, and then the rest have a different set of 6 strings in common. Only the last two strings change in either case.

Need to at a minimum remove the duplication here, but I think we need to brainstorm on a better way to parametrize this information in the first place.

cfme/tests/configure/test_analysis_profiles.py
```diff
@@ -24,13 +24,107 @@
...
+                "Everything",
+                "Settings",
+                "Configuration",
+                "Access Control",
+                "Tenants",
+                "Modify",
+                "Manage Quotas",
+                "Manage Quotas (My Company)",
+            ],
+            False,
```
mshriver: There appears to be an extra layer of lists in this structure, since there is only a single item in each. Do you expect further parametrization items, with multiple tuples?

cfme/tests/configure/test_analysis_profiles.py
```diff
@@ -24,13 +24,107 @@
...
+                "Everything",
+                "Automation",
+                "Automate",
+                "Customization",
+                "Dialogs",
+                "Modify",
+                "Copy",
+                "Copy ({})".format(TENANT_NAME),
+            ],
+            False,
```
mshriver: Why is every one of the second tuple entries `False`? Is it really needed?

cfme/tests/configure/test_analysis_profiles.py
```diff
@@ -333,3 +427,79 @@ def test_analysis_profile_description_validation(analysis_profile_collection):
...
+
+
+@test_requirements.rbac
+@pytest.mark.tier(1)
+@pytest.mark.ignore_stream('5.9')
+@pytest.mark.parametrize(
+    'product_features',
+    [  # Navigation for product features trees
+        # product_features tree for Managing Quotas for 'My Company' tenant
+        PRODUCT_FEATURES[0],
```
mshriver: here is where the extra layer of lists are in the above data structure. Why isn't PRODUCT_FEATURES, a list, passed directly to parametrize here? Why the extra layer and specific index references here? The comments here should be in the definition above, where there is no context as to what is being defined.

cfme/tests/configure/test_analysis_profiles.py
```diff
@@ -333,3 +427,79 @@ def test_analysis_profile_description_validation(analysis_profile_collection):
...
+    )
+    request.addfinalizer(tenant.delete)
+    role = appliance.collections.roles.instantiate(name='EvmRole-tenant_quota_administrator')
+    copied_role = role.copy(name="{role}_{name}".format(
+        role=role.name, name=fauxfactory.gen_alpha()))
+    request.addfinalizer(copied_role.delete)
+    view = navigate_to(copied_role, 'Details')
+
+    # Checks whether feature tree path is checked for given node
+    assert not view.features_tree.check_uncheck_node(True, *(product_features[0][0]))
```
mshriver: What is [0][0] expected to be here?

cfme/tests/configure/test_analysis_profiles.py
```diff
@@ -333,3 +427,79 @@ def test_analysis_profile_description_validation(analysis_profile_collection):
...
+    copied_role.update({'product_features': product_features})
+
+    # Checks whether feature tree path is unchecked for given node
+    assert view.features_tree.check_uncheck_node(True, *(product_features[0][0]))
```
mshriver: I think you mean to pass `False` as the first argument to `check_uncheck_node()` here to uncheck?

cfme/tests/configure/test_analysis_profiles.py
ganeshhubale: Yes @mshriver It is required because by default it's true for all and here this test case checks the where unchecking(false) particular tenant feature works or not.

cfme/tests/configure/test_analysis_profiles.py
ganeshhubale: @mshriver 
- I think ```check_uncheck_node()``` is used to check or uncheck the node.
- So in this test case i need to check only; node is checked or not. Hence I am using ```node_checked```.

cfme/tests/configure/test_analysis_profiles.py
ganeshhubale: I have added description for use different uses of ```product_feature``` tuple.

cfme/tests/configure/test_analysis_profiles.py
ganeshhubale: While updating role with ```product_feature``` tree. we need to pass product_feature as a list of tupple with true/fasle arguments to check or uncheck the node.

cfme/tests/configure/test_analysis_profiles.py
mshriver: Okay, but if its `False` in all cases, why doesn't the test just use `False` directly? Maybe without some kind of comments/labels its just opaque how its being used effectively.

I'm not quite sure what you mean by 'by default its true for all' - default where? Your parametrization of this data structure is local to this module, and there's no default in the data structure, they're just `False`.

cfme/tests/configure/test_analysis_profiles.py
```diff
@@ -24,13 +24,27 @@
...
-        return updated_files if not form_bug.blocks else None
-    else:
-        return events_list if not form_bug.blocks else None
+OPERATIONS = ["Add", "Edit", "Delete", "Copy"]
+
+PRODUCT_FEATURES_QUOTA = [
+    (
+        ["Everything", "Settings", "Configuration", "Access Control", "Tenants", "Modify",
+         "Manage Quotas"] + ["Manage Quotas ({tenant})".format(tenant=tenant)],
+        False,
```
mshriver: Question still stands - why do we need a tuple with two elements, when the second element is always `False`? Couldn't whatever is using this second tuple value, just use `False` directly instead?

cfme/tests/configure/test_analysis_profiles.py
```diff
@@ -24,13 +24,27 @@
...
 
-def events_check(updates=False):
-    form_bug = BZ(1485953, forced_streams=['5.7', '5.8', 'upstream'])
-    if updates:
-        return updated_files if not form_bug.blocks else None
-    else:
-        return events_list if not form_bug.blocks else None
+OPERATIONS = ["Add", "Edit", "Delete", "Copy"]
+
+PRODUCT_FEATURES_QUOTA = [
```
mshriver: This is much better parametrization than before, thanks!

in-line comments here explaining what this represents would make it even more readable in the future - is this a UI nav path? without a comment, I have to guess and verify my guess.

cfme/tests/configure/test_analysis_profiles.py
```diff
@@ -333,3 +345,72 @@ def test_analysis_profile_description_validation(analysis_profile_collection):
...
+    request.addfinalizer(tenant.delete)
+    role = appliance.collections.roles.instantiate(name='EvmRole-tenant_quota_administrator')
+    copied_role = role.copy(name="{role}_{name}".format(
+        role=role.name, name=fauxfactory.gen_alpha()))
+    request.addfinalizer(copied_role.delete)
+    view = navigate_to(copied_role, 'Details')
+    # node_checked: Checks whether feature tree path is checked for given node
+    # Example:- Tuple: product_feature = (['Everything', 'Automation', 'Automate', 'Customization',
+    # 'Dialogs', 'Modify', 'Add', 'Add (tenant_27WyY6qCHi)'], False)
+    # 'False' argument with product_feature is required while updating copied_roles. It unchecks the
```
mshriver: Great explanation, but I don't think that should be in the tuple from the parametrization. It should be trivial to  inject it when you're actually trying to make the update. 

```
coped_role.update(
    {'product_features': [(f, False) for f in product_features]
)
```

dajoRH:
*CFME QE Bot*

cfme/tests/configure/test_analysis_profiles.py
ganeshhubale: - Agree! I have taken product features as a list and then passed tuple to update the copied role: ```copied_role.update({'product_features': [(product_features, False)]})```
- Now I can able to remove indexing and passed ```product_features``` to ```node_checked```.

cfme/tests/configure/test_analysis_profiles.py
ganeshhubale: Removed tuples from parametrization.

cfme/tests/configure/test_analysis_profiles.py
ganeshhubale: - Used ```False``` directly in test.
- By default true means: the tree path or nodes in RBAC are already checked (true) to the elements on which i am performing this test.

cfme/tests/configure/test_analysis_profiles.py
```diff
@@ -24,13 +24,34 @@
...
-    if updates:
-        return updated_files if not form_bug.blocks else None
-    else:
-        return events_list if not form_bug.blocks else None
+TENANT_NAME = "tenant_{}".format(fauxfactory.gen_alphanumeric())
+
+# Operations performed on service dialogs with respect to RBAC permissions
+OPERATIONS = ["Add", "Edit", "Delete", "Copy"]
+
+# Tree path navigation (for service dialog permissions) to particular node of product feature in
```
mshriver: :ok_hand:

cfme/tests/configure/test_analysis_profiles.py
```diff
@@ -24,13 +24,34 @@
...
+# RBAC for roles
+PRODUCT_FEATURES_DIALOG = [
+    ["Everything", "Automation", "Automate", "Customization", "Dialogs", "Modify"]
+    + [op, "{operation} ({tenant})".format(operation=op, tenant=TENANT_NAME)]
+    for op in OPERATIONS
+]
+
+# Tree path navigation (for quota management permissions) to particular node of product feature in
```
mshriver: :ok_hand:

PR #8499, opened 2019-02-22
[1LP][WIPTEST] Do not collect kubevirt or vSphere-IMS providers

john-dupuy:
Adding `Metadata` for timelines tests so that they do not pickup vSphere-IMS providers. Also, fixing tests that were collecting against Kubevirt that shouldn't be. Some general cleanup of old BZs as well.  

Also including fixes for tests in `test_timelines.py`, these tests were failing due to a change in 5.10 and BZ 1670550. 

{{pytest: -v --long-running --color=yes --use-provider  complete cfme/tests/infrastructure/test_timelines.py }}

### DIFF
--- a/cfme/tests/cloud_infra_common/test_events.py
+++ b/cfme/tests/cloud_infra_common/test_events.py
@@ -8,16 +8,18 @@ from cfme.cloud.provider.gce import GCEProvider
 from cfme.control.explorer.policies import VMControlPolicy
 from cfme.infrastructure.provider import InfraProvider
 from cfme.infrastructure.provider.kubevirt import KubeVirtProvider
+from cfme.markers.env_markers.provider import providers
+from cfme.utils.providers import ProviderFilter
 from cfme.utils.wait import wait_for
 
 
+all_prov = ProviderFilter(classes=[InfraProvider, CloudProvider], required_fields=['provisioning'])
+excluded = ProviderFilter(classes=[KubeVirtProvider], inverted=True)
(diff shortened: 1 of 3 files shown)


jawatts:
Waiting for collection to run

cfme/tests/cloud_infra_common/test_events.py
```diff
@@ -18,6 +17,7 @@
     pytest.mark.provider([CloudProvider, InfraProvider],
```
mshriver: I think you should update this to create an excluded ProviderFilter, and combine it with the current ProviderFilter that you have defined here as kwargs to mark.provider.

cfme/tests/cloud_infra_common/test_events.py
```diff
@@ -7,15 +7,18 @@
...
 from cfme.infrastructure.provider import InfraProvider
-from cfme.infrastructure.provider.rhevm import RHEVMProvider
-from cfme.utils.blockers import BZ
+from cfme.infrastructure.provider.kubevirt import KubeVirtProvider
+from cfme.markers.env_markers.provider import providers
+from cfme.utils.providers import ProviderFilter
 from cfme.utils.wait import wait_for
 
 
+all_prov = ProviderFilter(classes=[InfraProvider, CloudProvider])
```
mshriver: This ProviderFilter should include the required_fields kwarg. The `gen_func` specified, `providers`, does not have a required_fields kwarg in its signature.

PR #8500, opened 2019-02-22
[WIPTEST] Fixup test_cloud_timelines

john-dupuy:
Removing BZ from `test_cloud_timeline_policy_event`, as it was discovered that this is not a CFME bug. This test can be remedied by changing the `emit_cmd` to `_power_off` rather than `_power_off_power_on`. 

Also, allowing tests to still run for [BZ 1670550](https://bugzilla.redhat.com/show_bug.cgi?id=1670550). Instead of checking against all `targets` we only check against the instance's timelines when this BZ applies. 

{{pytest: --use-provider complete cfme/tests/cloud/test_cloud_timelines.py }}


cfme/tests/cloud/test_cloud_timelines.py
```diff
@@ -272,17 +268,17 @@ def test_cloud_timeline_create_event(new_instance, soft_assert, azone):
         assignee: jdupuy
         initialEstimate: 1/4h
```
digitronik: Required:
Not in PR scope but can you please add `casecomponent`

cfme/tests/cloud/test_cloud_timelines.py
```diff
@@ -272,17 +268,17 @@ def test_cloud_timeline_create_event(new_instance, soft_assert, azone):
...
     """
-    targets = (new_instance, new_instance.provider, azone)
+    if BZ(1670550).blocks:
+        targets = (new_instance, )
```
digitronik: Optional:
I know this is `black` but not that much strictness :)

john-dupuy: This formatting is actually necessary for the test, since `_check_timelines` requires `targets` to be an iterable object.

cfme/control/explorer/policies.py
```diff
@@ -354,7 +349,7 @@ def assign_events(self, *events, **kwargs):
             events += self.assigned_events
         view = navigate_to(self, "Details")
         view.configuration.item_select("Edit this Policy's Event assignments")
-        view = self.create_view(EditPolicyEventAssignments)
+        view = self.create_view(EditPolicyEventAssignments, wait="5s")
         assert view.is_displayed
```
izapolsk: I suspect you don't need this assert since you wait for view above

cfme/tests/cloud/test_cloud_timelines.py
```diff
@@ -395,9 +391,13 @@ def test_cloud_timeline_delete_event(new_instance, soft_assert, azone):
     Polarion:
         assignee: jdupuy
         initialEstimate: 1/4h
+        casecomponent: Events
     """
     event = 'delete'
-    targets = (new_instance, new_instance.provider, azone)
+    if BZ(1670550).blocks:
```
izapolsk: please add everywhere short comment about the bug.

cfme/control/explorer/policies.py
john-dupuy: good point, I will remove that

cfme/tests/infrastructure/test_timelines.py
```diff
@@ -191,7 +196,14 @@ def _clone_vm(self):
 
     def _migrate_vm(self):
         logger.info('%r will be migrated.', self.vm.name)
-        return self.vm.migrate_vm()
+        view = navigate_to(self.vm, "Details")
+        vm_host = view.entities.summary('Relationships').get_text_of('Host')
+        hosts = [vds.name for vds in self.vm.provider.hosts.all() if vds.name not in vm_host]
+        if hosts:
```
mshriver: :+1:

cfme/tests/infrastructure/test_timelines.py
```diff
@@ -415,6 +427,7 @@ def test_infra_timeline_diagnostic(new_vm, soft_assert, mark_vm_as_appliance):
     vm_event.catch_in_timelines(soft_assert, targets)
 
 
+@pytest.mark.meta(blockers=[BZ(1622952, forced_streams=['5.10'])])
```
mshriver: No need for `forced_streams` here:

```
In [6]: BZ(1622952).blocks
Out[6]: True

In [7]: store.current_appliance.version
Out[7]: Version('5.10.1.2')
```

PR #8503, opened 2019-02-25
[RFR] Fix service bundles in bundle test

niyazRedhat:
Purpose or Intent
=================
Creating bundles in bundle is not supported now
https://bugzilla.redhat.com/show_bug.cgi?id=1678368

{{pytest: cfme/tests/services/test_generic_service_catalogs.py -k test_bundles_in_bundle -v}}

### DIFF
--- a/cfme/tests/services/test_generic_service_catalogs.py
+++ b/cfme/tests/services/test_generic_service_catalogs.py
@@ -68,6 +68,7 @@ def test_delete_catalog_item_deletes_service(appliance, catalog_item):
         service_catalogs.order()
 
 
+@pytest.mark.ignore_stream("5.10")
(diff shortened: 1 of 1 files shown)


cfme/tests/services/test_generic_service_catalogs.py
```diff
@@ -118,6 +118,8 @@ def test_service_generic_catalog_bundle(appliance, catalog_item):
     assert provision_request.is_succeeded(), msg
 
 
+@pytest.mark.uncollectif(lambda appliance: appliance.version > '5.9',
```
valaparthvi: Optional: You can use `@pytest.mark.ignore_stream("5.10")` if you don't want to collect this at all for 5.10 version.

PR #8504, opened 2019-02-25
[1LP][RFR] Fix event cloud object test

digitronik:

{{pytest: cfme/tests/automate/custom_button/test_cloud_objects.py::test_custom_button_events -vv}}

### DIFF
--- a/cfme/networks/cloud_network.py
+++ b/cfme/networks/cloud_network.py
@@ -24,7 +24,7 @@ from cfme.utils.wait import wait_for
 
 
 @attr.s
-class CloudNetwork(Taggable, BaseEntity):
+class CloudNetwork(Taggable, BaseEntity, CustomButtonEventsMixin):
     """Class representing cloud networks in cfme database"""
     in_version = ('5.8', version.LATEST)
     category = 'networks'
@@ -112,7 +112,7 @@ class CloudNetwork(Taggable, BaseEntity):
 
 
 @attr.s
-class CloudNetworkCollection(BaseCollection, CustomButtonEventsMixin):
+class CloudNetworkCollection(BaseCollection):
     """Collection object for Cloud Network object"""
     ENTITY = CloudNetwork
 
--- a/cfme/cloud/availability_zone.py
+++ b/cfme/cloud/availability_zone.py
@@ -75,7 +75,7 @@ class AvailabilityZoneDetailsAccordion(View):
         tree = ManageIQTree()
 
 
-class AvailabilityZoneView(BaseLoggedInPage, CustomButtonEventsMixin):
+class AvailabilityZoneView(BaseLoggedInPage):
     """Bare bones base view for page header matching"""
     @property
     def in_availability_zones(self):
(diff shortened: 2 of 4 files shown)


cfme/common/__init__.py
```diff
@@ -702,9 +702,17 @@ class CustomButtonEventsView(View):
 
     @property
     def is_displayed(self):
+        obj = self.context["object"]
+
+        if isinstance(obj.parent,
+                      self.browser.appliance.collections.object_store_containers.__class__):
```
john-dupuy: @digitronik is there a shorter way to get this `Class`? A shorter call would improve readability.

cfme/networks/cloud_network.py
```diff
@@ -24,7 +24,7 @@
 
 
 @attr.s
-class CloudNetwork(Taggable, BaseEntity):
+class CloudNetwork(Taggable, BaseEntity, CustomButtonEventsMixin):
     """Class representing cloud networks in cfme database"""
     in_version = ('5.8', version.LATEST)
```
john-dupuy: What is variable doing? I don't see it used in anywhere else in this file.

mshriver: https://github.com/ManageIQ/integration_tests/blob/master/cfme/common/__init__.py#L711

provides `get_button_events` method for this Entity.

cfme/common/__init__.py
mshriver: @john-dupuy there isn't really, because we're in a View class method, and need to get access to the appliance.

That being said, @digitronik its not really clear why this check is being made, what the purpose is here. Needs some in-line comments please.

digitronik: @john-dupuy: I agreed with you. I had two option that import class from storage model and second one collection. but now days, I love to use collection possibilities. I added comment.

@mshriver let me know if you found that i should go with first option. I will add change. :+1:

ganeshhubale:
LGTM!

cfme/common/__init__.py
mshriver: @digitronik thanks for adding comments here - This seems an odd lookup to me, that could easily break when the parent collection of an ObjectStoreContainer is not an ObjectStoreContainerCollection.

This could quickly grow to having to check multiple collection types, parent types, etc if the Entity doesn't have the 'name' attribute/property.

I think its reasonable to state in the docblock for `CustomButtonEventsMixin` that a `name` attribute/property is required to use the navigation/view provided by the mixin.  Classes inheriting it can then define that - in this case, `ObjectStoreContainer.name` could be a property that returns the key attribute.

cfme/networks/cloud_network.py
john-dupuy: Just for visibility, I was referring to the `in_version` variable - I am going to see if this variable is used elsewhere and whether or not it can be removed. Do you happen to know @digitronik?

cfme/common/__init__.py
digitronik: Done... added property.

PR #8505, opened 2019-02-25
[WIPTEST] fixed wait_for for fill function

lina-nikiforova:
{{ pytest: -v cfme/tests/configure/test_access_control.py::test_group_crud_with_tag[tag_expression-virtualcenter] }}

This test was failing as it waits for the wrong condition in case the program is `delete_whole_expression`. In this case `expression_text` should be `<new element>`, that means that whole expression was deleted.


dajoRH:
*CFME QE Bot*

lina-nikiforova:
This was fixed by another PR, the tests are passed and this one is no longer needed

PR #8509, opened 2019-02-26
[1LP][RFR] fix grouping tests by provider in test paralellizer

izapolsk:
fixing grouping tests by provider in order to speed up parallelized testing.
It turned out that provider "supportability" + DataProvider changes broke grouping tests by providers.

this PR means to fix this grouping issue.

### DIFF
--- a/cfme/markers/env_markers/provider.py
+++ b/cfme/markers/env_markers/provider.py
@@ -158,6 +158,15 @@ class DataProvider(object):
     type_name = attr.ib()
     version = attr.ib()
     klass = attr.ib(default=attr.Factory(data_provider_types, takes_self=True))
+    the_id = attr.ib(init=False)
+
+    def __attrs_post_init__(self):
+        ver = self.version if self.version else None
+        if ver:
+            the_id = "{}-{}".format(self.type_name, self.version)
+        else:
+            the_id = "{}".format(self.type_name)
+        object.__setattr__(self, "the_id", the_id)
 
     def one_of(self, *classes):
         return issubclass(self.klass, classes)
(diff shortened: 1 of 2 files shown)


cfme/markers/env_markers/provider.py
```diff
@@ -158,6 +158,15 @@ class DataProvider(object):
     type_name = attr.ib()
     version = attr.ib()
     klass = attr.ib(default=attr.Factory(data_provider_types, takes_self=True))
+    the_id = attr.ib(init=False)
+
+    def __attrs_post_init__(self):
+        ver = self.version if self.version else None
```
mshriver: `self.version or None`?

cfme/markers/env_markers/provider.py
```diff
@@ -158,6 +158,15 @@ class DataProvider(object):
     type_name = attr.ib()
     version = attr.ib()
     klass = attr.ib(default=attr.Factory(data_provider_types, takes_self=True))
+    the_id = attr.ib(init=False)
+
+    def __attrs_post_init__(self):
```
mshriver: Could this just be a DataProvider property instead of a custom init attr?

cfme/markers/env_markers/provider.py
izapolsk: agreed. this is copy/paste

cfme/markers/env_markers/provider.py
izapolsk: yup, do you think it's necessary ? 
we use attrs. so, I try using its stuff instead.

mshriver: I don't see a need to set it at DataProvider instantiation, which is when I would use this. To me, just a regular property is more straightforward.

cfme/markers/env_markers/provider.py
```diff
@@ -158,6 +158,15 @@ class DataProvider(object):
...
+        ver = self.version if self.version else None
+        if ver:
+            the_id = "{}-{}".format(self.type_name, self.version)
+        else:
+            the_id = "{}".format(self.type_name)
+        object.__setattr__(self, "the_id", the_id)
```
digitronik: :+1:

cfme/markers/env_markers/provider.py
```diff
@@ -159,6 +160,13 @@ class DataProvider(object):
     version = attr.ib()
     klass = attr.ib(default=attr.Factory(data_provider_types, takes_self=True))
 
+    @cached_property
```
mshriver: :ok_hand: :rocket:

cfme/markers/env_markers/provider.py
```diff
@@ -1,3 +1,4 @@
+from cached_property import cached_property
```
digitronik: @izapolsk this is third part package... I think its reason behind travis failure.

PR #8511, opened 2019-02-26
[1LP][RFR] fixed active snapshot detection for RHEV

lina-nikiforova:
{{ pytest: -v cfme/tests/infrastructure/test_snapshot.py::test_verify_revert_snapshot --long-running }}

According to the fix mentioned here https://bugzilla.redhat.com/show_bug.cgi?id=1561618#c6
"Active VM" is always the active snapshot, so instead of

before:
image

it looks like:
image

### DIFF
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -648,7 +648,16 @@ class InfraVm(VM):
             title = getattr(self, self.parent_vm.provider.SNAPSHOT_TITLE)
             view = navigate_to(self.parent_vm, 'SnapshotsAll')
             root_item = view.tree.expand_path(self.parent_vm.name)
-            return has_child(view.tree, '{} (Active)'.format(title), root_item)
+            from cfme.infrastructure.provider.rhevm import RHEVMProvider
+
+            if self.parent_vm.provider.one_of(RHEVMProvider):
+                child = view.tree.child_items(root_item)
+                last_snapshot = view.tree.child_items(child[0])[0]
+                return (len(child) == 1 and
+                        child[0].text == 'Active VM (Active)' and
+                        last_snapshot.text == title)
+            else:
+                return has_child(view.tree, '{} (Active)'.format(title), root_item)
 
         def create(self, force_check_memory=False):
             """Create a snapshot"""


mwperina:
Boris, could you please reply?

@borod108

cfme/infrastructure/virtual_machines.py
```diff
@@ -648,7 +648,16 @@ def active(self):
             title = getattr(self, self.parent_vm.provider.SNAPSHOT_TITLE)
             view = navigate_to(self.parent_vm, 'SnapshotsAll')
             root_item = view.tree.expand_path(self.parent_vm.name)
-            return has_child(view.tree, '{} (Active)'.format(title), root_item)
+            from cfme.infrastructure.provider.rhevm import RHEVMProvider
+
+            if self.parent_vm.provider.one_of(RHEVMProvider):
```
izapolsk: REQUIRED: I think you need to add BZ here. otherwise all rhv snapshot related tests will fail when it is fixed.

borod108: Can you elaborate please?

lina-nikiforova: @izapolsk I checked `5.9` and there's no such thing `'{} (Active)'.format(title)` for snapshots there as well. You only see a bunch of snapshots and `Active VM (Active)` which is the same as in RHEV itself.

I made this change only for RHEV provider.

lina-nikiforova: @digitronik I checked this test for `vsphere65-nested` and it's different to RHEV, that's why no change for it is required.

1. When there're no snapshots, it shows 0 for `vsphere65-nested` but 1 for RHEV as it always has `Active VM (Active)` <- that's exactly what is displayed in RHEV itself

2. Using `vsphere65-nested` it shows which snapshot is active

image

whereas for RHEV it's always `Active VM (Active)` <- how it's displayed in RHEV itself

digitronik: @lina-nikiforova just put one more condition with this check. so as this BZ on `rhv` fixed its should fallow older way like.
```
if BZ(1561618, forced_streams=['5.10']).blocks and self.parent_vm.provider.one_of(RHEVMProvider):
```

lina-nikiforova: @digitronik this BZ is about different issue that I can't reproduce in 5.10 and no snapshots are missing. 
What I am saying it's always `Active VM (Active)` for RHEV, so I don't understand why it will go older way depending on this BZ.

izapolsk:
@lina-nikiforova, PRT is empty. I have restarted it. Please let me know when PRT is over. I'll merge your PR

PR #8515, opened 2019-02-27
[RFR] fixed fixture scope mismatch

lina-nikiforova:
{{ pytest: -v cfme/tests/infrastructure/test_vm_reconfigure.py::test_vm_reconfig_add_remove_hw_hot  --long-running }}

Fixed the error
```
ScopeMismatch: You tried to access the 'function' scoped fixture 'provider' with a 'module' scoped request object, involved factories
cfme/fixtures/provider.py:283: def setup_provider_modscope(request, provider)
../.cfme/lib/python2.7/site-packages/_pytest/fixtures.py:242: def get_direct_param_fixture_func(request)
```

PR #8516, opened 2019-02-27
[1LP][RFR] Fix scopes for snapshot tests

apagac:
This is a fix for snapshot tests.

Something has changed in the framework and ScopeMismatch errors started to appear. Solved by changing scope from module to function.

{{ pytest: -v --long-running cfme/tests/infrastructure/test_snapshot.py --use-provider vsphere65-nested }}

### DIFF
--- a/cfme/tests/infrastructure/test_snapshot.py
+++ b/cfme/tests/infrastructure/test_snapshot.py
@@ -29,7 +29,7 @@ pytestmark = [
 ]
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture(scope="function")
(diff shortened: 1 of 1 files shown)


lina-nikiforova:
thanks, I faced this as well when checking tests for RHV.

I also have a similar PR for the other test #8515

apagac:
Moving to RFR. Failures occurring when reverting snapshot will be addressed in different PR.

No ScopeMismatch errors present.

lina-nikiforova:
`ScopeMismatch` error is fixed, so moving this PR further

PR #8517, opened 2019-02-27
[1LP][RFR] fixed test_create_without_description for 5.10

lina-nikiforova:
{{ pytest: -v cfme/tests/infrastructure/test_snapshot.py::test_create_without_description --long-running }}

NOTE: here I applied the same changes as in #8516 to get rid of the scope mismatch.

In `5.10` there's no longer error message to check for as "Create" button is simply disabled if there's no description.

image

### DIFF
--- a/cfme/tests/infrastructure/test_snapshot.py
+++ b/cfme/tests/infrastructure/test_snapshot.py
@@ -18,7 +18,6 @@ from cfme.utils.generators import random_vm_name
 from cfme.utils.log import logger
 from cfme.utils.path import data_path
 from cfme.utils.ssh import SSHClient
-from cfme.utils.version import current_version
 from cfme.utils.wait import wait_for
 
 pytestmark = [
@@ -27,13 +26,6 @@ pytestmark = [
     test_requirements.snapshot,
     pytest.mark.provider([RHEVMProvider, VMwareProvider], scope="module"),
 ]
-
-
-@pytest.fixture(scope="function")
-def domain(request, appliance):
-    dom = DomainCollection(appliance).create(name=fauxfactory.gen_alpha(), enabled=True)
-    request.addfinalizer(dom.delete_if_exists)
-    return dom
 
 
 def provision_vm(provider, template):
@@ -141,11 +133,16 @@ def test_create_without_description(small_test_vm):
         initialEstimate: 1/4h
         casecomponent: Infra
     """
-    snapshot = new_snapshot(small_test_vm, has_name=False, create_description=False)
-    with pytest.raises(AssertionError):
-        snapshot.create()
-    view = snapshot.parent_vm.create_view(InfraVmSnapshotAddView)
-    view.flash.assert_message('Description is required')
+    if small_test_vm.appliance.version >= '5.10':
+        # In 5.10 it's not possible to create a snapshot w/o description,"Create" button is disabled
+        view = navigate_to(small_test_vm, 'SnapshotsAdd')
+        assert view.create.disabled
+    else:
+        snapshot = new_snapshot(small_test_vm, has_name=False, create_description=False)
+        with pytest.raises(AssertionError):
+            snapshot.create()
+        view = snapshot.parent_vm.create_view(InfraVmSnapshotAddView)
+        view.flash.assert_message('Description is required')
 
 
 @pytest.mark.provider([VMwareProvider], override=True)


cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -29,7 +29,7 @@
 ]
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture(scope="function")
```
john-dupuy: Optional: no need to pass `scope="function"` since `pytest.fixture` is function scoped by default.

digitronik: yup... as appliance only inherited.

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -29,7 +29,7 @@
...
 def domain(request, appliance):
```
digitronik: Suggestion:
As per me we have one global fixture for `domain`
`https://github.com/ManageIQ/integration_tests/blob/master/cfme/fixtures/automate.py#L6`

Please have look it will helpful to remove this stuff. let me know your view on this.

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -141,11 +141,16 @@ def test_create_without_description(small_test_vm):
         initialEstimate: 1/4h
         casecomponent: Infra
     """
-    snapshot = new_snapshot(small_test_vm, has_name=False, create_description=False)
-    with pytest.raises(AssertionError):
-        snapshot.create()
-    view = snapshot.parent_vm.create_view(InfraVmSnapshotAddView)
-    view.flash.assert_message('Description is required')
+    if small_test_vm.appliance.version > '5.10':
```
digitronik: I would use >= not simply >.

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -141,11 +141,16 @@ def test_create_without_description(small_test_vm):
...
+        # In 5.10 it's not possible to create a snapshot w/o description,"Create" button is disabled
+        view = navigate_to(small_test_vm, 'SnapshotsAdd')
+        assert view.create.disabled
+    else:
+        snapshot = new_snapshot(small_test_vm, has_name=False, create_description=False)
+        with pytest.raises(AssertionError):
+            snapshot.create()
+        view = snapshot.parent_vm.create_view(InfraVmSnapshotAddView)
```
digitronik: Optional:
I am not tested `shanpshot` as entity but I think it should give you direct `create_view` method... Please check it one time.

cfme/tests/infrastructure/test_snapshot.py
lina-nikiforova: @john-dupuy thanks!

cfme/tests/infrastructure/test_snapshot.py
lina-nikiforova: thanks, I checked and it's not the case. I can access it only through `InfraVm`

cfme/tests/infrastructure/test_snapshot.py
lina-nikiforova: thanks! I ran a test `test_create_snapshot_via_ae` with a global fixture and it works

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -29,13 +28,6 @@
 ]
 
 
-@pytest.fixture(scope="function")
-def domain(request, appliance):
```
mshriver: https://github.com/ManageIQ/integration_tests/pull/8517/files#diff-11b93f3b87782fa7e6964884e93593f2R444

This fixture is still used by a test below - do you expect it to be covered by another global fixture somewhere?

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -141,11 +133,16 @@ def test_create_without_description(small_test_vm):
...
     """
-    snapshot = new_snapshot(small_test_vm, has_name=False, create_description=False)
-    with pytest.raises(AssertionError):
-        snapshot.create()
-    view = snapshot.parent_vm.create_view(InfraVmSnapshotAddView)
-    view.flash.assert_message('Description is required')
+    if small_test_vm.appliance.version >= '5.10':
+        # In 5.10 it's not possible to create a snapshot w/o description,"Create" button is disabled
+        view = navigate_to(small_test_vm, 'SnapshotsAdd')
+        assert view.create.disabled
```
mshriver: I think you should assert that the description field is empty before asserting that the create button is disabled.

cfme/tests/infrastructure/test_snapshot.py
mshriver: Nevermind, found the previous discussions on this that were resolved. I see its provided by another module.  DomainCollection is getting used otherwise directly as well.

PR #8518, opened 2019-02-27
[1LP][RFR] New test for automate: Checking schema creation without 'Type'

ganeshhubale:
- Adding test case for checking schema creation without providing 'Type'
- This test case is related to [BZ:1365442](https://bugzilla.redhat.com/show_bug.cgi?id=1365442) 

{{ pytest: cfme/tests/automate/test_class.py -k 'test_automate_schema_field_without_type' -vv }}

### DIFF
--- a/cfme/tests/automate/test_class.py
+++ b/cfme/tests/automate/test_class.py
@@ -3,6 +3,7 @@ import fauxfactory
 import pytest
 
 from cfme import test_requirements
+from cfme.automate.explorer.klass import ClassSchemaEditView
 from cfme.utils.update import update
 
 pytestmark = [test_requirements.automate]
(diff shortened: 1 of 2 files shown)


cfme/tests/automate/test_class.py
```diff
@@ -169,3 +170,34 @@ def test_class_display_name_unset_from_ui(get_namespace):
     with update(a_class):
         a_class.display_name = ""
     assert a_class.exists
+
+
+@test_requirements.automate
```
valaparthvi: No need to put this, requirement is already satisfied by pytestmark

ganeshhubale: Thanks @valaparthvi

cfme/tests/automate/test_class.py
```diff
@@ -169,3 +170,33 @@ def test_class_display_name_unset_from_ui(get_namespace):
...
+
+    Bugzilla:
+        1365442
+    """
+    with pytest.raises(AssertionError):
+        klass.schema.add_fields({'name': 'execute', 'data_type': 'String'})
+    view = klass.create_view(ClassSchemaEditView)
+    assert view.schema.save_button.disabled
+    assert view.schema.reset_button.disabled
+    assert not view.schema.cancel_button.disabled
```
john-dupuy: Question: Is there an instance when the cancel button is disabled? Why do we need this assertion?

ganeshhubale: @john-dupuy 
- If you are talking about automate ```instance``` then I would like to tell you that this is related to only ```schema```  of automate class.
-  So here in this test case, I just want to make sure that these buttons are in expected States while creating schema with no ```TYPE```

john-dupuy: @ganeshhubale I don't mean automate `instance`, I mean that I have never seen a case where the cancel button is disabled, since the user is always able to cancel their edits. Regardless, it doesn't hurt anything to have that assertion in there.

digitronik: @john-dupuy :100: 
@john-dupuy more assertion in automation; I like to accept. this is normal behavior that `cancel button never disable` but check is in positive side.

PR #8519, opened 2019-02-27
[1LP][RFR] Separate test_assign_and_unassign_tag for different providers and minimize parametrization 

valaparthvi:
This PR brings the following change:
1. Separate `test_assign_and_unassign_tag` for cloud and infra providers since few collections work only with specific providers and supported only on specific appliance versions.
2. Minimize parametrization for `test_views.py` and `test_generate_reports.py` since it's not required.

{{ pytest: cfme/tests/configure/test_tag.py::TestTagsViaREST::test_assign_and_unassign_tag cfme/tests/intelligence/reports/test_views.py cfme/tests/intelligence/reports/test_generate_report.py --use-template-cache -sqvvv }}

### DIFF
--- a/cfme/tests/configure/test_tag.py
+++ b/cfme/tests/configure/test_tag.py
@@ -2,8 +2,10 @@
 import fauxfactory
 import pytest
 
+from cfme.cloud.provider import CloudProvider
 from cfme.infrastructure.provider import InfraProvider
 from cfme.markers.env_markers.provider import ONE
+from cfme.markers.env_markers.provider import ONE_PER_CATEGORY
 from cfme.rest.gen_data import categories as _categories
 from cfme.rest.gen_data import service_templates as _service_templates
 from cfme.rest.gen_data import tags as _tags
@@ -15,6 +17,12 @@ from cfme.utils.rest import delete_resources_from_collection
 from cfme.utils.rest import delete_resources_from_detail
 from cfme.utils.update import update
 from cfme.utils.wait import wait_for
+
+
+CLOUD_COLLECTION = ["availability_zones", "cloud_networks",
+        "cloud_networks", "cloud_subnets", "flavors", "network_routers", "security_groups"]
+INFRA_COLLECTION = ["clusters", "hosts", "data_stores", "providers", "resource_pools",
+        "services", "service_templates", "tenants", "vms", ]
 
 pytestmark = [
     pytest.mark.provider(classes=[InfraProvider], selector=ONE),
@@ -230,10 +238,19 @@ class TestTagsViaREST(object):
         assert_response(appliance, http_status=400)
 
     @pytest.mark.tier(3)
-    @pytest.mark.parametrize(
-        "collection_name", ["clusters", "hosts", "data_stores", "providers", "resource_pools",
-        "services", "service_templates", "tenants", "vms", "availability_zones", "cloud_networks",
-        "cloud_networks", "cloud_subnets", "flavors", "network_routers", "security_groups"])
+    @pytest.mark.provider(
+        [CloudProvider, InfraProvider], selector=ONE_PER_CATEGORY, override=True
+    )
+    @pytest.mark.parametrize("collection_name", INFRA_COLLECTION + CLOUD_COLLECTION)
+    @pytest.mark.uncollectif(
+        lambda appliance, collection_name, provider: (
+            provider.one_of(CloudProvider) and collection_name in INFRA_COLLECTION
+        )
+        or (
+            (provider.one_of(InfraProvider) or appliance.version < "5.10")
+            and collection_name in CLOUD_COLLECTION
+        )
+    )
     def test_assign_and_unassign_tag(self, appliance, tags_mod, provider, services_mod,
             service_templates, tenants, vm, collection_name):
         """Tests assigning and unassigning tags.
--- a/cfme/tests/configure/test_tag.py
+++ b/cfme/tests/configure/test_tag.py
@@ -19,11 +19,25 @@ from cfme.utils.update import update
 from cfme.utils.wait import wait_for
 
 
-CLOUD_COLLECTION = ["availability_zones", "cloud_networks",
-        "cloud_networks", "cloud_subnets", "flavors", "network_routers", "security_groups"]
-INFRA_COLLECTION = ["clusters", "hosts", "data_stores", "providers", "resource_pools",
-        "services", "service_templates", "tenants", "vms", ]
-
+CLOUD_COLLECTION = [
+    "availability_zones",
+    "cloud_networks",
+    "cloud_subnets",
+    "flavors",
+    "network_routers",
+    "security_groups",
+]
+INFRA_COLLECTION = [
+    "clusters",
+    "hosts",
+    "data_stores",
+    "providers",
+    "resource_pools",
+    "services",
+    "service_templates",
+    "tenants",
+    "vms",
+]
 pytestmark = [
     pytest.mark.provider(classes=[InfraProvider], selector=ONE),
     pytest.mark.usefixtures('setup_provider')
--- a/cfme/tests/intelligence/reports/test_views.py
+++ b/cfme/tests/intelligence/reports/test_views.py
@@ -2,10 +2,8 @@
 import pytest
 
 from cfme import test_requirements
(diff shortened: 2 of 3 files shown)


cfme/tests/configure/test_tag.py
```diff
@@ -230,10 +238,19 @@ def test_create_tag_with_wrong_arguments(self, appliance):
...
+        [CloudProvider, InfraProvider], selector=ONE_PER_CATEGORY, override=True
+    )
+    @pytest.mark.parametrize("collection_name", INFRA_COLLECTION + CLOUD_COLLECTION)
+    @pytest.mark.uncollectif(
+        lambda appliance, collection_name, provider: (
+            provider.one_of(CloudProvider) and collection_name in INFRA_COLLECTION
+        )
+        or (
+            (provider.one_of(InfraProvider) or appliance.version < "5.10")
+            and collection_name in CLOUD_COLLECTION
```
john-dupuy: Just checking my understanding of this. So here, you are uncollecting if the provider is a `CloudProvider` and the `collection` is related to an Infra provider, OR if the provider is an `InfraProvider` and the `collection` is related to a Cloud provider OR if the version of the appliance is "5.9" and the `collection` is related to a Cloud provider. 

Is there a cleaner way to do this, so that the tests aren't collected in the first place? I think the version uncollection is okay, but their may be a way to avoid the other uncollections.

cfme/tests/configure/test_tag.py
```diff
@@ -16,6 +18,12 @@
 from cfme.utils.update import update
 from cfme.utils.wait import wait_for
 
+
+CLOUD_COLLECTION = ["availability_zones", "cloud_networks",
+        "cloud_networks", "cloud_subnets", "flavors", "network_routers", "security_groups"]
```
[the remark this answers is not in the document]
valaparthvi: Missed out on that one. Removing the extra collection.

cfme/tests/configure/test_tag.py
valaparthvi: Um, I could write separate test cases for both the providers, but that's not the best option. There's one more way - using `pytest_generate_tests(metafunc)` to parametrize test, but afaik we're avoiding it.

cfme/tests/intelligence/reports/test_views.py
```diff
@@ -46,6 +44,10 @@ def test_report_view(report, view_mode):
         caseimportance: high
         initialEstimate: 1/6h
     """
-    view = navigate_to(report, 'Details')
+    try:
+        view = navigate_to(report, 'Details')
+    except CandidateNotFound:
+        report.browser.refresh()
```
lina-nikiforova: Q: can this be achieved using `wait_for`?

valaparthvi: I receive CandidateNotFound error while navigating to that page, I am not sure how to use `wait_for` in this case, can you show me an example of it?

dajoRH:
Lint report for commit 17cc76b46bab73e818c122f0996c100376cfd9c4:

`cfme/tests/intelligence/reports/test_views.py`:
- :red_circle: Line 6:1: **F401** *'cfme.exceptions.CandidateNotFound' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/intelligence/reports/test_views.py
valaparthvi: Looks like it didn't work @digitronik. Do you have any other way on your mind?

lina-nikiforova: @valaparthvi I'm curious, if you navigate manually to this page, are all elements present at the first time?

valaparthvi: @lina-nikiforova There are no problems with other elements of the page, so they are present the first time, but here's the thing: If you check this [video](https://up1-vrutkovs.int.open.paas.redhat.com/#HBTUFO0ssN7NIDLt7Cmedw), when the report is queued, the report name shown in tree in `Generating report`, the page is being continuously refreshed until a proper name of report is generated, this name appears to be updated in the saved_reports but sometimes it does not reflect in the tree, which is why we encounter CandidateNotFound error, for which we need to refresh the page.

valaparthvi: @digitronik I have added exception handling inside the navigation step and you might see some piece of code getting repeated, but I am working on making it DRY.

digitronik: @valaparthvi I think this is not way... I understand you need browser refresh.
last try which is in my mind... can you please try this.
`prerequisite` is `Details` page of `report`. add resetter for detail like
```
        self.view.browser.refresh()
```

digitronik:
@valaparthvi can you please check `5.10` run?

valaparthvi:
@digitronik It is failing because there is an issue with rhos13 provider, I have talked with @mnadeem92 about it and we're working on it, the test should pass once that is done.

cfme/intelligence/reports/reports.py
```diff
@@ -699,6 +699,9 @@ def step(self, *args, **kwargs):
         )
         self.view.report_info.select()
 
+    def resetter(self, *args, **kwargs):
+        self.obj.browser.refresh()
```
mshriver: Instead of doing a browser refresh in resetter, the recently added `force` kwarg should be used on navigation - this will force re-navigation from the prerequisite, as this is more reliable than a refresh. Browser refreshes can have a significant impact on test execution time.

cfme/tests/intelligence/reports/test_views.py
```diff
@@ -46,6 +42,8 @@ def test_report_view(report, view_mode):
         caseimportance: high
         initialEstimate: 1/6h
     """
-    view = navigate_to(report, 'Details')
+    view = navigate_to(report, "Details")
```
mshriver: Here is where you can pass `force=True` to force navigation instead of a browser refresh.  Since this is the first step in the test, and you added a refresh resetter, I'm assuming the fixture is leaving the browser on this page already, and you want to refresh it.

If you really do want/need a browser refresh in this instance, you can make a browser refresh call here, instead of impacting all navigation to Report details with a resetter method that refreshes.

The `force` option should update the view, through navigation, refreshing the content in the process to account for the fixture behavior.

valaparthvi: @mshriver We've already done all that you mentioned here, refer this [review](https://github.com/ManageIQ/integration_tests/pull/8519#discussion_r261662321) by Angelina. I guess refreshing the browser for this particular test might be the way to go.

cfme/tests/intelligence/reports/test_views.py
```diff
@@ -1,41 +1,38 @@
...
     ).queue(wait_for_finish=True)
     yield report
     if report.exists:
         report.delete()
 
 
 @pytest.mark.rhv3
-@pytest.mark.parametrize('view_mode', ['Hybrid View', 'Graph View', 'Tabular View'])
-@pytest.mark.meta(blockers=[BZ(1401560)])
+@pytest.mark.parametrize("view_mode", ["Hybrid View", "Graph View", "Tabular View"])
```
digitronik: Optional:
@valaparthvi I think better to set proper ids for this parametrization in test name ["Hybrid view"] will come. Totally optional but you can separate  by `-` or `_`.

cfme/tests/intelligence/reports/test_views.py
```diff
@@ -46,6 +43,12 @@ def test_report_view(report, view_mode):
         caseimportance: high
         initialEstimate: 1/6h
     """
-    view = navigate_to(report, 'Details')
+    try:
+        view = navigate_to(report, "Details")
+    except CandidateNotFound:
+        report.browser.refresh()
+        view = navigate_to(report, "Details")
```
digitronik: @valaparthvi so this is full work around. @mshriver don't like `resetter` then Please add one Note Inline comment so in feature we can quick understand whats problem.

dajoRH:
Lint report for commit 9821b4512638c94fbaa92d10f17d6b3f570c6668:

`cfme/tests/intelligence/reports/test_views.py`:
- :red_circle: Line 60:5: **F631** *assertion is always true, perhaps remove parentheses?*

**No commit flaws detected**.

*CFME QE Bot*

PR #8525, opened 2019-03-01
[1LP][RFR] New test: Testing automate simulate 'Retry' button using state machines

ganeshhubale:
- This test case create schema which includes state machines
- Adding method for retry button and to execute it using 'State' machines
- Adding 'retry_button' to the 'Automatesimulation' view page
- Checking whether 'Retry' button is displayed after simulation

{{ pytest: cfme/tests/automate/test_method.py -k 'test_automate_simulate_retry' -v }}

### DIFF
--- a/cfme/tests/automate/test_method.py
+++ b/cfme/tests/automate/test_method.py
@@ -4,14 +4,24 @@ import pytest
 
 from cfme import test_requirements
 from cfme.automate.explorer.klass import ClassDetailsView
+from cfme.automate.simulation import simulate
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.update import update
 
-pytestmark = [test_requirements.automate]
+pytestmark = [test_requirements.automate, pytest.mark.tier(2)]
+
+
+@pytest.fixture(scope='function')
+def original_class(domain):
+    # Take the 'Request' class and copy it for own purpose.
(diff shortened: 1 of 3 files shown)


cfme/tests/automate/test_method.py
```diff
@@ -4,12 +4,23 @@
...
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.update import update
 
 pytestmark = [test_requirements.automate]
 
 
+@pytest.fixture(scope='function')
+def original_class(domain):
+    # Take the 'Request' class and copy it for own purpose.
+    domain.appliance.collections.domains.instantiate(name="ManageIQ").namespaces.instantiate(
```
valaparthvi: If `domain` is an entity of `Domains` collection, you could use `domain.parent.instantiate`, instead of `domain.appliance.collection.instantiate`

ganeshhubale: Awesome suggestion to improvise code!
Thanks @valaparthvi

ganeshhubale:
Agree! But now in this file, we only have 4 test cases. 3 of them are of tier(2). I am updating with this suggestion. But it may change if more test cases of tier(1) are getting added to this file. 
>     1. Optional: If all except one test case has tier 2, you can add tier in pytestmark.

dajoRH:
*CFME QE Bot*

PR #8529, opened 2019-03-04
[1LP][WIPTEST] Restore edit element nav

digitronik:
Purpose or Intent
=================
In old implementation means before #8462. edit naviagtion:
```
def step(self)
    def step(self):
        self.prerequisite_view.element.edit_element("Text Box")
```
step under steps means fake call. Expected navigation never gone run.

For now, restoring. FA owner need to decide this navigation needed or not. As per my observation its not useful, but I would like FA owner take look on this.

{{pytest: cfme/tests/automate/test_service_dialog.py -k "test_reorder_unsaved_elements or test_reorder_elements" -vv}}

### DIFF
--- a/cfme/automate/dialogs/dialog_element.py
+++ b/cfme/automate/dialogs/dialog_element.py
@@ -138,9 +138,10 @@ class Element(BaseEntity):
 
     def add_another_element(self, element):
         """Method to add element."""
-        view = navigate_to(self, 'Edit')
+        navigate_to(self.dialog, 'Edit')
         dragged_element = element.get('element_information').get('choose_type')
         box_label = element.get('element_information').get('ele_label')
+        view = self.create_view(EditElementView)
         view.dd.drag_and_drop(dragged_element, box_label)
         view.element.edit_element(dragged_element)
         view.fill(element)
@@ -159,7 +160,8 @@ class Element(BaseEntity):
             second_element: The second element to be added to the dialog.
             element_data: Already existing first element's data.
         """
-        view = navigate_to(self, 'Edit')
+        navigate_to(self.dialog, 'Edit')
+        view = self.create_view(EditElementView)
         if add_element:
             dragged_element = second_element.get('element_information').get('choose_type')
             box_label = second_element.get('element_information').get('ele_label')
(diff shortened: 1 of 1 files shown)


mshriver:
This either needs to be written as an issue, or resolved in the included commits. Just commenting out the step method entirely is not a 'fix', and without a `step` method there is no purpose to the navigation class existing at all.

cfme/automate/dialogs/dialog_element.py
```diff
@@ -159,7 +160,8 @@ def reorder_elements(self, add_element, second_element, element_data):
             second_element: The second element to be added to the dialog.
             element_data: Already existing first element's data.
         """
-        view = navigate_to(self, 'Edit')
+        navigate_to(self.dialog, 'Edit')
+        view = self.create_view(EditElementView)
```
mshriver: We shouldn't have to create the view after navigation, it should be returned by navigate_to.  This tells me there is a strange navigation pattern going on in the framework.

This looks like a method that should exist on the `Dialog` entity, not here on the `Element` itself.  Is this editing the dialog, or editing the element?  Why are we expecting an `EditElementView` when we navigated to `Dialog` 'Edit' destination.

digitronik: @mshriver method better to called widget which is responsible for `element` edit is part of `element` view. but upto now I understand that to edit element
1. need to edit dialog
2. from dialog edit page only we need to edit element. 

so element edit facility should be part of `DialogEditView`. I just avoid this to add normal fix :D. I thing FA owner should take care of major changes related to view. but I am on it. let me try.

mshriver: Gotcha, yeah I would rather actually fix the view structure problem, either you or the FA owner (ideally working together! :) ) taking on that task of better structuring these views. 

Sounds like EditElementView just needs to be nested in DialogEdit view, or that a conditional view needs to be used.

digitronik:
@mshriver @sshveta @niyazRedhat closing is PR... as to fix those two tests need to restructure... `Collection`, `View` and then `methods` also need to take care about all dependent `tests`. This model page used by most of tests. Its good that FA owner should take care of these failures.

sshveta:
Sounds Good Nikhil . @niyazRedhat Please work on this with @digitronik help.

niyazRedhat:
@sshveta sure

PR #8530, opened 2019-03-04
[1LP][RFR] Update tree for aws rbac

apagac:
This is a fix for test_aws_iam_auth_and_roles.py for 5.10

The navigation tree changed in 5.10 and the test was expecting roles that weren't available in our code. Roles are hardcoded, so I updated them to reflect the UI navigation possibilities for various roles. This fixes all test_group_roles errors for 5.10.

{{ pytest: -v --long-running cfme/tests/integration/test_aws_iam_auth_and_roles.py }}

### DIFF
--- a/cfme/roles.py
+++ b/cfme/roles.py
@@ -300,8 +300,9 @@ role_access_ui_510z = {
                            'Replicators', 'Projects', 'Topology'],
             'Infrastructure': ['Datastores', 'Networking', 'Providers', 'Virtual Machines', 'Hosts',
                                'Clusters', 'Topology', 'PXE', 'Resource Pools'],
-            'Physical Infrastructure': ['Providers', 'Racks', 'Switches', 'Servers', 'Topology'],
-            'Migration': None},
+            'Physical Infrastructure': ['Overview', 'Providers', 'Chassis', 'Racks', 'Switches',
+                                        'Servers', 'Storages', 'Topology'],
+            'Migration': ['Migration Plans', 'Infrastructure Mappings', 'Migration Settings']},
         'Configuration': ['Management'],
         'Control': ['Import / Export', 'Log', 'Explorer', 'Simulation'],
         'Monitor': {
@@ -312,7 +313,8 @@ role_access_ui_510z = {
         'Red Hat Insights': ['Rules', 'Overview', 'Inventory', 'Actions'],
         'Services': ['Requests', 'Workloads', 'Catalogs', 'My Services'],
         'Storage': {
-            'Block Storage': ['Volume Snapshots', 'Managers', 'Volume Backups', 'Volumes'],
+            'Block Storage': ['Volume Snapshots', 'Managers', 'Volume Backups', 'Volumes',
+                              'Volume Types'],
             'Object Storage': ['Managers', 'Object Store Containers', 'Object Store Objects']}
     },
     'evmgroup-administrator': {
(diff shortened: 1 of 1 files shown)


cfme/roles.py
```diff
@@ -300,8 +300,9 @@ def _remove_from_all(roles, r_page):
                            'Replicators', 'Projects', 'Topology'],
             'Infrastructure': ['Datastores', 'Networking', 'Providers', 'Virtual Machines', 'Hosts',
                                'Clusters', 'Topology', 'PXE', 'Resource Pools'],
-            'Physical Infrastructure': ['Providers', 'Racks', 'Switches', 'Servers', 'Topology'],
-            'Migration': None},
+            'Physical Infrastructure': ['Overview', 'Providers', 'Chassis', 'Racks', 'Switches',
+                                        'Servers', 'Storages', 'Topology'],
```
john-dupuy: Q: Does the order matter here? I noticed that `Switches` comes after `Storages` but before `Topology` in CFME.

cfme/roles.py
```diff
@@ -312,7 +313,8 @@ def _remove_from_all(roles, r_page):
         'Red Hat Insights': ['Rules', 'Overview', 'Inventory', 'Actions'],
         'Services': ['Requests', 'Workloads', 'Catalogs', 'My Services'],
         'Storage': {
-            'Block Storage': ['Volume Snapshots', 'Managers', 'Volume Backups', 'Volumes'],
+            'Block Storage': ['Volume Snapshots', 'Managers', 'Volume Backups', 'Volumes',
```
john-dupuy: Same Q here.

cfme/roles.py
```diff
@@ -357,8 +361,9 @@ def _remove_from_all(roles, r_page):
         'Compute': {
             'Clouds': ['Instances'],
             'Infrastructure': ['Datastores', 'Providers', 'Virtual Machines', 'Hosts', 'Clusters',
-                               'PXE', 'Resource Pools'],
-            'Physical Infrastructure': ['Providers', 'Racks', 'Switches', 'Servers', 'Topology']},
+                               'Networking', 'PXE', 'Resource Pools'],
+            'Physical Infrastructure': ['Providers', 'Chassis', 'Racks', 'Switches', 'Servers',
+                                        'Storages', 'Topology']},
         'Control': ['Explorer', 'Log', 'Simulation'],
```
john-dupuy: Q: Why no `Import/Export` for `Control` here?

cfme/roles.py
apagac: The test that uses this is using DeepDiff with 'ignore_order=True', so the order does not matter. I tried however to add the items as they are in CFME UI, just for looks.

cfme/roles.py
apagac: Because there is no 'Import/Export' option under 'Control' when you are logged in as a member of group 'evmgroup-auditor'.
You can see for yourself by creating new user in CFME and assigning group 'Evmgroup-auditor' to the user, then logging in as newly created user and navigating to 'Control'.

cfme/roles.py
```diff
@@ -412,7 +420,8 @@ def _remove_from_all(roles, r_page):
             'Clouds': ['Instances'],
             'Infrastructure': ['Datastores', 'Providers', 'Virtual Machines', 'Hosts', 'Clusters',
                                'Resource Pools'],
-            'Physical Infrastructure': ['Providers', 'Racks', 'Switches', 'Servers', 'Topology']},
+            'Physical Infrastructure': ['Providers', 'Chassis', 'Racks', 'Switches', 'Servers',
```
sshveta: In 5.10 the order is Server , Storage , Switch , Topology , if the order matters here

cfme/roles.py
john-dupuy: Makes sense -- thanks for the response!

cfme/roles.py
mshriver: @apagac great minds think alike, this was my exact thought process when creating this.

cfme/roles.py
mshriver: Ultimately it doesn't - explained in a comment above:

https://github.com/ManageIQ/integration_tests/pull/8530/files#r262910147

PR #8531, opened 2019-03-05
[1LP][RFR] clone dialog is broken. uncollecting tests until those are fixed

izapolsk:
{{ pytest: --long-running cfme/tests/control/test_actions.py::test_action_cancel_clone cfme/tests/infrastructure/test_vm_clone.py::test_vm_clone}}

### DIFF
--- a/cfme/tests/control/test_actions.py
+++ b/cfme/tests/control/test_actions.py
@@ -426,7 +426,7 @@ def test_action_prevent_event(request, vm, vm_off, policy_for_testing):
         pytest.fail("CFME did not prevent starting of the VM {}".format(vm.name))
 
 
-@pytest.mark.meta(blockers=[1439331])
+@pytest.mark.meta(blockers=[BZ(1439331)])
 @pytest.mark.provider(
     [VMwareProvider, RHEVMProvider, OpenStackProvider, AzureProvider],
     scope="module"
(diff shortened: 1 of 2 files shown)


lina-nikiforova:
@izapolsk should these blockers work only for `5.10`?
Also, this might be related https://bugzilla.redhat.com/show_bug.cgi?id=1622952

izapolsk:
@lina-nikiforova, as I see in PRT. this issue is relevant to 5.10 only.
as for that bug, Harpreet left reference to it in the bug from PR. so, I'm aware of it. thanks.

cfme/tests/control/test_actions.py
```diff
@@ -426,7 +426,7 @@ def test_action_prevent_event(request, vm, vm_off, policy_for_testing):
         pytest.fail("CFME did not prevent starting of the VM {}".format(vm.name))
 
 
-@pytest.mark.meta(blockers=[1439331])
+@pytest.mark.meta(blockers=[BZ(1439331)])
```
lina-nikiforova: Why do you keep his BZ? It's `CLOSED CURRENTRELEASE`

izapolsk: I just fixed bug. I didn't even check this. I think this is test owner responsibility.

lina-nikiforova:
@izapolsk `test_action_cancel_clone` is failing for `5.9`. Is there the same bug and it should be blocked or is it another issue?

izapolsk:
@lina-nikiforova, this 5.9 failure isn't related to that bug and caused by something else.

digitronik:
As this PR under `blocker-only` label LGTM :+1:

PR #8537, opened 2019-03-05
[RFR] fixed nodata tests and visual configuration view

izapolsk:
visual tests have been relied on data provided by NonJSEntities which was replaced with JSEntities long time ago. 
In addition visual configuration view widgets have had wrong ids.

### DIFF
--- a/cfme/tests/configure/test_visual_infra.py
+++ b/cfme/tests/configure/test_visual_infra.py
@@ -304,8 +304,7 @@ def test_infraprovider_noquads(request, set_infra_provider_quad):
     """
     view = navigate_to(InfraProvider, 'All')
     view.toolbar.view_selector.select('Grid View')
-    # Here data property will return an empty dict when the Quadrants option is deactivated.
-    assert not view.entities.get_first_entity().data
+    assert 'topRight' not in view.entities.get_first_entity().data.get('quad', {})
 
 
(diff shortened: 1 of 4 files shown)


izapolsk:
@valaparthvi, okay, ty

izapolsk:
added VersionPickers for 5.9/5.10 and  fixed one more similar test.

cfme/tests/configure/test_visual_infra.py
```diff
@@ -360,8 +357,7 @@ def test_vm_noquads(appliance, request, set_vm_quad):
     """
     view = navigate_to(appliance.collections.infra_vms, 'VMsOnly')
     view.toolbar.view_selector.select('Grid View')
-    # Here data property will return an empty dict when the Quadrants option is deactivated.
-    assert not view.entities.get_first_entity().data
+    assert 'topRight' not in view.entities.get_first_entity().data.get('quad', {})
```
valaparthvi: view.entities.get_first_entity().data.get('quad', {}) will anyway return empty dict, using quadicon instead of quad returns dictionary with some data.

izapolsk: where do you use quadicon ? I don't see it in data returned by js.

izapolsk:
failed paginator tests  are fixed in PR #8547

PR #8545, opened 2019-03-07
[1LP][RFR] Fix assert view.is_displayed for catalog

niyazRedhat:
Purpose or Intent
=================
{{pytest: cfme/tests/services/test_catalog.py -k  test_catalog_crud -v}}

### DIFF
--- a/cfme/services/catalogs/catalog.py
+++ b/cfme/services/catalogs/catalog.py
@@ -150,8 +150,7 @@ class CatalogCollection(BaseCollection):
         })
         view.add_button.click()
         catalog = self.instantiate(name=name, description=description, items=items)
-        view = self.create_view(CatalogsView)
-        assert view.is_displayed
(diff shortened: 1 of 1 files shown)


cfme/services/catalogs/catalog.py
```diff
@@ -151,7 +151,7 @@ def create(self, name, description, items=None):
         view.add_button.click()
         catalog = self.instantiate(name=name, description=description, items=items)
         view = self.create_view(CatalogsView)
-        assert view.is_displayed
```
izapolsk: you don't need assert since wait_displayed will fail otherwise.

digitronik: @niyazRedhat even I would not use `view.wait_displayed()` as in last step only you are creating `view`.
best way:
```
view = self.create_view(CatalogsView, wait="10s")
```

PR #8547, opened 2019-03-07
[1LP][RFR] fixed paginator is_displayed and a bunch of tests failing due to absent paginator

izapolsk:


cfme/tests/configure/test_visual_cloud.py
```diff
@@ -105,6 +105,8 @@ def test_cloud_grid_page_per_item(appliance, request, page, value, set_grid):
     limit = appliance.user.my_settings.visual.grid_view_limit
     view = navigate_to(page, 'All', use_resetter=False)
     view.toolbar.view_selector.select('Grid View')
+    if not view.entities.paginator.is_displayed:
+        pytest.skip("This page doesn't have entities and/or paginator")
```
digitronik: @izapolsk 
1. I am suspect about skipping test as its title pointing `page_per_item`
2. There may be changes of BZ else `uncollect` will be best option.

izapolsk: @digitronik, sorry, I didn't get what you mean. What exactly you don't like ?

tests use paginator. if paginator is absent on some page, test fails. that's incorrect.
so, now tests are skipped if there is no paginator on page.

digitronik: It was my bad... after checking failure `screenshot` i understand that  paginator not available with `0` entities.

mshriver:
PRT has some tests failing due to assertion failures, but not due to any changes here. Pagination checks and is_displayed update look good, thanks @izapolsk

PR #8548, opened 2019-03-07
[1LP][RFR]Automating test_tenant_unique_catalog

mnadeem92:

Automating manual test test_tenant_unique_catalog
{{pytest: cfme/tests/configure/test_access_control.py::test_tenant_unique_catalog -v}}

### DIFF
--- a/cfme/tests/configure/test_access_control.py
+++ b/cfme/tests/configure/test_access_control.py
@@ -82,6 +82,18 @@ def tag_value(appliance, category, tag, request):
     return tag_for_create, tag_for_update
 
 
+@pytest.fixture(scope='module')
+def catalog_obj(appliance):
+    catalog_name = fauxfactory.gen_alphanumeric()
+    catalog_desc = "My Catalog"
+
+    cat = appliance.collections.catalogs.create(name=catalog_name, description=catalog_desc)
+    yield cat
+
+    if cat.exists:
+        cat.delete()
+
+
 # User test cases
 @pytest.mark.sauce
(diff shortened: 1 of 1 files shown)


cfme/tests/configure/test_access_control.py
```diff
@@ -1446,7 +1445,24 @@ def test_tenant_unique_catalog():
...
+    view.fill({
+        'name': cat1.name,
+        'description': cat1.description
+    })
+    view.add_button.click()
+    view.flash.wait_displayed(timeout=20)
+    assert view.flash.read() == [msg]
+
+    @request.addfinalizer
+    def _delete_catalog():
```
digitronik: Optional:
@mnadeem92 best practice while adding `finalizer`; add it after object creation. It will give readability in bigger test.

cfme/tests/configure/test_access_control.py
```diff
@@ -1446,7 +1445,24 @@ def test_tenant_unique_catalog():
         initialEstimate: 1/2h
         startsin: 5.5
     """
-    pass
+    catalog_name = fauxfactory.gen_alphanumeric()
+    catalog_desc = "My Catalog"
+    msg = "Name has already been taken"
+
+    cat1 = appliance.collections.catalogs.create(name=catalog_name, description=catalog_desc)
```
digitronik: Optional:
As one catalog is setup for this test. I would personally move it as fixture. So in future if someone want to use catalog then he/she can directly use.

PR #8552, opened 2019-03-07
[NOTEST][RFR] Common exists method in BaseEntity

valaparthvi:
1. Write a common `exists` method in `BaseEntity` that does simple navigation to `Details` page and returns `True` if there are no exceptions. (as suggested by @mshriver )
2. Delete all `exists` method that used this way of checking existence. Other methods that use a different way have been kept as is.

### DIFF
--- a/cfme/modeling/base.py
+++ b/cfme/modeling/base.py
@@ -7,7 +7,6 @@ from widgetastic.exceptions import NoSuchElementException
 from widgetastic.utils import VersionPick
 from widgetastic_patternfly import CandidateNotFound
 
-
 from cfme.exceptions import BackupNotFoundError
 from cfme.exceptions import FlavorNotFound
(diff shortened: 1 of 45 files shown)


john-dupuy:
Imports need to be reordered to address Travis failures, but otherwise LGTM -- good improvement!

cfme/modeling/base.py
```diff
@@ -160,6 +176,31 @@ def collections(self):
 
         return EntityCollections.for_entity(self, spec)
 
+    @property
+    def exists(self):
+        try:
+            navigate_to(self, "Details")
+        except (
+            BackupNotFoundError,
```
mshriver: I'm down to run with this for now, but I'm thinking we might have to provide a common `NotFoundException` class for collecting these in a more useful manner.

PR #8553, opened 2019-03-07
[1LP][RFR][NOTEST] Reassign control tests and move manual cases

john-dupuy:
Bulk reassign of control test cases to myself. Also moving some leftover manual cases.


cfme/tests/test_manual.py
```diff
@@ -1487,23 +1487,6 @@ def test_orchestration_link_mismatch():
     pass
 
 
-@pytest.mark.manual
```
mshriver: :tada:

PR #8554, opened 2019-03-07
[RFR] removing fixtured collections from test_compliance

john-dupuy:
also adding some polarion metadata

{{ pytest: --long-running --use-template-cache --color=yes --use-provider vsphere6-nested cfme/tests/control/test_compliance.py }}


john-dupuy:
1 test is failing which is unrelated to this change.

PR #8559, opened 2019-03-08
[RFR]Fix updates

jaryn:
Fix updates tests failing on auth error. Use the sprout_user_key from the cmdline.

{{py.test: --sprout-user-key sprout cfme/tests/cli/test_appliance_update.py::test_update_yum}}

### DIFF
--- a/cfme/tests/cli/test_appliance_update.py
+++ b/cfme/tests/cli/test_appliance_update.py
@@ -8,8 +8,8 @@ from cfme.fixtures.cli import provider_app_crud
 from cfme.fixtures.cli import provision_vm
 from cfme.fixtures.cli import update_appliance
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
+from cfme.test_framework.sprout.client import AuthException
 from cfme.test_framework.sprout.client import SproutClient
-from cfme.test_framework.sprout.client import SproutException
 from cfme.utils.appliance import find_appliance
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.conf import cfme_data
@@ -71,8 +71,16 @@ def appliance_preupdate(old_version, appliance):
         if a.startswith(old_version):
             usable.append(Version(a))
     usable.sort(reverse=True)
-    apps, pool_id = sp.provision_appliances(count=1, preconfigured=True,
-                                            lease_time=180, version=str(usable[0]))
+    try:
+        apps, pool_id = sp.provision_appliances(count=1,
+                                                preconfigured=True,
+                                                lease_time=180,
+                                                version=str(usable[0]))
+    except AuthException:
+        msg = ('Sprout credentials key or yaml maps missing or invalid,'
+               'unable to provision appliance version %s for preupdate'.format(str(usable[0])))
+        logger.exception(msg)
+        pytest.skip(msg)
 
     apps[0].db.extend_partition()
     urls = cfme_data["basic_info"][update_url]
--- a/cfme/tests/cli/test_appliance_update.py
+++ b/cfme/tests/cli/test_appliance_update.py
@@ -65,7 +65,7 @@ def appliance_preupdate(old_version, appliance):
     repo file for update"""
 
     usable = []
-    sp = SproutClient.from_config()
+    sp = SproutClient.from_config(sprout_user_key=pytest.config.option.sprout_user_key or None)
     available_versions = set(sp.call_method('available_cfme_versions'))
     for a in available_versions:
         if a.startswith(old_version):
--- a/cfme/tests/cli/test_appliance_update.py
+++ b/cfme/tests/cli/test_appliance_update.py
@@ -71,12 +71,8 @@ def appliance_preupdate(old_version, appliance):
         if a.startswith(old_version):
             usable.append(Version(a))
     usable.sort(reverse=True)
-    try:
-        apps, pool_id = sp.provision_appliances(count=1, preconfigured=True,
-                                                lease_time=180, version=str(usable[0]))
-    except Exception as e:
-        logger.exception("Couldn't provision appliance with following error:{}".format(e))
-        raise SproutException('No provision available')
+    apps, pool_id = sp.provision_appliances(count=1, preconfigured=True,
+                                            lease_time=180, version=str(usable[0]))
 
     apps[0].db.extend_partition()
     urls = cfme_data["basic_info"][update_url]


cfme/tests/cli/test_appliance_update.py
```diff
@@ -65,18 +65,14 @@ def appliance_preupdate(old_version, appliance):
     repo file for update"""
 
     usable = []
-    sp = SproutClient.from_config()
+    sp = SproutClient.from_config(sprout_user_key=pytest.config.option.sprout_user_key or None)
```
mshriver: I think it would be better to actually check this first. If no sprout_user_key is passed, then this ends up failing because it requires auth.

You can do something like this to make it behave in a more predictable way:
```
sprout_key = pytest.config.option.get('sprout_user_key')
if sprout_key is None:
    pytest.skip('Missing sprout auth key, can not run setup')
SproutClient.from_config(sprout_key)
```

jaryn: Not really. There is additional logic in the from_config (reading the env vars)

jaryn: I also don't understand why SKIP is better then failing the test with ERROR. I have better info about the failure from the ERROR -- the traceback.

mshriver: Sure, not a great pattern overall then for using SproutClient.from_config.

cfme/tests/cli/test_appliance_update.py
```diff
@@ -65,18 +65,14 @@ def appliance_preupdate(old_version, appliance):
...
         if a.startswith(old_version):
             usable.append(Version(a))
     usable.sort(reverse=True)
-    try:
-        apps, pool_id = sp.provision_appliances(count=1, preconfigured=True,
-                                                lease_time=180, version=str(usable[0]))
-    except Exception as e:
-        logger.exception("Couldn't provision appliance with following error:{}".format(e))
-        raise SproutException('No provision available')
+    apps, pool_id = sp.provision_appliances(count=1, preconfigured=True,
```
mshriver: Why remove the exception handling here? I think this could be a spot to cleanly skip the test when it can't provision via sprout. The exception can still happen when the option isn't passed, now we'll raise some exception other than SproutException.

jaryn: The exception AuthException was that got logged, but I had to dig quite a bit to find the correct log. From the trackerbot you cannot just click and read the log with the traceback of the AuthException, only the traceback:of the SproutException, which is not too useful.
https://cfmeqe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/downstream-59z-tests-master/638/artifact/integration_tests/log/slave12-slave12-cfme.log
That's why I am removing the rise of SproutException and just letting propagate the original error.

mshriver: Gotcha, understood about the exception levels and logging. I still think it would be better to capture this AuthException, still use logger.exception() to capture the full exception in logs, and then call pytest.skip. Can let other exception types just go through and fail the test with Error - but in this known failure mode, we can opt to skip the test instead, because we know we just don't have the correct information for running the test.

jaryn: As I wrote above, I prefer to get ERROR to get SKIP. Skip has no traceback associated. Traceback is valuable to me because it gives more information about what went wrong. I o often can see much clearly what went wrong from the traceback. Moreover the SKIP message goes just to the log, which is not immediately visible from trackerbot by simple clicking here and there. 

Skip is good to call when user means to skip the test intentionaly, but not when user fails to set something when intending to run the tests.

mshriver: @izapolsk @jawatts @digitronik I'd like to get some more opinions here on how we're handling skips.

Jaroslav, understood about the content of the artifactor report when it comes to skip vs error, but I don't want to let fixtures blow up in error just so that something is in the artifactor report. I don't think its that onerous to look at the cfme.log output to determine why a test was skipped, or see what exception was logged.

We can update artifactor to include/display more when there is a test skip, so that there is less process to see these.  You may not be selecting the appropriate options for artifactor, because with 'Skipped' checked in an artifactor report, I see a short traceback for teach test function that was skipped, and a log snipped that includes the skip message as well.

Artifactor use / enhancements aside, we regularly use `pytest.skip` when, during setup, an expected resource is not present or is unavailable. To me, any test where a predictable exception is happening, should be skipped or failed, with a `pytest.skip` or `pytest.fail` call directly, instead of letting the exception raise. Especially so in setup/teardown. 

Assertion errors would be the exception to this, as that's a 'natural' test failure mode. Things like SproutException, AuthException, AttributeError, KeyError, IndexError, are poor failure modes for tests that could have been a pytest.skip or pytest.fail.

jaryn: Checking the log is tedious: I think it is in the case when miq-runtest was used when the log is split into several files and one needs to first find which log should look (which slave it executed on), then go to Jenkins artifacts page, then go to correct slave log file and then search there. Instead of having displayed the traceback.

Take a look on http://infra-trackerbot2.cfme2.lab.eng.rdu2.redhat.com/ostriz/dini/?stream_report=downstream-59z&source=jenkins#cfme/tests/cli/test_appliance_update.py/test_update_scap_webui[5.9.4] You cannot find what you can see here:
https://cfmeqe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/downstream-59z-tests-master/638/artifact/integration_tests/log/slave12-slave12-cfme.log


> Artifactor use / enhancements aside, we regularly use pytest.skip when, during setup, an expected resource is not present or is unavailable. To me, any test where a predictable exception is happening, should be skipped or failed,

Skipped ... ok I get that I think there is no exclude option to py.test and that's why you are using skips for similar effect. When one needs to run the majority of tests but not a some of that needs some special configuration than SKIP may be OKish way to do that. But we should not let the test FAIL in case we couldn't setup some resource. FAIL != ERROR. Test should fail only when it clearly detects a bug in the product.

mshriver: Agreed on your last paragraph.  We overall count skips as failures in our pass %. Theoretically taking things to an extreme, suppose we intend to run all the tests but we've mis-configured many things and cfme_data.yaml just has one barely configured provider (some missing resources/templates).  We end up with many tests collected for the appliance version and parametrized for that single provider. Most of the tests are skipped in execution though - because we're missing things that are required on the one provider.  During collection we look only at the yaml, and then during execution we're actually trying to access things on the provider that maybe are disabled or not working. The tests get skipped during setup (maybe during run) because they're missing a prerequisite - in this case that prerequisite is a one-off appliance from sprout.

We want to collect the test as generically as possible - and then cleanly handle when more specific expectations are not met.  This could be taken to an extreme and all fixtures become a large `try/except Exception`, but I also think such 'diaper' exception handling is a poor pattern to follow as well.  My thought is that handling specific failure modes we can reasonably expect (length of available hosts is 2, sprout not available, test ldap server missing expected user) during setup by skipping is a good middle ground.  The failure modes that are skipped are explicit, not generic, so we can provide context to the error and skip the test.  If some unexpected error happens, it indicates something _really_ going wrong during execution, or some bad assumptions we haven't expected yet.

Thanks for the links, perfect, I see what you mean about the content of the traceback and log for that case not including the AuthException message. I'll have to look into that deeper, but that's a separate issue, and if you'd like to discuss enhancements to artifactor, that's awesome, please open an issue and we can continue discussion about how to handle/display that in the issue.

mshriver: I'm pushing a commit to wrap the AuthException in a pytest.skip call.

I do want this discussion to continue, and if you don't like the formatting/style/pattern that I push, please don't hesitate to open a rc-regression PR on Monday updating it.

Want to merge this for 17.46.0-rc though.

dajoRH:
Lint report for commit f1baa4a37e92d27933b4ba06f4ef2b2d7bc32ae2:

`cfme/tests/cli/test_appliance_update.py`:
- :bangbang: Line 79:9: **E124** *closing bracket does not match visual indentation*

**No commit flaws detected**.

*CFME QE Bot*

PR #8561, opened 2019-03-11
[1LP][RFR] New Test: Testing imported domain refresh with non-super user

ganeshhubale:
- Testing imported domain refresh with non-super user
- Added new view for refreshing imported domains
    - DomainRefreshView
- Added new property for domain
    - Refresh
- Removed not used imports
- Removed ```<5.9``` dependencies

{{pytest: cfme/tests/automate/test_git_import.py -k 'test_refresh_git_current_user' -vv }}

### DIFF
--- a/cfme/automate/explorer/domain.py
+++ b/cfme/automate/explorer/domain.py
@@ -278,10 +278,6 @@ class Domain(BaseEntity, Fillable):
             view.flash.assert_message(
                 'Edit of Automate Domain "{}" was cancelled by the user'.format(self.name))
 
-    def delete_if_exists(self):
-        if self.exists:
-            self.delete()
-
     def refresh(self, branch_or_tag=None, git_branch=None, cancel=False):
         view = navigate_to(self, 'Refresh')
 
--- a/cfme/automate/explorer/domain.py
+++ b/cfme/automate/explorer/domain.py
@@ -282,15 +282,19 @@ class Domain(BaseEntity, Fillable):
         view = navigate_to(self, 'Refresh')
 
         # It refreshes the domain for default values of 'branch_or_tag' and 'git_branch'; if these
-        # values are not provided while calling 'refresh'.
-        select_branch_or_tag = branch_or_tag or view.branch_or_tag.selected_option
-        select_git_branch = git_branch or view.git_branches.selected_option
+        # values are not provided while calling 'refresh' then imported domain should be refreshed
+        # with default values of 'branch_or_tag' and 'git_branches'.
 
-        view.branch_or_tag.select_by_visible_text(select_branch_or_tag)
-        view.git_branches.select_by_visible_text(select_git_branch)
-        view.save_button.click()
+        changed = view.fill(
+            {
+                'branch_or_tag': branch_or_tag or view.branch_or_tag.selected_option,
+                'git_branches': git_branch or view.git_branches.selected_option
+            }
+        )
 
-        if cancel:
+        if changed and not cancel:
+            view.save_button.click()
+        else:
             view.cancel_button.click()
         view.flash.assert_no_error()
 
--- a/cfme/automate/explorer/domain.py
+++ b/cfme/automate/explorer/domain.py
@@ -7,6 +7,7 @@ from widgetastic.utils import Fillable
 from widgetastic.widget import Checkbox
 from widgetastic.widget import Text
 from widgetastic.xpath import quote
+from widgetastic_patternfly import BootstrapSelect
 from widgetastic_patternfly import Button
 from widgetastic_patternfly import CandidateNotFound
 from widgetastic_patternfly import Input
@@ -26,6 +27,20 @@ from widgetastic_manageiq import UpDownSelect
 def generate_updown(title):
     return './/*[(self::a or self::button) and @title={}]/*[self::img or self::i]'.format(
         quote(title))
+
+
+class DomainRefreshView(AutomateExplorerView):
+    title = Text("#explorer_title_text")
+    branch_or_tag = BootstrapSelect(id="branch_or_tag_select")
+    git_branches = BootstrapSelect(id="git_branches")
+
+    save_button = Button("Save")
+    cancel_button = Button("Cancel")
+
+    @property
+    def is_displayed(self):
+        return (self.in_explorer
+                and self.title.text == "Refreshing branch/tag for Git-based Domain")
 
 
 class DomainPriorityView(AutomateExplorerView):
@@ -263,6 +278,26 @@ class Domain(BaseEntity, Fillable):
             view.flash.assert_message(
                 'Edit of Automate Domain "{}" was cancelled by the user'.format(self.name))
 
+    def delete_if_exists(self):
+        if self.exists:
+            self.delete()
+
+    def refresh(self, branch_or_tag=None, git_branch=None, cancel=False):
+        view = navigate_to(self, 'Refresh')
+
+        # It refreshes the domain for default values of 'branch_or_tag' and 'git_branch'; if these
+        # values are not provided while calling 'refresh'.
+        select_branch_or_tag = branch_or_tag or view.branch_or_tag.selected_option
+        select_git_branch = git_branch or view.git_branches.selected_option
+
+        view.branch_or_tag.select_by_visible_text(select_branch_or_tag)
+        view.git_branches.select_by_visible_text(select_git_branch)
+        view.save_button.click()
+
+        if cancel:
+            view.cancel_button.click()
+        view.flash.assert_no_error()
+
 
 @attr.s
 class DomainCollection(BaseCollection):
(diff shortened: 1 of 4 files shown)


cfme/tests/automate/test_git_import.py
```diff
@@ -78,3 +91,70 @@ def test_automate_git_domain_displayed_in_service(appliance, imported_domain, br
...
+    view = navigate_to(tasks_collection, 'AllTasks')
+
+    # Collecting list of all tasks performed by users
+    all_tasks = view.tabs.alltasks.table.read()
+    for tasks in all_tasks:
+        if tasks['Task Name'] == 'Refresh git repository':
+            assert tasks['User'] == 'admin'
+            break
+    with new_user:
+
```
jaryn: the space should not be here but maybe above the with.

cfme/tests/automate/test_git_import.py
```diff
@@ -78,3 +91,70 @@ def test_automate_git_domain_displayed_in_service(appliance, imported_domain, br
...
+    # Refreshed imported domain by 'Admin' user
+    imported_domain.refresh(branch_or_tag='Branch', git_branch='origin/master')
+    view = imported_domain.create_view(DomainDetailsView)
+    view.flash.assert_message("Successfully refreshed!")
+    view = navigate_to(tasks_collection, 'AllTasks')
+
+    # Collecting list of all tasks performed by users
+    all_tasks = view.tabs.alltasks.table.read()
+    for tasks in all_tasks:
```
jaryn: shouldn't there be `task` instead of `tasks`?

cfme/tests/automate/test_git_import.py
```diff
@@ -13,25 +16,35 @@
...
 
 
+@pytest.fixture(scope="module")
+def new_user(appliance):
+    """This fixture creates new user which assigned with non-super group"""
+    group = appliance.collections.groups.instantiate(description='EvmGroup-administrator')
+    user = appliance.collections.users.create(
+        name="user_{}".format(fauxfactory.gen_alphanumeric().lower()),
+        credential=Credential(principal='uid{}'.format(fauxfactory.gen_alphanumeric(4)),
+                              secret='{password}'.format(password=fauxfactory.gen_alphanumeric(4))),
```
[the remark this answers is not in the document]
ganeshhubale: Agree! 
But according to me, it is more readable. So I am keeping it as it is.

dajoRH:
*CFME QE Bot*

cfme/automate/explorer/domain.py
```diff
@@ -264,6 +278,26 @@ def update(self, updates):
...
+
+    def refresh(self, branch_or_tag=None, git_branch=None, cancel=False):
+        view = navigate_to(self, 'Refresh')
+
+        # It refreshes the domain for default values of 'branch_or_tag' and 'git_branch'; if these
+        # values are not provided while calling 'refresh'.
+        select_branch_or_tag = branch_or_tag or view.branch_or_tag.selected_option
+        select_git_branch = git_branch or view.git_branches.selected_option
+
+        view.branch_or_tag.select_by_visible_text(select_branch_or_tag)
```
digitronik: I would avoid use of `or`
```
if branch_or_tag:
    view.branch_or_tag.select_by_visible_text(branch_or_tag)
```

cfme/tests/automate/test_git_import.py
```diff
@@ -78,3 +91,70 @@ def test_automate_git_domain_displayed_in_service(appliance, imported_domain, br
...
+
+    # Refreshed imported domain by 'Admin' user
+    imported_domain.refresh(branch_or_tag='Branch', git_branch='origin/master')
+    view = imported_domain.create_view(DomainDetailsView)
+    view.flash.assert_message("Successfully refreshed!")
+    view = navigate_to(tasks_collection, 'AllTasks')
+
+    # Collecting list of all tasks performed by users
+    all_tasks = view.tabs.alltasks.table.read()
+    for task in all_tasks:
```
digitronik: Q:
what if expected `task` not found in `all_tasks`. test will still pass. do you not think we need proper assertion?

cfme/automate/explorer/domain.py
```diff
@@ -264,6 +278,26 @@ def update(self, updates):
             view.flash.assert_message(
                 'Edit of Automate Domain "{}" was cancelled by the user'.format(self.name))
 
+    def delete_if_exists(self):
```
digitronik: @ganeshhubale now this method cover under `BaseEntity`. I think no need of it. please check it one time.

dajoRH:
*CFME QE Bot*

cfme/tests/automate/test_git_import.py
```diff
@@ -110,3 +122,80 @@ def test_automate_git_import_multiple_domains(request, appliance):
...
+
+    with new_user:
+        # Refreshed imported domain by non-super user
+        imported_domain.refresh(branch_or_tag='Branch', git_branch='origin/master')
+        view = navigate_to(tasks_collection, 'AllTasks')
+
+        # Collecting list of all tasks performed by users
+        all_tasks = view.tabs.alltasks.table.read()
+        expected_task = None
+        for task in all_tasks:
```
jaryn: This and the similar case above can be written bit better using for/else
https://book.pythontips.com/en/latest/for_-_else.html#else-clause
but this is optional

cfme/automate/explorer/domain.py
```diff
@@ -264,6 +278,22 @@ def update(self, updates):
             view.flash.assert_message(
                 'Edit of Automate Domain "{}" was cancelled by the user'.format(self.name))
 
+    def refresh(self, branch_or_tag=None, git_branch=None, cancel=False):
+        view = navigate_to(self, 'Refresh')
+
+        # It refreshes the domain for default values of 'branch_or_tag' and 'git_branch'; if these
+        # values are not provided while calling 'refresh'.
+        select_branch_or_tag = branch_or_tag or view.branch_or_tag.selected_option
```
mshriver: I don't think this is necessary logic. The behavior that you are trying to replicate here is the default behavior of the `View.fill` method.

You can simply fill the DomainRefreshView object that you have from navigating to `Refresh` destination.

```
changed = view.fill(
    {
        'branch_or_tag': branch_or_tag,
        'git_branches': git_branch
    }
)

if changed and not cancel:
    view.save_button.click()
else:
    view.cancel_button.click()
```

`View.fill` will skip taking any action on the widgets when the value to fill is `None`. Since the function takes default values of `None`, when the user doesn't pass them, the currently selected option (default) on those widgets will be left alone.

cfme/configure/access_control/__init__.py
```diff
@@ -335,27 +333,20 @@ def create(self, name=None, credential=None, email=None, groups=None, cost_cente
                 not having appropriate permissions OR update is not allowed
                 for currently selected role
         """
-        if self.appliance.version < "5.8":
-            user_blocked_msg = "Userid has already been taken"
-        else:
-            user_blocked_msg = ("Userid is not unique within region {}".format(
-                self.appliance.server.zone.region.number))
+        user_blocked_msg = ("Userid is not unique within region {}".format(
```
mshriver: :+1: thanks!

cfme/tests/automate/test_git_import.py
mshriver: Since I'm requesting some other changes, consider changing this implementation @ganeshhubale. It will clean this up and make it more readable.  No need to initialize the `expected_task` or explicitly check it.

cfme/automate/explorer/domain.py
ganeshhubale: @mshriver 
- Imported Domain refresh must work without providing ```branch_or_tag``` and  ```git_branches``` because after importing this domain. 
- Refresh view dynamically fetches the branches for that domain.
- Hence updating as below:
```
    def refresh(self, branch_or_tag=None, git_branch=None, cancel=False):
        view = navigate_to(self, 'Refresh')

        # It refreshes the domain for default values of 'branch_or_tag' and 'git_branch'; if these
        # values are not provided while calling 'refresh' then imported domain should be refreshed
        # with default values of 'branch_or_tag' and 'git_branches'.

        changed = view.fill(
            {
                'branch_or_tag': branch_or_tag or view.branch_or_tag.selected_option,
                'git_branches': git_branch or view.git_branches.selected_option
            }
        )

        if changed and not cancel:
            view.save_button.click()
        else:
            view.cancel_button.click()
        view.flash.assert_no_error()
```

cfme/tests/automate/test_git_import.py
ganeshhubale: @mshriver @jaryn 
- I have updated with ```for/else```. 
- Please take a look:
```
 all_tasks = view.tabs.alltasks.table.read()
    for task in all_tasks:
        if task['Task Name'] == 'Refresh git repository':
            assert task['User'] == 'admin'
            break
    else:
        raise NameError("Task not found")

```

dajoRH:
*CFME QE Bot*

dajoRH:
*CFME QE Bot*

dajoRH:
*CFME QE Bot*

PR #8575, opened 2019-03-13
[1LP][RFR] Add in policy simulation test case

john-dupuy:
Adding test coverage for Cloud/Infra policy simulation page for vms/instances. There are quite a few bugs associated with this page, so this test is designed to cover those [BZs](https://bugzilla.redhat.com/buglist.cgi?bug_status=NEW&bug_status=VERIFIED&bug_status=ASSIGNED&bug_status=MODIFIED&bug_status=ON_DEV&bug_status=ON_QA&bug_status=RELEASE_PENDING&bug_status=POST&email1=jdupuy%40redhat.com&emailassigned_to1=1&emailcc1=1&emailreporter1=1&emailtype1=exact&list_id=10025319). 

{{ pytest: --use-provider complete cfme/tests/cloud_infra_common/test_policy_simulation.py }}

### DIFF
--- a/cfme/cloud/instance/__init__.py
+++ b/cfme/cloud/instance/__init__.py
@@ -456,8 +456,8 @@ class InstanceCollection(VMCollection):
         # Pretty much same as image, but defining at VMCollection would only work for cloud
         # provider filter means we're viewing instances through provider details relationships
         provider = self.filters.get('provider')  # None if no filter, need for entity instantiation
-        view = navigate_to(self,
-                           'AllForProvider' if provider else 'All')
+        view = navigate_to(provider or self,
+                          'Instances' if provider else 'All')
         # iterate pages here instead of use surf_pages=True because data is needed
         entities = []
         for _ in view.entities.paginator.pages():  # auto-resets to first page
@@ -468,6 +468,15 @@ class InstanceCollection(VMCollection):
                  for e in page_entities
                  if e.data.get('provider') != '']  # safe provider check, archived shows no provider
             )
+
+        # filtering
+        if self.filters.get("names"):
+            names = self.filters["names"]
+            entities = [e for e in entities if e.name in names]
+        if self.filters.get("name"):
+            name = self.filters["name"]
+            entities = [e for e in entities if e.name == name]
+
         return entities
 
 
@@ -585,13 +594,31 @@ class Provision(CFMENavigateStep):
         self.prerequisite_view.toolbar.lifecycle.item_select('Provision Instances')
 
 
-@navigator.register(Instance, 'PolicySimulation')
+@navigator.register(Instance, "PolicySimulation")
 class PolicySimulation(CFMENavigateStep):
     VIEW = PolicySimulationView
-    prerequisite = NavigateToSibling('Details')
+    prerequisite = NavigateToSibling("Details")
 
     def step(self, *args, **kwargs):
-        self.prerequisite_view.toolbar.policy.item_select('Policy Simulation')
+        self.prerequisite_view.toolbar.policy.item_select("Policy Simulation")
+
+
+@navigator.register(InstanceCollection, "PolicySimulation")  # noqa
+class PolicySimulationOnCollection(CFMENavigateStep):
+    VIEW = PolicySimulationView
+
+    def prerequisite(self):
+        provider = self.obj.filters.get("provider")
+        if provider:
+            return navigate_to(provider, "Instances")
+        else:
+            return navigate_to(self.obj, "All")
+
+    def step(self, *args, **kwargs):
+        # click the checkbox of every object in the filtered collection
+        for entity in self.obj.all():
+            self.prerequisite_view.entities.get_entity(name=entity.name, surf_pages=True).check()
+        self.prerequisite_view.toolbar.policy.item_select("Policy Simulation")
 
 
 @navigator.register(Instance, 'SetOwnership')
--- a/cfme/common/vm_views.py
+++ b/cfme/common/vm_views.py
@@ -28,6 +28,7 @@ from widgetastic_manageiq import Checkbox
 from widgetastic_manageiq import DriftComparison
 from widgetastic_manageiq import ItemsToolBarViewSelector
 from widgetastic_manageiq import JSBaseEntity
+from widgetastic_manageiq import ManageIQTree
 from widgetastic_manageiq import MultiBoxSelect
 from widgetastic_manageiq import PaginationPane
 from widgetastic_manageiq import ParametrizedSummaryTable
@@ -545,15 +546,51 @@ class PolicySimulationView(BaseLoggedInPage):
     """
     Policy Simulation page for vms/instances
     """
+    title = Text("#explorer_title_text")
+
     @View.nested
     class form(View):  # noqa
-        policy = BootstrapSelect('policy_id')
+        policy_profile = BootstrapSelect("profile_id")
         # TODO policies table, ability to remove
         entities = View.nested(BaseNonInteractiveEntitiesView)
-        cancel_button = Button('Cancel')
+        cancel_button = Button("Cancel")
 
-    # TODO match quadicon
-    is_displayed = displayed_not_implemented
+    @property
+    def is_displayed(self):
+        vm_type = (
+            getattr(self.context["object"], "VM_TYPE", None) or
+            getattr(self.context["object"].ENTITY, "VM_TYPE", None)
+        )
+        expected_title = "{} Policy Simulation".format(vm_type)
+        return self.title.text == expected_title and len(self.form.entities.get_all()) > 0
+
+
+class PolicySimulationDetailsView(BaseLoggedInPage):
+    """
+    Policy simulation details page that appears after clicking a quadicon on the
+    PolicySimulationView page.
+    """
+    title = Text("#explorer_title_text")
+    back_button = Button("Back")
+
+    @View.nested
+    class options(View):  # noqa
+        out_of_scope = Checkbox(name="out_of_scope")
+        show_successful = Checkbox(name="passed")
+        show_failed = Checkbox(name="failed")
+
+    @View.nested
+    class details(View):  # noqa
+        tree = ManageIQTree("policy_simulation_treebox")
+
+    @property
+    def is_displayed(self):
+        expected_title = "{} Policy Simulation".format(self.context["object"].VM_TYPE)
+        return (
+            self.title.text == expected_title and
+            self.details.tree.is_displayed and
+            self.details.tree.root_item.text == self.context["object"].name
+        )
 
 
 class RightSizeView(BaseLoggedInPage):
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -35,6 +35,7 @@ from cfme.common.vm_views import CloneVmView
 from cfme.common.vm_views import EditView
 from cfme.common.vm_views import ManagementEngineView
 from cfme.common.vm_views import MigrateVmView
+from cfme.common.vm_views import PolicySimulationView
 from cfme.common.vm_views import ProvisionView
 from cfme.common.vm_views import PublishVmView
 from cfme.common.vm_views import RenameVmView
@@ -1115,6 +1116,14 @@ class InfraVmCollection(VMCollection):
                  for e in page_entities
                  if e.data.get('provider') != '']  # safe provider check, orphaned shows no provider
             )
+        # filtering
+        if self.filters.get("names"):
+            names = self.filters["names"]
+            entities = [e for e in entities if e.name in names]
+        if self.filters.get("name"):
+            name = self.filters["name"]
+            entities = [e for e in entities if e.name == name]
+
         return entities
 
 
@@ -1518,3 +1527,30 @@ class VmUtilization(CFMENavigateStep):
 
     def step(self, *args, **kwargs):
         self.prerequisite_view.toolbar.monitoring.item_select('Utilization')
+
+
+@navigator.register(InfraVm, 'PolicySimulation')
+class PolicySimulation(CFMENavigateStep):
+    VIEW = PolicySimulationView
+    prerequisite = NavigateToSibling('Details')
+
+    def step(self, *args, **kwargs):
+        self.prerequisite_view.toolbar.policy.item_select('Policy Simulation')
+
+
+@navigator.register(InfraVmCollection, "PolicySimulation")  # noqa
+class PolicySimulationOnCollection(CFMENavigateStep):
+    VIEW = PolicySimulationView
+
+    def prerequisite(self):
+        provider = self.obj.filters.get("provider")  # None if no filter
+        if provider:
+            return navigate_to(provider, "ProviderVms")
+        else:
+            return navigate_to(self.obj, "All")
+
+    def step(self, *args, **kwargs):
+        # click the checkbox of every object in the filtered collection
+        for entity in self.obj.all():
+            self.prerequisite_view.entities.get_entity(name=entity.name, surf_pages=True).check()
+        self.prerequisite_view.toolbar.policy.item_select("Policy Simulation")
--- a/cfme/tests/cloud_infra_common/test_policy_simulation.py
+++ b/cfme/tests/cloud_infra_common/test_policy_simulation.py
@@ -1,0 +1,76 @@
+# -*- coding: utf-8 -*-
+"""This module tests events that are invoked by Cloud/Infra VMs."""
+import pytest
+
+from cfme.cloud.provider import CloudProvider
+from cfme.common.vm_views import PolicySimulationDetailsView
+from cfme.common.vm_views import PolicySimulationView
+from cfme.infrastructure.provider import InfraProvider
+from cfme.markers.env_markers.provider import ONE_PER_CATEGORY
+from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.blockers import BZ
+
+pytestmark = [
+    pytest.mark.usefixtures('setup_provider'),
+    pytest.mark.tier(2),
+    pytest.mark.provider(
+        classes=[CloudProvider, InfraProvider],
+        selector=ONE_PER_CATEGORY,
+        scope="module"
+    ),
+]
+
+
+@pytest.mark.parametrize("navigation", ["provider_vms", "all_vms", "vm_summary"])
+@pytest.mark.uncollectif(
+    lambda navigation: navigation == "provider_vms" and (BZ(1686617).blocks or BZ(1686619).blocks),
+    reason="This test is blocked by the following BZs: {}, {}".format(
+        BZ(1686617).bug_id,
+        BZ(1686619).bug_id
+    )
+)
+def test_policy_simulation_ui(appliance, provider, navigation):
+    """
+    Bugzillas:
+        * 1670456, 1686617, 1686619, 1688359, 1550503
+    Polarion:
+        assignee: jdupuy
+        casecomponent: Control
+        caseimportance: medium
+        initialEstimate: 1/6hr
+    """
+    # get a vm
+    collection = provider.appliance.provider_based_collection(provider)
+    vm = collection.all()[0]
+    my_filter = {"name": vm.name}
+
+    # additional filter args for navigation via provider_vms and vm_summary
+    if navigation != "all_vms":
+        my_filter["provider"] = provider
+
+    # create the filtered collection
+    filtered_collection = collection.filter(my_filter)
+
+    # navigate using filtered_collection if not using the vm's summary page
+    if navigation != "vm_summary":
+        view = navigate_to(filtered_collection, "PolicySimulation")
+    # otherwise use the vm
+    else:
+        view = navigate_to(vm, "PolicySimulation", force=True)
+
+    # check the quadicon
+    assert view.form.entities.check_context_against_entities([vm])
+
+    assert view.form.cancel_button.is_displayed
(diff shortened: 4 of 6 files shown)


dajoRH:
Lint report for commit 915bcdc7a3fca83ab05250e2c1c535a097be2b80:

`cfme/common/vm_views.py`:
- :bangbang: Line 591:9: **E122** *continuation line missing indentation or outdented*
- :bangbang: Line 592:9: **E122** *continuation line missing indentation or outdented*
- :bangbang: Line 593:9: **E122** *continuation line missing indentation or outdented*
- :bangbang: Line 594:5: **E122** *continuation line missing indentation or outdented*

`cfme/tests/cloud_infra_common/test_policy_simulation.py`:
- :bangbang: Line 24:1: **E302** *expected 2 blank lines, found 1*

Hmmm


**No commit flaws detected**.

*CFME QE Bot*

john-dupuy:
Failure in 5.10 is unrelated to this PR.

cfme/common/vm_views.py
```diff
@@ -545,15 +546,52 @@ class PolicySimulationView(BaseLoggedInPage):
...
+    class details(View):  # noqa
+        tree = ManageIQTree("policy_simulation_treebox")
+
+    @property
+    def is_displayed(self):
+        expected_title = "{} Policy Simulation".format(self.context["object"].VM_TYPE)
+        return (
+            self.title.text == expected_title and
+            self.details.tree.is_displayed and
+            self.details.tree.root_item.text == self.context["object"].name
```
digitronik: :+1:

cfme/tests/cloud_infra_common/test_policy_simulation.py
```diff
@@ -0,0 +1,62 @@
...
+    Bugzillas:
+        * 1670456, 1686617, 1686619, 1688359, 1550503
+    Polarion:
+        assignee: jdupuy
+        casecomponent: Control
+        caseimportance: medium
+        initialEstimate: 1/6hr
+    """
+    collection = provider.appliance.provider_based_collection(provider)
+    vm = collection.instantiate('cu-24x7', provider)
```
digitronik: Required: 
@john-dupuy, I was also doing this in same way but now I using `all` method.

1. Need to avoid this hard code.
```
vm = collection.all()[0]
```
2. If `vm` not available then we need to create it or `skip` test.
3. It's `setup` part for this test. May be in future you would like to add more test related to `policy` so better to put this as `fixture`.

cfme/tests/cloud_infra_common/test_policy_simulation.py
```diff
@@ -0,0 +1,62 @@
...
+    if navigation == "provider_vms":
+        if BZ(1686617).blocks or BZ(1686619).blocks:
+            pytest.skip("Skipping due to blocker BZs, {}, {}".format(BZ(1686617), BZ(1686619)))
```
digitronik: Required:
You are using `BZ(1686617)` object... I think better to use simple string or `BZ(num).bug_id`

digitronik: More better if you put `uncollectif` with reason

cfme/tests/cloud_infra_common/test_policy_simulation.py
```diff
@@ -0,0 +1,62 @@
...
+        view = navigate_to(vm.provider, "ProviderVms")
+        # Click the checkbox for our VM
+        view.entities.get_entity(name=vm.name, surf_pages=True).check()
+        # Navigate to policy simulation page
+        view.toolbar.policy.item_select("Policy Simulation")
+        view = vm.create_view(PolicySimulationView, wait="10s")
```
digitronik: Optional:
I have one thought if you like please have a glance.
I would write one more navigation against `vm` `AllForProviderSimulationView`.

cfme/common/vm_views.py
```diff
@@ -545,15 +546,52 @@ class PolicySimulationView(BaseLoggedInPage):
...
+        policy_profile = BootstrapSelect("profile_id")
         # TODO policies table, ability to remove
         entities = View.nested(BaseNonInteractiveEntitiesView)
-        cancel_button = Button('Cancel')
+        cancel_button = Button("Cancel")
 
-    # TODO match quadicon
-    is_displayed = displayed_not_implemented
+    @property
+    def quadicon(self):
```
digitronik: Required:
I would not use `quadicon`
1. it's part of `PolicySimulationView` means, I can select multiple `vms` try simulation form `All VM page`. 
2. we have entities so I would use `view.entities` for all operations to increase readability and consistency.

cfme/common/vm_views.py
```diff
@@ -545,15 +546,52 @@ class PolicySimulationView(BaseLoggedInPage):
...
+        return self.form.entities.get_entity(name=self.context["object"].name)
+
+    @property
+    def is_displayed(self):
+        expected_title = "{} Policy Simulation".format(self.context["object"].VM_TYPE)
+        return self.title.text == expected_title and self.quadicon.is_displayed
```
digitronik: Optional:
Here you can check at least one `entity` available or not than particular `VM`. as per your test its common view from `All` and `Summery` vm page. `VM` name you can check in test.

cfme/tests/cloud_infra_common/test_policy_simulation.py
john-dupuy: @digitronik I have a couple of thoughts regarding this. 
1) In this test, I am specifically using the providers that have this VM on them. 
2) It is slower to first get a whole list of all the VMs through navigation and then select the first item from that list. Leaving the hard code is faster. 
3. I agree about putting this into a fixture.

cfme/tests/cloud_infra_common/test_policy_simulation.py
john-dupuy: @digitronik, I feel that this navigation is atypical. Since this test is specifically in reference to a BZ, I think it is good to be explicit about the navigation steps here too.

john-dupuy: I do think that there should be another view that is registered against a vm collection, so that we can select multiple vms for a policy simulation. However, I'd like to implement that in a different PR.

cfme/common/vm_views.py
john-dupuy: @digitronik Since `PolicySimulationView` is registered against an instance/vm, I think it is important to check the quadicon, since it is the only widget on the page that is specific to the instance/vm. 

If we have a policy simulation against a number of items, I think this should be a different view that is registered against the collection rather than the vm/instance and therefore has a different `is_displayed` method.

cfme/tests/cloud_infra_common/test_policy_simulation.py
digitronik: @john-dupuy agreed... please for now put `all` method. It am also facing same problem it took time due to UI iteration... I moved most of `all` method from `UI` to `REST`. I am going to move `vm` `all` method to `rest`. It is in my task. Please wait... for some time.

cfme/common/vm_views.py
digitronik: Myview:
@john-dupuy we register navigation destination against any `BaseEntity` not view.
Entity using `BaseNonInteractiveEntitiesView` means it capable to provide you `n` numbers of `vm/Instance` on that page. with such implementation we are restricting for single `vm/Instance`.

Except entity not changes in view as per me. Let me know why we need other view than this?
You can check entity name in `test` for more validation. so in future, we can parametrize this test in better way. even you are parametrizing this test with `all` and `detail` page.

dajoRH:
Lint report for commit 3ac5adae752289fdd1a319b6eab9a9f15caa6a04:

`cfme/cloud/instance/__init__.py`:
- :red_circle: Line 610:1: **F811** *redefinition of unused 'PolicySimulation' from line 601*
- :bangbang: Line 645:1: **E303** *too many blank lines (3)*

`cfme/common/vm_views.py`:
- :bangbang: Line 562:17: **E126** *continuation line over-indented for hanging indent*

- :red_circle: Line 12:1: **F401** *'navmazing.NavigateToObject' imported but unused*
- :red_circle: Line 1542:1: **F811** *redefinition of unused 'PolicySimulation' from line 1533*

Hmmm


**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
Lint report for commit 33f18b1baa5c0ad1e2cd85f482b4cc2e118f3d95:

`cfme/cloud/instance/__init__.py`:
- :red_circle: Line 606:1: **F811** *redefinition of unused 'PolicySimulation' from line 597*

- :red_circle: Line 1541:1: **F811** *redefinition of unused 'PolicySimulation' from line 1532*

`cfme/tests/cloud_infra_common/test_policy_simulation.py`:
- :bangbang: Line 17:33: **E261** *at least two spaces before inline comment*
- :bangbang: Line 17:34: **E262** *inline comment should start with '# '*

**No commit flaws detected**.

*CFME QE Bot*

cfme/common/vm_views.py
```diff
@@ -545,15 +546,52 @@ class PolicySimulationView(BaseLoggedInPage):
     """
     Policy Simulation page for vms/instances
     """
+
+    title = Text("#explorer_title_text")
```
ganeshhubale: You can remove blank line (549)

cfme/tests/cloud_infra_common/test_policy_simulation.py
```diff
@@ -0,0 +1,79 @@
...
+]
+
+
+@pytest.fixture
+def vm_and_filter(appliance, provider):
+    collection = provider.appliance.provider_based_collection(provider)
+    vm = collection.all()[0]
+    filter = {'name': vm.name, 'provider': provider}
+    yield vm, filter
+
```
ganeshhubale: @john-dupuy 
Here you have already accessed ```appliance``` by ```provider``` object(```provider.appliance```). I think you can remove ```appliance``` from arguments.

cfme/tests/cloud_infra_common/test_policy_simulation.py
```diff
@@ -0,0 +1,79 @@
...
+        classes=[CloudProvider, InfraProvider],
+        selector=ONE_PER_CATEGORY,
+        scope="module"
+    ),
+]
+
+
+@pytest.fixture
+def vm_and_filter(appliance, provider):
+    collection = provider.appliance.provider_based_collection(provider)
```
ganeshhubale: @john-dupuy 
Just a question:
- Do you really need this as a fixture?
Because this fixture does not create or teardown anything. So we can add this to test case itself and fetch this data.

cfme/tests/cloud_infra_common/test_policy_simulation.py
```diff
@@ -0,0 +1,79 @@
...
+    vm = collection.all()[0]
+    filter = {'name': vm.name, 'provider': provider}
+    yield vm, filter
```
ganeshhubale: @john-dupuy  
You can change ```filter``` variable to some other name. Because, it is **python keyword**.

cfme/tests/cloud_infra_common/test_policy_simulation.py
```diff
@@ -0,0 +1,79 @@
...
+        if navigation == "all_vms":
+            filter.pop("provider")
+
+        if vm.provider.one_of(InfraProvider):
+            view = navigate_to(appliance.collections.infra_vms.filter(filter), "PolicySimulation")
+        else:
+            view = navigate_to(
+                appliance.collections.cloud_instances.filter(filter),
+                "PolicySimulation"
+            )
```
ganeshhubale: Same here for ```filter``` variables

cfme/tests/cloud_infra_common/test_policy_simulation.py
```diff
@@ -0,0 +1,79 @@
...
+from cfme.common.vm_views import PolicySimulationDetailsView
+from cfme.common.vm_views import PolicySimulationView
+from cfme.infrastructure.provider import InfraProvider
+from cfme.markers.env_markers.provider import ONE_PER_CATEGORY
+from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.blockers import BZ
+
+pytestmark = [
+    pytest.mark.usefixtures('uses_infra_providers', 'uses_cloud_providers', 'setup_provider'),
+    pytest.mark.tier(2),
```
ganeshhubale: @john-dupuy 
Just a question:
- Do we need ```uses_infra_providers``` and ```uses_cloud_providers``` separately? Because as per my understanding, ```setup_provider``` is applicable for all the providers.

cfme/tests/cloud_infra_common/test_policy_simulation.py
john-dupuy: Good point!

cfme/tests/cloud_infra_common/test_policy_simulation.py
john-dupuy: @ganeshhubale Right now, this doesn't really need to be a fixture. However this is the first test for the policy simulation page, and it is likely that future tests will make use of this fixture.

cfme/tests/cloud_infra_common/test_policy_simulation.py
john-dupuy: I don't think so, I was copying this context from another test. Thanks for pointing this out.

cfme/cloud/instance/__init__.py
```diff
@@ -456,8 +456,8 @@ def all(self):
         # Pretty much same as image, but defining at VMCollection would only work for cloud
         # provider filter means we're viewing instances through provider details relationships
         provider = self.filters.get('provider')  # None if no filter, need for entity instantiation
-        view = navigate_to(self,
-                           'AllForProvider' if provider else 'All')
+        view = navigate_to(provider or self,
+                          'Instances' if provider else 'All')
         # iterate pages here instead of use surf_pages=True because data is needed
         entities = []
         for _ in view.entities.paginator.pages():  # auto-resets to first page
```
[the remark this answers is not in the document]
mshriver: Definitely want to use `pages()` here, as it resets to the first page and iterates through them - without this, you _could_ be starting on any page.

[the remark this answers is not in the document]
valaparthvi: Then I might have to re-fix my test case :roll_eyes: I was facing an issue where `paginator.pages()` was returning nothing since `view.paginator.exists` was returning `False`(check [here](https://github.com/ManageIQ/integration_tests/blob/master/widgetastic_manageiq/__init__.py#L1664)) even though the page was visible in the view.

dajoRH:
Lint report for commit 9d70c58c1d0f1542fe17f352aed23968290480c3:

`cfme/tests/cloud_infra_common/test_policy_simulation.py`:
- :bangbang: Line 56:101: **E501** *line too long (101 > 100 characters)*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/cloud_infra_common/test_policy_simulation.py
```diff
@@ -0,0 +1,82 @@
...
+            )
+    else:
+        view = navigate_to(vm, "PolicySimulation", force=True)
+    # check the quadicon
+    assert view.form.entities.check_context_against_entities([vm])
+
+    assert view.form.cancel_button.is_displayed
+    view.form.policy_profile.select_by_visible_text("OpenSCAP profile")
+    # now check that we can navigate to the right page
+    if vm.provider.one_of(InfraProvider) or not BZ(1688359, forced_streams=["5.9", "5.10"]).blocks:
```
jawatts: Do you need to force streams here? 

```
In[9]: BZ(1688359)
Out[9]: <cfme.utils.blockers.BZ at 0x7fc9ca681fd0>
In[10]: BZ(1688359).blocks
Out[10]: True

```

cfme/tests/cloud_infra_common/test_policy_simulation.py
```diff
@@ -0,0 +1,82 @@
...
+        view.form.entities.get_entity(name=vm.name).click()
+        view = vm.create_view(PolicySimulationDetailsView, wait="10s")
+        # check that the back button works
+        view.back_button.click()
+    # check that the cancel button works
+    if not BZ(1670456, forced_streams=["5.9", "5.10"]).blocks:
```
jawatts: same question here

cfme/tests/cloud_infra_common/test_policy_simulation.py
john-dupuy: Ah, I guess not!

cfme/infrastructure/virtual_machines.py
```diff
@@ -1518,3 +1527,40 @@ def VIEW(self):     # noqa
...
+class PolicySimulation(CFMENavigateStep):
+    VIEW = PolicySimulationView
+    prerequisite = NavigateToSibling('Details')
+
+    def step(self, *args, **kwargs):
+        self.prerequisite_view.toolbar.policy.item_select('Policy Simulation')
+
+
+@navigator.register(InfraVmCollection, "PolicySimulation")  # noqa
+class PolicySimulation(CFMENavigateStep):  # noqa
```
mshriver: The class itself should have a unique name, but the registered destination name can be the same on both the collection and entity.

cfme/infrastructure/virtual_machines.py
```diff
@@ -1518,3 +1527,40 @@ def VIEW(self):     # noqa
...
+    def prerequisite(self):
+        provider = self.obj.filters.get("provider")  # None if no filter
+        if provider:
+            return navigate_to(provider, "ProviderVms")
+        else:
+            return navigate_to(self.obj, "All")
+
+    def step(self, *args, **kwargs):
+        # if filter is not defined we don't want to navigate using every object
+        if not self.navigatable:
```
mshriver: I don't think I like this pattern, at least implemented at this level and not in Navmazing directly.

I'm also not convinced this is needed here: if the collection is not filtered - then I want to run policy simulation on the entire collection - the call to .all() will account for any filters.

cfme/infrastructure/virtual_machines.py
```diff
@@ -1518,3 +1527,40 @@ def VIEW(self):     # noqa
...
+            raise DestinationNotFound(
+                "Navigation to policy simulation page using a collection requires the collection "
+                "to be filtered."
+            )
+        # click the checkbox of every object in the filtered collection
+        for entity in self.obj.all():
+            self.prerequisite_view.entities.get_entity(name=entity.name, surf_pages=True).check()
```
mshriver: How does this behave if it checks an item on the first page, then surfs to the 3rd page and selects a 2nd item?

cfme/tests/cloud_infra_common/test_policy_simulation.py
```diff
@@ -0,0 +1,82 @@
...
+from cfme.markers.env_markers.provider import ONE_PER_CATEGORY
+from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.blockers import BZ
+
+pytestmark = [
+    pytest.mark.usefixtures('setup_provider'),
+    pytest.mark.tier(2),
+    pytest.mark.provider(
+        classes=[CloudProvider, InfraProvider],
+        selector=ONE_PER_CATEGORY,
```
mshriver: I'm thinking we want larger collection pool for this test - that we should test simulation against each provider type.

john-dupuy: @mshriver Since this is a UI test, my thought is that the specific provider is not so important, only that we have one of each type.

cfme/infrastructure/virtual_machines.py
john-dupuy: Agreed, I don't think it's necessary either.

cfme/infrastructure/virtual_machines.py
john-dupuy: Good question, in MIQ if you perform this manually then you're initial check is not preserved. However, in this case, I have seen that the `all()` method of the collection will display 1000 items per page so that everything is displayed. So surfing the page is not relevant here. When the `all()` method is updated to use rest or a DB query, we will need to account for this behavior.

cfme/tests/cloud_infra_common/test_policy_simulation.py
```diff
@@ -0,0 +1,82 @@
...
+    Bugzillas:
+        * 1670456, 1686617, 1686619, 1688359, 1550503
+    Polarion:
+        assignee: jdupuy
+        casecomponent: Control
+        caseimportance: medium
+        initialEstimate: 1/6hr
+    """
+    vm, my_filter = vm_and_filter
+    if navigation != "vm_summary":
```
mshriver: Please add some in-line comments here as to the assumptions being made. Difficult to read without referencing the navigation parametrization.

cfme/tests/cloud_infra_common/test_policy_simulation.py
```diff
@@ -0,0 +1,82 @@
...
+        if navigation == "all_vms":
+            my_filter.pop("provider")
+
+        if vm.provider.one_of(InfraProvider):
+            view = navigate_to(
+                appliance.collections.infra_vms.filter(my_filter),
```
mshriver: I don't see much need for the vm_and_filter fixture - and if you actually move that code to the top of this function, you'll already have the provider-based-collection object that you need to apply the filter to without having this layer of `provider.one_of` checking.  You can also create the filter dictionary based on the navigation parametrization, instead of popping 'provider' off at L53 above.

john-dupuy: I agree, that's a much cleaner way to do this! :+1:

PR #8576, opened 2019-03-14
[RFR] Remove condition to fix deleting expressions

apagac:
__Fixing__ `test_group_crud_with_tag` when deleting existing expressions.

The `wait_for` in question here checks also for `self.expression_text != "<new element>"`. When deleting expressions during testing, we use:

```
while self.any_expression_present():
        self.select_first_expression()
        self.click_remove()
```
so we are deleting expression after expression and in the end, we are left with `"<new element>"` string. The condition that I removed from here would never be True for this case.

{{ pytest: -v --long-running cfme/tests/control/test_basic.py::test_modify_condition_expression }}

### DIFF
--- a/widgetastic_manageiq/expression_editor.py
+++ b/widgetastic_manageiq/expression_editor.py
@@ -272,7 +272,7 @@ class ExpressionEditor(View, Pretty):
         before = self.expression_text.encode("utf-8").strip()
         prog()
         wait_for(
-            lambda: self.expression_text != "<new element>" and self.expression_text != before,
+            lambda: self.expression_text != before,
             handle_exception=True,
             num_sec=10,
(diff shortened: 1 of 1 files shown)


apagac:
@john-dupuy can you look at this please, according to git blame you added the wait_for in question.

lina-nikiforova:
I also faced the problem with this `<new_element>` and tried to fix it in this PR
https://github.com/ManageIQ/integration_tests/pull/8505/files#diff-72044b72b08158d08006e75775fe3bd3R283

john-dupuy:
@apagac Can you please run this change against `test_condition_crud`, that change was made in `ExpressionEditor` to address test failures in this test. c.f. #8103

PR #8578, opened 2019-03-14
[1LP][RFR] Move `delete_if_exists` to BaseEntity

valaparthvi:
1. Move `delete_if_exists` method to `BaseEntity`. The method raises a `NotImplementedError` if `delete` method has not been implemented for an entity.
2. Remove `delete_if_exists` from all other places.

{{ pytest: cfme/tests/automate/custom_button/test_generic_objects.py::test_custom_button_display cfme/tests/automate/test_class.py::test_class_crud cfme/tests/automate/test_domain.py::test_domain_crud cfme/tests/automate/test_namespace.py::test_wrong_namespace_name cfme/tests/automate/test_service_dialog.py::test_checkbox_dialog_element cfme/tests/cloud/test_providers.py::test_add_cancelled_validation_cloud  --use-template-cache -sqvvvv }}

### DIFF
--- a/cfme/modeling/base.py
+++ b/cfme/modeling/base.py
@@ -200,6 +200,21 @@ class BaseEntity(NavigatableMixin):
         else:
             return True
 
+    def delete_if_exists(self, *args, **kwargs):
+        """Combines ``.exists`` and ``.delete()`` as a shortcut for ``request.addfinalizer``
+
+        Raises
+            NotImplementedError: If the ``.delete()`` is not implemented for the object
+        Returns: True if object existed and delete was initiated, False otherwise
+        """
+        if self.exists:
+            try:
+                self.delete(*args, **kwargs)
+                return True
+            except AttributeError:
+                raise NotImplementedError("Delete method is not implemented.")
+        return False
+
 
 @attr.s
 class CollectionProperty(object):
--- a/cfme/automate/buttons.py
+++ b/cfme/automate/buttons.py
@@ -315,10 +315,6 @@ class BaseButton(BaseEntity, Updateable):
                 view.submit_button.click()
             view.flash.assert_no_error()
 
-    def delete_if_exists(self):
-        if self.exists:
-            self.delete()
-
 
 @attr.s
 class DefaultButton(BaseButton):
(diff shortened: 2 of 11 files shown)


cfme/modeling/base.py
```diff
@@ -200,6 +200,21 @@ def exists(self):
         else:
             return True
 
+    def delete_if_exists(self, *args, **kwargs):
```
digitronik: Q: why `*args` and `**kwargs`?

cfme/modeling/base.py
```diff
@@ -200,6 +200,21 @@ def exists(self):
...
+        """Combines ``.exists`` and ``.delete()`` as a shortcut for ``request.addfinalizer``
+
+        Raises
+            NotImplementedError: If the ``.delete()`` is not implemented for the object
+        Returns: True if object existed and delete was initiated, False otherwise
+        """
+        if self.exists:
+            try:
+                self.delete(*args, **kwargs)
```
izapolsk: OPTIONAL: I'd consider moving this to arg of delete call if it isn't sophisticated for sure. Something like `ignore_not_exists` or `ignore_errors`.

valaparthvi: @izapolsk Can you tell in a little more detail? I don't understand what you're suggesting.

izapolsk: @valaparthvi I mean I lean towards considering such `delete_if_exists`  method as redundant. I'd better add option 'ignore_errors' to delete method.
it's worth discussing this. sure thing it depends on how many changes we will have to put in.

@mshriver ^^

mshriver: Ah, good point @izapolsk.  I'd like to be specific in this case, consider: `def delete(self, ignore_missing)` or `ignore_dne` or `ignore_exists`.

cfme/modeling/base.py
mshriver: In case a unique entity wants to implement this method separately, something running against many entities can still make the same function call, and pass extra arguments where needed. The core implementation will just ignore the extra args.

cfme/modeling/base.py
mshriver: Even more discussion was had offline - because of the common use of this method with `addfinalizer` calls, a separate method is useful in order to avoid lambda expressions.  Secondly, the implementation of `delete` itself is not the same across all entities, and we would not be defining it (or the kwarg) at a common place.

PR #8582, opened 2019-03-15
[1LP][RFR] Increase IPA timeout

apagac:
__Fixing__ `test_login_evm_group` errors for 5.10 when running with IPA.

From my testing, it seems that for whatever reason the process of setting up IPA authentication method takes more time in 5.10 than it takes in 5.9. This causes timeouts on some auth tests when running against 5.10.
The proposed solution is the easiest one, I just upped the timeout globally by 10 seconds. This seems to work.

{{ pytest: -v --long-running cfme/tests/integration/test_cfme_auth.py::test_login_evm_group[external-freeipa01-uid-ldapuser1] }}

### DIFF
--- a/cfme/utils/appliance/__init__.py
+++ b/cfme/utils/appliance/__init__.py
@@ -203,7 +203,7 @@ class ApplianceConsoleCli(object):
     def __init__(self, appliance):
         self.appliance = appliance
 
-    def _run(self, appliance_console_cli_command, timeout=10):
+    def _run(self, appliance_console_cli_command, timeout=35):
         result = self.appliance.ssh_client.run_command(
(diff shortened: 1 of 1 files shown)


apagac:
Actually, this might not be enough. For `test_user_group_switching` with IPA, it took 29 seconds before the IPA was successfully set up. I measured this for both 5.9 and 5.10.
Setting this back to WIP for now, need to find out if other IPA tests work, then up the timeout to 35 I think.

apagac:
5.9: 19 seconds
5.10: 29 seconds

dajoRH:
*CFME QE Bot*

PR #8583, opened 2019-03-15
[1LP][RFR] provision cfme from azure

lina-nikiforova:
{{ pytest: -v cfme/tests/cloud/test_providers.py::test_create_azure_vm_from_azure_image --use-provider=complete  --long-running }}

Automation for `test_create_azure_vm_from_azure_image`. In fact, also automates `test_upload_azure_image_to_azure` as it's part of the test.

### DIFF
--- a/cfme/tests/cloud/test_providers.py
+++ b/cfme/tests/cloud/test_providers.py
@@ -1,12 +1,15 @@
 # -*- coding: utf-8 -*-
 # pylint: disable=E1101
 # pylint: disable=W0621
+import os
 import uuid
 
 import fauxfactory
 import pytest
+from six.moves.urllib.parse import urljoin
 from wait_for import wait_for
 from widgetastic.exceptions import MoveTargetOutOfBoundsException
+from wrapanapi import VmState
 
 from cfme import test_requirements
 from cfme.base.credential import Credential
@@ -22,18 +25,164 @@ from cfme.fixtures.provider import enable_provider_regions
 from cfme.fixtures.pytest_store import store
 from cfme.markers.env_markers.provider import ONE
 from cfme.rest.gen_data import arbitration_profiles as _arbitration_profiles
+from cfme.utils import appliance
+from cfme.utils import conf
+from cfme.utils import ssh
 from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.conf import credentials
+from cfme.utils.generators import random_vm_name
 from cfme.utils.providers import list_providers
 from cfme.utils.providers import ProviderFilter
 from cfme.utils.rest import assert_response
 from cfme.utils.update import update
+from cfme.utils.wait import wait_for_decorator
 
 pytestmark = [pytest.mark.provider([CloudProvider], scope="function")]
+
+# path, where powershell scripts are located and where cfme image is downloaded,
+# needed for test_create_azure_vm_from_azure_image
+SPACE = '/mnt/space/'
 
 
 @pytest.fixture(scope='function')
 def enable_regions(provider):
     enable_provider_regions(provider)
+
+
+@pytest.fixture(scope='function')
+def az_pwsh_vm(appliance):
+    """
+    azure_pwsh contains powershell and necessary scripts to upload vhd, create VM, get ip of the
+    resource and delete the VM.
+    Find the provider that contains that template.
+
+    The example of the configuration can be found in data/az_pwsh_cloudinit.cfg
+    """
+    filter_pwsh_template = ProviderFilter(required_fields=[['templates', 'powershell_vm']])
+    providers = list_providers(filters=[filter_pwsh_template])
+    if not providers:
+        pytest.skip("There's no provider that contains a template with powershell")
+
+    # If there's more than 1 provider that has the template, we select the first
+    provider = providers[0]
+
+    vm_name = random_vm_name(context="pwsh")
+    pwsh_vm = provider.data.templates.powershell_vm.name
+
+    collection = provider.appliance.provider_based_collection(provider)
+
+    try:
+        vm = collection.instantiate(vm_name, provider, pwsh_vm)
+        vm.create_on_provider(allow_skip="default")
+    except IndexError:
+        from cfme.exceptions import ItemNotFound
+        raise ItemNotFound('VM with powershell not found!')
+
+    vm.mgmt.ensure_state(VmState.RUNNING)
+
+    @wait_for_decorator(timeout="10m", delay=5)
+    def get_ip_address():
+        ip = vm.mgmt.ip
+        return ip is not None
+    yield vm
+
+    vm.cleanup_on_provider()
+
+
+@pytest.fixture
+def pwsh_ssh(az_pwsh_vm):
+    """Provide vm_ssh_client for ssh operations in the test."""
+    with ssh.SSHClient(hostname=az_pwsh_vm.ip_address,
+                       username=credentials['host_default']['username'],
+                       password=credentials['host_default']['password']) as vm_ssh_client:
+        yield vm_ssh_client
+
+
+@pytest.fixture
+def connect_az_account(pwsh_ssh):
+    """
+    Connect to Azure account to run further scripts, see
+    https://docs.microsoft.com/en-us/powershell/azure/authenticate-azureps
+    """
+    path_script = os.path.join(SPACE, 'connect_account.ps1')
+    connect = pwsh_ssh.run_command("pwsh {}".format(path_script), timeout=180)
+    assert connect.success, "Failed to connect to Azure account"
+
+
+@pytest.fixture(scope='function')
+def cfme_vhd(appliance, pwsh_ssh):
+    path_script = os.path.join(SPACE, 'get_ip.ps1')
+    ip_of_recourse = pwsh_ssh.run_command(
+        r'pwsh {}| grep -oE "([0-9]{{1,3}}\.){{3}}[0-9]{{1,3}}"'.format(path_script),
+        timeout=60).output.strip()
+    if ip_of_recourse is not None:
+        pytest.skip("The resource is taken by some other VM in Azure")
+    stream = appliance.version.stream()
+    try:
+        # need to add the trailing slash for urljoin to work correctly
+        url = '{}/'.format(conf.cfme_data['basic_info']['cfme_images_url'][stream])
+    except KeyError:
+        pytest.skip("Skipping since no such key found in yaml")
+
+    image = pwsh_ssh.run_command(
+        """wget -qO- {url} | grep -Po '(?<=href=")[^"]*' | grep azure""".format(url=url),
+        timeout=30).output.strip()
+    image_url = urljoin(url, image)
+    pwsh_ssh.run_command("wget {image} -P {dest}".format(dest=SPACE, image=image_url),
+                         timeout=180)
+
+    # unpack the archive
+    vhd = image.replace('zip', 'vhd')
+    pwsh_ssh.run_command("unzip {} -d {}".format(os.path.join(SPACE, image), SPACE),
+                         timeout=15 * 60)
+    yield vhd
+
+    pwsh_ssh.run_command("rm -f {}".format(os.path.join(SPACE, image)), timeout=180)
+    pwsh_ssh.run_command("rm -f {}".format(os.path.join(SPACE, vhd)), timeout=180)
+
+
+@pytest.fixture(scope='function')
+def upload_image_to_azure(cfme_vhd, pwsh_ssh):
+    path_script = os.path.join(SPACE, 'upload_vhd.ps1')
+    pwsh_ssh.run_command(
+        r"""sed -i '1s/.*/$BlobNameSource = "{vhd}"/' {script}"""
+            .format(script=path_script, vhd=cfme_vhd), timeout=30)
+
+    pwsh_ssh.run_command("pwsh {}".format(path_script), timeout=15 * 60)
+
+
+@pytest.fixture(scope='function')
+def vm_ip(cfme_vhd, pwsh_ssh):
+    # Create VM in Azure
+    path_script = os.path.join(SPACE, 'create_vm.ps1')
+    pwsh_ssh.run_command(
+        r"""sed -i '1s/.*/$BlobNameSource = "{vhd}"/' {script} &&
+        sed -i '2s/.*/$BlobNameDest = "{b_dest}"/' {script} &&
+        sed -i '3s/.*/$VMName = "{name}"/' {script}""".format(
+            script=path_script,
+            vhd=cfme_vhd,
+            b_dest=cfme_vhd.replace('azure', 'test'),
+            name=cfme_vhd.replace('.x86_64.vhd', '-vm')),
+        timeout=20)
+
+    pwsh_ssh.run_command("pwsh {}".format(path_script), timeout=600)
+
+    # get the ip of the resource
+    path_get_ip = os.path.join(SPACE, 'get_ip.ps1')
+    ip = pwsh_ssh.run_command(
+        r'pwsh {}| grep -oE "([0-9]{{1,3}}\.){{3}}[0-9]{{1,3}}"'.format(path_get_ip),
+        timeout=60).output.strip()
+    yield ip
+
+    # Delete the VM
+    with pwsh_ssh:
+        pwsh_ssh.run_command(
+            r"""sed -i '1s/.*/$VMName = "{name}"/' {script}""".format(
+                script=path_script,
+                name=cfme_vhd.replace('.x86_64.vhd', '-vm')),
+            timeout=20)
+
+        pwsh_ssh.run_command("pwsh {}".format(path_script), timeout=180)
 
 
 @pytest.mark.tier(3)
@@ -666,7 +815,6 @@ class TestProvidersRESTAPI(object):
             assert 'SecurityGroup' in security_groups[0]['type']
 
 
-
 @pytest.mark.provider([CloudProvider], override=True, selector=ONE)
 def test_tagvis_provision_fields(setup_provider, request, appliance, user_restricted, tag,
                                  soft_assert):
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit e9e422f15bdfb4b1aaf8d806afe040e166a79b36:

- :bangbang: Line 77:90: **E226** *missing whitespace around arithmetic operator*
- :bangbang: Line 80:78: **E231** *missing whitespace after ','*
- :bangbang: Line 90:76: **E226** *missing whitespace around arithmetic operator*
- :heavy_exclamation_mark: Line 97:84: **W291** *trailing whitespace*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/cloud/test_providers.py
```diff
@@ -36,6 +41,84 @@ def enable_regions(provider):
     enable_provider_regions(provider)
 
 
+@pytest.fixture(scope='module')
+def az_pwsh_vm():
+    # azure_pwsh contains powershell and necessary scripts
+    provider = get_crud_by_name('rhv42')
```
mshriver: no hardcoding a provider name here, needs to be picked from cfme_data.management_systems via looking for 'powershell_vm'.

cfme/tests/cloud/test_providers.py
```diff
@@ -36,6 +41,84 @@ def enable_regions(provider):
...
+    pwsh_vm = provider.data.powershell_vm
+    collection = provider.appliance.provider_based_collection(provider)
+    try:
+        return collection.instantiate(pwsh_vm, provider)
+    except IndexError:
+        from cfme.exceptions import ItemNotFound
+        raise ItemNotFound('VM with powershell not found!')
```
mshriver: This should be a `pytest.skip` call, we don't have the resource necessary to run the test.

dajoRH:
Lint report for commit 83744d1dc143a0ad86a174e4a5f3067f51a7af18:

* 3861a14d8e5b9dac0cb298064654eaaf005c30c1

*CFME QE Bot*

cfme/tests/cloud/test_providers.py
```diff
@@ -853,42 +963,11 @@ def test_market_place_images_azure():
...
-        title: Upload Azure image to Azure
-    """
-    pass
-
-
-@pytest.mark.manual
+@pytest.mark.long_running
 @pytest.mark.tier(1)
-def test_create_azure_vm_from_azure_image():
+@pytest.mark.provider([RHEVMProvider], selector=ONE_PER_TYPE, override=True,
```
digitronik: Q: just suspect why rhv provider?

lina-nikiforova: @digitronik because I need to use this provider as it has the necessary template to provision a VM with powershell and powershell scripts.

cfme/tests/cloud/test_providers.py
```diff
@@ -36,6 +46,107 @@ def enable_regions(provider):
     enable_provider_regions(provider)
 
 
+@pytest.fixture(scope='function')
+def az_pwsh_vm(provider, appliance, request):
+    # azure_pwsh contains powershell and necessary scripts
+    vm_name = random_vm_name(context="pwsh")
+    template = provider.data.get('powershell_vm')
+    vm = provider.appliance.collections.infra_vms.instantiate(vm_name, provider, template)
```
mshriver: Why infra_vms collection here, for what I'm assuming is a fixture used by tests parametrized on Azure?

You can use `appliance.provider_based_collection` to get the right collection type.

As a second note, although the ProviderFilter required_fields is used to only parametrize on providers with 'powershell_vm', I think it would be good to wrap that reference in a try/except that would skip the test should this fixture be used in other test functions that forget to include this required_field.

cfme/tests/cloud/test_providers.py
```diff
@@ -36,6 +46,107 @@ def enable_regions(provider):
...
+
+@pytest.fixture(scope='function')
+def cfme_vhd(appliance, pwsh_ssh):
+    ip_of_recourse = pwsh_ssh.run_command(
+        'cd /mnt/space && pwsh get_ip.ps1 | grep -oE "([0-9]{1,3}\.){3}[0-9]{1,3}"',
+        timeout=60).output.strip()
+    if ip_of_recourse is not None:
+        pytest.skip("The resource is taken by some other VM in Azure")
+    stream = appliance.version.stream()
+    url = conf.cfme_data['basic_info']['cfme_images_url'][stream]
```
mshriver: I think this check should be wrapped for a pytest.skip call when the url match isn't found for the stream

cfme/tests/cloud/test_providers.py
```diff
@@ -36,6 +46,107 @@ def enable_regions(provider):
...
+    image = pwsh_ssh.run_command(
+        """wget -qO- {url} | grep -Po '(?<=href=")[^"]*' | grep azure""".format(url=url),
+        timeout=30).output.strip()
+    image_url = '{}/{}'.format(url, image)
+    pwsh_ssh.run_command("cd /mnt/space && wget {image}".format(image=image_url), timeout=180)
```
mshriver: Consider defining the first part of this command in a variable and formatting it into the strings.

cfme/tests/cloud/test_providers.py
```diff
@@ -36,6 +46,107 @@ def enable_regions(provider):
...
+    yield vhd
+
+    pwsh_ssh.run_command("cd /mnt/space && rm -f {image}".format(image=image), timeout=180)
+    pwsh_ssh.run_command("cd /mnt/space && rm -f {vhd}".format(vhd=vhd), timeout=180)
+
+
+@pytest.fixture(scope='function')
+def upload_image_to_azure(cfme_vhd, pwsh_ssh):
+    pwsh_ssh.run_command(
+        """cd /mnt/space && sed -i "1s/.*/\$BlobNameSource = \\"{vhd}\\"/" upload_vhd.ps1"""
```
mshriver: The more I see /mnt/space used, the more I think this should be a module-level 'constant', something like `CD_SPACE`

cfme/tests/cloud/test_providers.py
mshriver: This should be written a bit more generically, as the fact that CFME QE has this template on a RHEV provider is unique to us.

I think there's an assumption here, that's not immediately apparent to me - there is an azure account required to run this.  Is the azure powershell VM already setup with the account credentials? 

I think that the fixtures for this should be structured to look for a specific definition in cfme-qe-yamls of where an azure powershell vm can be run. This could take different forms, but I think it should use the current structure of `provider.data.templates` to list the template name and creds under a specific key (following the pattern of full_template, console_template, small_template, etc).

The fixtures should look to identify any provider that has one of these VM templates defined, and somehow pick one.
The expected setup of the azure VM (credentials, installed packages) should be defined at least in a docblock, even better as a cloud-init file.

cfme/tests/cloud/test_providers.py
```diff
@@ -959,4 +1038,34 @@ def test_create_azure_vm_from_azure_image():
                   VM is completely Stopped in Azure.
         title: Create Azure VM from Azure image
     """
-    pass
+    app = appliance.IPAppliance.from_url(vm_ip)
+
+    username = credentials['azure_appliance']['username']
```
mshriver: Is this a different credential than the one use for the powershell VM itself? I'm not quite clear what credential is being pulled here and used for the SSH client to that VM's IP.  Can this credential key be defined with the template name (see above comment about provider parametrization)

cfme/tests/cloud/test_providers.py
lina-nikiforova: @mshriver right, I started to mess around with that because I wanted to provision a preallocated disk, however, it appears to be not supported in RHV, so I'm reverting to my original implementation

cfme/tests/cloud/test_providers.py
lina-nikiforova: @mshriver yes, it's different. I'm adding a comment there.

lina-nikiforova:
@mshriver thanks a lot for the comments, as I moved from having one VM to the template, I actually missed connecting to the Azure account :woman_facepalming: 
I hope I addressed other concerns as well

dajoRH:
Lint report for commit 01d81415f612d86fd49761c8c14c61912de1cbbe:

- :red_circle: Line 24:1: **F401** *'cfme.infrastructure.provider.rhevm.RHEVMProvider' imported but unused*
- :red_circle: Line 26:1: **F401** *'cfme.markers.env_markers.provider.ONE_PER_TYPE' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

cfme/tests/cloud/test_providers.py
```diff
@@ -22,20 +23,163 @@
...
+    for provider, values in providers_data.items():
+        templates = values.get('templates')
+        if templates:
+            az_template = templates.get('powershell_vm')
+            if az_template:
+                provider_key = provider
+
+    if not provider_key:
+        pytest.skip("There's no provider that contains a template with powershell")
+    provider = get_crud_by_name(provider_key)
```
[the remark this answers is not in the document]
lina-nikiforova: @izapolsk well, this works, it replaced the hardcoded provider as I did before https://github.com/ManageIQ/integration_tests/pull/8583#discussion_r266860703

cfme/tests/cloud/test_providers.py
```diff
@@ -22,20 +23,163 @@
...
+    resource and delete the VM.
+    Find the provider that contains that template.
+    TODO(anikifor): define the configuration of that VM as cloud-init file
+    """
+    providers_data = cfme_data.get("management_systems", {})
+
+    for provider, values in providers_data.items():
+        templates = values.get('templates')
+        if templates:
+            az_template = templates.get('powershell_vm')
```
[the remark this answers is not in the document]
lina-nikiforova: @izapolsk no, there's a template, from which we provision the VM. To recreate it we can use #8663

cfme/tests/cloud/test_providers.py
```diff
@@ -760,9 +965,12 @@ def test_upload_azure_image_to_azure():
 
 @pytest.mark.manual
```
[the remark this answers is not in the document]
lina-nikiforova: thanks! This is the result of my rebasing skills, my bad.

cfme/tests/cloud/test_providers.py
```diff
@@ -22,20 +23,163 @@
...
+    try:
+        url = conf.cfme_data['basic_info']['cfme_images_url'][stream]
+    except KeyError:
+        pytest.skip("Skipping since no such key found in yaml")
+
+    image = pwsh_ssh.run_command(
+        """wget -qO- {url} | grep -Po '(?<=href=")[^"]*' | grep azure""".format(url=url),
+        timeout=30).output.strip()
+    image_url = '{}/{}'.format(url, image)
+    pwsh_ssh.run_command("{cd} wget {image}".format(cd=CD_SPACE, image=image_url), timeout=180)
```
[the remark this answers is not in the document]
lina-nikiforova: @izapolsk  can you explain the benefits? Of course, I know the arguments but to me the current way is more clear.

cfme/tests/cloud/test_providers.py
```diff
@@ -22,20 +23,163 @@
...
+        yield vm_ssh_client
+
+
+@pytest.fixture
+def connect_az_account(pwsh_ssh):
+    """
+    Connect to Azure account to run further scripts, see
+    https://docs.microsoft.com/en-us/powershell/azure/authenticate-azureps
+    """
+    pwsh_ssh.run_command("{} pwsh connect_account.ps1".format(CD_SPACE), timeout=180)
```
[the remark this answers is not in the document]
lina-nikiforova: added

lina-nikiforova:
@izapolsk Thanks for the review! I'd also like to consider https://github.com/ManageIQ/integration_tests/pull/8583/files#diff-4f10bc4cf82a0f08488c8079d4cd0085R934

cfme/tests/cloud/test_providers.py
```diff
@@ -22,20 +23,164 @@
...
+    TODO(anikifor): define the configuration of that VM as cloud-init file
+    """
+    providers_data = cfme_data.get("management_systems", {})
+
+    for provider, values in providers_data.items():
+        templates = values.get('templates')
+        if templates:
+            az_template = templates.get('powershell_vm')
+            if az_template:
+                provider_key = provider
```
mshriver: Can't you call `break` at this point, since you found a `provider_key` with the template defined? No need to continue iterating through the list.

cfme/tests/cloud/test_providers.py
```diff
@@ -22,20 +23,164 @@
...
+CD_SPACE = 'cd /mnt/space && '
+
 
 @pytest.fixture(scope='function')
 def enable_regions(provider):
     enable_provider_regions(provider)
 
 
+@pytest.fixture(scope='function')
+def az_pwsh_vm(appliance, request):
```
mshriver: request is unused here, are you expecting indirect parametrization?

cfme/tests/cloud/test_providers.py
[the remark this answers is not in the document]
mshriver: This is only working by chance, when the 'name' field matches the provider key. You need to use `get_crud` here.

```
In [1]: providers.get_crud_by_name('rhv43')

Out[1]: RHEVMProvider(endpoints={'default': <cfme.infrastructure.provider.rhevm.RHEVMEndpoint object at 0x7f6664440110>, 'candu': <cfme.common.provider.CANDUEndpoint object at 0x7f6664440250>}, name='rhv43', key='rhv43', zone='default', start_ip='<REDACTED>', end_ip='<REDACTED>', provider_data=None)

In [2]: providers.get_crud_by_name('vsphere67-ims')
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-2-05d91c5e044c> in <module>()
----> 1 providers.get_crud_by_name('vsphere67-ims')

/home/mshriver/repos/integration_tests/cfme/utils/providers.py in get_crud_by_name(provider_name)
    341         if provider_data.get("name") == provider_name:
    342             return get_crud(provider_key)
--> 343     raise NameError("Could not find provider {}".format(provider_name))
    344 
    345 

NameError: Could not find provider vsphere67-ims

In [3]: providers.get_crud('vsphere67-ims')
Out[3]: VMwareProvider(endpoints={'default': <cfme.infrastructure.provider.virtualcenter.VirtualCenterEndpoint object at 0x7f6664421d90>}, name='vSphere 6.7 IMS', key='vsphere67-ims', zone='default', start_ip='<REDACTED>', end_ip='<REDACTED>', provider_data=None)
```

cfme/tests/cloud/test_providers.py
```diff
@@ -22,20 +23,164 @@
...
+    providers_data = cfme_data.get("management_systems", {})
+
+    for provider, values in providers_data.items():
+        templates = values.get('templates')
+        if templates:
+            az_template = templates.get('powershell_vm')
+            if az_template:
+                provider_key = provider
+
+    if not provider_key:
```
mshriver: `provider_key` is set within the `if` blocks above, and if there was no matching provider found, this will trigger an `UnboundLocalError`.

I think you need a bit more efficient structure here to handle these cases. Take advantage of the tools in `cfme.utils.providers`.

```
filter = cfme.utils.providers.ProviderFilter(required_fields=[['templates', 'powershell_vm']])
provider_objects_list = cfme.utils.providers.list_providers(filters=[filter])
```

cfme/tests/cloud/test_providers.py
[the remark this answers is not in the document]
mshriver: Please see my comment above though, as you should not have to actually call `get_crud` at all when you're just using `cfme.utils.providers` methods.

cfme/tests/cloud/test_providers.py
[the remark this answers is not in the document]
mshriver: I think I agree with @izapolsk here, it would be better to use such options (and in other commands provide direct paths).  There are no assertions against the ssh commands made below, and if the `cd` fails for some reason, all the subsequent commands will fail.

To me its more clear to provide full paths to the files we're manipulating, and you should be able to construct those with the constant.

cfme/tests/cloud/test_providers.py
```diff
@@ -727,42 +871,74 @@ def test_domain_id_validation(request, provider):
 
 @pytest.mark.manual
 @pytest.mark.tier(1)
-def test_upload_azure_image_to_azure():
+def test_sdn_nsg_arrays_refresh_azure():
     """
-    Upload image copied from RHCF3-11271 to azure using powershell
-    similiar the script in the setup section.
+    https://bugzilla.redhat.com/show_bug.cgi?id=1520196
```
mshriver: Please use `@pytest.mark.meta(automates=[BZ(1520196)])` to indicate test coverage.

cfme/tests/cloud/test_providers.py
```diff
@@ -727,42 +871,74 @@ def test_domain_id_validation(request, provider):
...
-               $LocalFile = $LocalFilePath + $LocalFileName
-               Add-AzureRmVhd -ResourceGroupName $ResourceGroupName -Destination
-               $BlobDestination -LocalFilePath $LocalFile -NumberOfUploaderThreads 8
+        initialEstimate: 1/6h
+        testSteps:
+            1. Add Network Security group on Azure with coma separated port ranges
+            `1023,1025` rule inbound/outbound ( ATM this feature is not allowed in
+            East US region of Azure - try West/Central)
+            2. Add such Azure Region into CFME
+            3. Refresh provider
```
mshriver: Need `expectedResults` for a manual test case, so its clear what is being validated or checked here.

cfme/tests/cloud/test_providers.py
```diff
@@ -727,42 +871,74 @@ def test_domain_id_validation(request, provider):
...
+    pass
+
+
+@pytest.mark.manual
+@pytest.mark.tier(2)
+def test_provider_flavors_azure():
+    """
+    Verify that the vm flavors in Azure are of the correct sizes and that
+    the size display in CFME is accurate.  This test checks a regression
+    of https://bugzilla.redhat.com/show_bug.cgi?id=1357086
```
mshriver: Please use docblock section to list BZ's associated with the manual case (no pytest marker at this time for manual cases)

```
Bugzilla:
    1357086
```

cfme/tests/cloud/test_providers.py
```diff
@@ -727,42 +871,74 @@ def test_domain_id_validation(request, provider):
...
+    when azure adds new sizes.  Only need to spot check a few values.
+    For current size values, you can check here:
+    https://azure.microsoft.com/en-us/documentation/articles/virtual-
+    machines-windows-sizes/
+
+    Polarion:
+        assignee: anikifor
+        casecomponent: Cloud
+        caseimportance: low
+        initialEstimate: 1/8h
```
mshriver: testSteps and expectedResults please, even if they're short.

cfme/tests/cloud/test_providers.py
```diff
@@ -727,42 +871,74 @@ def test_domain_id_validation(request, provider):
...
+    https://bugzilla.redhat.com/show_bug.cgi?id=1491330
+
+    Polarion:
+        assignee: anikifor
+        casecomponent: Cloud
+        caseimportance: medium
+        initialEstimate: 1/6h
+        testSteps:
+            1.Enable market place images
+            2.Verify the list of images
```
mshriver: Please split this to an expectedResults block, with a bit more detail on what is expected to be visible/not visible. Should enabling market place de-list the private images? Not sure from this test case.

cfme/tests/cloud/test_providers.py
```diff
@@ -727,42 +871,74 @@ def test_domain_id_validation(request, provider):
...
     pass
 
 
 @pytest.mark.manual
 @pytest.mark.tier(1)
-def test_create_azure_vm_from_azure_image():
+def test_market_place_images_azure():
     """
-    CFME image must be uploaded to Azure
+    https://bugzilla.redhat.com/show_bug.cgi?id=1491330
```
mshriver: Please use `Bugzilla` docblock section.

cfme/tests/cloud/test_providers.py
```diff
@@ -831,4 +1007,37 @@ def test_create_azure_vm_from_azure_image():
                   VM is completely Stopped in Azure.
         title: Create Azure VM from Azure image
     """
-    pass
+    app = appliance.IPAppliance.from_url(vm_ip)
```
mshriver: Consider using `appliance.Appliance.from_provider()` here, though you may not want to given your fixtures for `vm_ip`

cfme/tests/cloud/test_providers.py
```diff
@@ -831,4 +1007,37 @@ def test_create_azure_vm_from_azure_image():
...
+
+    # Credentials for the provisioned VM from CFME image, this is different to the VM that runs
+    # powershell scripts as Azure has specific requirements for login/password.
+    # These credentials are used in the script create_vm.ps1 to provision the VM.
+    username = credentials['azure_appliance']['username']
+    password = credentials['azure_appliance']['password']
+
+    with ssh.SSHClient(hostname=vm_ip,
```
mshriver: I would love to see an enhancement that allows for passing username/password to `IPAppliance.__init__`, setting local vars during class construction.  Some `IPAppliance.username` and `IPAppliance.password` properties could return these local vars (`self._username`).

Then `IPAppliance.ssh_client` could use those properties, defaulting to the current conf file lookup when nothing is passed.

Then, when you create an appliance object here to test against, you can still use `app.ssh_client`, and it will automatically use the passed credentials instead of the default.

Just a thought, for a framework enhancement that would assist with this kind of testing.  There's already some code there for handling openshift creds for appliances.

cfme/tests/cloud/test_providers.py
```diff
@@ -831,4 +1007,37 @@ def test_create_azure_vm_from_azure_image():
...
+
+        # permit root login over ssh for future appliance configuration
+        command = 'sed -i "s/.*PermitRootLogin.*/PermitRootLogin yes/g" /etc/ssh/sshd_config'
+        config = app_ssh_client.run_command(
+            'echo {} | sudo -S {}'.format(password, command), ensure_user=True)
+        assert config.success
+
+        # restart sshd to apply configuration changes
+        restart = app_ssh_client.run_command(
+            'echo {} | sudo -S systemctl restart sshd'.format(password), ensure_user=True)
```
mshriver: Consider adding a `SystemdService` attribute to IPAppliance for `sshd`

cfme/tests/cloud/test_providers.py
lina-nikiforova: definitely, I missed that

cfme/tests/cloud/test_providers.py
lina-nikiforova: no, I wanted to add finalizer but implemented it using `yield`, so it's indeed not needed

cfme/tests/cloud/test_providers.py
lina-nikiforova: @mshriver huge thanks! this is so much better.

cfme/tests/cloud/test_providers.py
lina-nikiforova: hmmm, as it's a manual test, I don't think it's correct.
I'm adding it like this
```
    Bugzilla:
        1520196
```
as I see it in other _manual_ tests as well

cfme/tests/cloud/test_providers.py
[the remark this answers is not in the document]
lina-nikiforova: @mshriver ok, I'm changing it

cfme/tests/cloud/test_providers.py
lina-nikiforova: well, I didn't want to do changes to manual tests as I think it's out of scope of this PR, but why not

cfme/tests/cloud/test_providers.py
lina-nikiforova: @mshriver totally agree, Mike. Although I would prefer much more to have smaller PRs and this one is growing already to editing manual tests that I didn't touch here. I can try to raise a follow-up PR to do such changes.

cfme/tests/cloud/test_providers.py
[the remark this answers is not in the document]
izapolsk: @lina-nikiforova, there are several reasons why I'd better see using arguments there.
 main reason is that you rely on default behavior of tools which can be easily changed and may vary from system to system or between versions. so, you need to exactly specify what you want this tool to do.

cfme/tests/cloud/test_providers.py
```diff
@@ -22,20 +23,160 @@
...
+    stream = appliance.version.stream()
+    try:
+        url = conf.cfme_data['basic_info']['cfme_images_url'][stream]
+    except KeyError:
+        pytest.skip("Skipping since no such key found in yaml")
+
+    image = pwsh_ssh.run_command(
+        """wget -qO- {url} | grep -Po '(?<=href=")[^"]*' | grep azure""".format(url=url),
+        timeout=30).output.strip()
+    image_url = '{}/{}'.format(url, image)
```
izapolsk: I'd consider using urljoin or something similar

cfme/tests/cloud/test_providers.py
```diff
@@ -22,20 +23,160 @@
...
+        yield vm_ssh_client
+
+
+@pytest.fixture
+def connect_az_account(pwsh_ssh):
+    """
+    Connect to Azure account to run further scripts, see
+    https://docs.microsoft.com/en-us/powershell/azure/authenticate-azureps
+    """
+    connect = pwsh_ssh.run_command("pwsh {}connect_account.ps1".format(SPACE), timeout=180)
```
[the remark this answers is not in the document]
lina-nikiforova: @izapolsk thanks, gonna update

cfme/tests/cloud/test_providers.py
lina-nikiforova: @izapolsk right, makes sense

lina-nikiforova:
@izapolsk I just pushed the change with the path as per your comments.
The resource is taken by another VM, I can look into it more later but it definitely worked when I tested.

cfme/tests/cloud/test_providers.py
mshriver: oh, absolutely don't add that here, was just throwing the idea out there :)

PR #8585, opened 2019-03-18
[1LP][RFR] ndhandre manual move

digitronik:
1. Move multi-tenancy test to respective file and change owner
2. Remove one test as its cover with other.

### DIFF
--- a/cfme/tests/automate/custom_button/test_service_objects.py
+++ b/cfme/tests/automate/custom_button/test_service_objects.py
@@ -471,9 +471,10 @@ def test_custom_button_on_vm_resource_detail(context):
             2. In normal UI - OPS, Provision test service using any infra provider
             (nvc55 recommended)
             3. In SSUI, Check custom button on VM resource details page
+            4. Check button execution
 
     Bugzilla:
-        1427430
+        1427430, 1450473, 1454910
     """
     pass
 
--- a/cfme/tests/automate/custom_button/test_service_objects.py
+++ b/cfme/tests/automate/custom_button/test_service_objects.py
@@ -466,13 +466,60 @@ def test_custom_button_on_vm_resource_detail(context):
         startsin: 5.9
         casecomponent: CustomButton
         tags: custom_button
+        setup:
+            1. Add Provider
+            2. Refresh Provider; Data should be collected.
+            3. Create Simple Dialog for catalog
+                * Navigate to Automation > Automate > Customization
+                * Click on All Dialogs from a sidebar
+                * From toolbar select "Add a new Dialog"
+                * Fill Dialog's name and description
+                * Drag TextBox (we can select other as well)
+                * Save changes
+            4. Create Catalog
+                * Navigate to Services > Catalogs
+                * Click on Catalogs from a sidebar
+                * From toolbar Configuration select "Add New Catalog"
+                * Fill name and description
+                * Save changes
+            5. Create a Catalog item
+                * Navigate to Services > Catalogs
+                * From sidebar select All Catalogs > catalog (created above)
+                * From toolbar select Add New catalog item
+                * Select Provider
+                * Fill Name, description, catalog and dialog (created above)
+                * Select VM name proper template etc...
+            6. Order Catalog
+                * Navigate to Services > Catalogs
+                * Click on Service catalogs from a sidebar
+                * Order catalog from this Page
         testSteps:
-            1. Add custom button for VM/Instance object from automation
-            2. In normal UI - OPS, Provision test service using any infra provider
-            (nvc55 recommended)
-            3. In SSUI, Check custom button on VM resource details page
-            4. Check button execution
-
+            1. Add custom button group for VM/Instance object from automation
+                * Navigate to Automate > Automation > Customization
+                * Click on Buttons from a sidebar
+                * Select VM/Instance
+                * From configuration select Add new Button Group
+                * Fill text, hover, icon, icon color
+                * Save change
+            2. Add custom button in above group
+                * Under this Group Click on configuration and select Add new Button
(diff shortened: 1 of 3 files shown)


cfme/tests/configure/test_access_control.py
```diff
@@ -1926,3 +1926,26 @@ def test_tenant_visibility_miq_ae_namespaces_all_parents():
...
+        initialEstimate: 1/2h
+        tags: cfme_tenancy
+        testSteps:
+            1. Add RHOS provider and perform refresh
+            2. Navigate to Compute -> Clouds -> Providers
+            3. Click on RHOS provider which was added
+            4. In the Relationships table click on Cloud Tenants
+            5. Configuration ->Create Cloud tenant
+            6. Edit created tenant
+            7. Try to delete it
```
jawatts: **Required** For manual only test cases, we should have `testSteps` and `expectedResults`

digitronik: I think its optional... why we need expected result for each step?

digitronik: @mnadeem92 can you please help me here? Actually, this assigned to me but need to move you.

mnadeem92: expectedResult seems optional here as test step already mentioned to execute the CRUD (create/edit and delete) operation, but if you want to include it then expectedResult would be something like "CRUD operation should executed successfully".

mshriver: Yes, manual test cases should have testSteps and expectedResults.

What do you expect to see after the entity is edited? what fields should be edited? should those be reflected immediately, or after some delay? Should the changes be verified anywhere else in the UI?

After delete, should the entity's removal be verified from anywhere else in the UI? Should the same page just be refreshed?

mshriver: This has already been automated: https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/cloud/test_tenant.py

mshriver:
I don't see how this manual test is any different than the automated test that already exists:

https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/cloud/test_tenant.py

Please add to this automated test, or add new cases there, instead of creating a new manual test case.

cfme/tests/configure/test_access_control.py
digitronik: @mshriver thanks for pointing this. :+1: removed

@mshriver still not agreed with you... about expectedResult required. As per me totally optional and its should contributes privilege. How he or she want to write own test. 

Lets take step is "Add provider"
In some contributor mind provider successfully added some think about data collected or not. In this case expected result he must added so other can easy understand. 

Its not possible to collect all observations. Yup good to add expected results but its should optional not strict like "delete provider" is step and contributes need to write "deleted or not".

expectedResults indicate tabular form but same thing can achieve by simple single line statement.

jawatts: @digitronik There does not need to be a result for every test step, some can be left blank. For example:
https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/cloud/test_tag_mapping.py#L57

I believe it should be explicitly stated in `expectedResults` on what the tester is looking for in regards to the test. It also creates consistency across all the manual tests

mshriver: I'm not expecting the test author to "collect all observations" - but there should be some expected result of the test step. As @jawatts has pointed out not all expectedResults lines have to be populated, and we've certainly accepted many PR's with only an expectedResult on the last test step.

I don't plan on being very strict with non-CFME contributors to integration_tests, that want to define some manual test. As long as their docblock is thorough enough, we can push commits that fill out the polarion metadata relevant to the test.

cfme/tests/automate/custom_button/test_service_objects.py
```diff
@@ -471,9 +471,10 @@ def test_custom_button_on_vm_resource_detail(context):
             2. In normal UI - OPS, Provision test service using any infra provider
             (nvc55 recommended)
             3. In SSUI, Check custom button on VM resource details page
+            4. Check button execution
```
mshriver: This should be an expectedResult, along with step 3 as well from the looks of it.

Is the validate here just that the button is visible? Should it be clickable? What should the custom button do? Step 1 is not clear here, what the button should be or do.

Please add some more detail here - "check button execution" is very generic, and I think there's an assumption here that the button has been clicked.   Neither step 2 nor step 3 actually instruct the tester to click the button - just to "check" the custom button, which I assume is an expectedResult that the button is visible.

cfme/tests/configure/test_access_control.py
digitronik: @jawatts  that what I speak... optional, If expectedResults steps are optional means expectedResult tag is also optional for manual or automated.

Above test was just curd... so add/edit/delete operation. I won't find anything required part of expectedResult. I was not even the owner of test just moving as it was with my name with proper FA owner permission. 

Yup, we can ask for the expected result when contributor statement not clear. I don't think force is a better way as creating a test is the owner right. Every contributor has own way of testing.

cfme/tests/automate/custom_button/test_service_objects.py
digitronik: @mshriver  I tried to add explanation as per my understanding. Have a glance.

cfme/tests/automate/custom_button/test_service_objects.py
```diff
@@ -466,14 +466,62 @@ def test_custom_button_on_vm_resource_detail(context):
         startsin: 5.9
         casecomponent: CustomButton
         tags: custom_button
+        setup:
```
mshriver: hell yeah. This will be the easiest automation conversion ever ;)

PR #8586, opened 2019-03-18
[1LP][RFR][NOTEST] updated BZ for test_provision_vm_to_virtual_network

lina-nikiforova:
There's another BZ that blocks the execution of this test


lina-nikiforova:
@digitronik true, although it actually was uncollected in PRT

PR #8591, opened 2019-03-18
[1LP][RFR] docblock for create_view

mshriver:
just adding a docblock here

### DIFF
--- a/cfme/utils/appliance/__init__.py
+++ b/cfme/utils/appliance/__init__.py
@@ -3148,6 +3148,14 @@ class NavigatableMixin(object):
         return self.appliance.browser.widgetastic
 
     def create_view(self, view_class, o=None, override=None, wait=None):
+        """Create a view object given the class and contextual parameters
+
+        Args:
+            view_class (cls): widgetastic.widget.View based class to create an instance of
+            o (obj): Entity object that is the context for the view class (normally linked by nav)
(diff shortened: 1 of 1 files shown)


cfme/utils/appliance/__init__.py
```diff
@@ -3145,6 +3145,14 @@ def browser(self):
         return self.appliance.browser.widgetastic
 
     def create_view(self, view_class, o=None, override=None, wait=None):
+        """Create a vew object given the class and contextual parameters
```
jawatts: typo in view

cfme/utils/appliance/__init__.py
```diff
@@ -3145,6 +3145,14 @@ def browser(self):
...
+
+        Args:
+            view_class (cls): widgetastic.widget.View based class to create an instance of
+            o (obj): Entity object that is the context for the view class (normally linked by nav)
+            override (dict): Dictionary of override values for the contextual object (o)
+            wait: Value to pass to wait_displayed, default of None skips wait_displayed check
```
digitronik: Mike while correcting above put `bool` please.

mshriver: This doesn't have to be a string, which is why I didn't specify. it can be int or string. What types are allowed is determined by `wait_for`

PR #8596, opened 2019-03-19
[1LP][RFR] Updated is_diplayed of AutomateImportExportView

ganeshhubale:
- Fixed reset button xpath
- Added ```reset_title``` widget to read title

{{ pytest: cfme/tests/automate/test_git_import.py -k 'test_automate_git_domain_removed_from_disk' -vv }}

### DIFF
--- a/cfme/base/ui.py
+++ b/cfme/base/ui.py
@@ -1670,8 +1670,8 @@ class AutomateSimulation(CFMENavigateStep):
 
 
 class AutomateImportExportBaseView(BaseLoggedInPage):
-    # TODO This is currently overiding the base flash and should be renamed and efforts made
-    # to update assocaited tests
+    # TODO: This is currently overiding the base flash and should be renamed and efforts made
+    # to update associated tests
     flash = FlashMessages('div.import-flash-message')
     title = Text('.//div[@id="main-content"]//h1')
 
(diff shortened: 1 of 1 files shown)


cfme/base/ui.py
```diff
@@ -1709,8 +1709,9 @@ class import_git(View):     # noqa
     reset_all = VersionPick({
         Version.lowest(): Image(
             './/img[starts-with(@alt, "Reset all components in the following domains:")]'),
-        '5.10': Button(title='Reset all components in the following domains:')
+        '5.10': Button(title='Reset all components in the following domains: RedHat, ManageIQ')
     })
+    reset_title = Text('//*[@id="main_div"]/div[3]/h3[4]')
```
sshveta: Can the locator be more specific ?

ganeshhubale: @sshveta 
I have already tried to make it more specific than existing but I did not find more specific ```xpath``` than this one. Please let me know if you have any solution around it.

sshveta: :+1:

izapolsk: @ganeshhubale, isn't there any better xpath ?  
What I don't like in it
1. it uses indexes. So, any slight change of html page may break it.
2. Why do you use full xpath ? As far as I see your view is based on other views. So, I assume you can use relative locator here.

izapolsk:
@ganeshhubale, @mshriver, are you sure this BZ has too be automated at all ?

ganeshhubale:
@izapolsk 
I agree; it is static. But I am automating because of the BZ around it. 
It would be better to have test case around it. 
Please let me know your views @izapolsk @mshriver whether we are automating it or not.

mshriver:
@ganeshhubale it does seem a bit frivolous to have a test case for this, when it looks more like a statement that should be in the `View.is_displayed` method. 

Is adding this title check to the page's `is_displayed` method not a sufficient validation that the title is correct now? Ultimately, this BZ is then tested by any CRUD test hitting this page.

ganeshhubale:
@mshriver 
I am agree with you. I checked the ```is_displayed``` of view - ```AutomateImportExportView``` does not have this statement check.
I will update this is_displayed and  execute test which uses that view under this PR.
Thanks @mshriver @izapolsk

dajoRH:
Lint report for commit b1947b0c7f2e8b65bc7a8430cb83e974d6325c7d:

`cfme/base/ui.py`:
- :bangbang: Line 1727:17: **E126** *continuation line over-indented for hanging indent*

**No commit flaws detected**.

*CFME QE Bot*

cfme/base/ui.py
```diff
@@ -1709,12 +1709,22 @@ class import_git(View):     # noqa
...
     })
+    reset_title = Text(
+        ".//div[contains(@class, 'import-or-export')]/h3[contains(text(), 'Reset all')]"
+    )
 
     @property
     def is_displayed(self):
-        return self.in_import_export and self.export_all.is_displayed
+        # TODO: Remove check if the fix for BZ 1686054 is back ported to 5.9z
+        if self.appliance.version < "5.10":
```
mshriver: PRT highlights the bug here - you're in a View class, which doesn't have an appliance. You can use `self.browser.product_version`

PR #8598, opened 2019-03-19
[1LP][RFR] New Test: Checking domain import via git with or without top level domain

ganeshhubale:
- Test case added for checking domain imports with or without top level domain
- Removed test cases which was already exist for importing domain
- Merged two test cases into one

{{ pytest: cfme/tests/automate/test_git_import.py -k 'test_domain_import_git' -vv }}

### DIFF
--- a/cfme/tests/automate/test_git_import.py
+++ b/cfme/tests/automate/test_git_import.py
@@ -110,3 +110,69 @@ def test_automate_git_import_multiple_domains(request, appliance):
         domain = repo.import_domain_from(branch="origin/master")
         request.addfinalizer(domain.delete_if_exists)
         assert not domain.exists
+
+
+@pytest.mark.tier(2)
+@pytest.mark.parametrize(
+    ("url", "param_type", "param_value", "verify_ssl"),
+    [
+        (
+            'https://github.com/ramrexx/CloudForms_Essentials.git',
+            'branch',
+            'origin/cf4.1',
+            True
+        ),
+        (
+            'https://github.com/RedHatQE/ManageIQ-automate-git.git',
+            'tag',
+            '0.1',
+            False
+        ),
+        (
+            "https://github.com/RedHatQE/ManageIQ-automate-git.git",
+            "branch",
+            "origin/master",
+            False,
+        ),
+        (
+            "https://github.com/ganeshhubale/ManageIQ-automate-git.git",
+            "branch",
+            "origin/test",
+            False,
+        ),
+    ],
+    ids=["with_branch", "with_tag", "with_top_level_domain", "without_top_level_domain"],
+)
+def test_domain_import_git(
+    request, appliance, url, param_type, param_value, verify_ssl
+):
+    """This test case Verifies that a domain can be imported from git and Importing domain from git
+       should work with or without the top level domain directory.
+
+    Polarion:
+        assignee: ghubale
+        initialEstimate: 1/6h
+        caseimportance: medium
+        caseposneg: positive
+        testtype: functional
+        startsin: 5.7
+        casecomponent: Automate
+        tags: automate
+        title: Test automate git domain import top level directory
+        testSteps:
+            1. Enable server role: git Repositories Owner
+            2. Navigate to Automation > Automate > Import/Export
+            3. Create a Git Repository with the contents of a domain directory without including
+               the domain directory.
+        expectedResults:
+            1.
+            2.
+            3. Import should work with or without the top level domain directory.
+
+    Bugzilla:
+        1389823
+    """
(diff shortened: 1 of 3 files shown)


cfme/tests/automate/test_git_import.py
```diff
@@ -78,3 +78,69 @@ def test_automate_git_domain_displayed_in_service(appliance, imported_domain, br
...
+            1.
+            2.
+            3. Import should work with or without the top level domain directory.
+
+    Bugzilla:
+        1389823
+    """
+    repo = AutomateGitRepository(url=url, verify_ssl=verify_ssl, appliance=appliance)
+    domain = repo.import_domain_from(**{param_type: param_value})
+    request.addfinalizer(domain.delete)
```
john-dupuy: Question: Why not use `delete_if_exists` here?

ganeshhubale: Thanks for this question @john-dupuy I was expecting this one.
- So if you see I am asserting domain for it's existence.(```assert domain.exists```). In case, If domain exists then test gets passed and ```delete``` will remove the domain.
- In case, If domain does not exists(**which means test gets failed**) then there is no need of teardown.
- Actually ```delete_if_exists``` is first checking the domain exists or not then it calls ```delete``` method.
- I hope you are agree with me :)

cfme/tests/automate/test_git_import.py
```diff
@@ -78,3 +78,69 @@ def test_automate_git_domain_displayed_in_service(appliance, imported_domain, br
...
+        (
+            "https://github.com/ganeshhubale/ManageIQ-automate-git.git",
+            "branch",
+            "origin/test",
+            False,
+        ),
+    ],
+    ids=["with_branch", "with_tag", "with_top_level_domain", "without_top_level_domain"],
+)
+def test_domain_import_git(
```
valaparthvi: Don't you need to enable `git_owner` server role here?

ganeshhubale: Already enabled globally.

dajoRH:
*CFME QE Bot*

cfme/tests/automate/test_git_import.py
```diff
@@ -78,3 +78,69 @@ def test_automate_git_domain_displayed_in_service(appliance, imported_domain, br
     view.modal.apply.click()
     assert view.provisioning_entry_point.value == ("/{}/Service/Provisioning/StateMachines/"
         "ServiceProvision_Template/CatalogItemInitialization".format(imported_domain.name))
+
+
+@pytest.mark.tier(2)
+@pytest.mark.parametrize(
+    ("url", "param_type", "param_value", "verify_ssl"),
```
john-dupuy: Optional: To avoid the lengthy in the marker of a test case, you could place this elsewhere in the file and write it as a `namedtuple`. See [here](https://github.com/ManageIQ/integration_tests/blob/86a6f2acd36508fecf57e12477be9d1bb16246ec/cfme/tests/control/test_basic.py#L94) for an example.

cfme/tests/automate/test_git_import.py
```diff
@@ -78,3 +78,69 @@ def test_automate_git_domain_displayed_in_service(appliance, imported_domain, br
...
+            False,
+        ),
+        (
+            "https://github.com/ganeshhubale/ManageIQ-automate-git.git",
+            "branch",
+            "origin/test",
+            False,
+        ),
+    ],
+    ids=["with_branch", "with_tag", "with_top_level_domain", "without_top_level_domain"],
```
john-dupuy: Nice ids :+1:

dajoRH:
*CFME QE Bot*

cfme/tests/automate/test_git_import.py
```diff
@@ -110,3 +110,69 @@ def test_automate_git_import_multiple_domains(request, appliance):
...
+        expectedResults:
+            1.
+            2.
+            3. Import should work with or without the top level domain directory.
+
+    Bugzilla:
+        1389823
+    """
+    repo = AutomateGitRepository(url=url, verify_ssl=verify_ssl, appliance=appliance)
+    domain = repo.import_domain_from(**{param_type: param_value})
```
jaryn: this is an equivalent to ` domain = repo.import_domain_from(param_type=param_value)` which is _the_ way how we usually call functions with keywords

ganeshhubale: I tried it but getting error:
```
--> 177     domain = repo.import_domain_from(**{param_type: param_value})
    178     request.addfinalizer(domain.delete)

ipdb> domain = repo.import_domain_from(param_type=param_value)
*** TypeError: import_domain_from() got an unexpected keyword argument 'param_type'
ipdb> 
```

dajoRH:
*CFME QE Bot*

PR #8599, opened 2019-03-19
[1LP][WIPTEST] Automate test control icons simulation

john-dupuy:
Purpose or Intent
=================
Automating `test_control_icons_simulation` and moving it into a new module. This requires some widgets to be added to the view for the Control->Simulation page, so I've done that. 

Depends on minor widgetastic.patternfly PRs [here](https://github.com/RedHatQE/widgetastic.patternfly/pull/85) and [here](https://github.com/RedHatQE/widgetastic.patternfly/pull/88) 

{{ pytest: --long-running --use-provider complete --use-template-cache cfme/tests/control/test_control_simulation.py }}

### DIFF
--- a/cfme/tests/control/test_control_simulation.py
+++ b/cfme/tests/control/test_control_simulation.py
@@ -17,7 +17,8 @@ FILL_DATA = {
     "event_type": "Datastore Operation",
     "event_value": "Datastore Analysis Complete",
     "filter_type": "By Clusters",
-    "filter_value": "Cluster"
+    "filter_value": "Cluster",
+    "submit_button": True
 }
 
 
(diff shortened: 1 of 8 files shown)


cfme/tests/control/test_control_simulation.py
```diff
@@ -0,0 +1,63 @@
...
+        expectedResults:
+            1.
+            2.
+            3.
+            4.
+            5. All the icons should be present
+    """
+    view = navigate_to(appliance.server, "ControlSimulation")
+    view.fill(FILL_DATA)
+    view.wait_displayed()
```
sshveta: Is this required ? You have implemented WaitFillViewStrategy that waits for filling each field.

cfme/tests/control/test_control_simulation.py
```diff
@@ -0,0 +1,63 @@
...
+    view.submit_button.click()
+    # Now check all the icons
+    assert view.simulation_results.squash_button.is_displayed
+    # Check the tree icons
+    tree = view.simulation_results.tree
+    # Check the root_item
+    assert tree.image_getter(tree.root_item) is not None
```
valaparthvi: Can you not simply do `assert tree.image_getter(tree.root_item)` ?

cfme/tests/control/test_control_simulation.py
john-dupuy: @sshveta, probably not required, I will remove this.

cfme/tests/control/test_control_simulation.py
john-dupuy: Good point! :smiley:

widgetastic_manageiq/__init__.py
```diff
@@ -1388,6 +1388,15 @@ def __locator__(self):
         )
 
 
+class SquashButton(Button):
+    """ This button requires a specific locator"""
+
+    def __locator__(self):
```
mshriver: I think this would be better implemented as `ROOT = ParametrizedLocator('.//div[contains(@class, "btn") and @id="squash_button" {@locator_conditions}]')

PR #8600, opened 2019-03-20
[1LP][RFR] New Test: Checking git import for multiple domains

ganeshhubale:
- Importing of multiple domains from a single git repository is not allowed

{{ pytest: cfme/tests/automate/test_git_import.py -k 'test_automate_git_import_multiple_domains' -vv }}

### DIFF
--- a/cfme/tests/automate/test_git_import.py
+++ b/cfme/tests/automate/test_git_import.py
@@ -79,13 +79,11 @@ def test_automate_git_domain_displayed_in_service(appliance, imported_domain, br
     assert view.provisioning_entry_point.value == ("/{}/Service/Provisioning/StateMachines/"
         "ServiceProvision_Template/CatalogItemInitialization".format(imported_domain.name))
 
-from cfme.base.ui import AutomateImportExportView, AutomateImportExportBaseView
-from cfme.automate.import_export import GitImportSelectorView
 
 @pytest.mark.tier(3)
-def test_automate_git_import_multiple_domains(appliance):
+def test_automate_git_import_multiple_domains(request, appliance):
     """
-    Import of multiple domains from a single git repo is not allowed
+    Importing of multiple domains from a single git repository is not allowed.
 
     Polarion:
         assignee: ghubale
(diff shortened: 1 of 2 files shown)


cfme/tests/automate/test_git_import.py
```diff
@@ -78,3 +78,41 @@ def test_automate_git_domain_displayed_in_service(appliance, imported_domain, br
     view.modal.apply.click()
     assert view.provisioning_entry_point.value == ("/{}/Service/Provisioning/StateMachines/"
         "ServiceProvision_Template/CatalogItemInitialization".format(imported_domain.name))
+
+
+@pytest.mark.tier(3)
+@pytest.mark.parametrize(
+    ("url", "param_type", "param_value", "verify_ssl"),
```
john-dupuy: Question: Why are you parametrizing this test against a single parameter? I think this would better suited to a fixture, unless you plan on adding more parameters in the future.

ganeshhubale: @john-dupuy 
This test case will not need more than existing parameters. Also I am actually dealing with only one git import url. As per my understanding, It would be better to have fixture; if I need to deal with multiple different urls with different parameters which  is not required in this test case. Hence keeping as it is.

cfme/tests/automate/test_git_import.py
```diff
@@ -78,3 +78,41 @@ def test_automate_git_domain_displayed_in_service(appliance, imported_domain, br
...
+    [("https://github.com/ganeshhubale/ManageIQ-automate-git", "branch", "origin/master", True)],
+    ids=["multiple_domains"],
+)
+def test_automate_git_import_multiple_domains(
```
valaparthvi: Do you need to enable `git_owner` server role here?

cfme/tests/automate/test_git_import.py
valaparthvi: I agree with @john-dupuy  here. Why use parametrization if there is only one group of parameters? We use parameters because we need to test a single test case with multiple scenarios, and I only see one scenario here. You could simply define those variables inside the test case, I don't think there will be any major impact on test case if you do so.

cfme/tests/automate/test_git_import.py
ganeshhubale: Already [enabled globally.](https://github.com/ManageIQ/integration_tests/pull/8600/files#diff-12d53d617972cfe42469893013d596d0R9)

valaparthvi: Oh, my bad. Didn't see it there.

PR #8602, opened 2019-03-20
[1LP][RFR] New Test: Managing tenant quota with RBAC for different users

ganeshhubale:
- Managing tenant quota with RBAC for different users

 {{ pytest: cfme/tests/cloud/test_tenant.py -k 'test_dynamic_product_feature_for_tenant_quota' -vv }}


cfme/tests/cloud/test_tenant.py
```diff
@@ -41,3 +61,159 @@ def test_tenant_crud(tenant):
...
+    collection = appliance.collections.tenants
+
+    # Here TENANT[0] is the first tenant(parent tenant) from the tenant's list
+    tenant = collection.create(
+        name="{}".format(TENANTS[0]),
+        description="tenant_des{}".format(fauxfactory.gen_alphanumeric()),
+        parent=collection.get_root_tenant(),
+    )
+    yield tenant
+    if tenant.exists:
```
valaparthvi: You can use `delete_if_exists` method here.

cfme/tests/cloud/test_tenant.py
```diff
@@ -41,3 +61,159 @@ def test_tenant_crud(tenant):
...
+            4.
+            5.
+            6.
+            7.
+            8.
+            9.
+            10. Logged in with users created
+            10a. login with alpha and SEE IF ALPHA USER CAN ABLE TO SET QUOTA OF omega_tenant
+                 FOR CPU - 5('Manage Quotas' option should be available)
+            11b. login with omega and SEE QUOTA GETS CHANGED OR NOT. THEN TRY TO CHANGE QUOTA
```
valaparthvi: There are no 11 and 12 `testSteps` here, are you missing those steps in `testSteps`?

cfme/tests/cloud/test_tenant.py
```diff
@@ -41,3 +61,159 @@ def test_tenant_crud(tenant):
...
+            12c. Here as per role_omega permissions, omega must not able change its own quota or
+                 other tenants quota.
+
+    Bugzilla:
+        1655012, 1468795
+    """
+    # Here TENANT[1] is the second tenant(child tenant) from the tenant's list
+    child_tenant = appliance.collections.tenants.create(
+        name="{}".format(TENANTS[1]),
+        description="tenant_des{}".format(fauxfactory.gen_alphanumeric()),
```
valaparthvi: I would prefer using the `child_tenant` name somewhere in the description instead of some randomly generated data. Totally optional though.

cfme/tests/cloud/test_tenant.py
```diff
@@ -41,3 +61,159 @@ def test_tenant_crud(tenant):
...
+        parent=new_tenant,
+    )
+    request.addfinalizer(child_tenant.delete)
+
+    role_ = []
+
+    # Instantiating existing role - 'EvmRole-tenant_administrator' to copy it to new role
+    role = appliance.collections.roles.instantiate(name='EvmRole-tenant_administrator')
+    for i in range(2):
```
valaparthvi: Why not use a single `for` loop for role, group, and user creation?

cfme/base/ui.py
```diff
@@ -392,9 +392,10 @@ class database(Accordion):  # noqa
     @property
     def in_configuration(self):
         return (
-            self.accordions.settings.is_displayed and
-            self.accordions.accesscontrol.is_displayed and
-            self.accordions.diagnostics.is_displayed and
+            self.navigation.currently_selected == [] and
+            self.accordions.settings.is_displayed or
```
john-dupuy: Question: Why were these `and` statements changed to `or`? Is only a single accordion now displayed?

cfme/configure/access_control/__init__.py
```diff
@@ -1007,7 +998,8 @@ class AllRolesView(ConfigurationView):
     def is_displayed(self):
         return (
             self.accordions.accesscontrol.is_opened and
-            self.title.text == 'Access Control Roles'
+            self.title.text == 'Access Control Roles' or
+            self.title.text == 'Role "{}"'.format(self.context['object'].name)
```
john-dupuy: Question: Why are you checking against a specific `Role` in the `AllRolesView`?

cfme/configure/access_control/__init__.py
```diff
@@ -1077,8 +1069,6 @@ def update(self, updates):
         # Instead of inserting a blind refresh, just disable this until the bug is resolved since
```
john-dupuy: Question: Is this comment still pertinent? If not, I suggest removing it.

cfme/tests/cloud/test_tenant.py
```diff
@@ -41,3 +61,154 @@ def test_tenant_crud(tenant):
...
+
+
+@pytest.fixture(scope="module")
+def new_tenant(appliance):
+    """This fixture creates new tenant under root tenant(My Company)"""
+    collection = appliance.collections.tenants
+
+    # Here TENANT[0] is the first tenant(parent tenant) from the tenant's list
+    tenant = collection.create(
+        name="{}".format(TENANTS[0]),
```
john-dupuy: No need to use `format` here, you can just do `name=TENANTS[0]`

cfme/tests/cloud/test_tenant.py
john-dupuy: @valaparthvi @ganeshhubale It's my understanding that the numbering in `testSteps` must match that of `expectedResults`. Am I missing something here?

cfme/tests/cloud/test_tenant.py
```diff
@@ -41,3 +61,154 @@ def test_tenant_crud(tenant):
...
+                 IMPOSED BY ALPHA USER.
+            12c. Here as per role_omega permissions, omega must not able change its own quota or
+                 other tenants quota.
+
+    Bugzilla:
+        1655012, 1468795
+    """
+    # Here TENANT[1] is the second tenant(child tenant) from the tenant's list
+    child_tenant = appliance.collections.tenants.create(
+        name="{}".format(TENANTS[1]),
```
john-dupuy: Same comment as above ^

cfme/tests/cloud/test_tenant.py
```diff
@@ -41,3 +61,154 @@ def test_tenant_crud(tenant):
...
+        user_.append(user)
+        request.addfinalizer(user.delete_if_exists)
+
+    # Updating first role for first user with PRODUCT_FEATURES tree. It restrict user from the
+    # tenants for which this role does not have access.
+    # Here PRODUCT_FEATURES[0] is the tree path navigation for updating RBAC role for tenant1
+    # (parent tenant) from TENANTS list.
+    # role_[0] is the first role from the list of two role created above; Providing 'False' while
+    # updating role means it restrict user from accessing that particular tenant.
+    role_[0].update({'product_features': [(PRODUCT_FEATURES[0], False)]})
```
john-dupuy: Question: Why not loop over roles and PRODUCT_FEATURES here?

cfme/tests/cloud/test_tenant.py
```diff
@@ -41,3 +61,154 @@ def test_tenant_crud(tenant):
...
+            11b. login with omega and SEE QUOTA GETS CHANGED OR NOT. THEN TRY TO CHANGE QUOTA
+                 IMPOSED BY ALPHA USER.
+            12c. Here as per role_omega permissions, omega must not able change its own quota or
+                 other tenants quota.
+
+    Bugzilla:
+        1655012, 1468795
+    """
+    # Here TENANT[1] is the second tenant(child tenant) from the tenant's list
+    child_tenant = appliance.collections.tenants.create(
```
john-dupuy: Question: Is the creation of all these objects part of the test? Or is it part of the setup for the test? If the latter, they should be fixtures.

cfme/tests/cloud/test_tenant.py
valaparthvi: Yep, they should match @john-dupuy.

cfme/tests/cloud/test_tenant.py
```diff
@@ -41,3 +61,154 @@ def test_tenant_crud(tenant):
...
+            description="group_{}".format(fauxfactory.gen_alphanumeric()),
+            role=new_role.name,
+            tenant=tenant_[i]
+        )
+        request.addfinalizer(group.delete_if_exists)
+
+        # Creating two different users which are assigned with different groups
+        user = appliance.collections.users.create(
+            name="user_{}".format(fauxfactory.gen_alphanumeric().lower()),
+            credential=Credential(principal='uid{}'.format(fauxfactory.gen_alphanumeric(4)),
```
valaparthvi: Can you check formatting here? It looks a little messed up.

cfme/tests/cloud/test_tenant.py
```diff
@@ -41,3 +61,154 @@ def test_tenant_crud(tenant):
...
+            2.
+            3.
+            4.
+            5.
+            6.
+            7.
+            8.
+            9.
+            10. Logged in with users created
+            10a. login with alpha and SEE IF ALPHA USER CAN ABLE TO SET QUOTA OF omega_tenant
```
valaparthvi: @ganeshhubale Any updates here? I don't see testSteps 11 and 12 here, yet they are present in expectedResults.

digitronik:
@ganeshhubale I found some valid questions please have a glance. For not moving to WIPTEST

ganeshhubale:
@digitronik  I am on it. I will update soon :)

cfme/base/ui.py
ganeshhubale: As per my observations:
- ```is_displayed``` of ```ConfigurationView``` should be simple because whenever we try test cases with RBAC roles and if that user does not have the permission to settings or access-control or diagnostics then **in_configuration** gets failed.
- I encountered with this problem while automating this test case with RBAC for quota.
- Hence updated ```in_configuration``` with compulsory check of ```self.navigation.currently_selected == []``` with any of the remaining

cfme/configure/access_control/__init__.py
ganeshhubale: Observations:
- If we created new role by navigating ```Roles > Click on any role > select copy this role to new role from configuration drop-down > update with new name and save``` (Here navigation goes to all page of roles - **self.title.text == 'Access Control Roles'**)
- Then go that ```new role > select Edit this role from configuration drop-down and update product features tree and save``` (Here navigation goes to details page of roles - **self.title.text == 'Role "{}"'.format(self.context['object'].name**)
- And if we tried to move to new role then it checks that if we are on ```AllRolesView```. Hence need to check ```is_diplayed``` for specific role.

cfme/base/ui.py
john-dupuy: That makes sense, so it's dependent on the permissions of the user. Barring having a user dependent `is_displayed`, which could be very messy, I think this is a good approach.

cfme/configure/access_control/__init__.py
john-dupuy: @ganeshhubale I believe this is a bug. I have seen something similar in condition creation. When you edit a role and hit save, you should be taken to the details page for that Role,  and the role name should be highlighted in the accordion.

What I'm seeing here is that when a role is edited and you hit save, the role name isn't highlighted in the accordion, which should be a BZ.

cfme/tests/cloud/test_tenant.py
```diff
@@ -41,3 +61,152 @@ def test_tenant_crud(tenant):
...
+    )
+    yield child_tenant
+    child_tenant.delete_if_exists()
+
+
+def check_permissions(appliance, assigned_tenant):
+    """This function is used to check user permissions for particular tenant"""
+    view = navigate_to(appliance.collections.tenants, 'All')
+    count = view.table.row_count
+    for i in range(count):
```
digitronik: I think you try like this
```
for t in view.table:
    if t["Name"].text == "test":
        t.click()
        break
```
then create Detail page view. and then assert. 

I think better solution is navigate to `Details` destination against `assigned_tenant`. I don't understand why tenant test are not using it?

ganeshhubale: - I am agree with you on navigations.
- But I got issue related to tenant tree with respect to RBAC roles.
- I have created issue for it: https://github.com/ManageIQ/integration_tests/issues/8662
- It may change other tests related to tenant.
- So I have used this approach for this test case and added todo in this PR.

cfme/tests/cloud/test_tenant.py
```diff
@@ -41,3 +61,151 @@ def test_tenant_crud(tenant):
...
+        parent=new_tenant,
+    )
+    yield child_tenant
+    child_tenant.delete_if_exists()
+
+
+def check_permissions(appliance, assigned_tenant):
+    """This function is used to check user permissions for particular tenant"""
+    view = navigate_to(appliance.collections.tenants, 'All')
+    # TODO: Need to restructure tenant entity by considering RBAC roles
```
digitronik: Please add issue number like `TODO(GH-123)`. it will help us to add changes.

dajoRH:
*CFME QE Bot*

dajoRH:
*CFME QE Bot*

PR #8607, opened 2019-03-21
[RFR] Update quickstart, OS_NAME and OS_VERSION, make debuginfo optional

mshriver:
Add cli option to quickstart, making debuginfo an optional install
Update OS_NAME and OS_VERSION regex to be a bit more flexible, allowing for quotes around names/versions
Use VERSION_ID for OS_VERSION, since 'VERSION' field in os-release can contain things like 29 (workstation edition) instead of just 29

Tested with RHEL and Fedora os-release files

{{ pytest: cfme/scripting/tests/test_quickstart.py }}

### DIFF
--- a/cfme/scripting/quickstart/system.py
+++ b/cfme/scripting/quickstart/system.py
@@ -14,10 +14,8 @@ HAS_YUM = os.path.exists('/usr/bin/yum')
 HAS_APT = os.path.exists('/usr/bin/apt-get')
 
 OS_RELEASE_FILE = '/etc/os-release'
-OS_NAME_REGEX = r'NAME="([\w\W]+)"'
-OS_VERSION_REGEX = r'VERSION="([\w\W]+)"'
-OS_NAME = None
-OS_VERSION = None
+OS_NAME_REGEX = r'^NAME="?([\w\s]*?)"?$'
+OS_VERSION_REGEX = r'^VERSION_ID="?([\S]*?)"?$'
 
 REQUIRED_PACKAGES = None
 INSTALL_COMMAND = None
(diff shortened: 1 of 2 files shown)


cfme/scripting/quickstart/system.py
```diff
@@ -50,10 +48,12 @@
         for var in os_release.readlines():
             name = re.match(OS_NAME_REGEX, var)
             if name:
-                OS_NAME = name.group(1)
+                OS_NAME = name.group(1).rstrip('\n')
```
mshriver: rstrip just in case

dajoRH:
Lint report for commit 32ac7d912584e4f9d6092843697b7a851ef23167:

`cfme/scripting/quickstart/system.py`:
- :bangbang: Line 159:28: **E231** *missing whitespace after ','*

**No commit flaws detected**.

*CFME QE Bot*

cfme/scripting/quickstart/system.py
```diff
@@ -149,13 +149,16 @@ def retry_install_once(command, shell):
...
-    if DEBUG_INSTALL_COMMAND and DEBUG_PACKAGES:
-        retry_install_once('{} {}'.format(DEBUG_INSTALL_COMMAND, ' '.join(DEBUG_PACKAGES)),
-                        shell=True)
-    else:
+        installed = True
+    if DEBUG_INSTALL_COMMAND and DEBUG_PACKAGES and debuginfo_install:
+        retry_install_once('{} {}'.format(DEBUG_INSTALL_COMMAND, ' '.join(DEBUG_PACKAGES))
+                           ,shell=True)
+        installed = True
+    if not installed:
```
mshriver: The way the logic works here, this was only displaying if DEBUG_INSTALL_COMMAND and DEBUG_PACKAGES weren't set, and making debuginfo_install optional meant it was displayed every time. Original logic was not great anyway, with it not triggering on `INSTALL_COMMAND and REQUIRED_PACKAGES`

cfme/scripting/quickstart/system.py
```diff
@@ -149,13 +149,16 @@ def retry_install_once(command, shell):
...
 
 
-def install_system_packages():
+def install_system_packages(debuginfo_install=False):
+    installed = False
     if INSTALL_COMMAND and REQUIRED_PACKAGES:
         retry_install_once(INSTALL_COMMAND + REQUIRED_PACKAGES, shell=True)
-    if DEBUG_INSTALL_COMMAND and DEBUG_PACKAGES:
+        installed = True
+    if DEBUG_INSTALL_COMMAND and DEBUG_PACKAGES and debuginfo_install:
```
izapolsk: I believe we need to pass the buck to ansible in such scripts. we reinvent the wheel

mshriver: but we make the _best_wheels! :)  I don't disagree at all, the system package mgmt is the biggest headache of quickstart overall. Maybe we keep the components that are determining OS and install ansible, and then run the playbook for the given OS, instead of trying to build bash commands to run.

PR #8612, opened 2019-03-22
[1LP][RFR] Uncollect RHV hourly disk graph tests

nachandr:
The hourly graph tests for disk metrics for RHV should be uncollected since RHV doesn't collect disk metrics for hosts.

### DIFF
--- a/cfme/tests/candu/test_host_graph.py
+++ b/cfme/tests/candu/test_host_graph.py
@@ -107,7 +107,7 @@ def test_host_most_recent_hour_graph_screen(graph_type, provider, host, enable_c
     assert graph_data > 0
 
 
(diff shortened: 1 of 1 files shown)


cfme/tests/candu/test_host_graph.py
```diff
@@ -107,9 +107,9 @@ def refresh():
     assert graph_data > 0
 
 
-@pytest.mark.uncollectif(lambda provider, interval:
+@pytest.mark.uncollectif(lambda provider, graph_type, interval:
```
digitronik: Required:
As you are removing `interval` check please remove this as well. Please add run for test.

nachandr: Done

PR #8618, opened 2019-03-25
[WIPTEST] Fix test_action_start_virtual_machine_after_stopping

john-dupuy:
{{ pytest: --long-running --use-template-cache --use-provider rhos13 cfme/tests/control/test_actions.py::test_action_start_virtual_machine_after_stopping }}

### DIFF
--- a/cfme/tests/control/test_actions.py
+++ b/cfme/tests/control/test_actions.py
@@ -293,6 +293,7 @@ def test_action_start_virtual_machine_after_stopping(request, vm, vm_on, policy_
     request.addfinalizer(policy_for_testing.assign_events)
     # Stop the VM
     vm.mgmt.stop()
+    vm._prov.refresh_provider_relationships()
     # Wait for VM powered on by CFME
     try:
         vm.mgmt.wait_for_state(VmState.RUNNING, timeout=600, delay=5)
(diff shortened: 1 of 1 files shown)


john-dupuy:
Failing 5.9, but I will leave this RFR since we have branched 5.9.

lina-nikiforova:
@john-dupuy Q: why do we need to refresh relationships to see that VM has a new status? Is normal refresh not enough?

john-dupuy:
@lina-nikiforova I noticed that sometimes, especially for the rhos13, the new status isn't recognized without doing the provider refresh. This was causing the test to fail.

jawatts:
@john-dupuy I would check with dev here, I am wondering if a VM changing state should cause a provider refresh automatically. In Openshift, if a new pod is create, an automatic refresh is triggered.

john-dupuy:
Thanks @lina-nikiforova and @jawatts I am testing this now to see if it is a BZ, RHOS13 events do not seem to autorefresh.

john-dupuy:
After reaching out to devs and @mnadeem92, this appears to be a known issue. @Yadnyawalkya wrote a BZ against this https://bugzilla.redhat.com/show_bug.cgi?id=1692892 that highlights the same issue (but related to v2v). This should be fixed by: https://github.com/ManageIQ/manageiq-providers-openstack/pull/457

Closing this PR

aufi:
@john-dupuy The issue fixed in Openstack provider PR #457, haven't produced any failures in evm.log. It just sent queries to OpenStack with wrong time filters (which resulted in not getting events in response).

PR #8620, opened 2019-03-25
[1LP][RFR] Fix test_provisioning.

jaryn:
This patch requires https://github.com/ManageIQ/wrapanapi/pull/370

Remove a `pytest.raises` as it seems to me that is shouldn't be there. It causes:
```
request = <SubRequest 'modified_request_class' for <Function 'test_cloud_provision_from_template_with_attached_disks[openstack-13-1]'>>
domain = <cfme.automate.explorer.domain.Domain object at 0x7f93cade6fd0>
original_request_class = <cfme.automate.explorer.klass.Class object at 0x7f93cb1fdb10>

    def modified_request_class(request, domain, original_request_class):
        with pytest.raises(Exception, match="error: Error during 'Automate Class copy'"):
            # methods of this class might have been copied by other fixture, so this error can occur
>           original_request_class.copy_to(domain)
E           Failed: DID NOT RAISE <type 'exceptions.Exception'>

cfme/tests/cloud_infra_common/test_provisioning.py:341: Failed
```

Fix VM from volume provisioning and register finalizers to do proper cleanup in correct order after.

Fix `test_cloud_provision_from_template_with_attached_disks` by switching to the `block_device_mapping_v2` as the `block_device_mapping` didn't work. Also, set the first disk as bootable so nova won't complain about attaching non-bootable disk.

{{ py.test: --use-provider rhos13 -v cfme/tests/cloud_infra_common/test_provisioning.py }}

### DIFF
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -581,7 +581,8 @@ class BaseVMCollection(BaseCollection):
                 if provision_request.is_succeeded(method='ui'):
                     logger.info('Waiting for vm %s to appear on provider %s', vm.name,
                                 provider.key)
-                    wait_for(provider.mgmt.does_vm_exist, [vm.name], num_sec=600)
+                    wait_for(provider.mgmt.does_vm_exist, [vm.name],
+                             handle_exception=True, num_sec=600)
                 elif override:
                     logger.info('Overriding exception to check failure condition.')
                 else:
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -581,8 +581,7 @@ class BaseVMCollection(BaseCollection):
                 if provision_request.is_succeeded(method='ui'):
                     logger.info('Waiting for vm %s to appear on provider %s', vm.name,
                                 provider.key)
-                    wait_for(provider.mgmt.does_vm_exist, [vm.name],
-                             handle_exception=True, num_sec=600)
+                    wait_for(provider.mgmt.does_vm_exist, [vm.name], num_sec=600)
                 elif override:
                     logger.info('Overriding exception to check failure condition.')
                 else:
--- a/cfme/tests/cloud_infra_common/test_provisioning.py
+++ b/cfme/tests/cloud_infra_common/test_provisioning.py
@@ -367,6 +367,7 @@ def copy_domains(original_request_class, domain):
 
 
 # Not collected for EC2 in generate_tests above
+@pytest.mark.meta(blockers=[BZ(1713632)])
 @pytest.mark.parametrize("disks", [1, 2])
 @pytest.mark.provider([OpenStackProvider], required_fields=[['provisioning', 'image']],
                       override=True)
--- a/cfme/tests/cloud_infra_common/test_provisioning.py
+++ b/cfme/tests/cloud_infra_common/test_provisioning.py
@@ -3,7 +3,6 @@
 # in selenium (the group is selected then immediately reset)
 from textwrap import dedent
 
-import fauxfactory
 import pytest
 from riggerlib import recursive_update
 from widgetastic_patternfly import CheckableBootstrapTreeview as Check_tree
@@ -49,8 +48,9 @@ def vm_name():
 def instance_args(request, provider, provisioning, vm_name):
     """ Fixture to prepare instance parameters for provisioning
     """
-    inst_args = dict(template_name=provisioning.get('image', {}).get('image') or provisioning.get(
+    inst_args = dict(template_name=provisioning.get('image', {}).get('name') or provisioning.get(
         'template'))
+    assert inst_args.get('template_name')
 
     # Base instance info
     inst_args['request'] = {
@@ -336,20 +336,26 @@ def test_provision_from_template_using_rest(appliance, request, provider, vm_nam
 
 
 VOLUME_METHOD = ("""
+clone_options = {{
+  :image_ref => nil,
+  :block_device_mapping_v2 => [
+    {}
+  ]
+}}
+
 prov = $evm.root["miq_provision"]
-prov.set_option(
-    :clone_options,
-    {{ :block_device_mapping => [{}] }})
+prov.set_option(:clone_options, clone_options)
 """)
 
-ONE_FIELD = """{{:volume_id => "{}", :device_name => "{}"}}"""
-
-
-@pytest.fixture(scope="module")
-def domain(request, appliance):
-    domain = appliance.collections.domains.create(name=fauxfactory.gen_alphanumeric(), enabled=True)
-    request.addfinalizer(domain.delete_if_exists)
-    return domain
+ONE_FIELD = """{{
+  :boot_index => {boot_index},
+  :uuid => "{uuid}",
+  :device_name => "{device_name}",
+  :source_type => "volume",
+  :destination_type => "volume",
+  :volume_size => 1,
+  :delete_on_termination => false
+}}"""
 
 
 @pytest.fixture(scope="module")
@@ -405,37 +411,59 @@ def test_cloud_provision_from_template_with_attached_disks(
     if provider.one_of(AzureProvider):
         recursive_update(inst_args, {'environment': {'availability_zone': provisioning("av_set")}})
 
-    device_name = "/dev/sd{}"
+    device_name = "vd{}"
     device_mapping = []
 
-    with provider.mgmt.with_volumes(1, n=disks) as volumes:
-        for i, volume in enumerate(volumes):
-            device_mapping.append((volume, device_name.format(chr(ord("b") + i))))
-        # Set up automate
+    volumes = provider.mgmt.volume_configurations(1, n=disks)
 
-        method = modified_request_class.methods.instantiate(name="openstack_PreProvision")
+    @request.addfinalizer
+    def delete_volumes():
+        for volume in volumes:
+            provider.mgmt.delete_volume(volume)
 
+    # Set up automate
+    for i, volume in enumerate(volumes, 0):
+        if i == 0:
+            provider.mgmt.capi.volumes.set_bootable(volume, True)
+            device_mapping.append(
+                {'boot_index': 0,
+                'uuid': volume,
+                'device_name': device_name.format(chr(ord("a") + i))})
+        else:
+            device_mapping.append(
+                {'boot_index': -1,
+                'uuid': volume,
+                'device_name': device_name.format(chr(ord("a") + i))})
+
+    method = modified_request_class.methods.instantiate(name="openstack_PreProvision")
+
+    view = navigate_to(method, 'Details')
+    former_method_script = view.script.get_value()
+    with update(method):
+        disk_mapping = []
+        for mapping in device_mapping:
+            disk_mapping.append(ONE_FIELD.format(**mapping))
+        method.script = VOLUME_METHOD.format(",\n".join(disk_mapping))
+
+    @request.addfinalizer
+    def _finish_method():
         with update(method):
-            disk_mapping = []
-            for mapping in device_mapping:
-                disk_mapping.append(ONE_FIELD.format(*mapping))
-            method.script = VOLUME_METHOD.format(", ".join(disk_mapping))
+            method.script = former_method_script
 
-        def _finish_method():
-            with update(method):
-                method.script = """prov = $evm.root["miq_provision"]"""
-        request.addfinalizer(_finish_method)
+    instance = appliance.collections.cloud_instances.create(vm_name,
+                                                            provider,
+                                                            form_values=inst_args)
 
-        instance = appliance.collections.cloud_instances.create(vm_name,
-                                                                provider,
-                                                                form_values=inst_args)
-
-        for volume_id in volumes:
-            soft_assert(vm_name in provider.mgmt.volume_attachments(volume_id))
-        for volume, device in device_mapping:
-            soft_assert(provider.mgmt.volume_attachments(volume)[vm_name] == device)
-        instance.mgmt.delete()  # To make it possible to delete the volume
+    @request.addfinalizer
+    def delete_vm_and_wait_for_gone():
+        instance.mgmt.delete()
         wait_for(lambda: not instance.exists_on_provider, num_sec=180, delay=5)
+
+    for volume_id in volumes:
+        assert vm_name in provider.mgmt.volume_attachments(volume_id)
+    for device in device_mapping:
+        assert (provider.mgmt.volume_attachments(device['uuid'])[vm_name]
+                == '/dev/{}'.format(device['device_name']))
 
 
 # Not collected for EC2 in generate_tests above
@@ -458,52 +486,54 @@ def test_provision_with_boot_volume(request, instance_args, provider, soft_asser
 
     image = inst_args.get('template_name')
 
-    with provider.mgmt.with_volume(1, imageRef=provider.mgmt.get_template_id(image)) as volume:
-        # Set up automate
-        method = modified_request_class.methods.instantiate(name="openstack_CustomizeRequest")
+    volume = provider.mgmt.create_volume(1, imageRef=provider.mgmt.get_template(image).uuid)
+    request.addfinalizer(lambda: provider.mgmt.delete_volume(volume))
+
+    # Set up automate
+    method = modified_request_class.methods.instantiate(name="openstack_CustomizeRequest")
+    view = navigate_to(method, 'Details')
+    former_method_script = view.script.get_value()
+    with update(method):
+        method.script = dedent('''\
+            $evm.root["miq_provision"].set_option(
+                :clone_options, {{
+                    :image_ref => nil,
+                    :block_device_mapping_v2 => [{{
+                        :boot_index => 0,
+                        :uuid => "{}",
+                        :device_name => "vda",
+                        :source_type => "volume",
+                        :destination_type => "volume",
+                        :volume_size => 1,
+                        :delete_on_termination => false
+                    }}]
+                }}
+            )
+        '''.format(volume))
+
+    @request.addfinalizer
+    def _finish_method():
         with update(method):
-            method.script = dedent('''\
-                $evm.root["miq_provision"].set_option(
-                    :clone_options, {{
-                        :image_ref => nil,
-                        :block_device_mapping_v2 => [{{
-                            :boot_index => 0,
-                            :uuid => "{}",
-                            :device_name => "vda",
-                            :source_type => "volume",
-                            :destination_type => "volume",
-                            :volume_size => 1,
-                            :delete_on_termination => false
-                        }}]
-                    }}
-                )
-            '''.format(volume))
+            method.script = former_method_script
 
-        @request.addfinalizer
-        def _finish_method():
-            with update(method):
-                method.script = """prov = $evm.root["miq_provision"]"""
+    instance = appliance.collections.cloud_instances.create(vm_name,
+                                                            provider,
+                                                            form_values=inst_args)
 
-        instance = appliance.collections.cloud_instances.create(vm_name,
-                                                                provider,
-                                                                form_values=inst_args)
-
-        request_description = 'Provision from [{}] to [{}]'.format(image,
-                                                                   instance.name)
-        provision_request = appliance.collections.requests.instantiate(request_description)
-        try:
-            provision_request.wait_for_request(method='ui')
-        except Exception as e:
-            logger.info(
-                "Provision failed {}: {}".format(e, provision_request.request_state))
-            raise
-        msg = "Provisioning failed with the message {}".format(
-            provision_request.row.last_message.text)
-        assert provision_request.is_succeeded(method='ui'), msg
-        soft_assert(instance.name in provider.mgmt.volume_attachments(volume))
-        soft_assert(provider.mgmt.volume_attachments(volume)[instance.name] == "/dev/vda")
+    @request.addfinalizer
+    def delete_vm_and_wait_for_gone():
         instance.mgmt.delete()  # To make it possible to delete the volume
         wait_for(lambda: not instance.exists_on_provider, num_sec=180, delay=5)
+
+    request_description = 'Provision from [{}] to [{}]'.format(image, instance.name)
+    provision_request = appliance.collections.requests.instantiate(request_description)
+    provision_request.wait_for_request(method='ui')
+
+    msg = "Provisioning failed with the message {}".format(
+        provision_request.row.last_message.text)
+    assert provision_request.is_succeeded(method='ui'), msg
+    soft_assert(instance.name in provider.mgmt.volume_attachments(volume))
+    soft_assert(provider.mgmt.volume_attachments(volume)[instance.name] == "/dev/vda")
 
 
 # Not collected for EC2 in generate_tests above
@@ -532,6 +562,9 @@ def test_provision_with_additional_volume(request, instance_args, provider, smal
         image_id = provider.mgmt.get_template(small_template.name).uuid
     except KeyError:
         pytest.skip("No small_template in provider data!")
+
+    view = navigate_to(method, 'Details')
+    former_method_script = view.script.get_value()
     with update(method):
         method.script = dedent('''\
             $evm.root["miq_provision"].set_option(
@@ -550,14 +583,26 @@ def test_provision_with_additional_volume(request, instance_args, provider, smal
         )
         '''.format(image_id))
 
+    @request.addfinalizer
     def _finish_method():
         with update(method):
-            method.script = """prov = $evm.root["miq_provision"]"""
-    request.addfinalizer(_finish_method)
+            method.script = former_method_script
 
-    instance = appliance.collections.cloud_instances.create(vm_name,
-                                                            provider,
-                                                            form_values=inst_args)
+    def cleanup_and_wait_for_instance_gone():
+        instance.mgmt.refresh()
+        prov_instance_raw = instance.mgmt.raw
+        instance_volumes = getattr(prov_instance_raw, 'os-extended-volumes:volumes_attached')
+
+        instance.cleanup_on_provider()
+        wait_for(lambda: not instance.exists_on_provider, num_sec=180, delay=5)
+
+        # Delete the volumes.
+        for volume in instance_volumes:
+            provider.mgmt.delete_volume(volume['id'])
+
+    instance = appliance.collections.cloud_instances.create(
+        vm_name, provider, form_values=inst_args)
+    request.addfinalizer(cleanup_and_wait_for_instance_gone)
 
     request_description = 'Provision from [{}] to [{}]'.format(small_template.name, instance.name)
     provision_request = appliance.collections.requests.instantiate(request_description)
@@ -573,20 +618,14 @@ def test_provision_with_additional_volume(request, instance_args, provider, smal
 
     instance.mgmt.refresh()
     prov_instance_raw = instance.mgmt.raw
-    try:
-        assert hasattr(prov_instance_raw, 'os-extended-volumes:volumes_attached')
-        volumes_attached = getattr(prov_instance_raw, 'os-extended-volumes:volumes_attached')
-        assert len(volumes_attached) == 1
-        volume_id = volumes_attached[0]["id"]
-        assert provider.mgmt.volume_exists(volume_id)
-        volume = provider.mgmt.get_volume(volume_id)
-        assert volume.size == 3
-    finally:
-        instance.cleanup_on_provider()
-        wait_for(lambda: not instance.exists_on_provider, num_sec=180, delay=5)
-        if "volume_id" in locals():  # To handle the case of 1st or 2nd assert
-            if provider.mgmt.volume_exists(volume_id):
-                provider.mgmt.delete_volume(volume_id)
+
+    assert hasattr(prov_instance_raw, 'os-extended-volumes:volumes_attached')
+    volumes_attached = getattr(prov_instance_raw, 'os-extended-volumes:volumes_attached')
+    assert len(volumes_attached) == 1
+    volume_id = volumes_attached[0]["id"]
+    assert provider.mgmt.volume_exists(volume_id)
+    volume = provider.mgmt.get_volume(volume_id)
+    assert volume.size == 3
 
 
 @test_requirements.tag
--- a/cfme/tests/cloud_infra_common/test_provisioning.py
+++ b/cfme/tests/cloud_infra_common/test_provisioning.py
@@ -50,7 +50,8 @@ def instance_args(request, provider, provisioning, vm_name):
     """
     inst_args = dict(template_name=provisioning.get('image', {}).get('name') or provisioning.get(
         'template'))
-    assert inst_args.get('template_name')
+    if not inst_args.get('template_name'):
+        pytest.skip(reason='template name not specified in the provisioning in config')
 
     # Base instance info
     inst_args['request'] = {
@@ -335,29 +336,6 @@ def test_provision_from_template_using_rest(appliance, request, provider, vm_nam
                            .format(ex.message))
 
 
-VOLUME_METHOD = ("""
-clone_options = {{
-  :image_ref => nil,
-  :block_device_mapping_v2 => [
-    {}
-  ]
-}}
-
-prov = $evm.root["miq_provision"]
-prov.set_option(:clone_options, clone_options)
-""")
-
-ONE_FIELD = """{{
-  :boot_index => {boot_index},
-  :uuid => "{uuid}",
-  :device_name => "{device_name}",
-  :source_type => "volume",
-  :destination_type => "volume",
-  :volume_size => 1,
-  :delete_on_termination => false
-}}"""
-
-
 @pytest.fixture(scope="module")
 def original_request_class(appliance):
     return (appliance.collections.domains.instantiate(name='ManageIQ')
@@ -442,8 +420,29 @@ def test_cloud_provision_from_template_with_attached_disks(
     with update(method):
         disk_mapping = []
         for mapping in device_mapping:
-            disk_mapping.append(ONE_FIELD.format(**mapping))
-        method.script = VOLUME_METHOD.format(",\n".join(disk_mapping))
+            one_field = dedent("""{{
+                :boot_index => {boot_index},
+                :uuid => "{uuid}",
+                :device_name => "{device_name}",
+                :source_type => "volume",
+                :destination_type => "volume",
+                :volume_size => 1,
+                :delete_on_termination => false
+            }}""")
+            disk_mapping.append(one_field.format(**mapping))
+
+        volume_method = dedent("""
+            clone_options = {{
+            :image_ref => nil,
+            :block_device_mapping_v2 => [
+                {}
+            ]
+            }}
+
+            prov = $evm.root["miq_provision"]
+            prov.set_option(:clone_options, clone_options)
+        """)
+        method.script = volume_method.format(",\n".join(disk_mapping))
 
     @request.addfinalizer
     def _finish_method():
--- a/cfme/tests/cloud_infra_common/test_provisioning.py
+++ b/cfme/tests/cloud_infra_common/test_provisioning.py
@@ -402,17 +402,13 @@ def test_cloud_provision_from_template_with_attached_disks(
 
     # Set up automate
     for i, volume in enumerate(volumes, 0):
+        device_mapping.append(
+            {'boot_index': 0 if i == 0 else -1,
+            'uuid': volume,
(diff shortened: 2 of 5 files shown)


dajoRH:
*CFME QE Bot*

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -47,8 +47,9 @@ def vm_name():
 def instance_args(request, provider, provisioning, vm_name):
     """ Fixture to prepare instance parameters for provisioning
     """
-    inst_args = dict(template_name=provisioning.get('image', {}).get('image') or provisioning.get(
+    inst_args = dict(template_name=provisioning.get('image', {}).get('name') or provisioning.get(
         'template'))
+    assert inst_args.get('template_name')
```
digitronik: @jaryn I think if `None` will come it will assert error. As per me its setup where we are taking `template` name from `provisioning` l.e. provider data.
I think better to `skip` if what you think?

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -336,9 +350,7 @@ def original_request_class(appliance):
 
 @pytest.fixture(scope="module")
 def modified_request_class(request, domain, original_request_class):
-    with pytest.raises(Exception, match="error: Error during 'Automate Class copy'"):
-        # methods of this class might have been copied by other fixture, so this error can occur
```
digitronik: ^^ are you sure about this?

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -378,38 +390,57 @@ def test_cloud_provision_from_template_with_attached_disks(
...
+    volumes = provider.mgmt.volume_configurations(1, n=disks)
+
+    @request.addfinalizer
+    def delete_volumes():
+        for volume in volumes:
+            provider.mgmt.delete_volume(volume)
 
-        method = modified_request_class.methods.instantiate(name="openstack_PreProvision")
+    # Set up automate
+    for i, volume in enumerate(volumes, 0):
```
digitronik: Note:
default `start` index always `0`

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -378,38 +390,57 @@ def test_cloud_provision_from_template_with_attached_disks(
...
+        if i == 0:
+            provider.mgmt.capi.volumes.set_bootable(volume, True)
+            device_mapping.append(
+                {'boot_index': 0,
+                'uuid': volume,
+                'device_name': device_name.format(chr(ord("a") + i))})
+        else:
+            device_mapping.append(
+                {'boot_index': -1,
```
digitronik: Q:
 why always `-1` index for volume other than bootable? Just suspect about this.

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -336,9 +350,7 @@ def original_request_class(appliance):
 
 @pytest.fixture(scope="module")
 def modified_request_class(request, domain, original_request_class):
-    with pytest.raises(Exception, match="error: Error during 'Automate Class copy'"):
-        # methods of this class might have been copied by other fixture, so this error can occur
-        original_request_class.copy_to(domain)
+    original_request_class.copy_to(domain)
     klass = (domain
```
ganeshhubale: @jaryn 
- We have [domain](https://github.com/ManageIQ/integration_tests/blob/master/cfme/fixtures/automate.py#L6) as global fixture.
- So you can [remove this local fixture of ```domain``` form this file](https://github.com/ManageIQ/integration_tests/pull/8620/files#diff-fc91ed43845421f3eda7a5bb17217dc7R335) and use global one.

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -430,53 +461,58 @@ def test_provision_with_boot_volume(request, instance_args, provider, soft_asser
...
 
+    request_description = 'Provision from [{}] to [{}]'.format(image, instance.name)
+    provision_request = appliance.collections.requests.instantiate(request_description)
+    try:
+        provision_request.wait_for_request(method='ui')
+    except Exception as e:
+        logger.info(
+            "Provision failed {}: {}".format(e, provision_request.request_state))
+        raise
+
```
ganeshhubale: You can think of **key-value** pair for formatting.

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -430,53 +461,58 @@ def test_provision_with_boot_volume(request, instance_args, provider, soft_asser
...
+    instance = appliance.collections.cloud_instances.create(vm_name,
+                                                            provider,
+                                                            form_values=inst_args)
+
+    @request.addfinalizer
+    def delete_vm_and_wait_for_gone():
         instance.mgmt.delete()  # To make it possible to delete the volume
         wait_for(lambda: not instance.exists_on_provider, num_sec=180, delay=5)
 
+    request_description = 'Provision from [{}] to [{}]'.format(image, instance.name)
```
digitronik: optional:
I would use keying here.

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -430,53 +461,58 @@ def test_provision_with_boot_volume(request, instance_args, provider, soft_asser
...
+    provision_request = appliance.collections.requests.instantiate(request_description)
+    try:
+        provision_request.wait_for_request(method='ui')
+    except Exception as e:
+        logger.info(
+            "Provision failed {}: {}".format(e, provision_request.request_state))
```
digitronik: optional:
while formatting `logger` string. I would use `%s`

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -430,53 +461,58 @@ def test_provision_with_boot_volume(request, instance_args, provider, soft_asser
...
+        raise
+
+    msg = "Provisioning failed with the message {}".format(
+        provision_request.row.last_message.text)
+    assert provision_request.is_succeeded(method='ui'), msg
```
digitronik: :100: thanks

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -308,13 +309,26 @@ def test_provision_from_template_using_rest(appliance, request, provider, vm_nam
 
 
 VOLUME_METHOD = ("""
+clone_options = {{
```
john-dupuy: Optional: I think this would be more readable if it was indented like `ONE_FIELD` below. 

Question: Why are these not at the top of the file?

cfme/common/vm.py
```diff
@@ -577,8 +577,7 @@ def create(self, vm_name, provider, form_values=None, cancel=False, check_existi
                 if provision_request.is_succeeded(method='ui'):
                     logger.info('Waiting for vm %s to appear on provider %s', vm.name,
                                 provider.key)
-                    wait_for(provider.mgmt.does_vm_exist, [vm.name],
-                             handle_exception=True, num_sec=600)
+                    wait_for(provider.mgmt.does_vm_exist, [vm.name], num_sec=600)
```
john-dupuy: Question: Why are you removing `handle_exception`?

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -378,38 +390,57 @@ def test_cloud_provision_from_template_with_attached_disks(
...
 
-        method = modified_request_class.methods.instantiate(name="openstack_PreProvision")
+    # Set up automate
+    for i, volume in enumerate(volumes, 0):
+        if i == 0:
+            provider.mgmt.capi.volumes.set_bootable(volume, True)
+            device_mapping.append(
+                {'boot_index': 0,
+                'uuid': volume,
+                'device_name': device_name.format(chr(ord("a") + i))})
```
john-dupuy: Question: I'm curious about why `chr` and `ord` are necessary here? I've never seen them used before.

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -378,38 +390,57 @@ def test_cloud_provision_from_template_with_attached_disks(
...
+        else:
+            device_mapping.append(
+                {'boot_index': -1,
+                'uuid': volume,
+                'device_name': device_name.format(chr(ord("a") + i))})
 
+    method = modified_request_class.methods.instantiate(name="openstack_PreProvision")
```
john-dupuy: Q: Here is it okay to use the hardcoded string? `"openstack_PreProvision"`?

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -378,38 +390,57 @@ def test_cloud_provision_from_template_with_attached_disks(
...
-        method = modified_request_class.methods.instantiate(name="openstack_PreProvision")
+    # Set up automate
+    for i, volume in enumerate(volumes, 0):
+        if i == 0:
+            provider.mgmt.capi.volumes.set_bootable(volume, True)
+            device_mapping.append(
+                {'boot_index': 0,
+                'uuid': volume,
+                'device_name': device_name.format(chr(ord("a") + i))})
+        else:
```
john-dupuy: Question: Why are the other volumes not set to `bootable`?

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -378,38 +390,57 @@ def test_cloud_provision_from_template_with_attached_disks(
...
+            device_mapping.append(
+                {'boot_index': -1,
+                'uuid': volume,
+                'device_name': device_name.format(chr(ord("a") + i))})
 
+    method = modified_request_class.methods.instantiate(name="openstack_PreProvision")
+    with update(method):
+        disk_mapping = []
+        for mapping in device_mapping:
+            disk_mapping.append(ONE_FIELD.format(**mapping))
```
john-dupuy: Optional: This could be reduce LOC by making use of a list comprehension.

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -378,38 +390,57 @@ def test_cloud_provision_from_template_with_attached_disks(
...
-        instance.mgmt.delete()  # To make it possible to delete the volume
+            method.script = """prov = $evm.root["miq_provision"]"""
+
+    instance = appliance.collections.cloud_instances.create(vm_name,
+                                                            provider,
+                                                            form_values=inst_args)
+
+    @request.addfinalizer
+    def delete_vm_and_wait_for_gone():
+        instance.mgmt.delete()
```
john-dupuy: Question: Why not use `cleanup_on_provider`?

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -430,53 +461,58 @@ def test_provision_with_boot_volume(request, instance_args, provider, soft_asser
...
-        # Set up automate
-        method = modified_request_class.methods.instantiate(name="openstack_CustomizeRequest")
+    volume = provider.mgmt.create_volume(1, imageRef=provider.mgmt.get_template(image).uuid)
+    request.addfinalizer(lambda: provider.mgmt.delete_volume(volume))
+
+    # Set up automate
+    method = modified_request_class.methods.instantiate(name="openstack_CustomizeRequest")
+    with update(method):
+        method.script = dedent('''\
+            $evm.root["miq_provision"].set_option(
```
john-dupuy: If this is a hard-coded string, it'd be best to treat it like VOLUME_METHOD or ONE_FIELD and place it at the top of the file.

cfme/tests/cloud_infra_common/test_provisioning.py
john-dupuy: @digitronik Why is this? Just curious.

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -544,20 +592,14 @@ def _finish_method():
...
-        instance.cleanup_on_provider()
-        wait_for(lambda: not instance.exists_on_provider, num_sec=180, delay=5)
-        if "volume_id" in locals():  # To handle the case of 1st or 2nd assert
-            if provider.mgmt.volume_exists(volume_id):
-                provider.mgmt.delete_volume(volume_id)
+
+    assert hasattr(prov_instance_raw, 'os-extended-volumes:volumes_attached')
+    volumes_attached = getattr(prov_instance_raw, 'os-extended-volumes:volumes_attached')
+    assert len(volumes_attached) == 1
+    volume_id = volumes_attached[0]["id"]
```
john-dupuy: Optional: I don't see any need to create these variables if they are just used for the assertion.

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -378,38 +390,57 @@ def test_cloud_provision_from_template_with_attached_disks(
...
-        instance = appliance.collections.cloud_instances.create(vm_name,
-                                                                provider,
-                                                                form_values=inst_args)
-
-        for volume_id in volumes:
-            soft_assert(vm_name in provider.mgmt.volume_attachments(volume_id))
-        for volume, device in device_mapping:
-            soft_assert(provider.mgmt.volume_attachments(volume)[vm_name] == device)
-        instance.mgmt.delete()  # To make it possible to delete the volume
+            method.script = """prov = $evm.root["miq_provision"]"""
```
john-dupuy: Question: reason for triple quotes here?

cfme/tests/cloud_infra_common/test_provisioning.py
jaryn: True

cfme/tests/cloud_infra_common/test_provisioning.py
jaryn: again, the topic of this PR is not to beatifulize the the code. I wanted to make things working. So I was not mangling with everything I saw and could mangle.

Perhaps someone found this more readable then mixing apostrophes and quotes.

cfme/tests/cloud_infra_common/test_provisioning.py
jaryn: Because delete will raise an exception when VM that is expected to be there will not be there, so we would see this as test ERROR or FAIL

cfme/tests/cloud_infra_common/test_provisioning.py
jaryn: This is questionable. This way you can easily see what is the  `method.script` set to, without having to scroll to the top of the file.

cfme/tests/cloud_infra_common/test_provisioning.py
jaryn: Could be done, but again, I was making current things working. I didn't want to end up in reimplementing everything.

cfme/tests/cloud_infra_common/test_provisioning.py
jaryn: Again, this is just some part of already written code that I just had to shuffle around to make tests working and properly clean after.

@john-dupuy I agree what the @digitronik's says because it is shorter than .format(). Logger methods are formatting the messages themselves with the older `"%s"`, so it would look like `logger.info("Provision failed %s: $s", e, provison_request.request_state)`

cfme/tests/cloud_infra_common/test_provisioning.py
jaryn: yeah, this looks odd. But I think this may be what the tests was actually intended to tested. I may try to make this more straightforward.

dajoRH:
*CFME QE Bot*

cfme/tests/cloud_infra_common/test_provisioning.py
john-dupuy: In my opinion, it is better to follow the existing pattern of the codebase, where if you see something in all caps, you know that it is probably defined at the top of the file. This way, static variables are all in a single location for a test module.

cfme/tests/cloud_infra_common/test_provisioning.py
john-dupuy: Based on the docblock of `cleanup_on_provider`, don't we want to avoid such errors in test teardown? @digitronik can you confirm this?

digitronik: Personally, I would avoid to direct method of `mgmt` as we have some Wrapper methods. It will not raise `NotFoundError` just it will log in logger.

cfme/tests/cloud_infra_common/test_provisioning.py
jaryn: Actually, I think this logger message doesn't help with anything and should be removed.

cfme/tests/cloud_infra_common/test_provisioning.py
jaryn: ?

cfme/tests/cloud_infra_common/test_provisioning.py
jaryn: I think these are some names of methods that come from CFME and it is required that they are copied to other domain than they are in by default. So we cannot randomize.

cfme/tests/cloud_infra_common/test_provisioning.py
jaryn: I think these are some names of methods that come from CFME and it is required that they are copied to other domain than they are in by default. So we cannot randomize.

cfme/tests/cloud_infra_common/test_provisioning.py
jaryn: But they are gonna be used just once. There is no need to make define it and make it a constant. It would have a meaning if such constant would be used from many places, but this is not the case.

It is the same as lambda. Lambdas are unnamed functions that are passed to some method as callback targets. There is no need to name them just to be called, so it won't clutter the namespace and one doesn't need to scroll to see what the callback does.

cfme/tests/cloud_infra_common/test_provisioning.py
jaryn: On a second look I don't find these odd. They are just the checks that the test is probably intended to check.

cfme/tests/cloud_infra_common/test_provisioning.py
john-dupuy: @jaryn I still think it is important to maintain consistency with the rest of the code base and test_module. You have two other methods strings `VOLUME_METHOD` and `ONE_FIELD`, why treat this method string differently? In addition, this approach is the one taken by the style guide https://www.python.org/dev/peps/pep-0008/#constants.

cfme/tests/cloud_infra_common/test_provisioning.py
jaryn: @digitronik If the test intends to  create a VM, and the VM disappears during the test or if it were never created, then there was either some bug in the tests code or there is a problem somewhere which should be fixed. Or someone went and deleted the VM. In any case like that I don't see a problem to make the test state ERROR as it failed to cleanup what it intended to create.

This approach will increase the visibility of problems. Or do you think it is good to just put the problems under the carpet (to the log) and put on a ping glasses and pretend everything is fine?

dajoRH:
Lint report for commit 48db857631cd1eb048150e15f54b74f95fc2265c:

`cfme/tests/cloud_infra_common/test_provisioning.py`:
- :red_circle: Line 6:1: **F401** *'fauxfactory' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/cloud_infra_common/test_provisioning.py
jaryn: I removed the hard encoded sttrings used for reseting the `method.script`. I changed the code so that it first stores the `method.script`, then modify, then restores it back

dajoRH:
Lint report for commit f84c52e9267367afeab8581705276f5e1fe0c052:

`cfme/tests/cloud_infra_common/test_provisioning.py`:
- :red_circle: Line 6:1: **F401** *'fauxfactory' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

requirements/frozen.py2.txt
```diff
@@ -328,8 +328,8 @@ Werkzeug==0.15.2
 widgetastic.core==0.34
 widgetastic.patternfly==0.1.0
 widgetsnbextension==3.4.2
-wrapanapi==3.1.2
-wrapt==1.11.1
-xmltodict==0.12.0
+git+git://github.com/jaryn/wrapanapi.git@test_provision_with_boot_volume_fixes
+wrapt==1.10.11
```
mshriver: Are these intentional downgrades?

jaryn: no. somehow I must have done a mistake when rebasing. thanks.

dajoRH:
*CFME QE Bot*

dajoRH:
*CFME QE Bot*

dajoRH:
Lint report for commit c2f7bcac51830a628052e4584b06829342dfe92a:

* c2f7bcac51830a628052e4584b06829342dfe92a

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

jaryn:
The tests that this PR is touching were in really bad shape before. I think I made a good progress. And without changing a thing on that PR it started failing because of some other commit to integration_tests broke something or new CFME change broke it.

I am constantly trying to fix things but this way my PRs are growing in size, I have to maintain them and it is just not good. That's why I would like this PR to get merged even with the 5.11 tests broken.

If I have time tomorrow, I can go and try to find the 5.11 exceptions in trackerbot to support my assumption the failures are not related to this PR

cfme/common/vm.py
mshriver: I don't think that's a good change to make in the common method for create. Some does_vm_exist methods may raise on some providers, and some may just return None/False. If, in a specific use for a test, you want this to raise exception on the very first call to this method, then you should do it custom in the test. You can pass `wait=False` and `find_in_cfme=False` when calling to have it return, and you can call this method yourself.

cfme/tests/cloud_infra_common/test_provisioning.py
mshriver: I disagree, and that's not the pattern we've applied to the test modules. When a necessary configuration value is not available, the test is skipped. The user can't know ahead of time that they explicitly don't want to run this test, how would they know this and trigger a skip?

Ideally the tests run with no errors - just failures and skips. I especially don't want to see errors in fixtures (even AssertionError) when we can easily see and error that could happen.

`provisioning` fixture is just giving us the AttrDict from cfme_data for the provider - and it itself will `pytest.skip` if that section is missing.

If we're then trying to lookup more detailed keys in provisioning, then we should skip if they're missing. We just don't have the requisite data for the provider that was parametrized for the test.

The fact that these keys are required for these tests - the provider marker should be using a more detailed ProviderFilter with `required_fields`. This will exclude providers from collection before we even get to this point.

cfme/tests/cloud_infra_common/test_provisioning.py
mshriver: This doesn't look like it needs to be done within the context manager? You could do this before, to reduce what's being done in `update` - just setting `method.script`.

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -389,38 +389,60 @@ def test_cloud_provision_from_template_with_attached_disks(
...
+                                                            provider,
+                                                            form_values=inst_args)
+
+    @request.addfinalizer
+    def delete_vm_and_wait_for_gone():
+        instance.mgmt.delete()
         wait_for(lambda: not instance.exists_on_provider, num_sec=180, delay=5)
 
+    for volume_id in volumes:
+        assert vm_name in provider.mgmt.volume_attachments(volume_id)
```
mshriver: Sure you want to change this from a `soft_assert`? It allows us to get way more feedback in the test if the volumes are not lining up. With only one failing and killing the test, we get very little info.

With the soft_assert, we can maybe see that the volume ID's are all jumbled, or all empty, or whatever the case may be. We also get information about if its just the volumes, or if the following device_mapping assertions fail as well.

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -442,53 +464,55 @@ def test_provision_with_boot_volume(request, instance_args, provider, soft_asser
...
+    @request.addfinalizer
+    def delete_vm_and_wait_for_gone():
         instance.mgmt.delete()  # To make it possible to delete the volume
         wait_for(lambda: not instance.exists_on_provider, num_sec=180, delay=5)
 
+    request_description = 'Provision from [{}] to [{}]'.format(image, instance.name)
+    provision_request = appliance.collections.requests.instantiate(request_description)
+    provision_request.wait_for_request(method='ui')
+
+    msg = "Provisioning failed with the message {}".format(
```
mshriver: At the risk of being snarky, above you disagreed with @john-dupuy suggestion to use constants for the ruby scripts on the basis of not polluting the namespace with named variables that are used once. Here you've named `request_description` and `msg` when they're only used once, immediately on the next line. Why not put these strings directly into their relevant function calls?

jaryn: 1) It seems to me that both cases are existing code that I was only moving around or shifting left/right.
2) These variables have limited scope and they are used immediately after their definition. Therefore one doesn't need to scroll too mulch to see their definition.

So what I would like to do rather then placing them on top of the file is to place them near they one-shot-only use. They don't really need to be defined in modscope AFAIK

cfme/tests/cloud_infra_common/test_provisioning.py
jaryn: Ok. As you wish.

cfme/common/vm.py
jaryn: Well if `does_vm_exist` can raise on some and return False on other , then I would call this a problem to solve.

cfme/tests/cloud_infra_common/test_provisioning.py
jaryn: @mshriver:
```
def test_soft_assert(soft_assert):
    for a in [False, "fff", True]:
        soft_assert(a).
```
```
____________________________________________ test_soft_assert ____________________________________________

item = <Function 'test_soft_assert'>

    @pytest.mark.hookwrapper
    def pytest_runtest_call(item):
        """pytest hook to handle :py:func:`soft_assert` fixture usage"""
        yield
        if 'soft_assert' in item.fixturenames and _thread_locals.caught_asserts:
>           raise SoftAssertionError(_thread_locals.caught_asserts)
E           SoftAssertionError: 
E           soft_assert(a) (cfme/tests/test_soft_assert.py:3)

```
Doesn't really help much.

cfme/tests/cloud_infra_common/test_provisioning.py
jaryn: I wouldn't call these symbols a constant and none of these does need to be on top of the module.

I didn't want to move things around much but OK. I will move the `ONE_FIELD` and `VOLUME_METHOD` to the placess where I think they do belong.

cfme/tests/cloud_infra_common/test_provisioning.py
mshriver: Understood, but I think you may have been missing the soft assert exception context in the results where soft assert fails. This provides as rich of a message as you want along with the soft_assert. It should be passed as a second option to the  method, and can include whatever information you want so that its clear what failed.

Example of proper use of soft_assert with context:
```
# this statement is in a for loop, where target is the iterated value
soft_assert(False, '0 occurrence of {} found on the timeline of {}'.format(
                        self.event, target))```

And the resulting traceback:
```E           0 occurrence of policy found on the timeline of Host(name=<redacted>, provider=VMwareProvider(<redacted>), hostname=None, ip_address=None, custom_ident=None, host_platform=None, ipmi_address=None, mac_address=None, credentials={}, ipmi_credentials=None, interface_type='lan') (cfme/tests/infrastructure/test_timelines.py:300)
E           0 occurrence of policy found on the timeline of VMwareProvider(<redacted>) (cfme/tests/infrastructure/test_timelines.py:300)

cfme/fixtures/soft_assert.py:64: SoftAssertionError
SoftAssertionError

0 occurrence of policy found on the timeline of Cluster(name=u'Cluster', provider=VMwareProvider(<redacted>) (cfme/tests/infrastructure/test_timelines.py:300)
0 occurrence of policy found on the timeline of Host(name=<redacted>, provider=VMwareProvider(<redacted>), hostname=None, ip_address=None, custom_ident=None, host_platform=None, ipmi_address=None, mac_address=None, credentials={}, ipmi_credentials=None, interface_type='lan') (cfme/tests/infrastructure/test_timelines.py:300)
0 occurrence of policy found on the timeline of VMwareProvider(<redacted>) (cfme/tests/infrastructure/tes...```


As far as debugging goes, I'm not inclined to make test coverage concessions in order to facilitate debugging. We can easily make local changes to facilitate automatic triggering of a debugger. In this case you could have set a break-point at the soft_assert call, and checked the state of the values being asserted before the call is made, instead of relying on automatic triggering.

We're trading valuable assertion data at the time of test for the possibility of triggering a PDB instance when running locally. If we immediately fail on the first assert here, then we miss asserting any remaining volumes that haven't been asserted against yet, and all of the device_mapping assertion below.  We would end up with a failed test in automation, with less information about what was wrong (or right) with the volumes and devices.

File: cfme/tests/cloud_infra_common/test_provisioning.py [hunk 23]
```diff
@@ -465,53 +486,55 @@ def test_provision_with_boot_volume(request, instance_args, provider, soft_asser
...
-        method = modified_request_class.methods.instantiate(name="openstack_CustomizeRequest")
+    volume = provider.mgmt.create_volume(1, imageRef=provider.mgmt.get_template(image).uuid)
+    request.addfinalizer(lambda: provider.mgmt.delete_volume(volume))
+
+    # Set up automate
+    method = modified_request_class.methods.instantiate(name="openstack_CustomizeRequest")
+    view = navigate_to(method, 'Details')
+    former_method_script = view.script.get_value()
+    with update(method):
+        method.script = dedent('''\
```
mshriver: This backslash is unnecessary within the dedent method call.

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -412,38 +390,81 @@ def test_cloud_provision_from_template_with_attached_disks(
...
+    @request.addfinalizer
+    def delete_volumes():
+        for volume in volumes:
+            provider.mgmt.delete_volume(volume)
+
+    # Set up automate
+    for i, volume in enumerate(volumes, 0):
+        if i == 0:
+            provider.mgmt.capi.volumes.set_bootable(volume, True)
+            device_mapping.append(
```
mshriver: Consider only setting bootable in this if, and using the same dictionary instead of repeating.

```
device_mapping.append(dict(
    boot_index=0 if i==0 else -1,
    uuid=volume,
    device_name=device_name.format(chr(ord('a') + i)))
)
~``

File: cfme/tests/cloud_infra_common/test_provisioning.py [hunk 25]
```diff
@@ -412,38 +390,81 @@ def test_cloud_provision_from_template_with_attached_disks(
...
+    volumes = provider.mgmt.volume_configurations(1, n=disks)
+
+    @request.addfinalizer
+    def delete_volumes():
+        for volume in volumes:
+            provider.mgmt.delete_volume(volume)
+
+    # Set up automate
+    for i, volume in enumerate(volumes, 0):
+        if i == 0:
```
mshriver: I think the questions in review in this block warrant more in-line comments of the intent here. Why boot_index is `-1` on the non-bootable disk(s), that the device_name 'counts' through 'vda', 'vdb', etc.

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -412,38 +390,81 @@ def test_cloud_provision_from_template_with_attached_disks(
...
+            :image_ref => nil,
+            :block_device_mapping_v2 => [
+                {}
+            ]
+            }}
+
+            prov = $evm.root["miq_provision"]
+            prov.set_option(:clone_options, clone_options)
+        """)
+        method.script = volume_method.format(",\n".join(disk_mapping))
```
mshriver: I think this is the only statement that should be in the `with update(method)` context. The for loop and volume_method string above don't modify or depend on the method object.

cfme/tests/cloud_infra_common/test_provisioning.py
mshriver: @jaryn

@john-dupuy and @digitronik have valid points here about the test strategy and failure modes here.

I don't think there's any need for the negative tone taken in your response, or fallacy of them intending to hide test failures.  I assure you they both have the best intent for _our_ success.

You have expressed in your comments an implicit assertion that you're relying on in calling this delete method directly.  This implicit assertion/assumption has additionally been placed in the shutdown phase of the test function, without any inline comments to the additional assertion that's in your test.

An assertion that the VM still exists at the end of the test function should be included in the call phase of the test function, not shutdown.

It should also be explicit in its mode of assertion - do you want to 1) assert that the VM is still visible in cloudforms, in a non-archived, non-orphaned state, 2) just still visible in cloudforms, 3) visible on the provider, 4) in a specific power state, etc.

cfme/tests/cloud_infra_common/test_provisioning.py
jaryn: Well it is and in same time it is not:
In [1]: from textwrap import dedent

In [2]: a=dedent("""\
   ...: a""")

In [3]: b=dedent("""
   ...: a""")

In [4]: a==b
Out[4]: False

In [5]: a
Out[5]: 'a'

In [6]: b
Out[6]: '\na'

The newline will certainly not break the ruby thingy, but ... we would have a newline on the start of the string.

cfme/tests/cloud_infra_common/test_provisioning.py
jaryn: ok

cfme/tests/cloud_infra_common/test_provisioning.py
jaryn: Well there is that `provider.mgmt.capi.volumes.set_bootable(volume, True)` that needs to be set only if `i == 0` so I need double ifs then. But whatever. I don't think this is a big deal.

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -412,38 +390,89 @@ def test_cloud_provision_from_template_with_attached_disks(
...
 
-        method = modified_request_class.methods.instantiate(name="openstack_PreProvision")
+    @request.addfinalizer
+    def delete_volumes():
+        for volume in volumes:
+            provider.mgmt.delete_volume(volume)
+
+    # Set up automate
+    for i, volume in enumerate(volumes, 0):
+        # note the boot_index specifies an ordering in which the disks are tried to
```
mshriver: :+1:

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -412,38 +390,89 @@ def test_cloud_provision_from_template_with_attached_disks(
...
+    @request.addfinalizer
+    def delete_vm_and_wait_for_gone():
+        instance.mgmt.delete()
         wait_for(lambda: not instance.exists_on_provider, num_sec=180, delay=5)
 
+    for volume_id in volumes:
+        attachments = provider.mgmt.volume_attachments(volume_id)
+        soft_assert(
+            vm_name in attachments,
+            'The vm {} not found among the attachemnts of volume {}:'.format(
```
mshriver: :+1: Nice, did this syntax help with tracebacks on any test failures you ran into?

jaryn: It might, but I didn't test it. These tests are already passing and I didn't try to introduce any failure.

cfme/tests/cloud_infra_common/test_provisioning.py
jaryn: Done

cfme/tests/cloud_infra_common/test_provisioning.py
jaryn: I am sorry that people might take this personally but I still don't understand why is `cleanup_on_provider` which is first checking for existence is any better than just `instance.mgmt.delete()`.

jaryn: OK. I changed it to cleanup_on_provider.

PR #8632, opened 2019-03-27
[NOTEST][RFR] Added caseautomation for mmojzis' manual tests

mmojzis:
Added caseautomation for mmojzis' manual tests and also changed owners of some tests

### DIFF
--- a/cfme/tests/test_manual.py
+++ b/cfme/tests/test_manual.py
@@ -612,7 +612,7 @@ def test_infrastructure_hosts_icons_states():
     Summary of the host
 
     Polarion:
-        assignee: mmojzis
+        assignee: anikifor
         casecomponent: Infra
         caseimportance: low
         initialEstimate: 1/3h
@@ -1267,7 +1267,7 @@ def test_storage_ebs_volume_crud():
     pass
 
 
-@pytest.mark.manual
+@pytest.mark.manual('manualonly')
 def test_ec2_deploy_cfme_image():
     """
     BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1413835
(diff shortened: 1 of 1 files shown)


mshriver:
@mmojzis I have a PR for pytest-polarion-collect to support separate marking for caseautomation, so that you can specify 'manualonly' in the marker to mark things as manualonly, and they will be marked as notautomated by default.

I will update the pytest-polarion-collect requirement in #8639, once that is done you should be able to update this branch to mark tests appropriately.

mshriver:
@mmojzis A new version of pytest-polarion-collect has been released, supporting notautomated or manualonly test case marking:

https://pypi.org/project/pytest-polarion-collect/

A test case marked with a bare marker (`@pytest.mark.manual()`) will result in a `notautomated` caseautomation value, and a marker with an arg (`@pytest.mark.manual('literallyanything')`) will result in a `manualonly` caseautomation value.

Please include the pytest-polarion-collect version update in this PR, and take advantage of it by marking the cases that you would like to be `manualonly` with a marker with an argument.

cfme/tests/test_manual.py
```diff
@@ -13393,6 +13430,7 @@ def test_infrastructure_provider_left_panel_titles():
         casecomponent: Infra
         caseimportance: low
         initialEstimate: 1/18h
+        caseautomation: manualonly
```
mshriver: With the new pytest-polarion-collect, all the other cases should have the caseautomation value you want with the existing marker.

This test case you will have to add `@pytest.mark.manual('manualonly')` to get 'manualonly' caseautomation value here via the marker.

mmojzis: @mshriver Okay, added manualonly arg to those tests which cannot be automated.

PR #8633, opened 2019-03-27
[1LP][RFR] Automate test: test_delete_advanced_search filter via REST

valaparthvi:
1. Add 2 new test cases:
  1. `test_delete_advanced_search_filter_from_collection`
  2. `test_delete_advanced_search_filter_from_detail`

{{ pytest: cfme/tests/webui/test_rest.py --use-template-cache -sqvvv }}

### DIFF
--- a/cfme/tests/webui/test_rest.py
+++ b/cfme/tests/webui/test_rest.py
@@ -1,7 +1,6 @@
 import fauxfactory
 import pytest
 
-from cfme.infrastructure.provider import InfraProvider
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.rest import delete_resources_from_collection
 from cfme.utils.rest import delete_resources_from_detail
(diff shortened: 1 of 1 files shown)


valaparthvi:
Moving the PR to [RFR] since we're no longer going to support 5.9 stream.

cfme/tests/webui/test_rest.py
```diff
@@ -0,0 +1,61 @@
...
+from cfme.utils.rest import delete_resources_from_detail
+
+
+@pytest.fixture(params=[True, False], ids=["global", "local"])
+def search_filter_obj(appliance, request):
+    filter_name = fauxfactory.gen_string("alphanumeric", 10)
+    filter_value = fauxfactory.gen_string("alphanumeric", 10)
+    param_filter = "Infrastructure Provider : Name"
+
+    view = navigate_to(InfraProvider, "All")
```
mshriver: instead of importing InfraProvider,  I think it would be better to navigate via `appliance.collections.infra_providers`

mshriver: I pushed a commit for this after testing locally.

PR #8638, opened 2019-03-28
[1LP][RFR] New test: Include test cases to assert removal of RSS and Timelines pages

john-dupuy:
Purpose or Intent
=================
Creating a new test for this [RFE BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1672933). Designed to test this [PR](https://github.com/ManageIQ/manageiq-ui-classic/pull/5271) on upstream builds. 

{{ pytest: --use-template-cache cfme/tests/intelligence/test_timelines_rss_removed.py }}

### DIFF
--- a/cfme/tests/intelligence/test_timelines_rss_removed.py
+++ b/cfme/tests/intelligence/test_timelines_rss_removed.py
@@ -1,0 +1,49 @@
+# -*- coding: utf-8 -*-
+import pytest
+from selenium.common.exceptions import NoSuchElementException
+
+from cfme.utils.appliance.implementations.ui import navigate_to
+
+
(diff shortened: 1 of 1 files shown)


izapolsk:
@john-dupuy @mshriver, do you think this stuff has to be automated ?

john-dupuy:
@izapolsk I'm not sure, I don't think it hurts to have it automated.

mshriver:
Its in response to some specific BZ's. I think long term, these tests, and any like them, should be replaced with something more programatically parametrized.  This applies to #8640 as well, which needs to be expanded even more to work for Kubevirt, Lenovo xClarity, and future providers.

PR #8642, opened 2019-03-29
[1LP][RFR][NOTEST] CB: test unique names

digitronik:

Purpose or Intent
=================
try to give unique names without changing common meaning.


mshriver:
Thanks for the targeted PR!

PR #8643, opened 2019-03-29
[RFR] Add supportability.yaml for MIQ master

mshriver:
Define supportability.yaml in integration_tests to specify what MIQ supports.


mshriver:
setting land speed records with that merge

PR #8645, opened 2019-03-29
[1LP][RFR] Update LogValidator usage

mshriver:
Don't pass hostname/username/password if its just the default or current test appliance
Its just duplicating SSHClient default behavior.

PRT failures in run 30978 are not related to the LogValidator/SSHClient changes, but issues with sprout or with the test.

### DIFF
--- a/cfme/utils/ssh.py
+++ b/cfme/utils/ssh.py
@@ -130,24 +130,21 @@ class SSHClient(paramiko.SSHClient):
         self.f_stdout = connect_kwargs.pop('stdout', sys.stdout)
         self.f_stderr = connect_kwargs.pop('stderr', sys.stderr)
 
-        # load the defaults for ssh
-        default_connect_kwargs = {
-            'timeout': 10,
-            'allow_agent': False,
-            'look_for_keys': False,
-            'gss_auth': False
-        }
-        # Load credentials and destination from confs, if connect_kwargs is empty
-        if not connect_kwargs.get('hostname'):
-            default_connect_kwargs['hostname'] = store.current_appliance.hostname
-            default_connect_kwargs['port'] = ports.SSH
-            default_connect_kwargs['username'] = conf.credentials['ssh']['username']
-            default_connect_kwargs['password'] = conf.credentials['ssh']['password']
-        default_connect_kwargs['port'] = connect_kwargs.pop('port', ports.SSH)
+        # load the defaults for ssh, including current_appliance and default credentials keys
+        compiled_kwargs = dict(
+            timeout=10,
+            allow_agent=False,
+            look_for_keys=False,
+            gss_auth=False,
+            hostname=store.current_appliance.hostname,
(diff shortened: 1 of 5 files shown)


jaryn:
Omitting the creds doesn't work. I am getting this:
```
        lv = LogValidator(log,
                          matched_patterns=[r'.*Codename: \w+$'],
                          hostname=appliance.hostname)
>       lv.fix_before_start()

cfme/tests/test_appliance.py:429: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
cfme/utils/log_validator.py:51: in fix_before_start
    self._remote_file_tail.set_initial_file_end()
cfme/utils/ssh.py:731: in set_initial_file_end
    with self as sshtail:
cfme/utils/ssh.py:723: in __enter__
    self.connect(**self._connect_kwargs)
cfme/utils/ssh.py:228: in connect
    conn = super(SSHClient, self).connect(**self._connect_kwargs)
../work-venv/lib/python2.7/site-packages/paramiko/client.py:437: in connect
    passphrase,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <SSHClient hostname=u'10.16.5.97' port=22>, username = 'jhenner', password = None, pkey = None
key_filenames = [], allow_agent = False, look_for_keys = False, gss_auth = False, gss_kex = False
gss_deleg_creds = True, gss_host = None, passphrase = None
```

As you can see, it is trying to use `username = 'jhenner'`

mshriver:
Great point @jaryn, thanks!  I've updated the logic around default options in `SSHClient.__init__`, so that setting of defaults for hostname/username/password/port is not limited to only when a hostname is passed in. I'd like your review/opinion of that updated logic.

cfme/utils/ssh.py
```diff
@@ -130,24 +130,21 @@ def __init__(self, stream_output=False, **connect_kwargs):
...
 
-        # load the defaults for ssh
-        default_connect_kwargs = {
-            'timeout': 10,
-            'allow_agent': False,
-            'look_for_keys': False,
-            'gss_auth': False
-        }
-        # Load credentials and destination from confs, if connect_kwargs is empty
-        if not connect_kwargs.get('hostname'):
```
mshriver: I think this was a bad pattern to follow - only setting default endpoint/credentials if hostname is not passed in.

Made the setting flatter in my update. Setup the entire dictionary of defaults, and then update that with anything that was passed in as connect_kwargs and has not been popped off in the attribute lookups at the start of the method.

PR #8648, opened 2019-04-01
[1LP][RFR] new test_tag_change_tag_name

lina-nikiforova:
{{ pytest: -v cfme/tests/configure/test_tag.py::test_updated_tag_name_on_vm }}

also moved tag creation from `test_tag_crud` to fixture to use it in the new test

### DIFF
--- a/cfme/configure/configuration/region_settings.py
+++ b/cfme/configure/configuration/region_settings.py
@@ -72,8 +72,9 @@ class Tag(Pretty, BaseEntity, Updateable):
     def update(self, updates):
         """ Update category method """
         view = navigate_to(self.parent, 'All')
+        view.table.row(name=self.name).click()
         view.table.fill({
-            self.display_name: {
+            self.name: {
                 'Name': updates.get('name'),
                 'Description': updates.get('display_name')
             }
--- a/cfme/tests/configure/test_tag.py
+++ b/cfme/tests/configure/test_tag.py
@@ -57,22 +57,32 @@ def category(appliance):
         cg.delete()
 
 
+@pytest.fixture
+def tag(category):
+    tag = category.collections.tags.create(
+        name=fauxfactory.gen_alphanumeric(8).lower(),
+        display_name=fauxfactory.gen_alphanumeric(32)
+    )
+    yield tag
+
+    tag.delete_if_exists()
+
+
 @pytest.mark.sauce
 @pytest.mark.tier(2)
-def test_tag_crud(category):
+def test_tag_crud(tag):
     """
     Polarion:
         assignee: anikifor
         initialEstimate: 1/8h
         casecomponent: Tagging
     """
-    tag = category.collections.tags.create(
-        name=fauxfactory.gen_alphanumeric(8).lower(),
-        display_name=fauxfactory.gen_alphanumeric(32).lower()
-    )
-    with update(tag):
-        tag.display_name = fauxfactory.gen_alphanumeric(32).lower()
-    tag.delete()
+    assert tag.exists
+
+    tag.update({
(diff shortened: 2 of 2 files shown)


cfme/tests/configure/test_tag.py
```diff
@@ -57,22 +57,27 @@ def category(appliance):
...
 
 
+@pytest.fixture
+def tag(category):
+    tag = category.collections.tags.create(
+        name=fauxfactory.gen_alphanumeric(8).lower(),
+        display_name=fauxfactory.gen_alphanumeric(32)
+    )
+    yield tag
+    if tag.exists:
```
valaparthvi: You can use `delete_if_exists` method instead.

cfme/tests/configure/test_tag.py
```diff
@@ -107,6 +112,52 @@ def test_map_tagging_crud(appliance, category, soft_assert):
                                       .format(map_tag_entity.label))
 
 
+def test_tag_change_tag_name(provider, tag, request):
```
sshveta: Can you change the test name to  "test_updated_tag_name_on_vm"

cfme/configure/configuration/region_settings.py
```diff
@@ -87,6 +87,17 @@ def delete(self, cancel=False):
         view.browser.handle_alert(cancel=cancel)
         view.flash.assert_no_error()
 
+    def change_name(self, name):
```
sshveta: Optional : I think we should rename this method to "update_tag_name"

cfme/tests/configure/test_tag.py
```diff
@@ -107,6 +112,52 @@ def test_map_tagging_crud(appliance, category, soft_assert):
...
+    tag.change_name(new_tag_name)
+
+    vm_tags = vm.get_tags()
+
+    # assert the tag was not changed in the UI
+    assert any(
+        tag.category.display_name == vm_tag.category.display_name and
+        tag.display_name == vm_tag.display_name
+        for vm_tag in vm_tags
+    ), 'tag is not assigned'
```
sshveta: BZ also states that vm should not show the tag when the tag is deleted .
Can we include that check too here ?

cfme/tests/configure/test_tag.py
lina-nikiforova: @sshveta ok, makes sense

cfme/tests/configure/test_tag.py
lina-nikiforova: @sshveta I don't see this in the BZ. The tag disappeared in the UI after the tag's name was changed, but it was a bug and was observer before the fix.

cfme/tests/configure/test_tag.py
```diff
@@ -57,22 +57,27 @@ def category(appliance):
...
         casecomponent: Tagging
     """
-    tag = category.collections.tags.create(
-        name=fauxfactory.gen_alphanumeric(8).lower(),
-        display_name=fauxfactory.gen_alphanumeric(32).lower()
-    )
-    with update(tag):
-        tag.display_name = fauxfactory.gen_alphanumeric(32).lower()
-    tag.delete()
+    assert tag.exists
```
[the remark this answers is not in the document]
lina-nikiforova: @sshveta now I realized what was supposed to happen here, update didn't work as expected, it created the new tag in the UI, so now I'm updating the method itself. Thanks

cfme/configure/configuration/region_settings.py
lina-nikiforova: @sshveta I removed this method as I realized that `update` method doesn't do what it is supposed to do. Now I fixed it and can utilize it in this test as well.

PR #8654, opened 2019-04-02
[1LP][RFR] items per page in 5.11

lina-nikiforova:
{{ pytest: -v cfme/tests/configure/test_paginator.py::test_paginator_details_page }}

200 and 500 are back! https://bugzilla.redhat.com/show_bug.cgi?id=1515952

I added this additional check in one test as this widget is the same in the whole UI.

### DIFF
--- a/cfme/tests/configure/test_paginator.py
+++ b/cfme/tests/configure/test_paginator.py
@@ -1,6 +1,7 @@
 import fauxfactory
 import pytest
 from widgetastic.exceptions import NoSuchElementException
+from widgetastic_patternfly import Dropdown
 
 from cfme import test_requirements
 from cfme.configure.configuration.region_settings import RedHatUpdates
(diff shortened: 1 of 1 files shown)


cfme/tests/configure/test_paginator.py
```diff
@@ -151,6 +159,15 @@ def test_paginator_details_page(appliance, place_info, schedule):
             table[0].click()
     assert check_paginator_for_page(view) == paginator_expected_result
 
+    if check_paginator_for_page(view):
+        paginator = view.paginator
+        items_selector = Dropdown(view, '{} Items'.format(paginator.items_per_page))
+        msg = 'Not all options are present in items per page'
+        if view.extra.appliance.version < '5.11':
+            assert set(items_selection_5_10).issubset(items_selector.items), msg
```
izapolsk: shouldn't those be equal ?

lina-nikiforova: @izapolsk good point. I agree, we don't want to see any other weird options introduced in this dropdown

cfme/tests/configure/test_paginator.py
```diff
@@ -129,7 +135,9 @@ def test_paginator_config_pages(appliance, place_info):
                               for set_type in details_pages])
 def test_paginator_details_page(appliance, place_info, schedule):
     """
-        Check paginator is visible for access control pages + schedules
+        Check paginator is visible for access control pages + schedules.
+        If paginator is present, check that all options are present in items per page.
+        https://bugzilla.redhat.com/show_bug.cgi?id=1515952
```
jawatts: @mshriver and I just had a dicussion on this, this should be under the `Bugzilla` key and just the id. So:

```
Bugzilla:
    151952
```

PR #8663, opened 2019-04-03
[1LP][RFR] added cloud-init script for vm with powershell

lina-nikiforova:


data/az_pwsh_cloudinit.cfg
```diff
@@ -0,0 +1,129 @@
+#cloud-config
+password: $pass
+chpasswd: { expire: False }
+ssh_pwauth: True
+
+run_cmd:
+    # Register the Microsoft RedHat repository
+  - [ curl, "https://packages.microsoft.com/config/rhel/7/prod.repo", |, sudo, tee,
+  /etc/yum.repos.d/microsoft.repo ]
+  - yum install -y powershell
```
mshriver: All of these commands should actually be in `[cmd, arg, 'arg']` syntax, its basically coincidence that these commands work without it.

data/az_pwsh_cloudinit.cfg
```diff
@@ -0,0 +1,141 @@
...
+        $StorageAccount = Get-AzStorageAccount -ResourceGroup $ResourceGroupName -Name $StorageAccountName
+
+        $SourceImageUri = "https://$StorageAccountName.blob.core.windows.net/$BlobSourceContainer/$BlobNameSource"
+        $Location = $StorageAccount.Location
+        $OSDiskName = $VMName
+
+        # Network
+        $Subnet1Name = "default"
+        $VNetAddressPrefix = "10.1.0.0/16"
+        $VNetSubnetAddressPrefix = "10.1.0.0/24"
```
digitronik: I think we can put this with Azure in data yaml?

lina-nikiforova: @digitronik it's not a secret value https://docs.microsoft.com/en-us/azure/virtual-network/virtual-networks-faq

```
What address ranges can I use in my VNets?
Any IP address range defined in RFC 1918. For example, 10.0.0.0/16. 
```

PR #8669, opened 2019-04-04
[1LP][RFR] Global fixture for custom button support. 

digitronik:
Purpose or Intent
=================
- Added global fixture for returning `service` and respective `vm` object.
- This fixture going to use in upcoming `custom button` test cases
- I found this functionality in one of service test so removed that local fixture added small improvement and move to global.
- Yup there is one more global fixture which order catalog but returning item name. which not useful for everyone. 

{{pytest: cfme/tests/services/test_add_remove_vm_to_service.py -v --long-running}}

### DIFF
--- a/cfme/fixtures/provider.py
+++ b/cfme/fixtures/provider.py
@@ -382,7 +382,7 @@ def big_template_modscope(provider):
     return _get_template(provider, 'big_template')
 
 
-@pytest.fixture(scope="function")
+@pytest.fixture(scope="module")
 def provisioning(provider):
     try:
         return provider.data['provisioning']
--- a/cfme/fixtures/service_fixtures.py
+++ b/cfme/fixtures/service_fixtures.py
@@ -30,7 +30,7 @@ def catalog(request, appliance):
     return _catalog(request, appliance)
 
 
-@pytest.fixture(scope="function")
+@pytest.fixture(scope="module")
 def catalog_item(appliance, provider, provisioning, dialog, catalog):
     catalog_item = create_catalog_item(appliance, provider, provisioning, dialog, catalog)
     return catalog_item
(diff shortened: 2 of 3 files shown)


cfme/fixtures/provider.py
```diff
@@ -382,7 +382,7 @@ def big_template_modscope(provider):
     return _get_template(provider, 'big_template')
 
 
-@pytest.fixture(scope="function")
+@pytest.fixture(scope="module")
```
digitronik: Note: To speed up automation; we need this. I don't want due to `provisioning` lower scope limitation, we can not create long setup fixture with module scope.

cfme/fixtures/service_fixtures.py
```diff
@@ -158,3 +158,27 @@ def order_service(appliance, provider, provisioning, dialog, catalog, request):
     vm_name = '{}0001'.format(catalog_item.prov_data['catalog']['vm_name'])
     vm = appliance.collections.infra_vms.instantiate(vm_name, provider)
     vm.cleanup_on_provider()
+
+
+@pytest.fixture(scope="module")
+def service_vm(appliance, provider, catalog_item):
```
sshveta: Most of the code looks like order_service fixture 
https://github.com/ManageIQ/integration_tests/blob/e76d390991c98d95ac740fb2116cd0f1e4d2ec09/cfme/fixtures/service_fixtures.py#L137

Can we tweak the same method to return service and vm name too ?

digitronik: @sshveta I already mention this possibility in `PR` explanation.

Actually, `order_service` is different than this fixture as its using `request` input for taking decision related to `catalog_item` and yielding `catalog_item`. 

I would like to use this fixture as `setup` for tests (custom button and some service test can use it), so I am `skipping test` if service provisioning `fails` and `order_service` asserting `success` mean it can be a part of test.

So, as per me better to put both separate fixtures and in future if FA owner like to combine he/she can work on it. Also If i add changes it can break service test which i don't want.

sshveta: Yes both fixtures are different in what they yield but with some testing can be merged . I think that's Niyaz's part and can be done later .

cfme/fixtures/provider.py
izapolsk: @digitronik, @mshriver, we need to be more accurate with changing scope of global fixtures. this PR broke 50 - 100 tests due to fixture scope mismatch. I'm reverting these changes - https://github.com/ManageIQ/integration_tests/pull/8736

and let me add that if we need to change something global, we must add all tests affected by this change.

@digitronik, please make sure your test isn't ok after scope fix

digitronik: @izapolsk i changed on basis of theory that `higher` scope fixture can used within `lower` scope. 

Ahhh understand; might be test which are using this has `provider` in `function` scope.
my bad; i'm not checked all service or other tests which are using this fixtures. ok; I will seat with @niyazRedhat for his test structure. `Provider` `function` scope increasing automation time... yup, i know still it will behave like function scope.
Thanks for quick resolution.

PR #8673, opened 2019-04-04
[1LP][RFR] This will add workaround for AMI image selection.

sbulage:
Adding new AMI usage due to BZ 1684203 for EC2 SSA testing.

PRT run:

{{ pytest: cfme/tests/cloud_infra_common/test_vm_instance_analysis.py -k test_ssa_vm -qsvvvv --use-provider ec2west --long-running }}

PR #8676, opened 2019-04-04
[RFR] adding support of ssa docker endpoint to provider.create_rest()

izapolsk:
ec2 provider had support of ssa endpoints only via UI. This PR brings support of ssa endpoints creation to rest


mshriver:
This was included in #8673

PR #8679, opened 2019-04-05
[1LP][RFR][NOTEST] CB: new manual tests

digitronik:

Purpose or Intent
=================
Previously only `VM` object supporting open url and new support going to be add in next release for now adding manual test.

- Added as per custom button test structure.... so as automate ids going to match with manual.

### DIFF
--- a/cfme/tests/automate/custom_button/test_generic_objects.py
+++ b/cfme/tests/automate/custom_button/test_generic_objects.py
@@ -337,7 +337,12 @@ def test_custom_button_open_url_evm_obj(setup_obj, button_group):
             3. Create a custom button with open_url option and respective method
             4. Navigate to object Detail page
             5. Execute custom button
-            6. Check new tab open or not with respective url
+        expectedResults:
+            1.
+            2.
+            3.
+            4.
+            5. New tab should open with respective url
 
     Bugzilla:
         1550002
--- a/cfme/tests/automate/custom_button/test_generic_objects.py
+++ b/cfme/tests/automate/custom_button/test_generic_objects.py
@@ -316,3 +316,30 @@ def test_custom_button_expression_evm_obj(appliance, request, setup_obj, button_
             assert not custom_button_group.is_displayed
             setup_obj.add_tag(tag)
             assert button.text in custom_button_group.items
+
+
+@pytest.mark.manual
+def test_custom_button_open_url_evm_obj(setup_obj, button_group):
+    """ Test Open url functionality of custom button.
+
+    Polarion:
+        assignee: ndhandre
+        initialEstimate: 1/2h
+        caseimportance: high
+        caseposneg: positive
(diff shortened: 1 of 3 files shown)


cfme/tests/automate/custom_button/test_generic_objects.py
```diff
@@ -316,3 +316,27 @@ def test_custom_button_expression_evm_obj(appliance, request, setup_obj, button_
             assert not custom_button_group.is_displayed
             setup_obj.add_tag(tag)
             assert button.text in custom_button_group.items
+
+
+@pytest.mark.manual
+def test_custom_button_open_url_evm_obj(setup_obj, button_group):
```
john-dupuy: Possible to parametrize this test against the different objects? Or is this something you'd like to wait to do until you automate the tests? 

Also could you reference the RFE BZ in here?

digitronik: yup @john-dupuy 
`evm` object need to set as per `object type`. so I need to modify fixture means need a method fixture where i can send object type. I won't like to automate before feature comes in downstream. yup easy to automate for me just waiting.

digitronik: @john-dupuy as you know I like to collect BZ. but this is RFE when just they take decision to add `open url` for other objects. I would not treat it as BZ. Actually `startsin` pointing this. but still if you want let me know.

digitronik: I think better to add let me add this :)

john-dupuy: I'd say it's optional, I've been adding them just as a reference to help understand the test case, but I wouldn't say that it's necessary.

digitronik: yup totally agree with you... better to have `BZ` collection against test. It helps me lot. just i search `BZ` number in `IDE` and I am getting related tests. personally its help me lot. I don't know about other team member. but i found some of us fallowing `Bugzilla` tag.

cfme/tests/automate/custom_button/test_generic_objects.py
```diff
@@ -316,3 +316,30 @@ def test_custom_button_expression_evm_obj(appliance, request, setup_obj, button_
...
+        startsin: 5.11
+        casecomponent: CustomButton
+        tags: custom_button
+        testSteps:
+            1. Create ruby method for url functionality
+            2. Create custom button group with the Object type
+            3. Create a custom button with open_url option and respective method
+            4. Navigate to object Detail page
+            5. Execute custom button
+            6. Check new tab open or not with respective url
```
jawatts: I would expect to see `testSteps` and `expectedResults` for a manual test case here.

jawatts:
waiting for travis

PR #8675, opened 2019-04-04
[1LP][RFR] Remove 59z refs

mshriver:
Remove 59z and older refs

Cleanup BZ's

PR #8680, opened 2019-04-05
[1LP][RFR][NOTEST] Add manual case for BZ 1595259

john-dupuy:
Adding a manual case for https://bugzilla.redhat.com/show_bug.cgi?id=1595259

This is for 5.11/upstream testing.

### DIFF
--- a/cfme/tests/generic_objects/test_instances.py
+++ b/cfme/tests/generic_objects/test_instances.py
@@ -254,3 +254,23 @@ def test_generic_objects_tag_rest(appliance, generic_object, tags):
         generic_object.remove_tag(tag)
         tag_available = generic_object.get_tags()
         assert tag.id not in [t.id for t in tag_available]
+
+
+@pytest.mark.manual
+@pytest.mark.ignore_stream("5.10")
+def test_import_export_generic_object():
+    """
+    Bugzillas:
+        * 1595259
+
+    Polarion:
+        assignee: jdupuy
+        initialEstimate: 1/6h
+        caseimportance: high
+        caseposneg: positive
+        testtype: functional
+        startsin: 5.11
+        casecomponent: GenericObjects
+
+    """
+    pass
--- a/cfme/tests/generic_objects/test_instances.py
+++ b/cfme/tests/generic_objects/test_instances.py
@@ -260,8 +260,8 @@ def test_generic_objects_tag_rest(appliance, generic_object, tags):
 @pytest.mark.ignore_stream("5.10")
 def test_import_export_generic_object():
     """
-    Bugzillas:
-        * 1595259
+    Bugzilla:
+        1595259
 
     Polarion:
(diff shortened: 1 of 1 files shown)


cfme/tests/generic_objects/test_instances.py
```diff
@@ -254,3 +254,23 @@ def test_generic_objects_tag_rest(appliance, generic_object, tags):
         generic_object.remove_tag(tag)
         tag_available = generic_object.get_tags()
         assert tag.id not in [t.id for t in tag_available]
+
+
+@pytest.mark.manual
+@pytest.mark.ignore_stream("5.10")
+def test_import_export_generic_object():
+    """
+    Bugzillas:
```
lina-nikiforova: Required: Let's the same style for referencing BZs
Should be
```
Bugzilla:
    1595259
```
See here, for example https://github.com/ManageIQ/integration_tests/pull/8675/files#diff-df310a5bec51c509260e4f0afbe8eed0R435

cfme/tests/generic_objects/test_instances.py
```diff
@@ -254,3 +254,23 @@ def test_generic_objects_tag_rest(appliance, generic_object, tags):
...
+        * 1595259
+
+    Polarion:
```
lina-nikiforova: Please add some test steps as this is a manual test and it's hard to know what to do for a person not familiar with this functionality

john-dupuy: @lina-nikiforova this is an upstream test, and as we don't have a build yet, I am unsure of the test steps myself. Perhaps I'm confused about the new procedure, do we wait until we have a build to merge these? I was under the impression that we want to make the test case now and then automate/add test steps later.

cfme/tests/generic_objects/test_instances.py
john-dupuy: Agreed, I'm going to open a PR to address this.

john-dupuy: Done by #8684, will probably need rebase after #8675 is merged

lina-nikiforova: @john-dupuy Thanks, can you follow the same pattern for these new tests, they are not in scope of the #8684

cfme/tests/generic_objects/test_instances.py
lina-nikiforova: @john-dupuy that's a valid point, however at least some general test steps would be useful as it will be updated later anyway.
Also, the test steps will be similar to these:
https://github.com/ManageIQ/manageiq/pull/18327
https://github.com/ManageIQ/manageiq/pull/18284

john-dupuy: Thanks @lina-nikiforova I will update this.

PR #8682, opened 2019-04-08
[1LP][RFR] New Test: Checking 'Q-task_id' for automate methods in automation logs

ganeshhubale:
- Automating test case which checks ```Q-task_id``` for automate methods

{{ pytest: cfme/tests/automate/test_method.py -k 'test_task_id_for_method_automation_log' -vv }}

### DIFF
--- a/cfme/tests/automate/test_method.py
+++ b/cfme/tests/automate/test_method.py
@@ -7,6 +7,7 @@ from cfme.automate.explorer.klass import ClassDetailsView
 from cfme.automate.simulation import simulate
 from cfme.rest.gen_data import service_templates as _service_templates
 from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.log_validator import LogValidator
 from cfme.utils.update import update
 from cfme.utils.wait import wait_for
 
@@ -227,17 +228,18 @@ def test_automate_simulate_retry(klass, domain, namespace, original_class):
 
 
 @pytest.fixture(scope='module')
-def service_catalog(request, dialog, catalog):
+def service_catalog_item(request, dialog, catalog):
     """This fixture is used to create generic service catalog item"""
     cat_item = _service_templates(
         request, dialog.appliance, service_dialog=dialog, service_catalog=catalog, num=1
     )[0]
     yield cat_item
-    cat_item.action.delete()
+    if cat_item.exists:
+        cat_item.action.delete()
 
 
 @pytest.mark.tier(1)
-def test_task_id_for_method_automation_log(service_catalog, appliance):
+def test_task_id_for_method_automation_log(service_catalog_item):
     """
     Polarion:
         assignee: ghubale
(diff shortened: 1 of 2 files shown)


cfme/tests/automate/test_method.py
```diff
@@ -222,3 +224,47 @@ def test_automate_simulate_retry(klass, domain, namespace, original_class):
...
+
+
+@pytest.fixture(scope='module')
+def service_catalog(request, dialog, catalog):
+    """This fixture is used to create generic service catalog item"""
+    cat_item = _service_templates(
+        request, dialog.appliance, service_dialog=dialog, service_catalog=catalog, num=1
+    )[0]
+    yield cat_item
+    cat_item.action.delete()
```
john-dupuy: Can you use `delete_if_exists` here?

cfme/tests/automate/test_method.py
```diff
@@ -222,3 +224,47 @@ def test_automate_simulate_retry(klass, domain, namespace, original_class):
...
+            1.
+            2. Task id should be included in automation log for method logs.
+
+    Bugzilla:
+        1592428
+    """
+    service_catalog.action.order()
+
+    # Need to wait for some seconds to get automation logs
+    wait_for(lambda: False, silent_failure=True, timeout=20)
```
john-dupuy: Why not wait_for the result success? This way if the automation log appears in less than 20 seconds you'll be able to catch it.

cfme/tests/automate/test_method.py
```diff
@@ -222,3 +224,47 @@ def test_automate_simulate_retry(klass, domain, namespace, original_class):
...
+    result = appliance.ssh_client.run_command(
+        "grep 'Q-task_id' /var/www/miq/vmdb/log/automation.log"
```
jawatts: Same comment about `LogValidator` as #8685 .

john-dupuy: Also, instead of this ssh command, you can make use of the `LogValidator`, this is a handy way to check any cfme log. 

https://github.com/ManageIQ/integration_tests/blob/master/cfme/utils/log_validator.py#L9

dajoRH:
*CFME QE Bot*

cfme/tests/automate/test_method.py
```diff
@@ -222,3 +225,51 @@ def test_automate_simulate_retry(klass, domain, namespace, original_class):
 
     # Checking whether 'Retry' button is displayed
     assert view.retry_button.is_displayed
+
+
+@pytest.fixture(scope='module')
+def service_catalog_item(request, dialog, catalog):
```
digitronik: I think you are searching for
https://github.com/ManageIQ/integration_tests/blob/master/cfme/fixtures/service_fixtures.py#L40
other type also available have a glance these service fixtures may be help you. Let me know.

ganeshhubale: @digitronik 
- I went through two fixtures which return catalog item: ```catalog_item``` and ```generic_catalog_item```
- ```catalog_item``` requires provider and I am here creating ```generic``` catalog item. So I cant use this fixture.
- ```generic_catalog_item``` fixture yields the object which does not allow me to order catalog item via rest. This object required to be order via UI. So this fixture also is not useful here.
- Hence keeping as it is.

PR #8609, opened 2019-03-22
[1LP][RFR] Testing Datastore Cluster Changes with PRT

jawatts:
{{ pytest: cfme/tests/control/test_actions.py cfme/tests/infrastructure/test_publish_vm_to_template.py cfme/tests/cloud_infra_common/test_vm_instance_analysis.py --long-running --use-provider vsphere }}

PR #8689, opened 2019-04-08
[RFR] Upstream builds templatized and cloned working for VMware, RHEV, and OSP

jawatts:
Requires support for vmware datastore clusters if we want to enable the new vmware provider:
https://github.com/ManageIQ/wrapanapi/pull/368

Needs the following merged first:
https://github.com/ManageIQ/integration_tests/pull/8609

### DIFF
--- a/cfme/utils/template/base.py
+++ b/cfme/utils/template/base.py
@@ -23,9 +23,11 @@ from cfme.utils import trackerbot
 from cfme.utils.conf import cfme_data
 from cfme.utils.conf import credentials
 from cfme.utils.log import logger
+from cfme.utils.net import resolve_hostname
 from cfme.utils.path import project_path
 from cfme.utils.providers import get_mgmt
 from cfme.utils.ssh import SSHClient
+from cfme.utils.wait import wait_for
 
 NUM_OF_TRIES = 3
 lock = Lock()
@@ -185,11 +187,35 @@ class ProviderTemplateUpload(object):
                 'password': credentials[cred_key]['password']}
 
     @cached_property
+    def raw_vm_ssh_client_args(self):
+        """ Returns credentials + hostname for ssh client auth for a temp_vm."""
+
+        def check_ip():
+            vm_name = (self.temp_vm_name if self.provider_data.type == 'rhevm' else
+                       self.template_name)
+            vm = self.mgmt.get_vm(vm_name)
+            ip = vm.ip
+            # get_ip_address might return None
+            return ip if ip and resolve_hostname(ip) else False
+
+        vm_ip, tc = wait_for(check_ip, delay=5, num_sec=600,
+                             message="waiting for {} to obtain DHCP"
+                             .format(self.temp_vm_name))
+
+        return {'hostname': vm_ip,
+                'username': credentials['ssh']['username'],
+                'password': credentials['ssh']['password']}
+
+    @cached_property
     def tool_client_args(self):
         tool_data = self.from_template_upload('tool_client')
         return {'hostname': tool_data.hostname,
             'username': credentials[tool_data['credentials']].username,
             'password': credentials[tool_data['credentials']].password}
+
+    @cached_property
+    def _vm_mgmt(self):
+        return self.mgmt.get_vm(self.template_name)
 
     @staticmethod
     def from_template_upload(key):
@@ -309,11 +335,11 @@ class ProviderTemplateUpload(object):
         client = Client(version='2', **client_kwargs)
         if self.image_name in [i.name for i in client.images.list()]:
             logger.info('Image "%s" already exists on %s, skipping glance_upload',
-                        self.image_name, self.provider_key)
+                        self.image_name, self.glance_key)
             return True
 
         glance_image = client.images.create(
-            name=self.template_name if self.provider_type == 'openstack' else self.image_name,
+            name=self.image_name,
             container_format='bare',
             disk_format='qcow2',
             visibility='public')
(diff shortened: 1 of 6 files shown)


dajoRH:
Lint report for commit 2e37a80a34725e1089412c425863a7085ad996a0:

`cfme/utils/template/openstack.py`:
- :red_circle: Line 3:1: **F401** *'cfme.utils.log.logger' imported but unused*
- :bangbang: Line 17:34: **E126** *continuation line over-indented for hanging indent*
- :bangbang: Line 21:47: **E261** *at least two spaces before inline comment*
- :bangbang: Line 21:48: **E262** *inline comment should start with '# '*

**No commit flaws detected**.

*CFME QE Bot*

cfme/utils/template/base.py
```diff
@@ -184,13 +186,37 @@ def ssh_client_args(self):
...
+            vm_name = (self.temp_vm_name if self.provider_data.type == 'rhevm' else
+                       self.template_name)
+            vm = self.mgmt.get_vm(vm_name)
+            ip = vm.ip
+            # get_ip_address might return None
+            return ip if ip and resolve_hostname(ip) else False
+
+        vm_ip, tc = wait_for(check_ip, delay=5, num_sec=600,
+                             message="waiting for {} to obtain DHCP"
+                             .format(self.temp_vm_name))
```
mshriver: This should use `vm_name` since its set above, dependent on the provider type.

cfme/utils/template/base.py
```diff
@@ -184,13 +186,37 @@ def ssh_client_args(self):
...
     @cached_property
     def tool_client_args(self):
         tool_data = self.from_template_upload('tool_client')
         return {'hostname': tool_data.hostname,
             'username': credentials[tool_data['credentials']].username,
             'password': credentials[tool_data['credentials']].password}
 
+    @cached_property
+    def _vm_mgmt(self):
+        return self.mgmt.get_vm(self.template_name)
```
mshriver: This ends up being a bit ambiguous as a property.  According to the logic in `raw_vm_ssh_client_args`, this name could depend on the provider type?

Without comments here, its not clear why its different for rhevm in some cases, but not others.

Is this vm_mgmt for the temp VM? for the raw VM?

cfme/utils/appliance/__init__.py
```diff
@@ -2379,6 +2379,10 @@ def configure_vm_console_cert(self, log_callback=None):
         if cert is None:
             raise Exception('vm_console:cert does not exist in cfme_data.yaml')
 
+        # Upstream builds do not have the required pip package, pyopenssl, installed
+        if not self.is_downstream:
```
mshriver: This needs to run for 5.11 as well.

cfme/utils/template/openstack.py
```diff
@@ -9,7 +10,44 @@ class OpenstackTemplateUpload(ProviderTemplateUpload):
     log_name = 'RHOS'
     image_pattern = re.compile(r'<a href="?\'?([^"\']*(?:rhos|openstack|rhelosp)[^"\'>]*)')
 
+    @log_wrap('Deploy template to vm - before templatizing')
+    def deploy_vm_from_template(self):
+        """Deploy a instance from the raw template"""
+        self.mgmt.get_template(self.image_name).deploy(
+            vm_name=self.template_name,
```
mshriver: If this uses temp_template_name, then its consistent with the rhv process, and your property for fetching ssh credentials can be more consistent if its trying to get the VM after its been deployed here.

cfme/utils/template/base.py
```diff
@@ -184,13 +186,37 @@ def ssh_client_args(self):
                 'username': credentials[cred_key]['username'],
                 'password': credentials[cred_key]['password']}
 
+    @cached_property
+    def raw_vm_ssh_client_args(self):
+        """ Returns credentials + hostname for ssh client auth for a temp_vm."""
+
+        def check_ip():
+            vm_name = (self.temp_vm_name if self.provider_data.type == 'rhevm' else
```
mshriver: Might be able to remove this confusion altogether by making a settable property for the temp VM, and setting it during the upload run() method. So after a template deploy, set `self._temp_vm_mgmt` to the vm object before returning True and moving on to further VM config.

cfme/utils/template/virtualcenter.py
```diff
@@ -18,8 +18,10 @@ class VMWareTemplateUpload(ProviderTemplateUpload):
     def upload_template(self):
         cmd_args = [
             "ovftool --noSSLVerify",
-            # prefer the datastore from template_upload
-            "--datastore={}".format(self.provider_data.provisioning.datastore),  # move later
+            # Deploy on the datastore for provision, move to template datastore later
+            # JUWATTS CHANGE THIS
```
mshriver: Change this when? Why? Linked to other PR's? GH issues?

cfme/utils/template/virtualcenter.py
```diff
@@ -18,8 +18,10 @@ class VMWareTemplateUpload(ProviderTemplateUpload):
...
+            "--datastore={}".format(self.provider_data.provisioning.datastore_cluster),
```
mshriver: I would think we would want to check for a few fields here in provisioning - `datastore_cluster`, `datastore`, or `allowed_datastores`?

mshriver: This is setup in #8609 , ignore this.

dajoRH:
*CFME QE Bot*

cfme/utils/template/virtualcenter.py
jawatts: This was due to #8609 , rebased so this is resolved.

PR #8419, opened 2019-01-30
[WIP] Moving parametrize structure with minimal global fixture

Yadnyawalkya:
{{pytest: cfme/tests/v2v/test_post_migrations.py -k test_migrations_vm_attributes --use-provider rhv42 --use-provider vsphere65-nested --provider-limit 2 -vvvv }}

This is just proposal, going to move other tests with same pattern form v2v directory.

PR #8690, opened 2019-04-08
[1LP][RFR] Post migration new structure and osp support

Yadnyawalkya:
{{pytest: cfme/tests/v2v/test_post_migrations.py --use-provider rhv42 --use-provider vsphere67-ims --provider-limit 2 -vvvv}}

closing https://github.com/ManageIQ/integration_tests/pull/8419

@sshveta  @digitronik  ^^

### DIFF
--- a/cfme/fixtures/v2v_fixtures.py
+++ b/cfme/fixtures/v2v_fixtures.py
@@ -7,6 +7,7 @@ from riggerlib import recursive_update
 from widgetastic.utils import partial_match
 
 from cfme.cloud.provider.openstack import OpenStackProvider
+from cfme.fixtures.provider import rhel7_minimal
 from cfme.fixtures.provider import setup_or_skip
 from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
@@ -255,6 +256,13 @@ def get_data(provider, component, default_value):
     return data
 
 
+def get_migrated_vm(src_vm_obj, target_provider):
+    """Returns the migrated_vm from target_provider"""
+    collection = target_provider.appliance.provider_based_collection(target_provider)
+    migrated_vm = collection.instantiate(src_vm_obj.name, target_provider)
+    return migrated_vm
+
+
 def infra_mapping_default_data(source_provider, provider):
     """
     Default data for infrastructure mapping form.
@@ -285,6 +293,23 @@ def infra_mapping_default_data(source_provider, provider):
         ],
     }
     return infra_mapping_data
+
+
+@pytest.fixture(scope="function")
+def mapping_data_vm_obj_mini(request, appliance, source_provider, provider):
+    """Fixture to return minimal mapping data and vm object for migration plan"""
+    infra_mapping_data = infra_mapping_default_data(source_provider, provider)
+    vm_obj = get_vm(request, appliance, source_provider, template=rhel7_minimal)
+
+    infrastructure_mapping_collection = appliance.collections.v2v_infra_mappings
+    mapping = infrastructure_mapping_collection.create(**infra_mapping_data)
+
+    @request.addfinalizer
+    def _cleanup():
+        vm_obj.cleanup_on_provider()
+        infrastructure_mapping_collection.delete(mapping)
+
+    return FormDataVmObj(infra_mapping_data=infra_mapping_data, vm_list=[vm_obj])
 
 
 @pytest.fixture(scope="function")
--- a/cfme/tests/v2v/test_post_migrations.py
+++ b/cfme/tests/v2v/test_post_migrations.py
@@ -1,191 +1,100 @@
-"""Test contains all post migration usecases"""
+"""Tests to validate post-migrations usecases"""
 import re
 
 import fauxfactory
 import pytest
 
 from cfme.base.login import BaseLoggedInPage
-from cfme.fixtures.provider import rhel7_minimal
+from cfme.cloud.provider.openstack import OpenStackProvider
+from cfme.fixtures.v2v_fixtures import get_migrated_vm
 from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.markers.env_markers.provider import ONE_PER_TYPE
 from cfme.markers.env_markers.provider import ONE_PER_VERSION
 from cfme.utils.appliance.implementations.ui import navigate_to
-from cfme.utils.appliance.implementations.ui import navigator
-from cfme.utils.log import logger
 from cfme.utils.wait import TimedOutError
-from cfme.utils.wait import wait_for
-
 
 pytestmark = [
     pytest.mark.provider(
-        classes=[RHEVMProvider],
+        classes=[RHEVMProvider, OpenStackProvider],
         selector=ONE_PER_VERSION,
-        required_flags=['v2v'],
-        scope="module"
+        required_flags=["v2v"],
+        scope="module",
     ),
     pytest.mark.provider(
         classes=[VMwareProvider],
         selector=ONE_PER_TYPE,
-        fixture_name='source_provider',
-        required_flags=['v2v'],
-        scope="module"
-    )
+        required_flags=["v2v"],
+        fixture_name="source_provider",
+        scope="module",
+    ),
+    pytest.mark.usefixtures("v2v_provider_setup")
 ]
 
 
-def get_migrated_vm_obj(src_vm_obj, target_provider):
-    """Returns the migrated_vm obj from target_provider."""
-    collection = target_provider.appliance.provider_based_collection(target_provider)
-    migrated_vm = collection.instantiate(src_vm_obj.name, target_provider)
-    return migrated_vm
-
-
-@pytest.mark.parametrize(
-    "form_data_vm_obj_single_datastore", [["nfs", "nfs", rhel7_minimal]], indirect=True
-)
-def test_migration_policy_tag(request, appliance, v2v_providers, host_creds, conversion_tags,
-                              form_data_vm_obj_single_datastore, soft_assert):
-    """Test policy to prevent source VM from starting if migration is complete
+@pytest.mark.tier(1)
+def test_migration_post_attribute(appliance, provider, mapping_data_vm_obj_mini, soft_assert):
+    """
+    Test to validate v2v post-migrations usecases
 
     Polarion:
         assignee: ytale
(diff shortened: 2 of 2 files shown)


cfme/tests/v2v/test_post_migrations.py
```diff
@@ -1,191 +1,97 @@
...
-
-@pytest.mark.parametrize(
-    "form_data_vm_obj_single_datastore", [["nfs", "nfs", rhel7_minimal]], indirect=True
-)
-def test_migration_policy_tag(request, appliance, v2v_providers, host_creds, conversion_tags,
-                              form_data_vm_obj_single_datastore, soft_assert):
-    """Test policy to prevent source VM from starting if migration is complete
+@pytest.mark.tier(1)
+def test_migration_post_attribute(
+        appliance, v2v_provider_setup, mapping_data_vm_obj_mini, soft_assert):
```
sshveta: Please remove from here and do pytest.mark.usefixtures("v2v_provider_setup").
Instead pass 'provider' paramater here which is same as target provider

cfme/tests/v2v/test_post_migrations.py
```diff
@@ -1,191 +1,97 @@
...
-    mapping = infrastructure_mapping_collection.create(form_data_vm_obj_single_datastore.form_data)
-
-    @request.addfinalizer
-    def _cleanup():
-        infrastructure_mapping_collection.delete(mapping)
-
-    migration_plan_collection = appliance.collections.v2v_plans
-    # vm_obj is a list, with only 1 VM object, hence [0]
-    vm_obj = form_data_vm_obj_single_datastore.vm_list[0]
+    target_provider = [
```
sshveta: No need to do this , pass 'provider' in the method and it will automatically have the value that you can use .

cfme/tests/v2v/test_post_migrations.py
```diff
@@ -1,191 +1,97 @@
...
+    summary = source_view.entities.summary("Properties").get_text_of("Container")
+    source_cpu, source_socket, source_core, source_memory = re.findall("\d+", summary)
+    migration_plan_collection = appliance.collections.v2v_migration_plans
     migration_plan = migration_plan_collection.create(
         name="plan_{}".format(fauxfactory.gen_alphanumeric()),
         description="desc_{}".format(fauxfactory.gen_alphanumeric()),
-        infra_map=mapping.name,
-        vm_list=form_data_vm_obj_single_datastore.vm_list,
-        start_migration=True,
+        infra_map=mapping_data_vm_obj_mini.map_obj.name,
```
sshveta: Can we get mapping name from mapping_data_vm_obj_mini.infra_mapping_data.name ?
If yes then we will not need the mapping object

cfme/fixtures/v2v_fixtures.py
```diff
@@ -299,6 +309,24 @@ def infra_mapping_default_data(source_provider, provider):
...
+    infrastructure_mapping_collection = appliance.collections.v2v_infra_mappings
+    mapping = infrastructure_mapping_collection.create(**infra_mapping_data)
+
+    @request.addfinalizer
+    def _cleanup():
+        vm_obj.cleanup_on_provider()
+        infrastructure_mapping_collection.delete(mapping)
+
+    return FormDataVmObjMapObj(
+        infra_mapping_data=infra_mapping_data, vm_list=[vm_obj], map_obj=mapping)
```
sshveta: Can we get mapping name from mapping_data_vm_obj_mini.infra_mapping_data.name ?
If yes then we will not need the mapping object

cfme/fixtures/v2v_fixtures.py
```diff
@@ -267,6 +270,13 @@ def get_data(provider, component, default_value):
     return data
 
 
+def get_migrated_vm(src_vm_obj, target_provider):
+    """Returns the migrated_vm from target_provider"""
+    collection = target_provider.appliance.provider_based_collection(target_provider)
+    migrated_vm = collection.instantiate(src_vm_obj.name, target_provider)
+    return migrated_vm
```
sshveta: :100:

cfme/tests/v2v/test_post_migrations.py
Yadnyawalkya: Should works since we are passing same `infra_mapping_data` to the `infrastructure_mapping_collection` while creating collection object, make sense! :+1:

cfme/fixtures/v2v_fixtures.py
Yadnyawalkya: Yes, object was only to retrieve name so good to remove it.

cfme/tests/v2v/test_post_migrations.py
```diff
@@ -1,191 +1,94 @@
...
+    Test to validate v2v post-migrations usecases
 
     Polarion:
         assignee: ytale
         initialEstimate: 1/4h
+        caseimportance: critical
+        caseposneg: positive
+        testtype: functional
+        startsin: 5.10
         casecomponent: V2V
```
sshveta: Looks good , please add the test steps too.

cfme/tests/v2v/test_post_migrations.py
```diff
@@ -1,191 +1,94 @@
...
-    # vm_obj is a list, with only 1 VM object, hence [0]
-    vm_obj = form_data_vm_obj_single_datastore.vm_list[0]
+    src_vm_obj = mapping_data_vm_obj_mini.vm_list[0]
+    source_view = navigate_to(src_vm_obj, "Details")
+    summary = source_view.entities.summary("Properties").get_text_of("Container")
+    source_cpu, source_socket, source_core, source_memory = re.findall("\d+", summary)
+    migration_plan_collection = appliance.collections.v2v_migration_plans
     migration_plan = migration_plan_collection.create(
         name="plan_{}".format(fauxfactory.gen_alphanumeric()),
         description="desc_{}".format(fauxfactory.gen_alphanumeric()),
```
sshveta: We also need to provide target provider now for OSP to run.
https://github.com/ManageIQ/integration_tests/pull/8694/files#diff-12466afeba20d2a418487020feaecf7cR151

cfme/tests/v2v/test_post_migrations.py
Yadnyawalkya: Looks like I touched automation from long time, I don't remember I wrote teststeps before :D 
Adding it anyways, but do we still need it when we have added comments from which teststeps are fairly predictable, something like this https://github.com/ManageIQ/integration_tests/pull/8690/files/6c17c67b51b29753011eaa1dbc2aaffcced97820#diff-3e10df3dc335f2d18ff441a6bc8cba64R65 ?

cfme/tests/v2v/test_post_migrations.py
Yadnyawalkya: ahh, got you. thanks for searching this. :)

But I still wonder why we are externally adding target_provider in plan since we are not using that anywhere. I have an idea, let me see.

Yadnyawalkya: ahh, got you. thanks for searching this. :)

cfme/tests/v2v/test_post_migrations.py
sshveta: SO just add a summary of things that you are testing post migration in doc block , I see lot of asserts in test .

Yadnyawalkya: Yup. Added it, thanks :)

dajoRH:
*CFME QE Bot*

cfme/tests/v2v/test_post_migrations.py
```diff
@@ -1,191 +1,100 @@
...
-    return migrated_vm
-
-
-@pytest.mark.parametrize(
-    "form_data_vm_obj_single_datastore", [["nfs", "nfs", rhel7_minimal]], indirect=True
-)
-def test_migration_policy_tag(request, appliance, v2v_providers, host_creds, conversion_tags,
-                              form_data_vm_obj_single_datastore, soft_assert):
-    """Test policy to prevent source VM from starting if migration is complete
+@pytest.mark.tier(1)
```
jawatts: **Question** Does this not have to be parametrized over the different datastores anymore?

izapolsk: @jawatts, as I see this is different test.

cfme/tests/v2v/test_post_migrations.py
```diff
@@ -1,191 +1,100 @@
...
-        message="migration plan is in progress, be patient please",
-        delay=15,
-        num_sec=3600,
-    )
-    view.switch_to("Completed Plans")
-    view.wait_displayed()
-    migration_plan_collection.find_completed_plan(migration_plan)
+    assert migration_plan.in_progress
+    assert migration_plan.completed
+    assert migration_plan.successful
```
izapolsk: these 3 asserts looks a bit odd. does this mean that migration plan is in all 3 states at the same time ?

cfme/tests/v2v/test_post_migrations.py
izapolsk: @Yadnyawalkya, please confirm that you have merged two tests into one.
btw, please make sure your changes will be correctly tracked in polarion.

Yadnyawalkya: You both are correct here. :100:

> **Question** Does this not have to be parametrized over the different datastores anymore?

@jawatts : we are not using datastore anymore `mapping_data_vm_obj_mini` will take care of that.
 
> @jawatts, as I see this is different test.

@izapolsk : right, I added two tests into one.

cfme/tests/v2v/test_post_migrations.py
jawatts: @izapolsk I believe that it is asserting that it goes through the 3 states. For example, looking at `in_progress` there is a wait: https://github.com/ManageIQ/integration_tests/blob/master/cfme/v2v/migration_plans.py#L381

Yadnyawalkya: No @izapolsk, migration plan will never be in 3 states at same time.

I added 3 `assert` since I want this test to fail right on that moment if any of condition fails, I'm checking if plan is in progress (this will take 15-30min), if that plan is moved to completed tab (second line) and then if completed plan is success (last assert).

izapolsk: @jawatts, @Yadnyawalkya, sure thing, I  had the same suspicion but I wanted to lead you to another issue :). I think these properties are incorrectly named. their names should explain what is behind. @Yadnyawalkya, can you rename those in one of next PRs ?

PR #8691, opened 2019-04-08
[1LP][RFR] Cleanup and address test failures in test actions

john-dupuy:
Purpose or Intent
=================
`test_actions.py` is in dire need of an update, it has many failing tests and is in need of an upgrade to make use of methods that have since been included in our framework. 

{{ pytest: --use-provider vsphere65-nested --long-running --use-template-cache cfme/tests/control/test_actions.py }}

### DIFF
--- a/cfme/tests/control/test_actions.py
+++ b/cfme/tests/control/test_actions.py
@@ -11,10 +11,9 @@ Required YAML keys:
         nothing terrible happens, but provisioning can be then assigned to a datastore that does not
         work (iso datastore or whatever), therefore failing the provision.
 """
-from functools import partial
-
 import fauxfactory
 import pytest
+from _pytest.outcomes import Failed
 from wrapanapi import VmState
 
 from . import do_scan
@@ -31,9 +30,8 @@ from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.blockers import BZ
 from cfme.utils.generators import random_vm_name
 from cfme.utils.log import logger
-from cfme.utils.pretty import Pretty
+from cfme.utils.log_validator import LogValidator
 from cfme.utils.update import update
-from cfme.utils.virtual_machines import deploy_template
 from cfme.utils.wait import TimedOutError
 from cfme.utils.wait import wait_for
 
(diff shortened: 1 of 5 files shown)


cfme/tests/control/test_actions.py
```diff
@@ -188,20 +121,27 @@ def compliance_condition(appliance):
...
     return compliance_policy
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture
+def compliance_tag(appliance):
+    category = appliance.collections.categories.instantiate("Service Level")
+    tag = category.collections.tags.instantiate(name="gold", display_name="Gold")
+    return tag
+
```
ganeshhubale: - As I see ```compliance_tag``` is used in only one test case: ```test_action_check_compliance```. Also it is just initialization for tag and you need only ```appliance``` for it.
- So according to me, we can directly add it in test case rather than creating different fixture.

cfme/tests/control/test_actions.py
```diff
@@ -250,21 +190,15 @@ def vm_on(vm):
     """ Ensures that the VM is on when the control goes to the test."""
     vm.mgmt.wait_for_steady_state()
     vm.mgmt.ensure_state(VmState.RUNNING)
-    # Make sure the state is consistent
-    vm.crud.refresh_relationships(from_details=True)
-    vm.crud.wait_for_vm_state_change(desired_state=vm.crud.STATE_ON, from_details=True)
-    return vm
+    return
```
ganeshhubale: Question:
- Do we need this blank ```return``` here or you missed ```vm```?

cfme/tests/control/test_actions.py
```diff
@@ -250,21 +190,15 @@ def vm_on(vm):
...
-def vm_off(provider, vm):
+def vm_off(vm):
     """ Ensures that the VM is off when the control goes to the test."""
     vm.mgmt.wait_for_steady_state()
     vm.mgmt.ensure_state(VmState.STOPPED)
-    # Make sure the state is consistent
-    vm.crud.refresh_relationships(from_details=True)
-    vm.crud.wait_for_vm_state_change(desired_state=vm.crud.STATE_OFF, from_details=True)
-    return vm
+    return
```
ganeshhubale: Question:
- Same here for ```return``` 
- Don't we need to ```wait_for_vm_state_change```(here fixture doc string says that VM state should be OFF)?

ganeshhubale:
@john-dupuy 
I can see BZ: [**1439331**](https://bugzilla.redhat.com/show_bug.cgi?id=1439331) is veified. So please remove this BZ

cfme/tests/control/test_actions.py
```diff
@@ -95,8 +58,8 @@ def get_vm_object(appliance, vm_name):
         return None
 
 
-@pytest.fixture(scope="module")
-def vm_name(provider):
+@pytest.fixture
+def vm_name():
     return random_vm_name("action", max_length=16)
```
ganeshhubale: - As this fixture just returns random vm name string and you don't even need other fixture. We can define this as a global variable.

cfme/tests/control/test_actions.py
```diff
@@ -45,39 +41,6 @@
...
-        return self._prov.mgmt
-
-    def __getattr__(self, key):
-        """Creates partial functions proxying to mgmt_system.<function_name>(vm_name)"""
-        func = getattr(self._prov.mgmt, key)
-        return partial(func, self._vm)
-
 
 def get_vm_object(appliance, vm_name):
     """Looks up the CFME database for the VM.
```
ganeshhubale: Remove blank line no. 44

cfme/tests/control/test_actions.py
```diff
@@ -188,20 +121,27 @@ def compliance_condition(appliance):
         expression="fill_tag(VM and Instance.My Company Tags : Service Level, Gold)"
     )
     yield _compliance_condition
-    _compliance_condition.delete()
+    _compliance_condition.delete_if_exists()
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture
 def compliance_policy(vm_name, policy_name, appliance):
```
ganeshhubale: Question: 
- Why do we need to change the fixture scope of these fixtures?

ganeshhubale:
Optional:
- It would be better if you format this file using black. As I have seen this file. It is not well formatted.

cfme/tests/control/test_actions.py
john-dupuy: Since this is a function scoped object, the vm name will be different for each test case. I think it'd be better to get rid of this altogether, since the VM name can be accessed through `vm.name` I will refactor the module to get rid of this.

cfme/tests/control/test_actions.py
john-dupuy: This test module was originally designed to use a single vm for all of the action tests, while this is optimal in terms of speed, it makes these tests extremely flakey. I.e. if a single test fails then I was seeing all the tests fail. It is clearer to use a different VM for each test, and much easier to address test failures.

cfme/tests/control/test_actions.py
john-dupuy: I would prefer to leave this as a fixture. Since it is outside the scope of the test, there is no need to place it inside of the test function itself.

cfme/tests/control/test_actions.py
john-dupuy: There's no need to return `vm` with this fixture. All this fixture does is ensure that the `vm` is running.

cfme/tests/control/test_actions.py
john-dupuy: I was seeing failures related to `wait_for_vm_state_change` when running these tests. However, they may have been remedied by making `vm` function-scoped. However, since the tests are passing, I see no need to add this extra check back in there.

dajoRH:
*CFME QE Bot*

cfme/tests/control/test_actions.py
digitronik: @john-dupuy fyi in python if not returning means we are returning "None".

cfme/tests/control/test_actions.py
```diff
@@ -352,7 +262,7 @@ def test_action_suspend_virtual_machine_after_starting(request, vm, vm_off, poli
     policy_for_testing.assign_actions_to_event("VM Power On", ["Suspend Virtual Machine"])
     request.addfinalizer(policy_for_testing.assign_events)
     # Start the VM
```
digitronik: I would changes this inline comment. as vm started and we are just ensuring status.

cfme/tests/control/test_actions.py
```diff
@@ -31,188 +29,109 @@
...
-        fail_condition=None,
-        num_sec=600,
-        delay=15,
-    )[0]
-
-    return VMWrapper(provider, vm_name, api)
+    return vm
 
 
-@pytest.fixture(scope="module")
```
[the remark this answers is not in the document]
john-dupuy: @digitronik These tests were failing in part due to the fact that they were using the same VM for each test. If a test fails, then the state of the VM got messed up and a bunch of tests would fail.

cfme/tests/control/test_actions.py
```diff
@@ -31,188 +29,109 @@
...
-def vm(request, provider, setup_provider_modscope, small_template_modscope, vm_name):
-    return _get_vm(request, provider, small_template_modscope.name, vm_name)
+@pytest.fixture
+def vm(request, provider, setup_provider_modscope, small_template_modscope):
+    return _get_vm(request, provider, small_template_modscope.name)
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture
```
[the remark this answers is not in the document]
john-dupuy: See comment above, I can try a test with module scope to see if my changes have made these tests more stable. I agree that having a vm for each test function is costly in terms of time. I can try a PRT run with these module scoped.

cfme/tests/control/test_actions.py
```diff
@@ -31,188 +29,109 @@
...
 def policy_profile_name():
     return "action_testing: policy profile {}".format(fauxfactory.gen_alphanumeric())
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture
```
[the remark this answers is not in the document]
john-dupuy: See comments below, testing now on PRT with function scope.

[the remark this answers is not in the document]
john-dupuy: @digitronik See the results from the recent PRT run, 4 out of 15 tests failed when using module scope. The problem is that if a test fails, it is hard to anticipate at which point it will fail and what cleanup will be necessary if you're using the same VM for every test. For that reason, I am going to revert to using the function scope.

dajoRH:
*CFME QE Bot*

cfme/tests/control/test_actions.py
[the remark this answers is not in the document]
john-dupuy: @digitronik I have since figured out what was causing the issue. The `test_action_prevent_vm_retire` was retiring the VM, and the test was broken so it didn't catch the fact that the retirement was not prevented, so it marked it as a pass. Should be fixed now.

john-dupuy:
Tests still failing due to a recent version bump in wt.core from `v0.32.1` to `v0.34`, the issue has to do with the `Table` widget and reading tags from the `Smart Management` table on a VM's summary page.

john-dupuy:
Failing tests should be fixed by https://github.com/RedHatQE/widgetastic.core/pull/150

digitronik:
> Failing tests should be fixed by [RedHatQE/widgetastic.core#150](https://github.com/RedHatQE/widgetastic.core/pull/150)

@john-dupuy can you point Brandon branch temporarily? It will help to check `wt.core` changes and we can forward current PR.

dajoRH:
*CFME QE Bot*

mshriver:
Approved - but you need to update both frozen and frozen_docs for wt.core:

https://github.com/ManageIQ/integration_tests/blob/master/requirements/frozen_docs.py2.txt#L328

PR #8692, opened 2019-04-08
[1LP][RFR] Updating template test to cover vmware provider

kedark3:
Purpose or Intent
=================

__Extending__ template cloning test to cover VMware provider.

{{pytest: cfme/tests/infrastructure/test_publish_vm_to_template.py --long-running }}

### DIFF
--- a/cfme/tests/infrastructure/test_publish_vm_to_template.py
+++ b/cfme/tests/infrastructure/test_publish_vm_to_template.py
@@ -3,14 +3,15 @@ from wrapanapi import VmState
 
 from cfme.infrastructure.provider import InfraProvider
 from cfme.infrastructure.provider.rhevm import RHEVMProvider
+from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.markers.env_markers.provider import ONE_PER_TYPE
-from cfme.utils.blockers import BZ
 from cfme.utils.generators import random_vm_name
 from cfme.utils.log import logger
 from cfme.utils.virtual_machines import deploy_template
 
 pytestmark = [
-    pytest.mark.provider([InfraProvider])
+    pytest.mark.provider([InfraProvider],
+    required_fields=[['templates', 'small_template']])
 ]
 
 
@@ -19,8 +20,11 @@ def vm_crud(provider):
     collection = provider.appliance.provider_based_collection(provider)
     vm_name = random_vm_name(context='pblsh')
     vm = collection.instantiate(vm_name, provider)
-    deploy_template(vm.provider.key, vm_name, provider.data['small_template'],
-                    timeout=2500)
+    try:
+        deploy_template(vm.provider.key, vm_name,
+            provider.data.templates.small_template.name, timeout=2500)
+    except (KeyError, AttributeError):
+        pytest.skip("Skipping as small_template could not be found on the provider")
     vm.wait_to_appear(timeout=900, load_details=False)
     yield vm
 
@@ -32,10 +36,9 @@ def vm_crud(provider):
 
 
 @pytest.mark.rhv2
-@pytest.mark.provider([RHEVMProvider], override=True, scope="module",
+@pytest.mark.provider([RHEVMProvider, VMwareProvider], override=True, scope="module",
                       required_fields=[['templates', 'small_template']],
                       selector=ONE_PER_TYPE)
-@pytest.mark.meta(blockers=[BZ(1622952, forced_streams=['5.10'])])
 def test_publish_vm_to_template(request, setup_provider, vm_crud):
     """ Try to publish VM to template.
     Steps:


cfme/tests/infrastructure/test_publish_vm_to_template.py
```diff
@@ -19,8 +19,12 @@ def vm_crud(provider):
...
     vm_name = random_vm_name(context='pblsh')
     vm = collection.instantiate(vm_name, provider)
-    deploy_template(vm.provider.key, vm_name, provider.data['small_template'],
-                    timeout=2500)
+    if provider.one_of(VMwareProvider):
+        deploy_template(vm.provider.key, vm_name,
+            provider.data['templates']['small_template']['name'], timeout=2500)
+    else:
+        deploy_template(vm.provider.key, vm_name, provider.data['small_template'],
+                        timeout=2500)
```
jawatts: **Required** If a provider does not have these key, it raise a `KeyError`. We can handle this two ways:
1) Use a try/except
2)  Break out getting the template name and use a `.get()`, setting it to none if the key is not found. Then check to make sure the template name is not none.

cfme/tests/infrastructure/test_publish_vm_to_template.py
```diff
@@ -19,8 +19,15 @@ def vm_crud(provider):
     collection = provider.appliance.provider_based_collection(provider)
     vm_name = random_vm_name(context='pblsh')
     vm = collection.instantiate(vm_name, provider)
-    deploy_template(vm.provider.key, vm_name, provider.data['small_template'],
-                    timeout=2500)
+    try:
```
jawatts: I think it be good to put a `required_field` in the `pytestmark` requiring `small_template`. 

Example:
`required_fields=[['templates', 'small_template']]`

cfme/tests/infrastructure/test_publish_vm_to_template.py
```diff
@@ -19,8 +20,15 @@ def vm_crud(provider):
...
+        if provider.one_of(VMwareProvider):
+            deploy_template(vm.provider.key, vm_name,
+                provider.data['templates']['small_template']['name'], timeout=2500)
+        else:
+            deploy_template(vm.provider.key, vm_name, provider.data['small_template'],
+                            timeout=2500)
```
john-dupuy: `2500 s` is a long time, does this ever take that long?

cfme/tests/infrastructure/test_publish_vm_to_template.py
```diff
@@ -19,8 +20,15 @@ def vm_crud(provider):
...
+    except KeyError:
+        pytest.skip("Skipping as small_template could not be found on the provider")
     vm.wait_to_appear(timeout=900, load_details=False)
```
john-dupuy: Same question here about the time

cfme/tests/infrastructure/test_publish_vm_to_template.py
kedark3: Its kinda safe upper bound. During jenkins runs things can take longer time. Specially when there are lot of vms being built. I never personally saw it. But we've been using such upper limits in many places.

cfme/tests/infrastructure/test_publish_vm_to_template.py
digitronik: @kedark3 no need to catch `KeyError` as your required_fields will take care of this. what you say?

cfme/tests/infrastructure/test_publish_vm_to_template.py
```diff
@@ -3,14 +3,15 @@
...
+from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.markers.env_markers.provider import ONE_PER_TYPE
-from cfme.utils.blockers import BZ
 from cfme.utils.generators import random_vm_name
 from cfme.utils.log import logger
 from cfme.utils.virtual_machines import deploy_template
 
 pytestmark = [
-    pytest.mark.provider([InfraProvider])
+    pytest.mark.provider([InfraProvider]),
```
mshriver: This is not the proper way to filter the providers, you only want one mark call here.

Should look like:
```
pytestmark = [pytest.mark.provider([InfraProvider], required_fields=[['templates', 'small_template']])]
```

cfme/tests/infrastructure/test_publish_vm_to_template.py
```diff
@@ -19,8 +20,15 @@ def vm_crud(provider):
     collection = provider.appliance.provider_based_collection(provider)
     vm_name = random_vm_name(context='pblsh')
     vm = collection.instantiate(vm_name, provider)
-    deploy_template(vm.provider.key, vm_name, provider.data['small_template'],
-                    timeout=2500)
+    try:
+        if provider.one_of(VMwareProvider):
+            deploy_template(vm.provider.key, vm_name,
+                provider.data['templates']['small_template']['name'], timeout=2500)
+        else:
```
mshriver: This looks like yaml inconsistency that should be fixed in the yaml.

cfme/tests/infrastructure/test_publish_vm_to_template.py
```diff
@@ -19,8 +20,15 @@ def vm_crud(provider):
...
+            deploy_template(vm.provider.key, vm_name, provider.data['small_template'],
+                            timeout=2500)
+    except KeyError:
```
mshriver: Since you have a mix of calls here, I'd use `except (KeyError, AttributeError)`

cfme/tests/infrastructure/test_publish_vm_to_template.py
kedark3: Will open up a yamls PR.

izapolsk:
looks good but PRT failed and PRT command narrowed down list of providers. I updated it and restarted PRT in order to see complete results.

kedark3:
@izapolsk started PRT again. I had to reference `.name` for small template to make it work. It should pass now.

kedark3:
@izapolsk PRT passing now.

PR #8694, opened 2019-04-09
[1LP][RFR]Ansible test refactoring

sshveta:
{{ pytest: cfme/tests/v2v/test_v2v_ansible.py  --use-provider={rhv42,osp13-ims} --use-provider vsphere67-nested --provider-limit 2 -vvvv --long-running}}

### DIFF
--- a/cfme/tests/v2v/test_v2v_ansible.py
+++ b/cfme/tests/v2v/test_v2v_ansible.py
@@ -2,13 +2,15 @@ import fauxfactory
 import pytest
 
 from cfme import test_requirements
-from cfme.fixtures.provider import rhel7_minimal
 from cfme.cloud.provider.openstack import OpenStackProvider
+from cfme.fixtures.provider import rhel7_minimal
 from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
-from cfme.markers.env_markers.provider import ONE_PER_VERSION, ONE_PER_TYPE
+from cfme.markers.env_markers.provider import ONE_PER_TYPE
+from cfme.markers.env_markers.provider import ONE_PER_VERSION
 from cfme.utils.appliance.implementations.ui import navigate_to
-from cfme.utils.conf import cfme_data, credentials
+from cfme.utils.conf import cfme_data
+from cfme.utils.conf import credentials
 from cfme.utils.wait import wait_for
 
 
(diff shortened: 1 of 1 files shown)


cfme/tests/v2v/test_v2v_ansible.py
```diff
@@ -21,7 +20,7 @@
         server_roles=["+embedded_ansible"]
     ),
     pytest.mark.provider(
-        classes=[RHEVMProvider],
+        classes=[OpenStackProvider, RHEVMProvider],
```
lina-nikiforova: @sshveta  As I understand, this test now should run on OpenStack as well, can you add it to PRT?

sshveta: Hi Angeline , OSP has some provider issue in rest api which is being investigated by @Yadnyawalkya . Till then we can execute this only for RHV

Yadnyawalkya:
> LGTM 
> 
> @Yadnyawalkya can you please confirm. is there any problem with `rhos-ims` provider setup

Good point, so problem was associated with private key content which we use in API request. To check if content is valid or not we found only one way (possibly there are other ways too which I'm not aware of) we connect CFME to conversion host with that key and if access is successful the key got valid content. This is not possible here in `rhos-ims` case since it is not supported version for IMS, so not affected at least for issue I mentioned.

cfme/tests/v2v/test_v2v_ansible.py
```diff
@@ -90,12 +90,28 @@ def _cleanup():
...
+        testSteps:
+            1. Enable embedded ansible role
+            2. Create repository
+            3. Create credentials
+            4. Create ansible catalog item with provision.yml playbook
+            5. Create ansible catalog item with retire.yml playbook
+            6. Migrate VM from vmware to RHV/OSP using the above catalog items
+    """
+
+    creds = credentials[source_provider.data.templates.get("rhel7_minimal").creds]
```
mshriver: **Requred**

To be safe from `TypeError`, this should be `(...).templates.get('rhel7_minimal', {})['creds']`

Otherwise, this doesn't appear to be a `required_field` when marking/filtering providers, and this call should be wrapped with try/except to call `pytest.fail` or `pytest.skip` when the lookup fails.

cfme/tests/v2v/test_v2v_ansible.py
```diff
@@ -117,62 +133,33 @@ def test_migration_playbooks(request, appliance, v2v_providers, host_creds, conv
...
-        vm_list=form_data_vm_obj_single_datastore.vm_list,
-        start_migration=True,
+        vm_list=mapping_data_vm_obj_single_datastore.vm_list,
+        target_provider=provider,
         pre_playbook=provision_catalog.name,
         post_playbook=retire_catalog.name,
+        pre_checkbox=True,
+        post_checkbox=True
     )
+    assert migration_plan.plan_started
```
mshriver: Do these need wait_for or anything? Does it instantly go through all these states? 

Just some in-line comments here explaining the assumptions and why these states can be immediately asserted against, if there's no chance of a race condition causing test failure.

PR #8695, opened 2019-04-09
[1LP][RFR] New Test: Testing embedded method for notify feature

ganeshhubale:
- This test case, tests feature - ```notify``` which helps to provide notifications like info, error, warning etc using embedded method.

{{ pytest: cfme/tests/automate/test_method.py -k 'test_method_for_log_and_notify' -vv}}

### DIFF
--- a/cfme/automate/explorer/method.py
+++ b/cfme/automate/explorer/method.py
@@ -255,7 +255,10 @@ class EmbeddedView(View):
 
     @property
     def is_displayed(self):
-        return self.title.text == "Select Item"
+        return (self.title.text == "Select Item" and
+                self.tree.has_path("Datastore", "RedHat (Locked)") and
+                self.tree.has_path("Datastore", "ManageIQ (Locked)")
+                )
 
 
 class MethodAddView(AutomateExplorerView):
--- a/cfme/automate/explorer/method.py
+++ b/cfme/automate/explorer/method.py
@@ -31,6 +31,7 @@ from cfme.utils.blockers import BZ
 from cfme.utils.timeutil import parsetime
 from cfme.utils.wait import wait_for
 from widgetastic_manageiq import Input
+from widgetastic_manageiq import ManageIQTree
 from widgetastic_manageiq import ScriptBox
 from widgetastic_manageiq import SummaryFormItem
 
@@ -229,6 +230,34 @@ class PlaybookInputParameters(View):
         return self.all_vars
 
 
+class EntryPoint(Text, ClickableMixin):
+    def fill(self, value):
+        """Used to select the automate method from automate domain"""
+        if value:
+            self.click()
+            self.parent_view.embedded_view.wait_displayed('15s')
+            self.parent_view.embedded_view.tree.click_path(*value)
+            return True
+        return None
+
+
+class EmbeddedView(View):
+    """This view is for embedding automate method"""
+    title = Text(
+        './/div[contains(@class, "modal-header ng-scope")]/h4[contains(@class, "modal-title")]'
+    )
+    include_domain = Text(
+        './/div[contains(@class, "bootstrap-switch-container")]/span[contains(@class, '
+        '"bootstrap-switch-label")]'
+    )
+    close = Text('.//button[contains(@ng-click, "closeModal")]')
+    tree = ManageIQTree("treeview-entrypoint_selection")
+
+    @property
+    def is_displayed(self):
+        return self.title.text == "Select Item"
+
+
 class MethodAddView(AutomateExplorerView):
     title = Text('#explorer_title_text')
 
@@ -252,6 +281,10 @@ class MethodAddView(AutomateExplorerView):
     escalate_privilege = BootstrapSwitch('provisioning_become_enabled')
     verbosity = PlaybookBootstrapSelect('provisioning_verbosity')
     playbook_input_parameters = PlaybookInputParameters()
+
+    # Add embedded method
+    embedded_method = EntryPoint('//*[@id="automate-inline-method-select"]//button')
+    embedded_view = View.nested(EmbeddedView)
 
     add_button = Button('Add')
     cancel_button = Button('Cancel')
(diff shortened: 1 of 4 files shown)


dajoRH:
*CFME QE Bot*

dajoRH:
*CFME QE Bot*

dajoRH:
*CFME QE Bot*

cfme/automate/explorer/method.py
```diff
@@ -228,6 +229,34 @@ def read(self):
...
+    include_domain = Text(
+        './/div[contains(@class, "bootstrap-switch-container")]/span[contains(@class, '
+        '"bootstrap-switch-label")]'
+    )
+    close = Text('.//button[contains(@ng-click, "closeModal")]')
+    tree = ManageIQTree("treeview-entrypoint_selection")
+
+    @property
+    def is_displayed(self):
+        return self.title.text == "Select Item"
```
lina-nikiforova: Are there any other things we can rely on for `is_displayed?` we usually have more

ganeshhubale: @lina-nikiforova 
Updated **is_displayed** as below:
```
    def is_displayed(self):
        return (self.title.text == "Select Item" and
                self.tree.has_path("Datastore", "RedHat (Locked)") and
                self.tree.has_path("Datastore", "ManageIQ (Locked)")
                )
```

cfme/automate/explorer/method.py
```diff
@@ -228,6 +229,37 @@ def read(self):
...
 
+class EntryPoint(Text, ClickableMixin):
+    def fill(self, value):
+        """Used to select the automate method from automate domain"""
+        if value:
+            self.click()
+            self.parent_view.embedded_view.wait_displayed('15s')
+            self.parent_view.embedded_view.tree.click_path(*value)
+            return True
+        return None
```
mshriver: The fill method should return boolean, and that boolean should represent whether or not a change was made on the view.

This needs to compare `value` to the currently selected path in the tree, clicking a new one and returning `True`, or just returning `False` if the currently selected path matches `value`

cfme/automate/explorer/method.py
```diff
@@ -228,6 +229,37 @@ def read(self):
...
+
+
+class EmbeddedView(View):
```
mshriver: It looks to me like this should potentially use the wt.pf `Modal` view, which provides several properties and methods for interacting with the modal.

Please evaluate using it, keeping in mind that if `ROOT` locator doesn't exactly match, it can easily be changed in this local subclass.

cfme/automate/explorer/method.py
```diff
@@ -228,6 +229,37 @@ def read(self):
...
+        './/div[contains(@class, "bootstrap-switch-container")]/span[contains(@class, '
+        '"bootstrap-switch-label")]'
+    )
+    close = Text('.//button[contains(@ng-click, "closeModal")]')
+    tree = ManageIQTree("treeview-entrypoint_selection")
+
+    @property
+    def is_displayed(self):
+        return (self.title.text == "Select Item" and
+                self.tree.has_path("Datastore", "RedHat (Locked)") and
```
mshriver: I think these strings should be constructed from the `appliance.product_name` attribute, since its related to productization of ManageIQ.

Will they always be `(Locked)` or is that state dependent?

cfme/automate/explorer/method.py
```diff
@@ -228,6 +229,37 @@ def read(self):
         return self.all_vars
 
 
+class EntryPoint(Text, ClickableMixin):
+    def fill(self, value):
+        """Used to select the automate method from automate domain"""
+        if value:
+            self.click()
+            self.parent_view.embedded_view.wait_displayed('15s')
```
mshriver: This is making a big assumption about the parent view for this widget.  From the structure you have here, are these `EntryPoint` widgets displayed _on_ the `EmbeddedView` modal?  If so, they should be a widget defined within `EmbeddedView`.

You can use `ParametrizedView` and `ParametrizedLocator` to be able to define multiple `EmbeddedView` instances with different locators for `EntryPoint` that's displayed on them.

cfme/automate/explorer/method.py
ganeshhubale: Yes. They always be locked as those are default domains.

dajoRH:
*CFME QE Bot*

cfme/automate/explorer/method.py
ganeshhubale: I am removing this widget as we have common widget ```EntryPoint```

dajoRH:
Lint report for commit cfd0e05d49a16cf535416ab75cedb08a11b68c0d:

`cfme/automate/explorer/method.py`:
- :heavy_exclamation_mark: Line 249:50: **W504** *line break after binary operator*
- :heavy_exclamation_mark: Line 250:68: **W504** *line break after binary operator*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

cfme/automate/explorer/method.py
```diff
@@ -229,6 +232,26 @@ def read(self):
...
+        './/div[contains(@class, "bootstrap-switch-container")]/span[contains(@class, '
+        '"bootstrap-switch-label")]'
+    )
+    close = Text('.//button[contains(@ng-click, "closeModal")]')
+    tree = ManageIQTree("treeview-entrypoint_selection")
+
+    @property
+    def is_displayed(self):
+        return (self.title.text == "Select Item"
+                and self.tree.has_path("Datastore", "RedHat (Locked)")
```
mshriver: I bet the MIQ upstream builds don't have this path :)

mshriver: You can trigger this check based on `appliance.product_name` or `is_downstream`

ganeshhubale: Hey @mshriver 
- I am removing this(```self.tree.has_path("Datastore", "RedHat (Locked)")```) as we have ```self.tree.has_path("Datastore", "ManageIQ (Locked)")``` in both upstream and downstream builds.

dajoRH:
*CFME QE Bot*

PR #8696, opened 2019-04-09
[1LP][RFR]Automate test: test_reports_custom_tags

valaparthvi:
1. Automate test case: `test_reports_custom_tags`
2. Add a fixture to create the custom tags
3. Remove the test from manual file.

{{pytest: cfme/tests/intelligence/reports/test_reports.py::test_reports_custom_tags --use-template-cache -sqvvvv}}

### DIFF
--- a/cfme/tests/intelligence/reports/test_reports.py
+++ b/cfme/tests/intelligence/reports/test_reports.py
@@ -1,3 +1,4 @@
+import fauxfactory
 import pytest
 
 from cfme import test_requirements
@@ -5,6 +6,32 @@ from cfme.rest.gen_data import users as _users
 from cfme.utils.rest import assert_response
 
 pytestmark = [test_requirements.report, pytest.mark.tier(3), pytest.mark.sauce]
+
+
+@pytest.fixture
+def create_custom_tag(appliance):
+    # cannot create a category with uppercase in the name
+    category_name = fauxfactory.gen_alphanumeric().lower()
+
+    # command to create the new category
+    setup_command = (
+        '\'cat = Classification.create_category!(name: "{cat_name}", '
+        'description: "description_{cat_name}", read_only: true);'
+        'cat.add_entry(name: "{cat_name}_entry", description: "{cat_name}_entry_description")\''
+    ).format(cat_name=category_name)
+
+    # command to delete the created category
+    teardown_command = "'cat = Classification.find_by_name(\"{}\");cat.delete()'".format(
+        category_name
+    )
+
+    output = appliance.ssh_client.run_rails_command(setup_command)
+    assert output.success
+
+    yield category_name
+
+    output = appliance.ssh_client.run_rails_command(teardown_command)
+    assert output.success
 
 
 @pytest.fixture(scope="function")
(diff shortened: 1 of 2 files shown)


cfme/tests/intelligence/reports/test_reports.py
```diff
@@ -43,3 +70,37 @@ def test_non_admin_user_reports_access_rest(appliance, request, rbac_api):
...
+        "menu_name": "Custom Category Report {}".format(category_name),
+        "title": "Custom Category Report Title {}".format(category_name),
+        "base_report_on": "Availability Zones",
+        "report_fields": [
+            "Cloud Manager.My Company Tags : description_{}".format(category_name),
+            "VMs.My Company Tags : description_{}".format(category_name),
+        ],
+    }
+    report = appliance.collections.reports.create(**report_data)
+    request.addfinalizer(report.delete)
```
john-dupuy: Question: Able to use `delete_if_exists` here?

cfme/tests/intelligence/reports/test_reports.py
```diff
@@ -43,3 +70,37 @@ def test_non_admin_user_reports_access_rest(appliance, request, rbac_api):
...
+    assert report.exists
```
john-dupuy: Isn't this assertion implied by the create? I.e. if it doesn't exist the test has failed at creation and this assertion won't be hit anyway.

cfme/tests/intelligence/reports/test_reports.py
```diff
@@ -43,3 +70,37 @@ def test_non_admin_user_reports_access_rest(appliance, request, rbac_api):
...
+                ii. cat = Classification.create_category!(
+                    name: "rocat1", description: "read_only cat 1", read_only: true)
+                iii. cat.add_entry(name: "roent1", description: "read_only entry 1")
+        testSteps:
+            1. Create a new report with the newly created custom tag/category.
+        expectedResults:
+            1. Report must be created successfully.
+    """
+    category_name = create_custom_tag
+    report_data = {
```
john-dupuy: Optional: Thoughts on moving this into a fixture?

cfme/tests/intelligence/reports/test_reports.py
valaparthvi: Same comment as I mentioned [here](https://github.com/ManageIQ/integration_tests/pull/8697#discussion_r275235853).

cfme/tests/intelligence/reports/test_reports.py
valaparthvi: Creating the report is a part of the test and creating a separate fixture that just returns the dictionary data seems a little futile imo. But if in the future, I add more test cases that require report data, I will move it to fixture.

cfme/tests/intelligence/reports/test_reports.py
valaparthvi: When the report is created, but we do not check if the report exists inside the `create` method. And checking the report existence is a part of testing, so I asserted it. We could do use `flash.assert_no_error`, but that already happens in `create` method, so I tested `report.exists`.

cfme/tests/intelligence/reports/test_reports.py
```diff
@@ -7,6 +8,32 @@
 pytestmark = [test_requirements.report, pytest.mark.tier(3), pytest.mark.sauce]
 
 
+@pytest.fixture
+def create_custom_tag(appliance):
+    # cannot create a category with uppercase in the name
+    category_name = fauxfactory.gen_alphanumeric().lower()
+
```
ganeshhubale: Question:
```cannot create a category with uppercase in the name```(Is it a bug or feature?)

Comment (valaparthvi) [reply cfme/tests/intelligence/reports/test_reports.py]: Neither I guess. But that is how the category name should be. If you use uppercase, it will throw a validation error.

**Comment by ganeshhubale** [comment]
LGTM

File: cfme/tests/intelligence/reports/test_reports.py [hunk 5]
```diff
@@ -7,6 +8,32 @@
...
+    # command to create the new category
+    setup_command = (
```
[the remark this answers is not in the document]
valaparthvi: It worked. I had tested earlier with both REST and UI, but turns out I always missed out on adding entried to the categories, so it never appeared in the list. I'll make the changes.

PR #8697, opened 2019-04-09
[1LP][RFR]Automate testcase: test_query_custom_category

valaparthvi:
1. Automate test: test_query_custom_category
2. Add a fixture to create a custom category.

{{pytest: cfme/tests/configure/test_tag_category.py::test_query_custom_category_via_api --use-template-cache -sqvvv }}

### DIFF
--- a/cfme/tests/configure/test_tag_category.py
+++ b/cfme/tests/configure/test_tag_category.py
@@ -2,6 +2,7 @@
 import fauxfactory
 import pytest
 
+from cfme import test_requirements
 from cfme.rest.gen_data import categories as _categories
 from cfme.utils.appliance.implementations.ui import navigator
 from cfme.utils.rest import assert_response
@@ -9,6 +10,17 @@ from cfme.utils.rest import delete_resources_from_collection
 from cfme.utils.rest import delete_resources_from_detail
 from cfme.utils.update import update
 from cfme.utils.wait import wait_for
+
+
+@pytest.fixture
+def custom_category(appliance):
+    category = appliance.collections.categories.create(
+        name=fauxfactory.gen_alphanumeric(8).lower(),
+        description=fauxfactory.gen_alphanumeric(32),
+        display_name=fauxfactory.gen_alphanumeric(32),
+    )
+    yield category
+    category.delete_if_exists()
 
 
 @pytest.mark.tier(2)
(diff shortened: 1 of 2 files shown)


dajoRH:
*CFME QE Bot*

cfme/tests/configure/test_tag_category.py
```diff
@@ -11,6 +12,17 @@
...
 
+@pytest.fixture
+def custom_category(appliance):
+    category = appliance.collections.categories.create(
+        name=fauxfactory.gen_alphanumeric(8).lower(),
+        description=fauxfactory.gen_alphanumeric(32),
+        display_name=fauxfactory.gen_alphanumeric(32),
+    )
+    yield category
+    category.delete()
```
john-dupuy: Question: Why not use `delete_if_exists`?

valaparthvi: If you look at the test, I am not performing any destructive activity or any other activity on the object, there is no chance that the object might get deleted during testing, then why use `delete_if_exists`, if I know for sure that the object exists? What do you think?

john-dupuy: Perhaps I am mistaken, but my thought is that if the creation of the category fails, then we can use `delete_if_exists` so that we don't try to delete a now nonexistent category. I suppose that if the creation fails the test will error out anyway though.

valaparthvi: > Perhaps I am mistaken, but my thought is that if the creation of the category fails, then we can use delete_if_exists so that we don't try to delete a now nonexistent category.

Hmm, I didn't realize this could be one of the reasons for using `delete_if_exists`.

>  I suppose that if the creation fails the test will error out anyway though.

True. I think we need to agree upon a common practice. Um, probably using `delete_if_exists` will be a good practice.

valaparthvi: On second thought, if it encounters an error in setup, it will never reach tear down, it will stop the test execution with an error. I think I'll keep the `delete` method as is.

john-dupuy: Yep, makes sense, I wanted to check my understanding of `delete_if_exists`. Thanks for clearing this up!

digitronik: I agree with both of you... good practice is `delete_if_exists` as a part of `tear-down`. why? if create fails then automation will break accepted. If we do some destructive things perform in tests on particular entity and we don't want that effect on our tear-down. so we use delete entity which we created if exist after test at tear-down. what if same fixture `custom_category` someone use in test where he/she delete category?

valaparthvi: Hmm, if the fixture was going to be used by some other test, it might cause a problem at teardown. I guess `delete_if_exists` wins this argument. Maybe using `delete` is a good idea if the teardown happens within the test and we know for sure that the object will not get deleted in the process.

PR #8712, opened 2019-04-12
[1LP][RFR] Automate test: test_populate_default_dialog_values

valaparthvi:
1. Automate a new test: test_populate_default_dialog_values
2. Add fixtures to create setup for the test.
3. Add `rest` test requirement.

{{ pytest: cfme/tests/services/test_rest_services.py::test_populate_default_dialog_values --use-template-cache -sqvvv --long-running }}


cfme/tests/services/test_rest_services.py
```diff
@@ -35,6 +36,7 @@
 pytestmark = [
     pytest.mark.long_running,
     test_requirements.service,
+    test_requirements.rest,
     pytest.mark.tier(2),
```
ganeshhubale: Question:
Can we add two ```test_requirements``` globally?

cfme/tests/services/test_rest_services.py
```diff
@@ -2055,3 +2057,186 @@ def test_delete_cart_from_collection(self, cart):
...
+    method = klass.methods.create(
+        name=fauxfactory.gen_alphanumeric(),
+        display_name=fauxfactory.gen_alphanumeric(),
+        location="inline",
+        script=script,
+    )
+
+    klass.schema.add_fields({"name": "meth", "type": "Method", "data_type": "Integer"})
+
+    instance = klass.instances.create(
```
ganeshhubale: Question:
Why do you need ```data_type``` as ```Integer``` here(Actually don't need to mention ```data_type```. By default it takes ```string```)?

cfme/tests/services/test_rest_services.py
```diff
@@ -2055,3 +2057,186 @@ def test_delete_cart_from_collection(self, cart):
...
+                                "position": 0,
+                                "dynamic": False,
+                                "read_only": True,
+                                "type": "DialogFieldDropDownList",
+                                "resource_action": {"resource_type": "DialogField"},
+                            },
+                            {
+                                "name": "dynamic",
+                                "description": "",
+                                "data_type": "integer",
```
ganeshhubale: Question:
Can we remove ```description``` field(if it is not required field)?

dajoRH:
*CFME QE Bot*

cfme/tests/services/test_rest_services.py
valaparthvi: I saw more than 1 requirements were added in many places which is why I added it here as well, since we're testing services via REST.

cfme/tests/services/test_rest_services.py
valaparthvi: It throws error afair if you pass Integer instead of string.

ganeshhubale:
LGTM But need to be clear about adding [more than one requirement globally](https://github.com/ManageIQ/integration_tests/pull/8712#discussion_r275256267).

cfme/tests/services/test_rest_services.py
```diff
@@ -2105,3 +2107,184 @@ def test_deny_service_ordering_via_api(
     with pytest.raises(APIException):
         template.action.order()
     assert_response(appliance, http_status=400)
+
+
+@pytest.fixture(scope="module")
+def set_run_automate_method(appliance):
+    reset_setting = appliance.advanced_settings["product"][
```
digitronik: optional:
In state of specific with parsing i would make copy of current and restore at tear-down. but its good :+1:

cfme/tests/services/test_rest_services.py
```diff
@@ -2105,3 +2107,184 @@ def test_deny_service_ordering_via_api(
...
+    yield instance
+    method.delete()
+    instance.delete()
+
+
+@pytest.fixture(scope="module")
+def get_service_template(appliance, request, automate_env_setup):
+    instance = automate_env_setup
+    data = {
+        "buttons": "submit,cancel",
```
digitronik: space please? I am suspect is it required in such format?

cfme/tests/services/test_rest_services.py
valaparthvi: I am restoring the original state at tear-down. I don't understand what you're suggesting.

digitronik: btw its optional I was saying copy of `appliance.advanced_settings`

valaparthvi: Oh. I usually restore the appliance in the original state after the test is done.

cfme/tests/services/test_rest_services.py
```diff
@@ -2105,3 +2106,185 @@ def test_deny_service_ordering_via_api(
...
+    Bugzilla:
+        1635673
+        1650252
+        1639413
+        1660237
+    """
+    if auth_type == "token":
+        # obtaining authentication token
+        auth_token_response = requests.get(
+            "https://admin:smartvm@{}/api/auth".format(appliance.hostname), verify=False
```
izapolsk: I doubt this has to be hardcoded. please use yaml or appliance values instead

valaparthvi: Do you mean something like this?
```
credentials = appliance.user.credential
auth_token_response = requests.get(
    appliance.url_path("/api/auth"),
    auth=(credentials.principal, credentials.secret),
    verify=False
)
```

dajoRH:
Lint report for commit 63d0748a905e719e203272fe7319bf902224952c:

`cfme/tests/services/test_rest_services.py`:
- :red_circle: Line 6:1: **F401** *'requests' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

PR #8714, opened 2019-04-12
[1LP][RFR] Removed local Ansible fixture and fixed is_displayed.

sbulage:
PRT Run:

{{ pytest: cfme/tests/configure/test_log_depot_operation.py -k test_collect_log_depot -qsvvv --long-running }}

### DIFF
--- a/cfme/configure/configuration/diagnostics_settings.py
+++ b/cfme/configure/configuration/diagnostics_settings.py
@@ -230,7 +230,7 @@ class CollectLogsBase(Pretty, NavigatableMixin, Updateable):
             flash_message = "Log Depot Settings were saved"
         else:
             logger.info('Settings were not updated')
-        view = self.create_view(flash_view, wait=10)  # implicit assert
+        view = self.create_view(flash_view, wait=20)  # implicit assert
         view.flash.assert_message(flash_message)
 
     @property
@@ -381,7 +381,7 @@ class ServerCollectLogsView(ServerDiagnosticsView):
             # compare with the selection in accordion as it can be a slave server
             self.title.text ==
             self.form_expected_title(name=name,
-                                     sid=sid,
+                                     sid=int(sid[1]),
                                      current='' if current is None else ' (current)')
         )
 
--- a/cfme/configure/configuration/diagnostics_settings.py
+++ b/cfme/configure/configuration/diagnostics_settings.py
@@ -230,7 +230,7 @@ class CollectLogsBase(Pretty, NavigatableMixin, Updateable):
             flash_message = "Log Depot Settings were saved"
         else:
             logger.info('Settings were not updated')
-        view = self.create_view(flash_view, wait=20)  # implicit assert
+        view = self.create_view(flash_view, wait=10)  # implicit assert
         view.flash.assert_message(flash_message)
 
     @property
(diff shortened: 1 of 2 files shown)


sbulage:
All tests passed except one because of Dropbox Credential issue.

cfme/tests/configure/test_log_depot_operation.py
```diff
@@ -77,24 +73,33 @@ def pytest_generate_tests(metafunc):
                 hostname: ftp.example.com
                 use_for_log_collection: True
     """
-    if metafunc.function.__name__ == 'test_collect_unconfigured':
+    if metafunc.function.__name__ == "test_collect_unconfigured":
```
izapolsk: please don't do that

izapolsk:
@sbulage, please don't change formatting and quote marks. it doesn't bring any value but changes a lot of lines.

cfme/tests/configure/test_log_depot_operation.py
sbulage: I am reverting `black` changes.

cfme/configure/configuration/diagnostics_settings.py
```diff
@@ -230,7 +230,7 @@ def update(self, updates, cancel=False, reset=False):
             flash_message = "Log Depot Settings were saved"
         else:
             logger.info('Settings were not updated')
-        view = self.create_view(flash_view, wait=10)  # implicit assert
+        view = self.create_view(flash_view, wait=20)  # implicit assert
```
lina-nikiforova: I think you don't need to increase the wait since you fixed `is_displayed`. Can you check it?

cfme/tests/configure/test_log_depot_operation.py
```diff
@@ -220,71 +219,6 @@ def check_ftp(appliance, ftp, server_name, server_zone_id, check_ansible_logs=Fa
             )
 
 
-@pytest.fixture(scope="module")
-def wait_for_ansible(appliance):
```
lina-nikiforova: Thanks for removing this and using global fixtures!

cfme/tests/configure/test_log_depot_operation.py
```diff
@@ -19,14 +19,13 @@
...
 from cfme.utils.blockers import BZ
-from cfme.utils.conf import cfme_data
 from cfme.utils.ftp import FTPClient
 from cfme.utils.ssh import SSHClient
 from cfme.utils.update import update
 from cfme.utils.virtual_machines import deploy_template
 from cfme.utils.wait import wait_for
 
-pytestmark = [test_requirements.log_depot]
+pytestmark = [pytest.mark.long_running, test_requirements.log_depot]
```
lina-nikiforova: :+1:

cfme/configure/configuration/diagnostics_settings.py
```diff
@@ -381,7 +381,7 @@ def is_displayed(self):
             # compare with the selection in accordion as it can be a slave server
             self.title.text ==
             self.form_expected_title(name=name,
-                                     sid=sid,
+                                     sid=int(sid[1]),
```
lina-nikiforova: can you add an inline comment why do we need to convert to `int` and select the second element?

sbulage: The `form_expected_title` collecting `name`, `sid`, and `current`. But here it is sending `sid` in '[1]' format. At the end we are getting '[[1]]' which is failing to assert. By converting it into `int` solved the problem.

lina-nikiforova: thanks, what I meant is: can you add a small inline comment in the code?

cfme/configure/configuration/diagnostics_settings.py
sbulage: Reset to default time.

cfme/configure/configuration/diagnostics_settings.py
sbulage: Updated.

sbulage:
@izapolsk Updated with required changes.

PR #8715, opened 2019-04-13
[1LP][RFR] make scavenge providers task not to lock provisioning queue

izapolsk:
We encounter issue with scavenge managed providers task. It blocks all provisioning requests for some time when it is kicked off by celery beat.
I'm going to do the following in order to get this issue over with
1. don't do provider scan for appliances with life time less than 2 hours
2. check that appliance is alive 
3. add hard limit for this task and decrease soft limit
4. enable priorities for tasks and set such tasks lower priorities.


izapolsk:
prioritization of tasks will be done in another PR. it requires updating sprout.

PR #8716, opened 2019-04-15
[RFR] Add BZ and workaround for this test

john-dupuy:
Purpose or Intent
=================
Adding BZ metadata to `test_control_assign_actions_to_event`, which is failing due to https://bugzilla.redhat.com/show_bug.cgi?id=1700070. This PR also brings a workaround for this BZ. 

{{ pytest: --long-running --use-template-cache cfme/tests/control/test_basic.py::test_control_assign_actions_to_event }}

### DIFF
--- a/cfme/tests/control/test_basic.py
+++ b/cfme/tests/control/test_basic.py
@@ -17,6 +17,7 @@ from cfme.control.explorer import conditions
 from cfme.control.explorer import policies
 from cfme.control.explorer.alert_profiles import AlertProfileDetailsView
 from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.blockers import BZ
 from cfme.utils.log import logger
 from cfme.utils.update import update
 from cfme.utils.wait import wait_for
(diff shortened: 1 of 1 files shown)


cfme/tests/control/test_basic.py
```diff
@@ -526,6 +529,11 @@ def test_control_assign_actions_to_event(request, policy_class, policy, action):
         request.addfinalizer(policy.assign_events)
     else:
         prefix = policy.TREE_NODE if not policy.TREE_NODE == "Vm" else policy.TREE_NODE.upper()
+        if (
+            policy.TREE_NODE == "Physical Infrastructure" and
+            BZ(1700070, forced_streams=["5.10"]).blocks
```
mshriver: I don't think you want to use `forced_streams` here, this BZ is blocking for me without it, and it'll keep this workaround from auto-undoing itself when the bug is fixed.

```
In [2]: BZ(1700070).blocks
Out[2]: True
```

john-dupuy: Ah! Wasn't working this morning when I checked, I guess the BZ got triaged shortly thereafter, I'll add it in

PR #8717, opened 2019-04-15
[RFR] Default to vddk 6.5 in vddk_url fixture

john-dupuy:
Purpose or Intent
=================
A few tests are skipping because we don't have a `vddk_url` for VSphere 6.7. This PR will make tests use `vddk v6_5` for these skipped tests. 

{{pytest: --long-running --use-template-cache --use-provider vsphere67-nested cfme/tests/control/test_alerts.py::test_alert_hardware_reconfigured }}

### DIFF
--- a/cfme/fixtures/appliance.py
+++ b/cfme/fixtures/appliance.py
@@ -16,6 +16,7 @@ import pytest
 
 from cfme.test_framework.sprout.client import SproutClient
 from cfme.utils import conf
+from cfme.utils.log import logger
 
 
 @contextmanager
@@ -199,9 +200,10 @@ def get_vddk_url(provider):
     except (KeyError, AttributeError):
         pytest.skip('VDDK URL/Version not found in cfme_data.basic_info')
     if url is None:
-        pytest.skip("There is no vddk url for this VMware provider version")
-    else:
-        return url
+        logger.warning("Using VDDK {}, as VDDK {} was unavailable".format("v6_5", vddk_version))
+        url = conf.cfme_data.basic_info.vddk_url.get("v6_5")
+
+    return url
 
 
 @pytest.fixture(scope="function")
--- a/cfme/fixtures/appliance.py
+++ b/cfme/fixtures/appliance.py
@@ -195,13 +195,20 @@ def get_vddk_url(provider):
         major = str(provider.version)
         minor = "0"
     vddk_version = "v{}_{}".format(major, minor)
+
     try:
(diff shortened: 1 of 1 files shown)


cfme/fixtures/appliance.py
```diff
@@ -199,9 +200,10 @@ def get_vddk_url(provider):
     except (KeyError, AttributeError):
         pytest.skip('VDDK URL/Version not found in cfme_data.basic_info')
     if url is None:
-        pytest.skip("There is no vddk url for this VMware provider version")
-    else:
-        return url
+        logger.warning("Using VDDK {}, as VDDK {} was unavailable".format("v6_5", vddk_version))
+        url = conf.cfme_data.basic_info.vddk_url.get("v6_5")
```
mshriver: Its not guaranteed this key will be available either. This is a good fallback lookup, but the skip clause still needs to exist.

mshriver:
```
try:
    vddk_urls =  conf.cfme_data.basic_info.vddk_url
except (KeyError, AttributeError):
    pytest.skip('no vddk urls available')
key = vddk_version
if vddk_version not in vddk_urls:
    logger.warning('using default vddk url')
    key = 'v6_5'
url = vddk_urls.get(key)
```

I'm thinking a structure like this is what we want here, so that we can log and set defaults, and skip on vddk's not available.

cfme/fixtures/appliance.py
```diff
@@ -194,14 +195,22 @@ def get_vddk_url(provider):
...
+
     try:
-        url = conf.cfme_data.basic_info.vddk_url.get(vddk_version)
+        vddk_urls = conf.cfme_data.basic_info.vddk_url
     except (KeyError, AttributeError):
-        pytest.skip('VDDK URL/Version not found in cfme_data.basic_info')
+        pytest.skip("VDDK URLs not found in cfme_data.basic_info")
+
+    if vddk_version not in vddk_urls:
+        logger.warning("Using VDDK {}, as VDDK {} was unavailable".formate("v6_5", vddk_version))
```
mshriver: typo, `format`

mshriver: ah, also, almost forgot. With logger, we want to pass it the strings to format in.

`logger.warning('my message with arg: %s', arg_to_format_in_to_log)`

john-dupuy: Thanks!

PR #8718, opened 2019-04-16
[1LP][RFR] Update snapshot strings

apagac:
__Updating__ strings related to snapshot manipulation. 'ManageIQ' has changed to 'CFME'.

This fixes all failing snapshot tests.

{{ pytest: -v --long-running cfme/tests/infrastructure/test_snapshot.py --use-provider vsphere67-nested }}

### DIFF
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -685,7 +685,8 @@ class InfraVm(VM):
             if not cancel:
                 # TODO: test this in test_snapshot_crud, just assert_no_error here
                 view.flash.assert_message(
-                    "Delete Snapshot initiated for 1 VM and Instance from the ManageIQ Database"
+                    "Delete Snapshot initiated for 1 VM and Instance from the "
+                    "{} Database".format(self.parent_vm.appliance.product_name)
                 )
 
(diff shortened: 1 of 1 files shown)


cfme/infrastructure/virtual_machines.py
```diff
@@ -685,7 +685,7 @@ def delete(self, cancel=False):
             if not cancel:
                 # TODO: test this in test_snapshot_crud, just assert_no_error here
                 view.flash.assert_message(
-                    "Delete Snapshot initiated for 1 VM and Instance from the ManageIQ Database"
+                    "Delete Snapshot initiated for 1 VM and Instance from the CFME Database"
```
mshriver: This should likely use `appliance.product_name` property in order to be compatible with upstream and downstream. This was likely left as MIQ by accident downstream, and has been 'fixed' in the downstream images. Once we test upstream images, this will still be 'MIQ' in the appliance under test.

apagac: I was thinking something similar. Will fix.

Also, there's another issue here I need to fix before RFR, with `test_create_snapshot_via_ae`.

apagac:
PRT failures are unrelated to this fix.

Revert snapshot tests have a long standing problem with ssh client.
Suspended VM operation has changed behaviour in 5.10.

PR #8672, opened 2019-04-04
[1LP][RFR] Add Requirement Meta

mshriver:
Updating all test requirement markers with metadata to populate polarion (or any other test tool requirement item)

PR #8719, opened 2019-04-16
[RFR] fix two smoke tests

digitronik:

Purpose or Intent
=================
to fix
```
 ValueError: invalid literal for int() with base 10:
```

### DIFF
--- a/cfme/tests/test_appliance.py
+++ b/cfme/tests/test_appliance.py
@@ -152,8 +152,7 @@ def test_memory_total(appliance):
         testtype: functional
         casecomponent: Appliance
     """
-    result = appliance.ssh_client.run_command(
-        r'free -g | grep Mem: | awk \'{ print $2 }\'').output
+    result = appliance.ssh_client.run_command(r"free -g | grep Mem: | awk '{ print $2 }'").output
     assert int(result) >= 6
 
 
@@ -168,7 +167,8 @@ def test_cpu_total(appliance):
         casecomponent: Appliance
     """
     result = appliance.ssh_client.run_command(
-        r'lscpu | grep ^CPU\(s\): | awk \'{ print $2 }\'').output
+        r"lscpu | grep ^CPU\(s\): | awk '{ print $2 }'"
+    ).output
     assert int(result) >= 4
 
 


dajoRH:
Lint report for commit beb36963e96cb998c8f21ef309f4e0b3f00bf5d8:

`cfme/tests/test_appliance.py`:
- :bangbang: Line 174:1: **E303** *too many blank lines (3)*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/test_appliance.py
```diff
@@ -152,9 +152,8 @@ def test_memory_total(appliance):
         testtype: functional
         casecomponent: Appliance
     """
-    result = appliance.ssh_client.run_command(
-        r'free -g | grep Mem: | awk \'{ print $2 }\'').output
-    assert int(result) >= 6
+    result = appliance.ssh_client.run_command(r"free -g | grep Mem: | awk '{ print $2 }'").output
+    assert int(result.strip()) >= 6
```
mshriver: I did the same thing in a branch, but the strip is unnecessary here:
```
In [7]: int(app.ssh_client.run_command(r"free -g | grep Mem: | awk '{ print $2 }'").output)
Running command "free -g | grep Mem: | awk '{ print $2 }'"
Out[7]: 11
```

mshriver: Also, this fix is in 8672 already:
https://github.com/ManageIQ/integration_tests/pull/8672/files#diff-07e1dc45e2c0264896d4bceb289d1fa0L155

But I'm not opposed to merging separately.

mshriver: Now that I say that, using single quotes still causes issues in 8672, I'll update.

digitronik: I think srip is better before conversion :). @mshriver if you are going to merge #8672 then good to close this. I just found this while reviewing one of PR.

mshriver: Cool. I'd disagree about the strip, because its unnecessary, the `int()` call handles it.  Similar to how we don't use `dict.get(key, None)` because `None` is the default implicit behavior, here its an extra call to strip that's implicitly handled by the int() conversion.

PR #8723, opened 2019-04-17
[1LP][RFR] Fixed AssertionError while copying method/instance/class to domain

ganeshhubale:
This PR includes following changes:
- Removing flash message assertions from module page
- Automated test case which was related to copy ```automate class/instance/method``` to new domain
- Asserted those removed statements in this test case
- This fix will save more number of test cases related to quota and automate from failures
- Fixed Error:

```E           AssertionError: assert exact match of message: Copy selected Automate <class 'cfme.automate.explorer.instance.Instance'> was saved. 
E            Available messages: [u'Copy selected Automate Instance was saved']

/var/ci/cfme_venv/lib/python2.7/site-packages/widgetastic_patternfly/__init__.py:299: AssertionError
AssertionError
assert exact match of message: Copy selected Automate <class 'cfme.automate.explorer.instance.Instance'> was saved. 
 Available messages: [u'Copy selected Automate Instance was saved']```
```
{{ pytest: cfme/tests/automate/test_domain.py -k 'test_copy_to_domain' -v }}

### DIFF
--- a/cfme/tests/automate/test_domain.py
+++ b/cfme/tests/automate/test_domain.py
@@ -4,6 +4,9 @@ import pytest
 
 from cfme import test_requirements
 from cfme.automate.explorer.domain import DomainAddView
+from cfme.automate.explorer.instance import InstanceCopyView
+from cfme.automate.explorer.klass import ClassCopyView
+from cfme.automate.explorer.method import MethodCopyView
 from cfme.exceptions import OptionNotAvailable
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.blockers import BZ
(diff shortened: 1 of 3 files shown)


cfme/tests/automate/test_domain.py
```diff
@@ -291,3 +294,68 @@ def test_object_attribute_type_in_automate_schedule(appliance):
...
+def test_copy_to_domain(domain):
+    """This test case checks whether automate class, instance and method are successfully copying to
+    domain.
+
+    Polarion:
+        assignee: ghubale
+        casecomponent: Automate
+        caseimportance: low
+        initialEstimate: 1/15h
+        startsin: 5.10
```
valaparthvi: Why change it to 5.10 from 5.9? I think if the feature was introduced in 5.9, it should be kept that way as an FYI, even if we will no longer test 5.9.

cfme/tests/automate/test_domain.py
```diff
@@ -291,3 +294,68 @@ def test_object_attribute_type_in_automate_schedule(appliance):
...
+    original_klass = miq.classes.instantiate("MiqAe")
+
+    # Copy this Class to custom domain
+    original_klass.copy_to(domain=domain)
+    klass = domain.browser.create_view(ClassCopyView)
+    klass.flash.wait_displayed()
+    klass.flash.assert_message("Copy selected Automate Class was saved")
+
+    # Instantiating Instance - 'quota_source' from 'ManageIQ' domain
+    original_instance = miq.classes.instantiate("QuotaMethods").instances.instantiate(
```
valaparthvi: Optional: You're using `miq.classes.instantiate("QuotaMethods")` more than once here, you could put it in a variable and then use it. I don't prefer this a lot, but we have been practicing this in our codebase.

cfme/tests/automate/test_domain.py
ganeshhubale: Yeah agree! I know about the ```startsin```.  But as we are not supporting ```5.9```. So intentionally, It have changed to 5.10. Although that is not required field in polarion tags.

PR #8725, opened 2019-04-17
[RFR] Update package versions, miq-version in particular

mshriver:
Used freeze_all.py, so it pulled in a bunch of other stuff too.


requirements/frozen_docs.py2.txt
```diff
@@ -91,7 +90,6 @@ oauthlib==2.1.0
 os-service-types==1.3.0
 oslo.i18n==3.22.1
 oslo.utils==3.37.1
-ovirt-engine-sdk-python==4.3.0
```
izapolsk: I'm wondering if it really not necessary

izapolsk: it is still there - https://github.com/ManageIQ/wrapanapi/blob/master/requirements.txt#L15

requirements/frozen_docs.py2.txt
```diff
@@ -21,7 +21,6 @@ bottle-sqlite==0.1.3
 cached-property==1.5.1
 certifi==2018.10.15
 cffi==1.11.5
-cfme-testcases==0.12
```
izapolsk: we still use this, aren't we ?

mshriver: Correct - but its not needed for docs (or at least isn't in the docs template)

PR #8729, opened 2019-04-18
[1LP][RFR] Optimize assign events for policies

john-dupuy:
Purpose or Intent
=================
The current implementation of `BootstrapSwitchSelect`'s fill method is quite slow (taking around 10 s before a fill when assigning events to a policy), since it looks at each switch on the page and decides whether or not it's selected. 

This PR modifies the fill method of `BootstrapSwitchSelect` that allows for a faster fill of the page. It is less cautious than the original fill method since it does not look at the events which are already selected. 

This PR also introduces an `unassign_events` function that takes in the names of the events you'd like to unassign from the policy. I have modified tests to use this function for cleanup, as it is more explicit (and faster) than using `assign_events()`  

{{ pytest: --long-running --use-template-cache cfme/tests/control/test_basic.py::test_assign_two_random_events_to_control_policy cfme/tests/control/test_basic.py::test_control_assign_actions_to_event }}

### DIFF
--- a/cfme/control/explorer/policies.py
+++ b/cfme/control/explorer/policies.py
@@ -335,6 +335,7 @@ class BasePolicy(BaseEntity, Updateable, Pretty):
 
         Args:
             events: Events which need to be assigned.
+        Kwargs:
             extend: Do not uncheck existing events.
         """
         events = list(events)
@@ -344,12 +345,45 @@ class BasePolicy(BaseEntity, Updateable, Pretty):
         view = navigate_to(self, "Details")
         view.configuration.item_select("Edit this Policy's Event assignments")
         view = self.create_view(EditPolicyEventAssignments, wait="20s")
+        # decide if any events are already assigned
+        events = [event for event in events if not self.is_event_assigned(event)]
+        # select events
         changed = view.fill({"events": events})
         if changed:
             view.save_button.click()
+            view.flash.assert_success_message('Policy "{}" was saved'.format(self.description))
         else:
             view.cancel_button.click()
-        view.flash.assert_success_message('Policy "{}" was saved'.format(self.description))
+            view.flash.assert_success_message(
+                'Edit of Policy "{}" was cancelled by the user'.format(self.description)
+            )
+
+    def unassign_events(self, *events, **kwargs):
+        """Unassign events from this Policy.
+
+        Args:
+            events: Events which need to be unassigned.
+        Kwargs:
+            extend: Do not uncheck existing events.
+        """
+        events = list(events)
+        if kwargs.pop("extend", False):
+            events += self.assigned_events
+        view = navigate_to(self, "Details")
+        view.configuration.item_select("Edit this Policy's Event assignments")
+        view = self.create_view(EditPolicyEventAssignments, wait="20s")
+        # decide if events are assigned
+        events = [event for event in events if self.is_event_assigned(event)]
+        # unassign the events which are assigned
+        changed = view.fill({"events": events})
+        if changed:
+            view.save_button.click()
+            view.flash.assert_success_message('Policy "{}" was saved'.format(self.description))
+        else:
+            view.cancel_button.click()
+            view.flash.assert_success_message(
+                'Edit of Policy "{}" was cancelled by the user'.format(self.description)
+            )
 
     def is_event_assigned(self, event):
         return event in self.assigned_events
--- a/cfme/tests/cloud_infra_common/test_events.py
+++ b/cfme/tests/cloud_infra_common/test_events.py
@@ -73,7 +73,11 @@ def test_vm_create(request, appliance, vm_crud, provider, register_event):
     request.addfinalizer(policy.delete)
 
     policy.assign_events("VM Create Complete")
-    request.addfinalizer(policy.assign_events)
+
+    @request.addfinalizer
+    def _cleanup():
+        policy.unassign_events("VM Create Complete")
+
     policy.assign_actions_to_event("VM Create Complete", action)
 
     profile = appliance.collections.policy_profiles.create(
--- a/cfme/tests/control/test_basic.py
+++ b/cfme/tests/control/test_basic.py
@@ -165,7 +165,6 @@ EVENTS = [
     "Host Disconnect",
     "Host Maintenance Enter Request",
     "Host Maintenance Exit Request",
-    "Host Provision Complete",
     "Host Reboot Request",
     "Host Removed from Cluster",
     "Host Reset Request",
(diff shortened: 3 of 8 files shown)


dajoRH:
Lint report for commit c6ee9ececb0dd8616fa590176a5ffb0dbd99b385:

`cfme/tests/control/test_basic.py`:
- :bangbang: Line 525:9: **E306** *expected 1 blank line before a nested definition, found 0*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
Lint report for commit daf5ca8699ce9cc2866632a3cf8948202c364a13:

`cfme/tests/cloud_infra_common/test_events.py`:
- :bangbang: Line 76:5: **E306** *expected 1 blank line before a nested definition, found 0*

`cfme/tests/control/test_basic.py`:
- :bangbang: Line 528:9: **E306** *expected 1 blank line before a nested definition, found 0*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/control/test_basic.py
```diff
@@ -506,7 +505,7 @@ def test_assign_two_random_events_to_control_policy(control_policy, control_poli
         initialEstimate: 1/4h
     """
     random_events = random.sample(EVENTS, 2)
-    control_policy.assign_events(*random_events)
+    control_policy.assign_events(*random_events, fast=True)
```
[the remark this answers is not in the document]
john-dupuy: As this is leftover from a previous commit, thanks for pointing this out. Shouldn't be there.

widgetastic_manageiq/__init__.py
```diff
@@ -394,23 +394,41 @@ def _values_to_add(self, values):
...
+    def fill(self, values):
+        if values:
+            for value in values:
+                switch = self.switch_by_text(value)
+                if switch.selected:
+                    switch.fill(False)
+                else:
+                    switch.fill(True)
+            return True
+        else:
```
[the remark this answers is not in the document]
john-dupuy: I think we can read the specific switch and make sure that it's value changed, that's a good idea

cfme/tests/intelligence/test_reports_with_timelines.py
```diff
@@ -125,7 +125,7 @@ def generate_policy_event(request, appliance, provider, vm_crud, register_event)
     request.addfinalizer(policy.delete)
 
     policy.assign_events("VM Create Complete")
-    request.addfinalizer(policy.assign_events)
+
```
izapolsk: do you think unassign isn't necessary here ?

widgetastic_manageiq/__init__.py
```diff
@@ -394,18 +394,20 @@ def _values_to_add(self, values):
         return list(set(values) - set(self.selected_text))
 
     def fill(self, values):
-        if set(values) == set(self.selected_text):
-            return False
+        if values:
```
izapolsk: your implementation looks good. but I cannot get why we pass only switch names and change their values to opposite.

I guess we should pass list of switches and their values instead. But this stays apart of this PR

cfme/tests/intelligence/test_reports_with_timelines.py
john-dupuy: In this case, no assign is necessary, since the policy is deleted at the end of the test anyway. You can delete a policy without unassigning events.

widgetastic_manageiq/__init__.py
john-dupuy: Hmm... yeah, I'll think about this for a future PR. I think initially this was done because the switch only has either on or off, so passing it's value is kind of unnecessary.

cfme/control/explorer/policies.py
```diff
@@ -344,12 +345,40 @@ def assign_events(self, *events, **kwargs):
...
-        view.flash.assert_success_message('Policy "{}" was saved'.format(self.description))
+            view.flash.assert_success_message(
+                'Edit of Policy "{}" was cancelled by the user'.format(self.description)
+            )
+
+    def unassign_events(self, *events, **kwargs):
+        """Unassign events from this Policy. """
+        events = list(events)
+        extend = kwargs.pop("extend", False)
+        if extend:
```
mshriver: You can put the `kwargs.pop()` call in-line here since we aren't otherwise using `extend`

cfme/tests/cloud_infra_common/test_events.py
```diff
@@ -73,7 +73,11 @@ def test_vm_create(request, appliance, vm_crud, provider, register_event):
     request.addfinalizer(policy.delete)
 
     policy.assign_events("VM Create Complete")
-    request.addfinalizer(policy.assign_events)
+
+    @request.addfinalizer
+    def _cleanup():
+        policy.unassign_events("VM Create Complete")
```
mshriver: Given the repeated string use, it would be good to have an enumeration module for these event strings. That way, should they have to be version picked, we're only doing it in one place.

Eliminates chances of typos as well.

Not for this PR, but a later enhancement to consider.

widgetastic_manageiq/__init__.py
```diff
@@ -394,18 +394,20 @@ def _values_to_add(self, values):
...
+            before = []
+            after = []
+            for value in values:
+                switch = self.switch_by_text(value)
+                before.append(switch.read())
+                if switch.selected:
```
mshriver: `switch.fill(not switch.selected)`

mshriver: BOOOOOOOOM

john-dupuy: :dancing_men: :+1:

cfme/tests/cloud_infra_common/test_events.py
john-dupuy: Definitely agree here, something to implement in the future.

widgetastic_manageiq/__init__.py
```diff
@@ -394,18 +394,20 @@ def _values_to_add(self, values):
...
+                    switch.fill(False)
+                else:
+                    switch.fill(True)
+                after.append(switch.read())
+            return before != after
```
mshriver: This could be optimized a bit more I think, but there are other considerations for how this should be implemented to fill.

Basically, you could have a sentinel boolean indicating a change, and update that sentinel when before != after. Then you don't have to compile or compare the whole list. Once sentinel is set, it won't get unset, so you can short-circuit the check of before/after anyway and skip the second read.

```
updated = False
for value in values:
    switch = self.switch_by_text(value)
    before = switch.read()
    switch.fill(not switch.selected)
    if not updated and before != switch.read():
        updated = True  # once set, the statement above will short-circuit
return updated
```

PR #8684, opened 2019-04-08
[1LP][RFR] Make BZ references uniform

john-dupuy:
We have many places in tests where we reference BZs by different names. We should choose a uniform name for consistency in our code base. Since, `Bugzilla` seemed to be used the most, I chose that one. 

The scope of this PR is just to make the docblock BZ reference uniform. Later PRs will make use of a script to add meta data, such as `automates` to the test function. 

For reference the proper way to format BZs in the docblock going forward will be:

```python

def test_bugzilla(): 
    """
    Bugzilla:
        <bug_id1>
        <bug_id2>
        <bug_id3>
        ...
    """
    pass
```
Just the number (BZ ID) without a `*` or anything else preceding. I am working on a flake plugin to check this field. Each BZ id related to a test case should appear on its own line.

PR #8730, opened 2019-04-18
[1LP][RFR] Add bugzilla-docstrings to pre-commit

john-dupuy:
Purpose or Intent
=================

This PR adds the flake8 plugin[ `bugzilla-docstrings`](https://pypi.org/project/bugzilla-docstrings/) to our pre-commit config. This plugin makes sure that BZs placed under `Bugzilla:` are just numbers, and that each BZ is on its own line, as described by #8684.  

Note that this plugin does not check and make sure that the `Bugzilla:` field is present in the docstring, since not all tests are written against BZs. It also does not currently look for variations of Bugzilla, such as `BZ`, or `Bugzillas`. It only looks at what is written below `Bugzilla:`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,7 +25,7 @@ repos:
     exclude: ^(sprout|scripts|cfme/fixtures/rdb.py)
   - id: flake8
     language_version: python3.6
-    additional_dependencies: [polarion-docstrings,bugzilla-docstrings]
+    additional_dependencies: [polarion-docstrings, bugzilla-docstrings]
 - repo: https://github.com/asottile/pyupgrade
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -25,7 +25,7 @@ repos:
     exclude: ^(sprout|scripts|cfme/fixtures/rdb.py)
   - id: flake8
     language_version: python3.6
-    additional_dependencies: [polarion-docstrings]
+    additional_dependencies: [polarion-docstrings,bugzilla-docstrings]
```
mshriver: whitespace here please, guess this isn't linted. oh the irony!

PR #8735, opened 2019-04-19
[1LP][RFR] removing unneeded fixture since there is present such global fixture

izapolsk:


mshriver:
PRT failures are not directly related, as the fixtures are identical.

PR #8738, opened 2019-04-22
[1LP][RFR] CB service vm display test

digitronik:
Purpose or Intent
=================
- Trying to satisfy custom button requirements for testing thing on UI and SSUI VM objects
- Added some views and restore old implementation
- Added basic test for service vm objects so i can go for other after this PR. 

{{pytest: cfme/tests/automate/custom_button/test_service_vm_custom_button.py -v}}

### DIFF
--- a/cfme/fixtures/service_fixtures.py
+++ b/cfme/fixtures/service_fixtures.py
@@ -194,9 +194,11 @@ def service_vm(appliance, provider, catalog_item):
     if not provision_request.is_finished():
         pytest.skip("Failed to provision service '{}'".format(catalog_item.name))
 
-    service = MyService(appliance, catalog_item.name, vm_name)
+    service = MyService(appliance, catalog_item.name, vm_name=vm_name)
     yield service, vm
 
     vm.cleanup_on_provider()
     if service.exists:
         service.delete()
+    if provision_request.exists:
+        provision_request.remove_request_rest()
--- a/cfme/fixtures/service_fixtures.py
+++ b/cfme/fixtures/service_fixtures.py
@@ -201,4 +201,4 @@ def service_vm(appliance, provider, catalog_item):
     if service.exists:
         service.delete()
     if provision_request.exists:
-        provision_request.remove_request_rest()
+        provision_request.remove_request(method="rest")
--- a/cfme/services/requests.py
+++ b/cfme/services/requests.py
@@ -185,6 +185,7 @@ class Request(BaseEntity):
             view.breadcrumb.click_location(view.breadcrumb.locations[1], handle_alert=True)
         view.flash.assert_no_error()
 
+    @variable
     def remove_request(self, cancel=False):
         """Opens the specified request and deletes it - removes from UI
         Args:
@@ -193,7 +194,7 @@ class Request(BaseEntity):
         view = navigate_to(self, 'Details')
         view.toolbar.delete.click(handle_alert=not cancel)
 
-    @variable(alias="rest")
+    @remove_request.variant("rest")
     def remove_request_rest(self):
         if "service" in self.rest.request_type:
             request = self.appliance.rest_api.collections.service_requests.find_by(id=self.rest.id)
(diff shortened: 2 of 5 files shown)


digitronik:
> Looks good , one question . Is there a separate test case to execute the custom button ?
@sshveta No... I just trying to make setup for tests.... like navigation and views. After this i need to just copy and past my custom button testing codes. like execute+visibility+dialog+enablement etc... 

I am crating setup for `service vm` testing from my last two PRs.

sshveta:
> > Looks good , one question . Is there a separate test case to execute the custom button ?
> > @sshveta No... I just trying to make setup for tests.... like navigation and views. After this i need to just copy and past my custom button testing codes. like execute+visibility+dialog+enablement etc...
> 
> I am crating setup for `service vm` testing from my last two PRs.

:+1:

cfme/services/requests.py
```diff
@@ -193,6 +193,16 @@ def remove_request(self, cancel=False):
         view = navigate_to(self, 'Details')
         view.toolbar.delete.click(handle_alert=not cancel)
 
+    @variable(alias="rest")
+    def remove_request_rest(self):
```
mshriver: This is not properly implemented for varmeth.

This is not the original implementation of `remove_request`, that already exists on L188.

This should be decorated with `@remove_request.variant('rest')`.

The original `remove_request` method at L188 should be decorated with `@variable(alias='rest')` to register the method with varmeth.

The function name itself is fine, can't have the names match.

cfme/fixtures/service_fixtures.py
```diff
@@ -176,9 +176,11 @@ def service_vm(appliance, provider, catalog_item):
...
 
-    service = MyService(appliance, catalog_item.name, vm_name)
+    service = MyService(appliance, catalog_item.name, vm_name=vm_name)
     yield service, vm
 
     vm.cleanup_on_provider()
     if service.exists:
         service.delete()
+    if provision_request.exists:
+        provision_request.remove_request_rest()
```
mshriver: With a properly implemented varmeth, this should be called as `provision_request.remove_request(method='rest')`

cfme/tests/automate/custom_button/test_service_vm_custom_button.py
```diff
@@ -0,0 +1,119 @@
...
+                * From sidebar select All Catalogs > catalog (created above)
+                * From toolbar select Add New catalog item
+                * Select Provider
+                * Fill Name, description, catalog and dialog (created above)
+                * Select VM name proper template etc...
+            6. Order Catalog
+                * Navigate to Services > Catalogs
+                * Click on Service catalogs from a sidebar
+                * Order catalog from this Page
+        testSteps:
```
mshriver: I don't think its good practice to try and capture the test steps like this in automated test cases.

The level of detail is great for a manual case, but in automation the underlying methods that are achieving these actions through function calls in the test could change, without any mechanism for updating this text.  It becomes a fruitless game of trying to keep the two in sync, so that this description is accurate.

I would prefer no testSteps/expectedResults for automated tests, but am happy to allow for them.  That being said I think this goes much too far in trying to capture details that are not actually specified by the automation called here in the test.

mshriver:
Please expand PRT coverage, since new views/navigation classes are being added to SSUI and UI services modules.

cfme/tests/automate/custom_button/test_service_vm_custom_button.py
digitronik: @mshriver I added these steps for one of the manual tests. where I tried to avoid such details and step by step explanation. Now that work helpful to me just copy and past with small changes :).

This is an unstopable debate... I would like to reduce these. Thanks :+1:

digitronik:
need https://github.com/ManageIQ/integration_tests/pull/8776 for 5.11

cfme/services/requests.py
digitronik: @mshriver thanks for the above comment... I added changes just one thing; L188 l.e. main method `remove_request` implemented with UI so better to decorate with variable only so it will act as the main method and default. Let me know your view.

digitronik:
@mshriver added changes please have a glance...

PR #8741, opened 2019-04-22
[1LP][RFR] Schedule migration refactor

Yadnyawalkya:
{{pytest: cfme/tests/v2v/test_v2v_schedule_migrations.py --use-provider rhv42 --use-provider vsphere67-ims --provider-limit 2 -vvvv}}

@sshveta @digitronik  needs your reviews!

Tested locally and it works. PRT should fail until https://github.com/ManageIQ/integration_tests/pull/8690 gets merge since we are using mini fixture.

### DIFF
--- a/cfme/tests/v2v/test_v2v_schedule_migrations.py
+++ b/cfme/tests/v2v/test_v2v_schedule_migrations.py
@@ -1,83 +1,68 @@
-"""Test to validate basic navigations.Later to be replaced with End-to-End functional testing."""
+"""Tests to validate schedule migration usecases"""
 import fauxfactory
 import pytest
 
-from cfme.fixtures.provider import rhel7_minimal
+from cfme.cloud.provider.openstack import OpenStackProvider
+from cfme.fixtures.v2v_fixtures import get_migrated_vm
 from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.markers.env_markers.provider import ONE_PER_TYPE
 from cfme.markers.env_markers.provider import ONE_PER_VERSION
 from cfme.utils.appliance.implementations.ui import navigate_to
-from cfme.utils.log import logger
-from cfme.utils.wait import wait_for
 
 pytestmark = [
     pytest.mark.provider(
-        classes=[RHEVMProvider],
+        classes=[RHEVMProvider, OpenStackProvider],
         selector=ONE_PER_VERSION,
-        required_flags=['v2v'],
+        required_flags=["v2v"],
         scope="module"
     ),
     pytest.mark.provider(
         classes=[VMwareProvider],
         selector=ONE_PER_TYPE,
-        fixture_name='source_provider',
-        required_flags=['v2v'],
+        fixture_name="source_provider",
+        required_flags=["v2v"],
         scope="module"
-    )
+    ),
+    pytest.mark.usefixtures("v2v_provider_setup")
 ]
 
 
-@pytest.mark.parametrize('form_data_vm_obj_single_datastore', [['nfs', 'nfs', rhel7_minimal]],
- indirect=True)
-def test_single_vm_scheduled_migration(request, appliance, v2v_providers, host_creds,
-                                    conversion_tags, soft_assert,
-                                    form_data_vm_obj_single_datastore):
+@pytest.mark.tier(3)
+def test_schedule_migration(appliance, provider, mapping_data_vm_obj_mini, soft_assert):
     """
+    Test to validate schedule migration plan
+
     Polarion:
-        assignee: kkulkarn
+        assignee: ytale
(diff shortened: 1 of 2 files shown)


digitronik:
@Yadnyawalkya for now moving to `WIPTEST` please fill free to move RFR as PRT clear.

cfme/v2v/migration_plans.py
```diff
@@ -511,8 +511,13 @@ def create(
             'pre_checkbox': pre_checkbox,
             'post_checkbox': post_checkbox
         })
+
+        # Schedule migration check
         view.schedule.wait_displayed()
-        view.schedule.fill(start_migration)
+        start_selection = ["Start migration immediately"
```
sshveta: Required :
start_selection = "Start migration immediately" if start_migration else "Save migration plan to run later"

cfme/tests/v2v/test_v2v_schedule_migrations.py
```diff
@@ -1,83 +1,78 @@
...
+    view = navigate_to(migration_plan_collection, "NotStarted")
+    view.plans_not_started_list.schedule_migration(migration_plan.name)
+    soft_assert("Migration scheduled" in view.plans_not_started_list.get_clock(migration_plan.name))
 
-    # wait until plan is in progress
-    wait_for(func=view.plan_in_progress, func_args=[migration_plan.name],
-        message="migration plan is in progress, be patient please",
-        delay=5, num_sec=1800)
+    # Extra wait_for until schedule time
+    view = navigate_to(migration_plan, "InProgress")
```
sshveta: Required : 
 assert migration_plan.plan_started 
instead of  view = navigate_to(migration_plan, "InProgress")

cfme/tests/v2v/test_v2v_schedule_migrations.py
```diff
@@ -1,83 +1,78 @@
...
+    wait_for(
```
sshveta: This is not required with migration_plan.plan_started

cfme/tests/v2v/test_v2v_schedule_migrations.py
Yadnyawalkya: Good idea, tested locally, it works!

cfme/tests/v2v/test_v2v_schedule_migrations.py
```diff
@@ -1,83 +1,68 @@
...
 
-    view.switch_to("Completed Plans")
-    view.wait_displayed()
-    migration_plan_collection.find_completed_plan(migration_plan)
-    logger.info("For plan %s, migration status after completion: %s, total time elapsed: %s",
-        migration_plan.name,
-        view.migration_plans_completed_list.get_vm_count_in_plan(migration_plan.name),
-        view.migration_plans_completed_list.get_clock(migration_plan.name))
-    assert view.migration_plans_completed_list.is_plan_succeeded(migration_plan.name)
+    assert migration_plan.plan_started
```
izapolsk: python properties shouldn't wait for some state. it has to be done by methods. it is not obvious. please turn such properties into methods like 
```migration_play.wait_play_started```

Comment (izapolsk) [reply cfme/tests/v2v/test_v2v_schedule_migrations.py]: @mshriver, could you share your thoughts on above ?

Comment (Yadnyawalkya) [reply cfme/tests/v2v/test_v2v_schedule_migrations.py]: > python properties shouldn't wait for some state. it has to be done by methods. it is not obvious. please turn such properties into methods like
> `migration_play.wait_play_started`

properties like `wait_play_started` totally make sense though, we need to change our existing tests for this so asking @sshveta before changing it.

**Comment by Yadnyawalkya** [comment]
@izapolsk  PRT should fail now since we got new method of enabling conversion host (For 5.11.0.1 method is in UI but buggy, for 5.10.4 we should get new with fixes in 5.10.4.1).

File: cfme/tests/v2v/test_v2v_schedule_migrations.py (hunk 4 above)
Comment (mshriver) [reply cfme/tests/v2v/test_v2v_schedule_migrations.py]: We discussed this offline, but I don't think @Yadnyawalkya was on the call.

I agree completely with @izapolsk, its unexpected behavior when calling a property, and creates strange patterns in tests.

**Comment by mshriver** [comment]
5.4.0.1 and 5.11.0.2 are available internally, I'll restart PRT when both will be used.

File: cfme/v2v/migration_plans.py [hunk 5]
```diff
@@ -479,7 +479,7 @@ def create(
             post_playbook: (string) post-migration playbook name
             pre_checkbox: checkbox premigration playbook
             post_checkbox: post migration checkbox
-            start_migration: (string) text of radio button to choose
+            start_migration: (bool) flag to start migration
```
mshriver: Great improvement here.

mshriver:
The test is failing on navigation, `Could not do Waiting for view [AllInfraMappingView] to display`

PR #8742, opened 2019-04-23
[1LP][RFR] Automate test: test_rest_metric_rollups

valaparthvi:
Changes:
1. Automate test: test_rest_metric_rollups - not enabling candu, because we only want to check if the endpoint works and does not give any error.
2. Move `test_settings_collections` to `test_query_simple_collections`, and also test `automate_workspaces` collections which wasn't tested earlier.

{{ pytest: cfme/tests/test_rest.py -k "test_query_simple_collections or test_rest_metric_rollups" --use-template-cache -sqvvv }}

### DIFF
--- a/cfme/tests/test_rest.py
+++ b/cfme/tests/test_rest.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 """This module contains REST API specific tests."""
 import random
+from datetime import datetime
+from datetime import timedelta
 
 import fauxfactory
 import pytest
@@ -175,14 +177,12 @@ def _collection_not_in_this_version(appliance, collection_name):
 @pytest.mark.parametrize("collection_name", COLLECTIONS_ALL)
 @pytest.mark.uncollectif(
     lambda appliance, collection_name:
-        collection_name in COLLECTIONS_OMITTED or
+        collection_name == "metric_rollups" or
         _collection_not_in_this_version(appliance, collection_name)
 )
 def test_query_simple_collections(appliance, collection_name):
     """This test tries to load each of the listed collections. 'Simple' collection means that they
     have no usable actions that we could try to run
-    Steps:
-        * GET /api/<collection_name>
     Metadata:
         test_flag: rest
 
@@ -191,11 +191,22 @@ def test_query_simple_collections(appliance, collection_name):
         casecomponent: Rest
         caseimportance: high
         initialEstimate: 1/3h
+        testSteps:
+            1. Send a GET request: /api/<collection_name>
+        expectedResults:
+            1. Must receive a 200 OK response.
     """
     collection = getattr(appliance.rest_api.collections, collection_name)
-    assert_response(appliance)
-    collection.reload()
-    list(collection)
+
+    if collection_name in COLLECTIONS_OMITTED:
+        # the "settings" and automate_workspaces collections are untypical
+        # as they don't have "resources" and for this reason can't be reloaded (bug in api client)
+        appliance.rest_api.get(collection._href)
+        assert_response(appliance)
+    else:
+        assert_response(appliance)
+        collection.reload()
+        list(collection)
 
 
 @pytest.mark.tier(3)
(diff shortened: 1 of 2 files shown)


dajoRH:
Lint report for commit b9fbb39db942707db8690cd1b79dcee62b37ed12:

`cfme/tests/test_rest.py`:
- :red_circle: Line 202:9: **F841** *local variable 'body' is assigned to but never used*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/test_rest.py
```diff
@@ -175,14 +177,12 @@ def _collection_not_in_this_version(appliance, collection_name):
 @pytest.mark.parametrize("collection_name", COLLECTIONS_ALL)
 @pytest.mark.uncollectif(
     lambda appliance, collection_name:
-        collection_name in COLLECTIONS_OMITTED or
+        collection_name == "metric_rollups" or
```
john-dupuy: @valaparthvi Can you get rid of `COLLECTIONS_IN_59` or change it to `COLLECTIONS_IN_510`?

cfme/tests/test_rest.py
```diff
@@ -1119,3 +1112,37 @@ def test_find_created_events(self, appliance, vm_obj, gen_events, provider, soft
                 evt_col.get(id=found_evts[-1].id)
             except (IndexError, ValueError):
                 soft_assert(False, "Couldn't get event {} for vm {}".format(evt, vm_name))
+
+
+@pytest.mark.tier(3)
+@pytest.mark.parametrize("interval", ["hourly", "daily"])
+@pytest.mark.parametrize("resource_type", ["VmOrTemplate", "Service"])
```
john-dupuy: Optional, the double parametrization can be done in a single statement:

`@pytest.mark.parametrize(("param1, param2"), [("value11", "value12"), ("value21", "value22")]`)

cfme/tests/test_rest.py
valaparthvi: I was thinking of doing it, but then we will get a 5.11 build in the future, and there might be some collections which are a part of 5.11, but not 5.10, so I thought I'd make changes to it, once we have 5.11, in case they're required/

cfme/tests/test_rest.py
valaparthvi: I think parametrizing this way will not give me the combinations that I require.
I need 4 combinations - Service-hourly, Service-weekly, VmOrTemplate-hourly and VmOrTemplate-weekly
Using 2 parametrize statements helps in getting the combination.

cfme/tests/test_rest.py
```diff
@@ -191,11 +191,22 @@ def test_query_simple_collections(appliance, collection_name):
...
+            1. Must receive a 200 OK response.
     """
     collection = getattr(appliance.rest_api.collections, collection_name)
-    assert_response(appliance)
-    collection.reload()
-    list(collection)
+
+    if collection_name in COLLECTIONS_OMITTED:
+        # the "settings" and automate_workspaces collections are untypical
+        # as they don't have "resources" and for this reason can't be reloaded (bug in api client)
```
digitronik: Q: we have Bugzilla for it? If yes, why we are blocking?

valaparthvi: The bug is in the manageiq_client library, might not be a bug, but the feature has probably not been implemented in the library.

Collections not in `COLLECTION_OMITTED` can be easily accessed with `appliance.rest_api.collections.<collection_name>.all` since there is a `resource` attribute in the response, but we cannot access omitted collections such as `settings` and `automate_workspaces` as easily, since there is no `resource` attribute in the response.

Test this: `GET /api/providers` and `GET /api/automate_workspaces`, observe both the responses and look for `resources` attribute.

dajoRH:
*CFME QE Bot*

cfme/tests/test_rest.py
digitronik: @valaparthvi I think better to suggestion from devel for this.

PR #8746, opened 2019-04-23
[1LP][RFR] Add in a wait_for LogValidator

john-dupuy:
Adding a simple `wait_for_log_validation` for the `LogValidator`. This can be used in the following way, 

```python,
from cfme.utils.log_validator import LogValidator
test = LogValidator("/root/test.log", matched_patterns=[".*blah"])
test.fix_before_start()
test.wait_for_log_validation(timeout=300, delay=1)
# make a file on appliance with "[----] blah" on the line and we should find it
```

### DIFF
--- a/cfme/utils/log_validator.py
+++ b/cfme/utils/log_validator.py
@@ -84,7 +84,9 @@ class LogValidator(object):
             if pattern not in self.matches:
                 pytest.fail('Expected pattern {} did not match'.format(pattern))
 
-    def wait_for_log_validation(self, **kwargs):
+    def wait_for_log_validation(
+            self, delay=5, num_sec=180, message="waiting for log validation", **kwargs
+    ):
         """ Wait for log validation, takes the kwargs as wait_for. This function will reduce
             duplicate functions in tests that wait_for log_validation. It is necessary to create
             this function since _verify_match_logs raise pytest.fail() when it fails to find the
@@ -98,4 +100,4 @@ class LogValidator(object):
                 return True
             except Failed:
                 return False
-        wait_for(validate, **kwargs)
+        wait_for(validate, delay=delay, num_sec=num_sec, message=message, **kwargs)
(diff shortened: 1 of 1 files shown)


cfme/utils/log_validator.py
```diff
@@ -81,3 +83,14 @@ def _verify_match_logs(self):
         for pattern in self.matched_patterns:
             if pattern not in self.matches:
                 pytest.fail('Expected pattern {} did not match'.format(pattern))
+
+    def wait_for_log_validation(self, **kwargs):
```
jaryn: Cannot one just use

`wait_for(log_validator.validate_logs, handle_exception=True)` instead of using the `wait_for_log_validation`?

john-dupuy: @jaryn That's what I thought at first too, however, if you look at `_verify_match_logs()`, it raises `pytest.fail` if there are no matches, so we must catch that exception and have our function raise `True` or `False` in order to use wait for.

lina-nikiforova: @john-dupuy I see. Maybe worth adding some comments to your function about it?

lina-nikiforova:
PRT failure is not related to the changes

lina-nikiforova:
@john-dupuy there's a newline, I was confused by the way the file is displayed here. Sorry for confusing you as well :)

cfme/utils/log_validator.py
```diff
@@ -81,3 +83,19 @@ def _verify_match_logs(self):
...
+
+            Note that you must call fix_before_start() before making use of this function.
+        """
+        def validate():
+            try:
+                self.validate_logs()
+                return True
+            except Failed:
+                return False
+        wait_for(validate, **kwargs)
```
jawatts: **Required** What do you think about passing a default `delay`, `num_sec`, and a `message`?

PR #8747, opened 2019-04-23
[RFR] fixed diagnostics & logs views and relevant tests

izapolsk:
{{pytest: --long-running cfme/tests/configure/test_log_depot_operation.py}}

### DIFF
--- a/cfme/configure/configuration/diagnostics_settings.py
+++ b/cfme/configure/configuration/diagnostics_settings.py
@@ -2,6 +2,7 @@ import attr
 from navmazing import NavigateToAttribute
 from navmazing import NavigateToSibling
 from widgetastic.exceptions import RowNotFound
+from widgetastic.utils import WaitFillViewStrategy
 from widgetastic.widget import Text
 from widgetastic.widget import View
 from widgetastic_patternfly import BootstrapSelect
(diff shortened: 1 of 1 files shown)


cfme/configure/configuration/diagnostics_settings.py
```diff
@@ -219,14 +220,14 @@ def update(self, updates, cancel=False, reset=False):
         if reset:
             view.reset_button.click()
             flash_message = 'All changes have been reset'
-            flash_view = ServerCollectLogsEditView
```
mshriver: For the record (merging this), I don't quite like this pattern of setting these view attributes for use in flash message validation.  We prefer to have these flash message checks in the CRUD test for related collection/entity - and the flash message is available by using a generic `BaseLoggedInView`. 

Ultimately standardizing that pattern is an enhancement on the framework overall.

cfme/configure/configuration/diagnostics_settings.py
```diff
@@ -425,18 +430,21 @@ def is_displayed(self):
 
 
 class CollectLogsBasicEntities(View):
+    fill_strategy = WaitFillViewStrategy("5s")
```
mshriver: :tada: :heart: :100:

cfme/configure/configuration/diagnostics_settings.py
izapolsk: I entirely agree

PR #8749, opened 2019-04-23
[1LP][RFR] Retry ssh connection if failed.

jaryn:
When ssh connection to additional appliances (for example in the update
tests) is being made, it could fail, because SSH was not fully up and
running.

This patch adds re-trying to create the ssh connections for 2 minutes.

The reraise was used to log-and-ignore the exceptions when running an
SSH command, but this looks to me as a bad practice as it just burries
the problems in the logs.

{{ py.test: -v -m smoke -k "not test_cloud_provider_crud" }}

### DIFF
--- a/cfme/utils/appliance/__init__.py
+++ b/cfme/utils/appliance/__init__.py
@@ -986,7 +986,7 @@ ExecStartPre=/usr/bin/bash -c "ipcs -s|grep apache|cut -d\  -f2|while read line;
         logger.debug('Waiting for SSH to %s to become connective.',
                      self.hostname)
         self.wait_for_ssh()
-        logger.debug('SSH to %s ready.', self.hostname)
+        logger.debug('SSH port on %s ready.', self.hostname)
 
         # IPAppliance.ssh_client only connects to its address
         if self.openshift_creds:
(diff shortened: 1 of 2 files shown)


cfme/utils/appliance/__init__.py
```diff
@@ -1012,17 +1012,25 @@ def ssh_client(self):
...
+
+        def create_ssh_connection():
+            ssh_client = ssh.SSHClient(**connect_kwargs)
+            try:
+                ssh_client.get_transport().is_active()
+                logger.info('default appliance ssh credentials are valid')
+            except Exception as e:
+                if self.is_dev:
+                    raise Exception('SSH access on a dev alliance, (unsupported)')
+                logger.error(e)
```
mshriver: This can be cleanly logged with `logger.exception('default appliance ssh credentials failed, establish ssh connection using ssh private key')`.  Can clean that up while we're making changes here anyway :)

cfme/utils/appliance/__init__.py
```diff
@@ -1012,17 +1012,25 @@ def ssh_client(self):
...
-            ssh_client = self.ssh_client_with_privatekey()
+
+        def create_ssh_connection():
+            ssh_client = ssh.SSHClient(**connect_kwargs)
+            try:
+                ssh_client.get_transport().is_active()
+                logger.info('default appliance ssh credentials are valid')
+            except Exception as e:
+                if self.is_dev:
+                    raise Exception('SSH access on a dev alliance, (unsupported)')
```
mshriver: This is a typo in the original message anyway - should be `'SSH access unsupported on dev appliance'`

mshriver: It also makes more sense to check this before even calling wait_for on this method, since we don't expect it to work in the first place?

jaryn: @mshriver I don't know. I guess the check is in the `except` for some reason. Perhaps we want to shoot the warning only when things go wrong.

cfme/utils/appliance/__init__.py
```diff
@@ -1012,17 +1012,25 @@ def ssh_client(self):
...
+                logger.exception(
+                    ('default appliance ssh credentials failed: {}, '
+                     'trying establish ssh connection using ssh private key').format(e))
+                ssh_client.close()
+
+                ssh_client = self.ssh_client_with_privatekey()
+            assert ssh_client.run_command('true').success
+            return ssh_client
+
+        ssh_client = wait_for(func=create_ssh_connection, delay=5, timeout=120).out
```
izapolsk: I don't see handing assert

mshriver: There's some handling of the exceptions within `create_ssh_connection`, so I think this wait-retry is still functional, just limited in scope for what will be handled and what will immediately fail.

PR #8750, opened 2019-04-23
[RFR] Remove tests automated in test_relationships for SCVMM

john-dupuy:
Purpose or Intent
=================
Removing some of the basic manual cases that are automated by the `test_relationships` tests for SCVMM. I am using this PR to test the relationships tests on PRT. Also making a small change for `test_no_dvd_ruins_refresh`

{{ pytest: --use-provider ms_scvmm --use-template-cache --long-running cfme/tests/cloud_infra_common/test_relationships.py cfme/tests/infrastructure/test_scvmm_specific.py }}

### DIFF
--- a/cfme/tests/infrastructure/test_scvmm_specific.py
+++ b/cfme/tests/infrastructure/test_scvmm_specific.py
@@ -14,18 +14,18 @@ pytestmark = [
 
 
 @pytest.fixture
-def testing_vm_without_dvd(provider, small_template):
-    vm_name = "test_no_dvd_{}".format(fauxfactory.gen_alpha())
+def vm(provider, small_template):
+    vm_name = "test-scvmm-{}".format(fauxfactory.gen_alpha())
     vm = provider.appliance.collections.infra_vms.instantiate(
-        vm_name, provider, small_template.name)
-    vm.create_on_provider()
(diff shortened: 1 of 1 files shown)


john-dupuy:
One test failing which has a FIXME, other than that PRT is green.

cfme/tests/infrastructure/test_scvmm_specific.py
```diff
@@ -14,18 +14,23 @@
...
+        vm_name, provider, small_template.name
+    )
+    vm.create_on_provider(find_in_cfme=True)
     yield vm
     vm.cleanup_on_provider()
 
 
+@pytest.fixture
+def remove_dvd_from_vm(vm):
+    vm.mgmt.disconnect_dvd_drives()
```
digitronik: Q:
It will work as a single use... but a fixture has function scope so I'm suspect that if in future you used the same fixture for other tests then may you will get an error.

I don't know how `wrapanapi` handle this but I think one check is a better option (already disconnected or not).

john-dupuy: Thanks Nikhil, on second thought this doesn't really need to be a fixture, it is very specific to the `test_no_dvd_ruins_refresh` test, so I'll put it in there.

PR #8751, opened 2019-04-23
[RFR] Automate test case for testing scvmm mac address

john-dupuy:
Purpose or Intent
=================
Automating `test_vm_mac_scvmm` 

{{ pytest: --use-provider complete --long-running --use-template-cache cfme/tests/infrastructure/test_scvmm_specific.py::test_vm_mac_scvmm }}

### DIFF
--- a/cfme/tests/infrastructure/test_scvmm_specific.py
+++ b/cfme/tests/infrastructure/test_scvmm_specific.py
@@ -4,6 +4,7 @@ import pytest
 
 from cfme import test_requirements
 from cfme.infrastructure.provider.scvmm import SCVMMProvider
+from cfme.utils.appliance.implementations.ui import navigate_to
 
 
 pytestmark = [
@@ -38,14 +39,13 @@ def test_no_dvd_ruins_refresh(provider, vm):
     vm.wait_to_appear()
 
 
-@pytest.mark.manual
 @pytest.mark.tier(1)
-def test_vm_mac_scvmm():
+def test_vm_mac_scvmm(provider):
     """
     Bugzilla:
         1514461
 
-    Test case covers this BZ - we can"t get MAC ID of VM at the moment
+    Test case covers this BZ - we can't get MAC ID of VM at the moment
 
     Polarion:
         assignee: jdupuy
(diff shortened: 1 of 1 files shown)


cfme/tests/infrastructure/test_scvmm_specific.py
```diff
@@ -82,7 +81,10 @@ def test_vm_mac_scvmm():
             1.
             2. MAC Address should match what is in SCVMM
     """
-    pass
+    collection = provider.appliance.provider_based_collection(provider)
+    vm = collection.all()[0]
+    # compare mac_address in CFME to that from SCVMM
+    assert vm.mac_address == vm.mgmt.raw["VirtualNetworkAdapters"][0]["PhysicalAddress"]
```
jaryn: I don't like these `[0]`. If there are more vms, the test can start false-failing.

john-dupuy: @jaryn Why would this fail with more vms? It will just take the first one. Every provider has at least `cu-24x7` on it. For the `VirtualNetworkAdapters`, most VMs have only one of those, but the API returns a list, so we take the first one.

jaryn: @john-dupuy what makes you think api returns objects in same order as web-ui?

john-dupuy: @jaryn Good point, I can modify this to get a list of mac addresses and ensure that the one in the web-ui is in that list.

cfme/tests/infrastructure/test_scvmm_specific.py
```diff
@@ -82,7 +81,11 @@ def test_vm_mac_scvmm():
             1.
             2. MAC Address should match what is in SCVMM
     """
-    pass
+    collection = provider.appliance.provider_based_collection(provider)
+    vm = collection.all()[0]
+    # compare mac_address in CFME to that from SCVMM
+    mac_addresses = [entry["PhysicalAddress"] for entry in vm.mgmt.raw["VirtualNetworkAdapters"]]
+    assert vm.mac_address in mac_addresses
```
mshriver: There is an inherit assumption here being made, that the `mac_address` property is supplied via a UI lookup.

Reading the original testSteps, its intended that the address from the UI is compared to what's in SCVMM itself.

I think it would be better to directly navigate to the `vm` 'Details' page and read its `SummaryTable` for the MAC, then look for _that_ within the `mac_addresses` list.

A TODO here with an issue in wrapanapi to support getting the MAC for specific VM's would be great. That way we don't have this loose check, and can positively assert that a specific VM's MAC address is mapped correctly.

john-dupuy: @mshriver that is because of https://github.com/ManageIQ/integration_tests/blob/f0dd9b215a687cb75c8b939ab3108403c05eec3e/cfme/common/vm.py#L191, i.e. the mac_address is obtained via a UI lookup already. And then we are looking for that in the mac_addresses list (which is obtained from SCVMM). Given the `TODO` of the `mac_address` property in the `cfme/common/vm.py`, I can just explicitly do what that function does here, to avoid having to do it later when the `TODO` is fulfilled.

PR #8753, opened 2019-04-24
[1LP][RFR] CB catalog crud test and small refactor. 

digitronik:
Purpose or Intent
=================
- Adding custom button on catalog crud test
- Moved `AutomateRadioGroup` to `miq.wt` so we can share it in other files.
- Added some refactor in view; I would like catalog button params should similar to main custom button model page.
- Added methods for button; for group already we have.
- Added one workaround; as i don't want to guess flash msg.

### DIFF
--- a/cfme/automate/buttons.py
+++ b/cfme/automate/buttons.py
@@ -25,10 +25,10 @@ from cfme.utils.appliance.implementations.ui import CFMENavigateStep
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.appliance.implementations.ui import navigator
 from cfme.utils.update import Updateable
+from widgetastic_manageiq import AutomateRadioGroup
 from widgetastic_manageiq import FonticonPicker
 from widgetastic_manageiq import MultiBoxOrderedSelect
 from widgetastic_manageiq import PotentiallyInvisibleTab
-from widgetastic_manageiq import RadioGroup
 from widgetastic_manageiq import SummaryFormItem
 from widgetastic_manageiq.expression_editor import ExpressionEditor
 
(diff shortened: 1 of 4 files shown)


digitronik:
PRT failures will fixed in https://github.com/ManageIQ/integration_tests/pull/8755

widgetastic_manageiq/__init__.py
```diff
@@ -5349,3 +5349,24 @@ def fill(self, value):
...
+
+    @property
+    def selected(self):
+        names = self.button_names
+        for name in names:
+            bttn = self.browser.element(self.BUTTON.format(quote(name)))
+            if bttn.get_attribute("checked") is not None:
+                return name
+        else:
+            return names[0]
```
jaryn: I see you have been just moving this piece of code, but it doesn't seem to do what I would expect. I find strange it returns `names[0]`if it finds no item is selected.

digitronik: aaaaaaahhh thanks @jaryn . I just moved but thank need to check it. :+1:

cfme/automate/buttons.py
```diff
@@ -54,27 +53,6 @@
 ]
 
 
-class AutomateRadioGroup(RadioGroup):
```
lina-nikiforova: Are you sure it needs to be removed?
I see it's used in the same file in `class ButtonFormAnsibleView(View)`
`inventory = AutomateRadioGroup(locator=".//input[@name='inventory']/..")`

lina-nikiforova: ok, I see it's moved to ` widgetastic_manageiq/__init__.py`

cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -79,38 +84,60 @@ class modal(View):  # noqa
...
-    class options(WaitTab):    # noqa
-        TAB_NAME = 'Options'
-
-        btn_text = Input(name='name')
-        btn_hvr_text = Input(name='description')
-        select_dialog = BootstrapSelect('dialog_id')
-        btn_image = FonticonPicker('button_icon')
+    class options(WaitTab):  # noqa
+        btn_type = BootstrapSelect("button_type")
+        form = ConditionalSwitchableView(reference="type")
```
mshriver: Shouldn't this reference be 'btn_type', given that's the name of the `BootstrapSelect` widget above?

digitronik: It's my bad; you are not missing. I renamed `type` to `btn_type` to protect `type` default method and miss this. :) **The Eagle Eyes**

PR #8754, opened 2019-04-24
[1LP][RFR] Automate config manager tests for REST

valaparthvi:
Changes:
1. Add config_manager CRUD tests for REST
2. Add `create_rest`, `delete_rest` and `rest_api_entity` methods for Config Manager

{{ pytest: cfme/tests/infrastructure/test_config_management_rest.py -k "test_config_manager_create or test_config_manager_edit or test_config_manager_delete_details" --use-template-cache -sqvvv }}

### DIFF
--- a/cfme/infrastructure/config_management.py
+++ b/cfme/infrastructure/config_management.py
@@ -1,4 +1,6 @@
 # -*- coding: utf-8 -*-
+from manageiq_client.api import APIException
+from manageiq_client.api import Entity as RestEntity
 from navmazing import NavigateToAttribute
 from navmazing import NavigateToSibling
 from widgetastic.exceptions import NoSuchElementException
@@ -22,6 +24,7 @@ from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.appliance.implementations.ui import navigator
 from cfme.utils.log import logger
 from cfme.utils.pretty import Pretty
+from cfme.utils.rest import assert_response
 from cfme.utils.update import Updateable
 from cfme.utils.version import LATEST
 from cfme.utils.version import LOWEST
@@ -400,6 +403,74 @@ class ConfigManager(Updateable, Pretty, Navigatable):
                 )
                 # check the provider is indeed deleted
                 assert not self.exists
+
+    @property
+    def rest_api_entity(self):
+        """Returns the rest entity of config manager provider"""
+        # Since config manager provider is slightly different from other normal providers,
+        # we cannot obtain it's rest entity the normal way, instead, we use Entity class
+        # of manageiq_client library to instantiate the rest entity.
+        provider_id = self.appliance.rest_api.collections.providers.get(
+            name=self.ui_name
+        ).provider_id
+
+        return RestEntity(
+            self.appliance.rest_api.collections.providers,
+            data={
+                "href": self.appliance.url_path(
+                    "/api/providers/{}?provider_class=provider".format(provider_id)
+                )
+            },
+        )
+
+    def create_rest(self):
+        """Create the config manager in CFME using REST"""
+        if "ansible_tower" in self.key:
+            config_type = "AnsibleTower"
+        elif "satellite" in self.key:
+            config_type = "Foreman"
+
+        payload = {
+            "type": "ManageIQ::Providers::{}::Provider".format(config_type),
+            "url": self.url,
+            "name": self.name,
+            "verify_ssl": self.ssl,
+            "credentials": {
+                "userid": self.credentials.view_value_mapping["username"],
+                "password": self.credentials.view_value_mapping["password"],
+            },
+        }
+        try:
+            self.appliance.rest_api.post(
+                api_endpoint_url=self.appliance.url_path(
+                    "/api/providers/?provider_class=provider"
+                ),
+                **payload
+            )
+        except APIException as err:
+            raise AssertionError("Provider wasn't added: {}".format(err))
+
+        response = self.appliance.rest_api.response
+        if not response:
+            raise AssertionError(
+                "Provider wasn't added, status code {}".format(response.status_code)
+            )
+
+        assert_response(self.appliance)
+        return True
+
+    def delete_rest(self):
+        """Deletes the config manager from CFME using REST"""
+        try:
+            self.rest_api_entity.action.delete()
+        except APIException as err:
+            raise AssertionError("Provider wasn't deleted: {}".format(err))
+
+        response = self.appliance.rest_api.response
+        if not response:
+            raise AssertionError(
+                "Provider wasn't deleted, status code {}".format(response.status_code)
+            )
 
     @property
     def exists(self):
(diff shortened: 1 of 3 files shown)


cfme/tests/infrastructure/test_config_management_rest.py
```diff
@@ -230,3 +231,74 @@ def test_authentications_options(self, appliance, config_manager):
         collection = appliance.rest_api.collections.authentications
         assert 'credential_types' in collection.options()['data']
         assert_response(appliance)
+
+
+@pytest.fixture(scope="function")
```
nachandr: Any particular reason why this fixture is marked function scoped and not module scoped (considering that all the tests use this fixture) ?

cfme/tests/infrastructure/test_config_management_rest.py
```diff
@@ -230,3 +231,74 @@ def test_authentications_options(self, appliance, config_manager):
...
+def test_config_manager_edit(request, config_manager_rest):
+    """Test editing a config manager using REST API.
+
+    Metadata:
+        test_flag: rest
+
+    Polarion:
+        assignee: pvala
+        casecomponent: Rest
+        caseimportance: high
```
nachandr: Another observation , but otherwise the PR looks good.

The test_config_manager_edit test has been marked to be of high importance , whereas the remaining tests are marked to be of medium importance.

cfme/tests/infrastructure/test_config_management_rest.py
```diff
@@ -230,3 +231,74 @@ def test_authentications_options(self, appliance, config_manager):
...
+        casecomponent: Rest
+        caseimportance: medium
+        initialEstimate: 1/10h
+        testSteps:
+            1. Add Ansible Tower via REST
+        expectedResults:
+            1. Provider must be added and validated successfully.
+
+    Bugzilla:
+        1621888
```
nachandr: This BZ reference could be removed now that the BZ is closed.

dajoRH:
*CFME QE Bot*

cfme/tests/infrastructure/test_config_management_rest.py
valaparthvi: I have kept it as metadata, just for a reference.

cfme/tests/infrastructure/test_config_management_rest.py
valaparthvi: There is no particular reason, I required the config manager to be created for every test, so I was using function scope, but I guess using module scope doesn't make much difference either. Will change the scope.

dajoRH:
*CFME QE Bot*

cfme/infrastructure/config_management.py
```diff
@@ -401,6 +404,75 @@ def delete(self, cancel=False, wait_deleted=True, force=False):
...
+            data={
+                "href": self.appliance.url_path(
+                    "/api/providers/{}?provider_class=provider".format(provider_id)
+                )
+            },
+        )
+
+    def create_rest(self):
+        """Create the config manager in CFME using REST"""
+        if "ansible_tower" in self.key:
```
digitronik: Q: is there any better way than this check?... suspect about `name` or `key` they can be different as per me.

cfme/tests/infrastructure/test_config_management_rest.py
```diff
@@ -230,3 +231,66 @@ def test_authentications_options(self, appliance, config_manager):
...
+
+    yield rest_entity
+
+    if rest_entity.exists:
+        rest_entity.action.delete()
+        rest_entity.wait_not_exists()
+
+
+@pytest.mark.tier(3)
+def test_config_manager_create(config_manager_rest):
```
digitronik: Please add your rest requirement.

cfme/infrastructure/config_management.py
valaparthvi: I checked different properties of the config manager and `key` seemed like the best bet. But we could use `config_manager.yaml_data['type']` if not this.

cfme/tests/infrastructure/test_config_management_rest.py
valaparthvi: We have a requirement `tower` in pytestmark, do we need to add rest requirement there? And if yes, we'll have to add it to the whole file.

digitronik: Its depends on you. Decided this test you want to put under `tower` FA or `rest`. Global mark will mark it as `tower` else add local marker.

cfme/tests/infrastructure/test_config_management_rest.py
```diff
@@ -230,3 +231,69 @@ def test_authentications_options(self, appliance, config_manager):
...
+    config_manager_obj.create_rest()
+    assert_response(config_manager_obj.appliance)
+    rest_entity = config_manager_obj.rest_api_entity
+    rest_entity.reload()
+
+    yield rest_entity
+
+    if rest_entity.exists:
+        rest_entity.action.delete()
+        rest_entity.wait_not_exists()
```
digitronik: optional:
but not a good idea to wait for delete in tear-down... it if fails then actual test will fail.

valaparthvi: On 5.10, tower takes a little time to get deleted, due to which the one of the test fails. Need to use `wait_not_exists`.

cfme/tests/infrastructure/test_config_management_rest.py
```diff
@@ -230,3 +231,69 @@ def test_authentications_options(self, appliance, config_manager):
...
+    Bugzilla:
+        1621888
+    """
+    assert "?provider_class=provider" in config_manager_rest.href
+    assert "::Provider" in config_manager_rest.type
+
+
+@pytest.mark.tier(3)
+@test_requirements.rest
+def test_config_manager_edit(request, config_manager_rest):
```
mshriver: I think you'll end up with a duplicate test name here from test_config_management.py.  I think your test function names here should include `_rest`

cfme/tests/infrastructure/test_config_management_rest.py
```diff
@@ -230,3 +231,69 @@ def test_authentications_options(self, appliance, config_manager):
...
+    yield rest_entity
+
+    if rest_entity.exists:
+        rest_entity.action.delete()
+        rest_entity.wait_not_exists()
+
+
+@pytest.mark.tier(3)
+@test_requirements.rest
+def test_config_manager_create(config_manager_rest):
```
mshriver: Can we collapse all three of these test cases into a single `test_config_manager_rest_crud()` function please?

valaparthvi: I was going to do that, but then we need to test delete operation with POST and DELETE, so I had to keep that test case separate, and keeping create and update methods in a single test seemed weird, so I decided to keep these tests separate. I can merge create and update if you think that'd be nicer.

PR #8758, opened 2019-04-24
[1LP][RFR] Update cluster property to fetch info through REST

nachandr:
{{ pytest: cfme/tests/infrastructure/test_timelines.py }}

It looks like this is the only usage of vm_object.cluster.So, I'd like this test to be run on PRT against the new PR.

PRT results:
I'm still waiting on PRT results since Sprout is unable to provision an appliance at the moment.
I have tested this PR locally using this code segment .
http://pastebin.test.redhat.com/758518

With these code changes, the relevant information is fetched via REST API and not UI ; Using REST API speeds up the process .

### DIFF
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -982,8 +982,13 @@ class InfraVm(VM):
 
     @property
     def cluster(self):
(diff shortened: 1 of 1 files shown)


digitronik:
Tested Locally... working fine.

PR #8768, opened 2019-04-29
[1LP][RFR] fixed is_displayed for ServerCollectLogsView

lina-nikiforova:
{{ pytest: cfme/tests/configure/test_log_depot_operation.py -qsvvv --long-running }}

the previous solution didn't work if the server's number is more than 1 digit


lina-nikiforova:
PRT failure is not related to this change.
The cause is UI change:
5.10:
image

5.11:
image

izapolsk:
restarted PRT against whole test suite in order to see full pictrure

PR #8771, opened 2019-04-30
[1LP][RFR] New Test: Testing automate method of sending emails

ganeshhubale:
- This test case checks whether while sending mails automate methods considers person in cc and bcc.

{{ pytest: cfme/tests/automate/test_method.py::test_send_email_method -vv }}

### DIFF
--- a/cfme/tests/automate/test_method.py
+++ b/cfme/tests/automate/test_method.py
@@ -6,6 +6,7 @@ from cfme import test_requirements
 from cfme.automate.explorer.klass import ClassDetailsView
 from cfme.automate.simulation import simulate
 from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.blockers import BZ
 from cfme.utils.log_validator import LogValidator
 from cfme.utils.update import update
 from cfme.utils.wait import wait_for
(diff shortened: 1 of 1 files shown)


cfme/tests/automate/test_method.py
```diff
@@ -264,3 +265,96 @@ def test_task_id_for_method_automation_log(request, generic_catalog_item):
...
+                \nto = "{mail_to}"
+                \nsubject = "Hello"
+                \nbody = "Hi"
+                \nbcc = "{mail_bcc}"
+                \ncc = "{mail_cc}"
+                \ncontent_type = "message"
+                \nfrom = "cfadmin@cfserver.com"
+             '''
+    script = script.format(mail_cc=mail_cc, mail_bcc=mail_bcc, mail_to=mail_to)
+    script += (
```
digitronik: Q; why not it's part of the script directly?

cfme/tests/automate/test_method.py
```diff
@@ -264,3 +265,96 @@ def test_task_id_for_method_automation_log(request, generic_catalog_item):
...
+        testtype: functional
+        startsin: 5.10
+        casecomponent: Automate
+
+    Bugzilla:
+        1688500
+        1702304
+    """
+    # Resetting role - 'notifier' to default value
+    request.addfinalizer(
```
digitronik: I'm suspect why not where we enable `notifier`?

cfme/tests/automate/test_method.py
ganeshhubale: I tried adding this part to script it self but while ```formatting``` that multi-line string it gives error and leads to change in syntax of ruby code. Hence updated as above.

cfme/tests/automate/test_method.py
```diff
@@ -264,3 +265,90 @@ def test_task_id_for_method_automation_log(request, generic_catalog_item):
...
+    # Adding instance to call automate method - send_email
+    instance = klass.instances.create(
+        name=fauxfactory.gen_alphanumeric(),
+        display_name=fauxfactory.gen_alphanumeric(),
+        description=fauxfactory.gen_alphanumeric(),
+        fields={'execute': {'value': method.name}}
+    )
+
+    # Setting Outgoing SMTP E-mail Server
+    klass.appliance.server.settings.update_smtp_server(
```
valaparthvi: You can directly use the `smtp_test` fixture instead of this.
Refer this: https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/configure/test_email.py#L7

cfme/tests/automate/test_method.py
```diff
@@ -264,3 +265,90 @@ def test_task_id_for_method_automation_log(request, generic_catalog_item):
...
+
+
+@pytest.mark.meta(blockers=[BZ(1704439)])
+@pytest.mark.meta(server_roles="+notifier")
+def test_send_email_method(request, klass):
+    """
+    Polarion:
+        assignee: ghubale
+        initialEstimate: 1/20h
+        caseimportance: high
```
valaparthvi: No need of mentioning the `caseimportance` if it is high and `testtype` if it is functional.

cfme/tests/automate/test_method.py
```diff
@@ -264,3 +265,82 @@ def test_task_id_for_method_automation_log(request, generic_catalog_item):
...
+            "class": klass.name,
+            "instance": instance.name,
+        },
+        message="create",
+        request="Call_Instance",
+        execute_methods=True,
+    )
+    result.validate_logs()
+
+    # ToDo: Update get_emails for mail_cc and mail_bcc. It does not considers cc and bcc addresses
```
digitronik: Required fall TODO guideline; Thanks for adding this.

ganeshhubale: Added GitHub issue: https://github.com/ManageIQ/integration_tests/issues/8820

PR #8777, opened 2019-05-02
[1LP][RFR] Changes related to CFME 5.11

jaryn:
It seems that in CFME 5.11, the postgress is no longer a one from SCL.
We need to change some path prefixes and service names to because of
that.

There are also some problems with the tests verifying ssl chains. I don't know the reason for the openssl command to fail so I am disabling these tests as the curl connection works in the end.

CFME 5.11 uses netfilter instead of iptables so checking whether iptables rules exist is wrong on 5.11. We can rather check whether the firewalld services are present in the managiq zone and whether this zone is attached to the eth0 interface.

{{py.test: -v --long-running cfme/tests/test_appliance.py}}

### DIFF
--- a/cfme/tests/test_appliance.py
+++ b/cfme/tests/test_appliance.py
@@ -5,6 +5,7 @@ import os
 import pytest
 
 from cfme.utils import conf
+from cfme.utils.blockers import BZ
 from cfme.utils.log_validator import LogValidator
 from cfme.utils.wait import wait_for_decorator
 
@@ -187,6 +188,7 @@ def test_cpu_total(appliance):
     assert int(result.output) >= 4
 
 
+@pytest.mark.meta(blockers=[BZ(1712929, forced_streams=['5.11'])])  # against RHEL
 @pytest.mark.ignore_stream("upstream")
 def test_certificates_present(appliance, soft_assert):
     """Test whether the required product certificates are present.
--- a/cfme/tests/test_appliance.py
+++ b/cfme/tests/test_appliance.py
@@ -84,7 +84,7 @@ def test_evm_running(appliance):
     'evmserverd',
     'evminit',
     'sshd',
-    'postgresql',
+    'db_service',
 ])
 @pytest.mark.uncollectif(
     lambda appliance: appliance.is_pod)
@@ -98,14 +98,7 @@ def test_service_enabled(appliance, service):
         testtype: functional
         casecomponent: Appliance
     """
-    if service == 'postgresql':
-        service = appliance.db_service.unit_name
-    if appliance.os_version >= '7':
-        cmd = 'systemctl is-enabled {}'.format(service)
-    else:
-        cmd = 'chkconfig | grep {} | grep -q "5:on"'.format(service)
-    result = appliance.ssh_client.run_command(cmd)
-    assert result.success, result.output
+    assert getattr(appliance, service).enabled
 
 
 @pytest.mark.ignore_stream("upstream")
--- a/cfme/tests/test_appliance.py
+++ b/cfme/tests/test_appliance.py
@@ -99,7 +99,7 @@ def test_service_enabled(appliance, service):
         casecomponent: Appliance
     """
     if service == 'postgresql':
-        service = '{}-postgresql'.format(appliance.db.postgres_version)
+        service = str(appliance.db.service_name)
     if appliance.os_version >= '7':
         cmd = 'systemctl is-enabled {}'.format(service)
     else:
--- a/cfme/tests/test_appliance.py
+++ b/cfme/tests/test_appliance.py
@@ -302,7 +302,9 @@ def test_appliance_console_packages(appliance):
         initialEstimate: 1/4h
         casecomponent: Appliance
     """
-    assert appliance.ssh_client.run_command('scl --list | grep -v rh-ruby').success
+    if appliance.ssh_client.run_command('which scl').success:
+        # We have the scl command. Therefore we need to check the packages.
+        assert appliance.ssh_client.run_command('scl --list | grep -v rh-ruby').success
 
 
 @pytest.mark.manual
--- a/cfme/tests/test_appliance.py
+++ b/cfme/tests/test_appliance.py
@@ -61,8 +61,7 @@ def test_firewalld_running(appliance):
         initialEstimate: 1/4h
         casecomponent: Appliance
     """
-    result = appliance.ssh_client.run_command('systemctl status firewalld').output
-    assert 'active (running)' in result
+    assert appliance.firewalld.is_active
 
 
 @pytest.mark.uncollectif(lambda appliance: appliance.is_pod)
@@ -110,13 +109,8 @@ def test_service_enabled(appliance, service):
 
 @pytest.mark.ignore_stream("upstream")
 @pytest.mark.uncollectif(lambda appliance: appliance.is_pod)
-@pytest.mark.parametrize('proto,port', [
-    ('tcp', 22),
-    ('tcp', 80),
-    ('tcp', 443),
-])
-def test_iptables_rules(appliance, proto, port):
-    """Verifies key iptable rules are in place
+def test_firewalld_services_are_active(appliance):
+    """Verifies key firewalld services are in place
 
     Polarion:
         assignee: jhenner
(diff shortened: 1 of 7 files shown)


cfme/utils/appliance/db.py
```diff
@@ -508,7 +514,8 @@ def enable_internal(self, region=0, key_address=None, db_password=None, ssh_pass
             rbt_repl = {
```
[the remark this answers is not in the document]
mshriver: Seems appropriate - but this may be relevant when running against a non-virtualized appliance (local MIQ dev server)

dajoRH:
Lint report for commit f2690161930331336128fafffd46437c47834552:

`cfme/utils/appliance/services.py`:
- :bangbang: Line 118:1: **E302** *expected 2 blank lines, found 1*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/test_appliance.py
```diff
@@ -108,15 +109,41 @@ def test_service_enabled(appliance, service):
...
+        'firewall-cmd --permanent --zone={} --list-services'.format(manageiq_zone))
+    assert {'ssh', 'http', 'https'} <= set(result.output.split())
+
+    default_iface_zone = appliance.ssh_client.run_command(
+        "firewall-cmd --get-zone-of-interface {}".format(default_iface)
+    ).output.strip()
+    assert default_iface_zone == manageiq_zone
+
+
+@pytest.mark.meta(blockers=[BZ(1712944, forced_streams='5.10')])
```
mshriver: `forced_streams` isn't needed here, and I don't think it should be used. This BZ blocks on 5.10 appliances without it.

```
In [4]: BZ(1712944).blocks
Out[4]: True

In [5]: store.current_appliance.version
Out[5]: Version('5.10.5.0')
```

cfme/tests/test_appliance.py
```diff
@@ -197,12 +222,14 @@ def test_certificates_present(appliance, soft_assert):
...
     ).success
 
-    for cert in known_certs:
-        assert appliance.ssh_client.run_command("test -f '{}'".format(cert)).success
-        assert appliance.ssh_client.run_command(
-            "openssl verify -CAfile {ca_cert} '{cert_file}'"
-            .format(ca_cert=rhsm_ca_cert, cert_file=cert)
-        )
+    if appliance.version < '5.11':
+        # Until BZ(1712929) against RHEL is solved, keep this branch not ommitted on 5.11.
```
mshriver: This should be controlled by a `BZ` object and its `blocks` method, not just documented in a comment.

cfme/tests/test_appliance.py
```diff
@@ -125,22 +152,20 @@ def test_iptables_rules(appliance, proto, port):
...
+    def get_def_iface_zone():
+        default_iface_zone_cmd = appliance.ssh_client.run_command(
+            "firewall-cmd --get-zone-of-interface {}".format(default_iface)
+        )
+        assert default_iface_zone_cmd.success
+        return default_iface_zone_cmd.output.strip()
+
+    assert get_def_iface_zone() == manageiq_zone
+
+    assert appliance.ssh_client.run_command("systemctl restart firewalld").success
```
mshriver: Please add a systemd object to appliance, and interact with firewalld through it. We have a Systemd unit class to handle this in a more consistent/programmatic way - You can directly call restart and wait for it to be running again after.

cfme/tests/test_appliance.py
jaryn: @mshriver Are you sure it is going to work when the BZ is reported against RHEL?

cfme/utils/appliance/__init__.py
```diff
@@ -309,14 +310,15 @@ class IPAppliance(object):
     auditd = SystemdService.declare(unit_name='auditd')
     chronyd = SystemdService.declare(unit_name='chronyd')
     collectd = SystemdService.declare(unit_name='collectd')
-    db_service = SystemdService.declare(unit_name=ApplianceDB.service_name)
+    # db_service unit_name is initialized in the __init__()
```
jaryn: This is not true with the most recent state of this PR and needs to be deleted.

cfme/tests/test_appliance.py
jaryn: @mshriver it is not blocking
```
In [3]: BZ(1712929).blocks
Out[3]: False

In [6]: cfme.utils.appliance.current_appliance.version
Out[6]: Version('5.11.0.5')
```

cfme/tests/test_appliance.py
```diff
@@ -99,7 +99,7 @@ def test_service_enabled(appliance, service):
         casecomponent: Appliance
     """
     if service == 'postgresql':
-        service = '{}-postgresql'.format(appliance.db.postgres_version)
+        service = str(appliance.db_service.unit_name)
```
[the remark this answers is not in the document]
mshriver: I don't think this is a required change. There's no harm in the call to `str()` here, and the test is being explicit in what its expecting `service` to be - a string, not some object type. Its possible unit_name is not actually a string type.

cfme/tests/test_appliance.py
```diff
@@ -108,15 +108,17 @@ def test_service_enabled(appliance, service):
     assert result.success, result.output
 
 
+@pytest.fixture
+def default_iface(appliance):
+    default_iface = appliance.ssh_client.run_command(
+        "ip r | awk '/^default/ { print $5 }'")
+    return default_iface
```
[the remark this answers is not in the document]
mshriver: I would actually like to see this as a cached appliance property.  I don't see it changing during tested unless explicitly so for a test, in which case that cache can be invalidated.

Its not really a fixture, as its not doing any test setup or cleanup, its just getting information about the appliance. Almost by definition an appliance property :)

cfme/tests/test_appliance.py
mshriver: For now using forced streams is fine, with a comment that its RHEL related.

We _can_ update `BZ.blocks` to also resolve the OS version and BZ product, but currently it won't block on these RHEL bugs because they won't have versions set, and other markers that the `blocks` method uses to resolve a CFME blocker.

cfme/tests/test_appliance.py
mshriver: @jaryn I see you added the service object, but aren't using it here. Miss a commit?

cfme/tests/test_appliance.py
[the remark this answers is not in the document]
digitronik: @mshriver accept is not required comment :)
I don't understand why extra call? I remember one PR to avoid the extra call and what should be other probable `type` of `unit_name`? I'm not able to understand.

cfme/utils/appliance/services.py
```diff
@@ -114,3 +112,19 @@ def daemon_reload(self, log_callback=None):
...
+
+
+@attr.s
+class SystemdService(BaseService):
+    unit_name = attr.ib(type=str)
+
+
+@attr.s
+class DynaService(BaseService):
+    service_name_getting_cmd = attr.ib(type=str)
```
mshriver: I think we can find a better name for this attribute.  Good call on splitting the classes here.

The unit_name can be a cached_property, to reduce the number of SSH calls to fetch this.  We can delete the cached entry on upgrades, when it may change.

I'm also considering that while its nice to be able to use the envvar to set this value, having to define the unit in this way, and fetch it via costly ssh, is _too_ costly. We might be better off just version picking it - while less dynamic, this does not change often at all.

cfme/utils/appliance/__init__.py
```diff
@@ -1035,6 +1041,13 @@ def create_ssh_connection():
         store.ssh_clients_to_close.append(ssh_client)
         return ssh_client
 
+    @cached_property
```
mshriver: :rocket:

cfme/utils/appliance/__init__.py
```diff
@@ -346,6 +348,10 @@ class IPAppliance(object):
         'sssd': '/etc/sssd/sssd.conf'
     }
 
+    @cached_property
+    def db_service(self):
```
mshriver: :rocket:

cfme/tests/test_appliance.py
```diff
@@ -98,25 +98,37 @@ def test_service_enabled(appliance, service):
...
+        assignee: jhenner
+        initialEstimate: 1/4h
+        testtype: functional
+        casecomponent: Appliance
+        upstream: no
+    """
+    manageiq_zone = "manageiq"
+    result = appliance.ssh_client.run_command(
+        'firewall-cmd --permanent --zone={} --list-services'.format(manageiq_zone))
+    assert {'ssh', 'http', 'https'} <= set(result.output.split())
```
mshriver: :+1:

PR #8778, opened 2019-05-02
[1LP][RFR] Add miq command for updating qe-test-coverage flag and generating BZ reports

john-dupuy:
This PR adds a `miq` command to check tests marked with the `pytest.mark.meta(automates=[<bug_id>])` and `pytest.mark.meta(coverage=[<bug_id>])` and output information about them. 

The basic usage is a follows:
```python,
miq bz <command> <directory>
```
where <command> is one of report, list, or coverage, and <directory> is the testing directory, e.g.
"cfme/tests/control/"

To list BZs that this script would set coverage for, you can do,
```python,
miq bz coverage <directory>
```
This will print out the ids of BZs that could have qe_test_coverage switched to '+'
To then set qe_test_coverage on those BZs, you can do,
```python,
miq bz coverage --set <directory>
```

**Note to reviewers**: if you'd like to test this command, it is easiest to setup a test function in a new file (e.g. cfme/utils/tests/test_bz_coverage.py)  
```python,
@pytest.mark.meta(coverage=[1692537])
def test_automates():
    pass
```
The run `miq bz coverage cfme/utils/tests/test_bz_coverage.py`. This should be output at the end of the test run, 
```
The following BZs should have qe_test_coverage set to '+': 
    id: 1692537, qe_test_coverage: -
```
You can change the ID to anything you'd like to check if it works. Note that running it with `--set` will actually change the coverage flag, so be wary about running with that option.

### DIFF
--- a/cfme/fixtures/bzs.py
+++ b/cfme/fixtures/bzs.py
@@ -1,10 +1,9 @@
 # -*- coding: utf-8 -*-
 """Collection of fixtures for simplified work with bzs.
 
-The main purpose of this file is to add a pytest option for generates a BZ report, that
-gives information about the BZs that appear as coverage/automates metadata in the test_functions
+The main purpose of this file is to add a pytest option which generates a BZ report. This option
+gives information about the BZs that appear as coverage/automates metadata in test functions.
 """
-import pytest
 import yaml
 
 from cfme.fixtures.pytest_store import store
@@ -22,7 +21,6 @@ def pytest_addoption(parser):
     )
 
 
-@pytest.mark.trylast
 def pytest_collection_modifyitems(session, config, items):
     if not config.getvalue("generate_bz_report"):
         return
--- a/cfme/fixtures/bzs.py
+++ b/cfme/fixtures/bzs.py
@@ -1,0 +1,54 @@
+# -*- coding: utf-8 -*-
+"""Collection of fixtures for simplified work with bzs.
+
+The main purpose of this file is to add a pytest option for generates a BZ report, that
+gives information about the BZs that appear as coverage/automates metadata in the test_functions
+"""
+import pytest
+import yaml
+
+from cfme.fixtures.pytest_store import store
+from cfme.utils.blockers import BZ
+
+
+def pytest_addoption(parser):
+    group = parser.getgroup('Blocker options')
+    group.addoption(
+        '--generate-bz-report',
+        action='store_true',
+        default=False,
+        dest='generate_bz_report',
+        help='Generate a BZ report based on the automates/coverage metadata of test cases.'
+    )
+
+
+@pytest.mark.trylast
+def pytest_collection_modifyitems(session, config, items):
+    if not config.getvalue("generate_bz_report"):
+        return
+    store.terminalreporter.write("Loading automated/covered BZs ...\n", bold=True)
+    bz_list = []
+    for item in items:
+        if "automates" not in item._metadata and "coverage" not in item._metadata:
+            continue
+        # get list of bzs with coverage
+        if "automates" in item._metadata:
+            bz_list.extend(item._metadata["automates"])
+        if "coverage" in item._metadata:
+            bz_list.extend(item._metadata["coverage"])
+
+    if bz_list:
+        # remove duplicate BZs
+        bz_list = list(dict.fromkeys(bz_list))
+        # remove references that are an instance of BZ
+        bz_list = [bug.bug_id if isinstance(bug, BZ) else bug for bug in bz_list]
+        # get BZ info
+        info = BZ.bugzilla.get_bz_info(bz_list)
+        # output BZ info to yaml
+        with open("bz-report.yaml", "w") as outfile:
+            yaml.dump(info, outfile, default_flow_style=False)
+    else:
+        store.terminalreporter.write(
+            "ERROR: No BZs marked with 'automates'/'coverage' in that test module. A report will "
+            "not be generated.\n", bold=True
+        )
--- a/cfme/scripting/bz.py
+++ b/cfme/scripting/bz.py
@@ -1,0 +1,186 @@
+"""
+Scripts for dealing with Bugzilla metadata, listing BZs with coverage, and setting qe_test_coverage
+flag. This script looks at test-case metadata for the flags "automates" and "coverage", and fetches
+information about the BZs listed there.
+
+For usage see, "miq bz --help"
+
+The basic usage of this command is via
+    .. code-block:: python
+    >>> miq bz <command> <directory>
+where <command> is one of report, list, or coverage, and <directory> is the testing directory, e.g.
+"cfme/tests/control/"
+
+To list BZs that this script would set coverage for, you can do,
+    .. code-block:: python
+    >>> miq bz coverage <directory>
+This will print out the ids of BZs that could have qe_test_coverage switched to '+'
+To then set qe_test_coverage on those BZs, you can do,
+    .. code-block:: python
+    >>> miq bz coverage --set <directory>
+"""
+import os
+import sys
+from collections import namedtuple
+
+import click
+import pytest
+import yaml
+
+from cfme.utils.blockers import BZ
+from cfme.utils.log import logger
+
+STATUS = {
+    "open_bzs": {"val": "true", "text": "are open"},
+    "closed_bzs": {"val": "false", "text": "are closed"},
+    "all_bzs": {"val": None, "text": "have coverage"}
+}
+
+
+def get_report(directory):
+    click.echo("Generating a BZ report in bz-report.yaml")
+    pytest.main([
+        "--use-provider", "complete",
+        "--long-running",
+        "--use-template-cache",
+        "--collect-only",
+        "-q",
+        "--generate-bz-report", directory
+    ])
+    # read the generated yaml
+    try:
+        with open("bz-report.yaml", "r") as stream:
+            info = yaml.load(stream, Loader=yaml.BaseLoader)
+    except IOError:
+        msg = (
+            "ERROR: File bz-report.yaml not found, something went wrong during report generation.\n"
+            "       Likely no BZs were found in {} with 'automates'/'coverage',"
+            " so a report wasn't generated.".format(directory)
+        )
+        click.secho(msg, err=True, bold=True, fg="red")
+        sys.exit(0)
+    return info
+
+
+def get_qe_test_coverage(info, open_only=True):
+    """
+    Given info (what is returned from yaml.load on bz-report.yaml),
+    This function screens and returns only BZs which are OPEN, if open_only=True, otherwise it
+    returns all BZs
+    """
+    BZTestCoverage = namedtuple("BZTestCoverage", ["id", "qe_test_coverage"])
+
+    bz_list = []
+    for bug_id in info.keys():
+        if open_only and info[bug_id]["is_open"] == "false":
+            continue
+        # get qe_test_coverage_flag
+        qe_test_coverage = "?"
+        for flag in info[bug_id]["flags"]:
+            if flag["name"] == "qe_test_coverage":
+                qe_test_coverage = flag["status"]
+                break
+        # append the BZ
+        bz_list.append(BZTestCoverage(id=bug_id, qe_test_coverage=qe_test_coverage))
+
+    return bz_list
+
+
+def cleanup():
+    click.echo("Removing the BZ report file, bz-report.yaml")
+    try:
+        os.remove("bz-report.yaml")
+    except OSError:
+        logger.exception("bz-report.yaml not found")
+
+
+@click.group(help="Functions for generating reports on BZs included in test suite metadata")
+def main():
+    pass
+
+
+@main.command(help="Generate BZ report on BZs that have coverage given a directory")
+@click.argument("directory", default="cfme/tests/")
+def report(directory):
+    get_report(directory)
+
+
+@main.command(help="List open/closed BZs that have test coverage")
+@click.argument("directory", default="cfme/tests/")
+@click.option(
+    "--all",
+    "-a",
+    "bz_status",
+    is_flag=True,
+    help="list all BZs with coverage",
+    default=True,
+    show_default=True,
+    flag_value="all_bzs",
+)
+@click.option(
+    "--open",
+    "-o",
+    "bz_status",
+    is_flag=True,
+    help="list open BZs with coverage",
+    default=False,
+    show_default=True,
+    flag_value="open_bzs"
+)
+@click.option(
+    "--closed",
+    "-c",
+    "bz_status",
+    is_flag=True,
+    help="list closed BZs with coverage",
+    default=False,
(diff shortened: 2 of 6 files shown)


dajoRH:
Lint report for commit b129767963ed8fdf3d24e9076d70d2ac4eafa64c:

* 1e3c0b8d2979d2972d2e78b0b754682798a3c4c2

*CFME QE Bot*

cfme/scripting/bz.py
```diff
@@ -0,0 +1,149 @@
...
+@main.command(help="Generate BZ report on BZs that have coverage given a directory")
+@click.argument("directory", default="cfme/tests/")
+def report(directory):
+    logger.info("Generating a BZ report in bz-report.yaml")
+    get_report(directory)
+
+
+@main.command(help="List open/closed BZs that have test coverage")
+@click.argument("directory", default="cfme/tests/")
+@click.option("--open", "open_bzs", is_flag=True, help="list open BZs")
```
digitronik: optional:
I would use short name also like for `--open` `-o` and `--all` `-a`. As click provides you an easy facility for this.

cfme/scripting/bz.py
```diff
@@ -0,0 +1,149 @@
...
+        val = "false"
+        text = "are closed"
+
+    if not all_bzs:
+        bz_list = [bug_id for bug_id in info.keys() if info[bug_id]["is_open"] == val]
+    else:
+        bz_list = [bug_id for bug_id in info.keys()]
+
+    if bz_list:
+        print("The following BZ's {}: \n{}".format(text, "\n".join(bz_list)))
```
digitronik: Required:
Please use `click.echo()` in state of `print`. Its good practice for cli application writing.

cfme/scripting/bz.py
```diff
@@ -0,0 +1,149 @@
...
+    else:
+        print("I found no BZ's that {} BZ's".format(text))
```
digitronik: same as above ^^

cfme/scripting/bz.py
```diff
@@ -0,0 +1,149 @@
...
+@click.argument("directory", default="cfme/tests/")
+def report(directory):
+    logger.info("Generating a BZ report in bz-report.yaml")
+    get_report(directory)
+
+
+@main.command(help="List open/closed BZs that have test coverage")
+@click.argument("directory", default="cfme/tests/")
+@click.option("--open", "open_bzs", is_flag=True, help="list open BZs")
+@click.option("--all", "all_bzs", is_flag=True, help="list all BZs with coverage")
```
digitronik: Optional:
I think the default should be `--all` and `open` or `close` should be optional. It will give you more flexibility with command. I still not tested locally.

cfme/scripting/bz.py
```diff
@@ -0,0 +1,149 @@
...
+        print("The following BZ's {}: \n{}".format(text, "\n".join(bz_list)))
+    else:
+        print("I found no BZ's that {} BZ's".format(text))
+    cleanup()
+
+
+@main.command(help="Set QE test coverage flag based on automates/coverage metadata")
+@click.argument("directory", default="cfme/tests")
+@click.option(
+    "--set",
```
digitronik: I would use short name for `--set` to `-s`
```
@click.option(
   "-s",
    "--set",
    is_flag=True,
    help="Set QE test coverage on BZs that are marked in coverage/auotmates test metadata",
    default=False,
    show_default=True
)
```

cfme/utils/bz.py
```diff
@@ -207,6 +209,58 @@ def resolve_blocker(self, blocker, version=None, ignore_bugs=None, force_block_s
...
+
+        # logout
+        self.bugzilla.logout()
+
+    def get_bz_info(self, idlist):
+        bz_creds = credentials.get("bugzilla", None)
+
+        # login to bzapi
+        if not bz_creds:
+            raise BugzillaError("No bugzilla creds available")
```
digitronik: Please use the full form in Notifications..

Optional:
In state raise an error; I would just echo statement like "Bugzilla credentials not found" and exit
```
sys.exit(0)
```

cfme/utils/bz.py
```diff
@@ -207,6 +209,58 @@ def resolve_blocker(self, blocker, version=None, ignore_bugs=None, force_block_s
...
+        try:
+            self.bugzilla.login(bz_creds.get("username", None), bz_creds.get("password", None))
```
digitronik: Required:
default is `None` if not found "username".

cfme/utils/bz.py
```diff
@@ -207,6 +209,58 @@ def resolve_blocker(self, blocker, version=None, ignore_bugs=None, force_block_s
                 return bug
         return None
 
+    def set_flags(self, idlist, flags):
+        bz_creds = credentials.get("bugzilla", None)
+
+        # login to bzapi
+        if not bz_creds:
```
digitronik: Optional: I would not raise error just with msg exit.

cfme/scripting/bz.py
```diff
@@ -0,0 +1,149 @@
...
+    pytest.main([
+        "--use-provider", "complete",
+        "--long-running",
+        "--use-template-cache",
+        "--collect-only",
+        "--generate-bz-report", directory
+    ])
+    # read the generated yaml
+    with open("bz-report.yaml", "r") as stream:
+        info = yaml.load(stream, Loader=yaml.BaseLoader)
```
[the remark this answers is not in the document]
john-dupuy: I am using `load` here because this yaml file has objects in it like `!!python/object:xmlrpclib.DateTime` that `safe_load` doesn't recognize. That is also why I have `Loader=yaml.BaseLoader`, so that it parses those as strings. Do you know of a way I can work around this?

cfme/utils/bz.py
john-dupuy: Oh yeah, good point, I'll remove that

cfme/utils/bz.py
john-dupuy: Hmm.. the trouble is that these commands are called during a pytest run so they will show up as INTERNALERROR, and then the code after the pytest run will try to execute, even with a system exit. I raising the BZ error is informative and shows the user why (and where) the command failed.

cfme/utils/bz.py
john-dupuy: See my comment about this above

cfme/scripting/bz.py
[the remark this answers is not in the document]
digitronik: aahhh we are using old pckg `PyYAML` need to move on `ruamel.yaml`.

cfme/fixtures/bzs.py
```diff
@@ -0,0 +1,55 @@
...
+    )
+
+
+@pytest.mark.trylast
+def pytest_collection_modifyitems(session, config, items):
+    if not config.getvalue("generate_bz_report"):
+        return
+    store.terminalreporter.write("Loading automated/covered BZs ...\n", bold=True)
+    bz_list = []
+    for item in items:
```
mshriver: Just general comment about formatting and structure here. Consider using comprehension like below - but note that its kinda depends on whether or not `item._metadata['type']`  always returns a list.

```python
relevant_items = [i for i in items if 'automates' not in i._metadata and 'coverage' not in i._metadata]
bz_list = [i._metadata.get('automates') for i in relevant_items]  # will put some None's in the list
# using extend here depends on the metadata value being a list
[bz_list.extend(i._metadata.get('coverage', [None]) for i in relevant_items]  
bz_list = set([bug.bug_id if isinstance(bug, BZ) else bug for bug in bz_list])  # eliminate duplicates and convert instances
bz_list.remove(None)  # drop the irrelevant None
```

cfme/fixtures/bzs.py
```diff
@@ -0,0 +1,55 @@
...
+        info = BZ.bugzilla.get_bz_info(bz_list)
+        # output BZ info to yaml
+        with open("bz-report.yaml", "w") as outfile:
+            yaml.dump(info, outfile, default_flow_style=False)
+    else:
+        store.terminalreporter.write(
+            "ERROR: No BZs marked with 'automates'/'coverage' in that test module. A report will"
+            "not be generated.\n", bold=True
+        )
+        pytest.exit("No BZs found marked with 'automates' or 'coverage', exiting...")
```
mshriver: Should we exit the session here? What if I want to run tests _and_ generate a report while doing so? This would kill my testrun if I selected tests that don't happen to have any coverage/automates markers.

cfme/utils/bz.py
mshriver: Not sure what comment above you're talking about - can you put an in-line comment here as to why we're raising and not just logging.

cfme/utils/bz.py
```diff
@@ -207,6 +209,58 @@ def resolve_blocker(self, blocker, version=None, ignore_bugs=None, force_block_s
...
+            logger.info("Got %s from updating %s", result, bug)
+
+        # logout
+        self.bugzilla.logout()
+
+    def get_bz_info(self, idlist):
+        bz_creds = credentials.get("bugzilla", None)
+
+        # login to bzapi
+        if not bz_creds:
```
mshriver: Can you create a common method for this, since its used in any method actually contacting bugzilla, and we're not having to do the same logic blocks.

mshriver: Additionally, it seems prime for a contextmanger use, since you're explicitly calling login/logout before/after BZ queries.

cfme/fixtures/bzs.py
john-dupuy: I agree, we shouldn't exit the session here, this was implemented because it wasn't clear to the user why the report wasn't generated when running `miq bz`, but I will exit the session there, so that we can still do `pytest --generate-report` and not have it exit if the tests don't have any coverage/automates markers.

cfme/utils/bz.py
```diff
@@ -207,6 +209,58 @@ def resolve_blocker(self, blocker, version=None, ignore_bugs=None, force_block_s
...
+        try:
+            self.bugzilla.login(bz_creds.get("username"), bz_creds.get("password"))
+        except BugzillaError:
+            logger.exception("Unable to login to Bugzilla with those creds.")
+            raise
+
+        logger.info("Getting information about the following BZ's: %s", idlist)
+
+        # build info
+        info = {}
```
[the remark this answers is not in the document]
john-dupuy: I tried this but this breaks my current implementation. the keys of info are the bug ids, and the values are the AttrDict (i.e. the information about the bug id)

dajoRH:
*CFME QE Bot*

cfme/fixtures/bzs.py
```diff
@@ -0,0 +1,54 @@
+# -*- coding: utf-8 -*-
+"""Collection of fixtures for simplified work with bzs.
+
+The main purpose of this file is to add a pytest option for generates a BZ report, that
```
jawatts: Typo here

cfme/fixtures/bzs.py
```diff
@@ -0,0 +1,54 @@
...
+    group.addoption(
+        '--generate-bz-report',
+        action='store_true',
+        default=False,
+        dest='generate_bz_report',
+        help='Generate a BZ report based on the automates/coverage metadata of test cases.'
+    )
+
+
+@pytest.mark.trylast
```
jawatts: **Question** Do we want to mark this with `trylast` here?

jawatts: **Question** Do we want to mark this with `trylast` here?

dajoRH:
*CFME QE Bot*

cfme/scripting/bz.py
```diff
@@ -0,0 +1,186 @@
...
+    is_flag=True,
+    help="Set QE test coverage on BZs that are marked in coverage/auotmates test metadata",
+    default=False,
+    show_default=True
+)
+def coverage(directory, set_bzs):
+    info = get_report(directory)
+
+    # get list of open BZs that are open and should have test coverage set
+    bz_list = get_qe_test_coverage(info, open_only=True)
```
jawatts: **Required** I think that we shouldn't just restrict this to open bugs. What if the user wants to set the coverage for a bug that has been closed.

cfme/utils/bz.py
jawatts: I think this is still applicable here

cfme/fixtures/bzs.py
john-dupuy: Yeah this is leftover from some code I was looking at when writing this function. It is not necessary, so I'll remove it.

cfme/fixtures/bzs.py
john-dupuy: Just realized I never responded to this. While this reduces lines of code, in my opinion, what I have there now is a bit more readable. What do you think?

cfme/scripting/bz.py
john-dupuy: I'm going to implement something similar to the `miq list bz` so that the user can choose to audit `all`, only `open`, or only `closed` bzs.

cfme/utils/bz.py
john-dupuy: Sorry, missed that message @mshriver. We're raising the error so that the user will know why their command has failed if there are no bugzilla credentials available

cfme/scripting/bz.py
```diff
@@ -0,0 +1,221 @@
...
+    is_flag=True,
+    help="Audit test coverage on only closed BZs",
+    default=False,
+    show_default=True,
+    flag_value="closed_bzs"
+)
+def coverage(directory, set_bzs, bz_status):
+    info = get_report(directory)
+
+    # get list of open BZs that are open and should have test coverage set
```
jawatts: @john-dupuy I think the comment needs updating since a change has been made to show whatever the user wants

john-dupuy: updated

cfme/scripting/bz.py
```diff
@@ -0,0 +1,221 @@
...
+    "--closed",
+    "-c",
+    "bz_status",
+    is_flag=True,
+    help="Audit test coverage on only closed BZs",
+    default=False,
+    show_default=True,
+    flag_value="closed_bzs"
+)
+def coverage(directory, set_bzs, bz_status):
```
mshriver: In running `miq bz coverage`, I'm seeing some misleading/conflicting information about the BZ's I can/should update.

```
The following BZs should have qe_test_coverage set to '+': 
    id: 1443411, qe_test_coverage: +
```

As shown in the output, 1443411 has qe_test_coverage set to `+`.  Turns out there is a clone of this BZ that has qe_test_coverage set to `-`.

We need to either not take BZ clones into effect when checking/setting coverage, or we need to identify and include them in the output instead of the marked BZ number.

john-dupuy: As I've written it, this does nothing with clones at the moment. Perhaps this could be a future enhancement. I think this is just a "doh" moment where I forgot to check if the BZs had their coverage set to "+" in the coverage audit script (`get_qe_test_coverage`). I added a condition in there to prevent this from happening.

PR #8779, opened 2019-05-03
[1LP][RFR] fix setting configuration nav 511

digitronik:

Purpose or Intent
=================

{{pytest: cfme/tests/configure/}}

### DIFF
--- a/cfme/base/login.py
+++ b/cfme/base/login.py
@@ -1,4 +1,5 @@
 from widgetastic.exceptions import NoSuchElementException
+from widgetastic.widget import Text
 from widgetastic.widget import View
 from widgetastic_patternfly import FlashMessages
 from widgetastic_patternfly import NavDropdown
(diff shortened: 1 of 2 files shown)


digitronik:
As per PRT run navigation problem solve. failures are not related to changes. Need to fix views like Server and others.

cfme/base/login.py
```diff
@@ -15,6 +16,7 @@ class BaseLoggedInPage(View):
     CSRF_TOKEN = '//meta[@name="csrf-token"]'
     flash = FlashMessages('.//div[@id="flash_msg_div"]')
     help = NavDropdown('.//li[./a[@id="dropdownMenu1"]]|.//li[./a[@id="help-menu"]]')
+    configuration_settings = Text('.//li[./a[@title="Configuration"]]')      # 5.11+
```
izapolsk: @digitronik, I'd rather wrap up this item and old item in VersionPick. you wouldn't need to rename it when 5.10 was dumped.

digitronik: @izapolsk I would like to wrap but for  `< 5.10` `Configuration` is part of Dropdown `settings` [1] and now we have a button beside setting dropdown. so; I choose this way to differentiate `select` and `click` in navigation steps. Please let me know if you have any idea in mind.

[1] https://github.com/ManageIQ/integration_tests/pull/8779/files#diff-7c97dce5eefb958f90c137c66bd8f857R20

mshriver: I'm going to merge this as is because its workable. We can improve as we decide a better way to handle direct use of `view.settings.fill`, nested/included views, etc.

PR #8780, opened 2019-05-03
[1LP][RFR] New Test: Removal of openshift "Deployment" from automate domain

ganeshhubale:
- Testing removal of openshift "Deployment" from automate domain - ManageIQ

{{ pytest: cfme/tests/automate/test_namespace.py -k 'test_remove_openshift_deployment_in_automate' -vv }}


jawatts:
@izapolsk Talked a bit offline with @mshriver about this. Maybe we can move this to a different test case that may be already checking something similar? If not, I think we should keep it in. I do agree it may add to our maintenance, but this is attached to a BZ. Since we are running `composite-uncollect `, it won't be run everytime

PR #8782, opened 2019-05-03
[1LP][RFR] Automate Test: test_rbac_move_custom_report

valaparthvi:
1. Add a new method `move_reports` to move custom reports to a different menu.
2. Automate a test case: test_rbac_move_custom_report
3. Add 2 fixtures: `get_custom_report` and `rbac_user`.
3. Blackify test_menus.py (kindly ignore this for now)
4. Add test_requirements.
5. Add a new view `AllReportMenusView` and a navigator to the view.
6. Remove `handle_alert` from `logout` function which raises error on logout.

{{ pytest: cfme/tests/intelligence/reports/test_menus.py::test_rbac_move_custom_report --use-template-cache -sqvvv }}

### DIFF
--- a/cfme/intelligence/reports/menus.py
+++ b/cfme/intelligence/reports/menus.py
@@ -32,9 +32,7 @@ class AllReportMenusView(CloudIntelReportsView):
         )
 
 
-class EditReportMenusView(CloudIntelReportsView):
-    title = Text("#explorer_title_text")
-    reports_tree = ManageIQTree("menu_roles_treebox")
+class EditReportMenusView(AllReportMenusView):
     # Buttons
     save_button = Button("Save")
     reset_button = Button("Reset")
@@ -74,7 +72,7 @@ class ReportMenu(BaseEntity):
 
     def go_to_group(self, group_name):
         self.group = group_name
-        view = navigate_to(self, "EditReportMenus")
+        view = navigate_to(self, "Edit")
         assert view.is_displayed
         return view
 
@@ -133,6 +131,10 @@ class ReportMenu(BaseEntity):
         view = self.go_to_group(group)
         view.default_button.click()
         view.save_button.click()
+        flash_view = self.create_view(AllReportMenusView)
+        assert flash_view.flash.assert_message(
+            'Report Menu for role "{}" was saved'.format(group)
+        )
 
     @contextmanager
     def manage_subfolder(self, group, folder, subfolder):
@@ -161,6 +163,10 @@ class ReportMenu(BaseEntity):
             # If no exception happens, save!
             view.commit_button.click()
             view.save_button.click()
+            flash_view = self.create_view(AllReportMenusView)
+            flash_view.flash.assert_message(
+                'Report Menu for role "{}" was saved'.format(group)
+            )
 
     @contextmanager
     def manage_folder(self, group, folder=None):
@@ -193,6 +199,10 @@ class ReportMenu(BaseEntity):
             # If no exception happens, save!
             view.manager.commit()
             view.save_button.click()
+            flash_view = self.create_view(AllReportMenusView)
+            flash_view.flash.assert_message(
+                'Report Menu for role "{}" was saved'.format(group)
+            )
 
     def move_reports(self, group, folder, subfolder, *reports):
         """ Moves a list of reports to a given menu
@@ -208,9 +218,11 @@ class ReportMenu(BaseEntity):
         with self.manage_subfolder(group, folder, subfolder) as selected_menu:
             selected_options = selected_menu.parent_view.report_select.all_options
 
-            if set(selected_options) & set(reports):
-                # If the report is already present, we stay on the same page after fill is complete.
-                cancel_view = self.create_view(EditReportMenusView)
+            diff = set(selected_options) & set(reports)
+            if diff and (len(diff) == len(reports)):
+                cancel_view = self.create_view(AllReportMenusView)
+                # If all the reports to be moved are already present, raise an exception to exit.
+                raise FolderManager._BailOut
 
             # fill method replaces all the options in all_options with the value passed as argument
             # We do not want to replace any value, we just want to move the new reports to a given
@@ -219,15 +231,11 @@ class ReportMenu(BaseEntity):
             selected_menu.parent_view.report_select.fill(reports)
 
         if cancel_view:
-            assert cancel_view.is_displayed
-            cancel_view.cancel_button.click()
-            message = 'Edit of Report Menu for role "{}" was cancelled by the user'
-        else:
-            message = 'Report Menu for role "{}" was saved'
-
-        view = self.create_view(AllReportMenusView)
-        assert view.is_displayed
-        view.flash.assert_message(message.format(group))
+            cancel_view.flash.assert_message(
+                'Edit of Report Menu for role "{}" was cancelled by the user'.format(
+                    group
+                )
+            )
 
 
 @attr.s
@@ -237,15 +245,16 @@ class ReportMenusCollection(BaseCollection):
     ENTITY = ReportMenu
 
 
-@navigator.register(ReportMenu)
+@navigator.register(ReportMenu, "Edit")
 class EditReportMenus(CFMENavigateStep):
     VIEW = EditReportMenusView
-    prerequisite = NavigateToAttribute("appliance.server", "CloudIntelReports")
+    prerequisite = NavigateToAttribute(
+        "appliance.collections.intel_report_menus", "All"
+    )
 
     def step(self, *args, **kwargs):
         self.view.edit_report_menus.tree.click_path(
-            "All EVM Groups",
-            self.obj.group
+            "All EVM Groups", self.obj.group
         )
 
 
--- a/cfme/intelligence/reports/menus.py
+++ b/cfme/intelligence/reports/menus.py
@@ -195,12 +195,8 @@ class ReportMenu(BaseEntity):
             selected_options = selected_menu.parent_view.report_select.all_options
             for report in reports:
                 if report in selected_options:
-                    raise Exception(
-                        "Report already present in the menu {}/{}. Cannot move.".format(
-                            folder, subfolder
-                        )
-                    )
-
+                    # pass, since the report is already present
+                    pass
             # fill method replaces all the options in all_options with the value passed as argument
             # We do not want to replace any value, we just want to move the new reports to a given
             # menu. This is a work-around for that purpose.
--- a/cfme/intelligence/reports/menus.py
+++ b/cfme/intelligence/reports/menus.py
@@ -195,7 +195,11 @@ class ReportMenu(BaseEntity):
             selected_options = selected_menu.parent_view.report_select.all_options
             for report in reports:
                 if report in selected_options:
-                    raise Exception("Report already present. Cannot move.")
+                    raise Exception(
+                        "Report already present in the menu {}/{}. Cannot move.".format(
+                            folder, subfolder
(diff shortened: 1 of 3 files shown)


cfme/intelligence/reports/menus.py
```diff
@@ -180,6 +180,30 @@ def manage_folder(self, group, folder=None):
...
+            reports: List of reports that are to be moved.
+        """
+        if not isinstance(reports, list):
+            reports = [reports]
+
+        with self.manage_subfolder(group, folder, subfolder) as selected_menu:
+            selected_options = selected_menu.parent_view.report_select.all_options
+            for report in reports:
+                if report in selected_options:
+                    raise Exception("Report already present. Cannot move.")
```
john-dupuy: Thoughts on making a custom expression for this scenario, e.g. `ReportCannotBeMovedException`, also some information about where the report is already present might be helpful.

valaparthvi: Definitely! That's a good idea! I'll see if I can implement that.

valaparthvi: Actually, this error is raised when the report is already present in the place where it is being moved. I'll modify the exception message to make things clearer.

cfme/intelligence/reports/menus.py
```diff
@@ -180,6 +180,34 @@ def manage_folder(self, group, folder=None):
...
+            reports: List of reports that are to be moved.
+        """
+        if not isinstance(reports, list):
+            reports = [reports]
+
+        with self.manage_subfolder(group, folder, subfolder) as selected_menu:
+            selected_options = selected_menu.parent_view.report_select.all_options
+            for report in reports:
+                if report in selected_options:
+                    raise Exception(
```
digitronik: Q:
why not just pass? its user wants to `a`, `b`, `c` report and `b` already in the menu. Then a better idea to just move `a` and `c` and for `b` in iteration just pass.

cfme/intelligence/reports/menus.py
```diff
@@ -180,6 +180,30 @@ def manage_folder(self, group, folder=None):
...
+        """
+        if not isinstance(reports, list):
+            reports = [reports]
+
+        with self.manage_subfolder(group, folder, subfolder) as selected_menu:
+            selected_options = selected_menu.parent_view.report_select.all_options
+            for report in reports:
+                if report in selected_options:
+                    # pass, since the report is already present
+                    pass
```
digitronik: I think you took my suggestion in the wrong way :)

```
            for report in reports:
                if report in selected_options:
                    # pass, since the report is already present
                    pass
```
what is the use of this for loop?

valaparthvi: I understand your question now :sweat_smile: 
I think the new changes will now make a sense of why I am using the loop. If the report is already present, we stay on the same page. To get out of that page, we have to click cancel. I had added a new view(`AllReportMenusView) and navigator which were not present earlier and modified the `move_reports` method.

digitronik: Let me know as PRT cleared... LGTM :+1:

cfme/intelligence/reports/menus.py
```diff
@@ -180,6 +194,43 @@ def manage_folder(self, group, folder=None):
...
+                    # If the report is already present, we stay on the same page.
+                    cancel_view = self.create_view(EditReportMenusView)
+
+            # fill method replaces all the options in all_options with the value passed as argument
+            # We do not want to replace any value, we just want to move the new reports to a given
+            # menu. This is a work-around for that purpose.
+            reports.extend(selected_options)
+            selected_menu.parent_view.report_select.fill(reports)
+
+        view = self.create_view(AllReportMenusView)
```
digitronik: Please move this above L232. even I don't think you need to create this `AllReportMenusView` you are using only `flash` which is coming from Base view but its good practice that creates a spacific view :+1:

cfme/intelligence/reports/menus.py
```diff
@@ -180,6 +194,43 @@ def manage_folder(self, group, folder=None):
...
+        """
+        if not isinstance(reports, list):
+            reports = [reports]
+
+        cancel_view = ''
+
+        with self.manage_subfolder(group, folder, subfolder) as selected_menu:
+            selected_options = selected_menu.parent_view.report_select.all_options
+            for report in reports:
+                if report in selected_options:
```
digitronik: If logic is just to check report is present in `selected_options` or not then I would just use `set` operations.

cfme/intelligence/reports/menus.py
```diff
@@ -180,6 +194,43 @@ def manage_folder(self, group, folder=None):
...
 
+    def move_reports(self, group, folder, subfolder, reports):
+        """ Moves a list of reports to a given menu
+        Args:
+            group: User group
+            folder: Parent of the subfolder where reports are to be moved.
+            subfolder: Subfolder under which the reports are to be moved.
+            reports: List of reports that are to be moved.
+        """
+        if not isinstance(reports, list):
```
digitronik: optional:
I would use `*reports` as an argument to avoid this.

cfme/intelligence/reports/menus.py
```diff
@@ -198,3 +247,12 @@ def step(self, *args, **kwargs):
...
         )
+
+
+@navigator.register(ReportMenusCollection, "All")
+class ReportMenus(CFMENavigateStep):
+    VIEW = AllReportMenusView
+    prerequisite = NavigateToAttribute("appliance.server", "CloudIntelReports")
+
+    def step(self, *args, **kwargs):
+        self.view.edit_report_menus.tree.click_path("All EVM Groups")
```
mshriver: This is a bit of an odd pattern - using the local view to navigate from prerequisite - if I can use widgets on the view that should result from this step method, why is the step method necessary?

Is `edit_report_menus` available in `self.prerequisite_view` for use here? I see its not defined in `AllReportMenusView` directly.

cfme/intelligence/reports/menus.py
```diff
@@ -180,6 +194,41 @@ def manage_folder(self, group, folder=None):
...
+            # menu. This is a work-around for that purpose.
+            reports.extend(selected_options)
+            selected_menu.parent_view.report_select.fill(reports)
+
+        if cancel_view:
+            assert cancel_view.is_displayed
+            cancel_view.cancel_button.click()
+            message = 'Edit of Report Menu for role "{}" was cancelled by the user'
+        else:
+            message = 'Report Menu for role "{}" was saved'
```
mshriver: Do we not have to click any save buttons here?

cfme/tests/intelligence/reports/test_menus.py
```diff
@@ -5,11 +5,14 @@
 import yaml
 
 from cfme import test_requirements
+from cfme.base.credential import Credential
+from cfme.rest.gen_data import users as _users
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.path import data_path
 
+pytestmark = [pytest.mark.tier(3), test_requirements.report]
```
mshriver: :+1:

cfme/tests/intelligence/reports/test_menus.py
```diff
@@ -35,21 +38,27 @@ def report_menus(group, appliance):
 
 def crud_files_reports():
     result = []
-    if not report_crud_dir.exists:
-        report_crud_dir.mkdir()
-    for file_name in report_crud_dir.listdir():
+    if not REPORT_CRUD_DIR.exists:
```
mshriver: This data_path based object is a `LocalPath` type, and `exists` is a bound method, not a property. 

```
In [14]: from cfme.utils.path import data_path

In [15]: DIR = data_path.join('reports_crud')

In [16]: DIR.exists
Out[16]: <bound method LocalPath.exists of local('/home/mshriver/repos/integration_tests/data/reports_crud')>

In [17]: DIR.exists()
Out[17]: True

```

Additionally, LocalPath has a method to do what you're trying to do here, called `ensure`. It takes boolean kwarg `dir` for a directory vs a file.

```
In [19]: DIR = data_path.join('junk_path')

In [20]: DIR.exists()
Out[20]: False

In [21]: DIR.ensure(dir=True)
Out[21]: local('/home/mshriver/repos/integration_tests/data/junk_path')

In [22]: DIR.exists()
Out[22]: True
```

cfme/tests/intelligence/reports/test_menus.py
```diff
@@ -35,21 +38,27 @@ def report_menus(group, appliance):
...
+    with REPORT_CRUD_DIR.join(request.param).open(mode="r") as rep_yaml:
         return yaml.safe_load(rep_yaml)
 
 
-@pytest.mark.tier(3)
+@pytest.fixture(scope="function")
+def get_custom_report(appliance, custom_report_values):
+    custom_report = appliance.collections.reports.create(**custom_report_values)
+    yield custom_report
+    custom_report.delete()
```
mshriver: Safer to call `delete_if_exists` here, which should be available on the entity as long as the exists method is implemented.

cfme/tests/intelligence/reports/test_menus.py
```diff
@@ -86,8 +94,10 @@ def test_shuffle_first_level(appliance, group, report_menus):
     view = navigate_to(appliance.collections.reports, "All")
     tree = view.reports.tree.read_contents()[1]
     # Select some folder that has at least 3 children
-    folders = map(lambda item: item[0],
-                  filter(lambda item: isinstance(item[1], list) and len(item[1]) >= 3, tree))
+    folders = map(
```
mshriver: Consider actually re-writing this as comprehension instead of using `lambda` (x2), `map`, and `filter`.

```
folders = {i: i[0] for i in tree if isinstance(i[1], list) and len(i[1]) >=3}

**Comment by dajoRH** [comment]
*CFME QE Bot*

File: cfme/intelligence/reports/menus.py (hunk 8 above)
Comment (valaparthvi) [reply cfme/intelligence/reports/menus.py]: Don't need to click any buttons, `manage_subfolder` context manager takes care of that.

File: cfme/intelligence/reports/menus.py (hunk 7 above)
Comment (valaparthvi) [reply cfme/intelligence/reports/menus.py]: It is defined in the prerequisite_view itself.

> why is the step method necessary?
I am not quite sure how to explain this, but I think if you understand why the `AllReportMenusView` view was added, it might help. It has been defined to assert flash messages, this was not checked earlier. `edit_report_menus` is defined in `CloudIntelReportsView`(parent_view), so it can be accessed by both `view` and `prerequisite_view`. Step method is necessary to click the `edit_report_menus` tree.

File: cfme/tests/intelligence/reports/test_menus.py (hunk 10 above)
Comment (valaparthvi) [reply cfme/tests/intelligence/reports/test_menus.py]: Ah okay! I had copy-pasted the whole thing from another test file, didn't check it in details :/

File: cfme/intelligence/reports/menus.py (hunk 7 above)
Comment (mshriver) [reply cfme/intelligence/reports/menus.py]: I was asking mostly rhetorically :)

A navigation step should use the prerequisite view to navigate, not the view that's the result of executing the `step` method.

File: cfme/base/login.py [hunk 13]
```diff
@@ -103,7 +103,6 @@ def logged_out(self):
 
     def logout(self):
         self.settings.select_item('Logout')
-        self.browser.handle_alert(wait=None)
```
mshriver: https://github.com/ManageIQ/integration_tests/pull/8848#discussion_r288146353

valaparthvi: Why do we have that `handle_alert`? I don't remember seeing any alert box on logging out.

PR #8788, opened 2019-05-06
[1LP][RFR] Fix storage navigation

digitronik:

Purpose or Intent
=================
- manager details

{{pytest: cfme/tests/storage/}}

### DIFF
--- a/cfme/modeling/base.py
+++ b/cfme/modeling/base.py
@@ -199,6 +199,25 @@ class BaseEntity(NavigatableMixin):
                 raise NotImplementedError("Delete method is not implemented.")
         return False
 
+    @property
+    def expected_details_title(self):
+        """Provides expected Details/Summary Page title; which mostly used in EntityDetailsViews
+
+        .. code-block:: python
+            expected_title = self.context['object'].expected_details_title
+        """
+        return "{} (Summary)".format(self.name)
+
+    @property
+    def expected_details_breadcrumb(self):
+        """Provides expected Details/Summary Page breadcrumb active location;
+        which mostly used in EntityDetailsViews
+
+        .. code-block:: python
+            expected_breadcrumb = self.context['object'].expected_details_breadcrumb
+        """
+        return "{}{}".format(self.name, " (Summary)" if self.appliance.version < "5.11" else "")
+
 
 @attr.s
 class CollectionProperty(object):
--- a/cfme/storage/manager.py
+++ b/cfme/storage/manager.py
@@ -108,11 +108,11 @@ class StorageManagerDetailsView(StorageManagerView):
     """The details page for Storage Manager or Provider"""
     @property
     def is_displayed(self):
-        expected_title = '{} (Summary)'.format(self.context['object'].name)
+        obj = self.context['object']
 
         return(
-            self.title.text == expected_title and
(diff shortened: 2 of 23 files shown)


digitronik:
failures not related to changes.... fixed only navigation to the manager.

cfme/storage/manager.py
```diff
@@ -109,10 +111,13 @@ class StorageManagerDetailsView(StorageManagerView):
     @property
     def is_displayed(self):
         expected_title = '{} (Summary)'.format(self.context['object'].name)
+        active_location = VersionPicker(
```
izapolsk: I'd advice to use appliance.version in this case

digitronik:
Added all other Details page navigation fix for storage model pages.

cfme/storage/manager.py
```diff
@@ -109,10 +109,14 @@ class StorageManagerDetailsView(StorageManagerView):
     @property
     def is_displayed(self):
         expected_title = '{} (Summary)'.format(self.context['object'].name)
-
+        active_location = (
```
mshriver: 1. it would be great to get rid of how many times this is implemented, optional here.

2.  I think it would be cleaner to just use expected_title, and format into it. Then we don't have multiple references to the context object's name attribute, or extra names.
```
expected_title = '{}{}'.format(self.context['object'].name, 
                                             ' (Summary)' if self.browser.product_version <= '5.10' else '')
```

digitronik: @mshriver as per new implementation in `breadcrumb` we are not getting `(Summary)` and title is same; so we need two separate references.
```
5.10
title = 'obj_name (summary)'
breadcrumb= 'obj_name (summary)'

5.11
title = 'obj_name (summary)'
breadcrumb= 'obj_name'
```

cfme/modeling/base.py
```diff
@@ -199,6 +199,21 @@ def delete_if_exists(self, *args, **kwargs):
...
+        """The method used for collecting expected title and expected breadcrumb active location
+        on Summary/ Details page of Entity.
+
+        Returns: tuple of expected title and breadcrumb active location
+        """
+
+        expected_title = "{} (Summary)".format(self.name)
+        expected_breadcrumb = (
+            expected_title
+            if self.browser.product_version < "5.11"
```
mshriver: Keep in mind the context - this has been moved from a view to an entity.

Here you want to use `self.appliance.version`, instead of trying to use the browser.

cfme/modeling/base.py
```diff
@@ -199,6 +199,21 @@ def delete_if_exists(self, *args, **kwargs):
                 raise NotImplementedError("Delete method is not implemented.")
         return False
 
+    def expected_summary_attributes(self):
```
mshriver: I think it would be better to create two properties for `entity.expected_title` and `entity.expected_breadcrumb`. I'm not 100% on those names, maybe `summary_<thing>`, or `details_<thing>`.

Whatever it is, we want to be able to make it consistent across all entities to access common entity-based expected strings. Example - collection title/breadcrumb, CRUD method page titles.

izapolsk:
please ignore one redundant check. it was added manually for testing purposes.

PR #8792, opened 2019-05-07
[1LP][RFR][NOTEST] Added new manual test case

ganeshhubale:
Added manual test case for RFE: https://bugzilla.redhat.com/show_bug.cgi?id=1533263

### DIFF
--- a/cfme/tests/cloud_infra_common/test_quota_manual.py
+++ b/cfme/tests/cloud_infra_common/test_quota_manual.py
@@ -445,7 +445,6 @@ def test_quota_with_reconfigure_resize_disks():
     Polarion:
         assignee: ghubale
         casecomponent: Infra
-        caseimportance: high
         initialEstimate: 1/6h
         testSteps:
             1. Add VMware provider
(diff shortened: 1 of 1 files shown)


cfme/tests/cloud_infra_common/test_quota_manual.py
```diff
@@ -435,3 +435,30 @@ def test_orphaned_archived_vms_get_excluded_from_used_quota_counts():
...
+        testSteps:
+            1. Add VMware provider
+            2. Provision a VM
+            3. Set tenant quota limit for storage
+            4. Resize the disk of the VM over quota limit
+        expectedResults:
+            1.
+            2.
+            3.
+            4. VM reconfiguration request for resize the disk must be exceeded with proper reason.
```
john-dupuy: Question: What does `must be exceeded` mean in this context? Is the request denied because the VM is over the quota limit? 

Small typo: `VM reconfiguration request for` _resizing_ `the disk`

ganeshhubale: Yes. It must deny that request because resizing disk over the quota limit.

PR #8793, opened 2019-05-07
[RFR] added block for BZ1707328

lina-nikiforova:
{{ pytest: -v cfme/tests/configure/test_tag.py::test_map_tagging_crud }}


john-dupuy:
failing PRT, but I believe this is because the BZ was not yet triaged, a retest run should be fine.

PR #8794, opened 2019-05-07
[1LP][RFR]Fix Import/Export for reports and widgets

valaparthvi:
1. Fix Import/Export for reports and widgets
2. Delete `import_export.py` and move Views and Navigators to different files
3. Add `import_report` and `export_report` to ReportsCollection
4. Add `import_widget` and `export_widget` to DasboardReportWidgetsCollection
5. Automate test cases and remove corresponding manual test cases:
    1. `test_import_widget`
    2. `test_import_report`
    3. `test_export_widget`
    4. `test_export_report`
6. Add yaml files for report and widget.


cfme/intelligence/reports/reports.py
```diff
@@ -451,6 +468,45 @@ def create(self, **values):
...
+    ):
+        """
+        Import yaml files containing report data
+        Args:
+            filename (str): Complete path to the file
+            preserve_owner (bool): If true, original owner of the report will be preserved
+            overwrite (bool): If true, a possible duplicate of the report will be overwritten
+        """
+        view = navigate_to(self, "ImportExport")
+        assert view.is_displayed
```
digitronik: Please remove this... navigation will take care of this.

cfme/intelligence/reports/reports.py
```diff
@@ -451,6 +468,45 @@ def create(self, **values):
...
+            }
+        )
+        view.upload_button.click()
+        view.flash.assert_no_error()
+
+    def export_report(self, reports):
+        """
+        Export custom report
+        Args:
+            reports (str or lst): a single report name or a list of report names
```
digitronik: typo

cfme/intelligence/reports/reports.py
```diff
@@ -451,6 +468,45 @@ def create(self, **values):
...
+        """
+        if not isinstance(reports, list):
+            reports = [reports]
+
+        view = navigate_to(self, "ImportExport")
+        assert view.is_displayed
```
digitronik: same as above

cfme/intelligence/reports/widgets/__init__.py
```diff
@@ -132,6 +135,46 @@ def create(self, widget_class, *args, **kwargs):
...
 
+    def import_widget(self, filename, cancel=False):
+        """
+        Import yaml files containing widget data
+        Args:
+            filename (str): Complete path to the file
+            cancel (bool): If true, widgets will not be imported
+        """
+        view = navigate_to(self, "ImportExport")
+        assert view.is_displayed
```
digitronik: ^^

cfme/intelligence/reports/widgets/__init__.py
```diff
@@ -132,6 +135,46 @@ def create(self, widget_class, *args, **kwargs):
         view.flash.assert_no_error()
         return dashboard_widget
 
+    def import_widget(self, filename, cancel=False):
+        """
+        Import yaml files containing widget data
+        Args:
+            filename (str): Complete path to the file
```
digitronik: Optional:
I would use `filepath`

cfme/intelligence/reports/widgets/__init__.py
```diff
@@ -132,6 +135,46 @@ def create(self, widget_class, *args, **kwargs):
...
+            cancel (bool): If true, widgets will not be imported
+        """
+        view = navigate_to(self, "ImportExport")
+        assert view.is_displayed
+
+        view.fill({"upload_file": filename})
+        view.upload_button.click()
+        view.flash.assert_no_error()
+
+        import_view = self.create_view(ImportExportWidgetsCommitView)
```
digitronik: optional:
I think better to add some `wait` as its upload operation may be some time we can face some lag.

cfme/tests/intelligence/reports/test_import_export_reports_widgets.py
```diff
@@ -0,0 +1,125 @@
...
+            2. Check if report created with import is same as the expected report.
+
+    """
+    file_path, report = report_file
+    collection = appliance.collections.reports
+
+    collection.import_report(file_path)
+    view = collection.create_view(ImportExportCustomReportsView)
+    assert view.is_displayed
+    view.flash.assert_message("Imported Report: ", partial=True)
```
digitronik: Q: why a partial message match? Is it not possible to replicate the exact flash message?

valaparthvi: The flash messages changes slightly, depending on the data that was imported, so I had partially asserted the message, but I guess it will be fine considering we know what data is getting imported.

cfme/tests/intelligence/reports/test_import_export_reports_widgets.py
```diff
@@ -0,0 +1,125 @@
...
+    _, widget = widget_file
+    collection = appliance.collections.dashboard_report_widgets
+    collection.create(
+        widget_class=getattr(collection, widget.TITLE.upper()),
+        title=widget.title,
+        description=widget.description,
+        filter=widget.filter,
+        active=widget.active,
+    )
+    collection.export_widget(widget.title)
```
mshriver: This needs additional assertion/verification. The `export_widget` method itself only asserts there is no error type flash message, is there a success flash message to be validated here? Can we further verify the result of selecting widget export?

cfme/tests/intelligence/reports/test_import_export_reports_widgets.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+    _, report = report_file
+    collection = appliance.collections.reports
+    collection.create(
+        menu_name=report.menu_name,
+        title=report.title,
+        base_report_on="VMs and Instances",
+        report_fields=["Archived", "Autostart", "Boot Time"],
+    )
+    collection.export_report(report.menu_name)
```
mshriver: Same as widget export - I think this needs some additional validation.

cfme/intelligence/reports/widgets/__init__.py
```diff
@@ -249,3 +326,12 @@ class BaseEditDashboardWidgetStep(BaseDashboardWidgetDetailsStep):
...
         self.view.configuration.item_select("Edit this Widget")
+
+
+@navigator.register(DashboardReportWidgetsCollection, "ImportExport")
+class ImportExportWidgets(CFMENavigateStep):
+    VIEW = ImportExportWidgetsView
+    prerequisite = NavigateToAttribute("appliance.server", "CloudIntelReports")
+
+    def step(self, *args, **kwargs):
+        self.view.import_export.tree.click_path("Import / Export", "Widgets")
```
mshriver: Please use `self.prerequisite_view` in the step method for the navigation destinations in this PR. There is more than one spot where the widget being accessed is in the (common ancestor) prerequisite view.  Its a better pattern to use the prerequisite view to navigate to the target view for this navigation class.

cfme/intelligence/reports/widgets/__init__.py
```diff
@@ -132,6 +136,41 @@ def create(self, widget_class, *args, **kwargs):
...
+            view.flash.assert_message("Widget import cancelled")
+        else:
+            import_view.commit_button.click()
+            import_view.flash.assert_no_error()
+
+    def export_widget(self, *widgets):
+        """
+        Export custom widgets
+        Args:
+            widgets (str, list): a single widget name or a list of widget names
```
mshriver: I think a better way to describe this type of arg is `A single or many widget names`

mshriver: Its actually always a tuple, regardless of whether one or many strings are passed.

cfme/tests/intelligence/reports/test_import_export_reports_widgets.py
valaparthvi: There is no message displayed when we export the widget, it just gets exported. I don't think there is a sure-shot way of further testing this. If we could check the files downloaded, we could further verify it, but doing that seems more trouble than worth for.

cfme/tests/intelligence/reports/test_import_export_reports_widgets.py
valaparthvi: Same as above.

cfme/intelligence/reports/widgets/__init__.py
mshriver: @valaparthvi this is a required update for me to merge. Thanks!

cfme/intelligence/reports/reports.py
```diff
@@ -267,6 +270,30 @@ def is_displayed(self):
...
+
+    overwrite = Checkbox("overwrite")
+    preserve_owner = Checkbox("preserve_owner")
+
+    items_for_export = Select(id="choices_chosen")
+    export_button = Button(id="export_button")
+
+    @property
+    def is_displayed(self):
+        return (
```
jaryn: Often it happens that something in CFME changes and the `is_displayed` doesn't return `True`.  Then one needs to find why. If you created
```suggestion
result = (
            self.in_intel_reports,
            self.title.text == "Import / Export",
            self.subtitle.text == "Custom Reports",
            self.import_export.tree.currently_selected == ["Import / Export", "Custom Reports"])
)
return all(results)
```
Then it would be slighly easier to debug why the return value is not what is expected. In the debuger we could just check the results tuple

valaparthvi: Hmm, that's nice! Will add that. Thanks @jaryn  :)

mshriver: To be clear, there is also a clear downside to the above strategy. While useful for debugging, it has an impact on efficiency and time cost of the UI checks.

When chaining together `and` statements in python, once one statement returns `False`, the chain stops. So if `in_intel_reports` is `False`, and we store and then call `all()`, we still run every other check, reading multiple widgets and making string/list comparisons.  If we instead are chaining those calls with `and`, we would have saved the additional time of reading 3 other widgets and doing 3 str/list comparisons.

This is why we use chained `and` pattern in **every** `is_displayed` method in CFME, for efficiency. If you have a failing check and need to use the debugger, then you can easily make this change if that helps you debug, to determine what's failing.  You can also just set a breakpoint and read all the widgets to compare them manually/visually.

dajoRH:
*CFME QE Bot*

PR #8803, opened 2019-05-08
[1LP][RFR] Automate test: test_delete_automate_domain_via_api

valaparthvi:
Changes:
1. Add a fixture `domain_rest` to create domain.
2. Add 2 tests:
    1. `test_delete_automate_domain_from_collection`
    2. `test_delete_automate_domain_from_detail`

{{ pytest: cfme/tests/automate/test_rest.py -k "test_delete_automate_domain"  --use-template-cache -sqvvv }}

### DIFF
--- a/cfme/tests/automate/test_rest.py
+++ b/cfme/tests/automate/test_rest.py
@@ -1,14 +1,25 @@
 # -*- coding: utf-8 -*-
 """REST API specific automate tests."""
+import fauxfactory
 import pytest
 
 from cfme import test_requirements
+from cfme.utils.rest import delete_resources_from_collection
+from cfme.utils.rest import delete_resources_from_detail
 
 
-pytestmark = [test_requirements.rest]
+pytestmark = [test_requirements.rest, pytest.mark.tier(3)]
 
 
-@pytest.mark.tier(3)
+@pytest.fixture(scope="function")
+def domain_rest(appliance, domain):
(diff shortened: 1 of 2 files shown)


cfme/tests/automate/test_rest.py
```diff
@@ -29,3 +40,26 @@ def _do_query(**kwargs):
     filtered_depth = _do_query(depth='-1', search_options='state_machines')
     assert len(full_depth) > len(more_depth) > len(rest_api.collections.automate)
     assert len(full_depth) > len(filtered_depth) > 0
+
+
+@pytest.mark.ignore_stream("5.10")
+@pytest.mark.parametrize("method", ["POST", "DELETE"])
+def test_delete_automate_domain_from_detail(appliance, domain_rest, method):
```
john-dupuy: Optional: As far as I can tell, `test_delete_automate_domain` can be parametrized against `from_detail` and `from_collection`, though there may be some complication with `method`, so it may be more trouble than it's worth. I'll leave it as optional.

valaparthvi: Yep, more trouble than worth for! Actually, this is the pattern that is used throughout our code base, define 2 separte tests for deleting it via from_detail and from_collection. I could use `uncollectif` if I integrate these 2 tests, but this seems like the way to go!

PR #8805, opened 2019-05-08
[1LP][RFR] Automate test: test_welcoming_page

valaparthvi:
Changes:
1. Add a new widget `add_button`
2. Automate test:  `test_welcoming_page`

{{ pytest: cfme/tests/webui/test_general_ui.py::test_welcoming_page --use-template-cache -sqvvv }}

### DIFF
--- a/cfme/base/login.py
+++ b/cfme/base/login.py
@@ -19,6 +19,7 @@ class BaseLoggedInPage(View):
     configuration_settings = Text('.//li[./a[@title="Configuration"]]')      # 5.11+
     settings = SettingsNavDropdown('.//li[./a[@id="dropdownMenu2"]]')
     navigation = VerticalNavigation('#maintab')
+    add_button = Text('//*[@id="main_div"]//div/a')
 
     @property
     def is_displayed(self):
--- a/cfme/base/login.py
+++ b/cfme/base/login.py
@@ -19,7 +19,6 @@ class BaseLoggedInPage(View):
     configuration_settings = Text('.//li[./a[@title="Configuration"]]')      # 5.11+
     settings = SettingsNavDropdown('.//li[./a[@id="dropdownMenu2"]]')
     navigation = VerticalNavigation('#maintab')
-    add_button = Text('//*[@id="main_div"]//div/a')
 
     @property
     def is_displayed(self):
--- a/cfme/common/provider_views.py
+++ b/cfme/common/provider_views.py
@@ -303,7 +303,7 @@ class ProviderEntitiesView(BaseEntitiesView):
 
 class ProvidersView(BaseLoggedInPage):
     """
-     represents Main view displaying all providers
+    represents Main view displaying all providers
     """
     @property
     def is_displayed(self):
@@ -314,7 +314,7 @@ class ProvidersView(BaseLoggedInPage):
     sidebar = View.nested(ProviderSideBar)
     search = View.nested(Search)
     including_entities = View.include(ProviderEntitiesView, use_parent=True)
-    add_button = Text('//*[@id="main_div"]//div/a')
+    add_button = Button('Add a Provider')
 
     @View.nested
     class my_filters(Accordion):  # noqa
--- a/cfme/common/provider_views.py
+++ b/cfme/common/provider_views.py
@@ -314,6 +314,7 @@ class ProvidersView(BaseLoggedInPage):
     sidebar = View.nested(ProviderSideBar)
     search = View.nested(Search)
     including_entities = View.include(ProviderEntitiesView, use_parent=True)
+    add_button = Text('//*[@id="main_div"]//div/a')
 
     @View.nested
     class my_filters(Accordion):  # noqa
--- a/cfme/tests/webui/test_general_ui.py
+++ b/cfme/tests/webui/test_general_ui.py
@@ -269,10 +269,18 @@ def test_welcoming_page(appliance, has_no_providers):
 def test_add_button_on_provider_all_page(
     appliance, provider_type, add_view, has_no_providers
 ):
+    """
+    This test checks if the `Add a Provider` button is displayed on a providers all page
+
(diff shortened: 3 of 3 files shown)


cfme/base/login.py
```diff
@@ -19,6 +19,7 @@ class BaseLoggedInPage(View):
     configuration_settings = Text('.//li[./a[@title="Configuration"]]')      # 5.11+
     settings = SettingsNavDropdown('.//li[./a[@id="dropdownMenu2"]]')
     navigation = VerticalNavigation('#maintab')
+    add_button = Text('//*[@id="main_div"]//div/a')
```
john-dupuy: There are many pages that use this class but not all of them have the `add_button` widget on them. Also, why not use the button widget for this? 

My thought is that this button should be on the `InfraProvidersView` page.  After doing the navigatation via `view = navgiate_to(appliance.server, "LoggedIn")`, you then can create the view and then and assert that the button is present. What do you think?

cfme/tests/webui/test_general_ui.py
```diff
@@ -226,6 +227,28 @@ def test_infrastructure_filter_20k_vms(appliance, create_20k_vms):
     assert items_amount >= 20000, 'Vms count is less than should be filtered'
 
 
+@pytest.mark.ignore_stream("5.10")
+@pytest.mark.tier(2)
+def test_welcoming_page(appliance, has_no_providers):
```
jawatts: I think this should be taken a step further and parameterized for all the compute types.

valaparthvi: @jawatts The idea is good, but I think it would be better to write a separate test for it. This test is about checking the welcome page that the user first sees on login when there are no providers.

valaparthvi: @jawatts That is a good idea, but this test checks the view user first sees when they login into an appliance that has no providers. I think parametrizing the test would slightly deviate from the case that we're trying to test. But I can add a separate test for this. Let me know what you think.

jawatts: @valaparthvi that sounds good to me. If you could add a `TODO` somewhere, that be great.

cfme/common/provider_views.py
```diff
@@ -314,6 +314,7 @@ def is_displayed(self):
     sidebar = View.nested(ProviderSideBar)
     search = View.nested(Search)
     including_entities = View.include(ProviderEntitiesView, use_parent=True)
+    add_button = Text('//*[@id="main_div"]//div/a')
```
john-dupuy: Q: Is there a reason that `Button` can't be used here?

valaparthvi: Oh no! I forgot to change that to `Button`. Will do that right away!

cfme/tests/webui/test_general_ui.py
valaparthvi: Cool! I'll add a new test case along with this.

cfme/common/provider_views.py
valaparthvi: @john-dupuy I think I might now have a reason to use `Text` instead of `Button`. I cannot use locator/xpath of the element with `Button`, and using xpath is a sure-shot way to access the element. I think I'll stick to `Text`.

john-dupuy: @valaparthvi Ah I see, I'm surprised it doesn't work. You should be able to do `Button("Add a Provider")` and the locator should take care of the rest. 

```python,
    # TODO: Handle input value the same way as text for other tags
    def __locator__(self):
        return (
            './/*[(self::a or self::button or (self::input and (@type="button" or @type="submit")))'
            ' and contains(@class, "btn") {}]'.format(self.locator_conditions))
```

john-dupuy: Let me try and see

john-dupuy: Yep, just was able to confirm, with `Button` you pass the text of the button, not the xpath, so using `Button("Add a Provider")` will work. The `__init__` function of `Button` will build the xpath for you. 

https://github.com/RedHatQE/widgetastic.patternfly/blob/cf5f288254dca7f5ae00f17a50cd377119198c6a/src/widgetastic_patternfly/__init__.py#L125-L143

cfme/tests/webui/test_general_ui.py
```diff
@@ -226,6 +231,63 @@ def test_infrastructure_filter_20k_vms(appliance, create_20k_vms):
...
+    ("provider_type", "add_view"),
+    [
+        ("infra_providers", InfraProviderAddView),
+        ("cloud_providers", CloudProviderAddView),
+        ("containers_providers", ContainerProviderAddView),
+        ("physical_providers", PhysicalProviderAddView),
+    ],
+    ids=["infra", "cloud", "container", "physical"],
+)
+def test_add_button_on_provider_all_page(
```
valaparthvi: @jawatts Here's the test that we talked about.

cfme/common/provider_views.py
```diff
@@ -314,6 +314,7 @@ def is_displayed(self):
     sidebar = View.nested(ProviderSideBar)
     search = View.nested(Search)
     including_entities = View.include(ProviderEntitiesView, use_parent=True)
+    add_button = Button('Add a Provider')
```
valaparthvi: @john-dupuy It worked like magic!

cfme/common/provider_views.py
valaparthvi: @john-dupuy It worked like magic!

cfme/tests/webui/test_general_ui.py
valaparthvi: @jawatts I have added the test, you can review it [here](https://github.com/ManageIQ/integration_tests/pull/8805/files#diff-b97f49d8a6659114ea4b051eb76698e1R269).

PR #8808, opened 2019-05-08
[RFR] Fix parallelizer decode

mshriver:
We're hitting an exception trying to call decode on the exception object, because the call to print_message was passing the wrong object.

Changed things up a bit, no longer trying to pass the exception itself to print_message.

Cleaned up some print_message calls otherwise as well.

### DIFF
--- a/cfme/fixtures/parallelizer/__init__.py
+++ b/cfme/fixtures/parallelizer/__init__.py
@@ -314,14 +314,13 @@ class ParallelSession(object):
                 yield
                 if polls % poll_report_modulo == 0:
                     remaining_time = int(poll_num_sec - (time() - start_time))
-                    self.print_message(
-                        '{} still shutting down, '
-                        'will continue polling for {} seconds '
-                        .format(slaveid.decode('ascii'), remaining_time), blue=True)
+                    self.print_message('{} shutting down, will continue polling for {} seconds'
+                                       .format(slaveid.decode('ascii'), remaining_time),
+                                       blue=True)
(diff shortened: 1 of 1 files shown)


cfme/fixtures/parallelizer/__init__.py
```diff
@@ -581,14 +580,13 @@ def provs_of_tests(test_group):
...
                 app = slave.appliance
-                self.print_message(
-                    'cleansing appliance', slave, purple=True)
+                self.print_message('removing providers from appliance', slave, purple=True)
                 try:
                     app.delete_all_providers()
                 except Exception as e:
-                    self.print_message(
-                        'cloud not cleanse', slave, red=True)
-                    self.print_message('error:', e, red=True)
```
mshriver: Passing `e` here was triggering an exception in print_message, as we were calling `decode` on it if it wasn't a string.

There's an assumption in print_message where its calling decode on anything that isn't a string, assuming its binary string.

I'm not addressing the behavior in print_message at the moment, as it will soon be impacted by py3 conversion.

dajoRH:
Lint report for commit d1638156e08a8d0cc46b1692ca8e27acf42af854:

`scripts/sync_template_tracker.py`:
- :red_circle: Line 13:1: **F401** *'tabulate.tabulate' imported but unused*
- :bangbang: Line 156:101: **E501** *line too long (115 > 100 characters)*

**No commit flaws detected**.

*CFME QE Bot*

PR #8811, opened 2019-05-09
[1LP][RFR][NOTEST] Add test steps to the replication tests.

jaryn:
__Updating the replication tests__ with the testSteps

### DIFF
--- a/cfme/tests/test_replication.py
+++ b/cfme/tests/test_replication.py
@@ -14,6 +14,16 @@ def test_replication_powertoggle():
         casecomponent: Replication
         caseimportance: critical
         initialEstimate: 1/12h
+        testSteps:
+            1. Have a VM created in the provider in the Remote region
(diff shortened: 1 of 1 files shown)


cfme/tests/test_replication.py
```diff
@@ -170,23 +199,10 @@ def test_replication_appliance_add_multi_subscription():
...
         initialEstimate: 1/4h
         startsin: 5.7
-    """
-    pass
-
-
-@pytest.mark.manual
-@test_requirements.replication
-@pytest.mark.tier(1)
-def test_replication_appliance_set_type_remote_ui():
```
jaryn: This test is quite redundant as the its steps are included in every test needing the replication working.

jawatts:
@jaryn Since `testSteps` are being added here, I would expect to see `expectedResults`. Was there a reason behind that here?

PR #8816, opened 2019-05-10
[WIPTEST] Fixed host relationships for scvmm

ganeshhubale:
{{pytest: cfme/tests/cloud_infra_common/test_relationships.py -k 'test_host_relationships' --use-provider=scvmm --long-running -vv}}

Fixed Error:
```

>       ][-1]  # FIXME this is raising an IndexError for being out of range, list must be empty
E       IndexError: list index out of range

cfme/infrastructure/cluster.py:182: IndexError
IndexError
list index out of range
```


ganeshhubale:
No error in current automation runs.

PR #8818, opened 2019-05-11
[1LP][RFR] Put in order storage tests

mmojzis:
Deleted duplicate tests from test_manual.py
Reassigned several tests
Moved some tests from test_manual.py
@apagac Why were some storage snapshot tests assigned to you?

### DIFF
--- a/cfme/tests/storage/test_object_store_container.py
+++ b/cfme/tests/storage/test_object_store_container.py
@@ -2,6 +2,7 @@
 import fauxfactory
 import pytest
 
+from cfme import test_requirements
 from cfme.cloud.provider.openstack import OpenStackProvider
 from cfme.utils.wait import wait_for
 
@@ -50,3 +51,26 @@ def test_add_remove_tag(container):
     container.remove_tag(added_tag)
     tag_available = container.get_tags()
     assert not tag_available
+
+
+@pytest.mark.manual
+@test_requirements.storage
+def test_storage_object_store_container_edit_tag_openstack():
+    """
+    Requirs:
+    OpenstackProvider
+    1) Add Object Store Container
+    2) go to summery pages
+    2) add tag : [Policy > Edit Tags]
+    3) Verify the tag is assigned
+    4) remove tag: [Policy > Edit Tags]
+    5) Verify the tag is removed
+
+    Polarion:
+        assignee: mmojzis
+        caseimportance: medium
+        casecomponent: Cloud
+        initialEstimate: 1/8h
+        startsin: 5.7
+    """
+    pass
--- a/cfme/tests/storage/test_object_store_container.py
+++ b/cfme/tests/storage/test_object_store_container.py
@@ -57,20 +57,24 @@ def test_add_remove_tag(container):
 @test_requirements.storage
 def test_storage_object_store_container_edit_tag_openstack():
     """
-    Requirs:
-    OpenstackProvider
-    1) Add Object Store Container
-    2) go to summery pages
-    2) add tag : [Policy > Edit Tags]
-    3) Verify the tag is assigned
-    4) remove tag: [Policy > Edit Tags]
-    5) Verify the tag is removed
+    Requires:
+        OpenstackProvider
 
     Polarion:
-        assignee: mmojzis
+        assignee: anikifor
         caseimportance: medium
         casecomponent: Cloud
         initialEstimate: 1/8h
         startsin: 5.7
+        testSteps:
+            1. Add Object Store Container
+            2. go to summery pages
+            3. add tag : [Policy > Edit Tags]
+            4. remove tag: [Policy > Edit Tags]
+        expectedResults:
+            1.
+            2.
+            3. Verify the tag is assigned
+            4. Verify the tag is removed
     """
     pass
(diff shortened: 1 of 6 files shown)


apagac:
Seems like it happened because @mshriver wanted to assign all testcases to FA owners, and he saw the string 'snapshot' so it landed on me :)

mshriver:
@mmojzis @apagac That's correct - some tests hit two different focus areas, and I just picked one when making assignee required. Thanks for re-assigning!

cfme/tests/storage/test_volume.py
```diff
@@ -157,6 +159,25 @@ def test_storage_volume_crud(appliance, provider, from_manager):
                                unblock=lambda provider: provider.one_of(EC2Provider))])
 @pytest.mark.tier(1)
 def test_storage_volume_attach_detach(appliance, provider, instance_fixture, from_manager):
+    """ Test storage volume attach/detach
+
+    prerequisites:
+        * Storage provider
+        * Instance
+
+    Steps:
```
izapolsk: please move under Polarion section following our parsing agreements.

cfme/tests/storage/test_volume.py
```diff
@@ -173,6 +194,32 @@ def test_storage_volume_attach_detach(appliance, provider, instance_fixture, fro
     volume.delete()
 
 
+@pytest.mark.manual
+@test_requirements.storage
+def test_storage_volume_attached_delete():
+    """
+    Requires:
+    RHCF3-21779 - test_storage_volume_attach[openstack]
+    Steps to test:
```
izapolsk: same thing

cfme/tests/storage/test_volume.py
mmojzis: Sorry, I don't understand. Can you please give me an example?
Thanks

dajoRH:
*CFME QE Bot*

cfme/tests/storage/test_volume.py
mmojzis: Both fixed

cfme/tests/storage/test_object_store_container.py
```diff
@@ -50,3 +51,29 @@ def test_add_remove_tag(container):
...
+        assignee: anikifor
+        caseimportance: medium
+        casecomponent: Cloud
+        initialEstimate: 1/8h
+        startsin: 5.7
+
+        Requires:
+            OpenstackProvider
+
+        Steps:
```
mshriver: This needs to be `testSteps`, and please put `expectedResults` as well, splitting what's here between the two.

mshriver: Apply to all below. Please see `conf/polarion_tools.yaml` for the keys, or other tests for examples.

mmojzis: Added for manual tests.
But now I don't actually know what to do when there is multiple steps test and after which test which result is expected.
For automated tests I didn't add expectedResults as I don't think it's needed when it's all documented in the code? @mshriver

dajoRH:
*CFME QE Bot*

cfme/tests/storage/test_object_store_container.py
mshriver: We don't require testSteps/expectedResults for automation - Just inline comments, and comments as to the content of the test in the regular docblock, out of the Polarion section.

I don't quite follow your other concern, but I think you might be getting at not having expected results for some teststeps.  You can do the following:

```
testSteps:
  1. step one
  2. step two
  3. step three
expectedResults:
  1. thing happens
  2. 
  3. other thing happens
```

mmojzis: So expectedResults number should be same number as the step where do I expect that result? So I can have for example 10 steps, but I want expectedResult only for step 7 and 10 so in expected results I will have
```
expectedResults:
    7. something is equal something
    10. something is deleted
```
Like this?

mshriver: 95% correct - you still need to have empty `1.`, `2.`, etc

mmojzis: @mshriver Reworked it I hope correctly.

cfme/tests/storage/test_volume.py
```diff
@@ -157,6 +157,25 @@ def test_storage_volume_crud(appliance, provider, from_manager):
...
 def test_storage_volume_attach_detach(appliance, provider, instance_fixture, from_manager):
+    """ Test storage volume attach/detach
+    Polarion:
+        assignee: mmojzis
+        initialEstimate: 1/4h
+        casecomponent: Cloud
+        startsin: 5.7
+        caseimportance: high
+
+        prerequisites:
```
mshriver: Two related things here:
Polarion flake parser sees the empty line as the end of the polarion metadata section.

1. prerequisites isn't a valid polarion key, if you want to include it in polarion , it should be `'setup:'`.
2. testSteps should not have an empty line before it, it needs to immediately follow caseimportance (or whatever previous polarion metadata line may be)

cfme/tests/storage/test_object_store_container.py
mshriver: Looking much better, thanks!! One last comment about empty lines in the docblocks and you'll be good to go.

PR #8821, opened 2019-05-13
[RFR] Add a wait_for in getting the server_names

john-dupuy:
This test was failing with the following error:

```
cfme/tests/cli/test_evmserverd.py:47: 

.0 = <listiterator object at 0x7f284f5e9390>

    server[server_name_key].rstrip('*')  # evm* shows up in status
>   for server in appliance.ssh_client.status["servers"]
        }
E       KeyError: 'Server'

cfme/tests/cli/test_evmserverd.py:47: KeyError
```

Adding a wait_for should in getting the server_name should remedy this. 

{{ pytest: --long-running --use-template-cache cfme/tests/cli/test_evmserverd.py::test_evmserverd_stop }}

### DIFF
--- a/cfme/tests/cli/test_evmserverd.py
+++ b/cfme/tests/cli/test_evmserverd.py
@@ -42,10 +42,22 @@ def test_evmserverd_stop(appliance, request):
 
     server_name_key = 'Server'
 
+    @wait_for_decorator(timeout="2m", delay=5)
+    def get_server_names():
+        """ Wait for the server name to appear on the appliance.
+            This test was experiencing KeyErrors before this wait was included.
+        """
+        return all(
+            [bool(
+                server.get(server_name_key, False)
+            ) for server in appliance.ssh_client.status["servers"]]
+        )
+
     server_names = {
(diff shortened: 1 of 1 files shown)


cfme/tests/cli/test_evmserverd.py
```diff
@@ -42,10 +42,19 @@ def test_evmserverd_stop(appliance, request):
 
     server_name_key = 'Server'
 
+    @wait_for_decorator(timeout="2m", delay=5)
+    def get_server_names():
+        for server in appliance.ssh_client.status["servers"]:
+            if server.get(server_name_key):
```
mshriver: This can likely be flattened a bit to: `return bool(server.get(server_name_key, False))`

Please include a docblock here of what this is waiting for, the behavior we're trying to mitigate. This will stop waiting once a single element in the status has `server_name_key`, why doesn't it wait for _all_ of them?

john-dupuy: Good point, thanks for pointing out my faulty logic :smile: !!

PR #8825, opened 2019-05-14
[WIP][NOTEST]Enable black on pre-commit

valaparthvi:
Changes:
1. Enable black to run on `cfme/tests` in pre-commit

How black works while running pre-commit locally?  
1. Add your changes to git.
2. Commit your changes to git - If the files don't comply with Black formatting,  pre-commit will fail at Black and apply Black to the files that were added to git, else Black passes, just like any other pre-commit hook.
3. If the Black hook failed in step 2, re-add your changes to git and commit, Black should pass this time.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,7 @@ repos:
   - id: black
     args: [--safe, --line-length, '100']
     language_version: python3.6
-    <<: *exclude
+    exclude: ^(cfme/(?!tests)|scripts)
 - repo: https://github.com/pre-commit/pre-commit-hooks
   rev: v2.0.0
   hooks:


valaparthvi:
@mshriver @izapolsk Can you review this?
Travis currently fails because pre-commit applied black on all the files and thus it failed. But this means it's working!

@mshriver I haven't applied black to tests module, like we talked, will take it forward once you both approve these changes.

digitronik:
@valaparthvi 
My personal opinion... not to go with a black hook in pre-commit for now.
This makes large changes even if you took `cfme/tests`

My personal suggestion;
step-1: Mike will announce that start applying black in a team call.
step-2: Each FA owner will start applying black to respective model pages and tests.
Even, we can add `co-author` for blackifying PR; so we can move with collaboration.
step-3: When we clear step-1 and step-2; we are free to enable black hook.

valaparthvi:
@digitronik I agree this will be a huge change, and also that if we apply black to all the files in a single PR, contributions by other people will be destroyed in some way, definitely do not want to do that. I like your suggestion. However, if we're actively starting to apply black to all the files, I don't think keeping this hook enabled is such a bad idea.

mshriver:
@digitronik I disagree, trying to have all contributors apply black to areas they own will end up with fragmented conversion, and convoluted regex expressions in our exclude statement.  I don't expect @valaparthvi to be contributing all the PR's that apply `black`, but I don't want to levy a requirement on FA owners to apply this to modules they 'own'.

We should not accept PR's that apply black to modules, but don't include those modules in pre-commit's `exclude` or `files` lines.  Then we aren't consistently applying formatting, and the next time that file is changed, the contributor might not follow the black formatting. pre-commit is necessary to enforce the formatting once its converted.

@valaparthvi do you have the pre-commit hooks installed locally? This PR should include the change to pre-commit **and** the changes to the files now covered by the plugin.

digitronik:
@mshriver :+1:

valaparthvi:
@mshriver I have pre-commit enabled locally, but it did not apply black to the files when I committed these changes, it simply skipped, since this config file was one of the excluded files.

dajoRH:
*CFME QE Bot*

PR #8826, opened 2019-05-14
[1LP][RFR] test_unassigned button behavior for service obj

digitronik:

Purpose or Intent
=================
- BZ-1653195
- BZ over BZ :) 1709870

{{pytest: cfme/tests/automate/custom_button/test_service_objects.py::test_custom_button_unassigned_behavior_objs -v}}

### DIFF
--- a/cfme/tests/automate/custom_button/test_service_objects.py
+++ b/cfme/tests/automate/custom_button/test_service_objects.py
@@ -1,8 +1,11 @@
 import fauxfactory
 import pytest
+from widgetastic_patternfly import Button
 from widgetastic_patternfly import Dropdown
 
 from cfme import test_requirements
+from cfme.infrastructure.provider.virtualcenter import VMwareProvider
+from cfme.markers.env_markers.provider import ONE
 from cfme.services.myservice import MyService
 from cfme.tests.automate.custom_button import CustomButtonSSUIDropdwon
 from cfme.tests.automate.custom_button import log_request_check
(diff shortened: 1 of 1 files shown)


cfme/tests/automate/custom_button/test_service_objects.py
```diff
@@ -713,3 +716,67 @@ def test_custom_button_open_url_service_obj(objects, button_group):
...
+
+    with appliance.context.use(ViaUI):
+        button = gp.buttons.create(
+            text="btn_{}".format(fauxfactory.gen_alphanumeric(3)),
+            hover="btn_hover_{}".format(fauxfactory.gen_alphanumeric(3)),
+            system="Request",
+            request="InspectMe",
+        )
+        assert button.exists
+        request.addfinalizer(button.delete_if_exists)
```
jaryn: I am not sure but I am worried that the button wont't be deleted in the correct context.

cfme/tests/automate/custom_button/test_service_objects.py
```diff
@@ -713,3 +716,67 @@ def test_custom_button_open_url_service_obj(objects, button_group):
...
+
+
+@pytest.mark.provider([VMwareProvider], override=True, scope="module", selector=ONE)
+def test_custom_button_unassigned_behavior_objs(
+    appliance, setup_provider, unassigned_btn_setup, request
+):
+    """ Test unassigned custom button behavior
+
+    Note: Service unassigned custom button should display on SSUI but not OPS UI.
+    For other than service objects also fallows same behaviour l.e. not display on OPS UI.
```
sshveta: Typo : 
Note: Service unassigned custom button should display on SSUI but not OPS UI.
    For other than service objects also **follows** same behaviour **i.e.** not display on OPS UI.

cfme/tests/automate/custom_button/test_service_objects.py
```diff
@@ -713,3 +716,67 @@ def test_custom_button_open_url_service_obj(objects, button_group):
...
+
+
+@pytest.fixture(params=["Service", "Other"], scope="module")
+def unassigned_btn_setup(request, appliance, provider, service):
+    if request.param == "Service":
+        btn_type = request.param
+        obj = MyService(appliance, name=service.name)
+        destinations = [ViaUI, ViaSSUI]
+    else:
+        btn_type = "Provider"
```
sshveta: All the other categories(Vm , Instance) will come under Provider ?

cfme/tests/automate/custom_button/test_service_objects.py
```diff
@@ -713,3 +716,67 @@ def test_custom_button_open_url_service_obj(objects, button_group):
...
+    Polarion:
+        assignee: ndhandre
+        initialEstimate: 1/6h
+        caseimportance: medium
+        caseposneg: positive
+        testtype: functional
+        startsin: 5.8
+        casecomponent: CustomButton
+        testSteps:
+            1. Create unassigned custom button on service and one other custom button object.
```
sshveta: I created one button under "Unassigned" for Service and it shows up in OSP UI service .
SO the behaviour is same for OSP and SUI right? 
Let me know if I am checking at the wrong place.
Appliance : https://10.8.196.131 , check button on service "name"

digitronik: https://bugzilla.redhat.com/show_bug.cgi?id=1709870

For more info
https://bugzilla.redhat.com/show_bug.cgi?id=1653195

cfme/tests/automate/custom_button/test_service_objects.py
digitronik: I don't think we need to cover more objects. we have 28 means 27 other than service. so I just choose one other than service.

cfme/tests/automate/custom_button/test_service_objects.py
digitronik: I'm waiting for the final call from PM or some Developer on BZ.
Just need to change one line in assert and test will be ready to merge.

cfme/tests/automate/custom_button/test_service_objects.py
sshveta: In that case we can just provide the second parameter as "Provider"

cfme/tests/automate/custom_button/test_service_objects.py
digitronik: @sshveta as per talk with Harpreet it should not but still no PM ack on behavior. I  would like to go with this for now. As changes appear I will take care of those.

cfme/tests/automate/custom_button/test_service_objects.py
```diff
@@ -713,3 +716,66 @@ def test_custom_button_open_url_service_obj(objects, button_group):
...
+            2. Check destinations OPS UI should not display unassigned button but SSUI should.
```
sshveta: Steps say "OPS UI should not display unassigned button" but you are asserting button.exists

digitronik: @sshveta 
`L771` point `button.exists` means check button created properly or not it's not checking on the destination page.
`L781` is actual check where we are checking OPS and SSUI difference.

digitronik:
@sshveta rest please have a look on fixture

https://github.com/ManageIQ/integration_tests/blob/e2fa612c8acd84004e51a3743050aaacc2487441/cfme/tests/automate/custom_button/test_service_objects.py#L40

1. It's not a matter of automating BZ... I'm not doing that in this PR (Test covers those BZs)
2. It should behave some for REST or UI that's why I added PR.
3. still expected behavior not ack by PM so need Zita PR and for now, I would like its part of my testing.  If they change mind and I found it's ok Its one line changes for me to fix this test.

cfme/tests/automate/custom_button/test_service_objects.py
```diff
@@ -713,3 +716,66 @@ def test_custom_button_open_url_service_obj(objects, button_group):
         1550002
     """
     pass
+
+
+@pytest.fixture(params=["Service", "Provider"], scope="module")
+def unassigned_btn_setup(request, appliance, provider, service):
```
sshveta: Ok got it .so this service fixture is creating service through REST

PR #8829, opened 2019-05-14
[1LP][RFR] Catalog level unassigned btn test

digitronik:
Purpose or Intent
=================
-BZ 1653195

{{pytest: cfme/tests/automate/custom_button/test_catalog_custom_button.py::test_custom_button_unassigned_behavior_catalog_level -v}}

### DIFF
--- a/cfme/fixtures/service_fixtures.py
+++ b/cfme/fixtures/service_fixtures.py
@@ -202,3 +202,26 @@ def service_vm(appliance, provider, catalog_item):
         service.delete()
     if provision_request.exists:
         provision_request.remove_request(method="rest")
+
+
+@pytest.fixture(scope="module")
+def service_generic(appliance, generic_catalog_item):
+    """ This is global fixture to order generic service and return service and catalog item"""
+
+    service_catalogs = ServiceCatalogs(
+        appliance, catalog=generic_catalog_item.catalog, name=generic_catalog_item.name
+    )
+    provision_request = service_catalogs.order()
+    logger.info("Waiting for service provision request for service %s", generic_catalog_item.name)
+    provision_request.wait_for_request()
+
+    if not provision_request.is_finished():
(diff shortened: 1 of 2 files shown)


cfme/fixtures/service_fixtures.py
```diff
@@ -202,3 +202,26 @@ def service_vm(appliance, provider, catalog_item):
...
+    logger.info("Waiting for service provision request for service %s", generic_catalog_item.name)
+    provision_request.wait_for_request()
+
+    if not provision_request.is_finished():
+        pytest.skip("Failed to provision service '{}'".format(generic_catalog_item.name))
+
+    service = MyService(appliance, generic_catalog_item.dialog.label)
+    yield service, generic_catalog_item
+
+    if service.exists:
```
jaryn: I don't understand why is it necessary to check for the `service` and `provision_request` existence before the fixture deletes it.

I think as the fixture created it, it is responsible for deleting it. Not something else. So the check seems as redundant.

digitronik: @jaryn it's for checking is `service` exists after `test` or not? like some contributors do some operations like `delete` a spacific object in a test then teardown will fail in the absence of an object. So in teardown better to check object exists or not and then `delete`. The teardown is not part of the test. it's a cleanup script for test setup.

cfme/fixtures/service_fixtures.py
```diff
@@ -202,3 +202,26 @@ def service_vm(appliance, provider, catalog_item):
         service.delete()
     if provision_request.exists:
         provision_request.remove_request(method="rest")
+
+
+@pytest.fixture(scope="module")
+def service_generic(appliance, generic_catalog_item):
```
sshveta: Optional : 
Fixture name can be changed to generic_service

PR #8831, opened 2019-05-14
[1LP][RFR] Automate BZ: Add in test_case for buttons on action-edit page

john-dupuy:
This PR is meant to provider automated test coverage for the BZ 1708434.

Regarding the comment below, @jawatts and I spent some time debugging this and decided that a change to the `is_opened` property in `cfme.utils.bz` was necessary. The bug this test case was written against is on the latest downstream build. New BZs that are written against the latest build are considered to be upstream bugs, since they are also broken in the upstream build. Upstream bugs are also considered to be bugs that have no version string. However, in this case, for the downstream build, we want to consider the states of `POST` and `MODIFIED` as open for the BZ, since the bug is not fixed in the latest downstream build. This is remedied by changing `and` to `or`. 

{{ pytest: --use-template-cache cfme/tests/control/test_bugs.py::test_edit_action_buttons }}

### DIFF
--- a/cfme/tests/control/test_bugs.py
+++ b/cfme/tests/control/test_bugs.py
@@ -512,6 +512,6 @@ def test_edit_action_buttons(action_for_testing):
 
     assert view.save_button.disabled
     assert view.reset_button.disabled
-    # prep for deletion of the action during teardown
+    # click the cancel button and navigate to "Details" so that the action can be deleted
     view.cancel_button.click()
     navigate_to(action_for_testing, "Details")
--- a/cfme/tests/control/test_bugs.py
+++ b/cfme/tests/control/test_bugs.py
@@ -181,6 +181,17 @@ def setup_disk_usage_alert(appliance):
     assert not result.failed
 
 
+@pytest.fixture
+def action_for_testing(appliance):
+    action_ = appliance.collections.actions.create(
+        fauxfactory.gen_alphanumeric(),
+        action_type="Tag",
+        action_values={"tag": ("My Company Tags", "Department", "Accounting")}
+    )
+    yield action_
+    action_.delete()
+
+
 @pytest.mark.meta(blockers=[BZ(1155284)])
 def test_scope_windows_registry_stuck(request, appliance, infra_provider):
     """If you provide Scope checking windows registry, it messes CFME up. Recoverable.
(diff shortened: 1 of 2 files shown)


john-dupuy:
For some reason this BZ, 1708434, is not blocking for CFME 5.11, even with forced streams. 

```
BZ(1708434, forced_streams=["5.11"]).blocks
Out[10]: False```

This is causing the test failure in 5.11

**Comment by john-dupuy** [comment]
Oh yeah, thanks @jaryn I got side tracked by that issue and didn't explain the main reason for the PR, which is to automate a test case for a BZ!

File: cfme/tests/control/test_bugs.py [hunk 1]
```diff
@@ -481,3 +492,26 @@ def test_accordion_after_condition_creation(appliance, condition_class):
...
+        caseimportance: medium
+        initialEstimate: 1/30h
+    """
+    view = navigate_to(action_for_testing, "Edit")
+
+    assert view.save_button.disabled
+    assert view.reset_button.disabled
+    # prep for deletion of the action during teardown
+    view.cancel_button.click()
+    navigate_to(action_for_testing, "Details")
```
jawatts: **Quesiton** Why not put this in `action_for_testing`?

john-dupuy: Good suggestion, I'll do that!

cfme/tests/control/test_bugs.py
```diff
@@ -181,6 +182,20 @@ def setup_disk_usage_alert(appliance):
...
+@pytest.fixture
+def action_for_testing(appliance):
+    action_ = appliance.collections.actions.create(
+        fauxfactory.gen_alphanumeric(),
+        action_type="Tag",
+        action_values={"tag": ("My Company Tags", "Department", "Accounting")}
+    )
+    yield action_
+    view = action_.create_view(EditActionView, wait="10s")
+    view.cancel_button.click()
```
mshriver: I think this cleanup should be a bit safer - check that the view is displayed, or at least that the button is displayed before trying to click it. Its very possible when pytest returns for these statements, the view won't be present and we'll hit exceptions in teardown.

john-dupuy: After discussing this a bit offline, we've decided to move this cleanup (clicking the `cancel` button) back to the test itself, so that the fixture can be used in future tests.

PR #8832, opened 2019-05-14
[RFR] Update wt.pf, fix test_provisioning fixtures, add remote_console role

mshriver:
This exception doesn't always occur during the fixture, which obviously causes test failures.  Only occurring on the rhos provider tests, so only including it for PRT.

{{ pytest: --long-running --use-provider rhos13 cfme/tests/cloud_infra_common/test_provisioning.py }}


mshriver:
PRT failures unrelated, and missing `remote_console` role breaks many tests trying to modify roles, merging.

PR #8833, opened 2019-05-14
[1LP][RFR] EC2 storage snapshot tests

mmojzis:
Volume.status wasn't work, because elements.read was checking only first page of paginator so I added get_entity with surf_pages to target page of the entity and then elements are read.
My previous PR for storage tests made snapshot/backup tests broken by deleting storage_manager argument and also renamed size argument to volume_size - both is fixed in this PR.
Also added EC2Provider snapshot tests.

### DIFF
--- a/cfme/storage/volume.py
+++ b/cfme/storage/volume.py
@@ -402,6 +402,10 @@ class Volume(BaseEntity, CustomButtonEventsMixin, Updateable, Taggable):
             self.refresh()
         else:
             view.browser.refresh()
+        try:
+            view.entities.get_entity(name=self.name, surf_pages=True)
+        except ItemNotFound:
+            return False
 
         for item in view.entities.elements.read():
             if self.name in item['Name']:
--- a/cfme/storage/volume.py
+++ b/cfme/storage/volume.py
@@ -322,10 +322,16 @@ class Volume(BaseEntity, CustomButtonEventsMixin, Updateable, Taggable):
                  timeout=1000,
                  fail_func=self.refresh)
 
-    def create_snapshot(self, name, cancel=False, reset=False):
+    def create_snapshot(self, name, cancel=False, reset=False, from_manager=False):
         """create snapshot of cloud volume"""
         snapshot_collection = self.appliance.collections.volume_snapshots
-        view = navigate_to(self, 'Snapshot')
+        if from_manager:
+            view = navigate_to(self.parent.manager, 'Volumes')
+            view.entities.get_entity(surf_pages=True, name=self.name).check()
+            view.toolbar.configuration.item_select('Create a Snapshot of selected Cloud Volume')
+            view = view.browser.create_view(VolumeSnapshotView, additional_context={'object': self})
+        else:
+            view = navigate_to(self, 'Snapshot')
 
         changed = view.snapshot_name.fill(name)
 
@@ -403,13 +409,10 @@ class Volume(BaseEntity, CustomButtonEventsMixin, Updateable, Taggable):
         else:
             view.browser.refresh()
         try:
-            view.entities.get_entity(name=self.name, surf_pages=True)
+            ent = view.entities.get_entity(name=self.name, surf_pages=True)
+            return ent.data["status"]
         except ItemNotFound:
             return False
-
-        for item in view.entities.elements.read():
-            if self.name in item['Name']:
-                return str(item['Status'])
 
     @property
     def size(self):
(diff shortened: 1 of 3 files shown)


cfme/storage/volume.py
```diff
@@ -402,6 +408,10 @@ def status(self):
             self.refresh()
         else:
             view.browser.refresh()
+        try:
+            view.entities.get_entity(name=self.name, surf_pages=True)
+        except ItemNotFound:
+            return False
 
         for item in view.entities.elements.read():
```
digitronik: As you are using `get_entity`  now no need of this iteration over the table.
```
try:
    ent = view.entities.get_entity(name=self.name, surf_pages=True)
    return ent.data["status"]
except ItemNotFound:
    return False
```

cfme/storage/volume_snapshot.py
```diff
@@ -152,6 +152,10 @@ def status(self):
         """
         view = navigate_to(self.parent, 'All')
         view.toolbar.view_selector.select("List View")
+        try:
+            view.entities.get_entity(name=self.name, surf_pages=True)
+        except ItemNotFound:
+            return False
 
         for item in view.entities.elements.read():
```
digitronik: same as above

cfme/storage/volume.py
mmojzis: Nice, I tried to find where that entity has stored elements data...But haven't checked .data.
Thanks and reworked.

dajoRH:
*CFME QE Bot*

mmojzis:
@mshriver Please can you explain what does that travis fail mean?

mshriver:
```
   def test_match_files(pyver):
        frozen = reqs(HERE.joinpath("frozen.{pyver}.txt".format(pyver=pyver)))
        frozen_doc = reqs(HERE.joinpath("frozen_docs.{pyver}.txt".format(pyver=pyver)))
        expected_frozen_doc = {k: v for k, v in frozen.items() if k in frozen_doc}
>       assert expected_frozen_doc == frozen_doc
E       AssertionError: assert {'Babel': '2....'2.1.11', ...} == {'Babel': '2.6...'2.1.11', ...}
E         Omitting 329 identical items, use -vv to show
E         Differing items:
E         {'wrapanapi': '3.1.8'} != {'wrapanapi': '3.1.6'}
```

Wrapanapi is in the docs requirements files as well as the regular requirements files. We test that the versions of dependent packages are equal in docs vs regular.  Its telling you that there's a difference between the two files.

I don't think your updates to requirements files are necessary, you should rebase.

mmojzis:
@mshriver Okay, it passed for 5.10 now.

cfme/storage/volume.py
```diff
@@ -402,10 +408,11 @@ def status(self):
             self.refresh()
         else:
             view.browser.refresh()
-
-        for item in view.entities.elements.read():
-            if self.name in item['Name']:
-                return str(item['Status'])
+        try:
+            ent = view.entities.get_entity(name=self.name, surf_pages=True)
+            return ent.data["status"]
```
jawatts: **Question** Will the key 'status' always be there, perhaps we should catch `KeyError` as well?

cfme/tests/storage/test_volume_snapshot.py
```diff
@@ -24,16 +26,34 @@
 STORAGE_SIZE = 1
 
 
+@pytest.fixture(params=["from_manager", "from_volume"])
+def from_manager(request):
```
jawatts: **Question** Why not paramterize the test on this? A fixture should be used test setup

something like:

```
@pytest.mark.parametrize('snapshot_create_from', [True, False], ids=['from_manager', 'from_volume'])
def test_storage_snapshot_create_cancelled_validation(volume, from_manager):
```
I see you are trying to reduce code duplication, awesome. But for readability, I would prefer it to be marked for each test

cfme/storage/volume.py
mmojzis: Status should always be there as it's column in that table I am trying to access.

jawatts:
@mmojzis Looks good to me but could you look into the 5.11 failure.

mmojzis:
> @mmojzis Looks good to me but could you look into the 5.11 failure.
@jawatts 
None of my tests are working in 5.11. AFAIK there is different structure so it's failing.
I can focus on 5.11 fixing after all my storage tests are automated.

PR #8835, opened 2019-05-14
[1LP][RFR] Remove appliance dependencies from cfme.utils.bz and move them to cfme.utils.blockers

john-dupuy:
I recently faced some confusion in dealing with the BZ objects. A lot of it stemmed from the fact that `BugWrapper` in `cfme.utils.bz` gives information relative to the appliance that you're using for testing. My thought is that `cfme.utils.bz` should give information about the BZ _independent_ of the appliance on which you're testing. This appliance specific stuff should be evaluated in the `BZ()` blocker class in `cfme.utils.blockers`.

### DIFF
--- a/cfme/utils/blockers.py
+++ b/cfme/utils/blockers.py
@@ -182,6 +182,31 @@ class BZ(Blocker):
         return self.data
 
     @property
+    def can_test_on_current_upstream_appliance(self):
+        bug = self.data
+        if not bug.can_test_on_upstream:
+            return False
+        if version.appliance_is_downstream():
+            return False
+
+        change_states = {"POST", "MODIFIED"}
+        # With these states, the change is in upstream
+        if bug.status not in {"POST", "MODIFIED", "ON_QA", "VERIFIED", "RELEASE_PENDING"}:
+            return False
+        history = bug.get_history_raw()["bugs"][0]["history"]
+        changes = []
+        # We look for status changes in the history
+        for event in history:
+            for change in event["changes"]:
+                if change["field_name"].lower() != "status":
+                    continue
+                if change["added"] in change_states:
+                    changes.append(event["when"])
+                    return event["when"] < version.appliance_build_datetime()
+        else:
+            return False
+
+    @property
     def blocks(self):
         try:
             bug = self.data
@@ -191,7 +216,8 @@ class BZ(Blocker):
             if bug.is_opened:
                 result = True
             if bug.upstream_bug:
-                if not version.appliance_is_downstream() and bug.can_test_on_upstream:
+                if (not version.appliance_is_downstream() and
+                        self.can_test_on_current_upstream_appliance):
                     result = False
             if not result and version.appliance_is_downstream():
                 if bug.fixed_in is not None:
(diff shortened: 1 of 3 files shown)


dajoRH:
*CFME QE Bot*

mshriver:
@jawatts hit some issues with local testing, moving to WIPTEST

cfme/utils/blockers.py
```diff
@@ -181,6 +181,33 @@ def bugzilla_bug(self):
             return None
         return self.data
 
+    @property
+    def can_test_on_current_upstream_appliance(self):
```
john-dupuy: @mshriver and I discussed a name change for this function and also making it free from references to the appliance, with the appliance specific stuff coming in the `blocks` function.

john-dupuy: On second glance, however, the property `can_test_on_upstream` already provides a function free from references to the appliance. This property is designed for use within the `blocks` function, and therefore, needs the reference to the appliance. 

It is answering the question: "if I have a bug, and that bug is in a "POST" or "MODIFIED" state, and if I have an upstream appliance, is the build date of my current upstream appliance after the fix for the bug was merged?" If so, then it will return `True`, because the bug's fix is included in your current upstream appliance. If you happen to be on an older version of an upstream appliance, then it will return false.

john-dupuy: Adding a docblock for this function, so that's its intent is more clear

cfme/utils/blockers.py
```diff
@@ -181,6 +181,44 @@ def bugzilla_bug(self):
...
+        "If I have a bug, and that bug is in a "POST" or "MODIFIED" state, and if I have an
+        upstream appliance, is the build date of my current upstream appliance after
+        the fix for the bug was merged?"
+
+        If so, then it will return True, because the bug's fix is included in your current
+        upstream appliance.
+
+        If you happen to be on an older version of an upstream appliance that doesn't have the fix,
+        then it will return False.
+        """
```
mshriver: :tada: :tada:

cfme/utils/blockers.py
```diff
@@ -181,6 +181,44 @@ def bugzilla_bug(self):
...
+        history = bug.get_history_raw()["bugs"][0]["history"]
+        changes = []
+        # We look for status changes in the history
+        for event in history:
+            for change in event["changes"]:
+                if change["field_name"].lower() != "status":
+                    continue
+                if change["added"] in change_states:
+                    changes.append(event["when"])
+                    return event["when"] < version.appliance_build_datetime()
```
mshriver: I'm hitting a `TypeError` here because `version.appliance_build_datetime()` is returning `None`.

Creating an appliance object using `Appliance.from_provider`, with a MIQ nightly build, results in an appliance object with Version('master'), which has no build datetime.

john-dupuy: Moving to WIPTEST until this is resolved

john-dupuy: This should be fixed by a small change to the `build` property of an appliance object

PR #8828, opened 2019-05-14
[RFR] Move manual webmks console testcases

apagac:
__Moving__ manual testcases from `test_manual.py` to respective file.

I choose different approach this time. When testing consoles, there is considerable amount of combinations of operating systems, browsers and providers, all being tested manually at this time. Originally we had a testcase in polarion for every combination. Instead of copying all that from `test_manual.py`, I choose to describe the testing matrix in the comment of just one testcase.

Also, minor changes in testcase naming were performed.

Please let me know what you think.

PR #8840, opened 2019-05-15
[1LP][RFR] Move manual vmrc console testcases

apagac:
__Moving__ manual testcases `from test_manual.py` to newly created `test_vmrc_console.py`

Very similar to PR #8828. 

Made one testcase with testing matrix, then there are some BZ testcases. Minor naming changes.


cfme/tests/infrastructure/test_vmrc_console.py
```diff
@@ -0,0 +1,195 @@
+# -*- coding: utf-8 -*-
+import pytest
+
+from cfme import test_requirements
+
+
+@test_requirements.vmrc
+@pytest.mark.manual
```
lina-nikiforova: Suggestion: you can put this in
```
pytestmark = [
    pytest.mark.manual
]
```
for the whole file

PR #8843, opened 2019-05-16
[WIPTEST] Bump wait_for to display less trash in the traceback.

jaryn:
The newer wait_for doesn't display its own traceback frames which saves space on screen.

{{ py.test: -sv --use-provider 'complete' -m smoke }}

### DIFF
--- a/requirements/frozen.py2.txt
+++ b/requirements/frozen.py2.txt
@@ -320,7 +320,7 @@ uritemplate==3.0.0
 urllib3==1.24.1
 virtualenv==16.4.3
 vspk==5.3.2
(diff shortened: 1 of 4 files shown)


requirements/frozen.py2.txt
```diff
@@ -320,7 +320,7 @@ uritemplate==3.0.0
 urllib3==1.24.1
 virtualenv==16.4.3
 vspk==5.3.2
-wait-for==1.0.11
+wait-for==1.0.12
```
mshriver: Need to update the version in the docs frozen files as well.

Thanks for the enhancement!!

PR #8846, opened 2019-05-16
fix miq run test exit code

izapolsk:


cfme/scripting/runtest.py
```diff
@@ -12,4 +12,4 @@ def main():
         os.execl(sys.executable, sys.executable, *sys.argv)
     else:
         import pytest
-        pytest.main()
+        return pytest.main()
```
mshriver: :man_facepalming:

PR #8848, opened 2019-05-17
[1LP][RFR] Fixed path of quota related product feature tree

ganeshhubale:
{{ pytest: cfme/tests/cloud/test_tenant.py::test_dynamic_product_feature_for_tenant_quota -vv }}

### DIFF
--- a/cfme/base/login.py
+++ b/cfme/base/login.py
@@ -103,7 +103,6 @@ class BaseLoggedInPage(View):
 
     def logout(self):
         self.settings.select_item('Logout')
-        self.browser.handle_alert(wait=None)
         self.extra.appliance.user = None
 
     @property
--- a/cfme/base/login.py
+++ b/cfme/base/login.py
@@ -103,6 +103,7 @@ class BaseLoggedInPage(View):
 
     def logout(self):
         self.settings.select_item('Logout')
+        self.browser.handle_alert(wait=None)
         self.extra.appliance.user = None
 
     @property
--- a/cfme/tests/automate/test_domain.py
+++ b/cfme/tests/automate/test_domain.py
@@ -9,6 +9,7 @@ from cfme.automate.explorer.klass import ClassCopyView
 from cfme.automate.explorer.method import MethodCopyView
 from cfme.exceptions import OptionNotAvailable
 from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.blockers import BZ
 from cfme.utils.update import update
 
 pytestmark = [test_requirements.automate]
@@ -247,6 +248,7 @@ def test_domain_lock_unlock(domain, appliance):
 
 
 @pytest.mark.tier(1)
+@pytest.mark.meta(blockers=[BZ(1686762)])
 def test_object_attribute_type_in_automate_schedule(appliance):
     """
     Polarion:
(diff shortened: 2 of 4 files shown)


cfme/tests/automate/test_domain.py
```diff
@@ -247,6 +248,7 @@ def test_domain_lock_unlock(domain, appliance):
 
 
 @pytest.mark.tier(1)
+@pytest.mark.meta(blockers=[BZ(1686762, forced_streams=['5.10'])])
```
digitronik: I don't think you need `force_streams`.
```
In [2]: BZ(1686762).blocks
Out[2]: True
In [3]: appliance.version                                                                                                                         
Out[3]: Version('5.10.5.0')
```

cfme/tests/automate/test_git_import.py
```diff
@@ -7,11 +7,13 @@
 from cfme.automate.import_export import AutomateGitRepository
 from cfme.base.credential import Credential
 from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.blockers import BZ
 
 
 pytestmark = [
     test_requirements.automate,
     pytest.mark.meta(server_roles="+git_owner"),
+    pytest.mark.meta(blockers=[BZ(1713643, forced_streams=['5.11'])])
```
digitronik: Please check this as well

cfme/tests/cloud/test_tenant.py
```diff
@@ -104,9 +90,23 @@ def check_permissions(appliance, assigned_tenant):
...
 
 
+@pytest.fixture(scope="module")
+def product_features(appliance):
+    """Tree path navigation (for quota management permissions) to particular node of product feature
+     in RBAC for roles
+     """
+    config = "Settings,Configuration" if appliance.version < "5.11" else "Main Configuration"
+    product_feature = [
+        "Everything,{config},Access Control,Tenants,Modify,Manage Quotas,Manage Quotas ({tenant})".
```
digitronik: optional:
nice hack :) but I will try to avoid this for readability.
totally optional

cfme/base/login.py
```diff
@@ -103,7 +103,6 @@ def logged_out(self):
 
     def logout(self):
         self.settings.select_item('Logout')
-        self.browser.handle_alert(wait=None)
```
mshriver: This needs to be here, I understand its causing the below exception, but removing this call is not the correct fix. Please see #8877  for the fix for this.

```
>           self.logger.info('handling alert: %r', popup.text)
E           AttributeError: 'NoneType' object has no attribute 'text'

cfme/utils/appliance/implementations/common.py:80: AttributeError
AttributeError
'NoneType' object has no attribute 'text'
```

cfme/tests/cloud/test_tenant.py
mshriver: Yeah I'm not much of a fan of this either, using a string and then splitting it.  Consider just building the list directly.

```
product_feature = ['Everything']
product_feature.extend(['Settings', 'Configuration'] if appliance.version < '5.11' else ['Main Configuration])
product_feature.extend(['Access', 'others', Manage Quotas ({tenant})' for tenant in TENANTS])
```

mshriver: Also, this isn't that useful as a fixture, since there is no setup or teardown. 

This could be a regular method if used in more than one test, or just written into the test directly.

mshriver: Additionally it looks like `Settings` is now `User Settings` as well, at least in 5.10.5.

ganeshhubale: Added in test itself.

PR #8851, opened 2019-05-17
[1LP][RFR] New Test: Check whether service is getting retired by automate method

ganeshhubale:
- This test case creates one catalog item and orders it and tries to retire that service provisioned via automate method: ```create_retire_request```

{{ pytest: cfme/tests/automate/test_common_methods.py::test_service_retirement_from_automate_method -vv }}

### DIFF
--- a/cfme/tests/automate/test_common_methods.py
+++ b/cfme/tests/automate/test_common_methods.py
@@ -2,6 +2,7 @@
 """This module contains tests that test the universally applicable canned methods in Automate."""
 from datetime import date
 from datetime import timedelta
+from textwrap import dedent
 
 import fauxfactory
 import pytest
@@ -182,21 +183,10 @@ def test_miq_password_decrypt(klass):
     assert result.validate()
 
 
-@pytest.fixture(scope='module')
-def original_class(domain):
-    # Take the 'Request' class and copy it for own purpose.
-    domain.parent.instantiate(name="ManageIQ").namespaces.instantiate(
-        name="System"
-    ).classes.instantiate(name="Request").copy_to(domain.name)
-    klass = domain.namespaces.instantiate(name="System").classes.instantiate(name="Request")
-    yield klass
-    klass.delete_if_exists()
-
-
 @pytest.mark.tier(1)
 @pytest.mark.meta(automates=[1700524])
 @pytest.mark.ignore_stream("5.10")
-def test_service_retirement_from_automate_method(request, generic_catalog_item, original_class):
+def test_service_retirement_from_automate_method(request, generic_catalog_item, custom_instance):
     """
     Bugzilla:
         1700524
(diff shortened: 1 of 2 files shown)


dajoRH:
*CFME QE Bot*

dajoRH:
I detected some fixture changes in commit 68d25715c04f001af701b6382b442f5e81986dfe

The *local* fixture **`original_class`** is used in the following files:
- **cfme/tests/automate/test_common_methods.py**
    - *test_service_retirement_from_automate_method*

cfme/tests/automate/test_common_methods.py
```diff
@@ -180,3 +180,89 @@ def test_miq_password_decrypt(klass):
...
+    Bugzilla:
+        1700524
+
+    Polarion:
+        assignee: ghubale
+        initialEstimate: 1/8h
+        caseposneg: positive
+        startsin: 5.11
+        casecomponent: Automate
+        testSteps:
```
john-dupuy: Since you are including `testSteps` may as well include expected results too.

cfme/tests/automate/test_common_methods.py
```diff
@@ -180,3 +180,89 @@ def test_miq_password_decrypt(klass):
...
+        "request = $evm.execute(:create_retire_request, service)\n"
+        "$evm.log(:info, 'Create request for create_retire_request #{request}')"
+    )
+
+    # Adding method - send_email for sending mails
+    method = original_class.methods.create(
+        name=fauxfactory.gen_alphanumeric(),
+        display_name=fauxfactory.gen_alphanumeric(),
+        location='inline',
+        script=script)
```
john-dupuy: _Optional_ formatting,
```python,
    method = original_class.methods.create(
        name=fauxfactory.gen_alphanumeric(),
        display_name=fauxfactory.gen_alphanumeric(),
        location='inline',
        script=script
    )
```

cfme/tests/automate/test_common_methods.py
```diff
@@ -180,3 +180,89 @@ def test_miq_password_decrypt(klass):
...
+
+    # Adding instance to call automate method - send_email
+    instance = original_class.instances.create(
```
john-dupuy: Any cleanup necessary for this instance?

cfme/tests/automate/test_common_methods.py
```diff
@@ -180,3 +180,89 @@ def test_miq_password_decrypt(klass):
...
+
+
+@pytest.fixture(scope='module')
+def original_class(domain):
+    # Take the 'Request' class and copy it for own purpose.
+    domain.parent.instantiate(name="ManageIQ").namespaces.instantiate(
+        name="System"
+    ).classes.instantiate(name="Request").copy_to(domain.name)
+    klass = domain.namespaces.instantiate(name="System").classes.instantiate(name="Request")
+    return klass
```
john-dupuy: I would `yield` the `klass` since it's in a fixture

cfme/tests/automate/test_common_methods.py
ganeshhubale: - Actually there is no cleanup required for instance as it is child of klass/namespace/domain. Instance/method is automatically gets deleted once we delete klass/namespace/domain.
- It will consume more time as it will delete method - instance - klass- namespace - domain
- But I am getting these question again angain hence adding cleanup.

john-dupuy: Apologies for the repeated question! Maybe to avoid this question in the future you could put an inline comment explaining why no cleanup is necessary? It would certainly help jog my memory :smile:

ganeshhubale: - I am not saying repeated comments from you only but it is obvious that all are always asking about that as I am creating something there(like method/instance)
- But thanks for inline comment suggestion, I will start adding in coming PRs

cfme/tests/automate/test_common_methods.py
```diff
@@ -180,3 +180,99 @@ def test_miq_password_decrypt(klass):
         execute_methods=True,
     )
     assert result.validate()
+
+
+@pytest.fixture(scope='module')
+def original_class(domain):
```
digitronik: @ganeshhubale I think you can use this global fixture in place of this
https://github.com/ManageIQ/integration_tests/blob/c6913cc5bf0ff2f0c8186c5a84be0f9ea4a8e620/cfme/fixtures/automate.py#L40

cfme/tests/automate/test_common_methods.py
```diff
@@ -180,3 +180,99 @@ def test_miq_password_decrypt(klass):
...
+    service_request = generic_catalog_item.appliance.rest_api.collections.service_templates.get(
+        name=generic_catalog_item.name
+    ).action.order()
+    request.addfinalizer(lambda: service_request.action.delete())
+    wait_for(lambda: service_request.request_state == "finished", fail_func=service_request.reload,
+             timeout=180, delay=10)
+
+    # Ruby code to execute create_retire_request
+    script = (
+        "service = $evm.root['service']\n"
```
digitronik: optional:
but in such cases; I would suggest to use `textwrap` `dedent`. It will help to manage script in proper way.
you can refer

https://github.com/ManageIQ/integration_tests/blob/c6913cc5bf0ff2f0c8186c5a84be0f9ea4a8e620/cfme/tests/automate/custom_button/test_service_vm_custom_button.py#L46

cfme/tests/automate/test_common_methods.py
```diff
@@ -180,3 +180,99 @@ def test_miq_password_decrypt(klass):
...
+    # Adding automate method
+    method = original_class.methods.create(
+        name=fauxfactory.gen_alphanumeric(),
+        display_name=fauxfactory.gen_alphanumeric(),
+        location='inline',
+        script=script
+    )
+    request.addfinalizer(method.delete_if_exists)
+
+    # Adding instance to call automate method
```
digitronik: I think, we can again use direct global fixture... which will create all above thing directly by just passing `ruby code` to it. 
This is most common usecase in automation
https://github.com/ManageIQ/integration_tests/blob/c6913cc5bf0ff2f0c8186c5a84be0f9ea4a8e620/cfme/fixtures/automate.py#L52

PR #8853, opened 2019-05-17
[RFR] Update template_upload to block providers on streams

mshriver:
Specify for each provider type what streams should not be uploaded.

Might be better to have this in cfme_data.yaml?

We want to block upload of certain streams to certain provider types based on image availability.


john-dupuy:
@mshriver I think this should be included as a field in `cfme_data.yaml`, that way everything will be in a centralized location. Code to block the provider LGTM :+1:

PR #8855, opened 2019-05-20
[1LP][WIPTEST]Adding requirement marker for test feature

mnadeem92:

### DIFF
--- a/cfme/test_requirements.py
+++ b/cfme/test_requirements.py
@@ -72,6 +72,12 @@ alert = pytest.mark.requirement(
     assignee_id='jdupuy',
 )
 
+app_console = pytest.mark.requirement(
+    "Appliance Console",
+    description='Appliance console operations support',
+    assignee_id='mnadeem'
+)
+
 auth = pytest.mark.requirement(
     "Authentication",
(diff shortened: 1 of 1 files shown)


cfme/test_requirements.py
```diff
@@ -402,3 +402,9 @@ def test_quota_alert():
     description='Support of VMWare in MIQ/CFME',
     assignee_id='kkulkarn'
 )
+
+black_console = pytest.mark.requirement(
```
mshriver: I think it would be better to give this a short name of `appliance_console`, or just  `cli`, instead of `black_console`.  `black_console` seems a bit ambiguous to me (and its just the default terminal emulator background color).

Additionally, please don't just add it to the end of the file - follow the existing alphabetized ordering that's already here and insert whatever name is used at the right spot.

@jawatts @jkrocil Any thoughts on the FA naming here?

mnadeem92: cli seems very generic word, how is sounds app_console ?

jawatts: @mshriver I like `appliance_console` here.

mshriver: Either `app_console` or `appliance_console` I'd approve/merge.

PR #8856, opened 2019-05-20
[1LP][RFR] Fix toolbar download button for details page

valaparthvi:
Changes:
1. Since 5.10, the title of the download button in the toolbar has been changed from `Download summary in PDF format` to `Print or export Summary`, change that for automation.

I didn't find any tests using this download feature, so not testing the changes, but I have tested it locally and it works fine.

### DIFF
--- a/cfme/ansible/playbooks.py
+++ b/cfme/ansible/playbooks.py
@@ -53,7 +53,7 @@ class PlaybookDetailsView(PlaybookBaseView):
     class toolbar(View):   # noqa
         configuration = Dropdown("Configuration")
         policy = Dropdown(text='Policy')
-        download_button = Button(title="Print or export summary")
+        download = Button(title="Print or export summary")
 
     breadcrumb = BreadCrumb(locator='.//ol[@class="breadcrumb"]')
     entities = View.nested(PlaybookDetailsEntities)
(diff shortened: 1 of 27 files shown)


cfme/ansible/playbooks.py
```diff
@@ -53,7 +53,7 @@ class PlaybookDetailsView(PlaybookBaseView):
     class toolbar(View):   # noqa
         configuration = Dropdown("Configuration")
         policy = Dropdown(text='Policy')
-        download_button = Button(title="Download summary in PDF format")
+        download_button = Button(title="Print or export summary")
```
john-dupuy: Since every other view calls this `download` can you change this to `download` as well?

valaparthvi: There are still a few(very few) views that use `download_button` instead of `download`, I think it's better if I leave that to the FA owner, but let me know if you think otherwise, changing the name in the remaining views won't be that difficult.

john-dupuy: In my opinion, since this button is unused in tests, it shouldn't harm anything. And it is better to maintain consistency in the codebase. What do you think? 

Maybe ping the FA owners here so they are aware of the change.

valaparthvi: There you go @john-dupuy, replaced all the `download_button` to `download`!

PR #8857, opened 2019-05-20
[RFR][NOTEST]Remove unnecessary tests

valaparthvi:


valaparthvi:
So, 2 of these test cases are already covered by automation, one of them tests a very rare customer specific BZ that I don't think needs testing anymore, and the other one, I had added while adding qe-test-coverage, but I don't think it needs a coverage.

PR #8543, opened 2019-03-06
[1LP][RFR] Tests tagging on provider side

mshriver:
Rebased from Ruslana Babyuk's original commits in #7320, updated for more recent test patterns.

I would like to leave any remaining test failures to the FA owner to address.

{{ pytest: --long-running -vvvv --use-provider complete cfme/tests/cloud/test_tag_mapping.py cfme/tests/configure/test_tag.py }}

PR #8858, opened 2019-05-20
[RFR] follow-up for #8543

lina-nikiforova:
As promised in #8543 I'm fixing the errors in tagging tests

{{ pytest: --long-running -vvvv --use-provider complete cfme/tests/cloud/test_tag_mapping.py  }}

I don't know how `def get_tags` worked before, but I saw sometime ago also empty name in Tag property, so I believe this fix going to fix more tests.

https://github.com/ManageIQ/wrapanapi/pull/382 This is also needed for this fix.


lina-nikiforova:
all failed tests are failed on `provider.refresh_provider_relationships(method='ui')` so it's not related to the changes in the PR and will be handled separately

PR #8865, opened 2019-05-24
[1LP][RFR] Adding cleanup script for old snapshot

mnadeem92:
This PR will delete all instance snapshot with name started
with "test_snapshot_" and created by >= 2 hours before.

This PR is dependent on : https://github.com/ManageIQ/wrapanapi/pull/385

### DIFF
--- a/scripts/cleanup_openstack_instance_snapshot.py
+++ b/scripts/cleanup_openstack_instance_snapshot.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python2
+# -*- coding: utf-8 -*-
+"""Cleanup unused instance snapshot from glance repository
+
+Usage: scripts/cleanup_openstack_instance_snapshot.py [optional list of provider keys]
+
+If no providers specified, it will cleanup all of them.
+
+"""
+import sys
+from datetime import datetime
+from datetime import timedelta
+
+import iso8601
+import tzlocal
+
+from cfme.utils.providers import get_mgmt
+from cfme.utils.providers import list_provider_keys
+
+local_tz = tzlocal.get_localzone()
+GRACE_TIME = timedelta(hours=2)
+time_limit = datetime.now(tz=local_tz) - GRACE_TIME
+
+
+def main(*providers):
+    for provider_key in providers:
+        print("Cleaning up {}".format(provider_key))
+        api = get_mgmt(provider_key).gapi
+        try:
+            images = api.images.list()
+        except Exception as e:
+            print("Connect to provider failed:{} {} {}".format(
+                provider_key, type(e).__name__, str(e)))
+            continue
+
+        for img in images:
+            if "test_snapshot_" in img.name \
+                    and iso8601.parse_date(img.created_at) < time_limit:
+                print("Deleting {}".format(img.name))
+                try:
+                    api.images.delete(img.id)
+                except Exception as e:
+                    print("Delete failed: {} {}".format(type(e).__name__, str(e)))
+
+
+if __name__ == "__main__":
+    provs = sys.argv[1:]
+    if provs:
+        main(*provs)
+    else:
+        main(*list_provider_keys("openstack"))
--- a/scripts/cleanup_openstack_instance_snapshot.py
+++ b/scripts/cleanup_openstack_instance_snapshot.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
 # -*- coding: utf-8 -*-
 """Cleanup unused instance snapshot from glance repository
 
@@ -7,45 +7,61 @@ Usage: scripts/cleanup_openstack_instance_snapshot.py [optional list of provider
 If no providers specified, it will cleanup all of them.
 
 """
-import sys
+import argparse
 from datetime import datetime
 from datetime import timedelta
 
-import iso8601
 import tzlocal
 
+from cfme.utils.log import add_stdout_handler
+from cfme.utils.log import logger
 from cfme.utils.providers import get_mgmt
 from cfme.utils.providers import list_provider_keys
 
-local_tz = tzlocal.get_localzone()
+LOCAL_TZ = tzlocal.get_localzone()
 GRACE_TIME = timedelta(hours=2)
-time_limit = datetime.now(tz=local_tz) - GRACE_TIME
+TIME_LIMIT = datetime.now(tz=LOCAL_TZ) - GRACE_TIME
+
+# log to stdout too
+add_stdout_handler(logger)
 
 
-def main(*providers):
-    for provider_key in providers:
-        print("Cleaning up {}".format(provider_key))
-        api = get_mgmt(provider_key).gapi
+def parse_cmd_line():
+    parser = argparse.ArgumentParser(argument_default=None)
+    parser.add_argument('--name', default='test_snapshot_',
+                        help='Starting pettern of snaphsot name '
+                             'e.g. --name test_ delete all snapshot starting with test_')
+    parser.add_argument('--providers', default=list_provider_keys("openstack"), nargs='+',
+                        help='List of provider keys e.g. --providers rhos13 rhos12'
+                        )
+    args = parser.parse_args()
+    return args
+
+
+def main(args):
+    """ Cleanup all snapshots name starting with test_snapshot and created by >= 2 hours before
+
+    :param providers: Lists provider keys
+    :return:
+    """
+    for provider_key in args.providers:
+        logger.info("Cleaning up {}".format(provider_key))
+        provider_obj = get_mgmt(provider_key)
+
         try:
-            images = api.images.list()
-        except Exception as e:
-            print("Connect to provider failed:{} {} {}".format(
-                provider_key, type(e).__name__, str(e)))
+            images = provider_obj.list_templates()
+        except Exception:
+            logger.exception("Unable to get the list of templates")
             continue
 
         for img in images:
-            if "test_snapshot_" in img.name \
-                    and iso8601.parse_date(img.created_at) < time_limit:
-                print("Deleting {}".format(img.name))
+            if img.name.startswith(args.name) and img.creation_time < TIME_LIMIT:
+                logger.info("Deleting {}".format(img.name))
                 try:
-                    api.images.delete(img.id)
-                except Exception as e:
-                    print("Delete failed: {} {}".format(type(e).__name__, str(e)))
+                    img.delete()
+                except Exception:
+                    logger.exception("Snapshot {} Deletion failed".format(img.name))
 
 
 if __name__ == "__main__":
-    provs = sys.argv[1:]
-    if provs:
-        main(*provs)
-    else:
-        main(*list_provider_keys("openstack"))
+    main(parse_cmd_line())
--- a/cfme/tests/cloud_infra_common/test_snapshots_rest.py
+++ b/cfme/tests/cloud_infra_common/test_snapshots_rest.py
@@ -73,7 +73,7 @@ def vm_snapshot(request, appliance, collection, vm):
     snap_desc = 'snapshot {}'.format(uid)
     request.addfinalizer(lambda: _delete_snapshot(vm, snap_desc))
     vm.snapshots.action.create(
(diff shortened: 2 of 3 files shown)


dajoRH:
Lint report for commit 6b81568e8d56a305b89b97601e57c4a5304d1647:

`scripts/cleanup_openstack_instance_snapshot.py`:
- :bangbang: Line 36:101: **E501** *line too long (126 > 100 characters)*

**No commit flaws detected**.

*CFME QE Bot*

scripts/cleanup_openstack_instance_snapshot.py
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env python2
```
kedark3: we may want to get it on py3 since we are moving everything else to py3.

scripts/cleanup_openstack_instance_snapshot.py
```diff
@@ -0,0 +1,50 @@
...
+# -*- coding: utf-8 -*-
+"""Cleanup unused instance snapshot from glance repository
+
+Usage: scripts/cleanup_openstack_instance_snapshot.py [optional list of provider keys]
+
+If no providers specified, it will cleanup all of them.
+
+"""
+import sys
+from datetime import datetime
```
kedark3: Is it okay to put it as 
`from datetime import datetime, timedelta` ?

scripts/cleanup_openstack_instance_snapshot.py
```diff
@@ -0,0 +1,50 @@
...
+from datetime import timedelta
+
+import iso8601
+import tzlocal
```
kedark3: If I am not wrong you need L14,L15 imports to be before those that are `from...`

scripts/cleanup_openstack_instance_snapshot.py
```diff
@@ -0,0 +1,51 @@
...
+        print("Cleaning up {}".format(provider_key))
+        api = get_mgmt(provider_key).gapi
+        try:
+            images = api.images.list()
+        except Exception as e:
+            print("Connect to provider failed:{} {} {}".format(
+                provider_key, type(e).__name__, str(e)))
+            continue
+
+        for img in images:
```
kedark3: Suggestion: I'd write this as `for image in images:`

scripts/cleanup_openstack_instance_snapshot.py
mnadeem92: Imports should usually be on separate lines, but if we import 2 sub-module from same parent modules then we can right it on single line, but again i preferred to make it on separate line.
https://www.python.org/dev/peps/pep-0008/#imports

scripts/cleanup_openstack_instance_snapshot.py
mnadeem92: No, we cant move it  as the datetime is in built pyton library, so it should come first where as iso8601 and tzlocal are 3rd-party library, so it should come after that.

scripts/cleanup_openstack_instance_snapshot.py
mnadeem92: I guess it is trivial and should be of coder choice.

scripts/cleanup_openstack_instance_snapshot.py
mnadeem92: right now, our all other scripts are based on python2, i would suggest , we can migrate all at once when we are ready for py3.

scripts/cleanup_openstack_instance_snapshot.py
kedark3: Got it.

scripts/cleanup_openstack_instance_snapshot.py
kedark3: I think I will still vote on making it py3 now than to migrate later. Let's see what the other reviewers have to say.

apagac:
Two notes from me:

1. You change the snapshot naming in `cloud_infra_common/test_snapshots_rest.py` but there are snapshot getting created also in `tests/infrastructure/test_snapshot.py`. If you want to be consistent, it would be better to change the naming also there.
Note that in fact it doesn't really matter, because snapshot cleanup works fine on all providers except RHOS.
2. I see you are using api. I think that right now there is no way of getting a list of images from RHOS provider using CFME api, see bz 1712850.

mnadeem92:
@apagac Thank you for the input
1. I have changed the snapshot name to comply with our naming convention as per below reference, such that it will clean those un-used snapshot as well who gets created by some one during their manual testing.
>> https://mojo.redhat.com/docs/DOC-1077222
Thank you for pointing out, i will change the name from tests/infrastructure/test_snapshot.py as well.

2. Agree with you there is no way to  list the snapshot details by using instance object in CFME, which is why the cleanup is not working in case of OSP provider, And for the same reason i have used here Glance API client to filter out the snapshot list.

scripts/cleanup_openstack_instance_snapshot.py
```diff
@@ -0,0 +1,51 @@
...
+GRACE_TIME = timedelta(hours=2)
+time_limit = datetime.now(tz=local_tz) - GRACE_TIME
+
+
+def main(*providers):
+    for provider_key in providers:
+        print("Cleaning up {}".format(provider_key))
+        api = get_mgmt(provider_key).gapi
+        try:
+            images = api.images.list()
```
digitronik: I'm suspect due to `gapi`; is this implementation for `openstack` only?

If yes why we are not using `list_templates()`?

scripts/cleanup_openstack_instance_snapshot.py
```diff
@@ -0,0 +1,51 @@
...
+        except Exception as e:
+            print("Connect to provider failed:{} {} {}".format(
```
digitronik: I would log it

scripts/cleanup_openstack_instance_snapshot.py
```diff
@@ -0,0 +1,51 @@
...
+                provider_key, type(e).__name__, str(e)))
+            continue
+
+        for img in images:
+            if "test_snapshot_" in img.name \
+                    and iso8601.parse_date(img.created_at) < time_limit:
```
digitronik: Depends on the above question,
I would add a property `create_at` to `wrapanapi` `OpenstackImage` class.

scripts/cleanup_openstack_instance_snapshot.py
```diff
@@ -0,0 +1,51 @@
...
+
+import iso8601
+import tzlocal
+
+from cfme.utils.providers import get_mgmt
+from cfme.utils.providers import list_provider_keys
+
+local_tz = tzlocal.get_localzone()
+GRACE_TIME = timedelta(hours=2)
+time_limit = datetime.now(tz=local_tz) - GRACE_TIME
```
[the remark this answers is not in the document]
mnadeem92: Done

scripts/cleanup_openstack_instance_snapshot.py
```diff
@@ -0,0 +1,63 @@
...
+    :param providers: Lists provider keys
+    :return:
+    """
+
+    for provider_key in providers:
+        logger.info("Cleaning up {}".format(provider_key))
+        provider_obj = get_mgmt(provider_key)
+
+        try:
+            images = provider_obj.list_templates()
```
digitronik: No need to capture exception it will handle by `wrapanapi`. also if you handle an exception you will get an error at `L48` use object before initialization (`images`)

digitronik:
For now, moving to WIPTEST  as it needs wrapanapi PR

scripts/cleanup_openstack_instance_snapshot.py
mnadeem92: we need this exception to continue cleanup on multiple provider such that script wont exited in case if one provider is not reachable, 
further at L48 will never execute if it goes in exception and it again continue to the for loop and check for other provider.

scripts/cleanup_openstack_instance_snapshot.py
```diff
@@ -0,0 +1,63 @@
...
+        except Exception as e:
+            logger.exception("Connect to provider failed:{} {} {}".format(
+                provider_key, type(e).__name__, str(e)))
+            continue
+
+        for img in images:
+            if "test_snapshot_" in img.name \
```
digitronik: optional:
please avoid the use of `\`

scripts/cleanup_openstack_instance_snapshot.py
mshriver: Not correct that all the scripts are py2. Cleanup scripts are explicitly py3 because of their use of starmap.

scripts/cleanup_openstack_instance_snapshot.py
```diff
@@ -0,0 +1,63 @@
...
+                except Exception as e:
+                    logger.exception("Delete failed: {} {}".format(type(e).__name__, str(e)))
+
+
+if __name__ == "__main__":
+    provs = sys.argv[1:]
+    if provs:
+        main(*provs)
+    else:
+        main(*list_provider_keys("openstack"))
```
mshriver: :+1:

scripts/cleanup_openstack_instance_snapshot.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+
+    for provider_key in providers:
+        logger.info("Cleaning up {}".format(provider_key))
+        provider_obj = get_mgmt(provider_key)
+
+        try:
+            images = provider_obj.list_templates()
+        except Exception as e:
+            logger.exception("Connect to provider failed:{} {} {}".format(
```
mshriver: You don't need to do all this, `logger.exception` logs the exception context. You just need to provide a plain language description of what the failure is.

scripts/cleanup_openstack_instance_snapshot.py
```diff
@@ -0,0 +1,63 @@
...
+            continue
+
+        for img in images:
+            if "test_snapshot_" in img.name \
+                    and iso8601.parse_date(img.creation_time) < TIME_LIMIT:
+                logger.info("Deleting {}".format(img.name))
+                try:
+                    img.delete()
+                except Exception as e:
+                    logger.exception("Delete failed: {} {}".format(type(e).__name__, str(e)))
```
mshriver: Again, `logger.exception` will handle the exception context here, just log that the delete failed - and log the `img.name` property so that the message is useful.

scripts/cleanup_openstack_instance_snapshot.py
mshriver: This can certainly be re-formatted to avoid.

Alternatively, this could be applied much earlier, when making the API query in the first place.

`images = [i for i in provider_obj.list_templates() if i.name.startswith('test_snapshot_')]`

Even better, expand the script to use ArgParse and take a CLI arg of what image names to remove, so that its not hardcoded as 'test_snapshot'. :)  But that's optional.

mnadeem92: Thank you for suggesting ArgParse idea, Implemented it.  For your other suggestion of using list comprehension , it will require looping 2 times, once as you mentioned above and next when i need to loop the images for deleting it. So i have keep it as it is and removed the concern "\" :)

scripts/cleanup_openstack_instance_snapshot.py
mshriver: pre-commit configuration is using reorder-python-imports, which enforces one import per line.  Not my personal preference, but the standard we chose to adhere to since its enforced automatically.

scripts/cleanup_openstack_instance_snapshot.py
```diff
@@ -0,0 +1,67 @@
...
+from cfme.utils.log import add_stdout_handler
+from cfme.utils.log import logger
+from cfme.utils.providers import get_mgmt
+from cfme.utils.providers import list_provider_keys
+
+LOCAL_TZ = tzlocal.get_localzone()
+GRACE_TIME = timedelta(hours=2)
+TIME_LIMIT = datetime.now(tz=LOCAL_TZ) - GRACE_TIME
+
+# log to stdout too
```
mshriver: :tada:

PR #8841, opened 2019-05-15
[1LP][RFR]v2v cancel migration tests

sshveta:
{{pytest: cfme/tests/v2v/test_v2v_cancel_migrations.py --use-provider rhv43 --use-provider vsphere67-ims --provider-limit 2 -vvvv}}

PR #8868, opened 2019-05-24
[1LP][RFR] Cancel migration and check if instance, volumes and ports get deleted

Yadnyawalkya:
{{pytest: cfme/tests/v2v/test_v2v_cancel_migrations.py --use-provider osp13-ims --use-provider vsphere67-ims --provider-limit 2 -vvvv}}

PRT should pass after - 
- [x] @sshveta's PR get merged https://github.com/ManageIQ/integration_tests/pull/8841
- [x] Wrapenapi PR get merged https://github.com/ManageIQ/wrapanapi/pull/383
- [x] After `osp13-ims`'s keystone services have public IP address

_keystone service issue -
I am using `get_vm` function of wrapenapi which works on kapi aka keystone api in backend. When wrapenapi try to connect to kapi for authentication, since `osp13-ims`'s keystone endpoint exposes private IP we are unable to connect there. I have raised this concerns to @mnadeem92  and he might replace private IP with public IP if possible or redeploy OSP13-ims so that keystone will have public IP. Meanwhile, we can always use `rhos13` if we want to check if mentioned wrapenapi code works but we can not use this provider for end to end migration since this is not suppose to be target provider for IMS._

### DIFF
--- a/cfme/tests/v2v/test_v2v_cancel_migrations.py
+++ b/cfme/tests/v2v/test_v2v_cancel_migrations.py
@@ -93,3 +93,73 @@ def test_dual_vm_cancel_migration(request, appliance, soft_assert, provider,
         soft_assert(request_details_list.is_cancelled(vm))
         soft_assert(request_details_list.progress_percent(vm) < 100.0 or
                     "Virtual machine migrated" not in request_details_list.get_message_text(vm))
+
+
+@pytest.mark.tier(2)
+@pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


mshriver:
Marking WIPTEST as the PR has 3 open dependencies.

Yadnyawalkya:
We are not considering backend process usecase atm due to its uncertain behavior, we have asked devel to fix that (https://bugzilla.redhat.com/show_bug.cgi?id=1666799#c26). We do have workaround but this have lots of hardcoding which does not look good so better to skip it atm.

PR #8870, opened 2019-05-27
[WIPTEST]fix failing test_report_view

valaparthvi:
Changes:
1. Change the report which is the subject of this test. Old report sometimes failed for cloud, the changed report will work for both infra and cloud providers.

{{ pytest: cfme/tests/intelligence/reports/test_views.py --use-provider=rhos13 --use-template-cache -sqvvv }}

### DIFF
--- a/cfme/tests/intelligence/reports/test_views.py
+++ b/cfme/tests/intelligence/reports/test_views.py
@@ -19,8 +19,8 @@ pytestmark = [
 def report(appliance):
     report = appliance.collections.reports.instantiate(
         type="Configuration Management",
-        subtype="Hosts",
-        menu_name="Virtual Infrastructure Platforms",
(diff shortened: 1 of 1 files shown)


jaryn:
Can we select both providers for the py.test run?

valaparthvi:
That's the thing @jaryn. Earlier with the old report, we were generating the report for `Virtual Infrastructure Platforms`. If the test first ran with cloud provider, it would fail since it would not find any relevant data and generate an empty report on queueing. But if the test ran with infra provider first and then ran with cloud provider on the same appliance, it would to pass, since it would pick up old data of infra provider.

I needed a report that would work for both the providers. I need to test with cloud provider only to make sure that the test passes even if we run it on a fresh appliance.

PR #8871, opened 2019-05-27
[1LP][RFR] CB- roles Selection test

digitronik:
Purpose or Intent
=================
- Created widget as use of checkbox was difficult
- Added support of role in custom button model
- Added test for BZ-1703588

{{pytest: cfme/tests/automate/custom_button/test_buttons.py::test_custom_button_role_selection}}

### DIFF
--- a/cfme/automate/buttons.py
+++ b/cfme/automate/buttons.py
@@ -115,7 +115,7 @@ class ButtonFormCommon(AutomateCustomizationView):
                 return [(i,) for i in range(1, 6)]
 
         role_show = BootstrapSelect(id="visibility_typ")
-        roles = RolesSelector()
+        roles = RolesSelector(locator="//label[contains(text(),'User Roles')]/../div/table")
 
     cancel_button = Button("Cancel")
 
@@ -523,6 +523,7 @@ class ButtonCollection(BaseCollection):
 
         if roles:
             view.advanced.role_show.fill("<By Role>")
+            view.advanced.roles.wait_displayed("20s")
             view.advanced.roles.fill(roles)
         else:
             view.advanced.role_show.fill("<To All>")
--- a/cfme/automate/buttons.py
+++ b/cfme/automate/buttons.py
@@ -28,6 +28,7 @@ from widgetastic_manageiq import AutomateRadioGroup
 from widgetastic_manageiq import FonticonPicker
 from widgetastic_manageiq import MultiBoxOrderedSelect
 from widgetastic_manageiq import PotentiallyInvisibleTab
+from widgetastic_manageiq import RolesSelector
 from widgetastic_manageiq import SummaryFormItem
 from widgetastic_manageiq.expression_editor import ExpressionEditor
 
@@ -113,7 +114,8 @@ class ButtonFormCommon(AutomateCustomizationView):
             def all(cls, browser):
                 return [(i,) for i in range(1, 6)]
 
-        # TODO: Role Access
+        role_show = BootstrapSelect(id="visibility_typ")
+        roles = RolesSelector()
 
     cancel_button = Button("Cancel")
 
@@ -183,6 +185,7 @@ class ButtonDetailView(AutomateCustomizationView):
     type = SummaryFormItem("Object Attribute", "Type")
 
     show = SummaryFormItem("Visibility", "Show")
+    user_roles = SummaryFormItem("Visibility", "User Roles")
 
     @property
     def is_displayed(self):
@@ -220,9 +223,15 @@ class BaseButton(BaseEntity, Updateable):
                         "hosts": updates.get("hosts"),
                     },
                 },
-                "advanced": {"system": updates.get("system"), "request": updates.get("request")},
+                "advanced": {
+                    "system": updates.get("system"),
+                    "request": updates.get("request"),
+                    "role_show": updates.get("role_show"),
+                    "roles": updates.get("roles"),
+                },
             }
         )
+
         if changed:
             view.save_button.click()
         else:
@@ -293,6 +302,16 @@ class BaseButton(BaseEntity, Updateable):
                 view.submit_button.click()
             view.flash.assert_no_error()
 
+    @property
+    def user_roles(self):
+        """This property use to check roles
+
+        Return: `To All` if button not assigned to specific role else `list` of roles
+        """
+        view = navigate_to(self, "Details")
+        show = view.show.read()
+        return show if show == "To All" else view.user_roles.read().split(", ")
+
 
 @attr.s
 class DefaultButton(BaseButton):
@@ -311,6 +330,7 @@ class DefaultButton(BaseButton):
         attributes: Button attribute
         visibility: Visibility expression in terms of tag and its value
         enablement: Enablement expression in terms of tag and its value
+        roles: button assigned to specific roles
     """
 
     group = attr.ib()
@@ -325,6 +345,7 @@ class DefaultButton(BaseButton):
     attributes = attr.ib(default=None)
     visibility = attr.ib(default=None)
     enablement = attr.ib(default=None)
+    roles = attr.ib(default=None)
 
 
 @attr.s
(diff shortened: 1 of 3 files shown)


dajoRH:
Lint report for commit 0fc189bb26545ec5caa0418f1cd6d482a2ed4878:

- :bangbang: Line 5424:15: **E225** *missing whitespace around operator*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/automate/custom_button/test_buttons.py
```diff
@@ -484,7 +484,49 @@ def _clean():
...
 
-@pytest.mark.manual('manualonly')
+@pytest.mark.tier(3)
+def test_custom_button_role_selection(appliance, request):
+    """Test Icon and Request are required field while adding custom button.
+
+    Polarion:
+        assignee: ndhandre
+        initialEstimate: 1/6h
+        caseimportance: medium
```
lina-nikiforova: Q: maybe it should be high as BZ is high?

cfme/tests/automate/custom_button/test_buttons.py
```diff
@@ -484,7 +484,49 @@ def _clean():
...
+        startsin: 5.8
+        casecomponent: CustomButton
+        testSteps:
+            1. Add custom button with specific roles and verify from summery page
```
lina-nikiforova: typo: summary

widgetastic_manageiq/__init__.py
```diff
@@ -5368,3 +5368,59 @@ def select(self, name):
...
+            if not self.role_selected(el):
+                el.click()
+
+    def clear_all(self):
+        """Clear all selected roles"""
+        for el in self._elements.values():
+            if self.role_selected(el):
+                el.click()
+
+    def fill(self, roles):
```
lina-nikiforova: Can you explain why is it `fill` method. As I see, it selects specific roles unlike `select_all`

cfme/tests/automate/custom_button/test_buttons.py
digitronik: @lina-nikiforova thanks for pointing this; we are just checking updates so I would like to go with a medium

widgetastic_manageiq/__init__.py
digitronik: @lina-nikiforova as per widgetatic guidelines widget have `fill` and `read` and I'm not selecting all roles... I want to select specific roles.
1. `select_all` will select all available roles
2. `fill` method will select specific roles and it will help if we use `view.fill`

lina-nikiforova: @digitronik yep, I understood what it does but had a questioned the name. Thanks for the explanation, I think all good :)

dajoRH:
*CFME QE Bot*

cfme/automate/buttons.py
```diff
@@ -183,6 +185,7 @@ class ButtonDetailView(AutomateCustomizationView):
     type = SummaryFormItem("Object Attribute", "Type")
 
     show = SummaryFormItem("Visibility", "Show")
+    user_roles = SummaryFormItem("Visibility", "User Roles")
```
mshriver: TODO:

This view should be using `SummaryForm`, which is parametrized and will really reduce the number of individual widgets we have to define here.

widgetastic_manageiq/__init__.py
```diff
@@ -5374,5 +5374,61 @@ def select(self, name):
...
 
+class RolesSelector(Widget):
+    """This widget for custom button role selection"""
+
+    ROOT = "//div[@class='col-md-10']//table"
+    CELLS = "./tbody/tr/td"
+    INPUTS = "./tbody/tr/td/input"
+
+    def __init__(self, parent, logger=None):
+        Widget.__init__(self, parent=parent, logger=logger)
```
mshriver: This constructor is unnecessary, just let Widget's constructor resolve.

widgetastic_manageiq/__init__.py
```diff
@@ -5374,5 +5374,61 @@ def select(self, name):
         return False
 
 
+class RolesSelector(Widget):
+    """This widget for custom button role selection"""
+
+    ROOT = "//div[@class='col-md-10']//table"
```
mshriver: I don't think this is a good locator for this widget, as I'm not sure these classes will be very consistent. Is there nothing else in the DOM that can be used to identify this table?

widgetastic_manageiq/__init__.py
```diff
@@ -5374,5 +5374,61 @@ def select(self, name):
...
+    CELLS = "./tbody/tr/td"
+    INPUTS = "./tbody/tr/td/input"
+
+    def __init__(self, parent, logger=None):
+        Widget.__init__(self, parent=parent, logger=logger)
+
+    @property
+    def _elements(self):
```
mshriver: I'd like to give this property a more descriptive name, like `_cell_input_map`, since this isn't simply element objects.

widgetastic_manageiq/__init__.py
```diff
@@ -5374,5 +5374,61 @@ def select(self, name):
...
+        br = self.browser
+        return {
+            br.text(name): el for name, el in zip(br.elements(self.CELLS), br.elements(self.INPUTS))
+        }
+
+    @property
+    def roles(self):
+        """all available roles"""
+        return list(self._elements.keys())
```
mshriver: unnecessary `list()` call here, `keys()` will return a list even on an empty dict.

widgetastic_manageiq/__init__.py
```diff
@@ -5374,5 +5374,61 @@ def select(self, name):
...
+
+    def fill(self, roles):
+        """select roles"""
+        if set(roles) == set(self.currently_selected):
+            return False
+        for role in roles:
+            el = self._elements.get(role)
+            if not self.role_selected(el):
+                el.click()
+        return True
```
mshriver: I'd like to see a `read()` implementation here, returning `self.currently_selected`.

cfme/automate/buttons.py
digitronik: we will...

widgetastic_manageiq/__init__.py
digitronik: for `py2` it will return `list` but for `py3` `dict_keys` so for compatibility, we need to convert.

widgetastic_manageiq/__init__.py
digitronik: thanks, it will give a meaningful name to non-user property

widgetastic_manageiq/__init__.py
digitronik: I know... you can say it was for initial implementation. I forgot to replace it with locator as an argument. Now locator is an argument with `table` tag.

yup, we don't have supporting things in `dom` but I would like to construct `XPath` with child elements support. I construct it with the help of `label` l.e. `User Roles`.
```
locator=".//label[contains(text(),'User Roles')]/../div/table"
```

mshriver: Since this is a widget specific to that selector, I also don't think its a good idea to force the view to pass that entire locator. The locator needed to be more accurate than just the div class, but since the widget is specific for MIQ it can be a ParametrizedLocator, where the constructor just takes the label text. Contains vs exact match could be an additional kwarg

mshriver: its also good practice to wrap any `text()` use in normalized-space() call as well.

cfme/automate/buttons.py
```diff
@@ -113,7 +114,8 @@ class attribute(ParametrizedView):  # noqa
             def all(cls, browser):
                 return [(i,) for i in range(1, 6)]
 
-        # TODO: Role Access
+        role_show = BootstrapSelect(id="visibility_typ")
+        roles = RolesSelector(locator="//label[contains(text(),'User Roles')]/../div/table")
```
mshriver: I would suggest using a ParametrizedLocator here instead of having to form and pass an entire locator for a highly specific widget. But will accept as-is.

mshriver: I would suggest using a ParametrizedLocator here instead of having to form and pass an entire locator for a highly specific widget. But will accept as-is.

PR #8872, opened 2019-05-27
[1LP][RFR] Moved targeted refresh tests to cloud

mmojzis:
Moved targeted refresh tests to cloud and moved Steps under Polarion in docstrings.

### DIFF
--- a/cfme/tests/cloud/test_targeted_refresh.py
+++ b/cfme/tests/cloud/test_targeted_refresh.py
@@ -1,0 +1,200 @@
+import pytest
+
+from cfme import test_requirements
+
+pytestmark = [
+    test_requirements.ec2,
+]
+
+
+@pytest.mark.manual
+def test_ec2_targeted_refresh_instance():
+    """
+    Polarion:
+        assignee: mmojzis
+        casecomponent: Cloud
+        initialEstimate: 1 1/6h
+        startsin: 5.9
+
+        testSteps:
+            1. Instance CREATE
+            2. Instance RUNNING
+            3. Instance STOPPED
+            4. Instance UPDATE
+            5. Instance DELETE - or - Instance TERMINATE
+    """
+    pass
+
+
+@pytest.mark.manual
+def test_ec2_targeted_refresh_floating_ip():
+    """
+    Polarion:
+        assignee: mmojzis
+        casecomponent: Cloud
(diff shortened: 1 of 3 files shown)


cfme/tests/cloud/test_targeted_refresh.py
```diff
@@ -0,0 +1,201 @@
+import pytest
+
+from cfme import test_requirements
+
+pytestmark = [
+    test_requirements.ec2,
+    pytest.mark.ignore_stream('5.10', 'upstream')
```
lina-nikiforova: but in the tests it's `startsin: 5.9`...

mmojzis: I don't actually know what it means...It was in test_manual.py so I copied it here.

cfme/tests/cloud/test_targeted_refresh.py
```diff
@@ -0,0 +1,201 @@
...
+@pytest.mark.manual
+def test_ec2_targeted_refresh_instance():
+    """
+    Polarion:
+        assignee: mmojzis
+        casecomponent: Cloud
+        initialEstimate: 1 1/6h
+        startsin: 5.9
+
+        testSteps:
```
lina-nikiforova: testSteps should be numbered and have expectedResult (as it's a manual test).
Examples can be found in other manual test cases.

cfme/tests/cloud/test_targeted_refresh.py
lina-nikiforova: I believe it will not collect the test against 5.10. If it's not desired, it needs to be removed.

cfme/tests/cloud/test_targeted_refresh.py
lina-nikiforova: If you move tests, you can just move. I'd leave it in the description or add tests steps _and_ expected results if you want to make it better

mmojzis: Actually testSteps == expectedResults for now. I create instance in testStep and then just check it's created. I would only duplicate test steps to expectedResults...

mmojzis:
@lina-nikiforova Deleted pytest.mark.ignore_stream and testSteps are numbered.

PR #8873, opened 2019-05-28
[1LP][RFR] Skip maintenance zone in ZoneCollection

lina-nikiforova:
This is the new feature in 5.11, more details can be found here https://github.com/ManageIQ/manageiq/issues/17489

I found this test that was affected but maybe there are others:

{{ pytest: -v cfme/tests/configure/test_zones.py }}

### DIFF
--- a/cfme/base/__init__.py
+++ b/cfme/base/__init__.py
@@ -223,7 +223,11 @@ class ZoneCollection(BaseCollection, sentaku.modeling.ElementMixin):
         parent = self.filters.get('parent')
         for zone in zone_collection:
             zone.reload(attributes=['region_number'])
-            if parent and zone.region_number != parent.number:
+            # starting in 5.11 there's a maintenance zone that is not shown in the UI but
+            # is visible in API. We need to skip it for the correct navigation.
+            # https://bugzilla.redhat.com/show_bug.cgi?id=1455145
+            maintenance = 'Maintenance Zone'
+            if (parent and zone.region_number != parent.number) or zone.description == maintenance:
                 continue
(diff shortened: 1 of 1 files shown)


cfme/base/__init__.py
```diff
@@ -223,7 +223,11 @@ def all(self):
         parent = self.filters.get('parent')
         for zone in zone_collection:
             zone.reload(attributes=['region_number'])
-            if parent and zone.region_number != parent.number:
+            # starting in 5.11 there's a maintenance zone that is not shown in the UI but
+            # is visible in API. We need to skip it for the correct navigation.
+            # https://bugzilla.redhat.com/show_bug.cgi?id=1455145
+            maintenance = 'Maintenance Zone'
```
digitronik: optional:
I would avoid catching this in a variable better to use directly.

lina-nikiforova: I was struggling how to format this :(

digitronik:
Thanks for a run :+1: test fails not related to changes... 

@lina-nikiforova would you like to fix that test in the same PR? only need to replace
`Description can't be blank` with `Description is required` as per fist look. You can even go for new PR.
https://github.com/ManageIQ/integration_tests/blob/f717f65a004392eb1046b9ae388a817ea499ab75/cfme/tests/configure/test_zones.py#L169

lina-nikiforova:
@digitronik thanks a lot! I'm looking into it, if it's a quick fix, I can include it here

lina-nikiforova:
That's a bit weird... This test passes locally
image

Let's leave it like this for now and I will check the automation runs.

PR #8875, opened 2019-05-28
[1LP][RFR]Adding global test requirement for test cases

mnadeem92:


cfme/tests/test_replication.py
```diff
@@ -2,9 +2,12 @@
 
 from cfme import test_requirements
 
+pytestmark = [
```
digitronik: please in single line

PR #8877, opened 2019-05-28
[1LP][RFR] Fix AttributeError in HandleModalsMixin.handle_alert

mshriver:
When there is no wait, there's no check against the fact that the modal is actually present.

Addressing the following failure mode specifically:
```
>           self.logger.info('handling alert: %r', popup.text)
E           AttributeError: 'NoneType' object has no attribute 'text'

cfme/utils/appliance/implementations/common.py:80: AttributeError
AttributeError
'NoneType' object has no attribute 'text'
```

Local testing looks good, included some cases that failed in automation for PRT.

Not necessarily trying to resolve all PRT failures that might happen in these modules.

{{ pytest: cfme/tests/configure/test_access_control.py cfme/tests/configure/test_landing_page.py cfme/tests/infrastructure/test_quota.py }}

### DIFF
--- a/cfme/configure/access_control/__init__.py
+++ b/cfme/configure/access_control/__init__.py
@@ -1008,7 +1008,7 @@ class Role(Updateable, Pretty, BaseEntity):
 
     pretty_attrs = ['name', 'product_features']
 
-    name = attr.ib(default=None)
+    name = attr.ib()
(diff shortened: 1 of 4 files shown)


jawatts:
LGTM

PR #8881, opened 2019-05-29
[1LP][RFR] Updated test case of automate domain disable and lock

ganeshhubale:
- Updated domain disable and lock test case
- Added functionality check:
    - If we try to execute automate method of disabled domain using simulation then it does not work(gives error in automation logs)

{{ pytest: cfme/tests/automate/test_domain.py -k 'test_domain_lock_disabled' -vv }}

### DIFF
--- a/cfme/tests/automate/test_domain.py
+++ b/cfme/tests/automate/test_domain.py
@@ -7,9 +7,10 @@ from cfme.automate.explorer.domain import DomainAddView
 from cfme.automate.explorer.instance import InstanceCopyView
 from cfme.automate.explorer.klass import ClassCopyView
 from cfme.automate.explorer.method import MethodCopyView
+from cfme.automate.simulation import simulate
 from cfme.exceptions import OptionNotAvailable
 from cfme.utils.appliance.implementations.ui import navigate_to
-from cfme.utils.blockers import BZ
+from cfme.utils.log_validator import LogValidator
 from cfme.utils.update import update
 
 pytestmark = [test_requirements.automate]
@@ -70,7 +71,7 @@ def test_domain_edit_enabled(domain, appliance):
 
 
 @pytest.mark.tier(2)
-def test_domain_lock_disabled(request, appliance):
+def test_domain_lock_disabled(klass):
     """
     Polarion:
         assignee: ghubale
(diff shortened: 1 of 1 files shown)


dajoRH:
*CFME QE Bot*

cfme/tests/automate/test_domain.py
```diff
@@ -79,16 +80,56 @@ def test_domain_lock_disabled(request, appliance):
...
+        execute_methods=True,
+    )
+    result.validate_logs()
+
+    klass.namespace.domain.lock()
+    view = navigate_to(klass.namespace.domain, 'Details')
     assert 'Disabled' in view.title.text
     assert 'Locked' in view.title.text
 
+    # Need to unlock the domain to perform teardown on domain, namespace, class
```
sshveta: Where is the teardown happening btw ? 
Required : Delete the domain , namespace and class after test .

ganeshhubale: Hi @sshveta,

- Actually  domain, namespace and class(i.e. klass) are [global fixtures](https://github.com/ManageIQ/integration_tests/blob/master/cfme/fixtures/automate.py#L5).
- But here domain locking is comes under exception case and to delete ```domain``` we need to unlock it first. Hence added this step.

john-dupuy: @ganeshhubale Does `instance` need any teardown?

ganeshhubale: No. Actually, Instances or methods are under class. So their teardown happens when we delete the class itself.

cfme/tests/automate/test_domain.py
```diff
@@ -79,16 +80,56 @@ def test_domain_lock_disabled(request, appliance):
...
+    instance = klass.instances.create(
+        name=fauxfactory.gen_alphanumeric(),
+        display_name=fauxfactory.gen_alphanumeric(),
+        description=fauxfactory.gen_alphanumeric(),
+        fields={'execute': {'value': method.name}}
+    )
+
+    result = LogValidator(
+        "/var/www/miq/vmdb/log/automation.log",
+        matched_patterns=[r".*ERROR.*"],
```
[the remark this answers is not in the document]
ganeshhubale: - According to me, there is no need to add that entire error line in ```matched patterns``` because logs generated in this session(like after executing automate method via simulation) have this only this error and if I got any other error then also it must fail. 
- So better to check no error only.
- Please let me know you thoughts @john-dupuy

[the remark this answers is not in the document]
john-dupuy: Yep that makes sense to me! Thanks for the clarification.

PR #8883, opened 2019-05-29
[1LP][RFR] New Test: Test generic object service associations using automate method

ganeshhubale:
{{ pytest: cfme/tests/automate/test_method.py -k 'test_automate_generic_object_service_associations' -vv }}

Depend on: #8900

### DIFF
--- a/cfme/tests/automate/test_method.py
+++ b/cfme/tests/automate/test_method.py
@@ -353,7 +353,7 @@ def test_send_email_method(smtp_test, klass):
 
 
 @pytest.fixture(scope="module")
-def generic_definition(appliance):
+def generic_object_definition(appliance):
     # Creating generic object using REST
     with appliance.context.use(ViaREST):
         definition = appliance.collections.generic_object_definitions.create(
@@ -363,22 +363,22 @@ def generic_definition(appliance):
             associations={"vms": "Vm", "services": "Service"}
         )
         yield definition
-        if definition.exists:
-            definition.delete()
+        definition.delete_if_exists()
 
 
 @pytest.fixture
-def service_req(request, generic_catalog_item):
+def go_service_request(generic_catalog_item):
     # Creating generic service
     service_request = generic_catalog_item.appliance.rest_api.collections.service_templates.get(
         name=generic_catalog_item.name
     ).action.order()
-    request.addfinalizer(service_request.action.delete)
+    yield
+    service_request.action.delete()
 
 
 @pytest.mark.tier(1)
-def test_automate_generic_object_service_associations(appliance, klass, service_req,
-                                                      generic_definition):
+def test_automate_generic_object_service_associations(appliance, klass, go_service_request,
+                                                      generic_object_definition):
     """
     Polarion:
         assignee: ghubale
@@ -392,7 +392,7 @@ def test_automate_generic_object_service_associations(appliance, klass, service_
     """
     # Ruby code
     script = 'go_class = $evm.vmdb(:generic_object_definition).find_by(:name => "{name}")\n'.format(
-        name=generic_definition.name
+        name=generic_object_definition.name
     )
     script = script + (
         'load_balancer = go_class.create_object(:name => "Test Load Balancer", :location => '
--- a/cfme/tests/automate/test_method.py
+++ b/cfme/tests/automate/test_method.py
@@ -5,7 +5,9 @@ import pytest
 from cfme import test_requirements
 from cfme.automate.explorer.klass import ClassDetailsView
 from cfme.automate.simulation import simulate
+from cfme.utils.appliance.implementations.rest import ViaREST
 from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.appliance.implementations.ui import ViaUI
 from cfme.utils.blockers import BZ
 from cfme.utils.log_validator import LogValidator
 from cfme.utils.update import update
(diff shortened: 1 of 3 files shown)


cfme/tests/automate/test_method.py
```diff
@@ -348,3 +350,99 @@ def test_send_email_method(smtp_test, klass):
...
+            display_name=fauxfactory.gen_alphanumeric(),
+            description=fauxfactory.gen_alphanumeric(),
+            fields={'execute': {'value': method.name}}
+        )
+
+        result = LogValidator(
+            "/var/www/miq/vmdb/log/automation.log",
+            matched_patterns=[
+                r'.*XYZ go object: #<MiqAeServiceGenericObject.*',
+                r'.*XYZ load balancer got service: #<MiqAeServiceService.*'
```
sshveta: Typo : "go service"

cfme/tests/automate/test_method.py
```diff
@@ -348,3 +350,99 @@ def test_send_email_method(smtp_test, klass):
...
+            attributes={"location": "string"},
+            associations={"vms": "Vm", "services": "Service"}
+        )
+        yield definition
+        if definition.exists:
+            definition.delete()
+
+
+@pytest.fixture
+def service_req(request, generic_catalog_item):
```
sshveta: Optional: It will be good to rename the fixture to go_service_request

cfme/tests/automate/test_method.py
```diff
@@ -348,3 +350,99 @@ def test_send_email_method(smtp_test, klass):
...
+    )
+    script = script + (
+        'load_balancer = go_class.create_object(:name => "Test Load Balancer", :location => '
+        '"Mahwah")\n'
+        '$evm.log("info", "XYZ go object: #{load_balancer.inspect}")\n'
+        'service = $evm.vmdb(:service).first\n'
+        'load_balancer.services = [service]\n'
+        'content_type = "message"\n'
+        'load_balancer.save!\n'
+        '$evm.log("info", "XYZ load balancer got service: #{load_balancer.services.first.inspect}")'
```
sshveta: Should this be "XYZ load balancer go service" ?

cfme/tests/automate/test_method.py
```diff
@@ -348,3 +350,99 @@ def test_send_email_method(smtp_test, klass):
...
+            attributes_values={
+                "namespace": klass.namespace.name,
+                "class": klass.name,
+                "instance": instance.name,
+            },
+            message="create",
+            request="Call_Instance",
+            execute_methods=True,
+        )
+        result.validate_logs()
```
sshveta: Can you tell how are we asserting on go association with service . Can we check the service page and see if GO is displayed there ?

cfme/tests/automate/test_method.py
ganeshhubale: I have used this code directly from BZ and I think it is written for understanding in logs.
Please suggest your thoughts if you want it in other way :)

cfme/tests/automate/test_method.py
ganeshhubale: https://bugzilla.redhat.com/show_bug.cgi?id=1410920#c3
```
In the tail'ed log:
There should be no ERROR lines related to the execution.
There should be these two lines:
<AEMethod gotest> XYZ go object: #<MiqAeServiceGenericObject ....something...>
<AEMethod gotest> XYZ load balancer got service: #<MiqAeServiceService:....something....>
If there is "XYZ load balancer got service: nil", then this bug was reproduced.
```
On the basis of comment, I am verifying logs.

I am not much familiar with generic objects. So please let me know other ways to improve implementation of this test case.

cfme/tests/automate/test_method.py
ganeshhubale: I think it's not type: previously it was giving ```nil```(when BZ is reproduced). So it is saying if LoadBalancer ```got``` any service or not.

cfme/tests/automate/test_method.py
```diff
@@ -348,3 +350,99 @@ def test_send_email_method(smtp_test, klass):
...
+    # Creating generic object using REST
+    with appliance.context.use(ViaREST):
+        definition = appliance.collections.generic_object_definitions.create(
+            name="LoadBalancer_{}".format(fauxfactory.gen_numeric_string(3)),
+            description="LoadBalancer",
+            attributes={"location": "string"},
+            associations={"vms": "Vm", "services": "Service"}
+        )
+        yield definition
+        if definition.exists:
```
john-dupuy: You can use `delete_if_exists()` here.

cfme/tests/automate/test_method.py
```diff
@@ -348,3 +350,99 @@ def test_send_email_method(smtp_test, klass):
     # TODO(GH-8820): This issue should be fixed to check mails sent to person in 'cc' and 'bcc'
     # Check whether the mail sent via automate method really arrives
     wait_for(lambda: len(smtp_test.get_emails(to_address=mail_to)) > 0, num_sec=60, delay=10)
+
+
+@pytest.fixture(scope="module")
+def generic_definition(appliance):
```
john-dupuy: Why not call this a `generic_object_definition`? I think that would be a bit more descriptive.

cfme/tests/automate/test_method.py
```diff
@@ -348,3 +350,99 @@ def test_send_email_method(smtp_test, klass):
...
+            definition.delete()
+
+
+@pytest.fixture
+def go_service_request(request, generic_catalog_item):
+    # Creating generic service
+    service_request = generic_catalog_item.appliance.rest_api.collections.service_templates.get(
+        name=generic_catalog_item.name
+    ).action.order()
+    request.addfinalizer(service_request.action.delete)
```
john-dupuy: I prefer using `yield` in fixtures like this one, that way there's no need to pass `request` to the fixture. But I'll leave that as optional.

john-dupuy:
Awaiting PRT success for approval

cfme/tests/automate/test_method.py
kedark3: How about replacing `XYZ` with `test_automate_generic_object_service_associations` ? More easy to find in logs, IMO.

ganeshhubale: I don't think, we need to change this and I am not getting how it will  be easy to find in log by changin it to ```test_automate_generic_object_service_associations``` because at the end we will grep or search the string we want in logs.

kedark3: It would make it less ambiguous, because one cannot be sure if there will be another entry in the logs including `XYZ` but can be more sure that your test name won't be logged by anything apart from your test. It is just a suggestion.

PR #8887, opened 2019-05-30
[RFR] test coverage for BZ1443411

lina-nikiforova:
{{ pytest: -v cfme/tests/infrastructure/test_snapshot.py::test_delete_active_vm_snapshot --use-template-cache  --use-provider=complete --long-running }}

### DIFF
--- a/cfme/tests/infrastructure/test_snapshot.py
+++ b/cfme/tests/infrastructure/test_snapshot.py
@@ -116,6 +116,28 @@ def test_snapshot_crud(small_test_vm, provider):
 
 @pytest.mark.rhv3
 @pytest.mark.provider([RHEVMProvider], override=True)
+@pytest.mark.meta(automates=[BZ(1443411)])
+def test_delete_active_vm_snapshot(small_test_vm):
+    """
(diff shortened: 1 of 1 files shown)


cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -114,6 +114,28 @@ def test_snapshot_crud(small_test_vm, provider):
     snapshot.delete()
 
 
+@pytest.mark.rhv3
+@pytest.mark.provider([RHEVMProvider], override=True)
```
digitronik: As we are getting `RHEVM` from provider parametrization. I would just uncollect for `VMwareProvider` than override. what do you think?

mshriver: We prefer direct marking, not uncollection. Its the intended use of `override`, and is a much cleaner implementation than having to uncollect with lambda expressions.

lina-nikiforova: @mshriver thanks, I didn't know what is the better way

PR #8891, opened 2019-05-31
[1LP][RFR] Add test_fixauth_dryrun_has_feedback

jaryn:
Add test for BZ 1577303

### DIFF
--- a/cfme/tests/test_appliance.py
+++ b/cfme/tests/test_appliance.py
@@ -465,3 +465,25 @@ def test_ec2_deploy_cfme_image():
             4. CFME appliance should work
     """
     pass
+
+
+def test_fixauth_dryrun_has_feedback(temp_appliance_preconfig):
+    """
+    Check whether the fixauth says it is running in dry mode
+
+    Polarion:
+        assignee: jhenner
+        casecomponent: Appliance
+        initialEstimate: 1/60h
+
+    Bugzilla:
+        1577303
+    """
+    appliance = temp_appliance_preconfig
+    run_command = appliance.ssh_client.run_command
+    dry_run_message = (
+        'is executing in dry-run mode, and no actual changes will be made **')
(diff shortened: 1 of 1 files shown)


cfme/tests/test_appliance.py
```diff
@@ -435,3 +435,24 @@ def codename_in_stdout():
...
+        casecomponent: Appliance
+        initialEstimate: 1/60h
+
+    Testing BZ 1577303
+    """
+    appliance = temp_appliance_preconfig
+    run_command = appliance.ssh_client.run_command
+    dry_run_message = (
+        'is executing in dry-run mode, and no actual changes will be made **')
+    assert dry_run_message in run_command("fix_auth -d").output
```
digitronik: optional:
we can combine this 4 assertions but totally optional

jaryn: I don't like to combine them. I don't see a reason for doing so. I see just a reason for _not_ doing so. When they are separate, one can more easily see which one failed.

dajoRH:
*CFME QE Bot*

cfme/tests/test_appliance.py
```diff
@@ -435,3 +435,24 @@ def codename_in_stdout():
...
+def test_fixauth_dryrun_has_feedback(temp_appliance_preconfig):
+    """
+    Check whether the fixauth says it is running in dry mode
+
+    Polarion
+        assignee: jhenner
+        casecomponent: Appliance
+        initialEstimate: 1/60h
+
+    Testing BZ 1577303
```
digitronik: I missed this... @jaryn we started Bugzilla collection consistency under a common tag like 

https://github.com/ManageIQ/integration_tests/blob/c5cba3736dea0110a5d01a6d345d61214f4f6102/cfme/tests/automate/test_domain.py#L273

Also if you are automating you and add `automate` marker.

PR #8892, opened 2019-05-31
[1LP][RFR] new test to check template refresh relationships

lina-nikiforova:
{{ pytest: -v cfme/tests/cloud_infra_common/test_relationships.py::test_template_refresh_relationships }}

+ added `BZ.blocks` due to breadcrumb issue

### DIFF
--- a/cfme/tests/cloud_infra_common/test_relationships.py
+++ b/cfme/tests/cloud_infra_common/test_relationships.py
@@ -326,6 +326,28 @@ def test_host_refresh_relationships(provider, setup_provider):
     wait_for_relationship_refresh(provider)
 
 
+@pytest.mark.rhv3
+@pytest.mark.provider([InfraProvider])
+def test_template_refresh_relationships(appliance, provider, setup_provider):
+    """ Test that host refresh doesn't fail
+
+    Polarion:
+        assignee: ghubale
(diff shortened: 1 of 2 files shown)


cfme/tests/cloud_infra_common/test_relationships.py
```diff
@@ -326,6 +326,28 @@ def test_host_refresh_relationships(provider, setup_provider):
...
+def test_template_refresh_relationships(appliance, provider, setup_provider):
+    """ Test that host refresh doesn't fail
+
+    Polarion:
+        assignee: ghubale
+        casecomponent: Infra
+        caseimportance: high
+        initialEstimate: 1/8h
+        tags: relationship
+    """
```
ganeshhubale: I think, you should remove ```tags``` and ```caseimportance```(By default it is high).

lina-nikiforova: @ganeshhubale thanks for the review!
There's no harm in `caseimportance` - I think it's more clear.
What about the tag? I see it in all similar tests, so why should it be removed here?

ganeshhubale: I talked with @digitronik 
We can put it as it is optional.

PR #8893, opened 2019-06-03
[RFR] Removing test_appliance_console_screen test

mnadeem92:
This test was depend on screen package which
is not included in rhel8, i confirmed with author
of the TC, this TC is no more required for testing
any of the CFME feature, so removing it.

### DIFF
--- a/cfme/tests/cli/test_appliance_console.py
+++ b/cfme/tests/cli/test_appliance_console.py
@@ -731,36 +731,6 @@ def test_appliance_console_ha_setup_dc():
 
 @pytest.mark.manual
 @pytest.mark.tier(2)
-def test_appliance_console_screen():
-    """
-    Test new screen package added to appliances. Should be able to run a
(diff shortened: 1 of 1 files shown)


cfme/tests/cli/test_appliance_console.py
```diff
@@ -729,36 +729,6 @@ def test_appliance_console_ha_setup_dc():
     pass
 
 
-@pytest.mark.manual
-@pytest.mark.tier(2)
-def test_appliance_console_screen():
```
john-dupuy: @mnadeem92 the `screen` package is still present on 5.10.5.1 appliances, i.e. RHEL 7.6, shouldn't we still include this test case for 5.10?

mshriver:
Without this being linked to a customer BZ, I can't see a benefit from including it, especially as a manual test case. If we want to test this, we should do it by verifying the RPM is installed on the appliance (automated).

PR #8894, opened 2019-06-03
[1LP][RFR]Automate test: test_report_edit_secondary_display_filter AND update E…

valaparthvi:
1. Automate test: test_report_edit_secondary_display_filter
2. Fix primary_filter and secondary_filter so that right data can be accessed.
3. Enhance the ExpressionEditor `read` method to return a complete expression when required, earlier it only used to return the first expression of the multiple expressions present and remove 5.9 references from ExpressionEditor.

{{ pytest: cfme/tests/intelligence/reports/test_reports.py::test_report_edit_secondary_display_filter
 cfme/tests/control/test_bugs.py::test_policy_condition_multiple_ors cfme/tests/control/test_basic.py::test_modify_condition_expression
 --use-template-cache --long-running -sqvvv }}

### DIFF
--- a/cfme/intelligence/reports/reports.py
+++ b/cfme/intelligence/reports/reports.py
@@ -175,8 +175,15 @@ class ReportDetailsView(CloudIntelReportsView):
         user = SummaryFormItem("", "User")
         group = SummaryFormItem("", "EVM Group")
         updated_on = SummaryFormItem("", "Updated On")
-        report_schedule_data = Table('//*[@id="report_info"]/table[1]')
-        report_widgets_data = Table('//*[@id="report_info"]/table[2]')
+        # xpath is defined based on the value of the second title head of table
+        report_schedule_data = Table(
+            '//*[@id="report_info"]//th[contains(text(), "Name")]'
+            '/parent::tr/parent::thead/parent::table'
+        )
+        report_widgets_data = Table(
+            '//*[@id="report_info"]//th[contains(text(), "Title")]'
+            '/parent::tr/parent::thead/parent::table'
+        )
         queue_button = Button("Queue")
 
     @View.nested
--- a/cfme/intelligence/reports/reports.py
+++ b/cfme/intelligence/reports/reports.py
@@ -34,6 +34,7 @@ from cfme.utils.wait import wait_for
 from widgetastic_manageiq import InputButton
 from widgetastic_manageiq import PaginationPane
 from widgetastic_manageiq import ReportToolBarViewSelector
+from widgetastic_manageiq import SummaryFormItem
 from widgetastic_manageiq import Table
 from widgetastic_manageiq import WaitTab
 from widgetastic_manageiq.expression_editor import ExpressionEditor
@@ -78,8 +79,20 @@ class CustomReportFormCommon(CloudIntelReportsView):
         interval = BootstrapSelect("cb_interval")
         interval_size = BootstrapSelect("cb_interval_size")
         interval_end = BootstrapSelect("cb_end_interval_offset")
-        primary_filter = ExpressionEditor()
-        secondary_filter = ExpressionEditor()
+        primary_record_filter = Button("contains", "Record Filter")
+        secondary_display_filter = Button("contains", "Display Filter")
+
+        @View.nested
+        class primary_filter(ExpressionEditor):     # noqa
+            def child_widget_accessed(self, widget):
+                if self.parent.primary_record_filter.is_displayed:
+                    self.parent.primary_record_filter.click()
+
+        @View.nested
+        class secondary_filter(ExpressionEditor):   # noqa
+            def child_widget_accessed(self, widget):
+                if self.parent.secondary_display_filter.is_displayed:
+                    self.parent.secondary_display_filter.click()
 
     @View.nested
     class summary(WaitTab):  # noqa
@@ -152,6 +165,17 @@ class ReportDetailsView(CloudIntelReportsView):
     @View.nested
     class report_info(WaitTab):  # noqa
         TAB_NAME = "Report Info"
+        report_id = SummaryFormItem("", "ID")
+        title = SummaryFormItem("", "Title")
+        primary_filter = SummaryFormItem("", "Primary (Record) Filter")
+        secondary_filter = SummaryFormItem("", "Secondary (Display) Filter")
+        sort_by = SummaryFormItem("", "Sort By")
+        based_on = SummaryFormItem("", "Based On")
+        user = SummaryFormItem("", "User")
+        group = SummaryFormItem("", "EVM Group")
+        updated_on = SummaryFormItem("", "Updated On")
+        report_schedule_data = Table('//*[@id="report_info"]/table[1]')
+        report_widgets_data = Table('//*[@id="report_info"]/table[2]')
         queue_button = Button("Queue")
 
     @View.nested
--- a/cfme/intelligence/reports/reports.py
+++ b/cfme/intelligence/reports/reports.py
@@ -165,6 +165,7 @@ class ReportDetailsView(CloudIntelReportsView):
     @View.nested
     class report_info(WaitTab):  # noqa
         TAB_NAME = "Report Info"
+        # Keeping `group_title` empty since the summary form has no title
         report_id = SummaryFormItem("", "ID")
         title = SummaryFormItem("", "Title")
         primary_filter = SummaryFormItem("", "Primary (Record) Filter")
--- a/cfme/tests/intelligence/reports/test_reports.py
+++ b/cfme/tests/intelligence/reports/test_reports.py
@@ -152,8 +152,39 @@ def test_new_report_fields(appliance, based_on, request):
     assert report.exists
 
 
+@pytest.fixture
+def filter_report(appliance):
+    report_data = {
+        "title": "Testing report",
+        "menu_name": "testing report",
+        "base_report_on": "VMs and Instances",
+        "report_fields": [
+            "Active",
+            "EVM Custom Attributes : Name",
+            "EVM Custom Attributes : Region Description",
+            "EVM Custom Attributes : Region Number",
+            "Name",
+        ],
+        "consolidation": {
+            "group_records": [
+                "EVM Custom Attributes : Name",
+                "EVM Custom Attributes : Region Description",
+                "EVM Custom Attributes : Region Number",
+            ]
+        },
+        "filter": {
+            "primary_filter": "fill_field(VM and Instance : Active, IS NOT NULL)",
+            "secondary_filter": "fill_field(EVM Custom Attributes : Name, INCLUDES, A)",
+        },
+    }
+    report = appliance.collections.reports.create(**report_data)
+    yield report
+    report.delete_if_exists()
+
+
 @pytest.mark.tier(1)
-def test_report_edit_secondary_display_filter(appliance, request, soft_assert):
+@pytest.mark.meta(automates=[1565171])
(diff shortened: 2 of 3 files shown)


dajoRH:
*CFME QE Bot*

widgetastic_manageiq/expression_editor.py
```diff
@@ -125,6 +125,11 @@ class date_relative_form_view(View):  # noqa
     SELECT_SPECIFIC = ".//a[@title='Click to change to a specific Date/Time format']"
     SELECT_RELATIVE = ".//a[@title='Click to change to a relative Date/Time format']"
 
+    # This returns the whole expression unlike EXPRESSION_TEXT which only returns
+    # the first expression
+    COMPLETE_EXPRESSION_TEXT = '//*[@id="exp_editor_div"]//p'
```
john-dupuy: My thought is that this should be the `EXPRESSION_TEXT` and the current `EXPRESSION_TEXT` should be renamed to `FIRST_EXPRESSION` or something similar. It is a bit confusing to have `EXPRESSION_TEXT` and `COMPLETE_EXPRESSION_TEXT`. This could, however, result in a more major refactor. What do you think?

cfme/intelligence/reports/reports.py
```diff
@@ -152,6 +165,17 @@ class ReportDetailsView(CloudIntelReportsView):
     @View.nested
     class report_info(WaitTab):  # noqa
         TAB_NAME = "Report Info"
+        report_id = SummaryFormItem("", "ID")
```
john-dupuy: _Optional_ A brief comment about the blank string `""` might be helpful.

cfme/intelligence/reports/reports.py
```diff
@@ -78,8 +79,20 @@ class filter(WaitTab):  # noqa
...
         interval_size = BootstrapSelect("cb_interval_size")
         interval_end = BootstrapSelect("cb_end_interval_offset")
-        primary_filter = ExpressionEditor()
-        secondary_filter = ExpressionEditor()
+        primary_record_filter = Button("contains", "Record Filter")
+        secondary_display_filter = Button("contains", "Display Filter")
+
+        @View.nested
+        class primary_filter(ExpressionEditor):     # noqa
+            def child_widget_accessed(self, widget):
```
john-dupuy: _Question_ could you explain why `widget` is passed here? I haven't seen this used before and am curious as to how it works.

cfme/tests/intelligence/reports/test_reports.py
```diff
@@ -149,3 +150,87 @@ def test_new_report_fields(appliance, based_on, request):
...
+        initialEstimate: 1/6h
+        setup:
+            1. Create/Copy a report with secondary (display) filter.
+        testSteps:
+            1. Edit the secondary filter and test if the report was updated.
+        expectedResults:
+            1. Secondary filter must be editable and it must be updated.
+
+    Bugzilla:
+        1565171
```
john-dupuy: Can this be added as `automates` metadata?

cfme/tests/intelligence/reports/test_reports.py
```diff
@@ -149,3 +150,87 @@ def test_new_report_fields(appliance, based_on, request):
...
+                "EVM Custom Attributes : Region Number",
+            ]
+        },
+        "filter": {
+            "primary_filter": "fill_field(Instance : Active, IS NOT NULL)",
+            "secondary_filter": "fill_field(EVM Custom Attributes : Name, INCLUDES, A)",
+        },
+    }
+
+    report = appliance.collections.reports.create(**report_data)
```
john-dupuy: Since the initial report creation is part of setup, I would expect it to be a fixture. What do you think?

cfme/tests/intelligence/reports/test_reports.py
```diff
@@ -149,3 +150,87 @@ def test_new_report_fields(appliance, based_on, request):
...
+            1. Create/Copy a report with secondary (display) filter.
+        testSteps:
+            1. Edit the secondary filter and test if the report was updated.
+        expectedResults:
+            1. Secondary filter must be editable and it must be updated.
+
+    Bugzilla:
+        1565171
+    """
+    report_data = {
```
john-dupuy: _Optional_ I generally prefer `data` to be defined outside of the test case as a module constant, but that is mostly just preference.

john-dupuy:
@valaparthvi can you also run this test case on PRT:
 `cfme/tests/control/test_bugs.py::test_policy_condition_multiple_ors`, it may be impacted by your expression editor changes

cfme/intelligence/reports/reports.py
```diff
@@ -78,8 +79,20 @@ class filter(WaitTab):  # noqa
...
+                if self.parent.primary_record_filter.is_displayed:
+                    self.parent.primary_record_filter.click()
```
kedark3: Question: can this button be disabled? What is expected outcome of click() when its disabled?

cfme/tests/intelligence/reports/test_reports.py
kedark3: I think it should be a fixture as well. We had something similar in v2v, which we moved to fixtures. That helped us avoiding huge `data` dictionaries in test cases as we have above here.

cfme/tests/intelligence/reports/test_reports.py
```diff
@@ -149,3 +150,87 @@ def test_new_report_fields(appliance, based_on, request):
...
+            "secondary_filter": (
+                "fill_field(EVM Custom Attributes : Name, INCLUDES, A);"
+                " select_first_expression;click_or;fill_field"
+                "(EVM Custom Attributes : Region Description, INCLUDES, E)"
+            ),
+        }
+    }
+    report.update(data)
+
+    view = report.create_view(ReportDetailsView)
```
kedark3: **Optional**: I had similar situation in v2v, where I was assuming I am already on the page where I need to be(as that's how workflow we had), but that was assumption. It would be better to use `navigate_to` or maybe assert `view.is_displayed` before performing further operations.

cfme/tests/intelligence/reports/test_reports.py
```diff
@@ -149,3 +150,87 @@ def test_new_report_fields(appliance, based_on, request):
...
+
+    primary_filter = (
+        '( FIND Instance.Guest Applications : Name STARTS WITH "kernel" CHECK COUNT = 1'
+        ' OR FIND Instance.Guest Applications : Name STARTS WITH "kernel" CHECK COUNT = 1 )'
```
kedark3: Question: Why L222,223 both say FIND Instance.Guest Applications : Name STARTS WITH "kernel" CHECK COUNT = 1 ? Should there be something different?

widgetastic_manageiq/expression_editor.py
```diff
@@ -251,8 +263,11 @@ def delete_whole_expression(self):
             self.select_first_expression()
             self.click_remove()
 
-    def read(self):
+    def read(self, complete_expression=False):
         """Returns whole expression as represented visually."""
```
kedark3: **Question**
If we are keeping `complete_expression=False` then this comment should read ""Returns first expression as represented visually"" ?

widgetastic_manageiq/expression_editor.py
valaparthvi: I agree. I was thinking to make the changes you suggested but it would've led to a bigger refactor, probably breaking a few things, which is why I took this approach. But I don't mind working on it. Let me see how I can refactor further.

cfme/intelligence/reports/reports.py
valaparthvi: Okay so `child_widget_accessed` is a method that allows us to perform pre-requisite actions before accessing child-widgets of a Widget. In my case, I need to click on `primary_record_filter` to access Primary Filter and `secondary_display_filter` to access Secondary Filter. `child_widget_accessed` clicks those button for me.

The `widget` passed as an argument here is the child_widget we're trying to accessing.

cfme/tests/intelligence/reports/test_reports.py
valaparthvi: I just wanted more than 1 condition to deal with, so I added the same condition again. It's okay to add the same condition again, but I guess I can change to something else.

cfme/intelligence/reports/reports.py
valaparthvi: Nope, it will never be disabled. If we are not allowed to set a filter for the report, we wouldn't be allowed to navigate to that tab in the first place.

valaparthvi:
Hey @john-dupuy @kedark3, thank you for your reviews. I think I've addressed all of them. Let me know how they look.

widgetastic_manageiq/expression_editor.py
valaparthvi: Hey @john-dupuy, looks like it wasn't that big refactor after all. Hopefully, my current changes wouldn't break other tests. The test you provided passes locally. I'm assuming you had to write a separate function( `select_expression_text` ) to select the expression because of StaleElementReferenceException, I have added a comment to that function about the same. Let me know if that is not the case.

cfme/intelligence/reports/reports.py
john-dupuy: Many thanks for the explanation - makes sense! :+1:

cfme/intelligence/reports/reports.py
```diff
@@ -152,6 +165,18 @@ class ReportDetailsView(CloudIntelReportsView):
...
+        # Keeping `group_title` empty since the summary form has no title
+        report_id = SummaryFormItem("", "ID")
+        title = SummaryFormItem("", "Title")
+        primary_filter = SummaryFormItem("", "Primary (Record) Filter")
+        secondary_filter = SummaryFormItem("", "Secondary (Display) Filter")
+        sort_by = SummaryFormItem("", "Sort By")
+        based_on = SummaryFormItem("", "Based On")
+        user = SummaryFormItem("", "User")
+        group = SummaryFormItem("", "EVM Group")
+        updated_on = SummaryFormItem("", "Updated On")
```
digitronik: Q: why we are not using `SummaryForm` instate of `SummaryformItem`? It will reduce that much declaration.

cfme/intelligence/reports/reports.py
```diff
@@ -152,6 +165,18 @@ class ReportDetailsView(CloudIntelReportsView):
...
+        report_schedule_data = Table('//*[@id="report_info"]/table[1]')
+        report_widgets_data = Table('//*[@id="report_info"]/table[2]')
```
digitronik: Q: hummmmm I hate indexing in locator; can you find some relative locator for table?

cfme/intelligence/reports/reports.py
valaparthvi: Hi @digitronik, there is no group_title to pass to SummaryForm, so I had to individually define each item with SummaryFormItem.

cfme/intelligence/reports/reports.py
valaparthvi: Good that you caught this one @digitronik, xpaths weren't set properly here.
I have now set xpath on the basis of value of second title head in table.

cfme/intelligence/reports/reports.py
digitronik: ok :+1:  other option is parametrization but its fine

valaparthvi: I'm curious about the other implementation, can you elaborate?

digitronik: https://github.com/ManageIQ/integration_tests/blob/master/cfme/automate/buttons.py#L108
here repetition of key value pair.

PR #8521, opened 2019-02-28
[1LP][RFR] Fix minor version update tests.

jaryn:
In general this PR makes minor version update tests working.

Note for 2LP: It is probably futile, but would prefer to not squash the commits on approve to keep the track of what line changed for what reason.

 * fix picking a version previous of the current one for the update
 * In replicated environment, we can have two masters. In such case one of
them (possibly incorrect one) was selected by the tests. This PR
tries to select the the correct rest-api object matching the appliance
in the context.
 * make some update tests use the correct appliance (as using the default one  was wrong)
 * fixes related to browser stealing when tests have to use multiple appliances using the web ui.
 * fixes related to wait_for when checking in login page

This needs #8749

Tested already with: py.test: -v cfme/tests/cli/test_appliance_update.py
{{ py.test: -v cfme/tests/generic_objects/test_instances.py }}

PR #8895, opened 2019-06-03
[1LP][RFR] trying to turn off strict host key checking

izapolsk:
Needed by #8521


izapolsk:
it's possible that the issue is deeper and this fix just solves one of consequences instead of root cause.
I'm still looking into this issue.

mshriver:
Methods pulled from paramiko.SSHClient, modified to check `strict_host_key_checking`.

PR #8902, opened 2019-06-04
[RFR] Add in a wait_for between creating and removing the network security group

john-dupuy:
`test_manage_nsg_group` was failing because it didn't have time to register the event between creating and subsequently deleting the network security group. I have added a wait for that waits for a minute before deleting the network security group, this gives CFME time to receive the event from Azure. 

{{ pytest: --long-running --use-template-cache --use-provider azure cfme/tests/cloud/test_cloud_events.py::test_manage_nsg_group }}

### DIFF
--- a/cfme/tests/cloud/test_cloud_events.py
+++ b/cfme/tests/cloud/test_cloud_events.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 """This module tests only cloud specific events"""
+import time
+
 import pytest
 
 from cfme.cloud.provider.azure import AzureProvider
@@ -78,6 +80,8 @@ def test_manage_nsg_group(appliance, provider, register_event):
 
     # creating and removing network security group
     provider.mgmt.create_netsec_group(nsg_name, resource_group)
+    # wait for a minute before deleting the security group so CFME has time to receive the event
+    time.sleep(60)
     provider.mgmt.remove_netsec_group(nsg_name, resource_group)
 
 


cfme/tests/cloud/test_cloud_events.py
```diff
@@ -78,6 +79,8 @@ def rm_cmp(_, data):
 
     # creating and removing network security group
     provider.mgmt.create_netsec_group(nsg_name, resource_group)
+    # wait for a minute before deleting the security group so CFME has time to receive the event
+    wait_for(lambda: False, silent_failure=True, timeout=60)
```
jaryn: This is effectively a sleep. So
1) Why not to use sleep?
2) is there really no way how to check for the netsec group status?

john-dupuy: @jaryn We can use sleep here, I used `wait_for` instead for consistency with the codebase. In regards to 2, the creation checks that the group is created successfully. However, the problem is not that the group is successfully created, but that CFME needs time to receive the event about whether it was created. Because of the way that `register_event` is written it checks for the events of creation and deletion throughout the whole test, without the wait_for, the netsec group is deleted before the creation event is received, and we receive only the delete event.

mshriver: I'd rather explicitly use `sleep` here - we use wait_for whenever possible, but when there is no trigger condition for us waiting, and its literally time-bound, we should directly use sleep. The in-line comment justifies and explains the use over wait_for.

PR #8904, opened 2019-06-04
[RFR] Add forced streams for BZ's that aren't backported to 5.10 and some for 5.11

john-dupuy:
Adding forced streams to blocker BZs that have been fixed in 5.11 but whose fixes are not backported to 5.10. These tests are failing when they should be skipped. This is impacting `test_policy_simulation_ui` and `test_check_disk_allocation_size_scvmm`.

Also adding some forced_streams for some BZs in 5.11. 

{{ pytest: --use-template-cache --use-provider complete cfme/tests/cloud_infra_common/test_policy_simulation.py::test_policy_simulation_ui cfme/tests/infrastructure/test_scvmm_specific.py::test_check_disk_allocation_size_scvmm }}

### DIFF
--- a/cfme/tests/cloud_infra_common/test_policy_simulation.py
+++ b/cfme/tests/cloud_infra_common/test_policy_simulation.py
@@ -22,13 +22,17 @@ pytestmark = [
 
 
 @pytest.mark.parametrize("navigation", ["provider_vms", "all_vms", "vm_summary"])
-@pytest.mark.uncollectif(
-    lambda navigation: navigation == "provider_vms" and (BZ(1686617).blocks or BZ(1686619).blocks),
-    reason="This test is blocked by the following BZs: {}, {}".format(
-        BZ(1686617).bug_id,
-        BZ(1686619).bug_id
-    )
-)
+@pytest.mark.meta(blockers=[
+    BZ(1717483, forced_streams=['5.11']),
+    BZ(1686617,
+       forced_streams=["5.10"],
+       unblock=lambda navigation: navigation != 'provider_vms'),
+    BZ(1686619,
+       forced_streams=["5.10"],
+       unblock=lambda navigation: navigation != 'provider_vms'),
+    BZ(1717539,
+       unblock=lambda navigation: navigation != 'provider_vms')
+])
 def test_policy_simulation_ui(appliance, provider, navigation):
     """
     Bugzilla:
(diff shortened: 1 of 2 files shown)


cfme/tests/cloud_infra_common/test_policy_simulation.py
```diff
@@ -22,11 +22,18 @@
 
 
 @pytest.mark.parametrize("navigation", ["provider_vms", "all_vms", "vm_summary"])
+@pytest.mark.meta(blockers=[BZ(1717483, forced_streams=["5.11"])])
 @pytest.mark.uncollectif(
-    lambda navigation: navigation == "provider_vms" and (BZ(1686617).blocks or BZ(1686619).blocks),
+    lambda navigation: (
+        navigation == "provider_vms" and
+        (BZ(1686617, forced_streams=["5.10"]).blocks or
+         BZ(1686619, forced_streams=["5.10"]).blocks or
```
lina-nikiforova: Some questions here.

I don't see any clones of these two BZs for 5.10, do you think the fix will be backported? 

For 5.11 appliance forced_streams make no difference
image

john-dupuy: @lina-nikiforova I don't think the BZs will be backported, which is why I've included the forced streams here. Regarding `1717539`, this required the `forced_streams` at the time I wrote it, I will go ahead and remove it though.

lina-nikiforova: @john-dupuy in that case, why not to uncollect for the 5.10 stream instead of checking for the conditions that will always be True?

john-dupuy: @lina-nikiforova That is because of the first condition, we still want to run this test case if `navigation != "provider_vms"`. But you're right, it doesn't make sense to have the condition that always returns true, I will put in a check for the 5.10 stream.

cfme/tests/cloud_infra_common/test_policy_simulation.py
```diff
@@ -22,11 +22,17 @@
...
-    lambda navigation: navigation == "provider_vms" and (BZ(1686617).blocks or BZ(1686619).blocks),
+    lambda navigation, appliance: (
+        navigation == "provider_vms" and
+        (appliance.version < "5.11" or
+         BZ(1717539).blocks)
+    ),
     reason="This test is blocked by the following BZs: {}, {}".format(
         BZ(1686617).bug_id,
-        BZ(1686619).bug_id
+        BZ(1686619).bug_id,
```
mshriver: This is no longer true for this uncollect statement - these BZ's are no longer included in the lambda.

PR #8905, opened 2019-06-04
[1LP][RFR] Create a test case to test policy conditions with multiple OR statements

john-dupuy:
Creating a test case for the following customer BZ https://bugzilla.redhat.com/show_bug.cgi?id=1711352

{{ pytest: --use-template-cache --use-provider vsphere65-nested --long-running cfme/tests/control/test_bugs.py::test_policy_condition_multiple_ors }}

### DIFF
--- a/cfme/tests/control/test_bugs.py
+++ b/cfme/tests/control/test_bugs.py
@@ -197,7 +197,9 @@ def compliance_condition(appliance, virtualcenter_provider):
     try:
         vm_name = virtualcenter_provider.data["cap_and_util"]["capandu_vm"]
     except KeyError:
-        pytest.skip("No capandu_vm exists on virtualcenter_provider")
+        pytest.skip("Missing 'cap_and_util' field in {} provider data.".format(
+            virtualcenter_provider.key)
+        )
     expression = (
         "fill_field(VM and Instance : Name, =, {}); "
         "select_expression_text; "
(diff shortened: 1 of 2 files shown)


cfme/tests/control/test_bugs.py
```diff
@@ -192,6 +192,41 @@ def action_for_testing(appliance):
     action_.delete()
 
 
+@pytest.fixture
+def compliance_condition(appliance):
+    expression = (
+        "fill_field(VM and Instance : Name, =, cu-24x7); "
```
mshriver: We can't make this name assumption here, it needs to be pulled from the yamls. The fixture should be dependent on appliance and provider.

cfme/tests/control/test_bugs.py
```diff
@@ -515,3 +556,59 @@ def test_edit_action_buttons(action_for_testing):
...
+    vm_name = virtualcenter_provider.data["cap_and_util"]["capandu_vm"]
+    # check that it exists on provider
+    if not virtualcenter_provider.mgmt.does_vm_exist(vm_name):
+        pytest.skip("No capandu_vm available on virtualcenter_provider of name {}".format(vm_name))
+
+    vms = [all_vms.pop(all_vm_names.index(vm_name))]
+
+    # add some other vms to the list to run the policy simulation against
+    # we use this ugly logic in case there are not that many vms on the provider
+    if len(all_vms) > 3:
```
mshriver: slightly less messy:

if len(all_vms) == 1:
    vms.extend(all_vms[0])
else:
    vms.extend(all_vms[0:len(all_vms)%5]
```

mshriver: Even better solution found via discussion, since this `%` could result in zeros, `min(randint(1, len(all_vms)), 5)`

PR #8909, opened 2019-06-05
[1LP][RFR] fixed test_vm_filter_with_user_input

lina-nikiforova:
1. Unfortunately, this test was uncollected because of `required_fields=['large']` - we don't have such providers in yamls, therefore BZ 1715550 was missed.
So I removed it.

2. After that, there was scope mismatch, so I also fixed it.

3. The last problem was with the test itself. As described in the bug, nothing happens when 'Apply" is clicked, therefore this test will never fail as the filtering doesn't occur. I updated the test to check that after filtering the number of VMs is at least less than originally and now the test is failing.

4. Added BZ to block this test.

{{ pytest: -v cfme/tests/infrastructure/test_advanced_search_vms.py}}

### DIFF
--- a/cfme/tests/infrastructure/test_advanced_search_vms.py
+++ b/cfme/tests/infrastructure/test_advanced_search_vms.py
@@ -10,15 +10,16 @@ from widgetastic.exceptions import NoSuchElementException
 from cfme.infrastructure.provider import InfraProvider
 from cfme.markers.env_markers.provider import ONE
 from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.blockers import BZ
 
 pytestmark = [
     pytest.mark.tier(3),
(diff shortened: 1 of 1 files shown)


lina-nikiforova:
The failure of the test is as expected 
```
>       assert vms_after < vms_before
E       assert 23 < 23
```
But it was not blocked for neither 5.10 nor 5.11.

Adding forced_streams.

izapolsk:
since you change scope of common fixtures, I updated your prt command to run all tests in order to avoid scope mismatch issues. PRT has been restarted as well.

lina-nikiforova:
@izapolsk thanks. Noted for future PRTs.

lina-nikiforova:
PRT failed but no ScopeMismatch. Other errors that are not related to this change.

mshriver:
Agreed that the failures in PRT don't look directly related to the fixes here. But these fixes may have uncovered some bad behavior in MIQ :)

PR #8916, opened 2019-06-06
[1LP][RFR][NOTEST] Migration requirement and metadata

Yadnyawalkya:
PRT results are not relevant since we have
- [x] Removed RHV subcomponent  from tests
- [x] Update tests with v2v requirement
- [x] Updated polarion metadata


Yadnyawalkya:
Thank you @john-dupuy @sshveta @mshriver

PR #8918, opened 2019-06-06
[1LP][RFR] Enhance LogValidator for pattern match

digitronik:

Purpose or Intent
=================

- The current implementation of LogValidator is just to `search` some pattern in the log which is not sufficient for the custom button use case
- In the custom button after button execute request are fired in automation log and in our test, we are checking those. It's not a single request may be multiple.

- This checks can be part of `LogValidator`

- Shriver: removed from pytest, collecting more uses of LogValidator because of the scope of changes.
-k "test_custom_button_automate_infra_obj or test_custom_button_dialog_infra_obj or test_codename_in_log"

{{pytest: cfme/tests/test_appliance.py cfme/tests/cli/test_appliance_console.py cfme/tests/automate/test_method.py cfme/tests/services/test_catalog_item.py cfme/tests/control/test_actions.py cfme/tests/automate/custom_button/test_infra_objects.py --long-running }}

### DIFF
--- a/cfme/utils/log_validator.py
+++ b/cfme/utils/log_validator.py
@@ -47,18 +47,19 @@ class LogValidator(object):
         self.matched_patterns = kwargs.pop('matched_patterns', [])
 
         self._remote_file_tail = SSHTail(remote_filename, **kwargs)
-        self.matches = {}
+        self.matches = {key: 0 for key in self.matched_patterns}
 
     def fix_before_start(self):
+        """Start monitoring log before action"""
         self._remote_file_tail.set_initial_file_end()
 
-    def validate_logs(self):
+    def _validation(self):
         for line in self._remote_file_tail:
             if self._check_skip_logs(line):
                 continue
             self._check_fail_logs(line)
             self._check_match_logs(line)
-        self._verify_match_logs()
+        logger.info("Matches found: {}".format(self.matches))
 
     def _check_skip_logs(self, line):
         for pattern in self.skip_patterns:
@@ -77,12 +78,22 @@ class LogValidator(object):
         for pattern in self.matched_patterns:
             if re.search(pattern, line):
                 logger.info('Expected pattern {} was matched on line {}'.format(pattern, line))
-                self.matches[pattern] = True
+                self.matches[pattern] = self.matches[pattern] + 1
 
-    def _verify_match_logs(self):
-        for pattern in self.matched_patterns:
-            if pattern not in self.matches:
-                pytest.fail('Expected pattern {} did not match'.format(pattern))
+    def validate_logs(self):
+        """Validate log pattern"""
+        self._validation()
+        for pattern, count in self.matches.items():
+            if count == 0:
+                pytest.fail(
+                    'Expected pattern {} did not match; match count {}'.format(pattern, count)
+                )
+
+    @property
+    def patterns_match_count(self):
+        """It will return pattern match count dictionary"""
+        self._validation()
+        return self.matches
 
     def wait_for_log_validation(
             self, delay=5, num_sec=180, message="waiting for log validation", **kwargs
@@ -101,3 +112,25 @@ class LogValidator(object):
             except Failed:
                 return False
         wait_for(validate, delay=delay, num_sec=num_sec, message=message, **kwargs)
+
+    def wait_for_expected_patterns_match_count(
+        self,
+        expected_match,
+        timeout=180,
+        delay=5,
+        message="Check for expected match count",
+        **kwargs
+    ):
+        """ wait for expected match count found in respective log file.
+
+        Args:
+            expected_match (dir): expected patterns match count
+            timeout (int): timeout
+            delay (int): delay time
+            message (str): wait_for message
+        """
+
+        def _match():
+            return set(self.patterns_match_count) == set(expected_match)
+
+        wait_for(_match, timeout=timeout, message=message, delay=delay, **kwargs)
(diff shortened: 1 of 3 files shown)


dajoRH:
*CFME QE Bot*

digitronik:
511 one test failure not related to changes. I will take of view the latter.

cfme/utils/log_validator.py
```diff
@@ -77,12 +78,22 @@ def _check_match_logs(self, line):
...
+        """Validate log pattern"""
+        self._validation()
+        for pattern, count in self.matches.items():
+            if count == 0:
+                pytest.fail(
+                    'Expected pattern {} did not match; match count {}'.format(pattern, count)
+                )
+
+    @property
+    def patterns_match_count(self):
```
mshriver: I think the naming is a bit confusing here, it does not read clearly at all for me, but maybe its just me.

Consider changing the naming pattern/scheme here to be more clear, maybe using `number_of` or `count_of` in some way?

```
@property
def count_of_pattern_matches

def wait_for_pattern_count_match()
```

I think the docblocks need more examples, because its not obvious what is returned in the dictionary.

cfme/utils/log_validator.py
```diff
@@ -77,12 +78,22 @@ def _check_match_logs(self, line):
...
-    def _verify_match_logs(self):
-        for pattern in self.matched_patterns:
-            if pattern not in self.matches:
-                pytest.fail('Expected pattern {} did not match'.format(pattern))
+    def validate_logs(self):
+        """Validate log pattern"""
+        self._validation()
+        for pattern, count in self.matches.items():
+            if count == 0:
+                pytest.fail(
```
mshriver: I think this is a good opporunity to 'fix' this behavior, and not use pytest outside of a pytest context. I think we should instead be raising an `AssertionError` here

cfme/utils/log_validator.py
```diff
@@ -101,3 +112,25 @@ def validate():
...
+        expected_match,
+        timeout=180,
+        delay=5,
+        message="Check for expected match count",
+        **kwargs
+    ):
+        """ wait for expected match count found in respective log file.
+
+        Args:
+            expected_match (dir): expected patterns match count
```
mshriver: `dir` for expected match?  This docblock needs better descriptions here, as well as examples below.

Something like `'matches_and_counts (dict): patterns to match as dict keys, and their expected count as values'`

This could easily be a list of tuples as well, which might be a better way to structure such data.

Ultimately, since this is expecting the exact same data format as the LogValidator property, its a convenience method that I'm not sure adds a lot of value.

Did you consider actually changing how LogValidator.matched_patterns actually is stored, so that LogValidator.validate_logs is inherently validating the match and count?

It looks to me like we could default the `count` of a `matched_patterns` item to None (preserving current uses of LogValidator), but otherwise accept tuples of `(pattern, count)` in the `matched_patterns` list.  When the `count` is the default, `None`, we consider any number of matches a passing validation.

This way, the behavior of validation including the count would be inherent to a validation call, and the use case you've implemented here would simply be covered by `wait_for_log_validation`.

Then instead of separating

cfme/utils/log_validator.py
mshriver: Considering my comment below about an alternative design, I think this is a wrapper around something that should be a property anyway.

I think instead of this extra property, `LogValidator.matches` should be a property, which calls `self._validation` and then returns `self._matches`, which would be updated in `_check_match_logs` (instead of `self.matches` like now).

cfme/utils/log_validator.py
jawatts: :+1: Here

digitronik: If we like to avoid `pytest context` in such utilities. As per me, `AssertionError` is a bad idea. what do you think Mike? if method name should `validate` only with returning `bool` (True/ False).

digitronik:
@mshriver as per your comment its good time to refactor it. I would like to refactor more things

1. `fix_before_start` : its not providing clear meaning. many of us confused what is the purpose of it. Let's replace this with just `start_monitoring`

2. replace `validate_logs` with `validate` only with wait inbuild. No need of separate `wait_for_log_validation`. if user need it free pass just  validate(wait="60s")

3. `patterns_matched` property will return dict of matched pattern count. No need to add wait_for_patterns_matched. User will add own wait in tests.

cfme/utils/log_validator.py
digitronik: wait for matches not a part of LogValidator now. User will implement as per need.

cfme/utils/log_validator.py
digitronik: @mshriver now
`LogValidator.matches` is property... which used while validation. and `_matches` is for count collection.

digitronik:
> @mshriver as per your comment its good time to refactor it. I would like to refactor more things
> 
> 1. `fix_before_start` : its not providing clear meaning. many of us confused what is the purpose of it. Let's replace this with just `start_monitoring`
> 2. replace `validate_logs` with `validate` only with wait inbuild. No need of separate `wait_for_log_validation`. if user need it free pass just  validate(wait="60s")
> 3. `patterns_matched` property will return dict of matched pattern count. No need to add wait_for_patterns_matched. User will add own wait in tests.

First, two will be part of next sprint PR; which will refactor some methods.

In this PR we have one extra method which will return `dict` of `matches` (3)

cfme/utils/log_validator.py
mshriver: Discussed scope and needs offline, keeping this PR's scope reduced and will introduce more changes to LogValidator later.

cfme/utils/log_validator.py
mshriver: Leaving this behavior as-is for now, will update in further enhancement PR for LogValidator where callers are updated.

PR #8926, opened 2019-06-07
[RFR] Add a default severity for CFME 5.11

john-dupuy:
Purpose or Intent
=================
This BZ https://bugzilla.redhat.com/show_bug.cgi?id=1653427 (and this PR https://github.com/ManageIQ/manageiq-ui-classic/pull/5326) added alert severity back into CFME 5.11. This PR is setting the default severity for created alerts to "Info". "Severity" is a required field, so this is necessary for creating an alert in CFME 5.11. 

{{ pytest: --use-template-cache --long-running cfme/tests/control/test_basic.py::test_alert_crud }}

### DIFF
--- a/cfme/control/explorer/alerts.py
+++ b/cfme/control/explorer/alerts.py
@@ -25,7 +25,6 @@ from cfme.utils.appliance.implementations.ui import navigator
 from cfme.utils.log import logger
 from cfme.utils.pretty import Pretty
 from cfme.utils.update import Updateable
-from cfme.utils.version import current_version
 from widgetastic_manageiq import AlertEmail
 from widgetastic_manageiq import MonitorStatusCard
 from widgetastic_manageiq import SNMPForm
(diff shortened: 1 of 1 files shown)


cfme/control/explorer/alerts.py
```diff
@@ -376,6 +377,9 @@ def create(self, description, severity=None, active=None, based_on=None, evaluat
         view = navigate_to(self, "Add")
         if driving_event is None and evaluate is None:
             driving_event = view.driving_event.all_options[1].text
+        if severity is None and current_version() >= "5.11.0.7":
```
mshriver: We actually want to steer clear of using `current_version`, because it depends on 'magic' appliance creation.

We have direct access to the appliance object linked to our `AlertCollection`, and should use it instead. `self.appliance.version`

john-dupuy: Done

PR #8929, opened 2019-06-10
[1LP][RFR] added fileserver for remote file management

digitronik:
Purpose or Intent
=================
We need on storage for collecting customer dbs, Datastores, dialogs, Reports etc...

Needs: https://gitlab.cee.redhat.com/cfme-qe/cfme-qe-yamls/merge_requests/785


dajoRH:
Lint report for commit 0fabfeda11b2bcd87de9fda07340013133338d78:

`cfme/utils/fileserver.py`:
- :red_circle: Line 94:19: **F821** *undefined name 'FileNotFoundError'*
- :red_circle: Line 116:19: **F821** *undefined name 'FileExistsError'*
- :red_circle: Line 131:19: **F821** *undefined name 'FileNotFoundError'*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/intelligence/reports/test_import_export_reports_widgets.py
```diff
@@ -4,14 +4,21 @@
 from cfme import test_requirements
 from cfme.intelligence.reports.reports import ImportExportCustomReportsView
 from cfme.intelligence.reports.widgets import ImportExportWidgetsCommitView
-from cfme.utils.path import data_path
+from cfme.utils.fileserver import FileServer
```
jaryn: I don't like that you call this thing a FileServer while I think it really is a some kind of FTP client.

digitronik: I'm really bad at giving a name... I want one FileServer where we can store files and download those while running tests. Yes, its using FTP client. but FileServer I like at that time. please give me some suggestions.

jaryn: On second look I am not sure what this is and why should it exist. It seems that only non-dead-code method in the `FileServer` class is the `__init__` and download. So I am not sure why not to just use the client that there is in the same codebase.

It seems to me this could be named perhaps FTPClientWrapper, but as I told above, I don't know why does it exist on first place.

digitronik: aaaahhh `ftp client` need to send object of the file for `download` and `upload` even some method which I want to restrict. In `ftp.py` you will get the client but the method is more general like we are wrapping python core `ftplib` :). I don't know why we use `send_command` each time. I would like to put that as a client (**As it's more generic**) so created a new file and put restricted implementation you can refer PR header. why we need this and why not `ftp.py`

cfme/utils/fileserver.py
```diff
@@ -0,0 +1,190 @@
...
+        self.user = user if user else credentials[ftp_data.credentials]["username"]
+        self.passwd = passwd if passwd else credentials[ftp_data.credentials]["password"]
+        self.entrypoint = entrypoint or ftp_data.entrypoint
+
+        self.client = FTPClient(self.host, self.user, self.passwd, time_diff=False)
+        self.cwd = os.path.join(self.entrypoint, "miq", self.entity if entity else "")
+        self.client.cwd(self.cwd)
+
+    @property
+    def _entities(self):
```
sshveta: Required : Add docblock explaining the logic a bit

cfme/utils/fileserver.py
```diff
@@ -0,0 +1,190 @@
...
+            raise FTPException("{} not found".format(name))
+
+    def rm(self, name):
+        """ Remove or Delete remote FTP file
+        Arg:
+            name: name of remote file
+        Returns:
+            Success of the action
+        """
+        return self.client.dele(name)
```
sshveta: Required : Change this to return self.client.delete(name)

cfme/utils/fileserver.py
```diff
@@ -0,0 +1,190 @@
...
+    def mkdir(self, name):
+        """ Create remote FTP directory
+        Arg:
+            name: name of remote directory
+        Returns:
+            Success of the action
+        """
+        return self.client.mkd(name)
+
+    def rmd(self, name):
```
sshveta: How is this different from def rm () at line 97 ? 
We can use only one command and have one method ?

cfme/tests/intelligence/reports/test_import_export_reports_widgets.py
sshveta: From what I understand , it is using Ftp.py and adding some more methods on top to be used for downloading /uploading files . so yeah FTPClientWrapper or FTPwarapper seems more appropriate :+1:

cfme/utils/fileserver.py
digitronik: now it will take `dele` only as we are making a wrapper

cfme/tests/intelligence/reports/test_import_export_reports_widgets.py
digitronik: oK... I change complete implementation... @jaryn @sshveta have a look... I try to avoid this but nice to use old utility.

digitronik:
> The wrapper class is an acceptable change, but its not acceptable to me for us to remove import test yamls from the data directory, depriving upstream of test resources.

Added Entities in data yaml.
https://gitlab.cee.redhat.com/cfme-qe/cfme-qe-yamls/merge_requests/797

> Its great that 'our' team has an FTP server to host these files on, but it has its downsides too, as we lose version control of the data used in test. Without explicitly logging the content of those files in the test artifacts, we're unaware of what was imported when a test fails.

It will depend on users, I added some basic logging for providing more info. About version control; I think we are just putting testing files in `/data` (It's ok for .yamls but not good for .db, .zip, etc...) which can be part of FTP and user can monitor own files update as per test requirements.

> The fixtures should be written to look for an FTP server defined in the yaml, and should look for standardized keys in the yaml for where to locate files in the directory structure. If the FTP server isn't defined, they should use the local `/data` files.
> 
:+1: for now, I took an existing test for testing. I added changes in current fixtures which will first check `ftp` else path will pick up from `/data`.

In other implementation, I am going to add some `global` fixtures for entities; which will take the file name and yield path.

> There is too much coupling between FTPClientWrapper and the directory structure, in my opinion. I think it should not be hardcoding expected path components and should be getting those path components from yaml maps.

This done in [MR](https://gitlab.cee.redhat.com/cfme-qe/cfme-qe-yamls/merge_requests/797)
I'm waiting for PRT without MR (without FTP server) and we will restart PRT with MR :+1:

cfme/tests/intelligence/reports/test_import_export_reports_widgets.py
```diff
@@ -4,14 +4,26 @@
...
 
 @pytest.fixture()
 def widget_file(appliance):
-    file_path = data_path.join("ui/intelligence/import_widget.yaml").realpath().strpath
+
+    try:
+        fs = FTPClientWrapper(cfme_data.ftpserver.entities.reports)
+        file_path = fs.download("import_widget.yaml", "/tmp/import_widget.yaml")
+    except (FTPException, AttributeError) as e:
+        logger.warning("'import_widget.yaml' download failed with FTP: %s", e)
```
mshriver: Please use logger.exception here so we get better exception context/traceback automatically.

cfme/utils/ftp.py
```diff
@@ -172,10 +178,10 @@ def path(self):
             whole path for this file
 
         """
-        if self.parent_dir:
+        if isinstance(self.parent_dir, FTPDirectory):
             return self.parent_dir.path + self.name
         else:
-            return self.name
+            return os.path.join(self.parent_dir, self.name)
```
mshriver: Is this doing the same thing as L182 above, joining the parent and current name?

Now is a good time to change the `+` concatenation on L182 as well, and use `'{}{}'.format(self.parent_dir.path, self.name)`

cfme/tests/intelligence/reports/test_import_export_reports_widgets.py
mshriver: Pushed a commit for this, and moved the file names to variables since they're repeated.

cfme/utils/ftp.py
```diff
@@ -64,10 +69,7 @@ def path(self):
             whole path for this directory
 
         """
-        if self.parent_dir:
-            return self.parent_dir.path + self.name + "/"
-        else:
-            return self.name
+        return os.path.join(self.parent_dir.path if self.parent_dir else "", self.name)
```
mshriver: :tada:

PR #8930, opened 2019-06-10
[1LP][RFR] Added key pair tests for ec2

mmojzis:
Added key pair tests for ec2
It's failing for 5.11 - when I fix all 5.10 problems with automation I will start fixing 5.11.
Also it was failing in 5.10 when calling keypair.exists so I overriden that method and added wait_for which fixes it. I know this is dumb fix, but didn't find any better way. So if you know how to fix it better then please tell me. Thanks.

### DIFF
--- a/cfme/cloud/keypairs.py
+++ b/cfme/cloud/keypairs.py
@@ -2,17 +2,12 @@
 import attr
 from navmazing import NavigateToAttribute
 from navmazing import NavigateToSibling
-from navmazing import NavigationDestinationNotFound
 from widgetastic.exceptions import MoveTargetOutOfBoundsException
-from widgetastic.exceptions import NoSuchElementException
 from widgetastic.widget import View
 from widgetastic_patternfly import BootstrapNav
 from widgetastic_patternfly import BreadCrumb
 from widgetastic_patternfly import Button
-from widgetastic_patternfly import CandidateNotFound
 from widgetastic_patternfly import Dropdown
-
-
 
 from cfme.base.ui import BaseLoggedInPage
 from cfme.common import Taggable
@@ -23,7 +18,6 @@ from cfme.modeling.base import BaseEntity
 from cfme.utils.appliance.implementations.ui import CFMENavigateStep
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.appliance.implementations.ui import navigator
-from cfme.utils.wait import TimedOutError
 from cfme.utils.wait import wait_for
 from widgetastic_manageiq import Accordion
 from widgetastic_manageiq import BaseEntitiesView
@@ -35,7 +29,6 @@ from widgetastic_manageiq import Search
 from widgetastic_manageiq import SummaryTable
 from widgetastic_manageiq import Text
 from widgetastic_manageiq import TextInput
-
 
 
 class KeyPairToolbar(View):
@@ -192,24 +185,6 @@ class KeyPair(BaseEntity, Taggable):
         view.toolbar.configuration.item_select('Download private key')
         view.flash.assert_no_error()
 
-    @property
-    def exists(self):
-        try:
-            wait_for(lambda: navigate_to(self, "Details"), num_sec=10, delay=2,
-                     handle_exception=True)
-        except (
-            CandidateNotFound,
-            ItemNotFound,
-            KeyPairNotFound,
-            NameError,
-            NavigationDestinationNotFound,
-            NoSuchElementException,
-            TimedOutError,
-        ):
-            return False
-        else:
-            return True
-
 
 @attr.s
 class KeyPairCollection(BaseCollection):
@@ -245,7 +220,10 @@ class KeyPairCollection(BaseCollection):
                  fail_func=view.flush_widget_cache, handle_exception=True)
         assert view.is_displayed
         view.flash.assert_success_message(flash_message)
-        return self.instantiate(name, provider, public_key=public_key)
+        keypair = self.instantiate(name, provider, public_key=public_key)
+        if not cancel:
+            wait_for(lambda: keypair.exists, delay=2, timeout=240, fail_func=self.browser.refresh)
+        return keypair
 
     def all(self):
         view = navigate_to(self, 'All')
--- a/cfme/cloud/keypairs.py
+++ b/cfme/cloud/keypairs.py
@@ -2,12 +2,17 @@
 import attr
 from navmazing import NavigateToAttribute
 from navmazing import NavigateToSibling
+from navmazing import NavigationDestinationNotFound
 from widgetastic.exceptions import MoveTargetOutOfBoundsException
+from widgetastic.exceptions import NoSuchElementException
 from widgetastic.widget import View
 from widgetastic_patternfly import BootstrapNav
 from widgetastic_patternfly import BreadCrumb
 from widgetastic_patternfly import Button
+from widgetastic_patternfly import CandidateNotFound
 from widgetastic_patternfly import Dropdown
+
+
 
 from cfme.base.ui import BaseLoggedInPage
 from cfme.common import Taggable
@@ -18,6 +23,7 @@ from cfme.modeling.base import BaseEntity
 from cfme.utils.appliance.implementations.ui import CFMENavigateStep
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.appliance.implementations.ui import navigator
+from cfme.utils.wait import TimedOutError
 from cfme.utils.wait import wait_for
 from widgetastic_manageiq import Accordion
 from widgetastic_manageiq import BaseEntitiesView
@@ -29,6 +35,7 @@ from widgetastic_manageiq import Search
 from widgetastic_manageiq import SummaryTable
 from widgetastic_manageiq import Text
 from widgetastic_manageiq import TextInput
+
 
 
 class KeyPairToolbar(View):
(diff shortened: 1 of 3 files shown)


dajoRH:
Lint report for commit d8513b2d4874c74caa24ce251efdf06823ed2c6c:

`cfme/cloud/keypairs.py`:
- :bangbang: Line 17:1: **E303** *too many blank lines (3)*
- :bangbang: Line 41:1: **E303** *too many blank lines (3)*

**No commit flaws detected**.

*CFME QE Bot*

cfme/cloud/keypairs.py
```diff
@@ -185,6 +189,24 @@ def download_private_key(self):
         view.toolbar.configuration.item_select('Download private key')
         view.flash.assert_no_error()
 
+    @property
+    def exists(self):
+        try:
+            wait_for(lambda: navigate_to(self, "Details"), num_sec=10, delay=2,
```
digitronik: I think this should only check entity exists or not in the model page.
You can wait for the entity in the test. Maybe you are facing this problem with `create`?
If yes I think you can take an extra argument in the `create` method and wait for entity create there.

cfme/tests/cloud/test_keypairs.py
```diff
@@ -73,41 +78,55 @@ def test_keypair_crud_with_key(provider, appliance):
...
+@pytest.mark.tier(3)
+def test_keypair_create_validation(provider, appliance):
+    """ This will test validating the new of name of the key pair
+    Polarion:
+        assignee: mmojzis
+        casecomponent: WebUI
+        caseimportance: low
+        initialEstimate: 1/4h
+        testSteps:
+            1) Try to add key pair with empty name.
```
digitronik: please fallow formate `1.`

cfme/cloud/keypairs.py
```diff
@@ -185,6 +189,24 @@ def download_private_key(self):
...
+            ItemNotFound,
+            KeyPairNotFound,
+            NameError,
+            NavigationDestinationNotFound,
+            NoSuchElementException,
+            TimedOutError,
+        ):
+            return False
+        else:
+            return True
```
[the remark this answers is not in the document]
sshveta: Example : https://github.com/ManageIQ/integration_tests/blob/master/cfme/infrastructure/host.py#L230

cfme/tests/cloud/test_keypairs.py
```diff
@@ -73,41 +78,55 @@ def test_keypair_crud_with_key(provider, appliance):
...
+    """
+    keypair_collection = appliance.collections.cloud_keypairs
+    view = navigate_to(keypair_collection, 'Add')
+    view.fill({'provider': provider})
+    assert not view.form.add.active
 
 
+@pytest.mark.uncollectif(lambda provider: provider.one_of(EC2Provider))
```
sshveta: Question : Why are you uncollecting for EC2?

dajoRH:
*CFME QE Bot*

cfme/cloud/keypairs.py
mmojzis: Fixed by adding wait_for to create method.

cfme/tests/cloud/test_keypairs.py
mmojzis: Fixed

cfme/tests/cloud/test_keypairs.py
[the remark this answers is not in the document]
mmojzis: I changed uncollection to override provider to use only Openstack.
This test won't work for EC2 because of bug.
Also this test doesn't need to test multiple providers as it actually is testing something that doesn't belong to provider, but keypair cannot be added when there is no cloud provider.

cfme/cloud/keypairs.py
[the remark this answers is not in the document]
mmojzis: I fixed it by moving wait_for to create. Exists method was problematic only when using it after creation of keypair.

dajoRH:
Lint report for commit b607154c0e208281da14c89db71eee0c080b1ed4:

`cfme/cloud/keypairs.py`:
- :red_circle: Line 5:1: **F401** *'navmazing.NavigationDestinationNotFound' imported but unused*
- :red_circle: Line 7:1: **F401** *'widgetastic.exceptions.NoSuchElementException' imported but unused*
- :red_circle: Line 12:1: **F401** *'widgetastic_patternfly.CandidateNotFound' imported but unused*
- :red_circle: Line 24:1: **F401** *'cfme.utils.wait.TimedOutError' imported but unused*

`cfme/tests/cloud/test_keypairs.py`:
- :red_circle: Line 5:1: **F401** *'cfme.cloud.keypairs.KeyPairAllView' imported but unused*

Hmmm


**No commit flaws detected**.

*CFME QE Bot*

digitronik:
511 failures not related to code changes its `TagPageView` problem. It will solve as we fix this view.

cfme/tests/cloud/test_keypairs.py
```diff
@@ -73,41 +78,54 @@ def test_keypair_crud_with_key(provider, appliance):
...
         cancel=True
     )
-
     assert not keypair.exists
 
 
+@pytest.mark.provider([EC2Provider], override=True, scope="module", selector=ONE_PER_TYPE)
+@pytest.mark.tier(3)
+def test_keypair_create_validation(provider, appliance):
+    """ This will test validating the new of name of the key pair
```
jawatts: @mmojzis Could this be a bit more specific here? This test is really testing the validation that a key pair cannot be added with an empty name.

cfme/tests/cloud/test_keypairs.py
```diff
@@ -73,41 +78,54 @@ def test_keypair_crud_with_key(provider, appliance):
...
+        casecomponent: WebUI
+        caseimportance: low
+        initialEstimate: 1/4h
+        testSteps:
+            1. Try to add key pair with empty name.
+    """
+    keypair_collection = appliance.collections.cloud_keypairs
+    view = navigate_to(keypair_collection, 'Add')
+    view.fill({'provider': provider})
+    assert not view.form.add.active
```
jawatts: **Question** I think this test could be extended a bit more, perhaps with invalid key pair names? Thoughts?

mmojzis: Extended for invalid key pair name but that could be tested only on OpenStack. Also added test that checks public key.

cfme/tests/cloud/test_keypairs.py
mmojzis: Done.

cfme/tests/cloud/test_keypairs.py
```diff
@@ -73,41 +79,93 @@ def test_keypair_crud_with_key(provider, appliance):
...
+               'provider': provider.name,
+               'public_key': invalid_key})
+    view.form.add.click()
+    view = view.browser.create_view(KeyPairAllView)
+    if provider.one_of(EC2Provider):
+        view.flash.assert_message('Unable to create Key Pair "{}": Key is not in valid OpenSSH '
+                                  'public key format'.format(keypair_name))
+    elif provider.one_of(OpenStackProvider):
+        view.flash.assert_message('Unable to create Key Pair "{}": Keypair data is invalid: failed '
+                                  'to generate fingerprint'.format(keypair_name))
```
jawatts: @mmojzis I think we should check that the keypair also does not exist. So verify that we get the error message and check that the keypair was not added. What do you think?

mmojzis: Well, that's not so easy. Keypair is not created immediately so I would have to create wait_for for at least a minute to check whether this keypair is not in keypair array. I don't think that makes much sense. This error response comes directly from provider so if keypair is created then it's problem on provider side and not in CFME.

dajoRH:
*CFME QE Bot*

cfme/tests/cloud/test_keypairs.py
jawatts: @mmojzis  Does CFME create a keypair record in CFME when a request is made? Does it delete when the request fails? I think it is worth validating that there is no records in vmdb and that nothing appears in CFME.

mmojzis: No, that's why I have to wait until it's created. It calls provider and keypair is created on provider side. In CFME it's created by targeted refresh AFAIK. Provider sends data to CFME that it has a new record.
When request fails nothing is created as it's called directly from CFME UI to provider.

jawatts: Thanks @mmojzis

mmojzis: @jawatts Rebased...But it started failing on one test for no reason....

jawatts: @mmojzis looks like it failed here:
```[KeyPairDetailsView/toolbar/configuration]: Selecting 'Download private key' (/usr/lib/python2.7/site-packages/widgetastic_patternfly/__init__.py:1684)```

So there is no "Download private key" button on the details page. See this screenshot:
http://10.16.45.124/aRzZLu1b/artifacts/cfme/tests/cloud/test_keypairs.py/test_download_private_key[openstack-13]/filedump-exception_screenshot.png

Comment (jawatts) [reply cfme/tests/cloud/test_keypairs.py]: @mmojzis looks like it is under the `Configuration` drop down

Comment (jawatts) [reply cfme/tests/cloud/test_keypairs.py]: @mmojzis It looks like it is because the "Download private key" selection is disabled. 
```
In[19]: view.toolbar.configuration.item_enabled('Download private key')
Out[20]: False
```

mmojzis: @jawatts Thanks for helping me identifying that issue. I don't know how, but the problem was actually fixed by deleting some keypairs and it passed now.

PR #8932, opened 2019-06-10
[1LP][RFR]Add manual tests for RFEs

valaparthvi:


cfme/tests/intelligence/reports/test_reports_manual.py
```diff
@@ -360,3 +360,56 @@ def test_reports_sort_column():
...
+
+@pytest.mark.manual
+@test_requirements.report
+@pytest.mark.tier(0)
+def test_report_new_tags():
+    """
+    This test case tests report creation with new tags.
+
+    # TODO: New tags are often added as an RFE, parametrize and design the test
+    # such that it can accommodate any new tags added in the future.
```
digitronik: optional: I don't think we `#` in a docstring

valaparthvi: Oops, my bad :sweat_smile:
Will remove that.

PR #8936, opened 2019-06-11
[1LP][RFR] Update template upload for rhevm, name database disk

mshriver:
Depends on wrapanapi: ~~https://github.com/ManageIQ/wrapanapi/pull/390~~

### DIFF
--- a/cfme/utils/template/rhevm.py
+++ b/cfme/utils/template/rhevm.py
@@ -78,9 +78,10 @@ class RHEVMTemplateUpload(ProviderTemplateUpload):
             return
         rhevm_specs = cfme_data.template_upload.template_upload_rhevm
         disk_kwargs = dict(storage_domain=self.provider_data.template_upload.storage_domain,
-                           size=rhevm_specs.disk_size,
-                           interface=rhevm_specs.disk_interface,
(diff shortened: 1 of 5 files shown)


cfme/utils/template/rhevm.py
```diff
@@ -78,9 +78,10 @@ def add_disk_to_vm(self):
             return
         rhevm_specs = cfme_data.template_upload.template_upload_rhevm
         disk_kwargs = dict(storage_domain=self.provider_data.template_upload.storage_domain,
-                           size=rhevm_specs.disk_size,
-                           interface=rhevm_specs.disk_interface,
-                           format=rhevm_specs.disk_format)
+                           size=rhevm_specs.get('disk_size', 5000000000),
```
kedark3: I see we are defaulting to this value, do you think it would benefit us in any way to have that as a parameter on `add_disk_to_vm`?

kedark3:
LGTM, only one comment.

cfme/utils/template/rhevm.py
mshriver: Not in this case, since this is just a template upload method.

PR #8937, opened 2019-06-11
[1LP][RFR]Automate more tests Import/Export report

valaparthvi:
Changes:
1. Automate/add a few more tests:
    i. `test_import_duplicate_report`
    ii. `test_reports_invalid_file`
    iii. `test_widgets_invalid_file`
2. Remove tests from manual_file


cfme/tests/intelligence/reports/test_import_export_reports_widgets.py
```diff
@@ -15,56 +15,55 @@
 pytestmark = [pytest.mark.tier(1), test_requirements.report]
 
 
-@pytest.fixture()
-def widget_file(appliance):
-    yaml_name = "import_widget.yaml"
+def yaml_path(yaml_name):
+    """ Returns yaml path of the file with yaml_name name
+    """
+    yaml_name = yaml_name + ".yaml"
```
digitronik: Required:
`.format` please for string formatting. Also, try one line docstring

cfme/tests/intelligence/reports/test_import_export_reports_widgets.py
```diff
@@ -145,3 +140,60 @@ def test_export_report(appliance, report_file):
...
+    else:
+        view.flash.assert_message(
+            "Skipping Report (already in DB): [{}]".format(report.menu_name)
+        )
+
+
+@pytest.mark.parametrize("yaml_name", ["invalid_yaml", "invalid_report"])
+def test_reports_invalid_file(appliance, yaml_name):
+    """
+    Polarion:
```
digitronik: optional: Its look like a negative test. If yes, you can add it in polarion docstring

cfme/tests/intelligence/reports/test_import_export_reports_widgets.py
```diff
@@ -145,3 +140,60 @@ def test_export_report(appliance, report_file):
...
+        initialEstimate: 1/16h
+    """
+    with pytest.raises(AssertionError, match="Error", partial=True):
+        appliance.collections.reports.import_report(yaml_path(yaml_name))
+
+
+@pytest.mark.parametrize("yaml_name", ["invalid_yaml", "invalid_widget"])
+def test_widgets_invalid_file(appliance, yaml_name):
+    """
+    Polarion:
```
digitronik: same ^^

cfme/tests/intelligence/reports/test_import_export_reports_widgets.py
```diff
@@ -15,56 +15,51 @@
...
+    """ Returns yaml path of the file with yaml_name name"""
+    yaml_name = "{}.yaml".format(yaml_name)
+
     try:
         fs = FTPClientWrapper(cfme_data.ftpserver.entities.reports)
         file_path = fs.download(yaml_name, os.path.join("/tmp", yaml_name))
     except (FTPException, AttributeError):
         logger.exception("FTP download or YAML lookup of %s failed, defaulting to local", yaml_name)
-        file_path = data_path.join("ui/intelligence/import_widget.yaml").realpath().strpath
+        file_path = (data_path.join("ui/intelligence/{}".format(yaml_name)).realpath().strpath)
```
mshriver: You shouldn't have to use `format`, you're using `join()`. The `yaml_name` should be passed as second argument to `join`.

mshriver: Really, shouldn't even be passing `ui/intelligence`, you should be calling `data_path.join('ui', 'intelligence', yaml_name)`

cfme/tests/intelligence/reports/test_import_export_reports_widgets.py
```diff
@@ -145,3 +136,62 @@ def test_export_report(appliance, report_file):
...
+
+    if overwrite:
+        view.flash.assert_message("Replaced Report: [{}]".format(report.menu_name))
+    else:
+        view.flash.assert_message(
+            "Skipping Report (already in DB): [{}]".format(report.menu_name)
+        )
+
+
+@pytest.mark.parametrize("yaml_name", ["invalid_yaml", "invalid_report"])
```
mshriver: What is the functional difference between these two parameters?

In both cases, its just an invalid file path for import.

Can we assert any behavior against the appliance here - not just that there's an `AssertionError` raised by the `import_report` method? Can we expect a specific View to be displayed? Any evm.log or production.log messages? Any flash messages on the view?

cfme/tests/intelligence/reports/test_import_export_reports_widgets.py
```diff
@@ -145,3 +136,62 @@ def test_export_report(appliance, report_file):
         report_fields=["Archived", "Autostart", "Boot Time"],
     )
     collection.export_report(report.menu_name)
+
+
+@pytest.mark.tier(3)
+@pytest.mark.parametrize("overwrite", [True, False], ids=("overwrite", "replace"))
```
mshriver: Aren't `overwrite` and `replace` the same in this context? It looks to me like the `id` for the `False` parameter should be 'skipped'

cfme/tests/intelligence/reports/test_import_export_reports_widgets.py
valaparthvi: >What is the functional difference between these two parameters?

Both parameters have different data which gives a different error message when uploaded.

> Any flash messages on the view?

Yes, both files produce a different flash message, which we can use to assert.

cfme/fixtures/intelligence.py
```diff
@@ -0,0 +1,20 @@
...
+import pytest
+
+from cfme.utils.conf import cfme_data
+from cfme.utils.ftp import FTPClientWrapper
+from cfme.utils.ftp import FTPException
+from cfme.utils.log import logger
+
+
+@pytest.fixture
+def yaml_path(request, yaml_name):
```
mshriver: It looks like this was committed/pushed maybe halfway through writing?

This new fixture isn't registered in `cfme/test_framework/pytest_plugin.py`, which means its not available to any test modules.

The function still exists in test_import_export_reports_widgets.py below, not decorated as a fixture.

Fixtures should not take arguments, as they aren't called directly by the test (unless they return a function object).

Are you planning on writing a generic fixture that can be used by multiple test modules? If so they'll have to parametrize on yaml_name, not try to pass it or call this method directly.

cfme/tests/intelligence/reports/test_import_export_reports_widgets.py
mshriver: Okay, gotcha. I would like to see some docblock description of the expected content (or what's _wrong_ with the content) for the yaml files given by the parameter names.

The addition of flash message assertion certainly helps make it more clear what these files are for.

cfme/fixtures/intelligence.py
valaparthvi: Oh yes, I committed this one by mistake. I was working on writing a global fixture for reports and schedules. Removing it :grimacing:

PR #8938, opened 2019-06-11
[1LP][RFR] Simulation copy test

digitronik:

Purpose or Intent
=================

{{pytest: cfme/tests/automate/custom_button/test_buttons.py::test_simulated_object_copy_on_button}}

### DIFF
--- a/cfme/tests/automate/custom_button/test_buttons.py
+++ b/cfme/tests/automate/custom_button/test_buttons.py
@@ -6,6 +6,7 @@ import pytest
 from widgetastic_patternfly import Dropdown
 
 from cfme import test_requirements
+from cfme.automate.simulation import simulate
 from cfme.base.ui import AutomateSimulationView
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.markers.env_markers.provider import ONE_PER_TYPE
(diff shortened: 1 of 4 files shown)


cfme/tests/automate/custom_button/test_buttons.py
```diff
@@ -547,3 +547,61 @@ def test_attribute_override():
...
+    view = navigate_to(appliance.server, "AutomateSimulation")
+
+    simulation_data = {
+        "instance": "Automation",
+        "message": "test_bz",
+        "request": "InspectMe",
+        "target_type": target_type,
+        "target_object": target_obj,
+        "execute_methods": True,
+    }
```
john-dupuy: Is there any need for making a `Collection` for an `AutomateSimulation`? As far as I can tell, they are not really permanent objects, just temporary. But it may simplify the few test cases that use them. Do you think it's worth it?

digitronik: I thought same but I think we have one method.

digitronik: As it's not Entity based just one operation l.e. `simulate` so no need to implement collection for now. I improved some arguments like `appliance` needs for navigation.

PR #8942, opened 2019-06-11
[1LP][RFR] fixed CloudProviderInstancesView and tags

lina-nikiforova:
{{ pytest: -v cfme/tests/cloud_infra_common/test_tag_objects.py }}

PR #8944, opened 2019-06-11
[RFR] Add a widget to deal with sectioned bootstrap selects, fix action tests

john-dupuy:
Implementing a widget to deal with sectioned bootstrap selects, like the one present on the actions tagging page. 

This is meant to address the changes brought in CFME 5.11 with https://github.com/ManageIQ/manageiq-ui-classic/pull/5535

Since that page is not in the final state, this widget is more of a hack than a full-fledged implementation. The single bootstrap select will eventually be replaced by the tagging editor, which is essentially two bootstrap selects (one for the tag category, and one for the tag itself). This PR is a workaround until https://bugzilla.redhat.com/show_bug.cgi?id=1715397 is fixed. 

Fixes broken action tests which are impacting many tests that follow them.  

{{ pytest: --use-template-cache --long-running cfme/tests/control/test_basic.py::test_action_crud }}

### DIFF
--- a/widgetastic_manageiq/__init__.py
+++ b/widgetastic_manageiq/__init__.py
@@ -54,6 +54,7 @@ from widgetastic_patternfly import Button
 from widgetastic_patternfly import Dropdown
 from widgetastic_patternfly import Input
 from widgetastic_patternfly import NavDropdown
+from widgetastic_patternfly import SelectItemNotFound
 from widgetastic_patternfly import SelectorDropdown
 from widgetastic_patternfly import Tab
 from widgetastic_patternfly import VerticalNavigation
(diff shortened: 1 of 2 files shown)


john-dupuy:
@lina-nikiforova take a look when you get a chance, related to tagging widget changes.

widgetastic_manageiq/__init__.py
```diff
@@ -94,6 +95,83 @@ def currently_selected(self):
...
+    def all_options(self):
+        return [
+            self.SectionOption(
+                self.browser.text(
+                    self.browser.element(".//span[contains(@class, 'text')]", parent=el)
+                ),
+                el.get_attribute("data-original-index"),
+                el.get_attribute("data-optgroup"),
+            )
+            for el in self.browser.elements("./div/ul/li", parent=self)
```
mshriver: let's move this locator to a constant, OPTION_ITEM or something.

widgetastic_manageiq/__init__.py
```diff
@@ -94,6 +95,83 @@ def currently_selected(self):
...
+
+    @property
+    def all_options(self):
+        return [
+            self.SectionOption(
+                self.browser.text(
+                    self.browser.element(".//span[contains(@class, 'text')]", parent=el)
+                ),
+                el.get_attribute("data-original-index"),
+                el.get_attribute("data-optgroup"),
```
mshriver: These attributes are getting repeated, let's put them in constants where it makes sense.

widgetastic_manageiq/__init__.py
```diff
@@ -94,6 +95,83 @@ def currently_selected(self):
             return None
 
 
+class SectionedBootstrapSelect(BootstrapSelect):
```
mshriver: Ultimately I think we're going to want another wt.pf-react repo, or a separate module in wt.pf to store these widgets. Its fine here for now though.

widgetastic_manageiq/__init__.py
```diff
@@ -94,6 +95,83 @@ def currently_selected(self):
...
+        before = self.selected_option
+        if len(values) == 3:
+            section_text = values[1]
+        else:
+            section_text = values[0]
+        item_text = values[-1]
+        self.select_item_in_section(section_text, item_text)
+        after = self.selected_option
+        return before != after
+
```
mshriver: Any read() implementation any different than BoostrapSelect?

widgetastic_manageiq/__init__.py
john-dupuy: Agreed, @lina-nikiforova is working on the react version of this widget in #8942, this widget is just a special kind of `BootstrapSelect` however, no react necessary.

widgetastic_manageiq/__init__.py
john-dupuy: Nope, read is just 
```python,
    def read(self):
        if self.is_multiple:
            return self.all_selected_options
        else:
            return self.selected_option
```
`self.is_multiple` will be false for this widget, so it'll just return `self.selected_option`.

widgetastic_manageiq/__init__.py
```diff
@@ -94,6 +95,85 @@ def currently_selected(self):
...
+            self.Option(self.browser.text(el), el.get_attribute(self.GROUP_INDEX))
+            for el in self.browser.elements(self.SECTIONS, parent=self)
+        ]
+
+    @property
+    def all_options(self):
+        return [
+            self.SectionOption(
+                self.browser.text(self.browser.element(self.ITEM_TEXT, parent=el)),
+                el.get_attribute(self.ITEM_INDEX),
```
mshriver: Nice, I feel like this is much more readable with the context of _what_ the locator is

PR #8946, opened 2019-06-11
[RFR] views support in db client

izapolsk:
db client doesn't support views but some cf tables have been turned into views.
This PR should add support for views.

{{ pytest: --use-template-cache --long-running cfme/tests/candu/test_utilization_metrics.py }}


izapolsk:
PRT results look good. please ignore failed tests, those aren't related to my changes.

PR #8953, opened 2019-06-13
[1LP][RFR] Testing user requester for service or life cycle provision

ganeshhubale:
- Testing requester for service or life cycle provision are in logs

{{ pytest: cfme/tests/automate/test_service_automate.py -k 'test_user_requester_for_lifecycle_provision' -vv }}

### DIFF
--- a/cfme/tests/automate/test_service_automate.py
+++ b/cfme/tests/automate/test_service_automate.py
@@ -29,9 +29,12 @@ def new_users(appliance):
     ) for i in range(2)]
 
     yield users
-    if not BZ(1720273).blocks:
-        for user in users:
+    for user in users:
+        if not BZ(1720273).blocks:
             user.delete_if_exists()
+        else:
+            user = appliance.rest_api.collections.users.get(name=user.name)
+            user.action.delete()
 
 
 @pytest.fixture(scope='function')
--- a/cfme/tests/automate/test_service_automate.py
+++ b/cfme/tests/automate/test_service_automate.py
@@ -18,22 +18,20 @@ pytestmark = [test_requirements.automate]
 @pytest.fixture(scope="module")
 def new_users(appliance):
     """This fixture creates new users"""
-    users = []
-    for i in range(2):
-        user = appliance.collections.users.create(
-            name="user_{}".format(fauxfactory.gen_alphanumeric().lower()),
-            credential=Credential(principal='uid{}'.format(fauxfactory.gen_alphanumeric(4)),
-                                  secret=fauxfactory.gen_alphanumeric(4)),
-            email=fauxfactory.gen_email(),
-            groups=appliance.collections.groups.instantiate(
-                description="EvmGroup-super_administrator"),
-            cost_center="Workload",
-            value_assign="Database",
-        )
-        users.append(user)
+    users = [appliance.collections.users.create(
+        name="user_{}".format(fauxfactory.gen_alphanumeric().lower()),
+        credential=Credential(principal='uid{}'.format(fauxfactory.gen_alphanumeric(4)),
+                              secret=fauxfactory.gen_alphanumeric(4)),
+        email=fauxfactory.gen_email(),
+        groups=appliance.collections.groups.instantiate(description="EvmGroup-super_administrator"),
+        cost_center="Workload",
+        value_assign="Database",
+    ) for i in range(2)]
+
     yield users
-    if not BZ(1720273).blocks and user in users:
-        user.delete_if_exists()
+    if not BZ(1720273).blocks:
+        for user in users:
+            user.delete_if_exists()
 
 
 @pytest.fixture(scope='function')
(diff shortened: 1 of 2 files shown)


cfme/tests/automate/test_service_automate.py
```diff
@@ -0,0 +1,142 @@
...
+
+pytestmark = [test_requirements.automate]
+
+
+@pytest.fixture(scope="module")
+def new_users(appliance):
+    """This fixture creates new users"""
+    users = []
+    for i in range(2):
+        user = appliance.collections.users.create(
```
john-dupuy: **Optional** consider using a list comprehension here 
```python,
        users = [appliance.collections.users.create(
            name="user_{}".format(fauxfactory.gen_alphanumeric().lower()),
            credential=Credential(principal='uid{}'.format(fauxfactory.gen_alphanumeric(4)),
                                  secret=fauxfactory.gen_alphanumeric(4)),
            email=fauxfactory.gen_email(),
            groups=appliance.collections.groups.instantiate(
                description="EvmGroup-super_administrator"),
            cost_center="Workload",
            value_assign="Database",
        ) for i in range(2)]
```

cfme/tests/automate/test_service_automate.py
```diff
@@ -0,0 +1,142 @@
...
+                                  secret=fauxfactory.gen_alphanumeric(4)),
+            email=fauxfactory.gen_email(),
+            groups=appliance.collections.groups.instantiate(
+                description="EvmGroup-super_administrator"),
+            cost_center="Workload",
+            value_assign="Database",
+        )
+        users.append(user)
+    yield users
+    if not BZ(1720273).blocks and user in users:
```
john-dupuy: Shouldn't this BZ block the test from running?

cfme/tests/automate/test_service_automate.py
```diff
@@ -0,0 +1,142 @@
...
+    ).namespaces.instantiate(name="Provisioning").namespaces.instantiate(
+        name="StateMachines").classes.instantiate(
+        name="ServiceProvisionRequestApproval").methods.instantiate(
+        name="validate_request")
+    return method
+
+
+@pytest.mark.tier(3)
+@pytest.mark.provider([VMwareProvider], scope="module")
+@pytest.mark.meta(blockers=[BZ(1671563, forced_streams=['5.10'])])
```
john-dupuy: @ganeshhubale is BZ 1671563 going to be backported to 5.10? If not, you can just ignore the stream.

cfme/tests/automate/test_service_automate.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+    Polarion:
+        assignee: ghubale
+        casecomponent: Automate
+        caseimportance: high
+        initialEstimate: 1/6h
+        tags: automate
+
+    Bugzilla:
+         1671563
```
john-dupuy: **Required** add BZ 1720273 here.

cfme/tests/automate/test_service_automate.py
```diff
@@ -0,0 +1,142 @@
...
+    $evm.log(:info, "This is the user: #{user.userid}")
+
+    $evm.log("info", "Listing Root Object Attributes:")
+    $evm.root.attributes.sort.each { |k, v| $evm.log("info", "\t#{k}: #{v}") }
+    $evm.log("info", "===========================================")
+    """
+    infra_validate_request.update(updates={"script": script})
+    service_validate_request.update(updates={"script": script})
+
+    with new_users[0]:
```
john-dupuy: This `new_users` seems like a good candidate for test parametrization. Is there a reason you haven't done that here?

cfme/tests/automate/test_service_automate.py
ganeshhubale: - No. This BZ doesn't have any relationship with test case. 
- It gives problem while doing teardown of custom users in 5.11.

cfme/tests/automate/test_service_automate.py
ganeshhubale: Actually, BZ was reported on 5.9z but fixed in 5.11 only.
Hence using ignore stream.

cfme/tests/automate/test_service_automate.py
ganeshhubale: Actually, test case needs two users in the system and need to login one by one.

cfme/tests/automate/test_service_automate.py
john-dupuy: Right, if the tear down for the test is unable to occur, shouldn't we block the test case? Or is there no harm in having more than one user on an appliance after this test completes?

ganeshhubale: I am not sure about the blocking the test case only because of teardown.
According to me, there will not be any harm if more than one user in the appliance.

john-dupuy: Discussed offline about cleaning up via a REST query, that seems like the best option while this BZ is blocking.

john-dupuy:
Please check PRT failure in 5.11, moving to WIPTEST for the time being.

PR #8957, opened 2019-06-14
[RFR] Add cleanup to test provision from template using rest

john-dupuy:
Adding cleanup for `test_provision_from_template_using_rest`. This test case will create an instance/vm but does not clean it up when the test is finished. 

In testing this on PRT, I noticed a 5.11 BZ: 
https://bugzilla.redhat.com/show_bug.cgi?id=1720751

So I am adding that BZ as a blocker for the test on SCVMM. 

{{ pytest: --use-provider scvmm --long-running cfme/tests/cloud_infra_common/test_provisioning.py::test_provision_from_template_using_rest }}


john-dupuy:
Note that 5.11 PRT won't pass until BZ 1720751 `blocks` after it's triaged.

PR #8960, opened 2019-06-17
[1LP][RFR] New Test: Tesing quota check while provisioning services

ganeshhubale:
- Testing whether quota check is done for once while provisioning services
- This line: ```Getting Tenant Quota Values for:``` should be visible once in automation logs
- Added ```match_count``` function in ```logvalidator```

Waiting for this PR #8918 

{{ pytest: cfme/tests/automate/test_method.py -k 'test_automate_service_quota_runs_only_once' -vv }}

### DIFF
--- a/cfme/tests/automate/test_method.py
+++ b/cfme/tests/automate/test_method.py
@@ -472,4 +472,4 @@ def test_automate_service_quota_runs_only_once(appliance, generic_catalog_item):
     )
     provision_request = service_catalogs.order()
     provision_request.wait_for_request()
-    assert result.match_count()[result.matched_patterns[0]] == 1
+    assert result.matches[result.matched_patterns[0]] == 1
--- a/cfme/tests/automate/test_method.py
+++ b/cfme/tests/automate/test_method.py
@@ -462,9 +462,10 @@ def test_automate_service_quota_runs_only_once(appliance, generic_catalog_item):
     Bugzilla:
         1317698
     """
+    pattern = ".*Getting Tenant Quota Values for:.*"
     result = LogValidator(
         "/var/www/miq/vmdb/log/automation.log",
-        matched_patterns=[".*Getting Tenant Quota Values for:.*"]
+        matched_patterns=[pattern]
     )
     result.fix_before_start()
     service_catalogs = ServiceCatalogs(
@@ -472,4 +473,4 @@ def test_automate_service_quota_runs_only_once(appliance, generic_catalog_item):
     )
     provision_request = service_catalogs.order()
     provision_request.wait_for_request()
-    assert result.matches[result.matched_patterns[0]] == 1
+    assert result.matches[pattern] == 1
--- a/cfme/tests/automate/test_method.py
+++ b/cfme/tests/automate/test_method.py
@@ -5,6 +5,7 @@ import pytest
 from cfme import test_requirements
 from cfme.automate.explorer.klass import ClassDetailsView
 from cfme.automate.simulation import simulate
+from cfme.services.service_catalogs import ServiceCatalogs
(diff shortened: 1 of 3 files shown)


cfme/utils/log_validator.py
```diff
@@ -101,3 +101,12 @@ def validate():
             except Failed:
                 return False
         wait_for(validate, delay=delay, num_sec=num_sec, message=message, **kwargs)
+
+    def match_count(self):
```
john-dupuy: As it stands this function is a bit separate from the other utilities of the class. I.e. it is not making use of `_check_skip_logs`, `_check_fail_logs`, `_check_match_logs`, or `_verify_match_logs`. 

Perhaps, there can be a modification to `self.matches` to include not only the Boolean of whether or not the pattern was found, but also the count. Or there can be a separate dictionary defined in the class, `match_count` or something similar.

What do you think? My main concern is that this function is lacking some of the features that the typical `LogValidator` has already.

ganeshhubale: Hey @john-dupuy 
- I would like to keep it as a different method in ```LogValidator``` as it returns count of matches in logs.
- If we tried to modify the self.matches dict then again we will need to add new method to get match count. Because the purpose of ```_check_match_logs``` is to check if logs are matched or not.
- Modifying existing methods will created problem while returning values and it will affect other tests too.
- Please let me know, if we need any updates in existing method - ```match_count```
- Also @digitronik is also working on it. Please have a look: https://github.com/ManageIQ/integration_tests/pull/8918

john-dupuy: @ganeshhubale 

Thanks for the detailed reply. Modifying `self.matches` was just a suggestion, and I agree that it will require a larger refactor that isn't necessarily warranted. 

My main concern about this function is that it doesn't support the standard features of the function, i.e. `_check_skip_logs`, `_check_fail_logs`. So, for example, if there is a line you do **not** want added to the `match_count`, you should be able to pass a skip_pattern, and those lines won't be added to the count. 

Does that make sense? I think this could be done within your existing method `match_count`, but if you'd like to do this in a different PR, that's fine with me.

digitronik: The fact is current `LogValidator` is just for matching some string pattern from a log not exactly count. `LogValidator` must be flexible about actions and cover multiple purposes. Even I am facing a problem with `tails` it's not providing me exact start with `fix_before_start` method. when I'm handling a large amount of request like `50+`. We need to overwrite it or make wrapper. My automation working properly with clearing `automation.log` but its bad practice. I'm finding a way to handle 100+ request match count. @izapolsk

john-dupuy:
Marking this as WIPTEST until #8918 is merged.

cfme/tests/automate/test_method.py
```diff
@@ -446,3 +447,29 @@ def test_automate_generic_object_service_associations(appliance, klass, go_servi
...
+        "/var/www/miq/vmdb/log/automation.log",
+        matched_patterns=[".*Getting Tenant Quota Values for:.*"]
+    )
+    result.fix_before_start()
+    service_catalogs = ServiceCatalogs(
+        appliance, catalog=generic_catalog_item.catalog, name=generic_catalog_item.name
+    )
+    provision_request = service_catalogs.order()
+    provision_request.wait_for_request()
+    assert result.matches[result.matched_patterns[0]] == 1
```
digitronik: I would avoid accessing `dict` key with indexing. can we go like
```
pattern = ".*Getting Tenant Quota Values for:.*"
result = LogValidator(
        "/var/www/miq/vmdb/log/automation.log",
        matched_patterns=[pattern]
    )


assert result.matches[result.matched_patterns[pattern]] == 1
```

PR #8961, opened 2019-06-17
[1LP][RFR] Automate test: test_service_retirement_methods AND fix service retire…

valaparthvi:
Changes:
1. Add a new test: `test_service_retirement_methods`
2. Fix service retirement via REST failures.

{{ pytest: cfme/tests/services/test_rest_services.py -k "test_service_retirement_methods or test_retire_service_now or test_retire_service_future or test_retire_parent_service_now" --long-running --use-template-cache -sqvvv }}

### DIFF
--- a/cfme/tests/services/test_rest_services.py
+++ b/cfme/tests/services/test_rest_services.py
@@ -24,6 +24,7 @@ from cfme.rest.gen_data import services as _services
 from cfme.rest.gen_data import TEMPLATE_TORSO
 from cfme.rest.gen_data import vm as _vm
 from cfme.utils.blockers import BZ
+from cfme.utils.log_validator import LogValidator
 from cfme.utils.rest import assert_response
 from cfme.utils.rest import delete_resources_from_collection
 from cfme.utils.rest import delete_resources_from_detail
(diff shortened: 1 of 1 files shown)


cfme/tests/services/test_rest_services.py
```diff
@@ -425,6 +425,49 @@ def _finished(service):
...
+        Polarion:
+            assignee: pvala
+            casecomponent: Services
+            initialEstimate: 1/4h
+
+        Bugzilla:
+            1698480
+            1713477
+        """
+        service = services[0]
```
digitronik: Q: are you sure it will not raise `KeyError`?

cfme/tests/services/test_rest_services.py
```diff
@@ -425,6 +425,49 @@ def _finished(service):
...
+            appliance.rest_api.collections.services.action.retire(service)
+
+        assert_response(appliance)
+        wait_for(
+            lambda: hasattr(service, "retirement_state"),
+            fail_func=service.reload,
+            num_sec=50,
+            delay=5,
+        )
+        assert service.retirement_state == "initializing"
```
digitronik: Please try to validate Error msg form log as well.

cfme/tests/services/test_rest_services.py
valaparthvi: I am pretty sure it will not raise `KeyError`. `services` is a fixture that will create 3 services, even if there is an error, it will be caught before the test begins.

cfme/tests/services/test_rest_services.py
valaparthvi: Yes, I am currently working on that.

digitronik: so you used matched `matched_patterns` :) it's ok

PR #8879, opened 2019-05-28
[1LP][RFR]updating vmware test to automate vmware filters test and test vdswitch ui display

kedark3:
Purpose or Intent
=================

__Adding tests__ automated test for vmware_filters and test vdswitch ui display.

{{pytest: cfme/tests/infrastructure/test_vmware_provider.py }}

PR #8964, opened 2019-06-17
[1LP][RFR]Cd dvd/DIsk table updates and Tests

kedark3:
Purpose or Intent
=================
**IMPORTANT NOTE**: This PR includes commits from the PR #8879 as it was written in succession. 

__Adding tests__ for `test_vmware_cdrom_dropdown_not_blank`  This is going to fail on last assertion as this is a bug that is not fixed. But to avoid that, I have added blocker marker to it. Below GIF shows what would test do once bz is resolved.
Peek 2019-06-17 16-48

Merge after #8879 

{{pytest: cfme/tests/infrastructure/test_vmware_provider.py --use-provider complete -k 'test_vmware_cdrom_dropdown_not_blank' }}

### DIFF
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -66,7 +66,6 @@ from widgetastic_manageiq import SnapshotMemorySwitch
 from widgetastic_manageiq import SummaryTable
 from widgetastic_manageiq import Table
 from widgetastic_manageiq import TimelinesView
-from widgetastic_manageiq.vm_reconfigure import CDDVDTable
 from widgetastic_manageiq.vm_reconfigure import DisksTable
 
 
@@ -370,7 +369,7 @@ class InfraVmReconfigureView(BaseLoggedInPage):
     )
     cd_dvd_table = WTable('//div/table[./../h3[normalize-space(text())="CD/DVD Drives"]]',
         column_widgets={
-            "Host File": BootstrapSelect(id="isoNamde"),
+            "Host File": BootstrapSelect(id="isoName"),
             "Actions": Button(),
             # second action button, 'Cancel Add' or 'Delete' depending on context of row
             # https://github.com/RedHatQE/widgetastic.core/issues/95
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -15,6 +15,7 @@ from widgetastic.utils import Parameter
 from widgetastic.utils import partial_match
 from widgetastic.widget import NoSuchElementException
 from widgetastic.widget import ParametrizedView
+from widgetastic.widget import Table as WTable
 from widgetastic.widget import Text
 from widgetastic.widget import TextInput
 from widgetastic.widget import View
@@ -65,6 +66,7 @@ from widgetastic_manageiq import SnapshotMemorySwitch
 from widgetastic_manageiq import SummaryTable
 from widgetastic_manageiq import Table
 from widgetastic_manageiq import TimelinesView
+from widgetastic_manageiq.vm_reconfigure import CDDVDTable
 from widgetastic_manageiq.vm_reconfigure import DisksTable
 
 
@@ -351,7 +353,30 @@ class InfraVmReconfigureView(BaseLoggedInPage):
     cores_per_socket = BootstrapSelect(id='cores_per_socket_count')
     cpu_total = WInput()  # read-only, TODO widgetastic
 
-    disks_table = DisksTable()
+    disks_table = DisksTable('//div/table[./../h3[normalize-space(text())="Disks"]//button]',
+        column_widgets={
+            "Type": BootstrapSelect(id="hdType"),
+            "Mode": BootstrapSelect(id="hdMode"),
+            "Size": WInput(id="dvcSize"),
+            "ControllerType": BootstrapSelect(id="Controller"),
+            "Unit": BootstrapSelect(id="hdUnit"),
+            "Dependent": BootstrapSwitch(name="vm.cb_dependent"),
+            "Delete Backing": BootstrapSwitch(name="cb_deletebacking"),
+            "Actions": Button(),
+            # second action button, 'Cancel Add' or 'Delete' depending on context of row
+            # https://github.com/RedHatQE/widgetastic.core/issues/95
+            9: Button(),
+        }
+    )
+    cd_dvd_table = WTable('//div/table[./../h3[normalize-space(text())="CD/DVD Drives"]]',
+        column_widgets={
+            "Host File": BootstrapSelect(id="isoNamde"),
+            "Actions": Button(),
+            # second action button, 'Cancel Add' or 'Delete' depending on context of row
+            # https://github.com/RedHatQE/widgetastic.core/issues/95
+            3: Button('Connect'),
+        }
+    )
     affected_vms = Table('.//div[@id="records_div" or @id="miq-gtl-view"]//table')
 
     submit_button = Button('Submit', classes=[Button.PRIMARY])
--- a/cfme/tests/infrastructure/test_vmware_provider.py
+++ b/cfme/tests/infrastructure/test_vmware_provider.py
@@ -221,9 +221,8 @@ def test_vmware_inaccessible_datastore():
     pass
 
 
-@pytest.mark.manual
 @pytest.mark.tier(1)
-def test_vmware_cdrom_dropdown_not_blank():
+def test_vmware_cdrom_dropdown_not_blank(appliance, provider):
     """
     Test CD/DVD Drives dropdown lists ISO files, dropdown is not blank
 
@@ -249,6 +248,18 @@ def test_vmware_cdrom_dropdown_not_blank():
             4.Virtual machine is selected
             5.Dropdown of ISO files is not empty for CD/DVD Drive
     """
+    datastore_collection = appliance.collections.datastores
+    ds = [ds.name for ds in provider.data['datastores'] if ds.type == 'iso']
+    try:
+        iso_ds = datastore_collection.instantiate(name=ds[0], provider=provider)
+    except IndexError:
+        pytest.skip('No datastores found of type iso on provider {}'.format(provider.name))
+    iso_ds.run_smartstate_analysis()
+    vms_collections = appliance.collections.infra_vms
+    vm = vms_collections.instantiate(name='cu-24x7', provider=provider)
+    view = navigate_to(vm, 'Reconfigure')
+    from IPython import embed
+    embed()
     pass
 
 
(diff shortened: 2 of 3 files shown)


cfme/infrastructure/virtual_machines.py
```diff
@@ -351,7 +352,30 @@ class InfraVmReconfigureView(BaseLoggedInPage):
     cores_per_socket = BootstrapSelect(id='cores_per_socket_count')
     cpu_total = WInput()  # read-only, TODO widgetastic
 
-    disks_table = DisksTable()
+    disks_table = DisksTable('//div/table[./../h3[normalize-space(text())="Disks"]//button]',
+        column_widgets={
```
kedark3: @mshriver Please review this code that we wrote together.

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -249,7 +254,31 @@ def test_vmware_cdrom_dropdown_not_blank():
...
+    actions_column = view.cd_dvd_table[0]['Actions']
+    # First disconnect if already connected
+    if actions_column.text == 'Connect Disconnect':
+        actions_column.click()
+    # Confirm disconnect
+    if actions_column.text == 'Confirm':
+        actions_column.click()
+    # if 'Connect' option is present, click it
+    if actions_column.text == 'Connect':
+        actions_column.click()
```
jawatts: Is there a chance it isn't any of these? It would be good to assert that it is connected.

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -249,7 +254,31 @@ def test_vmware_cdrom_dropdown_not_blank():
...
+        iso_ds = datastore_collection.instantiate(name=ds[0], provider=provider)
+    except IndexError:
+        pytest.skip('No datastores found of type iso on provider {}'.format(provider.name))
+    iso_ds.run_smartstate_analysis()
+    vms_collections = appliance.collections.infra_vms
+    vm = vms_collections.instantiate(name='cu-24x7', provider=provider)
+    view = navigate_to(vm, 'Reconfigure')
+    actions_column = view.cd_dvd_table[0]['Actions']
+    # First disconnect if already connected
+    if actions_column.text == 'Connect Disconnect':
```
jawatts: Doing some local testing in 5.10, looks like it is "Disconnect" and not "Connect Disconnect"

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -249,7 +254,31 @@ def test_vmware_cdrom_dropdown_not_blank():
...
+    ds = [ds.name for ds in provider.data['datastores'] if ds.type == 'iso']
+    try:
+        iso_ds = datastore_collection.instantiate(name=ds[0], provider=provider)
+    except IndexError:
+        pytest.skip('No datastores found of type iso on provider {}'.format(provider.name))
+    iso_ds.run_smartstate_analysis()
+    vms_collections = appliance.collections.infra_vms
+    vm = vms_collections.instantiate(name='cu-24x7', provider=provider)
+    view = navigate_to(vm, 'Reconfigure')
+    actions_column = view.cd_dvd_table[0]['Actions']
```
jawatts: Let's put a comment about the index here.

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -249,7 +254,31 @@ def test_vmware_cdrom_dropdown_not_blank():
             4.Virtual machine is selected
             5.Dropdown of ISO files is not empty for CD/DVD Drive
     """
-    pass
+    datastore_collection = appliance.collections.datastores
+    ds = [ds.name for ds in provider.data['datastores'] if ds.type == 'iso']
+    try:
```
jawatts: **Question** Why did you go this route vs have an `if/else` here. I feel it would be better to check if the list is empty vs try to instantiate something that isn't there.

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -163,7 +164,11 @@ def test_vmware_reconfigure_vm_controller_type():
             3.Reconfigure VM opion should be enabled
             4.Controller type should be listed
     """
-    pass
+    vms_collections = appliance.collections.infra_vms
+    vm = vms_collections.instantiate(name='cu-24x7', provider=provider)
+    view = navigate_to(vm, 'Reconfigure')
+    row = view.disks_table[0]
```
jawatts: A comment here should be good saying that you are grabbing the first row in the table.

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -163,7 +164,11 @@ def test_vmware_reconfigure_vm_controller_type():
             3.Reconfigure VM opion should be enabled
             4.Controller type should be listed
     """
-    pass
+    vms_collections = appliance.collections.infra_vms
+    vm = vms_collections.instantiate(name='cu-24x7', provider=provider)
```
jawatts: **Required** I think we should check to make sure this VMs exists before continuing, there is a chance that this may not exist on the provider. In addition, we should have `cap_and_util` as a required field for the provider.

kedark3: Will add this.

cfme/tests/infrastructure/test_vmware_provider.py
kedark3: I don't have preference here. In either case I'd skip the test on `else` or `expect`. Do you think it will call the function before it checks if `ds[0]` actually exists and valid?

cfme/tests/infrastructure/test_vmware_provider.py
kedark3: If you read the text on `action_column.text` it does actually read text from both buttons. Give it a try. I did use only `Disconnect` first and didn't work.

cfme/tests/infrastructure/test_vmware_provider.py
kedark3: There will be something. It is usually Connected, hence trying to disconnect first. Certainly there is implicit assumption there too. But for what I saw, it was always connected. So flow I assumed is `Disconnect`->`Confirm`->`(Re)Connect`
Do we really need to assert it? wouldn't the `if` conditions take care of it?

kedark3: Adding a try/except for actions_column, for some VMs the table is empty I noticed.

cfme/tests/infrastructure/test_vmware_provider.py
kedark3: image

kedark3: As you can see that UI shows only `Disconnect` but `.text ` returns both.

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -18,7 +19,8 @@
     pytest.mark.provider([VMwareProvider],
                         required_fields=[['provisioning', 'template'],
                                         ['provisioning', 'host'],
-                                        ['provisioning', 'datastore']],
+                                        ['provisioning', 'datastore'],
+                                        (["cap_and_util", "capandu_vm"], "cu-24x7")],
```
mshriver: :+1:

cfme/tests/infrastructure/test_vmware_provider.py
jawatts: I think this may be a UI bug

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -221,9 +228,9 @@ def test_vmware_inaccessible_datastore():
     pass
 
 
-@pytest.mark.manual
 @pytest.mark.tier(1)
-def test_vmware_cdrom_dropdown_not_blank():
+# @pytest.mark.meta(blockers=[BZ(1689369, forced_streams=['5.10', '5.11'])])
```
jawatts: Why is this commented out?

cfme/tests/infrastructure/test_vmware_provider.py
jawatts: I think it be better to check if ds is not None and then try to instantiate the first index

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -249,7 +256,36 @@ def test_vmware_cdrom_dropdown_not_blank():
...
-    pass
+    datastore_collection = appliance.collections.datastores
+    ds = [ds.name for ds in provider.data['datastores'] if ds.type == 'iso']
+    try:
+        iso_ds = datastore_collection.instantiate(name=ds[0], provider=provider)
+    except IndexError:
+        pytest.skip('No datastores found of type iso on provider {}'.format(provider.name))
+    iso_ds.run_smartstate_analysis()
+    vms_collections = appliance.collections.infra_vms
+    vm = vms_collections.instantiate(name='cu-24x7', provider=provider)
```
jawatts: same check here for the vm existing

cfme/tests/infrastructure/test_vmware_provider.py
jawatts: still needs to be added here

cfme/tests/infrastructure/test_vmware_provider.py
jawatts: This was discussed offline. @john-dupuy and I did not see a better way for this. Maybe @mshriver or @izapolsk has a better one for these multiple if statements

cfme/tests/infrastructure/test_vmware_provider.py
kedark3: Oops, it shouldn't be.

kedark3:
@jawatts updated PR, please let me know if it looks good :) Would like to merge it today if possible.

cfme/tests/infrastructure/test_vmware_provider.py
jawatts: @kedark3 was thinking more like:
```if not vm.exists_on_provider:
        pytest.skip("Skipping test, cu-24x7 VM does not exist")```

I wouldn't say we want a fail test here, because the test isn't failing if the vm is not on the provider.

File: cfme/tests/infrastructure/test_vmware_provider.py (hunk 10 above)
Comment (jawatts) [reply cfme/tests/infrastructure/test_vmware_provider.py]: same update to the check here

File: cfme/tests/infrastructure/test_vmware_provider.py (hunk 5 above)
Comment (jawatts) [reply cfme/tests/infrastructure/test_vmware_provider.py]: leaving this as is, discussed offline

File: cfme/tests/infrastructure/test_vmware_provider.py (hunk 2 above)
Comment (mshriver) [reply cfme/tests/infrastructure/test_vmware_provider.py]: Yeah this is certainly an odd pattern.  If these `if` statements aren't `True`, then the test isn't actually clicking the button, and there's no indication in the test output.

I think it would be better to follow a pattern that actually asserts the expected states, not just an `if` check.

Reading this, it seems you could collapse it to `if actions_column.text in ['Connect', 'Confirm', 'Connect Disconnect']: actions_column.click()`.  This isn't the case though, because you're actually expecting this test to click the action column multiple times.

I think it should be more like:
```
assert actions_column.text == 'Connect Disconnect', 'CD/DVD action button not in expected state'
actions_column.click()
assert actions_column.text == 'Confirm', 'CD/DCD action button did not change to "Confirm"'
# etc
```

PR #8966, opened 2019-06-18
[1LP][RFR] Added new test cases

ganeshhubale:
- Added test coverage for automate FA

### DIFF
--- a/cfme/tests/automate/test_automate_manual.py
+++ b/cfme/tests/automate/test_automate_manual.py
@@ -1004,10 +1004,13 @@ def test_git_refresh_with_rapid_updates():
         startsin: 5.10
         casecomponent: Automate
         testSteps:
-            1. Have a git backed domain that imported cleanly (if possibly incompletely)
+            1. Have a git backed domain that imported cleanly
             2. Break the domain in Git, or notice a method isn't visible because its matching .yaml
                was never added to git
             3. Add an broken .yaml to git, push, etc, in a desperate attempt to fix the issue.
+               Note: There are different .yaml files for domain, namespace, class etc. So to break
+               this file; you can change file name from __domain__.yaml to __testdomain__.yaml(or
+               any) or you can change the code in the .yaml file
             4. Go to CF UI, Automate, Domain, "Refresh with a new branch or tag"
             5. Select suitable branch and "Save"
             6. Check evm.log
--- a/cfme/tests/automate/test_automate_manual.py
+++ b/cfme/tests/automate/test_automate_manual.py
@@ -930,3 +930,127 @@ def test_git_refresh_with_renamed_yaml():
         1716443
     """
     pass
+
+
+@pytest.mark.tier(2)
+def test_upload_blank_line():
+    """
+    Polarion:
+        assignee: ghubale
+        initialEstimate: 1/8h
+        caseposneg: negative
+        startsin: 5.10
+        casecomponent: Automate
+        testSteps:
+            1. Create blank zip(test.zip) and yaml(test.yml) file
+            2. Navigate to Automation > Automate > Import/Export and upload test.zip file
+            3. Navigate to Automation > Automate > Customization > Import/Export and upload test.yml
+        expectedResults:
+            1.
+            2. Error message should be displayed
+            3. Error message should be displayed
+
+    Bugzilla:
+        1720611
+    """
+    pass
+
+
+@pytest.mark.tier(1)
+def test_initiate_retirement_automate_method():
+    """
+    Polarion:
+        assignee: ghubale
+        initialEstimate: 1/8h
+        caseposneg: positive
+        startsin: 5.10
+        casecomponent: Automate
+        testSteps:
+            1. Create a writeable domain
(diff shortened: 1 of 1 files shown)


cfme/tests/automate/test_automate_manual.py
```diff
@@ -930,3 +930,127 @@ def test_git_refresh_with_renamed_yaml():
...
+    """
+    Polarion:
+        assignee: ghubale
+        initialEstimate: 1/8h
+        caseposneg: positive
+        startsin: 5.10
+        casecomponent: Automate
+        testSteps:
+            1. Have a git backed domain that imported cleanly (if possibly incompletely)
+            2. Break the domain in Git, or notice a method isn't visible because its matching .yaml
```
john-dupuy: What do you mean by breaking the Git domain?

cfme/tests/automate/test_automate_manual.py
```diff
@@ -930,3 +930,127 @@ def test_git_refresh_with_renamed_yaml():
...
+    pass
+
+
+@pytest.mark.tier(2)
+def test_git_refresh_with_rapid_updates():
+    """
+    Polarion:
+        assignee: ghubale
+        initialEstimate: 1/8h
+        caseposneg: positive
```
john-dupuy: Isn't this a negative test case?

cfme/tests/automate/test_automate_manual.py
```diff
@@ -930,3 +930,127 @@ def test_git_refresh_with_renamed_yaml():
...
+            8. It should re-pull or force user to do something if (5) is updated to block
+
+    Bugzilla:
+        1696396
+    """
+    pass
+
+
+@pytest.mark.tier(2)
+def test_simulation_copy_button():
```
john-dupuy: Isn't this automated by https://github.com/ManageIQ/integration_tests/pull/8938? @digitronik ?

cfme/tests/automate/test_automate_manual.py
```diff
@@ -930,3 +930,127 @@ def test_git_refresh_with_renamed_yaml():
...
+    """
+    Polarion:
+        assignee: ghubale
+        initialEstimate: 1/8h
+        caseposneg: positive
+        startsin: 5.10
+        casecomponent: Automate
+        testSteps:
+            1. Create a writeable domain
+            2. Create a /System/Request/retire_automation_service instance and set meth5 to
```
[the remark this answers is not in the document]
ganeshhubale: - This is no the typo. If you see the schema of class ```/System/Request``` in ManageIQ domain then you will the methods like meth1, meth2 upto meth7.
- So here I am just copying this default class and editing it's schema variable's values.

cfme/tests/automate/test_automate_manual.py
```diff
@@ -930,3 +930,127 @@ def test_git_refresh_with_renamed_yaml():
...
+def test_git_refresh_with_rapid_updates():
+    """
+    Polarion:
+        assignee: ghubale
+        initialEstimate: 1/8h
+        caseposneg: positive
+        startsin: 5.10
+        casecomponent: Automate
+        testSteps:
+            1. Have a git backed domain that imported cleanly (if possibly incompletely)
```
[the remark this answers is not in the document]
ganeshhubale: I had picked that as it is from BZ steps. Actually we want one git backed domain which is imported in CF cleanly(without errors). Hence removing sentence in the parentheses.

cfme/tests/automate/test_automate_manual.py
ganeshhubale: Breaking the git domain means to add broken .yaml or do not add .yaml file which is important while importing domain.
Or we can have a domain which is blank without any namespaces etc

cfme/tests/automate/test_automate_manual.py
ganeshhubale: - Yeah there is confusion in mind about it.
- Because if you see in this test case first we are checking if broking domain is imported or not(here we want error should be displayed) and at the end we are fixing this broken domain to import it successfully(here we want that fixed domain to import successfully).
- Hence keeping it as positive.

cfme/tests/automate/test_automate_manual.py
ganeshhubale: Nope this test case is different as it copy from simulation page and paste it in custom button. 
This(```test_simulation_copy_button```) test case just checking basic functionality of the copy button on simulation page respective to changes in the values on that page.

cfme/tests/automate/test_automate_manual.py
john-dupuy: Got it, could you add some explanation of that in the `testSteps`?

cfme/tests/automate/test_automate_manual.py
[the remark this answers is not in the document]
john-dupuy: Ah, my mistake, thanks for the explanation!

cfme/tests/automate/test_automate_manual.py
john-dupuy: Yeah, it's kind of a mix of positive and negative. I agree that the end result is positive, so positive makes the most sense.

PR #8968, opened 2019-06-18
[1LP][RFR] New Test: Testing quota source value

ganeshhubale:
- Testing quota source value in automation logs 

{{ pytest: cfme/tests/automate/test_instance.py  -k 'test_quota_source_value' -vv }}

### DIFF
--- a/cfme/tests/automate/test_instance.py
+++ b/cfme/tests/automate/test_instance.py
@@ -4,6 +4,7 @@ import pytest
 
 from cfme import test_requirements
 from cfme.automate.simulation import simulate
+from cfme.services.service_catalogs import ServiceCatalogs
 from cfme.utils.log_validator import LogValidator
 from cfme.utils.update import update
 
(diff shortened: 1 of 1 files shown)


cfme/tests/automate/test_instance.py
```diff
@@ -280,3 +281,78 @@ def test_check_system_request_calls_depr_conf_mgmt(appliance, copy_instance):
...
+        name=generic_catalog_item.name
+    )
+    request_description = 'Provisioning Service [{name}] from [{name}]'.format(
+        name=service_catalogs.name)
+    provision_request = copy_quota_instance.appliance.collections.requests.instantiate(
+        description=request_description)
+    service_catalogs.order()
+    provision_request.wait_for_request(method='ui')
+    request.addfinalizer(lambda: provision_request.remove_request(method="rest"))
+    result.validate_logs()
```
kedark3: **QUESTION** I am trying to understand how are you making sure logs are valid? Does `validate_logs()` returns a boolean that you can assert? I checked https://github.com/ManageIQ/integration_tests/blob/32faa0a0fba7dde3a611bc45b3ef653285896457/cfme/utils/log_validator.py#L55 and don't see a return value. 

I might be missing something but can you elaborate how/where you `assert` match?

cfme/tests/automate/test_instance.py
```diff
@@ -280,3 +281,78 @@ def test_check_system_request_calls_depr_conf_mgmt(appliance, copy_instance):
...
+    )
+    original_instance.copy_to(domain=domain)
+
+    instance = (
+        domain.namespaces.instantiate("System")
+        .namespaces.instantiate("CommonMethods")
+        .classes.instantiate("QuotaStateMachine")
+        .instances.instantiate("quota")
+    )
+    return instance
```
jawatts: **Required **@ganeshhubale I think it is worth adding a finalizer here to clean up this domain, we want to leave the system the same as when the test started. Do you agree here

ganeshhubale: @jawatts 
Actually, ```domain``` is global fixture(which creates custom domain) and we are copying default instance to this domain and domain`s teardown is already present there.
Once we delete the domain then all the things under it got deleted.

jawatts: @ganeshhubale Awesome, thanks!

PR #8969, opened 2019-06-18
[1LP][RFR] Add test case for BZ 1650104

john-dupuy:
Adding test case for https://bugzilla.redhat.com/show_bug.cgi?id=1650104

{{ pytest: --use-template-cache --long-running cfme/tests/generic_objects/test_definitions.py::test_upload_image_generic_object_definition }}

### DIFF
--- a/cfme/tests/generic_objects/test_definitions.py
+++ b/cfme/tests/generic_objects/test_definitions.py
@@ -76,3 +76,30 @@ def test_generic_objects_class_accordion_should_display_when_locale_is_french():
         1594480
     """
     pass
+
+
+@pytest.mark.ignore_stream("5.10")
+@pytest.mark.meta(blockers=[BZ(1650104)])
+def test_upload_image_generic_object_definition(appliance):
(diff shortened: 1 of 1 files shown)


cfme/tests/generic_objects/test_definitions.py
```diff
@@ -76,3 +76,30 @@ def test_generic_objects_class_accordion_should_display_when_locale_is_french():
...
+        casecomponent: GenericObjects
+    """
+    view = navigate_to(appliance.collections.generic_object_definitions, "Add")
+    view.custom_image_file.upload_chosen_file.click()
+    # make sure the flash assertion appears
+    view.flash.assert_message("No file chosen.")
+    # click button again
+    view.custom_image_file.upload_chosen_file.click()
+    # make sure only a single message is present
+    assert len(view.flash.read()) == 1
```
digitronik: optional:
I would use only `view.flash.messages` than read those for calculating length. but totally optional.

PR #8971, opened 2019-06-18
[1LP][RFR] BZ 1698431

sshveta:
Customer BZ automation

### DIFF
--- a/cfme/services/catalogs/catalog_items/__init__.py
+++ b/cfme/services/catalogs/catalog_items/__init__.py
@@ -31,6 +31,7 @@ from cfme.utils.blockers import BZ
 from cfme.utils.pretty import Pretty
 from cfme.utils.update import Updateable
 from cfme.utils.version import LOWEST
+from cfme.utils.version import VersionPicker
 from cfme.utils.wait import wait_for
 from widgetastic_manageiq import AutomateRadioGroup
 from widgetastic_manageiq import EntryPoint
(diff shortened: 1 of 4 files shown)


widgetastic_manageiq/__init__.py
```diff
@@ -5504,13 +5504,13 @@ def fill(self, value):
...
             return False
 
-        if self.textbox.is_enabled:
-            self.browser.click(self.textbox)
-        else:
-            self.group_btn.click()
-
-        self.tree.wait_displayed("10s")
-        self.tree.click_path(*value)
+        if value:
```
digitronik: Q: value is a required argument. then why we need this?

cfme/tests/services/test_service_customer_bz.py
```diff
@@ -0,0 +1,68 @@
...
+        catalog=catalog_item.catalog,
+        description="catalog_bundle",
+        display_in=True,
+        dialog=catalog_item.dialog,
+        provisioning_entry_point=prov_entry_point
+    )
+    view = navigate_to(catalog_bundle, "Edit")
+    assert view.basic_info.provisioning_entry_point.value == ("/Service/Provisioning/StateMachines/"
+                                                              "ServiceProvision_Template/"
+                                                              "CatalogItemInitialization")
```
digitronik: I think you need to close edit window after test.

widgetastic_manageiq/__init__.py
sshveta: When value is None , It still tries to clickpath on tree . Hence this check is needed

cfme/tests/services/test_service_customer_bz.py
sshveta: Done

cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -32,29 +31,17 @@
 from cfme.utils.pretty import Pretty
 from cfme.utils.update import Updateable
 from cfme.utils.version import LOWEST
+from cfme.utils.version import VersionPicker
 from cfme.utils.wait import wait_for
 from widgetastic_manageiq import AutomateRadioGroup
+from widgetastic_manageiq import EntryPoint
```
mshriver: :+1:

cfme/tests/services/test_service_customer_bz.py
```diff
@@ -0,0 +1,69 @@
...
+                         required_fields=[['provisioning', 'template'],
+                                          ['provisioning', 'host'],
+                                          ['provisioning', 'datastore']],
+                         scope="module"),
+]
+
+
+@pytest.mark.rhv3
+@pytest.mark.tier(2)
+def test_edit_bundle_entry_point(appliance, provider, catalog_item, request):
```
mshriver: I think there are better ways to indicate customer BZ driven tests than separate test modules.  Is there any other motivation between a separate test module instead of this test function being in cfme/tests/services/test_catalog_item.py?

sshveta: @mshriver : I created a separate module because we have started working on other service customer BZ's so I thought they can also be added here . All other service files already have lot of tests .

mshriver: I think this is ultimately going to cause confusion, and make it more difficult to locate and organize test functions. I'm willing to merge this as-is, but I think you and @niyazRedhat should really consider how these modules are organized, and what goes where. Maybe split catalog items and catalog bundles into separate modules to keep them a manageable size. Ultimately this is a pretty basic CRUD test for catalog bundle.

PR #8972, opened 2019-06-19
[1LP][RFR] Scsi controller vmware

kedark3:
Purpose or Intent
=================

__Adding tests__ for making sure SCSI controller - Appliance cfme-vsphere-paravirtual-*.ova has SCSI controller 
{{pytest: cfme/tests/infrastructure/test_vmware_provider.py -k "test_appliance_scsi_control_vmware" --long-running}}

### DIFF
--- a/cfme/tests/infrastructure/test_vmware_provider.py
+++ b/cfme/tests/infrastructure/test_vmware_provider.py
@@ -1,13 +1,18 @@
 # -*- coding: utf-8 -*-
 """Manual VMware Provider tests"""
+import os
 import re
+import tarfile
 
 import fauxfactory
 import pytest
+from six.moves.urllib import request as url_request
 
 from cfme import test_requirements
 from cfme.infrastructure.host import Host
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
+from cfme.markers.env_markers.provider import ONE_PER_TYPE
+from cfme.utils import conf
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.blockers import BZ
 from cfme.utils.log import logger
@@ -54,9 +59,11 @@ def test_vmware_provider_filters(appliance, provider, soft_assert):
         soft_assert(esx_platform in all_options, "ESX Platform does not exists in options")
 
 
-@pytest.mark.manual
 @pytest.mark.tier(3)
-def test_appliance_scsi_control_vmware():
+@pytest.mark.long_running
+@pytest.mark.ignore_stream("upstream")
+@pytest.mark.provider([VMwareProvider], selector=ONE_PER_TYPE, override=True)
+def test_appliance_scsi_control_vmware(request, appliance):
     """
     Appliance cfme-vsphere-paravirtual-*.ova has SCSI controller as Para
     Virtual
@@ -66,9 +73,34 @@ def test_appliance_scsi_control_vmware():
         casecomponent: Appliance
         caseimportance: critical
         initialEstimate: 1/4h
-    #TODO: yet to test this, once done, I will add steps. Test was not written by me originally.
     """
-    pass
+    try:
+        url = (conf.cfme_data.basic_info.cfme_images_url.cfme_paravirtual_url_format
+            .format(
+                baseurl=conf.cfme_data.basic_info.cfme_images_url.baseurl,
+                series=appliance.version.series(), ver=appliance.version
+            )
+        )
+    except AttributeError:
+        pytest.skip('Skipping as one of the keys might be missing in cfme_yamls.')
+    logger.info("Downloading ova file for parvirtual vsphere scsi controller test from %s", url)
+    filename = os.path.basename(url)
+
+    url_request.urlretrieve(url, filename)
+
+    @request.addfinalizer
+    def _cleanup():
+        if os.path.exists(filename):
+            os.remove(filename)
+
+    with tarfile.open(filename) as tar:
+        desc_member = tar.getmember("desc.ovf")
+        f = tar.extractfile(desc_member)
+        content = f.read()
+    assert content, "No content could be read from desc.ovf"
+    logger.debug("Desc file contains following text:%s" % content)
+    check_string = '<rasd:ResourceSubType>VirtualSCSI</rasd:ResourceSubType>'
+    assert check_string in str(content), "Given OVA does not have paravirtual scsi controller"
 
 
 @pytest.mark.tier(1)
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -66,7 +66,6 @@ from widgetastic_manageiq import SnapshotMemorySwitch
 from widgetastic_manageiq import SummaryTable
 from widgetastic_manageiq import Table
 from widgetastic_manageiq import TimelinesView
-from widgetastic_manageiq.vm_reconfigure import CDDVDTable
 from widgetastic_manageiq.vm_reconfigure import DisksTable
 
 
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -66,6 +66,7 @@ from widgetastic_manageiq import SnapshotMemorySwitch
 from widgetastic_manageiq import SummaryTable
 from widgetastic_manageiq import Table
 from widgetastic_manageiq import TimelinesView
+from widgetastic_manageiq.vm_reconfigure import CDDVDTable
 from widgetastic_manageiq.vm_reconfigure import DisksTable
 
 


dajoRH:
*CFME QE Bot*

dajoRH:
*CFME QE Bot*

dajoRH:
Lint report for commit edc69ef7fe2346c0de3e1f300f2319f48cbc5316:

`cfme/tests/infrastructure/test_vmware_provider.py`:
- :red_circle: Line 12:1: **F401** *'cfme.fixtures.soft_assert.soft_assert' imported but unused*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -68,7 +71,33 @@ def test_appliance_scsi_control_vmware():
         initialEstimate: 1/4h
     #TODO: yet to test this, once done, I will add steps. Test was not written by me originally.
     """
-    pass
+    url = ("{}/{}/{}/cfme-vsphere-paravirtual-{}-1.x86_64.vsphere.ova"
```
digitronik: please use keying for such customization
```
url = "{baseurl}/{series}/{ver}/cfme-vsphere-paravirtual-{ver}-1.x86_64.vsphere.ova".format(
baseurl=conf.cfme_data['basic_info']['cfme_images_url']['baseurl'],
series=appliance.version.series(),
ver=appliance.version
) 
```

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -68,7 +71,33 @@ def test_appliance_scsi_control_vmware():
...
+        .format(conf.cfme_data['basic_info']['cfme_images_url']['baseurl'],
+            appliance.version.series(), appliance.version, appliance.version))
+    logger.info("Downloading ova file for parvirtual vsphere scsi controller test from %s" % url)
+    ova_file = urllib.urlopen(url)
+    filename = "cfme-vsphere-paravirtual-{}.ova".format(appliance.version)
```
digitronik: optional:
rather than creating new name I would use basename form URL
```
os.path.basename(url)
```

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -68,7 +71,33 @@ def test_appliance_scsi_control_vmware():
...
+    with open(filename, 'wb') as output:
+        output.write(ova_file.read())
+
+    @request.addfinalizer
+    def _cleanup():
+        if os.path.exists(filename):
+            os.remove(filename)
+    tar = tarfile.open(filename)
+    for member in tar.getmembers():
+        if member.name == 'desc.ovf':
```
digitronik: I think you are looking for one `member` in `tarfile`. What you think, if we just pick that member rather than iteration over `members`
```
desc_member = tar.getmember("desc.ovf")
```

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -68,7 +71,33 @@ def test_appliance_scsi_control_vmware():
         initialEstimate: 1/4h
     #TODO: yet to test this, once done, I will add steps. Test was not written by me originally.
```
digitronik: you can resolve this TODO note :+1:

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -54,9 +58,8 @@ def test_vmware_provider_filters(appliance, provider, soft_assert):
         soft_assert(esx_platform in all_options, "ESX Platform does not exists in options")
 
 
-@pytest.mark.manual
 @pytest.mark.tier(3)
-def test_appliance_scsi_control_vmware():
+def test_appliance_scsi_control_vmware(request, appliance):
```
kedark3: @jawatts  I would like to run this test only once, irrespective of provider, what marker should I use? Asking since this test dosn't use any provider.

jawatts: Dicussed offline, will need to override the provider marker.

cfme/tests/infrastructure/test_vmware_provider.py
kedark3: Makes sense.

cfme/tests/infrastructure/test_vmware_provider.py
kedark3: Yep that was on my radar :D

cfme/tests/infrastructure/test_vmware_provider.py
digitronik: @jawatts @kedark3 I think override provider marker is not the best solution here; still it will try to add vsphere provider. The basic `pytest` concept is a `global marker` will always execute for each test in the respective file. So we need proper planning.

If we like to set this test `provider independent` (test will not try to add provider) in the same file then we need to remove a provider global marker and set on the remaining test.

kedark3: @jawatts @mshriver your opinion?

jawatts: @digitronik the global pytest marker in this case is running the test 3 times, for each vpshere version. What @kedark3 wants to do here, is run it against a vpshere provider, but just once. So
```
@pytest.mark.provider([VMwareProvider], selector=ONE_PER_TYPE, override=True)`

```

kedark3: Added that with ONE_PER_TYPE

digitronik: aaaaaaaahh I was confused with `test dosn't use any provider` actually steps also pointing same. :+1: then its ok... we can overwrite provider marker for a specific use case.

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -66,9 +71,33 @@ def test_appliance_scsi_control_vmware():
         casecomponent: Appliance
         caseimportance: critical
         initialEstimate: 1/4h
-    #TODO: yet to test this, once done, I will add steps. Test was not written by me originally.
     """
-    pass
+    url = ("{baseurl}/{series}/{ver}/cfme-vsphere-paravirtual-{ver}-1.x86_64.vsphere.ova"
```
mshriver: This image name pattern should also be pulled from `cfme_data`, or `miq_version`. It is 100% related to downstream images, since paravirtual images aren't built for upstream.

The test needs to ignore upstream, as it won't work - the `series` is master.

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -66,9 +71,33 @@ def test_appliance_scsi_control_vmware():
...
+        .format(baseurl=conf.cfme_data['basic_info']['cfme_images_url']['baseurl'],
+            series=appliance.version.series(), ver=appliance.version))
+    logger.info("Downloading ova file for parvirtual vsphere scsi controller test from %s" % url)
```
mshriver: This test needs to be marked long-running, this download could take quite a while.

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -66,9 +71,33 @@ def test_appliance_scsi_control_vmware():
...
+    ova_file = urllib.urlopen(url)
+    filename = os.path.basename(url)
+    with open(filename, 'wb') as output:
```
mshriver: I think you should use `request.urlretrieve` here, instead of this pattern.

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -66,9 +71,33 @@ def test_appliance_scsi_control_vmware():
...
+        output.write(ova_file.read())
+
+    @request.addfinalizer
+    def _cleanup():
+        if os.path.exists(filename):
+            os.remove(filename)
+    tar = tarfile.open(filename)
```
mshriver: Please use the context manager here when reading the tar archive, so that it gets closed.

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -66,9 +71,33 @@ def test_appliance_scsi_control_vmware():
...
+    desc_member = tar.getmember("desc.ovf")
+    f = tar.extractfile(desc_member)
+    content = f.read()
+    assert content, "No content could be read from desc.ovf"
+    logger.debug("Desc file contains following text:%s" % content)
+    scsi_controller_types = ['lsilogic', 'buslogic', 'lsilogicsas', 'virtualscsi']
+    content = content.lower()
+    counter = 0  # Count any mention of scsi_controller_type in ova XML content
+    for scsi_controller_type in scsi_controller_types:
```
mshriver: Maybe I'm reading this wrong, but I don't think that this is an accurate assertion for the test.

This is checking that three is at least one instance of any of the strings from `scsi_controller_types` in the content - which would yield a false positive for the purpose of the test.

Non-paravirtual images are built with `lsilogic` type SCSI controller, and paravirtual images are built with `virtualscsi` type SCSI controller.  This test needs to tightly assert that `virtualscsi` is the `ResourceSubType` for the `SCSI Controller`.  As I see it now, you could have downloaded the non-paravirtual image and ran this test, and it would pass, because the `lsilogic` controller would have triggered this count increasing.

Did you consider using `xml.etree` to actually parse the OVF file?

I think the minimum assertion here should be that there is one (at least) `ResourceSubType` element in the tree with a value of 'virtualscsi'.

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -66,9 +73,28 @@ def test_appliance_scsi_control_vmware():
         casecomponent: Appliance
         caseimportance: critical
         initialEstimate: 1/4h
-    #TODO: yet to test this, once done, I will add steps. Test was not written by me originally.
     """
-    pass
+    url = (conf.cfme_data['basic_info']['cfme_images_url']['cfme_paravirtual_url_format'])\
```
mshriver: This needs to be wrapped with a `pytest.skip` call if this value isn't found. Otherwise the test results in an ERROR with a KeyError.

Kind of personal opinion, but I think this looks a lot cleaner taking advantage of AttrDict. `conf.cfme_data.basic_info.cfme_images_url.cfme_paravirtual_url_format`. Note if you use this syntax, missing keys will result in AttributeError, not KeyError.

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -66,9 +73,28 @@ def test_appliance_scsi_control_vmware():
...
+        .format(baseurl=conf.cfme_data['basic_info']['cfme_images_url']['baseurl'],
+            series=appliance.version.series(), ver=appliance.version)
+    logger.info("Downloading ova file for parvirtual vsphere scsi controller test from %s" % url)
```
mshriver: The value to format should be passed as an argument, not using the `%` syntax.

`logger.info('message: %s', url)`

PR #8980, opened 2019-06-21
[1LP][RFR] Automating Ansible Service Verbosity tests

sbulage:
Removed few already automated tests `test_embed_tower_add_public_repo`  and added New test for `test_embed_tower_exec_play_stdout, test_embed_tower_enhanced_playbook_debug and test_service_ansible_verbosity` -->  Ansible Service with Verbosity.

PRT Run:

{{ pytest: cfme/tests/ansible/test_embedded_ansible_services.py -k test_service_ansible_verbosity -svvvv --long-running }}

### DIFF
--- a/cfme/tests/ansible/test_embedded_ansible_services.py
+++ b/cfme/tests/ansible/test_embedded_ansible_services.py
@@ -13,6 +13,8 @@ from cfme.markers.env_markers.provider import ONE_PER_TYPE
 from cfme.services.myservice import MyService
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.blockers import BZ
+from cfme.utils.log import logger
+from cfme.utils.log_validator import LogValidator
 from cfme.utils.update import update
 from cfme.utils.wait import wait_for
 
@@ -637,12 +639,26 @@ def test_embed_tower_exec_play_against(
 
 
 @pytest.mark.tier(2)
+@pytest.mark.parametrize(
+    "verbosity",
+    [
+        "1 (Verbose)",
+        "2 (More Verbose)",
+        "3 (Debug)",
+        "4 (Connection Debug)",
+        "5 (WinRM Debug)",
+        "0 (Normal)",
+    ],
+)
 @pytest.mark.meta(automates=[BZ(1460788)])
 def test_service_ansible_verbosity(
+    appliance,
+    request,
     ansible_catalog_item,
     ansible_service_catalog,
     ansible_service_request,
     ansible_service,
+    verbosity,
 ):
     """Check if the different Verbosity levels can be applied to service and
     monitor the std out
@@ -657,24 +673,39 @@ def test_service_ansible_verbosity(
         initialEstimate: 1/6h
         tags: ansible_embed
     """
+    pattern = '"verbosity"=>{}'.format(verbosity[0])
     with update(ansible_catalog_item):
-        ansible_catalog_item.provisioning = {"verbosity": "3 (Debug)"}
-        ansible_catalog_item.retirement = {"verbosity": "3 (Debug)"}
+        ansible_catalog_item.provisioning = {"verbosity": verbosity}
+        ansible_catalog_item.retirement = {"verbosity": verbosity}
+    # Log Validator
+    log = LogValidator(
+        "/var/www/miq/vmdb/log/evm.log", matched_patterns=[pattern]
+    )
+    # Start Log check or given pattern
+    log.fix_before_start()
+
+    @request.addfinalizer
+    def _revert():
+        with update(ansible_catalog_item):
+            ansible_catalog_item.provisioning = {"verbosity": "0 (Normal)"}
+            ansible_catalog_item.retirement = {"verbosity": "0 (Normal)"}
+
+        service = MyService(appliance, ansible_catalog_item.name)
+        if ansible_service_request.exists():
+            ansible_service_request.wait_for_request()
+            appliance.rest_api.collections.service_requests.action.delete(id=service_id.id)
+        if service.exists:
+            service.delete()
 
     ansible_service_catalog.order()
     ansible_service_request.wait_for_request()
-    view = navigate_to(ansible_service, "Details")
-    # '3' denotes Debug level verbosity which is passed in the catalog creation.
-    assert "3" == view.provisioning.details.get_text_of("Verbosity")
-    assert "3" == view.retirement.details.get_text_of("Verbosity")
-    # Resetting verbosity value back to '0'
-    with update(ansible_catalog_item):
-        ansible_catalog_item.provisioning = {"verbosity": "0 (Normal)"}
-        ansible_catalog_item.retirement = {"verbosity": "0 (Normal)"}
+    request_descr = "Provisioning Service [{0}] from [{0}]".format(ansible_catalog_item.name)
+    service_id = appliance.rest_api.collections.service_requests.get(description=request_descr)
+    # Searching string '"verbosity"=>0' in evm.log as Standard Output is being
+    # logging in evm.log
+    log.validate_logs()
+    logger.info("Pattern found {}".format(log.matched_patterns))
 
-    ansible_service_catalog.order()
-    ansible_service_request.wait_for_request()
     view = navigate_to(ansible_service, "Details")
-    # '0' denotes  level verbosity which is passed in the catalog creation.
-    assert "0" == view.provisioning.details.get_text_of("Verbosity")
-    assert "0" == view.retirement.details.get_text_of("Verbosity")
+    assert verbosity[0] == view.provisioning.details.get_text_of("Verbosity")
+    assert verbosity[0] == view.retirement.details.get_text_of("Verbosity")
--- a/cfme/tests/ansible/test_embedded_ansible_services.py
+++ b/cfme/tests/ansible/test_embedded_ansible_services.py
@@ -634,3 +634,43 @@ def test_embed_tower_exec_play_against(
 
     view = navigate_to(ansible_service, "Details")
     assert view.provisioning.results.get_text_of("Status") == "successful"
+
+
+@pytest.mark.tier(2)
+@pytest.mark.meta(automates=[BZ(1460788)])
+def test_service_ansible_verbosity(
+    request,
+    ansible_catalog_item,
+    ansible_service_catalog,
+    ansible_service_request,
+    ansible_service,
+):
+    """Check if the different Verbosity levels can be applied to service and
+    monitor the std out
+
+    Bugzilla:
+        1460788
+
+    Polarion:
+        assignee: sbulage
+        casecomponent: Ansible
+        caseimportance: medium
+        initialEstimate: 1/6h
+        tags: ansible_embed
+    """
+    with update(ansible_catalog_item):
+        ansible_catalog_item.provisioning = {"verbosity": "3 (Debug)"}
+        ansible_catalog_item.retirement = {"verbosity": "3 (Debug)"}
+
+    @request.addfinalizer
+    def _revert():
+        with update(ansible_catalog_item):
+            ansible_catalog_item.provisioning = {"verbosity": "0 (Normal)"}
+            ansible_catalog_item.retirement = {"verbosity": "0 (Normal)"}
+
+    ansible_service_catalog.order()
+    ansible_service_request.wait_for_request()
+    view = navigate_to(ansible_service, "Details")
+    # '3' denotes Debug level verbosity which is passed in the catalog creation.
+    assert "3" == view.provisioning.details.get_text_of("Verbosity")
+    assert "3" == view.retirement.details.get_text_of("Verbosity")
--- a/cfme/tests/ansible/test_embedded_ansible_services.py
+++ b/cfme/tests/ansible/test_embedded_ansible_services.py
@@ -673,36 +673,37 @@ def test_service_ansible_verbosity(
         initialEstimate: 1/6h
         tags: ansible_embed
     """
+    # Adding index 0 which will give pattern for e.g. pattern = "verbosity"=>0.
     pattern = '"verbosity"=>{}'.format(verbosity[0])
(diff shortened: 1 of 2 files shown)


cfme/tests/ansible/test_embedded_ansible_services.py
```diff
@@ -634,3 +634,42 @@ def _revert():
...
+    ansible_catalog_item,
+    ansible_service_catalog,
+    ansible_service_request,
+    ansible_service,
+):
+    """Check if the different Verbosity levels can be applied to service and
+    monitor the std out
+
+    Bugzilla:
+        1460788
```
digitronik: please add automate marker for this BZ. as its RFE and you are automating this.

sbulage: @digitronik I have updated PR with your changes.

cfme/tests/ansible/test_embedded_ansible_services.py
```diff
@@ -634,3 +634,43 @@ def _revert():
 
     view = navigate_to(ansible_service, "Details")
     assert view.provisioning.results.get_text_of("Status") == "successful"
+
+
+@pytest.mark.tier(2)
+@pytest.mark.meta(automates=[BZ(1460788)])
```
jawatts: :+1:

cfme/tests/ansible/test_embedded_ansible_services.py
```diff
@@ -634,3 +634,47 @@ def _revert():
...
+        ansible_catalog_item.provisioning = {"verbosity": "3 (Debug)"}
+        ansible_catalog_item.retirement = {"verbosity": "3 (Debug)"}
+
+    ansible_service_catalog.order()
+    ansible_service_request.wait_for_request()
+    view = navigate_to(ansible_service, "Details")
+    # '3' denotes Debug level verbosity which is passed in the catalog creation.
+    assert "3" == view.provisioning.details.get_text_of("Verbosity")
+    assert "3" == view.retirement.details.get_text_of("Verbosity")
+    # Resetting verbosity value back to '0'
```
mshriver: This is an assumption based on current settings - I would re-work the test a bit to actually assert this default expected configuration value, if we're going to consider it a default, before changing it to `3`

cfme/tests/ansible/test_embedded_ansible_services.py
```diff
@@ -634,3 +634,47 @@ def _revert():
 
     view = navigate_to(ansible_service, "Details")
     assert view.provisioning.results.get_text_of("Status") == "successful"
+
+
+@pytest.mark.tier(2)
+@pytest.mark.meta(automates=[BZ(1460788)])
+def test_service_ansible_verbosity(
```
jawatts: @sbulage After talking with Mike, this should be paramertized from `5-0` to test all the available options. Also add the finalizer back in to reset it to the default string as @mshriver suggested below. Sorry to steer your wrong with the logic of setting it back to 0 below.

cfme/tests/ansible/test_embedded_ansible_services.py
```diff
@@ -634,3 +634,47 @@ def _revert():
...
+    ansible_catalog_item,
+    ansible_service_catalog,
+    ansible_service_request,
+    ansible_service,
+):
+    """Check if the different Verbosity levels can be applied to service and
+    monitor the std out
```
jawatts: I don't see the std out monitored in this test?

cfme/tests/ansible/test_embedded_ansible_services.py
sbulage: Added Parametrization for all available Verbosity. Resetting back to `0 (Normal)`.

cfme/tests/ansible/test_embedded_ansible_services.py
sbulage: Stdout is logging into `evm.log` and using `LogValidator` finding expected pattern i.e. `verbosity` value and asserting it in UI as well.

cfme/tests/ansible/test_embedded_ansible_services.py
sbulage: Yes, @mshriver I agree with you, keeping `0 (Normal)` as a default and resetting back in finalizer.
Also, validating verbosity pattern in `evm.log` would give more information that `Verbosity` set in catalog item is being passed to execute playbook and based on that we are `Verbosity` value in the ordered `Service`.

cfme/tests/ansible/test_embedded_ansible_services.py
```diff
@@ -634,3 +636,76 @@ def _revert():
...
+        1460788
+
+    Polarion:
+        assignee: sbulage
+        casecomponent: Ansible
+        caseimportance: medium
+        initialEstimate: 1/6h
+        tags: ansible_embed
+    """
+    pattern = '"verbosity"=>{}'.format(verbosity[0])
```
jawatts: **Required** Please add a comment here on why you are grabbing index 0

cfme/tests/ansible/test_embedded_ansible_services.py
```diff
@@ -634,3 +636,76 @@ def _revert():
...
+            ansible_service_request.wait_for_request()
+            appliance.rest_api.collections.service_requests.action.delete(id=service_id.id)
+        if service.exists:
+            service.delete()
+
+    ansible_service_catalog.order()
+    ansible_service_request.wait_for_request()
+    request_descr = "Provisioning Service [{0}] from [{0}]".format(ansible_catalog_item.name)
+    service_id = appliance.rest_api.collections.service_requests.get(description=request_descr)
+    # Searching string '"verbosity"=>0' in evm.log as Standard Output is being
```
jawatts: Not searching 0 everytime here right? It should reflect what the user sets?

cfme/tests/ansible/test_embedded_ansible_services.py
```diff
@@ -634,3 +636,76 @@ def _revert():
...
+        ansible_catalog_item.provisioning = {"verbosity": verbosity}
+        ansible_catalog_item.retirement = {"verbosity": verbosity}
+    # Log Validator
+    log = LogValidator(
+        "/var/www/miq/vmdb/log/evm.log", matched_patterns=[pattern]
+    )
+    # Start Log check or given pattern
+    log.fix_before_start()
+
+    @request.addfinalizer
```
jawatts: So, a finalizer is going to be called after each test execution is finished. So in this case, just because it is parametrized, it is still going to run everytime because each parametrization is a separate test. 

**Required** I think the best bet here is to create a module scoped fixtures that reads the default setting in the CFME gui, yield, and sets it back after all tests have been executed. That way, this is only run once.

cfme/tests/ansible/test_embedded_ansible_services.py
sbulage: It's just a comment to help to understand why I am using this parameter, nothing is being related to verbosity 0.

cfme/tests/ansible/test_embedded_ansible_services.py
sbulage: Simple logic:
```
In [1]: verbosity = "0 (Normal)"

In [2]: print('"verbosity"=>{}'.format(verbosity[0]))
"verbosity"=>0

```

cfme/tests/ansible/test_embedded_ansible_services.py
sbulage: Not sure what you are telling, got confused here, where should I have finalizer there or not .?

sbulage: Considered the Verbosity `0 (Normal)`  as a default and setting back after every test execution in the finalizer.

cfme/tests/ansible/test_embedded_ansible_services.py
sbulage: should I add above lines there .? What do you think .?

cfme/tests/ansible/test_embedded_ansible_services.py
sbulage: It will gonna change as value of Verbosity is being changed.

jawatts: @sbulage I think just adding "Searching for "verbosity"=>0 (example) in evm.log as Standard Output is being logged to evm.log"

cfme/tests/ansible/test_embedded_ansible_services.py
jawatts: Just a simple comment why you are grabbing the index so if someone takes over this FA, it is clear to them.

sbulage: Adding index 0 which will give pattern for e.g. `pattern = "verbosity"=>0`, Doable .? @jawatts

cfme/tests/ansible/test_embedded_ansible_services.py
jawatts: So two part problem here:

1) We are assuming Verbosity 0 is the default, I don't think we should make that assumption. What if it is 0 right now, but it is changed to 3 in the future. We should read what the default is set to and use that to rest.

2) So for the finalizer. Every time this test is executed, the finalizer will run. But how it is currently written, it is not setting back to 0. It is setting it to whatever the verbosity is parametrized for this instance of the test. What really needs to happen, is to read what the default is set to and use that in the finalizer. But, it makes no sense to do that every since parameterization of this run. So what I am suggesting, if create a module scope fixture that:
- reads the default and saves that to a variable
- yields
- Set the default back after all iterations of the test are run.

cfme/tests/ansible/test_embedded_ansible_services.py
jawatts: :+1: that be a perfect comment

cfme/tests/ansible/test_embedded_ansible_services.py
sbulage: Okay Fixture which will read and reset it in finalizer. Got it.

cfme/tests/ansible/test_embedded_ansible_services.py
sbulage: Okay

cfme/tests/ansible/test_embedded_ansible_services.py
sbulage: @jawatts I have thought on this as all above procedure is part of code and not being used anywhere in the code, so keeping it in the test would give more impact.

Also, shuffled order of `Verbosity`, which will at the end add default `Verbosity` i.e. `"0 (Normal)"`. Which will act as `Finalizer`.

jawatts: @sbulage We are still assuming that `0` is the default and will always be the default. I don't think we should make the assumption. We should read what it is in the GUI first and reset it to that. Also, if the test fails at any point before `0`, that parameterization will not run. Also how the finalizer is written now, it is going to run every single test execution. So this finalizer is running 6 times. I don't think that is needed, that is why I suggested the fixture.

sbulage: Yeah, `Finalizer` will run after every test `parameter`, that's why I have removed that code from `Finalizer`. Updating PR now.

cfme/tests/ansible/test_embedded_ansible_services.py
```diff
@@ -634,3 +636,73 @@ def _revert():
...
+        if ansible_service_request.exists():
+            ansible_service_request.wait_for_request()
+            appliance.rest_api.collections.service_requests.action.delete(id=service_id.id)
+        if service.exists:
+            service.delete()
+
+    ansible_service_catalog.order()
+    ansible_service_request.wait_for_request()
+    request_descr = "Provisioning Service [{0}] from [{0}]".format(ansible_catalog_item.name)
+    service_id = appliance.rest_api.collections.service_requests.get(description=request_descr)
```
mshriver: Can you please add a comment here that this is used by the finalizer to delete the request? Its otherwise unused, and its a bit confusing why its being called here (after the finalizer has been defined).  Its also not the ID, but an object with the `id` attribute.

sbulage: Okay. Updating it.

sbulage:
> Looks good, thanks for working with @jawatts on getting an easy to read and maintain pattern for iterating through the verbosity levels.
> 
> I think some future improvements to the test could include:
> 
> 1. Reading initial state of the verbosity setting before the test, so we don't assume the default
> 2. Resetting to that default instead of resetting to `0` always
> 3. Additional log checks that the verbosity setting is actually taking effect (not just the setting is included in the log)

Yes, Mike agree with you but when you create new catalog_item, by default `0 Verbosity` is selected no a empty pattern and we are changing it to available patterns from `1-5 Verbosity`. So, in this case also we have to assume `0 Verbosity` as a `Default` one., if that makes sense .?

mshriver:
@sbulage Understood - but that is the default that MIQ uses _NOW_.  There's no guarantee that will _ALWAYS_ be the default, and if MIQ changes to level `1`, the test will still statically be setting it to `0` at the end of the test. It won't be restoring the value that was present at the start of the test, the MIQ default.

` So, in this case also we have to assume 0 Verbosity as a Default one., if that makes sense .?`

That's the problem - its an assumption that we are making, instead of actually reading what MIQ has set at the beginning of the test.

What if we started configuring our downstream appliances with `2` as a default value for this setting via advanced settings? Or just updated the module/session scoped global fixture that creates the catalog item used here? Then this test would be un-doing our intentional configuration, impacting tests that run after it.

sbulage:
> What if we started configuring our downstream appliances with `2` as a default value for this setting via advanced settings? Or just updated the module/session scoped global fixture that creates the catalog item used here? Then this test would be un-doing our intentional configuration, impacting tests that run after it.

@mshriver Right now, in the module scoped global fixture, which is keeping `0 (Normal)` Verbosity while creating `ansible_catalog_item` (global fixture). I think this is enough that would keep out intention of this test. ?

PR #8985, opened 2019-06-21
[1LP][RFR] Refactor schedules

valaparthvi:
{{ pytest: cfme/tests/intelligence/reports/test_crud.py::test_reports_crud_schedule_for_base_report_once cfme/tests/intelligence/reports/test_crud.py::test_crud_custom_report_schedule cfme/tests/intelligence/reports/test_reports_schedules.py --use-template-cache -sqvvv }}

### DIFF
--- a/cfme/intelligence/reports/reports.py
+++ b/cfme/intelligence/reports/reports.py
@@ -386,38 +386,23 @@ class Report(BaseEntity, Updateable):
     def saved_reports(self):
         return self.collections.saved_reports
 
-    def create_schedule(self, name=None, description=None, active=None,
-            timer=None, from_email=None, emails=None, email_options=None):
-
+    def create_schedule(self, **data):
         view = navigate_to(self, "ScheduleReport")
-        view.fill({
-            "name": name,
-            "description": description,
-            "active": active,
-            "run": timer.get("run"),
-            "time_zone": timer.get("time_zone"),
-            "starting_date": timer.get("starting_date"),
-            "hour": timer.get("hour"),
-            "minute": timer.get("minute"),
-            "emails_send": bool(emails),
-            "from_email": from_email,
-            "emails": emails,
-            "send_if_empty": email_options.get("send_if_empty"),
-            "send_txt": email_options.get("send_txt"),
-            "send_csv": email_options.get("send_csv"),
-            "send_pdf": email_options.get("send_pdf")
-        })
+        if "email" in data:
+            data["email"]["emails_send"] = True
+        view.fill(data)
 
         view.add_button.click()
         view.flash.assert_no_error()
-
+        data["report_filter"] = {
+            "filter_type": self.company_name,
+            "subfilter_type": self.subtype,
+            "report_type": self.menu_name,
+        }
         schedule = self.appliance.collections.schedules.instantiate(
-            name=name or self.menu_name,
-            description=description,
-            report_filter=(self.company_name, self.subtype, self.menu_name),
-            active=active,
-            emails=emails,
-            email_options=email_options,
+            name=data.pop("name", self.menu_name),
+            description=data.pop("description", self.menu_name),
+            **data
         )
 
         assert schedule.exists
--- a/cfme/intelligence/reports/reports.py
+++ b/cfme/intelligence/reports/reports.py
@@ -399,33 +399,25 @@ class Report(BaseEntity, Updateable):
         view = navigate_to(self, "ScheduleReport")
         if email:
             email["emails_send"] = True
-
-        data = {
-            "name": name,
-            "description": description,
-            "active": active,
-            "report_filter": {
+        schedule = self.appliance.collections.schedules.instantiate(
+            name=name or self.menu_name,
+            description=description or self.menu_name,
+            active=active,
+            report_filter={
                 "filter_type": self.company_name,
                 "subfilter_type": self.subtype,
                 "report_type": self.menu_name,
             },
-            "timer": timer,
-            "email": email,
-            "email_options": email_options,
-        }
-        view.fill(data)
+            email=email,
+            email_options=email_options
+        )
+        view.fill(schedule.fill_dict)
 
         if cancel:
             view.cancel_button.click()
 
         view.add_button.click()
         view.flash.assert_no_error()
-
-        schedule = self.appliance.collections.schedules.instantiate(
-            name=data.pop("name") or self.menu_name,
-            description=data.pop("description") or self.menu_name,
-            **data
-        )
 
         assert schedule.exists
         return schedule
--- a/cfme/intelligence/reports/reports.py
+++ b/cfme/intelligence/reports/reports.py
@@ -415,8 +415,8 @@ class Report(BaseEntity, Updateable):
 
         if cancel:
             view.cancel_button.click()
-
-        view.add_button.click()
+        else:
+            view.add_button.click()
         view.flash.assert_no_error()
 
         assert schedule.exists
--- a/cfme/intelligence/reports/reports.py
+++ b/cfme/intelligence/reports/reports.py
@@ -386,22 +386,44 @@ class Report(BaseEntity, Updateable):
     def saved_reports(self):
         return self.collections.saved_reports
 
-    def create_schedule(self, **data):
+    def create_schedule(
+        self,
+        name=None,
+        description=None,
+        active=True,
+        timer=None,
+        email=None,
+        email_options=None,
+        cancel=False,
+    ):
         view = navigate_to(self, "ScheduleReport")
-        if "email" in data:
-            data["email"]["emails_send"] = True
+        if email:
+            email["emails_send"] = True
+
+        data = {
+            "name": name,
+            "description": description,
+            "active": active,
+            "report_filter": {
+                "filter_type": self.company_name,
+                "subfilter_type": self.subtype,
+                "report_type": self.menu_name,
+            },
+            "timer": timer,
+            "email": email,
+            "email_options": email_options,
+        }
         view.fill(data)
+
+        if cancel:
+            view.cancel_button.click()
 
         view.add_button.click()
         view.flash.assert_no_error()
-        data["report_filter"] = {
-            "filter_type": self.company_name,
-            "subfilter_type": self.subtype,
-            "report_type": self.menu_name,
-        }
+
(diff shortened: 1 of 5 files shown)


dajoRH:
Lint report for commit dc3fce4e169b54fd500dd018f1c01f777b564535:

`cfme/intelligence/reports/schedules.py`:
- :red_circle: Line 62:11: **N801** *class names should use CapWords convention*
- :red_circle: Line 76:11: **N801** *class names should use CapWords convention*
- :red_circle: Line 83:11: **N801** *class names should use CapWords convention*
- :red_circle: Line 90:11: **N801** *class names should use CapWords convention*

**No commit flaws detected**.

*CFME QE Bot*

cfme/intelligence/reports/schedules.py
```diff
@@ -204,43 +208,18 @@ class ScheduleCollection(BaseCollection):
...
-                "send_pdf": email_options.get("send_pdf"),
-            }
-        )
-
+        view.fill(data)
         view.add_button.click()
         view = schedule.create_view(ScheduleDetailsView)
         assert view.is_displayed
-        view.flash.assert_success_message('Schedule "{}" was added'.format(name))
+        view.flash.assert_success_message(
```
sshveta: This (assert flash message) can be moved to the test

cfme/intelligence/reports/schedules.py
```diff
@@ -204,43 +208,18 @@ class ScheduleCollection(BaseCollection):
...
+            'Schedule "{}" was added'.format(data.get("name"))
+        )
         return schedule
```
sshveta: The views and assertions can be moved to tests and add_button in after_fill() method  in NewScheduleView(SchedulesFormCommon).

              view.fill(data)
              return self.instantiate(**data)

cfme/intelligence/reports/schedules.py
```diff
@@ -148,18 +161,9 @@ def __str__(self):
     report_filter = attr.ib()
     active = attr.ib(default=None)
     timer = attr.ib(default=None)
-    from_email = attr.ib(default=None)
```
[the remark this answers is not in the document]
valaparthvi: I have moved `from_email` to nested view `email`. If you look at the yaml, you will find it under `email` key.

cfme/intelligence/reports/schedules.py
valaparthvi: Is it really necessary to move to `after_fill`? 
 I agree to moving assertions to test, but I am not very convinced about using `after_fill` just to click the add_button. Is there a reason for doing this?

valaparthvi:
@sshveta  Moved all assertions to `test_reports_schedule_crud`. Added `assert_no_error` in place of them.

cfme/intelligence/reports/schedules.py
sshveta: Purpose of after_fill method is take/click actions on buttons on fill form .It is automatically called from the fill method .
So yes it is a good practice to click such buttons in after_fill .
Let's take @mshriver  opinion too

valaparthvi: Hey @sshveta this is the form I am dealing with, it is a simple form. I think you might be suggesting this because I have used Nested Views. I only used them so that it would be easier to deal with.
Screenshot from 2019-06-27 14-15-06

mshriver: I don't think I would use `after_fill` in this context.  We use `after_fill` for multi-part forms (lifecycle provisioning, migration wizard) where each 'part' of the form is modeled with a nested view, and the views click the action button to get to the next part of the form.

Here, we have a single-page form, and using `after_fill` would be breaking from our current pattern of clicking the `Add` or `Cancel` button within the create method.  This allows `create` methods to take a `cancel` kwarg, and cancel after filling the form. This kwarg is generally used in CRUD tests to trigger and validate a flash message.

In this PR, you have removed the named kwargs on the create methods in favor of `**data`, which at this time I don't think is a good pattern to follow. These create methods are in a way duplicating the instance attributes of the created entity that are passed to its constructor - but they provide useful hinting and defaults when using the methods.

@izapolsk @jawatts Thoughts on these two patterns? Do we want to change up how we're modeling forms?
(1. use of `after_fill` on single-page forms, 2. use of `**data`)

valaparthvi: @sshveta @mshriver @jawatts  About using `**data`, I changed that to explicitly defined arguments and I also made some changes to [docstring](https://github.com/ManageIQ/integration_tests/pull/8985/files#diff-07a83de3f6b8c9be43e1d7392d6a1ae3R143) for better understanding.

cfme/intelligence/reports/reports.py
```diff
@@ -386,38 +386,45 @@ def delete(self, cancel=False):
...
-        })
+            "report_filter": {
+                "filter_type": self.company_name,
+                "subfilter_type": self.subtype,
+                "report_type": self.menu_name,
+            },
+            "timer": timer,
+            "email": email,
+            "email_options": email_options,
+        }
```
sshveta: Looking good @valaparthvi .
Optional : the data part can be done in fill_dict method to be consistent with some other code .
Example : https://github.com/ManageIQ/integration_tests/blob/master/cfme/v2v/conversion_hosts.py#L142

valaparthvi: @sshveta implemented those changes.

cfme/intelligence/reports/reports.py
```diff
@@ -386,39 +386,38 @@ def delete(self, cancel=False):
...
+            },
+            email=email,
+            email_options=email_options
         )
+        view.fill(schedule.fill_dict)
+
+        if cancel:
+            view.cancel_button.click()
+
+        view.add_button.click()
```
mshriver: This needs to be in an else clause, else `cancel=True` would try to click cancel, and then add, and result in `NoSuchElement`

cfme/intelligence/reports/schedules.py
```diff
@@ -54,33 +55,45 @@ class SchedulesFormCommon(CloudIntelReportsView):
...
-    emails = AlertEmail()
-    send_if_empty = Checkbox("send_if_empty")
-    send_txt = Checkbox("send_txt")
-    send_csv = Checkbox("send_csv")
-    send_pdf = Checkbox("send_pdf")
     # Buttons
     cancel_button = Button("Cancel")
 
+    @View.nested
+    class timer(View):                      # noqa
```
mshriver: There should only be two spaces before the `#` character for an in-line comment. These should not be so far to the right.

cfme/intelligence/reports/schedules.py
```diff
@@ -148,33 +160,20 @@ def __str__(self):
...
-            .query(schedules.name)\
-            .filter(schedules.name == self.name)\
-            .count() > 0
-
     def update(self, updates):
         view = navigate_to(self, "Edit")
         changed = view.fill(updates)
         if changed:
             view.save_button.click()
+            view.flash.assert_no_error()
```
mshriver: Why only assert this when the form is saved? I think its good to keep this outside of this `if/else` block so that its always asserted.

cfme/intelligence/reports/schedules.py
```diff
@@ -148,33 +160,20 @@ def __str__(self):
...
-        view = self.create_view(ScheduleDetailsView, override=updates, wait='10s')
-        view.flash.assert_no_error()
-        if changed:
-            view.flash.assert_message(
-                'Schedule "{}" was saved'.format(updates.get("name", self.name)))
-        else:
-            view.flash.assert_message(
-                'Edit of Schedule "{}" was cancelled by the user'.format(self.name))
+
+        view = self.create_view(ScheduleDetailsView)
```
mshriver: Please use `wait` argument of `create_view()` to trigger an automatic wait_is_displayed call.

mshriver: Why is `override=updates` being removed from this create_view call? If the view's `is_displayed` methods depend on entity attributes that can be changed, then this will never be able to assert the view as it'll be using the original context object.

cfme/intelligence/reports/schedules.py
```diff
@@ -198,49 +197,56 @@ def enabled(self):
...
+            report_filter=report_filter,
+            email=email,
+            email_options=email_options
         )
 
+        view = navigate_to(self, "Add")
+        view.fill(schedule.fill_dict)
+        if cancel:
+            view.cancel_button.click()
         view.add_button.click()
```
mshriver: Again - if `cancel` was `True`, this will try to click add button after canceling the form.

cfme/tests/intelligence/reports/test_reports_schedules.py
```diff
@@ -157,7 +161,7 @@ def test_reports_create_schedule_send_report(smtp_test, schedule):
             1. Queueing the schedule must send the report via email to all the users.
     """
     schedule.queue()
-    emails_sent = ",".join(schedule.emails)
+    emails_sent = ",".join(schedule.email.get("to_emails"))
```
mshriver: This will cause a `TypeError`, the default for the `get` call has to be an iterable.

valaparthvi: I am not sure I understand what you're saying. Are you saying  I should assign a default value to the `get` call?
`schedule.email` is an iterable and so is `to_emails` and both are non-empty iterables, so at least for this case, I don't think we would be getting any TypeError.

cfme/intelligence/reports/schedules.py
valaparthvi: > Please use `wait` argument of `create_view()` to trigger an automatic wait_is_displayed call.

If I add `wait` in the `create_view()`, I wouldn't have anything left to do with that view. I checked other `update()` methods, where the view is used to assert a flash message, but since we have stopped asserting flash messages inside the view, I am not sure what to do with the view that was just created. Suggestions?

valaparthvi:
Hey @mshriver Thank you for the review! There were plenty of things that I had missed.

cfme/tests/intelligence/reports/test_reports_schedules.py
mshriver: correct, the `get()` call needs an iterable as its default, in case 'to_emails' isn't set.

mshriver:
Great work @valaparthvi

PR #8987, opened 2019-06-24
[1LP][RFR] added template cleanup on provider

lina-nikiforova:
Thanks @valaparthvi for pointing to this: templates were left on the provider after `test_publish_vm_to_template`.

{{ pytest: -v cfme/tests/infrastructure/test_publish_vm_to_template.py::test_publish_vm_to_template }}

### DIFF
--- a/cfme/tests/infrastructure/test_publish_vm_to_template.py
+++ b/cfme/tests/infrastructure/test_publish_vm_to_template.py
@@ -61,6 +61,10 @@ def test_publish_vm_to_template(request, setup_provider, vm_crud):
     template_name = random_vm_name(context='pblsh')
     template = vm_crud.publish_to_template(template_name)
 
-    request.addfinalizer(template.delete)
+    @request.addfinalizer
+    def _cleanup():
+        template.delete()
+        # also delete the template from the provider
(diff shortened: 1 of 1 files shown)


cfme/tests/infrastructure/test_publish_vm_to_template.py
```diff
@@ -61,6 +61,10 @@ def test_publish_vm_to_template(request, setup_provider, vm_crud):
     template_name = random_vm_name(context='pblsh')
     template = vm_crud.publish_to_template(template_name)
 
-    request.addfinalizer(template.delete)
+    @request.addfinalizer
+    def _cleanup():
+        template.delete()
+        # also delete the template from the provider
+        template.mgmt.delete()
```
valaparthvi: Question: What if you just delete the template using wrapanapi? Why first remove it from the appliance and then from provider, if all we want is to delete the template?

digitronik: As per my observation @valaparthvi, we use the same appliance for multiple tests... better tear down is clean from the appliance and then from the provider else, we need provider refresh.

PR #8989, opened 2019-06-24
[1LP][RFR] Adding Multi-Region FA requirement and appropriate tests

izapolsk:
Multi-Region FA.

This is first PR is the chain of PRs with tests related to Multi-Region/Central Administration FAs

### DIFF
--- a/cfme/test_requirements.py
+++ b/cfme/test_requirements.py
@@ -120,12 +120,6 @@ customer_stories = pytest.mark.requirement(
     assignee_id='ndhandre',
 )
 
-multi_tenancy = pytest.mark.requirement(
-    "MIQ Multi-Tenancy",
-    description='Support for Tenants in MIQ/CFME',
-    assignee_id='nachandr',
-)
-
 chargeback = pytest.mark.requirement(
     "Chargeback",
     description='Chargeback rates, calculations, and reports',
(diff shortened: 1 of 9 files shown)


izapolsk:
@john-dupuy, I can add it if you think it is necessary. from my point, it isn't crucial since it doesn't reduce readability or affect any .rules

cfme/test_requirements.py
```diff
@@ -420,3 +420,9 @@ def test_quota_alert():
     description='Support of VMWare in MIQ/CFME',
     assignee_id='kkulkarn'
 )
+
+multi_region = pytest.mark.requirement(
```
mshriver: This file is _roughly_ alphabetical, mind putting this in the middle of the list? (`multi-tenancy` is out of place too if you want to move it...)

cfme/tests/cloud_infra_common/test_power_control_manual.py
```diff
@@ -254,7 +260,6 @@ def test_power_operations_from_global_region():
             5.
             6. Power operations applicable for this vm should be working as expected
     """
-    pass
```
mshriver: I think I would like to require this `pass` statement on manual test cases

cfme/tests/cloud_infra_common/test_retirement.py
```diff
@@ -294,3 +296,27 @@ def test_resume_retired_instance(retire_vm, provider, remove_date):
 
     verify_retirement_date(retire_vm, expected_date=retire_date if retire_date else 'Never')
     assert retire_vm.is_retired is False
+
+
+@pytest.mark.manual('manualonly')
```
mshriver: We can't automate this?

izapolsk: good point, removing. everything in my FA is gonna be automated.

cfme/tests/cloud_infra_common/test_power_control_manual.py
izapolsk: since I have two votes against I obey :)

PR #8990, opened 2019-06-25
[1LP][RFR] Automate manual tests to see VM is attached to Ansible Service.

sbulage:
PRT Run:

{{ pytest: cfme/tests/ansible/test_embedded_ansible_services.py -k test_ansible_service_linked_vm -svvvv --long-running }}

### DIFF
--- a/cfme/tests/ansible/test_embedded_ansible_services.py
+++ b/cfme/tests/ansible/test_embedded_ansible_services.py
@@ -95,11 +95,8 @@ def ansible_policy_linked_vm(appliance, new_vm, ansible_linked_vm_action):
     new_vm.assign_policy_profiles(policy_profile.description)
     yield
 
-    if policy.exists:
-        policy.unassign_events("Tag Complete")
-        new_vm.unassign_policy_profiles(policy_profile.description)
-        policy_profile.delete_if_exists()
-        policy.delete()
+    policy_profile.delete_if_exists()
+    policy.delete_if_exists()
 
 
 @pytest.fixture
--- a/cfme/tests/ansible/test_embedded_ansible_services.py
+++ b/cfme/tests/ansible/test_embedded_ansible_services.py
@@ -7,12 +7,14 @@ from widgetastic_patternfly import BootstrapSelect
 
 from cfme import test_requirements
 from cfme.cloud.provider.ec2 import EC2Provider
+from cfme.control.explorer.policies import VMControlPolicy
 from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.markers.env_markers.provider import ONE_PER_TYPE
 from cfme.services.myservice import MyService
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.blockers import BZ
+from cfme.utils.generators import random_vm_name
 from cfme.utils.log import logger
 from cfme.utils.log_validator import LogValidator
 from cfme.utils.update import update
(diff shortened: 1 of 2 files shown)


cfme/tests/ansible/test_embedded_ansible_services.py
```diff
@@ -39,6 +42,69 @@
 ]
 
 
+@pytest.fixture()
+def ansible_linked_vm_action(
+    appliance, ansible_catalog_item, new_vm
+):
```
digitronik: I think we are not crossing 100 char limit

cfme/tests/ansible/test_embedded_ansible_services.py
```diff
@@ -39,6 +42,69 @@
...
+    with update(ansible_catalog_item):
+        ansible_catalog_item.provisioning = {"playbook": "add_single_vm_to_service.yml"}
+
+    action_collection = appliance.collections.actions
```
digitronik: optional:
please use it directly in create.

cfme/tests/ansible/test_embedded_ansible_services.py
```diff
@@ -39,6 +42,69 @@
...
+        fauxfactory.gen_alpha(), policies=[policy]
+    )
+    new_vm.assign_policy_profiles(policy_profile.description)
+    yield
+
+    if policy.exists:
+        policy.unassign_events("Tag Complete")
+        new_vm.unassign_policy_profiles(policy_profile.description)
+        policy_profile.delete()
+        policy.delete()
```
digitronik: optional:
can we try `delete_if_exists()` for both

cfme/tests/ansible/test_embedded_ansible_services.py
sbulage: Updated.

cfme/tests/ansible/test_embedded_ansible_services.py
sbulage: Updated as per directed.

cfme/tests/ansible/test_embedded_ansible_services.py
sbulage: @digitronik as we are removing assigned event couldn't able to use `delete_if_exists` directly there.

sbulage: Oops my bad I didn't see that included `BaseEntity` there, thanks @digitronik for pointing out.

jawatts: I still think this applies, we should check that the policy exists before deleting

cfme/tests/ansible/test_embedded_ansible_services.py
```diff
@@ -634,3 +696,34 @@ def _revert():
...
+        1448918
+
+    Polarion:
+        assignee: sbulage
+        casecomponent: Ansible
+        initialEstimate: 1/3h
+        tags: ansible_embed
+    """
+    added_tag = new_vm.add_tag()
+    request.addfinalizer(lambda: new_vm.remove_tag(added_tag))
```
mshriver: Since the VM is function scoped and is cleaned up after the test, is it worthwhile to spend the time removing the tag?

cfme/tests/ansible/test_embedded_ansible_services.py
```diff
@@ -634,3 +696,34 @@ def _revert():
 
     view = navigate_to(ansible_service, "Details")
     assert view.provisioning.results.get_text_of("Status") == "successful"
+
+
+@pytest.mark.tier(3)
+@pytest.mark.provider([VMwareProvider], override=True)
+@pytest.mark.usefixtures("setup_provider")
+def test_ansible_service_linked_vm(
```
jawatts: please add the automate marker for this BZ

cfme/tests/ansible/test_embedded_ansible_services.py
sbulage: Already updated with `policy_profile.delete_if_exists()`

cfme/tests/ansible/test_embedded_ansible_services.py
sbulage: Updating it.

cfme/tests/ansible/test_embedded_ansible_services.py
jawatts: @sbulage what about the policy though?

sbulage: @jawatts yes for Policy `delete_if_exists()` is available but here we are checking if policy is there then first unassign event and later deleting it.

cfme/tests/ansible/test_embedded_ansible_services.py
sbulage: My thought is whatever is added/created should be removed in `finalizer`, if you suggests to remove `remove_tag()` I will remove it.

sbulage: Removed Finalizer for `remove_tag()`

dajoRH:
*CFME QE Bot*

cfme/tests/ansible/test_embedded_ansible_services.py
sbulage: @john-dupuy @mshriver adding `policy.delete_if_exists()` is breaking tests at teardown. Adding back above lines to have test case working.

sbulage: Never Mind. Solved locator issue after debugging.

sbulage:
@jawatts @mshriver Updated PR with all suggestions.

PR #8992, opened 2019-06-25
[1LP][RFR] Update OpenstackProvider.view_value_mapping, v3 fields

mshriver:
The v3 keystone field is not displayed when keystone version v2 is selected.

Update OpenstackProvider.view_value_mapping to only include v3 attributes when using v3 endpoint.


mshriver:
Thanks for quick review guys. Looks like our rhos13 v2 endpoint still may have issues, but this resolves the framework behavior that blocks us from validating the endpoint.

PR #8993, opened 2019-06-25
[1LP][RFR] Block alert_profile_crud test due to BZ

john-dupuy:
This test is breaking a lot of the basic control tests because it gets stuck on the edit alert profile page due to https://bugzilla.redhat.com/show_bug.cgi?id=1723815

{{ pytest: --long-running --use-template-cache cfme/tests/control/test_basic.py }}


john-dupuy:
A few 5.11 tests are failing due a small string change that will be updated in a future PR. 

Tests are failing in 5.10, due to an event that was removed which is causing the browser to get stuck on the Assign events to policy page. The non-existent event has been removed and I will open a test case for this so it doesn't happen again.

PR #8994, opened 2019-06-25
[1LP][RFR]Fix test: test_contents

valaparthvi:
Changes:
1. Fix test `test_contents`.

{{ pytest: cfme/tests/configure/test_docs.py::test_contents --use-template-cache -sqvvv }}

### DIFF
--- a/cfme/tests/configure/test_docs.py
+++ b/cfme/tests/configure/test_docs.py
@@ -29,7 +29,7 @@ doc_titles = {
     'providers': 'managing providers',
     'rest': 'red hat cloudforms rest api',
     'scripting': 'scripting actions in cloudforms',
-    'vm_hosts': 'provisioning virtual machines and instances'}
+    'vm_instances': 'provisioning virtual machines and instances'}
 
 
 def pdf_get_text(file_obj, page_nums):
--- a/cfme/tests/configure/test_docs.py
+++ b/cfme/tests/configure/test_docs.py
@@ -12,6 +12,7 @@ from pdfminer.pdfpage import PDFPage
 from cfme import test_requirements
 from cfme.utils import version
 from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.blockers import BZ
 from cfme.utils.log import logger
 
 
@@ -28,7 +29,7 @@ doc_titles = {
     'providers': 'managing providers',
     'rest': 'red hat cloudforms rest api',
     'scripting': 'scripting actions in cloudforms',
-    'vm_hosts': 'provisioning virtual machines and hosts'}
+    'vm_hosts': 'provisioning virtual machines and instances'}
 
 
 def pdf_get_text(file_obj, page_nums):
@@ -108,7 +109,8 @@ def test_contents(appliance, soft_assert):
         else:
             expected.append('cloudforms')
             assert cur_ver.product_version() is not None
-            expected.append(cur_ver.product_version())
+            if not BZ(1723813).blocks:
+                expected.append(cur_ver.product_version())
 
         for exp_str in expected:
             soft_assert(exp_str in pdf_titlepage_text_low, "{} not in {}"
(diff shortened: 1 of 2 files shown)


widgetastic_manageiq/__init__.py
```diff
@@ -5749,3 +5749,15 @@ def fill(self, value):
...
     def is_displayed(self):
         return self.textbox.is_displayed
+
+
+class HelpNavDropdown(NavDropdown):
+    @property
+    def expandable(self):
+        """Returns true with a menu can be expanded by checking the value of data-toggle"""
+        try:
+            self.browser.element('./a[contains(@data-toggle, "dropdown")]', parent=self)
```
digitronik: optional: no need to add a parent as self it will self resolve by wt.

cfme/tests/configure/test_docs.py
```diff
@@ -108,7 +109,8 @@ def test_contents(appliance, soft_assert):
         else:
             expected.append('cloudforms')
             assert cur_ver.product_version() is not None
-            expected.append(cur_ver.product_version())
+            if not BZ(1723813, forced_streams=["5.11"]).blocks:
```
digitronik: Q: do you think we have to block this test with BZ for 511 than this workaround.

digitronik:
@mshriver @izapolsk can we add this expandable change in `wt.pfy` as optional to the current locator?

cfme/tests/configure/test_docs.py
mshriver: Agreed, I think this test should be blocked in its entirety, not just this conditional statement.

Additionally, `forced_streams` is not necessary here, as the BZ blocks against 5.11 without it.

```
BZ
In [2]: BZ(1723813).blocks
Out[2]: True

In [3]: from cfme.fixtures.pytest_store import store

In [4]: store.current_appliance.version
Out[4]: Version('5.11.0.10')
```

widgetastic_manageiq/__init__.py
mshriver: Correct, but not a bad thing to state explicitly. The `NavDropdown.expandable` method implementation does the same.

Overall - I would much rather see this done via a locator constant than by re-implementing the method with the exact same pattern and a different locator.

@valaparthvi Can you please open a wt.patternfly PR, updating `NavDropdown` to define and use a class attribute for this locator?  Then this class can simply define its `EXPANDABLE_LOC` in order to modify the base class.

Will look like:
```python
# in wt.patternfly:
class NavDropdown(Widget, ClickableMixin):
    EXPANDABLE_LOC = './a/span[contains(@class, "caret")]'
    # just highlighting what would change here
    @property
    def expandable(self):
        try: self.browser.element(self.EXPANDABLE_LOC, parent=self)

# in wt.manageiq
class HelpNavDropdown(NavDropdown):
    EXPANDABLE_LOC = './a[contains(@data-toggle, "dropdown")]'
```

cfme/tests/configure/test_docs.py
valaparthvi: Every other assertion is working fine for 5.11, except that the docs on 5.11 do not have expected product version.

valaparthvi:
@mshriver @digitronik 
1. I think it is better to skip a certain assertion rather than skipping the whole test since all other assertions are working fine.
2. Changes to widgetastic_manageiq were introduced to fix the test failure, so I'd rather keep them together.
3. Raised a [PR](https://github.com/RedHatQE/widgetastic.patternfly/pull/97) to wt.pf to enhance `NavDropdown`.
4. Tested these changes locally and they work fine. Since the changes haven't been merged into wt.pf yet, this PRT will fail.

dajoRH:
*CFME QE Bot*

cfme/tests/configure/test_docs.py
```diff
@@ -28,7 +29,7 @@
     'providers': 'managing providers',
     'rest': 'red hat cloudforms rest api',
     'scripting': 'scripting actions in cloudforms',
-    'vm_hosts': 'provisioning virtual machines and hosts'}
+    'vm_hosts': 'provisioning virtual machines and instances'}
```
digitronik: @valaparthvi `key` indicating `vm` and `hosts` ^^

valaparthvi:
@jawatts Idk why, but sometimes PRT fails for 5.10 and sometimes for 5.11. Doesn't pass for both :/

PR #8917, opened 2019-06-06
[RFR] Added VDDK 6.7 checks and updated getter method

sbulage:
PRT Run:

{{ pytest: cfme/tests/infrastructure/test_ssa_vddk.py -svvvv --long-running }}

PR #8997, opened 2019-06-25
[RFR] Add a UI workaround for setting up host creds in 5.11

john-dupuy:
Due to https://bugzilla.redhat.com/show_bug.cgi?id=1718209, we currently cannot setup host credentials in CFME 5.11 via REST. This is blocking a large number of tests that require host credentials for SSA, so as a workaround, we set them via the UI.

{{ pytest: --use-provider vsphere6-nested cfme/tests/control/test_compliance.py::test_check_files }}

@sbulage can you use this function regarding the changes you made in #8917? That way when the BZ is no longer blocking, it will again use REST.


sbulage:
@john-dupuy I am working on SSA RHEL VM fix, I will update it with those changes.

PR #8999, opened 2019-06-25
[1LP][RFR] Add a test case to check default policy events

john-dupuy:
In testing PRT on #8993, I noticed that some tests failed because the `EVENTS` defined in cfme.tests.control.test_basic.py were out of date. Here I've updated them and added a test case that will fail if they change. 

{{ pytest: --use-template-cache --long-running cfme/tests/control/test_basic.py::test_default_policy_events }}


cfme/tests/control/test_basic.py
```diff
@@ -640,6 +671,9 @@ def test_alert_profile_crud(request, appliance, alert_profile_class):
         casecomponent: Control
         caseimportance: critical
         initialEstimate: 1/12h
+
+    Bugzilla:
+        1723815
```
sshveta: Looks good , please also add 
@pytest.mark.meta(automates=[1723815]) to the test

john-dupuy: @sshveta it is there already :smiley:

sshveta: :+1:  Yeah just saw :)

izapolsk:
I guess we have to make up something better to verify that tests are based on correct data. but let's merge this for now.

PR #9004, opened 2019-06-26
[1LP][RFR] Fixed vsphere skiping issue

ganeshhubale:
- This PR fixes the issue of skipped test cases related to vsphere provider

### DIFF
--- a/cfme/tests/infrastructure/test_child_tenant.py
+++ b/cfme/tests/infrastructure/test_child_tenant.py
@@ -14,9 +14,10 @@ from cfme.utils.generators import random_vm_name
 pytestmark = [
     test_requirements.quota,
     pytest.mark.meta(server_roles="+automate"),
-    pytest.mark.usefixtures('setup_provider', 'uses_infra_providers'),
+    pytest.mark.usefixtures('setup_provider'),
     pytest.mark.long_running,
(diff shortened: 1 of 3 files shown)


ganeshhubale:
One test case failed on 5.11 which is not related the changes made in this PR. Hence moved to RFR.

cfme/tests/automate/test_git_import.py
```diff
@@ -45,7 +45,8 @@ def new_user(appliance):
         value_assign="Database",
     )
     yield user
-    user.delete_if_exists()
+    user = appliance.rest_api.collections.users.get(name=user.name)
+    user.action.delete()
```
digitronik: @ganeshhubale check exists with rest and then delete... but the reference of BZ as well

ganeshhubale: Actually I am keeping these changes as it was (```user.delete_if_exists()```) [this issue](https://bugzilla.redhat.com/show_bug.cgi?id=1720273) got fixed in 5.11.0.11

ganeshhubale:
> One test case failed on 5.11 which is not related the changes made in this PR. Hence moved to RFR.

This test case is passing locally.

mshriver:
PRT behavior does not appear related to the collection change here.

PR #9005, opened 2019-06-26
[RFR][NOMERGE] moving tests and wrapanapi to py3winrm

izapolsk:
since pywinrm used in wrapanapi for interacting with scvmm doesn't accept pull requests and doesn't fix issues. we are forced to clone it and fix existing py3 issues.

This PR intends to test our clone of pywinrm and replace old package with our one.


mshriver:
@izapolsk @jawatts has opened #9021 to include these updates and changes to `list_hosts` method calls, which was an additional PR included in wrapanapi 3.3.0.

PR #9006, opened 2019-06-26
[1LP][RFR] Adding new customer test case for ems switches bz 1688900 and  vcenter rebuilt bz 1719399

kedark3:
Purpose or Intent
=================

__Adding__ Adding new Test Case

 {{pytest: cfme/tests/infrastructure/test_vmware_provider.py -k "test_rebuilt_vcenter_duplicate_hosts or test_switches_class_present_ems"}}

### DIFF
--- a/cfme/tests/infrastructure/test_vmware_provider.py
+++ b/cfme/tests/infrastructure/test_vmware_provider.py
@@ -417,3 +417,47 @@ def test_esxi_reboot_not_orphan_vms(appliance, provider):
         power_state = view.entities.summary('Power Management').get_text_of('Power State')
         assert not power_state == 'orphaned'
         assert not power_state == 'archived'
+
+
+# Add test for 1719399
+
+@pytest.mark.tier(1)
+@pytest.mark.provider([VMwareProvider], selector=ONE_PER_TYPE, override=True)
+@pytest.mark.meta(automates=[1688900])
+def test_switches_class_present_ems(appliance, provider):
+    """
+    Under this customer requested enhancement, ems should have switches class.
+
+    Bugzilla:
+        1688900
+
+    Polarion:
+        assignee: kkulkarn
+        casecomponent: Infra
+        caseimportance: critical
+        initialEstimate: 1/2h
+        testtype: integration
+        testSteps:
+            1.Add VMware provider to CFME
+            2.SSH to CFME appliance and perform following steps in rails console
+                '''
+                $evm = MiqAeMethodService::MiqAeService.new(MiqAeEngine::MiqAeWorkspaceRuntime.new)
+                p = $evm.vmdb(:ems_infra).first
+                p.class.name
+                p.switches.first.class.name
+                '''
+        expectedResults:
+            1.Provider added successfully and is refreshed
+            2.p.switches.first.class.name returns exit status 0(success) and lists class name
+            containing HostVirtualSwitch
+    """
+    command = """'$evm = MiqAeMethodService::MiqAeService.new(MiqAeEngine::MiqAeWorkspaceRuntime.new);\
+                p = $evm.vmdb(:ems_infra).first;\
+                p.class.name;\
+                puts "class name [#{p.switches.first.class.name}]"'
+                """
+    result = appliance.ssh_client.run_rails_command(command)
+    logger.info("Output of Rails command was {}".format(result.output))
+    assert result.success, "SSH Command result was unsuccessful: {}".format(result)
+    if not result.output:
+        assert 'HostVirtualSwitch' in result.output
--- a/cfme/tests/infrastructure/test_vmware_provider.py
+++ b/cfme/tests/infrastructure/test_vmware_provider.py
@@ -407,10 +407,10 @@ def test_esxi_reboot_not_orphan_vms(appliance, provider):
                 vm.update_attributes(:uid_ems => SecureRandom.uuid);\
                 puts \"VM_ID=#{vm.id} name=[#{vm.name}] uid=#{vm.uid_ems}\"'"
     result = appliance.ssh_client.run_rails_command(command)
-    logger.info("Output of Rails command was {}".format(result.output))
     provider.refresh_provider_relationships()
     assert result.success, "SSH Command result was unsuccessful: {}".format(result)
     if not result.output:
+        logger.info("Output of Rails command was {}".format(result.output))
         vm_name = re.findall(r"\[.+\]", result.output)[0].split('[')[1].split(']')[0]
         vm = appliance.collections.infra_vms.instantiate(name=vm_name, provider=provider)
         view = vm.load_details(from_any_provider=True)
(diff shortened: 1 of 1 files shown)


cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -385,3 +386,47 @@ def test_esxi_reboot_not_orphan_vms(appliance, provider):
...
+    """
+    command = """'$evm = MiqAeMethodService::MiqAeService.new(MiqAeEngine::MiqAeWorkspaceRuntime.new);\
+                p = $evm.vmdb(:ems_infra).first;\
+                p.class.name;\
+                puts "class name [#{p.switches.first.class.name}]"'
+                """
+    result = appliance.ssh_client.run_rails_command(command)
+    logger.info("Output of Rails command was {}".format(result.output))
+    assert result.success, "SSH Command result was unsuccessful: {}".format(result)
+    if not result.output:
```
sshveta: I think this should be 
          
         if result.success:
             assert 'HostVirtualSwitch' in result.output

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -375,13 +376,99 @@ def test_esxi_reboot_not_orphan_vms(appliance, provider):
...
+        logger.info("Output of Rails command was {}".format(result.output))
         vm_name = re.findall(r"\[.+\]", result.output)[0].split('[')[1].split(']')[0]
         vm = appliance.collections.infra_vms.instantiate(name=vm_name, provider=provider)
         view = vm.load_details(from_any_provider=True)
         power_state = view.entities.summary('Power Management').get_text_of('Power State')
         assert not power_state == 'orphaned'
         assert not power_state == 'archived'
+
+
+# Add test for 1719399
```
sshveta: This can be removed now since you have added test

cfme/tests/infrastructure/test_vmware_provider.py
kedark3: resolved it slightly differently. Asserting output to be `non-empty` and if its empty, test would fail. Because ultimately I need output to be `non-empty` for final assertion to pass.

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -375,13 +376,97 @@ def test_esxi_reboot_not_orphan_vms(appliance, provider):
...
+            containing HostVirtualSwitch
+    """
+    command = """'$evm = MiqAeMethodService::MiqAeService.new(MiqAeEngine::MiqAeWorkspaceRuntime.new);\
+                p = $evm.vmdb(:ems_infra).first;\
+                p.class.name;\
+                puts "class name [#{p.switches.first.class.name}]"'
+                """
+    result = appliance.ssh_client.run_rails_command(command)
+    assert result.success, "SSH Command result was unsuccessful: {}".format(result)
+    logger.info("Output of Rails command was {}".format(result.output))
```
digitronik: optional:
assert will inform you... actually `result.__str__()` represent actual output only. but ok logger info

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -375,13 +376,97 @@ def test_esxi_reboot_not_orphan_vms(appliance, provider):
...
+    assert result.output, "Output is blank for ssh command"
```
digitronik: optional: No need of this asset as we are asserting `.success` and `HostVirtualSwitch` in `.output`

sshveta:
Please fix PRT failure for 5.11

kedark3:
@sshveta I think the class name in 511 is different. Need to check with developer. Please wait on merge until I hear from him. Thanks.

dajoRH:
*CFME QE Bot*

kedark3:
> @sshveta I think the class name in 511 is different. Need to check with developer. Please wait on merge until I hear from him. Thanks.

Developer said as soon as command is successful and returns some subclass of `Switch` that's fine.

cfme/tests/infrastructure/test_vmware_provider.py
jawatts: Agree here, the assert below is taking care of this

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -407,13 +407,96 @@ def test_esxi_reboot_not_orphan_vms(appliance, provider):
...
+    """
+    command = """'$evm = MiqAeMethodService::MiqAeService.new(MiqAeEngine::MiqAeWorkspaceRuntime.new);\
+                p = $evm.vmdb(:ems_infra).first;\
+                p.class.name;\
+                puts "class name [#{p.switches.first.class.name}]"'
+                """
+    result = appliance.ssh_client.run_rails_command(command)
+    assert result.success, "SSH Command result was unsuccessful: {}".format(result)
+    logger.info("Output of Rails command was {}".format(result.output))
+    assert 'switch' in result.output.lower()
```
jawatts: Between 5.10 and 5.11, different switch classes are returned. I think it be a better check to `VersionPick` here. Thoughts?

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -407,13 +407,96 @@ def test_esxi_reboot_not_orphan_vms(appliance, provider):
...
+    """
+    hosts_before = len(provider.hosts.all())
+    command = "'Host.all.each { |h| h.ems_id = nil; h.ems_ref = h.id.to_s; h.save! }'"
+    result = appliance.ssh_client.run_rails_command(command)
+    assert result.success, "SSH Command result was unsuccessful: {}".format(result)
+    logger.info("Output of Rails command was {}".format(result.output))
+    provider.refresh_provider_relationships()
+    # Using appliance.rest_api as hosts.all() do not return archived hosts, I need those too
+    hosts_after = len(appliance.rest_api.collections.hosts.all)
+    assert hosts_before == hosts_after
```
jawatts: I think we are assuming that there are no archived hosts at the start of this test. If there is an archived host, this will fail.

kedark3: Discussed offline, updated the `hosts_before` call

cfme/tests/infrastructure/test_vmware_provider.py
mshriver: When logging, please use the argument syntax, and don't directly format the string.
`logger.info('output of rails command: %s', result.output)`

cfme/tests/infrastructure/test_vmware_provider.py
mshriver: Agreed @jawatts, this assertion should be more specific. If we have a test specific to the EVM switch classes, it should be tightly asserting the proper switch class.

kedark3: Discussed with Justin and he asked to add comments. Added those above ^^

jawatts: ^ Unfortunately this is all we have here.

PR #9012, opened 2019-06-27
[1LP][RFR] Update Taggable.get_tags

mshriver:
Don't set the Tag.name attribute to the Category.display_name
Update test_vmware_vds_ui_tagging
Update add_tag to check for tag existing already

Marking RFR as PRT run 32832 passed `test_vmware_vds_ui_tagging`.  Including the test_tag_objects module as well for a broader run.

{{ pytest: cfme/tests/infrastructure/test_vmware_provider.py::test_vmware_vds_ui_tagging cfme/tests/cloud_infra_common/test_tag_objects.py cfme/tests/infrastructure/test_config_management.py::test_config_system_tag cfme/tests/infrastructure/test_config_management.py::test_ansible_tower_job_templates_tag cfme/tests/cloud_infra_common/test_provisioning.py::test_provision_with_tag cfme/tests/automate/custom_button/test_generic_objects.py cfme/tests/automate/custom_button/test_service_objects.py::test_custom_button_expression_service_obj cfme/tests/ansible/test_embedded_ansible_tags.py cfme/tests/cloud/test_keypairs.py::test_keypair_add_and_remove_tag --use-provider vsphere67-nested --use-provider ec2 -vvvv }}

### DIFF
--- a/cfme/common/__init__.py
+++ b/cfme/common/__init__.py
@@ -409,7 +409,8 @@ class Taggable(TaggableCommonBase):
     standardized widgetastic views
     """
 
-    def add_tag(self, tag=None, cancel=False, reset=False, details=True, dashboard=False):
+    def add_tag(self, tag=None, cancel=False, reset=False, details=True, dashboard=False,
+                exists_check=False):
         """ Add tag to tested item
 
         Args:
@@ -419,8 +420,9 @@ class Taggable(TaggableCommonBase):
             details (bool): set False if tag should be added for list selection,
                             default is details page
             dashboard (bool): Set to True if tag should be added via the Dashboard view
+            exists_check (bool): Set to True to check if tag exists before trying to add
         """
-        if tag and tag in self.get_tags():
+        if exists_check and tag and tag in self.get_tags():
             logger.warning('Trying to add tag [%s] already assigned to entity [%s]', tag, self)
             return tag
         if details and not dashboard:
--- a/cfme/common/__init__.py
+++ b/cfme/common/__init__.py
@@ -420,6 +420,9 @@ class Taggable(TaggableCommonBase):
                             default is details page
             dashboard (bool): Set to True if tag should be added via the Dashboard view
         """
+        if tag and tag in self.get_tags():
+            logger.warning('Trying to add tag [%s] already assigned to entity [%s]', tag, self)
+            return tag
         if details and not dashboard:
             step = 'EditTagsFromDetails'
         elif dashboard:
@@ -488,7 +491,7 @@ class Taggable(TaggableCommonBase):
             raise ItemNotFound('Details page does not exist for: {}'.format(self))
         except TimedOutError:
             raise ItemNotFound('Timed out navigating to details for: {}'.format(self))
-        tags = []
+        tags_objs = []
         entities = view.entities
         if hasattr(entities, 'smart_management'):
             tag_table = entities.smart_management
@@ -499,12 +502,12 @@ class Taggable(TaggableCommonBase):
             # check for users/groups page in case one tag string is returned
             for tag in [tags_text] if isinstance(tags_text, six.string_types) else list(tags_text):
                 tag_category, tag_name = tag.split(':')
-                category = self.appliance.collections.categories.instantiate(
-                    display_name=tag_category)
-                tags.append(category.collections.tags.instantiate(
-                    display_name=tag_name.strip(),
-                    name=tag_category.strip()))
-        return tags
+                # instantiate category first, then use its tags collection to instantiate tag
+                tags_objs.append(
+                    self.appliance.collections.categories.instantiate(display_name=tag_category)
+                    .collections.tags.instantiate(display_name=tag_name.strip())
+                )
+        return tags_objs
 
 
 class TaggableCollection(TaggableCommonBase):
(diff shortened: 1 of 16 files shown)


cfme/configure/configuration/region_settings.py
```diff
@@ -54,7 +54,7 @@ def is_displayed(self):
         )
 
 
-@attr.s
+@attr.s(cmp=False)
```
digitronik: :100:

cfme/common/__init__.py
```diff
@@ -420,6 +420,9 @@ def add_tag(self, tag=None, cancel=False, reset=False, details=True, dashboard=F
                             default is details page
             dashboard (bool): Set to True if tag should be added via the Dashboard view
         """
+        if tag and tag in self.get_tags():
+            logger.warning('Trying to add tag [%s] already assigned to entity [%s]', tag, self)
```
lina-nikiforova: why not `format` here?

digitronik: @lina-nikiforova for logger better to use normal format. If you check for conf you will get the complate idea. 

I understand good to use `.format` or `fstring` :+1:

lina-nikiforova: I have a concern here but I haven't tested. I doubt it will return True because I remember examples in our framework when `tag in get_tags()` doesn't work.

For example,
```
    tags = job_template.get_tags()
    msg = "Failed to setup a configuration system's tag"
    tag_info = '{}: {}'.format(tag.category.display_name, tag.display_name)
    assert tag_info in ['{}: {}'.format(t.category.display_name, t.display_name) for t in tags], msg
```

lina-nikiforova: @digitronik thanks, it's just a question :) because I briefly checked other logging messages we have in the codebase and I saw that format is used

lina-nikiforova: nevermind, I see you updated that method as well :)

cfme/common/__init__.py
```diff
@@ -499,12 +502,12 @@ def get_tags(self, tenant="My Company Tags"):
...
                 tag_category, tag_name = tag.split(':')
-                category = self.appliance.collections.categories.instantiate(
-                    display_name=tag_category)
-                tags.append(category.collections.tags.instantiate(
-                    display_name=tag_name.strip(),
-                    name=tag_category.strip()))
-        return tags
+                # instantiate category first, then use its tags collection to instantiate tag
+                tags_objs.append(
+                    self.appliance.collections.categories.instantiate(display_name=tag_category)
```
kedark3: Category is description right?

cfme/common/__init__.py
mshriver: @lina-nikiforova ones that you found that were formatting directly should be updated - thanks for asking/looking!

mshriver: yep!

This should now work with the update to `get_tags` and the `Tag.__eq__` method, and I did some cursory testing of `in`, `not in`, `==` comparisons of tag objects.

mshriver: I'll update `test_config_management.py` - I searched for references to `tag.name` that were expecting that to be set to the category display_name, but didn't locate that one.

mshriver: https://github.com/ManageIQ/integration_tests/pull/9012/commits/4bc2d023e0d7936ac327677f7df843dbe218678e

cfme/common/__init__.py
mshriver: Can you elaborate? The `TagCollection` has a `Category` as its parent, and the `Tag` has `name` and `display_name` attributes.

The attributes aren't named very well, but I'm not trying to change them in this scope.  MIQ uses `name` and `description`, where `name` is more akin to a key, and `description` is the displayed tag value.

kedark3: Never mind, this was in reference to our conversation offline about categories/name/description.

cfme/configure/configuration/region_settings.py
```diff
@@ -69,6 +69,22 @@ class Tag(Pretty, BaseEntity, Updateable):
     display_name = attr.ib()
     name = attr.ib(default=None)
 
+    def __eq__(self, other):
```
kedark3: :+1:

cfme/common/__init__.py
mshriver: Yeah there's a bit of a mismatch between the Tag/Category class attributes and what those fields are called in CFME. That should be pretty straightforward to change.

What's more complex is fixing the `get_tags()` method to resolve the class attributes for those tags that aren't displayed (name) when viewing tags on an entity.  Not for this PR :)

lina-nikiforova:
LGTM! Thank you for this PR, Mike!

cfme/common/__init__.py
izapolsk: I have a doubt that framework should silently return tag in this case. I'd better let it raise exception instead.
such change won't allow us test duplicate records test case for instance.

second thing to consider,  get_tags call may slow down this operation since it is UI based operation.

izapolsk: I found an answer what it is for. so, your changes look more clear.
however, I'd rather you either created another method f.e. `add_tag_unless_exist` or added some option to bypass `exists` check or added some decorator to handle `exists` case.
first option is more preferable since we already use it in similar cases like delete_if_exists

mshriver: Added a kwarg for `exists_check` and defaulted it to `False`. Passing `True` in some of the fixture uses.

PR #9002, opened 2019-06-26
[1LP][RFR] Fixed playbook method crud operations

ganeshhubale:
- This PR fixes crud operation of ```automate method type - playbook```
- Also changes in the naming of the automate method types for playbook and inline are resolved in this PR
- This PR updates test case with global fixture 
## PRT RUN
{{ pytest: cfme/tests/ansible/test_embedded_ansible_automate.py -k "test_automate_ansible_playbook_method_type_crud or test_automate_ansible_playbook_method_type" --long-running  -vvvvv }}

PR #9013, opened 2019-06-27
[1LP][WIPTEST] Small fix for automate method create function

john-dupuy:
Fixing a bug in the `create` method. 

{{ pytest: --long-running --use-provider vsphere65-nested cfme/tests/ansible/test_embedded_ansible_automate.py::test_alert_run_ansible_playbook }}


john-dupuy:
@ganeshhubale is working on a similar fix in #9002, I think his fix is better but I am leaving this open to see why this test case is failing on PRT.

john-dupuy:
Closing due to #9002  fix

PR #9016, opened 2019-06-27
[1LP][RFR] Refactored test_csv_imports.py

Yadnyawalkya:
{{pytest: cfme/tests/v2v/test_csv_import.py --use-provider={rhv43,osp13-ims} --use-provider vsphere67-ims --provider-limit 3 -vvvv}}

- Fixed non-csv test failure
- Added security group and flavor tests
- Fixed tests to accomodate OSP
- Fixed IndexError from network component to run OSP tests

### DIFF
--- a/cfme/tests/v2v/test_csv_import.py
+++ b/cfme/tests/v2v/test_csv_import.py
@@ -100,8 +100,8 @@ def check_vm_status(appliance, infra_map, error_text=None, filetype='csv', conte
         plan_view.instance_properties.table.wait_displayed()
         table_data = plan_view.instance_properties.table.read()[0]
         error_msg = {
-            "security_group": table_data["OpenStack Security Group"],
-            "flavor": table_data["OpenStack Flavor"]}
+            "security_group": str(table_data["OpenStack Security Group"]),
+            "flavor": str(table_data["OpenStack Flavor"])}
     plan_view.cancel_btn.click()
     return error_msg
 
--- a/cfme/tests/v2v/test_csv_import.py
+++ b/cfme/tests/v2v/test_csv_import.py
@@ -2,9 +2,11 @@ import tempfile
 
 import fauxfactory
 import pytest
+from widgetastic.exceptions import NoSuchElementException
 from widgetastic.exceptions import UnexpectedAlertPresentException
 
 from cfme import test_requirements
+from cfme.cloud.provider.openstack import OpenStackProvider
 from cfme.fixtures.v2v_fixtures import infra_mapping_default_data
 from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
@@ -16,7 +18,10 @@ from cfme.utils.wait import wait_for
 pytestmark = [
     test_requirements.v2v,
     pytest.mark.provider(
-        classes=[RHEVMProvider], selector=ONE_PER_VERSION, required_flags=["v2v"], scope="module"
+        classes=[RHEVMProvider, OpenStackProvider],
+        selector=ONE_PER_VERSION,
+        required_flags=["v2v"],
+        scope="module",
     ),
     pytest.mark.provider(
         classes=[VMwareProvider],
@@ -57,8 +62,8 @@ def migration_plan(appliance, infra_map, csv=True):
     return view
 
 
-def import_and_check(appliance, infra_map, error_text=None,
-                     filetype='csv', content=False, table_hover=False, alert=False):
+def import_and_check(appliance, infra_map, error_text=None, filetype='csv', content=False,
+                     table_hover=False, alert=False, security_group=False):
     plan_view = migration_plan(appliance, infra_map)
     temp_file = tempfile.NamedTemporaryFile(suffix='.{}'.format(filetype))
     if content:
@@ -76,13 +81,24 @@ def import_and_check(appliance, infra_map, error_text=None,
         else:
             # click on the checkbox to select VM (column 0)
             plan_view.vms.table[0][1].widget.click()
-        error_msg = plan_view.vms.popover_text.read()
+        if not security_group:
+            error_msg = plan_view.vms.popover_text.read()
     else:
         if alert:
             error_msg = plan_view.browser.get_alert().text
-            plan_view.browser.handle_alert()
+            try:
+                plan_view.browser.handle_alert()
+            except NoSuchElementException:
+                pass
         else:
             error_msg = plan_view.vms.error_text.text
+    if security_group:
+        plan_view.next_btn.click()
+        plan_view.instance_properties.table.wait_displayed()
+        table_data = plan_view.instance_properties.table.read()[0]
+        error_msg = {
+            "security_group": table_data["OpenStack Security Group"],
+            "flavor": table_data["OpenStack Flavor"]}
     plan_view.cancel_btn.click()
     return bool(error_msg == error_text)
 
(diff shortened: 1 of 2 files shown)


Yadnyawalkya:
5.11 failures are due to UI changes, all tests working for 5.10 against rhv43 and osp13-ims.

cfme/tests/v2v/test_csv_import.py
```diff
@@ -269,17 +253,36 @@ def test_csv_archived_vm(appliance, infra_map, archived_vm):
...
+        initialEstimate: 1/4h
+    """
+    security_group = provider.data.security_groups.admin[1]
+    flavor = provider.data.flavors[1]
+    content = "Name,Security Group,Flavor\n{valid_vm},{security_group},{flavor}".format(
+        valid_vm=valid_vm,
+        security_group=security_group,
+        flavor=flavor,
+    )
+    error_msg = {"security_group": security_group, "flavor": flavor}
```
sshveta: Question : Is this a positive test case ? on giving csv with security group and flavor do we get error message on UI ?

Yadnyawalkya: Yes this is positive testcase and we are not checking error in UI. I used existing `error_msg` and `error_text` variables just to re-use `import_and_check` function. 

Basically we are checking whatever security group and flavor we provide via CSV import should be there on instance properties section of infra map.

Yadnyawalkya: Yes this is positive testcase and we are not checking error in UI. I used existing `error_msg` and `error_text` variables just to re-use `import_and_check` function. 

Basically we are checking whatever security group and flavor we provide via CSV import (L281) should be there on instance properties section of infra map.

sshveta:
we need to fix it for 5.11 in next PR

cfme/tests/v2v/test_csv_import.py
```diff
@@ -269,17 +253,36 @@ def test_csv_archived_vm(appliance, infra_map, archived_vm):
...
+    """Test csv with openstack security group and flavor
+    Polarion:
+        assignee: ytale
+        caseposneg: positive
+        startsin: 5.10
+        casecomponent: V2V
+        customerscenario: true
+        initialEstimate: 1/4h
+    """
+    security_group = provider.data.security_groups.admin[1]
```
mshriver: What is this index for? What structure are you expecting in the provider data? Is this supposed to be a string or a map of values?

cfme/tests/v2v/test_csv_import.py
```diff
@@ -269,17 +253,36 @@ def test_csv_archived_vm(appliance, infra_map, archived_vm):
...
+    flavor = provider.data.flavors[1]
```
mshriver: These required provider data fields should be included in the provider marker for the test function.

Otherwise we'll end up picking providers from yaml that don't include these fields, and the test will fail with KeyError/AttributeError/IndexError here.

At a minimum, wrap these statements in `try:except:pytest.skip` pattern, but ideally use `ProviderFilter.required_fields` to limit provider collection in the first place.

cfme/tests/v2v/test_csv_import.py
```diff
@@ -57,8 +62,8 @@ def migration_plan(appliance, infra_map, csv=True):
     return view
 
 
-def import_and_check(appliance, infra_map, error_text=None,
-                     filetype='csv', content=False, table_hover=False, alert=False):
+def import_and_check(appliance, infra_map, error_text=None, filetype='csv', content=False,
```
mshriver: Just a general comment, not specific to the PR.

This seems like something that should be a method on the MigrationPlan object, maybe with the 'check' side left to a test function.

cfme/tests/v2v/test_csv_import.py
```diff
@@ -76,13 +81,24 @@ def import_and_check(appliance, infra_map, error_text=None,
...
             error_msg = plan_view.vms.error_text.text
+    if security_group:
+        plan_view.next_btn.click()
+        plan_view.instance_properties.table.wait_displayed()
+        table_data = plan_view.instance_properties.table.read()[0]
+        error_msg = {
+            "security_group": table_data["OpenStack Security Group"],
+            "flavor": table_data["OpenStack Flavor"]}
     plan_view.cancel_btn.click()
     return bool(error_msg == error_text)
```
mshriver: Again, general comment not specific to this PR, but something to consider as you work on these test modules.

These variables name/use is misleading, as `error_text` is expected to actually be a dictionary with very specific keys, not `text` as the name implies. Same thing with `error_msg`.  There's no docblock for the function explaining what its actually doing (or in-line comments explaining what is happening).

Additionally the test is making direct references to table columns after having read the dictionary, without asserting/checking that those columns are present. 

Additionally this would be a much more useful function if it was actually using deepdiff to compare the dictionaries. This will result in a diff that actually describes the differences between the dicts.  The assertion being here would also help, as then the test isn't asserting the return of some function, but the actual data.

cfme/tests/v2v/test_csv_import.py
Yadnyawalkya: we have two projects `admin` and `qe-auto` in OSP and each project have primary and secondary security group. I am accessing secondary group here from `admin` project which is suppose to be string.

mshriver: This needs to be described in the test, there's nothing here that describes those requirements.

cfme/tests/v2v/test_csv_import.py
Yadnyawalkya: :D I might picked up wrong function name here, we are checking condition in test function only https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/v2v/test_csv_import.py#L244

cfme/tests/v2v/test_csv_import.py
Yadnyawalkya: There are few things..

>These variables name/use is misleading, as error_text is expected to actually be a dictionary with very specific keys, not text as the name implies. Same thing with error_msg. 

Basically only in `test_csv_security_group_flavor` argument is dictionary, in this case I was using same `import_or_check` function so I used existing variables which I accepts misleading since it is not text as you noticed but I intentionally added that since it was exceptional.

>There's no docblock for the function explaining what its actually doing (or in-line comments explaining what is happening).

Let me update it.

>Additionally this would be a much more useful function if it was actually using deepdiff to compare the dictionaries. This will result in a diff that actually describes the differences between the dicts. The assertion being here would also help, as then the test isn't asserting the return of some function, but the actual data.

Got you, since it is only with one function I like to go with what I have but I totally agree and I will make sure I will use deepdiff when I compare dictionary from next time.

cfme/tests/v2v/test_csv_import.py
Yadnyawalkya: Make sense, let me update it.

cfme/tests/v2v/test_csv_import.py
Yadnyawalkya: Updated with ``Test csv with "secondary" openstack security group and flavor``

cfme/tests/v2v/test_csv_import.py
Yadnyawalkya: Updated "check" section in test :+1:

cfme/tests/v2v/test_csv_import.py
Yadnyawalkya: Updated functions docstring and used deepdiff.

cfme/tests/v2v/test_csv_import.py
Yadnyawalkya: Updated with try:except:pytest.skip

Yadnyawalkya:
OSP13-ims is down atm, I like to test deepdiff on that as well, moving to WIPTEST

cfme/tests/v2v/test_csv_import.py
Yadnyawalkya: @mshriver 5.11 was failing because deepdiff was comparing unicode string (value after reading table) with normal string (value of expected attributes) which was not equal, so I [added](https://github.com/ManageIQ/integration_tests/pull/9016/commits/776fa4628104d9e6021fa9e1f2f71c8f9465acbc) `str` conversion around values we are getting from table.

I could have just used `==` comparison too but I am preferring deepdiff right now.

Yadnyawalkya:
Ready to go, 5.10 on passing PRT, we are fixing 5.11 failures in https://github.com/ManageIQ/integration_tests/pull/9035.

PR #9020, opened 2019-06-28
[1LP][RFR] CB attribute override test

digitronik:

Purpose or Intent
=================

{{pytest: cfme/tests/automate/custom_button/test_buttons.py::test_attribute_override -vvv}}

### DIFF
--- a/cfme/tests/automate/custom_button/test_buttons.py
+++ b/cfme/tests/automate/custom_button/test_buttons.py
@@ -16,6 +16,7 @@ from cfme.tests.automate.custom_button import OBJ_TYPE_59
 from cfme.tests.automate.custom_button import TextInputDialogView
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.blockers import BZ
+from cfme.utils.log_validator import LogValidator
 from cfme.utils.update import update
 from cfme.utils.wait import TimedOutError
 from cfme.utils.wait import wait_for
@@ -565,9 +566,10 @@ def test_custom_button_language():
     pass
 
 
-@pytest.mark.manual
 @pytest.mark.tier(2)
-def test_attribute_override():
+@pytest.mark.meta(automates=[1651099])
+@pytest.mark.provider([VMwareProvider], override=True, selector=ONE_PER_TYPE)
+def test_attribute_override(appliance, request, provider, buttongroup):
     """ Test custom button attribute override
 
     Polarion:
@@ -589,7 +591,38 @@ def test_attribute_override():
     Bugzilla:
         1651099
     """
-    pass
+    attributes = [
+        ("class", "Request"),
+        ("instance", "TestNotification"),
+        ("message", "digitronik_msg"),
+        ("namespace", "/System"),
+    ]
+    req = "call_instance_with_message"
+    executions = {"simulation": "create", "execution": "digitronik_msg"}
+    group = buttongroup("PROVIDER")
+    button = group.buttons.create(
+        text="btn_{}".format(fauxfactory.gen_alphanumeric(3)),
+        hover="hover_{}".format(fauxfactory.gen_alphanumeric(3)),
+        system="Request",
+        request=req,
+        attributes=attributes,
+    )
+    request.addfinalizer(button.delete_if_exists)
+
+    # Check msg override for simulation and actual execution
+    for exe, msg in executions.items():
+        pattern = "[miqaedb:/System/Request/TestNotification#{msg}]".format(msg=msg)
+        log = LogValidator("/var/www/miq/vmdb/log/automation.log", matched_patterns=[pattern])
+        log.fix_before_start()
+
+        if exe == "simulation":
+            button.simulate(provider.name, request="call_instance_with_message")
+        else:
+            view = navigate_to(provider, "Details")
+            custom_button_group = Dropdown(view, group.hover)
+            custom_button_group.item_select(button.text)
+
+        log.wait_for_log_validation()
 
 
 @pytest.mark.meta(blockers=[BZ(1719282, unblock=lambda button_type: button_type != "User")])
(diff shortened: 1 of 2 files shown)


cfme/tests/automate/custom_button/test_buttons.py
```diff
@@ -589,7 +590,46 @@ def test_attribute_override():
     Bugzilla:
         1651099
```
john-dupuy: Can this BZ be added as metadata? I.e. `@pytest.mark.meta(automates=[1651099])`

cfme/tests/automate/custom_button/test_buttons.py
```diff
@@ -589,7 +590,46 @@ def test_attribute_override():
...
+
+    collection = appliance.collections.button_groups
+    group = collection.create(
+        text="grp_{}".format(fauxfactory.gen_alphanumeric(3)),
+        hover="hover_{}".format(fauxfactory.gen_alphanumeric(3)),
+        type=getattr(collection, "PROVIDER"),
+    )
+    request.addfinalizer(group.delete_if_exists)
+
+    button = group.buttons.create(
```
john-dupuy: Is there a reason you create these objects in the test rather than in fixtures? Talking about `button` and `group`.

cfme/tests/automate/custom_button/test_buttons.py
```diff
@@ -589,7 +590,46 @@ def test_attribute_override():
...
+    for exe, msg in executions.items():
+        pattern = "[miqaedb:/System/Request/TestNotification#{msg}]".format(msg=msg)
+        log = LogValidator("/var/www/miq/vmdb/log/automation.log", matched_patterns=[pattern])
+        log.fix_before_start()
+
+        if exe == "simulation":
+            button.simulate(provider.name, request="call_instance_with_message")
+        else:
+            view = navigate_to(provider, "Details")
+            custom_button_group = Dropdown(view, group.hover)
```
john-dupuy: Can this `Dropdown` be added in a parametrized view on the `provider` `Details` page? I have not seen widgets instantiated in tests before.

cfme/tests/automate/custom_button/test_buttons.py
digitronik: Yes, `button` object more specific about the test.
while writing the test. what we are testing; it should be part of test not fixture. Here we, are testing `attribute override` and attributes are passing while creating button please have a glance on `L608`.

Yup, a group we can move to fixture; we have one which I'm using where we can pass type of custom button.

cfme/tests/automate/custom_button/test_buttons.py
digitronik: Agreed with this... we thought same while the implementiong custom button testing arch. Then, I thought it should be TODO for me after complete every use cases. Its `inline` declaration of a `widget` and nothing bad in this. Custom button covers 28+ entities toolbars (All and Details) even I was getting a problem with the text while covering all entities.
I want similarity in custom button test.

cfme/tests/automate/custom_button/test_buttons.py
john-dupuy: That makes sense, thanks for the explanation.

cfme/tests/automate/custom_button/test_buttons.py
mshriver: Good thought @john-dupuy, I think you're right that we can just have a parametrized widget on the page that resolves a `Dropdown`

The direct instantiation here works because @digitronik is passing the view to the call, so `__new__` resolves the actual widget, and not just the `WidgetDescriptor`.

cfme/tests/automate/custom_button/test_buttons.py
```diff
@@ -589,7 +591,38 @@ def test_attribute_override():
...
+            custom_button_group.item_select(button.text)
+
+        log.wait_for_log_validation()
```
mshriver: You have defined `log` within the `for` loop above, and then call `wait_for_log_validation()` outside the loop.

I think you'll find, inspecting the test logs, that you're only asserting the last `msg` from the for loop.

You need to take advantage of your updates to `LogValidator` and its counting of the patterns.

You should create `LogValidator` before the loop, passing something like `matched_patterns=['[miqaedb{}]'.format(msg) for msg in executions.values()]` in order to register all of the relevant messages. Probably better to expand that statement though, so that you then have the list of expected strings to compare count against later.

Then fix the log, and iterate over `executions.keys()` to call the custom button `item_select` methods that trigger all the expected log messages.

Then use the `LogValidator` to assert that you got at least one of each message (default wait_for_log_validation behavior)

digitronik: > You have defined `log` within the `for` loop above, and then call `wait_for_log_validation()` outside the loop.

Mike `wait_for_log_validation()` is inside for loop. so its calling two times for both operation means button `execution` and `simulation`
> 
> I think you'll find, inspecting the test logs, that you're only asserting the last `msg` from the for loop.
> 
> You need to take advantage of your updates to `LogValidator` and its counting of the patterns.
> 
> You should create `LogValidator` before the loop, passing something like `matched_patterns=['[miqaedb{}]'.format(msg) for msg in executions.values()]` in order to register all of the relevant messages. Probably better to expand that statement though, so that you then have the list of expected strings to compare count against later.

I like this option; better to use `matches` property, but I need `wait_for` with it. Let's try this :+1: 
> 
> Then fix the log, and iterate over `executions.keys()` to call the custom button `item_select` methods that trigger all the expected log messages.
> 
> Then use the `LogValidator` to assert that you got at least one of each message (default wait_for_log_validation behavior)

izapolsk:
looks good to me but PRT is red. please check reason.

digitronik:
@izapolsk done :+1:

PR #9023, opened 2019-07-01
[1LP][RFR] execute cb on service vm

digitronik:

Purpose or Intent
=================

{{pytest: cfme/tests/automate/custom_button/test_service_vm_custom_button.py::test_custom_button_automate_service_vm}}


cfme/tests/automate/custom_button/test_service_vm_custom_button.py
```diff
@@ -228,3 +228,54 @@ def test_custom_button_with_dynamic_dialog_vm(
...
+        button = button_group.buttons.create(
+            text=fauxfactory.gen_alphanumeric(),
+            hover=fauxfactory.gen_alphanumeric(),
+            system="Request",
+            request="InspectMe",
+        )
+        request.addfinalizer(button.delete_if_exists)
+
+    # Check for UI and SSUI destinations.
+    for context in [ViaUI, ViaSSUI]:
```
john-dupuy: @digitronik thoughts on parameterizing this test case against `context`?

digitronik: right, we have a tradition to parametrize a test with ViaUI` and `ViaSSUI` with the help of `santaku`. but I'm thinking to make it more simple and inclusive.

1. In this case, it's possible also but no need of this as we are checking a single button on different `destinations`.
2. If we did, test time will dobule; which is not good when the test needs more `setup steps` or `timing`. `service_vm` fixture is in `functional` scope. I won't want to spin VM just for button execution on different destinations.

john-dupuy: That makes sense, thanks for the explanation!

cfme/tests/automate/custom_button/test_service_vm_custom_button.py
```diff
@@ -228,3 +228,54 @@ def test_custom_button_with_dynamic_dialog_vm(
             submit = "submit" if context is ViaUI else "submit_request"
             getattr(view, submit).click()
             log.wait_for_log_validation()
+
+
+@pytest.mark.meta(automates=[1427430, 1450473])
```
mshriver: 1454910?

mshriver: Talked to Nikhil, its a clone.

PR #9024, opened 2019-07-01
[1LP][RFR]Automate Test: test_new_report_fields

valaparthvi:
{{ pytest: cfme/tests/intelligence/reports/test_reports.py::test_new_report_fields --use-template-cache -sqvvv }}

### DIFF
--- a/cfme/tests/intelligence/reports/test_reports.py
+++ b/cfme/tests/intelligence/reports/test_reports.py
@@ -7,14 +7,6 @@ from cfme.utils.rest import assert_response
 
 pytestmark = [test_requirements.report, pytest.mark.tier(3), pytest.mark.sauce]
 
-REPORT_BASED_ON = [
-    ("Floating IPs", ["Address", "Status", "Cloud Manager : Name"]),
-    (
-        "Cloud Tenants",
-        ["Name", "My Company Tags : Owner", "My Company Tags : Cost Center"],
-    ),
-]
-
 
 @pytest.fixture
 def create_custom_tag(appliance):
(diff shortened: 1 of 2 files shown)


cfme/tests/intelligence/reports/test_reports.py
```diff
@@ -105,3 +115,36 @@ def test_reports_custom_tags(appliance, request, create_custom_tag):
     report = appliance.collections.reports.create(**report_data)
     request.addfinalizer(report.delete)
     assert report.exists
+
+
+@test_requirements.report
+@pytest.mark.tier(0)
+@pytest.mark.ignore_stream("5.10")
+@pytest.mark.parametrize("based_on", REPORT_BASED_ON)
+@pytest.mark.meta(automates=["1546927", "1504155"])
```
john-dupuy: @valaparthvi No string `""` necessary here, just the ID is fine.

cfme/tests/intelligence/reports/test_reports.py
```diff
@@ -105,3 +115,36 @@ def test_reports_custom_tags(appliance, request, create_custom_tag):
     report = appliance.collections.reports.create(**report_data)
     request.addfinalizer(report.delete)
     assert report.exists
+
+
+@test_requirements.report
+@pytest.mark.tier(0)
+@pytest.mark.ignore_stream("5.10")
+@pytest.mark.parametrize("based_on", REPORT_BASED_ON)
```
john-dupuy: Q: why parametrize the test case in this way? I.e. why not just use `REPORT_BASED_ON` in the test case? Just curious

valaparthvi: Um, no reason in particular actually. Earlier I was planning on using a dictionary, so I had parametrized it that way, and I guess I forgot to change that later. But better to change it, I guess.

PR #9026, opened 2019-07-02
[1LP][RFR] removed BZ1717902

lina-nikiforova:
The tests would fail (skipped due to blockers) anyway because of the other BZs.


lina-nikiforova:
PRT failures are not related to the change

PR #9027, opened 2019-07-02
[1LP][RFR] Automate test: test_infrastructure_hosts_icons_states

valaparthvi:
Changes:
1. Add a `rest_api_entity` property for Hosts.
2. Automate tests: test_infrastructure_hosts_icons_states

{{ pytest: cfme/tests/infrastructure/test_host.py::test_infrastructure_hosts_icons_states --use-template-cache -sqvvv }}


valaparthvi:
Note: I need help automating this test. I was able to manually test it, but to automate it, I need help figuring out how to access `vmdb_production` and then updating the host's power_state. I tried `appliance.ssh_client.run_command("psql -d -t vmdb_production)` but running that command tends to give `TimedOutError`.

cfme/infrastructure/host.py
```diff
@@ -456,6 +456,10 @@ def get_credentials_from_config(cls, credentials_config):
         return {cred_type: cls.Credential.from_config(cred) for
                 cred_type, cred in credentials_config.items()}
 
+    @property
+    def rest_api_entity(self):
+        return self.appliance.rest_api.collections.hosts.get(name=self.name)
```
digitronik: :+1:

mshriver:
@valaparthvi concerning DB updates - we must use the dbclient object that is attached to the appliance object for making these queries. This uses sqlalchemy under-the-hood to make the queries, instead of trying to run psql on the CLI over SSH.

cfme/tests/infrastructure/test_host.py
```diff
@@ -248,3 +248,63 @@ def test_250_vmware_hosts_loading(appliance, create_250_hosts, view_type):
...
+                and host(Host's Detail page) power_state changes after running the command:
+                `UPDATE hosts SET power_state = ':power_state' WHERE name=':host_name';`
+    """
+    # get host and host details
+    host = provider.hosts.all()[0]
+    host_name = host.name
+    reset_state = host.rest_api_entity.power_state
+
+    command = (
+        "psql vmdb_production -c \"UPDATE hosts SET power_state='{state}'"
```
mshriver: This must be done through `appliance.db` object.

Test pattern is good though - we don't actually want to try and modify the host states, so directly injecting the states that we want to test into the DB is a good alternative.

cfme/tests/infrastructure/test_host.py
```diff
@@ -248,3 +248,63 @@ def test_250_vmware_hosts_loading(appliance, create_250_hosts, view_type):
...
+    )
+    # change host power_state
+    result = appliance.ssh_client.run_command(
+        command.format(state=power_state, name=host_name)
+    )
+    assert result.success
+
+    # reset host power_state
+    request.addfinalizer(
+        lambda: appliance.ssh_client.run_command(
```
mshriver: I think this 'reset' should be done by running a provider refresh. Modifying that host state could have impacts on other VMDB objects, like archived/orphaned state of VMs. It would be good to actually do a full evm refresh to make sure the VMDB is fully updated.

valaparthvi: Hey @mshriver I'd rather not do a full evm refresh. The power state of a host cannot be changed while the provider is refreshing and refreshing a provider at the end of every test causes test failure after the first test passes. If you still insist I can add a fixture to do a full evm refresh once all the tests have been executed.

dajoRH:
*CFME QE Bot*

cfme/tests/infrastructure/test_host.py
```diff
@@ -248,3 +248,68 @@ def test_250_vmware_hosts_loading(appliance, create_250_hosts, view_type):
     view.entities.paginator.set_items_per_page(1000)
     view.toolbar.view_selector.select(view_type)
     wait_for(view.entities.get_first_entity, timeout=60, message='Wait for the view')
+
+
+@test_requirements.general_ui
```
jawatts: **Question** Any BZ associated with this?

cfme/tests/infrastructure/test_host.py
```diff
@@ -248,3 +248,68 @@ def test_250_vmware_hosts_loading(appliance, create_250_hosts, view_type):
...
+@pytest.mark.provider([RHEVMProvider], selector=ONE, scope="module", override=True)
```
jawatts: Is Vmware different? Curious why just RHEV here.

cfme/tests/infrastructure/test_host.py
valaparthvi: I'm not really sure. I picked this test from `test_manual.py` ,there was no BZ mentioned.

cfme/tests/infrastructure/test_host.py
valaparthvi: No particular reason. But it was explicitly mentioned in the test to use RHEV. But I guess there's no harm in adding other infra providers.

valaparthvi: Hey @jawatts, I've modified the test to use both infra providers.

jawatts: @valaparthvi Thanks!

jawatts:
@valaparthvi I am good with this, handing over to @mshriver for merge since there is an outstanding question for him to answer

cfme/tests/infrastructure/test_host.py
mshriver: @valaparthvi sorry for the delay in response - I think your idea  for a module level fixture for that evm refresh in teardown is a good idea.

You can protect against failure in the test cases following this refresh by using wait_for around the provider.is_refreshed property.

cfme/tests/infrastructure/test_host.py
```diff
@@ -248,3 +248,67 @@ def test_250_vmware_hosts_loading(appliance, create_250_hosts, view_type):
...
+                and host(Host's Detail page) power_state changes after running the command:
+                `UPDATE hosts SET power_state = ':power_state' WHERE name=':host_name';`
+    """
+    # get host and host details
+    host = provider.hosts.all()[0]
+    host_name = host.name
+    reset_state = host.rest_api_entity.power_state
+    hosts = appliance.db.client["hosts"]
+
+    # change host power_state
```
mshriver: :tada: Well done!

cfme/tests/infrastructure/test_host.py
```diff
@@ -248,3 +248,67 @@ def test_250_vmware_hosts_loading(appliance, create_250_hosts, view_type):
...
+    assert result == 1
+
+    # reset host power_state
+    @request.addfinalizer
+    def _finalize():
+        appliance.db.client.session.query(hosts).filter(hosts.name == host_name).update(
+            {hosts.power_state: reset_state}
+        )
+
+    # assert power_state from quadicon
```
mshriver: Good coverage!

PR #9031, opened 2019-07-03
[1LP][RFR] Refactor LogValidator

digitronik:

Purpose or Intent
=================
- Refactor method names as per converstation [PR-8918](https://github.com/ManageIQ/integration_tests/pull/8918#issuecomment-504006364)
    - `fix_before_start()` > `start_monitoring()`
    - `validate_logs()` > `validate()`
    - `wait_for_log_validation` > `validate(wait="....")`

{{pytest: cfme/tests -k 'test_custom_button_with_dynamic_dialog_vm or test_custom_button_automate_service_vm or test_domain_lock_disabled or test_custom_button_automate_infra_obj or test_custom_button_dialog_infra_obj or test_check_system_request_calls_depr_conf_mgmt or test_quota_source_value or test_task_id_for_method_automation_log or test_send_email_method or test_automate_generic_object_service_associations or test_automate_service_quota_runs_only_once or test_service_provisioning_email or test_codename_in_log or test_appliance_console_external_auth or test_appliance_console_external_auth_all or test_action_prevent_ssa or test_action_prevent_host_ssa or test_action_power_on_logged or test_action_power_on_audit or test_service_retirement_methods or test_service_ansible_verbosity or test_appliance_console_cli_external_auth or test_configuration_database_garbage_collection or test_blacklisted_container_events or test_pause_and_resume_single_provider_api or test_user_requester_for_lifecycle_provision or test_automate_relationship_trailing_spaces' -vv --long-running}}

### DIFF
--- a/cfme/tests/automate/test_instance.py
+++ b/cfme/tests/automate/test_instance.py
@@ -127,7 +127,6 @@ def test_automate_relationship_trailing_spaces(request, klass, namespace, domain
         initialEstimate: 1/10h
         caseimportance: medium
         caseposneg: positive
-        caseposneg: positive
         testtype: functional
         startsin: 5.9
         casecomponent: Automate
--- a/cfme/tests/automate/test_instance.py
+++ b/cfme/tests/automate/test_instance.py
@@ -127,6 +127,7 @@ def test_automate_relationship_trailing_spaces(request, klass, namespace, domain
         initialEstimate: 1/10h
         caseimportance: medium
         caseposneg: positive
+        caseposneg: positive
         testtype: functional
         startsin: 5.9
         casecomponent: Automate
@@ -206,7 +207,7 @@ def test_automate_relationship_trailing_spaces(request, klass, namespace, domain
         "/var/www/miq/vmdb/log/automation.log", matched_patterns=[".*{}.*".format(catch_string)],
         failure_patterns=[".*ERROR.*"]
     )
-    result.fix_before_start()
+    result.start_monitoring()
 
     # Executing the automate method of klass1 using simulation
     simulate(
@@ -218,7 +219,7 @@ def test_automate_relationship_trailing_spaces(request, klass, namespace, domain
             "instance": instance2.name,
         },
     )
-    result.validate_logs()
+    assert result.validate(wait="60s")
 
 
 @pytest.fixture(scope="module")
@@ -278,7 +279,7 @@ def test_check_system_request_calls_depr_conf_mgmt(appliance, copy_instance):
         appliance=appliance,
         request=copy_instance.name
     )
-    assert result.validate()
+    assert result.validate(wait="60s")
 
 
 @pytest.fixture(scope="module")
@@ -353,4 +354,4 @@ def test_quota_source_value(request, entity, search, copy_quota_instance, generi
     service_catalogs.order()
     provision_request.wait_for_request(method='ui')
     request.addfinalizer(lambda: provision_request.remove_request(method="rest"))
-    assert result.validate()
+    assert result.validate(wait="60s")
--- a/cfme/tests/automate/test_instance.py
+++ b/cfme/tests/automate/test_instance.py
@@ -272,13 +272,13 @@ def test_check_system_request_calls_depr_conf_mgmt(appliance, copy_instance):
     result = LogValidator(
         "/var/www/miq/vmdb/log/automation.log", matched_patterns=[".*{}.*".format(search)]
     )
-    result.fix_before_start()
+    result.start_monitoring()
     # Executing the automate instance - 'ansible_tower_job' using simulation
     simulate(
         appliance=appliance,
         request=copy_instance.name
     )
-    result.validate_logs()
+    assert result.validate()
 
 
 @pytest.fixture(scope="module")
@@ -341,7 +341,7 @@ def test_quota_source_value(request, entity, search, copy_quota_instance, generi
     result = LogValidator(
         "/var/www/miq/vmdb/log/automation.log", matched_patterns=[".*{}.*".format(search)]
     )
-    result.fix_before_start()
+    result.start_monitoring()
     service_catalogs = ServiceCatalogs(
         copy_quota_instance.appliance, catalog=generic_catalog_item.catalog,
         name=generic_catalog_item.name
@@ -353,4 +353,4 @@ def test_quota_source_value(request, entity, search, copy_quota_instance, generi
     service_catalogs.order()
     provision_request.wait_for_request(method='ui')
     request.addfinalizer(lambda: provision_request.remove_request(method="rest"))
-    result.validate_logs()
+    assert result.validate()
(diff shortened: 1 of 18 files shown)


digitronik:
I found some failures but not exactly related to changes... it's failing in other 511 builds too.

cfme/utils/log_validator.py
```diff
@@ -85,28 +110,22 @@ def matches(self):
...
-            this function since _verify_match_logs raise pytest.fail() when it fails to find the
-            match pattern in the logs.
+    def validate(self, wait=None, message="waiting for log validation", **kwargs):
+        """Validate log pattern
 
-            Note that you must call fix_before_start() before making use of this function.
+        Args:
+            wait: Wait for log validation (timeout)
+            message: Specific message.
+        Returns (bool): True if expected pattern matched in log else False
```
izapolsk: I'm not sure I understand when it will return False.

cfme/utils/log_validator.py
```diff
@@ -85,28 +110,22 @@ def matches(self):
...
+        Raise:
+            TimeOutError: If failed to match pattern in respective timeout
```
izapolsk: wrong exception. right exception is called TimedOutError

cfme/tests/automate/test_instance.py
```diff
@@ -127,6 +127,7 @@ def test_automate_relationship_trailing_spaces(request, klass, namespace, domain
         initialEstimate: 1/10h
         caseimportance: medium
         caseposneg: positive
+        caseposneg: positive
```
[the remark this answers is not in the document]
digitronik: aaah by mistake ctl+D

cfme/utils/log_validator.py
```diff
@@ -49,32 +60,46 @@ def __init__(self, remote_filename, **kwargs):
...
 
     def _check_match_logs(self, line):
         for pattern in self.matched_patterns:
             if re.search(pattern, line):
-                logger.info('Expected pattern {} was matched on line {}'.format(pattern, line))
+                logger.info("Expected pattern %s was matched on line %s", pattern, line)
                 self._matches[pattern] = self._matches[pattern] + 1
 
+    @property
+    def _validate(self):
```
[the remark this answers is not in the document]
digitronik: @izapolsk yup, It's not for a user; I used `_` before the name start. we have `validate` method. I moved it to `_is_valid`

cfme/utils/log_validator.py
digitronik: right actually I was returning an output of `func` which was wrong.
Now in case of wait, it will return `True` if wait_for success else `TimedOutError` and without wait it will return `bool`

cfme/utils/log_validator.py
```diff
@@ -1,17 +1,26 @@
...
-import pytest
-from _pytest.outcomes import Failed
-
 from .ssh import SSHTail
 from cfme.utils.log import logger
 from cfme.utils.wait import wait_for
 
 
+class LogValidatorException(Exception):
+    """Its custom exception related to LogValidator"""
```
john-dupuy: _Optional_ I would just leave this as `"""Custom exception for LogValidator"""`

cfme/utils/log_validator.py
```diff
@@ -37,8 +48,8 @@ class LogValidator(object):
                                   skip_patterns=['PARTICULAR_ERROR'],
                                   failure_patterns=['.*ERROR.*'],
                                   matched_patterns=['PARTICULAR_INFO'])
-          evm_tail.fix_before_start()
-          evm_tail.validate_logs()
+          evm_tail.start_monitoring()
+          evm_tail.validate()       # evm_tail.validate(wait="30s")
```
john-dupuy: Thoughts on adding a line or two about appropriate kwargs to pass for `wait`? E.g. 
```python,
"""
It is also possible for  wait for log validation, the proper syntax for doing so is:
evm_tail.validate(wait="30s")
The appropriate strings to pass to wait are ... x,y, z
"""
```

cfme/utils/log_validator.py
```diff
@@ -49,32 +60,46 @@ def __init__(self, remote_filename, **kwargs):
...
+                logger.info("Expected '%s' pattern not found", pattern)
+                return False
+        return True
+
     @property
     def matches(self):
-        """It will return pattern match count dictionary"""
+        """Collect match count in log
+
+        Returns (dict): Pattern match count dictionary
```
john-dupuy: _Optional_ thoughts on also giving example of the `key`, and `values` of the dict? I.e. Key is the `match` and value is the `count`?

cfme/utils/log_validator.py
```diff
@@ -85,28 +110,20 @@ def matches(self):
...
-        """ Wait for log validation, takes the kwargs as wait_for. This function will reduce
-            duplicate functions in tests that wait_for log_validation. It is necessary to create
-            this function since _verify_match_logs raise pytest.fail() when it fails to find the
-            match pattern in the logs.
+    def validate(self, wait=None, message="waiting for log validation", **kwargs):
+        """Validate log pattern
 
-            Note that you must call fix_before_start() before making use of this function.
+        Args:
+            wait: Wait for log validation (timeout)
```
john-dupuy: Question: why not call this `timeout`? Then we will match `wait_for` kwargs.

digitronik: to maintain similarity in framework

john-dupuy: Makes sense, I had forgot that we use `wait` for the `create_view`!

cfme/utils/log_validator.py
digitronik: John thanks for this, but we try to cover simple usecase as example in doc-string mainly when use `code-block`. I think its sufficient to pass usecase of logvalidator to user and after that he/she can checks other available methods like matches etc....

Let me know if you still want this; I will enhance this `code-block`.

john-dupuy: That's fine with me, I would consider it optional. As it matches with the usage of `wait` elsewhere in the codebase.

digitronik:
@izapolsk added changes please check those onetime.

cfme/utils/log_validator.py
```diff
@@ -1,17 +1,26 @@
...
 
-import pytest
-from _pytest.outcomes import Failed
-
 from .ssh import SSHTail
 from cfme.utils.log import logger
 from cfme.utils.wait import wait_for
 
 
+class LogValidatorException(Exception):
```
izapolsk: I'd ask you to follow pep 8 conventions in naming exceptions. 
https://www.python.org/dev/peps/pep-0008/#exception-names

digitronik: @izapolsk LogValidator not raising actual error. If `fail patterns` matched rather than fail test with `pytest`; we are raising custom exception. `Error` at end of Exception point actual error occur in code as per me. Let me know view on it.

izapolsk: actually this is not stopper but I'd rather use something like PatternNotMatchError.

PR #9033, opened 2019-07-03
[1LP][RFR] Fix test_provision_approval and Request views

mshriver:
rework its assertion patterns
Fix request view.is_displayed methods

PRT results:
Passing in 5.10, 1 pass + 1 failure + 2 BZ skip in 5.11

{{ pytest: --use-provider rhv43 --use-provider ec2west --long-running -vvvv cfme/tests/cloud_infra_common/test_provisioning.py::test_provision_approval }}

### DIFF
--- a/cfme/services/requests.py
+++ b/cfme/services/requests.py
@@ -15,6 +15,7 @@ from widgetastic_patternfly import Input
 from cfme.base.login import BaseLoggedInPage
 from cfme.common.vm_views import BasicProvisionFormView
 from cfme.common.vm_views import ProvisionView
+from cfme.exceptions import displayed_not_implemented
 from cfme.exceptions import ItemNotFound
 from cfme.exceptions import RequestException
 from cfme.modeling.base import BaseCollection
@@ -223,7 +224,7 @@ class Request(BaseEntity):
 
     @is_succeeded.variant('ui')
     def is_succeeded_ui(self):
-        return self.is_finished(method=('ui')) and self.row.status.text == 'Ok'
+        return self.is_finished(method='ui') and self.row.status.text == 'Ok'
 
     def copy_request(self, values=None, cancel=False):
         """Copies the request  and edits if needed
@@ -277,9 +278,10 @@ class RequestBasicView(BaseLoggedInPage):
 
     @property
     def in_requests(self):
-        return self.logged_in_as_current_user
-        # TODO uncomment after BZ 1472844 is fixed
-        # and  self.navigation.currently_selected == ['Services', 'Requests'] and
+        return (
+            self.logged_in_as_current_user
+            and self.navigation.currently_selected == ['Services', 'Requests']
+        )
 
 
 class RequestsView(RequestBasicView):
@@ -309,15 +311,15 @@ class RequestsView(RequestBasicView):
                 # it might be nice to add an option to fail at this point
                 continue
             elif len(rows) > 1:
-                raise RequestException(
-                    'Multiple requests with matching content found - be more specific!')
+                raise RequestException('Multiple requests with matching content found - '
+                                       'be more specific!')
             else:
                 # found the row!
                 row = rows[0]
                 logger.debug(' Request Message: %s', row.last_message.text)
                 return row
         else:
-            raise Exception("The requst specified by {} not found!".format(str(cells)))
+            raise Exception("The request specified by {} not found!".format(str(cells)))
 
     @property
     def is_displayed(self):
@@ -430,8 +432,12 @@ class RequestDetailsView(RequestsView):
 
     @property
     def is_displayed(self):
-        return (self.in_requests and
-                self.title.text == self.context['object'].rest.description)
+        expected_description = self.context['object'].rest.description
+        return (
+            self.in_requests and
+            self.breadcrumb.locations[-1] == expected_description and
+            self.title.text == expected_description
+        )
 
     breadcrumb = BreadCrumb()
     toolbar = RequestDetailsToolBar()
@@ -447,8 +453,9 @@ class RequestApprovalView(RequestDetailsView):
     def is_displayed(self):
         try:
             return (
-                self.breadcrumb.locations[1] == self.context['object'].rest.description and
-                self.breadcrumb.locations[2] == 'Request Approval')
+                self.breadcrumb.locations[-1] == 'Request Approval' and
+                self.breadcrumb.locations[-2] == self.context['object'].rest.description
+            )
         except Exception:
             return False
 
@@ -463,8 +470,9 @@ class RequestDenialView(RequestDetailsView):
     def is_displayed(self):
         try:
             return (
-                self.breadcrumb.locations[1] == self.context['object'].rest.description and
-                self.breadcrumb.locations[2] == 'Request Denial')
+                self.breadcrumb.locations[-1] == 'Request Denial' and
+                self.breadcrumb.locations[-2] == self.context['object'].rest.description
+            )
         except Exception:
             return False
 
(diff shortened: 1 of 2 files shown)


cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -240,22 +231,21 @@ def test_provision_approval(appliance, provider, vm_name, smtp_test, request,
...
+        wait_for(lambda: len(smtp_test.get_emails()) >= 2,
+                 num_sec=90,
+                 delay=3)
     except TimedOutError:
-        subjects = ",".join([normalize_text(m["subject"]) for m in smtp_test.get_emails()])
-        logger.error("expected: %s, got emails: %s", subject, subjects)
-        raise
+        pytest.fail('Did not receive at least 1 emails from provisioning request, received: {}'
+                    .format(smtp_test.get_emails()))
+    received_subjects = [normalize_text(m["subject"]) for m in smtp_test.get_emails()]
```
mshriver: I'm not the biggest fan of my own naming/pattern here, if anyone can suggest better names or a more clear way to format this I'm all for it.  Gets a little messy with applying `normalize_text` on everything, and having to iterate in multiple layers (checking for 2 strings, checking for partial match of each string in a list of received strings).

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -169,63 +168,55 @@ def test_provision_approval(appliance, provider, vm_name, smtp_test, request,
...
+                 delay=3)
     except TimedOutError:
-        subjects = ",".join([normalize_text(m["subject"]) for m in smtp_test.get_emails()])
-        logger.error("expected: %s, got emails: %s", subject, subjects)
-        raise
+        pytest.fail('Did not receive at least 2 emails from provisioning request, received: {}'
+                    .format(smtp_test.get_emails()))
+
+    expected_subject = normalize_text("your {} request is pending".format(vm_type))
+    from_subject = normalize_text("{} request from {}pending approval".format(vm_type, requester))
```
john-dupuy: Space between `{}` and `pending`

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -169,63 +168,55 @@ def test_provision_approval(appliance, provider, vm_name, smtp_test, request,
...
+
+    received_subjects = [normalize_text(m["subject"]) for m in smtp_test.get_emails()]
+    for subject in [expected_subject, from_subject]:
+        soft_assert(any(subject in r_sub for r_sub in received_subjects),
```
john-dupuy: _optional_ formatting as discussed above

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -240,22 +231,21 @@ def test_provision_approval(appliance, provider, vm_name, smtp_test, request,
...
+
     try:
-        wait_for(
-            lambda:
-            len(filter(
-                lambda mail:
-                normalize_text(subject) in normalize_text(mail["subject"]),
-                smtp_test.get_emails())) == 1,
-            num_sec=120, delay=5)
+        wait_for(lambda: len(smtp_test.get_emails()) >= 2,
```
john-dupuy: Why does `smtp_test.get_emails()` need to be called twice? Is there a reason it isn't stored as a variable?

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -240,22 +231,21 @@ def test_provision_approval(appliance, provider, vm_name, smtp_test, request,
...
+                 num_sec=90,
+                 delay=3)
     except TimedOutError:
-        subjects = ",".join([normalize_text(m["subject"]) for m in smtp_test.get_emails()])
-        logger.error("expected: %s, got emails: %s", subject, subjects)
-        raise
+        pytest.fail('Did not receive at least 1 emails from provisioning request, received: {}'
+                    .format(smtp_test.get_emails()))
+    received_subjects = [normalize_text(m["subject"]) for m in smtp_test.get_emails()]
+    from_subject = normalize_text("{} request from {}was approved".format(vm_type, requester))
```
john-dupuy: Space between `{}` and `was`

cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -269,24 +259,24 @@ def test_provision_approval(appliance, provider, vm_name, smtp_test, request,
...
     # Wait for e-mails to appear
-    def verify():
-        return (
-            len(filter(
-                lambda mail: normalize_text(subject) in normalize_text(mail["subject"]),
-                smtp_test.get_emails())) == len(vm_names)
-        )
     try:
-        wait_for(verify, message="email receive check", delay=5)
+        wait_for(lambda: len(smtp_test.get_emails()) >= expected_subject_count,
```
john-dupuy: _optional_ formatting

cfme/tests/cloud_infra_common/test_provisioning.py
valaparthvi: Hi @mshriver, these changes look fine to me in general, but I think the expected subject needs to be changed. This is the traceback I received from when I ran this test locally on a 511 appliance - 
```
E           SoftAssertionError: 
E           Expected subject [your virtual machine request is pending], not matched in received subjects [[u'request id 2 your virtual machine request is pending', u'request id 2 virtual machine request from vm provisioncfmeqecom pending approval']] (cfme/tests/cloud_infra_common/test_provisioning.py:204)
E           Expected subject [virtual machine request from vm provisioncfmeqecom pending approval], not matched in received subjects [[u'request id 2 your virtual machine request is pending', u'request id 2 virtual machine request from vm provisioncfmeqecom pending approval']] (cfme/tests/cloud_infra_common/test_provisioning.py:204)
E           Expected subject [your virtual machine request was approved], not matched in received subjects [[u'request id 2 virtual machine request from vm provisioncfmeqecom was approved pending quota validation', u'request id 2 your virtual machine request was approved pending quota validation']] (cfme/tests/cloud_infra_common/test_provisioning.py:248)
E           Expected subject [virtual machine request from vm provisioncfmeqecom was approved], not matched in received subjects [[u'request id 2 virtual machine request from vm provisioncfmeqecom was approved pending quota validation', u'request id 2 your virtual machine request was approved pending quota validation']] (cfme/tests/cloud_infra_common/test_provisioning.py:248)

cfme/fixtures/soft_assert.py:64: SoftAssertionError
```

I think you might want to use different messages depending on the appliance version.

cfme/services/requests.py
```diff
@@ -479,33 +486,35 @@ class form(BasicProvisionFormView):  # noqa
...
-
-    @property
-    def is_displayed(self):
-        try:
-            return (
-                self.breadcrumb.locations[1] == self.context['object'].rest.description and
-                self.breadcrumb.locations[2] == 'Edit VM Provision')
-        except Exception:
-            return False
+    # BZ 1726741 will impact the breadcrumb items
```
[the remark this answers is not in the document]
mshriver: Right now there isn't enough information in the breadcrumb to uniquely identify the page, so leaving it not implemented until that information is available.

cfme/tests/cloud_infra_common/test_provisioning.py
mshriver: The space should be included in the variable formatted into the string - IIRC 5.10 does not include the requester email

cfme/tests/cloud_infra_common/test_provisioning.py
mshriver: Agreed - I noticed that the use of `any()` here is impacting the traceback when the assertion fails.

mshriver: I think the traceback you've included here is from an older commit, before using `any()`. The only PRT failure is in 5.11 (run 33039). I'll still investigate though, thanks @john-dupuy @valaparthvi

cfme/tests/cloud_infra_common/test_provisioning.py
izapolsk: it is present in requester :). Could we just move that space from requested to here ?

mshriver: Its dependent on cloud vs infra, and I would rather set it there instead of having a separate if statement where these are needed.

cfme/tests/cloud_infra_common/test_provisioning.py
mshriver: We're calling it in the wait_for loop, but inside of the lambda, so wait_for isn't returning it when the wait is completed.  I have to call it again to actually get the contents of its return.

cfme/tests/cloud_infra_common/test_provisioning.py
mshriver: This is included in `requester`, will continue discussion on other comments concerning this and apply decision here too.

PR #9039, opened 2019-07-05
[1LP][RFR] fixed random tag selection for 5.11

lina-nikiforova:
{{ pytest: -v cfme/tests/configure/test_access_control.py::test_user_remove_tag }}

`all_options` selected some wrong (invalid) options, so I modified the locator.
The other problem was with encoding, that is also fixed but please let me know if there's a better solution.

### DIFF
--- a/cfme/common/__init__.py
+++ b/cfme/common/__init__.py
@@ -305,14 +305,21 @@ class TaggableCommonBase(object):
     """Class represents common functionality for tagging via collection and entities pages"""
 
     def _set_random_tag(self, view):
-        random_cat = random.choice(view.form.tag_category.all_options).text
-        # '*' is added in UI almost to all categoly while tag selection,
+        if self.appliance.version > "5.11":
+            random_cat = random.choice(view.form.tag_category.all_options)
+        else:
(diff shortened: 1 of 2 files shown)


cfme/common/__init__.py
```diff
@@ -305,14 +305,21 @@ class TaggableCommonBase(object):
     """Class represents common functionality for tagging via collection and entities pages"""
 
     def _set_random_tag(self, view):
-        random_cat = random.choice(view.form.tag_category.all_options).text
-        # '*' is added in UI almost to all categoly while tag selection,
+        if self.appliance.version > "5.11":
+            random_cat = random.choice(view.form.tag_category.all_options)
```
digitronik: optional: but just enthusiastic, is it not possible to implement `all_options` method of `ReactSelect` similar to `BootstrapSelect` means Options will be namedtuple of `text` and `web-element`?

john-dupuy: I agree with @digitronik here, is there a reason why this can't be done?

mshriver: Yeah this is a PITA, and obscure because its not really tied to the appliance version, its tied to which widget is being used

PR #9046, opened 2019-07-08
[1LP][RFR] UI change for Hosts & Clusters tab

lina-nikiforova:
{{ pytest: -v cfme/tests/configure/test_access_control.py::test_group_crud_with_tag }}


cfme/configure/access_control/__init__.py
```diff
@@ -268,22 +268,20 @@ def delete(self, cancel=True):
                 delete_user_txt))
 
         view.toolbar.configuration.item_select(delete_user_txt, handle_alert=cancel)
-        if not BZ(1721424, forced_streams=["5.11"]).blocks:
```
john-dupuy: @lina-nikiforova I see that this BZ was marked as a duplicate of 1720273, does that BZ no longer apply to this code? (I'm assuming not since PRT is green)

PR #9048, opened 2019-07-09
[1LP][RFR] adding new manual subscription tests

izapolsk:


cfme/tests/configure/test_config_manual.py
```diff
@@ -100,3 +100,110 @@ def test_configuration_dropdown_roles_by_server():
         1455283
     """
     pass
+
+
+@pytest.mark.manual
+@pytest.mark.tier(3)
+def test_replication_subscription_crud():
+    """
+    Add/Edit/Remove replication subscription
```
john-dupuy: @izapolsk Can you put these test steps in the `testSteps` field? With `expectedResults` included? Or is there a reason you've not done that here?

izapolsk: only reason is my laziness. I'm anyway gonna automate this stuff. fixed.

PR #9049, opened 2019-07-09
[1LP][RFR] added forced streams for BZ1707328

lina-nikiforova:
This BZ is also valid for 5.11.

{{ pytest: -v cfme/tests/configure/test_tag.py::test_map_tagging_crud }}

### DIFF
--- a/cfme/tests/configure/test_tag.py
+++ b/cfme/tests/configure/test_tag.py
@@ -121,7 +121,7 @@ def test_map_tagging_crud(appliance, category, soft_assert):
 
     map_tag_entity.delete()
     view = appliance.browser.create_view(navigator.get_class(map_tags_collection, 'All').VIEW)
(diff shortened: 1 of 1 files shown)


cfme/tests/configure/test_tag.py
```diff
@@ -121,7 +121,7 @@ def test_map_tagging_crud(appliance, category, soft_assert):
 
     map_tag_entity.delete()
     view = appliance.browser.create_view(navigator.get_class(map_tags_collection, 'All').VIEW)
-    if not BZ(1707328).blocks:
+    if not BZ(1707328, forced_streams=["5.10", "5.11"]).blocks:
```
john-dupuy: @lina-nikiforova Forced streams shouldn't be necessary for 5.10

```
BZ(1707328).blocks
Out[4]: True```

PR #9051, opened 2019-07-10
[1LP][RFR][NOTEST] Added manual test case for schedules

ganeshhubale:
Added test case for BZ - 1713072


cfme/tests/automate/test_automate_manual.py
```diff
@@ -989,3 +989,38 @@ def test_simulation_copy_button():
         1630800
     """
     pass
+
+
+@pytest.mark.tier(2)
```
john-dupuy: You can add coverage meta data here: `@pytest.mark.meta(coverage=[1713072])`

PR #9052, opened 2019-07-10
[1LP][RFR] Testing multiple automation requests

ganeshhubale:
- Testing multiple automation requests and denying

{{ pytest: cfme/tests/infrastructure/test_rest_automation_request.py -k 'test_multiple_automation_requests' -vv --long-running }}

### DIFF
--- a/cfme/tests/infrastructure/test_rest_automation_request.py
+++ b/cfme/tests/infrastructure/test_rest_automation_request.py
@@ -8,6 +8,7 @@ from cfme.infrastructure.provider import InfraProvider
 from cfme.markers.env_markers.provider import ONE
 from cfme.rest.gen_data import automation_requests_data as _automation_requests_data
 from cfme.rest.gen_data import vm as _vm
+from cfme.utils.blockers import BZ
 from cfme.utils.rest import assert_response
 from cfme.utils.rest import query_resource_attributes
 from cfme.utils.wait import wait_for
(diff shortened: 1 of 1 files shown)


dajoRH:
*CFME QE Bot*

cfme/tests/infrastructure/test_rest_automation_request.py
```diff
@@ -257,6 +258,23 @@ def test_edit_requests(self, collection, appliance, requests_pending, from_detai
...
+        """
+        Polarion:
+            assignee: ghubale
+            initialEstimate: 1/30h
+            casecomponent: Automate
+
+        Bugzilla:
+            1723830
+        """
+        requests_data = _automation_requests_data(vm, approve=False, num=2000)
```
john-dupuy: @ganeshhubale I see that the BZ was written when a customer submitted 60000 requests, is 2000 enough? Thoughts on parametrizing to more? e.g. `[2000, 10,000, 30,000, 60,000]`

ganeshhubale: Thanks @john-dupuy for going through BZ.
- According to me, parametrizing this test case with different number of requests is not actually a use case. Here we want to check if huge number of automation requests provisioned are getting denied or not.
- Yes, we can increase the number of requests up to the number we want. So I updated it to 60,000 and making this PR long_running(Although this is rest test case)

mshriver: Restarting PRT with `--long-running` included in the options, as this change caused the test to not be collected in run 33445

PR #9056, opened 2019-07-10
[1LP][WIPTEST]fixing provider refresh with adding wait_for so that hosts_after count is accurate

kedark3:
Purpose or Intent
=================

__Fixing__ Test test_rebuilt_vcenter_duplicate_hosts as it was passing with false_positives as the test was not waiting for provider to be refreshed. 
Also, adding ONE_PER_TYPE provider filter for some of the tests.
{{pytest: cfme/tests/infrastructure/test_vmware_provider.py -k "test_rebuilt_vcenter_duplicate_hosts"}}

### DIFF
--- a/cfme/tests/infrastructure/test_vmware_provider.py
+++ b/cfme/tests/infrastructure/test_vmware_provider.py
@@ -16,6 +16,7 @@ from cfme.utils import conf
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.blockers import BZ
 from cfme.utils.log import logger
+from cfme.utils.wait import wait_for
 
 pytestmark = [
     test_requirements.vmware,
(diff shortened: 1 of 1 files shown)


jawatts:
waiting on travis and PRT

cfme/tests/infrastructure/test_vmware_provider.py
```diff
@@ -502,6 +509,7 @@ def test_rebuilt_vcenter_duplicate_hosts(appliance, provider):
     assert result.success, "SSH Command result was unsuccessful: {}".format(result)
     logger.info('output of rails command: %s', result.output)
     provider.refresh_provider_relationships()
+    wait_for(provider.is_refreshed, func_kwargs={"refresh_delta": 120}, num_sec=300, delay=30)
```
john-dupuy: @kedark3 In my opinion, this should either be built in to `refresh_provider_relationships()` or a function should be available in `cfme.common.provider.py` that allows for this functionality (e.g. `provider.wait_for_provider_refresh(**kwargs)`). What do you think?

john-dupuy: PR LGTM though :+1:

jawatts: @john-dupuy I agree but this would be out of scope for this PR

PR #9059, opened 2019-07-11
[1LP][RFR] Fix template upload for scvmm

john-dupuy:
Fixing template upload for SCVMM provider.

### DIFF
--- a/cfme/utils/template/scvmm.py
+++ b/cfme/utils/template/scvmm.py
@@ -11,7 +11,7 @@ class SCVMMTemplateUpload(ProviderTemplateUpload):
 
     @property
     def library(self):
-        return self.template_upload_data.get('vhds', None)
+        return self.template_upload_data["vhds"]
 
     @property
     def vhd_name(self):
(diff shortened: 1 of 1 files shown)


cfme/utils/template/scvmm.py
```diff
@@ -11,55 +11,69 @@ class SCVMMTemplateUpload(ProviderTemplateUpload):
 
     @property
     def library(self):
-        return self.template_upload_data.get('vhds', None)
+        return self.template_upload_data.get('vhds')
```
digitronik: Q: what if `vhds` tag missing in `data_yaml`?

jaryn: This change doesn't change any behaviour as the `None` is a default value of the second parameter of `get`. So if exception should be thrown on data missing, then @john-dupuy should  use `self.template_upload_data['vhds']`

john-dupuy: @jaryn is correct, I was just removing the redundant specification of `None` here, do you think this should raise an error if that `key` isn't present?

john-dupuy: Changed to raise an error if key "vhds" isn't present. This was done so that we don't continue and try to download the file.

cfme/utils/template/scvmm.py
```diff
@@ -11,55 +11,69 @@ class SCVMMTemplateUpload(ProviderTemplateUpload):
 
     @property
     def library(self):
-        return self.template_upload_data.get('vhds', None)
+        return self.template_upload_data["vhds"]
```
kedark3: **Question** why are we changing this?

john-dupuy: So that it will raise an error if that field is not in the yamls.

kedark3: Got it. so we are expecting that error to be raised.

PR #9060, opened 2019-07-12
[WIPTEST] py3 sprout client fix

izapolsk:
wait_time can have value as string.

### DIFF
--- a/cfme/test_framework/sprout/client.py
+++ b/cfme/test_framework/sprout/client.py
@@ -104,6 +104,11 @@ class SproutClient(object):
             provider_type=None, lease_time=60, ram=None, cpu=None, **kwargs):
         # provisioning may take more time than it is expected in some cases
         wait_time = kwargs.pop('wait_time', 900)
+        try:
+            wait_time = int(wait_time)
+        except ValueError:
+            pass
+
         # If we specify version, stream is ignored because we will get that specific version
         if version:
             stream = get_stream(version)


cfme/test_framework/sprout/client.py
```diff
@@ -104,6 +104,11 @@ def provision_appliances(
             provider_type=None, lease_time=60, ram=None, cpu=None, **kwargs):
         # provisioning may take more time than it is expected in some cases
         wait_time = kwargs.pop('wait_time', 900)
+        try:
+            wait_time = int(wait_time)
+        except ValueError:
+            pass
```
john-dupuy: If the `ValueError` is hit, i.e. if `wait_time="10s"` or something, shouldn't we pass something else to `wait_for`? As `wait_for` will still try to use that string https://github.com/RedHatQE/wait_for/blob/e027318cd4202c5288246e12e591a86e3296115f/wait_for/__init__.py#L52 and then fail at https://github.com/RedHatQE/wait_for/blob/e027318cd4202c5288246e12e591a86e3296115f/wait_for/__init__.py#L176

Maybe would be best to add an update to `wait_for`?

izapolsk: I agree I thought of that but it would be more difficult :)
I'm gonna update wait_for in a bit

izapolsk:
closing this PR in favor of small tweak made in https://github.com/RedHatQE/wait_for/pull/21

PR #9065, opened 2019-07-12
[1LP][RFR] Fixed provision entry point issue

ganeshhubale:
- This PR fixes issue while selecting ```provision entry point``` tree
- Removed 5.11 blocker BZ

{{ pytest: cfme/tests/automate/test_git_import.py -k 'test_automate_git_domain_displayed_in_service' -vv }}

### DIFF
--- a/cfme/tests/automate/test_git_import.py
+++ b/cfme/tests/automate/test_git_import.py
@@ -6,7 +6,6 @@ from cfme import test_requirements
 from cfme.automate.explorer.domain import DomainDetailsView
 from cfme.base.credential import Credential
 from cfme.utils.appliance.implementations.ui import navigate_to
-from cfme.utils.blockers import BZ
 from cfme.utils.rest import assert_response
 
 
@@ -45,7 +44,6 @@ def new_user(appliance):
     user.delete_if_exists()
 
 
-@pytest.mark.meta(blockers=[BZ(1714493)])
 @pytest.mark.tier(1)
 def test_automate_git_domain_removed_from_disk(appliance, imported_domain):
     """
@@ -62,9 +60,8 @@ def test_automate_git_domain_removed_from_disk(appliance, imported_domain):
         '[ ! -d "/var/www/vmdb/data/git_repos{}" ]'.format(repo_path)).success
 
 
-@pytest.mark.meta(blockers=[BZ(1714493)])
 @pytest.mark.tier(2)
-def test_automate_git_domain_displayed_in_service(appliance, imported_domain):
+def test_automate_git_domain_displayed_in_service(appliance):
     """Tests if a domain is displayed in a service.
        Checks if the domain imported from git is displayed and usable in the pop-up tree in the
        dialog for creating services.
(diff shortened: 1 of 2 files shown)


widgetastic_manageiq/__init__.py
```diff
@@ -5744,7 +5744,9 @@ def fill(self, value):
 
         # Some EntryPoint need to apply tree selection
         if self.tree.is_displayed:
-            self.apply.click()
+            if include_domain:
+                self.parent.modal.include_domain.click()
```
digitronik: hummm.. `self.parent` means you are using `View` which holding `EntryPoint` right?
I think we need to include that checkbox as per of this widget.

widgetastic_manageiq/__init__.py
```diff
@@ -5744,7 +5744,9 @@ def fill(self, value):
...
+        self.apply.click()
```
digitronik: it should be under `tree.is_displayed` condition.

cfme/tests/automate/test_git_import.py
```diff
@@ -48,7 +47,6 @@ def new_user(appliance):
     user.delete_if_exists()
 
 
-@pytest.mark.meta(blockers=[BZ(1714493)])
```
mshriver: Please turn this into an `automates` marker instead of removing it.

cfme/tests/automate/test_git_import.py
```diff
@@ -65,9 +63,8 @@ def test_automate_git_domain_removed_from_disk(appliance, imported_domain):
         '[ ! -d "/var/www/vmdb/data/git_repos{}" ]'.format(repo_path)).success
 
 
-@pytest.mark.meta(blockers=[BZ(1714493)])
```
mshriver: same as above, change to `automates`

cfme/tests/automate/test_git_import.py
```diff
@@ -79,26 +76,26 @@ def test_automate_git_domain_displayed_in_service(appliance, imported_domain):
...
+        "GenericLifecycle",
+        "provision"
     )
-    view.modal.include_domain.fill(True)
-    view.modal.apply.click()
-    assert view.provisioning_entry_point.value == ("/{}/Service/Provisioning/StateMachines/"
-        "ServiceProvision_Template/CatalogItemInitialization".format(imported_domain.name))
+    view.provisioning_entry_point.fill(path, include_domain=True)
+    assert view.provisioning_entry_point.value == (
+        "/CloudForms_Essentials/Service/Generic/StateMachines/GenericLifecycle/provision")
```
mshriver: Looks like you should still be formatting `imported_domain.name` into this expected string, instead of hard-coding `'CloudForms_Essentials'` into this string.

mshriver: Consider picking items from `path` and using `'/'.join([imported_domain.name, path[2], path[3], path[4], path[5], path[6]])` - but note this won't produce a string with a leading `'/'`

ganeshhubale: Can we go with this @mshriver ?
```
assert view.provisioning_entry_point.value == (
        "/{}".format("/".join([imported_domain.name, path[2], path[3], path[4], path[5], path[6]]))
    )
```

cfme/tests/automate/test_git_import.py
```diff
@@ -79,27 +79,29 @@ def test_automate_git_domain_displayed_in_service(appliance, imported_domain):
         initialEstimate: 1/20h
         tags: automate
     """
+    url = "https://github.com/ramrexx/CloudForms_Essentials.git"
```
kedark3: **REQUIRED** : please move this to Yamls. Link to yamls shared in gchat.

ganeshhubale: @kedark3 
I am agree with you but this URL or git repo may vary if I want to change or update this test case
So for now I am forking this repo on my GitHub account and adding url of forked repo in Yamls.
PR: https://gitlab.cee.redhat.com/cfme-qe/cfme-qe-yamls/merge_requests/811

kedark3: I have merged that PR.

cfme/tests/automate/test_git_import.py
```diff
@@ -79,27 +80,29 @@ def test_automate_git_domain_displayed_in_service(appliance, imported_domain):
...
+        "Generic",
         "StateMachines",
-        "ServiceProvision_Template",
-        "CatalogItemInitialization"
+        "GenericLifecycle",
+        "provision"
+    )
+    view.provisioning_entry_point.fill(path, include_domain=True)
+    assert view.provisioning_entry_point.value == (
+        "/{}".format("/".join([imported_domain.name, path[2], path[3], path[4], path[5], path[6]]))
```
izapolsk: looks a bit not accurate. I guess the following would look better:
`[some_value, *path[1:]]`

izapolsk: and I would give a shot to list/tuple comparison instead. I mean you can split entry point instead of joining these values

widgetastic_manageiq/__init__.py
```diff
@@ -5730,12 +5732,12 @@ def value(self):
     def read(self):
         return self.value
 
-    def fill(self, value):
+    def fill(self, value, include_domain=False):
```
izapolsk: could you explain what it is for ?

cfme/tests/automate/test_git_import.py
ganeshhubale: - We need to include domain while selecting entry point in catalog item.
- This is useful if we have same tree but different domains available. So in this case we need to select domain prefix.


screen_cfme

widgetastic_manageiq/__init__.py
ganeshhubale: - We need to include domain while selecting entry point in catalog item.
- This is useful if we have same tree but different domains available. So in this case we need to select domain prefix.


screen_cfme

PR #9068, opened 2019-07-15
[1LP][RFR] block test with BZ-1729594

digitronik:

{{pytest: cfme/tests/automate/custom_button/test_container_objects.py::test_custom_button_dialog_container_obj --use-provider ocp-39-hawk  -v}}

### DIFF
--- a/cfme/tests/automate/custom_button/test_service_vm_custom_button.py
+++ b/cfme/tests/automate/custom_button/test_service_vm_custom_button.py
@@ -12,6 +12,7 @@ from cfme.utils.appliance import ViaSSUI
 from cfme.utils.appliance import ViaUI
 from cfme.utils.appliance.implementations.ssui import navigate_to as ssui_nav
 from cfme.utils.appliance.implementations.ui import navigate_to as ui_nav
+from cfme.utils.blockers import BZ
 from cfme.utils.log_validator import LogValidator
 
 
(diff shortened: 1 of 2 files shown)


cfme/tests/automate/custom_button/test_service_vm_custom_button.py
```diff
@@ -136,7 +137,8 @@ def test_custom_button_display_service_vm(request, appliance, service_vm, button
 
 @test_requirements.customer_stories
 @pytest.mark.tier(1)
-@pytest.mark.meta(automates=[1687061])
+# Dynamic dialog problem 1729594
+@pytest.mark.meta(automates=[1687061], blockers=[BZ(1729594), BZ(1729594)])
```
sshveta: Why do we have same BZ twice ?

mshriver: One blocks tests, one tracks automation for BZ test coverage.

When the test is no longer blocked by this BZ, blockers can be removed, automates continues to track that this test function/module/etc covers that BZ.

sshveta: @mshriver : Shouldn't it be this then ?

     @pytest.mark.meta(automates=[1687061, 1729594], blockers=[BZ(1729594)])

digitronik: @sshveta its was my bad its should `BZ(1729046), BZ(1729594)` :+1: 

about `automate` or `coverage` marker. It should not cover all BZ only `steps` not `setup` or `teardown`

about blocker; It should block as per version (5.10 and 5.11 but i missed; you can take it as `Human error`). Other thing about blocker, As reviewer my self feel bad about this; when, I asked someone to remove `force_stream` like thing as its depends on release version which depends on `triage`. contributors added this after adding BZ and we asked to remove it as per current status.

PR #9079, opened 2019-07-16
[1LP][RFR] RBAC tests refactor

apagac:
RBAC tests __refactor__.

Removed global `pytest.mark.provider` and `setup_provider` fixture, since most of the tests do not need a provider to be present.
Moved these markers to a couple of tests that do need to use provider.

Refactored usage of `group_collection` fixture. It was suggested we don't need to use a fixture to return a collection when we can reference that collection directly from `appliance`.

{{ pytest: -v --long-running cfme/tests/configure/test_access_control.py }}

### DIFF
--- a/cfme/tests/configure/test_access_control.py
+++ b/cfme/tests/configure/test_access_control.py
@@ -11,7 +11,7 @@ from cfme.configure.access_control import AddUserView
 from cfme.configure.tasks import TasksView
 from cfme.containers.provider.openshift import OpenshiftProvider
 from cfme.exceptions import RBACOperationBlocked
-from cfme.infrastructure.provider.virtualcenter import VMwareProvider
+from cfme.infrastructure.provider import InfraProvider
 from cfme.markers.env_markers.provider import ONE
 from cfme.services.myservice import MyService
 from cfme.utils.appliance.implementations.ui import navigate_to
@@ -20,15 +20,8 @@ from cfme.utils.log import logger
 from cfme.utils.update import update
 
 pytestmark = [
-    test_requirements.rbac,
-    pytest.mark.provider(classes=[VMwareProvider], selector=ONE),
-    pytest.mark.usefixtures('setup_provider')
+    test_requirements.rbac
 ]
-
-
-@pytest.fixture(scope='module')
-def group_collection(appliance):
-    return appliance.collections.groups
 
 
 def new_credential():
@@ -98,7 +91,7 @@ def catalog_obj(appliance):
 # User test cases
 @pytest.mark.sauce
 @pytest.mark.tier(2)
-def test_user_crud(appliance, group_collection):
+def test_user_crud(appliance):
     """
     Polarion:
         assignee: apagac
@@ -107,6 +100,7 @@ def test_user_crud(appliance, group_collection):
         tags: rbac
     """
     group_name = 'EvmGroup-user'
+    group_collection = appliance.collections.groups
     group = group_collection.instantiate(description=group_name)
 
     user = new_user(appliance, [group])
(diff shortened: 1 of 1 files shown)


apagac:
@mshriver can you please take a look :)

dajoRH:
Lint report for commit 64e7999203556896ad9e36f8cd14e5e0c0baf7f1:

- :red_circle: Line 505:24: **F821** *undefined name 'appliance'*

**No commit flaws detected**.

*CFME QE Bot*

apagac:
PRT result commentary: 5.11 has couple of problems that are unrelated to this PR; 5.10 has only one test failing and that needs to be fixed but is also unrelated to this PR.

cfme/tests/configure/test_access_control.py
```diff
@@ -506,7 +513,9 @@ def test_group_crud(group_collection):
 @pytest.mark.sauce
 @pytest.mark.tier(2)
 @test_requirements.tag
-def test_group_crud_with_tag(provider, tag_value, group_collection):
+@pytest.mark.provider(classes=[VMwareProvider], selector=ONE)
+@pytest.mark.usefixtures('setup_provider')
```
digitronik: In such cases I would inherit fixture directly like
 ```
def test_group_crud_with_tag(appliance, provider, setup_provider,  tag_value, ):
```

File: cfme/tests/configure/test_access_control.py [hunk 2]
```diff
@@ -1089,6 +1111,8 @@ def test_permissions_role_crud(appliance):
 
 
 @pytest.mark.tier(3)
+@pytest.mark.provider(classes=[VMwareProvider], selector=ONE)
+@pytest.mark.usefixtures('setup_provider')
```
digitronik: ^^

cfme/tests/configure/test_access_control.py
```diff
@@ -1976,6 +2004,7 @@ def test_tags_manual_features():
 
 @pytest.mark.manual
 @pytest.mark.provider([OpenshiftProvider], override=True)
```
digitronik: I think, no need to `override` provider now

cfme/tests/configure/test_access_control.py
apagac: Will do.

apagac:
Moved directly to RFR since no substantial changes were performed and I run tests with `setup_provider` locally.

cfme/tests/configure/test_access_control.py
```diff
@@ -525,6 +533,7 @@ def test_group_crud_with_tag(provider, tag_value, group_collection):
 
     path = 'VM_Template-Folder' if provider.key == 'vsphere55' else 'Discovered virtual machine'
```
kedark3: Suggestion: @apagac I know this is out of scope for this PR, but I think we can remove the `vsphere55` check as the provider is deprecated. What do you think?

cfme/tests/configure/test_access_control.py
```diff
@@ -1089,7 +1110,8 @@ def test_permissions_role_crud(appliance):
 
 
 @pytest.mark.tier(3)
-def test_permissions_vm_provisioning(appliance, provider):
+@pytest.mark.provider(classes=[VMwareProvider], selector=ONE)
```
kedark3: Do you need this provider specifically or any `InfraProvider` is ok? 
You may be able to use `pytest.mark.provider(classes=[InfraProvider], selector=ONE)`

cfme/tests/configure/test_access_control.py
jawatts: At least add a TODO here for now please

jawatts:
Outstanding questions and 1 TODO, but otherwise LGTM. Moving to WIPTEST for now.

cfme/tests/configure/test_access_control.py
apagac: Any Infra Provider should be enough for permissions testing or tag testing, there's no VMWare-specific stuff.

apagac:
Comments addressed, moving to RFR.

PR #9081, opened 2019-07-16
[1LP][RFR] adding snippets for quick multi-region env setup

izapolsk:


cfme/scripting/setup_env.py
```diff
@@ -206,5 +210,93 @@ def setup_replication_env(cfme_version, provider_type, provider, lease, sprout_p
...
+                for app in sprout_pool['appliances']:
+                    apps.append(IPAppliance(app['ip_address']))
+                sprout_client.set_pool_description(sprout_poolid, desc)
+            else:
+                sys.exit("Pool does not meet the minimum size requirements!")
+        else:
+            sys.exit("Pool not found!")
+
+    else:
+        print("Provisioning appliances")
```
john-dupuy: I got a comment about using `click.echo` rather than `print` on one of my PR's a while back. Some info on the difference is here: https://stackoverflow.com/questions/49906680/difference-between-print-and-click-echo-in-python-3. @digitronik is it worth changing `print` to `click.echo` here?

izapolsk: this is good improvement but it is out of scope. I follow existing code style. 
if we're going to use `click.echo` instead of `print`, we will have to update all our scripting stuff.

john-dupuy: Agreed, if we want to move to `click.echo`, we should do it in its own PR.

PR #9082, opened 2019-07-16
[1LP][RFR] Add BZ marker and update test_server_name patterns for filling

mshriver:
{{pytest: cfme/tests/configure/test_server_name.py }}

PRT RESULTS:
5.11 failure is addressed in wt.patternfly PR #98 and #9084  (which will include update to wt.patternfly)

### DIFF
--- a/cfme/tests/configure/test_server_name.py
+++ b/cfme/tests/configure/test_server_name.py
@@ -31,7 +31,7 @@ def test_server_name(request, appliance):
         appliance.rename(old_server_name)
 
     new_server_name = "RENAME-TEST"
-    assert view.server.basic_information.appliance_name.fill('updated-evm-name')
+    assert view.server.basic_information.appliance_name.fill(new_server_name)
     assert view.server.save.is_enabled
     view.server.save.click()  # no boolean return
     assert appliance.server.name == new_server_name
--- a/cfme/tests/configure/test_server_name.py
+++ b/cfme/tests/configure/test_server_name.py
@@ -9,7 +9,8 @@ from cfme.utils.appliance.implementations.ui import navigate_to
 pytestmark = [test_requirements.general_ui]
 
 
-@pytest.mark.tier(3)
+@pytest.mark.tier(1)
+@pytest.mark.meta(automates=[1635178])
 @pytest.mark.sauce
 def test_server_name(request, appliance):
     """Tests that changing the server name updates the about page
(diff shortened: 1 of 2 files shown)


cfme/tests/configure/test_server_name.py
```diff
@@ -23,27 +24,29 @@ def test_server_name(request, appliance):
...
-    appliance.server.settings.update_basic_information({'appliance_name': new_server_name})
-    flash_message = (
-        'Configuration settings saved for {} Server "{} [{}]" in Zone "{}"'.format(
-            appliance.product_name,
-            appliance.server.name,
-            appliance.server.sid,
-            appliance.server.zone.name))
-
-    view.flash.assert_message(flash_message)
+        appliance.rename(old_server_name)
```
john-dupuy: @mshriver It appears `rename` is a method of an `Appliance` instance but not an `IPAppliance` instance. https://github.com/ManageIQ/integration_tests/blob/976184c0df2ecee9e1baac5132facac5dd8552ef/cfme/utils/appliance/__init__.py#L2745

Do we want this to be method on `IPAppliance`?

mshriver: Agreed, moved.

cfme/tests/configure/test_server_name.py
```diff
@@ -23,27 +24,29 @@ def test_server_name(request, appliance):
...
+
+    new_server_name = "RENAME-TEST"
+    assert view.server.basic_information.appliance_name.fill(new_server_name)
```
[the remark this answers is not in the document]
mshriver: Tighter assertion of the actions taken in the test is pretty much always a positive thing in my book.

Here I validate that the form field was actually modified by simply asserting the fill() call.
Next I validate that the save button is enabled after the form is modified by asserting save.is_enabled.
Then I assert that the `appliance.server.name`, which is fetched from REST interface, matches the update I made.
Then I assert the UI flash message
Then I assert the about modal reflects the updated name.

cfme/tests/configure/test_server_name.py
```diff
@@ -23,27 +24,29 @@ def test_server_name(request, appliance):
...
+    assert view.server.save.is_enabled
+    view.server.save.click()  # no boolean return
+    assert appliance.server.name == new_server_name
+
+    view.flash.assert_success_message(
+        'Configuration settings saved for {} Server "{} [{}]" in Zone "{}"'
```
valaparthvi: Optional: Can we not make format this using keys?

mshriver: I would gladly merge a PR that changes this :) (already merged by the time I saw this)

PR #9083, opened 2019-07-16
[1LP][RFR] New container crud test and remove flash assert in BaseProvider

jawatts:
Removing flash message assertions for setup and delete in `BaseProvider`.
Add new containers provider crud test

{{ pytest: cfme/tests/containers/test_container_provider_crud.py --use-provider ocp-39-hawk }}

### DIFF
--- a/cfme/tests/containers/test_container_provider_crud.py
+++ b/cfme/tests/containers/test_container_provider_crud.py
@@ -1,0 +1,66 @@
+import fauxfactory
+import pytest
+
+from cfme import test_requirements
+from cfme.common.provider_views import ContainerProvidersView
+from cfme.containers.provider import ContainersProvider
+from cfme.markers.env_markers.provider import ONE_PER_VERSION
+from cfme.utils.blockers import BZ
+from cfme.utils.update import update
+
+
+pytestmark = [
+    pytest.mark.tier(1),
+    test_requirements.containers,
+    pytest.mark.provider(
+        [ContainersProvider],
+        scope='function',
+        selector=ONE_PER_VERSION),
+    pytest.mark.meta(
+        blockers=[BZ(1713504, forced_streams=['5.11'])]
+    ),
+]
+
+
+def test_container_provider_crud(request, appliance, has_no_providers, provider):
+
+    """
+    Polarion:
+        assignee: juwatts
+        caseimportance: critical
+        casecomponent: Containers
+        initialEstimate: 1/6h
+    """
+
+    provider.create()
+
+    request.addfinalizer(lambda: provider.delete_if_exists(cancel=False))
+
+    view = appliance.browser.create_view(ContainerProvidersView)
+
+    view.flash.assert_success_message('{} Providers "{}" was saved'.format(provider.string_name,
+                                                                           provider.name))
+
+    assert provider.exists
+
+    with update(provider):
+        provider.name = fauxfactory.gen_alpha(8).lower()
+
+    assert view.is_displayed
+
+    view.flash.assert_success_message(
+        'Containers Provider "{}" was saved'.format(provider.name))
+
+    assert provider.name == str(view.entities.get_first_entity().data.get('name', {}))
+
+    provider.delete(cancel=False)
+
+    assert view.is_displayed
+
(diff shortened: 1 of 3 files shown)


cfme/tests/containers/test_container_provider_crud.py
```diff
@@ -0,0 +1,71 @@
...
+        casecomponent: Containers
+        initialEstimate: 1/6h
+    """
+
+    if provider.exists:
+        provider.delete()
+
+    setup_or_skip(request, provider)
+
+    request.addfinalizer(lambda: provider.delete_if_exists(cancel=False))
```
mshriver: IIRC delete_if_exists default behavior is cancel=False, so you don't need a lambda here and can just pass the function.

cfme/tests/containers/test_container_provider_crud.py
```diff
@@ -0,0 +1,71 @@
...
+
+    """
+    Polarion:
+        assignee: juwatts
+        caseimportance: medium
+        casecomponent: Containers
+        initialEstimate: 1/6h
+    """
+
+    if provider.exists:
```
mshriver: I think instead of this pattern, you can use the `has_no_providers` fixture to clear them out before the test starts.

cfme/tests/containers/test_container_provider_crud.py
```diff
@@ -0,0 +1,71 @@
...
+]
+
+
+@pytest.mark.tier(3)
+def test_container_provider_crud(request, appliance, provider):
+
+    """
+    Polarion:
+        assignee: juwatts
+        caseimportance: medium
```
mshriver: I'd call it critical

cfme/tests/containers/test_container_provider_crud.py
```diff
@@ -0,0 +1,71 @@
...
+
+    view = appliance.browser.create_view(ContainerProvidersView)
+
+    success_add = '{} Providers "{}" was saved'.format(provider.string_name, provider.name)
+
+    view.flash.assert_success_message(success_add)
+
+    assert provider.exists
+
+    provider_edit_view = navigate_to(provider, 'Edit')
```
mshriver: Providers are `Updateable`, is there a reason you're explicitly filling the edit view here?

Its forcing you to use `new_provider_name` in the flash assert following, and then you're setting the name attribute outside of the `Updateable` context manager (L56).

cfme/tests/containers/test_container_provider_crud.py
```diff
@@ -0,0 +1,71 @@
...
+    if provider.exists:
+        provider.delete()
+
+    setup_or_skip(request, provider)
+
+    request.addfinalizer(lambda: provider.delete_if_exists(cancel=False))
+
+    view = appliance.browser.create_view(ContainerProvidersView)
+
+    success_add = '{} Providers "{}" was saved'.format(provider.string_name, provider.name)
```
mshriver: Consider not naming these strings (success_delete below too) since they're just used once in the flash message assert call.

cfme/tests/containers/test_container_provider_crud.py
jawatts: `delete_if_exists` passes *args and **kwargs to `delete`. Default behavior of delete is `cancel=True`. I think we need to change that to `False`. I will make and enchancement

cfme/tests/containers/test_container_provider_crud.py
jawatts: updated to use the `Updateable` context manager.

cfme/tests/containers/test_container_provider_crud.py
```diff
@@ -0,0 +1,67 @@
...
+
+    """
+    Polarion:
+        assignee: juwatts
+        caseimportance: critical
+        casecomponent: Containers
+        initialEstimate: 1/6h
+    """
+
+    setup_or_skip(request, provider)
```
mshriver: I don't think you want setup_or_skip, if it fails to add the provider, that's a test failure, not skip.

I would call provider.create() directly so that it gets added through the UI.

PR #9084, opened 2019-07-17
[1LP][RFR] Update NavDropdown use, parametrized locators, SSUI locator changes

mshriver:
~~NEEDS: https://github.com/RedHatQE/widgetastic.patternfly/pull/100~~

~~NEEDS: #9096 for passing 5.11 PRT~~

Update to use new NavDropdown pattern.

Updates/fixes to SSUI widget locators
Updates/fixes to expandable checks in latest MIQ.

{{ pytest: cfme/tests/configure/test_server_name.py cfme/tests/ssui/test_ssui_dashboard.py }}

### DIFF
--- a/cfme/base/login.py
+++ b/cfme/base/login.py
@@ -15,9 +15,10 @@ class BaseLoggedInPage(View):
     """
     CSRF_TOKEN = '//meta[@name="csrf-token"]'
     flash = FlashMessages('.//div[@id="flash_msg_div"]')
-    help = NavDropdown('.//li[./a[@id="dropdownMenu1"]]|.//li[./a[@id="help-menu"]]')
-    configuration_settings = Text('.//li[./a[@title="Configuration"]]')      # 5.11+
-    settings = SettingsNavDropdown('.//li[./a[@id="dropdownMenu2"]]')
+    # TODO don't use `help` here, its a built-in
+    help = NavDropdown(id="help-menu")
+    configuration_settings = Text('//li[.//a[@title="Configuration"]]')  # 5.11+
+    settings = SettingsNavDropdown(id="dropdownMenu2")
     navigation = VerticalNavigation('#maintab')
 
     @property
@@ -51,6 +52,7 @@ class BaseLoggedInPage(View):
     def logged_in_as_current_user(self):
         return self.logged_in_as_user(self.extra.appliance.user)
 
+    # TODO remove this property, it is erroneous. View properties should be returning data from UI
     @property
     def current_username(self):
         try:
--- a/cfme/base/ssui.py
+++ b/cfme/base/ssui.py
@@ -2,14 +2,13 @@ import time
 
 from navmazing import NavigateToSibling
 from selenium.webdriver.common.keys import Keys
-from widgetastic.utils import ParametrizedLocator
-from widgetastic.widget import ParametrizedView
 from widgetastic.widget import View
 from widgetastic_patternfly import Button
 from widgetastic_patternfly import FlashMessages
 from widgetastic_patternfly import Input
 from widgetastic_patternfly import NavDropdown
 from widgetastic_patternfly import Text
+from widgetastic_patternfly.utils import PFIcon
 
 from . import Server
 from cfme.base.credential import Credential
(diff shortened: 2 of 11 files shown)


requirements/frozen_docs.py3.txt
```diff
@@ -6,6 +6,7 @@
 adal==1.2.1
 aenum==2.1.2
 alabaster==0.7.12
+allure-pytest
```
mshriver: This is only added to support wt.pf git install for PRT, will be removed for review/merge.

cfme/base/login.py
```diff
@@ -15,9 +15,9 @@ class BaseLoggedInPage(View):
     """
     CSRF_TOKEN = '//meta[@name="csrf-token"]'
     flash = FlashMessages('.//div[@id="flash_msg_div"]')
-    help = NavDropdown('.//li[./a[@id="dropdownMenu1"]]|.//li[./a[@id="help-menu"]]')
-    configuration_settings = Text('.//li[./a[@title="Configuration"]]')      # 5.11+
-    settings = SettingsNavDropdown('.//li[./a[@id="dropdownMenu2"]]')
+    help = NavDropdown(id="help-menu")
+    configuration_settings = Text('//li[.//a[@title="Configuration"]]')  # 5.11+
```
kedark3: I see this is not new but what does it mean by `#5.11+` ? What happens on other stream?

cfme/base/ssui.py
```diff
@@ -25,29 +24,28 @@
...
+    ROOT = '//nav//li[@ng-if="vm.permissions.helpMenu"]'
+
+
 class SSUIBaseLoggedInPage(View):
     """This page should be subclassed by any page that models any other page that is available as
     logged in.
     """
     flash = FlashMessages('.//div[@id="flash_msg_div"]')
-    help = NavDropdown('.//li[./a[@id="dropdownMenu1"]]')
+    help = SSUIHelpNavDropdown()
```
kedark3: isn't `help` a keyword in python?

cfme/tests/configure/test_server_name.py
```diff
@@ -45,7 +46,8 @@ def _ensure_name_reset():
     )
 
     # CFME updates about box only after any navigation BZ(1408681) - closed wontfix
-    navigate_to(appliance.server, 'Dashboard')
+    view = navigate_to(appliance.server, 'Dashboard')
+    view.browser.refresh()
```
kedark3: Why do we need this refresh?

cfme/base/login.py
mshriver: This widget doesn't exist in 5.10, configuration is accessed through the settings widget.

cfme/base/ssui.py
mshriver: sure is, I was thinking it was just in ipython but its a built-in.  The way its used here I'm not too concerned, as it will never be called directly as a function.

 Okay leaving the changing of this to a non-keyword widget name in another PR? gotta audit for uses of it and change both UI and SSUI

cfme/tests/configure/test_server_name.py
mshriver: BZ 1408681, navigation alone is not consistent enough.

cfme/base/ssui.py
kedark3: Please add a TODO :)

mshriver: https://github.com/ManageIQ/integration_tests/commit/203bc8ab79a4fbf9f7ea042fce346fa774fb6049#diff-7c97dce5eefb958f90c137c66bd8f857R18

cfme/base/ssui.py
```diff
@@ -25,29 +24,29 @@
 from widgetastic_manageiq import SSUIVerticalNavigation
 
 
+class SSUISettingsNavDropdown(NavDropdown):
+    """iuser pficon-user to differentiate from help menu"""
+    ROOT = ('//nav'
+            '//li[.//a[contains(@class, "dropdown-toggle") and '
+            './/i[contains(@class, "{}")]] '
+            'and contains(@class, "dropdown")]'.format(PFIcon.icons.USER))
```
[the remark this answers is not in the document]
mshriver: Yep! This was certainly a bit of an odd pattern - though I understand why it was used, it was by far the most unique locator for the widget. But the element classes are accurate enough.

PR #9086, opened 2019-07-17
[1LP][RFR] Added entry for infra_hosts and assigned me "prichard" to it.

prichard77:
The purpose of this PR is to add an entry for the FA infra_hosts that is assigned to me. I'll be updating files in integration tests and will need to reference the infra_hosts FA.

### DIFF
--- a/cfme/test_requirements.py
+++ b/cfme/test_requirements.py
@@ -230,7 +230,7 @@ ipv6 = pytest.mark.requirement(
 
 infra_hosts = pytest.mark.requirement(
     "Infrastructure Hosts Support",
-    description='Support functionality for cfme hosts',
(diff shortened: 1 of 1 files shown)


prichard77:
I added an entry for infra_hosts in test_requirements.py

cfme/test_requirements.py
```diff
@@ -228,6 +228,12 @@ def test_quota_alert():
     assignee_id='',
 )
 
+infra_hosts = pytest.mark.requirement(
+    "Infrastructure Hosts Support",
+    description='Support functionality for cfme hosts',
```
john-dupuy: I would change `cfme` to `infrastructure`,

PR #9089, opened 2019-07-18
[1LP][RFR]Automate test: test_sent_text_custom_report_with_long_condition

valaparthvi:
1. Adding test_send_text_custom_report_with_long_condition to automate BZ 1677839.
2. Adding a fixture to create reports for `test_send_text_custom_report_with_long_condition` and `test_report_edit_secondary_display_filter`.

{{ pytest: cfme/tests/intelligence/reports/test_reports.py -k "test_send_text_custom_report_with_long_condition or  test_report_edit_secondary_display_filter" --use-template-cache -sqvvvv }}

### DIFF
--- a/cfme/tests/intelligence/reports/test_reports.py
+++ b/cfme/tests/intelligence/reports/test_reports.py
@@ -13,6 +13,63 @@ from cfme.utils.wait import wait_for
 pytestmark = [test_requirements.report, pytest.mark.tier(3), pytest.mark.sauce]
 
 
+FILTER_DATA = {
+    "title": "Testing report",
+    "menu_name": "testing report",
+    "base_report_on": "VMs and Instances",
+    "report_fields": [
+        "Active",
+        "EVM Custom Attributes : Name",
+        "EVM Custom Attributes : Region Description",
+        "EVM Custom Attributes : Region Number",
+        "Name",
+    ],
+    "consolidation": {
+        "group_records": [
+            "EVM Custom Attributes : Name",
+            "EVM Custom Attributes : Region Description",
+            "EVM Custom Attributes : Region Number",
+        ]
+    },
+    "filter": {
+        "primary_filter": "fill_field(VM and Instance : Active, IS NOT NULL)",
+        "secondary_filter": "fill_field(EVM Custom Attributes : Name, INCLUDES, A)",
+    },
+}
+
+LONG_CONDITION_DATA = {
+    "title": "Testing report",
+    "menu_name": "testing report",
+    "base_report_on": "VMs and Instances",
+    "report_fields": ["Name"],
+    "filter": {
+        "primary_filter": (
+            "fill_field({based_on} : Power State, = , on);"
+            "select_first_expression;click_or;fill_field("
+            "{based_on} : Datastore Path, INCLUDES, i);"
+            "select_first_expression;click_or;fill_field("
+            "{based_on}.Provider : Hostname, INCLUDES, env);"
+            "select_first_expression;click_or;fill_field("
+            "{based_on}.Provider : IP Address, INCLUDES, 1);"
+            "select_first_expression;click_or;fill_field("
+            "{based_on}.Provider : IP Address, INCLUDES, 2);"
+            "select_first_expression;click_or;fill_field("
+            "{based_on}.Provider : IP Address, INCLUDES, 4);"
+        ).format(based_on="VM and Instance")
+    },
+}
+
+
+@pytest.fixture
+def generate_report(appliance, request):
+    def _report(data):
+        report = appliance.collections.reports.create(**data)
+        request.addfinalizer(report.delete_if_exists)
+        return report
+
+    return _report
+
+
 @pytest.fixture
 def create_custom_tag(appliance):
     # cannot create a category with uppercase in the name
@@ -126,6 +183,7 @@ def test_reports_custom_tags(appliance, request, create_custom_tag):
             ["Name", "My Company Tags : Owner", "My Company Tags : Cost Center"],
         ),
     ],
+    ids=["floating_ips", "cloud_tenants"],
 )
 @pytest.mark.meta(automates=[1546927, 1504155])
 def test_new_report_fields(appliance, based_on, request):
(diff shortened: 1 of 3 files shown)


cfme/tests/intelligence/reports/test_reports.py
```diff
@@ -239,3 +274,55 @@ def test_report_edit_secondary_display_filter(appliance, filter_report, soft_ass
...
+    # assert that the mail was sent
+    wait_for(
+        lambda: len(smtp_test.get_emails(to_address=data["email"]["to_emails"])) == 1,
+        timeout=200,
+        delay=5,
+        msg="Mail was not sent. Some error occured.",
+    )
+
+    # assert that the pattern was not found in the logs
+    soft_assert(not log.validate(), "Found error message in the logs.")
```
john-dupuy: @valaparthvi You can use the `failure_pattern` keyword if you're trying to assert that a pattern **shouldn't** be found in the logs. https://github.com/ManageIQ/integration_tests/blob/861426751dd237366a91410fa0b76170d0351eb2/cfme/utils/log_validator.py#L57

cfme/tests/intelligence/reports/test_reports.py
```diff
@@ -239,3 +274,55 @@ def test_report_edit_secondary_display_filter(appliance, filter_report, soft_ass
...
+    # prepare LogValidator
+    pattern = ".*negative argument.*"
+    log = LogValidator("/var/www/miq/vmdb/log/evm.log", matched_patterns=[pattern])
+
+    log.start_monitoring()
+    schedule.queue()
+
+    # assert that the mail was sent
+    wait_for(
+        lambda: len(smtp_test.get_emails(to_address=data["email"]["to_emails"])) == 1,
```
john-dupuy: Outside of the scope of this PR but it'd be nice to have a `wait_for` built into this function. Or an extra utility of `SMTPCollectorClient` that will wait_for emails to arrive.

cfme/tests/intelligence/reports/test_reports.py
valaparthvi: Oh, that is how `failure_pattern` works! I had used it earlier, it wasn't working as I'd hoped, or at least that's what I thought. Will change that, thanks John! :)

cfme/tests/intelligence/reports/test_reports.py
valaparthvi: Hmm, that's a good idea! Let me see if I can implement that.

valaparthvi:
Hey @john-dupuy, so I made some major changes, instead of creating the report, I now download reports from the server(where I've uploaded them) and then import those reports. Importing the reports saves me a good amount of time. I have added `wait_for_emails` method to `SMTPCollectorClient`.
Let me know what you think.

cfme/tests/intelligence/reports/test_reports.py
```diff
@@ -1,14 +1,44 @@
...
 
 
+@pytest.fixture
+def get_report(appliance, request):
+    def _report(file_name, menu_name):
+        collection = appliance.collections.reports
+        file_name = "{}.yaml".format(file_name)
+
+        # download the report from server
+        fs = FTPClientWrapper(cfme_data.ftpserver.entities.reports)
```
john-dupuy: :tada: I like this pattern of downloading the report from the server rather than having the lengthy data yamls bloat our LOC in the codebase!

cfme/tests/intelligence/reports/test_reports.py
```diff
@@ -201,7 +204,8 @@ def test_report_edit_secondary_display_filter(appliance, filter_report, soft_ass
     Bugzilla:
         1565171
     """
-    filter_report.update(
+    report = get_report("filter_report", "test_filter_report")
```
john-dupuy: Q: why do you want the report to be instantiated during the test? I.e. why have you implemented the wrapper in `get_report`?

john-dupuy: Ah, now I see, because of the `menu_name` and `filter_name`

cfme/utils/smtp_collector_client.py
```diff
@@ -65,3 +66,12 @@ def get_emails(self, **filter):
 
     def get_html_report(self):
         return self._query(requests.get, "messages.html").text.strip()
+
+    def wait_for_emails(self, wait=60, **filter):
```
john-dupuy: :tada

cfme/tests/intelligence/reports/test_reports.py
```diff
@@ -1,14 +1,44 @@
...
 
+@pytest.fixture
+def get_report(appliance, request):
+    def _report(file_name, menu_name):
+        collection = appliance.collections.reports
+        file_name = "{}.yaml".format(file_name)
+
+        # download the report from server
+        fs = FTPClientWrapper(cfme_data.ftpserver.entities.reports)
+        file_path = fs.download(file_name, os.path.join("/tmp", file_name))
```
digitronik: optional: no need to specify location if you are downloading file in temp. `default` it will download in `temp` with same name as `file_name`

cfme/tests/intelligence/reports/test_reports.py
```diff
@@ -239,3 +243,53 @@ def test_report_edit_secondary_display_filter(appliance, filter_report, soft_ass
...
+    log.start_monitoring()
+    schedule.queue()
+
+    # assert that the mail was sent
+    assert (
+        len(smtp_test.wait_for_emails(wait=200, to_address=data["email"]["to_emails"]))
+        == 1
+    )
+    # assert that the pattern was not found in the logs
+    soft_assert(log.validate(), "Found error message in the logs.")
```
digitronik: Q: why `soft_assert`? I don't think you want to continue even if test fails here.

cfme/utils/smtp_collector_client.py
digitronik: :+1:

cfme/tests/intelligence/reports/test_reports.py
```diff
@@ -1,14 +1,44 @@
...
 from cfme.utils.rest import assert_response
 
 pytestmark = [test_requirements.report, pytest.mark.tier(3), pytest.mark.sauce]
 
 
+@pytest.fixture
+def get_report(appliance, request):
+    def _report(file_name, menu_name):
+        collection = appliance.collections.reports
+        file_name = "{}.yaml".format(file_name)
```
digitronik: It will be great if we wont matriculate name here `.yaml`; why not directly pass it form test? is there any spacial reason?

cfme/tests/intelligence/reports/test_reports.py
```diff
@@ -122,6 +149,7 @@ def test_reports_custom_tags(appliance, request, create_custom_tag):
             ["Name", "My Company Tags : Owner", "My Company Tags : Cost Center"],
         ),
     ],
+    ids=["floating_ips", "cloud_tenants"],
```
mshriver: :heart:

PR #9096, opened 2019-07-20
[1LP][WIPTEST] Fix Cloud Intel navigation

valaparthvi:
Purpose or Intent
=================

__Fixing__ navigation to 'Cloud Intel' because, since 5.11, "Cloud Intel" has been changed to "Overview", by adding a new property called `intel_name` to `Server` that returns appropriate name.

Few tests on 5.11.0.15 are failing because of [BZ 1724209](https://bugzilla.redhat.com/show_bug.cgi?id=1724209), but since they're passing for 5.10, I'm assuming these changes are alright.
{{ pytest: cfme/tests/intelligence/reports/test_crud.py -vvv }}

### DIFF
--- a/cfme/base/__init__.py
+++ b/cfme/base/__init__.py
@@ -132,6 +132,10 @@ class Server(BaseEntity, sentaku.modeling.ElementMixin):
         view.customlogos.save_button.click()
         view.flash.assert_no_error()
 
+    @property
+    def intel_name(self):
+        return "Overview" if self.appliance.version > "5.11" else "Cloud Intel"
+
 
 @attr.s
 class ServerCollection(BaseCollection, sentaku.modeling.ElementMixin):
--- a/cfme/base/__init__.py
+++ b/cfme/base/__init__.py
@@ -10,6 +10,8 @@ from cfme.utils import ParamClassName
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.log import logger
 from cfme.utils.pretty import Pretty
+from cfme.utils.version import Version
+from cfme.utils.version import VersionPicker
 
 
 @attr.s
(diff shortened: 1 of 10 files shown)


cfme/base/ui.py
```diff
@@ -470,7 +472,9 @@ class Chargeback(CFMENavigateStep):
     prerequisite = NavigateToSibling('LoggedIn')
 
     def step(self, *args, **kwargs):
-        self.prerequisite_view.navigation.select('Cloud Intel', 'Chargeback')
+        self.prerequisite_view.navigation.select(
+            self.view.context["object"].intel_name, "Chargeback"
```
mshriver: Within navigation destinations, you can/should reference the registered object via `self.obj`

cfme/base/__init__.py
```diff
@@ -132,6 +132,10 @@ def upload_custom_logo(self, file_type, file_data=None, enable=True):
         view.customlogos.save_button.click()
         view.flash.assert_no_error()
 
+    @property
```
mshriver: Consider using VersionPicker here, and defining it as a class attribute instead of a property method.  Then there isn't an explicit call to self.appliance.version, it'll pick itself when referenced.

```
intel_name = VersionPicker({Version.LATEST: 'Overview', Version.LOWEST: 'Cloud Intel'})
```

cfme/base/__init__.py
```diff
@@ -28,6 +30,7 @@ class Server(BaseEntity, sentaku.modeling.ElementMixin):
     current_username = sentaku.ContextualMethod()
     current_group_name = sentaku.ContextualMethod()
     group_names = sentaku.ContextualMethod()
+    intel_name = VersionPicker({"5.11": "Overview", Version.lowest(): "Cloud Intel"})
```
mshriver: :+1:

mshriver:
PRT results acceptable, `Cloud Intel` vs `Overview` change scoped for this PR is fixed.

PR #9097, opened 2019-07-22
[1LP][RFR]Automate test: test_tls_openssl_verify_mode

valaparthvi:
Purpose or Intent
=================

__Adding tests__ `test_tls_openssl_verify_mode` that covers BZ 1475553.

{{ pytest: cfme/tests/webui/test_general_ui.py::test_tls_openssl_verify_mode -vvv }}

PS: I feel the need to appreciate @sbulage's effort in helping me resolve errors with this PR. Thanks Satya :)

### DIFF
--- a/cfme/tests/webui/test_general_ui.py
+++ b/cfme/tests/webui/test_general_ui.py
@@ -316,17 +316,24 @@ def test_tls_openssl_verify_mode(temp_appliance_preconfig, request):
     # 1
     assert "openssl_verify_mode" not in appliance.advanced_settings["smtp"]
     # 2
-    appliance.server.settings.update_smtp_server({"start_tls": False})
-    # reset `Start TLS Automatically`
-    request.addfinalizer(
-        lambda: appliance.update_advanced_settings({"smtp": {"start_tls": "<<reset>>"}})
-    )
+    view = navigate_to(appliance.server, "Server")
+    old_tls = view.smtp_server.start_tls.read()
+
+    # start_tls will be set to False since it's default value is True,
+    # although value of start_tls doesn't really affect the value of openssl_verify_mode
+    appliance.server.settings.update_smtp_server({"start_tls": not old_tls})
+
     # 3
     wait_for(
         lambda: "openssl_verify_mode" in appliance.advanced_settings["smtp"],
         timeout=50,
         delay=2,
     )
+    assert appliance.advanced_settings["smtp"]["openssl_verify_mode"] == "none"
+
+    # reset `Start TLS Automatically` and assert that
+    # the value of openssl_verify_mode is still none.
+    appliance.server.settings.update_smtp_server({"start_tls": old_tls})
     assert appliance.advanced_settings["smtp"]["openssl_verify_mode"] == "none"
 
 
--- a/cfme/tests/webui/test_general_ui.py
+++ b/cfme/tests/webui/test_general_ui.py
@@ -10,6 +10,7 @@ from cfme.common.provider_views import PhysicalProviderAddView
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.markers.env_markers.provider import ONE_PER_TYPE
 from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.wait import wait_for
 
 
 pytestmark = [test_requirements.general_ui]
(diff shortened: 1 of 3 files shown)


cfme/tests/webui/test_general_ui.py
```diff
@@ -288,6 +289,48 @@ def test_add_button_on_provider_all_page(
     assert displayed_view.is_displayed
 
 
+@pytest.mark.tier(1)
+@pytest.mark.meta(automates=[1475553])
+@test_requirements.general_ui
```
sbulage: This can be removed as it is `globally` defined.

sbulage:
LGTM :+1:

cfme/tests/webui/test_general_ui.py
```diff
@@ -288,6 +289,47 @@ def test_add_button_on_provider_all_page(
...
+    Polarion:
+        assignee: pvala
+        casecomponent: WebUI
+        caseimportance: medium
+        initialEstimate: 1/8h
+        testSteps:
+            1. Check if key `openssl_verify_mode` is present in the advanced configuration.
+            2. Navigate to Configuration and toggle `Start TLS Automatically`
+                of Outgoing SMTP E-mail Server.
+            3. Again check for the presence of `openssl_verify_mode` and check it's value.
```
john-dupuy: small typo `check its value`

cfme/tests/webui/test_general_ui.py
```diff
@@ -288,6 +289,47 @@ def test_add_button_on_provider_all_page(
     assert displayed_view.is_displayed
 
 
+@pytest.mark.tier(1)
+@pytest.mark.meta(automates=[1475553])
+def test_tls_openssl_verify_mode(temp_appliance_preconfig, request):
```
jawatts: **Question** Why pull a temp_appliance for this case? Is this test destructive to an appliance?

cfme/tests/webui/test_general_ui.py
```diff
@@ -288,6 +289,47 @@ def test_add_button_on_provider_all_page(
...
+        1475553
+    """
+    # take a fresh appliance where smtp has never been touched
+    appliance = temp_appliance_preconfig
+    # 1
+    assert "openssl_verify_mode" not in appliance.advanced_settings["smtp"]
+    # 2
+    appliance.server.settings.update_smtp_server({"start_tls": False})
+    # reset `Start TLS Automatically`
+    request.addfinalizer(
```
jawatts: Since this is on a temp_appliance, I don't think the finalizer is necessary here. The appliance will be destroyed after this test method is ran.

cfme/tests/webui/test_general_ui.py
```diff
@@ -288,6 +289,47 @@ def test_add_button_on_provider_all_page(
...
+    )
+    # 3
+    wait_for(
+        lambda: "openssl_verify_mode" in appliance.advanced_settings["smtp"],
+        timeout=50,
+        delay=2,
+    )
+    assert appliance.advanced_settings["smtp"]["openssl_verify_mode"] == "none"
+
+
```
jawatts: I would also recommend switching the pattern of this test up a bit:
- Detect what the current setting is
- Change the key to the opposite of what is currently set
- Perform verification
- Change it back to the original
- Decorate a finalizer function that actually asserts that its changed back

Let me know what you think of that pattern

cfme/tests/webui/test_general_ui.py
valaparthvi: I need an appliance where SMTP has never been configured, I am using temp_appliance to make sure I get one.

cfme/tests/webui/test_general_ui.py
valaparthvi: Hmm, reading the current setting from UI seems like a good idea and the pattern now seems kinda obvious :sweat_smile: 
Will add this thing. Thanks for the suggestion!

cfme/tests/webui/test_general_ui.py
```diff
@@ -288,6 +289,50 @@ def test_add_button_on_provider_all_page(
...
+        {"start_tls": not view.smtp_server.start_tls.read()}
+    )
+
+    # 3
+    wait_for(
+        lambda: "openssl_verify_mode" in appliance.advanced_settings["smtp"],
+        timeout=50,
+        delay=2,
+    )
+    assert appliance.advanced_settings["smtp"]["openssl_verify_mode"] == "none"
```
jawatts: **Question** So regardless of the value of `start_tls`, `open_verify_mode` should be set to none correct? I think it is worth changing `start_tls` back to what the original value was and verifying that `start_tls` was changed back and `openssl_verify_mode` is still none. Thoughts?

valaparthvi: Correct, it will always be none, unless you change the value of `SSL Verify Mode`.

You need to either change `SSL Verify Mode` or `Start TLS Automatically` to introduce `openssl_verify_mode` into the advanced_config YAML.

>Thoughts?

Um, maybe. I guess there's no harm in adding those few more steps.

jawatts: Thanks @valaparthvi , I think it would make the test more complete.

PR #9100, opened 2019-07-22
[1LP][RFR] Use common view class for timelines view

john-dupuy:
Fixing the `is_displayed` properties in 5.11. I would also like to fix cluster and host views since those are also impacting these tests. 

{{ pytest: --use-provider vsphere6-nested cfme/tests/infrastructure/test_timelines.py::test_infra_timeline_create_event }}

### DIFF
--- a/cfme/cloud/instance/__init__.py
+++ b/cfme/cloud/instance/__init__.py
@@ -10,6 +10,7 @@ from widgetastic_patternfly import CheckableBootstrapTreeview
 from widgetastic_patternfly import Dropdown
 
 from cfme.base.login import BaseLoggedInPage
+from cfme.common import TimelinesView
 from cfme.common.vm import VM
 from cfme.common.vm import VMCollection
 from cfme.common.vm_views import EditView
@@ -34,7 +35,6 @@ from widgetastic_manageiq import CompareToolBarActionsView
 from widgetastic_manageiq import ManageIQTree
 from widgetastic_manageiq import Search
 from widgetastic_manageiq import SummaryTable
-from widgetastic_manageiq import TimelinesView
 
 
 class InstanceDetailsToolbar(View):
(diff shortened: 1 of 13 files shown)


widgetastic_manageiq/__init__.py
```diff
@@ -3106,12 +3106,12 @@ class chart(TimelinesChart):  # NOQA
...
     @property
     def is_displayed(self):
-        return self.title.text == "Timelines"
+        return self.title.text == "Timeline View"
 
     @property
     def is_timelines(self):
         """method to check title text for base Timelines without overriding is_displayed"""
-        return self.title.text == "Timelines"
+        return self.title.text == "Timeline View"
```
mshriver: same change backported to 5.10? if it changed in specific z-stream I'd use VersionPicker again.

john-dupuy: Yep, I believe that this is was just wrong and the `is_displayed` logic that uses this check got messed up as well.

john-dupuy: Actually this is due to BZ https://bugzilla.redhat.com/show_bug.cgi?id=1732517, I modified this method to workaround that BZ

cfme/common/__init__.py
```diff
@@ -818,3 +821,33 @@ def step(self, *args, **kwargs):
             table.click_at("Custom Button Events")
         else:
             raise DestinationNotFound("Custom button event count 0")
+
+
+class TimelinesView(View):
```
digitronik: It looks similar to `ServerTimelinesView` better to inherit and overwrite `is_displayed` method.

cfme/cloud/instance/__init__.py
```diff
@@ -170,7 +170,20 @@ def is_displayed(self):
 
 
 class InstanceTimelinesView(TimelinesView, CloudInstanceView):
-    pass
+
+    @property
+    def is_displayed(self):
+        try:
+            check_object = self.breadcrumb.locations
+        except NoSuchElementException:
```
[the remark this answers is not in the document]
john-dupuy: The problem is that in 5.10, there is no `breadcrumb` for the instance and infra VM timelines page. It is there on e.g. the provider's, host's and cluster's page but not this one. So this page needs to be specific to VM and instance.

cfme/common/__init__.py
john-dupuy: Agreed, I can make this change.

cfme/cloud/instance/__init__.py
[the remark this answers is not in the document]
john-dupuy: I tried using `VersionPicker` here, but it still tries to resolve the breadcrumb, which results in an `NoSuchElementException` error anyway, so I thought it best to simplify it to a `try/except` block.

PR #9102, opened 2019-07-23
[1LP][RFR] Fixed is displayed of cloud provider image view

ganeshhubale:
- This PR fixes ```is_displayed``` of view - ```CloudProviderImagesView```
- Fixed error:
```
E       assert False
E        +  where False = <cfme.cloud.provider.CloudProviderImagesView object at 0x7f9570f9a7d0>.is_displayed

cfme/tests/cloud_infra_common/test_relationships.py:208: AssertionError
AssertionError
assert False
 +  where False = <cfme.cloud.provider.CloudProviderImagesView object at 0x7f9570f9a7d0>.is_displayed
```

{{ pytest: cfme/tests/cloud_infra_common/test_relationships.py::test_cloud_provider_relationships[azure-Images] --use-provider=azure --long-running -vvv }}

### DIFF
--- a/cfme/cloud/provider/__init__.py
+++ b/cfme/cloud/provider/__init__.py
@@ -30,6 +30,8 @@ from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.appliance.implementations.ui import navigator
 from cfme.utils.log import logger
 from cfme.utils.pretty import Pretty
+from cfme.utils.version import Version
+from cfme.utils.version import VersionPicker
 from cfme.utils.wait import wait_for
 from widgetastic_manageiq import ItemsToolBarViewSelector
 from widgetastic_manageiq import TimelinesView
@@ -79,8 +81,9 @@ class CloudProviderImagesView(BaseLoggedInPage):
     """
     @property
     def is_displayed(self):
+        location = VersionPicker({Version.lowest(): 'Cloud Providers', '5.11': 'Compute'}).pick()
         return (
-            self.breadcrumb.locations[0] == 'Cloud Providers' and
+            self.breadcrumb.locations[0] == location and
             self.entities.title.text == '{} (All Images)'.format(self.context['object'].name))
 
     breadcrumb = BreadCrumb()
--- a/cfme/cloud/provider/__init__.py
+++ b/cfme/cloud/provider/__init__.py
@@ -81,10 +81,12 @@ class CloudProviderImagesView(BaseLoggedInPage):
     """
     @property
     def is_displayed(self):
-        location = VersionPicker({Version.lowest(): 'Cloud Providers', '5.11': 'Compute'}).pick()
+        location = VersionPicker({Version.lowest(): 'Cloud Providers',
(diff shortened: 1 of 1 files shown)


cfme/cloud/provider/__init__.py
```diff
@@ -79,8 +81,9 @@ class CloudProviderImagesView(BaseLoggedInPage):
     """
     @property
     def is_displayed(self):
+        location = VersionPicker({Version.lowest(): 'Cloud Providers', '5.11': 'Compute'}).pick()
         return (
-            self.breadcrumb.locations[0] == 'Cloud Providers' and
+            self.breadcrumb.locations[0] == location and
```
mshriver: I think you need to go further here, not just change the expected string for the first breadcrumb item.

Breadcrumbs are significantly different in 5.11/Ivanchuck, and I think the important part here is that the breadcrumb is including 'Cloud Providers'. That the breadcrumb includes 'Compute' doesn't really help validate that we're on Cloud Provider images.

Look at asserting against the last and second-to-last items in the breadcrumb, as they are what indicate more specifics about the page being viewed. Use `self.breadcrumb.locations[-1] == self.context['obj'].name` kind of assertions, where that last item may be different between the versions.

ganeshhubale: Hi @mshriver 
- I am agree with your approach of going with last and second-to-last items in breadcrumb.  last item is same as title of the page and also same in both versions.
- Breadcrumb in 5.11: ```Compute > Clouds > Providers > Cloud Providers > azure > azure (All Images)```
- Breadcrumb in 5.10: ```Cloud > Providers > azure (All Images)```
- Also using second-to-last item which also indicates the more specifics about the page being viewed.
So I updated is_diplayes as below:
```
    def is_displayed(self):
        location = VersionPicker({Version.lowest(): 'Cloud Providers',
                                  '5.11': self.context['object'].name}).pick()
        return (
            self.breadcrumb.locations[-2] == location and
            self.entities.title.text == self.breadcrumb.locations[-1])
```

cfme/cloud/provider/__init__.py
```diff
@@ -79,9 +81,12 @@ class CloudProviderImagesView(BaseLoggedInPage):
     """
     @property
     def is_displayed(self):
+        location = VersionPicker({Version.lowest(): 'Cloud Providers',
+                                  '5.11': self.context['object'].name}).pick()
         return (
-            self.breadcrumb.locations[0] == 'Cloud Providers' and
-            self.entities.title.text == '{} (All Images)'.format(self.context['object'].name))
+            self.breadcrumb.locations[-2] == location and
+            self.entities.title.text == self.breadcrumb.locations[-1]
```
digitronik: `active_location` ^^

dajoRH:
Lint report for commit 6cfd6ce8d5a69e6203e10363cc3e2c994e3765c9:

`cfme/cloud/provider/__init__.py`:
- :heavy_exclamation_mark: Line 87:55: **W504** *line break after binary operator*

**No commit flaws detected**.

*CFME QE Bot*

PR #9105, opened 2019-07-24
[1LP][RFR]Automate test: test_filtering_vm_with_multiple_ips

valaparthvi:
Purpose or Intent
=================
__Adding tests__ `test_filtering_vm_with_multiple_ips` to cover BZ: 1684681.

{{ pytest: cfme/tests/infrastructure/test_vm_rest.py::test_filtering_vm_with_multiple_ips -v }}

### DIFF
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -195,6 +195,7 @@ class BaseVM(
     @property
     def all_ip_addresses(self):
         """Fetches all IP Addresses of a VM, pingable or otherwise."""
+        # TODO: Implement sentaku for this property with ViaMGMT impl
         view = navigate_to(self, "Details", use_resetter=False)
         try:
             return view.entities.summary('Properties').get_text_of("IP Address")
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -193,6 +193,16 @@ class BaseVM(
         return find_pingable(self.mgmt)
 
     @property
+    def all_ip_addresses(self):
+        """Fetches all IP Addresses of a VM, pingable or otherwise."""
+        view = navigate_to(self, "Details", use_resetter=False)
+        try:
+            return view.entities.summary('Properties').get_text_of("IP Address")
+        except NameError:
+            # since some providers have plural 'Addresses'.
+            return view.entities.summary('Properties').get_text_of("IP Addresses").split(", ")
+
+    @property
     def mac_address(self):
         """Fetches MAC Address of VM"""
         # TODO: We should update this with wrapanapi method when it becomes available.
--- a/cfme/tests/infrastructure/test_vm_rest.py
+++ b/cfme/tests/infrastructure/test_vm_rest.py
@@ -1,4 +1,6 @@
 # -*- coding: utf-8 -*-
+from random import choice
+
 import fauxfactory
 import pytest
 from manageiq_client.filters import Q
@@ -7,7 +9,6 @@ from cfme import test_requirements
 from cfme.infrastructure.provider import InfraProvider
 from cfme.markers.env_markers.provider import ONE
 from cfme.rest.gen_data import vm as _vm
-from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.rest import assert_response
 from cfme.utils.rest import delete_resources_from_collection
 from cfme.utils.rest import delete_resources_from_detail
@@ -139,24 +140,10 @@ def test_delete_vm_from_collection(vm):
     delete_resources_from_collection([vm], not_found=True, num_sec=300, delay=10)
 
 
-@pytest.fixture
-def vm_ip(appliance, provider):
-    vm = appliance.collections.infra_vms.instantiate(
-        provider.data["cap_and_util"]["capandu_vm"], provider
-    )
-    view = navigate_to(vm, "Details")
-    rest_vm = appliance.rest_api.collections.vms.get(name=vm.name)
-    # get all the ipaddresses from VM Details page
-    ipaddrs = (
-        view.entities.summary("Properties").get_text_of("IP Addresses").split(", ")
-    )
-    return rest_vm, ipaddrs
-
-
 @pytest.mark.tier(1)
 @pytest.mark.ignore_stream("5.10")
 @pytest.mark.meta(coverage=[1684681])
-def test_filtering_vm_with_multiple_ips(appliance, provider, vm_ip, soft_assert):
+def test_filtering_vm_with_multiple_ips(appliance, provider):
     """
     Polarion:
         assignee: pvala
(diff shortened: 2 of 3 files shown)


cfme/tests/infrastructure/test_vm_rest.py
```diff
@@ -135,3 +137,48 @@ def test_delete_vm_from_collection(vm):
...
     delete_resources_from_collection([vm], not_found=True, num_sec=300, delay=10)
+
+
+@pytest.fixture
+def vm_ip(appliance, provider):
+    vm = appliance.collections.infra_vms.instantiate(
+        provider.data["cap_and_util"]["capandu_vm"], provider
+    )
+    view = navigate_to(vm, "Details")
+    rest_vm = appliance.rest_api.collections.vms.get(name=vm.name)
```
digitronik: I think we have direct method;
```
vm.get_vm_via_rest()
```

cfme/tests/infrastructure/test_vm_rest.py
```diff
@@ -135,3 +137,48 @@ def test_delete_vm_from_collection(vm):
...
+    # get all the ipaddresses from VM Details page
+    ipaddrs = (
```
digitronik: optional: I think better to add this as `property` in model page with will return list of assignee ips.

valaparthvi: Hmm, I was thinking the same, but then I thought it won't be useful to others so I didn't add one. If you think it'll be helpful to everyone, sure I'll add one! :)

valaparthvi: Thanks for this suggestion @digitronik, I realized I had written plenty of code that was really not necessary!

cfme/tests/infrastructure/test_vm_rest.py
```diff
@@ -139,24 +140,10 @@ def test_delete_vm_from_collection(vm):
...
-        view.entities.summary("Properties").get_text_of("IP Addresses").split(", ")
-    )
-    return rest_vm, ipaddrs
-
-
 @pytest.mark.tier(1)
 @pytest.mark.ignore_stream("5.10")
 @pytest.mark.meta(coverage=[1684681])
-def test_filtering_vm_with_multiple_ips(appliance, provider, vm_ip, soft_assert):
+def test_filtering_vm_with_multiple_ips(appliance, provider, soft_assert):
```
digitronik: unused `soft_assert` arg

cfme/tests/infrastructure/test_vm_rest.py
```diff
@@ -135,3 +138,38 @@ def test_delete_vm_from_collection(vm):
...
+        expectedResults:
+            1.
+            2. Selected VM must be present in the resources sent by response.
+
+    Bugzilla:
+        1684681
+    """
+    # 1
+    vm = appliance.collections.infra_vms.instantiate(
+        provider.data["cap_and_util"]["capandu_vm"], provider
```
kedark3: @valaparthvi do you think its a good idea to catch a KeyError here and use pytest.skip in case a provider doesn't have required keys?
Or how about: 
`required_fields=[["cap_and_util", "capandu_azone"]],` in pytest mark provider

kedark3: @valaparthvi do you think its a good idea to catch a KeyError here and use pytest.skip in case a provider doesn't have required keys?
Or how about: 
`required_fields=[["cap_and_util", "capandu_azone"]],` in pytest mark provider

cfme/tests/infrastructure/test_vm_rest.py
```diff
@@ -135,3 +138,38 @@ def test_delete_vm_from_collection(vm):
         casecomponent: Infra
     """
     delete_resources_from_collection([vm], not_found=True, num_sec=300, delay=10)
+
+
+@pytest.mark.tier(1)
+@pytest.mark.ignore_stream("5.10")
+@pytest.mark.meta(coverage=[1684681])
```
sshveta: Please use this for automated tests 
@pytest.mark.meta(automates=[1684681])

cfme/tests/infrastructure/test_vm_rest.py
john-dupuy: I agree here

valaparthvi: Hmm, `required_fields` is new to me, surely I can add it, thanks KK, John :)

cfme/tests/infrastructure/test_vm_rest.py
```diff
@@ -135,3 +138,44 @@ def test_delete_vm_from_collection(vm):
...
+
+    Bugzilla:
+        1684681
+    """
+    # 1
+    vm = appliance.collections.infra_vms.instantiate(
+        provider.data["cap_and_util"]["capandu_vm"], provider
+    )
+    # 2
+    result = appliance.rest_api.collections.vms.filter(
```
ganeshhubale: @valaparthvi 
- Don't you think this is confusing(```# 1, # 2``` - In my understanding, you are referencing steps or setup)?
- According to me, It will be better if you add ```testSteps``` as  **inline** comments and you can remove test steps and expected results from polarion data.
Please let me know your view.

kedark3: @ganeshhubale I think it does indeed reference to `testSteps` but I don't think she needs to remove anything from Polarion docblock as that is part of what we use for uploading tests to polarion.

cfme/tests/infrastructure/test_vm_rest.py
digitronik: same code need to move in model; they why it can not be part of model?

cfme/tests/infrastructure/test_vm_rest.py
valaparthvi: Hey @ganeshhubale I don't think it's that confusing, it seems fine to me. There are only 2 teststeps, I think I would rather keep this as is.

cfme/common/vm.py
```diff
@@ -192,6 +192,16 @@ def ip_address(self):
         """
         return find_pingable(self.mgmt)
 
+    @property
+    def all_ip_addresses(self):
```
jawatts: **Question** There is a method for each wrapanapi system called `all_ips`, I do see it seems to be limited on some providers. I think it be worth updating those methods if possible. A `TODO` here can be acceptable for now

mshriver: There is a bigger issue festering in this module I think, and that is all of these convenience methods/properties that are fetching data from the UI after navigation, and are not implemented in a contextual way.

I don't think switching them to wrapanapi queries directly is correct, as an existing TODO in the module suggests.

This is a place where, if we're going to have convenience methods for these meta values associated with an entity, they should be implemented using Sentaku in order to provide a consistent and predictable context for all of the values.  It will be too easy for their implementations to skew, and some to be using UI while others are using REST, or providers directly.

Provider direct queries should likely be a context of their own for Sentaku, that I don't believe we actually have yet. Separating `ViaRest` from `ViaMGMT`

cfme/tests/infrastructure/test_vm_rest.py
```diff
@@ -135,3 +138,44 @@ def test_delete_vm_from_collection(vm):
...
+
+
+@pytest.mark.tier(1)
+@pytest.mark.ignore_stream("5.10")
+@pytest.mark.meta(automates=[1684681])
+@pytest.mark.provider(
+    classes=[InfraProvider],
+    selector=ONE,
+    override=True,
+    required_fields=[["cap_and_util", "capandu_vm"]],
```
mshriver: :+1:

cfme/common/vm.py
valaparthvi: Hey @mshriver, I'm not sure if I understand you correctly, but are you suggesting we define all these convenient methods as contextual methods and properties?
`all_ips` is defined for almost all the providers in wrapanapi, but some of them return data and some don't.

valaparthvi: Hi @mshriver @jawatts I've currently added a TODO, because adding a new implementation and implementing sentaku will be a slightly bigger change. Can we please continue with a TODO for now?

PR #9110, opened 2019-07-24
[NOTEST][WIP]Adding ReportParser class and miq blame command

kedark3:
Purpose or Intent
=================

__Adding miq blame__ command that will be able to take a Json file as input and return list of test cases per assignee and also provide failed tests report. 

```
miq blame --help
Usage: miq blame [OPTIONS] COMMAND [ARGS]...

  Functions for generating reports on BZs included in test suite metadata

Options:
  --help  Show this message and exit.

Commands:
  failure-report  Returns a Dictionary containing failed test cases per user
  tests-report    Returns a Dictionary containing test cases per user

```
failure-report:
```
miq blame failure-report --help
Usage: miq blame failure-report [OPTIONS]

  Returns a Dictionary containing failed test cases per user

Options:
  --report-file TEXT
  --help              Show this message and exit.


```


digitronik:
If possible rename `blame` to `codecov` so I will use same command group for extension. I have one card which will provide `codecov` and cover commands like

1. automate and manual % as per assignee
2. automate and manual % as per FA
etc... 

Thanks for initialize this... I am facing problem with `requirements` but i will add basic implementation.

kedark3:
@digitronik I have no issue renaming. @mshriver @jawatts what you guys think?

dajoRH:
Lint report for commit 3c7a8638557ba526274d1d6c3d8001f9dd93e8b1:

`cfme/scripting/report_parser.py`:
- :red_circle: Line 0:0: **EPY3** *Usage of .iteritems*

**No commit flaws detected**.

*CFME QE Bot*

cfme/scripting/report_parser.py
```diff
@@ -0,0 +1,97 @@
+"""Script to parse test reports."""
+import json
+import smtplib
+from email.MIMEMultipart import MIMEMultipart
```
john-dupuy: Please update for py3 compatibility: `from email.mime.multipart import MIMEMultipart`

cfme/scripting/report_parser.py
```diff
@@ -0,0 +1,97 @@
...
+from email.MIMEText import MIMEText
```
john-dupuy: Same thing here

cfme/scripting/report_parser.py
kedark3: Done.

cfme/scripting/report_parser.py
kedark3: Done.

cfme/scripting/report_parser.py
```diff
@@ -0,0 +1,97 @@
...
+                        {test_case: result})
+        return assignee_wise_results
+
+
+@click.group("blame", help="Functions for generating reports on Tests grouped by user")
+def main():
+    pass
+
+
+@main.command("test-report", help="Returns a Dictionary containing failed test cases per user")
```
john-dupuy: @kedark3 I see in your PR description that you had two commands `test-report` and `failed-report`? However, here this command is reporting on failed test cases? Is this just leftover when the PR was in a different state?

john-dupuy: Thoughts on naming this something else? My concern is that this makes it sound like you're running a test. One thought is that the calling sequence could be `miq blame <user>` so that it makes it seem like you're literally blaming the person :smile: The all the other args could be optional. Not sure if it's worth the refactor though, so I'd like to hear your thoughts.

cfme/scripting/report_parser.py
```diff
@@ -0,0 +1,97 @@
...
+@click.option("--report-file", help="Path to the file with json data")
+@click.option("--email-config-yaml", help="If you decide to use this option,copy yaml file similar"
+    " to report_parser_email_conf.yaml.template with correct values and pass its path."
+    " No email would be sent if --user is not set.",
+    type=click.Path())
+@click.option("--user", help="Set this if you want to filter the report for a specific user",
+    default='')
+@click.option("--filter-by", type=click.Choice(['passed', 'failed', 'skipped']),
+    help="Defaults to 'failed'", default='failed')
+def test_results_per_assignee(report_file, email_config_yaml, user, filter_by):
```
john-dupuy: When I run `miq blame test-report`, I get an Unhandled TypeError that doesn't really explain why my command failed, it'd be useful to have it print some required about the required options.

cfme/scripting/report_parser.py
```diff
@@ -0,0 +1,97 @@
...
+    rp = ReportParser(report_file)
+    test_report_dict = rp.parse_report_per_user(filter_by, user)
+    # user is required before following routine would execute, else no email is sent
+    if test_report_dict and email_config_yaml and user:
```
john-dupuy: Is there anyway to generate the report for all users? Because if so I'd expect the default to do that.

cfme/scripting/report_parser.py
```diff
@@ -0,0 +1,97 @@
...
+        try:
+            with open(email_config_yaml, "r") as config_file:
+                email_config = yaml.safe_load(config_file)
+        except IOError:
+            logger.exception("Please provide valid file path for --email-config-yaml")
+        server = smtplib.SMTP(host=email_config['email_host'],
```
john-dupuy: I'd prefer formatting like:
```python,
        server = smtplib.SMTP(
            host=email_config['email_host'],
            port=email_config['email_port']
        )

File: cfme/scripting/report_parser.py [hunk 7]
```diff
@@ -0,0 +1,97 @@
...
+            port=email_config['email_port'])
+        email_msg = MIMEMultipart()
+        try:
+            email_msg['From'] = email_config['from_addr']
+            email_msg['To'] = '{}@redhat.com'.format(user)
+            email_msg['Subject'] = email_config['email_subject']
+            pretty_test_report_dict = '\n'.join(['{:<15} {:<10}'.format(k, v)
+                                        for k, v in test_report_dict[user].items()])
+            email_msg.attach(MIMEText(email_config['email_template']
```
john-dupuy: Same about formatting here.

cfme/scripting/report_parser.py
```diff
@@ -0,0 +1,97 @@
...
+@click.group("blame", help="Functions for generating reports on Tests grouped by user")
+def main():
+    pass
+
+
+@main.command("test-report", help="Returns a Dictionary containing failed test cases per user")
+@click.option("--report-file", help="Path to the file with json data")
+@click.option("--email-config-yaml", help="If you decide to use this option,copy yaml file similar"
+    " to report_parser_email_conf.yaml.template with correct values and pass its path."
+    " No email would be sent if --user is not set.",
```
john-dupuy: `No email will be sent if --user is not set` -> shouldn't this be under the `--user` field?

cfme/scripting/report_parser.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+@click.group("blame", help="Functions for generating reports on Tests grouped by user")
+def main():
+    pass
+
+
+@main.command("test-report", help="Returns a Dictionary containing failed test cases per user")
+@click.option("--report-file", help="Path to the file with json data")
+@click.option("--email-config-yaml", help="If you decide to use this option,copy yaml file similar"
```
john-dupuy: Any default value for this option? I suggest that the help message simply read `The path to your email configuration file, e.g. "report_parser_email_conf.yaml"`

cfme/scripting/report_parser.py
```diff
@@ -0,0 +1,97 @@
...
+        return assignee_wise_results
+
+
+@click.group("blame", help="Functions for generating reports on Tests grouped by user")
+def main():
+    pass
+
+
+@main.command("test-report", help="Returns a Dictionary containing failed test cases per user")
+@click.option("--report-file", help="Path to the file with json data")
```
john-dupuy: Any default value for this option?

cfme/scripting/report_parser.py
kedark3: Leftover.

kedark3: @john-dupuy I'd like to get one more review on this before I refactor. Don't want to refactor only to go back to what it was. I like your idea too. :)

cfme/scripting/report_parser.py
kedark3: Making --report-file required would make it go away.

cfme/scripting/report_parser.py
kedark3: So by default user is `''` so it fetches report for each user. And I am adding a `default_to_addr`  so that if user is not defined, email would go to `default_to_addr`.  Currently I derive email from username, but if we want to be explicit that would require more work. I understood the task as we will have our kerberos username in assignee.

cfme/scripting/report_parser.py
kedark3: No default values really. But I updated the message.

cfme/scripting/report_parser.py
kedark3: I think I can create something dummy, do we want that in the repo?

cfme/scripting/report_parser.py
kedark3: I marked that arg as required so it should not give TypeError now.

dajoRH:
*CFME QE Bot*

PR #9111, opened 2019-07-25
[1LP][RFR] Migration Throttling

sshveta:
{{ pytest: cfme/tests/v2v/test_migration_throttling.py --use-provider rhv43 --use-provider vsphere67-ims --provider-limit 2 -vvvv --long-running}}

Migration Throttling for v2v automated.


dajoRH:
Lint report for commit 87e06528af9065964125b7e769b9f83058c88187:

`cfme/tests/v2v/test_migration_throttling.py`:
- :red_circle: Line 4:1: **F811** *redefinition of unused 'pytest' from line 3*
- :red_circle: Line 15:1: **F401** *'cfme.v2v.conversion_hosts as mt' imported but unused*
- :heavy_exclamation_mark: Line 86:1: **W391** *blank line at end of file*

`cfme/v2v/migration_throttling.py`:
- :bangbang: Line 50:45: **E231** *missing whitespace after ':'*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/v2v/test_migration_throttling.py
```diff
@@ -0,0 +1,91 @@
...
+        fixture_name="source_provider",
+        required_flags=["v2v"],
+        scope="module",
+    ),
+    pytest.mark.usefixtures("v2v_provider_setup"),
+]
+
+
+@pytest.mark.parametrize(
+    "mapping_data_multiple_vm_obj_single_datastore",
```
john-dupuy: Can this variable name be shortened? Why not just call it `data`?

cfme/tests/v2v/test_migration_throttling.py
```diff
@@ -0,0 +1,91 @@
...
+        testtype: functional
+        startsin: 5.10
+    """
+    migration_throttling_collection = appliance.collections.v2v_throttling_settings
+    # Setting throttling at 2 concurrent migration per conversion host
+    migration_throttling_collection.set_max_migration_per_conv_host("2")
+
+    infrastructure_mapping_collection = appliance.collections.v2v_infra_mappings
+    mapping_data = mapping_data_multiple_vm_obj_single_datastore.infra_mapping_data
+    mapping = infrastructure_mapping_collection.create(**mapping_data)
```
john-dupuy: Can this `mapping` be moved into a fixture? Since the test is on throttling and not creating the mapping.

cfme/v2v/migration_throttling.py
```diff
@@ -0,0 +1,61 @@
...
+        Args:
+            migration_per_conv_host: maximum concurrent migration per conversion host
+            migration_per_provider: maximum concurrent migrtaion per provider
+    """
+    migration_per_conv_host = attr.ib(default=None)
+    migration_per_provider = attr.ib(default=None)
+
+
+@attr.s
+class MigrationThrottlingCollection(BaseCollection):
```
john-dupuy: Is there any `create` method for such a collection? It looks like `set_max_migration_per_conv_host` is being used like a `create` method.

cfme/tests/v2v/test_migration_throttling.py
sshveta: @john-dupuy : Infrastructure mapping takes mapping data which is different for each test and is passed from test parameters . Not able to create a fixture for it

cfme/v2v/migration_throttling.py
sshveta: Screenshot from 2019-07-26 12-23-58

Setting max values is the only operation permitted from this UI .

cfme/v2v/migration_throttling.py
```diff
@@ -0,0 +1,61 @@
...
+        return self.instantiate(migration_per_conv_host=migration_per_conv_host)
+
+
+@navigator.register(MigrationThrottlingCollection, "All")
+class MigrationThrottling(CFMENavigateStep):
+    prerequisite = NavigateToAttribute("appliance.server", "LoggedIn")
+    VIEW = MigrationSettingsView
+
+    def step(self):
+        self.prerequisite_view.navigation.select("Compute", "Migration", "Migration Settings")
```
john-dupuy: Migration has it's own tab in the latest build of 5.11
Screenshot from 2019-07-26 15-40-19

john-dupuy:
@sshveta please check PRT for 5.11, failed on a `Could not do Waiting for view [MigrationSettingsView] to display`

widgetastic_manageiq/__init__.py
```diff
@@ -5195,11 +5195,13 @@ def read_additional_info_popup(self, vm_name):
             el = self.browser.element(self.ITEM_ADDITIONAL_INFO_POPUP_LOCATOR)
             return {
                 "Status": self.browser.text("./h3", parent=el),
-                "Started": self.browser.text("./div[2]/div/div[1]", parent=el),
-                "Description": self.browser.text("./div[2]/div/div[2]", parent=el),
-                "Conversion Host": (self.browser.text("./div[2]/div/div[3]", parent=el)).split(
+                "Started": (self.browser.text("./div[2]/div/div[1]", parent=el)).split(": ")[1],
```
john-dupuy: Q: Is there any way to get around these indices? This could be difficult to maintain with these pages being updated often.

cfme/tests/v2v/test_migration_throttling.py
john-dupuy: @sshveta Any thoughts on this? My main concern is code readability.

john-dupuy:
Also can you rebase and squash your commits? Oh and why do all of your commits have the same commit message? Just curious if it's a workflow thing for you.

cfme/tests/v2v/test_migration_throttling.py
sshveta: @john-dupuy : There are lot of fixtures like these in v2v_fixtures.py with different data set like mapping_data_single_datastore, mapping_data_single_network, mapping_data_dual_vm_obj_dual_datastore, 
mapping_data_vm_obj_dual_nics and they create data as their name suggest .

john-dupuy: Got it, that's okay :+1:

widgetastic_manageiq/__init__.py
sshveta: Can I address that in a following PR ?

john-dupuy: Yep that's fine with me.

cfme/v2v/migration_throttling.py
mshriver: John is getting at an important issue here, but sort of for the wrong reasons.

A BaseCollection/BaseEntity derived class, modeling a collection of `things` :tm: in MIQ does not necessarily need a `create` method.  I would agree with him, that if this were a proper collection, that this method should be `create`.

But the real issue is that this isn't a collection, or an entity, its simply a configuration setting for Migration. Since there is no 'Migration' `BaseEntity` derived class encompassing this part of the appliance, its a setting that is global to the appliance, effectively.

In this case, it should inherit from `NavigatableMixin`, and not `BaseEntity`, like the appliance level settings:
https://github.com/ManageIQ/integration_tests/blob/master/cfme/configure/settings.py

Additionally, this view already exists, for Conversion Hosts:
https://github.com/ManageIQ/integration_tests/blob/master/cfme/v2v/conversion_hosts.py#L28

We should not be duplicating the view for this navigation destination or class. There should be a single navigation destination to `'MigrationSettings'`, where its returned `View` class has both tabs. Methods can be placed in the `MigrationSettings` class that inherits `NavigatableMixin`, and to which the navigation destination is registered, to update these settings by filling the tab.

sshveta: Hi @mshriver : Implemented Migration throttling inheriting """NavigatableMixin""".
Please take a look

john-dupuy:
@sshveta please check 5.11 failures.

sshveta:
> @sshveta please check 5.11 failures.

John , Test passed locally . Re-testing PR.

cfme/tests/v2v/test_migration_throttling.py
```diff
@@ -0,0 +1,89 @@
...
+from cfme import test_requirements
+from cfme.cloud.provider.openstack import OpenStackProvider
+from cfme.fixtures.provider import rhel69_template
+from cfme.fixtures.provider import rhel7_minimal
+from cfme.fixtures.provider import ubuntu16_template
+from cfme.infrastructure.provider.rhevm import RHEVMProvider
+from cfme.infrastructure.provider.virtualcenter import VMwareProvider
+from cfme.markers.env_markers.provider import ONE_PER_TYPE
+from cfme.markers.env_markers.provider import ONE_PER_VERSION
+from cfme.v2v.migration_settings import MigrationSettings as migration_settings
```
john-dupuy: So to distinguish between the class and the instance, I think you should keep the class name as `MigrationSettings` and call the instance `migration_settings`

cfme/tests/v2v/test_migration_throttling.py
```diff
@@ -0,0 +1,89 @@
...
+    Polarion:
+        assignee: sshveta
+        casecomponent: V2V
+        initialEstimate: 1/4h
+        caseimportance: high
+        caseposneg: positive
+        testtype: functional
+        startsin: 5.10
+    """
+    migration_settings(appliance).migration_throttling.set_max_migration_per_conv_host("2")
```
john-dupuy: For readability, I suggest
```python,
migration_settings = MigrationSettings(appliance)
migration_settings.migration_throttling.set_max_migration_per_conv_host("2")
```

cfme/v2v/migration_settings.py
```diff
@@ -25,28 +27,74 @@
 from widgetastic_manageiq import WaitTab
 
 
-class MigrationSettingsView(BaseLoggedInPage):
+class MigrationThrottling(Updateable, NavigatableMixin):
+
```
john-dupuy: Please add a docblock for this class

cfme/v2v/migration_settings.py
```diff
@@ -25,28 +27,74 @@
...
+        Args:
+            migration_per_conv_host: Integer
+        """
+        view = navigate_to(self.migration_settings, 'MigrationSettings')
+        view.tabs.migration_throttling.fill({'migration_per_conv_host': migration_per_conv_host})
+        view.tabs.migration_throttling.apply_btn.click()
+
+
+class MigrationSettings(Updateable, NavigatableMixin):
+
```
john-dupuy: Please add a docblock for this class

cfme/tests/v2v/test_migration_throttling.py
```diff
@@ -0,0 +1,89 @@
...
+
+@pytest.mark.parametrize(
+    "mapping_data_multiple_vm_obj_single_datastore",
+    [
+        ["nfs", "nfs", [rhel7_minimal, rhel69_template, ubuntu16_template]],
+    ],
+    indirect=True,
+)
+def test_migration_throttling(
+    request, appliance, mapping_data_multiple_vm_obj_single_datastore
```
kedark3: **Suggestion/Required**: `mapping_data_multiple_vm_obj_single_datastore` this and some other fixtures are missing `addfinalizer` that removes the vms that are created, can you please add it? Would be great if you can address in this PR but if not, open a new PR. Sooner the better so that tests cleanup properly.

cfme/tests/v2v/test_migration_throttling.py
```diff
@@ -0,0 +1,89 @@
...
+        infrastructure_mapping_collection.delete(mapping)
+
+    migration_plan_collection = appliance.collections.v2v_migration_plans
+    migration_plan = migration_plan_collection.create(
+        name="plan_{}".format(fauxfactory.gen_alphanumeric()),
+        description="desc_{}".format(fauxfactory.gen_alphanumeric()),
+        infra_map=mapping.name,
+        vm_list=mapping_data_multiple_vm_obj_single_datastore.vm_list,
+    )
+    assert migration_plan.wait_for_state("Started")
```
kedark3: Are we configuring all hosts to be conversion hosts? Can you point me to that code please?

cfme/tests/v2v/test_migration_throttling.py
```diff
@@ -0,0 +1,89 @@
...
+    # Check if conversion host is shown for each VM
+    for vm in vms:
+        try:
+            popup_text = request_details_list.read_additional_info_popup(vm)
+            # open__additional_info_popup function also closes opened popup in our case
+            request_details_list.open_additional_info_popup(vm)
+            conversion_host_popup.append(popup_text['Conversion Host'])
+        except NoSuchElementException:
+            # conversion host might be empty for throttled vm
+            # until it is waiting for other VM's to finish
```
kedark3: I was thinking it would be probably a good idea to make sure we have at least 2 conv hosts configured and then wait until we see conv host for each vm in its popup. 
That way, we don't assume that `conversion host might be empty for throttled vm` and we actually have conv host for the the vms before we go to assert statement at the end of the test.
what do you think?

cfme/v2v/migration_settings.py
```diff
@@ -25,28 +27,74 @@
...
+        self.migration_settings = migration_settings
+
+    def set_max_migration_per_conv_host(self, migration_per_conv_host):
+        """Migration throttling for conversion host
+        Args:
+            migration_per_conv_host: Integer
+        """
+        view = navigate_to(self.migration_settings, 'MigrationSettings')
+        view.tabs.migration_throttling.fill({'migration_per_conv_host': migration_per_conv_host})
+        view.tabs.migration_throttling.apply_btn.click()
```
kedark3: Would it be a good idea to also provide ability to read currently set value(a property or `get()`) ? So that we can assert the value we have set is correct.

cfme/v2v/migration_settings.py
```diff
@@ -25,28 +27,74 @@
...
+
+        @View.nested
+        class migration_throttling(WaitTab):  # noqa
+            fill_strategy = WaitFillViewStrategy("15s")
+            TAB_NAME = 'Migration Throttling'
+            including_entities = View.include(MigrationThrottlingForm, use_parent=True)
+
+        @View.nested
+        class conversion_hosts(WaitTab):  # noqa
+            fill_strategy = WaitFillViewStrategy("20s")
```
kedark3: How did we decide "15s" or "20s" ? Trying to understand out of curiosity.

cfme/v2v/migration_settings.py
```diff
@@ -25,28 +27,74 @@
...
 
-        @property
-        def is_displayed(self):
-            return self.title.text == "Configured Conversion Host"
+    @View.nested
+    class tabs(View):  # noqa
+        """The tabs on the page"""
+
+        @View.nested
+        class migration_throttling(WaitTab):  # noqa
```
kedark3: Do we not need a `is_displayed()` here? I see one in the `conversion_hosts` class below

cfme/tests/v2v/test_migration_throttling.py
sshveta: @kedark3 : VM's are being cleaned up at 
https://github.com/ManageIQ/integration_tests/blob/master/cfme/fixtures/v2v_fixtures.py#L384

cfme/tests/v2v/test_migration_throttling.py
sshveta: @kedark3 : https://github.com/ManageIQ/integration_tests/blob/master/cfme/fixtures/v2v_fixtures.py#L232

cfme/tests/v2v/test_migration_throttling.py
sshveta: @kedark3 : All the conversion hosts are configured . There is no control on how and which conversion host is picked .so for a test all three migrations can have two different conv hosts or it might wait for two conversions to finish (in which case it will be empty ) and then start third migration.

cfme/tests/v2v/test_migration_throttling.py
kedark3: Okay that's good. I did see some vms were not being cleaned up properly hence asked to check. Thanks!

cfme/v2v/migration_settings.py
sshveta: yes , We can read the value set in UI , Will cover that in separate test .

cfme/v2v/migration_settings.py
sshveta: In this case '15s' was less in some runs so I increased it a little .

cfme/tests/v2v/test_migration_throttling.py
kedark3: @sshveta thanks. My assumption was that if we have 2 conversion hosts with 2 migrations limit for each host, in that case for this test with three vms, if we wait until all vms are assigned a conversion host, we will see 2 vms assigned to host1 and 1 vm assigned to host2. That makes it more accurate to verify. See what I am saying? Would that be possible to implement or be too unpredictable?

cfme/v2v/migration_settings.py
kedark3: Got it. as long as you have sufficient confident that these values are good enough. :)

cfme/tests/v2v/test_migration_throttling.py
sshveta: So this sets limit for concurrent migration per conversion host .
Same host can also be picked for all the VM's but only two will run at a given time and third one will start only when one VM is done (if the same host is picked)

cfme/tests/v2v/test_migration_throttling.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+    migration_settings(appliance).migration_throttling.set_max_migration_per_conv_host("2")
+
+    infrastructure_mapping_collection = appliance.collections.v2v_infra_mappings
+    mapping_data = mapping_data_multiple_vm_obj_single_datastore.infra_mapping_data
+    mapping = infrastructure_mapping_collection.create(**mapping_data)
+
+    @request.addfinalizer
+    def _cleanup():
+        infrastructure_mapping_collection.delete(mapping)
```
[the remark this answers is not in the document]
kedark3: Thanks for adding this.

cfme/tests/v2v/test_migration_throttling.py
kedark3: I get that, but isn't there also supposed to be a feature that distributes the VM Migration across the available hosts evenly? I am almost sure we had that too.  Not trying to drag it too much, just see if we can improve this in any way. Thanks for cooperating.

cfme/v2v/migration_settings.py
kedark3: That would be good candidate for UI tests, but would it be possible to add that method along with this class while we are at it? We know we might need it/might wanna test it in future right?

cfme/tests/v2v/test_migration_throttling.py
sshveta: I am not aware of that feature @kedark3 , may be @Yadnyawalkya knows . Yadnya , thoughts ?

Yadnyawalkya: @kedark3  we do have such RFE (I searched, did not got that link but I am aware there is such RFE) but we don't have ability to choose conversion host per migration atm as @sshveta  mentioned.

dajoRH:
Lint report for commit af47cfb4188f602ed87b994608c51883e865e2d2:

`cfme/v2v/migration_settings.py`:
- :bangbang: Line 54:1: **E302** *expected 2 blank lines, found 1*

**No commit flaws detected**.

*CFME QE Bot*

cfme/v2v/migration_settings.py
sshveta: @kedark3 : Although I was not targeting to cover it in this test/PR , I have added the method .There will be more tests coming for this UI .

cfme/tests/v2v/test_migration_throttling.py
sshveta: Added related BZ in comment

sshveta:
5.10 tests are passing locally . Failure is sporadic , trying to check if a delay in click will fix the problem

cfme/v2v/migration_settings.py
```diff
@@ -25,28 +27,89 @@
...
+        """ Set migration throttling per conversion host
+        Args:
+            migration_per_conv_host: Integer
+        """
+        view = navigate_to(self.migration_settings, 'MigrationSettings')
+        view.tabs.migration_throttling.fill({'migration_per_conv_host': migration_per_conv_host})
+        view.tabs.migration_throttling.apply_btn.click()
+
+    def get_max_migration_per_conv_host(self):
+        """ Set migration throttling per conversion host
```
kedark3: Please update docblock `set` <-> `get`

cfme/v2v/migration_settings.py
kedark3: looks good!

cfme/tests/v2v/test_migration_throttling.py
```diff
@@ -0,0 +1,93 @@
...
+        scope="module",
+    ),
+    pytest.mark.usefixtures("v2v_provider_setup"),
+]
+
+
+@pytest.mark.parametrize(
+    "mapping_data_multiple_vm_obj_single_datastore",
+    [
+        ["nfs", "nfs", [rhel7_minimal, rhel69_template, ubuntu16_template]],
```
john-dupuy: Q: Why parametrize against a single parameter? Plans to parametrize against more?

sshveta: The fixtures like mapping_data_multiple_vm_obj_single_datastore in v2v_fixtures.py are written to take parameters from requests hence paramterized this test.

https://github.com/ManageIQ/integration_tests/blob/master/cfme/fixtures/v2v_fixtures.py#L467

PR #9113, opened 2019-07-25
[1LP][RFR] New Test: Test vm retire by new user

ganeshhubale:
## Purpose or Intent
- This PR contains new test case to check whether newly created user with role vm-user can able to retire the vm provisioned by admin

### PRT Run
 {{ pytest: cfme/tests/infrastructure/test_vm_power_control.py -k 'test_retire_vm_with_vm_user_role' --use-provider=complete --long-running -vvv }}

### DIFF
--- a/cfme/tests/infrastructure/test_vm_power_control.py
+++ b/cfme/tests/infrastructure/test_vm_power_control.py
@@ -2,7 +2,6 @@
 import random
 import time
 
-import fauxfactory
 import pytest
 
 from cfme import test_requirements
@@ -14,6 +13,7 @@ from cfme.infrastructure.provider.scvmm import SCVMMProvider
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.infrastructure.virtual_machines import VmsTemplatesAllView
 from cfme.markers.env_markers.provider import ONE_PER_TYPE
+from cfme.rest.gen_data import users as _users
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.blockers import BZ
 from cfme.utils.generators import random_vm_name
@@ -615,39 +615,38 @@ def test_guest_os_shutdown(appliance, provider, testing_vm_tools, ensure_vm_runn
                     "ui: {} should ==  orig: {}".format(new_last_boot_time, last_boot_time))
 
 
+@pytest.fixture(scope="function")
+def new_user(request, appliance):
+    user, user_data = _users(request, appliance, group="EvmGroup-vm_user")
+    yield appliance.collections.users.instantiate(
+        name=user[0].name,
+        credential=Credential(principal=user_data[0]["userid"], secret=user_data[0]["password"]),
+    )
+
+    if user[0].exists:
+        user[0].action.delete()
+
+
 @pytest.mark.tier(1)
 @pytest.mark.meta(automates=[1687597])
 @pytest.mark.provider([VMwareProvider], selector=ONE_PER_TYPE, override=True)
-def test_retire_vm_with_vm_user_role(request, appliance, testing_vm):
+def test_retire_vm_with_vm_user_role(new_user, appliance, testing_vm):
     """
+    Bugzilla:
+        1687597
+
     Polarion:
         assignee: ghubale
         initialEstimate: 1/8h
         caseposneg: positive
         startsin: 5.10
         casecomponent: Automate
-
-    Bugzilla:
-        1687597
+        setup:
+            1. Provision vm
+        testSteps:
+            1. Create custom user with 'EvmRole_vm-user' role
+            2. Retire VM by log-in to custom user
     """
-    # Using group - "EvmGroup-vm_user" because this is attached to role - "EvmRole-vm_user"
-    user_group = appliance.collections.groups.instantiate(description="EvmGroup-vm_user")
-
-    # Creating a user with 'EvmGroup-vm_user' group to retire VM provisioned by admin
-    new_user = appliance.collections.users.create(
-        name="user_{}".format(fauxfactory.gen_alphanumeric().lower()),
-        credential=Credential(
-            principal="uid{}".format(fauxfactory.gen_alphanumeric(4)),
-            secret="{password}".format(password=fauxfactory.gen_alphanumeric(4)),
-        ),
-        email=fauxfactory.gen_email(),
-        groups=user_group,
-        cost_center="Workload",
-        value_assign="Database",
-    )
-    user = appliance.rest_api.collections.users.get(name=new_user.name)
-    request.addfinalizer(user.action.delete)
-
     # Log in with new user to retire the vm
     with new_user:
         new = testing_vm.find_quadicon(from_any_provider=True)
--- a/cfme/tests/infrastructure/test_vm_power_control.py
+++ b/cfme/tests/infrastructure/test_vm_power_control.py
@@ -2,14 +2,18 @@
 import random
 import time
 
+import fauxfactory
 import pytest
 
 from cfme import test_requirements
+from cfme.base.credential import Credential
 from cfme.base.login import BaseLoggedInPage
 from cfme.infrastructure.provider import InfraProvider
 from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.infrastructure.provider.scvmm import SCVMMProvider
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
+from cfme.infrastructure.virtual_machines import VmsTemplatesAllView
+from cfme.markers.env_markers.provider import ONE_PER_TYPE
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.blockers import BZ
 from cfme.utils.generators import random_vm_name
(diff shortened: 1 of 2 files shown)


cfme/tests/infrastructure/test_vm_power_control.py
```diff
@@ -599,3 +603,46 @@ def test_guest_os_shutdown(appliance, provider, testing_vm_tools, ensure_vm_runn
...
+        casecomponent: Automate
+
+    Bugzilla:
+        1687597
+    """
+    # Using group - "EvmGroup-vm_user" because this is attached to role - "EvmRole-vm_user"
+    user_group = appliance.collections.groups.instantiate(description="EvmGroup-vm_user")
+
+    # Creating a user with 'EvmGroup-vm_user' group to retire VM provisioned by admin
+    new_user = appliance.collections.users.create(
```
valaparthvi: How about using REST to create the role, group, and user? That you will only need to create and instantiate them, no need to add finalizer to delete them.
If it doesn't affect your test, I think using REST will make your test case much faster.

cfme/tests/infrastructure/test_vm_power_control.py
```diff
@@ -599,3 +603,46 @@ def test_guest_os_shutdown(appliance, provider, testing_vm_tools, ensure_vm_runn
...
+        value_assign="Database",
+    )
+    user = appliance.rest_api.collections.users.get(name=new_user.name)
+    request.addfinalizer(user.action.delete)
+
+    # Log in with new user to retire the vm
+    with new_user:
+        new = testing_vm.find_quadicon(from_any_provider=True)
+        new.check()
+        view = testing_vm.create_view(VmsTemplatesAllView)
```
valaparthvi: Do you need to add `wait` to wait for `is_displayed`?

cfme/tests/infrastructure/test_vm_power_control.py
john-dupuy: I think the role, user, and group should be moved into fixtures (possibly only one fixture), since they are out of the scope of the test. And I definitely agree with using REST -- we should use rest anywhere possible to speed up our test cases. Thoughts?

cfme/tests/infrastructure/test_vm_power_control.py
john-dupuy: @valaparthvi I expect it is already on this page from the `find_quadicon` method, but might be good to add a wait for an extra assertion

ganeshhubale: @valaparthvi @john-dupuy 
-  I am already on that page. 
- According to me, I do not require wait for ```is_displayed``` of that view because I have already checked vm available on that page. It shows ```is_displayed``` is successful for that view. Otherwise it would have failed to check that vm.

cfme/tests/infrastructure/test_vm_power_control.py
```diff
@@ -609,3 +613,43 @@ def test_guest_os_shutdown(appliance, provider, testing_vm_tools, ensure_vm_runn
...
+        assignee: ghubale
+        initialEstimate: 1/8h
+        caseposneg: positive
+        startsin: 5.10
+        casecomponent: Automate
+        setup:
+            1. Provision vm
+        testSteps:
+            1. Create custom user with 'EvmRole_vm-user' role
+            3. Retire VM by log-in to custom user
```
john-dupuy: Small typo, should be `2.`

cfme/tests/infrastructure/test_vm_power_control.py
```diff
@@ -609,3 +613,43 @@ def test_guest_os_shutdown(appliance, provider, testing_vm_tools, ensure_vm_runn
...
+
+
+@pytest.fixture(scope="function")
+def new_user(request, appliance):
+    user, user_data = _users(request, appliance, group="EvmGroup-vm_user")
+    user = appliance.collections.users.instantiate(
+        name=user[0].name,
+        credential=Credential(principal=user_data[0]["userid"], secret=user_data[0]["password"]),
+    )
+    yield user
```
john-dupuy: Q: any cleanup necessary?

cfme/tests/infrastructure/test_vm_power_control.py
ganeshhubale: Thanks @john-dupuy

cfme/tests/infrastructure/test_vm_power_control.py
```diff
@@ -609,3 +613,45 @@ def test_guest_os_shutdown(appliance, provider, testing_vm_tools, ensure_vm_runn
...
+@pytest.fixture(scope="function")
+def new_user(request, appliance):
+    user, user_data = _users(request, appliance, group="EvmGroup-vm_user")
+    new_user = appliance.collections.users.instantiate(
+        name=user[0].name,
+        credential=Credential(principal=user_data[0]["userid"], secret=user_data[0]["password"]),
+    )
+    yield new_user
+    if user[0].exists:
+        user[0].action.delete()
```
john-dupuy: Can you not do `new_user.delete_if_exists()`?

ganeshhubale: Then it will delete via UI.
So I thought it would be better to delete this user via REST as we have created it using REST.

john-dupuy: Ah, gotcha! Sounds good!

cfme/tests/infrastructure/test_vm_power_control.py
```diff
@@ -609,3 +613,45 @@ def test_guest_os_shutdown(appliance, provider, testing_vm_tools, ensure_vm_runn
...
+
+@pytest.fixture(scope="function")
+def new_user(request, appliance):
+    user, user_data = _users(request, appliance, group="EvmGroup-vm_user")
+    new_user = appliance.collections.users.instantiate(
+        name=user[0].name,
+        credential=Credential(principal=user_data[0]["userid"], secret=user_data[0]["password"]),
+    )
+    yield new_user
+    if user[0].exists:
```
valaparthvi: We do not need this. `users` uses `create_skeleton` function, which will do the cleanup on its own. See this: https://github.com/ManageIQ/integration_tests/blob/master/cfme/rest/gen_data.py#L487

ganeshhubale: @valaparthvi
 - I went through this fixture(https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/intelligence/reports/test_menus.py#L133). Can you plese confirm why are we [using this](https://github.com/ManageIQ/integration_tests/pull/9113#discussion-diff-311385009R626) in that fixture?

valaparthvi: Ah, that was a mistake on my part. I recently realized that `create_skeleton` deletes the entities it creates, I am working on removing that and other places where I might've used it.

dajoRH:
I detected some fixture changes in commit 89fd9c3c12216ff338490b5cbf061bb87ca0250a

The *local* fixture **`new_user`** is used in the following files:
- **cfme/tests/infrastructure/test_vm_power_control.py**
    - *test_retire_vm_with_vm_user_role*

dajoRH:
I detected some fixture changes in commit fe0c7d725eb2822d52ac109574782f0d5ab002d8

The *local* fixture **`new_user`** is used in the following files:
- **cfme/tests/infrastructure/test_vm_power_control.py**
    - *test_retire_vm_with_vm_user_role*

dajoRH:
I detected some fixture changes in commit a0dc7b85c0b2c50e7ea9e8bd87c4b8ed0b0c379d

The *local* fixture **`new_user`** is used in the following files:
- **cfme/tests/infrastructure/test_vm_power_control.py**
    - *test_retire_vm_with_vm_user_role*

cfme/tests/infrastructure/test_vm_power_control.py
valaparthvi: Conclusion: Since we cannot delete a user if they're tied to a request on 5.11, we decided to explicitly delete the user here.

cfme/tests/infrastructure/test_vm_power_control.py
```diff
@@ -609,3 +613,45 @@ def test_guest_os_shutdown(appliance, provider, testing_vm_tools, ensure_vm_runn
...
+        casecomponent: Automate
+        setup:
+            1. Provision vm
+        testSteps:
+            1. Create custom user with 'EvmRole_vm-user' role
+            2. Retire VM by log-in to custom user
+    """
+    # Log in with new user to retire the vm
+    with new_user:
+        new = testing_vm.find_quadicon(from_any_provider=True)
```
sbulage: @ganeshhubale I would suggest here to create `view` by using `get_entity` by `testing_vm.name` and use `check()`. Your approach is right but it will be restrict to Grid View and will not allow you to check above behavior for any other view. @digitronik please confirm this. :)

ganeshhubale: Thanks @sbulage @digitronik

cfme/tests/infrastructure/test_vm_power_control.py
```diff
@@ -609,3 +613,45 @@ def test_guest_os_shutdown(appliance, provider, testing_vm_tools, ensure_vm_runn
...
+        new.check()
+        view = testing_vm.create_view(VmsTemplatesAllView)
+        assert view.toolbar.lifecycle.item_enabled("Retire selected items")
+        testing_vm.retire()
+        testing_vm.wait_for_vm_state_change(desired_state="retired", timeout=720, from_details=True)
```
sbulage: @ganeshhubale As test case related to `retire`, can you please `assert` the whether it is retired or not.?

PR #9114, opened 2019-07-25
[1LP][RFR] Nav work around for 5.11.0.16

digitronik:

## Purpose or Intent

In latest build `MyServiceView` is broken due to `VeriticalNav`; It will break all tests related to `service.` `currently_selected` returning value like  `['Services', 'My Services', 'Services', 'Providers']`. 

some local checks
```
ipdb> self.navigation.currently_selected                                          
['Services', 'My Services', 'Services', 'Providers']
ipdb> self.navigation.currently_selected                                          
['Storage', 'Block Storage', 'Managers']
ipdb> self.navigation.currently_selected                                          
['Services', 'Catalogs']
```
vertical_nav_bar

### PRT Run

{{pytest: cfme/tests/automate/custom_button/test_service_vm_custom_button.py::test_custom_button_with_dynamic_dialog_vm -v}}

### DIFF
--- a/cfme/services/myservice/ui.py
+++ b/cfme/services/myservice/ui.py
@@ -55,9 +55,16 @@ class MyServicesView(BaseLoggedInPage):
 
     @property
     def in_myservices(self):
+        # Slicing in currently_selected is workaround for BZ-1733489
+        nav_selected = (
+            self.navigation.currently_selected[:2]
+            if BZ(1733489).blocks
+            else self.navigation.currently_selected
+        )
         return (
             self.logged_in_as_current_user and
-            self.navigation.currently_selected == ["Services", "My Services"])
+            nav_selected == ["Services", "My Services"]
+        )
 
     @property
     def is_displayed(self):


cfme/services/myservice/ui.py
```diff
@@ -55,9 +55,12 @@ class MyServicesView(BaseLoggedInPage):
 
     @property
     def in_myservices(self):
+        # Adding workaround for 5.11.0.16.
+        # Its collecting ['Services', 'My Services', 'Services', 'Providers']
         return (
             self.logged_in_as_current_user and
-            self.navigation.currently_selected == ["Services", "My Services"])
+            self.navigation.currently_selected[0:2] == ["Services", "My Services"]
```
jaryn: Well this is quite minor, but you only need `self.navigation.currently_selected[:2]` there

digitronik: yup, but its ok in case of indexing

cfme/services/myservice/ui.py
```diff
@@ -55,9 +55,11 @@ class MyServicesView(BaseLoggedInPage):
 
     @property
     def in_myservices(self):
+        # Slicing in currently_selected is workaround for BZ-1733489
```
jawatts: I think i may be good to do BZ(1733489).blocks here? Thoughts?

digitronik: why we need blocks?
we are just limiting to index 1. If BZ fix anyway it will work fine.

digitronik:
Hey Justin... Added please have glance... we need it fast :+1:

PR #9117, opened 2019-07-26
[1LP][RFR] automate custom menu test

digitronik:

### PRT Run

{{pytest: cfme/tests/automate/custom_button/test_custom_menu.py}}

### DIFF
--- a/cfme/tests/automate/custom_button/test_custom_menu.py
+++ b/cfme/tests/automate/custom_button/test_custom_menu.py
@@ -1,15 +1,40 @@
 import pytest
 
 from cfme import test_requirements
+from cfme.utils.appliance.implementations.ui import navigate_to
 
 pytestmark = [test_requirements.custom_button]
+
+
+ADVANCE_SETTING_CUSTOM_MENU = {
+    "ui": {
+        "custom_menu": [
+            {
+                "type": "item",
+                "icon": "fa fa-cubes",
+                "id": "redhat",
+                "name": "RedHat",
+                "href": "https://www.redhat.com",
+                "rbac": "vm_explorer",
+            },
+            {
+                "type": "item",
+                "icon": "pficon pficon-project",
+                "id": "manageiq",
+                "name": "ManageIQ",
+                "href": "https://manageiq.org",
+                "rbac": "vm_explorer",
+            },
+        ]
+    }
+}
 
 
 @pytest.mark.manual
 @pytest.mark.tier(1)
 @pytest.mark.ignore_stream("5.10")
 @pytest.mark.meta(coverage=[1678151])
-def test_custom_menu_display():
+def test_custom_menu_display(appliance):
     """Add Custom Menu in Left Navigation bar as Admin
 
     Requirements for custom menu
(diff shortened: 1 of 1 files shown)


cfme/tests/automate/custom_button/test_custom_menu.py
```diff
@@ -54,4 +79,11 @@ def test_custom_menu_display():
     Bugzilla:
         1678151
     """
-    pass
+    appliance.update_advanced_settings(ADVANCE_SETTING_CUSTOM_MENU)
+    appliance.reboot()
+    view = navigate_to(appliance.server, "LoggedIn")
+
+    for menu in ["RedHat", "ManageIQ"]:
```
john-dupuy: Thoughts on parametrizing the test case against these two different menus? Probably not worth it due to increased time for the test to run, especially with the reboot. But just curious.

digitronik: Actually `reboot` time is not issue. If you check, we are updating both menu at a same time so no need of parametrization. this loop only checks are similar for both menus.

If we parametrized it will just like dummy test like performing same thing two time.

cfme/tests/automate/custom_button/test_custom_menu.py
```diff
@@ -54,4 +79,11 @@ def test_custom_menu_display():
...
+        view.navigation.select(menu)
+        assert view.navigation.currently_selected == [menu]
+        assert "id={}".format(menu.lower()) in appliance.browser.selenium.current_url
```
jawatts: **Question** Is there no cleanup after the test completes? I would like to see the menu being removed

cfme/tests/automate/custom_button/test_custom_menu.py
```diff
@@ -54,4 +79,11 @@ def test_custom_menu_display():
     Bugzilla:
         1678151
     """
-    pass
+    appliance.update_advanced_settings(ADVANCE_SETTING_CUSTOM_MENU)
+    appliance.reboot()
```
jawatts: **Question** Do we need a reboot here, or can we just do a evmserverd restart?

john-dupuy: @digitronik I just tested this and an evmserverd restart should work for this test, it adds the menus after the webui starts back up

jawatts: Going to mark this **Required**, I don't think we should be rebooting the appliance.

digitronik:
@john-dupuy thanks for checking this option; I was unknow about `advance setting change` I would like to change it. @jawatts

digitronik:
Ignore 5.10 fails... it will be ignore by other PR. which by mistaken, i add in that https://github.com/ManageIQ/integration_tests/pull/9159/files#diff-491bef3639003bd202f87514687ce746R10

dajoRH:
*CFME QE Bot*

cfme/tests/automate/custom_button/test_custom_menu.py
```diff
@@ -55,4 +88,12 @@ def test_custom_menu_display():
     Bugzilla:
         1678151
     """
-    pass
+    update_adv_setting_and_wait(appliance, ADVANCE_SETTING_CUSTOM_MENU)
+    request.addfinalizer(lambda: update_adv_setting_and_wait(appliance, {"ui": "<<reset>>"}))
```
jawatts: :+1:

PR #9123, opened 2019-07-29
[1LP][RFR] add test_appliance_contains_ansible_modules

jaryn:
Adding test checking presence of ansible modules

{{pytest: cfme/tests/test_appliance.py::test_appliance_contains_ansible_modules -v}}

### DIFF
--- a/cfme/tests/test_appliance.py
+++ b/cfme/tests/test_appliance.py
@@ -373,6 +373,26 @@ def test_appliance_exec_scripts():
     pass
 
 
+@pytest.mark.tier(2)
+@pytest.mark.meta(automates=[1678130])
+def test_appliance_contains_ansible_modules(appliance):
+    """
+    check that there are ansible modules included in the appliance.
+
+    Polarion:
+        assignee: jhenner
+        casecomponent: Appliance
+        caseimportance: high
+        initialEstimate: 1/6h
+        startsin: 5.9.2
+    Bugzilla:
+        1678130
+    """
+    assert appliance.ssh_client.run_command(
+        'stat /opt/rh/cfme-gemset/bundler/gems/cfme-conten*/content/ansible/roles'
+    ).success
+
+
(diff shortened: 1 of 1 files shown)


cfme/tests/test_appliance.py
```diff
@@ -373,6 +373,25 @@ def test_appliance_exec_scripts():
...
+    check that there are ansible modules included in the appliance.
+
+    Polarion:
+        assignee: jhenner
+        casecomponent: Appliance
+        caseimportance: high
+        initialEstimate: 1/6h
+        startsin: 5.9.2
+    Bugzilla:
+        1678130
```
digitronik: This test is for automating RFR right? Please add `automates` meta if possible.

jaryn: Oh, I forgot. THanks

cfme/tests/test_appliance.py
```diff
@@ -373,6 +373,26 @@ def test_appliance_exec_scripts():
...
+    """
+    assert appliance.ssh_client.run_command(
+        'stat /opt/rh/cfme-gemset/bundler/gems/cfme-conten*/content/ansible/roles'
```
sbulage: Along with this, You can also check for gems is available or not like:
```
[root@localhost vmdb]# bundle exec gem list | grep ansible
ansible_tower_client (0.20.1)
manageiq-providers-ansible_tower (0.1.0)
```

jaryn: @sbulage I am not sure it is a good idea. If I understand the bug, the package containing the files may be different depending on are we testing MiQ or CFME.

Also, the title of the BZ is to make sure the modules are there. It is not about checking some package is present. The package may contain the modules, or may not. Testing for the files presence is more specific.

sbulage: Okay.

PR #9053, opened 2019-07-10
[WIPTEST][NOMERGE] Drop py2.7 support

mshriver:
2to3 driven conversions, manual updates to travis configuration and requirements files.

Please also see #9088 for updates to how freeze is handled.

PR #9128, opened 2019-07-30
[RFR] Use ubuntu 14.04 for travis

john-dupuy:
just a band-aid for travis until #9053 is merged. Recently travis was updated to use ubuntu 16.04 by default, which breaks our quickstart. So here force it to use 14.04 until #9053 is merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ stages:
 - linting
 - test
 - deploy
+dist: trusty
 
 jobs:
   include:


.travis.yml
```diff
@@ -6,6 +6,7 @@ stages:
 - linting
 - test
 - deploy
+dist: trusty
```
digitronik: still if found as per doc Travis using `Ubuntu Trusty 14.04 default` even other thought; what its breaking? `Xenial` should be comparable for both `py2` and `py3` so just confused.

john-dupuy: @digitronik It fails in installing pycurl, this issue is not in py3.7, but I have seen it on all the PRs opened today.

PR #9132, opened 2019-07-31
[1LP][RFR] Test case updated for 5.10z

ganeshhubale:
## Purpose or Intent
- This BZ(1728706) is hot fix for 5.10z. Hence updating this test case for 5.10z
- Fixed Error for ```test_object_attributes```:
```
>               assert view.target_object.all_options > 0
E               TypeError: '>' not supported between instances of 'list' and 'int'

cfme/tests/automate/test_simulation.py:38: TypeError
TypeError
b"'>' not supported between instances of 'list' and 'int'"
```

### PRT Run

 {{ pytest: cfme/tests/automate/ -k 'test_user_requester_for_lifecycle_provision or test_object_attributes' --use-template-cache -qsvv }}

### DIFF
--- a/cfme/tests/automate/test_service_automate.py
+++ b/cfme/tests/automate/test_service_automate.py
@@ -7,7 +7,6 @@ from cfme.base.credential import Credential
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.provisioning import do_vm_provisioning
 from cfme.services.service_catalogs import ServiceCatalogs
-from cfme.utils.blockers import BZ
 from cfme.utils.generators import random_vm_name
 from cfme.utils.log_validator import LogValidator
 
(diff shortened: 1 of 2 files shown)


cfme/tests/automate/test_service_automate.py
```diff
@@ -30,11 +29,8 @@ def new_users(appliance):
 
     yield users
     for user in users:
-        if not BZ(1720273).blocks:
-            user.delete_if_exists()
-        else:
-            user = appliance.rest_api.collections.users.get(name=user.name)
-            user.action.delete()
+        user = appliance.rest_api.collections.users.get(name=user.name)
```
valaparthvi: This is a little out of scope of this PR, but how about using [`users`](https://github.com/ManageIQ/integration_tests/blob/master/cfme/rest/gen_data.py#L466) to create users? You'll probably need to change the group in `gen_data.users`, but then you will not need to delete the users that you create. You will only need to instantiate those users.

PR #9146, opened 2019-08-02
[1LP][RFR] fix py3 conversion issue

izapolsk:


izapolsk:
@mshriver, we will likely need to update other parts of framework as well

PR #9148, opened 2019-08-02
[1LP][RFR]Fix test: test_providers_summary and test_datastores_summary

valaparthvi:
## Purpose or Intent

- __Fixing__ `test_providers_summary` to pick correct name of `Hostname` for both the versions of appliance.
- __Fixiing__ `test_datastores_summary` to work for py3 upgrade.

{{ pytest: cfme/tests/intelligence/reports/test_canned_corresponds.py -k "test_providers_summary or test_datastores_summary" --use-provider=vsphere65-nested -vvvv }}

### DIFF
--- a/cfme/tests/intelligence/reports/test_canned_corresponds.py
+++ b/cfme/tests/intelligence/reports/test_canned_corresponds.py
@@ -56,8 +56,8 @@ def test_providers_summary(appliance, soft_assert):
         details_view = navigate_to(provider_object, 'Details')
         props = details_view.entities.summary("Properties")
 
-        hostname = ("Host Name", "Hostname")
-        soft_assert(props.get_text_of(hostname[0]) == provider[hostname[1]],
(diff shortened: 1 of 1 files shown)


lina-nikiforova:
Q: @valaparthvi I tried it locally and I see it fails from the second provider in the run. I feel like I've seen a bug for this. Is this the case?

valaparthvi:
Hey @lina-nikiforova, this is the bug https://bugzilla.redhat.com/show_bug.cgi?id=1504010
I'm working on fixing this thing, its been pending since a looooong time.

My current fix only fixes py3 and appliance version issue.

lina-nikiforova:
thanks, @valaparthvi ! I know it's not related to this PR but wasn't sure if I remember this bug correctly. Maybe it will be helpful to add this BZ number to the test (not required in this PR).

PR #9149, opened 2019-08-02
[1LP][RFR] New Test: Test miq stop and abort in state machines

ganeshhubale:
## Purpose or Intent
 - Test miq stop and miq abort in state machines

### PRT Run

{{ pytest: cfme/tests/automate/test_instance.py -k 'test_miq_stop_abort_with_state_machines' -vv }}

### DIFF
--- a/cfme/tests/automate/test_instance.py
+++ b/cfme/tests/automate/test_instance.py
@@ -10,10 +10,6 @@ from cfme.utils.update import update
 
 pytestmark = [test_requirements.automate]
 
-STATE1 = fauxfactory.gen_alpha()
-STATE2 = fauxfactory.gen_alpha()
-STATE3 = fauxfactory.gen_alpha()
-
 
 @pytest.mark.sauce
 @pytest.mark.tier(2)
@@ -365,8 +361,12 @@ def setup(domain, klass, namespace):
     """This fixture creates common domain, namespace, two classes, instance and method. This setup
     is common for both parameterized tests"""
 
+    state1 = fauxfactory.gen_alpha()
+    state2 = fauxfactory.gen_alpha()
+    state3 = fauxfactory.gen_alpha()
+
     klass.schema.add_fields(*[{'name': state, 'type': 'State'}
-                              for state in [STATE1, STATE2, STATE3]])
+                              for state in [state1, state2, state3]])
 
     # Adding method to execute via parent instance(state machine instance)
     method = klass.methods.create(
@@ -384,7 +384,7 @@ def setup(domain, klass, namespace):
         name=fauxfactory.gen_alphanumeric(),
         display_name=fauxfactory.gen_alphanumeric(),
         description=fauxfactory.gen_alphanumeric(),
-        fields={STATE3: {"value": f"METHOD::{method.name}"}}
+        fields={state3: {"value": f"METHOD::{method.name}"}}
     )
 
     # Creating other class for two child instances
@@ -394,9 +394,9 @@ def setup(domain, klass, namespace):
         description=fauxfactory.gen_alpha()
     )
     klass2.schema.add_fields(*[{'name': state, 'type': 'State'}
-                               for state in [STATE1, STATE2, STATE3]])
+                               for state in [state1, state2]])
 
-    yield klass2, instance
+    yield klass2, instance, state1, state2, state3
     method.delete_if_exists()
     instance.delete_if_exists()
     klass2.delete_if_exists()
@@ -419,8 +419,29 @@ def test_miq_stop_abort_with_state_machines(request, setup, process, domain, kla
         startsin: 5.9
         casecomponent: Automate
         tags: automate
+        setup:
+            1. Create automate domain, namespace, two classes - class A and class B
+            2. Class A and B should have schema of type - State
+            3. Create instance A1 and method A1 under class A. Here instance A1 should call method
+               A1
+            4. Create two instances B1, B2 and three methods B1, B2, B3 under class B. Here instance
+               B1 should call two methods B1 and B2. Also instance B2 should call method B3
+            5. Now Update instance A1 to call instance B1 and B2
+        testSteps:
+            1. Navigate to Automation > Automate > Simulation page and execute instance A1
+            2. If all the methods contain process - MIQ_STOP
+            3. If any method(in this test case - method B1, B2, B3) contains process - MIQ_ABORT
+        expectedResults:
+            1.
+            2. MIQ_STOP process only stops execution of current instance but it allows state machine
+               to execute other instances. So we are able to see execution of prent method -
+               method A1 - "Hello from method of parent instance"
+            3. MIQ_ABORT process stops execution of current instance as well as its parent instance
+              (here it stops instance B1 and then instance A1) which leads to no execution parent
+              instances - instance A1. So we are not able to see execution of parent method -
+              method A1 - "Hello from method of parent instance"
     """
-    klass2, instance = setup
+    klass2, instance, state1, state2, state3 = setup
 
     # Creating three child methods to execute via child instances
     child_method = list(map(lambda num: klass2.methods.create(
@@ -433,12 +454,12 @@ def test_miq_stop_abort_with_state_machines(request, setup, process, domain, kla
                 """
     ), ["first", "second", "third"]))
 
-    fields = [{STATE1: {'value': f"METHOD::{child_method[0].name}"},
-              STATE2: {'value': f"METHOD::{child_method[1].name}"}},
-              {STATE1: {'value': f"METHOD::{child_method[2].name}"}}]
+    fields = [{state1: {'value': f"METHOD::{child_method[0].name}"},
+              state2: {'value': f"METHOD::{child_method[1].name}"}},
+              {state1: {'value': f"METHOD::{child_method[2].name}"}}]
 
     # Creating two child instances
-    child_inst = list(map(lambda field: setup[0].instances.create(
+    child_inst = list(map(lambda field: klass2.instances.create(
         name=fauxfactory.gen_alphanumeric(),
         display_name=fauxfactory.gen_alphanumeric(),
         description=fauxfactory.gen_alphanumeric(),
@@ -455,10 +476,10 @@ def test_miq_stop_abort_with_state_machines(request, setup, process, domain, kla
     # Updating parent instance fields to execute child instances
     with update(instance):
         instance.fields = {
-            STATE1: {
+            state1: {
                 "value": f"/{domain.name}/{namespace.name}/{klass2.name}/{child_inst[0].name}"
             },
-            STATE2: {
+            state2: {
                 "value": f"/{domain.name}/{namespace.name}/{klass2.name}/{child_inst[1].name}"
             },
         }
--- a/cfme/tests/automate/test_instance.py
+++ b/cfme/tests/automate/test_instance.py
@@ -10,6 +10,10 @@ from cfme.utils.update import update
 
 pytestmark = [test_requirements.automate]
 
+STATE1 = fauxfactory.gen_alpha()
+STATE2 = fauxfactory.gen_alpha()
+STATE3 = fauxfactory.gen_alpha()
+
 
 @pytest.mark.sauce
 @pytest.mark.tier(2)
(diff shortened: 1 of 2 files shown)


cfme/tests/automate/test_instance.py
```diff
@@ -354,3 +354,117 @@ def test_quota_source_value(request, entity, search, copy_quota_instance, generi
...
+        testtype: functional
+        startsin: 5.9
+        casecomponent: Automate
+        tags: automate
+    """
+    state1 = fauxfactory.gen_alpha()
+    state2 = fauxfactory.gen_alpha()
+    state3 = fauxfactory.gen_alpha()
+    for state in [state1, state2, state3]:
+        klass.schema.add_fields({'name': state, 'type': 'State'})
```
valaparthvi: I think it will be better if you send all the fields at once instead of sending individual fields.
Something like: `klass.schema.add_fields(*[{'name': state, 'type': 'State'} for state in [state1, state2, state3]])` or any other suitable way.

cfme/tests/automate/test_instance.py
```diff
@@ -354,3 +354,117 @@ def test_quota_source_value(request, entity, search, copy_quota_instance, generi
...
+    # Creating other class for two child instances
+    klass2 = namespace.classes.create(
+        name=fauxfactory.gen_alpha(),
+        display_name=fauxfactory.gen_alpha(),
+        description=fauxfactory.gen_alpha()
+    )
+    request.addfinalizer(klass2.delete_if_exists)
+
+    for state in [state1, state2]:
+        klass2.schema.add_fields({'name': state, 'type': 'State'})
```
valaparthvi: Same as above, send all the fields at once instead of looping.

cfme/tests/automate/test_instance.py
```diff
@@ -354,3 +354,117 @@ def test_quota_source_value(request, entity, search, copy_quota_instance, generi
...
+
+    # Creating three child methods to execute via child instances
+    child_method = list(map(lambda num: klass2.methods.create(
```
valaparthvi: I saw that for the second parametrized test, new `state1`, `state2` and `state3` are created inside the same `klass` that was used for the first test..
So if everything until creating this child method is common for both the parameters, why not just use the same parent klass/method/instance for both the parametrized test?

cfme/tests/automate/test_instance.py
```diff
@@ -354,3 +354,117 @@ def test_quota_source_value(request, entity, search, copy_quota_instance, generi
...
+        display_name=fauxfactory.gen_alphanumeric(),
+        description=fauxfactory.gen_alphanumeric(),
+        fields=field
+    ), fields))
+
+    # Updating parent instance fields to execute child instances
+    with update(instance):
+        instance.fields = {
+            state1: {
+                "value": f"/{domain.name}/{namespace.name}/{klass2.name}/{child_inst[0].name}"
```
valaparthvi: How do you feel about using `"/".join(child_inst[0].tree_path_name_only[1:])` instead of this long string formating?

cfme/tests/automate/test_instance.py
ganeshhubale: Thanks @valaparthvi for this suggestion. 
I am making changes as per this suggestion.

cfme/tests/automate/test_instance.py
ganeshhubale: I would like to keep it as it is.
Because I find it more readable than using ```tree_path_name_only``` otherwise it will be very difficult for me to understand from which domain/namespace/class method is getting execute.

dajoRH:
I detected some fixture changes in commit 0e21efec4b1ff7de97a5357f17b554de648ca4c3

The *local* fixture **`setup`** is used in the following files:
- **cfme/tests/automate/test_instance.py**
    - *test_miq_stop_abort_with_state_machines*

cfme/tests/automate/test_instance.py
```diff
@@ -10,6 +10,10 @@
 
 pytestmark = [test_requirements.automate]
 
+STATE1 = fauxfactory.gen_alpha()
```
john-dupuy: Are these constants necessary? What is the reason you've decided not to put them in the fixture?

cfme/tests/automate/test_instance.py
```diff
@@ -354,3 +358,130 @@ def test_quota_source_value(request, entity, search, copy_quota_instance, generi
...
     assert result.validate(wait="60s")
+
+
+@pytest.fixture(scope='module')
+def setup(domain, klass, namespace):
+    """This fixture creates common domain, namespace, two classes, instance and method. This setup
+    is common for both parameterized tests"""
+
+    klass.schema.add_fields(*[{'name': state, 'type': 'State'}
+                              for state in [STATE1, STATE2, STATE3]])
```
john-dupuy: I.e. Same comment as above, can you do `for state in fauxfactory.gen_alpha()`?

cfme/tests/automate/test_instance.py
```diff
@@ -354,3 +358,130 @@ def test_quota_source_value(request, entity, search, copy_quota_instance, generi
...
+
+    # Adding method to execute via parent instance(state machine instance)
+    method = klass.methods.create(
+        name=fauxfactory.gen_alphanumeric(),
+        display_name=fauxfactory.gen_alphanumeric(),
+        location='inline',
+        script="""
+               \n$evm.log(:info, "Hello from method of parent instance")
```
john-dupuy: Are these newline characters `\n` necessary?

cfme/tests/automate/test_instance.py
john-dupuy: Oh I see, you are using them below, but is it possible to define them in the fixture? And have them be yielded. I will leave this as _optional_

cfme/tests/automate/test_instance.py
```diff
@@ -354,3 +358,130 @@ def test_quota_source_value(request, entity, search, copy_quota_instance, generi
...
+        location='inline',
+        script=f"""
+               \n$evm.log(:info, "This is method {num}")
+                \nexit {process.upper()}
+                """
+    ), ["first", "second", "third"]))
+
+    fields = [{STATE1: {'value': f"METHOD::{child_method[0].name}"},
+              STATE2: {'value': f"METHOD::{child_method[1].name}"}},
+              {STATE1: {'value': f"METHOD::{child_method[2].name}"}}]
```
john-dupuy: Should this be state 3? Also is it possible to use list comprehension here? Something like: 
```python,
fields = [ 
    { state: {"value": f"METHOD::{child_method[{i}].name}" for i, state in enumerate(states) }
]
```

john-dupuy: Ah, I see that they are different dictionaries, so probably this comment doesn't apply.

cfme/tests/automate/test_instance.py
```diff
@@ -354,3 +358,130 @@ def test_quota_source_value(request, entity, search, copy_quota_instance, generi
...
+    def finalize():
+        for method in child_method:
+            method.delete_if_exists()
+        for inst in child_inst:
+            inst.delete_if_exists()
+
+    # Updating parent instance fields to execute child instances
+    with update(instance):
+        instance.fields = {
+            STATE1: {
```
john-dupuy: Thoughts on using comprehension here?

cfme/tests/automate/test_instance.py
```diff
@@ -354,3 +358,130 @@ def test_quota_source_value(request, entity, search, copy_quota_instance, generi
...
+    method.delete_if_exists()
+    instance.delete_if_exists()
+    klass2.delete_if_exists()
+
+
+@pytest.mark.tier(1)
+@pytest.mark.meta(automates=[1441353])
+@pytest.mark.parametrize("process", ["miq_stop", "miq_abort"])
+def test_miq_stop_abort_with_state_machines(request, setup, process, domain, klass, namespace):
+    """
```
kedark3: **Suggestion:** Please add some description on what your test is trying to do.

cfme/tests/automate/test_instance.py
ganeshhubale: yes. It is required to keep ruby code in well format while putting it in inline method data.

cfme/tests/automate/test_instance.py
ganeshhubale: Yes @kedark3 
I agree with you here. Actually this is very complicated test case to understand. So I am adding ```testSteps``` here

cfme/tests/automate/test_instance.py
ganeshhubale: I have added these variable into the fixture and yielded

cfme/tests/automate/test_instance.py
ganeshhubale: Hey @john-dupuy 
- I tried to use comprehension here but getting error. 
```
ipdb> instace_fields ={state: { 
                "value": f"/{domain.name}/{namespace.name}/{klass2.name}/{child_inst[i].name}" 
           } for i, state in enumerate([state1, state2])} 
                                                                                                                                                                             
*** NameError: name 'domain' is not defined
ipdb> domain.name                                                                                                                                                            
'VdQbatotQz'                                                                                                                                                                        
```
- hence keeping as it is.

dajoRH:
I detected some fixture changes in commit 1369e69cc6ea01c674566df04ab4868b882b84ff

The *local* fixture **`setup`** is used in the following files:
- **cfme/tests/automate/test_instance.py**
    - *test_miq_stop_abort_with_state_machines*

cfme/tests/automate/test_instance.py
kedark3: Thanks Ganesh.

cfme/tests/automate/test_instance.py
```diff
@@ -354,3 +354,155 @@ def test_quota_source_value(request, entity, search, copy_quota_instance, generi
...
+        fields={state3: {"value": f"METHOD::{method.name}"}}
+    )
+
+    # Creating other class for two child instances
+    klass2 = namespace.classes.create(
+        name=fauxfactory.gen_alpha(),
+        display_name=fauxfactory.gen_alpha(),
+        description=fauxfactory.gen_alpha()
+    )
+    klass2.schema.add_fields(*[{'name': state, 'type': 'State'}
```
valaparthvi: Were you not adding only 2 fields to `klass2`? [This.](https://github.com/ManageIQ/integration_tests/pull/9149/commits/c0bcf2bec11c8ef1b96144bc0ad19a7791c2e576#diff-f16976f85bd8a542ad66c2f70c4c4a6fR413)

cfme/tests/automate/test_instance.py
```diff
@@ -354,3 +354,155 @@ def test_quota_source_value(request, entity, search, copy_quota_instance, generi
...
+            1. Create automate domain, namespace, two classes - class A and class B
+            2. Class A and B should have schema of type - State
+            3. Create instance A1 and method A1 under class A. Here instance A1 should call method
+               A1
+            4. Create two instances B1, B2 and three methods B1, B2, B3 under class B. Here instance
+               B1 should call two methods B1 and B2. Also instance B2 should call method B3
+            5. Now Update instance A1 to call instance B1 and B2
+        testSteps:
+            1. Navigate to Automation > Automate > Simulation page and execute instance A1
+            2. If all the methods contain process - MIQ_STOP
```
valaparthvi: Can you modify this test step a little? I think it should be more like - _" Execute all the methods containing process - MIQ_STOP."_
I think this would make more sense, let me know what you think.
Same for step3.

cfme/tests/automate/test_instance.py
```diff
@@ -354,3 +354,155 @@ def test_quota_source_value(request, entity, search, copy_quota_instance, generi
...
+    """
+    Bugzilla:
+        1441353
+
+    Polarion:
+        assignee: ghubale
+        initialEstimate: 1/8h
+        caseimportance: high
+        caseposneg: positive
+        testtype: functional
```
valaparthvi: OPTIONAL : `caseimportance`, `caseposneg`, `testtype` already have their respective default values, you can remove them.

cfme/tests/automate/test_instance.py
```diff
@@ -354,3 +354,155 @@ def test_quota_source_value(request, entity, search, copy_quota_instance, generi
...
+                \nexit {process.upper()}
+                """
+    ), ["first", "second", "third"]))
+
+    fields = [{state1: {'value': f"METHOD::{child_method[0].name}"},
+              state2: {'value': f"METHOD::{child_method[1].name}"}},
+              {state1: {'value': f"METHOD::{child_method[2].name}"}}]
+
+    # Creating two child instances
+    child_inst = list(map(lambda field: setup[0].instances.create(
```
valaparthvi: `setup[0]` or `klass2`?

cfme/tests/automate/test_instance.py
ganeshhubale: @valaparthvi 
- So if you see here ```step 1 : Navigate to Automation > Automate > Simulation page and execute instance A1``` is already indicates that you need to only execute that instance(A1).
- So step 2 and 3 are for understanding to check logs. Hence keeping as it is.
- Actually your sentence does not satisfy test case. Because it doesn't mean we are executing methods where all methods are containing ```MIQ_STOP``` or ```MIQ_ABORT```

dajoRH:
I detected some fixture changes in commit 6ecf406394dae4d580eb2387d9c4f95f01adbba6

The *local* fixture **`setup`** is used in the following files:
- **cfme/tests/automate/test_instance.py**
    - *test_miq_stop_abort_with_state_machines*

PR #9154, opened 2019-08-02
[1LP][RFR] Avoid warning from relative imports

mshriver:
Remove relative imports to avoid warnings in py3 for resolving those imports.

I also happen to prefer explicit imports, fight me? Or just challenge this preference in a comment, and maybe we avoid these warnings another way if we want to allow them.

### DIFF
--- a/cfme/automate/dialogs/dialog_box.py
+++ b/cfme/automate/dialogs/dialog_box.py
@@ -2,9 +2,8 @@ import attr
 from cached_property import cached_property
 from navmazing import NavigateToAttribute
 
-from . import AddBoxView
-from . import BoxForm
-from .dialog_element import ElementCollection
+from cfme.automate.dialogs import AddBoxView
+from cfme.automate.dialogs import BoxForm
 from cfme.modeling.base import BaseCollection
 from cfme.modeling.base import BaseEntity
 from cfme.modeling.base import parent_of_type
@@ -29,6 +28,7 @@ class Box(BaseEntity):
     box_label = attr.ib()
     box_desc = attr.ib(default=None)
 
+    from cfme.automate.dialogs.dialog_element import ElementCollection
     _collections = {'elements': ElementCollection}
 
     @cached_property
@@ -41,7 +41,7 @@ class Box(BaseEntity):
 
     @property
     def tab(self):
-        from .dialog_tab import Tab
+        from cfme.automate.dialogs.dialog_tab import Tab
         return parent_of_type(self, Tab)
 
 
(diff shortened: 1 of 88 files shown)


lina-nikiforova:
+1 to explicit imports :)

cfme/control/__init__.py
```diff
@@ -1,4 +1,4 @@
-from . import explorer  # noqa: F401
-from . import import_export  # noqa: F401
-from . import log  # noqa: F401
-from . import simulation  # noqa: F401
+from cfme.control import explorer  # noqa: F401
```
valaparthvi: I don't understand the use of importing these. Why can't we just leave the `__init__.py` empty?

cfme/automate/dialogs/dialog_box.py
```diff
@@ -29,6 +28,7 @@ class Box(BaseEntity):
     box_label = attr.ib()
     box_desc = attr.ib(default=None)
 
+    from cfme.automate.dialogs.dialog_element import ElementCollection
```
valaparthvi: Out of curiosity, are we using all these local imports to avoid circular dependencies?

cfme/services/service_catalogs/__init__.py
```diff
@@ -99,6 +99,6 @@ def fill(self, fill_data):
         return was_change
 
 
-from . import ui, ssui  # NOQA last for import cycles
+from cfme.services.service_catalogs import ui, ssui  # NOQA last for import cycles
```
valaparthvi: Not separating these imports?

cfme/control/__init__.py
mshriver: This allows for import of these modules though `cfme.control`.  If you would like to remove them I would certainly entertain that PR, but you'll have to identify where these imports are used in other modules and update them accordingly. I would not like to include that change in this PR.

cfme/automate/dialogs/dialog_box.py
mshriver: The path to the import won't impact whether or not the import is circular, but _where_ they are imported will. There are a lot of cross dependencies in these dialog modules, which is why imports are placed within the classes/methods.

cfme/services/service_catalogs/__init__.py
mshriver: Leaving them with `# NOQA` as a 1-line import for importscan.

kedark3:
@mshriver I am little too late to review this, but yeah I like this explicit imports. +1

PR #9157, opened 2019-08-05
[1LP][RFR] removed workaround for BZ 1703744

lina-nikiforova:
## Purpose or Intent
The BZ is verified and the workaround is no longer needed. However, I'd like to keep the BZ number nearby as it helps with finding qe_test_coverage - I'm not sure which test exactly automates this check.


john-dupuy:
LGTM, but can you please fix the travis failure? 

`cfme/common/provider_views.py:16:1: F401 'cfme.utils.blockers.BZ' imported but unused
`

PR #9158, opened 2019-08-05
[1LP][RFR] Fixed quota testing against reconfigure vm

ganeshhubale:
## Purpose or Intent
- This PR improves test case of testing root tenant - ```My Company``` quota against reconfiguring vm. 
- This test case now checks if quota is exceeded after vm reconfiguration above assigned qutoa.
### PRT Run
{{ pytest: cfme/tests/infrastructure/test_tenant_quota.py -k 'test_tenant_quota_vm_reconfigure' --use-provider=complete --long-running -qsvvv --use-template-cache }}


ganeshhubale:
@digitronik Actually there no need to check between ```original``` and ```new config```. I enhanced this test case to check if quota is exceeding after reconfiguring vm over assigned quota.

PR #9169, opened 2019-08-07
[RFR] fixed create_volume function for 5.11

lina-nikiforova:
## Purpose or Intent
Fixes 
```
E           assert exact match of message: success: Cloud Volume "xwKtzvVAhC" created. 
E            Available messages: []

/var/ci/cfme_venv3/lib64/python3.7/site-packages/widgetastic_patternfly/__init__.py:315: AssertionError
AssertionError
b'assert exact match of message: success: Cloud Volume "xwKtzvVAhC" created. \n Available messages: []'
```

### PRT Run
{{ pytest: -v cfme/tests/storage/test_volume.py::test_storage_volume_edit_tag --use-template-cache --use-provider=complete --long-running }}

PR #9171, opened 2019-08-07
[RFR] fixed all_options in tagging dropdown

lina-nikiforova:
## Purpose or Intent
Fixes
```
>           item = item.split(" *")[0]
E           TypeError: a bytes-like object is required, not 'str'

widgetastic_manageiq/__init__.py:2813: TypeError
TypeError
b"a bytes-like object is required, not 'str'"
```
I guess `encode("utf-8")` no longer needed with python3 and it works now without it.

### PRT Run

{{ pytest: -v cfme/tests/storage/test_volume.py::test_storage_volume_edit_tag --use-provider=complete --long-running }}

### DIFF
--- a/widgetastic_manageiq/__init__.py
+++ b/widgetastic_manageiq/__init__.py
@@ -2831,9 +2831,7 @@ class ReactSelect(Widget, ClickableMixin):
     def all_options(self):
         # need to open the dropdown to see all available options
         self.open()
-        options = [
-            self.browser.text(e).encode("utf-8") for e in self.browser.elements(self.ALL_OPTIONS)
-        ]
+        options = [self.browser.text(e) for e in self.browser.elements(self.ALL_OPTIONS)]
         self.close()
         return options
 


lina-nikiforova:
PRT failures:

```
E           assert exact match of message: success: Cloud Volume "bCeQVpyWKx" created. 
E            Available messages: []
```
is fixed in #9169

```
E                       widgetastic_patternfly.SelectItemNotFound: Could not find 'ec2-west EBS Storage Manager' in BootstrapSelect(locator='.//div[contains(@class, "bootstrap-select")]/select[normalize-space(@name)="storage_manager_id"]/..')
E                       These options are present: '<Choose>, RHOS13 Cinder Manager'
```
is the failure due to the order of execution. First test stopped at `Add New Cloud Volume` form and when the second test started it was already "here" - but the option in the dropdown is the residual from previous test. Simple refresh would solve the problem.
Anyway, it's not related to the changes in this PR.

PR #9172, opened 2019-08-07
[RFR] Fix broken readthedocs build

john-dupuy:
Our docs build has been broken for sometime, and this PR's goal is to fix it. With this PR, I am:

1. updating Sphinx to the latest version
2. Removing mocked imports from `frozen_docs.py3.txt`, the main one is `pycurl`
3. Making small config changes to `.readthedocs.yml`
4. Small changes to `docs/conf.py` to fix the build

I've setup a debug docs build that tracks this branch, you can view the builds here:
https://readthedocs.org/projects/integration-tests-debug/
And the built docs here:
https://integration-tests-debug.readthedocs.io/en/latest/


cfme/utils/apidoc.py
```diff
@@ -39,11 +40,13 @@ def purge_module_apidoc(sphinx, exception):
         return
 
     try:
-        sphinx.info(bold('cleaning autogenerated docs... '), nonl=True)
+        # sphinx.info(bold('cleaning autogenerated docs... '), nonl=True)
```
jawatts: should these be removed?

john-dupuy: Yeah, I think they should just be removed, with the change in Sphinx version, they result in an `AttributeError`. But I'm not 100% sure if I should change them to print statements or just remove them altogether

cfme/utils/apidoc.py
```diff
@@ -3,17 +3,14 @@
 To enable the optional cleaning, set ``clean_autogenerated_docs`` to ``True`` in docs/conf.py
 
 """
-import subprocess
-
-from sphinx.util.console import bold
-from sphinx.util.console import red
+from sphinx.ext import apidoc
```
mshriver: This import is failing in travis:
``` from sphinx.ext import apidoc
663ImportError: cannot import name 'apidoc' from 'sphinx.ext' ```

cfme/utils/apidoc.py
john-dupuy: I changed them to logger calls, as that is what `Sphinx` 2.1.2 uses

PR #9174, opened 2019-08-08
[1LP][RFR] ignore test_sdn_balancers_tagvis for 5.11

lina-nikiforova:
## Purpose or Intent
Load balancers are no longer there

### DIFF
--- a/cfme/tests/networks/test_sdn_balancers.py
+++ b/cfme/tests/networks/test_sdn_balancers.py
@@ -11,7 +11,8 @@ from cfme.utils.appliance.implementations.ui import navigate_to
 
 pytestmark = [
     pytest.mark.usefixtures('setup_provider'),
-    pytest.mark.provider([AzureProvider, EC2Provider, GCEProvider], scope='module')
+    pytest.mark.provider([AzureProvider, EC2Provider, GCEProvider], scope='module'),
+    pytest.mark.ignore_stream("5.11", "upstream")  # BZ 1672949 (RFE)
 ]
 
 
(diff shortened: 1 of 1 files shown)


cfme/tests/networks/test_sdn_balancers.py
```diff
@@ -68,6 +68,7 @@ def test_sdn_balancers_detail(provider, network_prov_with_load_balancers):
 
 
 # only one provider is needed for that test, used Azure as it has balancers
+@pytest.mark.ignore_stream("5.11", "upstream")  # BZ 1672949 (RFE)
```
kedark3: Is that only test you need to ignore or all the `test_sdn_balancers`?

lina-nikiforova: @kedark3 good catch! it's the whole module that needs to be blocked

lina-nikiforova: updated

kedark3: Awesome I approve.

lina-nikiforova:
@digitronik added. Not sure how it is used in Polarion though. We can continue the discussion offline.

sbulage:
> @digitronik added. Not sure how it is used in Polarion though. We can continue the discussion offline.

Nice question @lina-nikiforova, by adding `endsin` in the test case will ensure any user that this test case/feature not supported in upcoming CFME versions. We were using `endsin` to generate Test Runs for respective CFME version.

PR #9183, opened 2019-08-09
[1LP][RFR][NOTEST] Deal with last two jhenner manual tests to sort.

jaryn:
The test_optimize_memory_usage_by_making_object_in_hash is being deleted
here as I (jhenner) have no clue what is meant to be tested.

The test_puma_server is obsolete as per BZ 1752509

### DIFF
--- a/cfme/tests/test_appliance.py
+++ b/cfme/tests/test_appliance.py
@@ -373,6 +373,52 @@ def test_appliance_exec_scripts():
     pass
 
 
+@pytest.mark.manual
+@pytest.mark.tier(1)
+def test_puma_server():
+    """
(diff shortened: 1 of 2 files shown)


cfme/tests/test_appliance.py
```diff
@@ -373,6 +373,44 @@ def test_appliance_exec_scripts():
     pass
 
 
+@pytest.mark.manual
+@pytest.mark.tier(1)
+def test_puma_server():
+    """
+    From a bugPuma, by default is a threaded web server, which means each
```
john-dupuy: **Required** can you move these steps into `testSteps` and `expectedResults` in the polarion docblock?

PR #9189, opened 2019-08-12
[1LP][RFR] New Test: Testing automate method call in state schema

ganeshhubale:
## Purpose or Intent
- Testing automate method call via state schema type

### PRT Run
{{ pytest: cfme/tests/automate/test_method.py  -k 'test_automate_state_method' --use-template-cache -qsvv }}

### DIFF
--- a/cfme/tests/automate/test_method.py
+++ b/cfme/tests/automate/test_method.py
@@ -496,3 +496,70 @@ def test_embedded_method_selection():
         1718495
     """
     pass
+
+
+@pytest.mark.tier(1)
+def test_automate_state_method(klass):
+    """
+    You can pass methods as states compared to the old method of passing
+    instances which had to be located in different classes. You use the
+    METHOD:: prefix
+
+    Polarion:
+        assignee: ghubale
+        casecomponent: Automate
+        caseimportance: medium
+        initialEstimate: 1/4h
+        tags: automate
+        startsin: 5.6
+        testSteps:
+            1. Create an automate class that has one state.
+            2. Create a method in the class, make the method output
+               something recognizable in the logs
+            3. Create an instance inside the class, and as a Value for the
+               state use: METHOD::method_name where method_name is the name
+               of the method you created
+            4. Run a simulation, use Request / Call_Instance to call your
+               state machine instance
+        expectedResults:
+            1. Class created
+            2. Method created
+            3. Instance created
+            4. The method got called, detectable by grepping logs
+    """
+    state = fauxfactory.gen_alpha()
+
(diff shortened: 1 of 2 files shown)


cfme/tests/automate/test_method.py
```diff
@@ -496,3 +496,70 @@ def test_embedded_method_selection():
...
+            1. Class created
+            2. Method created
+            3. Instance created
+            4. The method got called, detectable by grepping logs
+    """
+    state = fauxfactory.gen_alpha()
+
+    klass.schema.add_fields({'name': state, 'type': 'State'})
+
+    method = klass.methods.create(
```
jaryn: you probably should remove all the items you are creating using finalizers

cfme/tests/automate/test_method.py
```diff
@@ -496,3 +496,70 @@ def test_embedded_method_selection():
...
+    )
+
+    instance = klass.instances.create(
+        name=fauxfactory.gen_alphanumeric(),
+        display_name=fauxfactory.gen_alphanumeric(),
+        description=fauxfactory.gen_alphanumeric(),
+        fields={state: {"value": f"METHOD::{method.name}"}}
+    )
+
+    result = LogValidator(
```
jaryn: This variable is misleadingly named.

cfme/tests/automate/test_method.py
ganeshhubale: yes @jaryn I agree with you. 
- But we don't need cleanup for child items of ```domain, namespaces and klass``` because these are global fixtures and they have there own cleanup.
- Actually, **instance and methods are child items** of ```klass``` and if we delete klass then those items gets deleted.
Hence keeping as it is.

cfme/tests/automate/test_method.py
ganeshhubale: As we are finally checking expected result using ```LogValidator```. Hence I am using ```result``` variable.
I am not getting how ```with``` will change approach.
Please explain more.

mshriver: @ganeshhubale , @jaryn  is referring to the LogValidator context manager, defined here:
https://github.com/ManageIQ/integration_tests/blob/master/cfme/utils/log_validator.py#L132

I would also suggest using it for testing, though @jaryn could improve adoption by adding an example to the docblock for `LogValidator` or the `waiting` method itself.

See test_appliance_console for an existing use of the context manager:
https://github.com/ManageIQ/integration_tests/blob/master/cfme/tests/cli/test_appliance_console.py#L312

mshriver: Leaving this as a TODO/enhancement, as the PR looks solid as is.

PR #9197, opened 2019-08-13
[1LP][RFR] fixed select in adding new orchestration template

lina-nikiforova:
## Purpose or Intent
Fixes 
```
E           selenium.common.exceptions.NoSuchElementException: Message: Could not find an element Locator(by='xpath', locator='.//div[contains(@class, "bootstrap-select")]/button[normalize-space(@data-id)="type"]/..')

/var/ci/cfme_venv3/lib64/python3.7/site-packages/widgetastic/browser.py:347: NoSuchElementException
NoSuchElementException
b'Message: Could not find an element Locator(by=\'xpath\', locator=\'.//div[contains(@class, "bootstrap-select")]/button[normalize-space(@data-id)="type"]/..\')\n'
```
~~in 5.11 but there's another frontend change that will fail the PRT.~~

I also had to create a new widget `ReactCodemirror` which is incomplete. I'd also like to implement `read` method there but so far tests pass and I can do it in a separate PR.

### PRT Run

{{ pytest: -v cfme/tests/services/test_orchestration_template.py::test_tag_orchestration_template --long-running }}

### DIFF
--- a/cfme/services/catalogs/orchestration_template.py
+++ b/cfme/services/catalogs/orchestration_template.py
@@ -18,7 +18,11 @@ from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.appliance.implementations.ui import navigator
 from cfme.utils.pretty import Pretty
 from cfme.utils.update import Updateable
+from cfme.utils.version import LOWEST
+from cfme.utils.version import VersionPicker
 from widgetastic_manageiq import PaginationPane
+from widgetastic_manageiq import ReactCodeMirror
+from widgetastic_manageiq import ReactSelect
 from widgetastic_manageiq import ScriptBox
 from widgetastic_manageiq import SummaryTable
 from widgetastic_manageiq import Table
@@ -49,11 +53,17 @@ class CopyTemplateForm(ServicesCatalogView):
 
 class TemplateForm(ServicesCatalogView):
     title = Text('#explorer_title_text')
-    template_type = BootstrapSelect("type")
+    template_type = VersionPicker({
+        "5.11": ReactSelect("type"),
+        LOWEST: BootstrapSelect("type")
+    })
     name = Input(name='name')
     description = Input(name="description")
     draft = Checkbox(name='draft')
-    content = ScriptBox(locator="//pre[@class=' CodeMirror-line ']/span")
+    content = VersionPicker({
+        "5.11": ReactCodeMirror(),
+        LOWEST: ScriptBox(locator="//pre[@class=' CodeMirror-line ']/span")
+    })
 
 
 class AddTemplateView(TemplateForm):
--- a/widgetastic_manageiq/__init__.py
+++ b/widgetastic_manageiq/__init__.py
@@ -1022,6 +1022,43 @@ class ScriptBox(Widget):
         )
 
 
+class ReactCodeMirror(Widget):
+    """Looks the same as ScriptBox but it is changed in 5.11
+    https://github.com/JedWatson/react-codemirror"""
+
+    ROOT = "//div[contains(@class,'miq-codemirror')]//textarea"
+    PARENT_DIV = "./.."  # we need to change visibility of this one
+    CLICK_DIV = "./../.."  # the widget has very nested divs,we need to click on that one ¯\_(ツ)_/¯
+
+    def __init__(self, parent, logger=None):
+        Widget.__init__(self, parent, logger=logger)
+
+    @property
+    def element(self):
+        return self.browser.element(self)
+
+    @property
+    def is_editable(self):
+        return self.element.get_attribute("contentEditable") == "true"
+
+    def make_editable(self):
+        if not self.is_editable:
+            self.browser.set_attribute("contentEditable", "true", self)
+
+    def fill(self, value):
+        # TODO(anikifor): remove the 2 workarounds below as soon as BZ 1740753 is verified
+        self.make_editable()
+        self.browser.set_attribute("style", "overflow: visible", self.PARENT_DIV)
+        self.browser.element(self.CLICK_DIV).click()
+        self.element.send_keys(value)
+        return True
+
+    def read(self):
+        # TODO: implement, the text is stored separately on each line in the page
+        # locator for read is somewhere there "//pre[@class=' CodeMirror-line ']/span"
+        raise NotImplementedError("Read method is not implemented yet")
+
+
 class SettingsGroupSubmenu(NavDropdown):
     """
     Widget for the Group 'Dropdown' menu that allows a user assigned to multiple groups


widgetastic_manageiq/__init__.py
```diff
@@ -1022,6 +1022,43 @@ def workaround_save_issue(self):
         )
 
 
+class ReactCodemirror(Widget):
```
lina-nikiforova: https://bugzilla.redhat.com/show_bug.cgi?id=1740753

lina-nikiforova:
I have the tests passed locally! The read method also requires some hacks and it will be easier to fix those test first and then deal with the read.

widgetastic_manageiq/__init__.py
```diff
@@ -1022,6 +1022,49 @@ def workaround_save_issue(self):
...
+    https://github.com/JedWatson/react-codemirror"""
+
+    def __init__(self, parent, locator=None, read_locator=None, logger=None):
+        Widget.__init__(self, parent, logger=logger)
+        self.locator = locator
+        self.parent_locator = f"{locator}/.."  # we need to change visibility of this one
+        self.click_locator = f"{self.parent_locator}/.."  # the widget has very nested divs
+        self.read_locator = read_locator
+
+    def __locator__(self):
```
digitronik: I would set `ROOT` which will auto resolve locator.

widgetastic_manageiq/__init__.py
```diff
@@ -1022,6 +1022,49 @@ def workaround_save_issue(self):
         )
 
 
+class ReactCodemirror(Widget):
+    """Looks the same as ScriptBox but it changede in 5.11
+    https://github.com/JedWatson/react-codemirror"""
+
+    def __init__(self, parent, locator=None, read_locator=None, logger=None):
```
digitronik: Q: It is not possible to take as single `locator` which point div of `CodeMirror` widget? and with help of this we can create other required locators like `read_locator`?

lina-nikiforova: hey, I think I will remove read_locator as it's more complicated as I originally thought.

However, there's no way you can build it from the main locator

```
locator="//div[contains(@class,'miq-codemirror')]//textarea",
read_locator="//pre[@class=' CodeMirror-line ']/span"
```
as you see, they are not related,and you send_keys in one place but read from another

widgetastic_manageiq/__init__.py
lina-nikiforova: thanks, I'll do that

widgetastic_manageiq/__init__.py
```diff
@@ -1022,6 +1022,48 @@ def workaround_save_issue(self):
...
+    https://github.com/JedWatson/react-codemirror"""
+
+    ROOT = "//div[contains(@class,'miq-codemirror')]//textarea"
+    PARENT_DIV = "./.."  # we need to change visibility of this one
+    CLICK_DIV = "./../.."  # the widget has very nested divs,we need to click on that one ¯\_(ツ)_/¯
+
+    def __init__(self, parent, logger=None):
+        Widget.__init__(self, parent, logger=logger)
+
+    def __locator__(self):
```
john-dupuy: @lina-nikiforova No need to define `__locator__` if you have defined a class constant of `ROOT` it will auto-resolve with the `Widget` class.

https://github.com/RedHatQE/widgetastic.core/blob/2bca6bfa4dfc5a098218ec0d9b251e1b3600bc92/src/widgetastic/widget/base.py#L648-L651

widgetastic_manageiq/__init__.py
```diff
@@ -1022,6 +1022,48 @@ def workaround_save_issue(self):
...
+    def make_editable(self):
+        if not self.is_editable:
+            self.browser.set_attribute("contentEditable", "true", self)
+
+    def fill(self, value):
+        self.make_editable()
+        self.browser.set_attribute("style", "overflow: visible", self.PARENT_DIV)
+        self.browser.element(self.CLICK_DIV).click()
+        self.element.send_keys(value)
+        return True
```
john-dupuy: Any way for the fill to fail? I.e. if nothing changes? Just wondering if this should ever return `False`.

widgetastic_manageiq/__init__.py
```diff
@@ -1022,6 +1022,48 @@ def workaround_save_issue(self):
...
+
+    def read(self):
+        # TODO: implement, the text is stored separately on each line in the page
+        # locator for read is somewhere there "//pre[@class=' CodeMirror-line ']/span"
+        pass
```
john-dupuy: Thoughts on raising `NotImplemented`?

widgetastic_manageiq/__init__.py
lina-nikiforova: @john-dupuy awesome, thanks! I didn't find it :)

widgetastic_manageiq/__init__.py
lina-nikiforova: hm, I am thinking if I could do that when implementing read method Then it will be possible to check that. But in my work on that I've never seen it sending the keys and they are not appearing. It'd fall with "Element no found" or "not intractable" or another selenium exception. I guess it's safe to leave it like that.

widgetastic_manageiq/__init__.py
lina-nikiforova: thanks!

cfme/services/catalogs/orchestration_template.py
```diff
@@ -18,7 +19,10 @@
 from cfme.utils.appliance.implementations.ui import navigator
 from cfme.utils.pretty import Pretty
 from cfme.utils.update import Updateable
+from cfme.utils.version import VersionPicker
 from widgetastic_manageiq import PaginationPane
+from widgetastic_manageiq import ReactCodemirror
```
izapolsk: please use proper camel case naming

lina-nikiforova: @izapolsk  hm, this should be correct according to the name of the original component https://github.com/JedWatson/react-codemirror
and there it says `Codemirror Component for React.js`

lina-nikiforova: well, it's actually inconsistent there

lina-nikiforova: changed :)

cfme/services/catalogs/orchestration_template.py
```diff
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import attr
+from miq_version import LOWEST
```
sbulage: **Suggestion:** 
To have uniformity, I would suggest to use `from cfme.utils.version import LOWEST`. I can see in the many test files we have this kinda implementation.

widgetastic_manageiq/__init__.py
```diff
@@ -1022,6 +1022,45 @@ def workaround_save_issue(self):
...
+    def __init__(self, parent, logger=None):
+        Widget.__init__(self, parent, logger=logger)
+
+    @property
+    def element(self):
+        return self.browser.element(self)
+
+    @property
+    def is_editable(self):
+        if self.element.get_attribute("contentEditable") == "true":
```
izapolsk: ``` return self.element.get_attribute("contentEditable") == "true" ```
would simplify this

widgetastic_manageiq/__init__.py
```diff
@@ -1022,6 +1022,45 @@ def workaround_save_issue(self):
         )
 
 
+class ReactCodeMirror(Widget):
+    """Looks the same as ScriptBox but it is changed in 5.11
+    https://github.com/JedWatson/react-codemirror"""
+
+    ROOT = "//div[contains(@class,'miq-codemirror')]//textarea"
```
izapolsk: could there be several such widgets on page ?

widgetastic_manageiq/__init__.py
```diff
@@ -1022,6 +1022,45 @@ def workaround_save_issue(self):
...
+        return self.browser.element(self)
+
+    @property
+    def is_editable(self):
+        if self.element.get_attribute("contentEditable") == "true":
+            return True
+        else:
+            return False
+
+    def make_editable(self):
```
izapolsk: I'd transform this method with above property into self.editable getter/setter pattern.
you can find example here - https://github.com/ManageIQ/integration_tests/blob/master/cfme/utils/appliance/__init__.py#L477

so, it would look like

```
code_mirror.editable = True
assert code_mirror.editable , "code mirror editable"
code_mirror.editable = False
assert not code_mirror.editable , "code mirror isn't editable" 
```

widgetastic_manageiq/__init__.py
```diff
@@ -1022,6 +1022,45 @@ def workaround_save_issue(self):
...
+        if not self.is_editable:
+            self.browser.set_attribute("contentEditable", "true", self)
+
+    def fill(self, value):
+        self.make_editable()
+        self.browser.set_attribute("style", "overflow: visible", self.PARENT_DIV)
+        self.browser.element(self.CLICK_DIV).click()
+        self.element.send_keys(value)
```
izapolsk: what will happen with previous contents of this view ? 
are there any cases when we would need to clear/replace previous contents ?

izapolsk:
@lina-nikiforova ^^

widgetastic_manageiq/__init__.py
lina-nikiforova: I haven't seen such page. So, I saw this widget in orchestration template and I think in cloud-init. Maybe somewhere else but it was always only one field for a script.

widgetastic_manageiq/__init__.py
lina-nikiforova: When this will be merged, I can remove this completely https://github.com/ManageIQ/manageiq-ui-classic/pull/6030

I'd prefer to merge it as is and then I will do a follow-up with removing these "hacks"

lina-nikiforova: The BZ is in post https://bugzilla.redhat.com/show_bug.cgi?id=1740753

widgetastic_manageiq/__init__.py
lina-nikiforova: well, nothing :)  I also haven't found such cases but if there're, I may need to account for that as well

lina-nikiforova:
@izapolsk I pushed one change and for the other we can either wait for the new build and then I can remove it or merge it now and I will remove it once the devs' PR is in the build. I prefer second option as it will allow tests to run already.

widgetastic_manageiq/__init__.py
izapolsk: okay, please add appropriate comment then

lina-nikiforova: @izapolsk done. There's TODO now :)

PR #9201, opened 2019-08-14
[1LP][RFR]Fix SavedReportDetailsView is_displayed

valaparthvi:
## Purpose or Intent

- __Fixing__ `is_displayed` method of `SavedReportDetailsView`. There are many reports whose title changes after they're generated.
See this screenshot to understand better:
Screenshot from 2019-08-14 14-39-10

Here `view.title.text` is `Saved Report "Processes for prod VMs sorted by CPU Time - Wed, 14 Aug 2019 07:54:53 +0000"`
but `expected_title` is `Saved Report "Processes for prod VMs sort by CPU Time - Wed, 14 Aug 2019 07:54:53 +0000"`

This is not a bug. There are other reports which have different name than it's `menu_name`, so instead of exactly matching the titles, we only check if `saved_report.queued_datetime_in_title` is present in `view.title.text`.

### PRT Run
{{ pytest: cfme/tests/intelligence/reports/test_generate_report.py -vvv }}

### DIFF
--- a/cfme/intelligence/reports/reports.py
+++ b/cfme/intelligence/reports/reports.py
@@ -247,15 +247,11 @@ class SavedReportDetailsView(CloudIntelReportsView):
 
     @property
     def is_displayed(self):
-        expected_title = 'Saved Report "{} - {}"'.format(
-            self.context["object"].report.title or self.context['object'].report.menu_name,
-            self.context["object"].queued_datetime_in_title
-        )
         return (
-            self.in_intel_reports and
-            self.reports.is_opened and
-            self.reports.tree.currently_selected == self.context["object"].tree_path and
-            self.title.text == expected_title
+            self.in_intel_reports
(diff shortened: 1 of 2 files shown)


cfme/intelligence/reports/reports.py
```diff
@@ -247,15 +247,11 @@ def all(cls, browser):
...
-        )
         return (
-            self.in_intel_reports and
-            self.reports.is_opened and
-            self.reports.tree.currently_selected == self.context["object"].tree_path and
-            self.title.text == expected_title
+            self.in_intel_reports
+            and self.reports.is_opened
+            and self.reports.tree.currently_selected == self.context["object"].tree_path
+            and self.context["object"].queued_datetime_in_title in self.title.text
```
digitronik: aaaaaaahh format variation :+1:

PR #9203, opened 2019-08-14
[1LP][RFR] Embedded Ansible 511 changes

sbulage:
## Purpose or Intent

- [x] Fixed Methods in appliance object.
- [x] Fixed CRUD tests for 5.11

### PRT Run

{{ pytest: cfme/tests/ansible/test_embedded_ansible_crud.py -svvvvv --long-running }}

### DIFF
--- a/cfme/fixtures/ansible_fixtures.py
+++ b/cfme/fixtures/ansible_fixtures.py
@@ -13,7 +13,8 @@ from cfme.utils.wait import wait_for
 @pytest.fixture(scope="module")
 def wait_for_ansible(appliance):
     appliance.server.settings.enable_server_roles("embedded_ansible")
-    appliance.wait_for_embedded_ansible()
+    if appliance.version < "5.11":
+        appliance.wait_for_embedded_ansible()
     yield
     appliance.server.settings.disable_server_roles("embedded_ansible")
 
--- a/cfme/fixtures/ansible_fixtures.py
+++ b/cfme/fixtures/ansible_fixtures.py
@@ -13,8 +13,7 @@ from cfme.utils.wait import wait_for
 @pytest.fixture(scope="module")
 def wait_for_ansible(appliance):
     appliance.server.settings.enable_server_roles("embedded_ansible")
-    if appliance.version < "5.11":
-        appliance.wait_for_embedded_ansible()
+    appliance.wait_for_embedded_ansible()
     yield
     appliance.server.settings.disable_server_roles("embedded_ansible")
 
--- a/cfme/tests/ansible/test_embedded_ansible_crud.py
+++ b/cfme/tests/ansible/test_embedded_ansible_crud.py
@@ -4,10 +4,7 @@ from cfme import test_requirements
 from cfme.utils.log import logger
 from cfme.utils.wait import wait_for
 
-pytestmark = [
-    pytest.mark.ignore_stream("upstream"),
-    test_requirements.ansible
-]
+pytestmark = [pytest.mark.ignore_stream("upstream"), test_requirements.ansible]
 
 
 @pytest.fixture(scope="function")
@@ -30,15 +27,16 @@ def test_embedded_ansible_enable(embedded_appliance):
         tags: ansible_embed
     """
     assert wait_for(lambda: embedded_appliance.is_embedded_ansible_running, num_sec=30)
-    assert wait_for(lambda: embedded_appliance.supervisord.is_active, num_sec=30)
-    assert wait_for(lambda: embedded_appliance.rabbitmq_server.running, num_sec=30)
-    assert wait_for(lambda: embedded_appliance.nginx.running, num_sec=30)
-    endpoint = "api" if embedded_appliance.is_pod else "ansibleapi"
+    if embedded_appliance.version < "5.11":
+        assert wait_for(lambda: embedded_appliance.supervisord.is_active, num_sec=30)
+        assert wait_for(lambda: embedded_appliance.rabbitmq_server.running, num_sec=30)
+        assert wait_for(lambda: embedded_appliance.nginx.running, num_sec=30)
+        endpoint = "api" if embedded_appliance.is_pod else "ansibleapi"
 
-    assert embedded_appliance.ssh_client.run_command(
-        'curl -kL https://localhost/{endp} | grep "AWX REST API"'.format(endp=endpoint),
-        container=embedded_appliance.ansible_pod_name,
-    )
+        assert embedded_appliance.ssh_client.run_command(
+            'curl -kL https://localhost/{endp} | grep "AWX REST API"'.format(endp=endpoint),
+            container=embedded_appliance.ansible_pod_name,
+        )
 
 
 @pytest.mark.tier(3)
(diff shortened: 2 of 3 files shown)


cfme/utils/appliance/__init__.py
```diff
@@ -1819,7 +1819,8 @@ def is_embedded_ansible_role_enabled(self):
 
     @property
     def is_embedded_ansible_running(self):
-        return self.is_embedded_ansible_role_enabled and self.supervisord.running
+        supervisord = self.supervisord.running if self.version < '5.11' else True
+        return self.is_embedded_ansible_role_enabled and supervisord
```
jaryn:
```suggestion
        return self.is_embedded_ansible_role_enabled and (self.version < '5.11' == self.supervisord.running)
```

cfme/utils/appliance/__init__.py
```diff
@@ -2040,7 +2041,7 @@ def server_roles(self, roles):
         self.update_advanced_settings({'server': server_data})
         timeout = 600 if enabling_ansible else 300
         wait_for(lambda: self.server_roles == roles, num_sec=timeout, delay=15)
-        if enabling_ansible:
+        if enabling_ansible and self.version < '5.11':
```
mshriver: I don't think these version checks are necessary. You can call `wait_for_embedded_ansible` on either appliance version, and the changes to `is_embedded_ansible_running` will handle the version changes.

cfme/fixtures/ansible_fixtures.py
```diff
@@ -13,7 +13,8 @@
 @pytest.fixture(scope="module")
 def wait_for_ansible(appliance):
     appliance.server.settings.enable_server_roles("embedded_ansible")
-    appliance.wait_for_embedded_ansible()
```
jaryn: Assuming we have fixed the `is_embedded_ansible_running`, then I don't see the reason why to make the `appliance.wait_for_embedded_ansible()` called only when version <5.11

cfme/utils/appliance/__init__.py
sbulage: @jaryn there is no `supervisord` service to check in terms of `appliance` `version` greater than or equal to 5.11 would fail and return `False`.

cfme/utils/appliance/__init__.py
sbulage: I will update `wait_for_embedded_ansible` change and remove other unnecessary changes.

cfme/fixtures/ansible_fixtures.py
mshriver: This ties in with my comment here [0] - Agreed, put the version check within the method, and then consistently call it when needed and let it check various services based on appliance version.

[0] https://github.com/ManageIQ/integration_tests/pull/9203#discussion_r313868573

sbulage: @jaryn @mshriver Removed this check and added into method only as suggested :smile:

cfme/utils/appliance/__init__.py
jaryn: @sbulage You are right. But that doesn't mean my suggestion is wrong.

sbulage: @jaryn I am not saying it is wrong :smile:. I will update it.

cfme/tests/ansible/test_embedded_ansible_crud.py
```diff
@@ -30,15 +27,16 @@ def test_embedded_ansible_enable(embedded_appliance):
...
-    endpoint = "api" if embedded_appliance.is_pod else "ansibleapi"
-
-    assert embedded_appliance.ssh_client.run_command(
-        'curl -kL https://localhost/{endp} | grep "AWX REST API"'.format(endp=endpoint),
-        container=embedded_appliance.ansible_pod_name,
-    )
+    if embedded_appliance.version < "5.11":
+        assert wait_for(lambda: embedded_appliance.supervisord.is_active, num_sec=30)
+        assert wait_for(lambda: embedded_appliance.rabbitmq_server.running, num_sec=30)
+        assert wait_for(lambda: embedded_appliance.nginx.running, num_sec=30)
```
valaparthvi: OPTIONAL: Don't you need to add `delay` as well? The default value appears to be 1.

cfme/tests/ansible/test_embedded_ansible_crud.py
```diff
@@ -81,16 +85,26 @@ def test_embedded_ansible_event_catcher_process(embedded_appliance):
...
+    else:
+        rpm_check = embedded_appliance.ssh_client.run_command(
+            "rpm -qa | grep 'ansible-runner'"
+        ).output
 
-    for data in result.splitlines():
-        logger.info("Checking service/process %s started or not", data)
-        assert "started" in data
+        for data in rpm_check.splitlines():
+            logger.info("Checking %s is present or not", data)
```
valaparthvi: **Optional:** Using f-string instead of this formatting.

cfme/tests/ansible/test_embedded_ansible_crud.py
sbulage: @valaparthvi It is only checking `service` `status` and returning `True` or `False` So delay is not require there.

cfme/tests/ansible/test_embedded_ansible_crud.py
sbulage: Updated.

cfme/tests/ansible/test_embedded_ansible_crud.py
```diff
@@ -4,10 +4,7 @@
 from cfme.utils.log import logger
 from cfme.utils.wait import wait_for
 
-pytestmark = [
-    pytest.mark.ignore_stream("upstream"),
-    test_requirements.ansible
-]
+pytestmark = [pytest.mark.ignore_stream("upstream"), test_requirements.ansible]
```
izapolsk: please don't reformat code w/o special need

cfme/utils/appliance/__init__.py
```diff
@@ -1832,7 +1832,9 @@ def is_embedded_ansible_role_enabled(self):
 
     @property
     def is_embedded_ansible_running(self):
-        return self.is_embedded_ansible_role_enabled and self.supervisord.running
+        return self.is_embedded_ansible_role_enabled and (
+            (self.version < "5.11") == self.supervisord.running
```
izapolsk: this statement looks weird please change it to something more readable

cfme/tests/ansible/test_embedded_ansible_crud.py
sbulage: It happened due to Black changes. Changing it.

sbulage:
> @sbulage, could you please explain why all those checks aren't valid for 5.11 ?

@izapolsk In CF 5.11 Embedded Ansible uses `Ansible Runner` in the back-end and not dependent on any of the Service. like `supervisord` or `nginx` etc. So, I have skipped those checks for CF 5.11.

cfme/utils/appliance/__init__.py
sbulage: @izapolsk Updated with previous logic, which much readable than current.

izapolsk: thanks

izapolsk: this old "new" logic looks much better but please pay attention to variable name next time.
your 'supervisord' variable retains process state not process itself. variable names should clearly describe their contents. I'd call it `process_state` or `is_running` or whatever.

and If I had to implement this logic, I'd write it in the following way 

```
return self.is_embedded_ansible_role_enabled and (self.version < '5.11' and self.supervisord.running)
```

PR #9205, opened 2019-08-14
[WIPTEST] Add podified to stream_excluder

mshriver:
Allows for including podified in ignore_stream markers, instead of using lambda and uncollectif

### DIFF
--- a/cfme/tests/automate/custom_button/__init__.py
+++ b/cfme/tests/automate/custom_button/__init__.py
@@ -9,21 +9,6 @@ from widgetastic_patternfly import Button
 from widgetastic_patternfly import Dropdown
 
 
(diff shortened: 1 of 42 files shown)


dajoRH:
I detected some fixture changes in commit 9bf7e750a89ba7ea3b3a0b049eb58e48e4674959

The *local* fixture **`start_evmserverd_after_module`** is used in the following files:
- **cfme/tests/cli/test_evmserverd.py**

mshriver:
Good thought @john-dupuy, that could also cover marking tests to ignore dev appliances.

mshriver:
Re-opening, I repurposed this branch. @john-dupuy comment is correct, this would be better is a separate ignore marker.

PR #9221, opened 2019-08-19
[1LP][RFR] Use contextmanager to close sockets

mshriver:
Our socket closing is a bit sloppy, leaving warnings for open socket resources.


mshriver:
@john-dupuy correct, we should not be seeing warnings of type:

``` ResourceWarning: unclosed socket.socket fd=4,
    family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6,
    laddr=('1.1.1.1', 65535), raddr=('1.1.1.2', 80)```

PR #9223, opened 2019-08-20
[1LP][RFR] Don't loosen DB

jaryn:
The tests started loosening the DB in far past (commit 6695364f56).
As the commit message states, it was a temporary solution to some
problem. It seems to me we do not need it any more and we should
rather use the default db permissions to do real testing.

Also, an invalid BZ#1741481 was reported because we are modifying the permissions in such a way that it doesn't allow replication

The unmodified permissions on CFME 5.11 are:

            # TYPE  DATABASE    USER  ADDRESS METHOD
            local   all         all           peer map=usermap
            local   replication all           peer map=usermap
            hostssl all         all   all     md5
            host    replication all   all     md5

### DIFF
--- a/cfme/utils/appliance/db.py
+++ b/cfme/utils/appliance/db.py
@@ -1,5 +1,3 @@
-from textwrap import dedent
-
 import attr
 import fauxfactory
 from cached_property import cached_property
@@ -228,47 +226,6 @@ class ApplianceDB(AppliancePlugin):
             message='appliance db ready', delay=20, num_sec=1200)
 
         self.logger.info('DB setup complete')
-
-    def loosen_pgssl(self, with_ssl=False):
-        """Loosens postgres connections"""
-
-        self.logger.info('Loosening postgres permissions')
-
-        # Init SSH client
-        client = self.appliance.ssh_client
-
-        # set root password
-        cmd = "psql -d vmdb_production -c \"alter user {} with password '{}'\"".format(
-            conf.credentials['database']['username'], conf.credentials['database']['password']
-        )
-        client.run_command(cmd)
-
-        pg_prefix = self.pg_prefix
-        # back up pg_hba.conf
-        client.run_command(
-            'mv {pg_prefix}/var/lib/pgsql/data/pg_hba.conf '
-            '{pg_prefix}/var/lib/pgsql/data/pg_hba.conf.sav'.format(pg_prefix=pg_prefix))
-
-        if with_ssl:
-            ssl = 'hostssl all all all cert map=sslmap'
-        else:
-            ssl = ''
-
-        # rewrite pg_hba.conf
-        write_pg_hba = dedent("""\
-        cat > {pg_prefix}/var/lib/pgsql/data/pg_hba.conf <<EOF
-        local all postgres,root trust
-        host all all 0.0.0.0/0 md5
-        hostssl all all all md5
-        {ssl}
-        EOF
-        """.format(ssl=ssl, pg_prefix=pg_prefix))
-        client.run_command(write_pg_hba)
-        client.run_command("chown postgres:postgres "
-            "{pg_prefix}/var/lib/pgsql/data/pg_hba.conf".format(pg_prefix=pg_prefix))
-
-        # restart postgres
-        return self.appliance.db_service.restart()
 
     def _run_cmd_show_output(self, cmd):
         """
--- a/cfme/tests/infrastructure/test_scvmm_specific.py
+++ b/cfme/tests/infrastructure/test_scvmm_specific.py
@@ -214,7 +214,7 @@ def test_create_appliance_on_scvmm_using_the_vhd_image(scvmm_appliance):
     # use appliance to get the stream so this test only runs once per test run
     # this will always use the downstream stable build, not the version of the appliance
     # configure the appliance
-    scvmm_appliance.configure(loosen_psql=False, fix_ntp_clock=False)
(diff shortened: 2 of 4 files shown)


cfme/utils/appliance/db.py
```diff
@@ -229,47 +227,6 @@ def setup(self, **kwargs):
...
-    def loosen_pgssl(self, with_ssl=False):
-        """Loosens postgres connections"""
-
-        self.logger.info('Loosening postgres permissions')
-
-        # Init SSH client
-        client = self.appliance.ssh_client
-
-        # set root password
-        cmd = "psql -d vmdb_production -c \"alter user {} with password '{}'\"".format(
```
mshriver: I think this needs to stay - set the postgres DB password to what's in the credentials file. This credentials file will later be used to create connections to the DB, and if its non-standard that will no longer work.

cfme/utils/appliance/db.py
```diff
@@ -229,47 +227,6 @@ def setup(self, **kwargs):
...
-        if with_ssl:
-            ssl = 'hostssl all all all cert map=sslmap'
-        else:
-            ssl = ''
-
-        # rewrite pg_hba.conf
-        write_pg_hba = dedent("""\
-        cat > {pg_prefix}/var/lib/pgsql/data/pg_hba.conf <<EOF
-        local all postgres,root trust
-        host all all 0.0.0.0/0 md5
```
mshriver: Per your PR comment about the default configuration, not including this line will prevent remote hosts from being able to connect over non-SSL connections.   Does the DB client in the framework still work without this? Is it using ssl for its connections?

mshriver:
Please specify in the PR description a set of tests that demonstrates the effectiveness of this change.

Although the pre-configured sprout images won't have this applied, you can locate (or write) and specify a test for PRT that uses temp_appliance fixture to get an unconfigured appliance, then run configuration on it and execute some DB calls against it through the framework.

jaryn:
Closing this and starting: https://github.com/ManageIQ/integration_tests/pull/9260 as I have found that we are using sqlachemy and thus my guess is we probably do some connections from outside of the appliance.

jaryn:
@mshriver: In commit 2950ae4 I was changing paths that is used by the DB permissions "loosening" code. The path used on CFME 5.10 is wrong though and the DB "loosening" was not effective there (we didn't know about that as the command failed silently).

We were rewriting the pg_hba.conf in the "loosening" DB permissions step (in case 5.11)
in such a way that it results in actually disallowing the replication. The
vanilla state of the pg_hba.conf allows replication. Allowance of db replication is required for db backup opertaions to work. This made restore tests not working on 5.11 but working on 5.10.

From the fact that we haven't had the loosening in effect on 5.10 for some time, the question to ask is: *Do we really need to modify the pg_hba.conf at all? *

The answer is that it seems it is not needed. For example the test `cfme/tests/control/test_basic.py::test_alerts_monitor_overview_page`, athough the CFME 5.10 Configured image is using
```
[root@host-192-168-200-28 ~]# cat /var/opt/rh/rh-postgresql95/lib/pgsql/data/pg_hba.conf 
# TYPE  DATABASE    USER  ADDRESS METHOD
local   all         all           peer map=usermap
local   replication all           peer map=usermap
hostssl all         all   all     md5
host    replication all   all     md5
```
The test is using this:
```
> /home/jhenner/work/miq/dont_rewrite_pg_hba/cfme/utils/db.py(163)engine()
-> return create_engine(self.db_url, echo_pool=True)
(Pdb) self.db_url
'postgresql://root:***CENSORED***@10.8.198.27:5432/vmdb_production'
```

And it seems that it can connect as on ostříž, you can see it is currently passing.

cfme/utils/appliance/db.py
jaryn: Well the password is set when creating a db in `enable_internal` in `cfme/utils/appliance`

cfme/utils/appliance/db.py
jaryn: It seems it is not needed. I explained this the comment to this PR.

john-dupuy:
@jaryn I see that no tests ran against PRT, can you please select some representative test cases to run against?

jaryn:
> @jaryn I see that no tests ran against PRT, can you please select some representative test cases to run against?

I don't know what you mean by representative test cases. 

1) Loosening is done in templates preparation time. I do not think we have any test for the code preparing the template.

2) I know "loosening" the db as it is (if really in effect) harms the restore tests (on 5.11), thus in PR#8655 i have had to make a workaround that sets the db permissions back to what they are in vanilla images to make this PR green. And I was told to make a separate PR for this "loosening"-removing commit.

3) As i wrote in the https://github.com/ManageIQ/integration_tests/pull/9223#issuecomment-526183513, loosening the db privs is **NOT** currently in effect for 5.10 images, but your test `cfme/tests/control/test_basic.py::test_alerts_monitor_overview_page` is PASSing

john-dupuy:
@jaryn, we have some test cases which will configure an appliance, I believe there is one for Azure and Ec2? But, this test case `test_create_appliance_on_scvmm_using_the_vhd_image` will also configure the appliance.

jaryn:
> @jaryn, we have some test cases which will configure an appliance, I believe there is one for Azure and Ec2? But, this test case `test_create_appliance_on_scvmm_using_the_vhd_image` will also configure the appliance.

well that test has the db privs loosening disabled:
`    scvmm_appliance.configure(loosen_psql=False, fix_ntp_clock=False)`

PR #9225, opened 2019-08-20
[1LP][RFR] Add customer scenario marker/filter, move tier and requirement markers

mshriver:
1.  Change how the tier and requirement markers are registered
2.  Remove unused node annotation logic
3.  Update pytest-polarion-collect and other polarion packages
4. Add a marker/filter for customerscenario meta in polarion

Changes to test_advanced_search are to repair its metadata collection and to prevent exceptions from pytest-polarion-collect during collection.


conf/polarion_tools.yaml
```diff
@@ -66,6 +66,7 @@ docstrings:
         # default for manual mark is 'notautomated'
         # @pytest.mark.manual('manualonly') to set 'manualonly'
         linkedWorkItems: "@pytest.mark.requirements"
+        customerscenario: "@pytest.mark.requirements.customer_stories"
```
mshriver: Not possible to set meta with these markers, evaluating if its worth supporting this.  Could also be set via BZ marker if by resolving items attached to the BZ.

izapolsk:
it's really great PR !. I love it

PR #9233, opened 2019-08-22
[RFR] fixed test_provider_log_level

lina-nikiforova:
## Purpose or Intent
1. I wanted to rewrite this using LogValidator as it seems like a better approach.
2. The test itself needed a change as it's hard to get debug outputs for certain providers (see https://bugzilla.redhat.com/show_bug.cgi?id=1633656)

Notes:
I had a problem with this test, when the log level is changed to `warn`, for some time there're still `info` messages - not a lot of time, it's hard to reproduce manually. I came up with that `def _no_info()` that works but I'm open to other suggestions.

### PRT Run
{{ pytest: -v cfme/tests/configure/test_logs.py::test_provider_log_level --use-provider=complete }}

### DIFF
--- a/cfme/tests/configure/test_logs.py
+++ b/cfme/tests/configure/test_logs.py
@@ -6,7 +6,8 @@ from cfme.cloud.provider.azure import AzureProvider
 from cfme.cloud.provider.ec2 import EC2Provider
 from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.infrastructure.provider.scvmm import SCVMMProvider
-from cfme.utils.blockers import BZ
+from cfme.utils.log_validator import FailPatternMatchError
+from cfme.utils.log_validator import LogValidator
 from cfme.utils.wait import wait_for
 
 pytestmark = [
@@ -108,12 +109,6 @@ def test_provider_log_updated(appliance, provider, log_exists):
     assert log_before != log_after, "Log hashes are the same"
 
 
-@pytest.mark.meta(blockers=[BZ(1633656,
-                               unblock=lambda provider: provider.one_of(AzureProvider, EC2Provider),
-                               forced_streams=["5.10", "upstream"]
-                               )
-                            ]
-                  )
 def test_provider_log_level(appliance, provider, log_exists):
     """
     Tests that log level in advanced settings affects log files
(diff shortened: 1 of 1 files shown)


cfme/tests/configure/test_logs.py
```diff
@@ -130,20 +125,38 @@ def test_provider_log_level(appliance, provider, log_exists):
...
     """
     assert log_exists, "Log file {}.log doesn't exist".format(provider.log_name)
     log_level = appliance.server.advanced_settings['log']['level_{}'.format(provider.log_name)]
-    # set log level to debug
+    log = f'/var/www/miq/vmdb/log/{provider.log_name}.log'
+    # set log level to info
     wait_for(lambda: appliance.server.update_advanced_settings(
-        {'log': {'level_{}'.format(provider.log_name): 'debug'}}), timeout=300)
+        {'log': {'level_{}'.format(provider.log_name): 'info'}}), timeout=300)
+    lv_info = LogValidator(log, matched_patterns=['.*INFO.*'])
```
john-dupuy: @lina-nikiforova can you make use of the new contextmanager for the `LogValidator`? https://github.com/ManageIQ/integration_tests/blob/d76fe2e735c0b45c63985bed87644e482dc98e42/cfme/utils/log_validator.py#L132-L136

lina-nikiforova: @john-dupuy I tried that but it gave me some error `AttributeError: __enter__`

cfme/tests/configure/test_logs.py
```diff
@@ -130,20 +125,38 @@ def test_provider_log_level(appliance, provider, log_exists):
         initialEstimate: 1/4h
         casecomponent: Configuration
         testSteps:
-            1. Change log level to debug
+            1. Change log level to warn
```
john-dupuy: _Optional_ these `testSteps` do not reflect the test since first you change to `info` and then change to `warn`. But since it's clear in the actual test what's going on, I leave that as _optional_

cfme/tests/configure/test_logs.py
```diff
@@ -108,12 +109,6 @@ def test_provider_log_updated(appliance, provider, log_exists):
     assert log_before != log_after, "Log hashes are the same"
 
 
-@pytest.mark.meta(blockers=[BZ(1633656,
-                               unblock=lambda provider: provider.one_of(AzureProvider, EC2Provider),
-                               forced_streams=["5.10", "upstream"]
-                               )
-                            ]
-                  )
 def test_provider_log_level(appliance, provider, log_exists):
```
john-dupuy: Thoughts on parametrizing this test against log level?

cfme/tests/configure/test_logs.py
john-dupuy: For `info` log level, do we also want to ensure that no `DEBUG` messages are present?

cfme/tests/configure/test_logs.py
lina-nikiforova: Thanks, I'd better change it to be clear :)

cfme/tests/configure/test_logs.py
lina-nikiforova: not a good idea considering why I'm rewriting this test :) originally I was trying to see DEBUG output but it wasn't shown for some providers, which turns out to be not a bug https://bugzilla.redhat.com/show_bug.cgi?id=1633656

cfme/tests/configure/test_logs.py
lina-nikiforova: Good idea, @john-dupuy . I'm adding it.

lina-nikiforova: @john-dupuy good idea! added.

PR #9234, opened 2019-08-22
[RFR] Fixed vm reconfigure quota test case

ganeshhubale:
## Purpose or Intent
- This test case needs to updated because in this test case we are checked if quota is getting  exceeded or not while reconfiguring vm above quota
- Here vm is not getting reconfigured because of quota exceeded. So removing assertion.

### PRT Run
{{ pytest: cfme/tests/infrastructure/test_quota_tagging.py::test_quota_vm_reconfigure --use-template-cache -qsvvv --use-provider=complete --long-running }}

### DIFF
--- a/cfme/tests/infrastructure/test_quota_tagging.py
+++ b/cfme/tests/infrastructure/test_quota_tagging.py
@@ -318,8 +318,8 @@ def test_quota_vm_reconfigure(
         )
     provision_request = appliance.collections.requests.instantiate(request_description)
     provision_request.wait_for_request(method="ui")
-    assert small_vm.configuration != original_config
     assert provision_request.row.reason.text == "Quota Exceeded"
+    assert small_vm.configuration == original_config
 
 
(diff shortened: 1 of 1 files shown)


cfme/tests/infrastructure/test_quota_tagging.py
```diff
@@ -318,7 +317,6 @@ def test_quota_vm_reconfigure(
         )
     provision_request = appliance.collections.requests.instantiate(request_description)
     provision_request.wait_for_request(method="ui")
-    assert small_vm.configuration != original_config
```
sbulage: **Suggestion:** 
Instead of assertion removal, you can use like this:` assert small_vm.configuration == original_config`.

As per your test, after vm reconfigure, quota gets exceeds then you can check with the original configuration of `small_vm` suggested above

sbulage: :+1:

PR #9238, opened 2019-08-22
[1LP][RFR] Add test_infrastructure_hosts_refresh_multi to test_host.py

prichard77:
##  Purpose or Intent
- __Updating tests__ for FA infra_hosts to cover refreshing multiple hosts. This was initially a manual test, but now it is automated.

{{ pytest: --use-provider complete --long-running cfme/tests/infrastructure/test_host.py::test_infrastructure_hosts_refresh_multi }}

### DIFF
--- a/cfme/common/provider.py
+++ b/cfme/common/provider.py
@@ -628,7 +628,7 @@ class BaseProvider(Taggable, Updateable, Navigatable, BaseEntity, CustomButtonEv
                                check_existing=True, validate_inventory=True)
 
     @variable(alias='rest')
-    def is_refreshed(self, refresh_timer=None, refresh_delta=600):
+    def is_refreshed(self, refresh_timer=None, refresh_delta=600, force_refresh=True):
         if refresh_timer:
             if refresh_timer.is_it_time():
                 logger.info(' Time for a refresh!')
@@ -639,7 +639,8 @@ class BaseProvider(Taggable, Updateable, Navigatable, BaseEntity, CustomButtonEv
             return False
         td = self.appliance.utc_time() - rdate
         if td > datetime.timedelta(0, refresh_delta):
-            self.refresh_provider_relationships()
+            if force_refresh:
+                self.refresh_provider_relationships()
             return False
         else:
             return True
--- a/cfme/tests/infrastructure/test_host.py
+++ b/cfme/tests/infrastructure/test_host.py
@@ -3,11 +3,13 @@ import random
 import socket
 
 import pytest
+from wait_for import TimedOutError
 
 from cfme import test_requirements
 from cfme.base.credential import Credential
 from cfme.common.host_views import HostsEditView
 from cfme.common.provider_views import ProviderNodesView
+from cfme.fixtures.provider import setup_or_skip
 from cfme.infrastructure.provider import InfraProvider
 from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.infrastructure.provider.scvmm import SCVMMProvider
@@ -15,6 +17,7 @@ from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.markers.env_markers.provider import ONE
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.conf import credentials
+from cfme.utils.log_validator import LogValidator
 from cfme.utils.wait import wait_for
 
 pytestmark = [
@@ -331,3 +334,59 @@ def test_hosts_not_displayed_several_times(appliance, provider, setup_provider):
     provider.wait_for_delete()
     provider.create()
     assert host_count == navigate_to(appliance.collections.hosts, "All").paginator.items_amount
+
+
+@pytest.fixture
+def setup_provider_min_hosts(request, provider, min_hosts=2):
+    num_hosts = len(provider.data.get('hosts', {}))
+    if num_hosts < min_hosts:
+        pytest.skip(f'Not enough hosts({num_hosts}) to run test. Need at least {min_hosts}')
+    """Function-scoped fixture to set up a provider"""
+    return setup_or_skip(request, provider)
+
+
+@test_requirements.infra_hosts
+def test_infrastructure_hosts_refresh_multi(appliance, setup_provider_min_hosts, provider):
+    """
+    Polarion:
+        assignee: prichard
+        casecomponent: Infra
+        caseimportance: low
+        initialEstimate: 1/6h
+        testSteps:
+            1. Navigate to the Compute > Infrastructure > Providers view.
+            2. Click on a provider guadicon, and then the hosts link along the top row of the view.
+            3. Select all hosts (need at least 2 hosts) by checking the box in upper left of
+               quadicons.
+            4. Click "Refresh Relationships and Power States" under the Configuration
+               dropdowm, and then click "OK" when prompted.
+        expectedResults:
+            1. Providers view is displayed.
+            2. Hosts view is displayed.
+            3.
+            4. "Refresh initiated for X Hosts from the CFME Database" is displayed in green
+               banner where "X" is the number of selected hosts. Properties for each host are
+               refreshed. Making changes to test pre-commithooks
+    """
+    hosts_view = navigate_to(provider.collections.hosts, "All")
+    num_hosts = len(hosts_view.entities.get_all())
+    if num_hosts < 2:
+        pytest.skip('not enough hosts in appliance UI to run test')
+    evm_tail = LogValidator('/var/www/miq/vmdb/log/evm.log',
+                            matched_patterns=[f"'Refresh Provider' successfully initiated for "
+                                              f"{num_hosts} Hosts"],
+                            hostname=appliance.hostname)
+    evm_tail.start_monitoring()
+    for h in hosts_view.entities.get_all():
+        h.check()
+    hosts_view.toolbar.configuration.item_select('Refresh Relationships and Power States',
+                                                 handle_alert=True)
+    hosts_view.flash.assert_success_message(
+        f'Refresh initiated for {num_hosts} Hosts from the CFME Database'
+    )
+    try:
+        wait_for(provider.is_refreshed, func_kwargs={'force_refresh': False}, num_sec=300,
+                 delay=10)
+    except TimedOutError:
+        pytest.fail("Hosts were not refreshed within given time")
+    assert evm_tail.validate(wait="30s")


cfme/tests/infrastructure/test_host.py
```diff
@@ -312,3 +312,39 @@ def _finalize():
...
     )
+
+
+@test_requirements.general_ui
+def test_infrastructure_hosts_refresh_multi(
+    appliance, setup_provider, provider
+):
+    """
+    Polarion:
+        assignee: pvala
```
john-dupuy: Shouldn't you be the assignee for this test case?

cfme/tests/infrastructure/test_host.py
```diff
@@ -312,3 +312,39 @@ def _finalize():
...
+            3. Click "Refresh Relationships and Power States" under the Configuration
+               dropdowm, and then click "OK" when prompted.
+        expectedResults:
+            1. Hosts view is displayed.
+            2.
+            3. "Refresh initiated for X Hosts from the CFME Database" is displayed in green
+               banner where "X" is the number of selected hosts. Properties for each host are
+               refreshed.
+    """
+    if len(provider.data.get('hosts', {})) < 2:
```
john-dupuy: _Optional_ you can use `provider.hosts.all()` to get host objects. However, using data will be faster anyway.

cfme/tests/infrastructure/test_host.py
```diff
@@ -312,3 +312,39 @@ def _finalize():
...
+        pytest.skip('not enough hosts to run test')
+    hosts_view = navigate_to(appliance.collections.hosts, "All")
```
john-dupuy: _Optional_ I would just call this `view`

john-dupuy: Also, I would navigate to the hosts through the provider object and not the appliance object, since this test will be parametrized against different providers.

cfme/tests/infrastructure/test_host.py
```diff
@@ -312,3 +312,39 @@ def _finalize():
...
+
+    assert hosts_view.is_displayed
+    [h.check() for h in hosts_view.entities.get_all()]
```
john-dupuy: _Optional_ I'm not a fan of list comprehension without assignment, though more LOC, I think it is more readable to use a standard `for` loop.

cfme/tests/infrastructure/test_host.py
```diff
@@ -312,3 +312,39 @@ def _finalize():
...
+    num_checked_hosts = len(hosts_view.entities.get_all())
```
john-dupuy: I would store this as a variable at the start of the test to reduce redundancy, it should always be the same as the number of hosts correct? (Unless a provider happened to have an inane no. of hosts on it)

cfme/tests/infrastructure/test_host.py
```diff
@@ -312,3 +312,39 @@ def _finalize():
...
+    hosts_view.toolbar.configuration.item_select('Refresh Relationships and Power States',
+                                                 handle_alert=True)
+    hosts_view.flash.assert_success_message(f'Refresh initiated for {num_checked_hosts} Hosts from'
```
john-dupuy: _Optional_ formatting
```python,
    hosts_view.flash.assert_success_message(
        f'Refresh initiated for {num_checked_hosts} Hosts from the CFME Database'
    )
```

john-dupuy:
Also @prichard77 if you only want PRT to run this test you should specify that in the description block. I've updated your PR description for this PR.

cfme/tests/infrastructure/test_host.py
```diff
@@ -312,3 +312,41 @@ def _finalize():
             actual_state, power_state
         ),
     )
+
+
+@test_requirements.general_ui
```
lina-nikiforova: I believe it should be `infra_hosts` FA. General UI is a different FA and has a different owner.

cfme/tests/infrastructure/test_host.py
```diff
@@ -312,3 +312,41 @@ def _finalize():
...
+def test_infrastructure_hosts_refresh_multi(
+    appliance, setup_provider, provider
+):
```
lina-nikiforova: The test name and parameters fit into one line, please format in on one line.

cfme/tests/infrastructure/test_host.py
```diff
@@ -312,3 +312,41 @@ def _finalize():
...
+    """
+    Polarion:
+        assignee: prichard
+        casecomponent: Infra
+        caseimportance: low
+        initialEstimate: 1/3h
```
lina-nikiforova: Are you sure it take 20 minutes to refresh 2 hosts? Provider setup is not included in `initialEstimate`

cfme/tests/infrastructure/test_host.py
```diff
@@ -312,3 +312,41 @@ def _finalize():
...
+        testSteps:
+            1. Navigate to the Compute > Infrastructure > Hosts view.
+            2. Select at least 2 hosts (more if possible) by checking the box in upper left of
```
lina-nikiforova: Optional: in the test I see that all hosts are refreshed but as it's clear what's going on, I leave it as optional

cfme/tests/infrastructure/test_host.py
```diff
@@ -312,3 +312,41 @@ def _finalize():
...
+            2.
+            3. "Refresh initiated for X Hosts from the CFME Database" is displayed in green
+               banner where "X" is the number of selected hosts. Properties for each host are
+               refreshed.
+    """
+    num_hosts = len(provider.data.get('hosts', {}))
+    if num_hosts < 2:
+        pytest.skip('not enough hosts to run test')
+    hosts_view = navigate_to(provider.collections.hosts, "All")
+    assert hosts_view.is_displayed
```
lina-nikiforova: I don't think it's necessary here, if the navigation is not successful, it will fail on `is_displayed` anyway

cfme/tests/infrastructure/test_host.py
```diff
@@ -312,3 +312,41 @@ def _finalize():
...
+    for h in hosts_view.entities.get_all():
```
lina-nikiforova: I'm pretty sure we don't have that many hosts in any provider but what do you think about using `surf_pages=True`?

cfme/tests/infrastructure/test_host.py
```diff
@@ -312,3 +312,41 @@ def _finalize():
...
+        h.check()
+    hosts_view.toolbar.configuration.item_select('Refresh Relationships and Power States',
+                                                 handle_alert=True)
+    hosts_view.flash.assert_success_message(
+        f'Refresh initiated for {num_hosts} Hosts from the CFME Database'
```
lina-nikiforova: I'm a bit concerned that refresh is made based on the number of hosts in the UI but `num_hosts` is determined based on the yaml file. Can you please explain?

lina-nikiforova: @john-dupuy well, here my concern is that for some reason not all hosts are in yamls (e.g. somebody added some test host), that way this test will fail.

cfme/tests/infrastructure/test_host.py
```diff
@@ -312,3 +312,41 @@ def _finalize():
...
+    if num_hosts < 2:
+        pytest.skip('not enough hosts to run test')
+    for h in hosts_view.entities.get_all():
+        h.check()
+    hosts_view.toolbar.configuration.item_select('Refresh Relationships and Power States',
+                                                 handle_alert=True)
+    hosts_view.flash.assert_success_message(
+        f'Refresh initiated for {num_hosts} Hosts from the CFME Database'
+    )
+    assert wait_for(provider.is_refreshed, func_kwargs={"refresh_delta": 120}, num_sec=300,
```
izapolsk: you don't need assert here since it will raise Timedout exception

lina-nikiforova: @izapolsk is it necessary to wrap? I agree with you on assert, but if we do just `wait_for(provider.is_refreshed...)` and it fails, the test fails as well.

prichard77: So we are not expecting each test to contain an assert? I am not arguing this point, just getting clarification. I am accustomed to each test having one, and only one assert (sometimes it makes sense to have more than one, but only in special cases). If the team is OK with tests passing with no asserts if they don't explicitly fail, I will comply.

prichard77: What is specifically wrong with the assert? Are we saying that the assert may cause/allow an incorrect test result, or do we just not like the explicit assert here. I'd definately want to fix anything that cases bad test results, but I don't see anything wrong with an explicit assert. It is what I am checking. If the wait_for times out, the test will fail. Otherwise the view is displayed and the test passes (exactly what I want). 
I have not tested the failure case (this test passes on the current builds), but the positive case seems to work. Do I need to cause the failure to make sure it would cause a fail result?

izapolsk: @lina-nikiforova, there is a difference between Failed test and Error(Broken) test. I'm certain you understand the difference. From my point, yes, we need to wrap wait_for calls used to check test codition into try/expect and raise pytest.fail() in order to mark test as Failed not Broken.
btw, I don't say that we need to wrap all wait_for calls. I mean only wait_for calls in test which check some test condition.

@prichard77, let me explain what will happen there.
1. if condition is True in wait_for, it will pass True to your assert and test will be passed.

2. If condition in False in wait_for, wait_for will call that method until is reaches timeout value.
wait_for will raise TimedOutException in such case. So, your assert will never get False.
so, it's redundant. Btw, your test will be marked as Error(Broken) instead of Failed in such case.

So, what I think you need to do
```
try:
  wait_for(provider.is_refreshed, func_kwargs={"refresh_delta": 120}, num_sec=300, delay=30)
except TimedOutError:
   pytest.fail("Blahblah wasn't refreshed within given time")
```

lina-nikiforova: @izapolsk thanks for the explanation! I agree with this point.

prichard77: I agree with the need to use try/except to better represent timeout as a failure and not error. I still question what is wrong with an assert? What about:
```
try:
  assert wait_for(provider.is_refreshed, func_kwargs={"refresh_delta": 120}, num_sec=300, delay=30)
except TimedOutError:
   pytest.fail("Blahblah wasn't refreshed within given time")

Comment (izapolsk) [reply cfme/tests/infrastructure/test_host.py]: @prichard77, there is nothing wrong with assert. it will just never be used in this particular case. because False will never be passed to it. so, it doesn't matter if it present or absent.

**Comment by prichard77** [comment]
Provider.is_refreashed is not doing what I wanted. I need to pretty much rewrite this test.

**Comment by prichard77** [comment]
The provider.is_refreshed is not doing anything useful for this test. It will give false positive. I have removed this from the test. I added a todo to add better checks. I spoke with Mike and he suggested looking for a "last refreshed" date for each host, storing it,  and comparing it after the refresh. But let's get this reviewed with only the check for the success banner. I may also add a check using the logValidator and checking evm logs.

**Comment by dajoRH** [comment]
Lint report for commit c463f940c62c55d5d3cbe47ec5a7182e7da23296:

`cfme/tests/infrastructure/test_host.py`:
- :bangbang: Line 351:5: **E265** *block comment should start with '# '*
- :heavy_exclamation_mark: Line 351:98: **W292** *no newline at end of file*
- :heavy_exclamation_mark: Line 351:98: **W292** *no newline at end of file*
- :heavy_exclamation_mark: Line 351:98: **W292** *no newline at end of file*

**No commit flaws detected**.

*CFME QE Bot*

**Comment by izapolsk** [comment]
please make your test RFR after adding appropriate check.

File: cfme/tests/infrastructure/test_host.py [hunk 15]
```diff
@@ -312,3 +312,40 @@ def _finalize():
...
+            1. Providers view is displayed.
+            2. Hosts view is displayed.
+            3.
+            4. "Refresh initiated for X Hosts from the CFME Database" is displayed in green
+               banner where "X" is the number of selected hosts. Properties for each host are
+               refreshed.
+    """
+    hosts_view = navigate_to(provider.collections.hosts, "All")
+    num_hosts = len(hosts_view.entities.get_all())
+    if num_hosts < 2:
```
mshriver: Just a thought, but not something to implement in this PR.

It would be great if ProviderFilter allowed for specifying a number of fields for required_fields. Right now it just supports looking for the presence of a field, and a matching value if supplied as a tuple.

For this PR, I think it would be better to use a fixture to get the number of hosts on the provider, skipping where there is only one. That way the skip happens in Setup, and not during Call phase.

The collection can actually be accessed directly via `provider.hosts`, an attribute is created by the BaseEntity class.

cfme/tests/infrastructure/test_host.py
mshriver: There is a method in the paginationpane widget for select all, I think we should use it here instead of iterating over like this.

Could be considered a separate test of pagination pane, could be parametrized, but I'm not sure that's worth it.

If we don't necessarily care how these items are selected, I'd use pagination select all, because its faster and a one-line call.

cfme/tests/infrastructure/test_host.py
mshriver: There's actually a subtle assertion going on here I guess, and that's the fact that the UI is displaying 2 or more hosts - its possible that the provider has more than two hosts in the REST collection used for `HostsCollection.all`, but there is  only one displayed on the UI.

cfme/tests/infrastructure/test_host.py
prichard77: This test is for refreshing multiple hosts, not all hosts. I'd like to leave the selection of the hosts similliar to the way it is, not using "all". I am going to take a closer look at how I determine the hosts (although similair code exists elsewhere). I could update the test to just use 2 hosts, but I figured we could test more if we had them and it would be a better test. Could add a refresh all test and use the suggested widget. I need to do some investigating as we are beginnig to go in circles with the PR.

prichard77:
I am going to move the chack for number of hosts back up into a fixture. I'll check the actual provider data (not UI) before even starting the test on a provider. Then I'll check again (if this is a supported requirement) in the test, that we have the same number of hosts displayed. I'll integrate this somehow into the list I iterate over to check the hosts. It's been suggested that I use an "all" widget, but this test isn't specifically for all hosts (I may add one later) just for multiple hosts.

dajoRH:
I detected some fixture changes in commit a46829722c15be392f83bcad278ffd39fa80bb94

The *local* fixture **`setup_provider_min_hosts`** is used in the following files:
- **cfme/tests/infrastructure/test_host.py**
    - *test_infrastructure_hosts_refresh_multi*

dajoRH:
I detected some fixture changes in commit 26583a1244dc3c910ba5b4abcd813ec51bba9b18

The *local* fixture **`setup_provider_min_hosts`** is used in the following files:
- **cfme/tests/infrastructure/test_host.py**
    - *test_infrastructure_hosts_refresh_multi*

dajoRH:
Lint report for commit 1b6a0cfd14942609476bf59650cb7a714835c943:

`cfme/tests/infrastructure/test_host.py`:
- :red_circle: Line 6:1: **F401** *'datetime' imported but unused*
- :red_circle: Line 353:5: **F841** *local variable 'current_refresh_date' is assigned to but never used*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
Lint report for commit e60f2c9aa30e3a3cdd1d0a1e2be35be97472b0f1:

`cfme/tests/infrastructure/test_host.py`:
- :red_circle: Line 6:1: **F401** *'datetime' imported but unused*
- :red_circle: Line 353:5: **F841** *local variable 'current_refresh_date' is assigned to but never used*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
I detected some fixture changes in commit e60f2c9aa30e3a3cdd1d0a1e2be35be97472b0f1

The *local* fixture **`setup_provider_min_hosts`** is used in the following files:
- **cfme/tests/infrastructure/test_host.py**
    - *test_infrastructure_hosts_refresh_multi*

dajoRH:
I detected some fixture changes in commit 5912a7ec53c7b77455bf25150d1e509063444175

The *local* fixture **`setup_provider_min_hosts`** is used in the following files:
- **cfme/tests/infrastructure/test_host.py**
    - *test_infrastructure_hosts_refresh_multi*

dajoRH:
I detected some fixture changes in commit 405789845dce0781e113719c3c042677cce671d6

The *local* fixture **`setup_provider_min_hosts`** is used in the following files:
- **cfme/tests/infrastructure/test_host.py**
    - *test_infrastructure_hosts_refresh_multi*

dajoRH:
I detected some fixture changes in commit ea7a074163ff467d8a5798a6e2bceda7acab69e4

The *local* fixture **`setup_provider_min_hosts`** is used in the following files:
- **cfme/tests/infrastructure/test_host.py**
    - *test_infrastructure_hosts_refresh_multi*

dajoRH:
*CFME QE Bot*

dajoRH:
I detected some fixture changes in commit 1aa2afccab7a336615d75dff6beb5efcec1a6ff8

The *local* fixture **`setup_provider_min_hosts`** is used in the following files:
- **cfme/tests/infrastructure/test_host.py**
    - *test_infrastructure_hosts_refresh_multi*

dajoRH:
I detected some fixture changes in commit c39a6034f140ea0226986d9f6813cd49dbb79c50

The *local* fixture **`setup_provider_min_hosts`** is used in the following files:
- **cfme/tests/infrastructure/test_host.py**
    - *test_infrastructure_hosts_refresh_multi*

cfme/tests/infrastructure/test_host.py
```diff
@@ -331,3 +334,59 @@ def test_hosts_not_displayed_several_times(appliance, provider, setup_provider):
...
+            2. Hosts view is displayed.
+            3.
+            4. "Refresh initiated for X Hosts from the CFME Database" is displayed in green
+               banner where "X" is the number of selected hosts. Properties for each host are
+               refreshed. Making changes to test pre-commithooks
+    """
+    hosts_view = navigate_to(provider.collections.hosts, "All")
+    num_hosts = len(hosts_view.entities.get_all())
+    if num_hosts < 2:
+        pytest.skip('not enough hosts in appliance UI to run test')
```
jawatts: This is duplicating the skip in the fixture above

jawatts: Hmm looking at Mike's comment above, I see there is a second assertion going on here.

cfme/common/provider.py
```diff
@@ -639,7 +639,8 @@ def is_refreshed(self, refresh_timer=None, refresh_delta=600):
             return False
         td = self.appliance.utc_time() - rdate
         if td > datetime.timedelta(0, refresh_delta):
-            self.refresh_provider_relationships()
+            if force_refresh:
+                self.refresh_provider_relationships()
```
jawatts: This is a good enhancement

cfme/tests/infrastructure/test_host.py
```diff
@@ -331,3 +334,59 @@ def test_hosts_not_displayed_several_times(appliance, provider, setup_provider):
...
     provider.create()
     assert host_count == navigate_to(appliance.collections.hosts, "All").paginator.items_amount
+
+
+@pytest.fixture
+def setup_provider_min_hosts(request, provider, min_hosts=2):
+    num_hosts = len(provider.data.get('hosts', {}))
+    if num_hosts < min_hosts:
+        pytest.skip(f'Not enough hosts({num_hosts}) to run test. Need at least {min_hosts}')
+    """Function-scoped fixture to set up a provider"""
```
jawatts: **Optional** please use `#` is you want this comment in here

prichard77: I was thinking the same when I copied it. Though maybe you guys were doing it differently for some reason. I remember to use # for single line comments (as I usually do) going forward. Maybe I'll go back and fix this in the place I copied it from also in another PR.

PR #9240, opened 2019-08-23
[WIPTEST] New Test: Testing null coalescing field values

ganeshhubale:
## Purpose or Intent
- This PR tests the ```null coalescing``` type of field which is used in inline automate method to specify values from schema
- This needs to automate because of BZ fixed in 5.11 - 1698184

### PRT Run
{{ pytest: cfme/tests/automate/test_method.py::test_null_coalescing_fields --use-template-cache -qsvvv}}


cfme/tests/automate/test_method.py
```diff
@@ -650,3 +652,88 @@ def test_method_for_log_and_notify(request, klass, notify_level, log_level):
...
+            1.
+            2.
+            3. Normal null coalescing behavior
+    """
+    var1, var2, var3, var4 = [fauxfactory.gen_alpha() for _ in range(4)]
+
+    klass.schema.add_fields(
+        *[
+            {"name": var, "type": "Attribute", "data_type": "String", "default_value": value}
+            for var, value in [[var1, "fred"], [var2, "barney"], [var3, " "]]
```
valaparthvi: **Optional**: With `var1` set to _fred_, I would love it if you could name `var2` _george_ :grinning:

cfme/tests/automate/test_method.py
```diff
@@ -650,3 +652,88 @@ def test_method_for_log_and_notify(request, klass, notify_level, log_level):
...
+            2.  Use Data Type null coalescing
+            3.  Make the default value something like this : ${#var3} || ${#var2} || ${#var1}
+        expectedResults:
+            1.
+            2.
+            3. Normal null coalescing behavior
+    """
+    var1, var2, var3, var4 = [fauxfactory.gen_alpha() for _ in range(4)]
+
+    klass.schema.add_fields(
```
valaparthvi:
```suggestion
    klass.schema.add_fields(
        *[
            {
                "name": var,
                "type": var_type,
                "data_type": "String",
                "default_value": value,
            }
            for var, value, var_type in [
                [var1, "fred", "Attribute"],
                [var2, "barney", "Attribute"],
                [var3, " ", "Attribute"],
                [var4, "", "Method"],
            ]
        ]
    )
```

PR #9245, opened 2019-08-23
[1LP][RFR]Fix failing tests and add ignore_stream wherever required

valaparthvi:
## Purpose or Intent

- __Fixing__ tests:
    1. `test_check_vm_retirement_requester` - test was failing due to a change in the request description, which is now fixed by using `find_by` instead of `get` action.
    2. Add ignore_stream("5.11") to test_rss.py and test_reports_with_timeline.py, since they have been removed from 5.11.
    3. Remove unnecessary code related to arbitration_profiles from cfme/tests/cloud/test_providers.py

### PRT Run
{{ pytest: cfme/tests/infrastructure/test_vm_retirement_rest.py::test_check_vm_retirement_requester  -vvv }}

### DIFF
--- a/cfme/tests/cloud/test_providers.py
+++ b/cfme/tests/cloud/test_providers.py
@@ -22,7 +22,6 @@ from cfme.common.provider_views import CloudProviderAddView
 from cfme.common.provider_views import CloudProvidersView
 from cfme.fixtures.provider import enable_provider_regions
 from cfme.markers.env_markers.provider import ONE
-from cfme.rest.gen_data import arbitration_profiles as _arbitration_profiles
 from cfme.utils import appliance
 from cfme.utils import conf
 from cfme.utils import ssh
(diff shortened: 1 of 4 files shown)


izapolsk:
Looks good to me. @valaparthvi, could you please explain why reports with timelines shouldn't be run against 5.11 ?

valaparthvi:
Hey @izapolsk RSS and Timelines have been removed from 5.11.

izapolsk:
@valaparthvi, please believe me that timelines are present in reports in 5.11. 
 So, please double check that such predefined timelines reports are removed in 5.11. 
In addition, if predefined timelines reports are removed I think we need to create new reports with timelines for those tests instead of disabling these tests.

@john-dupuy ^^

valaparthvi:
Hey @izapolsk,
The predefined timeline reports have not been removed from 5.11, but `Timelines` menu is.

The tests that I have marked with ignore_stream check if the timeline of these timeline-reports is accessible/properly displayed from Cloud Intel > Timelines. Since `Cloud Intel > Timelines` is no longer present in CF, these tests kinda render useless.

Although I think I can add a new test for this BZ.

john-dupuy:
@izapolsk Yep, @valaparthvi is correct, since Cloud Intel > Timelines has been removed, the test case `test_reports_with_timelines` will no longer be valid. As @izapolsk said, it appears that reports still have timelines associated with them, but the way they are accessed must have changed.

PR #9227, opened 2019-08-21
[1LP][RFR] generic object class level button tests

digitronik:
Signed-off-by: Nikhil Dhandre <ndhandre@redhat.com>

## Purpose or Intent
- Added edit related stuff
- adding custom button on generic object class level

PR #9254, opened 2019-08-26
[1LP][RFR]: Initial commit of test for service dialog BZ

john-dupuy:
Requires #9227. For testing purposes, I have rebased against @digitronik's PR/branch. Adding `qe_test_coverage` for BZ's 1729341 & 1743266. 

{{ pytest: --long-running cfme/tests/generic_objects/test_instances.py::test_generic_object_with_service_button }}

### DIFF
--- a/cfme/services/myservice/__init__.py
+++ b/cfme/services/myservice/__init__.py
@@ -3,7 +3,6 @@ import sentaku
 
 from cfme.common import Taggable
 from cfme.utils.appliance import Navigatable
-from cfme.utils.appliance import ViaREST
 from cfme.utils.update import Updateable
 
 
@@ -27,6 +26,7 @@ class MyService(Updateable, Navigatable, Taggable, sentaku.modeling.ElementMixin
     reconfigure_service = sentaku.ContextualMethod()
     launch_vm_console = sentaku.ContextualMethod()
     service_power = sentaku.ContextualMethod()
+    add_resource_generic_object = sentaku.ContextualMethod()
 
     def __init__(self, appliance, name=None, description=None, vm_name=None):
         self.appliance = appliance
@@ -36,19 +36,11 @@ class MyService(Updateable, Navigatable, Taggable, sentaku.modeling.ElementMixin
         self.parent = self.appliance.context
 
     @property
-    def rest_obj(self):
+    def rest_api_entity(self):
         return self.appliance.rest_api.collections.services.find_by(name=self.name)[0]
 
-    def add_resource_generic_object(self, gen_obj):
-        """ Add a generic object instance to the service """
-        with self.appliance.context.use(ViaREST):
-            self.rest_obj.action.add_resource(
-                resource=self.appliance.rest_api.collections.generic_objects.get(
-                    name=gen_obj.name
-                )._ref_repr()
-            )
 
-
-from cfme.services.myservice import ui, ssui  # NOQA last for import cycles
+from cfme.services.myservice import ui, ssui, rest  # NOQA last for import cycles
 importscan.scan(ui)
 importscan.scan(ssui)
+importscan.scan(rest)
--- a/cfme/tests/generic_objects/test_instances.py
+++ b/cfme/tests/generic_objects/test_instances.py
@@ -325,8 +325,8 @@ def test_generic_object_with_service_button(appliance, generic_object, button_wi
     """
     # add generic object to service
     myservice = MyService(appliance, name=generic_object.associations.get("services")[0].name)
-    myservice.add_resource_generic_object(generic_object)
-
+    with appliance.context.use(ViaREST):
+        myservice.add_resource_generic_object(generic_object)
     # now navigate to the details of the generic_object
     view = navigate_to(generic_object, "MyServiceDetails")
     view.toolbar.button(button_with_dialog.name).custom_button.click()
@@ -337,6 +337,6 @@ def test_generic_object_with_service_button(appliance, generic_object, button_wi
 
     # now for the actual test, make sure after hitting submit we're on the correct page
     try:
-        view = generic_object.create_view(MyServiceGenericObjectInstanceView, wait=10)
+        generic_object.create_view(MyServiceGenericObjectInstanceView, wait=10)
     except TimedOutError:
         pytest.fail("Could not wait for service's generic object view to displayed.")
--- a/cfme/services/myservice/rest.py
+++ b/cfme/services/myservice/rest.py
@@ -1,0 +1,13 @@
+from cfme.services.myservice import MyService
+from cfme.utils.appliance import MiqImplementationContext
+from cfme.utils.appliance.implementations.rest import ViaREST
+
+
+@MiqImplementationContext.external_for(MyService.add_resource_generic_object, ViaREST)
+def add_resource_generic_object(self, gen_obj):
+    """ Add a generic object instance to the service """
+    self.rest_api_entity.action.add_resource(
+        resource=self.appliance.rest_api.collections.generic_objects.get(
+            name=gen_obj.name
+        )._ref_repr()
+    )


dajoRH:
*CFME QE Bot*

cfme/tests/generic_objects/test_instances.py
```diff
@@ -42,6 +46,60 @@ def tags(request, appliance, categories):
...
+            description="testing dialog"
+        )
+        tab = service_dialog.tabs.instantiate("New tab")
+        box = tab.boxes.create(
+            box_label="box_{}".format(fauxfactory.gen_alpha()), box_desc="my box"
+        )
+        box.elements.create(element_data=[element_data])
+
+    # also create a view class for this dialog after it's accessed via a button
+    class TextBoxDialogView(TextInputDialogView):
```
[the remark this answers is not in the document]
digitronik: @john-dupuy I'm using some custom dialog view for testing custom button

https://github.com/ManageIQ/integration_tests/blob/58b757a679cf30a8efa63b708d13eee9b0ad7969/cfme/tests/automate/custom_button/__init__.py#L121

dajoRH:
I detected some fixture changes in commit 087cb7797d4d2f576d4e90c8291b8a698cbbe6cd

The *local* fixture **`textbox_dialog`** is used in the following files:
- **cfme/tests/generic_objects/test_instances.py**
    - *button_with_dialog*

The *local* fixture **`button_with_dialog`** is used in the following files:
- **cfme/tests/generic_objects/test_instances.py**

cfme/generic_objects/definition/button_groups.py
```diff
@@ -97,13 +97,15 @@ def create(self, name, description, request, image, button_type='Default',
...
             'display': display,
             'image': image,
-            'dialog': dialog,
             'open_url': open_url,
             'display_for': display_for,
             'request': request,
             'submit_version': submit_version,
             'system_message': system_message
         })
+        # note: for some reason this was not being filled if included in dict above
```
digitronik: Thanks for this :+1: 
John, I think better to move this as TODO so we will fix it.

cfme/services/myservice/__init__.py
```diff
@@ -34,6 +35,19 @@ def __init__(self, appliance, name=None, description=None, vm_name=None):
         self.vm_name = vm_name
         self.parent = self.appliance.context
 
+    @property
+    def rest_obj(self):
+        return self.appliance.rest_api.collections.services.find_by(name=self.name)[0]
+
+    def add_generic_object(self, gen_obj):
```
digitronik: optional: I would name this like `add_resource_generic_obj`. As we are linking resource to `service`.

cfme/tests/generic_objects/test_instances.py
```diff
@@ -284,3 +342,41 @@ def cleanup():
...
+        caseposneg: positive
+        testtype: functional
+        startsin: 5.10
+        casecomponent: GenericObjects
+    """
+    textbox_dialog, TextBoxDialogView = textbox_dialog
+    # add generic object to service
+    myservice = MyService(appliance, name=generic_object.associations.get("services")[0].name)
+    myservice.add_generic_object(generic_object)
+    # add myservice as part of the generic_objecct
```
digitronik: typo

cfme/tests/generic_objects/test_instances.py
```diff
@@ -284,3 +342,41 @@ def cleanup():
...
+    view = navigate_to(generic_object, "MyServiceDetails")
+    view.toolbar.button(button_with_dialog.name).custom_button.click()
+    view = generic_object.create_view(TextBoxDialogView, wait=10)
+    view.fill({"text_box": "asdf"})
+    wait_for(lambda: not view.submit.disabled, timeout="10s")
+    view.submit.click()
+
+    # now for the actual test, make sure after hitting submit we're on the correct page
+    view = generic_object.create_view(MyServiceGenericObjectInstanceView, wait=10)
+    assert view.is_displayed
```
digitronik: as we are asserting view; It will take care by `wait`

cfme/tests/generic_objects/test_instances.py
```diff
@@ -284,3 +342,41 @@ def cleanup():
...
+        testtype: functional
+        startsin: 5.10
+        casecomponent: GenericObjects
+    """
+    textbox_dialog, TextBoxDialogView = textbox_dialog
+    # add generic object to service
+    myservice = MyService(appliance, name=generic_object.associations.get("services")[0].name)
+    myservice.add_generic_object(generic_object)
+    # add myservice as part of the generic_objecct
+    generic_object.my_service = myservice
```
digitronik: If we are using `generic_object` global fixture then... this is done by that only

https://github.com/ManageIQ/integration_tests/blob/d776a12dd12e16475ac55d369a27ebecd29a6e7e/cfme/fixtures/generic_object.py#L39

dajoRH:
I detected some fixture changes in commit 89710b6ee0ba0a079b173006f643ea3c7362af83

The *local* fixture **`textbox_dialog`** is used in the following files:
- **cfme/tests/generic_objects/test_instances.py**

The *local* fixture **`button_with_dialog`** is used in the following files:
- **cfme/tests/generic_objects/test_instances.py**
    - *test_generic_object_with_service_button*

cfme/services/myservice/__init__.py
```diff
@@ -34,6 +35,19 @@ def __init__(self, appliance, name=None, description=None, vm_name=None):
...
+    def rest_obj(self):
+        return self.appliance.rest_api.collections.services.find_by(name=self.name)[0]
+
+    def add_resource_generic_object(self, gen_obj):
+        """ Add a generic object instance to the service """
+        with self.appliance.context.use(ViaREST):
+            self.rest_obj.action.add_resource(
+                resource=self.appliance.rest_api.collections.generic_objects.find_by(
+                    name=gen_obj.name
+                )[0]._ref_repr()
```
kedark3: any possibility of index errors here?

cfme/tests/generic_objects/test_instances.py
```diff
@@ -42,6 +45,58 @@ def tags(request, appliance, categories):
...
+
+    # # also create a view class for this dialog after it's accessed via a button
+    # class TextBoxDialogView(TextInputDialogView):
+    #     text_box = TextInput(locator=".//input")
+    #
+    #     @property
+    #     def is_displayed(self):
+    #         return self.submit.is_displayed and self.text_box.is_displayed
+
+    yield service_dialog
```
kedark3: Do we want to delete this as teardown?

cfme/tests/generic_objects/test_instances.py
```diff
@@ -284,3 +339,35 @@ def cleanup():
...
+        assignee: jdupuy
+        initialEstimate: 1/6h
+        caseimportance: high
+        caseposneg: positive
+        testtype: functional
+        startsin: 5.10
+        casecomponent: GenericObjects
+    """
+    # add generic object to service
+    myservice = MyService(appliance, name=generic_object.associations.get("services")[0].name)
```
kedark3: Any possibilities of uncaught exceptions with index or attribute?

cfme/tests/generic_objects/test_instances.py
```diff
@@ -284,3 +339,35 @@ def cleanup():
...
+    # now navigate to the details of the generic_object
+    view = navigate_to(generic_object, "MyServiceDetails")
+    view.toolbar.button(button_with_dialog.name).custom_button.click()
+    view = generic_object.create_view(TextInputDialogView, wait=10)
+    view.service_name.fill("Custom Button Execute")
+    wait_for(lambda: not view.submit.disabled, timeout="10s")
+    view.submit.click()
+
+    # now for the actual test, make sure after hitting submit we're on the correct page
+    view = generic_object.create_view(MyServiceGenericObjectInstanceView, wait=10)
```
kedark3: Did you want to assert something here?

john-dupuy: I got a comment from @digitronik `as we are asserting view; It will take care by wait`, so having an assert that the view is displayed after this line would be moot cause it wouldn't be hit if the view wasn't displayed.

cfme/tests/generic_objects/test_instances.py
john-dupuy: There should be none. Since the generic object is created with the service, it will have that association.

cfme/tests/generic_objects/test_instances.py
john-dupuy: woops, actually meant to delete this fixture as it is no longer used

cfme/services/myservice/__init__.py
john-dupuy: Nope. The rest collection returns a list, and since we are passing a generic_object to this method, this method will only fail if the generic object doesn't exist on the appliance

cfme/tests/generic_objects/test_instances.py
kedark3: Does `create_view` do that assertion? I remember creating view for v2v tests which has no way to assert, only navigating to a view would validate that. Maybe that has changed since I last used it.

john-dupuy: The test will fail with a TimedOutError if it exceeds the wait, we will have to nest it in a try/except if we want it to `Fail`

dajoRH:
I detected some fixture changes in commit fecdb1a3eb04b24e77988ca2280d253b6592832f

The *local* fixture **`button_with_dialog`** is used in the following files:
- **cfme/tests/generic_objects/test_instances.py**
    - *test_generic_object_with_service_button*

cfme/services/myservice/__init__.py
john-dupuy: however, this should probably use `.get`, rather than `find_by`

kedark3: thanks for clarifying

cfme/tests/generic_objects/test_instances.py
kedark3: For a test case it would make more sense if it fails as `AssertionError` than `TimedOutError` in my opinion. Cause in a test run, things time out all the time, and we usually see it as `Error` not `Failure` and we really care about reporting `Pass` or `Fail` , hence I'd suggest to have some mechanism to make it pass/fail.

john-dupuy: Yep, I agree, in this case it is the main assertion of the test so we should use a try/except, updating now.

dajoRH:
I detected some fixture changes in commit 22e03635ddb90006a4c2a4ea6ea75db46dd3f230

The *local* fixture **`button_with_dialog`** is used in the following files:
- **cfme/tests/generic_objects/test_instances.py**
    - *test_generic_object_with_service_button*

cfme/tests/generic_objects/test_instances.py
```diff
@@ -284,3 +305,38 @@ def cleanup():
...
+    view.toolbar.button(button_with_dialog.name).custom_button.click()
+    view = generic_object.create_view(TextInputDialogView, wait=10)
+    view.service_name.fill("Custom Button Execute")
+    wait_for(lambda: not view.submit.disabled, timeout="10s")
+    view.submit.click()
+
+    # now for the actual test, make sure after hitting submit we're on the correct page
+    try:
+        view = generic_object.create_view(MyServiceGenericObjectInstanceView, wait=10)
+    except TimedOutError:
```
kedark3: :+1:

cfme/generic_objects/definition/button_groups.py
```diff
@@ -97,13 +97,15 @@ def create(self, name, description, request, image, button_type='Default',
...
             'display': display,
             'image': image,
-            'dialog': dialog,
             'open_url': open_url,
             'display_for': display_for,
             'request': request,
             'submit_version': submit_version,
             'system_message': system_message
         })
+        # TODO: for some reason this was not being filled if included in dict above
```
mshriver: Fill order?

cfme/services/myservice/__init__.py
```diff
@@ -34,6 +35,19 @@ def __init__(self, appliance, name=None, description=None, vm_name=None):
         self.vm_name = vm_name
         self.parent = self.appliance.context
 
+    @property
+    def rest_obj(self):
```
mshriver: I think it would be better to be consistent with `rest_api_entity`, as this property name is already used for a few classes and is consistent with use of the word `entity`, though this is not _yet_ `BaseEntity` derived.

cfme/services/myservice/__init__.py
```diff
@@ -34,6 +35,19 @@ def __init__(self, appliance, name=None, description=None, vm_name=None):
...
+        return self.appliance.rest_api.collections.services.find_by(name=self.name)[0]
+
+    def add_resource_generic_object(self, gen_obj):
```
mshriver: Since there are already contextual methods defined for `MyService`, I think this should be added to a new `rest.py` context module.  

To do this you'll need to do a few things:
1. Add a contextualmethod above with the others, named as you wish
2. Create a new `rest.py` module with `ui.py` and `ssui.py`
3. Import this new module below with importscan
4. Decorate and define the method in `rest.py`

When the method is called, it'll resolve and call the ViaREST context automatically - or if you define this method for all three contexts, a test could be parametrized to test all three interfaces with the same test function.

cfme/tests/generic_objects/test_instances.py
```diff
@@ -284,3 +305,38 @@ def cleanup():
...
+    # add generic object to service
+    myservice = MyService(appliance, name=generic_object.associations.get("services")[0].name)
+    myservice.add_resource_generic_object(generic_object)
+
+    # now navigate to the details of the generic_object
+    view = navigate_to(generic_object, "MyServiceDetails")
+    view.toolbar.button(button_with_dialog.name).custom_button.click()
+    view = generic_object.create_view(TextInputDialogView, wait=10)
+    view.service_name.fill("Custom Button Execute")
+    wait_for(lambda: not view.submit.disabled, timeout="10s")
```
mshriver: You can avoid lambda here by passing `fail_cond=True`

cfme/tests/generic_objects/test_instances.py
mshriver: :+1: :+1:

cfme/services/myservice/__init__.py
mshriver: Check out what @valaparthvi is doing in #9329 !

PR #9260, opened 2019-08-28
[WIPTEST] Don't rewrite the pg_hba.conf when loosening it.

jaryn:
In commit 2950ae43d73c8 I was changing paths that is used by the DB permsissions "loosening" code. The path used on CFME 5.10 is wrong though and the DB "loosening" was not effective there (we didn't know about that as the command failed silently).

We were rewriting the pg_hba.conf in the "loosening" DB permissions step (in case 5.11)
in such a way that it results in actually disallowing the replication. The
vanilla state of the pg_hba.conf allows replication. Allowance of db replication is required for db backup opertaions to work. This made restore tests not working on 5.11 but working on 5.10.

This PR is fixing the path on 5.10 and also changing the code from doing complete rewrite to just adding to the pg_hba.conf, but from the fact that we haven't had the loosening in effect on 5.10 for some time, the question to ask is: *Do we really need to modify the pg_hba.conf at all? *

### DIFF
--- a/cfme/utils/appliance/db.py
+++ b/cfme/utils/appliance/db.py
@@ -34,7 +34,8 @@ class ApplianceDB(AppliancePlugin):
 
     @property
     def pg_prefix(self):
-        return '/opt/rh/rh-postgresql95/root' if self.appliance.version < '5.11' else ''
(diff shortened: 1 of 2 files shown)


jaryn:
I reopened the #9223 as it seems we don't really need to

PR #9261, opened 2019-08-28
[1LP][RFR] unassigned button association with rest

digitronik:

## Purpose or Intent

- Added instantiation with rest for custom button group
- Added new test for BZ-1737449
### PRT Run

{{pytest: cfme/tests/automate/custom_button/test_rest_custom_button.py::test_associate_unassigned_buttons_rest -v}}

### DIFF
--- a/cfme/automate/buttons.py
+++ b/cfme/automate/buttons.py
@@ -745,6 +745,31 @@ class ButtonGroup(BaseEntity, Updateable):
 
     _collections = {"buttons": ButtonCollection}
 
+    @classmethod
+    def from_id(cls, collection, id):
+        """Constructor for ButtonGroup
+        Args:
+            collection: parent
+            id: rest entity id
+        """
+        from cfme.tests.automate.custom_button import CLASS_MAP
+        gp, *_ = collection.appliance.rest_api.collections.custom_button_sets.find_by(id=id)
+        rest_btn_type = gp.set_data.get("applies_to_class")
+
+        for maps in CLASS_MAP.values():
+            if maps.get("rest") == rest_btn_type:
+                ui_btn_type = maps.get("ui")
+                break
+
+        return cls(
+            collection,
+            text=gp.name.split("|")[0],
+            hover=gp.description,
+            type=ui_btn_type,
+            display=gp.set_data.get("display"),
+            image=gp.set_data.get("button_icon"),
+        )
+
     @property
     def buttons(self):
         return self.collections.buttons
@@ -850,20 +875,17 @@ class ButtonGroupCollection(BaseCollection):
             assign_buttons=assign_buttons,
         )
 
+    def all(self):
+        """return all available custom button groups"""
+        return [
+            self.ENTITY.from_id(self, gp.id)
+            for gp in self.appliance.rest_api.collections.custom_button_sets.all
+        ]
+
     def delete_all(self):
         """This method will help to clean all custom button groups (rest)"""
         for btn_grp in self.appliance.rest_api.collections.custom_button_sets.all:
             btn_grp.action.delete()
-
-    def instantiate_with_rest(self, id):
-        from cfme.tests.automate.custom_button import CLASS_MAP
-        gp, *_ = self.appliance.rest_api.collections.custom_button_sets.find_by(id=id)
-        rest_btn_type = gp.set_data.get("applies_to_class")
-        for maps in CLASS_MAP.values():
-            if maps.get("rest") == rest_btn_type:
-                ui_btn_type = maps.get("ui")
-                break
-        return self.instantiate(text=gp.name, hover=gp.description, type=ui_btn_type)
 
 
 @navigator.register(ButtonGroupCollection, "All")
--- a/cfme/automate/buttons.py
+++ b/cfme/automate/buttons.py
@@ -10,6 +10,7 @@ from widgetastic.widget import ColourInput
 from widgetastic.widget import ConditionalSwitchableView
 from widgetastic.widget import ParametrizedView
 from widgetastic.widget import Select
+from widgetastic.widget import Table
 from widgetastic.widget import Text
 from widgetastic.widget import View
 from widgetastic_patternfly import BootstrapSelect
@@ -640,6 +641,7 @@ class ButtonGroupDetailView(AutomateCustomizationView):
         text_filter=lambda text: re.sub(r"\s+Display on Button\s*$", "", text),
     )
     hover = SummaryFormItem("Basic Information", "Hover Text")
+    assigned_buttons = Table('//h3[contains(text(), "Basic Information")]/following-sibling::table')
 
     @property
     def is_displayed(self):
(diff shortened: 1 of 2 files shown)


cfme/automate/buttons.py
```diff
@@ -853,6 +855,16 @@ def delete_all(self):
...
             btn_grp.action.delete()
 
+    def instantiate_with_rest(self, id):
+        from cfme.tests.automate.custom_button import CLASS_MAP
+        gp, *_ = self.appliance.rest_api.collections.custom_button_sets.find_by(id=id)
+        rest_btn_type = gp.set_data.get("applies_to_class")
+        for maps in CLASS_MAP.values():
+            if maps.get("rest") == rest_btn_type:
+                ui_btn_type = maps.get("ui")
+                break
```
sbulage: **Question:**
@digitronik can you please tell me why explicit use of `break`, any how it will return from `if` block .?

cfme/tests/automate/custom_button/test_rest_custom_button.py
```diff
@@ -159,3 +180,40 @@ def test_edit(self, buttons_groups, appliance, from_detail):
...
+        startsin: 5.10
+        casecomponent: CustomButton
+        tags: custom_button
+    """
+
+    # Add associate unassigned buttons to group with rest
+    set_data = group.set_data
+    set_data["button_order"] = [int(b.id) for b in buttons]
+    data = {"set_data": set_data}
+    group.action.edit(**data)
```
valaparthvi: Not asserting response?

digitronik: why we need to assert response? actually  test check is with UI buttons associated or not.

digitronik: added :+1: good to have more checks

dajoRH:
I detected some fixture changes in commit 895fb33f1c851c55e97521fabbd8b3f6a40f1e4f

The *local* fixture **`group`** is used in the following files:
- **cfme/tests/automate/custom_button/test_rest_custom_button.py**
    - *test_associate_unassigned_buttons_rest*

The *local* fixture **`buttons`** is used in the following files:
- **cfme/tests/automate/custom_button/test_rest_custom_button.py**
    - *test_associate_unassigned_buttons_rest*

cfme/tests/automate/custom_button/test_rest_custom_button.py
```diff
@@ -18,10 +20,29 @@
...
         lambda appliance, obj_type: obj_type not in OBJ_TYPE_59 and appliance.version < "5.10"
     ),
-    pytest.mark.parametrize("obj_type", OBJ_TYPE, ids=[obj.capitalize() for obj in OBJ_TYPE]),
+    pytest.mark.parametrize("obj_type", OBJ_TYPE, ids=[obj.capitalize() for obj in OBJ_TYPE],
+                            scope="module"),
 ]
 
 
+@pytest.fixture(scope="module")
+def group(request, appliance, obj_type):
```
mshriver: Might want to be more explicit in the naming here, since this is apparently returning a group REST ID, same below.

cfme/automate/buttons.py
```diff
@@ -853,6 +855,16 @@ def delete_all(self):
         for btn_grp in self.appliance.rest_api.collections.custom_button_sets.all:
             btn_grp.action.delete()
 
+    def instantiate_with_rest(self, id):
```
mshriver: I don't think this method should be added in this way, just to provide it for a REST based test.

What this is really providing is kwargs for instantiate via an ID, and I would like to separate that from the instantiation.

Ideally this could be added to BaseEntity, providing a way to instantiate by REST ID lookup, which would be advantageous to `BaseCollection.all` methods.

Consider, for now, creating a `ButtonGroup.kwargs_by_id` classmethod. That would provide the following pattern to fixtures/tests/framework:

```
def test_buttons(group):
    # using your fixture naming from below, but group here is actually a REST ID
    group_collection = appliance.collections.button_groups
    group_kwargs = group_collection.ENTITY.kwargs_by_id(group)
    group_entity = group_collection.instantiate(group_kwargs)
    assert group_entity.exists  # check it in the UI
    

```

cfme/tests/automate/custom_button/test_rest_custom_button.py
```diff
@@ -159,3 +180,41 @@ def test_edit(self, buttons_groups, appliance, from_detail):
...
+    group.action.edit(**data)
+    assert_response(appliance)
+
+    # Check association with UI
+    gp = appliance.collections.button_groups.instantiate_with_rest(group.id)
+
+    # Point to new object type so that teardown button(removed with rest) and new buttons, group
+    # (created with rest) reflect on UI without any selenium exception.
+    view = navigate_to(gp, "ObjectType")
+    view.browser.refresh()
```
mshriver: Why refresh here?

digitronik: @mshriver we are deleting and creating button with `rests` as per test steps... Its parametrized test. For first test it will work as navigation will start from beginning.  but, when we are on same page and other test (browser session) start run to display buttons (creation of new buttons and old test teardown means deletion) on page; we need browser refresh.

cfme/automate/buttons.py
digitronik: Interesting... let me try this. @mshriver I will ping; if found some problems with Entities.

dajoRH:
I detected some fixture changes in commit c2f240247a78b200f210045429c29f385526e4dc

The *local* fixture **`group_rest`** is used in the following files:
- **cfme/tests/automate/custom_button/test_rest_custom_button.py**
    - *test_associate_unassigned_buttons_rest*

The *local* fixture **`buttons_rest`** is used in the following files:
- **cfme/tests/automate/custom_button/test_rest_custom_button.py**
    - *test_associate_unassigned_buttons_rest*

cfme/tests/automate/custom_button/test_rest_custom_button.py
digitronik: not rest id actually but rest entity so i put `_rest`

cfme/automate/buttons.py
digitronik: @mshriver I like this idea... but added small changes like as we are writing `classmethod` better not return only kawars but instance. also I as you mentioned `all` method use case; I tried to cover it.

cfme/tests/automate/custom_button/test_rest_custom_button.py
mshriver: There is a `force` kwarg for `navigate_to` that should force navigation to the page, and won't allow is_displayed to keep it from navigating from the start of the 'stack' of navigation for that page.  I think it would be better to use it here, but at the very least an in-line comment explaining its need.

cfme/automate/buttons.py
```diff
@@ -848,6 +875,13 @@ def create(
             assign_buttons=assign_buttons,
         )
 
+    def all(self):
+        """return all available custom button groups"""
+        return [
+            self.ENTITY.from_id(self, gp.id)
+            for gp in self.appliance.rest_api.collections.custom_button_sets.all
```
mshriver: beautiful!

PR #9265, opened 2019-08-29
[1LP][RFR] Extend test: reload after deleting snapshot

psimovec:
{{ pytest: cfme/tests/infrastructure/test_snapshot.py::test_snapshot_crud -v --use-provider rhv43 --use-provider complete --long-running }}

### DIFF
--- a/cfme/tests/infrastructure/test_snapshot.py
+++ b/cfme/tests/infrastructure/test_snapshot.py
@@ -16,6 +16,7 @@ from cfme.utils.blockers import BZ
 from cfme.utils.conf import credentials
 from cfme.utils.generators import random_vm_name
 from cfme.utils.log import logger
+from cfme.utils.log_validator import LogValidator
 from cfme.utils.path import data_path
 from cfme.utils.ssh import SSHClient
 from cfme.utils.wait import wait_for
@@ -97,7 +98,9 @@ def test_memory_checkbox(small_test_vm, provider, soft_assert):
 
 
 @pytest.mark.rhv1
-@pytest.mark.meta(automates=[1608475])
+@pytest.mark.rhv3
+@test_requirements.rhev
+@pytest.mark.meta(automates=[1571291, 1608475])
 def test_snapshot_crud(small_test_vm, provider):
     """Tests snapshot crud
 
@@ -109,6 +112,11 @@ def test_snapshot_crud(small_test_vm, provider):
         casecomponent: Infra
         initialEstimate: 1/6h
     """
+    result = LogValidator(
+        "/var/www/miq/vmdb/log/evm.log",
+        failure_patterns=[r".*ERROR.*"],
+    )
+    result.start_monitoring()
     # has_name is false if testing RHEVMProvider
     snapshot = new_snapshot(small_test_vm, has_name=(not provider.one_of(RHEVMProvider)))
     snapshot.create()
@@ -116,6 +124,9 @@ def test_snapshot_crud(small_test_vm, provider):
     if provider.appliance.version >= "5.11" and provider.one_of(RHEVMProvider):
         assert snapshot.size
     snapshot.delete()
+    provider.refresh_provider_relationships()
+    wait_for(provider.is_refreshed, func_kwargs={'refresh_delta': 10}, timeout=600)
+    assert result.validate(wait="60s")
 
 
 @pytest.mark.rhv3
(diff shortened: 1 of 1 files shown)


dajoRH:
Lint report for commit bba601794a0c8f5e380800e3286ee921f3aaca48:

`cfme/tests/infrastructure/test_snapshot.py`:
- :bangbang: Line 143:1: **E302** *expected 2 blank lines, found 1*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
Lint report for commit 489044b1de328827b6f6f304278c81ef0e2ceb91:

`cfme/tests/infrastructure/test_snapshot.py`:
- :bangbang: Line 143:1: **E302** *expected 2 blank lines, found 1*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -109,13 +112,22 @@ def test_snapshot_crud(small_test_vm, provider):
         casecomponent: Infra
         initialEstimate: 1/6h
     """
+    result = LogValidator(
```
john-dupuy: _Optional_ consider using the `LogValidator`'s contextmanager

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -109,13 +112,22 @@ def test_snapshot_crud(small_test_vm, provider):
...
+        "/var/www/miq/vmdb/log/evm.log",
+        failure_patterns=[r".*ERROR.*"],
```
john-dupuy: Q: Is there any more specific `ERROR` message we could search for here? We wouldn't want this test to fail due to an error in an unrelated part of the appliance.

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -109,13 +112,22 @@ def test_snapshot_crud(small_test_vm, provider):
...
+    )
+    result.start_monitoring()
     # has_name is false if testing RHEVMProvider
     snapshot = new_snapshot(small_test_vm, has_name=(not provider.one_of(RHEVMProvider)))
     snapshot.create()
     # check for the size as "read" check
     if provider.appliance.version >= "5.11" and provider.one_of(RHEVMProvider):
         assert snapshot.size
     snapshot.delete()
+    wait_for(lambda: not snapshot.exists)
```
john-dupuy: Q: any `timeout` to assign here?

cfme/tests/infrastructure/test_snapshot.py
psimovec: I don't know about more specific errors that could happen, I followed verification steps https://bugzilla.redhat.com/show_bug.cgi?id=1571291#c13 .. and there is written that there should be no errors in evm.log.. So I guess multiple things could go wrong

cfme/tests/infrastructure/test_snapshot.py
psimovec: @john-dupuy trying if it could work, but I am losing wait="60s" in result.validate( ) by using contextmanager

cfme/tests/infrastructure/test_snapshot.py
psimovec: @john-dupuy default timeout is 2 mins.. Do you think that it would be better to increase the timeout? I don't know what would be reasonable one for snapshot removal.

cfme/tests/infrastructure/test_snapshot.py
john-dupuy: That's cool with me, left it _optional_ just to see if you wanted to implement it. If it doesn't work in this scenario, there's no need to shoehorn it in :+1:

cfme/tests/infrastructure/test_snapshot.py
john-dupuy: @psimovec I asked @lina-nikiforova the same question on one of her PR's and it seems like the same thing applies here. https://github.com/ManageIQ/integration_tests/pull/9375#discussion_r327011168

cfme/tests/infrastructure/test_snapshot.py
john-dupuy: @psimovec I just realized that the Snapshot `delete` method will also include the `wait_for`, so I would consider this one redundant. 

https://github.com/ManageIQ/integration_tests/blob/5dc69946af7d4f25ba1f0bc5a1a9a2e4978fce67/cfme/infrastructure/virtual_machines.py#L756-L775

cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -97,7 +98,9 @@ def test_memory_checkbox(small_test_vm, provider, soft_assert):
 
 
 @pytest.mark.rhv1
-@pytest.mark.meta(automates=[1608475])
+@pytest.mark.rhv3
```
mshriver: These are supposed to be RHV levels for testing (equivalent to tiers), and I don't think it makes sense to have more than one level (tier) marked on a case?

PR #9266, opened 2019-08-29
[1LP][RFR] Add sidebar accordion to Compute Infrastructure pages AND automate test: test_infrastructure_provider_left_panel_titles

valaparthvi:
## Purpose or Intent

- __Adding tests__ `test_infrastructure_provider_left_panel_titles`
- __Updating tests__ `test_run_host_analysis` which uses host details sidebar.
- __Enhancement__ Add sidebar accordions to the following pages
    1. Infrastructure > Providers
    2. Infrastructure > Clusters
    3. Infrastructure > Hosts
    4. Infrastructure > Resource Pools

Note: Dependent on : https://github.com/RedHatQE/widgetastic.patternfly/pull/114

### PRT Run
{{ pytest : cfme/tests/webui/test_general_ui.py::test_infrastructure_provider_left_panel_titles cfme/tests/infrastructure/test_host_analysis.py::test_run_host_analysis -vvv }}

### DIFF
--- a/cfme/tests/webui/test_general_ui.py
+++ b/cfme/tests/webui/test_general_ui.py
@@ -1,17 +1,25 @@
 import fauxfactory
 import pytest
+from widgetastic.utils import partial_match
 
 from cfme import test_requirements
 from cfme.base.ui import LoginPage
 from cfme.cloud.provider import CloudProvider
+from cfme.common.host_views import ProviderAllHostsView
 from cfme.common.provider_views import CloudProviderAddView
 from cfme.common.provider_views import ContainerProviderAddView
 from cfme.common.provider_views import InfraProviderAddView
 from cfme.common.provider_views import InfraProvidersView
 from cfme.common.provider_views import PhysicalProviderAddView
+from cfme.common.provider_views import ProviderDetailsView
+from cfme.common.provider_views import ProviderTimelinesView
 from cfme.infrastructure.config_management.ansible_tower import AnsibleTowerProvider
 from cfme.infrastructure.config_management.satellite import SatelliteProvider
 from cfme.infrastructure.provider import InfraProvider
+from cfme.infrastructure.datastore import ProviderAllDatastoresView
+from cfme.infrastructure.provider import ProviderClustersView
+from cfme.infrastructure.provider import ProviderTemplatesView
+from cfme.infrastructure.provider import ProviderVmsView
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.infrastructure.virtual_machines import InfraVmDetailsView
 from cfme.markers.env_markers.provider import ONE
@@ -20,10 +28,22 @@ from cfme.markers.env_markers.provider import ONE_PER_TYPE
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.blockers import BZ
 from cfme.utils.log_validator import LogValidator
+from cfme.utils.log import logger
 from cfme.utils.wait import wait_for
 
 
 pytestmark = [test_requirements.general_ui]
+
+ALL_OPTIONS = {
+    "properties": {"Summary": ProviderDetailsView, "Timelines": ProviderTimelinesView},
+    "relationships": {
+        "Clusters (": ProviderClustersView,
+        "Hosts (": ProviderAllHostsView,
+        "Datastores (": ProviderAllDatastoresView,
+        "VMs (": ProviderVmsView,
+        "Templates (": ProviderTemplatesView,
+    },
+}
 
 
 @pytest.fixture()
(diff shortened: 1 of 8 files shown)


cfme/tests/webui/test_general_ui.py
```diff
@@ -1,20 +1,38 @@
...
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.wait import wait_for
 
-
 pytestmark = [test_requirements.general_ui]
 
+ALL_OPTIONS = {
+    "properties": {"Summary": ProviderDetailsView, "Timelines": ProviderTimelinesView},
+    "relationships": {
+        "Clusters (": ProviderClustersView,
```
valaparthvi: `(` has been added at the end of every key as a workaround for [BZ 1752372](https://bugzilla.redhat.com/show_bug.cgi?id=1752372).

dajoRH:
*CFME QE Bot*

valaparthvi:
Failure in 511 PRT run is not relevant to the changes made by this PR.

cfme/tests/webui/test_general_ui.py
john-dupuy: @valaparthvi the bz linked was closed as `NOTABUG`; how is the `(` relevant to the BZ?

cfme/tests/webui/test_general_ui.py
```diff
@@ -465,6 +485,53 @@ def test_add_provider_button_accordion(has_no_providers, provider):
...
+    """
+    view = navigate_to(provider, "Details")
+    accordion = getattr(view.entities.sidebar, option)
+
+    for panel in ALL_OPTIONS[option]:
+        # Sometimes the panel doesn't enable immediately,
+        # using wait_for to give it some time to enable
+        wait_for(
+            lambda: not accordion.tree.is_disabled(partial_match(panel)),
+            fail_func=view.browser.refresh,
```
john-dupuy: Could you re-navigate rather than do a browser refresh?

cfme/tests/webui/test_general_ui.py
valaparthvi: @john-dupuy that bug was not always reproducible, so to be on the safe side I added the `(`.

cfme/tests/webui/test_general_ui.py
valaparthvi: Why so?

john-dupuy: I think I've been told before that we should avoid browser refreshes in tests? @jawatts or @mshriver is this something we should generally avoid?

jawatts: Correct. It adds a lot of time to our test runs.

valaparthvi: Okay. I added `view.toolbar.reload.click` instead, I hope that works. Let me know if otherwise :)

valaparthvi: @john-dupuy @jawatts I cannot use `view.toolbar.reload.click` because it is not available on some pages. I'll have to go back to using `view.browser.refresh`, let me know if you have some other solution :)

jawatts: @valaparthvi another option is to force renavigation

valaparthvi: Wouldn't navigate take longer than page refresh?

jawatts: Agree and we wouldn't want to renavigate for a `fail_func`.

valaparthvi: So are we okay with `view.browser.refresh`?

dajoRH:
*CFME QE Bot*

cfme/tests/webui/test_general_ui.py
jawatts: @valaparthvi yes, I am okay with it. Looking at the rest of the PR

jawatts:
Just waiting on PRT

valaparthvi:
These tests passed locally, but PRT keeps resulting into `Invalid` state for some reason :/
Just a heads up, PRT will fail for 511 because of `test_run_host_analysis` but that is not related to the changes made by this PR.

jawatts:
@valaparthvi looks like `test_general_ui` failed as well in 5.11

valaparthvi:
@jawatts So I figured what went wrong, I guess I'll have to open another PR to wt.patternfly :weary: 
[Here](https://github.com/RedHatQE/widgetastic.patternfly/blob/master/src/widgetastic_patternfly/__init__.py#L631), there is no parent, so it selects any element that matches the xpath without taking into account the scope. There is an extra element that matches the xpath on 511, which is causing this failure. I think defining the parent here should fix this.

cfme/tests/webui/test_general_ui.py
```diff
@@ -509,6 +529,53 @@ def test_provider_details_page_refresh_after_clear_cookies(
...
+            message="Wait until the panel enables",
+            handle_exception=True,
+            silent_failure=True
+        )
+        if not accordion.tree.is_disabled(partial_match(panel)):
+            accordion.tree.select(partial_match(panel))
+            test_view = provider.create_view(ALL_OPTIONS[option][panel], wait="60s")
+            soft_assert(test_view.is_displayed, f"{test_view} not displayed.")
+        else:
+            logger.info(f"'{panel}' was not tested, it did not enable.")
```
mshriver: I think this should be a pytest.skip or pytest.fail - the item was expected to enable why wouldn't that be a failure?

valaparthvi: There are few items that get enabled once the provider receives data, but there are items such as `Timelines` that take some 20-30 minutes to enable itself, so I decided to just log it, I don't want to fail/skip the test in this case. Is there a way to quickly enable Timelines?

valaparthvi:
Moving to RFR. 511 PRT failures are unrelated to these changes.

cfme/tests/webui/test_general_ui.py
mshriver: I'm not aware of a way to quickly enable timelines in this case.

I still am not comfortable with this result for the test though.  There will be no indication in the results or artifactor report that this occurred - you would have to inspect the cfme.log file on every test execution to check for this message and to know whether or not timelines was actually tested.

A pytest xfail may be your best option here - that its more or less and expected failure state. We need some indication in results that there was missing coverage for this test case.

That or test timelines separately, putting more investigation into how to enable the option quickly.

cfme/tests/webui/test_general_ui.py
```diff
@@ -509,6 +529,53 @@ def test_provider_details_page_refresh_after_clear_cookies(
...
+            fail_func=view.browser.refresh,
+            num=40,
+            delay=2,
+            message="Wait until the panel enables",
+            handle_exception=True,
+            silent_failure=True
+        )
+        if not accordion.tree.is_disabled(partial_match(panel)):
+            accordion.tree.select(partial_match(panel))
+            test_view = provider.create_view(ALL_OPTIONS[option][panel], wait="60s")
```
mshriver: Since you passed `wait` parameter, this will raise an exception before your soft assert if the view isn't displayed.

dajoRH:
I detected some fixture changes in commit d702bc8e244787c6e52513f9a67717a97c81b68e

The *local* fixture **`vm`** is used in the following files:
- **cfme/tests/webui/test_general_ui.py**

cfme/tests/webui/test_general_ui.py
valaparthvi: So John helped me trigger Timelines, activities like creating/deleting/reconfiguring VMs create an event and triggers Timelines. So I did that, I create VM as a part of setup.

dajoRH:
I detected some fixture changes in commit 32cd546367d7ddfc626117342b615d5e00462cac

The *local* fixture **`vm`** is used in the following files:
- **cfme/tests/webui/test_general_ui.py**

PR #9267, opened 2019-08-29
[WIPTEST] test_advanced_search -  doctrings get lost when transforming function to method

psimovec:
## Purpose or Intent
Make Polarion great again :sunglasses: 

Added quick fix that copies docstring of test function that is being wrapped up.

_Needs to be verified if Polarion can work with these tests now._

### DIFF
--- a/cfme/tests/webui/test_advanced_search.py
+++ b/cfme/tests/webui/test_advanced_search.py
@@ -177,6 +177,7 @@ def methodized(metafunc):
 
     def func(self, param, appliance):
         return metafunc(param, appliance)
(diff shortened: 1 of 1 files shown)


mshriver:
I referenced making this change here: https://github.com/ManageIQ/integration_tests/pull/9225#discussion_r318281675

While it _does_ work with the versions of pytest-polarion-collect and polarion-docstrings that we currently have, it does _not_ work with the latest versions of those packages.

We will need updates to pytest-polarion-collect/polarion-docstrings in order to attempt to resolve docstrings when the test functions are injected and the names don't match the pytest node IDs.

PR #9268, opened 2019-08-29
[1LP][RFR] New Test: Testing state machine variables

ganeshhubale:
## Purpose or Intent
- This PR includes testing of state machines variables by executing ```inline``` automate methods in state schema
- Removed test case - ```test_automate_retry_onexit_increases```. It is already automated test case - ```test_automate_schema_field_without_type```
### PRT Run
{{ pytest: cfme/tests/automate/test_class.py::test_state_machine_variable --use-template-cache -qsvvv }}

### DIFF
--- a/cfme/automate/explorer/method.py
+++ b/cfme/automate/explorer/method.py
@@ -457,9 +457,7 @@ class MethodCollection(BaseCollection):
 
         add_page = navigate_to(self, 'Add')
 
-        if self.browser.product_version < '5.11':
-            location = location
-        else:
+        if self.browser.product_version > '5.11' and location.islower():
             location = location.capitalize()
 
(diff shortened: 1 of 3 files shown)


cfme/automate/explorer/method.py
```diff
@@ -457,9 +457,7 @@ def create(
 
         add_page = navigate_to(self, 'Add')
 
-        if self.browser.product_version < '5.11':
-            location = location
-        else:
+        if self.browser.product_version > '5.11' and location.islower():
```
sbulage: **Suggestion:**
@ganeshhubale This logic could be like this --> `location.lower() if self.browser.product_version < '5.11' else location.capitalize()`

PR #9271, opened 2019-08-29
[1LP][RFR] New Test: Testing hash data in simulation tree on UI

ganeshhubale:
## Purpose or Intent
- Testing hash data generated by automate method

### PRT Run
{{ pytest: cfme/tests/automate/test_simulation.py::test_automate_simulation_result_has_hash_data --use-template-cache -qsvvv }}

### DIFF
--- a/cfme/tests/automate/__init__.py
+++ b/cfme/tests/automate/__init__.py
@@ -1,1 +1,67 @@
 # -*- coding: utf-8 -*-
+from textwrap import dedent
+
+user_list_hash_data = dedent(
+    """
+module Demo
+  module Automate
+    module System
+      module Request
+        module Test
+          class AvailableUsers
+            def initialize(handle = $evm)
+              @handle = handle
+            end
+
+            def main
+              fill_dialog_field(fetch_list_data)
+            end
+
+            private
+
+            def fetch_list_data
+              user1 = {:id => 1, :name => 'Fred', :age => 52, :email => "fred@bedrock.com"}
+              user2 = {:id => 2, :name => 'Barney', :age => 55, :email => "barney@bedrock.com" }
+              user3 = {:id => 3, :name => 'Wilma', :age => 39, :email => "wilma@berdock.com" }
+              user4 = {:id => 4, :name => 'Betty', :age => 38, :email => "betty@bedrock.com" }
+              user_list = { user1 => "#{user1[:name]} - ID #{user1[:id]}",
+                            user2 => "#{user2[:name]} - ID #{user2[:id]}",
+                            user3 => "#{user3[:name]} - ID #{user3[:id]}",
+                            user4 => "#{user4[:name]} - ID #{user4[:id]}"}
+              @handle.log(:info, "User List #{user_list}")
+              return nil => "<none>" if user_list.blank?
+
+              user_list[nil] = "<select>" if user_list.length > 1
+              user_list
+            end
+
+            def fill_dialog_field(list)
+              dialog_field = @handle.object
+
+              # sort_by: value / description / none
+              dialog_field["sort_by"] = "description"
+
+              # sort_order: ascending / descending
+              dialog_field["sort_order"] = "ascending"
+
+              # data_type: string / integer
+              dialog_field["data_type"] = "string"
+
+              # required: true / false
+              dialog_field["required"] = "true"
+
+              dialog_field["values"] = list
+              dialog_field["default_value"] = list.length == 1 ? list.keys.first : nil
+            end
+          end
+        end
+      end
+    end
+  end
+end
+
+if __FILE__ == $PROGRAM_NAME
+  Demo::Automate::System::Request::Test::AvailableUsers.new.main
+end
+        """
+)
--- a/cfme/tests/automate/test_simulation.py
+++ b/cfme/tests/automate/test_simulation.py
@@ -1,11 +1,15 @@
 # -*- coding: utf-8 -*-
+from textwrap import dedent
+
 import fauxfactory
 import pytest
 
 from cfme import test_requirements
 from cfme.automate.simulation import simulate
+from cfme.base.ui import AutomateSimulationView
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.blockers import BZ
+from cfme.utils.log_validator import LogValidator
 
 pytestmark = [test_requirements.automate, pytest.mark.tier(2)]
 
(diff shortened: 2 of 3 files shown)


cfme/tests/automate/test_simulation.py
```diff
@@ -86,3 +90,128 @@ def test_assert_failed_substitution(copy_class):
...
+            1. Create a Instance under /System/Request called ListUser, update it so that it points
+               to a ListUser Method
+            2. Create ListUser Method under /System/Request, paste the Attached Method
+            3. Run Simulation
+        expectedResults:
+            1.
+            2.
+            3. The UI should display the result objects
+    """
+    script = dedent(
```
john-dupuy: @ganeshhubale I think this script should be stored in `data/ruby_scripts/` and then imported. You could even just make it a python file with the same string. But it'd be good to keep all these scripts in a centralized location. What do you think?

cfme/tests/automate/test_simulation.py
```diff
@@ -86,3 +90,128 @@ def test_assert_failed_substitution(copy_class):
...
+        view.result_tree.click_path(
+            *(
+                f"ManageIQ/SYSTEM / PROCESS / {instance.klass.name}",
+                f"ManageIQ/System / {instance.klass.name} / Call_Instance",
+                f"{instance.domain.name}/System / {instance.klass.name} / {instance.name}",
+                "values",
+                "Hash",
+                "Key",
+            )
+        ).text
```
john-dupuy: Holy tree path, Batman! :open_mouth:

cfme/tests/automate/test_simulation.py
ganeshhubale: Hey @john-dupuy 
I agree but instead of creating files under ```data/ruby_scripts/```; I have added ruby scripts in already available file ```tests/automate/__init__.py```

john-dupuy: @ganeshhubale Thanks, that works too!

cfme/tests/automate/__init__.py
```diff
@@ -1 +1,67 @@
 # -*- coding: utf-8 -*-
+from textwrap import dedent
+
+script01 = dedent(
```
john-dupuy: Q: Is there any descriptive name to give to this script?

ganeshhubale: Yes. We can give name. I have updated with this one: ```user_list_hash_data```. Because this method creates user list and shows this hash data in simulation tree or in automation log

cfme/tests/automate/test_simulation.py
```diff
@@ -86,3 +88,64 @@ def test_assert_failed_substitution(copy_class):
...
+            1. Create a Instance under /System/Request called ListUser, update it so that it points
+               to a ListUser Method
+            2. Create ListUser Method under /System/Request, paste the Attached Method
+            3. Run Simulation
+        expectedResults:
+            1.
+            2.
+            3. The UI should display the result objects
+    """
+    from . import user_list_hash_data
```
izapolsk: what's the reason for having import statement here ?

cfme/tests/automate/test_simulation.py
```diff
@@ -86,3 +88,64 @@ def test_assert_failed_substitution(copy_class):
...
+                "instance": instance.name,
+            },
+            message="create",
+            request="Call_Instance",
+            execute_methods=True,
+        )
+    view = instance.create_view(AutomateSimulationView)
+    assert (
+        view.result_tree.click_path(
+            *(
```
izapolsk: you wrap a bunch of values into tuple and then roll it out into bunch of parameters again. what for ?

izapolsk:
good PR. I like it but I have a few questions.

cfme/tests/automate/test_simulation.py
ganeshhubale: I imported ```user_list_hash_data``` here. Because it is only related to this test case and will not be used in any other test case. 
Please let me know your views @izapolsk

cfme/tests/automate/test_simulation.py
ganeshhubale: After executing this test case. This type of tree generates in simulation tree view. I just have to check if this tree exists or not. Hence I am providing this path to check if it click able or not.
Please let me know if you have some other way to perform this part. 
Actually, I could have tried ```ha_path``` but I preferred clicking in that path.

cfme/tests/automate/__init__.py
```diff
@@ -1 +1,67 @@
...
+            end
+
+            def main
+              fill_dialog_field(fetch_list_data)
+            end
+
+            private
+
+            def fetch_list_data
+              user1 = {:id => 1, :name => 'Fred', :age => 52, :email => "fred@bedrock.com"}
```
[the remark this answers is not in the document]
ganeshhubale: Yes I agree with you @izapolsk . Actually, I am sure about it's copyrights. I got this code attached with BZ(https://bugzilla.redhat.com/show_bug.cgi?id=1445089#c0). 
How can we check it's copyrights?

cfme/tests/automate/test_simulation.py
izapolsk: @ganeshhubale, actually, according to pep8 all imports should be put in the head of file.
there are exceptional cases when we put imports inside some code when those cause some import issues.

cfme/tests/automate/test_simulation.py
izapolsk: click_path is correct. I'm talking about *() constuction. you wrap items into tuple and then roll out this tuple back to items.

cfme/tests/automate/test_simulation.py
ganeshhubale: Thanks @izapolsk :+1: 
I didn't get pep8 error for this import. So kept as it is.
I will update as you suggested.

cfme/tests/automate/test_simulation.py
ganeshhubale: Updated as below:
```
assert (
        view.result_tree.click_path(
                f"ManageIQ/SYSTEM / PROCESS / {instance.klass.name}",
                f"ManageIQ/System / {instance.klass.name} / Call_Instance",
                f"{instance.domain.name}/System / {instance.klass.name} / {instance.name}",
                "values",
                "Hash",
                "Key",
        ).text
        == "Key"
    )```

PR #9272, opened 2019-08-29
[1LP][RFR] Fix ValueError due to opening files with 'rw' mode

mshriver:
Seeing value error in the file defined below for PRT.

{{ pytest: --long-running --use-provider cmqe cfme/tests/containers/test_manageiq_ansible_custom_attributes.py::test_manageiq_ansible_edit_custom_attributes -k "openshift-3.9" }}


mshriver:
PRT tests passed, just haven't updated here yet.

PR #9276, opened 2019-08-29
[1LP][RFR] Add test_requirements.containers for all container tests

jawatts:
Add `test_requirements.containers` marker for all container tests


mshriver:
Obvious stat padding, but I guess I'll allow it.

PR #9286, opened 2019-08-30
[1LP][RFR] Remove old bz 1549529, allow tests to run

john-dupuy:
This should have been included in #9248, but I missed it. Essentially it is doing the same thing as that PR, but for a slightly different test case. 

{{ pytest: --use-provider rhv43 --long-running cfme/tests/control/test_actions.py::test_action_create_snapshots_and_delete_them }}

### DIFF
--- a/cfme/tests/control/test_actions.py
+++ b/cfme/tests/control/test_actions.py
@@ -595,8 +595,10 @@ def test_action_create_snapshot_and_delete_last(appliance, request, vm, vm_on, p
     )
 
 
-@pytest.mark.meta(blockers=[BZ(1748410, forced_streams=["5.10"])], automates=[1748410])
 @pytest.mark.provider([VMwareProvider, RHEVMProvider], scope="module")
+@pytest.mark.meta(blockers=[BZ(1748410, forced_streams=["5.10"])],
+                  automates=[1748410],
+                  unblock=lambda provider: provider.one_of(VMwareProvider))
 def test_action_create_snapshots_and_delete_them(request, appliance, vm, vm_on, policy_for_testing):
     """ This test tests actions 'Create a Snapshot' (custom) and 'Delete all Snapshots'.
 
--- a/cfme/tests/control/test_actions.py
+++ b/cfme/tests/control/test_actions.py
@@ -28,7 +28,6 @@ from cfme.tests.control import wait_for_ssa_enabled
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.blockers import BZ
 from cfme.utils.generators import random_vm_name
-from cfme.utils.log import logger
 from cfme.utils.log_validator import LogValidator
 from cfme.utils.update import update
 from cfme.utils.wait import TimedOutError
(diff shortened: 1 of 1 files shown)


cfme/tests/control/test_actions.py
```diff
@@ -597,22 +596,16 @@ def finalize():
 
 
 @pytest.mark.provider([VMwareProvider, RHEVMProvider], scope="module")
-@pytest.mark.meta(
-    blockers=[
-        BZ(
-            1549529,
-            forced_streams=["5.10", "upstream"],
```
ganeshhubale: I can see that this  BZ(1549529) is CLOSED CURRENTRELEASE but It will be better, if we add this BZ(1549529) under *Bugzilla* data.

john-dupuy: @ganeshhubale I didn't put it there because it is the same BZ as 1745065, the original bug was closed but it's still an issue.

ganeshhubale: Yes agree But I would put this BZ for more info around this test case. Although this is somehow related to this test case.

john-dupuy: Okay, I can put it in there :+1:

sshveta:
@john-dupuy - Please check PRT failures 
" E       wait_for.TimedOutError: Could not do 'wait for snapshots to be deleted' at /cfme/cfme_tests/cfme/tests/control/test_actions.py:657 in time"

john-dupuy:
Thanks @sshveta, I think this test case requires some more attention, moving this PR to WIPTEST for now.

john-dupuy:
@sshveta I believe there is a CFME BZ causing this failing test, I added a blocker and opened https://bugzilla.redhat.com/show_bug.cgi?id=1748410

cfme/tests/control/test_actions.py
```diff
@@ -596,23 +595,20 @@ def finalize():
...
 
 
+@pytest.mark.meta(blockers=[BZ(1748410, forced_streams=["5.10"])], automates=[1748410])
 @pytest.mark.provider([VMwareProvider, RHEVMProvider], scope="module")
-@pytest.mark.meta(
-    blockers=[
-        BZ(
-            1549529,
-            forced_streams=["5.10", "upstream"],
-            unblock=lambda provider: provider.one_of(VMwareProvider),
```
sshveta: Don't you want to unblock the test for Vmware ? The bug is only for RHV right ?

john-dupuy: Oh good catch, thanks @sshveta!

cfme/tests/control/test_actions.py
```diff
@@ -649,9 +647,6 @@ def create_one_snapshot(n):
             message="wait for snapshot %d to appear" % (n + 1),
             delay=5,
         )
-        current_snapshot = vm.current_snapshot_name
-        logger.debug("Current Snapshot Name: {}".format(current_snapshot))
```
digitronik: Q: any special reason to remove this logging and assertion?
Its checking `Active` state from `All Snapshot Tree`.

Other thought like not to assert thing in model page is better; but make sure we have test for this.

john-dupuy: This is due to one of the BZ's mentioned in the docblock https://bugzilla.redhat.com/show_bug.cgi?id=1745065

This BZ is not really relevant to this test case, but yeah I agree that we should have some automation to cover that BZ. I will work on a small test case to cover this.

PR #9288, opened 2019-09-03
[RFR] fixed the expected path for ansible logs in the collected logs

lina-nikiforova:
## Purpose or Intent
Log collection tests failed on checking ansible logs. Before there was a workaroud and now it seems to work the same for all archives.

### PRT Run

{{ pytest: -v cfme/tests/configure/test_log_depot_operation.py::test_collect_log_depot  --long-running }}

### DIFF
--- a/cfme/tests/configure/test_log_depot_operation.py
+++ b/cfme/tests/configure/test_log_depot_operation.py
@@ -22,8 +22,7 @@ from cfme.utils.ftp import FTPClient
 from cfme.utils.ssh import SSHClient
 from cfme.utils.update import update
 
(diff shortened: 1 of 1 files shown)


lina-nikiforova:
Marking as RFR based on https://bugzilla.redhat.com/show_bug.cgi?id=1751961

You can see the same error in the PRT as well
```
                        log_file_size = result.output.split()[0]
>                       assert int(log_file_size) > 0, "Log file is empty!"
E                       AssertionError: Log file is empty!
E                       assert 0 > 0
E                        +  where 0 = int('0')

cfme/tests/configure/test_log_depot_operation.py:178: AssertionError
```

PR #9293, opened 2019-09-03
[1LP][RFR] Add blocker BZ(1746931) -- In openstack provisioning.

jaryn:
Add blocker to Openstack provisioning tests.

### DIFF
--- a/cfme/tests/cloud_infra_common/test_provisioning.py
+++ b/cfme/tests/cloud_infra_common/test_provisioning.py
@@ -473,6 +473,7 @@ def test_cloud_provision_from_template_with_attached_disks(
 
 
 # Not collected for EC2 in generate_tests above
+@pytest.mark.meta(blockers=[BZ(1746931)])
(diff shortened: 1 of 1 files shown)


cfme/tests/cloud_infra_common/test_provisioning.py
```diff
@@ -473,6 +473,7 @@ def delete_vm_and_wait_for_gone():
 
 
 # Not collected for EC2 in generate_tests above
+@pytest.mark.meta(blockers=[BZ(1746931)])
```
john-dupuy: @jaryn I see that his BZ is related to Openstack Catalog Items, but not provisioning? Is provisioning also broken by this BZ?

jaryn: yes. there is a dup BZ to this I reported: 1748330

john-dupuy: Got it, thanks!

PR #9297, opened 2019-09-03
[RFR] Add in test case for BZ 1741310

john-dupuy:
Creating a test case for https://bugzilla.redhat.com/show_bug.cgi?id=1741310

{{ pytest: cfme/tests/webui/test_general_ui.py::test_add_provider_button_accordion }}

### DIFF
--- a/cfme/tests/webui/test_general_ui.py
+++ b/cfme/tests/webui/test_general_ui.py
@@ -7,11 +7,14 @@ from cfme.common.provider_views import ContainerProviderAddView
 from cfme.common.provider_views import InfraProviderAddView
 from cfme.common.provider_views import InfraProvidersView
 from cfme.common.provider_views import PhysicalProviderAddView
+from cfme.infrastructure.config_management.ansible_tower import AnsibleTowerProvider
+from cfme.infrastructure.config_management.satellite import SatelliteProvider
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.infrastructure.virtual_machines import InfraVmDetailsView
 from cfme.markers.env_markers.provider import ONE
 from cfme.markers.env_markers.provider import ONE_PER_TYPE
 from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.blockers import BZ
 from cfme.utils.wait import wait_for
 
 
@@ -431,6 +434,35 @@ def test_compliance_column_header(appliance, setup_provider, provider):
     next(hr for hr in table.browser.elements(table.HEADERS) if hr.text == "Compliant").click()
     # Page should not break after after clicking the compliant column
     assert view.is_displayed
+
+
+@pytest.mark.ignore_stream("5.10")
+@pytest.mark.meta(blockers=[BZ(1741310)], automates=[1741310])
+@pytest.mark.provider([AnsibleTowerProvider, SatelliteProvider])
+def test_add_provider_button_accordion(has_no_providers, provider):
+    """
+    Test that add_provider button is visible after clicking accordion on
+        Ansible Tower and Satellite Provider pages
+
+    Bugzilla:
+        1741310
+
+    Polarion:
+        assignee: pvala
+        casecomponent: WebUI
+        caseimportance: medium
+        initialEstimate: 1/30h
+        startsin: 5.11
+    """
+    view = navigate_to(provider, "AllOfType")
+    assert view.add_button.is_displayed
+    # now click somewhere on the accordion
+    view.sidebar.configured_systems.open()
+    # now click back at the providers page
+    view.sidebar.providers.open()
+    view.wait_displayed()
+    # assert that the button is still present
+    assert view.add_button.is_displayed
 
 
 @pytest.mark.manual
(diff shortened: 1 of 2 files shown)


cfme/infrastructure/config_management.py
```diff
@@ -170,6 +170,7 @@ class ConfigManagementAllView(ConfigManagementView):
     sidebar = View.nested(ConfigManagementSideBar)
     search = View.nested(Search)
     including_entities = View.include(ConfigManagementEntities, use_parent=True)
+    add_button = Button('Add a Provider')
```
mshriver: I think this should actually be on `ConfigManagementEntities` since its in that section of the page, when the table is not.

cfme/tests/webui/test_general_ui.py
```diff
@@ -339,6 +341,42 @@ def test_tls_openssl_verify_mode(temp_appliance_preconfig, request):
     assert appliance.advanced_settings["smtp"]["openssl_verify_mode"] == "none"
 
 
+@pytest.mark.ignore_stream("5.10")
+@pytest.mark.meta(blockers=[BZ(1741310)], automates=[1741310])
+@pytest.mark.parametrize("destination", ["AnsibleTowerExplorer", "ConfigManagement"])
+def test_add_provider_button_accordion(appliance, destination):
+    """
+    Test that add_provider button is visible after clicking accordion on
```
mshriver: Is this the button that's only displayed when there are no providers configured? I think you'll want to use `has_no_providers` fixture if so, to ensure the appliance is 'clean' when this runs.

john-dupuy: Does the `has_no_providers` fixture work for these extra providers? Ansible Tower and Config Manager -> Satellite?

cfme/infrastructure/config_management.py
john-dupuy: Makes sense, I will update this

john-dupuy: Done, also employed a similar strategy for AnsibleTower

cfme/tests/webui/test_general_ui.py
valaparthvi: Nope, `has_no_providers` won't work here. We'll have to write a separate fixture for it.

valaparthvi: **Required:**
I think writing a local fixture to remove these providers before the test starts would do for now, what do you think @john-dupuy  @mshriver ? Writing a cleanup function for ConfigManager providers like the one we have for BaseProvider seems a little out of scope of this PR and a little futile to me.

cfme/tests/webui/test_general_ui.py
```diff
@@ -339,6 +341,42 @@ def test_tls_openssl_verify_mode(temp_appliance_preconfig, request):
...
+        accordion_select = "configured_systems"
+
+    assert view.add_button.is_displayed
+    # now click somewhere on the accordion
+    getattr(view.sidebar, accordion_select).open()
+    # now click back at the providers page
+    view.sidebar.providers.open()
+    view.wait_displayed()
+    # assert that the button is still present
+    assert view.including_entities.add_button.is_displayed
```
valaparthvi: **Required:**
`view.including_entities` gives AttributeError. I think it should simply be `view.add_button.is_displayed`.

john-dupuy: Oh duh, I noticed this in local testing and though I pushed this fix -- I guess not!

cfme/tests/webui/test_general_ui.py
john-dupuy: Thanks @valaparthvi, let me look into this. Perhaps it will be easy to implement a fixture to clear the providers. I am also working on #9317, but haven't got very far yet, maybe this could be something that'd be easy to implement as a class method on `ConfigManager`.

dajoRH:
*CFME QE Bot*

cfme/tests/webui/test_general_ui.py
john-dupuy: @valaparthvi now that #9317, we can finally use `has_no_providers` with Config Managers :tada:

valaparthvi: #9317 got merged! That was quick :grinning:
@john-dupuy we already have a test [`test_add_button_on_provider_all_page`](https://github.com/ManageIQ/integration_tests/blob/61a7eb1f82ebce57186b3591b26901cf804fd1d1/cfme/tests/webui/test_general_ui.py#L275), can you see if we can merge this test into it?

valaparthvi:
Hey @john-dupuy, can you check the PRT runs? I think the PRT result is a false positive.

cfme/tests/webui/test_general_ui.py
john-dupuy: Ah, okay, let me look into it.

john-dupuy:
@valaparthvi the test was skipped because the BZ has not yet been fixed.

cfme/tests/webui/test_general_ui.py
john-dupuy: @valaparthvi the test case you linked is a bit different. Here according to the BZ the add provider button disappears if you click the somewhere in the accordion and then click back. That test case just looks like it's checking that the button shows up when there's no provider.

PR #9302, opened 2019-09-04
[1LP][RFR] Cat item copy

digitronik:
## Purpose or Intent
- Added Copy  method in catalog item
- Automate on bugzilla
- ** Note:** test will skip as we are automating that bug having open status**

### PRT Run
{{pytest: cfme/tests/automate/custom_button/test_catalog_custom_button.py::test_catalog_item_copy_with_custom_buttons}}

### DIFF
--- a/cfme/services/catalogs/catalog_items/__init__.py
+++ b/cfme/services/catalogs/catalog_items/__init__.py
@@ -360,15 +360,17 @@ class BaseCatalogItem(BaseEntity, Updateable, Pretty, Taggable):
     def copy(self, name=None):
         view = navigate_to(self, 'Copy')
 
+        # there is default name like `Copy of *`
         if name:
             view.name.fill(name)
 
+        copied_name = view.name.value
         view.add.click()
         view.flash.assert_no_error()
 
         # Catalog item can be any type
         item_args = self.__dict__
-        item_args["name"] = view.name.value
+        item_args["name"] = copied_name
         return self.__class__(**item_args)
 
     def add_button_group(self, **kwargs):
--- a/cfme/services/catalogs/catalog_items/__init__.py
+++ b/cfme/services/catalogs/catalog_items/__init__.py
@@ -248,6 +248,21 @@ class EditCatalogItemView(BasicInfoForm):
             wait_for(lambda: not self.save.disabled, timeout='10s', delay=0.2)
 
 
+class CopyCatalogItemView(ServicesCatalogView):
+    name = Input(name="name")
+
+    add = Button("Add")
+    cancel = Button("Cancel")
+
+    @property
+    def is_displayed(self):
+        return (
+            self.title.text == f'Service Catalog Item "{self.context["object"].name}"'
+            and self.catalog_items.is_opened
+            and self.catalog_items.is_dimmed
+        )
+
+
 class TabbedEditCatalogItemView(ServicesCatalogView):
     fill_strategy = WaitFillViewStrategy()
     save = Button('Save')
@@ -341,6 +356,20 @@ class BaseCatalogItem(BaseEntity, Updateable, Pretty, Taggable):
         view.flash.assert_success_message(VersionPick(
             {LOWEST: 'The selected Catalog Item was deleted',
              '5.11': 'The catalog item "{}" has been successfully deleted'.format(self.name)}))
+
+    def copy(self, name=None):
+        view = navigate_to(self, 'Copy')
+
+        if name:
+            view.name.fill(name)
+
+        view.add.click()
+        view.flash.assert_no_error()
+
+        # Catalog item can be any type
+        item_args = self.__dict__
+        item_args["name"] = view.name.value
+        return self.__class__(**item_args)
 
     def add_button_group(self, **kwargs):
         button_name = kwargs.get("text", "gp_{}".format(fauxfactory.gen_alpha()))
@@ -693,6 +722,15 @@ class CatalogItemEditStep(CFMENavigateStep):
 
     def step(self, *args, **kwargs):
         self.prerequisite_view.configuration.item_select('Edit this Item')
+
+
+@navigator.register(BaseCatalogItem, "Copy")
+class CatalogItemCopyStep(CFMENavigateStep):
+    VIEW = CopyCatalogItemView
+    prerequisite = NavigateToSibling("Details")
+
+    def step(self, *args, **kwargs):
+        self.prerequisite_view.configuration.item_select("Copy Selected Item")
 
 
 @navigator.register(BaseCatalogItem, 'AddButtonGroup')
--- a/cfme/tests/automate/custom_button/test_catalog_custom_button.py
+++ b/cfme/tests/automate/custom_button/test_catalog_custom_button.py
@@ -8,6 +8,7 @@ from cfme.utils.appliance import ViaSSUI
 from cfme.utils.appliance import ViaUI
 from cfme.utils.appliance.implementations.ssui import navigate_to as ssui_nav
 from cfme.utils.appliance.implementations.ui import navigate_to as ui_nav
+from cfme.utils.blockers import BZ
 
 
 pytestmark = [pytest.mark.tier(2), test_requirements.custom_button]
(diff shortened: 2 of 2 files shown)


cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -342,6 +357,20 @@ def delete(self):
...
+        if name:
+            view.name.fill(name)
+
+        view.add.click()
+        view.flash.assert_no_error()
+
+        # Catalog item can be any type
+        item_args = self.__dict__
+        item_args["name"] = view.name.value
+        return self.__class__(**item_args)
```
sbulage: Nice Implementation :100: :rocket:

cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -248,6 +248,21 @@ def after_fill(self, was_change):
             wait_for(lambda: not self.save.disabled, timeout='10s', delay=0.2)
 
 
+class CopyCatalogItemView(ServicesCatalogView):
+    name = Input(name="name")
+
```
sbulage: **Suggestion:**
Remove this extra line.

cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -248,6 +248,21 @@ def after_fill(self, was_change):
...
+    add = Button("Add")
+    cancel = Button("Cancel")
```
sbulage: **Optional:**
Make `button` variable/object names more readable like:
`add_btn` and `cancel_btn`

cfme/tests/automate/custom_button/test_catalog_custom_button.py
```diff
@@ -121,3 +122,41 @@ def test_custom_button_unassigned_behavior_catalog_level(appliance, generic_serv
...
+    btn_data = {
+        "text": gen_numeric_string(start="button_"),
+        "hover": gen_numeric_string(start="hover_"),
+        "image": "fa-user",
+    }
+
+    btn = generic_catalog_item.add_button(**btn_data)
+
+    # copy catalog item
+    new_cat_item = generic_catalog_item.copy()
```
sbulage: **Suggestion:**
Can you please compare the changes of two objects, as you are copying it.? It's better to check the content to make sure it is copied.

cfme/tests/automate/custom_button/test_catalog_custom_button.py
```diff
@@ -121,3 +122,41 @@ def test_custom_button_unassigned_behavior_catalog_level(appliance, generic_serv
...
+
+    # check for catalog item and button on copied
+    assert new_cat_item.exists
```
sbulage: **Suggestion:**
I think `teardown` for `new_cat_item` is required, what's say ?

cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -342,6 +357,20 @@ def delete(self):
             {LOWEST: 'The selected Catalog Item was deleted',
              '5.11': 'The catalog item "{}" has been successfully deleted'.format(self.name)}))
 
+    def copy(self, name=None):
+        view = navigate_to(self, 'Copy')
+
+        if name:
+            view.name.fill(name)
+
+        view.add.click()
```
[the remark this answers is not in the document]
digitronik: No it has default value... so `add` button will be active even we won't pass name

cfme/services/catalogs/catalog_items/__init__.py
digitronik: yup... I added this in state of `instantiation` because catalog item can be any `type` and instantiation will be hectic task in cases so I used this :dancer:

cfme/services/catalogs/catalog_items/__init__.py
digitronik: aaaah.... as per me when we write any `view` or you can say any `code`; to separate out difference generally we add such `blanks lines`. If you want i would like to remove :+1:

cfme/services/catalogs/catalog_items/__init__.py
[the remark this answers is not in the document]
sbulage: Okay :+1:

cfme/services/catalogs/catalog_items/__init__.py
digitronik: This is nice point :+1:..
here i fallowed older implementation of file; If you check `catalog_items/__init__.py` mostly we used `add` or `cancel` not adding extension like `btn`. I think both notation are good, but try to fallow single in model page (means sure about consistency for specific model page).

cfme/services/catalogs/catalog_items/__init__.py
sbulage: Yes you're right approach but we can have blank line between group of `objects` and `property/method`, So I would suggest remove it :smile:

cfme/tests/automate/custom_button/test_catalog_custom_button.py
digitronik: This test mainly for copy catalog item with button (`test_catalog_item_copy_with_custom_buttons`) so I'm more interested in button checks not in other attributes or entities along with catalog item.

But yes, we need one test `test_copy_catalog_item`. which will check other entities along with catalog item. As this PR merge @niyazRedhat  can you please add one test for this functionality?

cfme/services/catalogs/catalog_items/__init__.py
digitronik: hummm interesting :). @sbulage  can you please point me any guideline or python code styling instructions. I'm little bit confused about it. I found such thing in c-python code base as well :)

cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -342,6 +357,22 @@ def delete(self):
...
              '5.11': 'The catalog item "{}" has been successfully deleted'.format(self.name)}))
 
+    def copy(self, name=None):
+        view = navigate_to(self, 'Copy')
+
+        # there is default name like `Copy of *`
+        if name:
+            view.name.fill(name)
+
+        copied_name = view.name.value
```
john-dupuy: Q: Why not just use `name` here?

cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -342,6 +357,22 @@ def delete(self):
...
+        view.add.click()
+        view.flash.assert_no_error()
+
+        # Catalog item can be any type
+        item_args = self.__dict__
+        item_args["name"] = copied_name
+        return self.__class__(**item_args)
```
john-dupuy: Q: Do we have a common way that we copy instances (with updates) in integration_tests? I have seen `from copy import copy` used, but I like this method here.

cfme/services/catalogs/catalog_items/__init__.py
digitronik: so `copy` method of catalog item just asked `name`; means it copied same catalog item with new name. Catalog item name is optional else default it will take like `Copy of *`. so I'm just reading name from textbox here.

cfme/services/catalogs/catalog_items/__init__.py
digitronik: yes... copy provide `deepcopy` and `shadow` copy but it has some limitations. when object is complex; then i use this way.

cfme/services/catalogs/catalog_items/__init__.py
john-dupuy: Ah, duh! makes sense, thanks @digitronik :+1:

cfme/services/catalogs/catalog_items/__init__.py
izapolsk: I wouldn't change button names. we already follow some approach. code should look similar everywhere

izapolsk:
@digitronik, looks good but the test is skipped against 5.11 in PRT. please double check your command and etc

digitronik:
@izapolsk yes, its skipping as we blocked it with Bugzilla (opened). In purpose/Intent of PR, I added respective note.

Providing some traceback without blocking test in support of this PR

```
        request.addfinalizer(new_cat_item.delete_if_exists)
    
        # check for catalog item and button on copied
        assert new_cat_item.exists
>       assert new_cat_item.button_exists(btn)
E       AssertionError: assert False
E        +  where False = <bound method BaseCatalogItem.button_exists of GenericCatalogItem(name='Copy of item_S9ECr', catalog=Catalog(name='cat...alog(label='sSGBI4R2aS', description='my dialog sSGBI4R2aS'), domain='ManageIQ (Locked)', provider=None, subtype=None)>('button_371')
E        +    where <bound method BaseCatalogItem.button_exists of GenericCatalogItem(name='Copy of item_S9ECr', catalog=Catalog(name='cat...alog(label='sSGBI4R2aS', description='my dialog sSGBI4R2aS'), domain='ManageIQ (Locked)', provider=None, subtype=None)> = GenericCatalogItem(name='Copy of item_S9ECr', catalog=Catalog(name='cat_0UTvwxU9IL', description='my catalog', items=N...ialog(label='sSGBI4R2aS', description='my dialog sSGBI4R2aS'), domain='ManageIQ (Locked)', provider=None, subtype=None).button_exists

test_catalog_custom_button.py:163: AssertionError

```

digitronik:
aaaaaaah It turn to `ON_QA` let me rerun PRT :+1:

PR #9306, opened 2019-09-04
[1LP][RFR] Add a new test for control breadcrumbs

john-dupuy:
Adding a test case for this BZ https://bugzilla.redhat.com/show_bug.cgi?id=1740290

{{ pytest: --long-running cfme/tests/control/test_bugs.py::test_control_breadcrumbs }}

### DIFF
--- a/cfme/tests/control/test_bugs.py
+++ b/cfme/tests/control/test_bugs.py
@@ -36,6 +36,14 @@ BAD_CONDITIONS = [
 ]
 
 
+BREADCRUMB_LOCATIONS = dict(
+    ControlExplorer=["Control", "Explorer", "Policy Profiles", "All Policy Profiles"],
+    ControlSimulation=["Control", "Simulation", "Policies"],
+    ControlImportExport=["Control", "Import / Export"],
+    ControlLog=["Control", "Log"]
+)
+
+
 def create_policy(request, collection):
     args = (VMControlPolicy, fauxfactory.gen_alpha())
     kwargs = {}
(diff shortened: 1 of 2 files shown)


dajoRH:
I detected some fixture changes in commit 5376e074375ca1712369d15b2250cbe1e0d7c63f

The *local* fixture **`breadcrumb_locations`** is used in the following files:
- **cfme/tests/control/test_bugs.py**
    - *test_control_breadcrumbs*

cfme/tests/control/test_bugs.py
```diff
@@ -624,3 +635,29 @@ def test_policy_condition_multiple_ors(
...
+def test_control_breadcrumbs(appliance, page, breadcrumb_locations):
+    """
+    Test to make sure breadcrumbs are visible and properly displayed
+
+    Bugzilla:
+        1740290
+
+    Polarion:
+        assignee: jdupuy
+        caseimportance: high
```
sbulage: **Optional:**
By default `caseimportance` is `high`.

cfme/tests/control/test_bugs.py
```diff
@@ -234,6 +234,17 @@ def vm_compliance_policy_profile(appliance, compliance_condition):
     policy.delete()
 
 
+@pytest.fixture
+def breadcrumb_locations():
```
sbulage: **Suggestion:**
I think simply `Dictionary` of `locations` would okay rather than treating it as `fixture`.

cfme/tests/control/test_bugs.py
```diff
@@ -234,6 +234,17 @@ def vm_compliance_policy_profile(appliance, compliance_condition):
     policy.delete()
 
 
+@pytest.fixture
```
sbulage: **Question:**
Any specific reason to have this `Dictionary` as a `Fixture`.?

cfme/tests/control/test_bugs.py
john-dupuy: No specific reason, I really saw them as equivalent so if you'd prefer I can just make it a constant `dict`

john-dupuy: @sbulage done

cfme/tests/control/test_bugs.py
john-dupuy: In my view it is better to be explicit here

cfme/tests/control/test_bugs.py
```diff
@@ -624,3 +632,29 @@ def test_policy_condition_multiple_ors(
...
+        1740290
+
+    Polarion:
+        assignee: jdupuy
+        caseimportance: high
+        casecomponent: Control
+        initialEstimate: 1/30h
+        startsin: 5.11
+    """
+    view = navigate_to(appliance.server, "ControlExplorer")
```
digitronik: Q: why we are checking every parametrization `ControlExplorer`? 
I think better to make it part of `page`.

john-dupuy: Talked off line, this is a part of the BZ. We have to navigate to this page first to see the BZ. I will add a comment to help explain this.

digitronik: As per offline discussion this is valid condition as per BZ. 

I like to request you... just comment on this navigation :+1:

john-dupuy: done.

PR #9310, opened 2019-09-05
[1LP][RFR] custom css manual test steps update

psimovec:

### DIFF
--- a/cfme/tests/test_db_migrate_manual.py
+++ b/cfme/tests/test_db_migrate_manual.py
@@ -295,15 +295,17 @@ def test_update_webui_custom_css():
         assignee: jhenner
         casecomponent: Appliance
         caseimportance: medium
-        initialEstimate: 1/15h
+        initialEstimate: 1/6h
 
         setup:  1. provision appliance
                 2. ssh to the appliance
                 3. change custom CSS file - you can run command:
-        vmdb && cd public &&\
-        wget 'https://image.spreadshirtmedia.com/image-server/v1/designs/m11541836-6,width=100.png'\
-        && echo "html {cursor: url('m11541836-6,width=100.png'), pointer;}" >> custom.css
-                4. open webui and check css changed you made (different cursor)
+                    echo 'body {background-color: #FF00EF;}' >> /var/www/miq/vmdb/public/custom.css
+                4. add repo file to /etc/yum.repos.d/
+                5. add RHSM settings
+                6. update through webui
+                7. check webui is available
+                8. check customizations still work
 
         startsin: 5.9
     """
--- a/cfme/tests/test_db_migrate_manual.py
+++ b/cfme/tests/test_db_migrate_manual.py
@@ -297,15 +297,22 @@ def test_update_webui_custom_css():
         caseimportance: medium
         initialEstimate: 1/6h
 
-        setup:  1. provision appliance
-                2. ssh to the appliance
-                3. change custom CSS file - you can run command:
+        setup:     provision appliance
+
+        testSteps:
+                1. ssh to the appliance
+                2. change custom CSS file - you can run command:
                     echo 'body {background-color: #FF00EF;}' >> /var/www/miq/vmdb/public/custom.css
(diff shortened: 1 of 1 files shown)


cfme/tests/test_db_migrate_manual.py
```diff
@@ -288,19 +288,23 @@ def test_update_webui_custom_css():
...
-               check customization"s still work
+        initialEstimate: 1/15h
+
+        setup:  1. provision appliance
+                2. ssh to the appliance
+                3. change custom CSS file - you can run command:
+        vmdb && cd public &&\
+        wget 'https://image.spreadshirtmedia.com/image-server/v1/designs/m11541836-6,width=100.png'\
+        && echo "html {cursor: url('m11541836-6,width=100.png'), pointer;}" >> custom.css
+                4. open webui and check css changed you made (different cursor)
```
jaryn: The steps are not complete. After the step 4 we need to the
- add repo file to /etc/yum.repos.d/
- add RHSM settings
- update through webui
- check webui is available
- check customization"s still work

jaryn:
As there are some steps needed to fulfil the test, it is not good idea to change the cursor as the one you picked is damn imprecise as ... it was a potato. !@#$% (: Let's change some colours.

cfme/tests/test_db_migrate_manual.py
```diff
@@ -288,19 +288,23 @@ def test_update_webui_custom_css():
...
         caseimportance: medium
-        initialEstimate: 1/6h
-        setup: provision appliance
-               add custom css file
-               add repo file to /etc/yum.repos.d/
-               add RHSM settings
-               update through webui
-               check webui is available
-               check customization"s still work
+        initialEstimate: 1/15h
```
jaryn: I also would keep the time as it was as there is a lot of time needed for putting the repo file on the appliance and doing the update

cfme/tests/test_db_migrate_manual.py
```diff
@@ -288,19 +288,25 @@ def test_update_webui_custom_css():
...
         initialEstimate: 1/6h
-        setup: provision appliance
-               add custom css file
-               add repo file to /etc/yum.repos.d/
-               add RHSM settings
-               update through webui
-               check webui is available
-               check customization"s still work
+
+        setup:  1. provision appliance
```
john-dupuy: As far as I can tell, this should be part of the `testSteps` and not the `setup`? If so, please move to `testSteps` and give corresponding `expectedResults`

cfme/tests/test_db_migrate_manual.py
```diff
@@ -288,19 +288,29 @@ def test_update_webui_custom_css():
...
+        testSteps:
+                1. ssh to the appliance
+                2. change custom CSS file - you can run command:
+                    echo 'body {background-color: #FF00EF;}' >> /var/www/miq/vmdb/public/custom.css
+                3. add repo file to /etc/yum.repos.d/
+                4. add RHSM settings
+                5. update through webui
+
+        expectedResults:
+                webui is available
```
john-dupuy: @psimovec These expected results must line up with the `testSteps`

see e.g.

https://github.com/ManageIQ/integration_tests/blob/65501a812cecb79720712a390554f6e4ccf3e3cb/cfme/tests/automate/custom_button/test_service_objects.py#L589-L620

cfme/tests/test_db_migrate_manual.py
```diff
@@ -288,19 +288,32 @@ def test_update_webui_custom_css():
...
-               update through webui
-               check webui is available
-               check customization"s still work
+
+        setup:     provision appliance
+
+        testSteps:
+                1. ssh to the appliance
+                2. change custom CSS file - you can run command:
+                    echo 'body {background-color: #FF00EF;}' >> /var/www/miq/vmdb/public/custom.css
```
mshriver: Looking forward to you automating this too :) There's an `ssh_client` on the appliance that makes this nice and easy.

PR #9319, opened 2019-09-09
[WIPTEST] Fixed is displyed of script data in inline method

ganeshhubale:
## Purpose or Intent
- This PR fixes ```is_displayed``` for widget ```ScriptBox``` used in ```inline``` type of automate method.
- It's is_displayed is not working because of change in Xpath
- This leads to fail number test cases in automate FA.
### PRT Run
{{ pytest: cfme/tests/automate/test_method.py --use-template-cache -qsvvv }}


izapolsk:
@ganeshhubale, https://github.com/ManageIQ/integration_tests/pull/9328 this PR should temporarily fix your issue with sentaku. I think you need to double check who changed the code related to your tests.

dajoRH:
I detected some fixture changes in commit 0a88108774a36fbb04b180d197844cca0ace5a51

The *local* fixture **`generic_object_definition`** is used in the following files:
- **cfme/tests/automate/test_method.py**

widgetastic_manageiq/__init__.py
```diff
@@ -988,7 +988,7 @@ def __init__(self, parent, locator=None, item_name=None, logger=None):
 
     def __locator__(self):
         if not self.locator:
-            self.locator = "//textarea[contains(@id, 'method_data')]"
+            self.locator = "//textarea[contains(@id, 'method_data')]/.."
```
mshriver: If we want the parent element of the textarea, I think its much better to explicitly state it, not use this relative xpath reference.

You should be able to specify this like below, assuming the parent is a `div`:
```
self.locator = "//div[//textarea[contains(@id, 'method_data')]]"
```

The selected element will be the parent div - but only if it contains the child textarea.

PR #9327, opened 2019-09-09
[1LP][RFR] Default polarion status and approvals

mshriver:
Has been tested against stage polarion instance, polarion workflow restrictions are still applied in the transition of the workitems from draft to approved.

### DIFF
--- a/requirements/frozen.py3.txt
+++ b/requirements/frozen.py3.txt
@@ -130,7 +130,7 @@ docker-py==1.10.6
 docker-pycreds==0.4.0
 docutils==0.15.2
 dogpile.cache==0.7.1
-dump2polarion==0.40.0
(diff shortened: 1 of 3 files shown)


requirements/frozen.py3.txt
```diff
@@ -130,7 +130,7 @@ docker-py==1.10.6
 docker-pycreds==0.4.0
 docutils==0.15.2
 dogpile.cache==0.7.1
-dump2polarion==0.40.0
+dump2polarion==0.43.0
```
mshriver: There will be another release here with the tiny changes that support this.

mshriver: Content I wanted is in 0.44 (thanks @mkoura !!!!)

PR #9340, opened 2019-09-11
[1LP][RFR] Add test_infrastructure_hosts_navigation_after_download to test_host.py

prichard77:
## Purpose or Intent

- __Adding tests__ for FA infra_hosts to cover BZ 1738664. BZ was found in 5.11.0.22. 

### PRT Run
{{ pytest: --use-provider complete --long-running cfme/tests/infrastructure/test_host.py::test_infrastructure_hosts_navigation_after_download }}

### DIFF
--- a/cfme/tests/infrastructure/test_host.py
+++ b/cfme/tests/infrastructure/test_host.py
@@ -8,6 +8,7 @@ from wait_for import TimedOutError
 from cfme import test_requirements
 from cfme.base.credential import Credential
 from cfme.common.host_views import HostsEditView
+from cfme.common.provider_views import InfraProviderDetailsView
 from cfme.common.provider_views import ProviderNodesView
 from cfme.fixtures.provider import setup_or_skip
 from cfme.infrastructure.provider import InfraProvider
@@ -15,10 +16,12 @@ from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.infrastructure.provider.scvmm import SCVMMProvider
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.markers.env_markers.provider import ONE
+from cfme.tests.networks.test_sdn_downloads import handle_extra_tabs
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.conf import credentials
 from cfme.utils.log_validator import LogValidator
 from cfme.utils.wait import wait_for
+
 
 pytestmark = [
     pytest.mark.tier(3),
@@ -390,3 +393,30 @@ def test_infrastructure_hosts_refresh_multi(appliance, setup_provider_min_hosts,
     except TimedOutError:
         pytest.fail("Hosts were not refreshed within given time")
     assert evm_tail.validate(wait="30s")
+
+
+@test_requirements.infra_hosts
+@pytest.mark.meta(coverage=[1738664])
+@pytest.mark.parametrize(
+    "report_format", ["Download as Text", "Download as CSV", "Print or export as PDF"]
+)
+def test_infrastructure_hosts_bz_1738664(
+    appliance, setup_provider, provider, report_format
+):
+    """
+    Polarion:
+        assignee: prichard
+        casecomponent: Infra
+        caseimportance: low
+        initialEstimate: 1/3h
+    Bugzilla:
+        1738664
+
+    """
+    hosts_view = navigate_to(provider.collections.hosts, "All")
+    hosts_view.toolbar.download.item_select(report_format)
+    if report_format == "Print or export as PDF":
+        handle_extra_tabs(hosts_view)
+    hosts_view.navigation.select("Compute")
+    provider_view = provider.create_view(InfraProviderDetailsView)
+    assert provider_view.is_displayed
--- a/cfme/tests/infrastructure/test_host.py
+++ b/cfme/tests/infrastructure/test_host.py
@@ -18,6 +18,7 @@ from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.markers.env_markers.provider import ONE
 from cfme.tests.networks.test_sdn_downloads import handle_extra_tabs
 from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.blockers import BZ
 from cfme.utils.conf import credentials
 from cfme.utils.log_validator import LogValidator
 from cfme.utils.wait import wait_for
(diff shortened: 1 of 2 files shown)


cfme/tests/infrastructure/test_host.py
```diff
@@ -312,3 +315,30 @@ def _finalize():
             actual_state, power_state
         ),
     )
+
+
+@test_requirements.infra_hosts
+@pytest.mark.meta(coverage=[1738664])
```
john-dupuy: **Required** since this is a automated test case, you should use `automates` rather than `coverage`

cfme/tests/infrastructure/test_host.py
```diff
@@ -312,3 +315,30 @@ def _finalize():
...
+@pytest.mark.parametrize(
+    "report_format", ["Download as Text", "Download as CSV", "Print or export as PDF"]
+)
+def test_infrastructure_hosts_bz_1738664(
```
john-dupuy: Rather than putting the BZ in the title of the test I would call this something more descriptive like `test_infra_hosts_download_formats`

cfme/tests/infrastructure/test_host.py
```diff
@@ -312,3 +315,30 @@ def _finalize():
...
+    appliance, setup_provider, provider, report_format
+):
+    """
+    Polarion:
+        assignee: prichard
+        casecomponent: Infra
+        caseimportance: low
```
john-dupuy: Should this really be `low` case importance?

cfme/tests/infrastructure/test_host.py
```diff
@@ -312,3 +315,30 @@ def _finalize():
             actual_state, power_state
         ),
     )
+
+
+@test_requirements.infra_hosts
+@pytest.mark.meta(coverage=[1738664])
+@pytest.mark.parametrize(
+    "report_format", ["Download as Text", "Download as CSV", "Print or export as PDF"]
```
john-dupuy: **Required** I would add some ids so that the test case names aren't so long when they are generated from the parametrization. 

e.g. 
```python,
@pytest.mark.parametrize(
    "report_format", ["Download as Text", "Download as CSV", "Print or export as PDF"],
    ids=["txt", "csv", "pdf"]
)
```

cfme/tests/infrastructure/test_host.py
john-dupuy: @prichard77 Is this test case going to be backported to 5.10? If so, you should list that BZ here. If not, you must use

```python,
@pytest.mark.meta(blockers=[BZ(1738664, forced_streams=["5.10"])], coverage=[1738664])
```

cfme/tests/infrastructure/test_host.py
```diff
@@ -312,3 +315,30 @@ def _finalize():
...
+        caseimportance: low
+        initialEstimate: 1/3h
+    Bugzilla:
+        1738664
+
+    """
+    hosts_view = navigate_to(provider.collections.hosts, "All")
+    hosts_view.toolbar.download.item_select(report_format)
+    if report_format == "Print or export as PDF":
+        handle_extra_tabs(hosts_view)
```
john-dupuy: Just a thought, probably unnecessary, but should we do any cleanup of the downloaded report? I.e. delete it from the appliance after it's downloaded?

cfme/tests/infrastructure/test_host.py
```diff
@@ -312,3 +315,30 @@ def _finalize():
...
+    hosts_view.navigation.select("Compute")
```
john-dupuy: Rather than navigating in this manner, why not navigate using the provider object?
```python,
try:
    navigate_to(provider, "Details", wait_for_view=10)
except TimedOutError:
    pytest.fail("Couldn't navigate to provider details page after downloading report") 
```

Or will this break the test in some way that I'm not seeing?

dajoRH:
*CFME QE Bot*

cfme/tests/infrastructure/test_host.py
prichard77: Updated as suggested.

cfme/tests/infrastructure/test_host.py
prichard77: updated to high.

cfme/tests/infrastructure/test_host.py
prichard77: added ids.

cfme/tests/infrastructure/test_host.py
prichard77: Your last comment is correct. The bug is not reproduced unless I navigate as hosts_view.navigation.select("Compute") or something else that actually clicks on "Compute".

cfme/tests/infrastructure/test_host.py
prichard77: Will this keep PRT from running this against 5.10? Do I need to leave "coverage" here, or update to automate? And do I replace line 321 with your suggestion or append your line?

dajoRH:
*CFME QE Bot*

cfme/tests/infrastructure/test_host.py
john-dupuy: 1. Yeah, this will cause PRT to skip the test case due to the blocking BZs. Also applies to any automation/jenkins runs. 
2. sorry, you should use `automates` here, I copypastad this line
3. Replace

cfme/tests/infrastructure/test_host.py
john-dupuy: @prichard77 in that case I would put a `wait` in the `create_view` and then nest things in a `try/except` block that fails after `except TimedOutError`

dajoRH:
I detected some fixture changes in commit 8a659e72c7ee8919a4e7d629bf67bccf1bac8d33

The *local* fixture **`setup_provider_min_hosts`** is used in the following files:
- **cfme/tests/infrastructure/test_host.py**
    - *test_infrastructure_hosts_refresh_multi*

dajoRH:
Lint report for commit 8a659e72c7ee8919a4e7d629bf67bccf1bac8d33:

`cfme/tests/infrastructure/test_host.py`:
- :bangbang: Line 400:1: **E265** *block comment should start with '# '*
- :bangbang: Line 403:19: **E999** *SyntaxError: invalid syntax*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/infrastructure/test_host.py
prichard77: updated to test_infrastructure_hosts_navigation_after_download

cfme/tests/infrastructure/test_host.py
prichard77: Might be a good idea. I'll look into it.

prichard77: I looked in integration_tests for existing cleanup code for downloads. I did not find any. Then I spke with Mike briefly about this. It looks like we'll be OK without clean-up here. The downloads are not saved on the appliance or provider, but are saved on the machine where the browser is running. In this case a container that gets built each time a test run is kicked off. And these files are pretty small and will not be able to accumulate.

dajoRH:
I detected some fixture changes in commit c80507970765a310b0f6e8a4cb785a7a32ba58f2

The *local* fixture **`setup_provider_min_hosts`** is used in the following files:
- **cfme/tests/infrastructure/test_host.py**
    - *test_infrastructure_hosts_refresh_multi*

cfme/tests/infrastructure/test_host.py
john-dupuy: @prichard77 I would still expect this to be nested in a `try/except` block. The `create_view` will also need a `wait` because, otherwise the test will fail at L424

prichard77: I am having a little difficulty seeing why and where I need to put things in a try/except block. I see create_view and is_displayed used in several places without try/except. I have run the test several times against builds where it is supposed to fail (not fixed) and where it should pass. For my testing it passes and fails in the correct runs. I am not trying to argue, just to understand. Can you explain further? Is my test at risk of failing incorrectly 
, but I have not seen it because I am getting lucky with the timing or load on the system?

john-dupuy: Spoke offline, since @prichard77 is not using the `wait` kwarg, this implementation is :+1:

PR #9348, opened 2019-09-12
[1LP][RFR] Adding copy catalog item test

niyazRedhat:
{{pytest: cfme/tests/services/test_catalog_item.py::test_copy_catalog_item}}

Thanks, @digitronik  for adding copy method  #9302

### DIFF
--- a/cfme/tests/services/test_catalog_item.py
+++ b/cfme/tests/services/test_catalog_item.py
@@ -382,17 +382,18 @@ def test_service_reconfigure_in_distributed_environment():
     pass
 
 
-@pytest.mark.meta(coverage=[1678149])
-@pytest.mark.manual
+@pytest.mark.meta(automates=[1678149])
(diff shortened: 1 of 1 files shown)


cfme/tests/services/test_catalog_item.py
```diff
@@ -403,7 +404,12 @@ def test_copy_catalog_item():
             2.
             3. Able to copy catalog item
     """
-    pass
+    # copy catalog item
+    new_cat_item = generic_catalog_item.copy()
+    request.addfinalizer(new_cat_item.delete_if_exists)
+
+    # check for catalog item
+    assert new_cat_item.exists
```
mshriver: TODO: I think it would be good to have stronger assertions in this test case, instead of relying so heavily on the `copy` method in the framework.

By this, I mean asserting flash messages (here instead of in the CRUD method), asserting fields in the UI match, and parametrization against types of catalog items.

PR #9355, opened 2019-09-13
[1LP][RFR] Duplicate Infra mapping test

sshveta:
{{ pytest: cfme/tests/v2v/test_v2v_migrations_ui.py --use-provider rhv43 --use-provider vsphere67-ims --provider-limit 2 -vvvv --long-running -k test_duplicate_mapping_name }}

### DIFF
--- a/cfme/tests/v2v/test_v2v_migrations_ui.py
+++ b/cfme/tests/v2v/test_v2v_migrations_ui.py
@@ -517,5 +517,5 @@ def test_duplicate_mapping_name(appliance, mapping_data_vm_obj_mini):
     view = navigate_to(infrastructure_mapping_collection, "Add")
     view.general.name.fill(name)
     view.general.description.fill("description")
-    assert view.general.alert.read() == f"Infrastructure mapping {name} already exists"
+    view.general.flash.assert_message(f"Infrastructure mapping {name} already exists")
     view.general.cancel_btn.click()
--- a/cfme/tests/v2v/test_v2v_migrations_ui.py
+++ b/cfme/tests/v2v/test_v2v_migrations_ui.py
@@ -491,3 +491,32 @@ def test_duplicate_plan_name(appliance, mapping_data_vm_obj_mini, provider, requ
     view.general.description.fill("description")
     assert view.general.alert.read() == "Name {name} already exists".format(name=name)
     view.cancel_btn.click()
+
+
+@pytest.mark.tier(2)
+def test_duplicate_mapping_name(appliance, mapping_data_vm_obj_mini):
+    """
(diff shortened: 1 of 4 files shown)


dajoRH:
Lint report for commit 5e94f26e9d7e8290913139e6e97cd0e85fccccb0:

`cfme/tests/v2v/test_v2v_migrations_ui.py`:
- :heavy_exclamation_mark: Line 523:1: **W391** *blank line at end of file*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/v2v/test_v2v_migrations_ui.py
```diff
@@ -491,3 +491,32 @@ def _cleanup():
...
+            1. Infra map with duplicate name shows error message
+    """
+    name = mapping_data_vm_obj_mini.infra_mapping_data.get("name")
+    infrastructure_mapping_collection = appliance.collections.v2v_infra_mappings
+    # mapping created is cleaned up in mapping_data_vm_obj_mini fixture
+    view = navigate_to(infrastructure_mapping_collection, "Add")
+    view.general.name.fill(name)
+    view.general.description.fill("description")
+    assert view.general.alert.read() == (
+        "Infrastructure mapping {name} already exists".format(name=name))
```
jawatts: **Optional** Look into using f strings here

cfme/v2v/infrastructure_mapping.py
```diff
@@ -105,6 +105,7 @@ class general(InfraMappingCommonButtons):  # noqa
         plan_type = BootstrapSelect("targetProvider")
         name_help_text = Text(locator='.//div[contains(@id,"name")]/span')
         description_help_text = Text(locator='.//div[contains(@id,"description")]/span')
+        alert = Text('.//div[contains(@class, "alert")]')
```
jawatts: **Question** Why not use the `FlashMessage` class in widgetastic_patternfly? So all the methods would be available to you? Some updates would be needed to support v2v pages

john-dupuy: You could maybe inherit from that class in `wigetastic_manageiq` and overwrite the `messages` property since it looks like the locator here is a bit different. 

https://github.com/RedHatQE/widgetastic.patternfly/blob/master/src/widgetastic_patternfly/__init__.py#L252

sshveta: This alert is not same as flash message . hence flash messages cannot be used .
Screenshot from 2019-09-13 11-29-43

sshveta: Screenshot from 2019-09-13 11-30-54

PR #9356, opened 2019-09-13
[1LP][RFR] Increase timeout for creating the db

jaryn:
It seems that creating the db takes ~4 minutes to complete. It seems
that when our systems are more loaded, we are getting slightly over
4 minutes and many migration tests fail.

I tested
```
time appliance_conosle_cli --region {region} --internal \
                      --hostname localhost \
                      --username {username} --password {password} \
                      --dbname {dbname} --verbose --dbdisk {dbdisk}
```
with some configurations with these results:
```
	       CFME version
region	5.11.0.23 	5.11.0.24
---------------------------------
0	3m53.328s	4m16.664s
50	3m58.793s	3m40.494s
99	4m16.768s	3m49.669s
```

To make the test passing I also had to:
1) handle the exception risen from the connection checking code that happened when the machine was till refusing the connection attempts as there was no service listening yet.

1) increase timeout for waiting for LoginPage from default value of 120s.

    The failed tests on PRT seem to be failed because of sprout issues like for example destroyed modscope-fixture created VM.

1) add 5.10 to ignore streams to make smoke tests pass.

{{pytest: cfme/tests/test_db_migrate.py -v}}

### DIFF
--- a/cfme/tests/test_appliance.py
+++ b/cfme/tests/test_appliance.py
@@ -181,7 +181,7 @@ def test_cpu_total(appliance):
     assert int(result.output) >= 4
 
 
-@pytest.mark.meta(blockers=[BZ(1712929, forced_streams=['5.11'])])  # against RHEL
+@pytest.mark.meta(blockers=[BZ(1712929, forced_streams=['5.10', '5.11'])])  # against RHEL
 @pytest.mark.ignore_stream("upstream")
 def test_certificates_present(appliance, soft_assert):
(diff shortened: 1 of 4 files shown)


jaryn:
Actually no, it didn't. I failed to change the file in the py.line to cfme/tests/test_db_migrate.py and that's why PRT executed smoke tests and I though I have to solve failed smoke test as well.

So I have (hopefully) just fixed the py.line so the migration tests should be executed.

PR #9358, opened 2019-09-16
[WIPTEST] Update fixture use and fail condition in test_sdn_topology

mshriver:
{{ pytest: cfme/tests/networks/test_sdn_topology.py::test_topology_search --long-running --use-provider rhos13 }}


mshriver:
setup_provider_modscope is included via entity_collections - it appears something is off with RHOS for this test, because nothing is displayed in the UI.

lina-nikiforova:
The test is passed both locally and in jenkins, therefore I am concluding it was a temporary issue with the provider.

PR #9375, opened 2019-09-18
[1LP][RFR] added new test_delete_vm_on_provider_side

lina-nikiforova:
## Purpose or Intent
Automated https://bugzilla.redhat.com/show_bug.cgi?id=1592430

### PRT Run
{{ pytest: -v cfme/tests/infrastructure/test_vm_delete.py::test_delete_vm_on_provider_side }}

### DIFF
--- a/cfme/tests/infrastructure/test_vm_delete.py
+++ b/cfme/tests/infrastructure/test_vm_delete.py
@@ -1,0 +1,51 @@
+import pytest
+
+from cfme import test_requirements
+from cfme.infrastructure.provider.rhevm import RHEVMProvider
+from cfme.utils.generators import random_vm_name
+from cfme.utils.log_validator import LogValidator
+from cfme.utils.wait import TimedOutError
+from cfme.utils.wait import wait_for
+
+pytestmark = [
+    pytest.mark.provider([RHEVMProvider], scope="module"),
+    pytest.mark.usefixtures('setup_provider'),
+    test_requirements.rhev
+]
+
+
+@pytest.fixture(scope="module")
+def vm_test(provider):
+    collection = provider.appliance.provider_based_collection(provider)
+    vm_name = random_vm_name(context="del-test")
+    vm = collection.instantiate(vm_name, provider)
+    vm.create_on_provider(allow_skip="default", find_in_cfme=True)
+    vm.wait_to_appear(timeout=900, load_details=False)
+    yield vm
+
+    if vm.exists:
+        vm.cleanup_on_provider()
+
+
+@pytest.mark.rhv2
+@pytest.mark.meta(automates=[1592430])
+def test_delete_vm_on_provider_side(vm_test, provider):
+    """ Delete VM on the provider side and refresh relationships in CFME
+
+    Polarion:
+        assignee: anikifor
+        initialEstimate: 1/4h
+        casecomponent: Infra
+
+    Bugzilla:
+        1592430
+    """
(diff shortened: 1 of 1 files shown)


dajoRH:
I detected some fixture changes in commit 25e097d4ab1add0a67dd1edbdee8fe55513ebf3a

The *local* fixture **`vm_test`** is used in the following files:
- **cfme/tests/infrastructure/test_vm_delete.py**
    - *test_delete_vm_on_provider_side*

cfme/tests/infrastructure/test_vm_delete.py
```diff
@@ -0,0 +1,51 @@
...
+        assignee: anikifor
+        initialEstimate: 1/4h
+        casecomponent: Infra
+
+    Bugzilla:
+        1592430
+    """
+    logs = LogValidator("/var/www/miq/vmdb/log/evm.log", failure_patterns=[".*ERROR.*"])
+    logs.start_monitoring()
+    vm_test.mgmt.cleanup()
```
john-dupuy: Q: Why not use `cleanup_on_provider()`? They're are functionally the same

cfme/tests/infrastructure/test_vm_delete.py
```diff
@@ -0,0 +1,51 @@
...
+
+    Polarion:
+        assignee: anikifor
+        initialEstimate: 1/4h
+        casecomponent: Infra
+
+    Bugzilla:
+        1592430
+    """
+    logs = LogValidator("/var/www/miq/vmdb/log/evm.log", failure_patterns=[".*ERROR.*"])
```
john-dupuy: Q: Any specific error message we could search for here?

cfme/tests/infrastructure/test_vm_delete.py
lina-nikiforova: yes, I know that it's the same. My idea that it's more clear that delete happens on the provider's side. But looks like it was just me thinking that :) will change

cfme/tests/infrastructure/test_vm_delete.py
lina-nikiforova: No, in general, there should be no errors in the log for any reason. Even if there's some other reason for the error (not that vm was deleted), I'd much rather prefer the test to fail so I can investigate it.

dajoRH:
I detected some fixture changes in commit 2255b830b5a4eab144b09172a8bd3daec52a5042

The *local* fixture **`vm_test`** is used in the following files:
- **cfme/tests/infrastructure/test_vm_delete.py**
    - *test_delete_vm_on_provider_side*

cfme/tests/infrastructure/test_vm_delete.py
```diff
@@ -0,0 +1,51 @@
...
+    logs.start_monitoring()
+    vm_test.cleanup_on_provider()
```
jawatts: **Question** Do you think it is worth verifying that the vm was deleted?

lina-nikiforova: @jawatts don't think so... 
`cleanup` method calls `delete` and it's the wait_for for that
```
    def delete(self):
        """
        Removes the entity on the provider
        """
        self.ensure_state(VmState.STOPPED)
        self.logger.debug(' Deleting RHEV VM %s/%s', self.name, self.uuid)

        self.api.remove()

        wait_for(
            lambda: not self.exists,
            message="wait for RHEV VM '{}' deleted".format(self.uuid),
            num_sec=300
        )
        return True
```

jawatts: Yes! I thought so!

PR #9377, opened 2019-09-19
[1LP][WIP] BZ: 1720245: Adding dialog regex validation test 

niyazRedhat:
{{ pytest: cfme/tests/services/test_dialog_regex_validation_in_catalog.py -vvvv -k test_dialog_element_regex_validation }}

### DIFF
--- a/cfme/tests/services/test_dialog_regex_validation_in_catalog.py
+++ b/cfme/tests/services/test_dialog_regex_validation_in_catalog.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import fauxfactory
 import pytest
+from wait_for import TimedOutError
 from wait_for import wait_for
 
 from cfme import test_requirements
@@ -52,10 +53,15 @@ def dialog_cat_item(appliance, catalog):
     sd.delete_if_exists()
 
 
+@pytest.mark.tier(2)
+@pytest.mark.meta(automates=[1518971, 1720245])
+@pytest.mark.customer_scenario
 def test_dialog_element_regex_validation(appliance, dialog_cat_item):
     """Tests Service Dialog Elements with regex validation.
 
-    Testing BZ 1518971
+    Bugzilla:
+        1518971
+        1720245
 
     Polarion:
         assignee: nansari
@@ -66,10 +72,23 @@ def test_dialog_element_regex_validation(appliance, dialog_cat_item):
     catalog_item, ele_name = dialog_cat_item
     service_catalogs = ServiceCatalogs(appliance, catalog_item.catalog, catalog_item.name)
     view = navigate_to(service_catalogs, 'Order')
-    view.fields(ele_name).fill("!@#%&")
-    wait_for(lambda: view.submit_button.disabled, timeout=7)
-    view.fields(ele_name).fill("test_123")
-    wait_for(lambda: not view.submit_button.disabled, timeout=7)
+
+    for value in ["test_123", "!@#$%", ""]:
+        view.fields(ele_name).fill(value)
+        assert view.fields(ele_name).read() == value
+
+        if value in ["!@#$%"]:
+            expected_warning = 'Entered text should match the format: ^([a-z0-9]+_*)*[a-z0-9]+$'
+            assert view.fields(ele_name).input.warning == expected_warning
+        try:
+            wait_for(
+                lambda: view.submit_button.disabled,
+                fail_condition=False if value == "!@#$%" else True,
+                timeout=10,
+                delay=2
+            )
+        except TimedOutError:
+            pytest.fail("'{}' fail to match regex validation".format(value))
 
 
 @pytest.mark.manual
@@ -91,35 +110,6 @@ def test_dialog_text_area_element_regex_validation():
             1.
             2.
             3. Regex validation should work
-    """
-    pass
-
-
-@pytest.mark.meta(coverage=[1720245])
-@pytest.mark.manual
-@pytest.mark.ignore_stream('5.10')
-@pytest.mark.tier(2)
-def test_dialog_regex_validation_button():
-    """
-    Bugzilla:
-        1720245
-    Polarion:
-        assignee: nansari
-        casecomponent: Services
-        initialEstimate: 1/16h
-        startsin: 5.11
-        testSteps:
-            1. Add dialog with Regular Expression - "^[0-9]*$"
-            2. Create catalog and catalog item
-            3. Navigate to Order page of the service
-            4. Type "a" and it will show a message that does not satisfy the regex.
-            5. Clear the field
-        expectedResults:
-            1.
-            2.
-            3.
-            4.
-            5. Submit button should have become active when the validate field cleared
     """
     pass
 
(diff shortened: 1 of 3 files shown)


cfme/tests/services/test_dialog_regex_validation_in_catalog.py
```diff
@@ -70,6 +76,8 @@ def test_dialog_element_regex_validation(appliance, dialog_cat_item):
     wait_for(lambda: view.submit_button.disabled, timeout=7)
     view.fields(ele_name).fill("test_123")
     wait_for(lambda: not view.submit_button.disabled, timeout=7)
+    view.fields(ele_name).fill("")
+    wait_for(lambda: not view.submit_button.disabled, timeout=7)
```
valaparthvi: How about using `re` to match the pattern with the string that you use to fill? You can yield the pattern in `dialog_cat_item` along with `catalog_item` and `ele_name`  and then use it to assert if the submit_button is enabled or disabled.
Something like:
```suggestion
    for string in ["!@#$%", "test_123", ""]:
        view.fields(ele_name).fill(string)
        soft_assert(wait_for(
            lambda: view.submit_button.disabled,
            fail_condition=True if (re.match(pattern="^([a-z0-9]+_*)*[a-z0-9]+$", string=string) or string=="") else False,
            timeout=7
        ))
    wait_for(lambda: not view.submit_button.disabled, timeout=7)
```
I would've suggested parametrizing the test with various fill values("!@#$%", "test_123", "") but it will require creating service dialogs, catalogs and catalog items everytime and I don't think it's worth all the time, so you could add a `for` loop. instead.

john-dupuy: TIL about suggestions, thanks @valaparthvi

dajoRH:
Lint report for commit ca9ebfdebaeb4361b7453a7d5469ceb7e6f358f8:

`cfme/tests/services/test_dialog_regex_validation_in_catalog.py`:
- :red_circle: Line 4:1: **F401** *'time' imported but unused*
- :bangbang: Line 5:21: **E271** *multiple spaces after keyword*
- :bangbang: Line 84:20: **E126** *continuation line over-indented for hanging indent*

- :bangbang: Line 5669:1: **E302** *expected 2 blank lines, found 1*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
Lint report for commit 7e01810c7969b56f3080e93a6d8490f2981dfd2a:

`cfme/tests/services/test_dialog_regex_validation_in_catalog.py`:
- :red_circle: Line 2:1: **F401** *'time' imported but unused*
- :bangbang: Line 6:21: **E271** *multiple spaces after keyword*
- :bangbang: Line 85:20: **E126** *continuation line over-indented for hanging indent*

**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/services/test_dialog_regex_validation_in_catalog.py
```diff
@@ -66,10 +72,23 @@ def test_dialog_element_regex_validation(appliance, dialog_cat_item):
...
+        view.fields(ele_name).fill(value)
+        assert view.fields(ele_name).read() == value
+
+        if value in ["!@#$%"]:
+            expected_warning = 'Entered text should match the format: ^([a-z0-9]+_*)*[a-z0-9]+$'
+            assert view.fields(ele_name).input.warning == expected_warning
+        try:
+            wait_for(
+                lambda: view.submit_button.disabled,
+                fail_condition=False if value == "!@#$%" else True,
```
mshriver: You're setting boolean values based on the boolean result of an expression.

`fail_condition=(value != "!@#$%")`

cfme/tests/services/test_dialog_regex_validation_in_catalog.py
```diff
@@ -66,10 +72,23 @@ def test_dialog_element_regex_validation(appliance, dialog_cat_item):
     catalog_item, ele_name = dialog_cat_item
     service_catalogs = ServiceCatalogs(appliance, catalog_item.catalog, catalog_item.name)
     view = navigate_to(service_catalogs, 'Order')
-    view.fields(ele_name).fill("!@#%&")
-    wait_for(lambda: view.submit_button.disabled, timeout=7)
-    view.fields(ele_name).fill("test_123")
-    wait_for(lambda: not view.submit_button.disabled, timeout=7)
+
+    for value in ["test_123", "!@#$%", ""]:
```
mshriver: This special character string is used three times, let's name it and use that name instead.

cfme/tests/services/test_dialog_regex_validation_in_catalog.py
```diff
@@ -66,10 +72,23 @@ def test_dialog_element_regex_validation(appliance, dialog_cat_item):
...
+        view.fields(ele_name).fill(value)
+        assert view.fields(ele_name).read() == value
+
+        if value in ["!@#$%"]:
+            expected_warning = 'Entered text should match the format: ^([a-z0-9]+_*)*[a-z0-9]+$'
+            assert view.fields(ele_name).input.warning == expected_warning
```
mshriver: No validation that there is no warning for the 'good' value?

Does the warning go away if you replace the bad value with the good? This test will always be putting a good value in with no warning, then putting a bad value in with a warning, but never the opposite order.

I think, since this is a generic "regex_validation" test, it should hit more of these cases, maybe being split into a positive and negative test case function, where negative is testing multiple actions (entering bad value, modifying to another bad value, modifying to a good value, modifying back to a bad value) and validating button state/warning state between each.

At a minimum I would like to see more positive case validation in this test case, the above suggestion about test coverage could be done as a follow-up.

cfme/tests/services/test_dialog_regex_validation_in_catalog.py
```diff
@@ -66,10 +72,23 @@ def test_dialog_element_regex_validation(appliance, dialog_cat_item):
...
+        try:
+            wait_for(
+                lambda: view.submit_button.disabled,
+                fail_condition=False if value == "!@#$%" else True,
+                timeout=10,
+                delay=2
+            )
+        except TimedOutError:
+            pytest.fail("'{}' fail to match regex validation".format(value))
```
mshriver: I don't think this failure message is accurate - it failed to detect the correct submit_button state.

The failure of matching regex validation happens above in the assert statements against the warning message.

cfme/tests/services/test_dialog_regex_validation_in_catalog.py
```diff
@@ -66,10 +72,23 @@ def test_dialog_element_regex_validation(appliance, dialog_cat_item):
...
+
+    for value in ["test_123", "!@#$%", ""]:
+        view.fields(ele_name).fill(value)
+        assert view.fields(ele_name).read() == value
+
+        if value in ["!@#$%"]:
+            expected_warning = 'Entered text should match the format: ^([a-z0-9]+_*)*[a-z0-9]+$'
+            assert view.fields(ele_name).input.warning == expected_warning
+        try:
+            wait_for(
```
izapolsk: this logic looks not enough readable. could you make it more strightforward ?

dajoRH:
*CFME QE Bot*

PR #9378, opened 2019-09-19
[1LP][RFR] Automate: test_reports_service_unavailable AND add methods to ApplianceDB and FTPFileWrapper

valaparthvi:
## Purpose or Intent
- __Adding tests__
    1. `test_reports_service_unavailable`
- __Enhancement__
    1. `FTPFileWrapper` by adding `filesize` property.
    2. `ApplianceDB` by adding `restore_database` with the help of which one can restore a customer database onto the appliance.

- __Removing__
    1. Unused global fixture: `customer_db_migrate.py:customer_db_migrate`

### PRT Run

{{ pytest: cfme/tests/intelligence/reports/test_reports.py -k "test_reports_service_unavailable" --long-running }}


dajoRH:
I detected some fixture changes in commit de2cae0b3ffbc4a1e649c3b6a92cd6c713f0041d

The *local* fixture **`restore_db`** is used in the following files:
- **cfme/tests/intelligence/reports/test_reports.py**
    - *test_reports_service_unavailable*

dajoRH:
I detected some fixture changes in commit 7ab1ebc987b7f70795126092aca174cf118a0632

The *local* fixture **`restore_db`** is used in the following files:
- **cfme/tests/intelligence/reports/test_reports.py**
    - *test_reports_service_unavailable*

cfme/utils/appliance/__init__.py
```diff
@@ -453,7 +453,12 @@ def server(self):
     def _rest_api_server(self):
         shref = self.appliance.rest_api.server_info['server_href']
         results = self.appliance.rest_api.collections.servers.all
-        server, = (r for r in results if r.href == shref)
+        try:
+            server, = (r for r in results if r.href == shref)
+        except ValueError:
```
valaparthvi: This error is sometimes encountered while dealing with customer database.

cfme/utils/appliance/db.py
```diff
@@ -48,7 +48,11 @@ def address(self):
         if self.appliance.db_host:
             return self.appliance.db_host
         try:
-            db_addr = self.appliance.wait_for_host_address()
+            db_addr = (
+                self.appliance.wait_for_host_address()
```
valaparthvi: If the evmserver is down, `self_wait_for_host_address()` will take 200sec until it returns TimedOutError and `None` is assigned to `db_addr`, and since `.address` is used a lot, the test will take much much longer to execute if we have to wait for 200sec every time `.address` is called. Instead, we directly assign `None` to `db_addr` if evmserver is down.

dajoRH:
I detected some fixture changes in commit 3118c727bcef3c9146f81f3b9c8cc996f8b3740c

The *local* fixture **`restore_db`** is used in the following files:
- **cfme/tests/intelligence/reports/test_reports.py**
    - *test_reports_service_unavailable*

cfme/tests/intelligence/reports/test_reports.py
```diff
@@ -461,3 +484,40 @@ def test_reports_online_vms(appliance, setup_provider, provider, request, vm):
...
+            1. Load customer database on the appliance.
+        testSteps:
+            1. Navigate to Reports.
+            2. Navigate to `Details` page of the report with >31k rows.
+        expectedResults:
+            1. Reports must be accessible, there should be no 503 service unavailable error.
+            2. Details page must be accessible.
+    """
+    appliance = temp_appliance_preconfig
+    # 1.
```
sbulage: **Suggestion:**
I don't think comments `# 1.` required, while looking into steps, it will be clear what it is doing.

cfme/tests/intelligence/reports/test_reports.py
```diff
@@ -461,3 +484,40 @@ def test_reports_online_vms(appliance, setup_provider, provider, request, vm):
...
+    view = navigate_to(appliance.collections.reports, "All")
+    assert view.is_displayed
+    # 2.
```
sbulage: Same as above

cfme/tests/intelligence/reports/test_reports.py
valaparthvi: I don't really see any harm in keeping those comments as is, I would rather keep them Satya.

sbulage: @valaparthvi Yes, I agree with you, but I never saw such comments in our code base, Also, not sure whether it is documented anywhere :smile:

dajoRH:
I detected some fixture changes in commit 640896f0a2b95897f4a6dcb666eecad38f67416f

The *local* fixture **`restore_db`** is used in the following files:
- **cfme/tests/intelligence/reports/test_reports.py**
    - *test_reports_service_unavailable*

cfme/tests/intelligence/reports/test_reports.py
```diff
@@ -461,3 +484,40 @@ def test_reports_online_vms(appliance, setup_provider, provider, request, vm):
...
+    appliance = temp_appliance_preconfig
+
+    view = navigate_to(appliance.collections.reports, "All")
+    assert view.is_displayed
+
+    saved_report = appliance.collections.reports.instantiate(
+        type="Configuration Management",
+        subtype="Hosts",
+        menu_name="Host vLANs and vSwitches",
+    ).saved_reports.instantiate("06/17/19 11:46:59 UTC", "06/17/19 11:44:57 UTC", False)
```
john-dupuy: Q: Where are these dates taken from? And do they need to be hardcoded?

cfme/utils/appliance/__init__.py
john-dupuy: @valaparthvi Can you use

```python,
return self.server.rest_api_entity
```

Here instead? It includes the `ValueError` already. https://github.com/ManageIQ/integration_tests/blob/07175beba8c27d731a1e62d626dd6f4dd495aa84/cfme/base/__init__.py#L110-L115

Also can this `_rest_api_server` be a property?

valaparthvi: I don't think placing the ValueError code inside `rest_api_entity` would work. Afaik this is how the chain is:
`appliance.server` uses `_rest_api_server` and you will use `rest_api_entity` like `appliance.server.rest_api_entity`. We are getting the error in `_rest_api_server` so it doesn't make sense to move it to `rest_api_entity`. But your suggestion about making it property is good, I'll work on that.

dajoRH:
I detected some fixture changes in commit 662ff7f1de8208bf10c9c8236ff364becd4a8f16

The *local* fixture **`restore_db`** is used in the following files:
- **cfme/tests/intelligence/reports/test_reports.py**
    - *test_reports_service_unavailable*

john-dupuy:
Looking good but please check PRT results

valaparthvi:
Hey @john-dupuy, PRT is failing because of a change in FTP permissions, @digitronik is working on resolving that, other people are also facing the same error. You can continue reviewing the changes.

dajoRH:
*CFME QE Bot*

dajoRH:
I detected some fixture changes in commit 3b41c4c6530692e6545f05831c56b4e28ef287d3

The *local* fixture **`restore_db`** is used in the following files:
- **cfme/tests/intelligence/reports/test_reports.py**
    - *test_reports_service_unavailable*

cfme/utils/ftp.py
```diff
@@ -671,3 +671,13 @@ def download(self, target=None):
             target path
         """
         return self.client.download(self.name, target)
+
+    @property
+    def filesize(self):
+        """Return file size"""
+        # Switch to Binary mode
+        self.client.ftp.sendcmd("TYPE i")
+        size = self.client.ftp.size(self.path)
```
mshriver: I think this should be wrapped in a try:except:finally, so that the ASCII switch happens if/when the size call fails for some reason.

cfme/utils/appliance/__init__.py
mshriver: When is this hitting this condition - what are you trying to solve here?

This was explicitly raising the valueerror, because there is an issue here - the server href returned by the API is not actually found in the servers list. There should be one, and only one, server matching this. I don't think its good behavior to just fallback to the master server for this.

valaparthvi: I am hitting this condition when I restore the CU DB on the appliance.
The problem is fetching `appliance.rest_api.collections.servers.all` returns all the servers restored by the CU  DB, but `appliance.rest_api.server_info["server_href"]` still returns the old the server href. It looks like a bug, I'll file a bug, but until then can we keep it as a workaround? Or it doesn't look like a good idea?

valaparthvi: So, I found that deleting the `rest_api` cached property after restoring the database did the trick.

cfme/tests/intelligence/reports/test_reports.py
```diff
@@ -140,6 +140,29 @@ def rbac_api(appliance, request):
     )
 
 
+@pytest.fixture
+def restore_db(temp_appliance_preconfig, file_name):
+    db_file = FTPClientWrapper(cfme_data.ftpserver.entities.databases).get_file(file_name)
```
mshriver: This needs to be wrapped in a pytest.skip, and maybe some of the below calls in pytest.fail, to prevent this from resulting in an Error result when there's something amiss with the DB file fetching.

valaparthvi: **Question:** If something fails in the test setup fixture, should I use `pytest.skip` or` pytest.fail`?
When should I use `pytest.skip` and `pytest.fail`?

valaparthvi: Currently I've only used `pytest.skip` because I don't think I need to fail the test if it's setup failed. Let me know if you think otherwise.

dajoRH:
I detected some fixture changes in commit 2198da9cbcc86f94f0e1c78de3a0acc0f34ea53e

The *local* fixture **`restore_db`** is used in the following files:
- **cfme/tests/intelligence/reports/test_reports.py**
    - *test_reports_service_unavailable*

dajoRH:
I detected some fixture changes in commit 609a0820b14d06365ea29b9c236710c7c6ad1151

The *local* fixture **`restore_db`** is used in the following files:
- **cfme/tests/intelligence/reports/test_reports.py**
    - *test_reports_service_unavailable*

cfme/tests/intelligence/reports/test_reports.py
mshriver: Up to your discretion, but I encourage behavior like:

If a data lookup fails, or something necessary for the test is missing, I would skip.

If an *action* during setup fails, I would use pytest.fail, because something we expect to work doesn't.  The core reason may be because something is missing, but there's always room for nuance.

cfme/tests/intelligence/reports/test_reports.py
```diff
@@ -140,6 +141,35 @@ def rbac_api(appliance, request):
...
+        pytest.skip("Failed to fetch the file from FTP server.")
+
+    db_path = f"/tmp/{db_file.name}"
+
+    # Download the customer db on appliance
+    result = temp_appliance_preconfig.ssh_client.run_command(
+        f"curl -o {db_path} ftp://{db_file.link}"
+    )
+    if not result.success:
+        pytest.fail("Failed to download the file to the appliance.")
```
mshriver: :ok_hand:

cfme/tests/intelligence/reports/test_reports.py
valaparthvi: So I forgot to update the comment, but I used `pytest.fail` if the files fails to download on the appliance and `pytest.skip` if we fail to fetch the file from FTP server in the first step or if the downloaded file is broken.

PR #9380, opened 2019-09-19
[WIPTEST] Check flavor and security group from default and secondory

Yadnyawalkya:
## Purpose or Intent
Added security group and flavor related test to IMS

### PRT Run
{{pytest: cfme/tests/v2v/test_v2v_osp.py --use-provider osp13-ims --use-provider vsphere67-ims --provider-limit 2 -vvvv}}


Yadnyawalkya:
hey @john-dupuy, PRT failures are due to some OSP issue which @mnadeem92  is currently working on, we have tested it locally when OSP was up and it was working fine. You can begin with code review, please ignore PRT results atm, it should just work.

cfme/tests/v2v/test_v2v_osp.py
```diff
@@ -0,0 +1,105 @@
...
+        component.targets.pop()
+        component.targets.append(partial_match(provider.data.clusters[1]))
+        map_cluster = component.targets[1]
+
+    mapping = infrastructure_mapping_collection.create(**mapping_data)
+    src_vm_obj = mapping_data_vm_obj_single_datastore.vm_list[0]
+
+    # Selecting default flavor and security group selection
+    map_flavor = provider.data.flavors[0]
+    map_security_group = provider.data.security_groups.admin[0]
```
sshveta: These are not used anywhere . These values are overridden by new values in line 75 and 77 . You can remove these

cfme/tests/v2v/test_v2v_osp.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+    infrastructure_mapping_collection = appliance.collections.v2v_infra_mappings
+    mapping_data = mapping_data_vm_obj_single_datastore.infra_mapping_data
+
+    if project == "secondary":
+        # Changing target cluster (openstack project) to other than default
+        component = mapping_data["clusters"][0].targets
+        component.targets.pop()
+        component.targets.append(partial_match(provider.data.clusters[1]))
+        map_cluster = component.targets[1]
```
sshveta: I don't see this cluster being used in creating migration plan .

Yadnyawalkya: We are updating attribute of object from `mapping_data` so it should just work without creating new one.

cfme/tests/v2v/test_v2v_osp.py
Yadnyawalkya: No. Actually this are default values and we are only overridding those values when we have that specific `attribute` set by parameterization. Final permutation should be - 

[default cluster] + [flavor] + sec group
[default cluster] + flavor + [sec group]
[secondary cluster] + [flavor] + sec group
[secondary cluster] + flavor + [sec group]

Here values in square brakets are actual selection.

sshveta:
This was automated by another PR

PR #9384, opened 2019-09-20
[1LP][RFR] New Test: Testing domain version changed after upgrading appliance

ganeshhubale:
## Purpose or Intent
- After upgrading appliance from lower version to higher version.The change in datastore versions logged in evm logs.

### PRT Run
{{ pytest: cfme/tests/automate/test_domain.py::test_redhat_domain_sync_after_upgrade --long-running -vv }}

### DIFF
--- a/cfme/tests/automate/test_domain.py
+++ b/cfme/tests/automate/test_domain.py
@@ -1,3 +1,5 @@
+import os
+
 import fauxfactory
 import pytest
 
@@ -14,6 +16,8 @@ from cfme.rest.gen_data import tenants as _tenants
 from cfme.rest.gen_data import users as _users
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.blockers import BZ
+from cfme.utils.conf import cfme_data
+from cfme.utils.ftp import FTPClientWrapper
 from cfme.utils.log_validator import LogValidator
 from cfme.utils.update import update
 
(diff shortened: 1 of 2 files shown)


dajoRH:
*CFME QE Bot*

cfme/tests/automate/test_domain.py
```diff
@@ -514,3 +516,57 @@ def test_automate_restrict_domain_crud(user, custom_instance):
...
+    # Download the customer db on appliance
+    result = temp_appliance_preconfig.ssh_client.run_command(
+        f"curl -o {db_path} ftp://{db_file.link}"
+    )
+    assert result.success
+
+    def _check_file_size(file_path, expected_size):
+        return temp_appliance_preconfig.ssh_client.run_command(
+            f"stat {file_path} | grep {expected_size}"
+        ).success
```
john-dupuy: _Optional_ since this method is small I don't think it needs to be in a nested function. But it's a minor point, so feel free to ignore me :smile:

cfme/tests/automate/test_domain.py
```diff
@@ -514,3 +516,57 @@ def test_automate_restrict_domain_crud(user, custom_instance):
...
+            3. Check last_startup.txt file
+        expectedResults:
+            1.
+            2. You should find this string in logs: RedHat domain version on disk differs from db
+               version
+            3. You should find this string in file: RedHat domain version on disk differs from db
+               version
+    """
+    db_file = FTPClientWrapper(cfme_data.ftpserver.entities.databases).get_file(file_name)
+    db_path = f"/tmp/{db_file.name}"
```
izapolsk: REQUIRED: please don't concatenate paths using string. there is os.path module for that.
in this particular case please use `os.path.join`

cfme/tests/automate/test_domain.py
```diff
@@ -514,3 +516,57 @@ def test_automate_restrict_domain_crud(user, custom_instance):
...
+    )
+    assert result.success
+
+    def _check_file_size(file_path, expected_size):
+        return temp_appliance_preconfig.ssh_client.run_command(
+            f"stat {file_path} | grep {expected_size}"
+        ).success
+
+    # assert the whole database was downloaded to make sure we do not test with broken database
+    assert _check_file_size(db_path, db_file.filesize)
```
izapolsk: 1. preceeding call won't finish until curl is over
2. if you want and need to check that original file and copied file are the same, please use md5sum or crc32 instead.
3. I'm not sure I understand what is  above function for since you run it only once

cfme/tests/automate/test_domain.py
```diff
@@ -514,3 +516,57 @@ def test_automate_restrict_domain_crud(user, custom_instance):
...
+
+    is_major = True if temp_appliance_preconfig.version > "5.11" else False
```
izapolsk: REQUIRED: bool(temp_appliance_preconfig.version > "5.11")

cfme/tests/automate/test_domain.py
valaparthvi: Hey @izapolsk I had originally written this piece of code for one of my tests. I wanted to make sure the whole file was downloaded, so I wrote the function, I wasn't aware of md5sum or crc32 checks. I'll see if I can implement them.

izapolsk: honestly, I doubt we need such check at all.

ganeshhubale: Discussed with @valaparthvi and as per your suggestion. I have removed the check.
Thank you :)

PR #9389, opened 2019-09-23
[1LP][RFR] New Test: Testing import of blank file 

ganeshhubale:
## Purpose or Intent
- Testing importing of blank file for datastore and dialog

### PRT Run
{{ pytest: cfme/tests/automate/test_file_import.py::test_upload_blank_file -vv }}

### DIFF
--- a/cfme/automate/import_export.py
+++ b/cfme/automate/import_export.py
@@ -36,6 +36,8 @@ class AutomateImportExportBaseView(BaseLoggedInPage):
 
 
 class AutomateImportExportView(AutomateImportExportBaseView):
+    error_flash = Text("//div[contains(@class, 'import-flash-message')]")
+
     @View.nested
     class import_file(View):  # noqa
         upload_file = FileInput(id="upload_file")
--- a/cfme/automate/import_export.py
+++ b/cfme/automate/import_export.py
@@ -16,6 +16,7 @@ from cfme.modeling.base import BaseEntity
 from cfme.utils.appliance.implementations.ui import CFMENavigateStep
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.appliance.implementations.ui import navigator
+from widgetastic_manageiq import V2VFlashMessages
 
 
 class AutomateImportExportBaseView(BaseLoggedInPage):
@@ -36,7 +37,7 @@ class AutomateImportExportBaseView(BaseLoggedInPage):
 
 
 class AutomateImportExportView(AutomateImportExportBaseView):
-    error_flash = Text("//div[contains(@class, 'import-flash-message')]")
+    flash = V2VFlashMessages(locator="div.import-flash-message")
 
     @View.nested
     class import_file(View):  # noqa
--- a/cfme/tests/automate/test_file_import.py
+++ b/cfme/tests/automate/test_file_import.py
@@ -41,11 +41,11 @@ def test_domain_import_file(import_datastore, import_data):
 
 @pytest.mark.tier(2)
 @pytest.mark.meta(automates=[1720611])
-@pytest.mark.parametrize("files", ["datastore_blank.zip", "dialog_blank.yml"],
+@pytest.mark.parametrize("upload_file", ["datastore_blank.zip", "dialog_blank.yml"],
                          ids=["datastore", "dialog"])
-@pytest.mark.uncollectif(lambda files: files == "dialog_blank.yml" and
+@pytest.mark.uncollectif(lambda upload_file: upload_file == "dialog_blank.yml" and
                          BZ(1720611, forced_streams=['5.10']).blocks)
-def test_upload_blank_file(appliance, files):
+def test_upload_blank_file(appliance, upload_file):
     """
     Bugzilla:
         1720611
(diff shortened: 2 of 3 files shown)


cfme/tests/automate/test_file_import.py
```diff
@@ -31,3 +37,55 @@ def test_domain_import_file(import_datastore, import_data):
...
     """
     assert import_datastore.exists
+
+
+@pytest.mark.tier(2)
+@pytest.mark.meta(automates=[1720611])
+@pytest.mark.parametrize("files", ["datastore_blank.zip", "dialog_blank.yml"],
+                         ids=["datastore", "dialog"])
+@pytest.mark.uncollectif(lambda files: files == "dialog_blank.yml" and
+                         BZ(1720611, forced_streams=['5.10']).blocks)
```
lina-nikiforova: I don't see that https://bugzilla.redhat.com/show_bug.cgi?id=1720611 will be fixed, there's no clone to 5.10, so you can always uncollect based on the appliance version.

cfme/tests/automate/test_file_import.py
```diff
@@ -31,3 +37,55 @@ def test_domain_import_file(import_datastore, import_data):
...
+    else:
+        # Import datastore file to appliance
+        datastore = appliance.collections.automate_import_exports.instantiate(
+            import_type="file", file_path=file_path
+        )
+        view = navigate_to(datastore.parent, "All")
+        with LogValidator("/var/www/miq/vmdb/log/production.log",
+                          failure_patterns=[".*FATAL.*"]).waiting(timeout=120):
+            view.import_file.upload_file.fill(datastore.file_path)
+            view.import_file.upload.click()
```
lina-nikiforova: @ganeshhubale  Should there be an assert for the flash message?

cfme/tests/automate/test_file_import.py
ganeshhubale: This BZ might get back-ported to 5.10. Hence keeping it as it is.

cfme/automate/import_export.py
```diff
@@ -36,6 +36,8 @@ def is_displayed(self):
 
 
 class AutomateImportExportView(AutomateImportExportBaseView):
+    flash = Text("//div[contains(@class, 'import-flash-message')]")
```
john-dupuy: Q: can this use the `FlashMessages` widget? May have to subclass to overwrite locators/xpath.

ganeshhubale: Hey @john-dupuy 
- I agree on using ```FlashMessage``` or ```FlashMessages``` to assert the error message. I tried these widgets with locator ```//div[contains(@class, 'import-flash-message')]``` but it does not read the message.
- So I have changed the name from flash to error_flash for this widget to avoid override of flash widget from BaseClass.
- Please let me know your view.

john-dupuy: @ganeshhubale I would have to check to make sure, but I think you can subclass and modify the `messages` xpath locator @sshveta did something similar with `V2VFlashMessages`

https://github.com/ManageIQ/integration_tests/blob/8ae591f505e95cb0291258fc47987c4ec33efb66/widgetastic_manageiq/__init__.py#L5864-L5876

cfme/tests/automate/test_file_import.py
john-dupuy: resolving based on L92 below

cfme/tests/automate/test_file_import.py
```diff
@@ -31,3 +37,56 @@ def test_domain_import_file(import_datastore, import_data):
...
+            view = navigate_to(import_export, "DialogImportExport")
+            view.upload_file.fill(file_path)
+            view.upload.click()
+            view.flash.assert_message('Error: the uploaded file is blank')
+    else:
+        # Import datastore file to appliance
+        datastore = appliance.collections.automate_import_exports.instantiate(
+            import_type="file", file_path=file_path
+        )
+        view = navigate_to(datastore.parent, "All")
```
john-dupuy: _Optional_ consider navigating via `appliance.collections.automate_import_exports` instead, it is a bit more explicit.

ganeshhubale: Sounds Good!

cfme/tests/automate/test_file_import.py
```diff
@@ -31,3 +37,56 @@ def test_domain_import_file(import_datastore, import_data):
...
+        # Import datastore file to appliance
+        datastore = appliance.collections.automate_import_exports.instantiate(
+            import_type="file", file_path=file_path
+        )
+        view = navigate_to(appliance.collections.automate_import_exports, "All")
+        with LogValidator("/var/www/miq/vmdb/log/production.log",
+                          failure_patterns=[".*FATAL.*"]).waiting(timeout=120):
+            view.import_file.upload_file.fill(datastore.file_path)
+            view.import_file.upload.click()
+            assert view.flash.assert_message("Error: import processing failed: domain: *")
```
izapolsk: I doubt you need assert here since you call assert_message

cfme/tests/automate/test_file_import.py
```diff
@@ -31,3 +37,56 @@ def test_domain_import_file(import_datastore, import_data):
             3. Import should work. Check imported or not.
     """
     assert import_datastore.exists
+
+
+@pytest.mark.tier(2)
+@pytest.mark.meta(automates=[1720611])
+@pytest.mark.parametrize("files", ["datastore_blank.zip", "dialog_blank.yml"],
```
izapolsk: sorry for nitpicking but why files ? as I see you pass files one by one. so, it is definitely "file". and since you are testing upload. I'd call it upload_file or blank_file or test_file.

PR #9391, opened 2019-09-23
[1LP][RFR] Automate test case for BZ 1650559

john-dupuy:
Automating test case for https://bugzilla.redhat.com/show_bug.cgi?id=1650559

Note that this test case doesn't require you to create the Ansible Playbook type generic object button, just that the fields are valid. A future test case will create these buttons.

This test case required some framework enhancements to allow for different types of generic object buttons. 

{{ pytest: --long-running cfme/tests/automate/generic_objects/buttons/test_generic_class_custom_button.py::test_ansible_playbook_generic_object_button }}

### DIFF
--- a/cfme/generic_objects/definition/button_groups.py
+++ b/cfme/generic_objects/definition/button_groups.py
@@ -25,8 +25,8 @@ class GenericObjectButton(BaseEntity, Updateable):
 
     name = attr.ib()
     description = attr.ib()
-    image = attr.ib()
     request = attr.ib()
+    image = attr.ib(default='fa-home')
     button_type = attr.ib(default='Default')
     display = attr.ib(default=True)
     dialog = attr.ib(default=None)
@@ -37,6 +37,11 @@ class GenericObjectButton(BaseEntity, Updateable):
     attributes = attr.ib(default=None)
     role = attr.ib(default=None)
     button_group = attr.ib(default=None)
+
+    # extra fields for ansible playbook type buttons
+    playbook_cat_item = attr.ib(default=None)
+    inventory = attr.ib(default=None)
+    hosts = attr.ib(default=None)
 
     def delete(self, cancel=False):
         """Delete generic object button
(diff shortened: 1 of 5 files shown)


cfme/generic_objects/definition/button_groups.py
```diff
@@ -25,8 +25,8 @@ class GenericObjectButton(BaseEntity, Updateable):
 
     name = attr.ib()
     description = attr.ib()
-    image = attr.ib()
     request = attr.ib()
+    image = attr.ib(default='fa-home')
```
mshriver: Consider defining this in wt.pf.PFIcons

cfme/generic_objects/definition/definition_views.py
```diff
@@ -192,6 +194,7 @@ class ButtonParameterForm(ParametersForm):
 
 class GenericObjectAddEditButtonView(GenericObjectDefinitionView):
     title = Text('#explorer_title_text')
+    form = ConditionalSwitchableView(reference="button_type")
```
mshriver: :tada:

cfme/generic_objects/definition/ui.py
```diff
@@ -112,12 +112,22 @@ def generic_objects(self):
     return self.collections.generic_objects
 
 
+@MiqImplementationContext.external_for(GenericObjectDefinition.generic_object_buttons.getter, ViaUI)
+def generic_object_buttons(self):
+    return self.collections.generic_object_buttons
```
mshriver: This is consistent with the property above, but I think you should take a hard look at the nuance here - this property, for ViaUI, is actually just returning something attached to the model of the generic object defintion - its not actually returning what generic object buttons are attached to the generic object definition _as shown by the CFME UI_.

john-dupuy: Yeah I thought it was weird to have it defined in this way, but wanted to remain consistent with existing framework. I will probably change in the near future!

PR #9392, opened 2019-09-23
[1LP][RFR] Optimize TestInstanceRESTAPI to execute faster

valaparthvi:
## Purpose or Intent
- __Updating tests__ 
    1. Optimize `TestInstanceRESTAPI` to use REST to wait for power changes, use `assert_reponse` and other minor changes.

Note: Dependent on #9387.
### PRT Run
{{ pytest: cfme/tests/cloud/test_instance_power_control.py -k "TestInstanceRESTAPI" --use-provider=complete --long-running -vvv }}

### DIFF
--- a/cfme/tests/cloud/test_instance_power_control.py
+++ b/cfme/tests/cloud/test_instance_power_control.py
@@ -97,7 +97,7 @@ def wait_for_ui_state_refresh(instance, provider, state_change_time, timeout=900
 def wait_for_pwr_state_change(instance, state_change_time, timeout=720):
     return wait_for(
         lambda: instance.rest_api_entity.state_changed_on != state_change_time,
-        num_sec=720,
+        num_sec=timeout,
         delay=30,
         message='Waiting for instance state refresh'
     ).out
@@ -681,7 +681,8 @@ class TestInstanceRESTAPI(object):
 
         # assert and wait until the power state change is reflected in REST
         assert testing_instance.wait_for_power_state_change_rest(
-            desired_state=testing_instance.STATE_SUSPENDED
+            desired_state=testing_instance.STATE_SUSPENDED,
+            delay=15
         )
         # check if the power state change is reflected on UI and provider
         wait_for_instance_state(soft_assert, testing_instance, state="suspended")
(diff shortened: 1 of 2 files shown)


valaparthvi:
510 tests are failing due to some reason, I'm not sure which, but the tests are passing locally.

cfme/tests/cloud/test_instance_power_control.py
```diff
@@ -94,13 +95,12 @@ def _fail_func():
...
-    def _wait_for():
-        vm = instance.rest_api_entity
-        if vm.state_changed_on != state_change_time:
-            return True
-
     return wait_for(
-        _wait_for, num_sec=timeout, delay=30, message='Waiting for instance state refresh')
+        lambda: instance.rest_api_entity.state_changed_on != state_change_time,
+        num_sec=720,
+        delay=30,
```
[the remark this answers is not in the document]
valaparthvi: Ay, good catch!

cfme/tests/cloud/test_instance_power_control.py
```diff
@@ -94,13 +95,12 @@ def _fail_func():
 
 
 def wait_for_pwr_state_change(instance, state_change_time, timeout=720):
```
john-dupuy: Thoughts on this being a method of `BaseEntity`?

john-dupuy: Also do you think the distinction between this method and `wait_for_power_state_change_rest` could be made a bit clearer by calling this method `wait_for_power_state_refresh`?

valaparthvi: > Thoughts on this being a method of BaseEntity?

Probably not, because we already have methods to check power state change viaUI and viaREST, we rarely have the need to check `last_changed_on` property.

cfme/tests/cloud/test_instance_power_control.py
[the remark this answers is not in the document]
izapolsk: num_sec accepts only int. there will be error if some test passes string. please fix this upfront.

cfme/tests/cloud/test_instance_power_control.py
```diff
@@ -555,13 +539,19 @@ def test_stop(self, provider, testing_instance, ensure_vm_running,
             caseimportance: high
             initialEstimate: 1/4h
         """
-        testing_instance.wait_for_instance_state_change(desired_state=testing_instance.STATE_ON)
+        testing_instance.wait_for_power_state_change_rest(desired_state=testing_instance.STATE_ON)
```
izapolsk: could you make such methods to use sentaku instead ? 
I don't insist to make such changes in this PR.

valaparthvi: Added a TODO for this.

PR #9400, opened 2019-09-24
[1LP][RFR] Automate: test_refresh_with_empty_iot_hub_azure

psimovec:
:heavy_check_mark:  requires https://github.com/ManageIQ/wrapanapi/pull/419
:stop_sign:  5.10.11 blocker found by test run https://bugzilla.redhat.com/show_bug.cgi?id=1756984 

{{ pytest: cfme/tests/cloud/test_providers.py::test_refresh_with_empty_iot_hub_azure -v --use-provider azure }}

### DIFF
--- a/cfme/tests/cloud/test_providers.py
+++ b/cfme/tests/cloud/test_providers.py
@@ -27,6 +27,7 @@ from cfme.utils import appliance
 from cfme.utils import conf
 from cfme.utils import ssh
 from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.blockers import BZ
 from cfme.utils.conf import credentials
 from cfme.utils.generators import random_vm_name
 from cfme.utils.log_validator import LogValidator
@@ -542,7 +543,7 @@ def test_azure_multiple_subscription(appliance, request, soft_assert):
 
 @pytest.mark.tier(3)
 @test_requirements.azure
-@pytest.mark.meta(automates=[1495318])
+@pytest.mark.meta(automates=[1495318], blockers=[BZ(1756984)])
 @pytest.mark.provider([AzureProvider], scope="function", override=True, selector=ONE)
 def test_refresh_with_empty_iot_hub_azure(request, provider, setup_provider):
     """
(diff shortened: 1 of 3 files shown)


cfme/tests/cloud/test_providers.py
```diff
@@ -538,6 +539,42 @@ def test_azure_multiple_subscription(appliance, request, soft_assert):
...
+               create an IoT Hub in Azure (using free tier pricing is good enough):
+               $ az iot hub create --name rmanes-iothub --resource-group iot_rg
+        testSteps:
+            1. refresh azure provider
+        expectedResults:
+            1. no errors found in logs
+    Bugzilla:
+        1495318
+    """
+    result = LogValidator(
```
lina-nikiforova: This thing fits one line nicely :)
`result = LogValidator("/var/www/miq/vmdb/log/evm.log", failure_patterns=[r".*ERROR.*"])`

cfme/tests/cloud/test_providers.py
```diff
@@ -538,6 +539,42 @@ def test_azure_multiple_subscription(appliance, request, soft_assert):
                                                                                      prov_b[0]))
 
 
+@pytest.mark.tier(1)
```
lina-nikiforova: should be 3, as 1 is the most important and this case has `caseimportance: low`

mshriver: Case Importance and tier are separate metadata.

Case importance is of low/medium/high/critical, and represents the how frequently the test case should be executed, along with how important it is when it fails.

Tier markers, integers 0, 1, 2, 3, equate to Case Level, which represents the test case layer from unit, function, system, to integration.

cfme/tests/cloud/test_providers.py
```diff
@@ -538,6 +540,39 @@ def test_azure_multiple_subscription(appliance, request, soft_assert):
...
+        expectedResults:
+            1. no errors found in logs
+    Bugzilla:
+        1495318
+    """
+    result = LogValidator("/var/www/miq/vmdb/log/evm.log", failure_patterns=[r".*ERROR.*"])
+    result.start_monitoring()
+    azure = provider.mgmt
+    if not azure.has_iothub():
+        azure.create_iothub("potatoiothub")
```
mshriver: This needs to be randomized. If we're running multiple instances of the test framework concurrently, they could collide.

cfme/tests/cloud/test_providers.py
```diff
@@ -538,6 +540,39 @@ def test_azure_multiple_subscription(appliance, request, soft_assert):
...
+        request.addfinalizer(lambda: azure.delete_iothub("potatoiothub"))
```

cfme/tests/cloud/test_providers.py
lina-nikiforova: just a note: they could collide even in that case too. In our azure we can't have more than 1 iothub (1 is free).

PR #9402, opened 2019-09-24
[NOTEST][WIP]initial commit with vault+sprout integration via dynaconf

kedark3:
## Purpose or Intent
- __Enhancement__ for integrating Vault with Sprout using dynaconf. This will allow us to get the provider object that fetches creds from Vault using dynaconf. This is still a WIP as we need to upgrade our Vault server first before we could test it. Although I tested it with a different vault server and it worked alright.


sprout/vault/vault.py
```diff
@@ -0,0 +1,53 @@
...
+# to create role_id and secret_id which together lets you authenticate to AppRole
+vault_approle_token = os.environ.get("VAULT_APPROLE_TOKEN", None)
+vault_url = os.environ.get("VAULT_URL_FOR_DYNACONF", None)
+
+
+def _login_and_renew_token(url, token):
+    """Log into Vault, renew the token, and return the Vault client"""
+    vault = Client(url=url, token=token, verify=False)
+    if not vault.is_authenticated():
+        return None
```
izapolsk: shouldn't it login in this case ?

sprout/vault/vault.py
```diff
@@ -0,0 +1,53 @@
+import os
+
+from dynaconf import LazySettings
+from hvac import Client
+
+VAULT_APPROLE = 'cfme-qe-infra'
```
izapolsk: I'd turn all this stuff into class

sprout/vault/approle_setup.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+# This file briefly describes the steps and can execute those as well - to create Approle Token.
+# make sure to have exported VAULT_ADDR=https://addr:port
+# and VAULT_SKIP_VERIFY=true to disable ssl verification
+
+echo 'login with kerberos - make sure you are admin by reading listed policies'
+vault login -method=ldap -tls-skip-verify=true username=<user>
```
izapolsk: shouldn't it ask for this value or look for it in env variables ?

sprout/appliances/utils/providers.py
```diff
@@ -0,0 +1,71 @@
...
+
+import six
+
+from cfme.common.provider import all_types
+from cfme.exceptions import UnknownProviderType
+from cfme.utils import conf
+from cfme.utils.log import logger
+from sprout.vault.vault import settings
+
+providers_data = conf.cfme_data.get("management_systems", {})
```
izapolsk: sprout is deployed along with test framework. do we really need to add it's own providers and vault files ?

PR #9407, opened 2019-09-25
[1LP][RFR] Removed ignore stream and updated global fixture

ganeshhubale:
## Purpose or Intent
- BZ(1700524) is fixed for 5.10.11.0. So need to remove ```ignore_stream```.
- Improved length of the instance and method names created in custom instance global fixture.
### PRT Run
{{ pytest: cfme/tests/automate/test_common_methods.py::test_service_retirement_from_automate_method --use-template-cache -qsvvv }}

### DIFF
--- a/cfme/fixtures/automate.py
+++ b/cfme/fixtures/automate.py
@@ -61,11 +61,11 @@ def custom_instance(request_cls):
     type of automate method."""
     def method(ruby_code):
         meth = request_cls.methods.create(
-            name=fauxfactory.gen_alphanumeric(start="meth_", length=6), script=ruby_code
+            name=fauxfactory.gen_alphanumeric(start="meth_", length=10), script=ruby_code
         )
 
         instance = request_cls.instances.create(
-            name=fauxfactory.gen_alphanumeric(start="inst_", length=6),
+            name=fauxfactory.gen_alphanumeric(start="inst_", length=10),
             fields={"meth1": {"value": meth.name}},
         )
         return instance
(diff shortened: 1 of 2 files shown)


cfme/tests/automate/test_common_methods.py
```diff
@@ -189,11 +189,11 @@ def test_miq_password_decrypt(klass):
 
 @pytest.mark.tier(1)
 @pytest.mark.meta(automates=[1700524])
-@pytest.mark.ignore_stream("5.10")
 def test_service_retirement_from_automate_method(request, generic_catalog_item, custom_instance):
```
valaparthvi: There is a customer portal case assigned to the BZs you mentioned, can you confirm if this is a customer case, and if so, can you add `@pytest.mark.customer_scenario`?

cfme/fixtures/automate.py
```diff
@@ -61,11 +61,11 @@ def custom_instance(request_cls):
     type of automate method."""
     def method(ruby_code):
         meth = request_cls.methods.create(
-            name=fauxfactory.gen_alphanumeric(start="meth_", length=6), script=ruby_code
+            name=fauxfactory.gen_alphanumeric(start="meth_", length=10), script=ruby_code
```
valaparthvi: Good one!

john-dupuy:
Moving to 1LP as https://github.com/ManageIQ/integration_tests/pull/9407#discussion_r328588072 has been addressed

PR #9415, opened 2019-09-26
[1LP][WIPTEST] fix ansible repo enable check

digitronik:

## Purpose or Intent
- Fix ansible 5.10 problem as it took time to enable ansible
```
DropdownDisabled
b'Dropdown "Configuration" is not enabled'
```
- Now if all `items` in `Dropdown` disabled then over all Dropdown disabled.
 

### PRT Run

{{pytest: cfme/tests/ansible/test_embedded_ansible_crud.py}}

### DIFF
--- a/cfme/ansible/repositories.py
+++ b/cfme/ansible/repositories.py
@@ -256,10 +256,13 @@ class RepositoryCollection(BaseCollection):
             delete_on_update=delete_on_update,
             update_on_launch=update_on_launch)
 
-        wait_for(lambda: repository.exists,
-                 fail_func=repo_list_page.browser.selenium.refresh,
-                 delay=5,
-                 timeout=900)
+        wait_for(
+            lambda: repository.exists,
+            fail_func=repo_list_page.browser.refresh,
+            delay=5,
+            timeout=1080,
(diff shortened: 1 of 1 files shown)


digitronik:
Assertion error in 5.10 not related to code changes
```
E       AssertionError: assert 1 == 2
E        +  where 2 = len(('setup-2019-09-26-07:26:15.log', 'profile'))
```

cfme/ansible/repositories.py
```diff
@@ -256,10 +256,13 @@ def create(self, name, url, description=None, scm_credentials=None, scm_branch=N
             delete_on_update=delete_on_update,
             update_on_launch=update_on_launch)
 
-        wait_for(lambda: repository.exists,
-                 fail_func=repo_list_page.browser.selenium.refresh,
-                 delay=5,
-                 timeout=900)
+        wait_for(
+            lambda: repository.exists,
+            fail_func=repo_list_page.selenium.refresh,
```
mshriver: Why is selenium refresh being called here, refresh should be called through browser, like it is in the view below.

I would also keep the delay at 5s.

digitronik: @mshriver  done

PR #9417, opened 2019-09-26
[WIP] Adding snapshot deletion check in test_migration_post_attribute

mnadeem92:

Adding snapshot deletion check after migration.

{{pytest: cfme/tests/v2v/test_post_migrations.py -k "test_migration_post_attribute" --use-provider osp13-ims --use-provider vsphere67-ims --provider-limit 2 -vvvv}}


cfme/tests/v2v/test_post_migrations.py
```diff
@@ -57,6 +58,22 @@ def test_migration_post_attribute(appliance, provider, mapping_data_vm_obj_mini,
...
     source_cpu, source_socket, source_core, source_memory = re.findall(r"\d+", summary)
+    snap_name = '{}_snap_'.format(src_vm_obj)
+    snap_desc = 'v2v_snap_test'
+    snapshots_before = src_vm_obj.snapshots.count
+
+    src_vm_obj.snapshots.action.create(name=snap_name + '1',
+                                       description=snap_desc, memory=False,)
+    src_vm_obj.snapshots.action.create(name=snap_name + '2',
+                                       description=snap_desc, memory=False,)
+    src_vm_obj.snapshots.action.create(name=snap_name + '3',
```
sshveta: This can be done inside a for loop 
   

         for x in range(0, 2):
             src_vm_obj.snapshots.action.create(name=snap_name + x,
                                       description=snap_desc, memory=False,)

cfme/tests/v2v/test_post_migrations.py
```diff
@@ -57,6 +58,19 @@ def test_migration_post_attribute(appliance, provider, mapping_data_vm_obj_mini,
     source_view = navigate_to(src_vm_obj, "Details")
     summary = source_view.entities.summary("Properties").get_text_of("Container")
     source_cpu, source_socket, source_core, source_memory = re.findall(r"\d+", summary)
+    snap_name = '{}_snap_'.format(src_vm_obj)
+    snap_desc = 'v2v_snap_test'
+    snapshots_before = src_vm_obj.total_snapshots
+
+    for x in range(3):
+        src_vm_obj.rest_api_entity.snapshots.action.create(name=snap_name + str(x),
+                                                           description=snap_desc, memory=False,)
```
sshveta: Suggestion : Remove line 61 and 62 

        for x in range(3):
            src_vm_obj.rest_api_entity.snapshots.action.create(name=f"{src_vm_obj}_snap_{x}",
                                                               description="v2v_snap_test",
                                                               memory=False,)

cfme/tests/v2v/test_post_migrations.py
```diff
@@ -57,6 +58,19 @@ def test_migration_post_attribute(appliance, provider, mapping_data_vm_obj_mini,
...
+    wait_for(
+        lambda: int(src_vm_obj.total_snapshots) == int(snapshots_before) + 3,
```
sshveta: Does this work for both OSP and RHV ?

cfme/tests/v2v/test_post_migrations.py
```diff
@@ -102,3 +116,7 @@ def test_migration_post_attribute(appliance, provider, mapping_data_vm_obj_mini,
     soft_assert(source_core == target_core)
     # Test7: memory on source and target vm
     soft_assert(source_memory == target_memory)
+    # Test8: check snapshot should be destroyed after migration
+    expected_count = 1 if provider.one_of(RHEVMProvider) else 0
```
sshveta: Why are we expecting 1 snapshot for RHV ?

cfme/tests/v2v/test_post_migrations.py
mnadeem92: Here the src_vm_obj will always be vpshere vm object and it will work with both OSP and RHV.

cfme/tests/v2v/test_post_migrations.py
mnadeem92: After migration of a VM to RHV, there is always a snapshot gets created to store its current status, however the same is not true in case of OSP.

mnadeem92: Please have a look to below comment
>> https://bugzilla.redhat.com/show_bug.cgi?id=1744563#c5

cfme/tests/v2v/test_post_migrations.py
```diff
@@ -57,6 +61,18 @@ def test_migration_post_attribute(appliance, provider, mapping_data_vm_obj_mini,
...
+    snapshots_before = src_vm_obj.total_snapshots
+
+    for x in range(3):
+        src_vm_obj.rest_api_entity.snapshots.action.create(name=f"{src_vm_obj.name}_snap_{x}",
+                                                           description='v2v_snap_test',
+                                                           memory=False,)
+    wait_for(
+        lambda: int(src_vm_obj.total_snapshots) == int(snapshots_before) + 3,
+        num_sec=800,
+        message="wait for snapshot appear",
```
john-dupuy: small grammar comment: `waiting for snapshots to appear`

cfme/tests/v2v/test_post_migrations.py
```diff
@@ -57,6 +61,18 @@ def test_migration_post_attribute(appliance, provider, mapping_data_vm_obj_mini,
     source_view = navigate_to(src_vm_obj, "Details")
     summary = source_view.entities.summary("Properties").get_text_of("Container")
     source_cpu, source_socket, source_core, source_memory = re.findall(r"\d+", summary)
+    snapshots_before = src_vm_obj.total_snapshots
+
+    for x in range(3):
```
john-dupuy: Do we really need to make 3 snapshots? Why not just 1 or 2?

cfme/tests/v2v/test_post_migrations.py
```diff
@@ -102,3 +118,7 @@ def test_migration_post_attribute(appliance, provider, mapping_data_vm_obj_mini,
     soft_assert(source_core == target_core)
     # Test7: memory on source and target vm
     soft_assert(source_memory == target_memory)
+    # Test8: check snapshot should be removed after migration
+    expected_count = 1 if provider.one_of(RHEVMProvider) else 0
+    ui_snap_count = target_view.entities.summary("Properties").get_text_of("Snapshots")
+    soft_assert(ui_snap_count == expected_count)
```
sshveta: soft assert is not needed here as this is the last assert statement

cfme/tests/v2v/test_post_migrations.py
sshveta: Ok please add a comment about that

cfme/tests/v2v/test_post_migrations.py
mnadeem92: There is no hard limit of choosing it, we can not consider 1 as in case of RHV the snapshot count always be 1 after migration , so it is hard to tell if this is the expected snapshot or the old snapshot. We may take 2 but again probability of old snapshot migrated is 50-50, So i have kept it 3 to be more accurate.

cfme/tests/v2v/test_post_migrations.py
mnadeem92: Done

izapolsk:
@mnadeem92, PR looks good but PRT doesn't have any successful test. Please check results.
one more thing, soft_assert isn't replacement for assert. please use it only in certain cases where it is really necessary.

cfme/tests/v2v/test_post_migrations.py
izapolsk: please use assert instead.  soft_assert should be used only in case when there can be several asserts in one test and we would like to see results of all failed asserts.

one more thing, if you put assert, please add description to it explaining what's wrong.

shs8313 (approved):

PR #9422, opened 2019-09-26
[RFR] Add browser isolation marker for test items

mshriver:
Browser isolation is needed at a lower level than wrapped around every test case via the CLI option.

This adds a marker (to auth tests first) that enforces setup/teardown based browser isolation.

### DIFF
--- a/cfme/markers/isolation.py
+++ b/cfme/markers/isolation.py
@@ -1,0 +1,26 @@
+"""Marker for browser isolation on specific tests"""
+import pytest
+
+from cfme.test_framework.browser_isolation import browser_implementation_quits
+from cfme.utils.log import logger
+
+
+# Add Marker
+def pytest_configure(config):
+    config.addinivalue_line('markers', 'browser_isolation: Mark a test case for browser isolation')
+
+
+@pytest.mark.hookwrapper(tryfirst=True)
+def pytest_runtest_setup(item):
+    if item.get_marker('browser_isolation'):
+        logger.info('Browser isolation for marker in setup')
+        browser_implementation_quits(item)
+    yield
+
+
+@pytest.mark.hookwrapper(trylast=True)
+def pytest_runtest_teardown(item, nextitem):
+    yield
+    if item.get_marker("browser_isolation"):
+        logger.info('Browser isolation for marker in teardown')
+        browser_implementation_quits(item)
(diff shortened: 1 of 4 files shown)


cfme/tests/integration/test_cfme_auth.py
```diff
@@ -118,6 +118,7 @@ def user_obj(appliance, auth_user, user_type):
 @pytest.mark.uncollectif(lambda auth_user: not any([True for g in auth_user.groups or []
                                                    if 'evmgroup' in g.lower()]),
                          reason='No evm group available for user')
+@pytest.mark.browser_isolation
```
lina-nikiforova: maybe just put it on top of the file for all tests in this module?

mshriver: Yeah I think that should work fine - I was in the mentality of a fixture, and the scope not working right - but it should still operate at the function level just fine.

cfme/markers/isolation.py
```diff
@@ -0,0 +1,26 @@
...
+def pytest_configure(config):
+    config.addinivalue_line('markers', 'browser_isolation: Mark a test case for browser isolation')
+
+
+@pytest.mark.hookwrapper(tryfirst=True)
+def pytest_runtest_setup(item):
+    if item.get_marker('browser_isolation'):
+        logger.info('Browser isolation for marker in setup')
+        browser_implementation_quits(item)
+    yield
```
sbulage: **Suggestion:**
Blank `yield` is not require.

mshriver: Sure do in these hookwrappers.  Without yielding, you'll see this:

```
___________________________ ERROR at setup of test_login_evm_group[ldap-ad_dev_2012-cn-ldapuser1] ____________________________

item = <Function 'test_login_evm_group[ldap-ad_dev_2012-cn-ldapuser1]'>

    @pytest.mark.hookwrapper(tryfirst=True)
    def pytest_runtest_setup(item):
        """Pytest hook ensuring that failures caused by NotImplementedError show up as skips instead"""
        outcome = yield
        try:
>           outcome.get_result()
E           TypeError: 'NoneType' object is not an iterator

```

sbulage: Thanks @mshriver :smile:

PR #9423, opened 2019-09-26
[RFR] add new label to release script

mshriver:
Adding label to the release script.

Thanks @john-dupuy for the idea!

### DIFF
--- a/cfme/utils/release.py
+++ b/cfme/utils/release.py
@@ -49,6 +49,7 @@ VALID_LABELS = [
     "RHV",
     "sprout",
     "tech-debt",
+    "test-automation",
     "test-cleanup",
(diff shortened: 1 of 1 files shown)


cfme/utils/release.py
```diff
@@ -49,6 +49,7 @@ def clean_commit(commit_msg):
     "RHV",
     "sprout",
     "tech-debt",
+    "test-automation",
```
kedark3: How is this different from new-test-or-feature? what is the use case?

john-dupuy: It is for PR's that are automating manual test cases, so that we can collect metrics about number of manual cases automated per sprint.

PR #9425, opened 2019-09-26
[1LP][RFR] Download ansible logs

sshveta:
Downloading Pre and Post migration logs after migration

### DIFF
--- a/cfme/tests/v2v/test_v2v_ansible.py
+++ b/cfme/tests/v2v/test_v2v_ansible.py
@@ -164,9 +164,7 @@ def test_migration_playbooks(request, appliance, source_provider, provider,
     assert migration_plan.wait_for_state("Successful")
 
     # Downloading pre and post migration logs
-    completed_plans_view = navigate_to(migration_plan, "Complete")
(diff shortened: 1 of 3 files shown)


cfme/tests/v2v/test_v2v_ansible.py
```diff
@@ -162,5 +163,14 @@ def _cleanup():
     assert migration_plan.wait_for_state("Completed")
     assert migration_plan.wait_for_state("Successful")
 
+    # Downloading pre and post migration logs
+    completed_plans_view = navigate_to(migration_plan, "Complete")
```
john-dupuy: @sshveta thoughts on making this a method of `migration_plan`? Do you anticipate this being used in the future?

kedark3: I think it will be a good idea to have methods to download pre/post migration logs separately.

sshveta: @john-dupuy , Sure yeah we might use navigation to Completed plans in other tests , will create a method @kedark3 , method for downloading pre/post logs is not useful , We will be using it only in this test

PR #9426, opened 2019-09-26
[WIPTEST] Test migration without conversion hosts

Yadnyawalkya:
## Purpose or Intent
- __Adding test__ to check migration without setting conversion host entry 

## pytest
{{ pytest: cfme/tests/v2v/test_v2v_migrations_ui.py --use-provider rhv43 --use-provider vsphere67-ims --provider-limit 2 -vvvv -k 'test_migration_with_no_conversion' }}

### DIFF
--- a/cfme/tests/v2v/test_v2v_migrations_ui.py
+++ b/cfme/tests/v2v/test_v2v_migrations_ui.py
@@ -519,3 +519,34 @@ def test_duplicate_mapping_name(appliance, mapping_data_vm_obj_mini):
     view.general.description.fill("description")
     view.general.flash.assert_message(f"Infrastructure mapping {name} already exists")
     view.general.cancel_btn.click()
+
+
+def test_migration_with_no_conversion(appliance, provider, mapping_data_vm_obj_mini):
+    """
+    Test Migration plan without setting conversion hosts
+
+    Polarion:
+        assignee: sshveta
+        initialEstimate: 1/2h
+        caseimportance: high
+        caseposneg: negative
+        testtype: functional
+        startsin: 5.10
+        casecomponent: V2V
+    """
+    # Delete existing conversion entries from cloudforms
+    delete_hosts = appliance.ssh_client.run_rails_command(
+        "'MiqTask.delete_all; ConversionHost.delete_all'")
+    if not delete_hosts.success:
+        pytest.skip(
+            "Failed to delete all conversion hosts: {}".format(delete_hosts.output))
+
+    migration_plan_collection = appliance.collections.v2v_migration_plans
+    migration_plan = migration_plan_collection.create(
+        name="plan_{}".format(fauxfactory.gen_alphanumeric()),
+        description="desc_{}".format(fauxfactory.gen_alphanumeric()),
+        infra_map=mapping_data_vm_obj_mini.infra_mapping_data.get("name"),
+        target_provider=provider,
+        vm_list=mapping_data_vm_obj_mini.vm_list,
+    )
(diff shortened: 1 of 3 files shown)


cfme/v2v/migration_plans.py
```diff
@@ -438,6 +439,12 @@ def successful(self):
         view = navigate_to(self, "Complete")
         return view.plans_completed_list.is_plan_succeeded(self.name)
 
+    def stucked_in_conversion(self):
```
kedark3: Does it really gets stuck or failed?

cfme/tests/v2v/test_v2v_migrations_ui.py
```diff
@@ -519,3 +519,34 @@ def test_duplicate_mapping_name(appliance, mapping_data_vm_obj_mini):
...
+        assignee: sshveta
+        initialEstimate: 1/2h
+        caseimportance: high
+        caseposneg: negative
+        testtype: functional
+        startsin: 5.10
+        casecomponent: V2V
+    """
+    # Delete existing conversion entries from cloudforms
+    delete_hosts = appliance.ssh_client.run_rails_command(
```
john-dupuy: I would expect this to be in a fixture. Since it is cleaning up the appliance in order to run the test. Could be a global fixture like `has_no_providers`, something like `has_no_conversion_hosts`. What do you think?

cfme/tests/v2v/test_v2v_migrations_ui.py
```diff
@@ -519,3 +519,34 @@ def test_duplicate_mapping_name(appliance, mapping_data_vm_obj_mini):
...
+
+    migration_plan_collection = appliance.collections.v2v_migration_plans
+    migration_plan = migration_plan_collection.create(
+        name="plan_{}".format(fauxfactory.gen_alphanumeric()),
+        description="desc_{}".format(fauxfactory.gen_alphanumeric()),
+        infra_map=mapping_data_vm_obj_mini.infra_mapping_data.get("name"),
+        target_provider=provider,
+        vm_list=mapping_data_vm_obj_mini.vm_list,
+    )
+    assert migration_plan.wait_for_state("Stucked")
```
john-dupuy: @Yadnyawalkya the past tense of stick is stuck. English is annoying and has dumb rules with exceptions all over the place

cfme/v2v/migration_plans.py
```diff
@@ -375,7 +375,8 @@ class MigrationPlan(BaseEntity):
     MIGRATION_STATES = {'Started': lambda self: self.plan_started(),
                         'In_Progress': lambda self: self.in_progress(),
                         'Completed': lambda self: self.completed(),
-                        'Successful': lambda self: self.successful()}
+                        'Successful': lambda self: self.successful(),
+                        'Stucked': lambda self: self.stucked_in_conversion()}
```
john-dupuy: Same here about `stucked`, I would expect `self.stuck_in_conversion`

widgetastic_manageiq/__init__.py
```diff
@@ -5447,6 +5448,17 @@ def get_progress_percent(self, plan_name):
         ]
         return dict(list(zip(desc, val)))
 
+    def is_plan_failed(self, plan_name):
```
john-dupuy: @Yadnyawalkya why do we need `is_plan_failed` and `is_plan_succeeded`? I would also expect this to always return a boolean (and not error_text). It is consistent with our framework to have this be the case (e.g. `is_displayed`).

dajoRH:
*CFME QE Bot*

PR #9428, opened 2019-09-27
[RFR] Fix rest-api-entity for myservice

john-dupuy:
`myservice.rest_api_entity` was returning 

```{TypeError}getattr(): attribute name must be string```

due to an unnecessary index `[0]`

### DIFF
--- a/cfme/services/myservice/__init__.py
+++ b/cfme/services/myservice/__init__.py
@@ -39,8 +39,8 @@ class MyService(Updateable, Navigatable, Taggable, sentaku.modeling.ElementMixin
     @property
     def rest_api_entity(self):
         try:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by jawatts** [comment]
Pound for pound, this is the best PR I have ever seen

File: cfme/services/myservice/__init__.py [hunk 1]
```diff
@@ -39,7 +39,7 @@ def __init__(self, appliance, name=None, description=None, vm_name=None):
     @property
     def rest_api_entity(self):
         try:
-            return self.appliance.rest_api.collections.services.get(name=self.name)[0]
```
jawatts: This will no longer raise a `IndexError`, it will raise a `ValueError`

john-dupuy: good call, thanks! :+1:

PR #9433, opened 2019-10-01
[1LP][RFR] New Test: Testing CRUD operations on imported domains

ganeshhubale:

## Purpose or Intent
- Testing crud operation on imported domains with respect to different sources
### PRT Run
{{ pytest: cfme/tests/automate/test_file_import.py::test_crud_imported_domains --use-template-cache -qsvv }}

### DIFF
--- a/cfme/automate/explorer/domain.py
+++ b/cfme/automate/explorer/domain.py
@@ -240,12 +240,6 @@ class Domain(BaseEntity, Fillable, Updateable):
             domains_view = self.create_view(DomainListView)
             assert domains_view.is_displayed
             domains_view.flash.assert_no_error()
-            domains_view.flash.assert_message(
-                'Automate Domain "{}": Delete successful'.format(self.description or self.name))
-
-        # TODO(BZ-1704439): Remove the work-around once this BZ got fixed
-        if BZ(1704439).blocks:
-            self.browser.refresh()
 
     def lock(self):
         # Ensure this has correct data
@@ -292,13 +286,6 @@ class Domain(BaseEntity, Fillable, Updateable):
         view = self.create_view(DomainDetailsView, override=updates)
         assert view.is_displayed
         view.flash.assert_no_error()
-        if changed:
-            text = (updates.get('description', self.description)
-                    or updates.get('name', self.name))
-            view.flash.assert_message('Automate Domain "{}" was saved'.format(text))
-        else:
-            view.flash.assert_message(
-                'Edit of Automate Domain "{}" was cancelled by the user'.format(self.name))
 
     def refresh(self, branch_or_tag=None, git_branch=None, cancel=False):
         view = navigate_to(self, 'Refresh')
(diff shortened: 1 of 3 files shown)


cfme/tests/automate/test_file_import.py
```diff
@@ -31,3 +36,68 @@ def test_domain_import_file(import_datastore, import_data):
...
+    if import_data.file_name == "bz_1753586_user.zip":
+        with update(domain):
+            domain.description = fauxfactory.gen_alpha()
+        domain.delete()
+        view.flash.assert_message(f'Automate Domain "{domain.description}": Delete successful')
+    elif import_data.file_name == "bz_1753586_user_locked.zip":
+        with update(domain):
+            domain.description = fauxfactory.gen_alpha()
+        domain.delete()
+        view.flash.assert_message(f'Automate Domain "{domain.description}": Delete successful')
```
john-dupuy: @ganeshhubale Since these `if, elif` blocks are doing the same thing, you can combine them.

ganeshhubale: My bad!
Thanks @john-dupuy :+1:

cfme/tests/automate/test_file_import.py
```diff
@@ -90,3 +92,63 @@ def test_upload_blank_file(appliance, upload_file):
...
+            3. From within the new Domain_Test02 dir, edit the __domain__.yaml file so that the
+               'name' attribute equals Domain_Test02
+               NOTE: Do not change the 'source' attribute from 'system' to 'user' this time.
+            4. Archived the new Domain_Test02 dir and upload to CloudForms Appliance
+        expectedResults:
+            1.
+            2.
+            3.
+            4. The imported domains should be Custom Domains or these imported domains should gets
+               deleted from db
```
valaparthvi: The test case is simple enough and I really don't think we need testSteps, besides testSteps doesn't seem to be related to what your test case is actually doing there. I'd rather the testSteps be removed, they are a little confusing. Let me know what you think.

ganeshhubale: Removed testSteps and expectedResults.

cfme/automate/explorer/domain.py
```diff
@@ -292,13 +286,6 @@ def update(self, updates):
         view = self.create_view(DomainDetailsView, override=updates)
         assert view.is_displayed
         view.flash.assert_no_error()
-        if changed:
-            text = (updates.get('description', self.description)
-                    or updates.get('name', self.name))
-            view.flash.assert_message('Automate Domain "{}" was saved'.format(text))
```
jawatts: **Required** agree with removing this in this module but I think there should be a test checking these messages flash messages.

ganeshhubale: @jawatts 
I agree. I am thinking of updating testcase ```test_domain_crud``` to check domain delete flash message.

PR #9434, opened 2019-10-01
[1LP][RFR] marked test_azure_multiple_subscription with 2 providers

lina-nikiforova:
## Purpose or Intent
The test was still skipped in jenkins, I hope it will fix it.

### PRT Run
{{ pytest: -v cfme/tests/cloud/test_providers.py::test_azure_multiple_subscription --use-provider=complete --long-running }}

### DIFF
--- a/cfme/markers/env_markers/provider.py
+++ b/cfme/markers/env_markers/provider.py
@@ -15,6 +15,7 @@ from cfme.utils.pytest_shortcuts import fixture_filter
 from cfme.utils.version import Version
 
 ONE = 'one'
+SECOND = 'second'
 ALL = 'all'
 LATEST = 'latest'
 ONE_PER_VERSION = 'one_per_version'
@@ -294,6 +295,14 @@ def providers(metafunc, filters=None, selector=ALL, fixture_name='provider'):
     if selector == ONE:
         if matching_provs:
             allowed_providers = [matching_provs[0]]
+        else:
+            allowed_providers = []
+    elif selector == SECOND:
+        if matching_provs:
+            try:
+                allowed_providers = [matching_provs[1]]
+            except IndexError:
+                pytest.skip("no second provider was found")
         else:
             allowed_providers = []
     elif selector == LATEST:
--- a/cfme/tests/cloud/test_providers.py
+++ b/cfme/tests/cloud/test_providers.py
@@ -23,6 +23,7 @@ from cfme.common.provider_views import CloudProviderAddView
 from cfme.common.provider_views import CloudProvidersView
 from cfme.fixtures.provider import enable_provider_regions
 from cfme.markers.env_markers.provider import ONE
+from cfme.markers.env_markers.provider import SECOND
 from cfme.utils import appliance
 from cfme.utils import conf
 from cfme.utils import ssh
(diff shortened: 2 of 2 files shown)


cfme/markers/env_markers/provider.py
```diff
@@ -296,6 +297,11 @@ def get_valid_providers(provider):
             allowed_providers = [matching_provs[0]]
         else:
             allowed_providers = []
+    elif selector == SECOND:
+        if matching_provs:
+            allowed_providers = [matching_provs[1]]
```
john-dupuy: Any chance of an index error if the `len` of matching_provs is 1?

cfme/tests/cloud/test_providers.py
```diff
@@ -502,7 +503,10 @@ def test_azure_subscription_required(request, provider):
 @pytest.mark.tier(2)
 @test_requirements.azure
 @pytest.mark.provider([AzureProvider], scope="function", override=True, selector=ONE)
```
izapolsk: I guess you don't need override here.

cfme/tests/cloud/test_providers.py
```diff
@@ -502,7 +503,10 @@ def test_azure_subscription_required(request, provider):
...
-def test_azure_multiple_subscription(appliance, request, soft_assert):
+@pytest.mark.provider([AzureProvider], fixture_name="second_provider", selector=SECOND)
```
izapolsk: @lina-nikiforova, I'm curios how we can test this at all since we have only one azure provider in cfme_data ?  As far as I understand two subscriptions implies two azure providers.  
I guess we need to add one more azure provider with another subscription to test this.

@mshriver, @john-dupuy, what's your opinion ?

john-dupuy: @izapolsk don't we have 2? I see `azure` and `azure_msdn` in the yamls

izapolsk: @john-dupuy, you're right my bad, I incorrectly interpreted search results.

cfme/tests/cloud/test_providers.py
lina-nikiforova: @izapolsk I got 
```
cfme/markers/env_markers/provider.py:428: in process_env_mark
    "specifying the override flag [{}]".format(metafunc.function.__name__)
E   Exception: You have an override provider without specifying the override flag [test_azure_multiple_subscription]
```

cfme/markers/env_markers/provider.py
lina-nikiforova: I think it will cause an error at setup

cfme/tests/cloud/test_providers.py
izapolsk: @lina-nikiforova, weird, sorry, I was told that override could be removed as redundant after recent changes.
@mshriver, is this an issue in provider marker or I'm just wrong ?

cfme/markers/env_markers/provider.py
john-dupuy: I was curious so I checked after commenting out `azure_msdn` in yamls,
```
cfme/markers/env_markers/provider.py:402: in providers_by_class
    return providers(metafunc, filters=[pf], selector=selector, fixture_name=fixture_name)
cfme/markers/env_markers/provider.py:302: in providers
    allowed_providers = [matching_provs[1]]
E   IndexError: list index out of range
```
I think it would be helpful caught this error and gave the user some indication that setup failed because there was only 1 provider available. What do you think?

PR #9440, opened 2019-10-03
[1LP][RFR] New Test: Test embedded method copy to new domain

ganeshhubale:
## Purpose or Intent
 - Testing embedded method attached with inline automate method is getting copied to new domain
- Updated test steps by going through BZ. This does not belogs to ansible. So changed case component and assignee.
### PRT Run
{{ pytest: cfme/tests/automate/test_method.py::test_copy_with_embedded_method --use-template-cache -qsvv}}

### DIFF
--- a/cfme/tests/automate/test_method.py
+++ b/cfme/tests/automate/test_method.py
@@ -797,3 +797,49 @@ def test_automate_user_has_groups(request, appliance, custom_instance):
                 "instance": instance.name,
             },
         )
+
+
+@pytest.mark.tier(2)
+@pytest.mark.meta(automates=[1592140])
+def test_copy_with_embedded_method(request, appliance, klass):
+    """
+    When copying a method within the automate model the copied method
+    does not have the Embedded Methods that are a part of the source method
+
+    Bugzilla:
+        1592140
+
+    Polarion:
+        assignee: ghubale
+        casecomponent: Automate
+        initialEstimate: 1/2h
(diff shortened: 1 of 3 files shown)


cfme/tests/automate/test_method.py
```diff
@@ -797,3 +797,48 @@ def test_automate_user_has_groups(request, appliance, custom_instance):
...
+    request.addfinalizer(domain.delete_if_exists)
+    method.copy_to(domain.name)
+
+    copied_method = (
+        domain.namespaces.instantiate(klass.namespace.name)
+        .classes.instantiate(klass.name)
+        .methods.instantiate(method.name)
+    )
+    view = navigate_to(copied_method, "Details")
+    assert view.embedded_method_table.read()[0]["Path"] == f"/{'/'.join(path[2:])}"
```
john-dupuy: Q: Shouldn't you also assert that the copied methods path matches the original methods path? In the UI?

ganeshhubale: Hi @john-dupuy 
- This test case scenario is related to checking ```embedded method path``` is copied to new domain or not after copying that method.
- So I created one method and selected embedded method and went to method's details page and checked if path is present or not then I copied that method to new domain and went to method's details page in new domain and checked if path is still there or not.
- Please let me know if you mean something else. So that we can test this scenario in more better way :)

PR #9446, opened 2019-10-04
[1LP][RFR] New Test: Testing domain import with or without connection

ganeshhubale:

## Purpose or Intent
- Testing domain import with or without internet connectivity to appliance
### PRT Run
{{ pytest: cfme/tests/automate/test_git_import.py::test_automate_git_domain_import_connection -vvv }}

### DIFF
--- a/cfme/tests/automate/test_automate_manual.py
+++ b/cfme/tests/automate/test_automate_manual.py
@@ -112,34 +112,6 @@ def test_automate_check_quota_regression():
 
 
 @pytest.mark.tier(3)
-@pytest.mark.manual('manualonly')
-def test_automate_git_domain_import_with_no_connection():
-    """
-
-    Polarion:
-        assignee: ghubale
-        casecomponent: Automate
-        caseimportance: medium
-        initialEstimate: 1/6h
-        tags: automate
-        startsin: 5.7
-        testSteps:
-            1. Import a Git Domain into Automate
-            2. Server the connection to the GIT Server from the appliance
-               (Disable VPN or some other trick)
-            3. List all the Automate Domains using Automate-> Explorer
-        expectedResults:
-            1.
-            2.
-            3. The domain should be displayed properly
-
-    Bugzilla:
-        1391208
-    """
-    pass
-
-
-@pytest.mark.tier(3)
 def test_automate_git_import_without_master():
     """
     Git repository doesn't have to have master branch
--- a/cfme/tests/automate/test_git_import.py
+++ b/cfme/tests/automate/test_git_import.py
@@ -378,3 +378,47 @@ def test_automate_git_import_case_insensitive(request, appliance):
     domain = appliance.collections.domains.instantiate(name="testdomain")
     request.addfinalizer(domain.delete_if_exists)
     assert domain.exists
+
+
+@pytest.mark.tier(3)
+@pytest.mark.meta(automates=[1391208])
+@pytest.mark.parametrize("connection", [True, False], ids=["with_connection", "without_connection"])
+def test_automate_git_domain_import_connection(request, temp_appliance_preconfig, connection):
+    """
+    Bugzilla:
+        1391208
+
+    Polarion:
+        assignee: ghubale
+        casecomponent: Automate
+        caseimportance: medium
+        initialEstimate: 1/6h
+        tags: automate
+        startsin: 5.7
+        testSteps:
+            1. Import a Git Domain into Automate
+            2. Server the connection to the GIT Server from the appliance
+               (Disable VPN or some other trick)
+            3. List all the Automate Domains using Automate-> Explorer
+        expectedResults:
+            1.
+            2.
+            3. The domain should be displayed properly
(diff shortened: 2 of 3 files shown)


cfme/tests/automate/test_git_import.py
```diff
@@ -378,3 +378,68 @@ def test_automate_git_import_case_insensitive(request, appliance):
...
+            3. The domain should be displayed properly
+    """
+    # Server role 'git_owner' needs to enabled explicitly on temp_appliance_preconfig
+    temp_appliance_preconfig.server.settings.enable_server_roles('git_owner')
+    if connection:
+        # Checking datastore import from github repository with internet connectivity
+        repo = temp_appliance_preconfig.collections.automate_import_exports.instantiate(
+            import_type="git", url=url, verify_ssl=True
+        )
+        domain = repo.import_domain_from(**{param_type: param_value})
```
valaparthvi: If you're only going to use `param_type` and `param_value` when `connection` is True, you might want to directly declare them inside the test instead of using them as parameters.

cfme/tests/automate/test_git_import.py
```diff
@@ -378,3 +378,68 @@ def test_automate_git_import_case_insensitive(request, appliance):
...
     assert domain.exists
+
+
+@pytest.mark.tier(3)
+@pytest.mark.meta(automates=[1391208])
+@pytest.mark.parametrize(
+    ("url", "param_type", "param_value", "connection"),
+    [
+        (
+            GIT_REPO_URL,
```
valaparthvi: If the value of `url`  for both the parameters is same i.e. `GIT_REPO_URL`, why are you parametrizing it? Why not just directly use the value?

cfme/tests/automate/test_git_import.py
```diff
@@ -378,3 +378,68 @@ def test_automate_git_import_case_insensitive(request, appliance):
...
+            import_type="git", url=url, verify_ssl=True
+        )
+        domain = repo.import_domain_from(**{param_type: param_value})
+        temp_appliance_preconfig.ssh_client.run_command("echo '8.8.8.8 github.com' >> /etc/hosts")
+        request.addfinalizer(domain.delete)
+        assert domain.exists
+    else:
+        # Checking datastore import from github repository without internet connectivity
+        temp_appliance_preconfig.ssh_client.run_command("echo '8.8.8.8 github.com' >> /etc/hosts")
+        view = navigate_to(temp_appliance_preconfig.collections.automate_import_exports, "All")
```
valaparthvi: Instead of all these steps you could use `pytest.raises`.
```suggestion
with pytest.raises("Error during repository fetch: hostname does not match certificate"):
    domain = repo.import_domain_from(**{param_type: param_value})
```

cfme/tests/automate/test_git_import.py
```diff
@@ -378,3 +378,68 @@ def test_automate_git_import_case_insensitive(request, appliance):
...
+        view.import_git.fill({"url": url, "verify_ssl": True})
+        view.import_git.submit.click()
+        view.flash.assert_message(
+            "Error during repository fetch: hostname does not match certificate"
+        )
```
valaparthvi: Instead of all these steps you could use `pytest.raises`.
```suggestion
with pytest.raises("Error during repository fetch: hostname does not match certificate"):
    domain = repo.import_domain_from(**{param_type: param_value})
```

cfme/tests/automate/test_git_import.py
ganeshhubale: Removed parameterization

cfme/tests/automate/test_git_import.py
ganeshhubale: Hi @valaparthvi 
- I tried as you said with **pytest.raise(AssertionError/Exception)**  but I am getting below error. Hence going with existing one:
```*** wait_for.TimedOutError: Could not do 'lambda defined as `ret, _ = wait_for(lambda: self.is_displayed, timeout=timeout, delay=delay)`' at /home/ghubale/cfme_workspace/integration_tests/.cfme_venv/lib64/python3.7/site-packages/widgetastic/widget/base.py:506 in time
ipdb>                                                                                                                                                                        ```
- Please let me know, if you have any other way here

Comment (valaparthvi) [reply cfme/tests/automate/test_git_import.py]: [Here](https://github.com/ManageIQ/integration_tests/blob/master/cfme/automate/import_export.py#L142) in `import_domain_from` method, if you add a `view.flash.assert_no_error()` after `view.import_git.submit.click()`, you will be able to use `pytest.raises(AssertionError)` .

File: cfme/tests/automate/test_automate_manual.py [hunk 5]
```diff
@@ -111,34 +111,6 @@ def test_automate_check_quota_regression():
     pass
 
 
-@pytest.mark.tier(3)
-@pytest.mark.manual('manualonly')
```
mshriver: Proved this marker wrong! :)

PR #9450, opened 2019-10-04
[1LP][RFR]test_sdn_port_net_prov for azure is blocked by BZ

mmojzis:


cfme/tests/networks/test_sdn_ports.py
```diff
@@ -44,6 +45,7 @@ def test_sdn_port_detail_name(provider, appliance):
             pass
 
 
+@pytest.mark.meta(blockers=[BZ(1718495, forced_streams=['5.11'])])
```
john-dupuy: @mmojzis I see that this BZ is verified in CFME 5.11, why is `forced_streams` necessary here?

mmojzis: @john-dupuy Thanks for pointing that out. I tested it with already verified BZ and then forgot to change to correct BZ. It should be okay now.

cfme/tests/networks/test_sdn_ports.py
```diff
@@ -44,6 +45,7 @@ def test_sdn_port_detail_name(provider, appliance):
             pass
 
 
+@pytest.mark.meta(blockers=[BZ(1758593, forced_streams=['5.11'])])
```
jawatts: We do not need the force stream here:
```
In[4]: appliance.version
Out[4]: Version('5.11.0.27')
In[5]: from cfme.utils.blockers import BZ
In[6]: BZ(1758593).blocks
Out[6]: True
```

mmojzis: Forced stream removed.

mmojzis:
> One small change but otherwise it looks good

@jawatts Changed.

PR #9451, opened 2019-10-04
[1LP][RFR] Extend test_refresh_relationships: check for errors in evm.log

psimovec:
There should be no errors during refreshing relationships.
By checking for errors in a more specific test case of this one [1] 5.10.11 blocker [2] was found. (Which will also block 5.10 of this test for now.)

[1] - https://github.com/ManageIQ/integration_tests/pull/9400
[2] - https://bugzilla.redhat.com/show_bug.cgi?id=1756984

{{ pytest:  cfme/tests/cloud_infra_common/test_relationships.py -v --use-provider complete  }}

### DIFF
--- a/cfme/tests/cloud_infra_common/test_relationships.py
+++ b/cfme/tests/cloud_infra_common/test_relationships.py
@@ -10,6 +10,7 @@ from cfme.cloud.flavor import ProviderFlavorAllView
 from cfme.cloud.provider import CloudProvider
 from cfme.cloud.provider import CloudProviderImagesView
 from cfme.cloud.provider import CloudProviderInstancesView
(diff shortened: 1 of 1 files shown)


psimovec:
[openstack-13] fails with error in log:
[----] E, [2019-10-08T07:39:29.881768 #23452:2af690ea8c40] ERROR -- : MIQ(ManageIQ::Providers::Openstack::StorageMana...rted because [RESOURCE_LOCKED - cannot obtain exclusive access to locked queue 'miq-192.168.200.22-nova' in vhost '/']

or in report:
http://10.16.45.124/pr9451-r35178-downstream-510z-KwwmQc/artifacts/report.html

PR #9453, opened 2019-10-07
[1LP][RFR] Add test_requirements for jhenner's tests

jaryn:
### Purpose or Intent
Marking tests with test_requirements

Some requirements do not exist yet. I presume they will get added to Polarion by the importer.
Some requirements were changed. I presume I will have to delete the old ones from polarion by hand.

Some of the touched tests failed but I was not changing any tests logic so they fail for reasons unrelated to this PR. Perhaps I should just have set this as NOTEST.

### DIFF
--- a/cfme/test_requirements.py
+++ b/cfme/test_requirements.py
@@ -74,6 +74,12 @@ alert = pytest.mark.requirement(
     assignee_id='jdupuy',
 )
 
+appliance = pytest.mark.requirement(
+    "Appliance",
+    description='Appliance',
+    assignee_id='jhenner'
+)
+
 app_console = pytest.mark.requirement(
     "Appliance Console",
     description='Appliance console operations support',
@@ -322,6 +328,13 @@ rest = pytest.mark.requirement(
     assignee_id='pvala',
 )
 
+restore = pytest.mark.requirement(
+    "CFME Database Backup Restore",
+    description='Database backup and restore tests',
+    assignee_id='jhenner',
+)
+
+
 retirement = pytest.mark.requirement(
     "VM/Instance Retirement",
     description='Retirement of VMs/Instances and visibility of objects after retirement',
@@ -344,6 +357,12 @@ satellite = pytest.mark.requirement(
     "Configuration Management: Satellite",
     description='Configuration Management providers, specifically Red Hat Satellite',
     assignee_id='tpapaioa',
+)
+
+scheduled_ops = pytest.mark.requirement(
+    'Scheduled Operations',
+    description='Scheduled Operations',
+    assignee_id='jhenner',
 )
 
 scvmm = pytest.mark.requirement(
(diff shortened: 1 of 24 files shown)


cfme/test_requirements.py
```diff
@@ -414,9 +433,15 @@ def test_quota_alert():
     assignee_id='nachandr',
 )
 
-upgrade = pytest.mark.requirement(
-    "CFME Upgrade and Database Migrations",
-    description='Upgrading application and migrating databases',
+update = pytest.mark.requirement(
```
john-dupuy: @jaryn Why split `upgrade` into two? In the sign-off sheet they are listed under one FA. I'm just curious.

jaryn: @john-dupuy Well I must say I didn't think it through enough yet. I wanted to put more emphasis on the tests are of different kinds.

I am not sure what are the requirements meant for. Perhaps it is not good to split markers into two.  I marked the tests that require some kind of CFME code as update. Tests that are requiring doing db:migration are marked with migration.

There is this difference between the tests:

### The update tests:
When only the `z` in CFME version `x.y.z` is changing, the DB scheme should not change. This implies the list of migrations should remain the same. `rake db:migrate` will do nothing (well perhaps there should be a test for this).

These tests generally do some equivalent of yum update only + some stuff checking some things.

For these tests, we need to start with appliance of CFME version x.y.(z-1)

### The migration tests:
Generally they use some db_dump file of some older version and this is used to pre-fill the DB of various deployments. Then they do rake db:migrate to check the migrations are working with the sample DBs.

Then there are some tests like test_upgrade_multi_replication_inplace which requires to boot some appliances, do the update, do the migration and perhaps some additional check.

For these tests, we need to start with appliance of CFME version x.(y-1).z

### Considerations
We also need to consider that with CFME 5.11, we cannot yum update but the update part is actually dumping the db and restoring the db on different appliance (which is part of what the restore tests doing) and then db:migrate. Perhaps these should be marked with side_by_side_update.

Perhaps the polarion marker should still be upgrade for all kinds of these but I think I may like to categorize the tests with some non-polarion marker.

john-dupuy: @jaryn Thanks for the detailed explanation of your thought process here. I think the distinction is important and based on what you've described, it makes sense for these to be separate requirements. By the way, the requirements are directly mapped to those in Polarion so by making these two new requirements in integration_tests they will be added as requirements into Polarion (and then any tests marked with these requirements in integration tests will become linked work items under the requirements in Polarion). 

I am not sure of the best approach for the update tests of 5.11, but I think they can be left as `update` for now.

jaryn: @john-dupuy I am sorry I don't understand.

You say that the tests should be left as "update" but they were "upgrade". Together with the first paragraph it looks like you agree with the split. But you switched this PR back to WIP. This seems to me as contradiction.

dajoRH:
*CFME QE Bot*

PR #9454, opened 2019-10-07
[1LP][RFR]Added manual tests for ec2 url_endpoint and ec2 regions

mmojzis:
Added manual tests for ec2 url_endpoint and ec2 regions.
For url endpoint I need to create VPC and then add it to yamls so I could automate it.
For regions I think there is no way how to automate this test case.


mmojzis:
Tests are only manual. Failed checks are not related to this PR.

PR #9459, opened 2019-10-08
[1LP][RFR] BZ build report

john-dupuy:
Adding a new `miq bz` command, `miq bz build-report`, this command will build a BZ report on BZ's that have gone into the current appliance version (i.e. their `fixed_in` version matches that of the current appliance). 

```
Usage: miq bz build-report [OPTIONS]

  Generate BZ report on bugs going into a z-stream

Options:
  -s, --stream TEXT      Stream for which to get BZs (e.g. 5.10.11.0)
                         [default: <current-appliance-stream>]
  -q, --query-file TEXT  Template query file  [default: fixed_in_query.yaml]
  -o, --output           Show information about BZs in the terminal  [default:
                         False]
  -f, --filename TEXT    Filename into which the report is written (must be
                         yaml file)
  --help                 Show this message and exit.

```

### DIFF
--- a/cfme/scripting/bz.py
+++ b/cfme/scripting/bz.py
@@ -26,15 +26,21 @@ from collections import namedtuple
 import click
 import pytest
 import yaml
+from bugzilla_data import BugzillaData
 
 from cfme.utils.blockers import BZ
 from cfme.utils.log import logger
+from cfme.utils.path import data_path
+from cfme.utils.version import current_version
 
 STATUS = {
     "open_bzs": {"val": "true", "text": "are open", "coverage_text": " open"},
     "closed_bzs": {"val": "false", "text": "are closed", "coverage_text": " closed"},
     "all_bzs": {"val": None, "text": "have coverage", "coverage_text": ""}
 }
+
+QUERY_PATH = data_path.join("/bugzilla-queries/")
+BZ_URL = "bugzilla.redhat.com"
 
 
 def get_report(directory):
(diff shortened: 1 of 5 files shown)


john-dupuy:
travis is failing due to a matplotlib dependency issue, spoke with @digitronik about whether it's necessary to have this plotting functionality and we are exploring options of having matplotlib optional/ as a plugin so that jenkins/prt etc will not have to have it as a library.

john-dupuy:
@digitronik Remove plotting functionality, that is now only for the scripts included in bugzilla-data https://github.com/john-dupuy/bugzilla-data/blob/master/scripts/make_plot.py

cfme/scripting/bz.py
```diff
@@ -110,6 +116,51 @@ def report(directory):
...
+@click.option(
+    "--filename",
+    "-f",
+    "fname",
+    default=f"bz-report-{current_version().vstring}.yaml",
+    help="Filename into which the report is written (must be yaml file)"
+)
+def build_report(stream, query_file, output, fname):
+    query_path = QUERY_PATH.join(query_file)
+    bz_data = BugzillaData(query_path, BZ_URL, "component", credential_file="None")
```
digitronik: As per my understanding. I we not logged in to `python-bugzilla` then more chances to miss `private` bug.

data/bugzilla-queries/fixed_in_query.yaml
```diff
@@ -0,0 +1,16 @@
...
+        - summary
+        - component
+        - description
+        - status
+        - qa_contact
+        - creator
+        - assigned_to
+        - version
+        - fixed_in
+    fixed_in:
```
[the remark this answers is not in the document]
john-dupuy: So this yaml file actually just serves as a template. When the command is run, it pulls the current version of your appliance that's specified in `env.local.yaml` and modifies the query with it's version. Or if you pass the `--stream/--version` argument it will use that version.

cfme/scripting/bz.py
john-dupuy: Updated bugzilla-data and this PR to allow for this

john-dupuy:
For some reason the modules `cfme/scripting/bz.py` and `cfme/scripting/miq.py` are causing `test_modules_importable.py` to hang in travis. I cannot seem to reproduce this locally on Fedora 30 with python 3.7.4. Perhaps it has something to do with running ubuntu with python 3.7.1 rather than 3.7.4.

For now I've marked them as failing modules.

cfme/scripting/bz.py
```diff
@@ -26,16 +26,23 @@
...
+from cfme.utils.version import current_version
 
 STATUS = {
     "open_bzs": {"val": "true", "text": "are open", "coverage_text": " open"},
     "closed_bzs": {"val": "false", "text": "are closed", "coverage_text": " closed"},
     "all_bzs": {"val": None, "text": "have coverage", "coverage_text": ""}
 }
 
+QUERY_PATH = data_path.join("/bugzilla-queries/")
+BZ_URL = "bugzilla.redhat.com"
```
digitronik: I would take this url from `data_yaml` than hard-codding it.

john-dupuy: good idea -- let me do that. I am confused by the travis failures as well, that test runs fine locally.

dajoRH:
*CFME QE Bot*

PR #9464, opened 2019-10-09
[1LP][RFR] cleanup volume on target provider

sshveta:
{{ pytest: cfme/tests/v2v/test_v2v_migrations.py --use-provider rhv43 --use-provider vsphere67-ims --provider-limit 2 -vvvv --long-running -k rhevm-4.3-virtualcenter-6.7-mapping_data_dual_vm_obj_dual_datastore0 }}

### DIFF
--- a/cfme/fixtures/v2v_fixtures.py
+++ b/cfme/fixtures/v2v_fixtures.py
@@ -378,6 +378,17 @@ def __set_conversion_instance_for_osp_ui(appliance, source_provider,  # noqa
                                        osp_ca_cert=tls_cert_key)
 
 
+def cleanup_target(provider, migrated_vm):
+    """Helper function to cleanup instances and associated volumes from openstack"""
+    if provider.one_of(OpenStackProvider):
+        volumes = []
+        vm = provider.mgmt.get_vm(migrated_vm.name)
+        for vol in vm.raw._info['os-extended-volumes:volumes_attached']:
+            volumes.append(vol['id'])
+        migrated_vm.cleanup_on_provider()
+        provider.mgmt.delete_volume(*volumes)
+
+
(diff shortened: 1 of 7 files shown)


cfme/tests/v2v/test_v2v_ansible.py
```diff
@@ -170,4 +167,10 @@ def _cleanup():
     view.flash.assert_no_error()
 
     migrated_vm = get_migrated_vm(src_vm_obj, provider)
+
+    @request.addfinalizer
+    def _cleanup():
+        infrastructure_mapping_collection.delete(mapping)
```
jaryn: Hmmm...
I do not like that the mapping is left without any finalizer until here. I was trying to figure out why you are registering the finalizer after line 169, but did not succeed. Can you please explain?

sshveta: @jaryn :  I have added another line in finalizer method that needs `migrated_vm` which is at line 169 so moved the finalizer down

cfme/fixtures/v2v_fixtures.py
```diff
@@ -378,6 +378,18 @@ def __set_conversion_instance_for_osp_ui(appliance, source_provider,  # noqa
...
 
 
+def cleanup_target(provider, migrated_vm):
+    """Helper function to cleanup instances and associated volumes from openstack"""
+    if provider.one_of(OpenStackProvider):
+        volumes = []
+        for vm in provider.mgmt.list_vms():
+            if migrated_vm.name == vm.name:
+                for vol in vm.raw._info['os-extended-volumes:volumes_attached']:
+                    volumes.append(vol['id'])
```
izapolsk: I guess you need to add `break` in order to stop iteration when you found necessary vm. btw, afaik there should be some special method like get_vm() which returns concrete vm .

PR #9474, opened 2019-10-11
[1LP][RFR] Add planned-in to requirements for polarion

mshriver:
An update to requirement meta, including 'planned in' ID's for polarion.

Default in yaml are the default supported versions of the framework.

Specific requirements can set the ID's as needed.

### DIFF
--- a/cfme/test_requirements.py
+++ b/cfme/test_requirements.py
@@ -59,7 +59,6 @@ ansible = pytest.mark.requirement(
     "Ansible Automation",
     description='Embedded Ansible Automation Requirement',
     assignee_id='sbulage',
-    planned_in_ids="5_11",
 )
 
 access = pytest.mark.requirement(
--- a/cfme/test_requirements.py
+++ b/cfme/test_requirements.py
@@ -43,7 +43,7 @@ Supported requirement metadata fields (and example data)
             "assignee_id": "mkourim",
             "category_ids": "category_id1, category_id2",
             "due_date": "2018-09-30",
-            "planned_in_ids": "planned_id1, planned_id2",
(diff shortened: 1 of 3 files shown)


cfme/test_requirements.py
```diff
@@ -43,7 +43,7 @@ def test_quota_alert():
             "assignee_id": "mkourim",
             "category_ids": "category_id1, category_id2",
             "due_date": "2018-09-30",
-            "planned_in_ids": "planned_id1, planned_id2",
+            "planned_in_ids": "planned_id1 planned_id2",
```
kedark3: removed comma on purpose?

cfme/test_requirements.py
```diff
@@ -54,16 +54,19 @@ def test_quota_alert():
 """
 import pytest
 
+
 ansible = pytest.mark.requirement(
     "Ansible Automation",
     description='Embedded Ansible Automation Requirement',
     assignee_id='sbulage',
+    planned_in_ids="5_11",
```
digitronik: is it not `5_10` Mike?

cfme/test_requirements.py
izapolsk: @mshriver, do we use different separators in different fields ?  look at category_ids above. it has comma separator

mshriver: Not sure about category_ids specifically, and if you read the comment in its entirety you'll see the message from Martin that this is potentially an inaccurate example.

In testing, polarion xunit import accepts space delineated planned in ID's (which is different from the displayed planned-in value). Comma separated resulted in xunit import failures.

PR #9476, opened 2019-10-14
[1LP][RFR] resetter should point all active services

digitronik:
## Purpose or Intent

- Fix 
```
AttributeError: 'MyServiceAll' object has no attribute 'prerequisite_view'
```
- `resetter` should never use `prerequisite_view`; it can use `view` as its post operation
- We had this resetter to point `ActiveServices` so restoring old line.

### PRT Run

{{py.test: cfme/tests/services/test_service_performance.py::test_services_performance cfme/tests/automate/custom_button/test_service_objects.py::test_custom_button_automate_service_obj[Service-Submit_all-ViaUI] -vv}}

### DIFF
--- a/cfme/services/myservice/ui.py
+++ b/cfme/services/myservice/ui.py
@@ -369,7 +369,7 @@ class MyServiceAll(CFMENavigateStep):
         self.prerequisite_view.navigation.select('Services', 'My Services')
 
     def resetter(self, *args, **kwargs):
-        self.prerequisite_view.navigation.select('Services', 'My Services')
+        self.view.myservice.tree.click_path('Active Services')
 
 
 @navigator.register(MyService, 'Details')


jaryn:
I think this may break `cfme/tests/services/test_service_performance.py::test_services_performance` with Firefox. I am going to check that.

cfme/services/myservice/ui.py
```diff
@@ -369,7 +369,7 @@ def step(self, *args, **kwargs):
         self.prerequisite_view.navigation.select('Services', 'My Services')
 
     def resetter(self, *args, **kwargs):
-        self.prerequisite_view.navigation.select('Services', 'My Services')
+        self.view.myservice.tree.click_path('Active Services')
```
jaryn: I suggest to fix the problem like that:
```suggestion
        self.view.navigation.select('Services', 'My Services')
```
Because otherwise the `cfme/tests/services/test_service_performance.py::test_services_performance` will fail as the services tree will not get displayed in FF.

digitronik: @jadhaj if you check steps in navigation its only doing that work `select('Services', 'My Services')` then why we need to do same thing in `resetter`?

secondly, we are testing these thing in chrome not in firefox. Yup, I know most of navigation and widgets not working properly on firefox.

mshriver: @jaryn interesting observation, are you sure this isn't a BZ in CF?

Going through the navigation tree really is a re-navigation, not a resetter action.  If the test needs specific navigation behavior, that should be handled in the test.

"Normal" navigation reset behavior, I think, would just be to click the accordion tree, not completely re-navigate.

mshriver:
As @jaryn predicted this breaks the service performance test. That test should be updated to navigate without resetter.

PR #9478, opened 2019-10-14
[WIPTEST]Adding is_configured property

mnadeem92:


cfme/utils/appliance/__init__.py
```diff
@@ -366,6 +366,11 @@ def _version_tostr(x):
             for k in set(self.CONFIG_MAPPING.values())
             if k in self.__dict__})
 
+    @property
+    def is_configured(self):
+        """ Check if the appliance is configured or unconfigured."""
+        return self.db.has_database
```
john-dupuy: If the DB is configured shouldn't we also check if the DB `has_tables` 

```suggestion
        return self.db.has_database and self.db.has_tables
```

PR #9483, opened 2019-10-15
[1LP][RFR] added proper skip if the datastore is not available for ssa

lina-nikiforova:
## Purpose or Intent
The test failed instead of skipping because the whole dropdown was disabled because both options are disabled. Second is to remove the datastore and it's not possible when there are VMs or templates.

### PRT Run
{{ pytest: -v cfme/tests/infrastructure/test_datastore_analysis.py::test_run_datastore_analysis  --long-running }}


lina-nikiforova:
@digitronik yes, that's exactly what happens

PR #9484, opened 2019-10-15
[1LP][RFR] Automated BZ1656308, Add Ansible repository in another zone.

sbulage:
## Purpose or Intent

- [x] This will automate Embedded Ansible behavior after moving to another Zone(New).

### PRT Run

{{pytest: cfme/tests/ansible/test_embedded_ansible_basic.py::test_embed_tower_repo_add_new_zone  --long-running -svvvv }}

### DIFF
--- a/cfme/tests/ansible/test_embedded_ansible_basic.py
+++ b/cfme/tests/ansible/test_embedded_ansible_basic.py
@@ -131,6 +131,22 @@ def credentials_collection(appliance):
     return appliance.collections.ansible_credentials
 
 
+@pytest.fixture(scope="function")
+def new_zone(appliance):
+    zone_collection = appliance.collections.zones
+    zone = zone_collection.create(
+        name=fauxfactory.gen_alphanumeric(5),
+        description=fauxfactory.gen_alphanumeric(8)
+    )
+    server_settings = appliance.server.settings
+    server_settings.update_basic_information({'appliance_zone': zone.name})
+    yield zone
+
+    # Resetting appliance to 'default' zone.
+    server_settings.update_basic_information({'appliance_zone': "default"})
+    zone.delete()
+
+
 @pytest.mark.rhel_testing
 @pytest.mark.tier(1)
 def test_embedded_ansible_repository_crud(ansible_repository, wait_for_ansible):
@@ -416,3 +432,59 @@ def test_embedded_ansible_repository_playbook_link(ansible_repository):
     # Asserting 'PlaybookRepositoryView' which is navigated from Repository to Playbooks view.
     # Playbooks view is coming from Repository view which is not existing Playbooks view.
     assert view.is_displayed
+
+
+@pytest.mark.tier(2)
+@pytest.mark.meta(automates=[1656308])
+def test_embed_tower_repo_add_new_zone(appliance, ansible_repository, new_zone, request):
+    """
+    Test whether repository or credentials on New Zone.
+
+    Bugzilla:
+        1656308
+
+    Polarion:
+        assignee: sbulage
+        casecomponent: Ansible
+        initialEstimate: 1/2h
+        testSteps:
+            1. Configure a CFME appliance with the Embedded Ansible provider
+            2. Create a new zone
+            3. Move the appliance into the new zone
+            4. Add an embedded Ansible repository or credential
+        expectedResults:
+            1. Check Embedded Ansible Role is started.
+            2.
+            3.
+            4. Check Repository or Credentials were added.
+    """
+    # Check Ansible repository via fixture is created or not.
+    assert ansible_repository.exists
+    # Check Appliance is moved to newly created Zone.
+    assert new_zone.description == appliance.server.zone.description
+
+    # Creating new Repository after moving appliance to new zone.
+    repositories = appliance.collections.ansible_repositories
+    try:
+        playbooks_yaml = cfme_data.ansible_links.playbook_repositories
+        playbook_name = getattr(request, 'param', 'embedded_ansible')
+    except (KeyError, AttributeError):
+        message = "Missing ansible_links content in cfme_data, cannot setup repository"
+        logger.exception(message)  # log the exception for debug of the missing content
+        pytest.skip(message)
+    repository = repositories.create(
+        name=fauxfactory.gen_alpha(),
+        url=getattr(playbooks_yaml, playbook_name),
+        description=fauxfactory.gen_alpha(),
+        scm_branch="second_playbook_branch"
+    )
+
+    view = navigate_to(repository, "Details")
+    wait_for(
+        lambda: repository.status == "successful",
+        num_sec=120,
+        delay=5,
+        fail_func=view.toolbar.refresh.click
+    )
+    request.addfinalizer(repository.delete_if_exists)
+    assert repository.exists
--- a/cfme/tests/ansible/test_embedded_ansible_manual.py
+++ b/cfme/tests/ansible/test_embedded_ansible_manual.py
@@ -431,32 +431,6 @@ def test_automate_ansible_playbook_method_type_verbosity():
 
 
 @pytest.mark.tier(2)
-def test_embed_tower_repo_add_remote_zone():
-    """
-    Test whether repository or credentials on New Zone.
-
-    Bugzilla:
-        1656308
-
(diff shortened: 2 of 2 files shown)


dajoRH:
I detected some fixture changes in commit 242794eca8ec93c284182de4494d60fa6a1476f8

The *local* fixture **`new_zone`** is used in the following files:
- **cfme/tests/ansible/test_embedded_ansible_basic.py**
    - *test_embed_tower_repo_add_new_zone*

cfme/tests/ansible/test_embedded_ansible_basic.py
```diff
@@ -131,6 +131,22 @@ def credentials_collection(appliance):
...
+        description=fauxfactory.gen_alphanumeric(8)
+    )
+    server_settings = appliance.server.settings
+    server_settings.update_basic_information({'appliance_zone': zone.name})
+    yield zone
+
+    # Resetting appliance to 'default' zone.
+    server_settings.update_basic_information({'appliance_zone': "default"})
+    zone.delete_if_exists
+
```
ganeshhubale: I think, it should be function call: ```zone.delete_if_exists()```

cfme/tests/ansible/test_embedded_ansible_basic.py
```diff
@@ -131,6 +131,22 @@ def credentials_collection(appliance):
     return appliance.collections.ansible_credentials
 
 
+@pytest.fixture(scope="function")
+def new_zone(appliance):
+    zone_collection = appliance.collections.zones
+    zone = zone_collection.create(
```
ganeshhubale: **zone_collection** is used only once. I would use it directly rather than assigning to new variable.

cfme/tests/ansible/test_embedded_ansible_basic.py
```diff
@@ -416,3 +432,57 @@ def test_embedded_ansible_repository_playbook_link(ansible_repository):
...
+            scm_branch="second_playbook_branch"
+        )
+    except (KeyError, AttributeError):
+        message = "Missing ansible_links content in cfme_data, cannot setup repository"
+        logger.exception(message)  # log the exception for debug of the missing content
+        pytest.fail(message)
+    view = navigate_to(repository, "Details")
+    wait_for(
+        lambda: view.entities.summary("Properties").get_text_of("Status") == "successful",
+        timeout=60,
```
ganeshhubale: We can make use of **status** property here like: ```lambda: repository.status == "successful",```

ganeshhubale:
Please have a look at PRT failures.

sbulage:
@ganeshhubale there was a minor change, fixed now.

john-dupuy:
@sbulage PRT still failing with

```>       zone, = zone_res
E       ValueError: not enough values to unpack (expected 1, got 0)

cfme/base/__init__.py:263: ValueError
ValueError
b'not enough values to unpack (expected 1, got 0)'```

File: cfme/tests/ansible/test_embedded_ansible_basic.py [hunk 4]
```diff
@@ -416,3 +432,58 @@ def test_embedded_ansible_repository_playbook_link(ansible_repository):
...
+        repository = repositories.create(
+            name=fauxfactory.gen_alpha(),
+            url=getattr(playbooks_yaml, playbook_name),
+            description=fauxfactory.gen_alpha(),
+            scm_branch="second_playbook_branch"
+        )
+    except (KeyError, AttributeError):
+        message = "Missing ansible_links content in cfme_data, cannot setup repository"
+        logger.exception(message)  # log the exception for debug of the missing content
+        pytest.fail(message)
```
john-dupuy: Since the data is missing, I would expect a skip rather than a fail in this scenario

sbulage:
> @sbulage PRT still failing with
> 
> ```
> E       ValueError: not enough values to unpack (expected 1, got 0)
> 
> cfme/base/__init__.py:263: ValueError
> ValueError
> b'not enough values to unpack (expected 1, got 0)'
> ```

Yeah, I have found the problem, I will update the with fix.

cfme/tests/ansible/test_embedded_ansible_basic.py
sbulage: You're right it should ideally skip test, but if I use `pytest.fail` then we should get the exact error or information why test failed which skip won't tell.

john-dupuy: @sbulage `skip` can pass information about why it skipped, just pass the `message` string in the same way your doing with `fail` here.

sbulage: @john-dupuy Agree that `skip` will also give information but in the logs. Also, this test is related to CRUD so I think failing rather skipping is more appropriate .?

john-dupuy: @sbulage While the test is crud, for it to fail, in my opinion, something must go wrong on the CFME side. In this case, nothing has gone wrong on the CFME side, it's just that our yamls don't have the `ansible_links` content. Because I this, I think a skip is more appropriate.

sbulage: Yup !!! as discussed offline, updating it. :smile:

sbulage:
> @sbulage PRT still failing with
> 
> ```
> E       ValueError: not enough values to unpack (expected 1, got 0)
> 
> cfme/base/__init__.py:263: ValueError
> ValueError
> b'not enough values to unpack (expected 1, got 0)'
> ```

Due to above failure--> `cfme/tests/configure/test_zones.py -->  test_zone_add_maxlength` is also failing so added into pytest run.

sbulage:
Fix Zone related changes in separate PR: #9628.

john-dupuy:
@sbulage hitting a different error in 5.10 now
```
>               'Could not find an element {}'.format(repr(locator))) from None
E           selenium.common.exceptions.NoSuchElementException: Message: Could not find an element './a'

.cfme_venv/lib64/python3.7/site-packages/widgetastic/browser.py:352: NoSuchElementException
NoSuchElementException
b"Message: Could not find an element './a'\n"
```

tpapaioa:
PR #9628 assumes that zone names are unique. This is not true in general (i.e., in multi-region configurations). My PR #9590 reverts that change and introduces other changes that deal with 'ValueError: not enough values to unpack (expected 1, got 0)' exceptions. The create_zone() method in cfme/tests/configure/test_zones.py creates the new zone and ensures that the Zone instance's id attribute is set.

sbulage:
> PR #9628 assumes that zone names are unique. This is not true in general (i.e., in multi-region configurations). My PR #9590 reverts that change and introduces other changes that deal with 'ValueError: not enough values to unpack (expected 1, got 0)' exceptions. The create_zone() method in cfme/tests/configure/test_zones.py creates the new zone and ensures that the Zone instance's id attribute is set.

Thanks @tpapaioa, I might have added temporary solution then. 💯

sbulage:
> @sbulage hitting a different error in 5.10 now
> 
> ```
> >               'Could not find an element {}'.format(repr(locator))) from None
> E           selenium.common.exceptions.NoSuchElementException: Message: Could not find an element './a'
> 
> .cfme_venv/lib64/python3.7/site-packages/widgetastic/browser.py:352: NoSuchElementException
> NoSuchElementException
> b"Message: Could not find an element './a'\n"
> ```

Yeah @john-dupuy, and at tear down only, whole test is passing 😄

sbulage:
Test on CF5.10 is passing locally, not getting issue related to `NoSuchElementException`.

sbulage:
Test is successfully passed locally ran using wharf driver.

john-dupuy:
Test case looks good, I think we should try to track down the 5.10 issue before merging however.

dajoRH:
I detected some fixture changes in commit 1c6189bf66a71f58002989aa083ab93ae2bc5207

The *local* fixture **`new_zone`** is used in the following files:
- **cfme/tests/ansible/test_embedded_ansible_basic.py**
    - *test_embed_tower_repo_add_new_zone*

mshriver:
Also concerned about the 5.10 behavior in PRT - the screenshot shows it still on the playbook page, when its supposed to be deleting a zone.

PR #9494, opened 2019-10-16
[1LP][RFR] Add methods to add and delete folders and subfolders

valaparthvi:
## Purpose or Intent

- __Fixing__ 
    1. `add_folder` and `add_subfolder` of ReportMenus
    2. Fix `add` method of `FolderManager` widget of wt.manageiq
- __Extending__ 
    1. Add `remove_folder` and `remove_subfolder` to ReportMenus
    2. Add ` delete` method to `FolderManager` widget of wt.manageiq
- __Adding tests__ 
    1. `test_reports_menu_folder_crd` to `test_menus.py`

### PRT Run
{{ pytest: cfme/tests/intelligence/reports/test_menus.py -k "test_custom_reports_menu" -vvvv }}

### DIFF
--- a/cfme/tests/intelligence/reports/test_menus.py
+++ b/cfme/tests/intelligence/reports/test_menus.py
@@ -1,12 +1,14 @@
 # -*- coding: utf-8 -*-
 import random
 
+import fauxfactory
 import pytest
 
 from cfme import test_requirements
 from cfme.base.credential import Credential
 from cfme.rest.gen_data import users as _users
 from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.blockers import BZ
 from cfme.utils.conf import cfme_data
 from cfme.utils.ftp import FTPClientWrapper
 from cfme.utils.path import data_path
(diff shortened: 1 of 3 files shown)


widgetastic_manageiq/__init__.py
```diff
@@ -3893,11 +3894,15 @@ def selected_field(self):
...
             return self.selected_field_element.text.encode("utf-8").strip()
 
-    def add(self, subfolder):
-        self.add_subfolder()
-        wait_for(lambda: self.selected_field_element is not None, num_sec=5, delay=0.1)
-        self.browser.double_click(self.selected_field_element, wait_ajax=False)
-        self.browser.handle_alert(prompt=subfolder)
+    def add(self, folder_name):
+        self.add_button.click()
+        wait_for(lambda: self.browser.elements(self._new_field) is not None, num_sec=5, delay=0.1)
```
digitronik: is it ok to use like `bool(self.browser.elements(self._new_field))`? actually elements return `list` of web element

valaparthvi: I could use that bool statement, I was just trying to make it work so I didn't realize I could have also enhanced that piece of code. Will change that!

cfme/tests/intelligence/reports/test_menus.py
```diff
@@ -259,3 +261,48 @@ def test_reset_report_menus(appliance, get_custom_report, group, report_menus):
     assert not appliance.collections.reports.instantiate(
         type=folder, subtype=subfolder, menu_name=get_custom_report.menu_name
     ).exists
+
+
+@pytest.mark.parametrize("group", GROUPS)
+@pytest.mark.meta(coverage=[1762363])
+def test_custom_reports_menu_crd(
```
mshriver: Please use `test_custom_reports_menu_crud`

valaparthvi: I wasn't updating anything here, which is why I went with `crd`.

PR #9495, opened 2019-10-16
[1LP][RFR] Archive vm custom button

digitronik:
### PRT Run
{{pytest: cfme/tests/automate/custom_button/test_service_objects.py::test_custom_button_dialog_service_archived -vv --use-provider vsphere67-nested}}

### DIFF
--- a/cfme/tests/automate/custom_button/test_service_objects.py
+++ b/cfme/tests/automate/custom_button/test_service_objects.py
@@ -16,6 +16,7 @@ from cfme.utils.appliance import ViaUI
 from cfme.utils.appliance.implementations.ssui import navigate_to as ssui_nav
 from cfme.utils.appliance.implementations.ui import navigate_to as ui_nav
 from cfme.utils.blockers import BZ
+from cfme.utils.log_validator import LogValidator
 from cfme.utils.wait import TimedOutError
 from cfme.utils.wait import wait_for
 
@@ -462,8 +463,13 @@ def test_custom_button_role_access_service(context):
     pass
 
 
-@pytest.mark.manual
-def test_custom_button_dialog_service_archived():
+@test_requirements.customer_stories
+@pytest.mark.meta(automates=[BZ(1439883)])
+@pytest.mark.provider([VMwareProvider], override=True, scope="module")
+@pytest.mark.uncollectif(lambda button_group: "GENERIC" in button_group)
+def test_custom_button_dialog_service_archived(
+    request, appliance, provider, setup_provider, service_vm, button_group, dialog
+):
     """ From Service OPS check if archive vms"s dialog invocation via custom button. ref: BZ1439883
 
     Polarion:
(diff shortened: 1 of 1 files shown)


dajoRH:
*CFME QE Bot*

cfme/tests/automate/custom_button/test_service_objects.py
```diff
@@ -462,8 +463,13 @@ def test_custom_button_role_access_service(context):
     pass
 
 
-@pytest.mark.manual
-def test_custom_button_dialog_service_archived():
+@test_requirements.customer_stories
+@pytest.mark.meta(automates=[BZ(1439883)])
+@pytest.mark.provider([VMwareProvider], override=True, scope="module")
```
john-dupuy: Why `module` scope here? If it is module scope, why not put it at the top of the file in `pytestmark`?

cfme/tests/automate/custom_button/test_service_objects.py
```diff
@@ -484,7 +490,62 @@ def test_custom_button_dialog_service_archived():
...
+        button = group.buttons.create(
+            text=fauxfactory.gen_alphanumeric(start="btn_"),
+            hover=fauxfactory.gen_alphanumeric(start="hover_"),
+            dialog=dialog,
+            system="Request",
+            request="InspectMe",
+        )
+    request.addfinalizer(button.delete_if_exists)
+
+    for with_vm in [True, False]:  # [vm, archive_vm]
```
john-dupuy: This seems like a parameter to me -- why have you decided not to parametrize?

cfme/tests/automate/custom_button/test_service_objects.py
```diff
@@ -484,7 +490,62 @@ def test_custom_button_dialog_service_archived():
...
+        if not with_vm:
+            # Make vm archive by deleting vm from provider side
+            vm.mgmt.delete()
+            vm.wait_for_vm_state_change(
+                desired_state="archived", timeout=720, from_details=False, from_any_provider=True
+            )
+
+        for context in [ViaUI, ViaSSUI]:  # check execution with UI and SSUI
```
john-dupuy: Same question about parametrization here.

cfme/tests/automate/custom_button/test_service_objects.py
digitronik: The test was simple means to check button execution on `service` having vm in `archive` state... but just on loop give me more checks :D.

cfme/tests/automate/custom_button/test_service_objects.py
digitronik: I think we discuss this before also... Trying to avoid create/delete(form provider side) `VM` operations as we just need to execute button on SSUI with same setup.

cfme/tests/automate/custom_button/test_service_objects.py
digitronik: Actually most of test in this file using `generic service` no need of provider that why I avoid mark provider globally. The second why module yup no need of it even function scope work but it will not effect. can i move to fictional scope?

john-dupuy: Yeah kind of a minor point but I would leave it at `function` scope.

cfme/tests/automate/custom_button/test_service_objects.py
```diff
@@ -484,7 +490,62 @@ def test_custom_button_dialog_service_archived():
...
+                # Check for request in automation log
+                try:
+                    wait_for(
+                        lambda: log.matches[request_pattern] == 1,
+                        timeout=180,
+                        message="wait for expected match count",
+                        delay=5,
+                    )
+                except TimedOutError:
+                    assert False, f"Expected '1' requests; found '{log.matches[request_pattern]}'"
```
john-dupuy: _Optional_ I suggest using `pytest.fail` rather than `assert False` here

PR #9501, opened 2019-10-17
[1LP][RFR] miq appliance minor upgrade 511 support

digitronik:
## Purpose or Intent
- added support for `minor` upgrade of 5.11.z
- Major need more work... for now its not supporting
- Avoid some operation which not needed for minor upgrade.


digitronik:
@jadhaj please take look..

PR #9505, opened 2019-10-17
[1LP][RFR] Automate test-pwd tests, delete manual cases

john-dupuy:
Automating some manual test cases that test password types for db auth users. Deleting a manual case that is covered by `test_login.py`

{{ pytest: cfme/tests/integration/test_db_auth.py }}

### DIFF
--- a/cfme/tests/integration/test_auth_manual.py
+++ b/cfme/tests/integration/test_auth_manual.py
@@ -723,93 +723,16 @@ def test_session_timeout():
 
 @pytest.mark.manual
 @pytest.mark.tier(2)
-def test_pwd_trailing_whitespace():
-    """
-    Test changing password to one with trailing whitespace.
-
-    Polarion:
-        assignee: jdupuy
-        casecomponent: Auth
-        caseimportance: medium
-        initialEstimate: 1/8h
-    """
-    pass
-
-
-@pytest.mark.manual
-@pytest.mark.tier(2)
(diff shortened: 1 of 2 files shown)


cfme/tests/integration/test_auth_manual.py
```diff
@@ -719,97 +719,3 @@ def test_session_timeout():
...
-        casecomponent: Auth
-        caseimportance: medium
-        initialEstimate: 1/8h
-    """
-    pass
-
-
-@pytest.mark.manual
-@pytest.mark.tier(2)
-def test_login_invalid_user():
```
sshveta: Nice PR ! Is this TC also automated already ?

john-dupuy: There are tests that cover bad passwords and the like in `test_login.py` but nothing that covers an invalid user. I am not really certain how to determine if a user is invalid and the test case doesn't provide many details. Possibly it's a user that just isn't defined in the DB, which would be easiest enough to automate.

So, I guess I will leave this test case for now and automate it in a separate PR.

dajoRH:
I detected some fixture changes in commit b7095b26bf6f887e69f07dff285295dcf34b3261

The *local* fixture **`user`** is used in the following files:
- **cfme/tests/integration/test_db_auth.py**
    - *test_db_user_pwd*

PR #9506, opened 2019-10-17
[1LP][RFR] OSP flavor and security group

sshveta:
Added a test to migrate a VM with different flavor and security group in a different project in OSP and check if the migrated VM got the correct flavor and security group.

{{ pytest: cfme/tests/v2v/test_v2v_migrations_ui.py --use-provider osp13-ims --use-provider vsphere65-nested --provider-limit 2 -vvvv --long-running -k test_v2v_custom_attribute }}

### DIFF
--- a/cfme/tests/v2v/test_v2v_migrations_ui.py
+++ b/cfme/tests/v2v/test_v2v_migrations_ui.py
@@ -571,7 +571,8 @@ def test_migration_with_no_conversion(appliance, delete_conversion_hosts, source
     assert "no conversion host" in view.progress_card.get_error_text(migration_plan.name)
 
 
-@pytest.mark.provider([OpenStackProvider], scope='function')
+@pytest.mark.provider([OpenStackProvider], selector=ONE_PER_VERSION,
+                      required_flags=["v2v"], override=True, scope='module')
 @pytest.mark.parametrize("attribute", ["flavor", "security_group"])
 @pytest.mark.parametrize(
     "mapping_data_vm_obj_single_datastore", [["nfs", "nfs", rhel7_minimal]], indirect=True)
--- a/cfme/tests/v2v/test_v2v_migrations_ui.py
+++ b/cfme/tests/v2v/test_v2v_migrations_ui.py
@@ -571,7 +571,7 @@ def test_migration_with_no_conversion(appliance, delete_conversion_hosts, source
     assert "no conversion host" in view.progress_card.get_error_text(migration_plan.name)
 
 
-@pytest.mark.uncollectif(lambda provider: provider.one_of(RHEVMProvider))
+@pytest.mark.provider([OpenStackProvider], scope='function')
 @pytest.mark.parametrize("attribute", ["flavor", "security_group"])
 @pytest.mark.parametrize(
     "mapping_data_vm_obj_single_datastore", [["nfs", "nfs", rhel7_minimal]], indirect=True)
--- a/cfme/tests/v2v/test_v2v_migrations_ui.py
+++ b/cfme/tests/v2v/test_v2v_migrations_ui.py
@@ -2,10 +2,13 @@
 import fauxfactory
 import pytest
 from widgetastic.exceptions import NoSuchElementException
+from widgetastic.utils import partial_match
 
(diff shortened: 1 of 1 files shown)


cfme/tests/v2v/test_v2v_migrations_ui.py
```diff
@@ -566,3 +569,74 @@ def _cleanup():
...
+    infrastructure_mapping_collection = appliance.collections.v2v_infra_mappings
+    mapping_data = mapping_data_vm_obj_single_datastore.infra_mapping_data
+
+    # Changing target cluster to other than default(admin to qe-auto)
+    component = mapping_data["clusters"][0]
+    # Removed 'admin' (default project) from component
+    component.targets.pop()
+    # Add 'qe-auto' (different project) to component
+    component.targets.append(partial_match(provider.data.clusters[1]))
+    map_cluster = component.targets[0]
```
john-dupuy: Q: can you do?

```suggestion
component.targets = [partial_match(provider.data.clusters[1])]
map_cluster = component.targets[0]
```

cfme/tests/v2v/test_v2v_migrations_ui.py
```diff
@@ -566,3 +569,74 @@ def _cleanup():
...
+    # create mapping with 'qe-auto' project
+    mapping = infrastructure_mapping_collection.create(**mapping_data)
+    src_vm_obj = mapping_data_vm_obj_single_datastore.vm_list[0]
+
+    # Selecting second flavor and security group selection since first is default
+    map_flavor = provider.data.flavors[1] if attribute == "flavor" else provider.data.flavors[0]
```
john-dupuy: Perhaps add a small comment about what flavors these indexes correspond to in yamls

cfme/tests/v2v/test_v2v_migrations_ui.py
```diff
@@ -566,3 +569,74 @@ def _cleanup():
...
+        infra_map=mapping.name,
+        osp_security_group=map_security_group,
+        osp_flavor=map_flavor,
+        target_provider=provider,
+        vm_list=mapping_data_vm_obj_single_datastore.vm_list,
+    )
+    assert migration_plan.wait_for_state("Started")
+    assert migration_plan.wait_for_state("In_Progress")
+    assert migration_plan.wait_for_state("Completed")
+    assert migration_plan.wait_for_state("Successful")
```
john-dupuy: Do we need to check each step of the way? Can we just assert success?

sshveta: We cannot @john-dupuy . Each of these are different methods .Only In_progress wait for plan to finish , rest all just checks plan status. So we have to go sequentially . 
Initially it was a single method that did everything but lot of tests requires only few of these states so made them individual methods.

cfme/tests/v2v/test_v2v_migrations_ui.py
sshveta: Done

sshveta:
> @sshveta Thoughts on moving the migration_plan creation and its steps to a fixture? Nice test case! I have some small questions

@johndupuy : Migration plan takes infrastructure mapping as parameter and in this test case flavor and security group too .Hence not making it a fixture

mshriver:
False-positive PRT, osp13-ims didn't setup and tests were skipped. restarting.

sshveta:
> False-positive PRT, osp13-ims didn't setup and tests were skipped. restarting.

@mshriver : OSP13-ims is being redeployed currently . Tests won't pass

cfme/tests/v2v/test_v2v_migrations_ui.py
```diff
@@ -566,3 +569,73 @@ def _cleanup():
     view = navigate_to(migration_plan, "InProgress")
     assert not view.progress_card.is_plan_started(migration_plan.name)
     assert "no conversion host" in view.progress_card.get_error_text(migration_plan.name)
+
+
+@pytest.mark.provider([OpenStackProvider], selector=ONE_PER_VERSION,
```
mshriver: Thanks @sshveta !!

PR #9508, opened 2019-10-17
[1LP][RFR] Add parameterized fixture to test_infrastructure_hosts_navigation_aft…

prichard77:
…er_download to cover appliance and provider host collections

## Purpose or Intent

- __Updating test__  test_infrastructure_hosts_navigation_after_download. Adding support for both provider and appliance host collections.

### PRT Run
{{ pytest: --use-provider complete --long-running cfme/tests/infrastructure/test_host.py::test_infrastructure_hosts_navigation_after_download}}

### DIFF
--- a/cfme/tests/infrastructure/test_host.py
+++ b/cfme/tests/infrastructure/test_host.py
@@ -409,7 +409,7 @@ def test_infrastructure_hosts_refresh_multi(appliance, setup_provider_min_hosts,
     ids=["txt", "csv", "pdf"]
 )
 def test_infrastructure_hosts_navigation_after_download(
-    appliance, setup_provider, provider, report_format
+    appliance, setup_provider, provider, report_format, hosts_collection
 ):
     """
     Polarion:
@@ -421,10 +421,20 @@ def test_infrastructure_hosts_navigation_after_download(
         1738664
 
     """
-    hosts_view = navigate_to(provider.collections.hosts, "All")
+    if hosts_collection == "provider":
+        hosts_view = navigate_to(provider.collections.hosts, "All")
+    elif hosts_collection == "appliance":
+        hosts_view = navigate_to(appliance.collections.hosts, "All")
     hosts_view.toolbar.download.item_select(report_format)
     if report_format == "Print or export as PDF":
         handle_extra_tabs(hosts_view)
     hosts_view.navigation.select("Compute")
-    provider_view = provider.create_view(InfraProviderDetailsView)
-    assert provider_view.is_displayed
+    if hosts_collection == "provider":
+        provider_view = provider.create_view(InfraProviderDetailsView)
+        assert provider_view.is_displayed
+    elif hosts_collection == "appliance":
+        assert hosts_view.is_displayed
+
+@pytest.fixture(params=["provider", "appliance"])
+def hosts_collection(request):
+    return request.param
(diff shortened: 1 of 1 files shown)


dajoRH:
I detected some fixture changes in commit b8aece8b1fc6cf3bf8bf503da1959f1c4b648999

 The *local* fixture **`hosts_collection`** was changed, but I didn't find where it's used.

cfme/tests/infrastructure/test_host.py
```diff
@@ -421,10 +421,21 @@ def test_infrastructure_hosts_navigation_after_download(
...
-    assert provider_view.is_displayed
+    if hosts_collection == "provider":
+        provider_view = provider.create_view(InfraProviderDetailsView)
+        assert provider_view.is_displayed
+    elif hosts_collection == "appliance":
+        assert hosts_view.is_displayed
+
+
+@pytest.fixture(params=["provider", "appliance"])
+def hosts_collection(request):
```
john-dupuy: Why use this method over a standard parametrization? Are you concerned about the test order? By standard parametrization I mean,

```python
@pytest.mark.parametrize("hosts_collection", ["provider", "appliance"])
```

At the top of the test function.

prichard77: Fixtures must be parameterized using this format AFAIK. The test is already paramtererised. This way we get the 3 cases in the test each run for provider and appliance, 6 tests total for each provider.

john-dupuy: You're right about fixtures having to be parametrized using this format. But this parametrizing doesn't need a fixture. If you put `@pytest.mark.parametrize("hosts_collection", ["provider", "appliance"])`, you should get the same 6 test cases (for each provider) that you do from from parametrized fixture. You can have two (or more) parametrizations on a single test case and pytest will create a "test matrix" for all the parameters.

Note that the ordering may be different.

cfme/tests/infrastructure/test_host.py
```diff
@@ -421,10 +422,16 @@ def test_infrastructure_hosts_navigation_after_download(
         1738664
 
     """
-    hosts_view = navigate_to(provider.collections.hosts, "All")
+    if hosts_collection == "provider":
+        hosts_view = navigate_to(provider.collections.hosts, "All")
+    elif hosts_collection == "appliance":
```
john-dupuy: _Optional_ since there are only 2 parameters this can just be `else:`

cfme/tests/infrastructure/test_host.py
```diff
@@ -421,10 +422,16 @@ def test_infrastructure_hosts_navigation_after_download(
...
     hosts_view.toolbar.download.item_select(report_format)
     if report_format == "Print or export as PDF":
         handle_extra_tabs(hosts_view)
     hosts_view.navigation.select("Compute")
-    provider_view = provider.create_view(InfraProviderDetailsView)
-    assert provider_view.is_displayed
+    if hosts_collection == "provider":
+        provider_view = provider.create_view(InfraProviderDetailsView)
+        assert provider_view.is_displayed
+    elif hosts_collection == "appliance":
```
john-dupuy: Same comment here: can just be `else:`

john-dupuy:
Also I just noticed that with all providers this parametrizes to 42 test cases! Does this need to be run for every single infra provider? Or can we run it just a single infra provider or just a single infra provider type?

prichard77:
This only runs on 2 providers, 12 tests (at least when I run it locally). Maybe I need to look more closely. I thought these lines restricted the procviders ```pytestmark = [
    pytest.mark.tier(3),
    pytest.mark.provider([InfraProvider],
                         required_fields=['hosts'], scope='module',
                         selector=ONE_PER_VERSION),
]```

PR #9512, opened 2019-10-18
[1LP][RFR] Adding new tests of VM reconfig resize disk

niyazRedhat:
{{ pytest: cfme/tests/infrastructure/test_vm_reconfigure.py  --use-provider vsphere6-nested -vvvv --long-running -k test_vm_reconfig_resize_disk  }}

- Automating VM reconfiguration resize disk

- Removing test_vm_reconfig_attach_iso_vsphere67_nested because it's already automated

- This PR will automate 30 test cases

### DIFF
--- a/cfme/infrastructure/virtual_machines.py
+++ b/cfme/infrastructure/virtual_machines.py
@@ -400,7 +400,7 @@ class InfraVmReconfigureView(BaseLoggedInPage):
         column_widgets={
             "Type": BootstrapSelect(id="hdType"),
             "Mode": BootstrapSelect(id="hdMode"),
-            "Size": WInput(id="dvcSize"),
+            "Size": WInput(locator='//*[@id="dvcResize" or @id="dvcSize"]'),
             "ControllerType": BootstrapSelect(id="Controller"),
             "Unit": BootstrapSelect(id="hdUnit"),
             "Dependent": BootstrapSwitch(name="vm.cb_dependent"),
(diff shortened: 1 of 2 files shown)


cfme/infrastructure/virtual_machines.py
```diff
@@ -667,6 +680,13 @@ def get_changes_to_fill(self, other_configuration):
                 changes['disks'].append({'action': 'delete', 'disk': disk, 'delete_backing': None})
             elif disk not in self.disks and disk in other_configuration.disks:
                 changes['disks'].append({'action': 'add', 'disk': disk})
+            elif disk in self.disks and disk in other_configuration.disks:
+                old_disk = [d for d in self.disks if d == disk][0]
+                new_disk = [d for d in other_configuration.disks if d == disk][0]
```
psimovec: [Optional] 
it would look better to me to use here something like
next(d for d in self.disks if d == disk)
it won't generate full list, just find first value :)

cfme/infrastructure/virtual_machines.py
```diff
@@ -667,6 +680,13 @@ def get_changes_to_fill(self, other_configuration):
                 changes['disks'].append({'action': 'delete', 'disk': disk, 'delete_backing': None})
             elif disk not in self.disks and disk in other_configuration.disks:
                 changes['disks'].append({'action': 'add', 'disk': disk})
+            elif disk in self.disks and disk in other_configuration.disks:
```
psimovec: [Optional]
I think that there could be just 'else:', as it's iterating through these two lists. It is already checked if a disk is just in one of the lists, so this condition should be always True (if the lists are not modified during this for loop)

cfme/tests/infrastructure/test_vm_reconfigure.py
```diff
@@ -248,58 +249,63 @@ def test_vm_reconfig_add_remove_hw_hot(
...
-            3. Go to VM reconfiguration
-            4. Click on Add Disk -> select disk_type and disk_mode , save and submit
-            5. Go to VM reconfiguration and resize the disk
-            6. Check the changes in VM reconfiguration page
-            7. Remove the disk
-    """
-    pass
+@pytest.mark.uncollectif(
+    lambda disk_mode, vm_state: disk_mode == 'independent_nonpersistent' and vm_state == 'hot',
+    reason='The disk resize operation not supported for hot-independent_nonpersistent')
```
mshriver: :ok_hand:

PR #9531, opened 2019-10-23
[1LP][RFR] Updates to test_catalog.py and Remove Flash Assert in Catalog Entity

jawatts:
This PR addresses two updates:
1) Updating test_catalog.py
2) Removing the flash message success assertions from the catalog entity and moves them to the crud test

{{ pytest: cfme/tests/services/test_catalog.py::test_catalog_crud }}

### DIFF
--- a/cfme/services/catalogs/catalog.py
+++ b/cfme/services/catalogs/catalog.py
@@ -17,7 +17,10 @@ from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.appliance.implementations.ui import navigator
 from cfme.utils.pretty import Pretty
 from cfme.utils.update import Updateable
+from cfme.utils.version import LOWEST
+from cfme.utils.version import VersionPicker
 from widgetastic_manageiq import MultiBoxSelect
+from widgetastic_manageiq import ReactTextInput
 
 
 class CatalogsMultiBoxSelect(MultiBoxSelect):
@@ -34,8 +37,14 @@ class CatalogsMultiBoxSelect(MultiBoxSelect):
 class CatalogForm(ServicesCatalogView):
     title = Text('#explorer_title_text')
 
-    name = Input(name='name')
-    description = Input(name="description")
+    name = VersionPicker({
+        "5.11": ReactTextInput(name='name'),
+        LOWEST: Input(name='name')
+    })
+    description = VersionPicker({
+        "5.11": ReactTextInput(name="description"),
+        LOWEST: Input(name='description')
+    })
     assign_catalog_items = CatalogsMultiBoxSelect(
         move_into="Move Selected buttons right",
         move_from="Move Selected buttons left",
--- a/cfme/services/catalogs/catalog.py
+++ b/cfme/services/catalogs/catalog.py
@@ -111,20 +111,12 @@ class Catalog(BaseEntity, Updateable, Pretty, Taggable):
             view.cancel_button.click()
         view = self.create_view(DetailsCatalogView, override=updates, wait='10s')
         view.flash.assert_no_error()
-        if changed:
-            view.flash.assert_message(
-                'Catalog "{}" was saved'.format(updates.get('name', self.name)))
-        else:
-            view.flash.assert_message(
-                'Edit of Catalog "{}" was cancelled by the user'.format(self.name))
 
     def delete(self):
         view = navigate_to(self, "Details")
         view.configuration.item_select('Remove Catalog', handle_alert=True)
         view = self.create_view(CatalogsView, wait='10s')
         view.flash.assert_no_error()
-        view.flash.assert_success_message(
-            'Catalog "{}": Delete successful'.format(self.description or self.name))
 
 
 @attr.s
--- a/cfme/tests/services/test_catalog.py
+++ b/cfme/tests/services/test_catalog.py
@@ -5,14 +5,18 @@ import pytest
 import cfme.tests.configure.test_access_control as tac
 from cfme import test_requirements
 from cfme.services.catalogs.catalog import CatalogsView
+from cfme.services.catalogs.catalog import DetailsCatalogView
+from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.blockers import BZ
 from cfme.utils.update import update
+
 
 pytestmark = [test_requirements.service, pytest.mark.tier(2)]
 
 
 @pytest.mark.rhel_testing
 @pytest.mark.sauce
-def test_catalog_crud(appliance):
+def test_catalog_crud(request, appliance):
     """
     Polarion:
         assignee: nansari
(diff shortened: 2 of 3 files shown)


cfme/tests/services/test_catalog.py
```diff
@@ -5,31 +5,55 @@
...
     """
+    # Create Catalog
     catalog_name = fauxfactory.gen_alphanumeric()
     cat = appliance.collections.catalogs.create(name=catalog_name, description='my catalog')
+    request.addfinalizer(cat.delete_if_exists)
 
     view = cat.create_view(CatalogsView)
     assert view.is_displayed
-    view.flash.assert_success_message('Catalog "{}" was saved'.format(catalog_name))
+    if appliance.version >= '5.11':
```
jawatts: I think this may be a bug, @niyazRedhat is checking with the dev on this. Name was displayed in 5.10 but in 5.11 it was removed. But on the delete message, it uses the description.

cfme/tests/services/test_catalog.py
```diff
@@ -5,31 +5,55 @@
...
+    view = cat.create_view(DetailsCatalogView)
+    assert view.wait_displayed()
+    view.flash.assert_message(f'Edit of Catalog "{catalog_name}" was cancelled by the user')
+    assert cat.description == "my edited description"
+
+    # Delete Catalog
     cat.delete()
+    view = cat.create_view(CatalogsView)
+    view.flash.assert_success_message(
+        f'Catalog "{cat.description}": Delete successful')
```
jawatts: @niyazRedhat why does it use the description in 5.10 and 5.11, instead of the catalog name.

cfme/tests/services/test_catalog.py
```diff
@@ -5,31 +5,59 @@
...
+        saved_message = f'Catalog "{catalog_name}" was saved'
+    view.flash.assert_success_message(saved_message)
 
+    # Edit Catalog
+    update_descr = 'my edited description'
     with update(cat):
-        cat.description = 'my edited description'
+        cat.description = update_descr
+
+    assert cat.description == update_descr
```
digitronik: optional:
Along with `object` `attribute` check, I think we can check `cat.exist` (UI check)

cfme/tests/services/test_catalog.py
```diff
@@ -5,31 +5,59 @@
...
+
+    view.flash.assert_success_message(saved_message)
+
+    view = navigate_to(cat, 'Edit')
+    view.fill(value={'description': 'test_cancel'})
+    view.cancel_button.click()
+    view = cat.create_view(DetailsCatalogView)
+    assert view.wait_displayed()
```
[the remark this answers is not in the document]
john-dupuy: Unless I'm mistaken, this changes the structure of the `assert`. It changes from a test _failure_ to a test _error_. Don't we want a test failure in this case?

cfme/tests/services/test_catalog.py
```diff
@@ -5,31 +5,59 @@
...
         initialEstimate: 1/8h
         tags: service
     """
+    # Create Catalog
     catalog_name = fauxfactory.gen_alphanumeric()
     cat = appliance.collections.catalogs.create(name=catalog_name, description='my catalog')
+    request.addfinalizer(cat.delete_if_exists)
 
     view = cat.create_view(CatalogsView)
     assert view.is_displayed
```
john-dupuy: why not `view.wait_displayed()`?

cfme/tests/services/test_catalog.py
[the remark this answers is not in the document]
digitronik: I think in both case we will face error `TimedOutError`. I think assertion is better option for `bool`.

cfme/tests/services/test_catalog.py
jawatts: :+1: Added in the create section

cfme/tests/services/test_catalog.py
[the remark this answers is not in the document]
jawatts: added

cfme/tests/services/test_catalog.py
jawatts: BZ 1765107 was written

cfme/tests/services/test_catalog.py
jawatts: updated to use `wait` are in `create_view`

john-dupuy: Wait, I don't see the added `wait`? Has it been pushed?

widgetastic_manageiq/__init__.py
```diff
@@ -4547,6 +4547,32 @@ def data_for_timestamp(self, timestamp):
...
+class ReactTextInput(TextInput):
+
+    # The clear method from the WebElement class in the Selenium package was not clearing the react
+    # text field. arguments[0] is the widget name
+    clear_text_field = """\
+        (function(elem){
+            var elem_name = document.getElementsByName(elem);
+            var elem_id = elem_name[0].id;
+            document.getElementById(elem_id).value = "";
+        }(arguments[0]));
```
jawatts: I am not stuck on this current implementation. If someone had a better way to do this, will be happy to see!

digitronik: If you like; its one of good hack over `TextInput` or `TextArea`

```
In [12]: from selenium.webdriver.common.keys import Keys  
In [13]: el.send_keys(Keys.CONTROL + "a")                 
In [14]: el.send_keys(Keys.DELETE)
```
only better thing; It will remove unique widget dependency.

widgetastic_manageiq/__init__.py
```diff
@@ -4547,6 +4547,32 @@ def data_for_timestamp(self, timestamp):
...
+    """
+
+    def clear(self):
+        self.browser.execute_script(self.clear_text_field, self.name)
+
+    def fill(self, value):
+        current_value = self.value
```
digitronik: Optional: I would like to avoid variable declaration if its not reusable.

cfme/services/catalogs/catalog.py
```diff
@@ -34,8 +37,14 @@ class CatalogsMultiBoxSelect(MultiBoxSelect):
 class CatalogForm(ServicesCatalogView):
     title = Text('#explorer_title_text')
 
-    name = Input(name='name')
-    description = Input(name="description")
+    name = VersionPicker({
+        "5.11": ReactTextInput(name='name'),
+        LOWEST: Input(name='name')
```
digitronik: Q: is `ReactTextInput` not working for both versions?
If working; I don't think we need this `VersionPicker`.

cfme/tests/services/test_catalog.py
```diff
@@ -5,31 +5,65 @@
...
         tags: service
     """
+    # Create Catalog
     catalog_name = fauxfactory.gen_alphanumeric()
     cat = appliance.collections.catalogs.create(name=catalog_name, description='my catalog')
+    request.addfinalizer(cat.delete_if_exists)
 
-    view = cat.create_view(CatalogsView)
+    view = cat.create_view(CatalogsView, wait="10s")
     assert view.is_displayed
```
izapolsk: I guess this is_displayed isn't necessary after adding `wait=`

cfme/tests/services/test_catalog.py
```diff
@@ -5,31 +5,65 @@
...
-    view.flash.assert_success_message('Catalog "{}" was saved'.format(catalog_name))
+    if BZ(1766276, forced_streams=['5.11']).blocks:
+        saved_message = f"Catalog was saved"
```
izapolsk: this isn't formatted string

cfme/tests/services/test_catalog.py
```diff
@@ -5,31 +5,65 @@
...
+
+    assert cat.description == update_descr
+
+    view.flash.assert_success_message(saved_message)
+
+    view = navigate_to(cat, 'Edit')
+    view.fill(value={'description': 'test_cancel'})
+    view.cancel_button.click()
+    view = cat.create_view(DetailsCatalogView, wait="10s")
+    assert view.is_displayed
```
izapolsk: ^^

cfme/services/catalogs/catalog.py
jawatts: It does work for both versions and I originally had it as just the `ReactTextInput` but @mshriver wanted to see it be explicit

widgetastic_manageiq/__init__.py
```diff
@@ -4547,6 +4547,32 @@ def data_for_timestamp(self, timestamp):
...
+    # text field. arguments[0] is the widget name
+    clear_text_field = """\
+        (function(elem){
+            var elem_name = document.getElementsByName(elem);
+            var elem_id = elem_name[0].id;
+            document.getElementById(elem_id).value = "";
+        }(arguments[0]));
+    """
+
+    def clear(self):
```
mshriver: So I think there's a good enhancement here for `Widgetastic.widget.input.TextInput`.

Defining `TextInput.clear`, and modifying its fill implementation to use `self.clear` instead of `self.browser.clear` directly would allow for subclassing widgets to define their own clear mechanisms, without having to re-define `TextInput.fill` locally.

This would need to be done in wt.core, and this PR isn't blocked by this - but it would allow for a slimmer implementation here.

PR #9534, opened 2019-10-24
[1LP][RFR] New Test: Testing automate method called more than expected

ganeshhubale:

## Purpose or Intent
- Testing customer scenario where automate method called multiple times where it need not be called.

### PRT Run
{{ pytest: cfme/tests/automate/test_service_dialog.py::test_automate_methods_from_dynamic_dialog_should_run_as_per_designed --long-running --use-template-cache -qsvvv}}

### DIFF
--- a/cfme/tests/automate/test_service_dialog.py
+++ b/cfme/tests/automate/test_service_dialog.py
@@ -4,7 +4,11 @@ import pytest
 
 from cfme import test_requirements
 from cfme.automate.dialogs.service_dialogs import DialogsView
+from cfme.fixtures.automate import DatastoreImport
+from cfme.services.service_catalogs import ServiceCatalogs
+from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.appliance.implementations.ui import navigator
+from cfme.utils.log_validator import LogValidator
 from cfme.utils.update import update
 
 pytestmark = [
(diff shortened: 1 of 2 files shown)


cfme/tests/automate/test_service_dialog.py
```diff
@@ -418,3 +422,60 @@ def test_dialog_items_default_values_on_different_screens():
...
+        catalog=catalog,
+        dialog=sd,
+    )
+    request.addfinalizer(catalog_item.delete_if_exists)
+    service_catalogs = ServiceCatalogs(appliance, catalog_item.catalog, catalog_item.name)
+    patterns = [
+        ".*CC- dialog_instance1 value=.*",
+        ".*CC- dialog_instance2 value=.*",
+        ".*CC- dialog_instance3 value=.*",
+    ]
```
jawatts: @ganeshhubale I think we may need to do some more investigation here. Looking at the BZ, the problem was:
```
 method for Element A
  method for Element B
  method for Element C
  method for Element A
  method for Element B
  method for Element C```
was being logged.

The log validator in this case I believe would pass in this context since each pattern was hit. But really we need to be checking the count of those instances in the log. They should be logged only once each.  Let me know your thoughts here.

Comment (ganeshhubale) [reply cfme/tests/automate/test_service_dialog.py]: Thanks for this suggestion :)
I have updated test case as below. Please let me know if we need more changes:
```
patterns = [
        ".*CC- dialog_instance1 value=.*",
        ".*CC- dialog_instance2 value=.*",
        ".*CC- dialog_instance3 value=.*",
    ]

    # Checking if automates method gets triggered by three instances for once after ordering catalog
    # item
    result = LogValidator(
        "/var/www/miq/vmdb/log/automation.log", matched_patterns=patterns
    )
    result.start_monitoring()
    view = navigate_to(service_catalogs, "Order")
    for i in range(len(patterns)):
        assert result.matches[patterns[i]] == 1
```

File: cfme/tests/automate/test_service_dialog.py [hunk 2]
```diff
@@ -418,3 +422,64 @@ def test_dialog_items_default_values_on_different_screens():
...
+
+    # Checking if automates method gets triggered by three instances for once after ordering catalog
+    # item
+    result = LogValidator(
+        "/var/www/miq/vmdb/log/automation.log", matched_patterns=patterns
+    )
+    result.start_monitoring()
+    view = navigate_to(service_catalogs, "Order")
+    for i in range(len(patterns)):
```
mshriver: This seems like total overkill to me, why not just iterate over the patterns?

```
for pattern in patterns:
    assert result.matches[pattern] == 1
```

cfme/tests/automate/test_service_dialog.py
```diff
@@ -417,3 +421,64 @@ def test_dialog_items_default_values_on_different_screens():
...
+    for pattern in patterns:
+        assert result.matches[pattern] == 1
+
+    with LogValidator("/var/www/miq/vmdb/log/automation.log", failure_patterns=patterns).waiting(
+        timeout=120
+    ):
+        # Values like 'label1'(value of ele_label), 'label2' and label3 are element names of three
+        # different text boxes attached with service dialog
+        for ele_name in [ele_label, "label2", "label3"]:
+            # Checking if automate method is not triggered after updating values of dialog widgets
```
mshriver: clever, doing this within logvalidator to wrap your validation!

jawatts:
Looks good! waiting on travis and PRT! nice work @ganeshhubale

PR #9539, opened 2019-10-24
[1LP][RFR][NOTEST] Added BZ param in the already automated tests.

sbulage:
## Purpose or Intent

- [x] Added BZ parameter.


jawatts:
Failures in PRT not due to this change

PR #9545, opened 2019-10-26
[1LP][RFR] Add model pages and tests for Overview > Optimization

valaparthvi:
## Purpose or Intent
- __Adding tests__ 
    `test_optimization.py`
- __Enhancement__ 
    1. Add views and navigators for Overview > Optimization page.
    2. Add collections and entities for Optimization Reports [Work in question?]

### PRT Run
{{ pytest: cfme/tests/intelligence/test_optimization.py -vvv }}


dajoRH:
Lint report for commit c5c2eccc62e412af75c5187b9f7fe8ff512ce69c:

`cfme/intelligence/optimization.py`:
- :bangbang: Line 57:1: **E302** *expected 2 blank lines, found 1*
- :bangbang: Line 64:1: **E302** *expected 2 blank lines, found 1*
- :bangbang: Line 64:1: **E112** *expected an indented block*
- :bangbang: Line 64:1: **E999** *IndentationError: expected an indented block*
- :bangbang: Line 71:1: **E302** *expected 2 blank lines, found 1*
- :bangbang: Line 89:1: **E302** *expected 2 blank lines, found 1*
- :heavy_exclamation_mark: Line 98:111: **W292** *no newline at end of file*

`cfme/tests/intelligence/test_optimization.py`:
- :red_circle: Line 5:12: **F821** *undefined name 'navigate_to'*
- :heavy_exclamation_mark: Line 12:76: **W292** *no newline at end of file*

Hmmm


**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
Lint report for commit d2c85631b40a62f2e3da9cfcc615970de35580c4:

`cfme/tests/intelligence/test_optimization.py`:
- :red_circle: Line 5:12: **F821** *undefined name 'navigate_to'*
- :heavy_exclamation_mark: Line 12:76: **W292** *no newline at end of file*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
Lint report for commit 3b7668956396c83d2459fd8e2bb78b594c7ad581:

`cfme/tests/intelligence/test_optimization.py`:
- :red_circle: Line 5:12: **F821** *undefined name 'navigate_to'*
- :heavy_exclamation_mark: Line 12:76: **W292** *no newline at end of file*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
Lint report for commit cf673e51f32871086ba92152398f6c70e5bc3f35:

`cfme/tests/intelligence/test_optimization.py`:
- :red_circle: Line 5:12: **F821** *undefined name 'navigate_to'*
- :heavy_exclamation_mark: Line 12:76: **W292** *no newline at end of file*

**No commit flaws detected**.

*CFME QE Bot*

valaparthvi:
Moving this to RFR, since PRT failure is expected because of problems with uncollectif.

cfme/tests/intelligence/test_optimization.py
```diff
@@ -0,0 +1,68 @@
...
+    "Host CPU Trends (last week)",
+    "Host Memory Trends (last week)",
+    "Offline VMs with Snapshot",
+    "Top CPU Consumers (weekly)",
+    "Top Memory Consumers (weekly)",
+    "VMs with Volume Free Space >= 75%",
+]
+
+
+@pytest.mark.meta(coverage=[1769346])
```
john-dupuy: Should be `automates`

valaparthvi: Um I thought of that, but the bug is only a small part of the test. I would usually use `automates` if the whole bug was automated by the test, not if the but was a small part of it. But let me know if you think otheriwse :)

mshriver: Maybe more discussion happened offline - I'd agree with @john-dupuy, automates does not imply that the entire (or even a majority) of the test content is concerned with the BZ. Merely that the BZ root cause is automated by this test case.

cfme/intelligence/optimization.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+@attr.s
+class OptimizationReport(BaseEntity):
+    menu_name = attr.ib()
+    last_run_at = attr.ib(default=None)
+    runs = attr.ib(default=None)
+
+    def __attrs_post_init__(self):
+        self._collections = {
```
mshriver: This should not be done in `__attrs_post_init__`, the Base model classes handle this via direct definition within the class.

https://cfme-tests.readthedocs.io/en/master/guides/collections.html#changes-to-previous-model-versions

cfme/intelligence/optimization.py
```diff
@@ -0,0 +1,157 @@
...
+            "saved_reports": SavedReportsCollection,
+            "reports": ReportsCollection,
+        }
+
+    def queue(self):
+        view = navigate_to(self.parent, "All")
+        row = view.table.row(report_name=self.menu_name)
+        row.action.widget.click()
+        view.flash.assert_no_error()
+        last_run_at = row["Last Run at"].text
```
mshriver: I think this should be read before the action widget is clicked, to avoid a race between the two.

cfme/intelligence/optimization.py
```diff
@@ -0,0 +1,157 @@
...
+        wait_for(
+            lambda: row["Last Run at"].text != last_run_at,
+            fail_func=view.refresh.click,
+            num_sec=60,
+            message="Wait for report queue completion",
+            delay=1,
+        )
+        view.refresh.click()
+        self.last_run_at = row["Last Run at"].text
```
mshriver: This should be a property on the entity class, not set during this CRUD type method.  This would facilitate implementation of the property via sentaku, for fetching via UI or via REST.  No need to implement via sentaku in this PR, but this instance attribute should be implemented via a property.

cfme/intelligence/optimization.py
```diff
@@ -0,0 +1,157 @@
...
+
+@attr.s
+class OptimizationReportsCollection(BaseCollection):
+    ENTITY = OptimizationReport
+
+
+@attr.s
+class OptimizationSavedReport(BaseEntity):
+    name = attr.ib()
+    last_run_at = attr.ib()
```
mshriver: This to me seems like a property of the entity, not an attribute to be defined on its instantiation.  Can it be uniquely identified by the name only?

cfme/intelligence/optimization.py
```diff
@@ -0,0 +1,157 @@
...
+        row.click()
+        view = self.create_view(OptimizationReportAllView)
+        return self.saved_optimization_reports.instantiate(
+            name=view.table.row(last_run_at=self.last_run_at)["Report"].text,
+            last_run_at=self.last_run_at,
+            username=self.appliance.user.credential.principal,
+        )
+
+    @cached_property
+    def saved_optimization_reports(self):
```
mshriver: What's the advantage to defining this cached property, when the reports collection can just be directly accessed via `self.collections` by whatever is calling this property?

cfme/intelligence/optimization.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+@attr.s
+class OptimizationSavedReport(BaseEntity):
+    name = attr.ib()
+    last_run_at = attr.ib()
+    username = attr.ib(default=None)
+
+    @property
+    def parent_obj(self):
```
mshriver: This is an ambiguous property to me, given this name.   The `cfme.modeling.base.parent_of_type` method is for doing this kind of lookup in a more declarative way, where you would specifically look for an instance in the entity's ancestor tree that is of type `OptimizationReport`.

It actually seems to me that if this is needed, it could actually be a class attribute passed on instantiation when an OptimizationReport is queued. Either would work, functionally.

cfme/intelligence/optimization.py
```diff
@@ -0,0 +1,157 @@
...
+@navigator.register(OptimizationReportsCollection, "All")
+class OptimizationAll(CFMENavigateStep):
+    VIEW = OptimizationView
+    prerequisite = NavigateToAttribute("appliance.server", "LoggedIn")
+
+    def step(self, *args, **kwargs):
+        self.prerequisite_view.navigation.select("Overview", "Optimization")
+
+
+@navigator.register(OptimizationReport, "SavedAll")
```
mshriver: Is there any 'All' destination for OptimizationSavedReportsCollection, or can that collection only be viewed through the parent OptimizationReport?

cfme/tests/intelligence/test_optimization.py
```diff
@@ -0,0 +1,69 @@
+import pytest
+from widgetastic.utils import attributize_string
+
+from cfme import test_requirements
+from cfme.utils.blockers import BZ
+
+pytestmark = [test_requirements.report, pytest.mark.tier(1), pytest.mark.ignore_stream("5.10")]
+
+REPORTS = [
```
mshriver: Optional for this PR, but consider defining this list with the report entity classes, if they're consistent report types that may (will) be used in multiple tests. If they're just specific parametrization report types for this test module, then ignore me :)

cfme/tests/intelligence/test_optimization.py
```diff
@@ -0,0 +1,69 @@
...
+    "Top Memory Consumers (weekly)",
+    "VMs with Volume Free Space >= 75%",
+]
+
+
+@pytest.mark.meta(automates=[1769346])
+@pytest.mark.uncollectif(
+    lambda menu_name: menu_name in ["Top CPU Consumers (weekly)", "Top Memory Consumers (weekly)"]
+    and BZ(1769346, forced_streams=["5.11"]).blocks,
+    reason="Breadcrumb for these reports is incorrect which is why the exists method fails.",
```
mshriver: :tada:

mshriver: This _might_ be better defined as a blocker meta, with an unblock lambda. Then the BZ.blocks is implicit.

mshriver: Defining as a blocker meta would provide more accurate test result metrics as well - we would see that there are two broken test cases in the test result (skipped via BZ) versus with this, we are just missing two test cases that _should_ be covered.

cfme/tests/intelligence/test_optimization.py
```diff
@@ -0,0 +1,69 @@
...
+)
+@pytest.mark.parametrize("menu_name", REPORTS, ids=[attributize_string(i) for i in REPORTS])
```
mshriver: :+1: Good idea, makes these parameter ID's much more consistent.

cfme/intelligence/optimization.py
valaparthvi: Good thought @mshriver! The table on `OptimizationSavedReport` All page had `Last Run at`, so I just picked that name for it, but `run_at` seems much better, less confusion. And we cannot access the report since most of the saved reports have the same name, we need to use `run_at` property, it's unique for every report.
Changing `last_run_at` attribute to `run_at` for `OptimizationSavedReport`.

cfme/intelligence/optimization.py
valaparthvi: I think `parent_of_type` wouldn't be the right thing to use here. I am not sure what exactly this function does, but from what I understand, it takes an object and a possible parent class, and if the object or some ancestor of the object is an instance of the possible parent class, it returns that matched object.
But instead of all that, I think simply using `parent.parent` would do the trick for me.

cfme/intelligence/optimization.py
valaparthvi: No, there's no `All` destination for it, you will require OptimizationReport to access the All page.

cfme/tests/intelligence/test_optimization.py
valaparthvi: Ignoring you. :grimacing:

cfme/tests/intelligence/test_optimization.py
valaparthvi: Using `pytest.skip` instead.

valaparthvi:
These tests are not applicable to 5.10, so we can ignore the failed 510 run.

cfme/intelligence/optimization.py
```diff
@@ -0,0 +1,154 @@
...
+
+    _collections = {
+        "optimization_saved_reports": OptimizationSavedReportsCollection,
+        "saved_reports": SavedReportsCollection,
+        "reports": ReportsCollection,
+    }
+
+    @property
+    def last_run_at(self):
+        view = navigate_to(self.parent, OptimizationView)
```
mshriver: Why is the view class being passed as the destination here? Is there some black magic in navmazing I missed out on?? :)

I think you want `navigate_to(self.parent, 'All')` here, which should result in an instance of `OptimizationView`.

mshriver:
@valaparthvi only a small comment about the navigation call.  Please work with @digitronik to get this merged tomorrow since I'll be on PTO.  Otherwise its looking awesome!

PR #9551, opened 2019-10-28
[RFR] Update automate button group create, add button wait

mshriver:
Test failures with the screenshot showing a filled form and an active 'Add' button, but the button was never clicked.

Wrapping the 'Add' button click in a wait.

{{ pytest: --use-provider complete --long-running cfme/tests/automate/custom_button/test_buttons.py::test_custom_button_order_sort cfme/tests/automate/custom_button/test_infra_objects.py::test_custom_button_dialog_infra_obj cfme/tests/automate/custom_button/test_generic_objects.py::test_custom_button_automate_evm_obj}}

### DIFF
--- a/cfme/automate/buttons.py
+++ b/cfme/automate/buttons.py
@@ -19,12 +19,16 @@ from widgetastic_patternfly import Input
 
 from cfme.automate import AutomateCustomizationView
 from cfme.base.ui import AutomateSimulationView
+from cfme.exceptions import CFMEException
 from cfme.modeling.base import BaseCollection
 from cfme.modeling.base import BaseEntity
 from cfme.utils.appliance.implementations.ui import CFMENavigateStep
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.appliance.implementations.ui import navigator
+from cfme.utils.log import logger
 from cfme.utils.update import Updateable
+from cfme.utils.wait import TimedOutError
+from cfme.utils.wait import wait_for
 from widgetastic_manageiq import AutomateRadioGroup
 from widgetastic_manageiq import FonticonPicker
 from widgetastic_manageiq import MultiBoxOrderedSelect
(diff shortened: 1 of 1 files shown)


cfme/automate/buttons.py
```diff
@@ -860,11 +873,23 @@ def create(
...
+                     handle_exception=True)
+        except TimedOutError:
+            logger.exception('Timed out waiting for add button on button group form')
+            raise CFMEException('Custom button group add form button did not activate')
         view.add_button.click()
         view = self.create_view(ButtonGroupObjectTypeView)
 
         view.flash.assert_no_error()
-        view.flash.assert_message('Button Group "{}" was added'.format(hover))
+        view.flash.assert_success_message('Button Group "{}" was added'.format(hover))
```
jawatts: This should be in the crud test but out of scope for this PR

mshriver: Agreed, I'd like @digitronik to address this, as well as the 5.10.12 failures in PRT for `ButtonCollection.create()`

digitronik:
@mshriver thanks for try. The failures are not going to solve with `wait_for`; Actually, on 5.10 group not allow to add second button. `group` scope is module when try to create button under same group second time its fails. This is bug we already raised ([BZ-1755229](https://bugzilla.redhat.com/show_bug.cgi?id=1755229)). In our [run](http://infra-trackerbot2.cfme2.lab.eng.rdu2.redhat.com/ostriz/dini/?csrfmiddlewaretoken=LLnSQpQrbnQCn1GZ91FqYIWT7MQzodnCrDlA8CFX6jVii79PeBOjlddy8kIYqvpg&report_type=artifactor&test_filter=&build=5.10.12.1-20191022171907_f883bde&submit=Submit#) we are getting good results... you know reason multiple runs :D.

I'm avoiding to move `button_group` to `fuctional` scope as you know we need to change lots of things. but i will try if bug not going to address in next release.

Yup about actual flash should not assert in model page. I'm going to remove it. It was from start and upto now we won't got single failure due to flash.

PR #9553, opened 2019-10-28
[WIPTEST] DRAFT skip/fail on meta blocker resolving that triggers exception

mshriver:
I'd like to handle exceptions during blocker resolution in a better way, not just letting it result in an ERROR state for the marked test item.

Draft state right now, the pytest.skip call in the pytest hook didn't skip the item in local testing...

Also adding a `repr` for the blocker classes so the exception messages are more readable and informative.


dajoRH:
*CFME QE Bot*

PR #9565, opened 2019-10-31
[1LP][RFR] automated BZ 1726590

lina-nikiforova:
## Purpose or Intent
Automated BZ https://bugzilla.redhat.com/show_bug.cgi?id=1726590 for provisioning with the selected "Linked Clone" option from the template with preallocated disks. 
~Requires wrapanapi PR https://github.com/ManageIQ/wrapanapi/pull/424~

Requires yaml update.

### PRT Run
{{ pytest: -v cfme/tests/infrastructure/test_provisioning_dialog.py::test_linked_clone_default }}

### Notes
Actually, PRT run is useless because the BZ is still on dev. I tested locally and the test currently fails as expected.

### DIFF
--- a/cfme/tests/infrastructure/test_provisioning_dialog.py
+++ b/cfme/tests/infrastructure/test_provisioning_dialog.py
@@ -465,3 +465,50 @@ def test_vmware_default_placement(provisioner, prov_data, provider, setup_provid
     )
     assert 'Datacenter' == provider.mgmt.get_vm(vm_name).raw.parent.parent.name, (
         'The new vm is not placed in the Datacenter root directory!')
+
+
+@pytest.mark.rhv2
+@pytest.mark.provider([RHEVMProvider], override=True,
+                      required_fields=[['cap_and_util', 'capandu_vm']])
+@pytest.mark.meta(automates=[1726590], blockers=[BZ(1726590, forced_streams=["5.10"])])
+def test_linked_clone_default(provisioner, provisioning, provider, prov_data, vm_name):
+    """ Tests provision VM from a template with the selected "Linked Clone" option.
+    The template must have preallocated disks (at least one) for this test.
+
+    Required_fields is set to [['cap_and_util', 'capandu_vm']] because template for this VM has
+    a preallocated disk for sure.
+
+    Bugzilla:
+        1726590
+
+    Metadata:
+        test_flag: provision
+
+    Polarion:
+        assignee: anikifor
+        caseimportance: high
+        casecomponent: Provisioning
+        initialEstimate: 1/4h
+    """
+    prov_data['catalog']['vm_name'] = vm_name
+    prov_data['catalog']['linked_clone'] = True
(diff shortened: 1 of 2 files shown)


cfme/tests/infrastructure/test_provisioning_dialog.py
```diff
@@ -465,3 +465,50 @@ def test_vmware_default_placement(provisioner, prov_data, provider, setup_provid
...
+
+@pytest.mark.rhv2
+@pytest.mark.provider([RHEVMProvider], override=True,
+                      required_fields=[['cap_and_util', 'capandu_vm']])
+@pytest.mark.meta(automates=[1726590], blockers=[BZ(1726590, forced_streams=["5.10"])])
+def test_linked_clone_default(provisioner, provisioning, provider, prov_data, vm_name):
+    """ Tests provision VM from a template with the selected "Linked Clone" option.
+    The template must have preallocated disks (at least one) for this test.
+
+    Required_fields is set to [['cap_and_util', 'capandu_vm']] because template for this VM has
```
kedark3: Isn't that a big CU-24x7 VM? Maybe we can discuss offline and get a small_template that has preallocated disk. 128MB of preallocated is better than bigger ones.

cfme/tests/infrastructure/test_provisioning_dialog.py
```diff
@@ -465,3 +465,50 @@ def test_vmware_default_placement(provisioner, prov_data, provider, setup_provid
...
+        casecomponent: Provisioning
+        initialEstimate: 1/4h
+    """
+    prov_data['catalog']['vm_name'] = vm_name
+    prov_data['catalog']['linked_clone'] = True
+    # should be automatic but due to limited storage on rhv sometimes it may fail because of that
+    prov_data['environment'] = {'automatic_placement': True}
+    rhv = provider.mgmt
+
+    def _get_template_with_preallocated_disk():
```
kedark3: I like this approach you are using here, but I was wondering, can we just do this:
1 . Add a preallocated small template on each of our provider
2 . update yamls with a field sparse: True/False under the templates 
3. Read yamls to figure out the sparse template is present. 

This is sort of approach we also use for console_template. And its something I would prefer in your case too.

john-dupuy: +1 I think this is a good idea.

lina-nikiforova: yes, we can. 
I thought about it in the beginning but wanted to have a more versatile and reliable approach. I like it and reading actual data seems better to me, but we can go the easy way and just create a small template and put its name in the yamls. It doesn't need to be on each provider, this is a RHV specific case, as I understand, so it can be only on the latest RHV.

Kedar, I have contacted you about this template.

kedark3: This approach is something we can use globally to make things more dynamic in the future.

PR #9567, opened 2019-11-01
[1LP][RFR] Add stringification function to BaseProvider class

prichard77:
## Purpose or Intent

- __Enhancement__ for BaseProvider class

### DIFF
--- a/cfme/common/provider.py
+++ b/cfme/common/provider.py
@@ -119,6 +119,9 @@ class BaseProvider(Taggable, Updateable, Navigatable, BaseEntity, CustomButtonEv
     def __eq__(self, other):
         return type(self) is type(other) and self.key == other.key
 
+    def __str__(self):
+        return self.name
+
     @property
     def data(self):
         """ Returns yaml data for this provider.


prichard77:
This is a pretty simple update. However Mike, understandably, wants to do some testing and investigating to ensure this will not affect existing code. I'm setting this to RFR but am not expecting it to get merged until Mike gets back from PTO. It's not urgent.

john-dupuy:
I found one usage of `str(provider)` in the trackerbot code, this PR may have some adverse affects there: https://github.com/ManageIQ/integration_tests/blob/046c0380fd418015d41b9193e0ee8b41f4bd2240/cfme/utils/trackerbot.py#L217

@izapolsk can you take a look?

john-dupuy:
Ah nevermind, that is a different provider subclass, shouldn't break anything then. https://github.com/ManageIQ/integration_tests/blob/046c0380fd418015d41b9193e0ee8b41f4bd2240/cfme/utils/trackerbot.py#L342

PR #9568, opened 2019-11-01
[1LP][RFR][NOTEST] Add BZ links to the test cases

jaryn:
Add BZ links to covering cases

### DIFF
--- a/cfme/tests/configure/test_register_appliance.py
+++ b/cfme/tests/configure/test_register_appliance.py
@@ -199,12 +199,10 @@ def test_rh_registration(appliance, request, reg_method, reg_data, proxy_url, pr
 
 
 @pytest.mark.rhel_testing
+@pytest.mark.meta(coverage=[1694073])
 def test_rhsm_registration_check_repo_names(
         temp_appliance_preconfig_funcscope, soft_assert, appliance):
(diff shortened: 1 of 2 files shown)


cfme/tests/configure/test_register_appliance.py
```diff
@@ -196,6 +196,8 @@ def test_rh_registration(appliance, request, reg_method, reg_data, proxy_url, pr
 def test_rhsm_registration_check_repo_names(
         temp_appliance_preconfig_funcscope, soft_assert, appliance):
     """ Checks default rpm repos on a fresh appliance
+    Bugzilla:
+        1694073
```
mshriver: Please put them in coverage/automates meta markers if that applies. This is a nice way to list BZ's that are somehow related to the bug, maybe the original RFE, but if there is specific coverage for these, marking them is more useful for tooling around automatically determining our coverage.

jaryn: done

john-dupuy: No need to remove them in the future, but they are not necessary like the markers are.

john-dupuy:
Not really blockers-only, but it's bz related.

PR #9571, opened 2019-11-04
[RFR] Update test case meta

mshriver:
These tests aren't getting the right states/meta in polarion because of duplicate names, bad parametrization, or re-titling.

Update polarion_tools.yaml to prevent setting title via flake plugin.

### DIFF
--- a/cfme/tests/infrastructure/test_snapshot.py
+++ b/cfme/tests/infrastructure/test_snapshot.py
@@ -573,8 +573,6 @@ def test_sui_create_snapshot():
         caseimportance: medium
         initialEstimate: 1/4h
         startsin: 5.9
-        title: SUI: Create snapshot from vm details page, snapshot page
-               and service details page
         setup:
             1. Add an infra provider
             2. Create and order a service catalog which provisions a VM
@@ -664,6 +662,7 @@ def test_snapshot_timeline_new_vm():
 @test_requirements.snapshot
 @test_requirements.ssui
 @pytest.mark.tier(2)
+@pytest.mark.meta(coverage=[1600043])
 def test_ssui_snapshot_memory_checkbox():
     """
     Test "snapshot vm memory" checkbox when creating snapshot for powered off vm.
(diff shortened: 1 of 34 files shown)


cfme/tests/infrastructure/test_snapshot.py
```diff
@@ -573,7 +573,6 @@ def test_sui_create_snapshot():
         caseimportance: medium
         initialEstimate: 1/4h
         startsin: 5.9
-        title: SUI: Create snapshot from vm details page, snapshot page
                and service details page
```
prichard77: Need to delete this line also.

cfme/tests/test_appliance.py
```diff
@@ -24,7 +24,6 @@
     'rhn-check',
     'rhnlib',
 ])
-@test_requirements.appliance
```
prichard77: I am curious, why are we deleting all of the "@test_requirements.appliance" lines for tests in this file. After the deletion, most of the tests don't have a @test_requirements marker and I see the casecomponent in the Polarion docblock is set to Appliance?

prichard77: I guess above is because there is no "appliance" section in test_requirements.py?

cfme/tests/infrastructure/test_snapshot.py
jawatts: nice catch

mshriver: indeed nice catch! I thought flake had caught me on all of these, but I guess this one didn't trigger?

cfme/tests/test_appliance.py
mshriver: because the marker is included at the module level, L13 above.

jawatts: @prichard77 @mshriver moved it to the module pytestmark

cfme/tests/test_manual.py
```diff
@@ -123,8 +123,7 @@ def test_sui_stack_service_vm_detail_page_should_show_correct_data():
         casecomponent: SelfServiceUI
         caseimportance: medium
         initialEstimate: 1/4h
-        title: SUI : Stack Service VM detail page should show correct data
-    """
+            """
```
jawatts: quotes are off here

PR #9576, opened 2019-11-05
[1LP][RFR] New Test: Checking dialog export flash message on while editing dialog

ganeshhubale:
## Purpose or Intent
- Checking flash message on wrong page
### PRT Run
{{ pytest: cfme/tests/automate/test_service_automate.py -k 'test_import_dialog_file_without_selecting_file' --use-template-cache -qsvvv}}

### DIFF
--- a/cfme/automate/dialog_import_export.py
+++ b/cfme/automate/dialog_import_export.py
@@ -19,6 +19,7 @@ class DialogImportExportView(BaseLoggedInPage):
     title = Text("#explorer_title_text")
     upload_file = FileInput(id="upload_file")
     upload = Button(id="upload_service_dialog_import")
+    export = Text('//button[text()="Export" and @name="button"]')
 
     def in_import_export(self):
         return (
--- a/cfme/tests/automate/test_service_automate.py
+++ b/cfme/tests/automate/test_service_automate.py
@@ -381,7 +381,7 @@ def test_service_retire_automate():
 
 
 @pytest.mark.tier(2)
-@pytest.mark.meta(automates=[1740796])
+@pytest.mark.meta(automates=[1740796], blockers=[BZ(1740796, forced_streams=["5.10"])])
 def test_import_dialog_file_without_selecting_file(appliance, dialog):
     """
     Bugzilla:
@@ -404,5 +404,4 @@ def test_import_dialog_file_without_selecting_file(appliance, dialog):
     view = navigate_to(import_export, "DialogImportExport")
     view.export.click()
     view.flash.assert_message("At least 1 item must be selected for export")
-    if not BZ(1740796, forced_streams=['5.10']).blocks:
-        dialog.update({'label': fauxfactory.gen_alphanumeric()})
+    dialog.update({'label': fauxfactory.gen_alphanumeric()})
--- a/cfme/tests/automate/test_service_automate.py
+++ b/cfme/tests/automate/test_service_automate.py
@@ -404,6 +404,5 @@ def test_import_dialog_file_without_selecting_file(appliance, dialog):
     view = navigate_to(import_export, "DialogImportExport")
     view.export.click()
     view.flash.assert_message("At least 1 item must be selected for export")
-    if BZ(1740796).blocks:
-        with pytest.raises(AssertionError, match="At least 1 item must be selected for export"):
-            dialog.update({'label': fauxfactory.gen_alphanumeric()})
+    if not BZ(1740796, forced_streams=['5.10']).blocks:
+        dialog.update({'label': fauxfactory.gen_alphanumeric()})
--- a/cfme/tests/automate/test_service_automate.py
+++ b/cfme/tests/automate/test_service_automate.py
@@ -5,10 +5,13 @@ import fauxfactory
 import pytest
 
 from cfme import test_requirements
+from cfme.automate.dialog_import_export import DialogImportExport
 from cfme.base.credential import Credential
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.provisioning import do_vm_provisioning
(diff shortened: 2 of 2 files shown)


cfme/tests/automate/test_service_automate.py
```diff
@@ -397,4 +400,10 @@ def test_import_dialog_file_without_selecting_file():
             1. Flash message: "At least 1 item must be selected for export"
             2. Error flash message should not appear
     """
-    pass
+    import_export = DialogImportExport(appliance)
+    view = navigate_to(import_export, "DialogImportExport")
+    view.export.click()
+    view.flash.assert_message("At least 1 item must be selected for export")
+    if BZ(1740796).blocks:
+        with pytest.raises(AssertionError, match="At least 1 item must be selected for export"):
```
sshveta: This will only check the flash message till the BZ is fixed right ? 
If you are automating this BZ it should be tested even after the BZ is verified so you can block the test until the BZ is fixed and once the BZ is fixed it should check that the flash message is not shown anymore .
What do you think ?

john-dupuy: Yeah I would say that this test shouldn't run until the BZ is fixed, because this is the main assertion of the test. The bz should just be in the `blockers` metadata

ganeshhubale: @john-dupuy 
- According to me, we don't need to skip this entire test case. Although clicking on export button without selecting dialog gives expected error flash message.
@sshveta 
- I agree with you. I changed test case like below:
```
 if not BZ(1740796).blocks:
        dialog.update({'label': fauxfactory.gen_alphanumeric()})
```
Please le me know, if we need any other changes.

john-dupuy: @ganeshhubale While this solution is good once the BZ is fixed, right now the test case is passing. It should be failing because the BZ is still present. In my view, a test case for a specific BZ shouldn't pass until the BZ is fixed. This is why I think it is better to block the whole test case.

ganeshhubale: @john-dupuy 
- I agree that this BZ is not yet fixed. But this BZ is only related to this issue - ```Error flash message after updating dialog```
- So I am skipping execution of this part until this BZ got fixed. Once it fixed, it will check if there is error flash message after updating dialog.
- Please let me know your view.

cfme/automate/dialog_import_export.py
```diff
@@ -19,6 +19,7 @@ class DialogImportExportView(BaseLoggedInPage):
     title = Text("#explorer_title_text")
     upload_file = FileInput(id="upload_file")
     upload = Button(id="upload_service_dialog_import")
+    export = Text('//button[text()="Export" and @name="button"]')
```
john-dupuy: Should this be a button widget?

cfme/tests/automate/test_service_automate.py
john-dupuy: Chatted offline, eventually settled on blocking the test case with `blockers`

cfme/automate/dialog_import_export.py
ganeshhubale: I tried with button widget and different xpath but not able to access. So I used Text widget.

cfme/tests/automate/test_service_automate.py
```diff
@@ -378,8 +381,8 @@ def test_service_retire_automate():
 
 
 @pytest.mark.tier(2)
-@pytest.mark.meta(coverage=[1740796])
-def test_import_dialog_file_without_selecting_file():
+@pytest.mark.meta(coverage=[1740796], blockers=[BZ(1740796, forced_streams=["5.10"])])
```
sshveta: This test case is not automating the BZ yet , hence no need to block . 
If you are automating the BZ then you have to check that the flash message is not shown after selecting a dialog and editing it  and then the blocker is needed. 

Also , you need to add `automated` instead of `coverage`

cfme/tests/automate/test_service_automate.py
```diff
@@ -397,4 +400,8 @@ def test_import_dialog_file_without_selecting_file():
             1. Flash message: "At least 1 item must be selected for export"
             2. Error flash message should not appear
     """
-    pass
+    import_export = DialogImportExport(appliance)
+    view = navigate_to(import_export, "DialogImportExport")
+    view.export.click()
+    view.flash.assert_message("At least 1 item must be selected for export")
+    dialog.update({'label': fauxfactory.gen_alphanumeric()})
```
john-dupuy: Ah thanks @sshveta, yeah @ganeshhubale you need to check the flash message after the update too

cfme/tests/automate/test_service_automate.py
ganeshhubale: This test case is automating this BZ by checking service dialog update.
I am changing coverage to automates

cfme/tests/automate/test_service_automate.py
ganeshhubale: We don't need to check flash message after dialog update. As we already have 
```view.flash.assert_no_error()``` in **update** method. On update, it gives ```AssertionError: assert_no_error: error: At least 1 item must be selected for export```

john-dupuy: Ah, thanks for the clarification!

izapolsk:
@ganeshhubale, Looks good to me but this test is skipped for both streams. is it correct ?

ganeshhubale:
Yes. As per discussion with @john-dupuy @sshveta . we are skipping this test case for both streams. But it automates the BZ scenario.

izapolsk:
@ganeshhubale, since I cannot recognize that based on PRT results, I can just hope so :)

PR #9577, opened 2019-11-05
[RFR] [NOTEST] Parametrize the nor/rightsize tests, eliminate duplication

mshriver:
These tests are duplicates, and should be provider parametrized.


mshriver:
Discussed this change with the test owner, @tpapaioa, and he's going to follow this up with some metadata updates for test steps, and WONTFIX BZ limitations.

PR #9580, opened 2019-11-05
[1LP][WIPTEST] Allow dummy appliances to be used with the parallelizer

john-dupuy:
This helps to see what slaves get which tests during a parallelized run. To use it you can do 

```pytest --collect-only --dummy-appliance --num-dummies N cfme/tests```

{{ pytest: --collect-only --dummy-appliance --num-dummies 4 cfme/tests/ }}

### DIFF
--- a/cfme/utils/appliance/__init__.py
+++ b/cfme/utils/appliance/__init__.py
@@ -3179,6 +3179,8 @@ class DummyAppliance(object):
     build = 'dummyappliance'
     managed_known_providers = []
     collections = attr.ib(default=attr.Factory(collections_for_appliance, takes_self=True))
+    url = 'http://dummies.r.us'
+    is_dummy = attr.ib(default=True)
 
     @property
     def is_downstream(self):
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: cfme/utils/appliance/__init__.py [hunk 1]
```diff
@@ -3179,6 +3179,8 @@ class DummyAppliance(object):
     build = 'dummyappliance'
     managed_known_providers = []
     collections = attr.ib(default=attr.Factory(collections_for_appliance, takes_self=True))
+    url = 'http://dummies.r.us'
```
jawatts: :laughing:

mshriver: Now I want to buy this domain

izapolsk:
@john-dupuy, nice Pr but I see that all tests have been sent to one slave. I'd like to double check this. So, I slightly updated your PRT command and restarted PRT

izapolsk:
I still see that only one slave is used in collectonly run.

PR #9592, opened 2019-11-07
[RFR] Add BZ Blocker test_appliance_console_scap

jawatts:
{{ pytest: cfme/tests/cli/test_appliance_console.py::test_appliance_console_scap }}


jawatts:
At the time of running PRT, bug was not triaged. It is now blocking:
```
In[2]: from cfme.utils.blockers import BZ
In[3]: BZ(1769901).blocks
Out[3]: True
In[4]: appliance.version
Out[4]: Version('5.11.0.28')

```

PR #9598, opened 2019-11-11
[RFR] fix sprout client

izapolsk:
bz scripting stuff needs lazy evaluation. otherwise, it makes miq sprout check current appliance

### DIFF
--- a/cfme/scripting/bz.py
+++ b/cfme/scripting/bz.py
@@ -122,7 +122,7 @@ def report(directory):
     "--version",
     "-v",
     "version",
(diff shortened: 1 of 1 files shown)


cfme/scripting/bz.py
```diff
@@ -122,7 +122,7 @@ def report(directory):
     "--version",
     "-v",
     "version",
-    default=current_version().vstring,
+    default=lambda: current_version().vstring,
```
mshriver: :man_facepalming:

PR #9602, opened 2019-11-11
[RFR][1LP] Add ensure_checked() and ensure_unchecked() functions to JSBaseEntity.

prichard77:
## Purpose or Intent.

- __Enhancement__ for selecting entities in the JSBaseEntity class. Using new functions ensure_checked() or ensure_unchecked() will make sure entities are unchecked before selecting them, and checked before un-selecting them.

### PRT Run
{{ pytest: --use-provider complete --long-running cfme/tests/infrastructure/test_host.py::test_infrastructure_hosts_crud
cfme/tests/infrastructure/test_host.py::test_infrastructure_hosts_refresh_multi
cfme/tests/infrastructure/test_host.py::test_compare_hosts_from_provider_allhosts }}

### DIFF
--- a/widgetastic_manageiq/__init__.py
+++ b/widgetastic_manageiq/__init__.py
@@ -3290,6 +3290,18 @@ class JSBaseEntity(View, ReportDataControllerMixin):
         else:
             return getattr(self, "_name", None)
 
+    def ensure_checked(self):
+        if not self.is_checked:
+            self.check()
+        if not self.is_checked:
+            raise WidgetOperationFailed(f'Checking of Entity {self} unsuccessful.')
+
+    def ensure_unchecked(self):
+        if self.is_checked:
+            self.uncheck()
+        if self.is_checked:
+            raise WidgetOperationFailed(f'Unchecking of Entity {self} unsuccessful.')
+
     def check(self):
         self._call_item_method("select")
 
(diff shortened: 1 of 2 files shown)


prichard77:
The failures in PRT are not the tests that are affected by this update. I am still not sure what is wrong with my pytest command that is causing all tests in the modiule to get run.

widgetastic_manageiq/__init__.py
```diff
@@ -3290,6 +3290,18 @@ def name(self):
         else:
             return getattr(self, "_name", None)
 
+    def ensure_checked(self):
+        if not self.is_checked:
```
valaparthvi: Can you use `wait_for` here?

widgetastic_manageiq/__init__.py
```diff
@@ -3290,6 +3290,18 @@ def name(self):
...
+            self.check()
+        if not self.is_checked:
+            raise WidgetOperationFailed(f"Checking of Entity {self} unsuccessful.")
+
+    def ensure_unchecked(self):
+        if self.is_checked:
```
valaparthvi: Same comment as above.

widgetastic_manageiq/__init__.py
prichard77: I don't believe wait_for is needed here. I have tested this fairly well and the check functionality is essentially instantaneous.

PR #9604, opened 2019-11-11
[1LP][RFR] Manual test cleanup and qe-test-coverage test cases

valaparthvi:
## Purpose or Intent

- __Adding tests__ 
    1. test_widget_generate_content_via_rest - [BZ 1761836](https://bugzilla.redhat.com/show_bug.cgi?id=1761836)
    2. test_service_refresh_dialog_fields_default_values - [BZ 1730813](https://bugzilla.redhat.com/show_bug.cgi?id=1730813)
    3. test_schedule_automation_request - [BZ 1740340](https://bugzilla.redhat.com/show_bug.cgi?id=1740340)
    4. test_update_roles_via_rest_name_change - [BZ 1596142](https://bugzilla.redhat.com/show_bug.cgi?id=1596142)
    5. test_compare_vm_from_datastore_relationships -  [BZ 1733120](https://bugzilla.redhat.com/show_bug.cgi?id=1733120)

- __Removing tests__
    1. test_automation_request_task [already automated]
    2. test_date_should_be_change_in_editing_reports_scheduled [already automated]
    3. test_custom_reports_with_timelines

### DIFF
--- a/cfme/tests/test_rest_manual.py
+++ b/cfme/tests/test_rest_manual.py
@@ -145,14 +145,14 @@ def test_service_refresh_dialog_fields_default_values():
 
 @pytest.mark.manual
 @pytest.mark.tier(2)
-@pytest.mark.meta(coverage=[1486765, 1730813])
+@pytest.mark.meta(coverage=[1486765, 1740340])
 @pytest.mark.parametrize(
     "scheduler", ["2", "2019-08-14 17:41:06 UTC"], ids=["number_of_days", "exact_time"]
 )
 def test_schedule_automation_request(scheduler):
     """
     Bugzilla:
-        1730813
+        1740340
         1486765
 
     Polarion:
--- a/cfme/tests/test_rest_manual.py
+++ b/cfme/tests/test_rest_manual.py
@@ -3,6 +3,8 @@
 import pytest
 
 from cfme import test_requirements
+from cfme.utils.appliance.implementations.rest import ViaREST
+from cfme.utils.appliance.implementations.ui import ViaUI
 
 
 @pytest.mark.manual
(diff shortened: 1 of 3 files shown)


cfme/tests/test_rest_manual.py
```diff
@@ -94,3 +66,148 @@ def test_notification_url_parallel_requests():
         1700378
     """
     pass
+
+
+@pytest.mark.manual
+@pytest.mark.meta(coverage=[1761836])
+@pytest.mark.tier(3)
+@pytest.mark.parametrize("implementation", ["UI", "Rest"])
```
izapolsk: There are already defined appropriate classes. and please use `context` instead for better readability
`@pytest.mark.parametrize('context', [ViaSSUI, ViaUI])`

cfme/tests/test_rest_manual.py
```diff
@@ -94,3 +66,148 @@ def test_notification_url_parallel_requests():
...
+def test_widget_generate_content_via_rest(implementation):
```
izapolsk: I'd not add `via_rest` to tests which will be run with several contexts. in addition such contexts will be reflected in test names.

valaparthvi: My test objective is to test `generate_content` action via REST hence `via_rest` in the name but I also need to check if the action yielded results for both UI and REST, hence the parametrization.

cfme/tests/test_rest_manual.py
mshriver: @valaparthvi I think there was a misunderstanding here - there are actual ViaUI / ViaRest classes that should be parametrized against here, and those context classes can be used directly.  Your update still just parametrizes the strings.

Example:
https://github.com/ManageIQ/integration_tests/blob/d752d1822102c5a8c246babf1b07b95e5d0d1ec5/cfme/tests/cloud/test_instance_reconfigure.py#L18

valaparthvi: I understood that, but um okay here's my question: If I use the classes instead of strings, will I also need to use those contexts? Because I am only using this parametrization so that I know which way to verify my response. Should I stil use the classes?

izapolsk: @valaparthvi, you aren't required to use these classes in test if it isn't necessary. It will only be used for generating tests.

PR #9616, opened 2019-11-13
[1LP][RFR] New Test: Test tag delete using automate method

ganeshhubale:
## Purpose or Intent
-  Testing ```tag_delete``` method introduced for deletion of tags under category
- This test case also goes through methods like ```test_create```, ```tag_exists``` etc
- **Note**: Test cases ```test_archived_instance_status``` and ```test_orphaned_instance_status``` are already automated in test case ```test_power_options_on_archived_instance_all_page```

### PRT Run
{{ pytest: cfme/tests/automate/test_method.py::test_delete_tag_from_category --use-template-cache -qsvvv }}

### DIFF
--- a/cfme/tests/automate/__init__.py
+++ b/cfme/tests/automate/__init__.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 from textwrap import dedent
 
+import fauxfactory
+
 user_list_hash_data = dedent(
     """
 module Demo
@@ -64,4 +66,24 @@ if __FILE__ == $PROGRAM_NAME
   Demo::Automate::System::Request::Test::AvailableUsers.new.main
 end
         """
+)
+
+tag_values = [fauxfactory.gen_alphanumeric(start="tag_", length=12).lower(), "{tag_exist}"]
+tag_delete_from_category = (
+    f"""
+    # Create tag under category - 'Location'
+    $evm.execute(
+'tag_create', 'location', :name => '{tag_values[0]}', :description => 'Testing tag {tag_values[0]}'
+)
+    # Check if tag exists
+    tag_exist = $evm.execute('tag_exists?', 'location', '{tag_values[0]}')
+    $evm.log(:info, "Tag exists: #{tag_values[1]}")
+
+    # Delete the tag
+    $evm.execute('tag_delete', 'location', '{tag_values[0]}')
+
+    # Check if deleted tag exists
+    tag_exist = $evm.execute('tag_exists?', 'location', '{tag_values[0]}')
+    $evm.log(:info, "Tag exists: #{tag_values[1]}")
+    """
 )
--- a/cfme/tests/automate/test_method.py
+++ b/cfme/tests/automate/test_method.py
@@ -4,6 +4,7 @@ from textwrap import dedent
 import fauxfactory
 import pytest
 
+from . import tag_delete_from_category
 from cfme import test_requirements
 from cfme.automate.explorer.klass import ClassDetailsView
 from cfme.automate.simulation import simulate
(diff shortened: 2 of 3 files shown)


cfme/tests/automate/__init__.py
```diff
@@ -65,3 +67,23 @@ def fill_dialog_field(list)
...
+    $evm.execute(
+'tag_create', 'location', :name => '{tag_values[0]}', :description => 'Testing tag {tag_values[0]}'
+)
+    # Check if tag exists
+    tag_exist = $evm.execute('tag_exists?', 'location', '{tag_values[0]}')
+    $evm.log(:info, "Tag exists: #{tag_values[1]}")
+
+    # Delete the tag
+    $evm.execute('tag_delete', 'location', '{tag_values[0]}')
+
```
sbulage: **Question:**
@ganeshhubale do you think logging about the tag successfully removed or not would more helpful .?

cfme/tests/automate/__init__.py
```diff
@@ -65,3 +67,23 @@ def fill_dialog_field(list)
...
+    # Check if deleted tag exists
+    tag_exist = $evm.execute('tag_exists?', 'location', '{tag_values[0]}')
+    $evm.log(:info, "Tag exists: #{tag_values[1]}")
```
sbulage: **Question/Suggestion**
@ganeshhubale everywhere you use `{tag_vaules[0]}`, why here `{tag_values[1]}`, is there any specific reason.? If any could please add more information around it.

cfme/tests/automate/__init__.py
```diff
@@ -65,3 +67,23 @@ def fill_dialog_field(list)
...
+'tag_create', 'location', :name => '{tag_values[0]}', :description => 'Testing tag {tag_values[0]}'
+)
+    # Check if tag exists
+    tag_exist = $evm.execute('tag_exists?', 'location', '{tag_values[0]}')
+    $evm.log(:info, "Tag exists: #{tag_values[1]}")
+
+    # Delete the tag
+    $evm.execute('tag_delete', 'location', '{tag_values[0]}')
+
+    # Check if deleted tag exists
```
sbulage: **Suggestion:**
IMO, you can check same immediate after perform delete operation .?

cfme/tests/automate/test_method.py
```diff
@@ -842,3 +843,37 @@ def test_copy_with_embedded_method(request, appliance, klass):
     )
     view = navigate_to(copied_method, "Details")
     assert view.embedded_method_table.read()[0]["Path"] == embedded_method_path
+
+
+@pytest.mark.tier(2)
+@pytest.mark.meta(blockers=[BZ(1744514)], automates=[1744514])
+@pytest.mark.customer_scenario
+def test_delete_tag_from_category(custom_instance):
```
sbulage: **Suggestion:**
Whole test is around CRUD via Automate method, test case name should be something like: `test_automate_method_delete_tag_from_category` or `test_automate_delete_tag_from_category` to have consistency around it.

cfme/tests/automate/__init__.py
ganeshhubale: I am doing that using ```tag_exists```. Hence no need of extra log after removing tag.

cfme/tests/automate/__init__.py
ganeshhubale: No need of more info. It's just list of two values used in ruby code: ```tag_name``` and ```ruby_variable```(So that formatting with fstring will be easy)

sbulage: Okay.

cfme/tests/automate/__init__.py
ganeshhubale: I am using ```tag_exists```. So that this is also gets tested and we will easily get to know if tag is successfully deleted or not.

cfme/tests/automate/test_method.py
ganeshhubale: I don't think we need to change test name. It is already under in automate  related files. hence keeping as it is.

sbulage: Yes, @ganeshhubale I agree with you about the test is present in the same file, but looking all of your tests in the same file have suggested name, implementation is up to you :smile:

mshriver: As long as there isn't duplication of the test case names across test modules, I think the naming here is fine, as some context is given by the module.

PR #9617, opened 2019-11-13
[1LP][RFR] Adding a new test to refresh ssui browser page

niyazRedhat:
{{ pytest: cfme/tests/ssui/test_ssui_service_catalogs.py  -k 'test_refresh_ssui_page' --long-running --use-template-cache -sqvvv }}


izapolsk:
@niyazRedhat, from my point, this test should change something before refresh in order to check that refresh really does something.

PR #9627, opened 2019-11-15
[1LP][RFR] New Test: Testing proper error flash message for playbook type domain imports

ganeshhubale:

## Purpose or Intent
- Testing proper error flash message for playbook type domain imports

### PRT Run
{{ pytest: cfme/tests/ansible/test_embedded_ansible_automate.py::test_import_domain_containing_playbook_method --use-template-cache -qsvv --long-running }}

### DIFF
--- a/cfme/tests/ansible/test_embedded_ansible_automate.py
+++ b/cfme/tests/ansible/test_embedded_ansible_automate.py
@@ -2,6 +2,7 @@ import fauxfactory
 import pytest
 
 from cfme import test_requirements
+from cfme.automate.import_export import FileImportSelectorView
 from cfme.automate.simulation import simulate
 from cfme.control.explorer import alert_profiles
 from cfme.fixtures.automate import DatastoreImport
@@ -12,6 +13,7 @@ from cfme.services.service_catalogs.ui import OrderServiceCatalogView
 from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.conf import cfme_data
 from cfme.utils.conf import credentials
+from cfme.utils.ftp import FTPClientWrapper
 from cfme.utils.log_validator import LogValidator
 from cfme.utils.update import update
 from cfme.utils.wait import TimedOutError
@@ -418,3 +420,56 @@ def test_variable_pass(request, appliance, setup_ansible_repository, import_data
         provision_request = appliance.collections.requests.instantiate(request_description)
         provision_request.wait_for_request(method="ui")
         request.addfinalizer(provision_request.remove_request)
+
+
+@pytest.mark.tier(1)
+@pytest.mark.ignore_stream("5.10")
+@pytest.mark.meta(automates=[1677575])
+@pytest.mark.parametrize(
+    "import_data",
+    [DatastoreImport("bz_1677575.zip", "bz_1677575", None)],
+    ids=["datastore"],
+)
+def test_import_domain_containing_playbook_method(request, appliance, setup_ansible_repository,
+                                                  import_data):
+    """This test case tests support of Export/Import of Domain with Ansible Method
+
+    Bugzilla:
+        1677575
+
+    Polarion:
+        assignee: ghubale
+        initialEstimate: 1/8h
+        caseimportance: high
+        caseposneg: positive
+        testtype: functional
+        startsin: 5.11
+        casecomponent: Automate
+        tags: automate
+        setup:
+            1. Add playbook repository
+            2. Create a new automate method with playbook type.
+            3. Fill the required fields, for instance repository and playbook.
+            4. Export this datastore.
+            5. Playbook method fields are stored as a names instead of IDs. (this is not
+               possible via automate need to check manually in method yaml)
+        testSteps:
+            1. Import the exported datastore and change name of playbook in method yaml to invalid
+               playbook name(Note: These test steps needs to execute manually and then import
+               datastore)
+        expectedResults:
+            1. Proper error should be displayed while importing datastore with invalid playbook
+    """
+    # Download datastore file from FTP server
+    fs = FTPClientWrapper(cfme_data.ftpserver.entities.datastores)
+    file_path = fs.download(import_data.file_name)
+
+    # Import datastore file to appliance
+    datastore = appliance.collections.automate_import_exports.instantiate(
+        import_type="file", file_path=file_path
+    )
+    domain = datastore.import_domain_from(import_data.from_domain, import_data.to_domain)
+    request.addfinalizer(domain.delete_if_exists)
+    view = appliance.browser.create_view(FileImportSelectorView)
+    error_msg = "Playbook 'invalid_1677575.yml' not found in repository 'test_playbooks_automate'"
+    assert error_msg in view.flash.read()[0]
--- a/cfme/tests/ansible/test_embedded_ansible_automate.py
+++ b/cfme/tests/ansible/test_embedded_ansible_automate.py
@@ -329,7 +329,7 @@ def setup_ansible_repository(appliance, wait_for_ansible):
         timeout=60,
         fail_func=view.toolbar.refresh.click
     )
-    yield
+    yield repository
     repository.delete_if_exists()
 
 
@@ -471,5 +471,9 @@ def test_import_domain_containing_playbook_method(request, appliance, setup_ansi
     domain = datastore.import_domain_from(import_data.from_domain, import_data.to_domain)
     request.addfinalizer(domain.delete_if_exists)
     view = appliance.browser.create_view(FileImportSelectorView)
-    error_msg = "Playbook 'invalid_1677575.yml' not found in repository 'test_playbooks_automate'"
-    assert error_msg in view.flash.read()[0]
+    # setup_ansible_repository.name is ansible repository name already there in datastore yml which
+    # we are going to import.
+    error_msg = (
+        f"Playbook 'invalid_1677575.yml' not found in repository '{setup_ansible_repository.name}'"
+    )
+    view.flash.assert_message(text=error_msg, partial=True)
--- a/cfme/automate/import_export.py
+++ b/cfme/automate/import_export.py
@@ -71,6 +71,7 @@ class AutomateImportExportView(AutomateImportExportBaseView):
 
 
 class FileImportSelectorView(AutomateImportExportBaseView):
+    flash = V2VFlashMessages(locator="div.import-flash-message")
     import_into = BootstrapSelect(id="selected_domain_to_import_to")
     import_from = BootstrapSelect(
         locator=".//div[contains(@class, 'bootstrap-select importing-domains')]"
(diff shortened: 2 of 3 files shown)


dajoRH:
Lint report for commit 69aed87e3d5c6ad6eb7de5f5b3c8a85dc3e827e4:

`cfme/tests/ansible/test_embedded_ansible_automate.py`:
- :heavy_exclamation_mark: Line 462:26: **W291** *trailing whitespace*
- :heavy_exclamation_mark: Line 463:94: **W291** *trailing whitespace*
- :heavy_exclamation_mark: Line 464:84: **W291** *trailing whitespace*
- :heavy_exclamation_mark: Line 465:98: **W291** *trailing whitespace*
- :heavy_exclamation_mark: Line 466:98: **W291** *trailing whitespace*

Hmmm


**No commit flaws detected**.

*CFME QE Bot*

cfme/tests/ansible/test_embedded_ansible_automate.py
```diff
@@ -419,3 +421,43 @@ def test_variable_pass(request, appliance, setup_ansible_repository, import_data
...
+                                                  import_data):
+    """This test case tests support of Export/Import of Domain with Ansible Method
+
+    Bugzilla:
+        1677575
+
+    Polarion:
+        assignee: ghubale
+        initialEstimate: 1/8h
+        caseimportance: high
```
sbulage: **Required:**
@ganeshhubale No need to mention these Polarion tags which are by default. --> `caseimportance: high`, `caseposneg: positive` and `testtype: functional`

cfme/tests/ansible/test_embedded_ansible_automate.py
```diff
@@ -419,3 +421,43 @@ def test_variable_pass(request, appliance, setup_ansible_repository, import_data
...
+    "import_data",
+    [DatastoreImport("bz_1677575.zip", "bz_1677575", None)],
+    ids=["datastore"],
+)
+def test_import_export_domain_with_ansible_method(request, appliance, setup_ansible_repository,
+                                                  import_data):
+    """This test case tests support of Export/Import of Domain with Ansible Method
+
+    Bugzilla:
+        1677575
```
jaryn: I think this is redundant now, when we are using the automates marker

cfme/tests/ansible/test_embedded_ansible_automate.py
```diff
@@ -419,3 +421,43 @@ def test_variable_pass(request, appliance, setup_ansible_repository, import_data
...
+    """
+    # Download datastore file from FTP server
+    fs = FTPClientWrapper(cfme_data.ftpserver.entities.datastores)
+    file_path = fs.download(import_data.file_name)
+
+    # Import datastore file to appliance
+    datastore = appliance.collections.automate_import_exports.instantiate(
+        import_type="file", file_path=file_path
+    )
+    domain = datastore.import_domain_from(import_data.from_domain, import_data.to_domain)
```
sbulage: **Question:**
Why you're explicitly writing these steps.?
There is fixture to achieve this steps. --> [https://github.com/ManageIQ/integration_tests/blob/master/cfme/fixtures/automate.py#L81-L110](url)

ganeshhubale: Yeah. But I cant use this fixture. If you see [here](https://github.com/ManageIQ/integration_tests/blob/master/cfme/fixtures/automate.py#L106); I am not importing domain successfully, I am checking for error flash message on that page.

cfme/tests/ansible/test_embedded_ansible_automate.py
ganeshhubale: Added it for more clarity while looking into doc string. Although we have choice to keep it in polarion doc or remove.

cfme/tests/ansible/test_embedded_ansible_automate.py
ganeshhubale: I didn't hear this before. I am adding as it is part of doc string.

cfme/tests/ansible/test_embedded_ansible_automate.py
sbulage: @ganeshhubale Afaik, this isn't doc string, these are parameters for Polarion to set. We kept few parameters by default to make clarity.

cfme/tests/ansible/test_embedded_ansible_automate.py
sbulage: Okay.

cfme/tests/ansible/test_embedded_ansible_automate.py
```diff
@@ -419,3 +421,56 @@ def test_variable_pass(request, appliance, setup_ansible_repository, import_data
...
+    request.addfinalizer(domain.delete_if_exists)
+    view = appliance.browser.create_view(FileImportSelectorView)
+    error_msg = "Playbook 'invalid_1677575.yml' not found in repository 'test_playbooks_automate'"
```
sshveta: Is there a way to avoid hardcoding repository name ?

cfme/tests/ansible/test_embedded_ansible_automate.py
```diff
@@ -419,3 +421,56 @@ def test_variable_pass(request, appliance, setup_ansible_repository, import_data
...
+        setup:
+            1. Add playbook repository
+            2. Create a new automate method with playbook type.
+            3. Fill the required fields, for instance repository and playbook.
+            4. Export this datastore.
+            5. Playbook method fields are stored as a names instead of IDs. (this is not
+               possible via automate need to check manually in method yaml)
+        testSteps:
+            1. Import the exported datastore.
+            2. Change name of playbook in method yaml to invalid playbook name
```
sshveta: Where is this happening in the test ?

cfme/tests/ansible/test_embedded_ansible_automate.py
mshriver: Its certainly optional in the docblocks. automates/coverage/blockers meta marks are actually used by the scripting tools.  The docblock is a good place to put BZ's that are related to the case, but aren't necessarily part of automates/blockers.

cfme/tests/ansible/test_embedded_ansible_automate.py
ganeshhubale: @sshveta 
Yeah. I agree over here. But while importing datastore; we are not able to change repository name from datastore yamls. Hence need to have repository, which is there in datastore. As I commented:
```# Repository name(test_playbooks_automate) is static because this is already in the datastore
    # imported. If we used(fauxfactory.gen_alpha()) here then datastore import will fail with no
    # ansible repository available.```

File: cfme/tests/ansible/test_embedded_ansible_automate.py (hunk 5 above)
Comment (ganeshhubale) [reply cfme/tests/ansible/test_embedded_ansible_automate.py]: @sshveta 
We cant do those things with automation. I mean, we can not export datastore. Hence I did those things manually and importig that datastore.
Updated test steps with note:
```
testSteps:
            1. Import the exported datastore and change name of playbook in method yaml to invalid
               playbook name(Note: These test steps needs to execute manually and then import
               datastore)
        expectedResults:
            1. Proper error should be displayed while importing datastore with invalid playbook
```

File: cfme/tests/ansible/test_embedded_ansible_automate.py (hunk 4 above)
Comment (sshveta) [reply cfme/tests/ansible/test_embedded_ansible_automate.py]: @ganeshhubale : Try reading the datastore file in this way 
https://github.com/ManageIQ/integration_tests/blob/master/cfme/fixtures/service_fixtures.py#L253

If that doesn't work please add a comment  that this repository name is coming from the datastore edited manually .

**Comment by dajoRH** [comment]
I detected some fixture changes in commit 03475ec009ef17e488fcc08fa1236bfd9b850dd3

The *local* fixture **`setup_ansible_repository`** is used in the following files:
- **cfme/tests/ansible/test_embedded_ansible_automate.py**
    - *test_variable_pass*
    - *test_import_domain_containing_playbook_method*

**Comment by ganeshhubale** [comment]
@sshveta I added comment for ansible repo name. As reading this value from yml would be so specific to yml file of method. Because in dastastore there are yml files of domain, namespace, class, instance and methods.

File: cfme/tests/ansible/test_embedded_ansible_automate.py [hunk 6]
```diff
@@ -418,3 +420,60 @@ def test_variable_pass(request, appliance, setup_ansible_repository, import_data
...
+    )
+    domain = datastore.import_domain_from(import_data.from_domain, import_data.to_domain)
+    request.addfinalizer(domain.delete_if_exists)
+    view = appliance.browser.create_view(FileImportSelectorView)
+    # setup_ansible_repository.name is ansible repository name already there in datastore yml which
+    # we are going to import.
+    error_msg = (
+        f"Playbook 'invalid_1677575.yml' not found in repository '{setup_ansible_repository.name}'"
+    )
+    assert error_msg in view.flash.read()[0]
```
john-dupuy: Can you use `view.flash.messages`? 

```suggestion
    assert error_msg in view.flash.messages
```

ganeshhubale: Hey @john-dupuy Here we are getting one message.
- By using ```messages``` we are only getting object of flash message in list:
```
ipdb> view.flash.messages                                                                                                                                                    
[<widgetastic_manageiq.V2VFlashMessage object at 0x7ff493d76b90>]

```
- So if we used as you suggested then we need to change it like:
```
ipdb> assert error_msg in view.flash.messages[0].read()                                                                                                                      
```
Hence keeping as it is.

valaparthvi: Does `view.flash.assert_message(message)` work for error messages? Have you tried it?

ganeshhubale: So @valaparthvi 
I am not checking exact whole flash message which is nearly 3-4 line long. I am asserting for error message; I need for this test case. So I can not use **assert_message**. 
Please let me know, if you have any other view.

ganeshhubale: So as discussed offline with @valaparthvi .
Tried with ```partial=True``` and ```assert_message``` which worked as expected.

dajoRH:
I detected some fixture changes in commit c8a141aea87d0b473c606a9641e7ee188f54fe69

The *local* fixture **`setup_ansible_repository`** is used in the following files:
- **cfme/tests/ansible/test_embedded_ansible_automate.py**
    - *test_variable_pass*
    - *test_import_domain_containing_playbook_method*

cfme/tests/ansible/test_embedded_ansible_automate.py
mshriver: @tpapaioa is also currently working on supporting regex compiled expressions for assert_message, which would be useful here. Though I don't want to block your PR on that enhancement - you can change from partial to regex once its released if you'd like - at a later time.

PR #9638, opened 2019-11-18
[RFR] no more merkyl. use LogValidator instead

izapolsk:

PR #9639, opened 2019-11-18
[1LP][RFR] No more session fixtures

izapolsk:
there is a plan to implement appliance markers and generate appliances based on those markers. 
We should get rid of session scoped fixuters using appliance fixture. This PR is one of steps we have to undertake to achieve above.

based on https://github.com/ManageIQ/integration_tests/pull/9638


dajoRH:
Lint report for commit 1783bffc252d1c476495d524dd5d318c655af6a1:

`cfme/fixtures/browser.py`:

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
*CFME QE Bot*

dajoRH:
*CFME QE Bot*

cfme/fixtures/cfme_data.py
```diff
@@ -4,5 +4,5 @@
 
 
 @pytest.fixture(scope="session")
```
jawatts: is this suppose to be session still?

izapolsk: yes. I have to get rid of session scoped fixtures which use appliance. the rest can be left as is.

cfme/fixtures/cfme_data.py
```diff
@@ -4,5 +4,5 @@
...
-def cfme_data(request):
+def cfme_data():
```
mshriver: Wow, I don't think we're actually using this anywhere.

PR #9655, opened 2019-11-22
[1LP][RFR] SSUI page provider check tests

digitronik:
### PRT Run
{{pytest: cfme/tests/services/test_dialog_element_in_catalog.py::test_provider_field_should_display_in_vm_details_page_in_ssui -v}}

### DIFF
--- a/cfme/services/myservice/ssui.py
+++ b/cfme/services/myservice/ssui.py
@@ -22,6 +22,7 @@ from cfme.utils.version import VersionPicker
 from cfme.utils.wait import wait_for
 from widgetastic_manageiq import Notification
 from widgetastic_manageiq import SSUIAppendToBodyDropdown
+from widgetastic_manageiq import SSUICardPFInfoStatus
 from widgetastic_manageiq import SSUIDropdown
 from widgetastic_manageiq import SSUIlist
 from widgetastic_manageiq import SSUIPaginationPane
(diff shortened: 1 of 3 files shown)


cfme/services/myservice/ssui.py
```diff
@@ -175,6 +176,11 @@ class MyServiceVMDetailsView(MyServicesView):
     # TODO: This view needs enhancement by FA owner.
     #  TimelinesChart Widget not supporting completely need improvements as per SSUI.
 
+    # I won't found better way than indexing on this Page.
```
mshriver: Too bad, thanks for at least commenting it.

cfme/tests/services/test_dialog_element_in_catalog.py
```diff
@@ -55,7 +57,7 @@ def service_dialog(appliance, widget_name):
 
 
 @pytest.fixture(scope="function")
-def catalog_item(appliance, service_dialog, catalog):
+def catalog_item_local(appliance, service_dialog, catalog):
```
mshriver: You actually don't need to change the name of the fixture, pytest will resolve the most local one matching the node's scope.

widgetastic_manageiq/__init__.py
```diff
@@ -5946,3 +5946,17 @@ def messages(self):
         except NoSuchElementException:
             pass
         return result
+
+
+class SSUICardPFInfoStatus(Widget):
```
mshriver: Consider subclassing `widgetastic.widget.GenericLocatorWidget` when setting up a widget like this, as it takes care of mapping ROOT to the passed locator.

cfme/tests/services/test_dialog_element_in_catalog.py
digitronik: Actually `pytest` took local `catalog_item` fixture  instate of `catalog_item` global fixture; when we call global `service_vm`.

PR #9667, opened 2019-11-25
[1LP][RFR] service order after prov deletion

digitronik:
### PRT Run
{{pytest: cfme/tests/services/test_service_catalogs.py::test_order_service_after_deleting_provider --long-running -v}}

### DIFF
--- a/cfme/tests/services/test_service_catalogs.py
+++ b/cfme/tests/services/test_service_catalogs.py
@@ -3,6 +3,8 @@ import pytest
 
 from cfme import test_requirements
 from cfme.infrastructure.provider import InfraProvider
+from cfme.infrastructure.provider.rhevm import RHEVMProvider
+from cfme.markers.env_markers.provider import ONE
 from cfme.markers.env_markers.provider import ONE_PER_TYPE
 from cfme.services.service_catalogs import ServiceCatalogs
 from cfme.services.workloads import VmsInstances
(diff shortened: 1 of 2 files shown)


cfme/tests/services/test_service_catalogs.py
```diff
@@ -202,3 +204,35 @@ def test_advanced_search_registry_element(request, appliance):
...
+@pytest.mark.tier(2)
+@pytest.mark.provider([RHEVMProvider], selector=ONE)
+def test_order_service_after_deleting_provider(appliance, provider, setup_provider, catalog_item):
+    """
+    Polarion:
+        assignee: nansari
+        casecomponent: Services
+        testtype: functional
+        initialEstimate: 1/4h
+        startsin: 5.8
```
kedark3: Do we care about 5.8 anymore? Would it make sense to say 5.9 or 5.10 even? Just a thought.

digitronik: I think better to put exact if we know.

cfme/tests/services/test_service_catalogs.py
```diff
@@ -202,3 +204,35 @@ def test_advanced_search_registry_element(request, appliance):
...
+        tags: service
+    """
+    template_name = catalog_item.prov_data["catalog"]["catalog_name"]["name"]
+    template_id = appliance.rest_api.collections.templates.find_by(name=template_name)[0].id
```
john-dupuy: If the template does not exist on the provider can we skip the test?

digitronik: @john-dupuy I'm retrieving this from fixture return l.e. `catalog_item` for finding `id` .
`catalog_item` fixture using `provisioning`; which going to call first.

https://github.com/ManageIQ/integration_tests/blob/1f5035262ae7d4c5f43e6fe1f33276ce5b7b4236/cfme/fixtures/service_fixtures.py#L50

It should taken care by `provisioning`; which checking for `provisioning` key in `yaml`

PR #9674, opened 2019-11-25
[1LP][RFR] Change from_provider to use get_mgmt instead of get_crud

jawatts:


cfme/utils/appliance/__init__.py
```diff
@@ -2562,22 +2562,22 @@ def from_provider(cls, provider_key, vm_name, **kwargs):
...
             browser_steal: Setting of the browser_steal attribute.
         """
-        from cfme.utils.providers import get_crud
-        from cfme.containers.provider.openshift import OpenshiftProvider
+        from wrapanapi.systems.container import Openshift as OpenShiftSystem
+        from cfme.utils.providers import get_mgmt
         from cfme.utils import conf
 
-        provider = get_crud(provider_key)
+        provider = get_mgmt(provider_key)
```
john-dupuy: Can we call it `provider_mgmt` to avoid confusion?

mshriver:
Looks good, I shouldn't have allowed @john-dupuy to change it to an actual provider object. We didn't hit it until hooks were added to appliance stack push/pop. 

Below PR is for wrapping the hook execution and reducing default ssh wait time:
https://github.com/ManageIQ/integration_tests/pull/9675

Also discussing a separate change to where the hook methods are registered, to restrict them to being executed within a pytest context - when we don't have dummy appliance objects, and to reduce the load up time and actions when using appliance objects in local consoles.

cfme/utils/appliance/__init__.py
```diff
@@ -2636,10 +2637,8 @@ def configure(self, setup_fleece=False, log_callback=None, on_openstack=False, *
 
         # Defer to the IPAppliance.
         super(Appliance, self).configure(log_callback=log_callback, on_openstack=on_openstack)
-        # And do configure the fleecing if requested
-        if setup_fleece:
-            self.configure_fleecing(log_callback=log_callback)
 
+    #  TODO Can we remove this?
```
john-dupuy: For vsphere this method is provided by the fixtures https://github.com/ManageIQ/integration_tests/blob/0fc21aaf6c6d633bfe52c1d36e9a045e377520e0/cfme/fixtures/appliance.py#L261-L278

Unless we want to keep it as a convenience method for SSA testing. @sbulage do you make use of this method during 
your SSA testing?

sbulage: @john-dupuy No I didn't write it.

mshriver: Thanks for pointing this out @john-dupuy, @jawatts and I discussed it and decided because of the larger assumptions and actions in configure_fleecing, we would leave it in place for now.

PR #9675, opened 2019-11-25
[RFR] Add exception diaper to appliance hooks, reduce ssh wait time

mshriver:
Wrap exceptions on appliance stack hook execution.

I'm also looking at moving the hook registration into pytest_configure, but need to thoroughly test the stack push order in the existing appliance plugin to make sure the hook is injected before anything is pushed to the stack.

### DIFF
--- a/cfme/utils/appliance/__init__.py
+++ b/cfme/utils/appliance/__init__.py
@@ -1584,7 +1584,7 @@ ExecStartPre=/usr/bin/bash -c "ipcs -s|grep apache|cut -d\  -f2|while read line;
             logger.exception('Exception clearing cached_property "guid"')
         return str(result).rstrip('\n')  # should return UUID from stdout
 
-    def wait_for_ssh(self, timeout=600):
+    def wait_for_ssh(self, timeout=30):
         """Waits for appliance SSH connection to be ready
 
         Args:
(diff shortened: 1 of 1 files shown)


cfme/utils/appliance/__init__.py
```diff
@@ -2884,34 +2884,44 @@ def __init__(self):
         super(ApplianceStack, self).__init__()
 
     def push(self, obj):
-        was_before = self.top
+        stack_parent = self.top
```
mshriver: Other than the name change, I changed to using getattr() instead of ternary (getattr is safe against a None).

I also wrapped the hook method call in a general except block.

cfme/utils/appliance/__init__.py
```diff
@@ -1584,15 +1584,15 @@ def update_guid(self, log_callback=None):
...
+    def wait_for_ssh(self, timeout=30):
         """Waits for appliance SSH connection to be ready
 
         Args:
             timeout: Number of seconds to wait until timeout (default ``600``)
         """
         wait_for(func=lambda: self.is_ssh_running,
                  message='appliance.is_ssh_running',
-                 delay=5,
+                 delay=1,
```
mshriver: There's a 10s wait within the net_check used by is_ssh_running, so by the time this wait_for wants to try again, the time will have lapsed anyway. Just drop this to a minimum.

izapolsk:
@mshriver, I'm gonna turn appliance hooks into pytest hooks. so, please leave this with me.

PR #9676, opened 2019-11-26
[1LP][RFR] fix slave id representation in logs

izapolsk:
slave id appears as bytes in logs everywhere. That's a bit annoying. 
This PR should make it appear as string in logs.

{ pytest: --long-running --use-template-cache --collect-only --dummy-appliance --num-dummies 10 }

### DIFF
--- a/cfme/fixtures/parallelizer/__init__.py
+++ b/cfme/fixtures/parallelizer/__init__.py
@@ -109,7 +109,8 @@ class SlaveDetail(object):
 
     appliance = attr.ib()
     worker_config = attr.ib()
-    id = attr.ib(default=attr.Factory(lambda: next(SlaveDetail.slaveid_generator)))
+    id = attr.ib(default=attr.Factory(lambda: next(SlaveDetail.slaveid_generator)),
+                 repr=lambda value: value.decode('utf-8'))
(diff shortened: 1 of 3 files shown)


cfme/fixtures/parallelizer/__init__.py
```diff
@@ -109,7 +109,8 @@ class SlaveDetail(object):
 
     appliance = attr.ib()
     worker_config = attr.ib()
-    id = attr.ib(default=attr.Factory(lambda: next(SlaveDetail.slaveid_generator)))
+    id = attr.ib(default=attr.Factory(lambda: next(SlaveDetail.slaveid_generator)),
+                 repr=lambda value: value.decode('utf-8'))
```
john-dupuy: @izapolsk can you make this change in the `slaveid_generator`? Something like 

```
slaveid_generator = (f'slave{i:02d}'.encode('utf-8') for i in count())
```

izapolsk: @john-dupuy, I can't I'm under impression it is intentionally made `bytes` for interprocess communication and etc

john-dupuy: Ah gotcha, okay, I was curious

PR #9680, opened 2019-11-27
[1LP][RFR] Fixed test_add_button testcase

dgaikwad:
## Purpose or Intent

- __Fixing__ test_add_button because it is currently checking assert message with button name. It should check with rover name and removed version check which is not required.
Note: You can introduce Screenshots/Gifs for providing more information.

### PRT Run

{{pytest: cfme/tests/services/test_catalog_item.py::test_add_button -v}}

### DIFF
--- a/cfme/tests/services/test_catalog_item.py
+++ b/cfme/tests/services/test_catalog_item.py
@@ -15,7 +15,6 @@ from cfme.utils.log import logger
 from cfme.utils.log_validator import LogValidator
 from cfme.utils.update import update
 
-
 pytestmark = [test_requirements.service, pytest.mark.tier(3), pytest.mark.ignore_stream("upstream")]
 
 
@@ -132,12 +131,14 @@ def test_add_button(catalog_item, appliance):
         casecomponent: Services
         tags: service
     """
-    button_name = catalog_item.add_button()
+    btn_data = {
+        "text": fauxfactory.gen_numeric_string(start="btn_"),
+        "hover": fauxfactory.gen_numeric_string(15, start="btn_hvr_"),
+        "image": "fa-user",
+    }
+    catalog_item.add_button(**btn_data)
     view = appliance.browser.create_view(BaseLoggedInPage)
-    if appliance.version.is_in_series('5.8'):
-        message = 'Button "{}" was added'.format(button_name)
-    else:
-        message = 'Custom Button "{}" was added'.format(button_name)
+    message = f'Custom Button "{btn_data["hover"]}" was added'
     view.flash.assert_success_message(message)
 
 


digitronik:
@dgaikwad please sign your commit with `gpg`

cfme/tests/services/test_catalog_item.py
```diff
@@ -134,10 +134,8 @@ def test_add_button(catalog_item, appliance):
     """
     button_name = catalog_item.add_button()
     view = appliance.browser.create_view(BaseLoggedInPage)
-    if appliance.version.is_in_series('5.8'):
-        message = 'Button "{}" was added'.format(button_name)
-    else:
-        message = 'Custom Button "{}" was added'.format(button_name)
+
+    message = 'Custom Button "{}" was added'.format(button_name)
```
sbulage: **Suggestion:** 
@dgaikwad Instead of `.format`, can you please use `f'` string .?
Like this --> `message = f'Custom Button "{button_name}" was added'`

dgaikwad: Done.

cfme/services/catalogs/catalog_items/__init__.py
```diff
@@ -452,7 +452,7 @@ def add_button(self, **kwargs):
         view.add.click()
         view = self.create_view(DetailsCatalogItemView, wait=5)
         view.flash.assert_no_error()
-        return text
+        return hover
```
mshriver: Thanks so much for tackling a test fix! Things tend to get complex when the parameters and return values for framework code changes.

The test that has been modified along with this change is not the only use of this method within the framework, and the other uses are using this return value to then make other calls - like to `button_exists` below, which takes the text, not the hover.

Check out `cfme/tests/automate/custom_button/test_catalog_custom_button.py::test_catalog_item_copy_with_custom_buttons`, which is using this return, and kind of highlights how there is duplicated coverage for the fixed test.

A small scope fix would be to actually pass text/hover attributes to the add_button() call in the failing test.  That way you can use the hover attribute that you passed to validate the flash message.  You can also consider just removing the test completely, as the behavior is covered by the above test function.

I think the correct fix, from a framework perspective, is that this method should be returning an instance of `cfme.automate.button.DefaultButton`, and that all callers should be updated to handle that return.  That way all of the attributes are available for further use and validation.

cfme/tests/services/test_catalog_item.py
```diff
@@ -134,10 +134,8 @@ def test_add_button(catalog_item, appliance):
     """
     button_name = catalog_item.add_button()
     view = appliance.browser.create_view(BaseLoggedInPage)
-    if appliance.version.is_in_series('5.8'):
-        message = 'Button "{}" was added'.format(button_name)
-    else:
-        message = 'Custom Button "{}" was added'.format(button_name)
+
+    message = f'Custom Button "{button_name}" was added'
```
digitronik: I like @mshriver idea but you need to major changes in model page also taken care about navigation as catalog level and automate buttons are different.

Optional: (hot fix)
As  `add_button` method accepting `**kwargs`
```
btn_data = {
    "text": gen_numeric_string(start="btn_"),
    "hover": gen_numeric_string(15, start="btn_hvr_"),
    "image": "fa-user",
}
catalog_item.add_button(**btn_data)
view = appliance.browser.create_view(BaseLoggedInPage)
message = f'Custom Button "{btn_data["hover"]}" was added'
```

dgaikwad:
@mshriver changes are done based on @digitronik comment so now we do not need to change the return value of add_function() function.

mshriver:
Great fix, thanks @dgaikwad

PR #9683, opened 2019-11-27
[1LP][RFR] fixing error in artifactor shutdown appearing due to changes in selenium

izapolsk:
capabilities don't have platform and version in currently used selenium. so, this code throws exception. 
It seems those fields were renamed into browserVersion and browserPlatform. Thus this PR fixes that issue.


mshriver:
Nice find!!

PR #9684, opened 2019-11-27
[1LP][RFR] Fix dashboard tests.

jaryn:
Recent changes to widgetastic broke the dashboard tests:
```
/var/ci/cfme_venv3/lib64/python3.7/site-packages/widgetastic/widget/base.py:1199: AttributeError
AttributeError
b'This is not an instance of widgets. You need to call this object and pass the required parameters of the view.'
```

There probably actually was a problem that ParametrizedView class was
used where just View should have been used.

When testing this PR, I also found out that the `cfme/tests/intelligence/reports/test_widgets.py::test_widgets_on_dashboard` fails on CFME 5.11 because it misses RSS which was removed in 5.11. I made creating of that conditional and this fixes the test.

{{pytest: cfme/tests/intelligence/test_dashboard.py cfme/tests/intelligence/reports/test_widgets.py -v}}

### DIFF
--- a/cfme/dashboard.py
+++ b/cfme/dashboard.py
@@ -14,6 +14,7 @@ from widgetastic.widget import Widget
 from widgetastic.xpath import quote
 from widgetastic_patternfly import Button
 from widgetastic_patternfly import Dropdown
+from widgetastic_patternfly import Tab
 
 from cfme.common import BaseLoggedInPage
 from cfme.modeling.base import BaseCollection
@@ -24,7 +25,6 @@ from cfme.utils.appliance.implementations.ui import navigator
 from cfme.utils.timeutil import parsetime
 from cfme.utils.wait import wait_for
 from widgetastic_manageiq import Table
-from widgetastic_manageiq import WaitTab
 
 
 # TODO: Move this into widgetastic_patternfly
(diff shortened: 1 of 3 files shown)


cfme/dashboard.py
```diff
@@ -385,7 +385,7 @@ def all(self, content_type=None):  # widgets
         view = self.dashboard_view
         result = []
         # TODO: Idiomatize the following line
-        for (widget_name, ) in view.widgets.view_class.all(view.browser):
+        for (widget_name, ) in view.widgets.all(view.browser):
```
mshriver: Surprised this does work without passing a TITLE parameter to `view.widgets`, since its still decorated as a nested 
parametrized view

dajoRH:
Lint report for commit ecb2a34ff08afd95fda0cdec83657358139ff8a6:

- :red_circle: Line 3660:61: **F821** *undefined name 'widgets'*

**No commit flaws detected**.

*CFME QE Bot*

dajoRH:
I detected some fixture changes in commit 6de6402af4ce16dac7dac06b4bf76d2a135e296a

The *local* fixture **`custom_widgets`** is used in the following files:
- **cfme/tests/intelligence/reports/test_widgets.py**

widgetastic_manageiq/__init__.py
```diff
@@ -3657,22 +3657,20 @@ def remove_dashboard_widget(self, widget):
...
-        return list(set(values) - set(self.all_dashboard_widgets))
+        return values - set(self.all_dashboard_widgets)
 
     def fill(self, values):
         if isinstance(values, str):
             values = [values]
-        if set(values) == set(self.all_dashboard_widgets):
+        values = set(values)
+
+        if values == set(self.all_dashboard_widgets):
```
john-dupuy:
```suggestion
        if set(values) == set(self.all_dashboard_widgets):
```

PR #9693, opened 2019-11-28
[1LP][RFR] New Customer Scenario Test: test_miq_schedule_validation_failed

valaparthvi:
## Purpose or Intent

- __Adding tests__ 
    1. Add `test_miq_schedule_validation_failed` for [customer BZ ](https://bugzilla.redhat.com/show_bug.cgi?id=1740229).

### PRT Run
{{ pytest: cfme/tests/intelligence/reports/test_reports_schedules.py -k "test_miq_schedule_validation_failed" }}

### DIFF
--- a/cfme/tests/intelligence/reports/test_reports_schedules.py
+++ b/cfme/tests/intelligence/reports/test_reports_schedules.py
@@ -1,3 +1,5 @@
+import os
+
 import fauxfactory
 import pytest
 import yaml
@@ -7,6 +9,9 @@ from cfme.intelligence.reports.schedules import NewScheduleView
 from cfme.intelligence.reports.schedules import ScheduleDetailsView
 from cfme.rest.gen_data import users as _users
 from cfme.utils.appliance.implementations.ui import navigate_to
+from cfme.utils.conf import cfme_data
+from cfme.utils.ftp import FTPClientWrapper
+from cfme.utils.log_validator import LogValidator
 from cfme.utils.path import data_path
 from cfme.utils.wait import wait_for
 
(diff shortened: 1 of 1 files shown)


cfme/tests/intelligence/reports/test_reports_schedules.py
```diff
@@ -246,3 +251,50 @@ def test_reports_schedules_user(appliance, request, user, schedule_data):
...
+        pytest.fail("Failed to download the file to the appliance.")
+
+    # In case the failure pattern is encountered, the following piece of code will raise
+    # a TimedOutError from appliance.wait_for_web_ui
+    # The matched_patterns will only be found on successful database restore.
+    with LogValidator(
+        "/var/www/miq/vmdb/log/evm.log",
+        failure_patterns=[
+            ".* ERROR .*Validation failed: MiqSchedule: Name has already been taken  Method.*"
+        ],
```
ganeshhubale: Are you sure? we have extra space here: taken  Method.

valaparthvi: Yes, it is correct.

john-dupuy: Do we need to search for the whole string? Can we just search for the `.* ERROR .*Validation failed: MiqSchedule:`. My concern is that this could change upstream and then we may be missing test failures.

valaparthvi: Um, I'm not sure I completely understand your concern. Have you faced something like this before where a log message changed in upstream?

john-dupuy: I haven't, probably not really something worth fretting over. But that log message looks like a typo upstream and it could be fixed at some point.

valaparthvi: Oh I see, is it because of that extra space? Um, I guess I can make some change there.

valaparthvi: So I added a few `.*` :grimacing:

PR #9696, opened 2019-11-29
[1LP][RFR] automated BZ 1594817

lina-nikiforova:
## Purpose or Intent
Automated https://bugzilla.redhat.com/show_bug.cgi?id=1594817

### PRT Run
{{ pytest: -v cfme/tests/infrastructure/test_providers.py::test_rhv_custom_attributes_after_refresh }}

### DIFF
--- a/cfme/tests/infrastructure/test_providers.py
+++ b/cfme/tests/infrastructure/test_providers.py
@@ -15,6 +15,7 @@ from cfme.infrastructure.provider.rhevm import RHEVMEndpoint
 from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.infrastructure.provider.virtualcenter import VirtualCenterEndpoint
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
+from cfme.markers.env_markers.provider import ONE
 from cfme.markers.env_markers.provider import ONE_PER_VERSION
 from cfme.utils.appliance.implementations.ui import navigate_to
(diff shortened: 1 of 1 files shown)


cfme/tests/infrastructure/test_providers.py
```diff
@@ -375,6 +376,48 @@ def _refresh_provider():
...
+            1.
+            2.
+            3. The custom attribute is still there
+    """
+    # get the name of any VM on the provider
+    view = navigate_to(provider, 'ProviderVms')
+    vm_name = view.entities.all_entity_names[0]
+
+    vm = f"Vm.where(name: '{vm_name}').last"
+
```
ganeshhubale: **Optional:**
It would be better if we remove **vm_name** variable. As it used once. So we can simply go with: ```vm = f"Vm.where(name: '{view.entities.all_entity_names[0]}').last"```

cfme/tests/infrastructure/test_providers.py
```diff
@@ -375,6 +376,48 @@ def _refresh_provider():
...
+    # set the attributes to the VM
+    assert appliance.ssh_client.run_rails_console(f"{vm}.miq_custom_set('mykey', 'myval')").success
+
```
ganeshhubale: **Question:**
Can we use **fauxfactory** values instead of ```mykey``` and ```myval```(Or these are default values)?

PR #9699, opened 2019-11-29
[1LP][RFR] Remove the BZ 1704835 workaround and mark with automates.

jaryn:
The BZ is fixed.

I think tests may need this to pass:
https://github.com/ManageIQ/integration_tests/pull/9646

### DIFF
--- a/cfme/tests/cli/test_appliance_update.py
+++ b/cfme/tests/cli/test_appliance_update.py
@@ -300,6 +300,7 @@ def test_update_replicated_webui(get_replicated_appliances_with_providers, appli
 
 
 @pytest.mark.ignore_stream("upstream")
+@pytest.mark.meta(coverage=[1704835])
 @pytest.mark.parametrize("update_strategy", [update_appliance, do_yum_update], ids=["webui", "yum"])
 def test_update_ha(ha_appliances_with_providers, appliance, update_strategy, request, old_version):
     """ Tests updating an appliance with providers using webui, also confirms that the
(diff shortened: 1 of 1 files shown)


cfme/tests/cli/test_appliance_update.py
```diff
@@ -121,6 +119,7 @@ def do_yum_update(appliance):
     appliance.wait_for_web_ui()
 
 
+@pytest.mark.meta(coverage=[1714236])
```
sshveta: This should be automates instead ? Since this is automated

cfme/tests/cli/test_appliance_update.py
```diff
@@ -300,6 +299,7 @@ def test_update_replicated_webui(get_replicated_appliances_with_providers, appli
 
 
 @pytest.mark.ignore_stream("upstream")
+@pytest.mark.meta(coverage=[1704835])
```
sshveta: Same as above 
@pytest.mark.meta(automates=[1704835])

sshveta: Also you can add bugzilla docblock

jaryn: thanks done

cfme/tests/cli/test_appliance_update.py
jaryn: thanks done

john-dupuy:
@sshveta moving to 1LP as both comments are addressed.

PR #9714, opened 2019-12-03
[1LP][RFR] Automate: test_delete_generated_report

valaparthvi:
## Purpose or Intent

- __Adding tests__ 
    1. `test_delete_generated_report`
- __Enhancements__
    1. Add `RowNotFound` exception to the list of exceptions for `exists` method in `BaseEntity`.
    2. Replace `runs` attribute of `OptimizationReport` to property.

### PRT Run
{{ pytest: cfme/tests/intelligence/test_optimization.py -k "test_delete_generated_report" -vvv }}

### DIFF
--- a/cfme/tests/intelligence/test_optimization.py
+++ b/cfme/tests/intelligence/test_optimization.py
@@ -1,7 +1,10 @@
+from datetime import datetime
+
 import pytest
 from widgetastic.utils import attributize_string
 
 from cfme import test_requirements
+from cfme.utils.appliance.implementations.ui import navigate_to
 from cfme.utils.blockers import BZ
 
 pytestmark = [test_requirements.report, pytest.mark.tier(1), pytest.mark.ignore_stream("5.10")]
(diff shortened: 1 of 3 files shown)


cfme/tests/intelligence/test_optimization.py
```diff
@@ -69,4 +71,51 @@ def test_delete_generated_report():
...
+        )
+
+    view = navigate_to(optimization_report.collections.saved_reports, "All")
+    # This asserts the report exists for SavedReports accordion.
+    # If the report did not exist, RowNotFound error would be raised here.
+    row = next(
+        view.table.rows(
+            run_at__contains=run_at_time, name__contains=opt_saved_report.parent.parent.menu_name
+        )
+    )
```
john-dupuy: I don't see where the assert is happening. Won't this error out with a `RowNotFound` error rather than an `AssertionError`?

valaparthvi: What I meant by the comment was, if the row is found, we can assert that it exists/is_displayed, else it would raise a RowNotFound error. But I guess wrapping the statement in try and except block would be a good idea.

cfme/tests/intelligence/test_optimization.py
```diff
@@ -69,4 +72,56 @@ def test_delete_generated_report():
...
+        )
+
+    view = navigate_to(optimization_report.collections.saved_reports, "All")
+
+    # This asserts the report exists for SavedReports accordion.
+    # If the report does not exist, RowNotFound error would be raised here.
+    try:
+        row = next(
+            view.table.rows(
+                run_at__contains=run_at_time,
```
mshriver: :tada: Nice use of row filtering!

PR #9722, opened 2019-12-04
[1LP][RFR] Automate: test_authorization_header_sql_response

valaparthvi:
## Purpose or Intent

- __Adding tests__ 
    1. `test_authorization_header_sql_response`

### PRT Run
{{ pytest: cfme/tests/test_providerless_rest.py -k "test_authorization_header_sql_response" -vvv }}

### DIFF
--- a/cfme/tests/test_providerless_rest.py
+++ b/cfme/tests/test_providerless_rest.py
@@ -1,6 +1,7 @@
 """This module contains REST API specific tests which do not require a provider setup.
 For tests that do require provider setup, add them to test_rest.py"""
 import pytest
+from manageiq_client.api import APIException
 
 from cfme import test_requirements
 from cfme.utils.wait import wait_for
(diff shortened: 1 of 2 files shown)


cfme/tests/test_rest.py
```diff
@@ -1306,3 +1307,42 @@ def test_release_server_info(appliance):
...
+                {
+                "error": {
+                    "kind": "unauthorized",
+                    "message": ("PG::CharacterNotInRepertoire: ERROR:"
+                                "  invalid byte sequence for encoding \"UTF8\": 0xb5\n:"),
+                    "klass": "Api::AuthenticationError"
+                    }
+                }
+    """
+    appliance.rest_api._session.headers = {"Authorization": "Basic testing"}
```
digitronik: @valaparthvi you are overtiring attributes of `ManageIQClient` instance.
I think this is special case where we need different header than `{'Content-Type': 'application/json; charset=utf-8'}`. I would instantiate separate instance of `MangeIQClient` than this workaround. what you say?

mshriver: Great suggestion @digitronik. Alternatively this needs to be reset on the session with a request.addfinalizer call.

valaparthvi: Good catch @digitronik @mshriver ! Changed the test to use a seperate instance.

PR #9510, opened 2019-10-18
[1LP][WIP] Post upgrade test run support

digitronik:
## Purpose or Intent
- First step towards upgrade testing
- Adding `post-upgrade` support (for now it will support only `Minor` upgrade)
- Tested `5.11.0.25` upgrade to `5.11.z` l.e. latest
- Tested `5.10.10.0` upgrade to `5.10.z` l.e latest

For Reviewer: job `ndhandre-pr-9510`

PR #9728, opened 2019-12-05
[1LP][RFR] upgrade method for appliance

digitronik:
## Purpose or Intent
- As per @mshriver suggestion moving `upgrade` method for appliance in this PR. Actually, It part of https://github.com/ManageIQ/integration_tests/pull/9510;

### DIFF
--- a/cfme/utils/appliance/__init__.py
+++ b/cfme/utils/appliance/__init__.py
@@ -1327,6 +1327,56 @@ ExecStartPre=/usr/bin/bash -c "ipcs -s|grep apache|cut -d\  -f2|while read line;
 
         return result
 
+    def upgrade(self, upgrade_to, cfme_only=True, reboot=False):
+        """ Upgrade an appliance
+        Args:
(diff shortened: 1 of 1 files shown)


cfme/utils/appliance/__init__.py
```diff
@@ -1327,6 +1327,56 @@ def update_rhel(self, *urls, **kwargs):
...
+            result = self.ssh_client.run_command(f"yum -y update{cmd}", timeout=3600)
+        except socket.timeout:
+            logger.error(f"SSH timed out while updating appliance: {result.output}")
+
+        self.db_service.restart()
+        self.evmserverd.start()
+
+        # May be chance to update kernel with all update.
+        if reboot or not cfme_only:
+            self.reboot()
```
jawatts: @digitronik could we move the reboot before the db and evmserver service restarts/start? Like:
```
 # May be chance to update kernel with all update.
 if reboot or not cfme_only:
     self.reboot()
else:
    self.db_service.restart()
    self.evmserverd.start()
```
Those services will be restarted with a reboot anyway. Thoughts?

digitronik: sure :+1:

PR #9744, opened 2019-12-09
[1LP][RFR] Handle a 'None' BZ because of no creds

mshriver:


cfme/utils/bz.py
```diff
@@ -234,6 +234,12 @@ def get_bz_info(self, idlist, clones=True):
         # build info
         info = {}
         for bug_id, bug in zip(idlist, self.bugzilla.getbugs(idlist)):
+            # safety first
+            if bug is None:
```
john-dupuy: Can you do? 

```suggestion
            if not bug:
```

mshriver: Sure

PR #9756, opened 2019-12-11
[1LP][RFR] New Test: Test overwrite domain using rake command

ganeshhubale:

## Purpose or Intent
- Test overwrite domain using rake
### PRT Run
{{ pytest: cfme/tests/automate/test_file_import.py -k 'test_overwrite_import_domain' --use-template-cache -qsvv }}

### DIFF
--- a/cfme/tests/automate/test_file_import.py
+++ b/cfme/tests/automate/test_file_import.py
@@ -1,3 +1,5 @@
+import os
+
 import fauxfactory
 import pytest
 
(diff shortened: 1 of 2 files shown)


dajoRH:
I detected some fixture changes in commit 867848489acf998d4c7b185460bc8ce23a7e7812

The *local* fixture **`local_domain`** is used in the following files:
- **cfme/tests/automate/test_file_import.py**
    - *test_overwrite_import_domain*

dajoRH:
I detected some fixture changes in commit a1df91e271efa9a9633f7bbdcac6d59b3e977be3

The *local* fixture **`local_domain`** is used in the following files:
- **cfme/tests/automate/test_file_import.py**
    - *test_overwrite_import_domain*

cfme/tests/automate/test_file_import.py
```diff
@@ -201,3 +203,76 @@ def test_automate_import_attributes_updated(setup_automate_model, import_datasto
...
+@pytest.fixture(scope='module')
+def local_domain(appliance):
+    """This fixture used to create automate domain - Datastore/Domain"""
+    # Domain name should be static to match with name of domain imported using rake command
+    domain = appliance.collections.domains.create(
+        name="bz_1753860",
+        description=fauxfactory.gen_alpha(),
+        enabled=True)
+    yield domain
+    domain.delete_if_exists()
```
valaparthvi: I think it would be better if you use REST to delete the domain.
```
if domain.rest_api_entity.exists:
    domain.rest_api_entity.action.delete()
```
_The reason being_, if the test case fails/errors out/stops after disabling the domain, it will not be able to delete the domain via UI, since domain name in the tree would change to `domain_name (Disabled)`, and its tree path would also change to `["Datastore", "domain_name (Disabled)"].

To delete a domain, it first checks the path in the bootstrap tree. Since `.enabled` property of `local_domain` does not get updated when you enable or disable it and since the original domain.enabled value is `True`, it tries to search for ["Datastore", local_domain.name] in the bootstrap tree path, but does not find it and hence is not able to delete the domain.

With REST this problem does not arise, so it doesn't matter whether a domain is enabled or disabled, it will delete it if it exists as long as it is not locked.

dajoRH:
I detected some fixture changes in commit 3584e87350550a7e868e51ad2b10f15d5a712128

The *local* fixture **`local_domain`** is used in the following files:
- **cfme/tests/automate/test_file_import.py**
    - *test_overwrite_import_domain*

cfme/tests/automate/test_file_import.py
```diff
@@ -201,3 +203,76 @@ def test_automate_import_attributes_updated(setup_automate_model, import_datasto
...
+    Bugzilla:
+        1753860
+
+    Polarion:
+        assignee: ghubale
+        initialEstimate: 1/8h
+        caseposneg: positive
+        casecomponent: Automate
+        testSteps:
+            1. Create custom domain, namespace, class, instance, method. Do not delete this domain.
```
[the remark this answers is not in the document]
ganeshhubale: These things needs to create manually by exporting datastore which is not possible in automation.

cfme/tests/automate/test_file_import.py
ganeshhubale: Thank you for detailed solution @valaparthvi 
It is very helpful. But it worked with 5.11 and givin some error in 5.10 as below:
```
ipdb> domain.rest_api_entity.action.all                                                              
['refresh_from_source']
```
It does not have delete action in 5.10 via REST

valaparthvi: Oh yes, so maybe in that case, instead of using REST, you could make sure that the domain is enabled before deleting it, or maybe modify the delete() method to delete a domain no matter it's status.

cfme/tests/automate/test_file_import.py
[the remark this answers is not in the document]
valaparthvi: Okay, so the datastore you're importing creates all these, correct?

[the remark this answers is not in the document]
ganeshhubale: yes

cfme/tests/automate/test_file_import.py
valaparthvi: Another solution would be something like this - 
```suggestion
@pytest.fixture(scope='module')
def local_domain(appliance):
    """This fixture used to create automate domain - Datastore/Domain"""
    # Domain name should be static to match with name of domain imported using rake command
    domain = appliance.collections.domains.create(
        name="bz_1753860", description=fauxfactory.gen_alpha(), enabled=True
    )
    yield domain
    
    domain.enabled = domain.rest_api_entity.enabled
    
    domain.delete()
```

The reason being, `treepath` of a domain is decided on the basis of it's `enabled` status. We can change `domain.enabled` to whatever it's current value is, that way the domain will be deleted for sure.

dajoRH:
I detected some fixture changes in commit 11d7bf97cd5304cb14526df4da1f4ccf301f1e12

The *local* fixture **`local_domain`** is used in the following files:
- **cfme/tests/automate/test_file_import.py**
    - *test_overwrite_import_domain*

dajoRH:
I detected some fixture changes in commit f70242964c1d04cea7e37b3a18a0f4b2f0e0dedf

The *local* fixture **`local_domain`** is used in the following files:
- **cfme/tests/automate/test_file_import.py**
    - *test_overwrite_import_domain*

PR #9703, opened 2019-12-02
[1LP][RFR] Automating basic VMRC Console Test

kedark3:
## Purpose or Intent

- __Adding new tests__ VMRC Test automation is added. 

Note: This PR is built on top of work from #9701  and #9595, please review those first so that I can rebase this.

### PRT Run

{{pytest: cfme/tests/infrastructure/test_vmrc_console.py}}

izapolsk: REQUIRES #9764 9764

PR #9764, opened 2019-12-12
[1LP][RFR] dismiss_alerts closes vmrc related alerts

izapolsk:
it should be disabled during vmrc tests. these changes will be tested by @kedark3 's PR #9703 .

### DIFF
--- a/cfme/common/vm.py
+++ b/cfme/common/vm.py
@@ -280,8 +280,14 @@ class BaseVM(
 
         view = navigate_to(self, 'Details')
 
-        # Click console button given by type
-        view.toolbar.access.item_select(console, handle_alert=invokes_alert)
+        # dismiss_any_alerts() call closed subsequent alerts needed for vmrc
+        # below code is needed to fix such issue
+        try:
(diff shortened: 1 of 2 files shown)


cfme/common/vm.py
```diff
@@ -280,8 +280,14 @@ def open_console(self, console='VM Console', invokes_alert=None):
 
         view = navigate_to(self, 'Details')
 
-        # Click console button given by type
-        view.toolbar.access.item_select(console, handle_alert=invokes_alert)
+        # dismiss_any_alerts() call closed subsequent alerts needed for vmrc
+        # below code is needed to fix such issue
+        try:
+            view.browser.IGNORE_SUBSEQUENT_ALERTS = True
```
izapolsk: would it be worthwhile to replace this try/finally with appropriate context manager ? :thinking:

john-dupuy: @izapolsk I think so, it could have uses elsewhere. But I'm okay with that being a separate PR.

PR #9770, opened 2019-12-16
[1LP][RFR] Automate: test_service_refresh_dialog_fields_default_values

valaparthvi:
## Purpose or Intent

- __Adding tests__ 
    1. Automate: `test_service_refresh_dialog_fields_default_values` and remove manual test.
- __Enhancement__
    1. Add `rest_api_entity` property to `ServiceDialogs`.

### PRT Run
{{ pytest: cfme/tests/services/test_rest_services.py -k "test_service_refresh_dialog_fields_default_values" --long-running -vvvv }}

### DIFF
--- a/cfme/tests/services/test_rest_services.py
+++ b/cfme/tests/services/test_rest_services.py
@@ -2434,3 +2434,68 @@ def test_embedded_ansible_cat_item_edit_rest(appliance, request, ansible_catalog
     service_template.reload()
     assert service_template.name == new_data["name"]
     assert service_template.description == new_data["description"]
+
+
+@pytest.mark.meta(automates=[1730813])
+@test_requirements.rest
+@pytest.mark.customer_scenario
+@pytest.mark.parametrize("file_name", ["testgear-dialog.yaml"], ids=[''])
+def test_service_refresh_dialog_fields_default_values(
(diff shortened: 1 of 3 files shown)


cfme/tests/services/test_rest_services.py
```diff
@@ -2434,3 +2434,68 @@ def test_embedded_ansible_cat_item_edit_rest(appliance, request, ansible_catalog
     service_template.reload()
     assert service_template.name == new_data["name"]
     assert service_template.description == new_data["description"]
+
+
+@pytest.mark.meta(automates=[1730813])
+@test_requirements.rest
+@pytest.mark.customer_scenario
+@pytest.mark.parametrize("file_name", ["testgear-dialog.yaml"], ids=[''])
+def test_service_refresh_dialog_fields_default_values(
```
ganeshhubale: **Optional**:
Don't you think we need add something in ```ids```?

valaparthvi: The test name will be like `test_service_refresh_dialog_fields_default_values[infra]`, so I'm okay with not giving it any ID here. Besides, I couldn't think of any good ID name so I kept it empty.

PR #9772, opened 2019-12-16
[1LP][RFR] Update password gem name

### LINKED ISSUE #8595: MiqPassword is going away
@mshriver I am in the process of finalizing the removal of MiqPassword in favor of ManageIQ::Password (a separate gem).  Technically the alias will stick around for some time so we don't break anything, but the goal is for it to go away.  I see a lot of references in here, and I was going to make a PR, but I know that the code here deals with multiple versions simultaneously and I don't know how to deal with that.  Just wanted to give you a heads up.

mshriver:
Thanks for the heads up @Fryguy

### DIFF
--- a/scripts/harden_security.py
+++ b/scripts/harden_security.py
@@ -77,6 +77,9 @@ def main():
     def update_db_yaml(address):
         with SSHClient(hostname=address, **ssh_creds) as client:
             client.run_command('cd /var/www/miq/vmdb')
+            # TODO Issue 8595, MiqPassword alias/gem will go away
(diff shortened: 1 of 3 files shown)


scripts/harden_security.py
```diff
@@ -77,6 +77,9 @@ def generate_key(address):
     def update_db_yaml(address):
         with SSHClient(hostname=address, **ssh_creds) as client:
             client.run_command('cd /var/www/miq/vmdb')
+            # TODO Issue 8595, MiqPassword alias/gem will go away
```
digitronik: :+1: thanks for this.

PR #9774, opened 2019-12-16
[1LP][RFR] New Test: test_crud_service_template_with_picture

valaparthvi:
## Purpose or Intent

- __Adding tests__ 
    1. Add new customer scenario test: `test_crud_service_template_with_picture`

- __Enhancement__ 
    1. Add `rest_api_entity` property to 
        1. `Dialog` entity
        2. `Catalog` entity

### PRT Run

{{ pytest: cfme/tests/services/test_rest_services.py -k "test_crud_service_template_with_picture" --long-running -vvvv }}

### DIFF
--- a/cfme/services/catalogs/catalog.py
+++ b/cfme/services/catalogs/catalog.py
@@ -9,6 +9,7 @@ from widgetastic_patternfly import Button
 from widgetastic_patternfly import Input
 
 from cfme.common import Taggable
+from cfme.exceptions import RestLookupError
 from cfme.modeling.base import BaseCollection
 from cfme.modeling.base import BaseEntity
 from cfme.services.catalogs import ServicesCatalogView
@@ -126,6 +127,13 @@ class Catalog(BaseEntity, Updateable, Pretty, Taggable):
         view.configuration.item_select('Remove Catalog', handle_alert=True)
         view = self.create_view(CatalogsView, wait='10s')
         view.flash.assert_no_error()
+
+    @property
+    def rest_api_entity(self):
+        try:
+            return self.appliance.rest_api.collections.service_catalogs.get(name=self.name)
+        except ValueError:
+            raise RestLookupError(f'No service catalog rest entity found matching name {self.name}')
 
 
 @attr.s
--- a/cfme/tests/services/test_rest_services.py
+++ b/cfme/tests/services/test_rest_services.py
@@ -2499,3 +2499,94 @@ def test_service_refresh_dialog_fields_default_values(
         expected_tag_0_function == response["tag_0_function"]["values"],
         "Default values for 'tag_0_function' did not match.",
     )
+
+
+@pytest.mark.tier(1)
+@pytest.mark.meta(automates=[1702479])
+@pytest.mark.customer_scenario
+@test_requirements.rest
+def test_crud_service_template_with_picture(appliance, request, service_catalog_obj, dialog):
+    """
+    Bugzilla:
+        1702479
+        1683723
+
+    Polarion:
+        assignee: pvala
+        casecomponent: Rest
+        initialEstimate: 1/4h
+        setup:
+            1. Create a service catalog and service dialog.
+        testSteps:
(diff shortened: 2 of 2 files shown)


cfme/services/catalogs/catalog.py
```diff
@@ -127,6 +128,13 @@ def delete(self):
         view = self.create_view(CatalogsView, wait='10s')
         view.flash.assert_no_error()
 
+    @property
+    def rest_api_entity(self):
+        try:
+            return self.appliance.rest_api.collections.service_catalogs.get(name=self.name)
+        except ValueError:
+            raise RestLookupError(f'No service catalog rest entity found matching ID {self.name}')
+
```
ganeshhubale: It should be ```matching name {self.name}```. Right?

cfme/tests/services/test_rest_services.py
```diff
@@ -2434,3 +2434,99 @@ def test_embedded_ansible_cat_item_edit_rest(appliance, request, ansible_catalog
...
+        casecomponent: Rest
+        initialEstimate: 1/4h
+        setup:
+            1. Create a service catalog and service dialog.
+        testSteps:
+            1. Create a service template via REST with picture.
+            2. Note the picture md5 by querying `picture` attribute of the service template.
+            3. Edit the service template via REST for a different picture.
+            4. Note the picture md5 by querying `picture` attribute of the service template.
+            3. Compare both the md5 from testStep 2 and 4.
```
ganeshhubale: It should be ```5. Compare both the md5 from testStep 2 and 4.``` Right?

cfme/tests/services/test_rest_services.py
```diff
@@ -2434,3 +2434,99 @@ def test_embedded_ansible_cat_item_edit_rest(appliance, request, ansible_catalog
...
+        expectedResults:
+            1. Service template created successfully without any error.
+            2.
+            3. Service template edited successfully without any error.
+            4.
+            5. Both the md5 values must be different.
+    """
+    data = {
+        "name": fauxfactory.gen_alpha(start="Name "),
+        "prov_type": "generic",
```
ganeshhubale: Can we write like this: ```fauxfactory.gen_alpha(start="Name_")``` or we need space?

cfme/tests/services/test_rest_services.py
```diff
@@ -2499,3 +2499,99 @@ def test_service_refresh_dialog_fields_default_values(
...
+    service_template.action.edit(**updated_data)
+    assert_response(appliance)
+
+    picture_2_md5 = appliance.rest_api.get(f"{service_template._href}?attributes=picture")[
+        "picture"
+    ]["md5"]
+
+    assert picture_1_md5 != picture_2_md5
+
+    service_template.action.delete()
```
digitronik: Q: why we need this deletion? I think it should be taken care by finalizer.

cfme/tests/services/test_rest_services.py
```diff
@@ -2499,3 +2499,99 @@ def test_service_refresh_dialog_fields_default_values(
...
+                "fqname": (
+                    "/Service/Provisioning/StateMachines/ServiceProvision_Template/"
+                    "CatalogItemInitialization"
+                ),
+                "dialog_id": dialog.rest_api_entity.id,
+            },
+        },
+        "service_type": "atomic",
+        "display": True,
+        "description": fauxfactory.gen_alpha(start="Description "),
```
digitronik: Required: `fauxfactory` default character limit is `10`

cfme/tests/services/test_rest_services.py
valaparthvi: I could let finalizer do the deletion, but then I thought since I was already creating and editing the template, I might as well delete it to justify the **crud** in the test name :grimacing:

digitronik: I think best practice is limit test as per test behavior. If deletion will fails means teardown of test will fail not actual steps. I think better to have separate test for crud operation test.

PR #9778, opened 2019-12-17
[1LP][RFR] Clear cached appliance DB properties on appliance upgrade

### LINKED ISSUE #2702: Appliance DB schema upgrade might cause chaos
The tables and columns are cached inside the Db() object. If we update an appliance and the DB schema changes, it will need `del appliance.db` in order to invalidate that cache.

mshriver:
FIXES #2702


john-dupuy:
Is there any test in the repo that can validate that this?

PR #9779, opened 2019-12-17
[1LP][RFR] Change test_child_tenant to use non super-admin role

### LINKED ISSUE #7385: [Bug] test_child_tenant_quota_enforce_via_lifecycle_infra might be causing problem with subsequent provision requests
This issue is quite complicated. I've already shared it with @anurag03. He investigated for a while but could not come up with a quick solution, so we decided together to track here for better transparency. I explained everything quite thoroughly to Anurag in my original email, therefore I'll just copy it here.

----------------------------------------------------------------------------------------------------------------------
Hello Anurag, I have a problem and I hope you could help me with solving it. I am reaching out to you since (according to git blame) you're an author of module cfme/tests/infrastructure/test_child_tenant.py. The following is a bit of a text wall, so brace youself.

Okay, so what went wrong? I probably began with adding test_child_tenant_quota_enforce_via_lifecycle_infra to integration_tests. Since you also added @rhv2 decorator, this test became part of regression suite that we use for RHV-CFME integration testing.
image
After that, I started noticing that some of our provisioning test cases fail with the following error:

TimedOutError: Could not do Request finished at /var/lib/jenkins/workspace/cfme-5.9-rhv-4.2-integration-test-dev/integration_tests/cfme/services/requests.py:54 in time

I noticed that those test cases were executed after some quota test cases, so I decided to execute Jenkins run with only three modules:
cfme/tests/infrastructure/test_child_tenant.py <- this one I suspected for potential cause of the problem
cfme/tests/infrastructure/test_project_quota.py <- and this one as well
cfme/tests/infrastructure/test_provisioning_dialog.py <- this is the module where provisioning requests timed out

You can check out the whole run [here](https://rhv-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/cfme-5.9-rhv-4.2-integration-test-dev/60/) yourself. Long story short, this is the outcome:
image

All those test cases without PASSED are those that FAILED. That means that all test cases from cfme/tests/infrastructure/test_child_tenant.py were successful. All the other test cases failed because provisioning requests got stuck at "Pending Approval" stage. This is what CFME appliance looked like after it:
image

I believe that introduction of cfme/tests/infrastructure/test_child_tenant.py has something to do with it. My first idea was that maybe that the user that was created during your test is still logged-in and tries to provision new VMs with his limited quotas. However, as you can see, all of the requests are made by Administrator, so that is not the case. To be completely honest, I have no idea what went wrong during that test. The only thing I know (or at least the evidence seems to suggest) that before executing cfme/tests/infrastructure/test_child_tenant.py, Administrator can make provision requests that get auto-approved. After that, it's no longer true.

Could you please help me and look into that? Any help would be highly appreciated. Thank you!

-----------------------------------------------------------------------------------------------------------------------------

It's very well possible that my assumptions are wrong. However, I could not find any better explanation for this so far.

mshriver:
FIXES #7385 

Per developer comments, the test should not be using super_admin for quota testing.

PRT failures aren't really related to the original issue that was blocking the test.  I'd like to work with @ganeshhubale (test assignee) to resolve these failures in a followup.

{{ pytest: cfme/tests/infrastructure/test_child_tenant.py::test_child_tenant_quota_enforce_via_lifecycle_infra --long-running }}

### DIFF
--- a/cfme/tests/infrastructure/test_child_tenant.py
+++ b/cfme/tests/infrastructure/test_child_tenant.py
@@ -8,13 +8,12 @@ from cfme.infrastructure.provider.rhevm import RHEVMProvider
 from cfme.infrastructure.provider.virtualcenter import VMwareProvider
 from cfme.markers.env_markers.provider import ONE_PER_TYPE
 from cfme.provisioning import do_vm_provisioning
-from cfme.utils.blockers import GH
 from cfme.utils.generators import random_vm_name
 
 pytestmark = [
     test_requirements.quota,
     pytest.mark.meta(server_roles="+automate"),
-    pytest.mark.usefixtures('setup_provider'),
+    pytest.mark.usefixtures('setup_provider_modscope'),
     pytest.mark.long_running,
     pytest.mark.provider([VMwareProvider, RHEVMProvider], scope="module",
                          required_fields=[["provisioning", "template"]], selector=ONE_PER_TYPE)
@@ -79,7 +78,7 @@ def new_child(appliance, new_tenant):
 def new_group(appliance, new_child, new_tenant):
     collection = appliance.collections.groups
     group = collection.create(description=fauxfactory.gen_alphanumeric(start="group_"),
-                              role='EvmRole-super_administrator',
+                              role='EvmRole-administrator',
                               tenant='My Company/{}/{}'.format(new_tenant.name, new_child.name))
     yield group
     if group.exists:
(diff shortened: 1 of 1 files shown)


dajoRH:
I detected some fixture changes in commit 190b9e878697627ebe3ef14bd09b1d01bbddc06d

The *local* fixture **`new_group`** is used in the following files:
- **cfme/tests/infrastructure/test_child_tenant.py**
    - *new_user*

